PR #525, opened 2021-06-12
RFC: pp motives and misc delab fixes

dselsam:
This PR adds support for printing motives both in and out of matches. It also fixes two other minor issues in the delaborator that were needed to get `Std.Range.forIn.loop` to roundtrip. 

- Feature 1: `pp.pi_motives`. Print motives that return `pi` types, `true` by default.
- Feature 2: `pp.dep_motives`. Print motives whose return type depends on one of the arguments, `false` by default.
- Fix 1: delaborating lambdas was skipping instance arguments that the elaborator is expecting to see. Note: this fix solves a recurring problem I have been having, but there may still be other inconsistencies between the elaborator and delaborator regarding implicit lambas.
- Fix 2: one missing `:term` annotation inside a `doElem` was preventing the delaborated syntax tree from roundtripping. Note: this omission may have been intentional, since it could still roundtrip when indirecting through the parser, and adding the annotation introduced one additional set of parentheses in one of the tests.

Happy to split this up into 4 mini-commits and add additional tests/documentation.


src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -106,7 +106,16 @@ register_builtin_option pp.proofs.withType : Bool := {
   group    := "pp"
   descr    := "(pretty printer) when eliding a proof (see `pp.proofs`), show its type instead"
 }
-
+register_builtin_option pp.pi_motives : Bool := {
```
Kha, line 109: We switched to camelCase option names at some point, though we haven't updated all old names yet

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
```diff
@@ -128,6 +128,26 @@ def delabAppExplicit : Delab := do
...
+  | skip
+  | name (n : Name)
+
+def returnsPi (motive : Expr) : Bool := do
+  match motive with
+  | Expr.lam name d b _ => returnsPi b
+  | Expr.forallE .. => true
+  | _ => false
+
+def isDepFun (motive : Expr) : Bool := do
```
Kha, line 142: The name is a bit misleading, it's more `isNonConstFun`...? Perhaps the use site could come with some examples. Also feel free to use `where` for helper functions that are probably not reusable.

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
```diff
@@ -137,15 +157,24 @@ def delabAppImplicit : Delab := whenNotPPOption getPPExplicit do
...
+      let opts ← getOptions
+      let argKind : ImplicitArgKind := match paramKinds with
+        | [ParamKind.implicit _ (some v)] => 
+          if !v.hasLooseBVars && v == arg then ImplicitArgKind.skip else ImplicitArgKind.provide
+        | ParamKind.implicit name none :: _  => do
+          if name == `motive && showMotive arg opts
+          then ImplicitArgKind.name name                        
+          else ImplicitArgKind.skip
+        | _ => ImplicitArgKind.provide
+      match argKind with
```
Kha, line 170: I'd suggest to fuse this `match` with the prior `let`, getting rid of `ImplicitArgKind`, which doesn't seem to be used anywhere else. You can also factor out the common `(fnStx, paramKinds.tailD [],` part, then the `skip` and `provide` cases are trivial enough that duplicating them should not be an issue.

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
```diff
@@ -277,7 +309,13 @@ def delabAppMatch : Delab := whenPPOption getPPNotation do
   | _,        #[] => failure
   | _,        _   =>
     let pats ← delabPatterns st
-    let stx ← `(match $[$st.discrs:term],* with $[| $pats,* => $st.rhss]*)
+    let stx ← do
+      let (piStx, lamMotive) := st.motive.get!
+      let opts ← getOptions
+      if showMotive lamMotive opts then
+        `(match $[$st.discrs:term],* : $(piStx) with $[| $pats,* => $st.rhss]*)
```
Kha, line 316:
```suggestion
        `(match $[$st.discrs:term],* : $piStx with $[| $pats,* => $st.rhss]*)
```
(unless I'm missing something)

tests/lean/run/pp_roundtrip_defs.lean
```diff
@@ -0,0 +1,26 @@
+import Lean
```
Kha, line 1: Please name test files in camelCase

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
```diff
@@ -606,7 +647,10 @@ partial def delabDoElems : DelabM (List Syntax) := do
       | Expr.lam _ _ body _ =>
         withBindingBodyUnusedName fun n => do
           if body.hasLooseBVars then
-            prependAndRec `(doElem|let $n:term ← $ma)
+            -- Note: the `$ma:term` annotation is required for elaboration to roundtrip
+            -- without indirecting through the parser. However, this annotation seems
+            -- to cause additional parentheses to be inserted in some cases.
```
Kha, line 652: I only see *dropped* parentheses in the changed test output?

Kha (approved):

src/Lean/PrettyPrinter/Delaborator/Basic.lean
dselsam: Fixed. Using these three now:
```lean
set_option pp.motives.pi true
set_option pp.motives.nonConst true
set_option pp.motives.all true
```

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
dselsam: Yes, fixed.

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
dselsam: I removed `ImplicitArgKind` at your suggestion. Now the `let` binds an `Option Syntax`, we match on this to compute the new `argStxs` and then all paths return together.

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
dselsam: Fixed.

tests/lean/run/pp_roundtrip_defs.lean
dselsam: Fixed.

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
dselsam: Yes, good. I removed this comment.

Issue #513, opened 2021-06-06
User attributes


leodemoura:
We should allow users to declare their own attributes. This feature is supported in Lean 3 and is used in a few mathlib metaprograms.

Kha:
Just to explain the current state: attributes can be registered only in `builtin_initialize`, i.e. initializers compiled into native code. However, this is *not* a good idea to do in plugins (i.e. anywhere outside the stdlib) because environment extensions (what attributes boil down to) are identified by an index assumed to be globally unique. Thus .olean files will not be compatible between packages using different sequences of attribute-declaring plugins, and in particular you will not be able to depend on two independent packages declaring (and using) attributes. If I'm not mistaken.

tydeu:
Looking through the environment code, I noticed that the globally unique index @Kha mentioned is an incrementing `Nat` that is simply the extension's index in the array of extensions. Could one way of solving at least the order problem @Kha presented be to use a `Name` to identify each extension? This would at least allow users to safely use plugins to register new attributes (until proper non-builtin attributes arrive).

PR #527, opened 2021-06-13
fix: allow plugins to run IO initializers

Vtec234:
We currently set `IO.initalizing := false` before loading plugins, so they don't get to do things like registering env extensions. This PR allows them to do so. Other problems (such as [indexing](https://github.com/leanprover/lean4/issues/513#issuecomment-855405596)) remain.

### DIFF
--- a/src/initialize/init.cpp
+++ b/src/initialize/init.cpp
@@ -20,6 +20,9 @@ namespace lean {
 extern "C" object* initialize_Init(object* w);
 extern "C" object* initialize_Lean(object* w);
 
+/* Initializes the Lean runtime. Before executing any code which uses the Lean package,
+you must first call this function, and then `lean::io_mark_end_initialization`. Inbetween
+these two calls, you may also have to run additional initializers for your own modules. */
 extern "C" void lean_initialize() {
     save_stack_info();
     initialize_util_module();
@@ -47,6 +50,9 @@ void finalize() {
 
 initializer::initializer() {
     lean_initialize();
+    /* Remark: We used to call `lean::io_mark_end_initialization` here, however this prevented
+    plugins from setting up global state such as environment extensions in their initializers.
+    See also `lean_initialize`. */
 }
 
 initializer::~initializer() {
--- a/src/shell/lean.cpp
+++ b/src/shell/lean.cpp
@@ -423,6 +423,7 @@ int main(int argc, char ** argv) {
 #endif
 
     try {
+        // Remark: This currently runs under `IO.initializing = true`.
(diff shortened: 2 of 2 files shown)


src/shell/lean.cpp
```diff
@@ -318,16 +318,6 @@ void load_plugin(std::string path) {
     // NOTE: we never unload plugins
 }
 
-class initializer {
```
Vtec234, line 321: Seems like dead code.

src/shell/lean.cpp
```diff
@@ -546,6 +536,8 @@ int main(int argc, char ** argv) {
         }
     }
 
+    lean::io_mark_end_initialization();
```
Vtec234, line 539: Now `init_search_path` (L 436) runs under `IO,initializing = false`. This didn't seem to cause trouble for me.

Kha:
Just curious, do you have a use case for it other than environment extensions, or do you want to enable them despite the issues?

Vtec234:
Yes, I'm experimenting with installing custom LSP request handlers in plugins. This should be guarded somehow to prevent concurrency trouble such as the main thread traversing installed handlers while an elab worker installs a new one. `IO.initializing` seemed like the natural guard to use.

More generally, it seemed like a bug that `initialize_Lean` can register env extensions while `initialize_MyPlugin` cannot. But if preventing env extension registration outside stdlib is actually the intended semantics of `IO.initializing` (and user attributes would be supported with a single env extension rather than multiple per-tag ones), I could probably just use another guard in the server.

Kha:
I see, makes sense. I saw us using the flag in that regard in other places, though for your use case I don't think there's strictly any chance for trouble - IO.Ref.get/set/modify are atomic after all. Anyway, LGTM, but I think I'll leave merging to @leodemoura.

src/initialize/init.cpp
```diff
@@ -50,8 +46,7 @@ void finalize() {
 }
 
 initializer::initializer() {
-    initialize();
-    lean::io_mark_end_initialization();
+    lean_initialize();
```
leodemoura, line 49: Could you please add a comment here saying that we used to have `lean::io_mark_end_initialization` here and why we moved it to another place?
It is just a matter of copying the explanation above as a comment here.
We should also add a reminder that people using Lean as a library in their code are now responsible for invoking `lean::io_mark_end_initialization` themselves.

src/shell/lean.cpp
leodemoura: Please add this remark as a comment it the code.

Issue #466, opened 2021-05-15
Partition Lean binaries into reasonably-sized shared libraries


Kha:
We currently link `lean` and `leanpkg` statically (as much as possible), which makes for very simple distribution. Interestingly, we can still `dlsym` symbols from these executables for use in the interpreter, but this does come with some caveats on specific platforms. Additionally, we are regularly hitting the maximum number of symbols exportable from a Windows DLL/PE (2^16). If we want to keep all (relevant) symbols reachable from the interpreter, we should partition the Lean & C++ code into reasonably-sized shared libraries, which will also reduce code duplication between `lean` and `leanpkg` (and potential future executables).

leodemoura:
Let's do this one after we refactor the compiler and delete some of the old C++ code.

PR #531, opened 2021-06-15
feat: update LLVM integration and turn it on by default

Kha:


Kha:
Looks mostly good so far, what's left to be done here is to bundle LLVM with the release builds. My plan for that is to copy the corresponding `libLLVM*` from https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.0 to our `lib/` (or `bin/` on Windows) and link `lean` dynamically against it with a corresponding rpath. Sounds good? /cc @joehendrix @tobiasgrosser

Kha:
...except there isn't actually a libLLVM in e.g. https://github.com/llvm/llvm-project/releases/download/llvmorg-12.0.0/clang+llvm-12.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz, only static archives. So is statically linking `lean` against LLVM the way to go then? We can do that, but only after https://github.com/leanprover/lean4/issues/466. And I assume we would still need to copy some other dependencies like `libunwind.so`...

tobiasgrosser:
> Looks mostly good so far, what's left to be done here is to bundle LLVM with the release builds. My plan for that is to copy the corresponding `libLLVM*` from https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.0 to our `lib/` (or `bin/` on Windows) and link `lean` dynamically against it with a corresponding rpath. Sounds good? /cc @joehendrix @tobiasgrosser

That sounds good to me. This might cost you ~5% in compile-time performance (a rough guess) but worrying about this is likely premature optimization. As long as everything works as expected, this is a good first step. We can always change this if further features are required.

Kha:
Ok, I found the flag that shows the external dependencies:
```bash
$ ./clang+llvm-12.0.0-x86_64-linux-gnu-ubuntu-16.04/bin/llvm-config --system-libs
-lrt -ldl -lpthread -lm -lz -ltinfo -lxml2
```
Seems like a bit much! Especially the last two don't necessarily seem to be available on all distributions, so people do custom builds of LLVM after all to exclude them: https://github.com/mun-lang/mun/issues/191. So, should we do that as well? :/

No idea what the situation on Windows is like btw.

Kha:
Alright, this is actually starting to look reasonable. For CI release builds, we now link statically against LLVM from Nixpkgs, which turns out to obviate the need for some LLVM dependencies like libxml2 (though they may come back when we depend on more parts of LLMV). The remaining dependencies, libz and ncurses, are also drawn from Nixpkgs and linked against statically. ncurses is compiled from source for this because only its dynamic build is cached, which takes ~6min but could be cached in our own binary cache.

The end result is that we get `lean` binaries with no more dependencies than before:
```
$ ldd build/stage1/bin/lean
	linux-vdso.so.1 (0x00007f27d49b5000)
	libdl.so.2 => /nix/store/h0p0h3rh1q4i2yavzm3yqi716s9yaj2f-glibc-2.27/lib/libdl.so.2 (0x00007f27d49aa000)
	libm.so.6 => /nix/store/h0p0h3rh1q4i2yavzm3yqi716s9yaj2f-glibc-2.27/lib/libm.so.6 (0x00007f27d4814000)
	libpthread.so.0 => /nix/store/h0p0h3rh1q4i2yavzm3yqi716s9yaj2f-glibc-2.27/lib/libpthread.so.0 (0x00007f27d47f3000)
	libc.so.6 => /nix/store/h0p0h3rh1q4i2yavzm3yqi716s9yaj2f-glibc-2.27/lib/libc.so.6 (0x00007f27d463d000)
	/lib64/ld-linux-x86-64.so.2 => /nix/store/h0p0h3rh1q4i2yavzm3yqi716s9yaj2f-glibc-2.27/lib64/ld-linux-x86-64.so.2 (0x00007f27d49b7000)
```
(the Nix paths are not actually embedded)
```
$ otool -L build/stage1/bin/lean
build/stage1/bin/lean:
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.60.2)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 1.0.0)
```
The increase in executable size seems to be modest for now (~8MB in compressed artifact size).

As for Windows, the link fails as expected because of number of new symbols, but otherwise there are no linker errors.

For local development (when `-DSTATIC=ON` is not set), we can simply link dynamically against a system installation of LLVM (or the one from `shell.nix`), which simplifies setup and should speed up linking. I've pushed the necessary fixes from this branch to do so when opting in with `-DLLVM=ON`, leaving this PR with the changed default and the CI setup open for when we actually start putting LLVM to use.

tydeu:
@Kha 

> No idea what the situation on Windows is like btw.

On Windows, I believe the standard procedure is to build LLVM from the source with `-DLLVM_BUILD_LLVM_DYLIB=ON` and `-DLLVM_LINK_LLVM_DYLIB=ON` in order to build the shared libraries and then dynamic link to them. See, for example, the LLVM build instructions for Haskell's [`llvm-hs`](https://github.com/llvm-hs/llvm-hs#installing-llvm).

DanielFabian:
Do we have a windows mitigation for the number of symbols? Because I'm starting to take a dependency on this branch and would love to know that it will be able to make it into master.

Kha:
Shelved for now

PR #532, opened 2021-06-15
Custom server request registration

Vtec234:
This allows registering custom LSP request handlers in the server via `builtin_initialize` or `initialize` in plugins. We maintain a global `IO.Ref` table of handlers.

For now registering request handlers as-you-go in standard Lean files (via the environment) is unsupported as the semantics of that are quite unclear. In particular, the watchdog needs to know which worker to route requests to, but the `Environment` object is only accessible within one given worker. We could require all custom requests to contain a `uri : DocumentUri` field. In any case I'm leaving this feature for later if it's desired.

Other possible features which might be useful in the future:
- custom notification handlers which run after the standard ones so that plugins can e.g. directly observe changes to the document state
- custom no-response handlers for requests with responding handlers already registered, for the same reason (observing the system state)

### DIFF
--- a/src/Lean/Server/Requests.lean
+++ b/src/Lean/Server/Requests.lean
@@ -55,52 +55,52 @@ abbrev RequestTask α := Task (Except RequestError α)
 abbrev RequestM := ReaderT RequestContext <| ExceptT RequestError IO
 
 namespace RequestM
-  open FileWorker
-  open Snapshots
+open FileWorker
+open Snapshots
 
-  def readDoc : RequestM EditableDocument := fun rc =>
-    rc.docRef.get
+def readDoc : RequestM EditableDocument := fun rc =>
+  rc.docRef.get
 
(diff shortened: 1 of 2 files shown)


src/Lean/Server/Requests.lean
```diff
@@ -0,0 +1,167 @@
...
+structure RequestContext where
+  srcSearchPath : SearchPath
+  docRef        : IO.Ref FileWorker.EditableDocument
+
+abbrev RequestTask α := Task (Except RequestError α)
+/-- Workers execute request handlers in this monad. -/
+abbrev RequestM := ReaderT RequestContext <| ExceptT RequestError IO
+
+namespace RequestM
+  open FileWorker
```
Kha, line 58: ditto

Kha (approved):

Kha:
Good work!

PR #539, opened 2021-06-21
fix: 32-bit Unicode name mangling

Vtec234:
Reported on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Incorrect.20Name.20Mangling).


src/Lean/Compiler/NameMangling.lean
```diff
@@ -4,35 +4,45 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+private def mangleAux : Nat → String.Iterator → String → String
   | 0,   it, r => r
   | i+1, it, r =>
     let c := it.curr
     if c.isAlpha || c.isDigit then
       mangleAux i it.next (r.push c)
     else if c = '_' then
       mangleAux i it.next (r ++ "__")
-    else if c.toNat < 255 then
+    else if c.toNat < 0x100 then
```
Vtec234, line 18: NB the bound increase, now `_xff` can be encoded as well.

src/Lean/Compiler/NameMangling.lean
```diff
@@ -4,35 +4,45 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+      let n := n % 0x1000
+      let r := r.push $ Nat.digitChar (n / 0x100)
+      let n := n % 0x100
+      let r := r.push $ Nat.digitChar (n / 0x10)
+      let r := r.push $ Nat.digitChar (n % 0x10)
       mangleAux i it.next r
+    else
+      let n := c.toNat
+      let r := r ++ "_u32"
+      let r := Nat.toDigits 16 n |>.foldl (fun r c => r.push c) r
```
Vtec234, line 37: Is it important in the 16-bit case that the encoding be constant-width? If yes, I can add justification to 8 characters here as well. Or alternatively, this last case could cover all Unicode with `_u` used always instead of `_u`/`_u32`.

digama0: I think it has to be fixed width because otherwise `_uabcd` is confusable with `_uabc` + `d`.

Vtec234: Good point. I'm not sure if we ever unmangle atm, but still it's better to keep the mangling injective.

Kha (approved):

leodemoura (approved):

PR #540, opened 2021-06-22
fix: move `elabCommand` parts that should happen only once into new function

Kha:


src/Lean/Elab/Command.lean
```diff
@@ -285,13 +277,38 @@ partial def elabCommand (stx : Syntax) : CommandElabM Unit := do
...
+
+/--
+`elabCommand` wrapper that should be used for the initial invocation, not for recursive calls after
+macro expansion etc.
+-/
+def elabCommandTopLevel (stx : Syntax) : CommandElabM Unit := do
+  let initMsgs ← modifyGet fun st => (st.messages, { st with messages := {} })
+  let initInfoTrees ← getResetInfoTrees
+  withLogging do
+    runLinters stx
```
Kha, line 289: We definitely should run linters only on the original command

tests/lean/infoTree.lean.expected.out
```diff
@@ -316,7 +316,11 @@
               id : {α : Type} → α → α @ ⟨34, 35⟩-⟨34, 37⟩
             val : Nat → Nat := id @ ⟨34, 28⟩-⟨34, 31⟩
     pair : A × A := ({ val := id }, { val := id }) @ ⟨34, 2⟩-⟨34, 6⟩
-def id.{u} : {α : Sort u} → α → α :=
-fun {α : Sort u} (a : α) => a
-[Elab.info] command @ ⟨37, 0⟩-⟨37, 9⟩ @ Lean.Elab.Command.elabPrint
-  id : {α : Sort u} → α → α @ ⟨37, 7⟩-⟨37, 9⟩
+def Nat.xor : Nat → Nat → Nat :=
+bitwise bne
+[Elab.info] command @ ⟨37, 0⟩-⟨38, 10⟩ @ Lean.Elab.Command.expandInCmd
```
Kha, line 321: This tree is now printed correctly and exactly once

src/Lean/Elab/Command.lean
```diff
@@ -285,13 +277,38 @@ partial def elabCommand (stx : Syntax) : CommandElabM Unit := do
...
+  -- then convert new traces to messages
   let mut msgs ← (← get).messages
   -- `stx.hasMissing` should imply `initMsgs.hasErrors`, but the latter should be cheaper to check in general
   if !showPartialSyntaxErrors.get (← getOptions) && initMsgs.hasErrors && stx.hasMissing then
     -- discard elaboration errors, except for a few important and unlikely misleading ones, on parse error
     msgs := ⟨msgs.msgs.filter fun msg =>
       msg.data.hasTag `Elab.synthPlaceholder || msg.data.hasTag `Tactic.unsolvedGoals⟩
-  modify ({ · with messages := initMsgs ++ msgs })
+  for tree in (← getInfoTrees) do
+    let fmt ← tree.format
```
Vtec234, line 304: Eager formatting?

tests/lean/infoTree.lean.expected.out
Vtec234: For a while we have been wrapping top-level commands in `ofCommandInfo`, even when they expand to multiple ones as here. Are any cases of multiple top-level info trees left? If not, maybe we can enforce this and get rid of all the annoying [loops](https://github.com/leanprover/lean4/blob/8454102fc4b0b709ecf068a9d6db8e2b3f2a737c/src/Lean/Server/FileWorker/RequestHandling.lean#L91).

src/Lean/Elab/Command.lean
Kha: Yes, I forgot that `trace.Elab.info` and `enableInfoTree` are separate... let's wait for #541

tests/lean/infoTree.lean.expected.out
Kha: We can't change `Command.State` since we want it to implement `MonadInfoTree`; at best we could assert there is a single root in `compileNextCmd` and then store that in a new `Snapshot` field

Vtec234: That might work, I'll give implementing that in the server a go later.

PR #541, opened 2021-06-22
perf: make trace[...] ... notation lazy

gebner:
The current `trace[...]` macro has the slightly counterintuitive behavior that the argument is in general evaluated lazily, but monadic `<- t` lift methods are not.  (As opposed to e.g. if-then-else.)

This PR makes `trace[...]` a doElem-macro instead, which expands in such a way that the arguments are evaluated completely lazily.  The reason this only works as a doElem-macro and not as a term-macro is because lift methods in the do syntax are expanded after doElem-expansion, but before term-expansion.  To prevent surprises in the evaluation order, there is only the doElem-macro and no term-macro.

Changing the syntax category breaks several uses of trace in the tests.  These are all copies of the same `print` function.

As discussed on zulip: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.5Brfc.5D.20make.20trace.5B.2E.2E.2E.5D.20lazy

### DIFF
--- a/tests/lean/run/trace.lean
+++ b/tests/lean/run/trace.lean
@@ -38,6 +38,8 @@ do traceCtx `module $ do {
    -- Messages are computed lazily. The following message will only be computed
    -- if `trace.slow is active.
    trace[slow] (m!"slow message: " ++ toString (slow b))
+   -- This is true even if it is a monad computation:
+   trace[slow] (m!"slow message: " ++ (← toString (slow b)))
 
 def run (x : M Unit) : M Unit :=
 withReader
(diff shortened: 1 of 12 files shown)


tests/lean/run/trace.lean
```diff
@@ -38,6 +38,8 @@ do traceCtx `module $ do {
    -- Messages are computed lazily. The following message will only be computed
    -- if `trace.slow is active.
    trace[slow] (m!"slow message: " ++ toString (slow b))
+   -- This is true even if it is a monad computation:
```
Kha, line 41: The official term for this is "lifted method" :)

Kha (approved):

PR #542, opened 2021-06-22
fix: info on non-atomic simp args

Kha:


src/Lean/Elab/InfoTree.lean
```diff
@@ -315,6 +315,20 @@ def resolveGlobalConstWithInfos [MonadResolveName m] [MonadEnv m] [MonadError m]
 @[inline] def withInfoContext [MonadFinally m] (x : m α) (mkInfo : m Info) : m α := do
   withInfoTreeContext x (fun trees => do return InfoTree.node (← mkInfo) trees)
 
+@[inline] def withSaveInfoContext [MonadFinally m] [MonadEnv m] [MonadOptions m] [MonadMCtx m] [MonadResolveName m] [MonadFileMap m] (x : m α) : m α := do
```
Kha, line 318: @leodemoura I thought I would need this at both `CommandElabM` and `TermElabM` but only the latter was needed. I can specialize it if you want.

Kha: The name is also a bit confusing given the functions above with similar names but that have nothing to do with `InfoTree.Context` (so maybe *they* should be renamed).

PR #543, opened 2021-06-22
fix: file worker: don't wait for running tasks

gebner:
If an elaboration task had an infinite loop, the file worker used to run forever.  Even after closing the file or restarting the server.

This fix was suggested by Sebastian.  Let's see how many sanitizers this breaks in CI.

Zulip discussion: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4


gebner:
Closing in favor of #544.  Sebastian seems to have spotted an additional problem with IO.Process.exit.

PR #545, opened 2021-06-23
doc: document a few tactics

Kha:
I *tried* make the docstrings exhaustive (where interesting for the majority of users) by consulting each implementation, but I'm sure I missed some variants and quirks. Some parts were taken from Lean 3. Just wanted to ensure this style makes sense before I continue with the heavy hitters.

### DIFF
--- a/src/Init/Notation.lean
+++ b/src/Init/Notation.lean
@@ -227,7 +227,7 @@ syntax (name := intro) "intro " notFollowedBy("|") (colGt term:max)* : tactic
 syntax (name := intros) "intros " (colGt (ident <|> "_"))* : tactic
 /--
 `rename t => x` renames the most recent hypothesis whose type matches `t` (which may contain placeholders) to `x`,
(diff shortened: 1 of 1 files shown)


src/Init/Notation.lean
```diff
@@ -204,42 +204,114 @@ macro_rules
...
+  | n + 1, 0 => tac
+  | ...
+  ```
+-/
 syntax (name := intro) "intro " notFollowedBy("|") (colGt term:max)* : tactic
+/-- `intros x...` behaves like `intro x...`, but then keeps introducing (anonymous) hypotheses until goal is not of a function type. -/
 syntax (name := intros) "intros " (colGt (ident <|> "_"))* : tactic
+/--
+`rename t => x` renames the most recent hypothesis whose type matches `t` (which may contain placeholders) to `x`,
+or fails if no such hyptothesis could be found. -/
```
gebner, line 230:
```suggestion
or fails if no such hypothesis could be found. -/
```

PR #551, opened 2021-07-01
Misc server changes

Vtec234:
- Add go-to (projection) and hover docs for structure fields
- Check prefix in `set_option` completion and only show options with that prefix
- Go-to-declaration now jumps to the parser while go-to-def jumps to the elaborator as suggested by @gebner

### DIFF
--- a/tests/lean/interactive/completionOption.lean
+++ b/tests/lean/interactive/completionOption.lean
@@ -1,2 +1,2 @@
-set_option pp
-           --^ textDocument/completion
+set_option pp.proofs
+                  --^ textDocument/completion
--- a/.gitignore
+++ b/.gitignore
@@ -15,7 +15,8 @@ compile_commands.json
 tasks.json
 settings.json
 .gdb_history
-.vscode
+.vscode/*
+!.vscode/settings.json
 *.produced.out
 CMakeSettings.json
 CppProperties.json
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "files.insertFinalNewline": true
+}
--- a/tests/lean/interactive/completionOption.lean.expected.out
+++ b/tests/lean/interactive/completionOption.lean.expected.out
@@ -1,68 +1,10 @@
 {"textDocument": {"uri": "file://completionOption.lean"},
- "position": {"line": 0, "character": 13}}
+ "position": {"line": 0, "character": 20}}
 {"items":
- [{"label": "pp.all",
-   "detail":
-   "(false), (pretty printer) display coercions, implicit parameters, proof terms, fully qualified names, universe, and disable beta reduction and notations during pretty printing"},
-  {"label": "pp.auxDecls",
-   "detail":
-   "(false), display auxiliary declarations used to compile recursive functions"},
-  {"label": "pp.binder_types",
-   "detail":
-   "(false), (pretty printer) display types of lambda and Pi parameters"},
-  {"label": "pp.coercions",
-   "detail": "(true), (pretty printer) hide coercion applications"},
-  {"label": "pp.explicit",
-   "detail": "(false), (pretty printer) display implicit arguments"},
-  {"label": "pp.full_names",
-   "detail": "(false), (pretty printer) display fully qualified names"},
-  {"label": "pp.inaccessibleNames",
-   "detail": "(false), display inaccessible declarations in the local context"},
-  {"label": "pp.macroStack",
-   "detail": "(false), dispaly macro expansion stack"},
-  {"label": "pp.motives.all",
-   "detail": "(false), (pretty printer) print all motives"},
-  {"label": "pp.motives.nonConst",
-   "detail":
-   "(false), (pretty printer) print all motives that are not constant functions"},
-  {"label": "pp.motives.pi",
-   "detail": "(true), (pretty printer) print all motives that return pi types"},
-  {"label": "pp.notation",
-   "detail":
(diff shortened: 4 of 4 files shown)


src/Lean/Server/Requests.lean
```diff
@@ -49,6 +49,7 @@ end RequestError
 structure RequestContext where
   srcSearchPath : SearchPath
   docRef        : IO.Ref FileWorker.EditableDocument
+  hLog          : IO.FS.Stream
```
Kha, line 52: I usually just use `dbg_trace` in request handlers fwiw, works well enough

src/Lean/Server/Completion.lean
```diff
@@ -242,13 +242,23 @@ private def dotCompletion (ctx : ContextInfo) (info : TermInfo) (expectedType? :
...
-    let opts ← getOptions
-    let items : Array CompletionItem := decls.fold (init := #[]) fun items name decl =>
-      items.push { label := name.toString, detail? := s!"({opts.get name decl.defValue}), {decl.descr}", documentation? := none }
-    return some { items := sortCompletionItems items, isIncomplete := true }
+      let partialName := stx[1].getId
+      -- HACK(WN): unfold the type so ForIn works
+      let (decls : Std.RBMap _ _ _) ← getOptionDecls
+      let opts ← getOptions
+      let mut items := #[]
+      for ⟨name, decl⟩ in (decls : Std.RBMap _ _ _) do
```
Kha, line 252: I'm assuming you don't need a second `(decls : ...)` here

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -56,15 +56,18 @@ partial def handleHover (p : HoverParams)
 
       return none
 
-open Elab in
-partial def handleDefinition (goToType? : Bool) (p : TextDocumentPositionParams)
+inductive GoToKind | declaration | definition | type
+deriving DecidableEq
```
Kha, line 60:
```suggestion
inductive GoToKind
  | declaration | definition | type
  deriving DecidableEq
```

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -92,16 +95,20 @@ partial def handleDefinition (goToType? : Bool) (p : TextDocumentPositionParams)
...
           if let Info.ofTermInfo ti := i then
             let mut expr := ti.expr
-            if goToType? then
+            if kind = type then
               expr ← ci.runMetaM i.lctx do
                 Meta.instantiateMVars (← Meta.inferType expr)
             if let some n := expr.constName? then
               return ← ci.runMetaM i.lctx <| locationLinksFromDecl i n
+          if let Info.ofFieldInfo fi := i then
+            if kind ≠ type then
```
Kha, line 104: Should we not support going to the type of a structure field?

Kha (commented):
Could you add tests please?

src/Lean/Server/FileWorker/RequestHandling.lean
Vtec234: I didn't add this since projections have Pi-types, but you're right, it is more convenient to just go to the codomain.

src/Lean/Server/Requests.lean
Vtec234: Oh, at some past point I had issues with that since the main thread's `stderr` wasn't inherited by task threads. It seems to work now, but I see no harm in leaving that log stream field.

Vtec234:
The latest change is not in the server but I thought it didn't warrant a separate PR. Just adds more `UInt` casts.

tests/lean/interactive/completionOption.lean
```diff
@@ -0,0 +1,2 @@
+set_option pp
```
Kha, line 1: We should use a prefix that is less likely to break the test when new options are added. Perhaps `pp.proofs`?

src/Lean/Server/Completion.lean
```diff
@@ -242,13 +242,23 @@ private def dotCompletion (ctx : ContextInfo) (info : TermInfo) (expectedType? :
...
+      let opts ← getOptions
+      let mut items := #[]
+      for ⟨name, decl⟩ in decls do
+        if partialName.isPrefixOf name ||
+          (match partialName, name with
+            | Name.str p₁ s₁ _, Name.str p₂ s₂ _ => p₁ == p₂ && s₁.isPrefixOf s₂
+            | _, _ => false) then
+          items := items.push
+            { label := name.toString
+              detail? := s!"({opts.get name decl.defValue}), {decl.descr}"
```
Kha, line 259: This is confusing, the default value may not be the current value. "(true), (pretty printer) ..." also looks a bit weird I think.

Kha: Whoops, I thought I double checked that this was indeed new code :smile:

tests/lean/interactive/completionOption.lean
```diff
@@ -0,0 +1,2 @@
...
+           --^ textDocument/completion
```
Kha, line 2: Btw, please make sure to end files with a trailing newline to avoid unnecessary diffs. Can we commit a `.vscode/` that does that?

src/Lean/Server/Completion.lean
Vtec234: I agree it looks a bit strange, but I also tried using the proper `documentation?` LSP field and with that you have to expand the popup to see the description, whereas with `details` it always shows. Note also that `opts.get name decl.defValue` is the *current* value (right?).

Kha: Ah, you are right about the current value

PR #553, opened 2021-07-02
fix: simp: respect `Config.ctorEq`

Kha:


Kha:
Ah, I've realized the actual injection theorems are added as regular simp theorems and thus can't be categorically turned off easily. So perhaps the configuration option should be removed instead.

leodemoura:
> Ah, I've realized the actual injection theorems are added as regular simp theorems and thus can't be categorically turned off easily. So perhaps the configuration option should be removed/renamed to `ctorNeq` instead.

I think we should remove it. If there is a need for disabling the simplification step for `ctor1 ... = ctor2 ...`, we add the `ctorNeq` option.

PR #558, opened 2021-07-07
fix: ToJson for single named field constructors

Vtec234:
They weren't round-tripping.


Kha:
The decision to change the JSON structure depending on whether all constructor arguments are named or not starts looking a bit questionable to me now, but not the topic of this PR.

PR #564, opened 2021-07-10
fix: mkAtomFrom: generate synthetic position like other `*From` functions

Kha:
Also consistently use binders as reference position for an elided binder type.
Before, type errors were always given extent 1, the length of the
synthetic `_` token.

### DIFF
--- a/src/Lean/Elab/LetRec.lean
+++ b/src/Lean/Elab/LetRec.lean
@@ -115,6 +115,8 @@ private def registerLetRecsToLift (views : Array LetRecDeclView) (fvars : Array 
 @[builtinTermElab «letrec»] def elabLetRec : TermElab := fun stx expectedType? => do
   let view ← mkLetRecDeclView stx
   withAuxLocalDecls view.decls fun fvars => do
+    for decl in view.decls, fvar in fvars do
+      addTermInfo (isBinder := true) decl.ref[0] fvar
     let values ← elabLetRecDeclValues view
     let body ← elabTermEnsuringType view.body expectedType?
     registerLetRecsToLift view.decls fvars values
--- a/tests/lean/interactive/goTo.lean
+++ b/tests/lean/interactive/goTo.lean
@@ -49,5 +49,6 @@ example : Nat :=
   let a := 1
 --v textDocument/definition
   a + b
+    --^ textDocument/definition
 where
   b := 2
--- a/tests/lean/interactive/goTo.lean.expected.out
+++ b/tests/lean/interactive/goTo.lean.expected.out
@@ -96,3 +96,12 @@
   {"start": {"line": 48, "character": 6}, "end": {"line": 48, "character": 7}},
   "originSelectionRange":
   {"start": {"line": 50, "character": 2}, "end": {"line": 50, "character": 3}}}]
+{"textDocument": {"uri": "file://goTo.lean"},
+ "position": {"line": 50, "character": 6}}
+[{"targetUri": "file://goTo.lean",
+  "targetSelectionRange":
+  {"start": {"line": 53, "character": 2}, "end": {"line": 53, "character": 3}},
+  "targetRange":
+  {"start": {"line": 53, "character": 2}, "end": {"line": 53, "character": 3}},
+  "originSelectionRange":
+  {"start": {"line": 50, "character": 6}, "end": {"line": 50, "character": 7}}}]


tests/lean/interactive/plainTermGoal.lean.expected.out
```diff
@@ -47,7 +47,7 @@
  "position": {"line": 23, "character": 6}}
 {"range":
  {"start": {"line": 23, "character": 6}, "end": {"line": 23, "character": 7}},
- "goal": "⊢ Type"}
+ "goal": "n : Nat\n⊢ Nat"}
```
Kha, line 50: @gebner A welcome improvment, I assume :smile: 
(input is
```
  fun n => Nat.ltOfLeOfLt (Nat.leAddRight n 41) (Nat.ltSuccSelf _)
    --^ $/lean/plainTermGoal
```
)

gebner: Yes, indeed!

(I just checked and Lean 3 also prints `⊢ Type` here.)

Vtec234: The goal on `n` having `n` as a hypothesis is a bit strange though, no?

Kha: Mmh, I suppose so! From the server's point of view, the term info is the same as on *usages* of `n`, in which case the goal makes more sense.

gebner: As far as I'm concerned, this is a not particularly important edge-case here and both behaviors seem fine to me.  The correct answer to the question "what goal does `n` fill?" is neither of them though.  The answer should be `⊢ Nat`.

leodemoura: I agree with @Vtec234, the goal looks very strange. The new goal looks like a bug to me. For me, the possible behaviors are
1- We show the goal for the implicit hole `n : _`, and the goal is just `⊢ Type` and it is in master right now.
2- We show the type of `n`, and make it clear it is a type and not a goal. In this case, the output should be just `n : Nat`. I think this is a bit of stretch since the request was `/lean/plainTermGoal`.
3- The request `lean/plainTermGoal` doesn't produce anything at the binder `n`.

gebner: I'm happy with anything that shows the local context.  Out of Leo's list, this rules out everything except 1 (the current behavior in master).

leodemoura: @gebner We could show the local context on 2. In this particular example, it would be just `⊢ n : Nat`. In the following example, 
```lean
variable (α : Type) (f : α → α)

#check fun a => f a
         --^ $/lean/plainTermGoal
```
The result would be
```lean
α : Type
f : α → α
⊢ 
a : α
```
As I said above, the name of the request is still a bit misleading.

Kha: There seems to be rough consensus that there isn't really any kind of "goal" that could be associated with this occurrence of `n`, nor really an "expected type" (which is what the UIs print above the result of this request). Given that, should we ignore `a` and show the same result as on e.g. `fun`? I'm not sure if that falls under Leo's 3), but it seems to fulfill Gabriel's condition.
Implementation-wise, I think we would need a new flag, say `TermInfo.showAsGoal`, to do that.

gebner: > @gebner We could show the local context on 2. In this particular example, it would be just `⊢ n : Nat`. In the following example, The result would be
> 
> ```lean
> α : Type
> f : α → α
> ⊢ 
> a : α
> ```

Ok, then I misunderstood you.  This is also useful.  It almost exactly matches what I suggested before:
```lean
α : Type
f : α → α
⊢ α
```

leodemoura: @gebner Yes, I just want to make it clear the output is not a goal, but the type of term/binder `a` in a particular local context.

leodemoura: @Kha What about displaying 
```lean
Binder type:
α : Type
f : α → α
⊢ α
```
in the Info panel. 
In the original example, the output would be
```lean
Binder type:
⊢ Nat
```
Another option is to just remove the "Expected", and just display "Type:".
The PR still has to be fixed since it is producing `"goal": "n : Nat\n⊢ Nat"` instead of `"goal": "⊢ Nat"`

leodemoura: It still bothers me that we are using "goal" in the LSP message for something that is not a goal.

leodemoura: That being said, I will be happy to merge the PR if we display 
```lean
Type:
α : Type
f : α → α
⊢ α
```
and 
```lean
Type:
⊢ Nat
```
in the examples above, even if the message is still using `"goal"` to name the JSON field.

Kha: Ok, we could do that with a different kind of new flag, say `TermInfo.isReference`, which renders goals that are fvars with that fvar excluded from the printed context. This flag could be useful for other in other cases as well, e.g. local go-to-definition.

Vtec234: What do you think about a separate `BinderInfo`? Binders aren't terms on their own, so maybe it's better not to store them as such. The associated `Expr` would be the type of the binder in a context before said binder is introduced.

Vtec234: Btw, there is also interaction with dependency. Right now we see:
```lean
def test₁ : (∃ (a : Nat), (a < 10)) → true = true :=
/-
x✝ : ∃ (a : Nat), a < 10
a : Nat
h : a < 10
⊢ Nat
     --v -/
  fun ⟨a, h⟩ => rfl
/-      --^
x✝ : ∃ (a : Nat), a < 10
a : Nat
h : a < 10
⊢ a < 10
-/
```
while ideally it would look like
```lean
/-
x✝ : ∃ (a : Nat), a < 10
⊢ Nat
     --v -/
  fun ⟨a, h⟩ => rfl
/-      --^
x✝ : ∃ (a : Nat), a < 10
a : Nat
⊢ a < 10
-/
```

Kha: > What do you think about a separate `BinderInfo`?

Sounds good as well, I see that was mentioned as an alternative in the original commit storing binder info: https://github.com/leanprover/lean4/commit/fb52ec8ef54ce673c009e2c1a6ea7184f02db3be

Kha: In the end I went with `isBinder` to avoid unnecessary code duplication... and because the otherwise apt name `BinderInfo` is already taken

Kha: Added go-to-definition for local variables to validate the design, and also because I really wanted that feature

PR #565, opened 2021-07-10
perf: extract cold paths in lean_free_small and mark noinline 

Kha:
While staring at the assembly output in https://github.com/leanprover/lean4/pull/555#issuecomment-877315772, I noticed that quite a bit of time on the hot path (deallocating allocation from the current thread's heap) is spent spilling registers, which are not actually used in the hot path, on the stack. Outlining the cold paths should relieve register and instruction cache pressure.

### DIFF
--- a/src/include/lean/lean.h
+++ b/src/include/lean/lean.h
@@ -40,12 +40,10 @@ extern "C" {
 #define LEAN_UNLIKELY(x) (__builtin_expect((x), 0))
 #define LEAN_LIKELY(x) (__builtin_expect((x), 1))
 #define LEAN_ALWAYS_INLINE __attribute__((always_inline))
-#define LEAN_NOINLINE __attribute__((noinline))
 #else
 #define LEAN_UNLIKELY(x) (x)
 #define LEAN_LIKELY(x) (x)
 #define LEAN_ALWAYS_INLINE
-#define LEAN_NOINLINE
 #endif
 
 #ifdef LEAN_RUNTIME_STATS
--- a/src/runtime/alloc.cpp
+++ b/src/runtime/alloc.cpp
@@ -10,6 +10,12 @@ Author: Leonardo de Moura
 #include <lean/alloc.h>
 #include <lean/lean.h>
 
+#if defined(__GNUC__) || defined(__clang__)
+#define LEAN_NOINLINE __attribute__((noinline))
+#else
+#define LEAN_NOINLINE
+#endif
+
 #define LEAN_PAGE_SIZE             8192        // 8 Kb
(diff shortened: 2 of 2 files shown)


Kha:
!bench

Kha:
Not bad! Let's try that for the dual function too.

Kha:
!bench

Kha:
!bench

Kha:
Oh, looks like I completely missed that in the first benchmark, *only* changes in instruction counts were reported, no actual wall time changes. But that partially seems to be the fault of above-average variance in that run, since changes < 10 stddev are not reported. All in all, it looks like each of the changes sped up every benchmark at least somewhat, with the biggest wins at ~5+5% in unionfind.

src/runtime/alloc.cpp
```diff
@@ -448,6 +458,17 @@ void * alloc(size_t sz) {
...
+    g_heap->m_to_export_list_size++;
+    if (g_heap->m_to_export_list_size > LEAN_MAX_TO_EXPORT_OBJS) {
+        LEAN_RUNTIME_STAT_CODE(g_num_exports++);
+        g_heap->export_objs();
+    }
+}
+
 static inline void dealloc_small_core(void * o) {
     LEAN_RUNTIME_STAT_CODE(g_num_small_dealloc++);
     if (LEAN_UNLIKELY(g_heap == nullptr)) {
```
Kha, line 474: @leodemoura Does this dynamic check exist for using Lean libraries from non-Lean code? Since for every program/thread started from Lean we eagerly initialize the heap I believe.
Even if the branch is perfectly predictable, we might want to check whether it still incurs some overhead, in which case we could make heap initialization the external user's burden, or possibly that of the Lean module initialization functions.

leodemoura: If I remember correctly the test above had two motivations. The one you pointed out, and C++ static initialization code. Both scenarios are not gone. The former may reappear in the future, but we can always instruct users to explicitly call an initialization function.

src/include/lean/lean.h
```diff
@@ -40,10 +40,12 @@ extern "C" {
 #define LEAN_UNLIKELY(x) (__builtin_expect((x), 0))
 #define LEAN_LIKELY(x) (__builtin_expect((x), 1))
 #define LEAN_ALWAYS_INLINE __attribute__((always_inline))
+#define LEAN_NOINLINE __attribute__((noinline))
```
leodemoura, line 43: I don't see a good reason for defining it here at `lean.h` (our public interface).

Kha: Inline in `alloc.cpp` then, until we have a need for it in other files?

PR #568, opened 2021-07-13
Infoview RPC

Vtec234:
(Ready to review.)

## Motivation

This PR implements infoview RPC. The problem this solves is that the infoview client often needs to execute functions with heavy-weight objects. For example, with an `Environment` in order to pretty-print an `Expr`. Serializing these into JSON would be incredibly inefficient, so instead we allow the client to refer to server-side objects using `RpcRef`s and execute RPC calls with these objects. As far as the client is concerned, these are opaque IDs.

## Implementation

To this end, we:
- Extend the file worker state with the concept of an RPC session. This is a table of objects kept alive by the server so that the client can refer to them. See sections below for details.
- Implement parametrised `deriving` with the syntax `deriving instance Class with { config } for Foo`. There's a test case [here](https://github.com/leanprover/lean4/pull/568/files#diff-a5395953ab5af87797bd9e8566322144b4b6bbb81f10f56a31478a68fe060d60) and example usage [here](https://github.com/Vtec234/lean4/commit/64399ee2756627f44f26648de8b01949467ff41d#diff-96f9ef9586f676ff71dab6f45c8bb6ed69a351c862885e43c24323c696118ef0R82). This makes it much nicer to auto-derive LSP encodings for structures containing `RpcRef`s.
- Introduce `Format.tag` for storing references to other objects (such as subexpressions) in `Format`s.

## Validation

When an opaque `RpcRef` is received from the client, we must validate that it refers to a real object and has the right type. We do so by looking it up in the RPC session [table](https://github.com/leanprover/lean4/pull/568/files#diff-5225092adf9703f2208b276d72487c6481012b3dce8ff7393f9d3cb5e5a74269R64). The table contains a fully-qualified type name of the object which is then validated at the use site. It may be possible to break by defining different types with the same name in separate modules and mixing them in RPC calls, but outside of this contrived scenario 99% of use cases should be safe. This approach is conservative because it will fail for defeq types. If we eventually compile type-safe infoview code from Lean this will be less of an issue, but currently the infoview is TypeScript.

## GC

We keep objects alive on the server for as long as the client may reference them by storing them in the RPC session. It is initialized and re-initialized when the client starts up, as well as when the client or server restarts, so refs should never become out-of-date. For long-lived client sessions, we rely on the client to tell the server when it no longer needs a reference so that the server can drop it. I implemented a prototype which hooks into a [`FinalizationRegistry`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry), effectively offloading the ref-counting of client objects to the JS GC.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -291,9 +291,6 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   set(LEANC_STATIC_LINKER_FLAGS "-lleancpp -lInit -lStd -lLean")
 elseif(${CMAKE_SYSTEM_NAME} MATCHES "Emscripten")
   set(LEANC_STATIC_LINKER_FLAGS "-lleancpp -lInit -lStd -lLean -lnodefs.js")
-elseif(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
-  # `-pie` defaulting is not consistent on Linux distributions, so let's default to off
-  set(LEANC_STATIC_LINKER_FLAGS "-no-pie -Wl,--start-group -lleancpp -lInit -lStd -lLean -Wl,--end-group -lbcrypt")
 else()
   # `-pie` defaulting is not consistent on Linux distributions, so let's default to off
   set(LEANC_STATIC_LINKER_FLAGS "-no-pie -Wl,--start-group -lleancpp -lInit -lStd -lLean -Wl,--end-group")
@@ -312,6 +309,12 @@ else()
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${LEAN_EXTRA_LINKER_FLAGS} -rdynamic")
   set(LEANC_STATIC_LINKER_FLAGS "${LEANC_STATIC_LINKER_FLAGS} -ldl")
   set(LEANC_SHARED_LINKER_FLAGS "${LEANC_SHARED_LINKER_FLAGS} -ldl")
+endif()
+
+# On Windows, add bcrypt for random number generation
+if(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
+  set(LEANC_STATIC_LINKER_FLAGS "${LEANC_STATIC_LINKER_FLAGS} -lbcrypt")
+  set(LEANC_SHARED_LINKER_FLAGS "${LEANC_SHARED_LINKER_FLAGS} -lbcrypt")
 endif()
 
 # Allow `lean` symbols in plugins without linking directly against it. If we linked against the


Kha:
Just FYI, un-drafting a PR (or changing the title) does not generate a GitHub notification. I only noticed it is ready while scanning the list of PRs :) .

src/Init/System/IO.lean
```diff
@@ -496,6 +496,29 @@ instance : MonadLift (ST IO.RealWorld) (EIO ε) := ⟨fun x s =>
 def mkRef (a : α) : IO (IO.Ref α) :=
   ST.mkRef a
 
+/-- Keeps an object alive by ensuring that its RC >= 1. Unlike holding typed `Ref`s,
+this structure is homogeneous. The ref can safely be shared between threads. -/
```
Kha, line 500: > The ref can safely be shared between threads

Is this the only reason it's a ref at all? It doesn't look like we ever write to it.

src/Init/System/IO.lean
```diff
@@ -496,6 +496,29 @@ instance : MonadLift (ST IO.RealWorld) (EIO ε) := ⟨fun x s =>
...
+  ptr : USize
+
+namespace UntypedRef
+
+/-- Creation is unsafe because it must be ensured that:
+- the `ref` field is never used to access the value as a type other than `α`
+- the type `α` is not a scalar -/
+unsafe def mkRefUnsafe {α : Type u} (a : α) : IO UntypedRef := do
+  let ref ← IO.mkRef (unsafeCast a)
+  let ptr := ptrAddrUnsafe a
```
Kha, line 512: Will multiple refs to the same object not share the same `ptr`, and thus the same `Lsp.RpcRef`?

src/Lean/Server/FileWorker/LspEncoding.lean
```diff
@@ -0,0 +1,239 @@
...
+      let fieldInsts (func : Name) := do fieldIds.mapM fun fid =>
+        `(Parser.Term.structInstField| $fid:ident := ← $(mkIdent func):ident a.$fid:ident)
+      let encFields ← fieldInsts ``lspEncode
+      let decFields ← fieldInsts ``lspDecode
+      `(
+        namespace $typeId:ident
+
+        private structure LspPacket where
+          $[($fieldIds : $fieldEncTIds)]*
+          deriving $(mkIdent ``Lean.FromJson), $(mkIdent ``Lean.ToJson)
```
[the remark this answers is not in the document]
Vtec234: Yes, it fails. Here is a small repro:
```lean
class Bar

structure Baz₁
  deriving Bar -- default handlers have not been implemented yet, class: 'Bar' types: [Baz₁]

elab "elabMe" : command => do
  let stx ← `(structure Baz₂
    deriving Bar)
  elabCommand stx
elabMe -- unknown constant 'Bar✝'
```
This is actually something I'd been wondering about regarding quotations. Generally, quotations seem to introduce `_hyg` on all idents which I suppose makes sense as there is no way in general for them to tell whether the ident is a resolvable global const or something local. But for this to work it seems everything which takes idents must then strip the macro scopes, which `deriving ..` doesn't. Should it? Is that the bug?

src/Lean/Server/FileWorker/LspEncoding.lean
```diff
@@ -0,0 +1,239 @@
...
+        let some fieldEncTName ← fieldEncT.constName?
+          | throwError "type '{fieldEncT}' is not constant"
+        fieldEncTs := fieldEncTs.push fieldEncTName
+
+      let typeId := mkIdent typeName
+
+      let fieldIds := fields.map mkIdent
+      let fieldEncTIds := fieldEncTs.map mkIdent
+
+      -- HACK(WN): the syntax this produces looks royally messed up but elaborates fine
```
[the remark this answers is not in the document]
Vtec234: The `trace.Elab.command` output is wonky, but it may actually just be broken formatting of hygienic idents:
```lean
[Elab.command] instance  : LspEncoding✝ (WithRpcRef✝ FooRef) Lsp.RpcRef✝where lspEncode✝ a✝ := encode✝ a✝lspDecode✝
  a✝ := decode✝ a✝
```
It's not high prio. I'll remove the comment.

src/Lean/Server/FileWorker.lean
```diff
@@ -97,10 +97,14 @@ structure WorkerContext where
...
-  docRef             : IO.Ref EditableDocument
-  pendingRequestsRef : IO.Ref PendingRequestMap
 
-abbrev WorkerM := ReaderT WorkerContext IO
+structure WorkerState where
+  doc             : EditableDocument
+  pendingRequests : PendingRequestMap
+  rpcSesh         : Option RpcSession
+
+-- NOTE(WN): StateT or StateRefT?
```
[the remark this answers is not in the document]
Vtec234: I thought `StateT` might be okay since we don't pile more than one state here, but I guess there's no good reason to avoid `StateRefT` either.

src/Init/System/IO.lean
Vtec234: I think my reasoning in one of the outsourced-GC prototypes was that the *action* of incing the RC should really be an IO action so as to not get erased by any const compiler magic. But since we have an `aliveRefs` table now which keeps things alive, I think said table could just contain `NonScalar`s. Is this what you're suggesting?

src/Lean/Elab/Deriving/Basic.lean
```diff
@@ -1,61 +1,69 @@
...
 open Command
 
-def DerivingHandler := (typeNames : Array Name) → CommandElabM Bool
+def DerivingHandler := (typeNames : Array Name) → (args? : Option Syntax) → CommandElabM Bool
+def DerivingHandlerNoArgs := (typeNames : Array Name) → CommandElabM Bool
 
 builtin_initialize derivingHandlersRef : IO.Ref (NameMap DerivingHandler) ← IO.mkRef {}
 
-def registerBuiltinDerivingHandler (className : Name) (handler : DerivingHandler) : IO Unit := do
+def registerBuiltinDerivingHandler' (className : Name) (handler : DerivingHandler) : IO Unit := do
```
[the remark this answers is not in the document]
Vtec234: I defaulted to non-parameterised handlers since that's most of them. Renamed the parametrised to a mouthful `registerBuiltinDerivingHandlerWithArgs`.

src/Init/System/IO.lean
Kha: Sounds good

src/Lean/Server/FileWorker/LspEncoding.lean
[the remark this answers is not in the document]
Kha: Ah, this is a bug in the name resolution used by `deriving`. The compile-time resolution information is stored as metadata on the identifier, but this specific resolution ignores that metadata. Fixed in https://github.com/leanprover/lean4/pull/583/files, but let's not worry about it for this PR.

src/CMakeLists.txt
```diff
@@ -264,6 +264,9 @@ else()
   set(COPY_LIBS ${COPY_LIBS} ${GMP_LIBRARIES})
   # dlopen
   set(EXTRA_LIBS ${EXTRA_LIBS} ${CMAKE_DL_LIBS})
+  if("${CMAKE_SYSTEM_NAME}" MATCHES "Windows")
+    set(EXTRA_LIBS ${EXTRA_LIBS} "bcrypt")
```
Vtec234, line 268: It seems we have to link this on Windows to get random bytes.

src/CMakeLists.txt
```diff
@@ -291,6 +291,9 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
   set(LEANC_STATIC_LINKER_FLAGS "-lleancpp -lInit -lStd -lLean")
 elseif(${CMAKE_SYSTEM_NAME} MATCHES "Emscripten")
   set(LEANC_STATIC_LINKER_FLAGS "-lleancpp -lInit -lStd -lLean -lnodefs.js")
+elseif(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
+  # `-pie` defaulting is not consistent on Linux distributions, so let's default to off
+  set(LEANC_STATIC_LINKER_FLAGS "-no-pie -Wl,--start-group -lleancpp -lInit -lStd -lLean -Wl,--end-group -lbcrypt")
```
Kha, line 296: Let's add `-lbcrypt` to `LEANC_STATIC_LINKER_FLAGS` in a separate block so as not to duplicate this line

Kha (approved):

src/CMakeLists.txt
Kha: And I guess it should be added to `LEANC_SHARED_LINKER_FLAGS` as well

PR #570, opened 2021-07-14
Dany/low level code gen

DanielFabian:
this is the branch to put away for archival.


Kha:
Stored at https://github.com/leanprover/lean4/tree/dany/low-level-code-gen

PR #571, opened 2021-07-14
refactor: move ir decoding helpers to separate header / implementation file

DanielFabian:
This is the PR we should apply.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ settings.json
 CMakeSettings.json
 CppProperties.json
 result
+.cache/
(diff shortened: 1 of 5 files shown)


leodemoura:
!bench

Kha:
Yeah, these accessors are all simple enough that they should just be `static inline`. We can do that when we need them in a header.

DanielFabian:
uff, so my moving the code around causes a sizable perf regression? that's really unfortunate, sorry.

Kha:
I'll close this for now then

PR #572, opened 2021-07-14
Fix splitOn and field hovers

Vtec234:
See [here](https://github.com/leanprover/lean4/commit/872228e60aec1158a7e5b6085b101a33f0799076#diff-624026c6ea9fe213570ea4a86deda51060ac39ab5abf28debc98b06a8f541015) for the hovers that were incorrect.

### DIFF
--- a/src/Init/Data/String/Basic.lean
+++ b/src/Init/Data/String/Basic.lean
@@ -345,23 +345,23 @@ namespace Substring
 
 /-- Given an offset of a codepoint into the substring,
 return the offset there of the next codepoint. -/
-@[inline] private def next : Substring → String.Pos → String.Pos
+@[inline] def next : Substring → String.Pos → String.Pos
   | ⟨s, b, e⟩, p =>
     let absP := b+p
     if absP = e then p else s.next absP - b
 
 /-- Given an offset of a codepoint into the substring,
 return the offset there of the previous codepoint. -/
-@[inline] private def prev : Substring → String.Pos → String.Pos
+@[inline] def prev : Substring → String.Pos → String.Pos
   | ⟨s, b, _⟩, p =>
     let absP := b+p
     if absP = b then p else s.prev absP - b
 
-private def nextn : Substring → Nat → String.Pos → String.Pos
+def nextn : Substring → Nat → String.Pos → String.Pos
   | ss, 0,   p => p
   | ss, i+1, p => ss.nextn i (ss.next p)
 
-private def prevn : Substring → String.Pos → Nat → String.Pos
+def prevn : Substring → String.Pos → Nat → String.Pos
   | ss, 0,   p => p
   | ss, i+1, p => ss.prevn i (ss.prev p)
 
@@ -390,7 +390,7 @@ or `s.bsize` if `c` doesn't occur. -/
   | ⟨s, b, e⟩, p => b + p == e
 
 @[inline] def extract : Substring → String.Pos → String.Pos → Substring
-  | ⟨s, b, _⟩, b', e' => if b' ≥ e' then ⟨"", 0, 1⟩ else ⟨s, b+b', b+e'⟩
+  | ⟨s, b, e⟩, b', e' => if b' ≥ e' then ⟨"", 0, 0⟩ else ⟨s, e.min (b+b'), e.min (b+e')⟩
 
 partial def splitOn (s : Substring) (sep : String := " ") : List Substring :=
   if sep == "" then


src/Lean/Elab/App.lean
```diff
@@ -785,9 +785,15 @@ private partial def elabAppFn (f : Syntax) (lvals : List LVal) (namedArgs : Arra
       f.getArgs.foldlM (fun acc f => elabAppFn f lvals namedArgs args expectedType? explicit ellipsis true acc) acc
   else
     let elabFieldName (e field : Syntax) := do
-      let newLVals := field.getId.eraseMacroScopes.components.map fun n =>
-        -- We use `none` here since `field` can't be part of a composite name
-        LVal.fieldName field (toString n) none e
+      let newLVals := 
+        -- We use `none` in `suffix?` since `field` can't be part of a composite name
+        if field.getId.hasMacroScopes then
+          -- We assume that names with macro scopes have no meaningful syntax associated
```
Vtec234, line 791: I'm not 100% sure this is a correct assumption, is this fine?

src/Lean/Syntax.lean
```diff
@@ -172,6 +172,24 @@ partial def getTailWithPos : Syntax → Option Syntax
   | node _ args         => args.findSomeRev? getTailWithPos
   | _                   => none
 
+open SourceInfo in
+/-- Splits an `ident` into its dot-separated components while preserving source info.
+For example, `` `foo.bla.boo `` ↦ `` [`foo, `bla, `boo] ``. -/
+def identComponents : Syntax → List Syntax
```
Vtec234, line 178: I hoped to also use this in `Term.resolveName'` but there we also have to compute a substring like `foo.bla` (if `foo.bla` was a namespaced identifier), so doesn't quite fit.

src/Lean/Syntax.lean
```diff
@@ -172,6 +172,24 @@ partial def getTailWithPos : Syntax → Option Syntax
...
+  | ident info rawStr val _ =>
+    rawStr.splitOn "." |>.map fun ss =>
+      let off := ss.startPos - rawStr.startPos
+      let info := match info with
+        | original lead pos trail _ =>
+          let lead := if off == 0 then lead else "".toSubstring
+          let trail := if ss.stopPos == rawStr.stopPos then trail else "".toSubstring
+          original lead (pos + off) trail (pos + off + ss.bsize)
+        | synthetic pos _ =>
+          synthetic (pos + off) (pos + off + ss.bsize)
```
Kha, line 188: I believe that in the synthetic case, all components should inherit the same info since synthetic syntax trees do not have any "physical extent". Perhaps that could also solve the Elab.App issue? And yes, we should amend `resolveName'` as well.

src/Lean/Syntax.lean
Kha: What about an optional parameter to keep the first `n` components as a single `Syntax`, then pass `n := id.components.length` from `resolveName'`, would that work?

src/Lean/Syntax.lean
```diff
@@ -172,6 +172,24 @@ partial def getTailWithPos : Syntax → Option Syntax
   | node _ args         => args.findSomeRev? getTailWithPos
   | _                   => none
 
+open SourceInfo in
+/-- Splits an `ident` into its dot-separated components while preserving source info.
+For example, `` `foo.bla.boo `` ↦ `` [`foo, `bla, `boo] ``. -/
+def identComponents : Syntax → List Syntax
+  | ident info rawStr val _ =>
+    rawStr.splitOn "." |>.map fun ss =>
```
Kha, line 180: Same issue in `resolveName'`, but this can break in presence of escaped identifiers, right? Another reason not to implement it twice :) .

src/Lean/Elab/App.lean
Vtec234: The problem is that `identComponents` operates on the `rawStr : Substring` field of an `ident`, while in `ident`s with macro scopes, `rawStr` may not match `val : Name`. For example, we have
```lean
deriving instance Repr for Syntax
elab "testme" : command => do
  let stx ← `(foo.test)
  trace[Elab.command] "{repr stx}"

/- [Elab.command] Lean.Syntax.ident (Lean.SourceInfo.synthetic 1064 1084) "foo.test".toSubstring `foo.test._@._hyg.996 [] -/
```
However in the current code we `eraseMacroScopes` anyway. So if it's an invariant that `rawStr.toString == repr val.eraseMacroScopes` then perhaps we can just not use `eraseMacroScopes`. In fact, I can build stage 2 assuming this.

src/Lean/Syntax.lean
Vtec234: If the escaped part includes dots, yes. Is there ever a legitimate use case for this? I would suggest to just forbid dots between French quotes.

src/Lean/Elab/App.lean
Vtec234: Oh well, it builds but the assumption is false. We also have `` Syntax.ident SourceInfo.none "reprSyntax._@._hyg.5".toSubstring `reprSyntax._@._hyg.5 ``. Perhaps the `rawStr` should really be part of `SourceInfo.original` and `val : Name` the sole source of truth.

Vtec234: Okay, now I'm assuming the `eraseMacroScopes` representation if original info is present, otherwise using just the `Name`.

tests/lean/infoTree.lean.expected.out
```diff
@@ -9,7 +9,7 @@
     ===>
     Prod✝ Nat Nat
       Nat × Nat : Type @ ⟨13, 18⟩†-⟨13, 27⟩ @ Lean.Elab.Term.elabApp
-        Prod : Type → Type → Type @ ⟨13, 18⟩†-⟨13, 22⟩†
+        Prod : Type → Type → Type @ ⟨13, 18⟩†-⟨13, 27⟩†
```
Vtec234, line 12: The synthetic spans are now all of the original source span.

src/Lean/Syntax.lean
tydeu: @Vtec234 Isn't the point of French quotes to specifically get around any limitations the syntax might impose? So would that kind of defeat the point? Also, dots can easily appear when dealing with autogenerated syntax kind names. For example: `notation "...." => id` will have the syntax kind "term....".

Vtec234: Not any -- you cannot quote `»` :) But the dots notation is a good counterexample, and we shouldn't forbid it. Thanks.

tydeu: Yeah, I know. :) I meant to say "particularly for ASCII", but I forgot to. XD There is a part of me that wonders though why there is not a way to escape "»" -- as then it would match the semantics of the `Name` object itself (any string). Alternatively, since it can't be quoted I am surprised "»" is not used instead of the underscore for what are intended to be inaccessible names (ex. `_private` or `_hyg` could be `»private` or `»hygiene`).

Vtec234:
!bench

Kha:
Nice work!

src/Lean/Elab/App.lean
Kha: > Perhaps the `rawStr` should really only be part of `SourceInfo.original` and `val : Name` the sole source of truth in `ident`s.

This is an interesting idea. So it would simply be something like
```
  | original (leading token trailing : Substring)
```
We *might* even be able to replace `Syntax.atom`'s `val : String` with it since at least quotations [completely ignore it](https://github.com/leanprover/lean4/blob/07a4ff2d0fb3704888e2e3857ca563732395b4cc/src/Lean/Elab/Quotation.lean#L268-L269), though this might not be true for all parts of the elaborator (yet). /cc @leodemoura

PR #579, opened 2021-07-19
feat: show macro expansions

Vtec234:
Seeing the macro expansion is quite useful for writing macros and quotations, but it can be a bit scary. We may want to put it behind a `server.macro_expansions` option?

Alternatively, we may want never to show this in hovers and instead make a new `Macro expansion` infoview pane.


Kha:
> instead make a new `Macro expansion` infoview pane.

Yeah, I think that sounds better. There we could show the whole stack of macro expansions. It's also somewhat related to the idea of a dedicated, interactive trace messages pane.

Would be even cooler to have an action for unfolding the macro in-place (in the editor and/or info view), of course.

Vtec234:
Let's resuscitate this when the interactive features are coming in.

PR #580, opened 2021-07-20
[WIP] trying to improve RC

leodemoura:


leodemoura:
!bench

leodemoura:
TODO: remove obsolete CMAKE config options, and document new object header.

Kha:
`perf diff` for `qsort`:
```
               +4.17%  [.] lean_dec_ref_cold
     6.60%     +1.40%  [.] l_checkSortedAux
    19.47%     +0.56%  [.] l_qsortAux___at_main___spec__2
     4.46%     -0.35%  [.] lean_alloc_small
     2.57%     +0.19%  [.] lean_array_push
     0.14%     +0.10%  [.] lean::dealloc
    56.79%     -0.09%  [.] l___private_qsort_0__partitionAux___at_main___spec__3
     1.99%     -0.09%  [.] lean_free_small
     0.06%     -0.05%  [.] _int_free
     1.51%     +0.03%  [.] l_mkRandomArray
     0.04%     +0.03%  [.] l_Nat_forM_loop___at_main___spec__4
     0.06%     +0.02%  [k] 0xffffffffa1803e83
               +0.02%  [.] strcmp
     0.33%     -0.01%  [.] lean::alloc
     0.03%     -0.01%  [.] _int_malloc
     0.03%     -0.01%  [.] malloc
     0.01%     +0.01%  [.] cfree@GLIBC_2.2.5
     4.43%             [.] lean_del
     1.44%             [.] lean_copy_expand_array
     0.01%             [.] std::basic_ios<wchar_t, std::char_traits<wchar_t> >::init
     0.01%             [.] __memset_avx2_erms
     0.01%             [.] lean_mark_persistent
```
It's hard to say exactly what's happening here, and which part got slower (it doesn't help that `perf diff` only reports deltas in *relative* run time). `lean_dec_ref_cold` simply replaced `lean_del`. One interesting aspect is that `lean_copy_expand_array` apparently got small enough that it was inlined into `lean_array_push`, but it looks like that actually improved run times.

Kha:
Regarding the CI failures, it looks like signed atomic ops are not available in those stdc++ (Linux Release, GCC 7.4.0) / libc++ (macOS, LLVM 10) releases :/

leodemoura:
@Kha Not sure whether we should merge or not. The reduction in the binary size was nice. The fluctuation at `qsort` seems a bit random. What do you think?

leodemoura:
> Regarding the CI failures, it looks like signed atomic ops are not available in those libstdc++ (Linux Release, GCC 7.4.0) / libc++ (macOS, LLVM 10) releases :/

I will try to push a fix. The error occurs in a place where I am using both `int` and `unsigned`.

Kha:
Oh, that could be it! I'm fine with the changes otherwise, `stdlib` is still our most important benchmark after all.

PR #581, opened 2021-07-21
[WIP] feat: generalized monad transformers

tydeu:
A preliminary draft of the changes discussed in the Zulip thread [here](https://leanprover.zulipchat.com/#narrow/stream/147302-lean4-dev/topic/Leanpkg.2B.2B/near/246400813). Currently just covers the basic generalization. The next step would be to split `MonadExcept` and `Alternative` up in order to support `throw`/`failure` in more restricted contexts than `Monad`.

### DIFF
--- a/src/Init/Control/Except.lean
+++ b/src/Init/Control/Except.lean
@@ -25,6 +25,11 @@ variable {ε : Type u}
   intro e
   simp [Except.map]; cases e <;> rfl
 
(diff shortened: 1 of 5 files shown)


leodemoura:
As discussed in the Zulip channel, we will get back to this PR in the future. I will close it for now since it will probably take several months until we are ready for it.

PR #583, opened 2021-07-22
fix: respect preresolved names at `resolveConst*`

Kha:
This makes sure we can properly quote e.g. `deriving` clauses and avoids a suspicious `eraseMacroScopes` call (though not at `Elab.Syntax`, since categories do not have to be declaration names)

### DIFF
--- a/src/Lean/Elab/BuiltinTerm.lean
+++ b/src/Lean/Elab/BuiltinTerm.lean
@@ -163,9 +163,7 @@ def elabScientificLit : TermElab := fun stx expectedType? => do
   | none     => throwIllFormedSyntax
 
 @[builtinTermElab doubleQuotedName] def elabDoubleQuotedName : TermElab := fun stx _ => do
-  match stx[1].isNameLit? with
-  | some val => toExpr (← resolveGlobalConstNoOverloadCore val)  -- TODO: restore info
-  | none     => throwIllFormedSyntax
+  toExpr (← resolveGlobalConstNoOverloadWithInfo stx[2])
 
 @[builtinTermElab typeOf] def elabTypeOf : TermElab := fun stx _ => do
   inferType (← elabTerm stx[1] none)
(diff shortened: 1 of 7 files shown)


Kha:
@leodemoura While I think this is definitely the correct way to go, there's no hurry in reviewing or merging it

src/Lean/Elab/BuiltinTerm.lean
```diff
@@ -164,7 +164,7 @@ def elabScientificLit : TermElab := fun stx expectedType? => do
 
 @[builtinTermElab doubleQuotedName] def elabDoubleQuotedName : TermElab := fun stx _ => do
   match stx[1].isNameLit? with
-  | some val => toExpr (← resolveGlobalConstNoOverloadWithInfo stx[1] val)
+  | some val => toExpr (← resolveGlobalConstNoOverloadCore val)  -- TODO: restore info
```
Kha, line 167: @leodemoura I think (double) quoted names should not be literals but use the regular `ident` parser. Otherwise they will not work correctly when quoted since they currently lack the preresolved information. Is there any reason not to change their parser?

leodemoura: I would need a context switch to answer this question. Could you please add comments here explaining the issue? It would also be useful to open a new issue with examples that do not work.
I am assuming this PR can be merged independently of this issue. That is, it is not introducing any new issue.

Kha: With the current state of the PR we would lose go-to-definition on double-quoted names, so we might want to resolve the `ident` issue first. Opened https://github.com/leanprover/lean4/issues/586.

Kha:
Added a straightforward fix for #586 that changes the syntax of double-quoted names, but not single quoted ones, after noticing that we do use the latter with non-idents such as `` `_ ``.

PR #585, opened 2021-07-24
feat: FromToJson for recursive inductives

Vtec234:
Also adds a VSCode setting to trim trailing whitespace.


Kha:
These derive handlers sure can get complicated :sweat_smile:

Vtec234:
Yeah, it's on my backlog to factor out some of the patterns appearing in almost all of them like creating a `match` (with enough flexibility) to `Deriving.Util` or even `Meta.Basic`.

PR #587, opened 2021-07-27
Zero-cost .olean loading via `mmap`

Kha:
This PR implements relocation-free loading of compacted regions (e.g. .olean files) by choosing a "random" but deterministic base address when storing and (trying to) map the stored region at this address via `mmap` when loading. If mapping at the specified address fails, e.g. because the address is already occupied by another module or other kind of mapping, or the address is not even valid on the current platform, we fall back to the current approach of copying the region into a new buffer and rewriting all pointers in it. The approach has been inspired by the paper [Efficient Communication and Collection with Compact Normal Forms](http://ezyang.com/papers/ezyang15-cnf.pdf).

For the specific base address computation, see https://github.com/kha/lean4/blob/281a9cf42eb5089d2a23a6aee6b0099807ec98cc/src/library/module.cpp#L55-L68. The current formula, designed for and tested on x64-64, gives us 2^35 possible addresses 4KB apart (update: now 2^31 addresses 64KB apart to appease Windows). [According to Wikipedia](https://en.wikipedia.org/wiki/Birthday_problem#Probability_of_a_shared_birthday_(collision)), this means that we would need to try and load `sqrt(2*2^35*ln(1/(1-0.5))) ~= 218000` (update: now ~55000) modules for a 50% chance of even just a single collision, assuming the distribution is truly uniform, other mapped objects in the address space are negligible, and ignoring that modules can be bigger than 4KB. For comparison, with addresses 2MB apart, this number falls to ~9600 modules.

TODO:
- [x] see if macOS CI likes it too
- [x] implement on Windows
- [x] move `terminator_object` to the start of each region to make it truly zero-cost
- [x] adjust build system: mmap-ed .olean files may not be modified, so we should always create new files and `mv` them over any previous version
- [x] add statistics about collisions


Kha:
!bench

Kha:
!bench

Kha:
>     * add statistics about collisions

```
$ lean --stats src/Lean.lean
...
number of imported modules:            403
number of memory-mapped modules:       403
```

Kha:
> ```diff
> + stdlib      maxrss                    -2%    (-41.9 σ)
> ```

Note that RSS (Resident Set Size) does include all memory mappings, evem though those pages can be shared between processes (and swapped out without writing them to swap), so no significant improvement was expected for this measure. `massif` says that the maximum heap size of `lean src/Lean.lean` fell from ~270MB to ~30MB.

Kha:
!bench

Kha:
mathport with `master`:
```
import took 2.3s
direct imports:                        #[Init, Mathlib.All]
number of imported modules:            1869
number of consts:                      273026
number of imported consts:             273026
number of local consts:                0
number of buckets for imported consts: 524288
trust level:                           0
number of extensions:                  52
```
Peak heap size: 1.2GB

---

mathport with this PR:
```
import took 1.45s
direct imports:                        #[Init, Mathlib.All]
number of imported modules:            1871
number of memory-mapped modules:       1871
number of consts:                      273387
number of imported consts:             273387
number of local consts:                0
number of buckets for imported consts: 273387
trust level:                           0
number of extensions:                  52
```
Peak heap size: 55.8MB :exclamation:

gebner:
Sweet!

> import took 2.3s
> import took 1.45s

Do you have an idea what's still taking up time?  Is this all just the initializers?

Kha:
>  Is this all just the initializers?

It's the setup ("finalization") of environment extensions (especially after optimizing the setup of `Environment.constants` with the previous commit). For the stdlib, it's mostly compiler-related extensions. For current mathport, it's mostly `[simp]`: https://profiler.firefox.com/public/bq3fmsr9c9w1ht0f20m35k639g39yj0vv4zrh3g/flame-graph/?globalTrackOrder=0&thread=0&v=6

gebner:
> For current mathport, it's mostly [simp]:

Ah, this is interesting.  Luckily there seem to be lots of low-hanging fruit there (insertAux is accidentally quadratic and compares the proof terms).

src/Lean/Environment.lean
```diff
@@ -616,10 +619,12 @@ partial def importModules (imports : List Import) (opts : Options) (trustLevel :
...
     let (_, s) ← importMods imports |>.run {}
-    -- (moduleNames, mods, regions)
+    let mut numConsts := 0
+    for mod in s.moduleData do
+      numConsts := numConsts + mod.constants.size
     let mut modIdx : Nat := 0
-    let mut const2ModIdx : HashMap Name ModuleIdx := {}
-    let mut constants : ConstMap := SMap.empty
+    let mut const2ModIdx : HashMap Name ModuleIdx := Std.mkHashMap (nbuckets := numConsts)
+    let mut constants : ConstMap := SMap.empty (nbuckets₁ := numConsts)
```
Kha, line 627: @leodemoura This was a low-hanging fruit for significantly speeding up this part of importing by removing the need for resizing the hashmaps. I want to try further simple optimizations of the hashmap such as enforcing a "load factor" (usually 0.75), but not in this PR.

leodemoura:
@Kha Could you please test again this PR with the new `Lib4` https://github.com/dselsam/mathport/releases/tag/v0.0.0.0?
@dselsam generated it using the commit above b25bb78.
The import time for the whole mathlib went from 1s to 0.5s on my machine using the master branch.

leodemoura:
@Kha I pushed additional perf improvements to `importModules`. It now takes 0.38s to load Mathlib on my machine.
```lean
~/projects/Lib4 $ LEAN_PATH=. time lean all.lean
0.38user 0.35system 0:00.73elapsed 99%CPU (0avgtext+0avgdata 1245872maxresident)k
0inputs+0outputs (0major+305951minor)pagefaults 0swaps
```

Kha:
It's now 0.45s for me, i.e. 1s faster :tada:

src/library/module.cpp
```diff
@@ -45,9 +45,11 @@ static char const * g_olean_header   = "oleanfile!!!!!!!";
 
 extern "C" object * lean_save_module_data(b_obj_arg fname, b_obj_arg mod, b_obj_arg mdata, object *) {
     std::string olean_fn(string_cstr(fname));
+    // we first write to a temp file and then move it to the correct path (possibly deleting an older file)
+    // so that we neither expose partially-written files nor modify possibly memory-mapped files
```
Kha, line 48: @leodemoura @tydeu Correct replacement of .olean files is now handled right inside Lean so that build systems do not have to worry about it. This should also obviate the need for file locking.

Kha: TBD: replacement on Windows, where memory-mapped files cannot be deleted...

src/library/module.cpp
```diff
@@ -37,20 +44,56 @@ namespace lean {
...
+        object_compactor compactor(reinterpret_cast<void *>(base_addr + strlen(g_olean_header) + sizeof(base_addr)));
+        compactor(mdata);
         out.write(g_olean_header, strlen(g_olean_header));
+        out.write(reinterpret_cast<char *>(&base_addr), sizeof(base_addr));
         out.write(static_cast<char const *>(compactor.data()), compactor.size());
         out.close();
+        while (std::rename(olean_tmp_fn.c_str(), olean_fn.c_str()) != 0) {
+#ifdef LEAN_WINDOWS
+            if (errno == EEXIST) {
+                // Memory-mapped files can be deleted starting with Windows 10, but *only* through the UTF-16 API
```
Kha, line 83: https://news.ycombinator.com/item?id=23745019 :exploding_head:

Kha:
Ready for lift-off.

PR #592, opened 2021-07-30
RPC refactor and `Std.ToFormat` housekeeping

Vtec234:
- This PR introduces some RPC changes, notably [`MonadRpcSession`](https://github.com/leanprover/lean4/compare/master...Vtec234:rpc-format?expand=1#diff-426776cbaf3cb9c32ef0e0eef4c86076d8cc8a8223065d605ecc806845b6499bR21) to break a dependency loop; and eager session initialization to enable sending refs in diagnostics.
- I also noticed that both `ToFormat.format` and `fmt` are exported and both are used semi-randomly, so here the latter is removed and only `format` is exported. I propose to disambiguate typeclass-`format` from `format`-being-implemented using namespaces instead, like `Foo.format (f: Foo) := Std.format f.field`. If this is not desirable, we could consistently use `fmt` and stop exporting `format` instead.

It also used to:
- (Postponed, see discussion.) Generalize `Format` tags from only allowing `Nat`. I am using a generic `EFormat` a lot in the server and agree with [Ed's comment](https://github.com/leanprover-community/lean/pull/276#issuecomment-634071755) that it is useful, in particular more convenient than the roughly isomorphic `Format x (HashMap Nat TagType)` where `Format` is `Nat`-tagged. To avoid duplication here I change the "real" type. The existing interface is mostly preserved with some tuning of optional params, ~though there are a few places where the type needs to be spelled out~.

### DIFF
--- a/src/Init/Data/Format/Basic.lean
+++ b/src/Init/Data/Format/Basic.lean
@@ -17,20 +17,19 @@ inductive Format.FlattenBehavior where
 
 open Format
 
-inductive Format (τ := Nat) where
-  | nil {}              : Format τ
-  | line {}             : Format τ
-  | text                : String → Format τ
-  | nest (indent : Int) : Format τ → Format τ
-  | append              : Format τ → Format τ → Format τ
-  | group               : Format τ → (behavior : FlattenBehavior := FlattenBehavior.allOrNone) → Format τ
+inductive Format where
+  | nil                 : Format
+  | line                : Format
+  | text                : String → Format
+  | nest (indent : Int) : Format → Format
+  | append              : Format → Format → Format
+  | group               : Format → (behavior : FlattenBehavior := FlattenBehavior.allOrNone) → Format
   /- Used for associating auxiliary information (e.g. `Expr`s) with `Format` objects. -/
-  | tag                  : τ → Format τ → Format τ
+  | tag                 : Nat → Format → Format
   deriving Inhabited
 
 namespace Format
-
-def fill (f : Format τ) : Format τ :=
+def fill (f : Format) : Format :=
   group f (behavior := FlattenBehavior.fill)
 
 @[export lean_format_append]
@@ -41,15 +40,15 @@ protected def appendEx (a b : Format) : Format :=
 protected def groupEx : Format → Format :=
   group
 
-instance : Append (Format τ)     := ⟨Format.append⟩
-instance : Coe String (Format τ) := ⟨Format.text⟩
+instance : Append Format     := ⟨Format.append⟩
+instance : Coe String Format := ⟨text⟩
 
-def join (xs : List (Format τ)) : Format τ :=
+def join (xs : List Format) : Format :=
   xs.foldl (·++·) ""
 
-def isNil : Format τ → Bool
-  | nil _ => true
-  | _     => false
+def isNil : Format → Bool
+  | nil => true
+  | _   => false
 
 private structure SpaceResult where
   foundLine              : Bool := false
@@ -64,9 +63,9 @@ private structure SpaceResult where
     let r₂ := r₂ (w - r₁.space);
     { r₂ with space := r₁.space + r₂.space }
 
-private def spaceUptoLine : Format τ → Bool → Nat → SpaceResult
-  | nil _,        flatten, w => {}
-  | line _,       flatten, w => if flatten then { space := 1 } else { foundLine := true }
+private def spaceUptoLine : Format → Bool → Nat → SpaceResult
+  | nil,          flatten, w => {}
+  | line,         flatten, w => if flatten then { space := 1 } else { foundLine := true }
   | text s,       flatten, w =>
     let p := s.posOf '\n';
     let off := s.offsetOfPos p;
@@ -76,16 +75,16 @@ private def spaceUptoLine : Format τ → Bool → Nat → SpaceResult
   | group f _,    _,       w => spaceUptoLine f true w
   | tag _ f,      flatten, w => spaceUptoLine f flatten w
 
-private structure WorkItem (τ) where
-  f : Format τ
+private structure WorkItem where
+  f : Format
   indent : Int
 
-private structure WorkGroup (τ) where
+private structure WorkGroup where
   flatten : Bool
   flb     : FlattenBehavior
-  items   : List (WorkItem τ)
+  items   : List WorkItem
 
-private partial def spaceUptoLine' : List (WorkGroup τ) → Nat → SpaceResult
+private partial def spaceUptoLine' : List WorkGroup → Nat → SpaceResult
   |   [],                         w => {}
   |   { items := [],    .. }::gs, w => spaceUptoLine' gs w
   | g@{ items := i::is, .. }::gs, w => merge w (spaceUptoLine i.f g.flatten w) (spaceUptoLine' ({ g with items := is }::gs))
@@ -94,10 +93,10 @@ private structure State where
   out    : String := ""
   column : Nat    := 0
 
-private def pushGroup (flb : FlattenBehavior) (items : List (WorkItem τ)) (gs : List (WorkGroup τ)) (w : Nat) : StateM State (List (WorkGroup τ)) := do
+private def pushGroup (flb : FlattenBehavior) (items : List WorkItem) (gs : List WorkGroup) (w : Nat) : StateM State (List WorkGroup) := do
   let k  := (← get).column
   -- Flatten group if it + the remainder (gs) fits in the remaining space. For `fill`, measure only up to the next (ungrouped) line break.
-  let g  := { flatten := flb == FlattenBehavior.allOrNone, flb := flb, items := items : WorkGroup τ }
+  let g  := { flatten := flb == FlattenBehavior.allOrNone, flb := flb, items := items : WorkGroup }
   let r  := spaceUptoLine' [g] (w-k)
   let r' := merge (w-k) r (spaceUptoLine' gs);
   -- Prevent flattening if any item contains a hard line break, except within `fill` if it is ungrouped (=> unflattened)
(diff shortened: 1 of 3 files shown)


Vtec234:
The current default tag type is `Empty`, which means `Format` is tag-less by default. In general we can functorially map `Format a`, but with `Empty` the following special case arises:
```lean
def Format.generalize (α) : Format Empty → Format α
  | nil _ => nil _
  | line _ => line _
  | text s => text s
  | nest indent a => nest indent (a.generalize α)
  | append a₁ a₂ => append (a₁.generalize α) (a₂.generalize α)
  | group a b => group (a.generalize α) b
  | tag t a => nomatch t
```
We can see that morally, if the constructor representations are equivalent, this function does nothing. However `trace.compiler.ir` seems to indicate that a whole traversal is emitted. One cannot just prove something like `fmt.generalize a = id fmt`, but I am wondering if there is a way to make this a no-op, or the compiler smart enough to do so. If not, the default should probably be `Format Nat` as it is now to avoid these `generalize` casts in the pretty printer.

Kha:
> but I am wondering if there is a way to make this a no-op, or the compiler smart enough to do so.

Currently there's `unsafeCast` only, I believe. In it's most general form, it would be great to have something like Haskell's [`coerce`](https://hackage.haskell.org/package/base-prelude-1.4/docs/BasePrelude.html#v:coerce) so that we can express that e.g. an array of a coercible type can itself be coerced in constant time.

src/Init/Data/Format/Basic.lean
```diff
@@ -163,60 +164,84 @@ private partial def be (w : Nat) : List WorkGroup → StateM State Unit
...
   st.out
 
-end Format
+@[export lean_format_pretty]
+protected def prettyEx (f : Format) (w : Nat := defWidth) : String :=
+  pretty f w
+
+def map (f : α → β) (fmt : Format α) : Format β :=
+  go fmt
+  where go : Format α → Format β
```
Kha, line 199: `where` is usually dedented

src/Init/Data/Format/Basic.lean
```diff
@@ -17,19 +17,20 @@ inductive Format.FlattenBehavior where
...
-inductive Format where
-  | nil                 : Format
-  | line                : Format
-  | text                : String → Format
-  | nest (indent : Int) : Format → Format
-  | append              : Format → Format → Format
-  | group               : Format → (behavior : FlattenBehavior := FlattenBehavior.allOrNone) → Format
+inductive Format (τ := Nat) where
+  | nil {}              : Format τ
+  | line {}             : Format τ
```
Kha, line 22: Why the `{}`? It looks like in (almost?) all cases `τ` can be inferred.

src/Lean/Server/FileWorker/Utils.lean
```diff
@@ -67,19 +69,40 @@ structure RpcSessionState where
...
   let ref := ⟨st.nextRef⟩
   let st' := { st with aliveRefs := st.aliveRefs.insert ref (typeName, obj)
                        nextRef := st.nextRef + 1 }
   (ref, st')
 
-def release (ref : Lsp.RpcRef) (st : RpcSessionState) : RpcSessionState :=
-  { st with aliveRefs := st.aliveRefs.erase ref }
+def release (st : RpcSessionState) (ref : Lsp.RpcRef) : Bool × RpcSessionState :=
+  let st' := { st with aliveRefs := st.aliveRefs.erase ref }
+  (st.aliveRefs.contains ref, st')
```
Kha, line 80:
```suggestion
  let released := st.aliveRefs.contains ref
  (released, { st with aliveRefs := st.aliveRefs.erase ref })
```
It shouldn't matter much for performance, but I find this order easier to read.

Kha (approved):
>     * I also noticed that both `ToFormat.format` and `fmt` are exported and both are used semi-randomly, so here the latter is removed and only `format` is exported. I propose to disambiguate typeclass-`format` from `format`-being-implemented using namespaces instead, like `Foo.format (f: Foo) := Std.format f.field`.

Fine with me. The motivation for `fmt` might have been stronger when we did not have format strings. /cc @leodemoura

src/Init/Data/Format/Basic.lean
Vtec234: There are quite a few occurrences where it fails due to instances like `Coe Format MessageData`. Resolution will not assign the MVar to elab `Format.nil` as `coe (Format.nil : Format Nat) : MessageData`.

src/Init/Data/Format/Basic.lean
```diff
@@ -163,60 +164,84 @@ private partial def be (w : Nat) : List WorkGroup → StateM State Unit
...
   let (_, st) := be w [{ flb := FlattenBehavior.allOrNone, flatten := false, items := [{ f := f, indent := 0 }] }] {};
   st.out
 
-end Format
+@[export lean_format_pretty]
+protected def prettyEx (f : Format) (w : Nat := defWidth) : String :=
+  pretty f w
+
+def map (f : α → β) (fmt : Format α) : Format β :=
+  go fmt
```
Vtec234, line 198:
```suggestion
    go fmt
```

src/Init/Data/Format/Basic.lean
```diff
@@ -163,60 +164,84 @@ private partial def be (w : Nat) : List WorkGroup → StateM State Unit
...
+    | nil _        => nil _
+    | line _       => line _
+    | text s       => text s
+    | nest i a     => nest i (go a)
+    | append a₁ a₂ => append (go a₁) (go a₂)
+    | group a b    => group (go a) b
+    | tag t a      => tag (f t) (go a)
+
+def mapM [Monad m] (f : α → m β) (fmt : Format α) : m (Format β) :=
+  go fmt
```
Vtec234, line 209:
```suggestion
    go fmt
```

src/Init/Data/Format/Basic.lean
Kha: Ah, the `Syntax arity` issue again :)

src/Init/Data/Format/Basic.lean
Kha: Ah, no, the `where` should literally be on the same level of `def` :smile:

gebner:
> I am using a generic EFormat a lot in the server [...]

In Lean 3 putting the sub-expression into the format object was the only realistic choice.  The pretty-printer went directly from expr to format.  Adding an extra generic tag to the format object was the least-effort way to propagate this information.  If this information was already available via a separate hash map, we would've just used that.

The pretty-printer is much more modular right now, so maybe it's worth revisiting these (accidental) design decisions.  Some questions (partly based on the `SubexprMap`/`ExprWithCtx` types in your rpc-widgets branch):

 - How do the generic tags generalize to other categories?  For example, when pretty-printing a command like `example : Nat := 42  #eval 1`, what would the type parameter for Format be?  Obviously you should be able to figure out the subspans of the commands, etc.  (This is not a purely theoretical question, `#print foo` might want to print a command syntax that you should be able to interact with.)
 - Could we actually use more than one tag type in practice?  Let's say expressions are pretty-printed as `Format ExprWithCtx` and commands are pretty-printed as `Format CommandWithCtx`, what tag type would a message containing both an expression and a command have?

Maybe I'm also confused about the intended use case.  Or is everything in Lean supposed to use `Format Nat`, and the only the widget code does a final `Format.map` at the end?

I think the extra indirection with HashMap (i.e., storing the address in the format object) could even turn out be useful, because it allows you to modify the expression or syntax object.  (Applying simp to a subexpression, or unfolding a macro, etc.)

Vtec234:
> How do the generic tags generalize to other categories?

In the generic variant we'd use `Format.map` to inject into a more general tag type. For example, from an `Expr` we might obtain a `Format ExprWithCtx` (more or less, actually it needs to be a lazy thunk to avoid eager traversals before the widget-user actually clicks anything). Then we can format a `MessageData` containing such an `Expr` in `MessageData.ofExpr`, and get a `Format MessageEmbed` with `inductive MessageEmbed | ofExpr (e : ExprWithCtx) | ofOtherStuffThatMessagesContain (..)`. (In my current code there is actually only one `MessageEmbed`, the `ofExpr`, but at least for `MessageData.ofGoal` we will likely want something special.)

> #print foo might want to print a command syntax that you should be able to interact with

It already does, with `MessageEmbed` :) One thing that does not work interactively is `trace.Elab.command` outputs, but these are `Syntax` which may not even be well-formed, so I'm not sure what to do there.

> what tag type would a message containing both an expression and a command have?

I think the first answer explains it.

> Maybe I'm also confused about the intended use case. Or is everything in Lean supposed to use Format Nat, and the only the widget code does a final Format.map at the end? I think the extra indirection with HashMap (i.e., storing the address in the format object) could even turn out be useful, because it allows you to modify the expression or syntax object. (Applying simp to a subexpression, or unfolding a macro, etc.)

Great question. I was already on the fence about this, and now having tried both out some more, I'm starting to become convinced that the indirect approach is better. Let's consider some properties we may want:
- Typing discipline to remember what the tags mean. Clearly `Format t` does this, but to an extent so does `MapFormat t := Format Nat x HashMap Nat t` with a sufficiently encapsulated interface.
- The ability to embed more specifically-tagged formats into more general ones, as in the `MessageData` example. With `Format t` we need to map the tree by traversing it (as even `coerce` wouldn't work given the actual change in runtime representation). With `MapFormat t` we are able to avoid that by having the top-level `HashMap` refer to lower-level ones while keeping the same `Format Nat` object along the way, but one s.t. the *meaning* of the Nat-tags changes as the `HashMap`s change. This is probably the most error prone part and the reason for typing the tags themselves, however it is not terribly bad.
- Since `Format t`s are sent to the client eagerly for messages and goals, we want `t` to be easy to compute, so it can't be `ExprWithCtx` directly but must rather be `Thunk ExprWithCtx`. `HashMap Nat ExprWithCtx` currently has the same problem (although eventually the delaborator will output it in the traversal that it has to do anyway) but a more general `SubexprMap := Nat -> Option ExprWithCtx` gets the laziness for free.
- And yes, a `MapFormat t` gives us more direct access to the embedded information.

Vtec234:
Okay, since it is no longer clear to me that we definitely want `Format a`, I restored the current version. I kept the `format`/`fmt` housekeeping since that seems cleaner in any case. Let's get the RPC stuff merged, and I will either re-introduce `Format a` or one of the other discussed solutions in a PR which actually uses it so it is clear to everyone what the requirements are and whether they're met.

PR #575, opened 2021-07-15
feat: pp.analyze heuristics for concise roundtripping

dselsam:
This PR adds an (on-by-default) flag `pp.analyze` which, when enabled, is called at the start of `delab` to produce an `OptionsPerPos` map that enables `delab` to produce syntax that (a) can be re-elaborated correctly and (b) concise. The analysis includes many heuristics, and is both unsound (not guaranteed to re-elaborate) and incomplete (syntax not guaranteed to be minimal). However, it is about as good as `pp.all` for roundtripping and usually much more concise. It also fixes a (rare) exponential blow-up in the delaborator (https://github.com/leanprover/lean4/blob/master/tests/lean/delabUnexpand.lean#L25), and also improves the default name delaboration to avoid using names that are strongly shadowed.

PR #596, opened 2021-08-03
Interactive goals & diagnostics

Vtec234:
Implements server-side support for interactive goals and diagnostics. ~~Currently the only supported mode of interaction is exploring `Expr`s (we can check the types of subexpressions and explicit arguments). We add a new `taggedMsg` field to the `Diagnostic` type which clients can ignore, while the infoview uses it to visualize things.~~

We support the following interactive objects:
- "code"; currently this is only ever a pretty-printed `Expr` but other objects such as general `Syntax` will also work, as long as they have associated `Elab.Info`
- goals; we can interact with goals and the "code" embedded in the hypothesis/goal types as well as `let`-bound values
- traces; messages with trace annotations can be explored interactively; they are pretty-printed lazily to make massive traces reasonably performant
- general messages; can contain code, goals, and traces

TODOs:
- [x] use `InfoTree`s instead of `Expr`s directly (depends at least on #575)
- [x] fix memory leaks in the RPC protocol
- [x] final code cleanup

### DIFF
--- a/src/Lean/Message.lean
+++ b/src/Lean/Message.lean
@@ -117,6 +117,8 @@ partial def formatAux : NamingContext → Option MessageDataContext → MessageD
   | nCtx, _,         withContext ctx d        => formatAux nCtx ctx d
   | _,    ctx,       withNamingContext nCtx d => formatAux nCtx ctx d
   | nCtx, ctx,       tagged t d               =>
+    /- Messages starting a trace context have their tags postfixed with `_traceCtx` so that
+    we can detect them later. Here, we do so in order to print the trace context class. -/
     if let Name.str cls "_traceCtx" _ := t then do
       let d₁ ← formatAux nCtx ctx d
       return f!"[{cls}] {d₁}"
--- a/src/Lean/Meta/PPGoal.lean
+++ b/src/Lean/Meta/PPGoal.lean
@@ -160,9 +160,9 @@ def ppGoal (mvarId : MVarId) : MetaM Format := do
           pure fmt
         else
           let fmt := addLine fmt
-          match ids, type? with
-          | _, none      => pure fmt
-          | _, some type => do
+          match type? with
+          | none      => pure fmt
+          | some type => do
             let typeFmt ← ppExpr type
             pure $ fmt ++ (Format.joinSep ids.reverse (format " ") ++ " :" ++ Format.nest indent (Format.line ++ typeFmt)).group
       let rec ppVars (varNames : List Name) (prevType? : Option Expr) (fmt : Format) (localDecl : LocalDecl) : MetaM (List Name × Option Expr × Format) := do
--- a/src/Lean/PrettyPrinter/Delaborator/Basic.lean
+++ b/src/Lean/PrettyPrinter/Delaborator/Basic.lean
@@ -43,11 +43,15 @@ structure Context where
   optionsPerPos  : OptionsPerPos
   currNamespace  : Name
   openDecls      : List OpenDecl
-  inPattern      : Bool := false -- true whe delaborating `match` patterns
+  inPattern      : Bool := false -- true when delaborating `match` patterns
   subExpr        : SubExpr
 
 structure State where
+  /-- We attach `Elab.Info` at various locations in the `Syntax` output in order to convey
+  its semantics. While the elaborator emits `InfoTree`s, here we have no real text location tree
+  to traverse, so we use a flattened map. -/
   infos    : Std.RBMap Pos Info compare := {}
+  /-- See `SubExpr.nextExtraPos`. -/
   holeIter : SubExpr.HoleIterator := {}
 
 -- Exceptions from delaborators are not expected. We use an internal exception to signal whether
--- a/src/Lean/PrettyPrinter/Delaborator/SubExpr.lean
+++ b/src/Lean/PrettyPrinter/Delaborator/SubExpr.lean
@@ -7,13 +7,12 @@ import Lean.Meta.Basic
 import Std.Data.RBMap
 
 /-!
-This file defines utilities for `MetaM` computations to traverse subexpressions
-of an expression in sync with the `Nat` "position" values that refers to them.
-We use a simple encoding scheme: every `Expr` constructor has at most 3 direct
-expression children. Considering an expressions type as well, we can injectively
-map a path of `childIdxs` to a natural number by computing the value of the 4-ary
-representation `1 :: childIdxs`, since n-ary representations without leading zeros
-are unique. Note that `pos` is initialized to `1` (case `childIdxs == []`).
+This file defines utilities for `MetaM` computations to traverse subexpressions of an expression
+in sync with the `Nat` "position" values that refer to them.  We use a simple encoding scheme:
+every `Expr` constructor has at most 3 direct expression children. Considering an expression's type
+to be one extra child as well, we can injectively map a path of `childIdxs` to a natural number
+by computing the value of the 4-ary representation `1 :: childIdxs`, since n-ary representations
+without leading zeros are unique. Note that `pos` is initialized to `1` (case `childIdxs == []`).
 -/
 
 namespace Lean.PrettyPrinter.Delaborator
@@ -122,8 +121,9 @@ def HoleIterator.next (iter : HoleIterator) : HoleIterator :=
     ⟨2*iter.top, maxChildren*iter.top⟩
   else ⟨iter.curr+1, iter.top⟩
 
-/-- The positioning scheme guarantees that there will be an infinite number of positions
-that are never used by subexpressions. We use these to attach additional `Elab.Info`. -/
+/-- The positioning scheme guarantees that there will be an infinite number of extra positions
+which are never used by `Expr`s. The `HoleIterator` always points at the next such "hole".
+We use these to attach additional `Elab.Info`. -/
 def nextExtraPos : m Pos := do
   let iter ← getThe HoleIterator
   let pos := iter.toPos


src/Init/Data/Format/Basic.lean
```diff
@@ -89,36 +88,35 @@ private partial def spaceUptoLine' : List WorkGroup → Nat → SpaceResult
...
+private partial def be (w : Nat) [Monad m] [MonadPrettyFormat m] : List WorkGroup → m Unit
   | []                           => pure ()
   |   { items := [],    .. }::gs => be w gs
   | g@{ items := i::is, .. }::gs => do
     let gs' (is' : List WorkItem) := { g with items := is' }::gs;
     match i.f with
     | nil => be w (gs' is)
-    | tag _ f => be w (gs' ({ i with f }::is))
+    | tag t f =>
+      withTag t <| be w [{ g with items := [{ i with f }] }]
```
Vtec234, line 116: I believe this prevents `be` from being tail-recursive. However I observed no performance degradation due to this with a fairly long `#synth` trace taken from binport `Lib4`. If this is a worry we could consider specializing `be`. cc @leodemoura

src/Lean/Widget/InteractiveGoals.lean
```diff
@@ -0,0 +1,137 @@
...
+Copyright (c) 2021 Microsoft Corporation. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+
+Authors: Wojciech Nawrocki
+-/
+import Lean.Widget.ExprWithCtx
+
+/-! RPC procedures for retrieving tactic and term goals with embedded `ExprText`s. -/
+-- TODO(WN): this module is mostly a slightly adapted copy of the corresponding `plainGoal/plainTemGoal` handlers
+-- unify them somehow? or delete the plain ones.
```
Vtec234, line 11: Any thoughts? I guess we want to keep `plainGoal` for infoview-less clients?

src/Lean/Widget/TaggedText.lean
```diff
@@ -0,0 +1,127 @@
...
+  | text (s : String)
+  /- Invariant: non-empty and never directly nested. `append #[tag _ append #[]]` is okay. -/
+  | append (as : Array (TaggedText α))
+  | tag (t : α) (a : TaggedText α)
+  deriving Inhabited, BEq
+
+namespace TaggedText
+
+partial def fromJson? [FromJson α] (j : Json) : Except String (TaggedText α) :=
+  (do
```
gebner, line 26: Is there a reason `deriving FromJson` doesn't work?

If we're doing a custom serializer, we could also pick a more compact representation like:
```typescript
type TaggedText<A> = TaggedText<A>[] | string | {tag: A, text: TaggedText<A>}
```

src/Lean/Widget/InteractiveDiagnostics.lean
```diff
@@ -0,0 +1,131 @@
...
+import Lean.Widget.TaggedText
+import Lean.Widget.ExprWithCtx
+
+namespace Lean.Widget
+open Lsp Server
+
+structure MsgEmbed where
+  -- TODO(WN): other variants, e.g. `ofGoal`
+  ef : ExprText
+  deriving Inhabited, RpcEncoding
```
gebner, line 19: We might want to pick a more descriptive name than `ef`, since this is part of the public RPC API.

src/Lean/Widget/InteractiveGoals.lean
```diff
@@ -0,0 +1,137 @@
...
+
+namespace Lean.Widget
+open Server
+
+structure InteractiveGoal where
+  hyps      : Array (String × ExprText)
```
gebner, line 17: Just so that I understand this correctly: when the client gets an interactive goal, it also gets ownership of all the `LazyExprWithCtx` references in the goal.  These references need to be released every time the goal is refreshed.  Isn't that a lot of requests?

src/Lean/Widget/InteractiveDiagnostics.lean
```diff
@@ -0,0 +1,131 @@
...
+      let endPos := if endPos.line > m.pos.line then { line := m.pos.line + 1, column := 0 } else endPos
+      text.leanPosToLspPos endPos
+    | none        => low
+  let range : Range := ⟨low, high⟩
+  let fullRange : Range := ⟨low, fullHigh⟩
+  let severity := match m.severity with
+    | MessageSeverity.information => DiagnosticSeverity.information
+    | MessageSeverity.warning     => DiagnosticSeverity.warning
+    | MessageSeverity.error       => DiagnosticSeverity.error
+  let source := "Lean 4 server"
```
gebner, line 108:
```suggestion
  let source := "Lean 4"
```
We have been patching this in the vscode extension for quite some while now.

gebner (commented):
> We add a new taggedMsg field to the Diagnostic type which clients can ignore [...]

Is this true?  Don't you get a memory leak if you don't release the RPC references?  (Also, diagnostics are updated really often.  This will cause a lot of release requests.)

(There's also the question of who owns these references in the vscode extension, i.e., who is responsible for releasing them.  If it is the infoview, what happens if it is closed?)

src/Init/Data/Format/Basic.lean
leodemoura: @Vtec234 I am not worried about performance degradation, but potential stack overflows.

src/Lean/Widget/ExprWithCtx.lean
```diff
@@ -0,0 +1,121 @@
...
+  env           : Environment
+  mctx          : MetavarContext
+  lctx          : LocalContext
+  options       : Options
+  currNamespace : Name
+  openDecls     : List OpenDecl
+  deriving Inhabited, RpcEncoding with { withRef := true }
+
+/-- We traverse expressions lazily when the client requests it. -/
+abbrev LazyExprWithCtx := Unit → IO ExprWithCtx
```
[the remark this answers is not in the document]
Vtec234: Thunks are pure, right? So it'd have to be `Thunk (IO ExprWithCtx)` which I guess caches the action but not its result, which isn't very useful. I'm not aware of a `ThunkIO`. `Task` would work but it's kind of weird to thread-pool this. As for using `LazyExprWithCtx` everywhere, I guess that's reasonable enough. I'll try it.

src/Lean/Widget/TaggedText.lean
Vtec234: `FromJson` can't be derived because of the `Array (TaggedText α)`, where I had to introduce a local recursive instance of `FromJson (TaggedText α)`. Currently `deriving FromJson` will write `def instFromJsonFoo` and then call that on recursive fields, but will not introduce a local instance. For this to work for `Array` we'd need to look for `Traversable` or something. Though I think we should have `Traversable` anyway, the introduce-local-instance approach avoids this. I can fix up the `deriving` handler to do it.

As for a more compact representation, I did try it but observed no performance upgrade (we are bottlenecked on the `TaggedText`-producing pretty-printing) so I stuck with the same one that `FromJson` would emit.

src/Lean/Widget/InteractiveGoals.lean
Vtec234: It's a lot of requests and I have an outstanding TODO to batch them into a `Release where refs : Array RpcRef`. I'll add that.

src/Lean/Widget/ExprWithCtx.lean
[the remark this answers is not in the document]
gebner: There's `unsafeIO`, so you can write a very reasonable `IO α → IO (Thunk (Except IO.Error α))` function.

Vtec234:
> Don't you get a memory leak if you don't release the RPC references? 

This is correct. We can have a `lean.hasInfoview` client capability s.t. refs are only sent if the capability is set.

src/Lean/Widget/ExprWithCtx.lean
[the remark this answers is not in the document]
leodemoura: I did not carefully look at this PR yet, but please try to avoid `unsafeIO` as much as possible. We had many issues with it in the past, the current implementation uses it in a single place: `profileitIO`, and we should probably remove it in the future. In the past, we used `unsafeIO` in other places, and if I remember correctly it "blew up" in our faces more than once. It is easy to write incorrect code when using `unsafeIO`. BTW, if I remember correctly, we had problems with a very similar function
```
constant asTask (act : IO α) (prio := Task.Priority.default) : IO (Task (Except IO.Error α))
```
Note that it is not marked as `unsafe`, but `act` is always executed. I think the same justification does not work for `IO α → IO (Thunk (Except IO.Error α))` since it would not make sense to run the IO action eagerly in this case. So, I am assuming the function `IO α → IO (Thunk (Except IO.Error α))` would have to be marked as `unsafe`.
The following link has a discussion of a similar issue in Haskell. https://mail.haskell.org/pipermail/haskell/2009-March/021064.html
@Kha Do you remember the details? I recall discussing these issues with you but I forgot the details.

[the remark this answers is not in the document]
gebner: This is interesting, I didn't realize that nondeterministic execution fell under the unsafe umbrella as well.

I still think it would still be a good idea to avoid the double references (plus the extra server roundtrip for the `LazyExprWithCtx.get` call).  A `ThunkIO` implementation that requires `IO` to access the thunk should still be fine, right?
```lean
def ThunkIO (α) := IO.Ref (Sum (IO α) α)

def ThunkIO.mk (a : IO α) : IO (ThunkIO α) := IO.mkRef (Sum.inl a)

def ThunkIO.get (t : ThunkIO α) : IO α := do
  match ← ST.Ref.get t with
    | Sum.inr a => a
    | Sum.inl fn =>
      let a ← fn -- potential double evaluation, but avoids locking
      t.set (Sum.inl a)
      a
```

gebner:
> This is correct. We can have a lean.hasInfoview client capability s.t. refs are only sent if the capability is set.

What happens if the infoview is closed?

src/Init/Data/Format/Basic.lean
Vtec234: Fixed. I had to add an additional fake tag stack size to `WorkItem` but it is tail-recursive again. The default `MonadPrettyFormat` does nothing on tags so should hopefully specialize away.

src/Lean/Widget/TaggedText.lean
Vtec234: I fixed the handler in #601 and will rebase on that, removing the manual encoding. I agree that our serialization scheme is quite inefficient (it is JSON...), but it's also hardly the bottleneck. For the sizes of messages where it would become problematic, it is far trumped by the message production time, i.e. things like delaboration and formatting. I think if we were to fix this while sticking to JSON rather than moving to e.g. Cap N' Proto, the only nice solution would be to write another `deriving FromJson` strategy (since the handlers are parametrised now). But the current handler is in dire need of a refactor already, so overall it would be quite a bit of work to fix something which isn't really the problem. Note though that I switched to a "flat" `TaggedText.append : Array (TaggedText) -> TaggedText` to avoid some unnecessary nesting, so there aren't as many nested objects (it was blowing the JS stack in VSCode internals).

gebner: > Note though that I switched to a "flat" TaggedText.append : Array (TaggedText) -> TaggedText to avoid some unnecessary nesting

Yes, this is great (same for `text` I hope).  It also gets rid of the "ugly" `as` name.

src/Lean/Widget/ExprWithCtx.lean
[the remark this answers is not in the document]
Vtec234: Now that we are using `Info`s which are produced by the delaborator eagerly, `LazyExprWithCtx` is no longer necessary. For lazy traces we use `MsgEmbed.lazyTrace`.

gebner:
There's also quite a bit of discussion on the corresponding vscode-lean4 PR: https://github.com/leanprover-community/vscode-lean4/pull/32

Vtec234:
This should now be in a reviewable state. There are some TODOs for which I will make issues, but the backend seems to be working.

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -133,54 +135,66 @@ partial def handleDefinition (kind : GoToKind) (p : TextDocumentPositionParams)
...
 open Elab in
-open Meta in
-partial def handlePlainTermGoal (p : PlainTermGoalParams)
-    : RequestM (RequestTask (Option PlainTermGoal)) := do
+partial def handlePlainGoal (p : PlainGoalParams)
+    : RequestM (RequestTask (Option PlainGoal)) := do
+  let t ← getInteractiveGoals p
+  t.map <| Except.map fun ⟨goals⟩ =>
+    if goals.isEmpty then
+      some { goals := #[], rendered := "no goals" }
```
gebner, line 162: I'll look at this branch more closely later.  However this is a breaking change: it now shows "goals accomplished" everywhere (even when there's no tactic proof).

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -133,54 +135,66 @@ partial def handleDefinition (kind : GoToKind) (p : TextDocumentPositionParams)
               return ← ci.runMetaM i.lctx <| locationLinksFromDecl i ei.elaborator
       return #[]
 
-open Elab in
-partial def handlePlainGoal (p : PlainGoalParams)
-    : RequestM (RequestTask (Option PlainGoal)) := do
+open RequestM in
+def getInteractiveGoals (p : Lsp.PlainGoalParams) : RequestM (RequestTask Widget.InteractiveGoals) := do
```
gebner, line 139:
```suggestion
def getInteractiveGoals (p : Lsp.PlainGoalParams) : RequestM (RequestTask (Option Widget.InteractiveGoals)) := do
```

src/Lean/Server/FileWorker/RequestHandling.lean
Vtec234: Ah thanks, that's a bug. I forgot about the `none` meaning not a proof and `#[]` meaning goals accomplished semantics.

src/Lean/Data/Json/Printer.lean
```diff
@@ -64,19 +64,35 @@ end
...
-  | num s      => s.toString
-  | str s      => renderString s
-  | arr elems  =>
-    let f := ",".intercalate (elems.map compress).toList
-    s!"[{f}]"
-  | obj kvs    =>
-    let ckvs := kvs.fold (fun acc k j => s!"{renderString k}:{compress j}" :: acc) []
-    let ckvs := ",".intercalate ckvs
-    s!"\{{ckvs}}"
+protected inductive CompressWorkItem
```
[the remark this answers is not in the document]
Vtec234: It doesn't seem possible to `open` a `private inductive` to get its constructors (`unknown namespace 'CompressWorkItem'`).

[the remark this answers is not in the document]
Kha: Interesting, I think that could be considered a bug

src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -224,10 +269,24 @@ def delab (currNamespace : Name) (openDecls : List OpenDecl) (e : Expr) (options
...
+    pure (stx, (← get).infos)
   catchInternalId Delaborator.delabFailureId
-    (Delaborator.delab.run { defaultOptions := opts, optionsPerPos := optionsPerPos, currNamespace := currNamespace, openDecls := openDecls } (Delaborator.SubExpr.mkRoot e))
+    (run
+      { defaultOptions := opts
+        optionsPerPos := optionsPerPos
+        currNamespace := currNamespace
+        openDecls := openDecls
+        subExpr := Delaborator.SubExpr.mkRoot e }
+      |>.run' { : Delaborator.State })
```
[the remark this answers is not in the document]
Vtec234: Perhaps readability? (Daniel wrote that part.) But it's also readable with a structure match.

src/Lean/Widget/TaggedText.lean
```diff
@@ -0,0 +1,107 @@
...
+
+def appendTag (acc : TaggedText α) (t₀ : α) (a₀ : TaggedText α) : TaggedText α :=
+  match acc with
+  | append as => append (as.push <| tag t₀ a₀)
+  | text ""   => tag t₀ a₀
+  | a         => append #[a, tag t₀ a₀]
+
+partial def map (f : α → β) : TaggedText α → TaggedText β :=
+  go
+where go : TaggedText α → TaggedText β
```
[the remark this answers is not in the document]
Vtec234: Changed, but I must note that this is one of few things where the Lean 3 syntax actually seems more natural/sensible to me. The old fixed-before-colon/recursive-after-colon makes sense once one gets used to it. It's kind of weird that `variable (a : T) in def foo` has a different behaviour than directly writing `def foo (a : T)` and indeed I forgot about it here.

src/Lean/Server/Snapshots.lean
```diff
@@ -110,29 +130,29 @@ def compileNextCmd (contents : String) (snap : Snapshot) : IO (Sum Snapshot Mess
...
-          severity := MessageSeverity.information
-          pos      := inputCtx.fileMap.toPosition snap.endPos
-          data     := output
-        }
+        messages := postCmdState.messages.add
+          { fileName := inputCtx.fileName
+            severity := MessageSeverity.information
+            pos      := inputCtx.fileMap.toPosition snap.endPos
+            data     := output
+          }
```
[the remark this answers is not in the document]
Vtec234: I'm not even sure why I changed it!

src/Lean/Widget/InteractiveDiagnostic.lean
```diff
@@ -0,0 +1,208 @@
...
+    let t ← pushEmbed <| EmbedFmt.expr ci ctx.lctx infos
+    return Format.tag t fmt
+  | _,    none,      ofGoal mvarId            => pure $ "goal " ++ format (mkMVar mvarId)
+  | nCtx, some ctx,  ofGoal mvarId            => withIgnoreTags <| ppGoal (mkPPContext nCtx ctx) mvarId
+  | nCtx, _,         withContext ctx d        => go nCtx ctx d
+  | _,    ctx,       withNamingContext nCtx d => go nCtx ctx d
+  | nCtx, ctx,       tagged t d               => do
+    -- tagged is *almost* perfect for detecting traces
+    -- expect for the following two other occurrences:
+    -- src/Lean/Elab/Term.lean:454
```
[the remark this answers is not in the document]
Vtec234: Thanks for spotting it, the comment was irrelevant anyway since we postfix `_traceCtx` now.

src/Lean/Server/Snapshots.lean
```diff
@@ -78,26 +92,32 @@ partial def parseAhead (contents : String) (snap : Snapshot) : IO (Array Syntax)
...
 -- isServer? conditionals and not be worth it due to how short it is.
-def compileNextCmd (contents : String) (snap : Snapshot) : IO (Sum Snapshot MessageLog) := do
-  let inputCtx := Parser.mkInputContext contents "<input>"
+def compileNextCmd (text : FileMap) (snap : Snapshot) : IO Snapshot := do
+  let inputCtx := Parser.mkInputContext text.source "<input>"
   let cmdState := snap.cmdState
   let scope := cmdState.scopes.head!
   let pmctx := { env := cmdState.env, options := scope.opts, currNamespace := scope.currNamespace, openDecls := scope.openDecls }
   let (cmdStx, cmdParserState, msgLog) :=
     Parser.parseCommand inputCtx pmctx snap.mpState snap.msgLog
```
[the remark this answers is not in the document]
Vtec234: Yeah, that would get rid of computing the `MessageLog` diff. But we do want `Snapshot.interactiveDiags` to contain every diagnostic up to a point so that we can send them easily without folding over all snapshots. So we'd have a `MessageLog` with messages just for this snapshot but an interactive diagnostics log for everything, which is perhaps a bit weird. What do you think?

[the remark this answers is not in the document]
Kha: I think that's fine, we can call it `allInteractiveDiags` or something similar. Not sure folding would even be a performance issue, concatenating mostly-empty arrays should be quite fast even with hundreds of snapshots.

gebner:
I've tried implementing the RPC protocol in the neovim plugin (https://github.com/Julian/lean.nvim/pull/118).  Some thoughts:

There is another problem with `$/lean/rpc/connect` being a notification instead of a request: you don't know if the request failed.  In particular, there is no way to tell if you are connected to an old lean version which does not support the request or if you just need to wait a little longer.  This should really be a request and not a notification, synchronous communication is what requests are made for.

The keep-alive mechanism is not particularly helpful for the neovim plugin.  It doesn't do anything that can't be done by sending a `$/lean/rpc/connect` request to close the current session.  (Sending a `$/lean/rpc/connect` notification might also be an option for the vscode implementation.)

I'd like to reiterate at this point that I think it is worth to get the basic design right from the beginning.  It is trivial to add new functions and fields to existing structures; and this is unlikely to break anything.  However the current approach really hardcodes the assumption that every editor has at most one infoview which uses at most one session, and doesn't use the RPC for anything else.  I would at least like to see a plan on how to support two infoviews.  There are all kinds of cool things on the horizon ([e.g. there could be a vscode API to embed a webview in the editor](https://github.com/microsoft/vscode/issues/85682)), and I would hate it if we couldn't make use of them due to basic design constraints.

The `ToJson Name` and `FromJson Name` instances surprisingly require a backtick.  This has the effect that you need to write ``"`Lean.Widget.getInteractiveGoals"`` as a method name.  I think the backtick should be removed from the instances.

The `names` field in `InteractiveHypothesis` should probably be an array.

The `getInteractiveTermGoal` function doesn't return the range (which the `plainTermGoal` request does).

I don't completely understand the `ExprWithCtx` API.  Is there even a way to get an `ExprWithCtx` reference?  From what I can tell, this seems to be supplanted by the new `InfoWithCtx` API, which seems to be less general (e.g. it doesn't have `inferType`).  Either way, if one of them is already obsolete, then this should be cleaned up before merging.

gebner:
Something else that would be super cool and might be hard to retrofit (because the InfoWithCtx is a reference, and we can't just add a field there) is semantic highlighting.  That is, information about which parts of a CodeWithInfos are keywords, literals, operators, etc.

Vtec234:
@gebner I am thinking more about backwards compatibility. We can preserve the interface when possible, but ultimately I do want to at least be able to make breaking changes. I am assuming that we do not have to worry about a new server<->old client situation because at least VSCode will auto-update extensions, and in Vim or Emacs we can expect users to update as well. This leaves new client<->old server in cases when one is working on a project that e.g. uses older mathlib. For this, I think we should start properly versioning the Lean server (currently on 0.0.1) and detect this in editor extensions when necessary, then use the API for that server version.

src/Lean/Server/Snapshots.lean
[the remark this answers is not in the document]
Vtec234: I tried this out now, but it is still clunky. The reason is that the `MessageLog` is part of `Command.State` which is designed for a "rolling" usage. So we'd have to leave most of the command state as-is but clear out just the message log. I would leave it as is, but added a comment clarifying it.

Vtec234:
All outstanding comments should be addressed now.

leodemoura:
@Vtec234 Is it ready to be merged?

src/Lean/PrettyPrinter/Delaborator/SubExpr.lean
```diff
@@ -97,6 +101,36 @@ def withNaryArg (argIdx : Nat) (x : m α) : m α := do
...
+  iter.curr
+
+def HoleIterator.next (iter : HoleIterator) : HoleIterator :=
+  if (iter.curr+1) == iter.top then
+    ⟨2*iter.top, maxChildren*iter.top⟩
+  else ⟨iter.curr+1, iter.top⟩
+
+/-- The positioning scheme guarantees that there will be an infinite number of positions
+that are never used by subexpressions. We use these to attach additional `Elab.Info`. -/
+def nextExtraPos : m Pos := do
```
leodemoura, line 127: Could you please add comments explaining why need `HoleIterator`s?

src/Lean/Message.lean
```diff
@@ -116,7 +116,12 @@ partial def formatAux : NamingContext → Option MessageDataContext → MessageD
   | nCtx, some ctx,  ofGoal mvarId            => ppGoal (mkPPContext nCtx ctx) mvarId
   | nCtx, _,         withContext ctx d        => formatAux nCtx ctx d
   | _,    ctx,       withNamingContext nCtx d => formatAux nCtx ctx d
-  | nCtx, ctx,       tagged _ d               => formatAux nCtx ctx d
+  | nCtx, ctx,       tagged t d               =>
+    if let Name.str cls "_traceCtx" _ := t then do
```
leodemoura, line 120: Could you please add a comment explaining why we need the `if` here?

src/Lean/Meta/PPGoal.lean
```diff
@@ -161,7 +161,6 @@ def ppGoal (mvarId : MVarId) : MetaM Format := do
         else
           let fmt := addLine fmt
           match ids, type? with
-          | [], _        => pure fmt
```
leodemoura, line 164: Is this a bug fix? Or, was this case unnecessary?

src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -36,19 +36,25 @@ The delaborator is extensible via the `[delab]` attribute.
...
 
 structure Context where
   defaultOptions : Options
   optionsPerPos  : OptionsPerPos
   currNamespace  : Name
   openDecls      : List OpenDecl
   inPattern      : Bool := false -- true whe delaborating `match` patterns
+  subExpr        : SubExpr
+
+structure State where
```
leodemoura, line 49: Could you please comment this ner structure?

src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -170,12 +188,40 @@ def withBindingBodyUnusedName {α} (d : Syntax → DelabM α) : DelabM α := do
...
 partial def delabFor : Name → Delab
   | Name.anonymous => failure
-  | k              => do
-    (delabAttribute.getValues (← getEnv) k).firstM id >>= annotateCurPos <|>
-      -- have `app.Option.some` fall back to `app` etc.
-      delabFor k.getRoot
+  | k              =>
+    (do let stx ← (delabAttribute.getValues (← getEnv) k).firstM id
+        let stx ← annotateCurPos stx
+        addTermInfo (← getPos) stx (← getExpr)
```
leodemoura, line 221: What is the overhead of using `addTermInfo` here? Is the slowdown significant? Just curious.

src/Lean/PrettyPrinter/Delaborator/Basic.lean
Vtec234: I wasn't sure what to say about `State` but I commented the fields.

src/Lean/Meta/PPGoal.lean
Vtec234: It was unnecessary as just above we check `ids.isEmpty`. Actually the match on `ids` is redundant, so I removed that as well now.

Vtec234:
> @Vtec234 Is it ready to be merged?

I think so. There are some smaller discussions left which can be addressed later.

src/Lean/PrettyPrinter/Delaborator/Basic.lean
Vtec234: I didn't test in a profiler, but in terms of wall time on both a long trace (~1s to print) and a long expression (~1.3s), I observed no difference with and without the `addTermInfo/addFieldInfo` calls.

PR #606, opened 2021-08-05
feat: borrow inference: preserve mutual tail calls

Kha:
While the current C backend guarantees TCO only for non-mutual tail calls, it is simple enough to extend the borrow inference heuristics to preserve them on mutually recursive calls as well for the benefit of other backends/eventually the clang 13 `[[musttail]]` attribute.

### DIFF
--- a/src/Lean/Compiler/IR/Borrow.lean
+++ b/src/Lean/Compiler/IR/Borrow.lean
@@ -137,6 +137,7 @@ def applyParamMap (decls : Array Decl) (map : ParamMap) : Array Decl :=
 
 structure BorrowInfCtx where
   env      : Environment
+  decls    : Array Decl  -- block of mutually recursive functions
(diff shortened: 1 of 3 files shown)


src/Lean/Compiler/IR/Borrow.lean
```diff
@@ -259,8 +260,7 @@ def preserveTailCall (x : VarId) (v : Expr) (b : FnBody) : M Unit := do
   let ctx ← read
   match v, b with
   | (Expr.fap g ys), (FnBody.ret (Arg.var z)) =>
-    if ctx.currFn == g && x == z then
-      -- dbgTrace ("preserveTailCall " ++ toString b) $ fun _ => do
+    if ctx.decls.any (·.name == g) && x == z then
```
Kha, line 263: @leodemoura I'm assuming that a linear scan here is likely faster/fast enough given the expected size of `decls`

PR #608, opened 2021-08-05
fix: actually interpret imported anonymous `[init]` decls

Kha:


src/Lean/Compiler/InitAttr.lean
```diff
@@ -98,7 +99,6 @@ def hasInitAttr (env : Environment) (fn : Name) : Bool :=
   (getInitFnNameFor? env fn).isSome
 
 def setBuiltinInitAttr (env : Environment) (declName : Name) (initFnName : Name := Name.anonymous) : Except String Environment :=
-  -- builtinInitAttr.setParam env declName initFnName -- TODO: use this one
-  regularInitAttr.setParam env declName initFnName
+  builtinInitAttr.setParam env declName initFnName
```
Kha, line 102: @leodemoura Just a bootstrapping oversight, I assume?

PR #612, opened 2021-08-06
fix: forallMetaTelescope shouldn't reduce after maxMVars

dselsam:
At least it doesn't seem like it should. The following test illustrates the issue: https://github.com/dselsam/lean4/blob/forallMetaTelescope/tests/lean/forallMetaTelescopeBounded.lean Without this PR, the `resultType` will have unfolded `Set`, exposing an arrow. With the PR, it simply prints `Set ?_uniq.1`. Context: the old behavior was complicating `isDefEq` queries inside `pp.analyze`: what I would have expected were trivial `Set ?x =?= Set x` queries became not-quite-as-trivial `?x -> Prop =?= Set x` queries.


leodemoura:
Thanks for finding this problem, and suggesting a fix and test. I will submit a fix soon.

PR #614, opened 2021-08-06
refactor: separate pp.funBinderTypes and pp.piBinderTypes

dselsam:


tests/lean/PPRoundtrip.lean.expected.out
```diff
@@ -40,7 +40,7 @@ Type → Type → Type
   [inst :
     ToString Nat],
   x = x
-∀ (x : Nat), x = x
+∀ x, x = x
```
dselsam, line 43: Previously, `pp.binderTypes false` was getting overridden by `pp.analyze`, which would still annotate all `pi`s to need binder types. Now, `pp.analyze` does not add annotations `pi` binders.

tests/lean/ppMotives.lean.expected.out
```diff
@@ -11,7 +11,7 @@ protected def Nat.add : Nat → Nat → Nat :=
 fun x x_1 =>
   Nat.brecOn (motive := fun x => Nat → Nat) x_1
     (fun x f x_2 =>
-      (match x_2, x : (x : Nat) → (x : Nat) → Nat.below (motive := fun x => Nat → Nat) x → Nat with 
+      (match x_2, x : (x x : Nat) → Nat.below (motive := fun x => Nat → Nat) x → Nat with
```
dselsam, line 14: The old output is wrong and the new one is right. (Note: I did not figure out why it used to print that way).

tests/lean/ppMotives.lean.expected.out
```diff
@@ -22,7 +22,7 @@ fun x x_1 x_2 x_3 =>
   | α, .(α), Eq.refl α, a => HEq.refl _
 theorem ex.{u} : ∀ {α β : Sort u} (h : α = β) (a : α), cast h a ≅ a :=
 fun x x_1 x_2 x_3 =>
-  match x, x_1, x_2, x_3 : Sort u → Sort u → ∀ x_5, ∀ x_6, cast x_5 x_6 ≅ x_6 with 
+  match x, x_1, x_2, x_3 : ∀ (x x_4 : Sort u) (x_5 : x = x_4) (x_6 : x), cast x_5 x_6 ≅ x_6 with
```
dselsam, line 25: Currently, `pp.analyze` doesn't actually see the `pi` type that gets annotated for matches, it only sees its original `lambda` version. So, it does not annotate binder types. The new behavior is better -- the "synthetic" `pi` now gets the binder types despite `pp.analyze` not annotating them.

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
```diff
@@ -455,13 +457,16 @@ private def shouldGroupWithNext : DelabM Bool := do
...
+  getPPBinderTypes (e : Expr) :=
+    if e.isForall then getPPPiBinderTypes else getPPFunBinderTypes
 
 private partial def delabBinders (delabGroup : Array Syntax → Syntax → Delab) : optParam (Array Syntax) #[] → Delab
   -- Accumulate names (`Syntax.ident`s with position information) of the current, unfinished
   -- binder group `(d e ...)` as determined by `shouldGroupWithNext`. We cannot do grouping
   -- inside-out, on the Syntax level, because it depends on comparing the Expr binder types.
   | curNames => do
-    if (← shouldGroupWithNext) then
+    if ← shouldGroupWithNext then
```
Kha, line 469: Ah, I know that I've use the latter style in the delaborator, but Leo prefers the one with parentheses. Just something to keep in mind for the future. We should fix all existing occurrences automatically at some point...

Kha (approved):

PR #618, opened 2021-08-09
feat: pp.instances and pp.instances.withType

dselsam:
New pp options for printing (as part of explicit applications) instance arguments in any of 2x2 ways (terms-vs-holes, types-vs-no-types). See the test and expected outputs below. I have found this extremely useful when debugging TC issues. Also, since instances are one of the main sources of blowup when using `pp.all=true`, the combination `pp.all=true, pp.instances=false` may be a useful level of detail in other contexts.

### DIFF
--- a/src/Lean/PrettyPrinter/Delaborator/Options.lean
+++ b/src/Lean/PrettyPrinter/Delaborator/Options.lean
@@ -95,6 +95,16 @@ register_builtin_option pp.proofs.withType : Bool := {
   group    := "pp"
   descr    := "(pretty printer) when eliding a proof (see `pp.proofs`), show its type instead"
 }
+register_builtin_option pp.instances : Bool := {
+  defValue := true
+  group    := "pp"
+  descr    := "(pretty printer) if set to false, replace inst-implicit arguments to explicit applications with placeholders"
+}
+register_builtin_option pp.instanceTypes : Bool := {
+  defValue := false
+  group    := "pp"
+  descr    := "(pretty printer) when printing explicit applications, show the types of inst-implicit arguments"
+}
 register_builtin_option pp.motives.pi : Bool := {
   defValue := true
   group    := "pp"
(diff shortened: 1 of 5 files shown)


tests/lean/PPInstances.lean.expected.out
```diff
@@ -0,0 +1,6 @@
+PPInstances.lean:27:68-27:73: warning: declaration uses 'sorry'
+@Module R E _ _ : Type
+@Module R E (_ : Semiring R) (_ : AddCommMonoid E) : Type
```
dselsam, line 3: This combination `pp.instances=false, pp.instances.withType=true` is great for spotting these nested instances. Wow, I didn't even realize that the nesting could get so deep. Here is a goal from binport that I have been struggling to synthesize:
```lean
(_ :
                HasLiftT
                  (@coeSort'
                    (@Subspace Real E (_ : Field Real) (_ : AddCommGroup E)
                      (_ : @Module Real E (_ : Semiring Real) (_ : AddCommMonoid E)))
                    (_ :
                    HasCoeToSort
                      (@Subspace Real E (_ : Field Real) (_ : AddCommGroup E)
                        (_ : @Module Real E (_ : Semiring Real) (_ : AddCommMonoid E))))
                    p)
                  E)
```

src/Lean/PrettyPrinter/Delaborator/Options.lean
```diff
@@ -95,6 +95,16 @@ register_builtin_option pp.proofs.withType : Bool := {
   group    := "pp"
   descr    := "(pretty printer) when eliding a proof (see `pp.proofs`), show its type instead"
 }
+register_builtin_option pp.instances : Bool := {
+  defValue := true
+  group    := "pp"
+  descr    := "(pretty printer) when printing explicit applications, show holes instead of inst-implicit arguments"
```
Kha, line 101: Meaning negated?

tests/lean/PPInstances.lean.expected.out
Kha: It looks good, but (minor nit) it feels a bit weird that `pp.instances.withType` does something when `pp.instances` is disabled. Maybe `pp.instanceTypes` would be a more appropriate name?

Kha (approved):

PR #622, opened 2021-08-11
Update lean3changes.md

BoltonBailey:
Above, you say "We do not use `,` anymore to separate the binders from the lambda expression body". If that's true, shouldn't this be a `=>`?


Vtec234:
That sentence is about Lean 3, where the notation was `fun x, x + 1`. I'd say it is correct.

PR #623, opened 2021-08-11
Generalize macro parameter syntax

Kha:
* allow any `stx:arg`, optionally preceded by a binding `$ident:`, as macro parameters
* adjust precedence of all `stx` prefix and postfix operators to `arg` so they can now be used as macro parameters without parentheses
* remove the now-redundant syntax (in `macro`) `strLit%ident`, it can be replaced with `ident:strLit`
* put parameters without a binding in `group` so that e.g. `macro x:ident noWs ...` doesn't break (`macro` doesn't know that `noWs` has arity 0)

### DIFF
--- a/tests/lean/weirdmacro.lean
+++ b/tests/lean/weirdmacro.lean
@@ -1,4 +1,0 @@
-macro term x:term : term => `($x)
-
-#check term 10
--- 10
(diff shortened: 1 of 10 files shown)


tests/lean/weirdmacro.lean
```diff
@@ -1,4 +0,0 @@
-macro term x:term : term => `($x)
```
Kha, line 1: @leodemoura I removed this test case since the leading `term` would now be interpreted as the actual parser. This test case only existed to check for absence of the previous "auto-quoting" feature, I don't think any user would write down something like this and expect something sensible to happen.

Issue #625, opened 2021-08-11
Pretty-printer inconsistent with MData-wrapped app functions


dselsam:
```lean
import Lean
open Lean Lean.PrettyPrinter

def foo : Unit → Unit := id
def x   : Unit := ()

def wrap (e : Expr) := mkMData {} e

def e : Expr := mkApp (wrap $ mkConst `foo) (mkConst `x)

@[appUnexpander foo] def unexpandFoo : Unexpander
  | `($(_) _) => `(myFoo)
  | _         => throw ()

#eval delab Name.anonymous [] e
/-
PANIC at Lean.PrettyPrinter.Delaborator.unexpandRegularApp Lean.PrettyPrinter.Delaborator.Builtins:199:52:
unreachable code has been reached
-/
```
The `isRegularApp` check unfolds `MData`s on the app function (https://github.com/leanprover/lean4/blob/master/src/Lean/PrettyPrinter/Delaborator/Builtins.lean#L192) but `unexpandRegularApp` does not (https://github.com/leanprover/lean4/blob/master/src/Lean/PrettyPrinter/Delaborator/Builtins.lean#L199). Note: although the latter line causes the `PANIC`, the behavior in `isRegularApp` is also incorrect since the `MData` may include additional pretty-printer options.

PR #626, opened 2021-08-11
fix: #625

dselsam:


src/Lean/PrettyPrinter/Delaborator/Builtins.lean
```diff
@@ -187,16 +187,28 @@ def shouldShowMotive (motive : Expr) (opts : Options) : MetaM Bool := do
   <||> (← getPPMotivesPi opts <&&> returnsPi motive)
   <||> (← getPPMotivesNonConst opts <&&> isNonConstFun motive)
 
+partial def withMDatas [Inhabited α] (x : DelabM α) : DelabM α := do
```
Kha, line 190: Should we extract a function `withMData` without recursion and use it in `delabMData`?

dselsam: Good idea. Will fix.

PR #627, opened 2021-08-11
feat: unexpander for Subtype

dselsam:
Note: we disable `pp.analyze` in the test because it doesn't "play well" with `Subtype` yet. We probably want a heuristic that anything of the form `@f α (fun (x : α) => ...) ...` annotates the binder type of `x` rather than explicitly specifying the implicit `α`.


Kha:
> We probably want a heuristic that anything of the form `@f α (fun (x : α) => ...) ...` annotates the binder type of `x` rather than explicitly specifying the implicit `α`.

SGTM in general

PR #634, opened 2021-08-18
chore: align `stx` precedence in `syntax` to the new one in `macro`

Kha:


Kha:
Thinking more about this, while it rejects `syntax "foo" term <|> "bar" term` instead of using the confusing interpretation `syntax "foo" (term <|> "bar") term`, the same issue is still present in e.g. `syntax ("foo" term <|> "bar" term)`. Even if we fixed the grammar there as well to mean `("foo" term) <|> ("bar" term)`, `<|>` is still a somewhat dangerous operator to even expose to users, at least in combination with `(...)`: syntax `match` will not actually be able to distinguish between the two cases since both use `nullKind`. This will need some more deliberation.

tydeu:
@Kha I think `(...)` creating null nodes may be the pain point here. Would it be possible to have `(...)` not create a node at all (and maybe have an alternate `&(...)` be the null generated expression). For example:

```lean
syntax ex1 := foo bar <|> bar baz
  -- equivalent to:
def ex1' := leading_parser (foo >> bar) <|> (bar >> baz) 

syntax ex2 := foo (bar <|> bar) baz
  -- equivalent to:
def ex2' := leading_parser foo >> (bar <|> bar) >> baz

syntax ex3a := foo (bar "+" bar) baz  
  -- equivalent to:
def ex3a' := leading_parser foo >> bar >> "+" >> bar >> baz 

syntax ex3b := foo &(bar "+" bar) baz  
  -- equivalent to:
def ex3b' := leading_parser foo >> node `null (bar >> "+" >> bar) >> baz

leodemoura:
> Even if we fixed the grammar there as well to mean ("foo" term) <|> ("bar" term), <|> is still a somewhat dangerous operator to even expose to users, at least in combination with (...): syntax match will not actually be able to distinguish between the two cases since both use nullKind. This will need some more deliberation.

I think we should generate an error message. It would tell users to create an auxiliary `syntax aux := ...` for each alternative otherwise syntax match will not work. Don't remember whether this issue affects other combinators.
 
BTW, if I remember correctly, `("foo" term) <|> ("bar" term)` does not create a `nullKind` node for each alternative. Other combinators do it (e.g. `many`), but not `<|>`.

leodemoura:
Linking relevant issue: https://github.com/leanprover/lean4/issues/638

leodemoura:
I am merging the PR, the actual change is mostly orthogonal to the discussion here. We can use #638 to continue the discussion.

PR #642, opened 2021-08-21
perf: short-circuit lean_string_append

Kha:
No need to allocate/copy when either side is the empty string


Kha:
!bench

Kha:
Huh, interesting

Kha:
!bench

Kha:
Oh well.

tydeu:
Interesting benchmarks! I guess it is at least 8 additional instructions for every copy for an edge case that might not happen that often. Thus, the former may outweigh the later.

PR #647, opened 2021-08-27
doc: lean4-mode with straight.el and use-package

hargoniX:
Since pulling in lean4-mode via use-package + straight.el is a tad more complicated than the usual
```
(use-package foo-mode
  :straight t)
```
I added some instructions about how I pull the package into my emacs configuration. While doing that I also updated the URL in the
package description to match the current repository and added the latest version of lsp-mode to the dependencies since it was missing before entirely. If the latest version is not the one this should depend on I could of course pull it down a bit as well.


Kha:
Thanks!

PR #657, opened 2021-09-07
small fixes to msys docs

lovettchris:


leodemoura:
merged

doc/make/msys2.md
```diff
@@ -22,12 +22,44 @@ Here are the commands to install all dependencies needed to compile Lean on your
 pacman -S make mingw-w64-x86_64-cmake mingw-w64-x86_64-ccache mingw-w64-x86_64-gcc git
 ```
 
+Then make sure the following is included in your PATH:
+
+```bash
+export PATH=$PATH:/mingw64/bin
```
Kha, line 28: This should definitely happen by default: https://github.com/msys2/MSYS2-packages/blob/85d54481612aea963af4ce822a6d9d98cb2a189b/filesystem/profile#L52
Is it clear under which circumstances it might not (which might lead to other issues)? Are you sure you started an "MSYS2 MinGW 64-bit" shell, not "MSYS2 MSYS"?

doc/make/msys2.md
```diff
@@ -22,12 +22,44 @@ Here are the commands to install all dependencies needed to compile Lean on your
...
 `cmake ../.. -G "Unix Makefiles"` instead of `cmake ../..`. This ensures that cmake will call `sh` instead
 of `cmd.exe` for script tasks.
 
 ## Install lean
 
 You can use the `install` ninja/make target to install Lean into, by default,
-`C:\\User Programs (x86)\\LEAN`. To change this, add `-DCMAKE_INSTALL_PREFIX=path/you/want`
-to your cmake invocation.
+`./build/release/stage1/msys64/lean/`. To change this, add `-DCMAKE_INSTALL_PREFIX=path/you/want`
+to your initial cmake invocation.
```
Kha, line 46: If I try `make install`, I get
```
Install the project...
-- Install configuration: "Release"
-- Installing: C:/Program Files (x86)/LEAN/include/lean
CMake Error at cmake_install.cmake:41 (file):
  file INSTALL cannot make directory "C:/Program Files
  (x86)/LEAN/include/lean": No such file or directory.
```
Not sure if that is a permission problem.

doc/make/msys2.md
melted: Yes, that is probably not best to recommend as MINGW64 sets up other things too.

lovettchris: Ah, you are right I was not using mingw64.exe.  My bad, I will fix it.

doc/make/msys2.md
lovettchris: Probably is given that "c:/Program Files" is usually not writable by non-admin users...

PR #659, opened 2021-09-08
Build & distribute release builds with Zig as stand-alone C compiler

Kha:
As they say, [Zig is an obvious choice for what C compiler to ship with your language](https://andrewkelley.me/post/zig-cc-powerful-drop-in-replacement-gcc-clang.html). In particular, this is a big step towards zero-setup native compilation on Windows.

We use Zig's cross-compilation feature to link the Linux build against the ancient glibc 2.15, obviating the previous workaround of using an older nixpkgs channel that combines a not-quite-recent glibc with a not-quite-recent clang. Other cross-compilation targets for platforms not supported by GH Actions would also be interesting, though I wasn't able to find one that works out of the box from Linux yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,7 +122,7 @@ jobs:
               echo "LEAN_VERSION_STRING=$LEAN_VERSION_STRING" >> $GITHUB_ENV
             fi
           fi
-          cmake .. ${{ matrix.CMAKE_OPTIONS }} "${OPTIONS[@]}"
+          cmake .. ${{ matrix.CMAKE_OPTIONS }} "${OPTIONS[@]-}"
           make -j4
       # de-Nix-ify binaries
       - name: Patch


Kha:
To give a bit more context and a progress update, this PR tries to build all of Lean with Zig, and then copy it to `bin/` and have `leanc` use it as a portable C compiler. Conceptually this is the simplest approach and it opens up the possibility of using Zig for cross-compiling Lean, as this PR already does for glibc 2.15 as mentioned above. However, in practice there still seem to be many hurdles to overcome for (non-cross-)compiling a nontrivial C++ project like Lean with it on all our platforms, see below.

An alternative approach would be to keep using system/Nix clang for compiling Lean itself and ship Zig only for compiling user Lean programs. Since Lean produces C code, not C++, and of a quite limited variety, it should be much easier to ensure that Zig can cope with it. However, we would still need to *link* Lean programs with the C++ standard library used for compiling the Lean stdlib, so we would need to ship its static library with Lean. That should work I think, but we could still run into compiler mismatches in advanced cases, e.g. LTO. And we would not be able to cross-compile Lean.

So the first approach would be preferable. The solved and unsolved issues with it are:
* [x] `zig c++` does not support exceptions anymore
  * [fixed upstream](https://github.com/ziglang/zig/issues/9591)
* [x] `zig ld` does not support `--whole-archive/-force_load` for building shared libraries from static libraries
  * workaround: hackily gather & pass .o files manually, use response file to avoid Windows cmdline length limit
* [x] macOS: `getopt.h` seems to be missing
  * workaround: we already have a custom implementation for MSVC, use it on all platforms
* [x] Windows: `pthread.h` seems to be missing
  * workaround: turn off `AUTO_THREAD_FINALIZATION` on Windows
  * @leodemoura there's also an old comment saying that we should use C++11's `thread_local` when it's ready, which I hope it is by now
* [x] Windows: Zig's `ntdef.h` defines `min`/`max` macros, which mingw64 didn't
  * solution: `#define NOMINMAX`
* [x] Windows: mingw64 `ccache` does not like our `zigcc/zigc++` bash wrapper scripts
  * not directly an issue with Zig
  * solution should be: use msys2 `ccache`, but it segfaults...
  * workaround: do not use ccache when building with Zig on Windows
    * [ ] might want to use ccache without Zig for non-release builds, or figure out how to store Zig's own cache on GitHub Actions
* [ ] macOS: can't find `utimensat`
  * wrong POSIX version, perhaps (somehow?) inherited [by Nix](https://github.com/NixOS/nixpkgs/issues/86299)?
* [ ] Windows: unclear linker problem linking `libleanshared.so`, possibly linking order related (but then why does it work on Linux, and on Windows without Zig?)
* [ ] Zig's caching is not quite compatible with `make`
  * when restoring a compile target from Zig's cache, the mtime is not updated, so make will remake the target again on the next run
  * should not be an issue for CI
* [ ] to be continued...?

And for the fun of it, let's try cross-compiling from x86_64-linux-gnu:
* [x] `aarch64-linux-gnu`: seems to work! :tada: 
* [ ] `aarch64-macos-gnu`: building Lean seems to work, but configuring GMP fails immediately?
```text
configure:5940: /home/sebastian/lean/lean/build/release/zigcc -O2 -pedantic  conftest.c >&5
warning(zld): unknown filetype for positional input file: '/home/sebastian/.cache/zig/o/de3616f6617a99f6465ca1549bd65449/conftest.o'
```
```bash
file /home/sebastian/.cache/zig/o/de3616f6617a99f6465ca1549bd65449/conftest.o
/home/sebastian/.cache/zig/o/de3616f6617a99f6465ca1549bd65449/conftest.o: LLVM bitcode, wrapper
```
Not sure why this issue would occur with a trivial file while building GMP, but not with Lean
* [ ] Emscripten: unclear to me what target to use, if any. `wasm32-wasi-musl` support looks much better, but AFAIK there are many features missing from WASI until it could run Lean
* [ ] to be continued...?

Kha:
I'll take the liberty to ping @andrewrk himself here in case he's interested in our little Zig experiment & progress :) ...

leodemoura:
> @leodemoura there's also an old comment saying that we should use C++11's thread_local when it's ready, which I hope it is by now

I hope so too :)

Kha:
Phew, good progress on Windows after many more fixes & workarounds (see above)! The test suite is now all green on my Windows machine, while on CI it randomly fails in the interactive tests with a single last error :thinking: .

Unfortunately, on macOS the tests uncovered many more linker issues that I don't see easy workaround for. I can open upstream issues for them, but unless someone is nice enough to specifically fix all of them for us, I think the only realistic solution is to ship without Zig on macOS for the time being.

Kha:
Obsoleted by #795, which did not run into any of the above issues apart from the Windows headers (which apparently are from the clang64 toolchain)

PR #662, opened 2021-09-08
Fix msys text based on code review feedback.

lovettchris:
Got some good code review feedback on my previous commit, this PR fixes those issues.

### DIFF
--- a/doc/make/msys2.md
+++ b/doc/make/msys2.md
@@ -30,16 +30,18 @@ gcc --version
 cmake --version
 ```
 
-Then follow the [generic build instructions](index.md) in the MSYS2 MinGW shell, using
-`cmake ../.. -G "Unix Makefiles"` instead of `cmake ../..`. This ensures that cmake will call `sh` instead
-of `cmd.exe` for script tasks.
+Then follow the [generic build instructions](index.md) in the MSYS2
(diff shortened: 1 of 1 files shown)


doc/make/index.md
```diff
@@ -45,6 +53,13 @@ This will compile the Lean library and binary into the `stage1` subfolder; see
 below for details. Add `-jN` for an appropriate `N` to `make` for a parallel
 build.
 
+To install the build run the following:
```
Kha, line 56: I believe one reason this wasn't mentioned previously is that `elan toolchain link` (see below) mostly obsoletes the need for a global installation. Users had recurrent issues with unclean or duplicate global installations, so we might not want to advertise this option. Not sure what's more convenient on Windows.

lovettchris: Ah, right, I'm not having much luck with `elan` in an msys2 environment.  Has that been tested?  When I ran all those `elan toolchain link ...` setup steps from `index.md`, typing `lean` yields `-bash: lean: command not found`

lovettchris: Found the problem, the path "elan" prints out is wrong for msys2.  Probably should fix that somehow, but I added the right path information to msys2.md and remove the make install notes from ubuntu.md.

PR #668, opened 2021-09-15
fix msys2 windows build so the windows apps support utf-8 file paths.

lovettchris:
fix: utf-8 file paths on Windows.

You can now do this on windows:

```
lean.exe --run D:\temp\foo\英語\bar\Hello.lean the quick brown 英語
[the, quick, brown, Φï▒Φ¬₧]
```

Notice that the file path can contain unicode chars and the arguments can be passed as utf-8.

Without this fix you get this:
```
lean.exe --run D:\temp\foo\英語\bar\Hello.lean the quick brown 英語
file 'D:\temp\foo\??\bar\Hello.lean' not found
```

where `Hello.lean` contains

```lean
#lang lean4
def main (n : List String) : IO UInt32 := do
    IO.println (toString n)
    pure 0
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.10)
+cmake_minimum_required(VERSION 3.11)
 # store all variables passed on the command line into CL_ARGS so we can pass them to the stage builds
 # https://stackoverflow.com/a/48555098/161659
 # MUST be done before call to 'project'


Kha:
Ahh, much simpler :) . Modifying the MSYS2 installation during build doesn't sound very "nice" though. Should we perhaps put it in the setup instructions/`ci.yml` instead? And we still need a solution for clang/Zig (I'd be fine with having an (automatic) solution for clang *only*).

lovettchris:
Yeah, what I'd like to be able to do is modify src/shell/CMakeLists.txt like this:
```
add_library(shell OBJECT lean.cpp lean.rc)
```
this works, it does invoke windres and compile the app manifest resource, and it tries to link it but then I get this compile error:

```
bash: line 1: C:/msys64/home/clovett/git/lean4/build/release/stage0/shell/CMakeFiles/shell.dir/lean.rc.obj: cannot execute binary file: Exec format error
```
Even though the resulting lean.rc.obj is byte for byte identical to the default-manifest.o.  

So it seems I'm hitting [msys issue #454](https://github.com/msys2/MSYS2-packages/issues/454) which claims to have been fixed in a [gcc-patch](https://gcc.gnu.org/legacy-ml/gcc-patches/2014-04/msg01382.html), but I'm not seeing it.  

That's why I'm replacing the system installed default-manifest.o.  And I want to do it during build and not in the github workflow because I want it to also work for devs setting up their own msys environment without adding any addition faffing about.

clang builds seems to have this problem.

Kha:
> The `cannot execute binary file: Exec format error` seems to be a bug in our build system, this is coming from `stdlib.make`.

You can look at the executed cmdline using `make ... VERBOSE=1`. CMake is likely concatenating the objects from `shell` using `;`, we would have to replace those with spaces [like here](https://github.com/kha/lean4/blob/e9b846f443b57e0a6d6baa92924682f738625061/src/runtime/CMakeLists.txt#L13) before passing it to `leanc` [here](https://github.com/kha/lean4/blob/e9b846f443b57e0a6d6baa92924682f738625061/src/shell/CMakeLists.txt#L27).

> I do not get this error in a little cmake test app. In a little cmake test app I get `.rsrc merge failure: multiple non-default manifests`, which I reported to msys guys in a new issue [msys2/MINGW-packages#9603 (comment)](https://github.com/msys2/MINGW-packages/issues/9603#issuecomment-920359379).

Is this with clang? The cmdline you posted in that issue does not specify a compiler.

lovettchris:
Cool, also making progress here: https://github.com/msys2/MINGW-packages/issues/9603#issuecomment-920659064 so it seems we may be able to get updated version of msys that will remove the requirement on patching the default-manifest.o file...  So I won't merge this until I get that resolved.

lovettchris:
Ok, VERBOSE=1 helps, this is the log output, we see it is invoking "leanmake"...  which I think is in turn invoking "lean.mk"...?
```
make[6]: Entering directory '/home/clovett/git/lean4/src'
"/C/msys64/home/clovett/git/lean4/build/release/stage1/bin/leanmake" bin PKG=Leanpkg BIN_NAME=leanpkg.exe LEAN=" C:/msys64/home/clovett/git/lean4/build/release/stage0/bin/lean.exe" OUT="../build/release/stage1/lib" LIB_OUT="../build/release/stage1/lib/lean" OLEAN_OUT="../build/release/stage1/lib/lean" BIN_OUT="C:/msys64/home/clovett/git/lean4/build/release/stage1/bin" LEAN_OPTS+="" LEANC_OPTS+=" -O3 -DNDEBUG" LEAN_AR="C:/msys64/mingw64/bin/ar.exe" MORE_DEPS+="C:/msys64/home/clovett/git/lean4/build/release/stage0/bin/lean.exe" CMAKE_LIKE_OUTPUT=1 LINK_OPTS='-lleanshared  -pthread'
bash: line 1: C:/msys64/home/clovett/git/lean4/build/release/stage1/shell/CMakeFiles/shell.dir/manifest.rc.obj: cannot execute binary file: Exec format error
```

lovettchris:
Ok, I pushed new changes to stop modifying default-manifest.o and this part of the build fails eroneously:

```
cd C:/msys64/home/clovett/git/lean4/stage0/src && make -f C:/msys64/home/clovett/git/lean4/build/release/stage0/stdlib.make lean "LEAN_SHELL=C:/msys64/home/clovett/git/lean4/build/release/stage0/shell/CMakeFiles/shell.dir/lean.cpp.obj;C:/msys64/home/clovett/git/lean4/build/release/stage0/shell/CMakeFiles/shell.dir/manifest.rc.obj"
```
saying:
```
bash: line 1: C:/msys64/home/clovett/git/lean4/build/release/stage0/shell/CMakeFiles/shell.dir/manifest.rc.obj: cannot execute binary file: Exec format error
```

something about stdlib.make is getting confused by the contents of `manifest.rc.obj`.

Kha:
Yes, you need to replace the `;` as described above. I thought `VERBOSE=1` would also show the commands of the inner Makefile, but apparently it doesn't.

Kha:
Btw, you shouldn't need to change stage0/ here. It has its own cmake files, so as long as a change isn't needed in the Lean executable that builds stage 1, it is sufficient to change src/ only.

lovettchris:
> Btw, you shouldn't need to change stage0/ here. It has its own cmake files, so as long as a change isn't needed in the Lean executable that builds stage 1, it is sufficient to change src/ only.

well theoretically someone could clone "lean" into a unicode path, then all hell would break loose in msys when using the stage0 version of lean.exe.

leodemoura:
> > Btw, you shouldn't need to change stage0/ here. It has its own cmake files, so as long as a change isn't needed in the Lean executable that builds stage 1, it is sufficient to change src/ only.
> 
> well theoretically someone could clone "lean" into a unicode path, then all hell would break loose in msys when using the stage0 version of lean.exe.

@lovettchris We try to avoid updates to stage0 in PR's since they require extra work when merging. For example, we often have to manually rebase, build, and replay all stage0 updates in a PR because they are out of sync with other changes made in the master branch. 

I am not concerned about the scenario you have described. It will be corrected later when we perform the regular update stage0 operations.

lovettchris:
Ah, I missed your semicolon fix before, I found another way to fix that using COMMAND_EXPAND_LISTS.

lovettchris:
> We try to avoid updates to stage0 in PR's since they require extra work when merging.

so that's what I was wondering, what is the process for updating stage0?  Do we have to keep it in sync always, or is it done in separate PR's.  I've definitely seen PR's that do both in one PR.   Not sure how I see me keeping it mirrored causes extra work though?

Kha:
> Ah, I missed your semicolon fix before, I found another way to fix that using COMMAND_EXPAND_LISTS.

Ooh, nice!

Kha:
You should only update stage 0 in a PR if it is necessary for the PR to build. All other changes will find their way automatically into stage 0 eventually after merging (usually in a matter of days) when it is updated for some other change.

lovettchris:
> You should only update stage 0 in a PR if it is necessary for the PR to build. All other changes will find their way automatically into stage 0 eventually after merging (usually in a matter of days) when it is updated for some other change.

ok, I undid changes to stage0.

Kha:
Manually copying stuff into stage 0 should already be considered an advanced operation, the intended way is to always go through `update-stage0` (which *can* update arbitrary files and thus increase the chance of merge conflicts dramatically). The last time I copied `src/CMakeLists.txt` into `stage0/` manually, it broke because I needed to copy other build files changed in a previous commit as well. Manually copying single lines should usually work if one is careful, but that just seems like unnecessary hassle if there is no need to do that in a PR, and something that can easily be forgotten on further changes.

Kha:
Basically stage0/ should be considered a single binary blob, which in fact is the case in other bootstrapping schemes, and which we tell Git and GitHub to do so in the `.gitattributes` file.

lovettchris:
Ok got it, it is just "being careful" process then, sounds good.  We should write this down someplace... do we have docs for team members somewhere?

leodemoura:
> do we have docs for team members somewhere?

We have some documentation here https://github.com/leanprover/lean4/blob/master/doc/make/index.md, but we definitely need more.

lovettchris:
> > do we have docs for team members somewhere?
> 
> We have some documentation here https://github.com/leanprover/lean4/blob/master/doc/make/index.md, but we definitely need more.

Maybe I should start a 'dev' folder that includes coding guidelines and tips & tricks, etc... `make` is only one part of the overall content we need there... but having a "developer guide" is common for projects like this...

leodemoura:
> > > do we have docs for team members somewhere?
> > 
> > 
> > We have some documentation here https://github.com/leanprover/lean4/blob/master/doc/make/index.md, but we definitely need more.
> 
> Maybe I should start a 'dev' folder that includes coding guidelines and tips & tricks, etc... `make` is only one part of the overall content we need there... but having a "developer guide" is common for projects like this...

Yes, a `doc/dev` folder is a great idea.

lovettchris:
I'll do the 'dev' doc updates in a separate PR.  This PR is ready for approval.

Kha:
Hmm, did the `COMMAND_EXPAND_LISTS` change break `leanrt`?

Kha:
Could you also add a test case, e.g. adding Unicode output to `tests/lean/run/IO_test.lean`?

Kha:
We should also make sure `leanc` builds Lean programs in the same way, but that doesn't have to be in this PR

lovettchris:
> Could you also add a test case, e.g. adding output to a Unicode file path to `tests/lean/run/IO_test.lean`?

good idea, will do.

lovettchris:
> Hmm, did the `COMMAND_EXPAND_LISTS` change break `leanrt`?

I guess so, I was afraid of that as there was a merge conflict on this line of the cmake.

lovettchris:
> > Hmm, did the `COMMAND_EXPAND_LISTS` change break `leanrt`?
> 
> I guess so, I was afraid of that as there was a merge conflict on this line of the cmake.

What is weird though is on this same branch both my WSL (Ubuntu) and MSYS2 builds work just fine... ?

image

lovettchris:
> > Could you also add a test case, e.g. adding output to a Unicode file path to `tests/lean/run/IO_test.lean`?
> 
> good idea, will do.

Can you give a quick description of the test organization, what is in root level "tests", and what goes into "interactive", "Reformat", "run", "server" and "trust0".  Is "run" short for "runtime" tests perhaps?

PR #669, opened 2021-09-15
fix: space before 'at' in location

dselsam:
@Kha:
- I added helpers `formatTactic` and `parenthesizeTactic`. I may be the only one who uses these, but I use them all the time and constantly add them to my branches. It would be nice if they were in core.
- I changed "at " to " at " in the location syntax so that e.g. `rw [Nat.add_zero] at this` does not format as `rw [Nat.add_zero]at this`. Is this the best way to do this?


Kha:
Yeah, LGTM

PR #670, opened 2021-09-15
Windows: explicitly export Lean functions only

Kha:
Saves us some more symbols by not exporting our C++ functions and avoids the `--export-all-symbols` flag not supported by Zig

### DIFF
--- a/src/stdlib.make.in
+++ b/src/stdlib.make.in
@@ -28,13 +28,13 @@ endif
 # These can be phony since the inner Makefile will have the correct dependencies and avoid rebuilds
 Init:
 # Use `+` to use the Make jobserver with `leanmake` for parallelized builds
-	+"${LEAN_BIN}/leanmake" lib PKG=Init $(LEANMAKE_OPTS)
+	+"${LEAN_BIN}/leanmake" lib PKG=Init $(LEANMAKE_OPTS) LEANC_OPTS+=-DLEAN_EXPORTING
 
 Std: Init
-	+"${LEAN_BIN}/leanmake" lib PKG=Std $(LEANMAKE_OPTS)
+	+"${LEAN_BIN}/leanmake" lib PKG=Std $(LEANMAKE_OPTS) LEANC_OPTS+=-DLEAN_EXPORTING
 
 Lean: Init Std
-	+"${LEAN_BIN}/leanmake" lib PKG=Lean $(LEANMAKE_OPTS)
+	+"${LEAN_BIN}/leanmake" lib PKG=Lean $(LEANMAKE_OPTS) LEANC_OPTS+=-DLEAN_EXPORTING
 
 # the following targets are all invoked by separate `make` calls; see src/CMakeLists.txt
 


Kha:
>
```
ld.exe: ../build/stage1/lib/temp/Leanpkg.o:Leanpkg.c:(.text+0xb59c): undefined reference to `lean_dec_ref_cold'
```
Ah-ha, that's how we knows it's working! So we should mark runtime functions with `dllexport` as well. Unfortunately, it gets worse. According to https://web.archive.org/web/20140808231508/http://blogs.msdn.com/b/russellk/archive/2005/03/20/399465.aspx, we should mark functions imported from DLLs with `__declspec(dllimport)` for performance and *must* do so for imported data, e.g. 0-ary `def`s. So when emitting C code, we would need to know which imports are/will be compiled into separate DLLs and which won't :scream: . That would be a pretty significant change, but note that it is mandatory if we want back MSVC support (/cc @lovettchris), and perhaps to support Zig's future custom linker in case they don't want to implement the MinGW semantics.

lovettchris:
I'd love to get back MSVC support :-) I played with it but kept running into bash shell invocations and custom make files... (leanc.in, stdlib.make.in, etc).  but I'm happy to test what you are trying to do with this PR if you don't have msys handy.

src/shell/lean.cpp
```diff
@@ -644,7 +644,7 @@ int main(int argc, char ** argv) {
         }
 
         if (c_output && ok) {
-            std::ofstream out(*c_output);
+            std::ofstream out(*c_output, std::ios_base::binary);
```
lovettchris, line 647: what is this fixing?

stage0/src/Lean/Data/Rat.lean
```diff
@@ -0,0 +1,125 @@
+/-
```
lovettchris, line 1: this appears to be a new file about rational numbers, what does that have to do with this PR ?

stage0/stdlib/Lean/Compiler/IR/EmitC.c
```diff
@@ -585,6 +585,7 @@ lean_object* l_Lean_IR_EmitC_emitDeclAux___lambda__1(lean_object*, lean_object*,
 static lean_object* l_Nat_forM_loop___at_Lean_IR_EmitC_emitDeclAux___spec__1___closed__2;
 lean_object* l_Nat_forM_loop___at_Lean_IR_EmitC_emitDeclAux___spec__1___boxed(lean_object*, lean_object*, lean_object*, lean_object*, lean_object*);
 lean_object* l_Lean_IR_EmitC_toCType(lean_object*);
+static lean_object* l_Lean_IR_EmitC_emitDeclAux___closed__1;
```
lovettchris, line 588: I'd love to move stage0 to a git submodule so we can prune it's history from time to time... hate seeing this code generated cruft mingled with real changes like this...

lovettchris (changes requested):
couple questions...

src/shell/lean.cpp
Kha: See commit message, we should produce the same `stage0/` on every platform. `EmitC` already uses `\n` explicitly, but the `ofstream` mangled it into `\r\n`.

stage0/src/Lean/Data/Rat.lean
Kha: Please ignore all changes in `stage0/` :) . I couldn't even find your comment at first in order to respond because they are all collapsed by default.

stage0/stdlib/Lean/Compiler/IR/EmitC.c
Kha: Leo and I talked about doing this, but it would introduce more coordination headaches, especially around PRs. We'll gather some first experience with a Lake submodule (which shouldn't change as much) soon, then we can reevaluate this.

Kha:
> I'd love to get back MSVC support :-) I played with it but kept running into bash shell invocations and custom make files... (leanc.in, stdlib.make.in, etc). but I'm happy to test what you are trying to do with this PR if you don't have msys handy.

Oh right, the Makefile... the only way we could get rid of it that I can see is to bootstrap with Lake, which I don't see happening in the near future.

Kha:
Same goes for this PR, it might be easier to add support for `--export-all` to Zig than to land this one...

lovettchris:
> > I'd love to get back MSVC support :-) I played with it but kept running into bash shell invocations and custom make files... (leanc.in, stdlib.make.in, etc). but I'm happy to test what you are trying to do with this PR if you don't have msys handy.
> 
> Oh right, the Makefile... the only way we could get rid of it that I can see is to bootstrap with Lake, which I don't see happening in the near future.

On the ELL compiler project we have a build process that produces new cmake files.  You just have configure_file(foo.cmake.in) instead off "foo.make.in") Any reason we could not do that?  Instead of using Make files.  Just go one step meta, generate a cmake instead.  On linux that will generate Make files and on MSVC that will generate msbuild files.

Kha:
When would this file be regenerated? The build graph potentially changes any time a `.lean` file is touched, wouldn't that be quite costly (and screw with incremental building, not sure)?

Kha:
Nice, looks like I got all symbols now. In my previous message I forgot that dllexport and dllimport are somewhat orthogonal, so if we add only the former, we still reap all the benefits mentioned in the original message. While dllimport would still be necessary for MVSC support, this is far from the only hurdle for that as mentioned.

@leodemoura Are you happy with these changes?

src/runtime/object.h
```diff
@@ -9,6 +9,8 @@ Author: Leonardo de Moura
 #include <lean/lean.h>
 #include "runtime/mpz.h"
 
+#define EXTERN extern "C" LEAN_EXPORT
```
leodemoura, line 12: Where are we using `EXTERN`?

leodemoura:
> @leodemoura Are you happy with these changes?

Yeah, I am happy with the changes. I had just a small question (above).

src/runtime/object.h
Kha: Ah, I wanted to use this to save some space, but not all runtime cpps include this header, so I didn't use it in the end.

Kha:
I found out that clang doesn't like this version, will move the annotations from .cpp to .h soon

src/Lean/Compiler/IR/EmitC.lean
```diff
@@ -110,15 +113,15 @@ def emitFnDeclAux (decl : Decl) (cppBaseName : String) (addExternForConsts : Boo
...
   let cppBaseName ← toCName decl.name
-  emitFnDeclAux decl cppBaseName addExternForConsts
+  emitFnDeclAux decl cppBaseName isExternal
 
 def emitExternDeclAux (decl : Decl) (cNameStr : String) : M Unit := do
   let cName := Name.mkSimple cNameStr
   let env ← getEnv
   let extC := isExternC env decl.name
-  emitFnDeclAux decl cNameStr (!extC)
+  emitFnDeclAux decl cNameStr extC
```
Kha, line 124: @leodemoura It makes more sense this way around, right? We want to annotate `[extern "..."]` constants with `extern`, but not `[extern inline ...]` ones. It doesn't look like this was used before, but now that I extended the parameter to functions, it is.

Kha (commented):
Ok, final revision hopefully. Clang's requirement that the (first) declaration needs to be annotated means that we now have some annotations in headers and some (when there are no header decls) in `.cpp`s. But it shouldn't matter when applying the natural "implement new runtime function by copying existing function from same file" approach.

PR #684, opened 2021-09-22
feat: hide non-exported symbols on Unix

Kha:
After #670, we may as well be a good Unix citizen and do the equivalent thing there.


Kha:
!bench

Kha:
!bench

Kha:
`libleanshared.so` got 0.47% smaller, yay!

PR #685, opened 2021-09-23
chore: move modToFilePath.go to modToDir

digama0:
The `modToDir` function, previously only an internal recursion in `modToFilePath`, is independently useful in `lake` (see leanprover/lake#14).


Kha:
I think calling `modToFilePath` with an empty extension is the way to go. It is the perfect name for the operation, and that it uses `withExtension` instead of a potential `addExtension` as stated in the linked PR is an independent (theoretical) issue.

PR #688, opened 2021-09-24
perf: reuse idle thread in favor of spawning new one

Kha:
While debugging the Zig/Windows TLS issue, I was wondering why every single interactive test was spawning `2 * $(nproc)` threads. This change should be a good idea in general... and it might avoid triggering the TLS issue, though that is of course not a proper solution. Not sure what to do in that case, we might want to investigate whether it is still an issue during practical use.

### DIFF
--- a/src/runtime/object.cpp
+++ b/src/runtime/object.cpp
@@ -635,6 +635,7 @@ struct scoped_current_task_object : flet<lean_task_object *> {
 class task_manager {
     mutex                                         m_mutex;
     unsigned                                      m_num_std_workers{0};
+    unsigned                                      m_idle_std_workers{0};
     unsigned                                      m_max_std_workers{0};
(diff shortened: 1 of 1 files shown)


Kha:
> it might avoid triggering the TLS issue

It does not.

leodemoura:
@Kha I took a quick look at the PR. Here is my understanding. Before this commit, any Lean program creating `Task`s (in some kind of loop) would eventually spawn `m_num_std_workers` threads even if there is at most a single `Task` running at any single point. So, I think it is a good PR.
BTW, I am assuming the `2*` is because of hyper-threading, correct? That is, `hardware_concurrency()` returns `2*num_cores` in machines with hyper-threading.

leodemoura (approved):

PR #702, opened 2021-09-30
fix some syntax and link in the docs

lovettchris:
doc: various doc updates

- Fix some some old documentation notation for :keyword:, :numref: and :ref: and so on and fix some broken links.  
- Added new doc on "types" which fans out to all the various topics about types.  
- Added some examples to lexical_structure and removed some unnecessary back ticks inside triple back-ticked blocks there (which may also have been a carry over from other doc formats).  
- Added missing documentation of `\uxxxx` syntax.  
- Added missing documentation on Floating point literal syntax.
- Added mention of new windows friendly `elan-init.ps1` install script.

### DIFF
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -3,6 +3,13 @@
 These instructions will walk you through setting up Lean using the "basic" setup and VS Code as the editor.
 See [Setup](./setup.md) for other ways and more details on setting up Lean.
 
+1. Install the latest Lean 4 nightly through [`elan`](https://github.com/leanprover/elan): in any bash-compatible shell, run
+    ```sh
+    $ curl https://raw.githubusercontent.com/leanprover/elan/master/elan-init.sh -sSf | sh -s -- --default-toolchain leanprover/lean4:nightly
+    ```
+    Alternatively, on Windows, download [the latest release](https://github.com/leanprover/elan/releases/latest/download/elan-x86_64-pc-windows-msvc.zip) and run the contained `elan-init.exe`.
+    See the [elan repo](https://github.com/leanprover/elan) for other installation options and details.
+    
 1. Install [VS Code](https://code.visualstudio.com/).
 
 1. Open VS Code and install the `lean4` extension.


Kha:
Note that it's not clear to me if we should even keep the "Lexical Structure" chapter (in its current form) since Lean 4's approach is so radically different from Lean 3's, but this is certainly an improvement.

lovettchris:
> Note that it's not clear to me if we should even keep the "Lexical Structure" chapter (in its current form) since Lean 4's approach is so radically different from Lean 3's, but this is certainly an improvement.

Every new programming language needs to explain the basic syntax to new users.  I found this kind of information critical to understanding how to read lean code.     What is your concern?  Is it because the syntax is also extensible?  We could add more of a pointer to that concept in this chapter then point to syntax.md.

doc/types.md
```diff
@@ -0,0 +1,59 @@
+# Types
+
+Every programming language needs a foundation in sound type theory.
+Lean has a rich provable type theory that is built on type universes.
+
+TODO: add a profound insight into why type universes are useful -
```
Vtec234, line 6: As far as I understand, type universes are mostly there to avoid self-referential paradoxes like Russell's paradox.

doc/types.md
```diff
@@ -0,0 +1,59 @@
+# Types
+
+Every programming language needs a foundation in sound type theory.
+Lean has a rich provable type theory that is built on type universes.
```
Vtec234, line 4: I'm not sure what it means for a type theory to be provable.

lovettchris: Ok, I'm just trying to say something interesting, how would you describe the Lean type system and how it is different from other programming languages?

Vtec234: I understand, and to be honest I don't think there's one good answer! But other parts of the manual already prove motivation for the type system, for example [here](https://leanprover.github.io/lean4/doc/decltypes.html). So I think it may well be okay to say nothing here and just point the reader at those other sections.

Vtec234: Btw, since universes are mostly a distraction for a beginner, if the language manual is meant to be somewhat of an introduction rather than an exhaustive reference, and if `types.md` is meant to be the entrypoint into documentation about types, I would put the universe syntax in one of the more detailed type theory sections.

doc/types.md
```diff
@@ -0,0 +1,59 @@
+# Types
+
+Every programming language needs a foundation in sound type theory.
```
tydeu, line 3: This is rather ***strong*** statement considering there are many widely used programming languages without types. And even those with types are rarely concerned about the soundness of their type theory. For example, Haskell's type theory is well known to be unsound, but it is fine with that.

tydeu: A precise statement might be "every theorem prover" or "strongly types programming language with integrated correctness proofs".

Kha:
Maybe move (significant) additions of new text into a separate PR to move this one forward?

doc/lexical_structure.md
```diff
@@ -96,32 +106,45 @@ Numeric Literals
 Numeric literals can be specified in various bases.
 
 ```
-   numeral    : `numeral10` | `numeral2` | `numeral8` | `numeral16`
-   numeral10  : [0-9]+
+   numeral    : numeral10 | numeral2 | numeral8 | numeral16
+   numeral10  : [-] [0-9]+
```
Kha, line 110: The `-` is not part of the token; it's just the regular `Neg.neg`

doc/lexical_structure.md
```diff
@@ -96,32 +106,45 @@ Numeric Literals
...
    numeral8   : "0" [oO] [0-7]+
-   numeral16  : "0" [xX] `hex_char`+
+   numeral16  : "0" [xX] hex_char+
+```
+
+Decimal literals are also possible:
+
 ```
+   decimal    : [-] [0-9]+ "." [0-9]+
+```
```
Kha, line 120: This seems to be subsumed by float literals

doc/syntax.md
```diff
@@ -1,16 +1,27 @@
 # Syntax Extensions
 
-Lean's syntax can be extended and customized by users at every level, ranging from basic "mixfix" notations to custom elaborators.
-In fact, all builtin syntax is parsed and processed using the same mechanisms and APIs open to users.
-In this section, we will describe and explain the various extension points.
-Significant syntax extensions already builtin into Lean such as the [`do` notation](./do.md) are described in subsections.
+[Lean's syntax](lexical_structure.md) can be extended and customized
```
Kha, line 3: Before reformatting existing text, we should agree on a standard format and ideally enforce it in editors.

PR #707, opened 2021-10-02
Start adding documentation for metaprogramming

bollu:
I wish to write down the examples of using LEAN4 macros that I feel are useful to get up and running. This is part 1 which parses the simplest non-trivial grammar I can think of. 

I know very little of how the macro system works internally, so corrections and elaborations would be very useful!

### DIFF
--- a/doc/metaprogramming.md
+++ b/doc/metaprogramming.md
@@ -35,7 +35,7 @@ This names our grammar and allows us to specify parsing rules associated with ou
 declare_syntax_cat brack
 ```
 
-Next, we specify the grammar using the `syntax <parse rule>` directive:
+Next, we specify the grammar using the `syntax <parse rule>` command:
 
 ```lean
 syntax "End" : brack


leodemoura:
Thanks a bunch for writing it down. I have only a few comments.

doc/metaprogramming.md
```diff
@@ -1 +1,139 @@
 # Metaprogramming
+
+Macros are a language feature that allows writing code that writes other code (metaprogramming).
+
+In LEAN4, macros are used pervasively. So much so that core language features such as `do` notation
```
leodemoura, line 5: We use "Lean 4" is the rest of the documentation.

doc/metaprogramming.md
```diff
@@ -1 +1,139 @@
...
+is implemented via macros! As a language user, macros are useful to easily
+embed domain-specific languages and to generate code at compile-time, to name a few uses.
+
+
+#### Balanced Parentheses as an Embedded Domain Specific Language
+
+Let's look at how to use macros to extend the LEAN parser and embed a language for building _balanced parentheses_.
```
leodemoura, line 12: "LEAN" => "Lean"

doc/metaprogramming.md
```diff
@@ -1 +1,139 @@
...
+We begin by defining an inductive data type of the grammar we wish to parse:
+
+```lean
+inductive Dyck: Type :=
+   | Round : Dyck -> Dyck -- ( <inner> )
+   | Flower : Dyck -> Dyck -- { <inner> }
+   | End : Dyck
+```
+
+We begin by declaring a _syntax category_ using the `declare_syntax_cat <category>` directive.
```
leodemoura, line 31: "directive" => "command"

doc/metaprogramming.md
```diff
@@ -1 +1,139 @@
...
+The above means that the string "End" lives in syntax category `brack`.
+
+Similarly, we declare the rules `"(" Dyck ")"` and `"{" Dyck "}"` using the rules:
+
+```lean
+syntax "(" brack ")" : brack
+syntax "{" brack "}" : brack
+```
+
+Finally, we need a way to build _LEAN4 terms_ from this grammar -- that is, we must translate out of this
```
leodemoura, line 53: "LEAN4" => "Lean 4"

Kha:
Note that this will not be visible unless added to SUMMARY.md. how should this be differentiated from syntax.md? Should we have a new folder/section for tutorials?

leodemoura:
> Should we have a new folder/section for tutorials?

I think this is a great idea.

doc/metaprogramming.md
```diff
@@ -1 +1,139 @@
...
+#### Balanced Parentheses as an Embedded Domain Specific Language
+
+Let's look at how to use macros to extend the Lean 4 parser and embed a language for building _balanced parentheses_.
+This language accepts strings given by the [BNF grammar](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form)
+
+```
+Dyck := 
+  "(" Dyck ")" 
+  | "{" Dyck '}'
+  | "End"
```
Kha, line 19: This token looks like a hack. How can we make the grammar more conventional? We could use `(brack)?`, but that is not as nice in BNF. An epsilon production would be more common there, but we don't have a direct encoding of that as a `ParserDescr` yet. Should we?

doc/metaprogramming.md
```diff
@@ -1 +1,139 @@
...
+```
+
+We begin by defining an inductive data type of the grammar we wish to parse:
+
+```lean
+inductive Dyck: Type :=
+   | Round : Dyck -> Dyck -- ( <inner> )
+   | Flower : Dyck -> Dyck -- { <inner> }
```
Kha, line 27: I don't think I've ever heard of flower brackets :) . Sounds very poetic though. Is it a common name?

doc/metaprogramming.md
bollu: Having a `?` would be very useful!

doc/metaprogramming.md
bollu: That's what I was taught as a child. It appears to be a colloquialism :) I'll be happy to change it to curly brackets if you'd like!

doc/metaprogramming.md
Kha: We could also modify the `syntax` command to accept an empty sequence of parsers, but an explicit epsilon parser might be the better idea

doc/metaprogramming.md
Kha: Yeah, feel free to change this file in the new PR so we have both files side by side

Issue #494, opened 2021-05-28
Pretty printer fails on syntax containing incQuotDepth


gebner:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

In custom quoting libraries, the `incQuotDepth` parser combinator is useful so that you can use `$x` for antiquotations.  However pretty-printing syntax that uses the `incQuotDepth` parser combinator fails.

### Steps to Reproduce

```lean
import Lean

constant f : Nat → Nat

syntax "f(" incQuotDepth(term) ")" : term

open Lean PrettyPrinter.Delaborator in
@[delab app.f] def delabF : Delab := do
  guard $ (← getExpr).getAppNumArgs == 1
  `(f($(← withAppArg delab)))

set_option pp.rawOnError true in
#check f 42
-- [Error pretty printing expression: parser 'incQuotDepth' was not found. Falling back to raw printer.]
-- f (OfNat.ofNat.{0} Nat 42 (instOfNatNat 42)) : Nat
```

I expected `f(42)` as output.

### Versions

Nightly from May 26.

PR #708, opened 2021-10-04
Add tutorial on parsing the arithmetic grammar using macros

bollu:
This is part 2, which covers a more complicated grammar, using identifiers, and "escaping" the macro context.

This also starts a new `doc/tutorial` subfolder. How should I link these with the manual?

### DIFF
--- a/doc/tutorial/metaprogramming-arith.md
+++ b/doc/tutorial/metaprogramming-arith.md
@@ -100,39 +100,22 @@ And some examples:
 ```lean,ignore
 #check [Arith| "x" * "y"] -- Arith.mul (Arith.symbol "x") (Arith.symbol "y") : Arith
 
-def bar := [Arith| "x" + "y"] -- add
-#print bar
-/-
-def bar : Arith :=
-Arith.add (Arith.symbol "x") (Arith.symbol "y")
--/
+#check [Arith| "x" + "y"] -- add
+-- Arith.add (Arith.symbol "x") (Arith.symbol "y") 
 
-def baz := [Arith| "x" + 20] -- symbol + int
-#print baz
-/-
-def baz : Arith :=
-Arith.add (Arith.symbol "x") (Arith.int 20)
--/
+#check  [Arith| "x" + 20] -- symbol + int
+-- Arith.add (Arith.symbol "x") (Arith.int 20)
 
 
-def quux_left := [Arith| "x" + "y" * "z" ] -- precedence
-#print quux_left
-/-
-def quux_left : Arith :=
+#check [Arith| "x" + "y" * "z" ] -- precedence
 Arith.add (Arith.symbol "x") (Arith.mul (Arith.symbol "y") (Arith.symbol "z"))
--/
-
-def quux_right := [Arith| "x" * "y" + "z"] -- precedence
-#print quux_right
+-- 
+#check [Arith| "x" * "y" + "z"] -- precedence
 -- Arith.add (Arith.mul (Arith.symbol "x") (Arith.symbol "y")) (Arith.symbol "z")
 
 
-def quuz := [Arith| ("x" + "y") * "z"] -- brackets
-#print quuz
-/-
-def quuz : Arith :=
-Arith.mul (Arith.add (Arith.symbol "x") (Arith.symbol "y")) (Arith.symbol "z")
--/
+#check [Arith| ("x" + "y") * "z"] -- brackets
+-- Arith.mul (Arith.add (Arith.symbol "x") (Arith.symbol "y")) (Arith.symbol "z")
 ```
 
 
@@ -154,31 +137,16 @@ macro_rules
 Let's test and see that we can now write expressions such as `x * y` directly instead of having to write `"x" * "y"`:
 
 ```lean,ignore
-def x_ident := [Arith| x ]
-#print x_ident
-/-
-def x_ident : Arith :=
-Arith.symbol "x"
--/
+#check [Arith| x ] -- Arith.symbol "x"
 
-def x_plus_y : Arith := [Arith| x + y]
-#print x_plus_y
-/-
-def x_plus_y : Arith :=
-Arith.add (Arith.symbol "x") (Arith.symbol "y")
--/
+#check [Arith| x + y] -- Arith.add (Arith.symbol "x") (Arith.symbol "y")
 ```
 
 We now show an unfortunate consequence of the above definitions. Suppose we want to build `(x + y) + z)`.
 Since we already have defined `x_plus_y` as `x + y`, perhaps we should reuse it! Let's try:
 
 ```lean,ignore
-def x_plus_y_plus_z := [Arith| x_plus_y + z]
-#print x_plus_y_plus_z
-/-
-def x_plus_y_plus_z : Arith :=
-Arith.add (Arith.symbol "x_plus_y") (Arith.symbol "z")
--/
+#check [Arith| x_plus_y + z] --Arith.add (Arith.symbol "x_plus_y") (Arith.symbol "z")
 ```
 
 Whoops, that didn't work! What happened? Lean treats `x_plus_y` _itself_ as an identifier! So we need to add some syntax
@@ -195,12 +163,7 @@ macro_rules
 Let's try our previous example:
 
 ```lean,ignore
-def x_plus_y_plus_z2 := [Arith| <[ x_plus_y ]>]
-#print x_plus_y_plus_z2
-/-
-def x_plus_y_plus_z2 : Arith :=
-x_plus_y
--/
+#check [Arith| <[ x_plus_y ]>] -- x_plus_y
 ```
 
 Perfect!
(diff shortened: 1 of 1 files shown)


doc/tutorial/metaprogramming-arith.md
```diff
@@ -0,0 +1,290 @@
...
+#print x_plus_y_plus_z
+/-
+def x_plus_y_plus_z : Arith :=
+Arith.Add (Arith.Symbol "x_plus_y") (Arith.Symbol "z")
+-/
+```
+
+Whoops, that didn't work! What happened? Lean treats `x_plus_y` _itself_ as an identifier! So we need to add some syntax
+to be able to "escape" the `fromArith%` context. Let's use the syntax `<[ $e:term ]>` to mean: evaluate `$e` as a real term,
+not an identifier. The macro looks like follows:
```
Kha, line 155: It is [possible, though not trivial](https://github.com/gebner/quote4/commit/c700ec841dc0d6da08eedc3dde7b4d85c3320de5), to reuse Lean's own antiquotation syntax for DSLs. I wonder if we could make that simpler, e.g. by hardcoding support for a new `$e:antiquot` antiquotation kind. /cc @gebner

doc/tutorial/metaprogramming-arith.md
```diff
@@ -0,0 +1,290 @@
...
+
+Perfect!
+
+In this tutorial, we expanded on the previous tutorial to parse a more
+realistic grammar with multiple levels of precedence, how to parse identifiers directly
+within a macro, and how to provide an escape from within the macro context.
+
+#### Full code listing
+
+```lean,ignore
```
Kha, line 183: I accidentally marked the full listing as ignored as well, but we should have at least those checked by CI

doc/tutorial/metaprogramming-arith.md
```diff
@@ -0,0 +1,290 @@
...
+We also write a macro rule for `fromArith%` that translates an identifier into
+a string, using `$(Lean.quote (toString x.getId))`.  (TODO: explain what
+`Lean.quote` does):
+
+```lean,ignore
+-- Have to use french quotes («arith») to refer to `arith` as
+-- `arith` is a keyword after declare_syntax_cat.
+-- We declare identifiers as living in the `arith` syntax category, and
+-- prove an interpretation for them with `fromArith%`
+syntax ident : «arith»
```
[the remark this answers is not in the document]
bollu: Sorry, I am unsure which part is not necessary --- The declaration `syntax ident : «arith»`? Or describing this in the tutorial?

doc/tutorial/metaprogramming-arith.md
```diff
@@ -0,0 +1,290 @@
...
+Let's parse another classic grammar, the grammar of arithmetic expressions with
+addition, multiplication, integers, and variables.  In the process, we'll learn
+how to:
+
+- Convert identifiers such as `x` into strings within a macro.
+- Add the ability to "escape" the macro context from within the macro. This is useful to interpret identifiers with their _original_ meaning (predefined values)
+  instead of their new meaning within a macro (treat as a symbol).
+
+Let's begin with the simplest thing possible. We'll define an AST, and use operators `:+` and `:*` to denote
+building an arithmetic AST. We use `:+` and `:*` instead of `+` and `*` to avoid clashing names.
```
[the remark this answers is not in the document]
bollu: [I asked about this on zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.E2.9C.94.20macro.20v.2Fs.20typeclass.20resolution/near/255492626). As I understand it, there's some ambiguity introduced by the rule:

```lean
syntax term : arith -- int for Arith.Int
```

I suppose I ought to change the rule to 

```
syntax num : arith -- strings for Arith.Int
```

to be more precise about the input `fromArith|` expects, so I'll do that.

doc/tutorial/metaprogramming-arith.md
```diff
@@ -0,0 +1,290 @@
...
+```lean,ignore
+declare_syntax_cat arith
+syntax term : arith -- int for Arith.Int
+syntax str : arith -- strings for Arith.Symbol
+syntax:50  arith ":+" arith : arith -- Arith.Add
+syntax:60 arith ":*" arith : arith -- Arith.Mul
+syntax "(" arith ")" : arith -- bracketed expressions
+```
+
+We define the macro `fromArith%` to translate the syntax category `arith` into an `Arith` inductive value that lives in `term`:
```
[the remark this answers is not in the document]
bollu: Done, thanks for the stylistic note!

doc/tutorial/metaprogramming-arith.md
bollu: Done

doc/tutorial/metaprogramming-arith.md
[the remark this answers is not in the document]
Kha: Apologies, I meant the part about quoting. In the Zulip code, there was a macro named `arith`, but that's not the case here.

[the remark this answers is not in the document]
bollu: Ah, I think I understand. One can say `syntax ident : arith` directly.

doc/tutorial/metaprogramming-arith.md
```diff
@@ -0,0 +1,290 @@
...
+Let's begin with the simplest thing possible. We'll define an AST, and use operators `:+` and `:*` to denote
+building an arithmetic AST. We use `:+` and `:*` instead of `+` and `*` to avoid clashing names.
+
+
+Here's the AST that we will be parsing:
+
+```lean,ignore
+-- example on parsing arith language via macros
+inductive Arith: Type :=
+   | Add : Arith -> Arith -> Arith -- e :+ f
```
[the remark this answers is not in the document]
bollu: Done, changed to lowercase

doc/tutorial/metaprogramming-arith.md
```diff
@@ -0,0 +1,290 @@
...
+   | Mul : Arith -> Arith -> Arith -- e :* f
+   | Int : Int -> Arith -- constant
+   | Symbol : String -> Arith -- variable
+```
+
+We declare a syntax category to describe the grammar that we will be parsing.
+See that we control the precedence of `:+` and `:*` by writing `syntax:50` for addition and `syntax:60` for multiplication,
+indicating that multiplication binds tighter than addition (higher the number, tighter the binding).
```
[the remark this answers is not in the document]
bollu: I've fixed the notation, expanded the explanation and added a link to the Lean docs for further information. I hope this is better now!

doc/tutorial/metaprogramming-arith.md
```diff
@@ -0,0 +1,333 @@
...
+Let's begin with the simplest thing possible. We'll define an AST, and use operators `+` and `*` to denote
+building an arithmetic AST.
+
+
+Here's the AST that we will be parsing:
+
+```lean,ignore
+-- example on parsing arith language via macros
+inductive Arith: Type where
+ | add : Arith -> Arith -> Arith -- e + f
```
Kha, line 20:
```suggestion
inductive Arith : Type where
  | add : Arith -> Arith -> Arith -- e + f
```
one more try :)

doc/tutorial/metaprogramming-arith.md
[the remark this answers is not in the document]
Kha: I literally meant ``"`[Arith|" arith "]"`` :) . See, for example, https://github.com/IPDSnelting/tba-2021/blob/b6390e55b768423d3266969e81d19290129c5914/TBA/While/Com.lean#L58. I'm afraid the current `Arith|` notation is confusing because inside of quotations it looks like a "quotation kind" (e.g. `` `(level| ...) ``).

doc/tutorial/metaprogramming-arith.md
Kha: Oh, this time I forgot the Unicode arrows myself

doc/tutorial/metaprogramming-arith.md
```diff
@@ -0,0 +1,333 @@
...
+-- a * b + c
+a * (b:60 + c:61):60
+(a:70 * b:71):70 + c
+```
+
+While parsing `a * (b + c)`, `(b + c)` is assigned a precedence `60` by the addition rule. However, multiplication expects
+the right argument to have precedence **at least** 71. Thus, this parse is invalid. In contrast, `(a * b) + c` assigns
+a precedence of `70` to `(a * b)`. This is compatible with addition which expects the left argument to have precedence
+**at least `60` ** (`70` is greater than `60`). Thus, the string `a * b + c` is parsed as `(a * b) + c`.
+For more details, please look at the [Lean manual on syntax extensions](https://leanprover.github.io/lean4/doc/syntax.html#notations-and-precedence).
```
Kha, line 73:
```suggestion
For more details, please look at the [Lean manual on syntax extensions](../syntax.md#notations-and-precedence).
```

doc/tutorial/metaprogramming-arith.md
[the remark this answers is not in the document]
bollu: Ah, I see :) Okay, I updated the notation to  `[Arith| ... ]`

doc/tutorial/metaprogramming-arith.md
bollu: The [latest commit]( https://github.com/bollu/lean4/blob/b152e9da2b81952403ecf6ee06640433c5259fe0/doc/tutorial/metaprogramming-arith.md?plain=1#L17-L23) should hopefully be correct  (spacing of ` : `, 2-spaces before `|`, and unicode arrows)

Kha (approved):
Thanks for bearing with me!

doc/tutorial/metaprogramming-arith.md
gebner: This is actually much easier than it looks like!  The biggest obstacles were issue #494 and figuring out how to set up the syntax so that `q($x)` parses.  If you're looking for things to improve:
1. Matching on antiquotations could be easier, I'm not sure if I like the category annotation (because it usually specifies the category and not a node).  But writing `` `($[$x]:antiquot) => ... `` would be much more ergonomic than what we have right now.
1. These two-part syntaxes (`do`/`←` in particular, but less so with `` `()``/`$`) are in general awkward to extend because a lot of code hard-codes the other part of the syntax.  (Indeed the really horrible nontrivial part of quote4 is the part which extends match...)
1. I was under the impression `$x` and `$(x)` parse to the same syntax, but `$((x))` does not.  This is because `getAntiquotTerm` returns the same result for the first two syntaxes.  It would be great if pattern-matching could distinguish `$x` and `$(x)`.
1. In the same vein, `$x.y` and `$x.1` surprisingly parse differently.
1. I'm not 100% sure that I implemented escaped antiquotations correctly.  There's very little high-level documentation on how to abuse antiquotations so I've been mostly guessing.  Presumably I need to add an `unescapeAntiquot` to support `$$$$$$$$$$$$x`.
1. Speaking of lots of dollars, it might be easier to understand and easier to write macros if they're counted the other way around.  That is, `$$$$$x` generates five nested antiquotation nodes, and you can write a macro like `` `(fromArith% $[$x]:antiquot) => return x``.

Some of these issues could be solved with better documentation so I think it would be good to introduce antiquotations already here.  That is, make it so that `fromArith% $x + 1` works.

doc/tutorial/metaprogramming-arith.md
```diff
@@ -0,0 +1,326 @@
...
+```
+And some examples:
+
+```lean,ignore
+def foo := [Arith| "x" * "y" ] -- mul
+#print foo
+/-
+def foo : Arith :=
+Arith.mul (Arith.symbol "x") (Arith.symbol "y")
+-/
```
gebner, line 106: This is a very long-winded and repetitive way to show the output of the macros.  I think a `#check` would be easier to read:

```suggestion
#check [Arith| "x" * "y"] -- Arith.mul (Arith.symbol "x") (Arith.symbol "y") : Arith
```

doc/tutorial/metaprogramming-arith.md
```diff
@@ -0,0 +1,326 @@
...
+
+def x_plus_y : Arith := [Arith| x + y]
+#print x_plus_y
+/-
+def x_plus_y : Arith :=
+Arith.add (Arith.symbol "x") (Arith.symbol "y")
+-/
+```
+
+We now show an unfortunate consequence of the above definitions. Suppose we want to build `(x :+ y) :+ z)`.
```
Vtec234, line 177:
```suggestion
We now show an unfortunate consequence of the above definitions. Suppose we want to build `(x + y) + z)`.
```

doc/tutorial/metaprogramming-arith.md
```diff
@@ -0,0 +1,326 @@
...
+
+```lean,ignore
+syntax ident : arith
+
+macro_rules
+  | `([Arith| $x:ident]) => `(Arith.symbol $(Lean.quote (toString x.getId)))
+```
+
+
+Let's test and see that we can now write expressions such as `x :* y` directly instead of having to write `"x" :* "y"`:
```
Vtec234, line 159:
```suggestion
Let's test and see that we can now write expressions such as `x * y` directly instead of having to write `"x" * "y"`:
```

Vtec234 (changes requested):

doc/tutorial/metaprogramming-arith.md
bollu: Reminder to self: change all the examples to `#check`.

PR #712, opened 2021-10-04
Improve do-notation error message when monad is unknown.

bollu:
The patch also eliminates the unused `hasBindInst` field of `ExtractMonadResult`.

Comparison of error messages:

-------------------------

```lean
partial def foo (u: Unit) : OptionM Nat := do
  return $ 42
```

- Before: `error: invalid 'do' notation, expected type is not available`
- After:
```
error: function expected at
  pure PUnit.unit
term has type
  Id PUnit
```
-------------------------

```lean
partial def bar (i: Int) := do
  return k
```

- Before: `do.lean:1:28: error: invalid 'do' notation, expected type is not available`
- After:  `do.lean:2:9: error: unknown identifier 'k'`

---------------------------

```lean
partial def bar (i: Int) := do
  return _
```

- Before: `error: invalid 'do' notation, expected type is not available`
- After: 
```
do.lean:2:2: error: typeclass instance problem is stuck, it is often due to metavariables
         Pure (?m.59 I)
```

------

Caveat emptor: I don't fully understand what `Do.lean` does. I tried to do the sane thing, which is to generate metavariables if we are unable to infer the monad type. I find the new error messages more precise, but perhaps this idea of generating metavariables was ruled out due to other concerns [eg. performance, or unsoundness?] I hesitated sending in the patch, since I wasn't sure if the unused `hasBindInst`  is a placeholder for further development of `do`-notation elaboration.

### DIFF
--- a/src/Lean/Elab/Do.lean
+++ b/src/Lean/Elab/Do.lean
@@ -87,7 +87,6 @@ private partial def hasLiftMethod : Syntax → Bool
 structure ExtractMonadResult where
   m            : Expr
   α            : Expr
-  hasBindInst  : Expr
   expectedType : Expr
   isPure       : Bool -- `true` when it is a pure `do` block. That is, Lean implicitly inserted the `Id` Monad.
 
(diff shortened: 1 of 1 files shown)


leodemoura:
Regarding the error message for
```lean
partial def foo (u: Unit) : OptionM Nat := do
  return $ 42
```
I think the new error message is still confusing. The key issue here is that `return` is not a function like in Haskell, but a notation. We can add code for producing a nice error message in this case. This should be useful since many users with Haskell experience will probably hit this issue. I can see them writing `return $ x + 1` instead of `return x + 1`.

PR #713, opened 2021-10-05
chore: update to LLVM 13

Kha:


Kha:
!bench

Kha:
Some ups and downs as usual, but the most significant change isn't measured (because it would take too long): compiling Lean without a hot ccache cache is now 21% faster on my machine [thanks to the new pass manager](https://blog.llvm.org/posts/2021-03-26-the-new-pass-manager/) :tada:

PR #718, opened 2021-10-08
Option section

paulch42:
Proposal for the `Option` section of the Lean 4 manual


leodemoura:
Thanks for the PR, but we are still trying to figure out how to organize the documentation.

PR #723, opened 2021-10-11
Widgets: `ToHtmlFormat`

### LINKED ISSUE #680: Widgets: `ToHtmlFormat`
As some of the Lean 3 widget features are [rolling in](https://github.com/leanprover/vscode-lean4/pull/37), we should decide on the next step. I think a good one would be to provide a typeclass which will allow users to visualize types as HTML. Specifically, we would have a `Html` type with something like
```lean
class ToHtmlFormat (α : Type u) where
  formatHtml : α → Html
```

On encountering a value of type which implements `ToHtmlFormat`, the infoview would render it as the HTML output. This is not yet the fully programmable widgets feature, but I think it will go a long way towards visualizations such as Dynkin diagrams ([zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Dynkin.20diagrams)).

Vtec234:
Closes #680. Besides the typeclass itself, we also provide a JSX-like syntax adapted from the webserver test to more easily write HTML.

### DIFF
--- a/src/Lean/Server/FileWorker/WidgetRequests.lean
+++ b/src/Lean/Server/FileWorker/WidgetRequests.lean
@@ -54,31 +54,17 @@ builtin_initialize
           exprExplicit?
           doc? := ← i.info.docString? : InfoPopup }
 
-open Meta in
-/-- Return an instance of `ToHtmlFormat tp` if there is one, otherwise `none`. -/
-private def hasToHtmlFormat? (tp : Expr) : MetaM (Option Expr) := do
-  let clTp ←
-    try
-      mkAppM ``ToHtmlFormat #[tp]
-    catch ex =>
-      throwError "failed to construct 'ToHtmlFormat {tp}':\n{ex.toMessageData}"
-  match (← trySynthInstance clTp) with
-  | LOption.some r => some r
-  | _ => none
-
 /-- Try to find an instance of `ToHtmlFormat` for the type of `val` and use it
 to produce HTML for the value. Return the HTML on success, `none` on failure. -/
 private unsafe def evalToHtmlFormatUnsafe? (val : Expr) : MetaM (Option Html) := do
-  let tp ← Meta.inferType val
-  if (← hasToHtmlFormat? tp).isNone then
-    return none
-  let n ← mkFreshUserName `_htmlOut
   let htmlOut ←
     try Meta.mkAppM ``ToHtmlFormat.formatHtml #[val]
     -- Return `none` if the expression is meaningless because, for example,
-    -- `ToHtmlFormat` isn't imported.
+    -- `ToHtmlFormat` isn't imported, or there is no instance for the type
+    -- of `val`.
     catch ex => return none
 
+  let n ← mkFreshUserName `_htmlOut
   let decl := Declaration.defnDecl {
     name        := n
     levelParams := []


gebner:
I would be much more enthusiastic about this feature if we had concrete ideas on how to make use of it.  It adds a very specific extension point that doesn't fit well with any of the applications I've seen for widgets so far.  There are roughly two general applications I know of:

1. Goal state visualization, like the popular Sudoku demo.  These add a widget to the whole proof state, and not to a type.  In fact, the API in this PR is not strong enough to support this particular demo (IIRC, the sudoku state is stored as one hypothesis per field).
2. Showing HTML trace messages from Lean.  Like the (cool) slides you had on Wednesday (sorry I couldn't come but it was almost midnight here).  Or like the graphical plots that Rob generated with the Mathematica interface.  Of course you can simulate a trace message by generating a dummy term and then hovering over it ... but that's unnecessarily complicated.

In the slides you cite the Alectryon paper, and I can see where you got the inspiration from.  However I think that this kind of visualization would make much more sense as an extension to `#eval` (implemented using a general HTML trace API that could also support other applications in category 2).  Note that the examples in the Alectryon paper are exclusively for the Coq equivalent of `#eval`.

I am not sure how useful an `#eval`-command for subterms of the tactic state actually is in practice.  In a proof, you rarely see `mkBinaryTree [1:6]`, instead you have `mkBinaryTree (xs ++ ys)` or something (to connect to one of the examples in the Alectryon paper).

Vtec234:
> Showing HTML trace messages from Lean. Like the (cool) slides you had on Wednesday (sorry I couldn't come but it was almost midnight here). Or like the graphical plots that Rob generated with the Mathematica interface. Of course you can simulate a trace message by generating a dummy term and then hovering over it ... but that's unnecessarily complicated.

The hovering UI is a bit of a red herring that I introduced only because it was the simplest way to display HTML. I have no intention to actually have the UX be this. We could definitely have a pane in the tactic state as well as a `#html` command. But I think we can do that in a follow-up PR. Here, let's just figure out how to transform terms to HTML. Which brings me to..

> In a proof, you rarely see mkBinaryTree [1:6], instead you have mkBinaryTree (xs ++ ys) or something

This point about open terms is a very good one. I think we should support them. On the other hand, I want to try the hardest we can to avoid meta stuff. Meta stuff is intimidating for the majority of non-power users. Of course we should add fully expressive meta-visualization as well, but this part of the widgets abstraction is explicitly meant to be easy to use if you've only ever written definitions and proofs. I wonder whether we could introduce partial evaluation magic to transform pure definitions into ones which work on, say `node red x y` where `x,y` are subtrees. Concretely, suppose `formatHtml (node red x y) = <circle red><left-edge>{formatHtml x}</left-edge><right-edge>{formatHtml y}>/right-edge</circle>`. Then we could replace calls to `formatHtml` on variables with a `CodeWithInfo` object which shows the variable. I think this could be really cool.

cpehle:
There is one pretty obvious use case for an extension like this: To support interactive plotting / diagram generation. Right now the vscode extension (probably prudently) escapes the string it renders. Without that it already is easy now to create svg based plots that dynamically update using ```#eval``` (i wrote some experimental code to do that, inspired by d3). This is really relevant especially for data / ml type applications.

gebner:
I completely forgot about this PR.  I believe we have already talked a bit about this since then, let me repeat it here.

> The hovering UI is a bit of a red herring that I introduced only because it was the simplest way to display HTML. I have no intention to actually have the UX be this. We could definitely have a pane in the tactic state as well as a #html command. But I think we can do that in a follow-up PR.

This PR already commits to this UX though: the HTML is added to the info popup (at least via separate RPC call).

If the plan is to add additional RPC calls to get HTML fragments for more objects (like diagnostics for example), then it would be good to make the API future-proof so that we can easily add support for interactive HTML fragments (that produce React components and make RPC calls) without changing all of diagnostics/popups/etc. again.

> Here, let's just figure out how to transform terms to HTML. Which brings me to..

I think that's the most controversial aspect of this PR.  The proposed mechanism is to evaluate every closed subterm in the VM if you open it in a popup, and pretty-print the evaluated VM object using a type class.  In other words, automatic `#html_eval` for every popup.

1. We don't want to call `#eval` on every subterm.  The expression may require a lot of runtime/memory to compute.  Evaluating the expression may also have side-effects (like panics, stack overflows, etc.) or even crash Lean.
2. Evaluating the subterm loses information.  You might want to print `42 / 100` differently than `0`.
3. Evaluation is not possible for 99% of mathlib, because the definitions depend on classical axioms.  Even if by pure chance some objects can be evaluated, it might be impossible to obtain a string representation from the VM object (e.g. `42 : ℝ` is computable, but all computable functions `ℝ → String` are constant).
4. Similarly, propositions and types cannot be htmlified.
5. Relatedly, evaluation doesn't work for open terms.  But I've already mentioned that.

The design in this PR with the `ToHtmlFormat` class is great for an `#html_eval` command, but not for visualization of terms in the tactic state.  (I'm not against an "eval" button in a popup though, but that should be a conscious action and also not the only way to print HTML.)

To make at least one productive suggestion, for the UI proposed in this PR I think it would be better to register HTML formatters for subterms using an attribute (though we might have to change the API for interactive stuff):
```lean
@[htmlFormatter] def printMyGadget : Expr → MetaM Html := ...
```

Maybe it would be easier to start with `#html_eval` though, there the existing ToHtmlFormat API is a good fit and there is also less uncertainty about the UX.

> On the other hand, I want to try the hardest we can to avoid meta stuff. Meta stuff is intimidating for the majority of non-power users.

I think a good strategy to address this issue is to expose a straightforward low-level layer which uses Meta etc., and then build a nicer user friendly layer on top of it.  There are many ideas to explore on how to make this more accessible, and having a low-level API available means that we can experiment easily with all of these approaches.  For example the partial evaluation approach you've suggested has some big issues (you can't compute the size of the tree and put that number in the SVG), but maybe there's a nice API to be built around it that still works nicely for tree visualization.

I wouldn't underestimate our users either.  Of course we have a lot of non-technical users, but they most likely don't know HTML either.  On the other hand we have quite a few power users who know both HTML and a lot of Lean.

Vtec234:
I also forgot about it :sweat_smile: I agree with Gabriel's proposed approach of first providing a powerful, low-level Meta API and then building a nice high-level interface on top. Let's work that out in another PR.

hargoniX:
I'd like to work with the HTML/JSX DSL part of this PR to generate some HTML for the new doc-gen. Would it be possible to at least get the DSL merged here? Or should I just copy the parts I need over into doc-gen with the appropriate licensing? @Vtec234

Vtec234:
@hargoniX I think we will want something like that in core Lean eventually, but for now please feel free to just copy it. It is Apache-licensed like all of Lean. When (if?) it becomes part of core Lean, you should hopefully be able to replace that with an `import` and maybe minor adaptations.

PR #725, opened 2021-10-13
remove manual shell scripting from our cmake files

lovettchris:
fix: remove manual shell scripting for compiling libleanrt with the additional compile flag "-ftls-model=local-exec".  I think cmake can easily do this just using target_compile_options.  Removing bash shell will eventually get us a cmake system that will work on windows.

This is proof that the cmake change works, there are two sets of object files, one for lean_rt-initial-exec and one for leanrt and they are different:
image


Kha:
Nice. Not sure why I couldn't find that option :sweat_smile: .

PR #729, opened 2021-10-15
fix: out-of-bounds array access

gebner:
This fixes a panic with the following file (and also every single time you write `macro_rules`):
```lean
macro_rules
```

Since `macro_rules` parses `many1` alternatives, this parses as a single alternative with no patterns.  And that case was not handled in `inferMacroRulesAltKind`.


Kha:
Does `$pat, $pats,*` work?

gebner:
That requires a comma, doesn't it?  In macro_rules you typically only have a single pattern.  Maybe matching on `$pat` instead could also work.

PR #733, opened 2021-10-20
Bundle clang+lld+libc++ for Linux

Kha:
Alternative approach to #659: instead of relying on Zig, we bundle everything necessary for standalone compilation ourselves. This way we're kept in control and can use a uniform, mature toolchain across all platforms. And hopefully avoid unexplainable errors that we could not reproduce without Zig.

Specifically, what we bundle in the case of x86-64 Linux is
```
stage1
├── [237M]  bin
│   ├── [141M]  clang
│   ├── [   3]  ld.lld -> lld
│   └── [ 96M]  lld
├── [ 47K]  include
│   └── [ 43K]  clang
│       ├── [2.8K]  limits.h
│       ├── [ 583]  stdalign.h
│       ├── [1.1K]  stdarg.h
│       ├── [7.1K]  stdatomic.h
│       ├── [ 897]  stdbool.h
│       ├── [3.5K]  stddef.h
│       ├── [ 857]  __stddef_max_align_t.h
│       ├── [ 22K]  stdint.h
│       └── [ 510]  stdnoreturn.h
└── [ 11M]  lib
    ├── [8.5K]  crt1.o
    ├── [2.3K]  crtbegin.o
    ├── [2.7K]  crtbeginS.o
    ├── [1.1K]  crtend.o
    ├── [1.1K]  crtendS.o
    ├── [2.7K]  crti.o
    ├── [2.5K]  crtn.o
    ├── [3.9M]  glibc
    │   ├── [ 23K]  libc_nonshared.a
    │   ├── [1.9M]  libc.so
    │   ├── [ 18K]  libdl.so
    │   ├── [  13]  libgcc_s.so -> libgcc_s.so.1
    │   ├── [ 98K]  libgcc_s.so.1
    │   ├── [1.7M]  libm.so
    │   ├── [143K]  libpthread.so
    │   └── [ 45K]  librt.so
    ├── [  18]  libatomic.so -> libatomic.so.1.2.0
    ├── [  18]  libatomic.so.1 -> libatomic.so.1.2.0
    ├── [ 30K]  libatomic.so.1.2.0
    ├── [1.8M]  libc++.a
    ├── [662K]  libc++abi.a
    ├── [  14]  libc++abi.so -> libc++abi.so.1
    ├── [  16]  libc++abi.so.1 -> libc++abi.so.1.0
    ├── [357K]  libc++abi.so.1.0
    ├── [ 12K]  libc++experimental.a
    ├── [  28]  libc++.so
    ├── [  13]  libc++.so.1 -> libc++.so.1.0
    ├── [1022K]  libc++.so.1.0
    ├── [2.9M]  libgcc.a
    ├── [ 948]  libgmp.la
    ├── [  16]  libgmp.so -> libgmp.so.10.3.2
    ├── [  16]  libgmp.so.10 -> libgmp.so.10.3.2
    └── [636K]  libgmp.so.10.3.2
```
As can be seen, the bundle size is dominated completely by the `clang` and `lld` executables. That Zig manages to combine both of them *and* their own compiler in an executable smaller than `clang` is an impressive feat and demonstrates that this bundle could still be optimized, though I don't think it's a dealbreaker for us.

A special issue on Linux is the glibc. We bundle a relatively old version of it, glibc 2.27 from Nixpkgs 19.03, so that Lean programs linked against it work on most somewhat recent distros. For running we always use the system glibc, which is why we put the bundled one in a separate directory.

Apart from the glibc, a Lean program built with this bundle does have more runtime dependencies, namely the libc++ and GMP DSOs:
```
$ ldd tests/compiler/534.lean.out
	linux-vdso.so.1 (0x00007ffd9dde4000)
	libc++.so.1 => not found
	libc++abi.so.1 => not found
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f5b35e59000)
	libgmp.so.10 => /usr/lib/x86_64-linux-gnu/libgmp.so.10 (0x00007f5b35dd5000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f5b35dcf000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f5b35db2000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f5b35d8f000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f5b35b9d000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f5b35fd2000)
```
For the tests, I simply set `LD_LIBRARY_PATH` to the bundle's `lib/`, but that isn't really an option in general. libc++ (BSD-like/MIT) we should probably link statically (like it seems to be the case with libstdc++ right now), but that is problematic with GMP because of its LGPL license. The best approach I can come up with is to use the "Windows approach": put `$ORIGIN` (the executable's own directory) in its rpath and copy libgmp.so next to it (which is basically free with hardlinks). The second step should probably be done by Lake, not leanc.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,7 +148,7 @@ jobs:
       - name: List Install Tree
         run: |
           # omit contents of src/, Init/, ...
-          tree --du -h lean-* | grep -Ev '(│.*){3}'
+          tree --du -h lean-* | grep -Ev '\.o?lean'
       - name: Pack
         if: ${{ startsWith(github.ref, 'refs/tags/v') && matrix.release }}
         run: |


Kha:
@gebner It looks like we should really compile LLVM from source after all... or we do the same thing as with the Lean executable and just copy those deps as well.

`stdio.h` is not a header required by Lean programs. I'm assuming that FFI components will require additional dependencies and thus will be compiled using external toolchains anyway.

gebner:
> @gebner It looks like we should really compile LLVM from source after all... or we do the same thing as with the Lean executable and just copy those deps as well.
> I'm assuming that FFI components will often require additional dependencies and thus will be compiled using external toolchains anyway.

I'm not sure I understand the intent behind this PR then.  If users are supposed to use their own toolchain anyhow, why do we bundle it?  Wouldn't it be easier to add an `elan install-clang` command that downloads a known good clang release for users on old linux distributions?

Kha:
Ah, the cpack configuration must be wrong, all the files listed above should be included! I'll take a look.

Regarding NixOS, it might be best to completely ignore the bundle and keep using stdenv? We can make it so that when `LEAN_CC` is defined, none of the new flags (like `--sysroot`) are applied.

tydeu:
@Kha you are correct, running the executable outside of mingw does crash due to not being able to find libgmp. I am suprised I never stumbled onto that before! 

More to your point, the copying should probably occur during the creation of a "distribution" of package, which Lake does not yet do, However, it is certainly something on the TODO list. Such a `dist` folder would also want to be free of intermediate or development build artifacts like trace files and give users the ability to hook in additional shared libraries to copy.

However, this does raise the question of whether it might be worth it in the medium to long term to move from libgmp to a different multi-precision library that can be linked statically to make Lean's core self-contained.

Kha:
Closing in favor of #795

PR #734, opened 2021-10-20
Cross-compiled build for Linux/aarch64

gebner:
This adds a new build for Linux/aarch64, which is cross-compiled on github actions.  The tests are run using CPU emulation.

The release tarball is called `lean-x.y.z-linux_aarch64.tar.gz`, this is supported by recent elan versions.

The main difficulty was getting stage0 to work as a cross-compiler (i.e., compile the stage0 compiler for amd64 but have it generate oleans and C files that run on aarch64).  It turns out that the oleans were not architecture-independent, because of the hash function we use for strings.  This function had a parameter of type `char const *`, on amd64 this is `signed char const *`, on aarch64 this is `unsigned char const *`.  I have changed it to `unsigned char` independent of the architecture (requiring a stage0 update).

(Thanks to @Kha for pointing me to the olean files as a source of troubles.  In order to debug the olean differences, I wrote a small and very incomplete olean parser in Lean: https://github.com/gebner/oleanparser)


leodemoura:
@gebner @Kha Should we close or merge this PR?

gebner:
I believe Sebastian wants to do this in a completely different way.

Kha:
For what it's worth, the first alpha release of Asahi Linux might increase the priority of this feature a little; personally I'm waiting for [better 16K pages support](https://github.com/AsahiLinux/docs/wiki/Software-known-to-have-issues-with-16k-page-size) in Emacs & libunwind. There are even [reports](https://twitter.com/marcan42/status/1505811806010613767) about Linux on M1 being faster than macOS. I'll try cross-compiling LLVM as soon as this PC has been upgraded to more than 8GB RAM.

PR #736, opened 2021-10-22
Bundle clang+lld+libc++ for Windows

Kha:
Building on top of #733. There are no official LLVM tarball releases for Windows and we need the mingw headers anyway, so we simply copy everything from mingw.


Kha:
Closing in favor of #795

Issue #738, opened 2021-10-25
Asynchronous printing hangs the server


tydeu:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

Printing text asynchronously (e.g., a `IO.println` run through `IO.asTask`) causes the Lean server to hang.

### Steps to Reproduce

Run the following code via the Lean server (e.g., in VSCode) to observe the server hanging:

```lean
#eval show IO _ from do 
  IO.wait <| ← IO.asTask <| IO.println "test"
```

**Expected behavior:**

To print "test" either as a diagnostic or in the output panel.

**Actual behavior:**

The server hangs.

**Reproduces how often:** 

Always.

### Versions

Windows 10 20H2
`Lean (version 4.0.0-nightly-2021-10-23, commit a741a3dfd422, Release)`

### Additional Information

I suspect this is actually due to the same problem as #640. That is, I suspect that the asynchronous print is not being properly redirected (due to being in a separate thread) and is thus printing to normal `stdout`,. This, as #640 shows, causes the server to hang.

kmill:
Just checked that this still hangs on 4.12.0, commit a3bc4d235961bfeacf9a0ed62b01319338b2c75f.
```lean4
#eval discard <| IO.asTask <| IO.println "test"
```

TwoFX:
Now that we have the `Async` monad, I expect that people will begin to hit this bug much more frequently ([here](https://leanprover.zulipchat.com/#narrow/channel/217875-Is-there-code-for-X.3F/topic/Parallel.20Execution.20of.20.60List.2Emap.60/near/529483927) is an example).

Kha:
Suggestion: in the language server, after retrieving the streams for LSP communication itself, set `g_stream_stdout` (i.e. the default stdout stream every thread starts with) to stderr and set `g_stream_stdin` to a closed stream.

robsimmons:
As @TwoFX predicted, I hit this because I was using Async. 

This throws a user-defined message into the LSP's `stderr` stream (in Workbench and in VSC, not in live.llo):

```
import Std.Async
open Std.Async

#eval (Async.block do
  IO.eprintln "this gets printed to an error toast" : IO Unit)
```

and this breaks communication between the LSP and Vscode (in Workbench, VSC, and live.llo:

```
import Std.Async
[...]

PR #744, opened 2021-10-27
feat: generalize IO task functions to `BaseIO := EIO Empty`

tydeu:
They could be further generalized (along the other primitives) by having them operate in the `EIO Empty` monad rather than `IO`. 

However, this would require a slightly broader API change. Namely, the addition of a `MonadLift (EIO Empty) (EIO ε)` instance  (for ease of use) and a demonstrations at `Except ε α` is inhabited. However, I wasn't sure if we want to go in that direction in the Lean core, so I have yet to include said generalization in this PR. . Note that I am taking a such an approach in Lake, though.

### DIFF
--- a/src/runtime/io.cpp
+++ b/src/runtime/io.cpp
@@ -822,7 +822,7 @@ extern "C" LEAN_EXPORT obj_res lean_io_as_task(obj_arg act, obj_arg prio, obj_ar
     return io_result_mk_ok(t);
 }
 
-/* {α β : Type} (f : α → BaseIO β) (a : BaseIO α) : β */
+/* {α β : Type} (f : α → BaseIO β) (a : α) : β */
 static obj_res lean_io_bind_task_fn(obj_arg f, obj_arg a) {
     object_ref r(apply_2(f, a, io_mk_world()));
     return object_ref(io_result_get_value(r.raw()), true).steal();


tydeu:
@Kha any comments on the PR? It is ready to merged in its current state (though there is further generalization as @gebner mentioned that could be done in the future).

Kha:
Yeah, sorry, I did look into the first `liftM` but then got distracted. The other ones seem much more fundamental; even if we special-case the `Empty` case, it would be nice if it Just Worked for other error coercions as well. This works very nicely in Rust afaik, but it certainly helps having dedicated syntax for error propagation and async bind instead of arbitrary HOFs.

In any case, we need to decide whether the added generality is worth the potential for bad type inference. Current evidence seems to suggest that only unrealistic example code is affected, so I'm fine with the PR as is.

As for the first `liftM`, I'm blaming the dubious double occurrence of `ε` in `bindTask` (which previously was a double occurrence of `IO.Exception`). Making the signature more strict fixes this and led me to a little refactoring of a dubious monad stack in the server, which looks like a beneficial change as well to me: https://github.com/Kha/lean4/commit/c3d6255b25b8fa54d5303141bfee7f94aa0d259e. I can push that after this PR is merged.

Kha (approved):

src/Init/System/IO.lean
```diff
@@ -31,6 +31,12 @@ def IO.RealWorld : Type := Unit
 -/
 def EIO (ε : Type) : Type → Type := EStateM ε IO.RealWorld
 
+@[inline] def EIO.liftEmpty (x : EIO Empty α) : EIO ε α :=
+  fun s => match x s with
+  | EStateM.Result.ok a s => EStateM.Result.ok a s
```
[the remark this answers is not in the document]
tydeu: @Kha Huh? The current version has the same indentation as `catchExceptions` below it. That suggestion would make the indentation inconsistent between them?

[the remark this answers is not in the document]
Kha: I guess in my mind they're both wrong then! But it's not the end of the world.

tydeu:
From the discussion here it seemed like the the generalization to `RealM := EIO Empty` (that @gebner suggested) was desirable, so I went ahead and did that. Doing so also allowed me to revert the changes to the test.  @Kha, does my implementation look fine?

src/Init/System/IO.lean
```diff
@@ -97,48 +175,39 @@ def sleep (ms : UInt32) : IO Unit :=
...
-    | t::ts, as =>
-      IO.bindTask t (fun a => go ts (a :: as)) prio
-    | [], as => IO.asTask (f as.reverse) prio
+/-- `IO` specialization of `EIO.mapTasks`. -/
+@[inline] def mapTasks (f : List α → IO β) (tasks : List (Task α)) (prio := Task.Priority.default) : RealM (Task (Except IO.Error β)) :=
+  EIO.mapTasks f tasks prio
 
 /-- Check if the task's cancellation flag has been set by calling `IO.cancel` or dropping the last reference to the task. -/
-@[extern "lean_io_check_canceled"] constant checkCanceled : IO Bool
+@[extern "lean_io_check_canceled"] constant checkCanceled : EIO ε Bool
```
Kha, line 195: Changing this from `EIO Empty` can't seriously be the solution to the lifting issue, can it?

src/Init/System/IO.lean
```diff
@@ -31,37 +31,61 @@ def IO.RealWorld : Type := Unit
...
+  | EStateM.Result.ok a s => EStateM.Result.ok a s
+
+instance : MonadLift RealM (EIO ε) := ⟨RealM.toEIO⟩
+
+@[inline] def EIO.toRealM (self : EIO ε α) : RealM (Except ε α) :=
+  fun s => match self s with
+  | EStateM.Result.ok a s     => EStateM.Result.ok (Except.ok a) s
+  | EStateM.Result.error ex s => EStateM.Result.ok (Except.error ex) s
+
+@[inline] def EIO.toRealM_ (self : EIO ε α) : RealM Unit :=
```
Kha, line 54: I don't think we use this naming style. I'm also not sure we should have this function.

src/Init/System/IO.lean
```diff
@@ -31,37 +31,61 @@ def IO.RealWorld : Type := Unit
...
 
-@[inline] def EIO.toIO' (x : EIO ε α) : IO (Except ε α) :=
-  EIO.toIO (fun _ => unreachable!) (observing x)
+@[inline] def EIO.toIO' (self : EIO ε α) : IO (Except ε α) :=
+  self.toRealM
 
-@[inline] def IO.toEIO (f : IO.Error → ε) (x : IO α) : EIO ε α :=
-  x.adaptExcept f
+@[inline] def IO.toEIO (f : IO.Error → ε) (self : IO α) : EIO ε α :=
+  self.adaptExcept f
```
Kha, line 77: Why this unnecessary churn? The commit guidelines say to adapt to the style of the surrounding code, not the reverse. If we want to change the style of the stdlib, we should have a discussion first.

Kha (commented):
I might have preferred for the `RealM` changes to go in a separate PR, but we should be able to cope. The signatures look good afaics. But in this case, I think we should do it right from the beginning and make sure that it is not defeq to `EIO Empty`, as Gabriel suggested. That is, conceptually `EIO` should be a monad stack `ExceptT E (RealM A)`, paying no mind to the actual inverse implementation.

I would also prefer a name that is not uncomfortably close to `Real`, such as `IOBase`.

src/Init/System/IO.lean
tydeu: @Kha This naming style is copied from Haskell. I currently have a minimum make do API for [`RealM`](https://github.com/leanprover/lake/blob/b2709a91741df37f2ee6d4af74eb9ceb63f88ae2/Lake/RealM.lean) in Lake and this was the last function missing from the introduction of `RealM` here that I use there. Thus, I stuck it here. I am happy to revert the change if you think it is inappropriate, but I can say, at the very less, it is useful to me (in Lake),

src/Init/System/IO.lean
tydeu: @Kha the arguments names are not used anywhere in the Lean core and there doesn't appear to be, as far as I could tell, a consistent naming convention (e.g., `x` was used here, `fn` is used below in `unsafe*`,  `act` was used in `asTask`).  As such, I was under the impression the names weren't significant. Thus, I just went with my usual style (`self` for arguments intended to be used with dot notation) and adjusted close similar functions to match in order to introduce a little more consistency across them. I am happy to revert the name changes if desired, but then I would love some guidelines as to what convention I should be following for argument names for functions.

src/Init/System/IO.lean
tydeu: I didn't do this elsewhere because this style, on the other hand, prevents automatic lifting of the monad in some cases. As such, there is a where negotiation between where `EIO e` provides better type inference and where `RealM` does.

gebner:
I also dislike the name `RealM`, not only does it clash with real numbers, the name is also pretty long and there's nothing particularly "real" about the actions it supports.  The `IO` monad enables mutating references, side-effects, input-output, calls to native libraries, and side-effects in general.  I think `IO` is the best name for this (exception-free) monad (mainly because it fills this niche in Haskell).  Another potential suggestion would be `Mut` or `MutM` for "mutation".

tydeu:
@gebner Fair enough. though I must say I somehow managed to feel the exact opposite on most of your points.

>  the name is also pretty long 

It's 5 characters? It matches the current minimum length of any monad with proper name (i.e., one ended with `M` -- e.g., `MetaM`/`CoreM`, etc.). Standard issue monads have even longer names (e.g., `OptionM`, `ExceptM`, `StateM`,  etc.).

> and there's nothing particularly "real" about the actions it supports. 

II allows for interaction with the real world?  To do be honest, I would think that would make it more deserving of the term "real" than even the real numbers (which, given the modern quantization of physics, likely aren't particularly "real" to begin with).

> The `IO` monad enables mutating references, side-effects, input-output, calls to native libraries, and side-effects in general. I think `IO` is the best name for this (exception-free) monad (mainly because it fills this niche in Haskell). 

In my view, you just illustrated why  `IO` is *not* a good name for the more general monad. As you pointed it out it enables many things beyond conventional input-output, making the name `IO` far to restrictive. I feel the current `IO` monad is a much more deserving of that name (as the errors it encapsulates are considerably more conventional input-output focused). I will grant you though, that using `IO` for it does have the advantage of familiarity for those coming from functional languages in the vein of Haskell.

> Another potential suggestion would be `Mut` or `MutM` for "mutation".

I feel that this name is similarly restrictive -- many of the things which use `RealM`/`IO` are not mutations (e.g., it can be simply reading info from the external environment). Furthermore, it doesn't make it clear how it is different from a pure mutation (e.g., `StateM`). Other names I would suggest would be some like `ExternM`/`ExtM`/`EnvM`/`SysM`/etc. That is, names that make it clear the monad is interfacing with some external / real / system environment. 

I think `SysM` would probably be the best alternatively as it doesn't clash with, as far as I know, any other unrelated concept (as most of things in the `System` namespace do operate within it).

Kha:
> By that, do you mean by wrapping it in a ` structure`?

A `def` is usually more comfortable for deriving instances, see `EIO`.

As for naming, I don't see any name here that suggests everything from mutable references to actual I/O, thus my suggestion that we stay with the current naming and use `IOBase`. As I don't expect this base monad to be used all that often, I don't mind the longer name.

>  I think IO is the best name for this (exception-free) monad (mainly because it fills this niche in Haskell).

I wouldn't call Haskell's `IO` exception-free because of the presence of async exceptions. In my first draft for the current `IO` API, I had `IO` as this exception-free base monad and used explicit `ExceptT _ IO`s on top of it, but Leo convinced me that the stack with the standard error type should have the common and shortest name `IO`. That is, I/O should be in `IO`.

Kha:
>  As I don't expect this base monad to be used all that often, I don't mind the longer name.

In particular, I'm wondering if we should have something like
```
def AsyncIOBase α := IOBase (Task α)
def AsyncEIO ε α  := AsyncIOBase (Except ε α)
def AsyncIO α     := AsyncEIO IO.Error α
```
as a higher abstraction. This would enable robust lifting of synchronous operations into the asynchronous monads.

tydeu:
@Kha 

> > By that, do you mean by wrapping it in a ` structure`?
> 
> A `def` is usually more comfortable for deriving instances, see `EIO`.

 The was my original intention (afterl all, it is what I used for for my [`RealM`](https://github.com/leanprover/lake/blob/b2709a91741df37f2ee6d4af74eb9ceb63f88ae2/Lake/RealM.lean) in Lake). However, I initially had some trouble getting the `MonadLift` instance to produce proper inferences, so I gave up, but that seems to have since resolved itself. I also I did not realize that `def`s created not defeq types -- I really need to familiarize myslef more wiith the nitty gritty of how the kernel works.

> As for naming, I don't see any name here that suggests everything from mutable references to actual I/O, thus my suggestion that we stay with the current naming and use `IOBase`. As I don't expect this base monad to be used all that often, I don't mind the longer name.

Is `BaseIO` fine as instead? It has prettier camel-casing imo and keeps the trend of the suffix indicating the kind of type it is. 

I also do suspect this will be used quite a lot considering that many uses of IO (e.g., motive references, FFIs) either can't error or have errors that don't fit neatly into that of `IO`. @gebner seem to also think it could see quite a lot of use.


> In particular, I'm wondering if we should have something like
> 
> ```
> def AsyncIOBase α := IOBase (Task α)
> def AsyncEIO ε α  := AsyncIOBase (Except ε α)
> def AsyncIO α     := AsyncEIO IO.Error α
> ```
> 
> as a higher abstraction. This would enable robust lifting of synchronous operations into the asynchronous monads.

I was actually work on a similar abstraction for Lake (testing it is how I came across issue #738). My only worry in designing that was that it lifts all synchronous operations to `Task`, which I am not sure is desirable (i.e., I believe you previously mentioned on Zulip that binds of pure Tasks can not be efficiently optimized). 

Also, fyi, I think you are missing a `Task` in the definition of `AsyncEIO`.

src/Init/System/IO.lean
Kha: It looks like a huge red flag to me. When is it ok to swallow IO exceptions?

src/Init/System/IO.lean
Kha: I used `x` for any monadic value, but I'm happy with Leo's `act`. `fn` is strange indeed.

Kha:
> I also I did not realize that `def`s created not defeq types

Well, they are still defeq, just not during typeclass inference.



> Is `BaseIO` fine as instead? It has prettier camel-casing imo and keeps the trend of the suffix indicating the kind of type it is.

Fine with me.

> I believe you previously mentioned on Zulip that binds of pure Tasks can not be efficiently optimized

That's true, but I think I now know how to do that. But that's a topic for another PR I suppose.



> Also, fyi, I think you are missing a `Task` in the definition of `AsyncEIO`.

It's based on `AsyncIOBase`.

tydeu:
@Kha 

> It's based on `AsyncIOBase`.

Oh, oops, it appears it is I who have misread! 😆 Sorry about that! Otherwise, sounds good, Will get on implemented the review soon.

src/Init/System/IO.lean
tydeu: @Kha when performing an action where you can't reasonably handle the error and/or you don't really care if it completes.

For example, printing out log messages. If the print errors, then the error itself can't be logged (because logging is the thing that is erroring). You also generally don't want the whole program to stop because you failed to print a log line. Thus, it is fine to swallow the IO error.

src/Init/System/IO.lean
tydeu: @Kha so how would you like me to handle this? Would you like me to revert these changes and use `x`  (or `act`) instead in the new functions? Or is it fine enough as is?

src/Init/System/IO.lean
Kha: I believe we can use `BaseIO` now without breaking the test

src/Init/System/IO.lean
Kha: We are already more strict about discarding return values in `do` than Haskell, so I don't think it's consistent to provide functions like this one. I don't like the "cute" naming style either. We have `discard` (or `let _ <-`) for this, let's be explicit when we use it.

src/Init/System/IO.lean
tydeu: @Kha 

> I'd suggest to change every `self` to `act` just to not introduce a fourth style. You're welcome to open a thread in #lean4-dev about naming guidelines going forward.

done

src/Init/System/IO.lean
tydeu: @Kha I removed the function

tydeu:
@Kha I believe I have addressed all your comments, so the PR should now be complete.

src/Init/System/IO.lean
Kha: Ah, I should have unresolved this one

tydeu: > I believe we can use `BaseIO` now without breaking the test

@Kha I am 99% we cannot, as I had previously tried to do so and it failed. But I will test again just to make sure.

tydeu: Well, it appears I am wrong, it does now work!

tydeu: As such, I am going to change `getNumHearbeats` as well.

Kha:
@leodemoura What do you think? This sensibly generalizes & simplifies primitives while the "convenience" `IO` layer is left (mostly) the same.

leodemoura:
> @leodemoura What do you think? This sensibly generalizes & simplifies primitives while the "convenience" `IO` layer is left (mostly) the same.

@Kha It looks good to me. I am going to merge.

PR #752, opened 2021-10-29
--load-dynlib

Kha:
This is a relatively simple change that exposes a flag for cross-platform `LD_PRELOAD` and adds a test for interpreting `@[extern]`. There is some overlap with `--plugin` of course, but since we are not quite sure about the intended semantics of that one while this one is quite clear, a separate flag seemed like the best solution.

I copied the more complex `foreign` setup from #733, so I assume the Windows segfault will come back now...

### DIFF
--- a/src/util/shell.cpp
+++ b/src/util/shell.cpp
@@ -301,7 +301,7 @@ void load_plugin(std::string path) {
 #ifdef LEAN_WINDOWS
     HMODULE h = LoadLibrary(path.c_str());
     if (!h) {
-        throw exception(sstream() << "error loading plugin " << path);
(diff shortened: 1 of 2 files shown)


Kha:
To access an extern decl, you first need to import it, which will run the module initializer (backed by native code in this case)

Anderssorby:
I'm trying to use this here to link to native shared libs:
https://github.com/yatima-inc/lean4/tree/acs/add-nix-ability-for-native-libs

Kha:
This is finally ready. As far as I can tell, it is not possible on Windows to link an executable against a shared library and then have that library use symbols from the executable at runtime. Thus we should always use libleanshared when integrating external C++ code as in the test.

PR #757, opened 2021-11-02
feat: Lean.githash 

Kha:


tydeu:
@Kha is this just waiting on Leo, or do you have more adjustments planned?

Kha:
It's ready from my side. The only caveat I can see is that when the hash changes, instead of only relinking `lean`, we now have to relink everything. But linking is stil not much of a bottleneck when building Lean afaik.

tydeu:
@Kha yeah linking is relatively quite fast. The huge bottleneck when building Lean is building the whole import chain. Running the tests has also gotten slower as we have climbed our way to over 1000 tests.

PR #758, opened 2021-11-02
doc: Structure update.

bollu:
Add docs about how to update structure fields via `with`

### DIFF
--- a/doc/struct.md
+++ b/doc/struct.md
@@ -217,7 +217,7 @@ def msg2 : MessageExt where
 
 ## Updating structure fields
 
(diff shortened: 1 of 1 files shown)


doc/struct.md
```diff
@@ -217,4 +217,11 @@ def msg2 : MessageExt where
 
 ## Updating structure fields
 
-TODO
\ No newline at end of file
+Structure fields can be updated using `{ <struct> with <field> := <new-value> }`:
```
Kha, line 220:
```suggestion
Structure fields can be updated using `{ <struct> with <field> := <new-value>, ... }`:
```

PR #759, opened 2021-11-02
fix: document highlight kind bug

lovettchris:
The vscode definitions are defined in index.d.ts as:

```typescript
/**
 * A document highlight kind.
 */
export enum DocumentHighlightKind {

    /**
     * A textual occurrence.
     */
    Text = 0,

    /**
     * Read-access of a symbol, like reading a variable.
     */
    Read = 1,

    /**
     * Write-access of a symbol, like writing to a variable.
     */
    Write = 2
}
```

### DIFF
--- a/src/Lean/Data/Lsp/LanguageFeatures.lean
+++ b/src/Lean/Data/Lsp/LanguageFeatures.lean
@@ -77,9 +77,9 @@ inductive DocumentHighlightKind where
 
 instance : ToJson DocumentHighlightKind where
  toJson
-   | DocumentHighlightKind.text => 1
-   | DocumentHighlightKind.read => 2
-   | DocumentHighlightKind.write => 3
+   | DocumentHighlightKind.text => 0
+   | DocumentHighlightKind.read => 1
+   | DocumentHighlightKind.write => 2
 
 structure DocumentHighlight where
   range : Range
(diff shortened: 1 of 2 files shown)


Kha:
Then `index.d.ts` is in violation of the [spec](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#documentHighlightKind) :) ...

gebner:
It's perfectly fine if vscode uses different enum values than the LSP spec (although unnecessarily confusing). The vscode-languageclient libraries correctly converts between the two: https://github.com/microsoft/vscode-languageserver-node/blob/b988c1e9d17fb11fec9700028da7bc3b99d17ef3/client/src/common/protocolConverter.ts#L758-L768

The LSP interface in Lean is standards-conforming and should not be changed.

lovettchris:
> It's perfectly fine if vscode uses different enum values than the LSP spec (although unnecessarily confusing). The vscode-languageclient libraries correctly converts between the two: https://github.com/microsoft/vscode-languageserver-node/blob/b988c1e9d17fb11fec9700028da7bc3b99d17ef3/client/src/common/protocolConverter.ts#L758-L768


Ah, nice find, now it makes more sense, thanks!

PR #763, opened 2021-11-03
Add nix ability for native shared libs

Anderssorby:
This adds a new recursive option `nativeSharedLibs` to `buildLeanPackage` in nix.
The nix part of --load-dynlib #752

### DIFF
--- a/nix/buildLeanPackage.nix
+++ b/nix/buildLeanPackage.nix
@@ -8,6 +8,8 @@ lib.makeOverridable (
   deps ? [ lean.Lean lean.Leanpkg ],
   # Static library dependencies. Each derivation `static` should contain a static library in the directory `${static}`.
   staticLibDeps ? [],
+  # Whether to wrap static library inputs in a -Wl,--start-group [...] -Wl,--end-group to ensure dependencies are resolved.
+  groupStaticLibs ? false,
   # Shared library dependencies included at interpretation with --load-dynlib and linked to. Each derivation `shared` should contain a 
   # shared library at the path `${shared}/${shared.libName or shared.name}` and a name to link to like `-l${shared.linkName or shared.name}`.
   # These libs are also linked to in packages that depend on this one.
@@ -57,13 +59,14 @@ with builtins; let
   srcRoot = src;
 
   # A flattened list of Lean-module dependencies (`deps`)
-  allExternalDeps = lib.foldr (dep: allExternalDeps: allExternalDeps // { ${dep.name} = dep; } // dep.allExternalDeps) {} deps;
+  allExternalDeps = lib.unique (lib.foldr (dep: allExternalDeps: allExternalDeps ++ [ dep ] ++ dep.allExternalDeps) [] deps);
   allNativeSharedLibs =
-    lib.unique (lib.flatten (nativeSharedLibs ++ (map (dep: dep.allNativeSharedLibs or []) (attrValues allExternalDeps))));
+    lib.unique (lib.flatten (nativeSharedLibs ++ (map (dep: dep.allNativeSharedLibs or []) allExternalDeps)));
+
   # A flattened list of all static library dependencies: this and every dep module's explicitly provided `staticLibDeps`, 
   # plus every dep module itself: `dep.staticLib`
   allStaticLibDeps = 
-    lib.unique (lib.flatten (staticLibDeps ++ (map (dep: [dep.staticLib] ++ dep.staticLibDeps or []) (attrValues allExternalDeps))));
+    lib.unique (lib.flatten (staticLibDeps ++ (map (dep: [dep.staticLib] ++ dep.staticLibDeps or []) allExternalDeps)));
 
   pathOfSharedLib = dep: dep.libPath or "${dep}/${dep.libName or dep.name}";
 
@@ -73,7 +76,7 @@ with builtins; let
   fakeDepRoot = runBareCommandLocal "${name}-dep-root" {} ''
     mkdir $out
     cd $out
-    mkdir ${lib.concatStringsSep " " ([name] ++ attrNames allExternalDeps)}
+    mkdir ${lib.concatStringsSep " " ([name] ++ (map (d: d.name) allExternalDeps))}
   '';
   print-lean-deps = writeShellScriptBin "print-lean-deps" ''
     export LEAN_PATH=${fakeDepRoot}
@@ -125,7 +128,7 @@ with builtins; let
     '';
   };
   singleton = name: value: listToAttrs [ { inherit name value; } ];
-  externalModMap = lib.foldr (dep: depMap: depMap // dep.mods) {} (attrValues allExternalDeps);
+  externalModMap = lib.foldr (dep: depMap: depMap // dep.mods) {} allExternalDeps;
   # Recursively build `mod` and its dependencies. `modMap` maps module names to
   # `{ deps, drv }` pairs of a derivation and its transitive dependencies (as a nested
   # mapping from module names to derivations). It is passed linearly through the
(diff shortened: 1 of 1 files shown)


Anderssorby:
I will probably squash this entire PR into a single commit with the correct format when things start to work as intended.

nix/buildLeanPackage.nix
```diff
@@ -154,17 +164,19 @@ in rec {
       ${lib.concatStringsSep " " (map (d: "${d}/*.a") allStaticLibDeps)} \
       -o $out/${name}.so
   '';
-  executable = runCommand executableName { buildInputs = [ stdenv.cc leanc ]; } ''
+  executable = runCommand executableName { 
+    buildInputs = [ stdenv.cc leanc ];
+  } ''
```
Kha, line 169: inconsistent style

nix/buildLeanPackage.nix
```diff
@@ -112,7 +121,7 @@ with builtins; let
       # make local "copy" so `drv`'s Nix store path doesn't end up in ccache's hash
       ln -s ${drv.c}/${drv.cPath} src.c
       # on the other hand, a debug build is pretty fast anyway, so preserve the path for gdb
-      leanc -c -o $out/$oPath $leancFlags -fPIC ${if debug then "${drv.c}/${drv.cPath} -g " else "src.c -O3 -DNDEBUG"}
+      leanc -c -o $out/$oPath $leancFlags -fPIC ${if debug then "${drv.c}/${drv.cPath} -g -v" else "src.c -O3 -DNDEBUG"}
```
Kha, line 124: leftover?

nix/buildLeanPackage.nix
```diff
@@ -8,6 +8,10 @@ lib.makeOverridable (
   deps ? [ lean.Lean lean.Leanpkg ],
   # Static library dependencies. Each derivation `static` should contain a static library in the directory `${static}`.
   staticLibDeps ? [],
+  # Shared library dependencies included at interpretation with --load-dynlib and linked to. Each derivation `shared` should contain a 
+  # shared library at the path `${shared}/${shared.libName or shared.name}` and a name to link to like `-l${shared.linkName or shared.name}`.
+  # These libs are also linked to in packages that depend on this one.
+  nativeSharedLibs ? [],
```
Kha, line 14: The "native" feels redundant

nix/buildLeanPackage.nix
```diff
@@ -154,17 +164,19 @@ in rec {
       ${lib.concatStringsSep " " (map (d: "${d}/*.a") allStaticLibDeps)} \
```
Kha, line 164: You're not actually linking against `nativeSharedLibs` here? Doing so correctly might actually be tricky - if we built a dependency into a shared library, we must not link against it again in `allStaticLibDeps`...

nix/buildLeanPackage.nix
Anderssorby: I think this is a very useful feature when debugging builds. Should I use a different option than `debug`?

nix/buildLeanPackage.nix
Anderssorby: Yes, I was unsure about this, but it is to distinguish it from shared libs that are written in lean. Do you want me to remove the native prefix?

nix/buildLeanPackage.nix
Anderssorby: Yes, that might be tricky, but that might be solved with some extra option to exclude certain static libraries from extra linking.

nix/buildLeanPackage.nix
tydeu: @Anderssorby there is no difference though? The shared library loaded through `--load-dynlib` *should* be a shared library written in Lean (though likely including functions with `@[extern]` definitions).

Anderssorby: I see your point, but one can link against shared lean libraries with the `deps` option which is what currently is the `leanSharedLibs` which includes recursive dependency loading.
The setup is a bit confusing, though, because you can either load things into the IR interpreter engine or link to it in a resulting library or executable.

tydeu: @Anderssorby To clarify, my point was that you want to link to the same native shared library both in `--load-dynlib` (for the interpreter) and in the actual build (of the shared Lean library or binary for the package and its dependents). However,, you also want to `--load-dynlib` the shared Lean libraries utilizing those native libraries.

 For instance, in your Blake3 package, there would be to options: 1) statically link the native blake3 to the shared Lean library for the Blake3 module and `--load-dynlib` just that (i.e., no native shared libraries), or 2) dynamically link the native blake3 library to the shared Lean library for Blake3 and `--load-dynlib` both (i.e., just the native library would be linked to the module, but both would be passed to `--load-dynlib`). 

Dependent package would then also need to respect this behavior (`--load-dynlib` the same libraries and link them to their shared libraries or binaries).

Kha:
So how about the following:
* Linking against foreign shared and static libraries is really already covered by `linkFlags`. The only missing part is to concatenate the linking flags of all (statically linked, i.e. currently all) dependencies when building `executable`.
* For shared libraries, in contrast, we should never statically link against Lean dependencies because it could lead to duplicate symbols if there is a dependency diamond. So let's always link against `deps`'s `sharedLib`'s + the current package's `linkFlags` when building `sharedLib`.
* At this point... should we simply load all dependencies with `--load-dynlib` by default? Note that loading the direct ones is sufficient since they will trigger loading transitive dependencies' shared libraries. This will ensure that dependencies' `extern`s will work and their tactics will run fast by default in exchange for some compilation overhead whenever dependencies change (which you will have to pay anyway in case your goal is to produce an executable). We can provide a flag to opt out, but I'm not sure there's a need for finer-grained control.

This looks like a relatively sane & simple approach to me... at least until we want to precompile code *within* a package, which will happen sooner or later in mathlib4. But that should be a strict extension of the above.

tydeu:
@Kha

> * For shared libraries, in contrast, we should never statically link against Lean dependencies because it could lead to duplicate symbols if there is a dependency diamond. So let's always link against `deps`'s `sharedLib`'s + the current package's `linkFlags` when building `sharedLib`.

Can Nix not resolve these dependency diamonds? Lake does statically link all Lean dependencies as the dependencies are flatten/unique'd.

Kha:
So if you statically link package A into packages B and C, which you dynamically link into package D, what happens? I believe the only correct answer is to dynamically  link A as well.

tydeu:
@Kha Currently, in Lake, you can't dynamically link one package into another, so that case doesn't exist. Also, as a result, I was talking more about building binaries, not shared libraries (as the later are only currently useful for `--load-dynlib`).

Kha:
Well, I was talking about shared libraries only as noted.

tydeu:
@Kha Oh! Sorry, I missed the "for shared libraries" modifier at the beginning of the bullet point. Sorry about that! 🤦‍♂️ Lake does still currently statically link shared libraries though, as the only use of them currently is `--load-dynlib`/`--plugin`, where duplicate symbols are not a concern (if the whole package tree is built into a single shared library).

Kha:
The same issue should also occur when using multiple `--load-dynlib`s, which effectively is the same thing as dynamic linking.

nix/buildLeanPackage.nix
Anderssorby: @tydeu Correct me if I'm wrong but I think what you are mentioning is already satisfied (option 2).

tydeu: @Anderssorby oh, sorry, I was going off the discussion below and hadn't notice that you had updated the building to include the shared libraries.

Anderssorby:
@Kha  @tydeu Is there anything more holding this PR back? For now I think it should be sufficient to remind users to not statically link into shared libraries twice.

Kha:
I would obviously prefer to merge a solution that supports diamonds, but since this is all experimental, I suppose it's fine. As mentioned above, we don't need to collect the transitive closure for shared libraries, though.

Anderssorby:
@Kha with static libraries wrapped in a `-Wl,--start-group` in executable I guess it will handle most diamonds (in static libs)?

Kha:
No, those are mostly independent issues. Diamonds are not a concern when linking statically. I don't really like putting everything in `--start-group`, it looks like a hack that should be replaced with a more local, principled solution. It also breaks on macOS where these flags are neither available nor necessary.

Anderssorby:
@Kha Okay. I can extract this as an option which is disabled by default for darwin/macOS. What would be a principled solution? Somehow scanning the static libs for dependencies and do topological ordering if possible?

Kha:
> I can extract this as an option which is disabled by default for darwin/macOS.

Okay, I'm fine with that if you do it uniformly for all linking steps, i.e. `sharedLib` and `withSharedLib` as well.

> What would be a principled solution?

We should first analyze what exactly is going wrong in your case. But one issue I can see now is that we discard the natural topological sorting in `allExternalDeps` when using an attrset instead of a list. We shouldn't do that.

Anderssorby:
@Kha Adding the all the static libs to the `withSharedLib` causes this error.
```
leanc> /nix/store/fp2h123793bxln4cjxy0mz08sf4nx2m3-binutils-2.35.1/bin/ld: /nix/store/gyb7kizg7kzvzd6q445yn0agxscb3a11-Init.Data.Float-cc/Init/Data/Float.o: undefined reference to symbol 'acos@@GLIBC_2.2.5'
leanc> /nix/store/fp2h123793bxln4cjxy0mz08sf4nx2m3-binutils-2.35.1/bin/ld: /nix/store/2jfn3d7vyj7h0h6lmh510rz31db68l1i-glibc-2.33-50/lib/libm.so.6: error adding symbols: DSO missing from command line
leanc> clang-13: error: linker command failed with exit code 1 (use -v to see invocation)
error: builder for '/nix/store/ylsvd813scmdrv6ddbbhrafvpfp4w0cy-leanc.drv' failed with exit code 1;
```

Kha:
Yeah, `withSharedStdlib` is a bit special/limited anyway right now because it only works if you have no other dependencies than the stdlib. Let's leave it that way for now.

Kha:
I still think https://github.com/leanprover/lean4/pull/763#issuecomment-968954205 is a better approach, but as you probably are the only user of this setup... we can leave it as is for now.

Issue #764, opened 2021-11-04
Help synthesize Prop and boolean expressions


lovettchris:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

It is surprising that something that looks so simple doesn't work.  I think folks from other
programming languages would be very surprised by this, can we fix it?

```lean
#eval (5 > 2) == true
```

Similarly this also fails:

```lean
#eval (5 > 2) = true
```

saying
```
failed to synthesize
  Decidable ((5 > 2) = (true = true))
```

### Steps to Reproduce

see above

**Expected behavior:** [What you expect to happen]

```
true
```

**Actual behavior:** [What actually happens]

```
failed to synthesize instance
  BEq Prop
```

**Reproduces how often:** [What percentage of the time does it reproduce?]

### Versions

You can get this information from copy and pasting the output of `lean --version`,
please include the OS and what version of the OS you're running.

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

tydeu:
@lovettchris 

> ```
> #eval (5 > 2) == true
> ```

This does not work because Lean is inferring `Prop` for both sides (as the `5 < 2 : Prop` comes first) and there is not (and theoretically cannot be) a `BEq` instance for  `Prop`. However, the inverse, `#eval true == (5 < 2) `, works fine (as Lean then infers `Bool` for both sides which solves the problem).

> ```
> #eval (5 > 2) = true
> ```


This one fails because Lean also lacks `Decidable` instance for `Prop` equality. As before, the inverse `#eval true = (5 > 2)` works. Here, though, such an instance could be made like so:

```lean
instance [d : Decidable (p ↔ q)] : Decidable (p = q) :=
  match d with
  | isTrue h => isTrue (propext h)
  | isFalse h => isFalse fun heq => h (heq ▸ Iff.rfl)

#eval (5 > 2) = true -- now produces `true` as desired
```

However, I suspect such an instance is missing because it has to rely on `propext`, which is an axiom.

lovettchris:
[...]

PR #768, opened 2021-11-05
feat: decidable Prop Eq based on propext

tydeu:
From #764


tydeu:
Oops, typo'ed the colon.

PR #769, opened 2021-11-05
chore: update Lake

tydeu:
Has an number of changes with the most important one being a fix for a bug that was breaking the [mathport](https://github.com/leanprover/mathport) build.


Kha:
So far I'm happy with the submodule :)

PR #24, opened 2019-07-30
one more misalignment

Kha:

PR #774, opened 2021-11-09
fix: pass `--trust` to frontend

tydeu:
During a discussion a few weeks ago on Lake issue [#24](https://github.com/leanprover/lake/issues/24), I discovered that the Lean shell doesn't actually pass its trust level to the frontend.  Since this appears to be a quick fix, I went ahead and made this PR.


Kha (approved):

tydeu:
cc @leodemoura

PR #784, opened 2021-11-12
feat: use Lake to resolve modules (WIP)

tydeu:
This is just a proof of concept of the idea we have been tossing around on the Zulip thread. The Lake changes needed to make it work are not even part of the PR yet However, this proof of concept does demonstrate a few design decision that I think are worth discussing before (and if) this change moves forward.

1. **Leanpkg support.** Most relevant Lean 4 projects have migrated to Lake. Thus, is it fine to drop leanpkg support in the server and simply have one code path which goes through Lake? Or should an alternate code path retaining the old behavior be kept for now?

2. **Resolving modules in the `lean` CLI.** Currently Lake is only used in the server to resolve modules. The direct `lean` command line interface will still just use `LEAN_PATH` as before. Conceivably, this could be changed as well by having a new `LEAN_MODUULES` environment variable that contains preresolved module to olean mappings (in JSON form). The `lean` executable could even conceivably call out to `lake print-paths` to build these on the fly if `LEAN_MODULES` is empty. Some road bumps to this approach are: (i) Lake does not exist in stage 0, so the latter strategy would not function there yet, (ii) *replacing* `LEAN_PATH` with `LEAN_MODULES` would require an update to an code not using Lake run Lean code. However, this could be avoided by still falling back to `LEAN_PATH` resolution for the time being (thus having both `LEAN_PATH` and `LEAN_MODULES` support). 

3. **Import-specific error messages.** In the current proof of concept, the new system is only used to provided per-import "could not find olean for X" messages. This could be further augmented by adding a per-module `errors` field to the interop, thus allowing the server to show per-import build errors.

4. **Missing import tolerance.** Should the server continue elaboration even if some modules are missing?

### DIFF
--- a/src/Lean/Elab/Import.lean
+++ b/src/Lean/Elab/Import.lean
@@ -14,10 +14,10 @@ def headerToImports (header : Syntax) : List Import :=
     let id      := stx[2].getId
     { module := id, runtimeOnly := runtime }
 
-def processHeader (header : Syntax) (opts : Options) (messages : MessageLog) (inputCtx : Parser.InputContext) (trustLevel : UInt32 := 0)
+def processHeader (header : Syntax) (opts : Options) (messages : MessageLog) (inputCtx : Parser.InputContext) (trustLevel : UInt32 := 0) (preresolved : NameMap System.FilePath := {})
     : IO (Environment × MessageLog) := do
   try
-    let env ← importModules (headerToImports header) opts trustLevel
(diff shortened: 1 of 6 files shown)


tydeu:
That time when you misspell `feat` as `fear` and don't notice until the bot informs you. 🤣

Kha:
Closing, we should revisit when we have more pressing needs for more granularity in LEAN_PATH

PR #522, opened 2021-06-10
Hover & go-to for syntax

Kha:
Stores elaborator (incl. macro) declaration names in the info tree and uses them in the server for the mentioned actions, falling back to the parser declaration name (assumed to be equal to the syntax kind). Restricted to the first, non-nested token of a syntax for now such as not to show hovers literally everywhere.

This draft shows the declaration name in place of a missing docstring since we don't exactly have many of those yet.

image

TBD:
- [ ] How to deal with overloaded elaborators, e.g. `expandParen` and `elabDeclaration`. We might want to split them up, incurring some hopefully insignificant backtracking overhead. `declaration` should probably become a syntax category anyway, which would also solve the issue that the logic is not triggered on e.g. the `def` token because it is too nested.
- [ ] The syntax/elaborator must actually be imported for this to work, which is kind of an issue for built-in syntax :scream:

PR #785, opened 2021-11-12
Record doc strings & declaration ranges of builtin parsers & elaborators

Kha:
Solves the second issue mentioned in https://github.com/leanprover/lean4/pull/522. For demonstration, also added short doc strings to most builtin term elaborators.

Lean is a powerful language, but some of its special syntax can be obscure for newcomers. I believe it is an open question of how helpful editor features like on-hover doc strings could be in teaching (or at least naming) these concepts, how much that helps more experienced users as well, and what the appropriate length of such doc strings should be (which surely is different depending on the experience level, so we would need to find some middle ground). This change allows us to start exploring these questions and refining the existing hover implementation. The tests have already uncovered one scenario where an unknown identifier should not fall back to the parent notation's doc string, for example.

### DIFF
--- a/src/Lean/Elab/BuiltinNotation.lean
+++ b/src/Lean/Elab/BuiltinNotation.lean
@@ -144,7 +144,7 @@ then aborted. -/
 
 /--
 `dbg_trace e; body` evaluates to `body` and prints `e` (which can be an
-interpolated string litera) to stderr. It should only be used for debugging. -/
+interpolated string literal) to stderr. It should only be used for debugging. -/
 @[builtinMacro Lean.Parser.Term.dbgTrace]  def expandDbgTrace : Macro := fun stx =>
   let arg  := stx[1]
   let body := stx[3]


Kha:
Ironically, some of the more obscure notations like nested `do` actions and cdot notation are not covered by this because they are not elaborators per se. We should be able to manually inject `TermInfo` nodes referencing "fake" elaborator decls for them, but I'll leave that as future work.

Kha:
Let's give the "temporarily fork Lake to fix the build" process a try. If there are no objections, I will merge both later today.

Kha:
Ah, this is a bit spooky: because of GitHub's "improper" separation of forks (I assume), the submodule keeps working even if the commit is not actually contained in leanprover/lake.

src/Lean/Elab/BuiltinNotation.lean
```diff
@@ -127,6 +142,9 @@ private def elabTParserMacroAux (prec lhsPrec : Syntax) (e : Syntax) : TermElabM
   | some code => `(if $cond then $body else panic! ("assertion violation: " ++ $(quote code)))
   | none => `(if $cond then $body else panic! ("assertion violation"))
 
+/--
+`dbg_trace e; body` evaluates to `body` and prints `e` (which can be an
+interpolated string litera) to stderr. It should only be used for debugging. -/
```
gebner, line 147:
```suggestion
interpolated string literal) to stderr. It should only be used for debugging. -/
```

PR #791, opened 2021-11-13
doc: add macro guide, light reorganizing

ammkrn:
Sebastian suggested moving the macro guide from Mathlib4 upstream into the Lean 4 manual (called `macro_overview.md` here). There was some discussion about where to put it; the section on metaprogramming seemed like the most appropriate place. I did a little bit of reorganizing to make some items more discoverable (e.g. the user-defined notation section).


doc/SUMMARY.md
```diff
@@ -24,8 +24,11 @@
 - [Tactics](./tactics.md)
 - [Syntax Extensions](./syntax.md)
   - [The `do` Notation](./do.md)
+  - [User-defined notation](./notation.md)
   - [String Interpolation](./stringinterp.md)
 - [Metaprogramming](./metaprogramming.md)
+  - [Macro Overview](./macro_overview.md)
+  - [A Guided Example](./metaprogramming_example.md)
```
Kha, line 31: There is some overlap between the topics, but I think these would better be placed under "Syntax Extensions". We should probably start differentiating between built-in and custom syntax extensions, but that doesn't have to be in this PR. The short `metaprogramming.md` can probably be removed.

doc/macro_overview.md
```diff
@@ -0,0 +1,385 @@
...
+
+## How macros are handled
+
+The general procedure is as follows:
+
+1. Lean parses the actual code in a given file, creating a Lean AST which
+   contains any unexpanded macros.
+
+2. Lean repeats the cycle (elaboration ~> (macro hygiene and expansion) ~>
+   elaboration...)
```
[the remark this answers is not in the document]
ammkrn: Good point. From pp.6-7 of the paper it looks like declarations and commands are the units for this, is that right?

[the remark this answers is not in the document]
Kha: Yes, with declarations being a subset of commands

ammkrn:
@Kha Thanks for the suggestions, they've been implemented in d692e41. I'll leave the last conversation open, you can close it if you're find with the way the SUMMARY file looks.

Kha:
Thanks!

doc/macro_overview.md
```diff
@@ -0,0 +1,385 @@
...
+allowing users to extend the language with new syntactic constructs without
+having to actually expand the core language, and b) allowing users to automate
+tasks that would otherwise be extremely repetitive, time-consuming, and/or
+error-prone.
+
+A motivating example is set builder notation. We would like to be able to write
+the set of natural numbers 0, 1, and 2 as just `{0, 1, 2}`. However, Lean does
+not natively support this syntax, and the actual definition of a set in Mathlib
+does not let us just declare sets in this manner; naively using the set API
+would force us to write `Set.insert 1 (Set.insert 2 (Set.singleton 3))`.
```
tydeu, line 25: Sorry for being very late to the conversation, but this wording seems awkward now that this documentation is part of the Lean core rather than Mathlib.

ammkrn: I did think about replacing this with the `List` literal macro in the prelude, but it's not as simple as the `Set` one. If you can find a better example from the prelude that demonstrates a similar idea you're welcome to change it; I think this isn't terrible since the macro is recreated in full below the text.

tydeu: Sorry, my point was more the mention of 'the actual definition of a set in Mathlib` -- it is somewhat awkward mention a third party library in the Lean core's documentation. Admittedly, the `Set` part is also a probably a little confusing to CS people as they would probably think of the `RBSet` /  `HashSet` from `Std` in the Lean core first, which is a very different kind of set than the `Set` from mathlib. 

I am little curious, though, why do you think the Set macro is simpler than the List one? They both can expand similarly `{0,1,2}` to `(Set.inserrt 1 (Set.insert 2 (Set.insert 3 ∅)` and `[0,1,2]` to `List.cons 1 (List.cons 2 (List.cons 3 ∅)`.

ammkrn: They expand similarly, but the definition of the list one is more complex than the set one: https://github.com/leanprover/lean4/blob/653ff184a817572b13d9a2a97aede3a3de7e925e/src/Init/Notation.lean#L190

tydeu: True, the actual implementation is  more complex (for optimization reasons), but for demonstration purposes they could be written essentially the same way.

ammkrn: I agree with the overall point you're making; if you want to swap the `Set` thing out to talk about/demo a simplified version of the list macro I'm all for it.

PR #797, opened 2021-11-15
feat: simpLocation

kim-em:
This refactors `evalSimp`, the implementation of the `simp [...] at ...` syntax, into two functions. Now `evalSimp` just handles the `Syntax` parsing, while `simpLocation` dispatches to the `simpGoal` tactic in `Lean.Meta.Tactic.Simp` according to a `location` argument.

I want this separation so that when writing other tactics, I can invoke the simplifier using a `location`, without having to synthesize a `Syntax`. (The specific application is in porting [`split_ifs`](https://github.com/leanprover-community/mathlib/blob/master/src/tactic/split_ifs.lean#L35).)

See also discussion on [zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/simpLocation).

### DIFF
--- a/src/Lean/Elab/Tactic/Simp.lean
+++ b/src/Lean/Elab/Tactic/Simp.lean
@@ -53,7 +53,7 @@ inductive Simp.DischargeWrapper where
   | default
   | custom (ref : IO.Ref Term.State) (discharge : Simp.Discharge)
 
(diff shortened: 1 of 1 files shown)


leodemoura:
@semorrison BTW, we have a `split` tactic in core Lean that splits `if` and `match` expressions. It has many features for `match`. See `split*.lean` tests at `tests/lean/run` for examples. That being said, it may be incompatible with the `splif_if` tactic.

PR #795, opened 2021-11-15
Bundle LLVM on all platforms

Kha:
Building on top of #733 and #736, add support for macOS to complete the trio.

```
├── [9.2K]  LICENSE
├── [7.8M]  bin
│   ├── [166K]  clang
│   ├── [1.5M]  lake
│   ├── [5.9M]  ld64.lld
│   ├── [ 16K]  lean
│   ├── [ 43K]  leanc
│   ├── [ 839]  leanmake
│   └── [218K]  leanpkg
├── [118K]  include
│   ├── [ 40K]  clang
│   │   ├── [2.8K]  limits.h
│   │   ├── [ 583]  stdalign.h
│   │   ├── [1.1K]  stdarg.h
│   │   ├── [7.1K]  stdatomic.h
│   │   ├── [ 897]  stdbool.h
│   │   ├── [3.5K]  stddef.h
│   │   ├── [ 22K]  stdint.h
│   │   └── [ 510]  stdnoreturn.h
│   └── [ 78K]  lean
│       ├── [ 259]  config.h
│       ├── [ 77K]  lean.h
│       ├── [ 551]  lean_gmp.h
│       └── [ 387]  version.h
├── [467M]  lib
│   ├── [762K]  clang
│   │       └── [762K]  lib
│   │           └── [762K]  darwin
│   │               └── [761K]  libclang_rt.osx.a
│   ├── [355M]  lean
│   │   ├── [ 24M]  Init
│   │   ├── [8.3M]  Lake
│   │   ├── [220M]  Lean
│   │   ├── [1.4M]  Leanpkg
│   │   ├── [5.4M]  Std
│   │   ├── [3.2M]  libInit.a
│   │   ├── [ 46M]  libLean.a
│   │   ├── [865K]  libStd.a
│   │   ├── [3.6M]  libleancpp.a
│   │   ├── [447K]  libleanrt.a
│   │   ├── [ 39M]  libleanshared.dylib
│   │   └── [3.4M]  src
│   │       ├── [488K]  Init
│   │       ├── [2.6M]  Lean
│   │       ├── [ 15K]  Leanpkg
│   │       ├── [ 82K]  Std
│   │       ├── [  64]  bin
│   │       ├── [ 160]  cmake
│   │       ├── [ 160]  include
│   │       │   └── [  64]  lean
│   │       ├── [  64]  initialize
│   │       ├── [  64]  kernel
│   │       ├── [136K]  lake
│   │       │   ├── [126K]  Lake
[...]

PR #800, opened 2021-11-16
Symlink to source from build dir for go-to-definition

Kha:
Someone might want to check that this will *not* do a source copy on msys2, which would not be quite as helpful...


Kha:
Ah, the CI failure will be fixed automatically after merging #795. So let's wait for that one.

Kha:
It does work on Windows using Developer Mode, except for https://github.com/leanprover/lean4/issues/810

PR #803, opened 2021-11-17
Fix error message from cpack

lovettchris:


Kha:
That line is currently in use (see absence of message in our release CI jobs), but we should be able to remove it starting with #795: https://github.com/leanprover/lean4/pull/795/commits/f1e13756bd4ea2ed300f001520316ae893cfa07d

lovettchris:
If I go back to https://github.com/leanprover/lean4/actions/runs/1466228779 I see the error in the builds called "Linux", "Linux Debug", "Linux fsanitize", and "Windows", but for some reason not in "macOS" or the NIX based build named "Linux release"...

image

But I see your LLVM PR is disabling the Pack step?

lovettchris:
And on macOs I see this:
image

so not sure why cpack doesn't complain on macOS but there is nothing matching "*.*" there.

which only leaves the NIX build which I don't have handy...

Kha:
Resolved with #795

PR #805, opened 2021-11-18
Expose API for sysroot search & compile flags

Kha:
Since `leanc` is not meant to be used for generic C compilation anymore, instead expose its flags in a convenient manner for Lake. Also expose a function inspired by Lake's code for locating the sysroot, which can also be useful for other tools (/cc @insightmind).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,8 +45,7 @@ jobs:
             os: macos-latest
             release: true
             shell: bash -euxo pipefail {0}
-            # Mojave, oldest maintained version as of 2021
-            CMAKE_OPTIONS: -DCMAKE_OSX_DEPLOYMENT_TARGET=10.14
+            CMAKE_OPTIONS: -DCMAKE_OSX_DEPLOYMENT_TARGET=10.15
             llvm-url: https://github.com/leanprover/lean-llvm/releases/download/13.0.0/lean-llvm-x86_64-apple-darwin.tar.zst
             prepare-llvm: script/prepare-llvm-macos.sh lean-llvm*
             binary-check: otool -L
--- a/script/prepare-llvm-macos.sh
+++ b/script/prepare-llvm-macos.sh
@@ -29,11 +29,11 @@ cp $SDK/usr/lib/libSystem.tbd stage1/lib/libc
 # use for linking, use system libs for running
 gcp llvm/lib/lib{c++,c++abi,unwind}.dylib stage1/lib/libc
 echo -n " -DLEAN_STANDALONE=ON"
-echo -n " -DCMAKE_C_COMPILER=$PWD/stage1/bin/clang -DCMAKE_CXX_COMPILER=$PWD/llvm/bin/clang++"
-# allow C++ code to include /usr since it needs quite a few more headers
-# need no-macro-redefined for weird clang stdint.h
-echo -n " -DLEAN_EXTRA_CXX_FLAGS='--sysroot $PWD/llvm --stdlib=libc++ -I $SDK/usr/include -Wno-macro-redefined'"
+# do not change C++ compiler; libc++ etc. being system libraries means there's no danger of conflicts,
+# and the custom clang++ outputs a myriad of warnings when consuming the SDK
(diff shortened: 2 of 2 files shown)


tydeu:
@Kha Neato! A few minor comments:

1. I would still really like to be able access the flags alone directly (without the include / library path). That way I can still use the richer `LeanInstall` of Lake to fill in the paths. The combined function is quite nice for a lot of simpler use cases (so I would keep it as well).

2. I would name `findLeanSysroot?` just `findSysroot?` since it is already in the `Lean` namespace.

src/Lean/Compiler/FFI.lean
```diff
@@ -0,0 +1,25 @@
...
+/-- Return C compiler flags for including Lean's headers. -/
+def getCFlags (leanSysroot : FilePath) : Array String :=
+  #["-I", (leanSysroot / "include").toString] ++ (getLeancExtraFlags ()).trim.splitOn
+
+@[extern "lean_get_linker_flags"]
+private constant getBuiltinLinkerFlags (linkStatic : Bool) : String
+
+/-- Return linker flags for linking against Lean's libraries. -/
+def getLinkerFlags (leanSysroot : FilePath) (linkStatic := true) (gmp := "-lgmp") : Array String :=
+  #["-L", (leanSysroot / "lib" / "lean").toString, gmp] ++ (getBuiltinLinkerFlags linkStatic).trim.splitOn
```
tydeu, line 23: My specific suggestion for the flag file is:

```suggestion
@[extern "lean_get_leanc_extra_flags"]
private constant getLeancExtraFlags : Unit → String

/-- Return C compiler flags for compiling C files outputted by Lean. -/
def extraCFlags : Array String :=
  getLeancExtraFlags () |>.trim.splitOn

/-- Return C compiler flags (with include path) for compiling C files outputted by Lean. -/
def cFlags (leanSysroot : FilePath) : Array String :=
  #["-I", (leanSysroot / "include").toString] ++ extraCFlags

@[extern "lean_get_linker_flags"]
private constant getBuiltinLinkerFlags (linkStatic : Bool) : String

/-- Return linker flags for linking against Lean's libraries. -/
def extraLinkerFlags (linkStatic := true) : Array String :=
  getBuiltinLinkerFlags linkStatic |>.trim.splitOn

/-- Return linker flags (with library paths) for linking against Lean's libraries. -/
def linkerFlags (leanSysroot : FilePath) (linkStatic := true) (gmp := "-lgmp") : Array String :=
  #["-L", (leanSysroot / "lib" / "lean").toString, gmp] ++ extraLinkerFlags
```

tydeu:
@Kha Also, a point of confusion for me: why are the `leanc` internal C/linker flags not part of the exposed flags?

Kha:
They wouldn't be all that internal then, would they!

Kha:
>     1. That way I can still use the richer `LeanInstall` of Lake to fill in the paths.

Please specify? Why does it matter whether Lean or Lake assembles the path if the outcome is the same?

Kha:
>     3\. Also, style-wise, why are the flag functions call `get*` (which is usually reserved for impure/lazy/moadic code)

I'm not aware of such a convention? In contrast, I believe that most functions should contain a verb.

tydeu:
> > ```
> > 1. That way I can still use the richer `LeanInstall` of Lake to fill in the paths.
> > ```
> 
> Please specify? Why does it matter whether Lean or Lake assembles the path if the outcome is the same?

Technically, the paths of `LeanInstall` are separate so the `oleanDir` doesn't necessarily have to be in the standard location. Though, admittedly, that is unlikely to occur.

tydeu:
> > ```
> > 3\. Also, style-wise, why are the flag functions call `get*` (which is usually reserved for impure/lazy/moadic code)
> > ```
> 
> I'm not aware of such a convention? In contrast, I believe that most functions should contain a verb.

I was basing this comment mostly on how the `Lean.version` definitions were patterned. I took the arguments more as templating for the flags (i.e., like a constructor which usually does not have verb).  I just generally associate the verb `get` in functional programming with state (after all that is what plain `get` deals with) not pure functions. But that's just my personal preference, feel free to ignore my comment.

PR #811, opened 2021-11-19
Show fully-qualified name in hover text

### LINKED ISSUE #749: Hover should show full name of declaration
When hovering over a declaration, including its namespaces in the hover would yield valuable information without taking up too much space. The type should probably be rendered as before, however, since it can potentially contain many names.

larsk21:
Show the fully-qualified name of a declaration when hovering over it. `unresolveNameGlobal` is changed to return the fully-qualified name when `pp.fullNames` is set to `true`, which is the description of this option.

The type shown in the hover is not changed.

semi-closes #749

### DIFF
--- a/src/Lean/PrettyPrinter/Delaborator/Builtins.lean
+++ b/src/Lean/PrettyPrinter/Delaborator/Builtins.lean
@@ -59,7 +59,11 @@ def delabSort : Delab := do
 
 
 def unresolveNameGlobal (n₀ : Name) : DelabM Name := do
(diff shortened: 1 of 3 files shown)


Kha:
Here is a potential issue with the new pp code:
```
constant foo : Nat

namespace Bar

constant foo : Nat

set_option pp.fullNames true
#check _root_.foo  -- should print `_root_.foo`

end Bar
```
I think the cleanest solution is to extend your code and duplicate a bit of logic from `unresolveNameCore`: if `pp.fullNames` is set, check if `resolveGlobalName n₀` resolves to `n₀`. If it doesn't, prepend `rootNamespace`.

PR #814, opened 2021-11-22
Fix LSP syntax highlighting in some cases

Garmelon:
This is a bit of a hack, but it seems to work better than the previous code.


Kha:
Thanks. I've added a more local fix with docs in https://github.com/leanprover/lean4/commit/226121433fa7d6964c94bda4fc2a9bd340c8f9ab

PR #817, opened 2021-11-23
fix: only check for ccache if variables aren't passed explicitly

typesanitizer:
The existing CMake will emit a spurious warning when someone
tries to explicitly pass CMAKE_C(XX)_COMPILER_LAUNCHER, such
as when they point it to sccache instead of ccache.

I couldn't find instructions on how to use linguist to update
`stage0/src/CMakeLists.txt`, so I've updated it manually.
However, that change doesn't show up by default as the file is marked
binary in `.gitattributes`.


Kha:
Thanks!

Issue #813, opened 2021-11-22
simp inconsistently simplifies Function.const


gebner:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

```lean
open Function
@[simp] theorem const_apply {y : β} {x : α} : const α y x = y := rfl
set_option trace.Meta.Tactic.simp true
example (p : Nat → Prop) : p (const Nat 1 2) := by simp -- simplifies
example (p : α → Prop) : p (const Nat a 2) := by simp -- DOES NOT simplify
example (f : α → α) (p : α → Prop) : p (const Nat (f a) 2) := by simp -- simplifies
```

**Expected behavior:** `simp` rewrites using the `const_apply` lemma in each of the three examples.  (At least I think that's a very desirable behavior as it makes the goal much more readable when there are no const-applications.)

What's happening here is that 1) `const` is an abbreviation, and 2) the discrimination tree module inserts the const-application in different ways because it reduces the expression until it reaches a bad key---andthe const-expression reduces to the different terms `?y`, `1`, `a`, `f a` here.

Metavariables are bad keys, so `const_apply` is inserted as `const * * * *`.  The match key in the first example is `const Nat Nat ofNat... ofNat...`, as `1` is a bad key because it is an application (which seems like an oversight though).  In the second example the match key is `a` because free variables are not bad keys and so the const-application is reduced.  In the third example there's again an application and the const is not reduced.

I am not sure if there's an easy fix here.  Some potential solutions come to mind:
 - Make `Function.const` semireducible.  (Generally speaking, adopt the rule that abbreviations should not reduce to bad keys.)
 - Insert multiple copies of the expression into the discrimination tree, that is, once before unfolding const and once afterwards.  (Generally, insert the version before unfolding if the definition body is a bad key.)  This approach obviously has some potential to blow up exponentially.

### Versions

Nightly from Nov 17.

gebner:
> Make `Function.const` semireducible.

[...]

PR #819, opened 2021-11-23
fix: unfold `x<y` in discrimination tree module

gebner:
While looking at #813, I noticed that the call to `isBadKey` was missing a `getAppFn`.  As a result the following fails on master (which we really want):
```lean
example {x y : Nat} (h : y > x) : x < y := by
  simp [h]
```

Unfolding reducible definitions in the discrimination tree has one caveat that I was not aware of before: simp won't apply lemmas that *overlap* with the reducible definition.  For example, it won't apply `(a = b) = true` in `a ≠ b`.  The only reason this works right now is because of the `ne_eq` simp lemma.  (This is also why this PR requires a stage0 update.)


leodemoura:
Merged, but had to regenerate "update stage0" step.

PR #823, opened 2021-11-24
feat: bundle `llvm-ar` with Lean

Kha:
Allows us to build static archives without a system toolchain. The size overhead is minimal: 89KB on Linux.

/cc @tydeu


tydeu:
@Kha Anything in particular you want my input on? Or was that just a heads up?

Kha:
Any issues you can see with using it in Lake. But no, there are not all that many design options here :smile: .

PR #830, opened 2021-11-27
fix: Update IR interpreter enum hygiene

joshuaseaton:
This change addresses -Wreturn-type and -Wimplicit-fallthrough
infractions in the IR interpreter. Moreover, default switch cases for
enum classes are removed to leverage compiler complaints of missing
cases, which ensures that all related switch logic gets thoughtfully
extended when the enums are extended themselves.


Kha:
Looks good, thanks!

PR #832, opened 2021-11-29
Generalize syntax of fvar-accepting tactics & make `this` a keyword 

### LINKED ISSUE #821: `this` cannot be used in macros
(alternative title: `this` should be hygiene-*bending*, not *breaking*)

```lean
macro "foo" : term => `(have := 1; this)
#check foo  -- unknown identifier 'this✝'
```

@leodemoura I believe we had a proposal once that "anaphoric" identifiers like `this` should be *keywords* instead, with an elaborator that unhygienically accesses the scope-less `this` assumption. I don't remember what became of that proposal though :) .

Kha:
Fixes #821

### DIFF
--- a/src/Lean/Elab/Tactic/ElabTerm.lean
+++ b/src/Lean/Elab/Tactic/ElabTerm.lean
@@ -136,7 +136,7 @@ def elabTermForApply (stx : Syntax) (mayPostpone := true) : TacticM Expr := do
     match (← Term.resolveId? stx (withInfo := true)) with
     | some e => return e
     | _      => pure ()
-  elabTerm stx none (mayPostpone := false)
+  elabTerm stx none mayPostpone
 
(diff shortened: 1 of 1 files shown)


src/Init/Notation.lean
```diff
@@ -201,6 +201,11 @@ macro_rules
 
 notation:50 e:51 " matches " p:51 => match e with | p => true | _ => false
 
+-- Declare `this` as a keyword that unhygienically binds to a scope-less `this` assumption (or other binding).
+-- The keyword prevents declaring a `this` binding except through metapgrogramming, as is done by `have`/`show`.
+/-- Special identifier introduced by "anonymous" `have : ...`, `suffices p ...` etc. -/
+macro "this" : term => Syntax.ident SourceInfo.none "this".toSubstring `this []
```
tydeu, line 207: Why does the ident not preserve the source info of the keyword?

Kha: No particular reason, thanks

PR #833, opened 2021-11-30
refactor: auto-generate (builtin)Parenthesizer attribute

gebner:
See discussion on Zulip: https://leanprover.zulipchat.com/#narrow/stream/147302-lean4-dev

There are two ways parenthesizers (and analogously formatters) can be registered in the system:
1. `@[combinatorParenthesizer parser]`
2. `@[parenthesizer kind]` (there is also the built-in variant `@[builtinParenthesizer kind]`)

The first way is only used at compile-time when automatically generating `*.parenthesizer` declarations.  For example, when we define a parser using `many1 ident`, then the parser compiler will compile this into the parenthesizer `many1.parenthesizer ident.parenthesizer` using the `@[combinatorParenthesizer many1] def many1.parenthesizer` definition.

The second way is only used at runtime, when dynamically looking up the correct parenthesizer to use for a given syntax.

At the moment, the `@[parenthesizer]` attribute is only generated for declarations tagged with the `@[termParser]` attribute (and `@[commandParser]`, etc.), assuming that the syntax kind is the same as the declaration name.  This has the effect that only syntax objects from a category parser can be printed, but not their subtrees.  For example `by tac; tac` can be printed but not `tac; tac`:
```lean
set_option trace.Elab.step true
example : True := by skip; trivial
-- failed to pretty print term
```

This PR changes when `@[parenthesizer]` attributes are added.  They are now added for any parser (that the parser compiler encounters) where we can statically determine the syntax kind.  That is, where the parser definition is literally `node kind p` (four other combinators are recognized as well) and `kind` evaluates to a name.  This generalizes the previous heuristic when leading_parser/trailing_parser is used, and fixes the `trace.Elab.step` issue.

### DIFF
--- a/tests/lean/traceTacticSteps.lean.expected.out
+++ b/tests/lean/traceTacticSteps.lean.expected.out
@@ -1,0 +1,44 @@
+[Elab.step] expected type: Sort ?u, term
+True
+[Elab.step] expected type: True, term
+by 
+  skip 
+  trivial
+[Elab.step] 
+  skip 
+  trivial
+[Elab.step] 
+  skip 
+  trivial
+[Elab.step] skip
+[Elab.step] trivial
+[Elab.step] (apply And.intro✝) <;> trivial
+[Elab.step] focus 
+  (
+    apply And.intro✝;
+    all_goals 
+      trivial)
+[Elab.step] 
+  (
+    apply And.intro✝;
+    all_goals 
+      trivial)
+[Elab.step] 
+  (
+    apply And.intro✝;
+    all_goals 
+      trivial)
+[Elab.step] (
+  apply And.intro✝;
+  all_goals 
+    trivial)
+[Elab.step] 
+  apply And.intro✝;
+  all_goals 
+    trivial
+[Elab.step] 
+  apply And.intro✝;
+  all_goals 
+    trivial
+[Elab.step] apply And.intro✝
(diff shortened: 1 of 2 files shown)


tests/lean/traceTacticSteps.lean.expected.out
```diff
@@ -0,0 +1,44 @@
...
+  skip 
+  trivial
+[Elab.step] 
+  skip 
+  trivial
+[Elab.step] skip
+[Elab.step] trivial
+[Elab.step] (apply And.intro✝) <;> trivial
+[Elab.step] focus 
+  (
```
Kha, line 17: Hmm, interesting formatting!

Kha (approved):

tests/lean/traceTacticSteps.lean.expected.out
gebner: Both newlines are because tacticSeq1Indented prints a line before every tactic.  We have quite a few interesting formatting corner cases like that:
```lean
def foo :=
  do
    putStrLn "hello world"

theorem bar :=
  by
    rfl

theorem baz :=
  by
    ·
      rfl
```

You can browse through the [mathport output](https://github.com/leanprover-community/mathlib3port/tree/master/Mathbin) to find more of these artworks.

PR #834, opened 2021-11-30
PoC: enable info trees by default

Kha:
If we want to generate a database for project-wide find all references and similar queries, we need to enable at least a subset of info state generation even when building from the cmdline. Let's try and see what kind of performance overhead this implies.


Kha:
!bench

Kha:
Great, looks like very reasonable overhead. But there's no harm in enabling it only when we really need it.

PR #838, opened 2021-12-01
Fix Nix build on macOS

Smaug123:
(Feel free to ignore this unsolicited PR if it's a pain to review.)

All I did to produce this PR was delete the existing `flake.lock` and allow `nix build` to generate a new one.

With the original `flake.lock`, I was unable to build on aarch64-darwin, Monterey:

```
error: flake 'git+file:///Users/patrick/Documents/GitHub/lean4copy' does not provide attribute 'packages.aarch64-darwin.defaultPackage.aarch64-darwin', 'legacyPackages.aarch64-darwin.defaultPackage.aarch64-darwin' or 'defaultPackage.aarch64-darwin' 
```

I can only assume that some dependency did not yet support Apple Silicon.

After this change, `nix build` Just Works (tm) except for right at the end where the code signing fails because it can't find `codesign_allocate`; I can almost certainly fix that locally by installing the Apple developer tools, but I'm currently trying to work out the correct fix to nixpkgs to ensure `codesign_allocate` is used from the Nix store rather than ambiently (https://github.com/NixOS/nixpkgs/issues/148189).

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "flake-utils": {
       "locked": {
-        "lastModified": 1623875721,
-        "narHash": "sha256-A8BU7bjS5GirpAUv4QA+QnJ4CceLHkcXdRp4xITDB0s=",
+        "lastModified": 1638122382,
+        "narHash": "sha256-sQzZzAbvKEqN9s0bzWuYmRaA03v40gaJ4+iL1LXjaeI=",
         "owner": "numtide",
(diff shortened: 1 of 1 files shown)


Kha:
Ah, interesting. I guess we're the first ones to try and run the wrapped linker outside of a stdenv derivation, which triggers the missing executable issue.

Smaug123:
Problem is fixed as of 171624d7a797976e79a546f754e8c4771ee52b20.

PR #842, opened 2021-12-03
doc: replace quickstart leanpkg info with info about lake.

lovettchris:
time to tell folks how to use lake.

### DIFF
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -36,7 +36,8 @@ You are set up!
 
 ## Create a Lean Project
 
-You can now create a Lean Project in a new folder, run `lake init foo` from the the "View > Terminal" to create a package, followed by `lake build` to get an executable version of your Lean program.
+You can now create a Lean project in a new folder. Run `lake init foo` from the "View > Terminal" to create a package, followed by `lake build` to get an executable version of your Lean program.
+On Linux/macOS, you first have to follow the instructions printed by the Lean installation or log out and in again for the Lean executables to be available in you terminal.
 
 Note: Packages **have** to be opened using "File > Open Folder..." for imports to work.
 Saved changes are visible in other files after running "Lean 4: Refresh File Dependencies" (`Ctrl+Shift+X`).


doc/dev/index.md
```diff
@@ -63,6 +63,14 @@ cd src
 # make `lean` etc. point to stage0 anywhere inside `src`
 elan override set lean4-stage0
 ```
+
+Note that `lake build` will not work on this kind of linked elan toolchain because
+it will be missing the bundled clang compiler and headers and libraries.
```
Kha, line 68: If you built Lean from a terminal, that same compiler should be usable for driving `leanc`. I don't want to worry about setting up the bundled toolchain on developers' machines, this was strictly intended for CI.

doc/quickstart.md
```diff
@@ -14,23 +14,24 @@ See [Setup](./setup.md) for other ways and more details on setting up Lean.
...
-    You should get a syntax-highlighted file with a "Lean Infoview" on the right that tells you the installed Lean version when placing your cursor on the last line.
+    You should get a syntax-highlighted file with a "Lean Infoview" on the right.
+    Add the following #eval statement which tells you the installed Lean version
+    when placing your cursor on the last line.
 
     ![successful setup](images/code-success.png)
 
-1. You are set up! Try opening a Lean folder containing a package file `leanpkg.toml`. You can create your own packages using `leanpkg init` on the command line.
-   Packages **have** to be opened using "File > Open Folder..." for imports to work.
-   Saved changes are visible in other files after running "Lean 4: Refresh File Dependencies" (`Ctrl+Shift+X`).
```
gebner, line 36: I don't think we should remove these important warnings.

doc/setup.md
```diff
@@ -21,21 +21,28 @@ $ elan default leanprover/lean4:nightly
...
 Foo/
-  A.lean  # further files, import via e.g. `import Foo.A`
-  A/...   # further nesting
-build/  # `leanpkg` output directory
+  A.lean       # further files, import via e.g. `import Foo.A`
+  A/...        # further nesting
+build/         # `lake` build output directory
+```
+
+After running `lake build` you will see a binary named `.\build\bin\Foo` and when you run it you should see the output:
```
gebner, line 42:
```suggestion
After running `lake build` you will see a binary named `./build/bin/Foo` and when you run it you should see the output:
```
I believe that will work on both Windows and Linux.

doc/dev/index.md
lovettchris: ok, but my Note above is true right? At least on windows it doesn't seem to work and I assume that's expected.  I can use still this local build in vscode as a lean --server though.

doc/setup.md
lovettchris: woops, fixed.

doc/quickstart.md
lovettchris: Ah, yes, I added back the note at the bottom.

doc/dev/index.md
Kha: You didn't tell me about your terminal or compiler, so can't really say anything about that. I assume it works out of the box in a mingw64 terminal, with `cc` in the PATH.

lovettchris: Oh, I see you are right, it does work in mingw64 .  Just not outside, whereas your normal toolchain install does work outside.

lovettchris: Adding clang to the PATH on windows might work, but I find these are missing from `lean-llvm-x86_64-w64-windows-gnu.tar.zst`:
- libc++.dll
- zlib1.dll

 do you know where you get those from? I don't see any mention of these in `\lean4\script\prepare-llvm-mingw.sh`
image

lovettchris: Ok, I added the disclaimer here where it belongs:
https://github.com/leanprover/lean4/pull/842/commits/fc9136f24347eb040157bc7a43bcb012efbea537

Kha: > Ok, I added the disclaimer here where it belongs

Then we can remove this paragraph?

Kha: > do you know where you get those from?

https://github.com/leanprover/lean4/blob/171624d7a797976e79a546f754e8c4771ee52b20/script/prepare-llvm-mingw.sh#L21

doc/quickstart.md
```diff
@@ -14,23 +14,27 @@ See [Setup](./setup.md) for other ways and more details on setting up Lean.
     del elan-init.ps1
     ```
     See the [elan repo](https://github.com/leanprover/elan) for other installation options and details.
```
Kha, line 16: Should we change this to installing elan from the VS Code extension?

doc/quickstart.md
```diff
@@ -14,23 +14,27 @@ See [Setup](./setup.md) for other ways and more details on setting up Lean.
...
-    
+
 1. Install [VS Code](https://code.visualstudio.com/).
 
-1. Open VS Code and install the `lean4` extension.
+1. Create a new folder called `Foo` and type:
```
Kha, line 20: I think it is Lake convention to have a lowercase package directory & name (/cc @tydeu) (not to be confused with the .lean filename/directory *inside* the package). You might also want to use `lake new` to combine the two steps.

doc/quickstart.md
```diff
@@ -14,23 +14,27 @@ See [Setup](./setup.md) for other ways and more details on setting up Lean.
...
-    ```lean
-    import Leanpkg
+    ![installing the vscode-lean4 extension](images/code-ext.png)
 
-    #eval Leanpkg.leanVersionString
-    ```
-    You should get a syntax-highlighted file with a "Lean Infoview" on the right that tells you the installed Lean version when placing your cursor on the last line.
+    You should get a syntax-highlighted file with a "Lean Infoview" on the right.
+    Add the following #eval statement which tells you the installed Lean version
+    when placing your cursor on the last line.
```
Kha, line 31: Why did you remove the code to be typed? Where is the file coming from now, is it opened automatically by VS Code?

doc/setup.md
```diff
@@ -21,21 +21,28 @@ $ elan default leanprover/lean4:nightly
 $ elan override set leanprover/lean4:stable
 ```
 
-### `leanpkg`
+NOTE: there are no `lean4:stable` releases yet, so for now
+just use `leanprover/lean4:nightly`
```
Kha, line 25: There sure are, does it not work for you?

doc/quickstart.md
Kha: In the end it's not clear to me if the quickstart should stick with the simpler stand-alone file setup or not.

doc/quickstart.md
tydeu: Correct.

doc/setup.md
Vtec234: As phrased I suppose this is incorrect given that stable releases do exist, but I would support noting that nightly is the recommended version to get. People come to Zulip every now and then with weird issues the solution to which is just "you're on stable and the version is very old, install nightly instead".

Kha: Well, this PR is now linked to an issue on the next milestone, so let's work on improving that part :) . However, we do actually recommend nightlies in the quickstart.

lovettchris: I went with the simpler vscode setup.  Please see new version, thanks.

doc/quickstart.md
lovettchris: Went with the simpler vscode setup, see new version.

doc/quickstart.md
lovettchris: Fixed

doc/quickstart.md
lovettchris: Done.

doc/dev/index.md
lovettchris: Thanks and paragraph removed.

doc/quickstart.md
```diff
@@ -3,38 +3,44 @@
...
+1. Create a new folder called `foo` and add a file named `hello.lean`
+containing the following:
+
     ```lean
-    lake init Foo
+    import Lake
+    #eval Lake.leanVersionString
+    ```
+
+    and a file named `lean-toolchain` containing just this:
```
Kha: Hmm, I concede that this is hardly better than before. What is still missing in the extension to make this work purely in VS Code, ideally without saving any file at all?

gebner: I think I must have missed the comment you're replying to, what was the previous proposal?  And what do you have in mind for the automatic set up?

Kha: The previous version installed elan from the cmdline to then run `lake init`. The ideal, no-hassle on-boarding I have in mind is:
1. install extension
2. File > New File
3. either "Select a language" > lean4 or save as `.lean` (does the latter default to Lean 4 when the Lean 3 extension is not installed?)
4. it should just work, where "just work" probably implies defaulting to nightly

lovettchris: Ok, love to get your input on this on then https://github.com/leanprover/vscode-lean4/pull/83

doc/make/msys2.md
```diff
@@ -74,6 +74,14 @@ The following linux command will do that:
 cp $(ldd lean.exe | cut -f3 -d' ' | grep mingw) .
 ```
 
+However, if you plan to use this build to compile lean programs
+to executable binaries using `lake build` in normal Windows command
+prompt outside of msys2 environment you will also need to add a windows
+version clang to your path.  This is done in the `.github\workflows\ci.yml`
+by downloading `lean-llvm-x86_64-w64-windows-gnu.tar.zst`.  You will
+see how these `clang` binaries are provided for you in a .elan toolchain
+install of Lean 4.
```
semorrison, line 83: I don't really understand these four lines, and I don't think pointing to a CI configuration file to imitate is the right approach here. :-) (And why did you write `.elan` instead of just `elan`?)

Kha:
I've cherry-picked the resolved parts onto master in c3a9860dc8cd6719ba95946070aa9a2b40c9d1f4.

lovettchris:
> I've cherry-picked the resolved parts onto master in [c3a9860](https://github.com/leanprover/lean4/commit/c3a9860dc8cd6719ba95946070aa9a2b40c9d1f4).

@Kha , rats, I really wanted to get rid of the user having to manually run curl https://raw.githubusercontent.com/leanprover/elan/master/elan-init.sh which is not needed if you just start with the VS code extension and let it do that for you.

doc/quickstart.md
Kha: Is this resolved now on the extension side?

lovettchris: Right, I will update this doc and PR for the new behavior of the vscode extension...

lovettchris: Ok, PR is updated.  I had to switch to `Leanpkg.leanVersionString` because this seems `Lake.leanVersionString` has been removed.  
```
unknown identifier 'Lake.leanVersionString'
```

tydeu: @lovettchris why  not just use `Lean.versionString`?

doc/make/msys2.md
lovettchris: You are right, I removed those details.

doc/quickstart.md
lovettchris: Doh!  Done, had to update the screenshot too.

lovettchris: See [quickstart.md](https://github.com/leanprover/lean4/blob/2286262eaa08e77ed0b55ade3d2a95c8801342d3/doc/quickstart.md)

Kha: Why do you insist on using a named file? I thought unnamed files just worked now?

lovettchris: Are you talking about moving step (2) after the current step (3) and making it like this?

3. Start VS code and select File/New File
4. Click "Select a Language" and type in "lean4" then paste in the following program:
```lean
import Lean
#eval Lean.versionString
```

Problem is I just tried this and it doesn't work:
```
'lean' is not recognized as an internal or external command,
operable program or batch file.
[Info  - 11:54:27 AM] Connection to server got closed. Server will restart.
'lean' is not recognized as an internal or external command,
```
we'll need a new issue opened to fix this.  But quite frankly I think creating a real file on disk is going to prepare them a lot better for playing with Lake Projects.

Kha: So https://github.com/leanprover/vscode-lean4/issues/92 isn't fixed after all?

lovettchris: That scenario works, just not together with bootstrapping apparently.

lovettchris: my PR on "[multi-folder support](https://github.com/leanprover/vscode-lean4/pull/104)" fixes it though.

Kha: Nice. I really think the steps in this file should be the quickest possible to get to a working Lean version, *especially* as long as we don't have a working web version. I don't think most people turn their very first Lean file into a package later, but we can add steps for that as well.

lovettchris: I agree, see updated [quickstart.md](https://github.com/leanprover/lean4/blob/2286262eaa08e77ed0b55ade3d2a95c8801342d3/doc/quickstart.md)

lovettchris:
I agree, see updated [quickstart.md](https://github.com/leanprover/lean4/blob/2286262eaa08e77ed0b55ade3d2a95c8801342d3/doc/quickstart.md)

doc/quickstart.md
```diff
@@ -3,33 +3,48 @@
...
+    you place your cursor at the end of the #eval statement.
 
     ![successful setup](images/code-success.png)
 
-1. You are set up! You can now also run `lake init foo` from the command line to create a package, followed by `lake build` to get an executable version of your Lean program.
+You are set up!
+
+## Create a Lean Project
+
+You can now create a Lean Project in a new folder, run `lake init foo` from the the "View > Terminal" to create a package, followed by `lake build` to get an executable version of your Lean program.
```
Kha, line 40:
```suggestion
You can now create a Lean project in a new folder. Run `lake init foo` from the "View > Terminal" to create a package, followed by `lake build` to get an executable version of your Lean program.
On Linux/macOS, you first have to follow the instructions printed by the Lean installation or log out and in again for the Lean executables to be available in you terminal.
```

Kha (approved):
Great! Happy to merge as soon as the fix is deployed.

doc/quickstart.md
```diff
@@ -3,33 +3,48 @@
...
 
 Note: Packages **have** to be opened using "File > Open Folder..." for imports to work.
 Saved changes are visible in other files after running "Lean 4: Refresh File Dependencies" (`Ctrl+Shift+X`).
+
+## Troubleshooting
+
+**The InfoView says "Waiting for Lean server to start..." forever.**
+
+Check that the VS Code Terminal is not prompting you with `Press ENTER key to start Lean:`.  If so click in the terminal window and pressthe  ENTER key.
```
Kha, line 49: Do we still have such prompts?

doc/quickstart.md
```diff
@@ -3,33 +3,48 @@
...
+    ```
+    info: syncing channel updates for 'nightly'
+    info: latest update on nightly, lean version nightly-2021-12-05
+    info: downloading component 'lean'
+    ```
+
+1. While it is installing you can paste the following Lean program into the new file:
+
     ```lean
+    import Lean
```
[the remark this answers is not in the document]
lovettchris: Fixed, thanks.

doc/quickstart.md
lovettchris: Thanks

lovettchris:
Yes, see https://github.com/lovettchris/vscode-lean4/blob/master/lean4-infoview/src/infoview/main.tsx#L57

Kha:
> Yes, see https://github.com/lovettchris/vscode-lean4/blob/master/lean4-infoview/src/infoview/main.tsx#L57

No, I meant the "Press ENTER key to start Lean" part.

Kha:
I see the prompt is still there in vscode-lean4, but I'm not entirely sure about its purpose

lovettchris:
Ah, yes, I was going to remove it by auto-exiting the process, then I wondered about "error conditions", if the elan install fails there will be an error message in that output window, and if I auto-close it the user won't know why it failed...

Kha:
Then, should we restrict it to non-zero exit codes from the installer?

PR #845, opened 2021-12-03
Report processing errors in file header as error progress

larsk21:
Extend the `LeanFileProgressProcessingInfo` with an `error` field to report that the processing of a file has stopped after a (fatal) error.

One implemented use case is the processing of a file header (imports), where the server already now stops processing the rest of the file. With these changes that behavior will be reported to the editor extension, which can then show the user that the rest of the file will not be processed.

Required by https://github.com/leanprover/vscode-lean4/pull/80.

### DIFF
--- a/src/Lean/Server/Watchdog.lean
+++ b/src/Lean/Server/Watchdog.lean
@@ -590,7 +590,7 @@ def initAndRunWatchdog (args : List String) (i o e : FS.Stream) : IO Unit := do
       capabilities := mkLeanServerCapabilities
       serverInfo?  := some {
         name     := "Lean 4 Server"
-        version? := "0.1"
+        version? := "0.1.1"
       }
       : InitializeResult
     }


src/Lean/Data/Lsp/Extra.lean
```diff
@@ -37,6 +37,7 @@ instance : ToJson WaitForDiagnostics :=
 
 structure LeanFileProgressProcessingInfo where
   range : Range
+  error : Bool := false
```
gebner, line 40: Would it make sense to go for an enum here instead (which would allow us to add other states in the future)?
```lean
inductive LeanFIleProgressKind
  | processing
  | importError
  deriving FromJson, ToJson
```

Kha: Sounds good to me. A potential further state could be "not processed because there was a `#exit` above" (in grey, perhaps?).

larsk21: Good idea. The only problem I see is that we report the progress as an `Array` of `LeanFileProgressProcessingInfo`. How would we aggregate multiple progresses with different kinds? I'm currently using `any`/`some` to search for an error somewhere in the array in the linked extension PR.

gebner: > How would we aggregate multiple progresses with different kinds?

Just paint each range separately. The "represent progress by the beginning line number" design is a hack that dates back to the time where the extension mined the diagnostics for progress information.  (The server used to send diagnostics with "processing..." text at the line it was currently at.)

The old Lean 3 extension also painted each range separately and it worked just fine: https://github.com/leanprover/vscode-lean/blob/faa1d9c76406f1ed31e62ee942505328be5908c3/src/taskgutter.ts#L27-L39

larsk21: Ah, I see, I will change it to an enum then.

Vtec234:
Let's also start bumping the Lean server version in `Watchdog.lean` on protocol changes. I think a minor bump suffices, so `0.1.1`.

larsk21:
> Let's also start bumping the Lean server version in `Watchdog.lean` on protocol changes. I think a minor bump suffices, so `0.1.1`.

What's the commit convention for version bump commits? Would this be a type `chore` commit?

Kha:
> Would this be a type chore commit?

Sounds good to me

src/Lean/Data/Lsp/Extra.lean
```diff
@@ -35,8 +35,23 @@ instance : FromJson WaitForDiagnostics :=
 instance : ToJson WaitForDiagnostics :=
   ⟨fun o => mkObj []⟩
 
+inductive LeanFileProgressKind
+  | processing | fatalError
+  deriving Inhabited, BEq
```
gebner, line 40: Is there a reason you opted for 1/2 instead of "processing"/"fatalError" as the json serialization?  There shouldn't be any performance issues with strings here, we already include the much longer file name and various keys in the message.

Vtec234: Fwiw, LSP tends to encode enums as numbers so I guess it matches?

larsk21: Yeah, that was my reasoning too, I took `DiagnosticSeverity` as an example.

Kha (approved):

src/Lean/Data/Lsp/Extra.lean
gebner: Ok, this makes sense.

Kha:
Is this part ready or should we wait for the Code PR?

gebner:
> Is this part ready or should we wait for the Code PR?

I think it is ready to merge.  The feature is backwards-compatible, so it doesn't depend on changes in the editor plugins.

PR #846, opened 2021-12-03
feat: missing monad-related utils

tydeu:
Two missing monad-related utilities I need/use in Lake that seemed appropriate to PR into Lean proper.

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -1386,6 +1386,9 @@ namespace ReaderT
 section
 variable {ρ : Type u} {m : Type u → Type v} {α : Type u}
 
+instance [MonadLiftT m n] : MonadLiftT (ReaderT ρ m) (ReaderT ρ n) where
+  monadLift x := fun r => liftM (x r)
+
 instance  : MonadLift m (ReaderT ρ m) where
   monadLift x := fun _ => x
 


src/Init/Prelude.lean
```diff
@@ -1386,6 +1386,9 @@ namespace ReaderT
 section
 variable {ρ : Type u} {m : Type u → Type v} {α : Type u}
 
+instance [MonadLiftT m n] : MonadLiftT (ReaderT ρ m) (ReaderT ρ n) where
```
Kha, line 1389: Do we already have any instances of this kind? The overlap with the existing `ReaderT` instance looks potentially problematic. If we have two types that each start with `n` `ReaderT`s, we should end up with `O(n²)` TC problems. Which might not a non-issue in practice.

tydeu: @Kha I think it being a `MonadLiftT` rather than a `MonadLift` instance helps avoid many of the TC problems that would otherwise exist while still providing the desired inferences. Also, the fact that it manages to work in the Lean codebase as whole (which does a lot of monadic shenanigans) without any hiccups makes me rather confident that it will be a non-issue in practice.

leodemoura: I am also concerned about this instance. @Kha already pointed out the overlap, and I am also concerned that 
- The PR is adding the new instance to the `Prelude`. 
- We have experienced a lot of problems with `MonadLift` instances in the past. 

BTW, it is great that it works on the Lean codebase, but all TC problems there have solutions. In the past, we often experienced timeouts in problems that do not have solutions.   

It is also not clear how useful this instance is in practice. Not sure it is worth the trouble of discussing it right now. Sorry, for bringing the following point over and over again, but we need to focus on the Mathlib port and all the problems we have to fix to make it happen. We don't need this instance in the Prelude to port Mathlib.

tydeu: @leodemoura Okay, fine by me. This PR was just suggestions for utilities I was already using in Lake that seem reasonable to port to core. If the instance is undesirable feel free to reject it and I will just continue using it only in Lake.

Kha:
I cherry-picked the `liftOption` commit

PR #852, opened 2021-12-06
feat: add missing mathematical scripts

Vtec234:
Adds abbreviations for missing mathcal/fraktur characters. These happen to live in other Unicode blocks than all the other ones and might render slightly differently on some fonts, but are useful to have anyway.


gebner:
Just for reference, we've had these abbreviations for a couple of years now in the vscode extension: https://github.com/leanprover/vscode-lean/pull/74  It is good if they are at least somewhat consistent.

PR #855, opened 2021-12-08
LSP: signature help 

lovettchris:
Work in progress, I've plumbed the LSP types and so on so the message is parsed, and a fake response is returned so we see this in vscode:

image

Now I just need the real Lean server implementation to give me the real expected parameter information...

### DIFF
--- a/src/Lean/Server/README.md
+++ b/src/Lean/Server/README.md
@@ -59,6 +59,6 @@ The test `tests/lean/infoTree.lean` shows how to inspect the info tree of a comm
 ## Code style
 
 Comments should exist to denote specifics of our implementation but, for the most part, we shouldn't copy whole
-paragraphs of protocol description over from the LSP specification to avoid unnecessary duplication. Bug simple comments
+paragraphs of protocol description over from the LSP specification to avoid unnecessary duplication. But simple comments
 on LSP structures and members is encouraged so that some amount of useful tooltips can be provided while editing
 LSP server code.


src/Lean/Server/Completion.lean
```diff
@@ -274,7 +274,10 @@ private def tacticCompletion (ctx : ContextInfo) : IO (Option CompletionList) :=
       items.push { label := tk.toString, detail? := none, documentation? := none }
     return some { items := sortCompletionItems items, isIncomplete := true }
 
-partial def find? (fileMap : FileMap) (hoverPos : String.Pos) (infoTree : InfoTree) : IO (Option CompletionList) := do
+partial def find? (fileMap : FileMap) (hoverPos : String.Pos) (infoTree : InfoTree) (trigger : CompletionTriggerKind): IO (Option CompletionList) := do
+  -- TODO: When user types CTRL+SPACE in "Append.app|end" with the cursor placed in the middle of the word like
+  -- that then the trigger is CompletionTriggerKind.Invoke and in that case it is supposed to return all
+  -- CompletionItems matching to the left of the cursor, so "app" in this example.
```
Kha, line 280: We should file an issue for this

src/Lean/Server/README.md
```diff
@@ -58,6 +58,7 @@ The test `tests/lean/infoTree.lean` shows how to inspect the info tree of a comm
 
 ## Code style
 
-Comments should exist to denote specifics of our implementation but, for
-the most part, we shouldn't copy comments over from the LSP specification
-to avoid unnecessary duplication.
+Comments should exist to denote specifics of our implementation but, for the most part, we shouldn't copy whole
+paragraphs of protocol description over from the LSP specification to avoid unnecessary duplication. Bug simple comments
```
Kha, line 62:
```suggestion
paragraphs of protocol description over from the LSP specification to avoid unnecessary duplication. But simple comments
```
Sounds good to me otherwise. @Vtec234 Do you agree?

src/Lean/Server/Completion.lean
lovettchris: Looks like you did, thanks: https://github.com/leanprover/lean4/issues/857

lovettchris:
@Kha, I'd love to know what this build failure means?
https://github.com/leanprover/lean4/runs/4452793380?check_suite_focus=true

Kha:
> @Kha, I'd love to know what this build failure means? https://github.com/leanprover/lean4/runs/4452793380?check_suite_focus=true

The filename `lean-toolchain.tar.zst` is a bit suspicious, isn't it? Please remove the `lean-toolchain` file you committed.

lovettchris:
> The filename `lean-toolchain.tar.zst` is a bit suspicious, isn't it? Please remove the `lean-toolchain` file you committed.

@Kha, Ah, that was it, thanks!

Kha:
I would like to hear more about how this is supposed to work & behave in general before we get more into the implementation. For example, when do we start showing the signature? Ideally before the first argument is typed, I assume, but at that point there is no `app` node yet that we could detect. It don't see a way for the server to figure out whether an application is even accepted at a location.

lovettchris:
@Kha would you prefer to have this discussion in Zulip first to flesh this out there?  Should it be in the lean4-dev stream perhaps?

src/Lean/Server/README.md
Vtec234: I established this style way back when because even with comments describing LSP structures, I found having the LSP spec open to the side anyway necessary for figuring out what should be sent when and in what format. In general I am very supportive of commenting everything, with this being the only exception due to the preceding reason combined with the fact that, one, copying the spec descriptions over consistently is just kind of a chore and, two, risks desync when the LSP spec does change. On the other hand if you find this genuinely useful in tooltips @lovettchris , I won't object to updating the server style.

lovettchris: The problem is I get no easy intellisense and I'm worried about LSP versioning.  The comments in our code will tell us if a signature has been changed since the code was written.  I don't know how well VS code keeps documentation for past versions and would rather not depend on that.

Vtec234: I see, diffing our own comments with the spec is also useful. Sounds good!

leodemoura:
We will not pursue this PR in the near future. So, I am closing it for now. We may reconsider it in the future.

PR #856, opened 2021-12-08
Further tweak option completion

Kha:


src/Lean/Parser/Command.lean
```diff
@@ -18,6 +18,10 @@ namespace Parser
 @[builtinTermParser] def Term.quot := leading_parser "`(" >> incQuotDepth (termParser <|> many1Unbox commandParser) >> ")"
 @[builtinTermParser] def Term.precheckedQuot := leading_parser "`" >> Term.quot
 
+-- preserve dangling `.` in the syntax tree when failing
+-- NOTE: `ident >> checkNoWsBefore >> "."` only succeeds on dangling `.`, so full parser in `optional` can never succeed
+def identCompl := leading_parser ident >> optional (checkNoWsBefore >> "." >> checkNoWsBefore >> ident)
```
Kha, line 23: @leodemoura I'd love it if there was a way for this to return an unwrapped `ident` on success, but I didn't find an appropriate encoding using the current combinators

src/Lean/Server/Completion.lean
```diff
@@ -246,24 +246,23 @@ private def dotCompletion (ctx : ContextInfo) (info : TermInfo) (expectedType? :
...
-    match pref?, name with
-    | Name.anonymous, _ => true
-    | p@(Name.str pn ps _), Name.str n s _ => if pn == n then ps.isPrefixOf s else isPrefixOf p n
-    | _, _ => false
+    let partialName := stx.getSubstring? (withLeading := false) (withTrailing := false) |>.get!
+    let partialName :=
+      if !partialName.str.atEnd partialName.stopPos && partialName.str[partialName.stopPos] == '.' then
+        partialName.toString ++ "."
+      else
+        partialName.toString
```
Kha, line 254: @leodemoura @larsk21 My first solution was to preserve the trailing dot in the syntax tree, but this solution via the original input is way easier. It should also be amenable to fixing #857, though that will require additional changes in `choose` below I believe.

larsk21:
Looks good to me, the only thing that's not working is

```
set_option 
         --^ textDocument/completion
```

to see all available options, which used to work after #824.

Kha:
@larsk21 Ah, I didn't know! But makes sense that there is no test case for that :) ...

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -26,8 +26,10 @@ partial def handleCompletion (p : CompletionParams)
   let text := doc.meta.text
   let pos := text.lspPosToUtf8Pos p.position
   -- dbg_trace ">> handleCompletion invoked {pos}"
-  -- NOTE: use `>=` since the cursor can be *after* the input
-  withWaitFindSnap doc (fun s => s.endPos >= pos)
+  -- NOTE: use `+ 1` since we sometimes want to consider invalid input technically after the command,
+  -- such as a trailing dot after an option name. This shouldn't be a problem since any subsequent
+  -- command starts with a keyword that (currently?) does not participate in completion.
+  withWaitFindSnap doc (·.endPos + 1 >= pos)
```
Kha, line 32: Btw, this is the reason trailing dot completion previously worked *only* on the last command - with any other command, this line would select the following snapshot instead. On the last command, it would then go on to ignore the trailing dot during matching, which is also fixed now.

Issue #858, opened 2021-12-08
lean4 Nix flake doesn't include standard library


MichaelBurge:
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

I was following the tutorial, and the `inputs.lean.url = github:leanprover/lean4;` dependency generated by the template doesn't seem to include the Lean standard library.

When I run the "hello world" example generated by the "Quick Start" example, I get an error about print and Init being missing. This makes me think the standard library is missing. 

### Steps to Reproduce

Reproduce with Nix flakes:
```
$ nix run github:leanprover/lean4 -- <(echo)
/dev/fd/63:1:0: error: unknown package 'Init'
```

Reproduce using Nix shell
```
$ nix-shell https://github.com/leanprover/lean4/archive/master.tar.gz -A nix
$ lean <(echo)
/dev/fd/63:1:0: error: unknown package 'Init
```

Does not reproduce using the [Quickstart](https://leanprover.github.io/lean4/doc/quickstart.html) install instructions
```
$ curl https://raw.githubusercontent.com/leanprover/elan/master/elan-init.sh -sSf | sh -s -- --default-toolchain leanprover/lean4:nightly
$ source $HOME/.elan/env
$ lean <(echo)
$
```
**Expected behavior:** [What you expect to happen]
I expect Lean 4 master built using the Nix flake to execute a blank file giving no output, just like the elan installed one does.

**Actual behavior:** [What actually happens]
I get a package import error.

**Reproduces how often:** [What percentage of the time does it reproduce?]
Reproduces every single time.

### Versions
You can get this information from copy and pasting the output of `lean --version`,
please include the OS and what version of the OS you're running.

```
$ nix run github:leanprover/lean4 -- --version
Lean (version 4.0.0, Release)

$ uname -a
Linux lam 5.4.0-91-generic #102-Ubuntu SMP Fri Nov 5 16:31:28 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04 LTS
Release:        20.04
Codename:       focal

$ nix --version
nix (Nix) 2.4
```

Kha:
[...]

PR #859, opened 2021-12-09
Set defaultApp to lean-all for interactive use

MichaelBurge:
See discussion in #858 for context.


Kha:
Thanks, I didn't even think of `defaultApp`. It might be a bit confusing though that `nix shell` doesn't use it, so I think we can just change `defaultPackage` instead.

PR #861, opened 2021-12-09
fix: add binder info nodes for parameter copies in body

### LINKED ISSUE #840: Missing definition nodes of parameters in info tree
Take the following program:
```lean
set_option pp.raw true
set_option trace.Elab.info true
example (a : Nat) : a = a := Eq.refl a
```
In the output, we have definition (`isBinder := true`) and usage nodes for occurrences of `a` in the header, but the `a` in the body uses a different FVarId `_uniq.8` with no corresponding definition node. This is an issue for e.g. the find-all-references request (/cc @Garmelon).
```
[Elab.info] command @ ⟨3, 0⟩-⟨3, 38⟩ @ Lean.Elab.Command.elabDeclaration
  Nat : Type @ ⟨3, 13⟩-⟨3, 16⟩ @ Lean.Elab.Term.elabIdent
    [.] `Nat : some Sort.{?_uniq.1} @ ⟨3, 13⟩-⟨3, 16⟩
    Nat : Type @ ⟨3, 13⟩-⟨3, 16⟩
  _uniq.2 : Nat @ ⟨3, 9⟩-⟨3, 10⟩
  Eq.{1} Nat _uniq.2 _uniq.2 : Prop @ ⟨3, 20⟩-⟨3, 25⟩ @ _aux_Init_Notation___macroRules_«term_=_»_2
    Macro expansion
    («term_=_» `a "=" `a)
    ===>
    (Term.binrel "binrel%" `Eq._@._hyg.7 `a `a)
      Eq.{1} Nat _uniq.2 _uniq.2 : Prop @ ⟨3, 20⟩†-⟨3, 25⟩ @ Lean.Elab.Term.elabBinRel
        _uniq.2 : Nat @ ⟨3, 20⟩-⟨3, 21⟩ @ Lean.Elab.Term.elabIdent
          _uniq.2 : Nat @ ⟨3, 20⟩-⟨3, 21⟩
        _uniq.2 : Nat @ ⟨3, 24⟩-⟨3, 25⟩ @ Lean.Elab.Term.elabIdent
          _uniq.2 : Nat @ ⟨3, 24⟩-⟨3, 25⟩
  Eq.refl.{1} Nat _uniq.8 : Eq.{1} Nat _uniq.8 _uniq.8 @ ⟨3, 29⟩-⟨3, 38⟩ @ Lean.Elab.Term.elabApp
    [.] `Eq.refl : some Eq.{1} Nat _uniq.8 _uniq.8 @ ⟨3, 29⟩-⟨3, 36⟩
    Eq.refl.{1} : forall {α : Type} (a : α), Eq.{1} α a a @ ⟨3, 29⟩-⟨3, 36⟩
    _uniq.8 : Nat @ ⟨3, 37⟩-⟨3, 38⟩ @ Lean.Elab.Term.elabIdent
      _uniq.8 : Nat @ ⟨3, 37⟩-⟨3, 38⟩
  _example : forall (a : Nat), Eq.{1} Nat a a @ ⟨3, 0⟩-⟨3, 7⟩
```
I see two solutions here:
* We manually add definition nodes for the new FVarIds when re-adding the parameters to the local context before elaborating the body. This will require some more work in the server to identify FVarIds that are "morally" the same, but it should be doable.
* We save and restore the original FVarIds when re-adding the parameters. I'm not completely sure it is safe to reuse out-of-scope FVarIds though.

@leodemoura What do you think?

Kha:
Fixes #840

### DIFF
--- a/src/Lean/Elab/LetRec.lean
+++ b/src/Lean/Elab/LetRec.lean
@@ -17,7 +17,7 @@ structure LetRecDeclView where
   attrs         : Array Attribute
   shortDeclName : Name
   declName      : Name
-  numParams     : Nat
+  binderIds     : Array Syntax
   type          : Expr
   mvar          : Expr -- auxiliary metavariable used to lift the 'let rec'
   valStx        : Syntax
@@ -46,11 +46,12 @@ private def mkLetRecDeclView (letRec : Syntax) : TermElabM LetRecView := do
       addAuxDeclarationRanges declName decl declId
       let binders := decl[1].getArgs
       let typeStx := expandOptType declId decl[2]
-      let (type, numParams) ← elabBinders binders fun xs => do
+      let (type, binderIds) ← elabBindersEx binders fun xs => do
           let type ← elabType typeStx
           registerCustomErrorIfMVar type typeStx "failed to infer 'let rec' declaration type"
+          let (binderIds, xs) := xs.unzip
           let type ← mkForallFVars xs type
-          pure (type, xs.size)
+          pure (type, binderIds)
       let mvar ← mkFreshExprMVar type MetavarKind.syntheticOpaque
       let valStx ←
         if decl.isOfKind `Lean.Parser.Term.letIdDecl then
@@ -62,7 +63,7 @@ private def mkLetRecDeclView (letRec : Syntax) : TermElabM LetRecView := do
         attrs         := attrs,
         shortDeclName := shortDeclName,
         declName      := declName,
-        numParams     := numParams,
+        binderIds     := binderIds,
         type          := type,
         mvar          := mvar,
         valStx        := valStx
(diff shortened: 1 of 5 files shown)


src/Lean/Elab/LetRec.lean
```diff
@@ -46,11 +46,12 @@ private def mkLetRecDeclView (letRec : Syntax) : TermElabM LetRecView := do
       addAuxDeclarationRanges declName decl declId
       let binders := decl[1].getArgs
       let typeStx := expandOptType declId decl[2]
-      let (type, numParams) ← elabBinders binders fun xs => do
+      let (type, binderIds) ← elabBindersEx binders fun xs => do
           let type ← elabType typeStx
           registerCustomErrorIfMVar type typeStx "failed to infer 'let rec' declaration type"
+          let (binderIds, xs) := xs.unzip
```
Kha, line 52: Even though we use `unzip` in both callers of `elabBindersEx`, I still prefer an array of pairs over two separate arrays in the signature for documentation purposes, and assume that any potential overhead from this is not a concern here.

leodemoura: An array of pairs is fine. I agree that the overhead is not a concern here.

PR #862, opened 2021-12-09
feat: `OfNat` instance for `Option`

leodemoura:
We often get errors such as
```
error: failed to synthesize instance
  OfNat (Option Nat) 0
```
They happen, for example, when we have a function `f (x : Nat) (val? : Option Nat := none)`, and we write `f a 0`. I have also seen a few users confused by this behavior because `f a a` works. That is, the coercion from `Nat` to `Option Nat` is used in this case, but not used when we have the polymorphic `0`. This issue is another instance of the more general issue "coercion and polymorphism produce unexpected results." That being said, we can avoid this problem by adding the instance
```lean
instance [o : OfNat α n] : OfNat (Option α) n where
  ofNat := some o.ofNat
```

One potential problem is that the `some` is now "hidden" inside the OfNat instance.
Example:
```lean
def f (x : Nat) (y? : Option Nat := none) := ...

variable (a : Nat)

#check f 1 Nat.zero
-- f 1 (some Nat.zero)

#check f 1 a
-- f 1 (some a)

#check f 1 0
-- f 1 0
```

### DIFF
--- a/src/Init/Data/Option/Basic.lean
+++ b/src/Init/Data/Option/Basic.lean
@@ -80,3 +80,6 @@ deriving instance BEq for Option
 
 instance [LT α] : LT (Option α) where
   lt := Option.lt (· < ·)
+
+instance [o : OfNat α n] : OfNat (Option α) n where
+  ofNat := some o.ofNat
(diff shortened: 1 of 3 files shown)


leodemoura:
We are discussing it at Zulip.

PR #865, opened 2021-12-10
Emacs mode lake

lecopivo:
A simple command for running `lake build` in lean4-mode

### DIFF
--- a/lean4-mode/lean4-lake.el
+++ b/lean4-mode/lean4-lake.el
@@ -16,7 +16,7 @@
   (interactive)
   (let* ((default-directory (file-name-as-directory (lean4-lake-find-dir-safe))))
     (with-existing-directory
-      (compile "lake build"))))
+      (compile (concat (lean4-get-executable "lake") " build")))))
 
 (defun lean4-lake-lib-paths ()
   "Returns list of paths to package's oleans and its dependencies."
--- a/lean4-mode/README.md
+++ b/lean4-mode/README.md
@@ -64,14 +64,15 @@ Set these with e.g. `M-x customize-variable`.
 Key Bindings and Commands
 =========================
 
-| Key                | Function                                                                        |
-|--------------------|---------------------------------------------------------------------------------|
-| <kbd>C-c C-k</kbd> | show the keystroke needed to input the symbol under the cursor                  |
-| <kbd>C-c C-d</kbd> | recompile & reload imports (`lean4-refresh-file-dependencies`)                  |
-| <kbd>C-c C-x</kbd> | execute Lean in stand-alone mode (`lean4-std-exe`)                              |
-| <kbd>C-c C-i</kbd> | toggle info view showing goals and errors at point (`lean4-toggle-info-buffer`) |
-| <kbd>C-c ! n</kbd> | flycheck: go to next error                                                      |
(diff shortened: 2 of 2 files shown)


Kha:
Thanks. Could you add the keyboard shortcut to `lean4-mode/README.md`?

lean4-mode/lean4-lake.el
```diff
@@ -0,0 +1,21 @@
...
+(defun lean4-lake-find-dir-safe ()
+  (or (lean4-lake-find-dir)
+      (error (format "cannot find lakefile.lean for %s" (buffer-file-name)))))
+
+(defun lean4-lake-build ()
+  "Call lake build"
+  (interactive)
+  (let* ((default-directory (file-name-as-directory (lean4-lake-find-dir-safe))))
+    (with-existing-directory
+      (compile "lake build"))))
```
Kha, line 19: You should probably use `(lean4-get-executable "lake")` here

lecopivo:
Sure, I fix that. I will also add ability to run `lake run <script>`. Would something else make sense?

tydeu:
@lecopivo

> I will also add ability to run `lake run <script>`.

Fyi, I plan to change this to `lake script run <script>` soon, so maybe hold off on that for just a little while?

lecopivo:
Sure, I can wait. 

Also, is there a planned support for unit testing in lake? Or is that ment to be done through scripts?

tydeu:
@lecopivo 

> Also, is there a planned support for unit testing in lake? Or is that ment to be done through scripts?

Yeah, testing will probably be done through a combination of a separate testing library, a test runner executable, and a script (like how [mocha](https://mochajs.org/) is generally used in JavaScript).

Kha:
You should use `lake env` (if there is a lakefile) for running `lean` with `LEAN_PATH` set up correctly.

lecopivo:
> You should use `lake env` (if there is a lakefile) for running `lean` with `LEAN_PATH` set up correctly.

Ohh I didn't about this, well at least I got some practice with elisp after a long time.

lecopivo:
Concerning keybindings, I'm not the biggest fan of the `C-c C-p C-l`. I picked it to be consistent with keybindings for leanpkg. What about having `C-c C-x` to be `lean4-std-exe` and `C-c C-l` be `lean4-lake-build`. Alternatively, we can use `C-c C-b` for `lean4-lake-build` as `lean4-message-boxes-toggle` does no longer exist.

Kha:
Either would be fine with me.

lecopivo:
I will go for `C-c C-l` for `lean4-lake-build`. I will add an option to rebuild lake project before running `lean4-std-exe`.

Kha:
I don't have a strong opinion on any of these since I never use them :) . But rebuilding might be problematic if some downstream module fails to build. Ideally we would only rebuild the imports like the server does via `lake print-paths`, but that's a bit more complicated.

Kha (changes requested):
> You should use `lake env` (if there is a lakefile) for running `lean` with `LEAN_PATH` set up correctly.

This is still missing I think

lecopivo:
Yes it is still missing. I will try to fix that and change the keybindings sometime this week. In the current form it solves my problem :) so I had priorities elsewhere.

PR #867, opened 2021-12-11
feat: allow opt-out of grouping in formatter

gebner:
Currently, the pretty-printer formats tactics proofs like this:
```lean
theorem foo : True → True :=
  by
    intro
    trivial
```

The main goal of this PR is to format theorems as follows (which is important for mathport because we use the formatter to generate the Lean 4 source code):
```lean
theorem foo : True → True := by
  intro
  trivial
```

The reason why the by is indented at the moment is that the formatter for category parsers automatically inserts fill+nest nodes around the produced format object (confusingly, this is called a group in the formatter).

We add a new no-op parser combinator called `ppAllowUngrouped` that instructs the category pretty-printer to omit this fill+nest operation, and add this combinator to by/do/fun.  This causes theorems with tactics proofs to be printed in the desired way
```lean
-- Omit group (i.e., fill+nest) in the enclosing category parser:
syntax ppAllowUngrouped "do " tacticSeq : term
```

There are also two new pretty-printing combinators which can do different things depending on whether the following parser was ungrouped or not.  `ppDedentIfGrouped` undos the nest step (but keeps the fill) if the parser is grouped (used in parentheses), and `ppHardLineUnlessUngrouped` prints a hard newline if the parser is grouped, and a soft newline otherwise (used in declarations after `:=`).

I have prepared a branch where all Lean source files have been reformatted with these changes: https://github.com/leanprover/lean4/compare/master...gebner:formatted?diff=split  There is of course still lots of low-hanging fruit left (and some more serious bugs).

There are several changes in this PR, I can split it up if desired:
 1. I have added a small script that reformats lean files ("feat: add script to reformat lean files").  In contrast to the `Reformat.lean` file in the tests directory, this also works if the lean file defines notation and hence requires elaboration.
 2. There was a bug in the `pushToken` function in the formatter which printed extra spaces, for example two spaces after `have` and very often trailing spaces ("fix: suppress extra spaces in formatter").
 3. The main change adding `ppAllowUngrouped` ("feat: allow opt-out of grouping in formatter")
 4. The rest is minor changes (such as adding spaces to notations, etc.), which was helpful to clean up the reformatted output and make it easier to see if the ungrouping change had unintended side effects.

### DIFF
--- a/tests/lean/ppite.lean.expected.out
+++ b/tests/lean/ppite.lean.expected.out
@@ -3,10 +3,12 @@ fun xs =>
   List.forM xs fun x =>
     if (x == 0) = true then do
       IO.println "foo"
-      IO.println "zero" else
+      IO.println "zero"
+    else
       if (x % 2 == 0) = true then do
         IO.println x
-        IO.println "even" else do
+        IO.println "even"
+      else do
         IO.println x
         IO.println "odd"
 if true = true then 1 else 0 : Nat
--- a/src/Init/Notation.lean
+++ b/src/Init/Notation.lean
@@ -137,12 +137,16 @@ infixr:100 " <$> " => Functor.map
 macro_rules | `($x <|> $y) => `(binop_lazy% HOrElse.hOrElse $x $y)
 macro_rules | `($x >> $y)  => `(binop_lazy% HAndThen.hAndThen $x $y)
 
-syntax (name := termDepIfThenElse) ppGroup(ppDedent("if " ident " : " term " then" ppSpace term ppDedent(ppSpace "else") ppSpace term)) : term
+syntax (name := termDepIfThenElse)
+  ppRealGroup(ppRealFill(ppIndent("if " ident " : " term " then") ppSpace term)
+    ppDedent(ppSpace) ppRealFill("else " term)) : term
 
 macro_rules
   | `(if $h:ident : $c then $t:term else $e:term) => `(let_mvar% ?m := $c; wait_if_type_mvar% ?m; dite ?m (fun $h:ident => $t) (fun $h:ident => $e))
 
-syntax (name := termIfThenElse) ppGroup(ppDedent("if " term " then" ppSpace term ppDedent(ppSpace "else") ppSpace term)) : term
+syntax (name := termIfThenElse)
+  ppRealGroup(ppRealFill(ppIndent("if " term " then") ppSpace term)
+    ppDedent(ppSpace) ppRealFill("else " term)) : term
 
 macro_rules
   | `(if $c then $t:term else $e:term) => `(let_mvar% ?m := $c; wait_if_type_mvar% ?m; ite ?m $t $e)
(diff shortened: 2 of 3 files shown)


tests/lean/Reformat.lean.expected.out
```diff
@@ -32,27 +32,25 @@ def idDelta {α : Sort u} (a : α) : α :=
 /- `idRhs` is an auxiliary declaration used to implement "smart unfolding". It is used as a marker. -/
 @[macroInline, reducible]
 def idRhs (α : Sort u) (a : α) : α :=
-  a 
+  a
 
-abbrev Function.comp {α : Sort u} {β : Sort v} {δ : Sort w} (f : β → δ) (g : α → β) : α → δ :=
-  fun x => f (g x)
+abbrev Function.comp {α : Sort u} {β : Sort v} {δ : Sort w} (f : β → δ) (g : α → β) : α → δ := fun x => f (g x)
```
Kha, line 37: Ah, that `fun` but not other terms can now be used in a single line is an interesting consequence... I think we should really introduce `ppLineUnlessUngrouped` and use it instead of `ppHardline...` but that doesn't have to be in this PR

tests/lean/PPRoundtrip.lean.expected.out
```diff
@@ -55,10 +55,10 @@ decide (1 < 2) ||
...
 typeAs Nat
   (let x := 1;
-  do 
-    discard (pure 2)
-    let y : Nat := 3
-    pure (x + y))
+  do
+  discard (pure 2)
+  let y : Nat := 3
+  pure (x + y))
```
Kha, line 61: Should we care about this one? Arguably the indent of `do` was wrong to begin with.

tests/lean/ppite.lean.expected.out
```diff
@@ -1,19 +1,13 @@
...
-            IO.println "odd"
+  List.forM xs fun x =>
+    if (x == 0) = true then do
+      IO.println "foo"
+      IO.println "zero" else
+      if (x % 2 == 0) = true then do
+        IO.println x
+        IO.println "even" else do
+        IO.println x
+        IO.println "odd"
```
Kha, line 11: Awesome!

Kha: ...oh, except for the `else`s? :)

gebner: Okay, I've pushed a changed that puts newlines after then and before/after else (by exposing Format.group as a pretty-printing combinator, which requires a stage0 update).

There's still the structure instance issue, where the formatter produces `{ a := 42 b:= 1 }`.  I've tried replacing `optional ", "` by `", " <|> (ppLine >> group skip)`.  This produces parseable syntax, but formats every structure instance in multi-line format...  Ideally we'd define structure instances in terms of `sepBy ", " structInstField (", " <|> (checkColGe >> ppLine))` (so that the final `ppLine` is omitted), but that requires big changes in the structure instance elaborator (with the accompanying bootstrapping problems).

PR #868, opened 2021-12-12
feat: use `BaseIO` more in `Init.System.IO`

tydeu:
Promote a number of error-less `IO` actions in `Init.Systen.IO` to use `BaseIO`.

### DIFF
--- a/tests/lean/doErrorMsg.lean.expected.out
+++ b/tests/lean/doErrorMsg.lean.expected.out
@@ -1,7 +1,7 @@
 doErrorMsg.lean:3:2-3:13: error: type mismatch
   IO.getStdin
 has type
-  IO IO.FS.Stream : Type
+  BaseIO IO.FS.Stream : Type
 but is expected to have type
   IO PUnit : Type
 doErrorMsg.lean:15:19-15:21: error: type mismatch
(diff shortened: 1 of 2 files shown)


src/Init/System/IO.lean
```diff
@@ -506,7 +507,7 @@ structure Child (cfg : StdioConfig) where
 
 @[extern "lean_io_process_spawn"] constant spawn (args : SpawnArgs) : IO (Child args.toStdioConfig)
 
-@[extern "lean_io_process_child_wait"] constant Child.wait {cfg : @& StdioConfig} : @& Child cfg → IO UInt32
+@[extern "lean_io_process_child_wait"] constant Child.wait {cfg : @& StdioConfig} : @& Child cfg → BaseIO UInt32
```
gebner, line 510: `waitpid` can return an error though?  (We also don't seem ensure that `Child.wait` is called exactly once.)

tydeu: @gebner Lean folds the error from `waitpid` into the exit code. As such, `Child.wait` cannot error.

Your comment also demonstrates one of the reasons why using `BaseIO` is so important. Without it, the user has to go look at the internal implementation to figure out whether something can actually error (as whether it can is not always intuitive).

Kha: Gabriel is right, we should not ignore the return value of `waitpid`. Waiting twice should fail.

tydeu: @Kha do you want push a change to update the implementation of `lean_io_process_child_wiat` and then I can update this PR accordingly?

Kha: done

tydeu:
This PR should be done now. What do you think @Kha?

PR #870, opened 2021-12-13
perf: server: do not rebuild full file without incremental sync

gebner:
Due to some bugs in the neovim LSP implementation, we want to disable incremental sync in the neovim plugin for now.  However it seems that without incremental sync, the Lean server reprocesses the whole file on every change.  This is because the `foldDocumentChanges` function claims that a fullChange always changes the first character in a file.

This PR computes the first position where the old and the new file differ instead.

From a performance point of view, I think we could just do the diff always.  Then we would also skip undo/redo pairs.  (We're already copying the whole source string for every single change..)  The `firstDiffPos` function can also be optimized further by doing byte comparisons instead of decoding UTF-8 (not that it is a bottleneck).

### DIFF
--- a/src/Lean/Server/Utils.lean
+++ b/src/Lean/Server/Utils.lean
@@ -105,23 +105,16 @@ def toFileUri (fname : System.FilePath) : Lsp.DocumentUri :=
 
 open Lsp
 
-/-- Returns the document contents with all changes applied, together with the position of the change
-which lands earliest in the file. Panics if there are no changes. -/
-def foldDocumentChanges (changes : @& Array Lsp.TextDocumentContentChangeEvent) (oldText : FileMap)
-  : FileMap × String.Pos :=
-  if changes.isEmpty then panic! "Lean.Server.foldDocumentChanges: empty change array" else
-  let accumulateChanges : FileMap × String.Pos → TextDocumentContentChangeEvent → FileMap × String.Pos :=
-    fun ⟨newDocText, minStartOff⟩ change =>
-      match change with
-      | TextDocumentContentChangeEvent.rangeChange (range : Range) (newText : String) =>
-        let startOff    := oldText.lspPosToUtf8Pos range.start
-        let newDocText  := replaceLspRange newDocText range newText
-        let minStartOff := minStartOff.min startOff
-        ⟨newDocText, minStartOff⟩
-      | TextDocumentContentChangeEvent.fullChange (newText : String) =>
-        ⟨newText.toFileMap, 0⟩
-  -- NOTE: We assume Lean files are below 16 EiB.
-  changes.foldl accumulateChanges (oldText, 0xffffffff)
+/-- Returns the document contents with the change applied. -/
+def applyDocumentChange (oldText : FileMap) : (change : Lsp.TextDocumentContentChangeEvent) → FileMap
+  | TextDocumentContentChangeEvent.rangeChange (range : Range) (newText : String) =>
+    replaceLspRange oldText range newText
+  | TextDocumentContentChangeEvent.fullChange (newText : String) =>
+    newText.toFileMap
+
+/-- Returns the document contents with all changes applied. -/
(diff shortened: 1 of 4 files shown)


src/Lean/Server/Utils.lean
```diff
@@ -105,6 +105,13 @@ def toFileUri (fname : System.FilePath) : Lsp.DocumentUri :=
 
 open Lsp
 
+partial def firstDiffPos (a b : String) : String.Pos :=
```
Vtec234, line 108: `a b` -> `old new` ?

Vtec234: It would also be good to note that this returns an offset into the UTF-8 encoding, e.g. by renaming to `firstDiffUtf8Pos`.

src/Lean/Server/Utils.lean
```diff
@@ -119,7 +126,8 @@ def foldDocumentChanges (changes : @& Array Lsp.TextDocumentContentChangeEvent)
         let minStartOff := minStartOff.min startOff
         ⟨newDocText, minStartOff⟩
       | TextDocumentContentChangeEvent.fullChange (newText : String) =>
-        ⟨newText.toFileMap, 0⟩
+        let newFileMap := newText.toFileMap
+        ⟨newFileMap, firstDiffPos newDocText.source newFileMap.source⟩
```
Vtec234, line 130: I was briefly confused on reading this code again, maybe `newDocText` wasn't such a good name. Perhaps `accDocText` for "accumulating" is better?

Vtec234 (approved):

Vtec234:
> From a performance point of view, I think we could just do the diff always. Then we would also skip undo/redo pairs.

I don't understand this.

> We're already copying the whole source string for every single change..

This may be true, although I would hope that since the accumulator is used linearly, FBIP could kick in.

src/Lean/Server/Utils.lean
gebner: Okay, I've renamed the function and added a docstring.  I thought tha `String.Pos` was always supposed to be UTF-8 though.

> a b -> old new ?

The function is completely symmetric in its arguments, so I'd rather not attach different meaning to them.

src/Lean/Server/Utils.lean
gebner: Good point.  I've refactored it and extracted the change application logic into a separate function, which avoids the variable naming issue.

There was already an issue due to the confusing variable names: `startOff` was computed by resolving the range against the original `oldText`, and not the current version.  (This wouldn't have affected the end result though.)

gebner:
> > From a performance point of view, I think we could just do the diff always. Then we would also skip undo/redo pairs.
> 
> I don't understand this.

If you type `e` followed by a backspace, this will generate two changes.  I haven't checked what vscode does internally, but neovim batches these changes and so might send a single didChange notification where first `e` is inserted and then deleted again.  Currently, the Lean LSP server would start reprocessing from the position where you inserted `e` even though the document did not change.

> > We're already copying the whole source string for every single change..
> 
> This may be true, although I would hope that since the accumulator is used linearly, FBIP could kick in.

FBIP isn't magic.  This is what the code looks like:
```lean
  let pre := text.source.extract 0 start
  let post := text.source.extract «end» text.source.bsize
  (pre ++ newText ++ post).toFileMap
```
`String.extract` copies the substring here.  It would be possible to write a `String.replace` function which does not allocate a new string if the argument is unshared, but it would still need to move the rest of the string that comes after the insertion (and potentially reallocate if the capacity is not enough).  All of that cost is of course dwarfed by the `toFileMap` function.

Kha:
> From a performance point of view, I think we could just do the diff always. Then we would also skip undo/redo pairs.

Sounds good to me

gebner:
> Sounds good to me

Great!  This simplifies the implementation a bit, because we no longer need to pass around the change position.

I've also moved the diff function to `Init.Data.String.Basic` as `String.firstDiffPos` (without the Utf8, since everything is UTF-8 there).

Vtec234:
Thanks! Lake test failure looks unrelated.

PR #876, opened 2021-12-14
chore: update Lake

tydeu:
Attempting to fix the broken `ffi-dep` test by reorganizing the example. Let's see if it works. 🙏


tydeu:
@leodemoura / @Kha this does appear to have fixed the `ffi-dep` failure -- its just `print-paths` memory leak that remains.

PR #878, opened 2021-12-15
feat: support syntax abbreviations in dynamic quotations

gebner:
It is constantly annoying me that I can't use syntax abbreviations in dynamic quotations:
```lean
syntax foo := "a" <|> "b"
-- ...

#check `(foo| a)
#check (· matches `(foo| a))
```

The reason this doesn't work at the moment is because the parserOfStack parser doesn't support ParserDescrs.  This PR replaces the evalParserConst function with the strictly more general mkParserOfConstant function from the Lean.Parser.Extension module (which can interpret parser descriptions).

I had to move some functions (including parserOfStack) to the Lean.Parser.Extension module and reorder some stuff there to satisfy the dependencies.


Kha:
Nice!!

PR #882, opened 2021-12-16
fix: use redirected stderr for `timeit` & `allocprof`

Kha:


gebner:
This is wonderful!  Is there any chance we can make `tout()` use `eprint` as well (so that the compiler output is redirected)?

Kha:
> Is there any chance we can make `tout()` use `eprint` as well (so that the compiler output is redirected)?

I'm sure it should be possible to implement a `std::streambuf` that is backed by `io_eprint` but wow, I'm staring at this API and it's dealing psychic damage to me.

gebner:
Ok, I'll give it a shot.

PR #886, opened 2021-12-17
chore: update Lake

tydeu:
A fix for a problem with mathport builds that I overlooked yesterday.


tydeu:
cc @leodemoura / @Kha

PR #890, opened 2021-12-17
fix: leanc: do not change sysroot

### LINKED ISSUE #828: Ensure LLVM bundling works for basic use cases
See discussion of issues at https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/C.20FFI.20usage.
- [x] https://github.com/leanprover/lake/issues/35
- [ ] https://github.com/leanprover/lake/issues/36

Kha:
Apparently clang is smart enough to look up runtime files missing from the sysroot in the installation directory automatically.

Resolves #828

/cc @tydeu

### DIFF
--- a/script/prepare-llvm-linux.sh
+++ b/script/prepare-llvm-linux.sh
@@ -36,8 +36,9 @@ OPTIONS=()
 echo -n " -DLEAN_STANDALONE=ON"
 echo -n " -DCMAKE_C_COMPILER=$PWD/stage1/bin/clang -DCMAKE_CXX_COMPILER=$PWD/llvm/bin/clang++ -DLEAN_CXX_STDLIB='-Wl,-Bstatic -lc++ -lc++abi -Wl,-Bdynamic'"
 # allow C++ code to include /usr since it needs quite a few more headers
+# set sysroot for both C++ and C to make sure we have all necessary runtime files, but don't embed in `leanc` so users can still link to system libs
 echo -n " -DLEAN_EXTRA_CXX_FLAGS='--sysroot $PWD/llvm -I/usr/include -I/usr/include/x86_64-linux-gnu'"
-echo -n " -DLEANC_INTERNAL_FLAGS='--sysroot ROOT -I ROOT/include/clang' -DLEANC_CC=ROOT/bin/clang"
+echo -n " -DLEANC_INTERNAL_FLAGS='-I ROOT/include/clang' -DLEANC_OPTS='--sysroot $PWD/stage1' -DLEANC_CC=ROOT/bin/clang"
 echo -n " -DLEANC_INTERNAL_LINKER_FLAGS='-L ROOT/lib -L ROOT/lib/glibc ROOT/lib/glibc/libc_nonshared.a -Wl,--as-needed -static-libgcc -Wl,-Bstatic -lgmp -lunwind -Wl,-Bdynamic -fuse-ld=lld'"
 # do not set `LEAN_CC` for tests
 echo -n " -DLEAN_TEST_VARS=''"
--- a/script/prepare-llvm-macos.sh
+++ b/script/prepare-llvm-macos.sh
@@ -35,8 +35,9 @@ echo -n " -DLEAN_STANDALONE=ON"
 # and the custom clang++ outputs a myriad of warnings when consuming the SDK
 echo -n " -DCMAKE_C_COMPILER=$PWD/stage1/bin/clang"
 echo -n " -DGMP_LIBRARIES=lib/libgmp.a -DGMP_INCLUDE_DIR=/usr/local/opt/gmp/include"
+# set sysroot for C to make sure we have all necessary runtime files, but don't embed in `leanc` so users can still link to system libs
 # need no-macro-redefined for weird clang stdint.h
-echo -n " -DLEANC_INTERNAL_FLAGS='--sysroot ROOT -I ROOT/include/clang -Wno-macro-redefined' -DLEANC_CC=ROOT/bin/clang"
+echo -n " -DLEANC_INTERNAL_FLAGS='-I ROOT/include/clang -Wno-macro-redefined' -DLEANC_OPTS='--sysroot $PWD/stage1' -DLEANC_CC=ROOT/bin/clang"
 echo -n " -DLEANC_INTERNAL_LINKER_FLAGS='-L ROOT/lib -L ROOT/lib/libc -fuse-ld=lld'"
 # do not set `LEAN_CC` for tests
 echo -n " -DLEAN_TEST_VARS=''"
--- a/script/prepare-llvm-mingw.sh
+++ b/script/prepare-llvm-mingw.sh
@@ -36,8 +36,10 @@ echo -n " -DLEAN_STANDALONE=ON"
 echo -n " -DCMAKE_C_COMPILER=$PWD/stage1/bin/clang.exe -DCMAKE_CXX_COMPILER=$PWD/llvm/bin/clang++.exe -DLEAN_CXX_STDLIB='-lc++ -lc++abi'"
 echo -n " -DSTAGE0_CMAKE_C_COMPILER=clang -DSTAGE0_CMAKE_CXX_COMPILER=clang++"
 # allow C++ code to include /usr since it needs quite a few more headers
+# allow C++ code to include /usr since it needs quite a few more headers
+# set sysroot for both C++ and C to make sure we have all necessary runtime files, but don't embed in `leanc` so users can still link to system libs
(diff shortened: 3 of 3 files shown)


Kha:
Test-driving on Ubuntu:
```sh
$ ./lean-4.0.0-linux/bin/leanc main.o -lssl -Wl,--verbose
ld.lld: /usr/lib/x86_64-linux-gnu/crt1.o
ld.lld: /usr/lib/x86_64-linux-gnu/crti.o
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/clang/13.0.0/lib/linux/clang_rt.crtbegin-x86_64.o
ld.lld: main.o
ld.lld: /usr/lib/x86_64-linux-gnu/libssl.so
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/glibc/libc_nonshared.a
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/libgmp.a
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/libunwind.a
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/lean/libleancpp.a
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/lean/libLean.a
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/lean/libInit.a
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/lean/libStd.a
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/lean/libleanrt.a
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/libc++.a
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/libc++abi.a
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/glibc/libm.so
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/glibc/libdl.so
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/libgmp.a
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/clang/13.0.0/lib/linux/libclang_rt.builtins-x86_64.a
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/glibc/libpthread.so
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/glibc/libc.so
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/clang/13.0.0/lib/linux/libclang_rt.builtins-x86_64.a
ld.lld: /tmp/fo/lean-4.0.0-linux/lib/clang/13.0.0/lib/linux/clang_rt.crtend-x86_64.o
ld.lld: /usr/lib/x86_64-linux-gnu/crtn.o
```
As we see, it does take the `crt` files from the sysroot (which should be fine, *those* at least should be universally compatible), but the clangrt from the bundle (since this PC does not have clang 13.0.0 installed).

gebner:
With this change `#include <stdio.h>` will silently compile with leanc under ubuntu, but fail under nixos, right?

tydeu:
@gebner yeah, but that is kind of the point though -- to allow packages to use easily system libraries (if they exist).

gebner:
> yeah, but that is kind of the point though -- to allow packages to use easily system libraries (if they exist).

The important part is the second half of my comment: *"but fail under nixos"*.  Allowing `leanc` to find libraries and headers in `/usr` encourages nonportable build scripts.  In other words, people will call `leanc` in their lakefiles when they should call `cc` instead.

tydeu:
@gebner That is what I was referring too -- the point is support nonportable build scripts. Depending on system libraries / tools is inherently non-portable. For instance, `cc` does not exist, by default, on native Windows (which Lean appears to be hoping to eventually support).

gebner:
> For instance, cc does not exist, by default, on native Windows or in MSYS2 with build essentials

But you *can* install it.  On the other hand there is afaict no way to make the bundled C compiler work with system libraries on nixos *at all*.  Using `leanc` instead of `cc` (as this PR encourages) hence just shifts around the breakage, from an easily fixable one on windows to an unfixable one on nixos (as far as the user is concerned).

In early versions of the bundled compiler, you could override the bundled clang by setting an environment variable (`LEAN_CC` iirc).  I am open to revisit that approach and set LEAN_CC in elan on nixos, but Sebastian was not excited about it.

tydeu:
@gebner Note that you can still set `LEAN_CC` to the change the compiler.  

To your other points: as far as I know, there is no way to install a `cc` at all on just plain Windows either (i.e., without the hackiy compatibility layer that is MSYS2) . Also, would not a build script for NixOS be using Nix and thus not be the target of this change?

Kha:
@gebner Actually, I think we should override `LEAN_CC` anyway in Nix elan to make the `NIX_LDFLAGS` magic work.

Kha:
> With this change `#include <stdio.h>` will silently compile with leanc under ubuntu, but fail under nixos, right?

We should still discourage it (as Lake e.g. does by using `c++` for `oFileTarget` by default), but this, as well as the issue of finding libraries on NixOS that we would have to solve anyway, should now be covered by https://github.com/NixOS/nixpkgs/pull/151328. If that looks good, I'd like to merge this PR soon.

gebner:
> Note that you can still set LEAN_CC to the change the compiler.

It's unfortunately not that easy, look at Sebastian's nixpkgs PR.  You also need to pass `-L .elan/toolchains/leanprover--lean4-.../lib/` to the linker, otherwise it won't find libc++, etc.  And there is no environment variable for linker arguments, so you need to wrap leanc.

> Also, would not a build script for NixOS be using Nix and thus not be the target of this change?

No, all the same reasons why you want to use elan still apply to nixos as well (you get binaries for the right version and you don't need to build them locally).

If you're thinking of the nix build scripts in this repository, that's something completely orthogonal.  It's an entirely different and incompatible build system, it doesn't help you if a project only has a lakefile.  Or if you want to debug a lakefile.

Personally I feel that the nix build system is also less attractive on nixos because it shares the system nix store.  If I want to remove old lean versions, then garbage collection will also clean up system packages.  If I want to add a binary cache, then I have to trust it completely (because it might also be used for system packages).  Particularly the last point makes me very reluctant to invest time in it.

gebner:
As expected, this change broke old nixos elan versions, but in a very unexpected way.  Apparently some headers are now included twice??

Nothing to do here except to upgrade your elan if you're on nixos and seeing the following error:
```
error: In file included from ./lean_packages/mathlib/./build/ir/Mathlib/Tactic/Lint.c:4:
In file included from /home/gebner/.elan/toolchains/leanprover--lean4---nightly-2021-12-22/include/lean/lean.h:10:
/home/gebner/.elan/toolchains/leanprover--lean4---nightly-2021-12-22/include/clang/stdint.h:175:10: warning: '__int_least32_t' macro redefined [-Wmacro-redefined]
# define __int_least32_t int32_t
         ^
/home/gebner/.elan/toolchains/leanprover--lean4---nightly-2021-12-22/include/clang/stdint.h:101:10: note: previous definition is here
# define __int_least32_t int64_t
         ^
/home/gebner/.elan/toolchains/leanprover--lean4---nightly-2021-12-22/include/clang/stdint.h:176:10: warning: '__uint_least32_t' macro redefined [-Wmacro-redefined]
# define __uint_least32_t uint32_t
         ^
/home/gebner/.elan/toolchains/leanprover--lean4---nightly-2021-12-22/include/clang/stdint.h:102:10: note: previous definition is here
# define __uint_least32_t uint64_t
         ^
/home/gebner/.elan/toolchains/leanprover--lean4---nightly-2021-12-22/include/clang/stdint.h:177:10: warning: '__int_least16_t' macro redefined [-Wmacro-redefined]
# define __int_least16_t int32_t
         ^
/home/gebner/.elan/toolchains/leanprover--lean4---nightly-2021-12-22/include/clang/stdint.h:103:10: note: previous definition is here
# define __int_least16_t int64_t
         ^
```

tydeu:
@gebner I also had some issues trying to use the new `leanc` to compile files for Alloy on Windows. I gave up and used `cc` instead, but seeing your issues makes me wonder if they may be connected. I had two problems, the first was that `leanc` could not find `mm_malloc.h` and then (after adding the MSYS2 clang headers to the include path) it did not know what `uint8_t` and `uint16_t` were in `lean.h` even after I manually inserted a `#include stdint.h` before `lean.h`. This later issue seems very similar to the error you are seeing.

Kha:
Ah, I did see the warnings at least once before: https://github.com/leanprover/lean4/blob/546bb8f053a95180abf6e7ab509a4b3f23b61155/script/prepare-llvm-macos.sh#L39
I don't know about the MSYS errors, but I should mention a possible alternative to this PR that is halfway between it and the former behavior: we set the linker's sysroot, but not the compiler's.

PR #895, opened 2021-12-18
feat: expose more version info (e..g., toolchain)

tydeu:
Since some of the version info used by Leanpkg (and Lake) was moved into the Lean core in #894, it seemed useful to move the rest. This way toolchain version information is available to any Lean program. I also touched up `Lean.versionString` to more closely mirror what is shown with `lean --version`. This included a fix to include special descriptions for release versions which have them (e.g., `4.0.0-m2` ).


tydeu:
cc @Kha / @leodemoura

Kha:
Ideally the origin should be configurable in cmake and filled in by CI

PR #899, opened 2021-12-21
fix: do not consider worker threads as idle during startup

Kha:
Without this change, enqueuing multiple tasks before the first worker
was started led to only a single worker being created. Now the first
increment and decrement happen under the task manager mutex, so
effectively the worker is never idle until it is out of tasks.

### DIFF
--- a/src/runtime/object.cpp
+++ b/src/runtime/object.cpp
@@ -705,10 +705,10 @@ class task_manager {
 
     void spawn_worker() {
         m_num_std_workers++;
-        m_idle_std_workers++;
         lthread([this]() {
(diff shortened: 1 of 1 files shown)


Kha:
!bench

Kha:
Apparently we missed the binarytrees regression in https://github.com/leanprover/lean4/commit/816dc1895fbd49b80e823095d8f48c49f8c468b4

Kha:
The interpreted regression looks reproducible, I assume it's a big regression in binarytrees. Nothing directly comes to mind where parallelism should hurt interpreter performance.

Issue #681, opened 2021-09-19
Widgets: improve `Info` delaboration


Vtec234:
When hovering over source code, the hover spans are determined in the LSP server from [`Info`](https://github.com/leanprover/lean4/blob/c2d33a1a582c4011f5334551b903df85aa32ddc7/src/Lean/Elab/InfoTree.lean) nodes emitted by the elaborator and stored in the `InfoTree`. Parts of the elaborator include special cases to emit the right `Info` for dot notation, `|>` pipes, and other syntactic constructs.

Analogously, in the [upcoming](https://github.com/leanprover/vscode-lean4/pull/37) infoview the spans of hoverable terms are shown exactly where the *delaborator* emits `Info`s. However, we are mostly missing the special cases which means the interactive output will be less accurate than the editor input. This issue is to track where in the delaborator these should be added.

Currently `Info`s are added for structure notation [field names](https://github.com/leanprover/lean4/blob/c2d33a1a582c4011f5334551b903df85aa32ddc7/src/Lean/PrettyPrinter/Delaborator/Builtins.lean#L223), and [generically](https://github.com/leanprover/lean4/blob/c2d33a1a582c4011f5334551b903df85aa32ddc7/src/Lean/PrettyPrinter/Delaborator/Basic.lean#L228) in `delabFor`.

Vtec234:
**Partial function application**.

Observe that in the goal state for
```lean
example : id 2 = id 2 := by
```
hovering over `id` shows the full application:

![screenshot](https://user-images.githubusercontent.com/13901751/133944508-c9d2bbd3-0e71-4361-b100-2584c76303c3.png)

Instead, we should have an `Info` node for just `id` there. This would be consistent with source code hovers:

![screenshot](https://user-images.githubusercontent.com/13901751/133944540-fa0b8bae-791e-456f-9345-b33bf67a8fe1.png)

tydeu:
@Vtec234 Honestly, I much prefer the type of the full application. One of the annoyances I often have when debugging type errors is that hovering over a type in code does *not* show a function's applied type but rather its bare type -- being able to see what the elaborator inferred for a term is quite helpful. That is, I would personally prefer that this behavior be ported to the source code hovers not the other way around

Vtec234:
> One of the annoyances I often have when debugging type errors is that hovering over a type in code does not show a function's applied type but rather its bare type

[...]

PR #902, opened 2021-12-21
fix: widgets: do not highlight entire expression in popup

gebner:
This is the last remaining issue from #681, as far as I can tell:

> Another issue is that we don't want a top-level span in the tooltip. For example if you click on `id 2`, then the tooltip shows `@id Nat 2`, which you can click on again to get `@id Nat 2` etc.

This PR removes this infinite highlight cycle.

### DIFF
--- a/src/Lean/Widget/InteractiveCode.lean
+++ b/src/Lean/Widget/InteractiveCode.lean
@@ -129,6 +129,7 @@ def exprToInteractiveExplicit (e : Expr) : MetaM CodeWithInfos := do
     openDecls := ← getOpenDecls
     fileMap := arbitrary
   }
+  let infos := infos.erase 1 -- remove highlight for entire expression in popups
   tagExprInfos ctx infos tt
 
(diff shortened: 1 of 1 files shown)


Vtec234:
@gebner I am not sure this is the best place to remove the info. Note that we also use `exprToInteractive` in popups if the info does not have a type (to be fair, I can't remember if that ever happens) but can't remove the info there because that function is also used in cases where we do want it. We could dedicate the `explicit` variant to just interactive popups (it's only used there anyway) but in that case should clearly mark it as such.

gebner:
> Note that we also use exprToInteractive in popups if the info does not have a type

I can't find that in the `infoToInteractive` function.  The expression itself is only formatted using `exprToInteractiveExplicit`.

> We could dedicate the explicit variant to just interactive popups (it's only used there anyway) but in that case should clearly mark it as such.

I kind of assumed that, so I think the best course of action would indeed be to document it there.

PR #903, opened 2021-12-21
feat: LSP request handler chaining

tydeu:
It was [discussed](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/IntellISense.20for.20Language.20Embeddings.3F/near/265672728) on Zulip that one way to support LSP middleware is to override the request handler (and then invoking the original handler from within). From what I can tell, removing the check is all that is necessary to permit this.

### DIFF
--- a/src/Lean/Server/Requests.lean
+++ b/src/Lean/Server/Requests.lean
@@ -170,7 +170,7 @@ def chainLspRequestHandler (method : String)
     let handle := fun j => do
       let t ← oldHandler.handle j
       let t ← t.map fun x => x.bind fun j => FromJson.fromJson? j |>.mapError fun e =>
-        IO.userError s!"failed to parse original LSP request handler for `{method}` when chaining: {e}"
+        IO.userError s!"Failed to parse original LSP response for `{method}` when chaining: {e}"
       let params ← liftExcept <| parseRequestParams paramType j
       let t ← handler params t
       t.map <| Except.map ToJson.toJson


gebner:
As I've said on Zulip, overriding is not the ideal solution here.  If we have multiple completion handlers, then we want to aggregate the results and not replace them.

If we're going for less principled temporary workarounds, then you can just copy&paste the `registerLspRequestHandler` function into alloy.

tydeu:
@gebner Aggregation only applies to completion handler though, right? For things like hovers or highlighting that would desire a handler overwrite, correct? Thus, this seemed like the most straightforward (and powerful) short term solution.

One improvement, though, could be to replace just the inner Lean request handler itself rather than the handler for the whole LSP method (as that is unlikely to change).

Vtec234:
Some response types are more compositional than others -- e.g. completions could be concatenated, while for a hover composition is less clear but could still be `Option` composition where the new handler runs iff the original returned `none` -- but regardless of the "natural" composition method, we should not impose it on users, I think. In particular users may want to run the original handler depending on some condition, or always run it and then compose, or always ignore and override it. Only allowing re-registrations does not make all of these possible. My suggestion would be to leave `register` as-is, and add something like the following (* you might have to fudge the types to make things work, not sure):
```lean
def registerChainedLspRequestHandler (method : String)
    paramType [FromJson paramType] [FileSource paramType]
    respType [ToJson respType]
    (handler : paramType → RequestTask respType → RequestM (RequestTask respType)) : IO Unit
```
which is almost the same but gets also the result of the original handler. Then the new handler can choose to ignore the original one and do something else, or to do something based on the result. For ignoring the original one might want to cancel its task. This technically still incurs the overhead of *creating* the task, so we might want an `overrideLspRequestHandler` variant much like the one you proposed here. Thoughts?

tydeu:
@Vtec234 Thanks for the comment! Some thoughts:

>  Only allowing re-registrations does not make all of these possible. 

What do you mean by this? I don't think either of us proposed just allowing re-registrations? Both of my proposals allowed for arbitrary registration and overriding and Gabriel's (as I understand it) was just aggregation with no re-registration at all?

> My suggestion would be to leave `register` as-is

When you say "as-is" do you mean with the check preventing overrides (as in Lean master) or without (as in this PR).

> And add something like the following" `def registerChainedLspRequestHandler ...` ... For ignoring the original one might want to cancel its task. This technically still incurs the overhead of _creating_ the task, so we might want an `overrideLspRequestHandler` variant much like the one you proposed here. Thoughts?

Another limitation of `registerChainedLspRequestHandler` is that it only permits chaining one request handler into another (not multiple).

Personally, I think calling one or more handlers from inside another would be simple, general, and sufficiently principled solution for composition. Even if a user wishes to chain a new handler with whatever handler was previously activate for a method, they can do so, during registration of the new handler, by calling `lookupLspRequestHandler` to get the current handler, use that old handler to build the new handler, and then register that new handler.

Vtec234:
> When you say "as-is" do you mean with the check preventing overrides (as in Lean master) or without (as in this PR).

I meant as in master, adding new functions instead.

> Another limitation of registerChainedLspRequestHandler is that it only permits chaining one request handler into another (not multiple).

Do you have a use case for calling a handler for *another* request kind from your own? It seems like at that point, if one needs so much extra information, one should use the Lean APIs directly.

> Even if a user wishes to chain a new handler with whatever handler was previously activate for a method, they can do so, during registration of the new handler, by calling lookupLspRequestHandler to get the current handler, use that old handler to build the new handler, and then register that new handler.

Yep that is pretty much what I imagined `registerChained` would do. It isn't much more than a convenience wrapper around this process which unpacks the data so you get a value of `respType` instead of JSON.

> Personally, I think calling one or more handlers from inside another would be simple, general, and sufficiently principled solution for composition.

How about: we add an `override : Bool` parameter to `register` which disables the check but also makes it explicit that overriding is happening, and add `registerChained` for convenience?

tydeu:
@Vtec234 

> > Another limitation of registerChainedLspRequestHandler is that it only permits chaining one request handler into another (not multiple).
> 
> Do you have a use case for calling a handler for _another_ request kind from your own? It seems like at that point, if one needs so much extra information, one should use the Lean APIs directly.

Ah sorry, my concern wasn't about requests of a *different* kind. I was saying the the `reigisterChained` signature only allows the handler to handle the result of the immediately preceding handler. In some cases you may wish to dispatch to multiple different handlers and the compose them in some complex way. 

For instance, for hovers, it may be desired to combine information from multiple different handlers in some complex way (e.g., the hover from Lean, the hover from an embedded DSL, the hover from some documentation extension, etc.) all within one new handler. Thus, the handler needs access to the result of all 3 as only it knows how to properly combine it.

---

> How about: we add an `override : Bool := False` default parameter to `register` which disables the check but also makes it explicit that overriding is happening, and add `registerChained` for convenience?

This sounds like a good compromise to me. It still enables the most general solution for those who need it, but provides a nice helper for the simpler cases.

@gebner, does this sound fine to you as well?

Vtec234:
> For instance, for hovers, it may be desired to combine information from multiple different handlers in some complex way (e.g., the hover from Lean, the hover from an embedded DSL, the hover from some documentation extension, etc.) all within one new handler. Thus, the handler needs access to the result of all 3 as only it knows how to properly combine it.

Is this for the embedded C in Alloy? It sounds like you would still just want to add one chained handler then. It could dispatch back to Lean when the cursor is in a Lean region, otherwise do whatever it needs to process C code. But it doesn't sound like you would want to add a separate handler *just for* C and then invoke that, as well as the Lean handler, from a third one.

tydeu:
@Vtec234 You are correct. For Alloy, the chaining approach is sufficient. For the example, though I was hypothesizing a further case where in addition to Lean and the DSL the was also an additional extension (e.g., one provided snippets, richer documentation or syntax highlighting) that wants to combine the information from the other two in some intelligent way.. However, I do imagine for most common cases, chaining would be sufficient.

gebner:
> > How about: we add an `override : Bool := False` default parameter to `register` which disables the check but also makes it explicit that overriding is happening, and add `registerChained` for convenience?
> 
> This sounds like a good compromise to me. It still enables the most general solution for those who need it, but provides a nice helper for the simpler cases.
> 
> @gebner, does this sound fine to you as well?

-1 for the override flag.  The problem is not that the override happens implicitly or accidentally, but that it happens at all.  I don't see any scenario where replacing the existing handler is the correct thing to do.  Even if you correctly implement the fallback to the normal completion handler, stuff will still break if somebody imports let's say Alloy and Alloy2.

Wojciech's chaining suggestion seems like the simplest workable solution to me.

What editors like vscode do is implement the aggregation logic for each request separately.  They send the request to all registered handlers.  Depending on the request type it either picks one result or the union, etc.  Each handler also has a priority,  which controls how the results are combined (only pick the highest-priority one / sort according to priority / ...).

Particularly for the completion request I think it would be very very valuable to have such an infrastructure in place, where you can register completion handlers for other syntax.  Right now the completion is hard-coded and only works for identifiers/options/namespaces/etc.  For example `conv` doesn't have auto-completion, and if it were defined in mathlib it would be impossible to add auto-completion for it.  Alloy isn't the only syntax that wants to add custom autocompletion.  I think it would be best if we had a way to register completion handlers in a general way, powerful enough that we can e.g. implement `set_option` completion using this infrastructure.

Kha:
This seems like a good time to mention that I would like to refactor the completion system to rely primarily on the syntax tree and use the info tree only when necessary, which would simplify extending it (but we also want an extensible info tree at some point). It's not very high on my to-do list though.

tydeu:
I implemented the chaining idea. 
 
> If we're going for less principled temporary workarounds, then you can just copy&paste the `registerLspRequestHandler` function into alloy.

This is actually not possible in current master because  the `RequestHandler` structure is `private`. I changed that so it is at least possible for plugins to experiment with such less principled things.

tydeu:
@Kha / @gebner / @Vtec234, what do you think of the PR now?

src/Lean/Server/Requests.lean
```diff
@@ -150,9 +150,35 @@ def registerLspRequestHandler (method : String)
...
+    paramType [FromJson paramType]
+    respType [FromJson respType] [ToJson respType]
+    (handler : paramType → RequestTask respType → RequestM (RequestTask respType)) : IO Unit := do
+  if !(← IO.initializing) then
+    throw <| IO.userError s!"Failed to chain LSP request handler for '{method}': only possible during initialization"
+  if let some oldHandler ← lookupLspRequestHandler method then
+    let handle := fun j => do
+      let t ← oldHandler.handle j
+      let t ← t.map fun x => x.bind fun j => FromJson.fromJson? j |>.mapError fun e =>
+        IO.userError s!"failed to parse original LSP request handler for `{method}` when chaining: {e}"
```
Vtec234, line 173:
```suggestion
        IO.userError s!"Failed to parse original LSP response for `{method}` when chaining: {e}"
```
Since I guess it's not the handler being parsed :)

Vtec234 (approved):

Issue #467, opened 2021-05-15
Revise closed term extraction


Kha:
Currently closed terms are evaluated eagerly at startup, which creates unnecessary overhead and linking dependencies. We want to move to bounded normalization at compile time, emit static code as in https://github.com/leanprover/lean4/pull/122 if normalization succeeded, and otherwise create static code for a thunk to implement lazy once-only evaluation.

Kha:
Note: we should double-check to not compile-time evaluate any values dependent on `getIsWindows` and similar functions.

hargoniX:
We did it!

PR #904, opened 2021-12-22
perf: do not specialize Inhabited

Kha:
We have many functions whose only typeclass parameter is `Inhabited`, for which specialization is not really productive and indeed can be counter-productive if it leads to code from compiled libraries (such as stdlib) being duplicated into interpreted libraries. A low-hanging fruit I noticed while looking at interpreter usage in mathlib4.

### DIFF
--- a/src/library/compiler/specialize.cpp
+++ b/src/library/compiler/specialize.cpp
@@ -184,7 +184,11 @@ environment update_spec_info(environment const & env, comp_decls const & ds) {
                 if (is_const(fn) && has_nospecialize_attribute(env, const_name(fn))) {
                     info.second.push_back(spec_arg_kind::Fixed);
                 } else {
-                    info.second.push_back(spec_arg_kind::FixedInst);
+                    type_checker tc(env, lctx);
+                    if (tc.is_prop(type))
+                        info.second.push_back(spec_arg_kind::FixedNeutral);
+                    else
+                        info.second.push_back(spec_arg_kind::FixedInst);
                 }
             } else {
                 type_checker tc(env, lctx);


Kha:
!bench

Kha:
!bench

Kha:
*what*

Kha:
@leodemoura The qsort speedup is pretty interesting, it looks like it is purely from passing around `Inhabited UInt32` as a `uint32_t` instead of repeatedly loading it from an extracted constant (I wonder why it didn't inline the instance?)
```
1211c1209
< static lean_object* _init_l___private_qsort_0__partitionAux___at_main___spec__4___boxed__const__1() {
---
> LEAN_EXPORT lean_object* l___private_qsort_0__partitionAux___at_main___spec__4(uint32_t x_1, uint32_t x_2, uint32_t x_3, lean_object* x_4, uint32_t x_5, uint32_t x_6) {
1214,1225c1212,1214
< uint32_t x_1; lean_object* x_2;
< x_1 = l_instInhabitedUInt32;
< x_2 = lean_box_uint32(x_1);
< return x_2;
< }
< }
< LEAN_EXPORT lean_object* l___private_qsort_0__partitionAux___at_main___spec__4(uint32_t x_1, uint32_t x_2, lean_object* x_3, uint32_t x_4, uint32_t x_5) {
```

leodemoura:
> I wonder why it didn't inline the instance?

@Kha Why did you expect it to be inlined?

Kha:
I assumed it might get inlined because of its small size, but I don't recall our exact inlining heuristics.

leodemoura:
> I assumed it might get inlined because of its small size, but I don't recall our exact inlining heuristics.

The code is here https://github.com/leanprover/lean4/blob/master/src/library/compiler/csimp.cpp#L1713

I think it is not worth investing too much time here since we will rewrite this code in Lean.

Kha:
I see, we skip all constants by default. That makes sense of course for anything more complicated than an unboxed literal.

Kha:
I suppose if we compiled the benchmarks with panic=abort semantics and activated LTO, LLVM might be able to completely get rid of the Inhabited instance. But that's not something I want to experiment with right now.

gebner:
> LLVM might be able to completely get rid of the Inhabited instance.

I'd be surprised if this worked.  Because the global variable is 0 before initialization (this is not UB!) and 1 after initialization.  And I doubt that LLVM can figure out that qsort is only ever called after initialization.

Kha:
Right, it might also need #467

gebner:
From what I can tell this PR has broken TCO in the `overflow1.lean` file.  (Or maybe it was one of the other specialization PRs, but in any case the specialize pass is the one that changes after updating stage0.)

Before:
```
[compiler.specialize] new code Array.foldlMUnsafe.fold._at.main._spec_4
fun as i stop b a =>
  let _x_4 := USize.decEq i stop;
  let _x_5 := decide (i = stop);
  Bool.casesOn _x_5
    (let _x_74 := Array.uget as i (_ : USize.toNat i < Array.size as);
    let _x_106 := Char.ofNat 122;
    let _x_107 := instDecidableEqChar _x_74 _x_106;
    Decidable.casesOn _x_107
      (fun x =>
        let _x_113 := Nat.add b 1;
        let _x_114 := USize.ofNat 1;
        let _x_115 := USize.add i _x_114;
        Array.foldlMUnsafe.fold._at.main._spec_3 as _x_115 stop _x_113 a)
      fun x =>
      let _x_109 := panic._at.main._spec_2 "z" a;
      let _x_4._do_jp := _x_109;
      Prod.casesOn _x_4._do_jp fun fst snd =>
        let _x_116 := USize.ofNat 1;
        let _x_117 := USize.add i _x_116;
        Array.foldlMUnsafe.fold._at.main._spec_3 as _x_117 stop fst snd)
    (b, a)
```

After:
```
[compiler.specialize] new code Array.foldlMUnsafe.fold._at.main._spec_2
fun _x_2 as i stop b a =>
  let _x_33 := USize.decEq i stop;
  let _x_34 := decide (i = stop);
  Bool.casesOn _x_34
    (let _x_35 := _x_2.2;
    let _x_36 := _x_35.1;
    let _x_37 := Array.uget as i (_ : USize.toNat i < Array.size as);
    let _x_38 := fun a =>
      let _x_42 := Char.ofNat 122;
      let _x_43 := instDecidableEqChar _x_37 _x_42;
      Decidable.casesOn _x_43
        (fun x =>
          let _x_44 := Nat.add b 1;
          (_x_44, a))
        fun x =>
        let _x_45 := instInhabited;
        panic "z" a;
    let _x_39 := fun a a_1 =>
      let _x_40 := USize.ofNat 1;
      let _x_41 := USize.add i _x_40;
      Array.foldlMUnsafe.fold._at.main._spec_2 _x_2 as _x_41 stop a a_1;
    _x_36 _x_38 _x_39 a)
    (let _x_46 := _x_2.1;
    let _x_47 := _x_46.2;
    let _x_48 := _x_47.1;
    _x_48 b a)
```

Kha:
Thanks, fixed in https://github.com/leanprover/lean4/commit/132898f7e7914be10dd3a70014cdc9b6456c875e

PR #908, opened 2021-12-23
Several formatter fixes.

gebner:
Also fixes a round-tripping error with the `calc` syntax in mathport: https://github.com/leanprover-community/mathport/issues/85

### DIFF
--- a/src/Init/NotationExtra.lean
+++ b/src/Init/NotationExtra.lean
@@ -196,6 +196,11 @@ macro_rules
     `($mods:declModifiers class $id $params* extends $[$parents:term],* $[: $ty]?
       attribute [instance] $ctor)
 
+/-- `· tac` focuses on the main goal and tries to solve it using `tac`, or else fails. -/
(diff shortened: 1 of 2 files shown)


src/Init/NotationExtra.lean
```diff
@@ -196,6 +196,11 @@ macro_rules
     `($mods:declModifiers class $id $params* extends $[$parents:term],* $[: $ty]?
       attribute [instance] $ctor)
 
+/-- `· tac` focuses on the main goal and tries to solve it using `tac`, or else fails. -/
+syntax ("·" <|> ".") ppHardSpace many1Indent(group(tactic ";"? ppLine)) : tactic
```
Kha, line 200: Could/should we change the initial `ppLine` in `tactiSeqIndented` to `ppHardlineUnlessUngrouped` so that callers can opt out of it?

Kha: Well, that really only makes sense for when the syntax prefix is exactly the indentation size, so further use cases don't seem all that likely.

PR #911, opened 2021-12-23
Delaboration of constants in function applications

hargoniX:
As described in https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Function.20application.20delaboration it would be nice to be able to (optionally) add info tags to the syntactic node representing the function in a function application, instead of just the entire function application, for automatic documentation generation purposes. However this would break the LSP's hovering behavior so this feature is disabled per default. Furthermore this PR fixes a few bugs along the delaboration/formatting pipeline which dropped the generated information at two places.

### DIFF
--- a/tests/lean/ppNotationCode.lean.expected.out
+++ b/tests/lean/ppNotationCode.lean.expected.out
@@ -34,11 +34,12 @@ fun x =>
         let discr_3 := Lean.Syntax.getArg discr_2 1;
         let rhs := discr_3;
         let lhs := discr;
-        do
-        let info ← Lean.MonadRef.mkInfoFromRefPos
-        let _ ← Lean.getCurrMacroScope
-        let _ ← Lean.getMainModule
-        pure (Lean.Syntax.node Lean.SourceInfo.none `«term_+++_» #[lhs, Lean.Syntax.atom info "+++", rhs])
+        let f := discr_1;
+        Lean.withRef f do
+          let info ← Lean.MonadRef.mkInfoFromRefPos
+          let _ ← Lean.getCurrMacroScope
+          let _ ← Lean.getMainModule
+          pure (Lean.Syntax.node Lean.SourceInfo.none `«term_+++_» #[lhs, Lean.Syntax.atom info "+++", rhs])
       else
         let discr := Lean.Syntax.getArg discr 1;
         throw ())


src/Init/Prelude.lean
```diff
@@ -2262,8 +2262,13 @@ abbrev Unexpander := Syntax → UnexpandM Syntax
...
 instance : MonadQuotation UnexpandM where
-  getRef              := pure Syntax.missing
-  withRef             := fun _ => id
+  getRef              := get
+  withRef             := fun ref m => do
+    let orig ← get
+    set ref
+    let res ← m
+    set orig
+    return res
```
gebner, line 2271:
```suggestion
    try m finally set orig
```

src/Lean/Elab/Notation.lean
```diff
@@ -55,13 +55,13 @@ def mkSimpleDelab (attrKind : Syntax) (vars : Array Syntax) (pat qrhs : Syntax)
     -- replace head constant with (unused) antiquotation so we're not dependent on the exact pretty printing of the head
     `(@[$attrKind:attrKind appUnexpander $(mkIdent c):ident]
       aux_def unexpand $(mkIdent c) : Lean.PrettyPrinter.Unexpander := fun
-       | `($$(_):ident $args*) => `($pat)
+       | `($$tk:ident $args*) => withRef tk `($pat)
        | _                     => throw ())
```
gebner, line 59:
```suggestion
       | `($$tk:ident $args*) => withRef tk `($pat)
       | _                    => throw ())
```
Please preserve the formatting.

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
```diff
@@ -100,11 +100,17 @@ def delabConst : Delab := do
           c := `_root_ ++ c
         else
           c := c₀
-      return mkIdent c
+      mkIdent c
```
gebner, line 103: This changes the behavior of the code (previously it was an early return).  But I think the new behavior is correct, since universe parameters don't have an effect on whether we need an explicit (`@`) annotation.

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
```diff
@@ -100,11 +100,17 @@ def delabConst : Delab := do
...
       `($(mkIdent c).{$[$(ls.toArray.map quote)],*})
 
-  maybeAddBlockImplicit stx
+  let stx ← maybeAddBlockImplicit stx
+  if (←getPPOption getPPTagSymbols) then
+    let stx ← annotateCurPos stx
+    addTermInfo (←getPos) stx (←getExpr)
+    stx
+  else
+    stx
```
gebner, line 113:
```suggestion
  let mut stx ← maybeAddBlockImplicit stx
  if (← getPPOption getPPTagSymbols) then
    stx ← annotateCurPos stx
    addTermInfo (← getPos) stx (← getExpr)
  stx
```

gebner:
> Furthermore this PR fixes a few bugs along the delaboration/formatting pipeline which dropped the generated information at two places.

This was the `fun x => 42` where the `x` wasn't tagged, right?  Do we also tag it with the correct position now?  Did you fix anything else?

It would be cool if we had tests for that (like `infoTree.lean` but for delaboration), but I guess that can go in another PR.

hargoniX:
The first place where they were dropped was in automatically generated delaborators for syntax that was defined using `notation` and derived keywords. The second one was in the formatter, which simply ignored the information that was now being preserved by the notation change.

src/Init/Prelude.lean
hargoniX: This change would actually break compilation since we are working in Prelude.lean where this is not available yet (I'll revert it).

src/Lean/PrettyPrinter/Delaborator/Options.lean
```diff
@@ -90,6 +90,11 @@ register_builtin_option pp.safeShadowing  : Bool := {
   group    := "pp"
   descr    := "(pretty printer) allow variable shadowing if there is no collision"
 }
+register_builtin_option pp.tagSymbols : Bool := {
```
gebner, line 93: The name here is a bit misleading, the option does not only affect symbols (which I associate with `+`, etc.) but any head symbol (such as `f` in `f a b`).  The terminology isn't great, we have lots of things that are called symbols.

src/Init/Prelude.lean
```diff
@@ -2251,7 +2251,7 @@ export Macro (expandMacro?)
 
 namespace PrettyPrinter
 
-abbrev UnexpandM := EStateM Unit Unit
+abbrev UnexpandM := EStateM Unit Syntax
```
Kha, line 2254:
```suggestion
abbrev UnexpandM := ReaderT Syntax (EStateM Unit Unit)
```
My bad.

src/Init/Prelude.lean
Kha, line 2271:
```suggestion
  getRef              := read
  withRef ref x       := withReader (fun _ => ref) x
  -- unexpanders should not need to introduce new names
```
to match ReaderT, also move comment

src/Init/Prelude.lean
```diff
@@ -2262,8 +2262,13 @@ abbrev Unexpander := Syntax → UnexpandM Syntax
 
 -- unexpanders should not need to introduce new names
```
Kha, line 2263:
```suggestion
```

src/Lean/Elab/Notation.lean
```diff
@@ -55,13 +55,13 @@ def mkSimpleDelab (attrKind : Syntax) (vars : Array Syntax) (pat qrhs : Syntax)
...
-       | _                     => throw ())
+       | `($$tk:ident $args*) => withRef tk `($pat)
+       | _                    => throw ())
   | `($c:ident)        =>
     let [(c, [])] ← Macro.resolveGlobalName c.getId | failure
     `(@[$attrKind:attrKind appUnexpander $(mkIdent c):ident]
       aux_def unexpand $(mkIdent c) : Lean.PrettyPrinter.Unexpander := fun
-       | `($$(_):ident) => `($pat)
+       | `($$tk:ident) => withRef tk `($pat)
        | _              => throw ())
```
Kha, line 65:
```suggestion
       | _             => throw ())
```

src/Lean/Elab/Notation.lean
```diff
@@ -55,13 +55,13 @@ def mkSimpleDelab (attrKind : Syntax) (vars : Array Syntax) (pat qrhs : Syntax)
     -- replace head constant with (unused) antiquotation so we're not dependent on the exact pretty printing of the head
     `(@[$attrKind:attrKind appUnexpander $(mkIdent c):ident]
       aux_def unexpand $(mkIdent c) : Lean.PrettyPrinter.Unexpander := fun
-       | `($$(_):ident $args*) => `($pat)
-       | _                     => throw ())
+       | `($$tk:ident $args*) => withRef tk `($pat)
```
Kha, line 58: It's not really a token, is it? And could you add an explanatory comment? It is not clear to me who actually consumes that ref.

Kha (changes requested):

Kha:
nit: you shouldn't capitalize the first word in commit messages

src/Lean/PrettyPrinter/Delaborator/Options.lean
hargoniX: Maybe `tagAllFunctions` ?

src/Lean/Elab/Notation.lean
```diff
@@ -53,16 +53,17 @@ def mkSimpleDelab (attrKind : Syntax) (vars : Array Syntax) (pat qrhs : Syntax)
     guard <| args.all (Syntax.isIdent ∘ getAntiquotTerm)
     guard <| args.allDiff
     -- replace head constant with (unused) antiquotation so we're not dependent on the exact pretty printing of the head
+    -- The reference is attached to the syntactic representation of the called function itself, not the entire function application
     `(@[$attrKind:attrKind appUnexpander $(mkIdent c):ident]
       aux_def unexpand $(mkIdent c) : Lean.PrettyPrinter.Unexpander := fun
-       | `($$(_):ident $args*) => `($pat)
-       | _                     => throw ())
+       | `($$fun_ref:ident $args*) => withRef fun_ref `($pat)
```
Kha, line 59: That doesn't match our naming convention, let's go with just `f`? Same below. And mind the indentation.

Kha: By the way, the *why* I was missing yesterday is that the quotation uses the reference as the source location of any newly introduced identifiers, which in this case is the function head.

src/Lean/PrettyPrinter/Delaborator/Options.lean
Kha: The `All` seems redundant. How about `tagAppFns`, in reference to `Expr.getAppFn`?

Kha: Would also be good to have a comment that if the constant is not part of an application, the tagging is unconditionally done by `delabFor`.

Kha (approved):

src/Lean/Elab/Notation.lean
```diff
@@ -53,16 +53,17 @@ def mkSimpleDelab (attrKind : Syntax) (vars : Array Syntax) (pat qrhs : Syntax)
...
+       | _                   => throw ())
   | `($c:ident)        =>
     let [(c, [])] ← Macro.resolveGlobalName c.getId | failure
     `(@[$attrKind:attrKind appUnexpander $(mkIdent c):ident]
       aux_def unexpand $(mkIdent c) : Lean.PrettyPrinter.Unexpander := fun
-       | `($$(_):ident) => `($pat)
-       | _              => throw ())
+       | `($$f:ident) => withRef f `($pat)
+       | _            => throw ())
   | _                  => failure
```
hargoniX, line 67: Now I'm unsure with the indentation here, is this correct? Should it be further to the left? Is there some written down style guide for this?

Kha:
Merged manually

PR #917, opened 2021-12-28
chore: use a different syntax kind for suffices-by

digama0:
There are other ways to address this issue, but this seems like the simplest fix. The problem that appears in mathport is that we want to replace `by exact $t` -> `$t`, but when this appears as a subterm of a suffices, like `suffices Nat by exact 1`, we get the malformed expression `suffices Nat 1`, because `suffices` takes only a term of the form `by $tacs`, not any term, even though the syntax kind of this `by` is the same as the usual one.

The fix implemented here is to have a separate syntax kind `byTactic'` for use in `show` and `suffices`, so that we cannot confuse `(byTactic'| by exact $t)` with `(term| by exact $t)`.

### DIFF
--- a/src/Lean/Parser/Term.lean
+++ b/src/Lean/Parser/Term.lean
@@ -82,9 +82,10 @@ def parenSpecial : Parser := optional (tupleTail <|> typeAscription)
 @[builtinTermParser] def anonymousCtor := leading_parser "⟨" >> sepBy termParser ", " >> "⟩"
 def optIdent : Parser := optional (atomic (ident >> " : "))
 def fromTerm   := leading_parser "from " >> termParser
-def sufficesDecl := leading_parser optIdent >> termParser >> ppSpace >> (fromTerm <|> byTactic')
+def showRhs := fromTerm <|> byTactic'
+def sufficesDecl := leading_parser optIdent >> termParser >> ppSpace >> showRhs
 @[builtinTermParser] def «suffices» := leading_parser:leadPrec withPosition ("suffices " >> sufficesDecl) >> optSemicolon termParser
-@[builtinTermParser] def «show»     := leading_parser:leadPrec "show " >> termParser >> ppSpace >> (fromTerm <|> byTactic')
+@[builtinTermParser] def «show»     := leading_parser:leadPrec "show " >> termParser >> ppSpace >> showRhs
 def structInstArrayRef := leading_parser "[" >> termParser >>"]"
 def structInstLVal   := leading_parser (ident <|> fieldIdx <|> structInstArrayRef) >> many (group ("." >> (ident <|> fieldIdx)) <|> structInstArrayRef)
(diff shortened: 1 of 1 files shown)


digama0:
I think the error is because this needs a bootstrap update. Help!

src/Lean/Parser/Term.lean
```diff
@@ -51,6 +51,7 @@ namespace Term
 /- Built-in parsers -/
 
 @[builtinTermParser] def byTactic := leading_parser:leadPrec ppAllowUngrouped >> "by " >> Tactic.tacticSeq
+def byTactic' := leading_parser:leadPrec ppAllowUngrouped >> "by " >> Tactic.tacticSeq
```
gebner, line 54:
```suggestion
def byTactic' := leading_parser "by " >> Tactic.tacticSeq
```

The `ppAllowUngrouped` only has an effect for category parsers.  I also don't think the precedences do anything for non-category parsers.

src/Lean/Parser/Term.lean
```diff
@@ -74,9 +75,9 @@ def parenSpecial : Parser := optional (tupleTail <|> typeAscription)
 @[builtinTermParser] def anonymousCtor := leading_parser "⟨" >> sepBy termParser ", " >> "⟩"
 def optIdent : Parser := optional (atomic (ident >> " : "))
 def fromTerm   := leading_parser "from " >> termParser
-def sufficesDecl := leading_parser optIdent >> termParser >> ppSpace >> (fromTerm <|> byTactic)
+def sufficesDecl := leading_parser optIdent >> termParser >> ppSpace >> (fromTerm <|> byTactic')
```
gebner, line 78:
```suggestion
def sufficesDecl := leading_parser optIdent >> termParser >> ppSpace >> showRhs
```
Might be worth it to add a common declaration for both suffices&show.

gebner:
> I think the error is because this needs a bootstrap update. Help!

Indeed, you just need to update stage0.  It's because the show-elaborator is implemented using a syntax matcher; and the syntax matcher uses the syntax as parsed by the stage0 compiler (since show is built-in notation).

src/Lean/Parser/Term.lean
```diff
@@ -51,6 +51,7 @@ namespace Term
 /- Built-in parsers -/
 
 @[builtinTermParser] def byTactic := leading_parser:leadPrec ppAllowUngrouped >> "by " >> Tactic.tacticSeq
+def byTactic' := leading_parser "by " >> Tactic.tacticSeq
```
leodemoura, line 54: Could you please add a comment explaining why `byTactic'` has been added? It can be based on the PR comment.

leodemoura:
Merged manually because of update stage0 conflicts.

leodemoura:
@digama0 Thanks for adding the comment.

PR #835, opened 2021-11-30
Implement single-file "find references" in LSP server

Garmelon:
This implementation will only find occurrences in the current file (even if the declaration is in another file). Works for global definitions and local variables.

The relevant issue #523 should probably not yet be closed because of the single-file limitation.

I still need to add a few tests.

PR #925, opened 2022-01-05
Implement "find references" in LSP server

### LINKED ISSUE #523: LSP: find references
Apart from the go to definition, it would be great to have the inverse, i.e. find references. It's a nicer experience than `grep` ;-)

Garmelon:
After #835, this PR implements cross-file "find references".

image

The way it works is:

1. `lean`, when called with `-i <file.ilean>`, generates an `.ilean` file (similar to `-o <file.olean>`) containing information about the file's definitions and usages (in JSON).
2. `leanbundle` is used to create an `.ilean-bundle` file combining the `.ilean` files of all active `.lean` files.
3. The LSP server's watchdog looks for all `.ilean-bundle` files in the olean path and loads them (allowing for dependencies' `.ilean-bundle`s to be loaded as well). It also listens for changed `.ilean-bundle` files in the current workspace via [DidChangeWatchedFiles notification](https://microsoft.github.io/language-server-protocol/specification.html#workspace_didChangeWatchedFiles) and loads/reloads/unloads the bundles.
4. LSP server workers also send their definitions and usages to the watchdog whenever they finish elaborating a file after it is opened or changed.
5. The watchdog collects all this information and uses it to respond to [references requests](https://microsoft.github.io/language-server-protocol/specification.html#textDocument_references).

This way, references of symbols can be found across files. Changes in open files are incorporated directly, while unopened files use the state from the last compilation. In the future, the [declaration](https://microsoft.github.io/language-server-protocol/specification.html#textDocument_declaration), [definition](https://microsoft.github.io/language-server-protocol/specification.html#textDocument_definition) and [typeDefinition](https://microsoft.github.io/language-server-protocol/specification.html#textDocument_typeDefinition) requests could also be implemented using this infrastructure, as well as finding unused definitions.

Creating the `.ilean-bundle` files is the responsibility of the build system, since the compiler or `leanbundle` doesn't know which `.ilean` files belong to active `.lean` files and which are only left-over artifacts from previous compilations. This PR modifies lean's makefiles to generate and collect `.ilean` files to `.ilean-bundle` files, one for `Init`, `Std` and `Lean`.

For `lake` integration, the `findSrcSearchPath` method and olean search path initialisation in `src/Lean/Server/Watchdog.lean` would need to be adjusted and `lake` would probably need to generate `.ilean-bundle` files. I'm not sure what exact steps would be necessary.

This PR is also missing tests.

Closes #523.


src/Lean/Server/References.lean
```diff
@@ -0,0 +1,243 @@
...
+def bundle (self : References) : Bundle :=
+  let base := self.bundles.toList.foldl (init := Bundle.empty) fun l (_, r) => l.addBundle r
+  base.addBundle self.overlayBundle
+
+def findAt? (self : References) (module : Name) (pos : Lsp.Position) : Option RefIdent := Id.run do
+  if let some refs := self.bundle.find? module then
+    return refs.findAt? pos
+  none
+
+def referingTo (self : References) (ident : RefIdent) (srcSearchPath : SearchPath)
```
Kha, line 225: Let's not repeat [HTTP's mistake](https://en.wikipedia.org/wiki/HTTP_referer#Etymology) :)

src/bin/leanbundle
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
```
Kha, line 1: For context, we were discussing whether this should be its own executable or part of `lean`. I believe that a separate executable that is *not* exposed by elan is the way to go. If we go with that, I'll rewrite it in Lean.

gebner:
> leanbundle is used to create an .ilean-bundle file combining the .ilean files of all active .lean files.

Can you elaborate a bit on the motivation for the bundles?  Is this because of the file size?  Or because of the loading time?

The bundling infrastructure seems like quite a bit of hassle, requiring modification to build systems, yet another name resolution system (.lean->.ilean-bundle in addition to .lean->.olean), new tools (leanbundle).  I think there was also a plan of maybe bundling the olean files as well in the future, it would be great if this worked in tandem (i.e. bundling both) instead of mix-and-match.

There's also the issue that you need to build the whole project first before find-references will work.  For example if I open a fresh checkout of mathlib4 and edit a file, then only the file plus its dependencies are going to be compiled.  It would be great if find-references worked on the already compiled files.

(I'm trying to understand the tradeoffs here.)

Related file format question: would it make sense to store ilean files using the olean format?

> It also listens for changed .ilean-bundle files in the current workspace via DidChangeWatchedFiles notification and loads/reloads/unloads the bundles.

Note that this is not (yet) supported in neovim: https://github.com/neovim/neovim/issues/16078  But we can probably hack in a custom command in lean.nvim.  (Emacs seems to support the notification.)

Garmelon:
The reason for the bundles is because if all `.ilean` files were loaded, this may include `.ilean` files for nonexistent (moved/deleted) `.lean` files from a previous compilation that were not yet cleaned up. An alternative could be for the build system to provide the server with a list of existing .lean files (or maybe for the server to watch but not open the .lean files itself?), and for the server to watch for changes in .ilean files. This would be similar to the bundle but not require all `.ilean` files to exist before find-references can be used. However, tracking all `.ilean` files belonging to a lean project more or less live sounds more complex than bundling ilean info during compilation.

For the mathlib4 example, the current find-references would only work for the file(s) opened in the editor, not its dependencies. A list-of-lean-files approach like above would probably help.

JSON was chosen as a format because it's easy to work with, including from other languages. For example, you could write a small python script analyzing which definitions are used more often than others.

Kha:
> The reason for the bundles is because if all `.ilean` files were loaded, this may include `.ilean` files for nonexistent (moved/deleted) `.lean` files from a previous compilation that were not yet cleaned up.

On second thought, this is a (very slight) issue for .oleans as well - the build system should not allow Lean to import .oleans of deleted .lean files (this should already be true in the server thanks to `print-paths`, but not on the cmdline). It seems to me that it would be good in general if the build system ensured the output directory was "clean" in this sense, at least eventually (e.g. after a full build). Alternatively, a more fine-grained `LEAN_PATH` as proposed by @tydeu would solve this problem as well. In either case, for this PR it should be safe to assume that "orphan" files are a negligible issue.

> However, tracking all .ilean files belonging to a lean project more or less live sounds more complex than bundling ilean info during compilation.

Do you mean this from the perspective of the server or the build system? For the server this should not be much harder than managing ilean updates from workers, no?

tydeu:
> On second thought, this is a (very slight) issue for .oleans as well - the build system should not allow Lean to import .oleans of deleted .lean files

This is more difficult than it sounds as it is perfectly valid to have a  `.olean` file without associated `.lean` files (e.g., binport). It is thus not clear how the build system should distinguish between modules with sources and those without them and/or how it should track module deletions. Both of these would likely require major changes to the way modules are currently managed by Lake and/or Lean. 

However, on a more optimistic note, they are probably solvable. The former. as @Kha noted, could be addressed by the finer grained `LEAN_PATH`/`LEAN_MODULES` I suggested and for the later Lake could save a module manifest to keep track of previously compiled modules (and then delete missing modules on rebuild).

gebner:
> The reason for the bundles is because if all `.ilean` files were loaded, this may include `.ilean` files for nonexistent (moved/deleted) `.lean` files from a previous compilation that were not yet cleaned up.

We could just check `pathExists` first.

> For the mathlib4 example, the current find-references would only work for the file(s) opened in the editor, not its dependencies. A list-of-lean-files approach like above would probably help.

It should also work if we just load all ilean files in build/lib.

I'm not arguing against bundling or changing the LEAN_PATH API.  But I think it would be better to do the simplest thing here, and discuss bundling / olean resolution changes separately on their own merits.

> JSON was chosen as a format because it's easy to work with, including from other languages. For example, you could write a small python script analyzing which definitions are used more often than others.

The big advantage of the olean format is that it can be loaded in O(1) time and O(1) memory on average, thanks to the mmap optimization that Sebastian implemented.  We will most likely put more and more information into ilean files and mathlib will also be huge, so this could be a significant issue.

Other tools could still use the data in json format, you just have to convert it first.  You could even do this without any new tools using something like this:
```
echo 'import Lean open Lean def main (args : List String) : IO Unit := do let (m, _) ← readModuleData args.head!; IO.println (toJson (unsafeCast m : ModuleRefs))' | lean --stdin --run . build/lib/Mathlib/Init/WF.ilean
```

gebner:
> On second thought, this is a (very slight) issue for .oleans as well - the build system should not allow Lean to import .oleans of deleted .lean files (this should already be true in the server thanks to print-paths, but not on the cmdline).

Note that this is also the behavior in Lean 3.  https://github.com/leanprover-community/lean/pull/150  I don't think it often comes up as an issue.

In Lean 4 I think it would be easiest to implement this in Lake.  Lake already needs to parse the dependencies anyhow, it can then also check that their .lean files exist (if they're supposed to exist, which you could neatly specify in the lakefile).  There's no need to hardcode this policy in leanc.

Kha:
> We will most likely put more and more information into ilean files and mathlib will also be huge, so this could be a significant issue.

Yes, this is quite likely to happen. Certainly if we ever start reading ileans from *workers* (because we moved infos such as docstrings from oleans to ileans), we definitely want to mmap them. But at least for the watchdog, this is not trivial because it would actually have to unmap these files to reload them, which we don't do anywhere else right now and can be surprisingly tricky because any references to the region have to be dropped first. So I believe we should tackle this optimization when the need for it arises.

Vtec234:
> In the future, the declaration, definition and typeDefinition requests could also be implemented using this infrastructure, as well as finding unused definitions.

I haven't read the code yet, but from the discussion here `.ilean` sounds like a custom format. Why not [LSIF](https://microsoft.github.io/language-server-protocol/specifications/lsif/0.6.0/specification/)?

Garmelon:
> I haven't read the code yet, but from the discussion here `.ilean` sounds like a custom format. Why not [LSIF](https://microsoft.github.io/language-server-protocol/specifications/lsif/0.6.0/specification/)?

LSIF seems to be a fairly low-level format, attaching LSP responses (including absolute file URLs?) to ranges in files. This works well for files that don't change, but would be difficult to work with here. Instead of attaching information to ranges, `.ilean` files attach them directly to the symbol names and use module names instead of file urls to identify files.

src/Lean/Parser/Basic.lean
```diff
@@ -1444,7 +1444,7 @@ def longestMatchStep (left? : Option Syntax) (startSize startLhsPrec : Nat) (sta
     (s.pushSyntax successNode, prio) -- put successNode back on the stack
 
 def longestMatchMkResult (startSize : Nat) (s : ParserState) : ParserState :=
-  if !s.hasError && s.stackSize > startSize + 1 then s.mkNode choiceKind startSize else s
+  if s.stackSize > startSize + 1 then s.mkNode choiceKind startSize else s
```
Kha, line 1447: @leodemoura FYI, this change ensures that `longestMatch` returns exactly one node (assuming the nested parsers do as well) no matter what. Before that, the elaborator would elaborate *all* partial parses of an ill-formed command, which would lead to an ill-formed info tree that this PR incidentally uncovered.

Garmelon:
The watchdog now loads all `.ilean` files instead of relying on bundles. As long as an `.ilean` file has a corresponding `.lean` file when a `references` request is sent, it will be used to generate the response (this part may still be slightly broken).

Kha:
!bench

Kha:
Merged! As a reminder, this will immediately affect only the Makefile workflow. I've yet to come up with a Nix solution... :) .

src/Lean/Server/Watchdog.lean
```diff
@@ -210,6 +224,10 @@ section ServerM
           fw.erasePendingRequest id
         if let Message.responseError id _ _ _ := msg then
           fw.erasePendingRequest id
+        if let Message.notification "$/lean/ileanInfo" params := msg then
+          if let some params := params then
+            if let Except.ok params := FromJson.fromJson? $ ToJson.toJson params then
+              handleIleanInfo fw params
```
Vtec234, line 230: I'm late to the party, but do we actually want to forward `$/lean/ileaninfo` to the client? Maybe there should be a `return` here. @Garmelon

Kha:
This now works at least manually:
```
$ nix build .#iTree -o src/.ilean  # re-run to update infos
$ LEAN_PATH=src/.ilean LEAN_SRC_PATH=.:src nix run .#emacs-dev ...
```
image
And yes, there's definitely a Unicode bug somewhere in lsp-mode.

src/Lean/Server/Watchdog.lean
Garmelon: True, but would a `return ← loop` from inside the `try` lead to some sort of leak because of all the nested `try`s? I guess the check for `$/lean/ileanInfo` can just be moved to the top.

Issue #287, opened 2021-01-21
Go To Definition for Imports


Kha:

leodemoura:
and "Go To Definition for notation/macros" :)

larsk21:
> and "Go To Definition for notation/macros" :)

What is the current state of this? I have noticed that "go to definition" already works on notation/macros such as `<$>` or `#[]`, but not on, for example, `<|`.

Kha:
> I have noticed that "go to definition" already works on notation/macros such as `<$>` or `#[]`, but not on, for example, `<|`.

Go-to for builtins should be covered by #937. `<|`, however, is an unfortunate special case: since it unfolds to syntax that is made up of "original" tokens only (more specifically, the first and last token are original), the editor instead jumps to `elabApp`.

Kha:
The relevant check is at https://github.com/leanprover/lean4/blob/60a70372af3821867364b6dfc31c897ca0ea6566/src/Lean/Server/InfoUtils.lean#L131, so this affects other users of `hoverableInfoAt?` as well. A potential fix could be to choose the innermost info node *that is not contained in an info node of the same range*, but that is not immediately implementable using the current `smallestInfo?`.

PR #929, opened 2022-01-07
"Go To Definition" for Imports

larsk21:
Include imports in the info tree of the header snap and use them to enable "go to definition" for imports.

see #287

### DIFF
--- a/src/Lean/Server/FileWorker.lean
+++ b/src/Lean/Server/FileWorker.lean
@@ -197,10 +197,10 @@ section Initialization
         env := headerEnv
         fileMap := m.text
       }) (Elab.InfoTree.node
-          (Elab.Info.ofCommandInfo { elaborator := `import, stx := headerStx })
+          (Elab.Info.ofCommandInfo { elaborator := `header, stx := headerStx })
           (headerStx[1].getArgs.toList.map (fun importStx =>
             Elab.InfoTree.node (Elab.Info.ofCommandInfo {
-              elaborator := Name.anonymous,
+              elaborator := `import
               stx := importStx
             }) #[].toPersistentArray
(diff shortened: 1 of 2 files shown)


src/Lean/Server/FileWorker.lean
```diff
@@ -193,9 +193,18 @@ section Initialization
...
-      -- TODO: maybe even fill with useful stuff
-      trees := #[Elab.InfoTree.node (Elab.Info.ofCommandInfo { elaborator := `import, stx := headerStx }) #[].toPersistentArray].toPersistentArray
+      trees := #[Elab.InfoTree.context ({
+        env := headerEnv
+        fileMap := m.text
+      }) (Elab.InfoTree.node
+          (Elab.Info.ofCommandInfo { elaborator := `import, stx := headerStx })
+          (headerStx[1].getArgs.toList.map (fun importStx =>
+            Elab.InfoTree.node (Elab.Info.ofCommandInfo {
+              elaborator := Name.anonymous,
```
Kha, line 203: The ``elaborator := `import`` part looks like a bad choice of name on my part, we should probably call that one `header` and this one `import`. If we do that, can we avoid the new big `if` in `handleDefinition` by checking if `hoverableInfoAt?` returned a `CommandInfo` with this elaborator?

larsk21: That would be the nicer solution, but `hoverableInfoAt?` doesn't seem to find the `CommandInfo` from the header.

Kha: Ah, that's probably because `withWaitFindSnap` skips the header snapshot. I *think* lifting that restriction should be fine.

PR #931, opened 2022-01-07
Make `set_option` completion replace typed partial name

larsk21:
#### Previous behaviour:
With
```
set_option trace.
```
accepting `trace.Meta.Tactic.simp` would result in
```
set_option trace.trace.Meta.Tactic.simp
```


larsk21:
The Linux check fails after trying to download some NixOS package (`500 - interval server error`) 🤔

PR #932, opened 2022-01-07
Fix duplicate in fakeDepRoot

Anderssorby:
This fixes an issue with diamond dependencies in nix fakeDepRoot.

### DIFF
--- a/nix/buildLeanPackage.nix
+++ b/nix/buildLeanPackage.nix
@@ -76,7 +76,7 @@ with builtins; let
   fakeDepRoot = runBareCommandLocal "${name}-dep-root" {} ''
     mkdir $out
     cd $out
-    mkdir ${lib.concatStringsSep " " ([name] ++ (map (d: d.name) allExternalDeps))}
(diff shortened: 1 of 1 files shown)


Kha:
So you do have two different versions of the same module? As I said on Zulip, that will be problematic no matter what. You really should avoid it.

Anderssorby:
This is a different, but similar issue. It fixes that different dependencies may depend on the same dependency.

Kha:
If it is literally the same dependency, I don't think this should happen. `allExternalDeps` itself is already `lib.unique`'d.

Anderssorby:
Yes, but for some reason it complains that `Init` and `Lean` already exists.

Kha:
Then you'll need to take a closer look at what's up with your `allExternalDeps`.
```
nix-repl> lib.unique [ hello hello ]
[ «derivation /nix/store/8dr2lhv3kriyw8qhg66dszr0vl3grxya-hello-2.10.drv» ]
```

Anderssorby:
@Kha It is caused by compiling with different lean versions. This should not happen however since each flake dependency's lean input follows the lean of it's parent.

Anderssorby:
I'm closing this since this is not the right solution to the problem.

PR #936, opened 2022-01-09
fix: race condition in RPC request handler

gebner:
@julian noticed that he non-determinstically gets an "RPC reference is not valid" error after removing a delay in lean.nvim: https://github.com/Julian/lean.nvim/issues/223#issuecomment-1008082188

It seems plausible that this is related to a race condition in the RPC call handler, which does not atomically update the RPC reference map.  This PR changes all modifications of the RPC reference map to use `Ref.modify` or `Ref.modifyGet`, which are atomic (albeit with busy-waiting, we should probably fix this at some point).


gebner:
cc @Vtec234

Julian:
I can confirm that with `Lean (version 4.0.0, commit 9caf4d6c3978, Release)` the issue seems gone (from ~10 times trying it).

Kha:
I believe this might be our first race condition bug (outside of the runtime)?

Vtec234:
Argh! I will accept the dubious honour, and thanks for fixing this.

PR #937, opened 2022-01-10
Go-to-definition for builtins

Kha:
Apparently I only tested docstring hovers with #785 because there was one missing ingredient for go-to-definition: we have to record the module names as well.

### DIFF
--- a/src/Lean/MonadEnv.lean
+++ b/src/Lean/MonadEnv.lean
@@ -151,10 +151,16 @@ builtin_initialize builtinModuleMap : IO.Ref (NameMap Name) ← IO.mkRef {}
 def addToBuiltinModuleMap (declName modName : Name) : IO Unit :=
   builtinModuleMap.modify (·.insert declName modName)
 
-def findModuleOf? [Monad m] [MonadEnv m] [MonadError m] [MonadLift IO m] (declName : Name) : m (Option Name) := do
+/--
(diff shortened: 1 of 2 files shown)


src/Lean/MonadEnv.lean
```diff
@@ -145,7 +145,19 @@ unsafe def evalConst [Monad m] [MonadEnv m] [MonadError m] [MonadOptions m] (α)
 unsafe def evalConstCheck [Monad m] [MonadEnv m] [MonadError m] [MonadOptions m] (α) (typeName : Name) (constName : Name) : m α := do
   ofExcept <| (← getEnv).evalConstCheck α (← getOptions) typeName constName
 
-def findModuleOf? [Monad m] [MonadEnv m] [MonadError m] (declName : Name) : m (Option Name) := do
+-- map built-in parsers/elaborators to module names even if not imported
+builtin_initialize builtinModuleMap : IO.Ref (NameMap Name) ← IO.mkRef {}
```
Kha, line 149: @leodemoura Does the solution make sense to you? If yes, I will rebase & merge.

Kha:
All things considered, go-to-definition is probably not nearly as important as docstring hover for builtins, so I'm closing this for now. Do object though if you think otherwise.

Eventually, I'd of course like to fix this by outright removing the need for builtins outside of bootstrapping.

PR #938, opened 2022-01-10
CI: try to improve lsan output

Kha:


Kha:
This run didn't fail of course, let's see when the next one does.

PR #942, opened 2022-01-11
feat: create lsp workspaces automatically

cipher1024:
LSP assumes that git repo = workspace. This feature gets LSP to identify `lakefile.lean` as the root of a workspace.

### DIFF
--- a/lean4-mode/lean4-mode.el
+++ b/lean4-mode/lean4-mode.el
@@ -178,7 +178,8 @@ enabled and disabled respectively.")
 (defun lean4-create-lsp-workspace ()
   "This will allow us to use emacs when a repo contains
 multiple lean packages"
-  (when-let ((root (vc-find-root "." "lakefile.lean")))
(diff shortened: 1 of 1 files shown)


lean4-mode/lean4-mode.el
```diff
@@ -172,7 +172,14 @@ enabled and disabled respectively.")
...
   (setq-local flycheck-disabled-checkers '())
   ;; Lean massively benefits from semantic tokens, so change default to enabled
-  (setq-local lsp-semantic-tokens-enable t))
+  (setq-local lsp-semantic-tokens-enable t)
+  (lean4-create-lsp-workspace))
+
+(defun lean4-create-lsp-workspace ()
+  "This will allow us to use emacs when a repo contains
+multiple lean packages"
+  (when-let ((root (vc-find-root "." "lakefile.lean")))
```
Kha, line 181: The `"."` looks suspicious to me, why not `(buffer-file-name)`?

PR #943, opened 2022-01-11
feat: auto-select lean 3 mode

cipher1024:
Some people split their time between Lean 3 and Lean 4. This tweak uses `elan` to check which version of Lean is used for a particular file and select the appropriate emacs mode.

### DIFF
--- a/lean4-mode/lean4-mode.el
+++ b/lean4-mode/lean4-mode.el
@@ -210,7 +210,8 @@ Invokes `lean4-mode-hook'.
 
 (defun lean--version ()
   (let ((version-line
-         (car (process-lines (lean4-get-executable "lean") "-v"))))
+         (car (last (process-lines (lean4-get-executable "lean")
+                                   "-v")))))
     (setq version-line (string-remove-prefix "Lean (version " version-line))
     (setq version-line (split-string version-line (rx (or "." " " ","))))
     (-take 3 version-line)))


lean4-mode/lean4-mode.el
```diff
@@ -208,9 +208,30 @@ Invokes `lean4-mode-hook'.
 ;;;###autoload
 (push '("\\.lean$" . lean4-mode) auto-mode-alist)
 
+(defun lean--version ()
+  (let ((version-line (car (process-lines "lean" "-v"))))
```
Kha, line 212: Please use `lean4-get-executable`. Also, have you tested this when elan first has to download Lean and prints its own output first?

lean4-mode/lean4-mode.el
```diff
@@ -208,9 +208,30 @@ Invokes `lean4-mode-hook'.
...
+  (message "Lean %s" (mapconcat 'identity (lean--version) ".")))
+(defalias 'lean4-show-version 'lean-show-version)
+
+(defun lean-select-mode ()
+  (if lean4-autodetect-lean3
+      (let ((version (lean--version)))
+        (cond ((equal (car version) "4") (lean4-mode))
+              ((equal (car version) "3") (lean-mode))))
+    (lean4-mode)))
+(defalias 'lean4-select-mode 'lean4-select-mode)
```
Kha, line 229: Typo? Why are these necessary anyway?

lean4-mode/lean4-mode.el
```diff
@@ -208,9 +208,30 @@ Invokes `lean4-mode-hook'.
 ;;;###autoload
 (push '("\\.lean$" . lean4-mode) auto-mode-alist)
```
Kha, line 209: Uh, don't you want to change this one too? How did you test this patch?

Kha (changes requested):

lean4-mode/lean4-mode.el
cipher1024: I hadn't thought of that. Is there a way of asking if elan if the required lean version needs to be downloaded?

lean4-mode/lean4-mode.el
cipher1024: They are mostly cosmetic. I just think that this feature conceptually belongs to more name spaces. They're not very important and I can remove them.

lean4-mode/lean4-mode.el
cipher1024: Good catch, thanks!

It is indeed poorly tested, I apologize. It is a version of something I have in my `init.el` file. In general, I don't have a systematic approach to testing my elisp code and any advice would be welcome.

lean4-mode/lean4-mode.el
Kha: I think it's sufficient to just parse the last line instead

lean4-mode/lean4-mode.el
Kha: I believe Soonho was the last person to know anything about unit-testing Elisp... :)

PR #949, opened 2022-01-13
Simple, opt-in precompilation scheme

Kha:
Extends the Nix-based build system with two parameters for controlling *precompilation*, by which I mean the automatic compilation of Lean modules to native code, which is then loaded whenever the module is imported in order to speed up execution.
https://github.com/leanprover/lean4/blob/26c3986a882da5a64144a92ab556466fece6ee73/nix/buildLeanPackage.nix#L17-L21

The second option ensures that we cap the number of maximum loaded libraries to (#modules in package + #transitive package dependencies):
```
$ ldd /nix/store/9d2yw0zgxx8ybr881mhhb6l3xcmz8ccl-Lean.Widget.so/Lean.Widget.so
	linux-vdso.so.1 (0x00007ffe4d763000)
	/nix/store/678nv2qy09767rsfmj0dkzad3jn10wqp-Init.so/Init.so (0x00007fd90cbd9000)
	/nix/store/i7kpq8javlwhd1qb5p9dsg3a8g9kcji3-Lean.Widget.InteractiveCode.so/Lean.Widget.InteractiveCode.so (0x00007fd90cbc5000)
	/nix/store/5d83q1nz3ic0hpkikxb1ghvb5bwkq6n2-Lean.Widget.InteractiveDiagnostic.so/Lean.Widget.InteractiveDiagnostic.so (0x00007fd90cb6d000)
	/nix/store/viqs1wnsqn05vi5b45xdxs2bsmmhfgy6-Lean.Widget.InteractiveGoal.so/Lean.Widget.InteractiveGoal.so (0x00007fd90caed000)
	/nix/store/789b0vkdqy7afiawm8dzqia4qqr17zy6-Lean.Widget.TaggedText.so/Lean.Widget.TaggedText.so (0x00007fd90cac6000)
...
	/nix/store/7gq4cvmv70yjap7507ksm93ihs6hmyfi-Std.so/Std.so (0x00007fd90be41000)
...
```
Monorepos such as mathlib might want to enable only the first option such that importing a part of it from another package does not lead to all of mathlib being loaded.

I decided to start with an implementation in Nix since I assumed it would be easier to do for me than extending Lake, and we can use it to benchmark the scheme [on the stdlib](https://github.com/Kha/lean4/commit/012530a603810d316d1da1e390947d317f46af6b):

```
$ nixprof diff e31fa4af012ddf573c1b21af47def4f430359b4a.json.1 02d0d58a1f477d832b8915a76628ee68724c4040.json.1
  diff [s]           drv
----------  -------  ---------------------------------------------------------
  +0.159    +100.0%  Lean.so
  +0.0882   +100.0%  Init.Core.so
  +0.0868   +100.0%  Lean.Compiler.IR.Basic.so
  -0.0821     -2.5%  Lean.Meta.Basic
  +0.0817   +376.1%  Lean.Elab.PreDefinition.WF-depRoot
  -0.0789    -33.9%  Lean.Elab.PreDefinition.Structural.Basic
  +0.065    +100.0%  Lean.Parser.Syntax.so
  -0.0573     -1.0%  Lean.Elab.Do
  -0.057      -1.3%  Lean.Elab.Term
  +0.054    +100.0%  Lean.Elab.so
  +0.0536   +100.0%  Init.so
  +0.0513   +100.0%  Lean.Meta.so
  +0.0494   +100.0%  Lean.Elab.Term.so
  +0.0492   +100.0%  Lean.Elab.Structure.so
  -0.0485     -4.7%  Lean.Compiler.IR.Basic
  +0.0484   +100.0%  Lean.Server.FileWorker.so
  +0.0482   +100.0%  Lean.Parser.Term.so
  +0.0481   +100.0%  Lean.Elab.Inductive.so
  +0.0478   +100.0%  Lean.Elab.Tactic.so
  +0.0477   +100.0%  Lean.Elab.MutualDef.so
...
 +20.7        +8.2%  total
```

Given that the amount of *compilation* is unchanged (and ccached as usual), the entire overhead is from the new linking steps, which using lld is around 40ms for most files - the few elevated measurings can be considered noise as they are not reproducible when run in isolation. All other differences should be considered noise as well - in particular, there is no actual speedup from precompilation in case of the stdlib because the used `lean` binary (from stage 0) usually contains all those symbols anyway. There is no noticeable overhead from the additional `--load-dynlib`s.

### DIFF
--- a/nix/bootstrap.nix
+++ b/nix/bootstrap.nix
@@ -89,17 +89,19 @@ rec {
         src = ../src;
         fullSrc = ../.;
         inherit debug;
-        builtin = true;
       } // args);
+      Init' = build { name = "Init"; deps = []; };
+      Std'  = build { name = "Std";  deps = [ Init' ]; };
+      Lean' = build { name = "Lean"; deps = [ Init' Std' ]; };
       attachSharedLib = sharedLib: pkg: pkg // {
         inherit sharedLib;
-        mods = mapAttrs (_: m: m // { inherit sharedLib; }) pkg.mods;
+        mods = mapAttrs (_: m: m // { inherit sharedLib; propagatedLoadDynlibs = []; }) pkg.mods;
       };
     in (all: all // all.lean) rec {
       inherit (Lean) emacs-dev emacs-package vscode-dev vscode-package;
-      Init = attachSharedLib leanshared (build { name = "Init"; deps = []; });
-      Std  = attachSharedLib leanshared (build { name = "Std";  deps = [ Init ]; });
-      Lean = attachSharedLib leanshared (build { name = "Lean"; deps = [ Init Std ]; });
+      Init = attachSharedLib leanshared Init';
+      Std  = attachSharedLib leanshared Std'  // { allExternalDeps = [ Init ]; };
+      Lean = attachSharedLib leanshared Lean' // { allExternalDeps = [ Init Std ]; };
       stdlib = [ Init Std Lean ];
       iTree = symlinkJoin { name = "ileans"; paths = map (l: l.iTree) stdlib; };
       Leanpkg = build { name = "Leanpkg"; deps = stdlib; linkFlags = ["-L${gmp}/lib -L${leanshared}"]; };
--- a/nix/buildLeanPackage.nix
+++ b/nix/buildLeanPackage.nix
@@ -21,8 +21,6 @@ lib.makeOverridable (
   precompilePackage ? precompileModules,
   # Lean plugin dependencies. Each derivation `plugin` should contain a plugin library at path `${plugin}/${plugin.name}`.
   pluginDeps ? [],
-  # set only when compiling the stdlib
-  builtin ? false,
   debug ? false, leanFlags ? [], leancFlags ? [], linkFlags ? [], executableName ? lib.toLower name,
   srcTarget ? "..#stage0", srcArgs ? "(\${args[*]})", lean-final ? lean-final' }:
 with builtins; let
@@ -86,8 +84,7 @@ with builtins; let
   pathOfSharedLib = dep: dep.libPath or "${dep}/${dep.libName or dep.name}";
 
   leanPluginFlags = lib.concatStringsSep " " (map (dep: "--plugin=${pathOfSharedLib dep}") pluginDeps);
-  pkgLeanLoadDynlibPaths = map pathOfSharedLib allNativeSharedLibs;
-  loadDynlibPathsOfDeps = deps: concatMap (d: lib.optional (!builtin && d ? sharedLib) (pathOfSharedLib d.sharedLib)) deps;
+  loadDynlibsOfDeps = deps: lib.unique (concatMap (d: d.propagatedLoadDynlibs) deps);
 
   fakeDepRoot = runBareCommandLocal "${name}-dep-root" {} ''
     mkdir $out
@@ -122,7 +119,7 @@ with builtins; let
     ileanPath = relpath + ".ilean";
     cPath = relpath + ".c";
     inherit leanFlags leanPluginFlags;
-    leanLoadDynlibFlags = map (p: "--load-dynlib=${p}") (pkgLeanLoadDynlibPaths ++ loadDynlibPathsOfDeps deps);
+    leanLoadDynlibFlags = map (p: "--load-dynlib=${pathOfSharedLib p}") (loadDynlibsOfDeps deps);
     buildCommand = ''
       dir=$(dirname $relpath)
       mkdir -p $dir $out/$dir $ilean/$dir $c/$dir
@@ -131,6 +128,7 @@ with builtins; let
     '';
   } // {
     inherit deps;
+    propagatedLoadDynlibs = loadDynlibsOfDeps deps;
   };
   compileMod = mod: drv: mkBareDerivation {
     name = "${mod}-cc";
@@ -146,11 +144,16 @@ with builtins; let
       leanc -c -o $out/$oPath $leancFlags -fPIC ${if debug then "${drv.c}/${drv.cPath} -g" else "src.c -O3 -DNDEBUG"}
     '';
   };
-  precompileMod = mod: obj: deps: mkSharedLib mod "${obj}/${obj.oPath} ${lib.concatStringsSep " " (map (d: "${d.sharedLib}/*") deps)}";
   mkMod = mod: deps:
     let drv = buildMod mod deps;
-        obj = compileMod mod drv; in
-      drv // { inherit obj; } // lib.optionalAttrs precompileModules { sharedLib = precompileMod mod obj deps; };
+        obj = compileMod mod drv;
+        # this attribute will only be used if any dependent module is precompiled
+        sharedLib = mkSharedLib mod "${obj}/${obj.oPath} ${lib.concatStringsSep " " (map (d: pathOfSharedLib d.sharedLib) deps)}";
+    in drv // {
+      inherit obj sharedLib;
+    } // lib.optionalAttrs precompileModules {
+      propagatedLoadDynlibs = [sharedLib];
+    };
   externalModMap = lib.foldr (dep: depMap: depMap // dep.mods) {} allExternalDeps;
   # Recursively build `mod` and its dependencies. `modMap` maps module names to
   # `{ deps, drv }` pairs of a derivation and its transitive dependencies (as a nested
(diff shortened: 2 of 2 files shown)


Kha:
In fact, it turns out that ~75% of the 40ms "linking time" are spent in Nixpkgs wrappers around clang and lld (and a good chunk of the remaining time in the dynamic linker loading the actual clang & lld) :) ... so the overhead in a Lake implementation with elan not taken from Nixpkgs should be even less.

gebner:
This seems like a practical scheme for tactics in mathlib.  It's also super useful in modules with FFI, because you don't need to make a separate package for the functions to work.

1. Have you measured the overhead for *loading* the thousands of shared libraries?
2. This doesn't work in the server yet, right?
3. Does this have any impact on using builtinNotation etc. from imported modules?

Kha:
> Have you measured the overhead for _loading_ the thousands of shared libraries?

Filtering out everything but the `lean`-invoking derivations, the overhead is 2.7s or 1.1% of the total build time. For the `Lean` module, i.e. the worst case, it seems to be about 30ms.

> This doesn't work in the server yet, right?

Uh, right! Sounds like a job for `lake serve`.

> Does this have any impact on using builtinNotation etc. from imported modules?

No, `--load-dynlib` does not change any semantics (modulo toolchain bugs...). Specifically, it does not run module initializers.

Kha:
> modulo toolchain bugs...

Yeah, about that... the interpreter assumes that all data symbols (i.e. compiled nullary functions) are already initialized, since it has no idea which are from `--load-dynlib`. We'll need to fix that or use `--plugin`.

gebner:
> For the Lean module, i.e. the worst case, it seems to be about 30ms.

That's great news, I believe the current import time is already 200ms.  A 15% increase doesn't sound so bad.

> Sounds like a job for lake serve a new extension to lake print-paths.

Either way looks good.  I'm beginning to think it would be a good idea to move the build imports logic to `lake serve` anyhow (and make `lake print-paths` a no-op).  The current set-up is confusing if you open multiple files at once: multiple lake build process get started simultaneously, the build status is reported in only one of the files, and only as a diagnostic showing the last line.  Ideally, the dependencies would be built inside `lake serve` with progress reports via `$/progress` (or a custom notification).

tydeu:
I'm fully in support of this approach!  🎉  (Per-module shared libraries where even already on my TODO list.)

> Either way looks good. I'm beginning to think it would be a good idea to move the build imports logic to `lake serve` anyhow (and make `lake print-paths` a no-op).

I also like this idea. One way to do this would be to embedded the watchdog into Lake and have it start up the file worker only after the imports are parsed and built. Doing this within Lake would then allow the builds to be scheduled together (preventing duplicate builds of imports) and allow the watchdog better insight into the Lake workspace configuration, which could be useful for other interactivity features.

Vtec234:
> One way to do this would be to embedded the watchdog into Lake and have it start up the file worker only after the imports are parsed and their modules built.

I support the idea of better integration with Lake, but I'm not sure the best way to do so would be coupling the codebases together more (assuming that is what you mean by 'embed'). I think we should better use something like IPC with JSON packets instead, and if anything eventually move the server into a package separate from core Lean and Lake.

tydeu:
> eventually move the server into a package separate from core Lean and Lake.

I think this would be a good long term plan. It could also involve abstracting certain components so it can also be used as a library of sorts.

> I support the idea of better integration with Lake, but I'm not sure the best way to do so would be coupling the codebases together more (assuming that is what you mean by 'embed'). 

By 'embed', I mean invoking the Lean watchdog code directly from within in Lake (and not migrate the server code into the Lake code base). I think this would be a richer alternative to IPC as it could retain functional guarantees about the codebase (e.g., keep the code monadic, rather than require IO-based communication)  eliminate the overhead of IPC packets, potentially reduce code duplication, and allow for tighter integration (e.g., logging, configuration, etc.).

Kha:
With #959, this now works fine for mathlib4. The overhead is similar, though surprisingly there are also some speedups even in the absence of complicated tactics:
```
  diff [s]           drv
----------  -------  ----------------------------------------
  -0.0976    -12.4%  Mathlib.Tactic.Lint.Simp
  -0.0882     -3.6%  Mathlib.Data.List.Basic
  -0.0812     -6.6%  Mathlib.Init.Data.Int.Order
  +0.0745   +100.0%  Mathlib.Mathport.Syntax.so
  +0.0697   +100.0%  Mathlib.Data.Prod.so
  +0.069    +100.0%  Mathlib.Init.Data.List.Lemmas.so
  +0.069    +100.0%  Mathlib.Data.UInt.so
  +0.0687   +100.0%  Mathlib.Data.ByteArray.so
  +0.0686   +100.0%  Mathlib.Data.Fin.Basic.so
  +0.0684   +100.0%  Mathlib.Tactic.NormNum.so
  +0.0682   +100.0%  Mathlib.Tactic.Ring.so
  +0.0681   +100.0%  Mathlib.Data.Subtype.so
  +0.068    +100.0%  Mathlib.Data.Option.Basic.so
  +0.068    +100.0%  Mathlib.Tactic.Spread.so
  +0.068    +100.0%  Mathlib.Data.List.Card.so
  +0.0679   +100.0%  Mathlib.Data.Option.Defs.so
  +0.0679   +100.0%  Mathlib.Logic.Function.Basic.so
  +0.0679   +100.0%  Mathlib.Algebra.Group.Basic.so
  +0.0678   +100.0%  Mathlib.Tactic.LibrarySearch.so
  +0.0677   +100.0%  Mathlib.Init.Set.so
  +0.0677   +100.0%  Mathlib.Data.List.Perm.so
  +0.0676   +100.0%  Mathlib.Tactic.Lint.Simp.so
  +0.0676   +100.0%  Mathlib.Data.List.Basic.so
  +0.0676   +100.0%  Mathlib.Algebra.Group.Defs.so
  +0.0675   +100.0%  Mathlib.Data.String.Lemmas.so
  +0.0675   +100.0%  Mathlib.Tactic.RunCmd.so
  +0.0674   +100.0%  Mathlib.Init.Data.List.Basic.so
...
  +7.84      +16.5%  total
```
For ` Mathlib.Tactic.Lint.Simp`, the speedup seems to come from `open private`'s `declsInModuleIdx` loop

Kha:
Note that this is with a warm ccache, otherwise there are another ~20% overhead from clang. While we don't ship ccache with Lean, I believe Lake's hash traces should already prevent recompilation if the .c file did not change?

Kha:
@tydeu I assume an implementation in Lake is not quite straightforward given the necessary mutual recursion between .olean, .o, and .so targets, so I should probably leave this part to you. I'm happy to assist though in any way I can.

Kha:
> > Sounds like a job for lake serve a new extension to lake print-paths.
> 
> Either way looks good. I'm beginning to think it would be a good idea to move the build imports logic to `lake serve` anyhow (and make `lake print-paths` a no-op).

That sounds like a nice improvement for the future, but in order to finish this part, I went with the minimal change of adding a new `LeanPaths` field for now.

src/Lean/Util/Paths.lean
```diff
@@ -14,8 +14,9 @@ namespace Lean
 open System
 
 structure LeanPaths where
-  oleanPath : SearchPath
-  srcPath   : SearchPath
+  oleanPath       : SearchPath
+  srcPath         : SearchPath
+  loadDynlibPaths : Array FilePath := #[]
```
tydeu, line 19: Why not `loadDynlibPath : SearchPath := []` for consistency?

Kha: That doesn't seem semantically appropriate, it's not a search path in any way. We simply load all of them.

tydeu: Ah, true, makes sense.

Kha:
People interested in this topic might also be interested in the ongoing discussion in https://github.com/leanprover/lake/pull/47. Input welcome.

Kha:
Okay, after one more revision I think I have a consistent story around dependencies:
* Instead of only checking direct dependencies for precompiled libraries, we now propagate the libraries properly, resetting the propagated set at each module with activated precompilation. Thus we still keep the cmdline as small as possible (the Pareto frontier of libraries if you will) but always load precompiled packages even if they are only indirect dependencies.
* If a package is not precompiled but a dependent package/module wants to be, we link it against the shared library of the whole package (as if `precompilePackage` was set). It would be too onerous to make the user override the precompilation settings for all dependencies.

PR #956, opened 2022-01-17
fix: provide reference implementation for `Array.modify`

gebner:
This got (presumably accidentally) removed during the Inhabited refactor.  Since I was at it, I've also removed the Inhabited arguments of modifyM and modifyOp.

### DIFF
--- a/src/Init/Data/Array/Basic.lean
+++ b/src/Init/Data/Array/Basic.lean
@@ -106,31 +106,35 @@ def shrink (a : Array α) (n : Nat) : Array α :=
   loop (a.size - n) a
 
 @[inline]
-def modifyM [Monad m] [Inhabited α] (a : Array α) (i : Nat) (f : α → m α) : m (Array α) := do
+unsafe def modifyMUnsafe [Monad m] (a : Array α) (i : Nat) (f : α → m α) : m (Array α) := do
   if h : i < a.size then
     let idx : Fin a.size := ⟨i, h⟩
     let v                := a.get idx
-    let a'               := a.set idx default
+    -- Replace a[i] by `box(0)`.  This ensures that `v` remains unshared if possible.
+    -- Note: we assume that arrays have a uniform representation irrespective
+    -- of the element type, and that it is valid to store `box(0)` in any array.
+    let a'               := a.set idx (unsafeCast ())
     let v ← f v
     pure <| a'.set (size_set a .. ▸ idx) v
   else
     pure a
 
-@[inline]
-unsafe def modifyUnsafe (a : Array α) (i : Nat) (f : α → α) : Array α :=
+@[implementedBy modifyMUnsafe]
+def modifyM [Monad m] (a : Array α) (i : Nat) (f : α → m α) : m (Array α) := do
   if h : i < a.size then
-    let idx : Fin a.size := ⟨i, h⟩
-    let v                := a.get idx
-    let a'               := a.set idx (unsafeCast ())
-    a'.set (size_set a .. ▸ idx) (f v)
+    let idx := ⟨i, h⟩
+    let v   := a.get idx
+    let v ← f v
+    pure <| a.set idx v
   else
-    a
-
-@[implementedBy modifyUnsafe]
-constant modify (a : Array α) (i : Nat) (f : α → α) : Array α
+    pure a
 
 @[inline]
-def modifyOp [Inhabited α] (self : Array α) (idx : Nat) (f : α → α) : Array α :=
+def modify (a : Array α) (i : Nat) (f : α → α) : Array α :=
+  Id.run <| modifyM a i f
+
+@[inline]
+def modifyOp (self : Array α) (idx : Nat) (f : α → α) : Array α :=
   self.modify idx f
 
 /-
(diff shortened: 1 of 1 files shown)


src/Init/Data/Array/Basic.lean
```diff
@@ -106,31 +106,32 @@ def shrink (a : Array α) (n : Nat) : Array α :=
...
 
 @[inline]
-def modifyM [Monad m] [Inhabited α] (a : Array α) (i : Nat) (f : α → m α) : m (Array α) := do
+unsafe def modifyMUnsafe [Monad m] (a : Array α) (i : Nat) (f : α → m α) : m (Array α) := do
   if h : i < a.size then
     let idx : Fin a.size := ⟨i, h⟩
     let v                := a.get idx
-    let a'               := a.set idx default
+    let a'               := a.set idx (unsafeCast ())
     let v ← f v
```
leodemoura, line 114: The unsafe `modifyM` may leave the array in an inconsistent state if `f` throws an exception.

gebner: Wouldn't the array just be discarded in that case?  The `a'` never escapes the `modifyM` function.

Note that the same situation also occurs in `mapMUnsafe`.

leodemoura: The situation is different at `mapUnsafe`,  we are storing a valid element in the array in the `map`.
Otherwise, I agree with you the value `a'` should not escape. If an exception is thrown, and the array is used in the "catch" block, then a destructive update will not be performed.
My concern is that `box(0)` may not be a valid element of the type, but this is not an issue for our current runtime. We should add a comment saying that the correctness of the code relies on the fact that the runtime uses a "generic" deletion procedure and it doesn't matter whether the array contains invalid elements or not.

gebner: > The situation is different at mapUnsafe, we are storing a valid element in the array in the map.

It is only superficially different since we do `(unsafeCast as : Array NonScalar)` at the beginning of `mapMUnsafe`.  After type erasure they both do the same thing and write `box(0)` to the array.  I can add a comment to both of the functions if you want.

> My concern is that box(0) may not be a valid element of the type, but this is not an issue for our current runtime. We should add a comment saying that the correctness of the code relies on the fact that the runtime uses a "generic" deletion procedure and it doesn't matter whether the array contains invalid elements or not.

Are you thinking of an implementation where `Array` can have different runtime representations depending on how it was constructed? (Like we had e.g. for `list expr` in Lean 3.)

I'm not sure if that would ever be a convenient API.  We'd presumably need to add a typeclass argument to `mkArray` that determines the runtime representation.  And if you forget the argument somewhere then you could easily end up with a non-specialized array.  Also all array access would need to branch on the representation, negating the potential performance advantage.

leodemoura: > It is only superficially different since we do (unsafeCast as : Array NonScalar) at the beginning of mapMUnsafe. After type erasure they both do the same thing and write box(0) to the array. I can add a comment to both of the functions if you want.

You are correct. I would appreciate it if could add a comment to both functions.

leodemoura: > Are you thinking of an implementation where Array can have different runtime representations depending on how it was constructed? (Like we had e.g. for list expr in Lean 3.)

No, we were considering generating type-specific deletion code. For example, for types such as `Prod Bool UInt16`. A specific deletion procedure would not have to execute 
```c
        object ** it  = lean_ctor_obj_cptr(o);
        object ** end = it + lean_ctor_num_objs(o);
        for (; it != end; ++it) dec(*it, todo);
        lean_free_small_object(o);
```
as it does now.

gebner: > No, we were considering generating type-specific deletion code.

I see, this looks like it could apply in a lot of cases (though I'll wait for the benchmarking results before getting excited).

This is interesting.  If we make use of type-specific information in the compiler, then this means that `(unsafeCast x : α)` is only safe if the runtime representation of `x` is a valid value for `α`. Therefore:
 - `(unsafeCast () : Nat)` is safe
 - `(unsafeCast () : Nat × Nat)` is unsafe (could crash with specialized deletion code)
 - `(unsafeCast "" : Nat × Nat)` is unsafe (could also crash)
 - `(unsafeCast (1, 2) : UInt32)` is unsafe (could leak)

In particular, the current implementation of `mapM` would remain safe but `modifyM` could become unsafe (and should then be reimplemented by first casting to `Array NonScalar`).

Kha: So there would basically be a subtyping relation of runtime types, along which it is safe to cast (as in languages with actual subtyping). Interesting.

gebner: Right, casting along the subtype relation is safe.  However we also need to cast in cases where the subtype relation is not satisfied, such as e.g. `(unsafeCast (unsafeCast (as : Array Bool) : Array NonScalar) : Array Bool)` (which is what happens in `mapMUnsafe`).

Note that `unsafeCast` could always lead to crashes.  However so far you could only get a crash by performing an operation on the casted value (e.g. accessing a constructor field, etc.).  What the type-specific optimization changes is that already `unsafeCast` itself is unsafe.  The mere presence of a runtime value that is invalid for the claimed type can cause crashes.

PR #958, opened 2022-01-17
chore: show declaration with sorry in `#eval`

gebner:
Tomas complained on Zulip that it is hard to figure out which sorry makes `#eval` fail.  If you have many sorrys in proofs, then it becomes hard to find the `sorry` in a definition which prevents evaluation.

This PR adds the name of the declaration containing sorry to the error message from `#eval`, and should make this search much easier.  In the case of `#eval (sorry : Nat)`, the error message shows `_eval._lambda_1`, which is not ideal but probably still clear enough for users to understand

### DIFF
--- a/src/library/compiler/ir_interpreter.cpp
+++ b/src/library/compiler/ir_interpreter.cpp
@@ -1047,8 +1047,9 @@ optional<name> get_sorry_dep(environment const & env, name const & n) {
 }
 
 object * run_boxed(environment const & env, options const & opts, name const & fn, unsigned n, object **args) {
-    if (get_sorry_dep(env, fn)) {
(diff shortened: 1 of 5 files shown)


gebner:
Unfortunately this doesn't help with compiled code.

Kha:
> Unfortunately this doesn't help with compiled code.

That indeed is a bit unfortunate especially in light of #949. I would prefer adding the current declaration as a parameter to `sorryAx`, much like `panic!`. A more general system along the lines of [`HasCallstack`](https://hackage.haskell.org/package/base-4.16.0.0/docs/GHC-Stack.html#g:2) would also be interesting. Having very generic helper functions trigger panics is a common annoyance for me.

PR #962, opened 2022-01-19
Make code blocks render properly

kl2792:


Kha:
See https://github.com/leanprover/lean4/pull/804#issuecomment-994926260

Issue #960, opened 2022-01-18
Fuzzy completion matching


Kha:
We currently match completions by exact (i.e. even case-sensitive) prefix. On the other hand, Lean 3 used a general fuzzy match algorithm that is much fuzzier than most other completion matches, I believe. clangd for example has a substring-based algorithm that looks more like what I would expect for completion, and is mentioned to be similar to VS Code's client-side matching (which I assume we can't make use of because of the enormous size of the global namespace): https://clang.llvm.org/extra//doxygen/FuzzyMatch_8cpp_source.html

PR #964, opened 2022-01-20
implement LSP workspace symbol request (ctrl+t in vscode)

### LINKED ISSUE #396: ctrl-t support in the Lean server
Lean 3 server implements this feature, and @DanielFabian pointed out it is super useful, and is not supported in the Lean 4 server yet.

Garmelon:
This is a fairly simple implementation using the ilean information that's already present for the references request.

- It uses a searching algorithm that case-insensitively looks for the query characters in the input string, in order but not necessarily consecutively (may return slightly unexpected results with unicode identifiers in some cases). VSCode's highlighting of the query in the results seems to do the same.
- It limits the amount of returned results to a maximum of 100.
- It doesn't try to sort the results in any meaningful order. If you're using this to find a specific symbol, you're probably entering enough information to only get a handful of results in the first place.
- Since ilean files don't contain the necessary information, every result has a `SymbolKind` of  `Constant`.

The searching algorithm is intentionally kept simple. See also #960

The PR also fixes find-references attempting to return references from non-existing .lean files and crashing in the process (by using `findModuleWithExt` instead of `findWithExt`).

Closes #396

### DIFF
--- a/src/Lean/Util/Path.lean
+++ b/src/Lean/Util/Path.lean
@@ -40,8 +40,10 @@ def findWithExt (sp : SearchPath) (ext : String) (mod : Name) : IO (Option FileP
 
 /-- Like `findWithExt`, but ensures the returned path exists. -/
 def findModuleWithExt (sp : SearchPath) (ext : String) (mod : Name) : IO (Option FilePath) := do
-  let root? ← sp.findM? fun p => do (modToFilePath p mod ext).pathExists
-  root?.map (modToFilePath · mod ext)
+  if let some path ← findWithExt sp ext mod then
+    if ← path.pathExists then
+      return some path
+  return none
 
 def findAllWithExt (sp : SearchPath) (ext : String) : IO (Array FilePath) := do
   let mut paths := #[]


src/Lean/Server/References.lean
```diff
@@ -210,6 +210,24 @@ def referringTo (self : References) (ident : RefIdent) (srcSearchPath : SearchPa
           result := result.push ⟨uri, range⟩
   result
 
+def definitionsMatching (self : References) (srcSearchPath : SearchPath) (filter : Name -> Bool)
+    (maxAmount : Option Nat := none) : IO $ Array (Name × Location) := do
+  let mut result := #[]
+  let mut amount := 0
```
Kha, line 216: `amount = result.size`?

src/Lean/Server/References.lean
```diff
@@ -210,6 +210,24 @@ def referringTo (self : References) (ident : RefIdent) (srcSearchPath : SearchPa
...
+  for (module, refs) in self.allRefs.toList do
+    if let some path ← srcSearchPath.findModuleWithExt "lean" module then
+      let uri := DocumentUri.ofPath <| ← IO.FS.realPath path
```
Kha, line 219: Doing hundreds of IO ops no matter the result size seems a bit wasteful, though perhaps we should first measure if it matters compared to the filtering overhead.

src/Lean/Server/References.lean
```diff
@@ -210,6 +210,24 @@ def referringTo (self : References) (ident : RefIdent) (srcSearchPath : SearchPa
...
+      for (ident, info) in refs.toList do
+        if let RefIdent.const name := ident then
+          if let some definition := info.definition then
```
Kha, line 222:
```suggestion
        if let (RefIdent.const name, some definition) := (ident, info.definition) then
```

src/Lean/Server/References.lean
```diff
@@ -210,6 +210,24 @@ def referringTo (self : References) (ident : RefIdent) (srcSearchPath : SearchPa
           result := result.push ⟨uri, range⟩
   result
 
+def definitionsMatching (self : References) (srcSearchPath : SearchPath) (filter : Name -> Bool)
+    (maxAmount : Option Nat := none) : IO $ Array (Name × Location) := do
```
Kha, line 214:
```suggestion
    (maxAmount? : Option Nat := none) : IO $ Array (Name × Location) := do
```

src/Lean/Util/Path.lean
```diff
@@ -30,13 +30,19 @@ abbrev SearchPath := System.SearchPath
...
+`ext` (`lean` or `olean`) corresponding to `mod`. Otherwise, return `none`. Does
+not check whether the returned path exists. -/
 def findWithExt (sp : SearchPath) (ext : String) (mod : Name) : IO (Option FilePath) := do
   let pkg := mod.getRoot.toString
   let root? ← sp.findM? fun p =>
     (p / pkg).isDir <||> ((p / pkg).withExtension ext).pathExists
   return root?.map (modToFilePath · mod ext)
 
+/-- Like `findWithExt`, but ensures the returned path exists. -/
+def findModuleWithExt (sp : SearchPath) (ext : String) (mod : Name) : IO (Option FilePath) := do
```
Kha, line 42: Let's base this one on `findWithExt` so that we're absolutely sure we always select the same package root

src/Lean/Util/Path.lean
```diff
@@ -30,13 +30,19 @@ abbrev SearchPath := System.SearchPath
 namespace SearchPath
 
 /-- If the package of `mod` can be found in `sp`, return the path with extension
-`ext` (`lean` or `olean`) corresponding to `mod`. Otherwise, return `none.` -/
+`ext` (`lean` or `olean`) corresponding to `mod`. Otherwise, return `none`. Does
+not check whether the returned path exists. -/
 def findWithExt (sp : SearchPath) (ext : String) (mod : Name) : IO (Option FilePath) := do
```
Kha, line 35: We should probably rename it to something like `getPathOfModuleFile?` or something... but not in this PR

Kha:
*yesss*
image

Kha:
I pushed a commit that should fix the Unicode issue, and have the matching ignore namespaces because the current greedy algorithm would otherwise often not even get to the declaration name. The results can still be quite unexpected, such as `String` first matching `instFileSourcePlainGoalParams`, but this is a good start.

src/Lean/Server/Watchdog.lean
```diff
@@ -368,8 +368,9 @@ def handleReference (p : ReferenceParams) : ServerM (Array Location) := do
 def handleWorkspaceSymbol (p : WorkspaceSymbolParams) : ServerM (Array SymbolInformation) := do
   let references ← (← read).references.get
   let srcSearchPath := (← read).srcSearchPath
-  let symbols ← references.definitionsMatching srcSearchPath (maxAmount := some 100)
-    fun name => containsCaseInsensitive p.query $ name.toString
+  let symbols ← references.definitionsMatching srcSearchPath (maxAmount := some 100) fun
+    | Name.str _ s _ => containsCaseInsensitive p.query s
+    | _              => false
```
gebner, line 373: Hmm, I always liked to do `ctrl-p # nat.ltlt` to find nat-specific lemmas.

Kha: Yeah, I removed it again, it's not too bad that way. It should get better when we prioritize consecutive matches.

Kha:
I pushed two further refinements that should be less controversial I hope. Now `fWE` works as well.

gebner:
BTW, the matching algorithm is surprisingly good (and fast!) for how simple it is.  I've added it to the mathlib docs (should be live in a few minutes), and it now only takes ~100ms to search through the 111k decls.

Kha:
We're still missing the filter for the current workspace, no? Actually, is `references` supposed to be local to the current workspace as well? It would be good to avoid loading the stdlib .ileans when we don't have to.

Kha:
We should probably look at other implementations, but I would have assumed that `symbol` was meant to be local because it is primarily about *navigation* while code completion is meant to *find anything* that can be used, regardless where it's from. That's not quite true for our current completion because it only lists imported things, but I would definitely like it to auto-import declarations in the future!

Having said that, I'm fine with whatever behavior we deem the most convenient.

gebner:
> Rust-analyzer seems to restrict the symbol search to symbols defined in the current workspace.

That's a non-standard extension: https://github.com/matklad/rust-analyzer/blob/1fd31f7f4cf6c65a407cb36b16a86c814c5783c3/docs/dev/lsp-extensions.md#workspace-symbols-filtering  I couldn't find any justification for the default setting in the introducing PR (https://github.com/rust-analyzer/rust-analyzer/pull/7698).

In RLS e.g. the symbol search includes the dependencies.  The typescript language server also returns symbols from the dependencies.

Kha:
Ooh, I like rust-analyzer's search suffixes. But it looks like the current behavior is fine then.

PR #966, opened 2022-01-20
[WIP] Emscripten WASM support

Anderssorby:
I'm working towards having a new working emscripten build with nix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,7 +23,7 @@ endforeach()
 include(ExternalProject)
 project(LEAN CXX C)
 
-if("${CMAKE_SYSTEM_NAME}" MATCHES "Emscripten")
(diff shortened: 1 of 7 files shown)


Kha:
Looks like the right direction, but I'm dismayed to learn that Emscripten is *not* set up as cross-compilation in Nixpkgs. We should not have Emscripten-specific attributes, and stuff like selecting the correct compiler should be handled by `stdenv`.

Anderssorby:
Well it is quite different from the other targets and needs a lot of special flags and handling so I would say having emscripten-specific attributes is mostly inevitable, at least temporarily. Ideally one could just change the target platform to wasm. I think this is a small price to pay.

Kha:
I'll close this as we discussed a different direction on Zulip: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/wasm.20nix.20emscripten.20build/near/270613397

PR #969, opened 2022-01-21
Extend implementation of `handleDocumentHighlight`

### LINKED ISSUE #957: Finish implementation of "document highlight" request
We already have a handler for `textDocument/documentHighlight`, but it only highlights the corresponding `do` of a `return`, which is a bit awkward as mentioned in https://github.com/leanprover/vscode-lean4/pull/103#issuecomment-1014834778. After #835, we have all the information we need to implement highlighting of identifiers as well.

larsk21:
Uses the "find references" implementation introduced in #925 to provide document highlights. Highlights are shown at least to the point where the highlight was requested, but possibly further if that part of the file is already parsed.

closes #957 

One known bug is that the highlight of the name of a `let rec` function, when requested from inside that function, will span the entire function body. This is probably caused by `findModuleRefs` returning multiple entries for the function name, one with an `fvar` id, one with a `const` name and another one with a `const` name and the postfix `_unsafe_rec`. The first entry contains the correct ranges for definition and usages, while the other two contain only a definition range spanning the entire function. @Garmelon Do you know what causes this?

### DIFF
--- a/src/Lean/Data/Lsp/Basic.lean
+++ b/src/Lean/Data/Lsp/Basic.lean
@@ -42,10 +42,7 @@ instance : LE Position := leOfOrd
 structure Range where
   start : Position
   «end» : Position
-  deriving Inhabited, BEq, Hashable, ToJson, FromJson
-
-instance : Ord Range where
-  compare a b := Ordering.lexOrdering [compare a.start b.start, compare a.end b.end]
+  deriving Inhabited, BEq, Hashable, ToJson, FromJson, Ord
 
 instance : LT Range := ltOfOrd
 instance : LE Range := leOfOrd
--- a/src/Init/Data/Ord.lean
+++ b/src/Init/Data/Ord.lean
@@ -12,14 +12,6 @@ inductive Ordering where
   | lt | eq | gt
 deriving Inhabited, BEq
 
-namespace Ordering
-
-def lexOrdering (orderings : List Ordering) : Ordering :=
-  orderings.foldr (init := Ordering.eq) fun ordering result =>
-    if ordering == Ordering.eq then result else ordering
-
-end Ordering
-
 
 class Ord (α : Type u) where
   compare : α → α → Ordering


Garmelon:
I'd assume it's because of how the desugaring(?) of `let rec` ends up in the info tree? It might be useful to be able to distinguish whether InfoTree nodes have a direct equivalent in the source file or not.

larsk21:
Yeah, I think so too. Would it help to check that the `elaborator` property of the `ElabInfo` is `` `ident``?

Kha:
I'm happy with the code of this PR as is. As @Garmelon siad, the let rec issue seems to be a more general issue with the info tree data.

larsk21:
Alright, I'm going to add some tests and then mark it as ready.

larsk21:
Another similar issue: All identifiers in a structure instance, like `{ foo := 42, foo2:= bar }`, are included twice. Since the duplicates are exactly equal, the highlights are shown as they should, but the interactive test fails due to the duplicate items.

Garmelon:
That sounds odd. Maybe the duplicates are because usages and definitions overlap? That could maybe be fixed in `combineFvars` while there is still a map from ranges to definitions, by checking whether each usage's range already exists as a definition.

Garmelon:
I tried to deduplicate references based on overlapping definitions and usages using this change, but it didn't work.
```diff
diff --git a/src/Lean/Server/References.lean b/src/Lean/Server/References.lean
index 91674febfd..468f814794 100644
--- a/src/Lean/Server/References.lean
+++ b/src/Lean/Server/References.lean
@@ -146,9 +146,22 @@ where
     | m, RefIdent.fvar id => RefIdent.fvar <| m.findD id id
     | _, ident => ident
 
+def dedupReferences (refs : Array Reference) : Array Reference := Id.run do
+  let mut defRanges : HashMap Lsp.Range Unit := HashMap.empty -- HashSet
+  for ref in refs do
+    if ref.isDeclaration then
+      defRanges := defRanges.insert ref.range ()
+
+  let mut result := #[]
+  for ref in refs do
+    if ref.isDeclaration || !(defRanges.contains ref.range) then
+      result := result.push ref
+  result
+
 def findModuleRefs (text : FileMap) (trees : Array InfoTree) (localVars : Bool := true)
     : ModuleRefs := Id.run do
-  let mut refs := combineFvars <| findReferences text trees
+  let mut refs := dedupReferences <| combineFvars <| findReferences text trees
   if !localVars then
     refs := refs.filter fun
       | { ident := RefIdent.fvar _, .. } => false
```

Deduplicating references indiscriminately seems to work, although one should probably ensure that references are not deduplicated across different idents:
```diff
diff --git a/src/Lean/Server/References.lean b/src/Lean/Server/References.lean
index 91674febfd..ceb8813b68 100644
--- a/src/Lean/Server/References.lean
+++ b/src/Lean/Server/References.lean
@@ -146,9 +146,17 @@ where
     | m, RefIdent.fvar id => RefIdent.fvar <| m.findD id id
     | _, ident => ident
 
+def dedupReferences (refs : Array Reference) : Array Reference := Id.run do
+  let mut refsByRange : HashMap Lsp.Range Reference := HashMap.empty
+  for ref in refs do
+      if ref.isDeclaration || !(refsByRange.contains ref.range) then
+        refsByRange := refsByRange.insert ref.range ref
+
+  return refsByRange.fold (init := #[]) fun refs _ ref => refs.push ref
+
 def findModuleRefs (text : FileMap) (trees : Array InfoTree) (localVars : Bool := true)
     : ModuleRefs := Id.run do
-  let mut refs := combineFvars <| findReferences text trees
+  let mut refs := dedupReferences <| combineFvars <| findReferences text trees
   if !localVars then
     refs := refs.filter fun
       | { ident := RefIdent.fvar _, .. } => false
```
I'm blaming the duplicates on the InfoTree :P

larsk21:
> I'm blaming the duplicates on the InfoTree :P

Me too :D 

But your second fix seems to work, although I would sort the deduped references by `range.start` so they have the same order as before (the hash map ruins the order) and to have a deterministic result.

> one should probably ensure that references are not deduplicated across different idents

Where would this be an issue? For structure `mk`s?

larsk21:
> It might not be an issue, but to me it feels like the right thing to do since it relies on less assumptions about other parts of the system and might be more robust in weird situations. Problems like overlapping definitions from different idents are probably best solved at an earlier step.

I agree. Would the following be a generally valid solution that we can use in `findModuleRefs`?

```
def dedupReferences (refs : Array Reference) : Array Reference := Id.run do
  let mut refsByRange : HashMap Lsp.Range (Array Reference) := HashMap.empty
  for ref in refs do
    match refsByRange.find? ref.range with
      | some rangeRefs => if ref.isDeclaration then
        refsByRange := refsByRange.insert ref.range (rangeRefs.push ref)
      | none =>
        refsByRange := refsByRange.insert ref.range #[ref]

  let dedupedRefs := refsByRange.fold (init := #[]) fun refs _ ref => refs.append ref
  return dedupedRefs.qsort (·.range.start < ·.range.start)
```

Garmelon:
Your implementation would still deduplicate two overlapping references from different idents into only one reference. Something like this might work (didn't test it), although it feels a bit overkill.
```lean
def dedupReferences (refs : Array Reference) : Array Reference := Id.run do
  let mut refsByIdAndRange : HashMap RefIdent (HashMap Lsp.Range Reference) := HashMap.empty
  for ref in refs do
    let refsByRange := refsByIdAndRange.findD ref.ident HashMap.empty
    if ref.isDeclaration || !(refsByRange.contains ref.range) then
      refsByIdAndRange := refsByIdAndRange.insert ref.ident <| refsByRange.insert ref.range ref

  let dedupedRefs := refsByIdAndRange.fold (init := #[]) fun result _ refsByRange =>
    refsByRange.fold (init := result) fun result _ ref =>
      result.push ref

  return dedupedRefs.qsort (·.range < ·.range)
```

You could derive `Ord` for `Lsp.Range` and sort by the range directly, instead of only sorting by the start.

By the way,  Github has lean syntax highlighting via \`\`\`lean.

larsk21:
Wouldn't this also work?
```lean
def dedupReferences (refs : Array Reference) : Array Reference := Id.run do
  let mut refsByIdAndRange : HashMap (RefIdent × Lsp.Range) Reference := HashMap.empty
  for ref in refs do
    refsByIdAndRange :=
      refsByIdAndRange.insert (ref.ident, ref.range) <|
      refsByIdAndRange.findD (ref.ident, ref.range) ref

  let dedupedRefs := refsByIdAndRange.fold (init := #[]) fun refs _ ref => refs.push ref
  return dedupedRefs.qsort (·.range < ·.range)
```

Garmelon:
Yes, that looks more elegant. It doesn't prioritize definitions, but that's probably not an issue. Not sure if the insertion every iteration has any negative impacts.

larsk21:
Should I add unit tests for `lexOrdering` and `dedupReferences` and if so, where should I put them? @Kha

src/Lean/Data/Lsp/Basic.lean
```diff
@@ -44,6 +44,12 @@ structure Range where
   «end» : Position
   deriving Inhabited, BEq, Hashable, ToJson, FromJson
 
+instance : Ord Range where
```
Garmelon, line 47: Doesn't `derive Ord` already behave this way? See #914

larsk21: Interesting, I wouldn't have thought that I changed the semantics of a related function when reordering the fields of structure.

Kha:
> Should I add unit tests for `lexOrdering` and `dedupReferences` and if so, where should I put them? @Kha

We don't really have a unit testing setup right now, apart from `#eval`s in an output test. If `dedupReferences` is (partially) covered by the new LSP test, that's sufficient. As for `lexOrdering`, yes, you can rely on structure fields being compared in order.

PR #979, opened 2022-01-28
Incremental ilean updates and go-to-definition fix

Garmelon:
This PR implements incremental ilean updates from the worker to the watchdog while the worker is going through the file.

It also fixes go-to-definition if the definition is in another, modified, open file. Previously, go-to-definition would jump to the definition's position from the last compilation, but now the server checks the ilean info and forwards the request to the watchdog only if no definition was found in the ilean info.

### DIFF
--- a/src/Lean/Server/FileWorker.lean
+++ b/src/Lean/Server/FileWorker.lean
@@ -98,6 +98,8 @@ section Elab
     if lastSnap.isAtEnd then
       publishDiagnostics m lastSnap.diagnostics.toArray ctx.hOut
       publishProgressDone m ctx.hOut
+      -- This will overwrite existing ilean info for the file, in case something
+      -- went wrong during the incremental updates.
       publishIleanInfoFinal m ctx.hOut <| s.snaps.insertAt 0 s.headerSnap
       return none
     publishProgressAtPos m lastSnap.endPos ctx.hOut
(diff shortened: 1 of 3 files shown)


src/Lean/Server/Watchdog.lean
```diff
@@ -514,6 +536,10 @@ section MessageHandling
           code := ErrorCode.internalError
           message := s!"Failed to process request {id}: {e}"
         }
+    if method == "textDocument/definition" || method == "textDocument/declaration" then
```
Kha, line 539: A comment as to why we try to handle this in the watchdog and why we still need the worker sometimes would be helpful here

src/Lean/Server/References.lean
```diff
@@ -210,6 +227,18 @@ def referringTo (self : References) (ident : RefIdent) (srcSearchPath : SearchPa
           result := result.push ⟨uri, range⟩
   result
 
+def definitionOf (self : References) (ident : RefIdent) (srcSearchPath : SearchPath)
```
Kha, line 230: Should be `definitionOf?`, ditto `findDefinition?`

src/Lean/Server/FileWorker.lean
```diff
@@ -113,20 +118,22 @@ section Elab
...
       -- Treat header processing errors as fatal so users aren't swamped with
       -- followup errors
-      let hOut := (←read).hOut
-      publishProgressAtPos m headerSnap.beginPos hOut (kind := LeanFileProgressKind.fatalError)
+      publishProgressAtPos m headerSnap.beginPos ctx.hOut (kind := LeanFileProgressKind.fatalError)
+      publishIleanInfoFinal m ctx.hOut #[headerSnap]
       AsyncList.nil
     else
-      let ctx ← read
+      publishIleanInfoUpdate m ctx.hOut <| snaps.insertAt 0 headerSnap
```
Kha, line 136: This will discard any previous ilean info in the watchdog because the doc version (presumably) increased, right? Could you add a small note in that regard?

PR #981, opened 2022-01-29
feat: inform server if widgets are available

### LINKED ISSUE #977: Trace messages missing in Emacs
```
[Meta.isDefEq] (trace hidden)
  [Meta.isDefEq] (trace hidden)
```
is not terribly helpful. We should show the full trace as before if there is no interactive support.

Vtec234:
Closes #977 . Clients that do provide widgets should add a `hasWidgets: true` field to their `initializationOptions` for the LSP server. /CC @gebner


Kha:
Thanks!

Kha:
@gebner Feel free to complain if I merged this too soon and the editors won't be adapted until the next nightly. We can always revert.

PR #985, opened 2022-02-01
chore: remove leanpkg

Kha:
It had a good run for something written/ported in a single week


tydeu:
The use-defined extensions, attribute, option, etc. tests/examples still seem useful, though, right? They just don't need to use Leanpkg.

Kha:
Yeah, we should rewrite them using Lake

PR #990, opened 2022-02-03
doc: document some do block patterns/sugar.

ammkrn:


doc/do.md
```diff
@@ -348,6 +348,44 @@ TODO: describe `forIn`
...
+#eval showUserInfo (some "JaneDoe") none
+```
+
+## If-let
+
+Inside a `do` block, users can employ the `if let` pattern to destructure actions:
+
+```lean
+def tryLength (input : Option String) : Except String Nat := do
+  if let some x ← input then return x.length else Except.error "No input!"
```
digama0, line 380: It's a bit odd to use `←` here since `input` is not monadic.

ammkrn: You right. I changed it to `IO (Option Nat)`.

Kha: The examples still *look* pretty pure; they should at least prefix monadic parameter names with "get" and use `pure <| some` throughout.

Kha:
Note that `if let` works outside of `do` as well, just not without the `else`

doc/do.md
ammkrn: @Kha 

You might have to clue me in on what the idiom is for `pure <| some`; the examples I used were supposed to mimic this: https://github.com/leanprover/lean4/blob/420f5bb315f7cf9b8ef628c0aea0049b43ae2402/src/Lean/Util/Path.lean#L42. I know there are more meaty examples in Meta/Elab, but they don't seem appropriate for a doc example covering syntax.

Kha: I only meant that you should use `pure <| some` consistently in both example blocks, which in fact should be mandatory now.

doc/expressions.md
```diff
@@ -419,33 +419,28 @@ Every computable definition in Lean is compiled to bytecode at definition time.
...
+    #reduce let x := 5; x + 3
+    #eval   let x := 5; x + 3
 
     def f x := x + 3
 
     #reduce f 5
     #eval   f 5
 
-    #reduce @nat.rec (λ n, Nat) (0 : Nat)
-                     (λ n recval : Nat, recval + n + 1) (5 : Nat)
```
Kha, line 434: What is the motivation for removing this? The `#eval` won't work, but `#reduce` should.

ammkrn: It's been reintroduced.

doc/do.md
Kha: > The examples still _look_ pretty pure; they should at least prefix monadic parameter names with "get"

So, what do you think? What I had in mind was `getInput` etc.

ammkrn: Sounds good.

PR #992, opened 2022-02-03
fix: ensure motive of `matches` is `Bool`

gebner:
Often I'm running into this issue with `matches`:
```lean
#check fun x : Option Nat => decide (x matches some .. : Prop)
-- failed to synthesize instance Decidable ...
```

To be honest, I don't why it even works at all with if-then-else because there the expected type is also Prop.

### DIFF
--- a/src/Init/Notation.lean
+++ b/src/Init/Notation.lean
@@ -203,7 +203,7 @@ macro_rules
     else
       `(%[ $elems,* | List.nil ])
 
-notation:50 e:51 " matches " p:51 => match e with | p => true | _ => false
+notation:50 e:51 " matches " p:51 => ((match e with | p => true | _ => false) : Bool)
 
 -- Declare `this` as a keyword that unhygienically binds to a scope-less `this` assumption (or other binding).
 -- The keyword prevents declaring a `this` binding except through metapgrogramming, as is done by `have`/`show`.


leodemoura:
> To be honest, I don't why it even works at all with if-then-else because there the expected type is also Prop.

@gebner The `if-then-else` notation waits for the type of the condition to be available.

```lean
macro_rules
  | `(if $c then $t:term else $e:term) => `(let_mvar% ?m := $c; wait_if_type_mvar% ?m; ite ?m $t $e)
```

Note that
```lean
set_option pp.match true

#check fun x : Option Nat => if x matches some .. then 0 else 1 -- works
#check fun x : Option Nat => if (x matches some .. : Prop) then 0 else 1 -- fail
```

The second fails because the type of the condition is `Prop`, and we fail to synthesize the `Decidable` instance
```lean
  Decidable (_check.match_1 (fun x => Prop) x (fun x => true = true) fun x => false = true)
```
It is the same problem we have with your example using `decide`.

```

**Comment by gebner** [comment]
> The if-then-else notation waits for the type of the condition to be available.

Ah, this explains why it worked most of the time but why I doubted my sanity when I couldn't reproduce the bug separately.  Before this PR, the match waited for the expected type to be available and the if-then-else waited for the type of the match to be available.  So whether the `matches` worked depended on whether `matches` or if-then-else succeeded to elaborate first without postponing.

```lean
def f (x : Nat) := if x matches 1 then x else 0 -- success
def g (x : Nat) := if x + 1 matches 1 then x else 0 -- fail?!?
```
(At least that was my theory before looking at the elab traces for the examples above.  `wait_if_type_mvar%` doesn't postpone in the `g` function, maybe because the type is a let??)

leodemoura:
> (At least that was my theory before looking at the elab traces for the examples above. wait_if_type_mvar% doesn't postpone in the g function, maybe because the type is a let??)

@gebner Good catch. Thanks. I will try to fix this issue today.

PR #1004, opened 2022-02-05
Handle overlapping definitions

Garmelon:
Overlapping definitions occur in places like structure definitions.

This PR makes the LSP consider *all* definitions/uses at the cursor position for certain requests. For example, finding references of a structure definition now returns uses of the structure name and its `mk` function. However, finding references from a use (and not the definition) of the structure doesn't return the uses of the `mk` function.

This is better than the previous behavior where the `mk` function's definition would often entirely shadow the structure's definition, leading to no references being found. However, it might not be entirely consistent as finding references of a found reference does not always yield the same set of references as the original query.

Solving this by always viewing exactly overlapping definitions as one definition might be desirable, but requires a bit more infrastructure than this (relatively quick) fix.


Garmelon:
CI doesn't seem to want to run for some reason

Garmelon:
Tried a force-push to get GitHub to try again

Issue #1012, opened 2022-02-10
Code folding support in the language server


SquidDev:
Several language constructs in Lean are often written without indentation, for instance `section`, `namespace`, `inductive` and `match`:

```lean
section sect

inductive Maybe (α : Type) : Type 
| none
| some (val : α)

example : Maybe α → α → α := λ m default =>
  match m with
  | Maybe.some x => x
  | Maybe.none => default

end sect
```

Sometimes it may be useful to fold these away (especially in the case of section/namespace). However editors often use indentation as a guide for foldable regions, and so these constructs are not considered foldable.

[LSP has support for custom folding ranges](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_foldingRange), which allows the language server to define its own regions. I've had a quick flick through `RequestHandling.lean`, and it definitely seems like it would be possible to implement for the basic cases above.

It's worth noting that folding range requires you to provide _all_ ranges (including those previously inferred via indentation), and I'm not sure how well this'd work with the rest of Lean's extensible syntax.

I'd be happy to help with/work on the implementation if useful, just felt it was worth getting some feedback/thoughts first!

Kha:
Feel free to give it a try if you want, you already found the right place to start, after all. You can always ask for help on Zulip if you're stuck.

For a first implementation, I would focus on folding namespaces, sections, and individual commands. The merit of term-level folds is not clear to me, especially since terms may not necessarily start at the beginning of the line. Imports and (doc) comments are another possible use case suggested by the spec.

PR #1014, opened 2022-02-11
LSP code folding support

SquidDev:
This adds code folding support to the language server (see #1012), making the following constructs foldable:
 - Sections and namespaces
 - Blocks of import/open statements
 - Multi-line commands (so mostly definitions)
 - Mutual definitions
 - Module-level doc comments
 - Top-level definition doc comments

<details>
<summary><strong>Example:</strong> Screenshot of VS Code with the test file open, and the folding handles visible.</summary>

image

</details>

Afraid I've not used Lean for a few years now (and obviously this was Lean 3), so sure I've made some mistakes. I'm not especially happy with how doc-comments are handled - don't know if there's a better way of poking the syntax?

### DIFF
--- a/src/Lean/Server/FileWorker/RequestHandling.lean
+++ b/src/Lean/Server/FileWorker/RequestHandling.lean
@@ -417,18 +417,9 @@ def handleSemanticTokensRange (p : SemanticTokensRangeParams)
 partial def handleFoldingRange (p : FoldingRangeParams)
   : RequestM (RequestTask (Array FoldingRange)) := do
   let doc ← readDoc
-  asTask do
-    let ⟨snaps, e?⟩ ← doc.allSnaps.updateFinishedPrefix
-    let mut stxs := snaps.finishedPrefix.map (·.stx)
-    match e? with
-    | some ElabTaskError.aborted =>
-      throw RequestError.fileChanged
-    | some (ElabTaskError.ioError e) =>
-      throw (e : RequestError)
-    | _ => pure ()
-
-    let lastSnap := snaps.finishedPrefix.getLastD doc.headerSnap
-    stxs := stxs ++ (← parseAhead doc.meta.text.source lastSnap).toList
+  let t ← doc.allSnaps.waitAll
+  mapTask t fun (snaps, _) => do
+    let stxs := snaps.map (·.stx)
     let (_, ranges) ← StateT.run (addRanges doc.meta.text [] stxs) #[]
     return ranges
   where
@@ -451,18 +442,13 @@ partial def handleFoldingRange (p : FoldingRangeParams)
         addRanges text sections stxs
       | _ => do
         if isImport stx then
-          addImportRanges text sections stx.getPos? stx.getTailPos? stxs
+          let (imports, stxs) := stxs.span isImport
+          let last := imports.getLastD stx
+          addRange text FoldingRangeKind.imports stx.getPos? last.getTailPos?
+          addRanges text sections stxs
         else
           addCommandRange text stx
           addRanges text sections stxs
-
-    addImportRanges (text : FileMap) sections startP endP stxs := do
-      if let stx::stxs := stxs then
-        if isImport stx then
-          return ← addImportRanges text sections (startP <|> stx.getPos?) (stx.getTailPos? <|> endP) stxs
-
-      addRange text FoldingRangeKind.imports startP endP
-      addRanges text sections stxs
 
     addCommandRange text stx :=
       match stx.getKind with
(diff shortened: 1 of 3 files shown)


src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -414,6 +414,87 @@ def handleSemanticTokensRange (p : SemanticTokensRangeParams)
...
+    let mut stxs := snaps.finishedPrefix.map (·.stx)
+    match e? with
+    | some ElabTaskError.aborted =>
+      throw RequestError.fileChanged
+    | some (ElabTaskError.ioError e) =>
+      throw (e : RequestError)
+    | _ => pure ()
+
+    let lastSnap := snaps.finishedPrefix.getLastD doc.headerSnap
+    stxs := stxs ++ (← parseAhead doc.meta.text.source lastSnap).toList
```
Kha, line 431: This is copied from `handleDocumentSymbol`, right? Let's put that in a new function... hmm... `getDocumentSyntaxWithParseAhead`?

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -414,6 +414,87 @@ def handleSemanticTokensRange (p : SemanticTokensRangeParams)
...
+          addCommandRange text stx
+          addRanges text sections stxs
+
+    addImportRanges (text : FileMap) sections startP endP stxs := do
+      if let stx::stxs := stxs then
+        if isImport stx then
+          return ← addImportRanges text sections (startP <|> stx.getPos?) (stx.getTailPos? <|> endP) stxs
+
+      addRange text FoldingRangeKind.imports startP endP
+      addRanges text sections stxs
```
Kha, line 465: If we float this line into `addRanges`, every branch would end in `addRanges text _ stxs` and we could rewrite it as a `for` loop, correct? Just a suggestion, it's fine as is.

Kha: The recursion here should also be avoidable using `stxs.partition isImport`

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -414,6 +414,87 @@ def handleSemanticTokensRange (p : SemanticTokensRangeParams)
...
+        -- When visiting a declaration, attempt to fold the doc comment
+        -- separately to the main definition.
+        -- We never fold other modifiers, such as annotations.
+        if let `($dm:declModifiers $decl) := stx then
+          if let some comment := dm[0].getOptional? then
+            addRangeFromSyntax text FoldingRangeKind.comment comment
+
+          addRangeFromSyntax text FoldingRangeKind.region decl
+          return
+
```
Kha, line 481:
```suggestion
        else
```

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -414,6 +414,87 @@ def handleSemanticTokensRange (p : SemanticTokensRangeParams)
...
+        addRangeFromSyntax text FoldingRangeKind.region stx
+      | _ =>
+        addRangeFromSyntax text FoldingRangeKind.comment stx
+
+    addRangeFromSyntax (text : FileMap) kind stx := addRange text kind stx.getPos? stx.getTailPos?
+
+    addRange (text : FileMap) kind start stop := do
+      if let (Option.some startP, Option.some endP) := (start, stop) then
```
Kha, line 489:
```suggestion
    addRange (text : FileMap) kind start? stop? := do
      if let (some start, some stop) := (start?, stop?) then
```

Kha (commented):
Great job, and fast! I only have some minor comments. Do tell me if you want to implement any other requests... :)

src/Lean/Server/FileWorker/RequestHandling.lean
SquidDev: > If we float this line into `addRanges`, every branch would end in `addRanges text _ stxs` and we could rewrite it as a `for` loop, correct? Just a suggestion, it's fine as is.

I don't think this is quite right, because sometimes our we recurse using `addImportRanges` and sometimes with `addRanges`, so it's unfortunately not quite a trivial for loop.

> The recursion here should also be avoidable using `stxs.partition isImport`

Do you mean `.span` rather than `.partition`? Yes, that might help a lot actually! - will give that a fiddle.

Kha: Yes, looks good!

src/Lean/Server/FileWorker/RequestHandling.lean
Kha: Interesting, inaccurate in what way?

> as we'd need to handle editors which don't support partial results

Which is all of them, unfortunately

SquidDev: Here is a relatively minimal case where parse-ahead fails. I can only reproduce when running under a debug build, though I imagine the same would happen with a release build if you had a term which takes sufficiently long to elaborate.

```lean
def foo :=
  0

infixl:65 " :=: " => Eq

theorem bar : 0 :=: 0 :=
  rfl
```

Editing the definition of `foo` will result in only `foo` having a folding region as, I assume, `infixl` hasn't been processed and so the definition of `bar` doesn't parse. I don't think there's much that can really be done about this :(.

Kha:
Sorry about the delay, I lost track of this PR. I see you've since updated it to wait for elaboration, so given that we do this for other requests such as semantic highlighting as well, I think this is good to go!

PR #1015, opened 2022-02-13
doc: document match h:e with syntax

ammkrn:
This feature doesn't seem to be documented yet.

### DIFF
--- a/doc/declarations.md
+++ b/doc/declarations.md
@@ -675,7 +675,7 @@ When a ``match`` has only one line, the vertical bar may be left out. In that ca
     def bar₄ (p : Nat × Nat) : Nat :=
     let ⟨m, n⟩ := p in m + n
 
-Information about the term being matched can be preserved in each branch using the syntax `match h:t with` (with no whitespace between the name of the hypothesis, the colon, and the term). For example, a user may want to match a term `ns : List Nat`, while tracking the hypotheses `ns = []` and `ns = h :: t` in each match arm:
+Information about the term being matched can be preserved in each branch using the syntax `match h:t with` (with no whitespace between the name of the hypothesis, the colon, and the term). For example, a user may want to match a term `ns : List Nat`, while tracking the hypothesis `ns = []` or `ns = h :: t` in the respective match arm:
 
 ```lean
 def foo (ns : List Nat) (h1 : ns ≠ []) (k : Nat -> Char) : Char :=


doc/declarations.md
```diff
@@ -675,6 +675,20 @@ When a ``match`` has only one line, the vertical bar may be left out. In that ca
     def bar₄ (p : Nat × Nat) : Nat :=
     let ⟨m, n⟩ := p in m + n
 
+Information about the term being matched can be preserved in each branch using the syntax `match h:t with` (with no whitespace between the name of the hypothesis, the colon, and the term). For example, a user may want to match a term `ns : List Nat`, while tracking the hypotheses `ns = []` and `ns = h :: t` in each match arm:
```
eric-wieser, line 678: Does this really require no whitespace?

Kha: Yes, because it would otherwise clash with the explicit type/convoy syntax `match ns : List Nat -> Char with ...`

digama0: Actually, it only requires no whitespace on the left. I like to write this as `match h: e with` which doesn't look as bad.

leodemoura: @Kha I agree the whitespace sensitivity here is quite annoying. What about moving the other `:` after the `with`?
That is, we would write `match ns with : Nat -> Char | ...`. 
I am also open to suggestions.

digama0: Not sure it's the right choice for lean, but in Coq the syntax for the match motive is `returns`, i.e. `match ns returns Nat -> Char with ...`. I think `as` or another keyword-like word could also work in that position.

leodemoura: @digama0 Thanks for the suggestion. I like the `as`. @Kha Any thoughts?

Kha: That would work, but `:` doesn't seem all that fitting for this purpose to begin with. I feel like even `match h@t` would be more appropriate; it's not quite consistent either, but at least it is pattern-related.

ammkrn: While we're throwing hats in the ring, `match (motive : ...) ns with` or `match (motiveType := ...) ns with` would be consistent with the `(generalizing := true)` syntax.

leodemoura: @Kha We use `:` for this purpose in other places. Examples: `cases` tactic, `generalize` tactic, equality proof for nested named patterns (e.g., `xs@h:pattern`). To be honest, I am quite happy with this part. We have to fix only the other `:`. Example:
```lean
match h:n as Nat -> Bool with
| 0 => ...
| i+1 => ...
```
looks resonable.

leodemoura: > While we're throwing hats in the ring, `match (motive : ...) ns with` or `match (motiveType := ...) ns with` would be consistent with the `(generalizing := true)` syntax.

I link the `(motive := ...)` suggestion too.

Kha: I like @ammkrn's suggestion given that this is a very rare use case, but `as` is fine as well

digama0: `as` isn't a keyword right now, though, right? It comes up every once in a while in list definitions where you have variables named `as` and `bs`, for example. I'm on board with `(motive := ...)` since it's a rare usage.

PR #1019, opened 2022-02-16
feat: add position to mod doc

xubaiw:
This adds position info to module doc comments, which is essential for doc-gen4 to render them properly. As mod doc is not used in Lean itself, this pull won't affect other parts of the code.


xubaiw:
The `moduleDoc.lean` segfaults at `printDoc` when tests are run in **stage1**, but no such error in stage2, and updating stage0 code may fix this.

leodemoura:
@xubaiw Yes, the failure is due to staging issues. I will merge manually and perform the `update-stage0`.

leodemoura:
Merged manually.

PR #1032, opened 2022-02-28
feat: add theory for ac normalization.

DanielFabian:
This is the first in a series of PRs getting `ac_refl` merged in. In this initial PR, I define a few type classes and prove soundness of a decision procedure that is used by an AC normalization tactic. 

It only operates on a single operator, at a time, because the tactic does these steps:
- go through the target expr top down, looking for occurrences of operators that are at least associative, and, optionally, commutative and idempotent. As well as potentially having neutral elements.
- find an occurrence that is unnormalized
- add an ac equality proof to the local context
- use the proof to rewrite the goal
- remove the proof from the local context
- and keep repeating until we reach a fixed point
- apply `rfl`

### DIFF
--- a/src/Lean/Meta/Tactic/AC/Main.lean
+++ b/src/Lean/Meta/Tactic/AC/Main.lean
@@ -86,50 +86,56 @@ def toACExpr (op l r : Expr) : MetaM (Array Expr × ACExpr) := do
 def buildNormProof (preContext : PreContext) (l r : Expr) : MetaM (Lean.Expr × Lean.Expr) := do
   let (vars, acExpr) ← toACExpr preContext.op l r
 
-  let (isNeutrals, context) ← mkContext vars
-  let acExprNormed := Data.AC.evalList ACExpr preContext $ Data.AC.norm (preContext, isNeutrals) acExpr
-  let tgt ← convertTarget vars acExprNormed
-  let lhs ← convert acExpr
-  let rhs ← convert acExprNormed
   let α ← inferType vars[0]
   let u ← getLevel α
+  let (isNeutrals, context) ← mkContext α u vars
+  let acExprNormed := Data.AC.evalList ACExpr preContext $ Data.AC.norm (preContext, isNeutrals) acExpr
+  let tgt := convertTarget vars acExprNormed
+  let lhs := convert acExpr
+  let rhs := convert acExprNormed
   let proof := mkAppN (mkConst ``Context.eq_of_norm [u]) #[α, context, lhs, rhs, ←mkEqRefl (mkConst ``Bool.true)]
   return (proof, tgt)
 where
-  mkContext (vars : Array Expr) : MetaM (Array Bool × Expr) := do
+  mkContext (α : Expr) (u : Level) (vars : Array Expr) : MetaM (Array Bool × Expr) := do
     let arbitrary := vars[0]
+    let zero := mkLevelZeroEx ()
+    let noneE := mkApp (mkConst ``Option.none [zero])
+    let someE := mkApp2 (mkConst ``Option.some [zero])
 
     let vars ← vars.mapM fun x => do
-      let isNeutral ←
+      let isNeutral :=
+        let isNeutralClass := mkApp3 (mkConst ``IsNeutral [u]) α preContext.op x
         match ←getInstance ``IsNeutral #[preContext.op, x] with
-        | none => pure (false, ←mkAppOptM ``Option.none #[←mkAppM ``IsNeutral #[preContext.op, x]])
-        | some isNeutral => pure (true, ←mkAppM ``Option.some #[isNeutral])
+        | none => (false, noneE isNeutralClass)
+        | some isNeutral => (true, someE isNeutralClass isNeutral)
 
-      return (isNeutral.1, ←mkAppM ``Variable.mk #[x, isNeutral.2])
+      return (isNeutral.1, mkApp4 (mkConst ``Variable.mk [u]) α preContext.op x isNeutral.2)
 
     let (isNeutrals, vars) := vars.unzip
     let vars := vars.toList
-    let vars ← mkListLit (←mkAppM ``Variable #[preContext.op]) vars
+    let vars ← mkListLit (mkApp2 (mkConst ``Variable [u]) α preContext.op) vars
 
-    let comm ←
+    let comm :=
+      let commClass := mkApp2 (mkConst ``IsCommutative [u]) α preContext.op
       match preContext.comm with
-      | none => mkAppOptM ``Option.none #[←mkAppM ``IsCommutative #[preContext.op]]
-      | some comm => mkAppM ``Option.some #[comm]
+      | none => noneE commClass
+      | some comm => someE commClass comm
 
-    let idem ←
+    let idem :=
+      let idemClass := mkApp2 (mkConst ``IsIdempotent [u]) α preContext.op
       match preContext.idem with
-      | none => mkAppOptM ``Option.none #[←mkAppM ``IsIdempotent #[preContext.op]]
-      | some idem => mkAppM ``Option.some #[idem]
+      | none => noneE idemClass
+      | some idem => someE idemClass idem
 
-    return (isNeutrals, ←mkAppM ``Lean.Data.AC.Context.mk #[preContext.op, preContext.assoc, comm, idem, vars, arbitrary])
+    return (isNeutrals, mkApp7 (mkConst ``Lean.Data.AC.Context.mk [u]) α preContext.op preContext.assoc comm idem vars arbitrary)
 
-  convert : ACExpr → MetaM Expr
-    | Data.AC.Expr.op l r => do mkAppM ``Data.AC.Expr.op #[←convert l, ←convert r]
-    | Data.AC.Expr.var x => mkAppM ``Data.AC.Expr.var #[mkNatLit x]
+  convert : ACExpr → Expr
+    | Data.AC.Expr.op l r => mkApp2 (mkConst ``Data.AC.Expr.op) (convert l) (convert r)
+    | Data.AC.Expr.var x => mkApp (mkConst ``Data.AC.Expr.var) $ mkNatLit x
 
-  convertTarget (vars : Array Expr) : ACExpr → MetaM Expr
-    | Data.AC.Expr.op l r => do mkAppM' preContext.op #[←convertTarget vars l, ←convertTarget vars r]
-    | Data.AC.Expr.var x => return vars[x]
+  convertTarget (vars : Array Expr) : ACExpr → Expr
+    | Data.AC.Expr.op l r => mkApp2 preContext.op (convertTarget vars l) (convertTarget vars r)
+    | Data.AC.Expr.var x => vars[x]
 
 def rewriteUnnormalized (mvarId : MVarId) : MetaM Unit := do
   let simpCtx :=


src/Lean/Meta/Tactic/AC/Main.lean
```diff
@@ -0,0 +1,337 @@
...
+    trace[Meta.AC] "picking norm strategy {MessageData.ofGoal mvarId}"
+    let mvarId ← addAcEq mvarId e target
+    let (h_ac, mvarId) ← intro mvarId `h_ac
+    let res ← rewrite mvarId (←getMVarType mvarId) (mkFVar h_ac)
+    let [] := res.mvarIds | throwError "no meta variables expected after rewrite"
+    let mvarId ← replaceTargetEq mvarId res.eNew res.eqProof
+    let mvarId ← clear mvarId h_ac
+    rewriteUnnormalized mvarId
+
+syntax (name := ac_refl) "ac_refl " : tactic
```
leodemoura, line 329: The syntax declaration should be at `Init/Notation`. Otherwise, users will have to import Lean to use the tactic.

src/Lean/Meta/Tactic/AC/Main.lean
```diff
@@ -0,0 +1,337 @@
...
+
+def addAcEq (mvarId : MVarId) (e : NormalizedExpr) (target : Expr) : M MVarId := do
+  let (proof, ty) ← buildProof e (←e.norm)
+  let goal ← withLocalDeclD `h_ac ty fun h_ac =>
+    mkForallFVars #[h_ac] target
+  let goal ← mkFreshExprMVar goal
+  assignExprMVar mvarId (mkApp goal proof)
+  return goal.mvarId!
+
+partial def rewriteUnnormalized (mvarId : MVarId) : M Unit :=
```
leodemoura, line 303: We should implement the tactic on top of `simp`.

leodemoura:
Here are two examples where `ac_refl` produces proofs that are rejected by the kernel.
```lean
import Lean

inductive Vector (α : Type u) : Nat → Type u where
  | nil  : Vector α 0
  | cons : α → Vector α n → Vector α (n+1)

def f (n : Nat) (xs : Vector α n) := xs

-- Dependent types trigger incorrect proofs
theorem ex1 (n m : Nat) (xs : Vector α (n+m)) (ys : Vector α (m+n)) : (f (n+m) xs, f (m+n) ys, n+m) = (f (n+m) xs, f (m+n) ys, m+n) := by
  ac_refl

-- Binders also trigger invalid proofs
theorem ex2 (n : Nat) : (fun x => n + x) = (fun x => x + n) := by
  ac_refl
```

src/Init/Data/AC.lean
```diff
@@ -0,0 +1,328 @@
...
+
+theorem Context.evalList_removeNeutrals (ctx : Context α) (e : List Nat) : evalList α ctx (removeNeutrals ctx e) = evalList α ctx e := by
+  induction e using List.two_step_induction with
+  | empty => rfl
+  | single =>
+    simp [removeNeutrals, removeNeutrals.loop]; split
+    case h_1 => rfl
+    case h_2 h _ _ => split at h <;> simp_all
+  | step x y ys ih =>
+    match h₁ : ContextInformation.isNeutral ctx x, h₂ : ContextInformation.isNeutral ctx y, h₃ : removeNeutrals.loop ctx ys with
```
leodemoura, line 286: We can use `<;>` to avoid this kind of big `match`. 
```lean
    cases h₁ : ContextInformation.isNeutral ctx x <;> cases h₂ : ContextInformation.isNeutral ctx y <;> cases h₃ : removeNeutrals.loop ctx ys 
      <;> simp [removeNeutrals, removeNeutrals.loop, h₁, h₂, h₃, evalList, ←ih] <;> (try simp [unwrap_isNeutral h₂ |>.2])
      <;> (try simp [unwrap_isNeutral h₁ |>.1])
```

src/Lean/Meta/Tactic/AC/Main.lean
DanielFabian: done

src/Init/Data/AC.lean
DanielFabian: done

src/Lean/Meta/Tactic/AC/Main.lean
DanielFabian: done

DanielFabian: > -- Binders also trigger invalid proofs
> theorem ex2 (n : Nat) : (fun x => n + x) = (fun x => x + n) := by
>   ac_refl
> ```

Most of those were relatively minor. The binder one is harder. I can't just universally quantify over the bound variable, because if it gets applied too much, it could easily cause an oscillation (or at least unnormalize an operator).

The rewrite has to only happen in context. Maybe I can rewrite the bound variable and then wrap it a few times in a `congrArg` or something to only rewrite in context here.

src/Lean/Meta/Tactic/AC/Main.lean
```diff
@@ -312,19 +312,30 @@ partial def rewriteUnnormalized (mvarId : MVarId) : M Unit :=
         assignExprMVar mvarId proof
       else throwError ""
     catch _ => throwError "cannot synthesize proof:\n{MessageData.ofGoal mvarId}"
-  | ProofStrategy.rfl tgt =>
-    trace[Meta.AC] "picking rfl strategy {MessageData.ofGoal mvarId}"
-    assignExprMVar mvarId (←mkAppM ``Eq.refl #[tgt])
+  | ProofStrategy.simp =>
+    trace[Meta.AC] "picking simp strategy {MessageData.ofGoal mvarId}"
+    let simpCtx ← Simp.Context.mkDefault
+    let newGoal ← simpTarget mvarId simpCtx
```
leodemoura, line 318: Note that `simpTarget` returns `none` when the goal has been closed (a good thing). 
It confuses me from time to time. @Kha and I talked about adding a new type (with a `closed` constructor), but didn't find a good name.

src/Lean/Meta/Tactic/AC/Main.lean
```diff
@@ -312,19 +312,30 @@ partial def rewriteUnnormalized (mvarId : MVarId) : M Unit :=
         assignExprMVar mvarId proof
       else throwError ""
     catch _ => throwError "cannot synthesize proof:\n{MessageData.ofGoal mvarId}"
-  | ProofStrategy.rfl tgt =>
-    trace[Meta.AC] "picking rfl strategy {MessageData.ofGoal mvarId}"
-    assignExprMVar mvarId (←mkAppM ``Eq.refl #[tgt])
+  | ProofStrategy.simp =>
+    trace[Meta.AC] "picking simp strategy {MessageData.ofGoal mvarId}"
+    let simpCtx ← Simp.Context.mkDefault
```
leodemoura, line 317: The default context contains all `[simp]` theorems registered in the system.
For `ac_refl`, a simpler context such as 
```
 {
      simpTheorems  := {}
      congrTheorems := (← getSimpCongrTheorems)
      config        := Simp.neutralConfig
   }
```
is probably more appropriate.

src/Lean/Meta/Tactic/AC/Main.lean
```diff
@@ -312,19 +312,30 @@ partial def rewriteUnnormalized (mvarId : MVarId) : M Unit :=
...
   | ProofStrategy.norm e =>
     trace[Meta.AC] "picking norm strategy {MessageData.ofGoal mvarId}"
     let mvarId ← addAcEq mvarId e target
     let (h_ac, mvarId) ← intro mvarId `h_ac
-    let res ← rewrite mvarId (←getMVarType mvarId) (mkFVar h_ac)
-    let [] := res.mvarIds | throwError "no meta variables expected after rewrite"
-    let mvarId ← replaceTargetEq mvarId res.eNew res.eqProof
-    let mvarId ← clear mvarId h_ac
-    rewriteUnnormalized mvarId
+    let simpCtx ← Simp.Context.mkDefault
```
leodemoura, line 326: See comment above about the context.

src/Lean/Meta/Tactic/AC/Main.lean
```diff
@@ -312,19 +312,30 @@ partial def rewriteUnnormalized (mvarId : MVarId) : M Unit :=
...
+    withMVarContext mvarId do
+    let simpCtx := { simpCtx with simpTheorems := ←simpCtx.simpTheorems.add #[] (mkFVar h_ac) }
+
+    trace[Meta.AC] "pre rewrite state:\n{MessageData.ofGoal mvarId}\n"
+    let mvarId ← simpTarget mvarId simpCtx
```
leodemoura, line 331: I think `Simp.main` function is more appropriate. It allows us to provide two functions `pre` and `post` that are invoked before/after we visit a term. We should invoke the ac normalization code in the `post`. 
Here is an example for `Simp.main`
https://github.com/leanprover/lean4/blob/master/src/Lean/Meta/Tactic/Split.lean#L19-L33

leodemoura: Here is an example where we check the "parent" term to decide whether we are going to "simplify" something or not.
https://github.com/leanprover/lean4/blob/master/src/Lean/Meta/Tactic/LinearArith/Simp.lean#L16-L24
You have to do something similar when invoking the AC normalizer in the post function. If the parent is the same operator, we should not normalize.

src/Lean/Meta/Tactic/AC/Main.lean
```diff
@@ -273,7 +273,7 @@ where
 
 inductive ProofStrategy
```
leodemoura, line 274: We don't need the `ProofStrategy`, we can do it by a single `simp` traversal where you invoke the normalization code from the `Simp.main` `post`.

src/Lean/Meta/Tactic/AC/Main.lean
leodemoura: The solution is to use `Simp.main`. It will apply all the congruence lemmas and `funext` for us.

src/Lean/Meta/Tactic/AC/Main.lean
DanielFabian: I am aware of that. It's just that the `simp` strategy is a finisher. So if it returns a new goal, `ac_refl` gives up and throws.

src/Lean/Meta/Tactic/AC/Main.lean
DanielFabian: ya, got it. It will also make the `findUnnormalized` easier, because it won't need to deal with multiple operators. Instead, it'll just try to match the head node and check if it's one of our operators. I'll look into it in the next few days, but probably not today.

leodemoura: I think you don't need `findUnnormalized`. Your `post` function can check whether the given expression is an AC operator, but the parent is not the same one.

DanielFabian: done

src/Lean/Meta/Tactic/AC/Main.lean
DanielFabian: done

src/Lean/Meta/Tactic/AC/Main.lean
DanielFabian: done

src/Lean/Meta/Tactic/AC/Main.lean
DanielFabian: done

src/Lean/Meta/Tactic/AC/Main.lean
DanielFabian: done

src/Lean/Meta/Tactic/AC/Main.lean
```diff
@@ -0,0 +1,178 @@
+/-
+Copyright (c) 2022 Microsoft Corporation. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Dany Fabian
+-/
+import Init.Data.AC
+import Lean.Meta.AppBuilder
+import Lean.Elab.Tactic.Basic
+import Lean.Elab.Tactic.Rewrite
```
Kha, line 9: I don't think we should import Lean.Elab from Lean.Meta?

leodemoura: Thanks for catching this. We should split this file into `Meta` and `Elab` (frontend) parts.

PR #1042, opened 2022-03-11
Generalize inferred namespace notation to function types

Kha:


src/Lean/Elab/App.lean
```diff
@@ -854,22 +854,23 @@ private partial def elabAppFn (f : Syntax) (lvals : List LVal) (namedArgs : Arra
...
+          | throwError "invalid dotted identifier notation, expected type is not of the form (... → C ...) where C is a constant{indentExpr expectedType}"
+        let idNew := declName ++ id.getId.eraseMacroScopes
+        unless (← getEnv).contains idNew do
+          throwError "invalid dotted identifier notation, unknown identifier `{idNew}` from expected type{indentExpr expectedType}"
+        let fConst ← mkConst idNew
+        addTermInfo f fConst
+        let s ← observing do
+          let e ← elabAppLVals fConst lvals namedArgs args expectedType? explicit ellipsis
+          if overloaded then ensureHasType expectedType? e else pure e
+        return acc.push s
```
Kha, line 873: The diff is terrible because I noticed that the branch had an odd indentation, so I took the chance to also streamlined the logic a bit using `let _:= _ | _`

gebner:
When I first read the PR title I thought that it would look up the name in the `Function` namespace.  (Which is a somewhat regularly requested feature for dot-notation, btw.)

leodemoura:
> When I first read the PR title I thought that it would look up the name in the `Function` namespace. (Which is a somewhat regularly requested feature for dot-notation, btw.)

For the `dot-notation`, I think this is a good idea, and I will add it.

For the extension implemented in this PR, I think the new tests make a compelling case for the approach implemented here.
What do you think?

gebner:
> For the extension implemented in this PR, I think the new tests make a compelling case for the approach implemented here.
What do you think?

To be honest, I don't find the new tests very compelling.  You can always simulate the behavior of this PR by writing `xs.foldl (.insert · ·) a` instead of `xs.foldl .insert a`, after all.  This example also rubs me the wrong way because what we actually want to write here is `a ++ xs`, and we're just working around the lack of a proper collections API.

It's still a reasonable change, I just don't see the big applications.

leodemoura:
@gebner Thanks for the feedback. I see your point. Perhaps, we should use `Function` as the namespace for arrows in both the dot-notation and this one. @Kha What do you think?

leodemoura:
BTW, just to make it clear, I am fully sold on using `Function` as the namespace for the old dot-notation.

Kha:
Something like `map .ok`, where none of the parameters already carry the namespace, would probably be a better example. I would definitely expect that one to work.

gebner:
Yeah, `xs.map .ok` or `xs.map .inl` are good examples.

leodemoura:
I have one example for using the `Function` namespace when the expected type is an arrow. 
We would be able to write `.const 5` instead of `Function.const 5`. I think it makes sense since `Function.const` is morally a "constructor" for functions.

Examples such as `xs.map .ok` can be written as `xs.map (.ok  ·)` without the extension in this PR.

Kha:
> I think it makes sense since `Function.const` is morally a "constructor" for functions.

Yes, there is some consistency with the new projection notation generalization. But if that is basically the only function constructor in practice, I find it hard to justify introducing a new feature just for it. In fact, I feel `const` is special enough (and less useful with any additional char you have to type for it, compared to its expanded form) that we should just `export` it from `Function`.

I can't say how often we could benefit from the alternative semantics implemented in this PR in comparison. But I would say there is a consistency argument to be made there as well - `(Except.ok ·)` and `Except.ok` are generally equivalent, so it feels like a "gotcha" to me if `(.ok ·)` and `.ok` are not.

leodemoura (approved):

PR #1044, opened 2022-03-16
Always use .zip compression for windows build artifacts.

lovettchris:
.tar.zst is not an easy format to handle on Windows.  .zip is a lot more standard (and is built into windows).

Having a .zip file means [other windows workflows](https://github.com/leanprover/vscode-lean4/pull/154) do not need to install or use MSYS2 for tar and zstd decompression.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -184,7 +184,7 @@ jobs:
           fi
           zip -rq9 pack/$dir.zip $dir
       - uses: actions/upload-artifact@v2
-        if: matrix.release
+        if: matrix.release && github.event_name == 'push' && github.repository == 'leanprover/lean4'
         with:
           name: build-${{ matrix.name }}
           path: pack/*


Kha:
This adds ~30s to the Windows build. The artifacts were not necessarily meant for public consumption, but if we want them to be, I suppose this is a fair tradeoff considering the entire Windows build takes ~30min. Even if the original use case for this PR is no more.

But please take care to preserve the semantics of the existing code. Your current changes would remove the Windows .tar.zst release tarball, which elan uses even on Windows.

Kha:
If we're okay with the extra 30s for each job, we should probably use .zip on all platforms in the non-release case. `zstd` isn't really a standard program on any of them.

gebner:
> so how hard would it be to make it use the .zip file

The reason we use tar.zst instead of tar.gz or even zip is because the tar.zstd is a lot smaller (on linux it's 100 megabytes smaller, on windows 40 megabytes).  It already takes long enough to download a nightly, we shouldn't make it even slower.

Kha:
> then our overall releases would be smaller

The cumulative size of the release, on the other hand, is basically irrelevant. There is no cost to hosting both formats... for us.

leodemoura:
What is the status of this PR?
We now have `.zip` and `.tar.zst` for each platform. 
https://github.com/leanprover/lean4/releases/tag/v4.0.0-m4

lovettchris:
> What is the status of this PR? We now have `.zip` and `.tar.zst` for each platform. https://github.com/leanprover/lean4/releases/tag/v4.0.0-m4

In the releases on github/leanprover-nightly yes, but this PR is about doing the same for the uploaded github artifacts so I can get a zip file in other workflows that use those github artifacts like the TPIL one.

Kha:
@lovettchris Do you still care about this? If so, could you fix the .tar.zst generation as mentioned above?

lovettchris:
@Kha yes I do, but what fix are you referring to exactly?

Kha:
The Windows workflow should create the same .tar.zst as before

leodemoura:
@lovettchris Are you still planning to address the issues on this PR? It is fine if you are not.

lovettchris:
Yes, I have incorporated the CR feedback as per discussion above, and I have reduced the size of the zip file, down from 216mb to 136mb.  I suspect this change will also help the TPIL workflow that is downloading these zip files as a build artifact since it will now always have a zip file there to work with.

.github/workflows/ci.yml
```diff
@@ -179,12 +179,12 @@ jobs:
...
           if [[ '${{ startsWith(github.ref, 'refs/tags/v') && matrix.release }}' == true || -n '${{ needs.set-nightly.outputs.nightly }}' ]]; then
             tar cf - $dir | zstd -T0 --no-progress -19 -o pack/$dir.tar.zst
-            zip -rq pack/$dir.zip $dir
           else
             tar cf - $dir | zstd -T0 --no-progress -o pack/$dir.tar.zst
           fi
+          zip -rq9 pack/$dir.zip $dir
       - uses: actions/upload-artifact@v2
-        if: matrix.release
+        if: matrix.release && github.event_name == 'push' && github.repository == 'leanprover/lean4'
```
Kha, line 187: Why? This is completely harmless.

lovettchris: the TPIL trigger triggers a build over there and we don't want TPIL builds picking up incomplete pull requests.

Kha:
It looks like the 30s overhead have now become 3min btw

.github/workflows/ci.yml
lovettchris: I always thought of github `build-artifact` as something that belongs to the team members who own the CI workflow.  Just downloading a github `build-artifact` is no simple task from the `gh` command line.

lovettchris:
> It looks like the 30s overhead have now become 3min btw

'tis true, zstd is insanely fast, running in 3 seconds on the previous Linux builds.  Of course, we'll get some of that time back if we don't do the actions/upload-artifact in all cases.

.github/workflows/ci.yml
lovettchris: I see I didn't realize you could download them from here https://github.com/leanprover/lean4/actions/runs/2585458255, that's cool.   For me personally I've never used these because I always build my PR locally and test it before pushing a PR.  But I could see the cross-platform aspect of the binaries could be handy for folks that want to share PR bits with other team members who are on different platforms...  I'm happy to put that back if you think it's useful.

lovettchris: PS: the other reason I'm playing with removing this is to see if it solves this kind of thing over in TPIL: https://github.com/leanprover/theorem_proving_in_lean4/actions/runs/2582579057

leodemoura: @lovettchris I want to clarify this part

> Leo wanted faster than nightly feedback on breaks to TPIL,

Nightly feedback is fine. The original issue I had is that I was not getting any feedback when changes on Lean broke TPIL.

lovettchris: Ok, with this information I can instead change TPIL to stop using these build-artifacts, and we can cancel this PR.  I will change the TPIL workflow to grab the zip files from https://github.com/leanprover/lean4-nightly/releases

leodemoura: Yes, the nightly builds are good enough. I will close this PR.

leodemoura:
See discussion above.

lovettchris:
and the response is to switch to TPIL triggered on nighly releases, see:
- https://github.com/leanprover/lean4/pull/1270
- https://github.com/leanprover/theorem_proving_in_lean4/pull/42

PR #1055, opened 2022-03-19
allow axiom tcs

### LINKED ISSUE #1054: Allow constants to be TCs
### Prerequisites

### Description

Constants are not alowed to be type classes, which, for drafting/prototyping would be really useful.

javra:
Attempts to close #1054

### DIFF
--- a/src/Lean/Class.lean
+++ b/src/Lean/Class.lean
@@ -65,7 +65,7 @@ def hasOutParams (env : Environment) (n : Name) : Bool :=
 private partial def checkOutParam : Nat → Array FVarId → Expr → Except String Bool
   | i, outParams, Expr.forallE _ d b _ =>
     if d.isOutParam then
-      let fvarId    := { name := Name.mkNum `_fvar outParams.size }
+      let fvarId := { name := Name.mkNum `_fvar outParams.size }
       let outParams := outParams.push fvarId
       let fvar      := mkFVar fvarId
       let b         := b.instantiate1 fvar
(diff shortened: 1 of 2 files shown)


leodemoura:
@javra Thanks for the PR. Could you please fix the commit messages?

src/Lean/Class.lean
```diff
@@ -84,7 +84,10 @@ def addClass (env : Environment) (clsName : Name) : Except String Environment :=
     | some decl@(ConstantInfo.inductInfo _) => do
       let b ← checkOutParam 1 #[] decl.type
       Except.ok (classExtension.addEntry env { name := clsName, hasOutParam := b })
-    | some _    => Except.error ("invalid 'class', declaration '" ++ toString clsName ++ "' must be inductive datatype or structure")
+    | some decl@(ConstantInfo.axiomInfo _) => do
+      let b ← checkOutParam 1 #[] decl.type
+      Except.ok (classExtension.addEntry env { name := clsName, hasOutParam := b })
```
javra, line 89: What's the sleekest way to avoid the code duplication here?

Kha: How about
```lean
def addClass (env : Environment) (clsName : Name) : Except String Environment := do
  if isClass env clsName then
    throw s!"class has already been declared '{clsName}'"
  let some decl := env.find? clsName
    | throw s!"unknown declaration '{clsName}'"
  unless decl matches .inductInfo .. || decl matches .axiomInfo .. do
    throw s!"invalid 'class', declaration '{clsName}' must be inductive datatype, structure, or constant"
  let b ← checkOutParam 1 #[] decl.type
  return classExtension.addEntry env { name := clsName, hasOutParam := b }
```

javra:
> Thanks for the PR. Could you please fix the commit messages?

Will squash them into one. I first did not intend this to become a PR ;)

PR #1059, opened 2022-03-22
fix: annotate binders in `intro` for hover / go to def

digama0:


digama0:
Not sure what happened in the windows build, it appears unrelated to this PR.

src/Lean/Elab/Tactic/BuiltinTactic.lean
```diff
@@ -146,17 +146,20 @@ partial def evalChoiceAux (tactics : Array Syntax) (i : Nat) : TacticM Unit :=
...
-    liftMetaTactic fun mvarId => do
-      let (_, mvarId) ← Meta.intro mvarId n
-      pure [mvarId]
+  introStep (ref : Option Syntax) (n : Name) : TacticM Unit := do
+    let fvar ← liftMetaTacticAux fun mvarId => do
+      let (fvar, mvarId) ← Meta.intro mvarId n
+      pure (fvar, [mvarId])
+    if let some stx := ref then
+      withMainContext do
+        Term.addLocalVarInfo stx (mkFVar fvar)
```
Kha, line 162: I didn't look too closely into the involved contexts, is there a reason we can't do this in the `liftMetaTacticAux` block?

Kha:
Thanks, looks like a sensible addition. Could you add a test or two as well?

src/Lean/Elab/Tactic/BuiltinTactic.lean
digama0: `Term.addLocalVarInfo` is in the `TermElabM` monad, and there does not seem to be an easy way to rewrite it to live in `MetaM` and/or open a local `TermElabM` context in the `MetaM`

digama0:
Is there a place for UI tests? The only visible change I am aware of is hover and go to definition info

Kha:
Yes, you can add them to https://github.com/leanprover/lean4/blob/master/tests/lean/interactive/hover.lean

src/Lean/Elab/Tactic/BuiltinTactic.lean
Kha: Right, I misread the signature. Looks fine then.

PR #1066, opened 2022-03-24
Cross-compiled build for Linux/aarch64

Kha:
Have CI create a full-featured aarch64 build by way of clang cross-compilation, bundling an aarch64 LLVM build @gebner provided together with other aarch64 dependencies taken from cache.nixos.org.

In contrast to #734, we don't try to run the test suite via qemu, which does take a while at ~30min, and requires further setup with `apt` that this PR otherwise avoids. As architecture-specific bugs should be quite rare after initial support, I think that's fine. We could introduce a formal tier system of supported platforms, where Tier 1 is the current platforms that are tested & published, whereas Tier 2 is only compiled & published.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,31 +14,6 @@ This is the repository for **Lean 4**, which is currently being released as mile
 
 See [Setting Up Lean](https://leanprover.github.io/lean4/doc/setup.html).
 
-## Supported Platforms
-
-### Tier 1
-
-Platforms built & tested by our CI, available as nightly & stable releases via elan (see above)
-
-* x86-64 Linux with glibc 2.27+
-* x86-64 macOS 10.15+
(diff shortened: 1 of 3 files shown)


Kha:
> We could introduce a formal tier system of supported platforms, where Tier 1 is the current platforms that are tested & published, whereas Tier 2 is only compiled & published.

[Added](https://github.com/Kha/lean4/blob/aarch64/README.md#supported-platforms), WDYT?

EdAyers:
Yeah I saw this PR and it got me thinking. I've been happily using a native arm64 build on my mac, but it would be nice to have elan download the arm64 versions without needing to emulate. I managed to use my mac as a lean4 github-actions-runner compiling native arm64 binaries .

PR #1067, opened 2022-03-24
fix: `fun (x) ...` should not be treated as a pattern

Kha:
As found by @larsk21

### DIFF
--- a/src/Lean/Elab/Binders.lean
+++ b/src/Lean/Elab/Binders.lean
@@ -295,16 +295,28 @@ partial def expandFunBinders (binders : Array Syntax) (body : Syntax) : MacroM (
         if binderBody.isNone then
           processAsPattern ()
         else
-          let idents  := binderBody[0]
+          let term    := binderBody[0]
           let special := binderBody[1]
           if special.isNone then
-            processAsPattern ()
+            match (← getFunBinderIds? term) with
+            | some idents =>
+              -- `fun (x ...) ...` ~> `fun (x : _) ...`
+              -- Interpret `(x ...)` as sequence of binders instead of pattern only if none of the idents
+              -- are defined in the global scope. Technically, it would be sufficient to only check the
+              -- first ident to be sure that the syntax cannot possibly be a valid pattern. However, for
+              -- consistency we apply the same check to all idents so that the possibility of shadowing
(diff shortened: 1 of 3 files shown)


leodemoura:
> The first identifier will most likely be a compound identifier in these cases

If we are inside of a namespace, the constructor may not be a compound identifier. Example:

```lean
def And.comm : p ∧ q → q ∧ p :=
  fun (intro h₁ h₂) => intro h₂ h₁
```

Kha:
@leodemoura We now have global name resolution in macros, do you see any remaining edge cases if we only do it when the leading ident is not in the global context? Because I can certainly empathize with the current behavior being confusing.

leodemoura:
> @leodemoura We now have global name resolution in macros, do you see any remaining edge cases if we only do it when the leading ident is not in the global context? Because I can certainly empathize with the current behavior being confusing.

Yes, the error message for
```lean
#check fun (α) [Repr α] => repr (α := α)
```
is really bad. The one for the following examples makes it clear that the term is being interpreted as a pattern.
```lean
#check fun (α β) [Repr α] => repr (α := α)
```

> do you see any remaining edge cases if we only do it when the leading ident is not in the global context?

Not sure, I would have to think about it, but shadowing is the first thing that comes to mind. That is, given `(a_1 ... a_n)`, `a_2` ... `a_n` can shadow global declarations, but `a_1` cannot.

src/Lean/Elab/Binders.lean
```diff
@@ -295,16 +295,28 @@ partial def expandFunBinders (binders : Array Syntax) (body : Syntax) : MacroM (
...
           if special.isNone then
-            processAsPattern ()
+            match (← getFunBinderIds? term) with
+            | some idents =>
+              -- `fun (x ...) ...` ~> `fun (x : _) ...`
+              -- Interpret `(x ...)` as sequence of binders instead of pattern only if none of the idents
+              -- are defined in the global scope. Technically, it would be sufficient to only check the
+              -- first ident to be sure that the syntax cannot possibly be a valid pattern. However, for
+              -- consistency we apply the same check to all idents so that the possibility of shadowing
+              -- a global decl is identical for all of them.
```
Kha, line 308: Changed to this more conservative choice

PR #1069, opened 2022-03-24
doc: Docstrings for LocalContext.lean

EdAyers:


src/Lean/LocalContext.lean
```diff
@@ -112,7 +127,10 @@ def empty : LocalContext := {}
 def isEmpty (lctx : LocalContext) : Bool :=
   lctx.fvarIdToDecl.isEmpty
 
-/- Low level API for creating local declarations. It is used to implement actions in the monads `Elab` and `Tactic`. It should not be used directly since the argument `(name : Name)` is assumed to be "unique". -/
+/-- Low level API for creating local declarations.
+It is used to implement actions in the monads `Elab` and `Tactic`.
+It should not be used directly since the argument `(name : Name)` is
+assumed to be "unique". -/
```
EdAyers, line 133: What is the recommended way to push to LocalContext?

digama0: looks like `LocalContext.mkLetDecl` and `LocalContext.mkLocalDecl` are the (somewhat low level) functions for doing this

EdAyers: I was also wondering what is the bad thing that happens when two variables with the same userName are present in the context. Seems like one would just clobber the other?

digama0: The issue is not overlapping userNames, it is overlapping unique names / `FVarId`s. Unique names have to be (globally?) unique and can be constructed by `mkFreshFVarId`. If you have overlapping user names, the outer one will be shadowed and displayed with the dagger symbol.

leodemoura: > What is the recommended way to push to LocalContext?

We recommend users to use the `MetaM` functions: `withLocalDecl`, `withLetDecl`, `forallTelescope`, `lambdaTelescope`, etc. We use these functions all over the place.

leodemoura: > Looks like `LocalContext.mkLetDecl` and `LocalContext.mkLocalDecl` are the (somewhat low level) functions for doing this. (The documentation says it's not recommended but based on use it looks more like it is an "unsafe" function with a precondition for correctness. The function `withLocalDecl` avoids the dependence on unique names, but lots of places use `mkLocalDecl` if they can't satisfy the callback structuring requirements of `withLocalDecl`.) The higher level way would be to call `intro1P`, `introN` or one of the other derivatives.

Yes, we use `intro*` functions for tactic programming (directory: `src/Lean/Meta/Tactic`). Most of the functions there take a `MVarId` (representing the goal) as an argument.

leodemoura: Most users should not need to use the `LocalContext` functions directly.

PR #1072, opened 2022-03-24
doc: MetaM

EdAyers:


src/Lean/Meta/Basic.lean
```diff
@@ -1297,6 +1334,7 @@ def isExprDefEq (t s : Expr) : MetaM Bool :=
     trace[Meta.isDefEq] "{t} =?= {s} ... {if b then "success" else "failure"}"
     return b
 
+/-- Determines whether two expressions are definitionally equal to each other. -/
```
EdAyers, line 1337: Something I haven't managed to figure out yet is where the flags go to do unification and matching.

leodemoura: Which flags?

leodemoura:
I am merging since the question above does not affect the PR.

src/Lean/Meta/Basic.lean
EdAyers: I mean, how can I perform isDefEq without any mvar unification happening, or with only mvar assignment happening on rhs.

leodemoura: Metavariables and their context have a "depth". Given a metavariable `?m` and a `MetavarContext` `mctx`, `?m` is not assigned if `?m.depth != mctx.depth`. We also have a combinator `withNewMCtxDepth x` which will bump the `mctx` while executing `x`. So, `withNewMCtxDepth (<- isDefEq a b)` is `isDefEq` without any mvar assignment happening. For matching, we create the term inside the `withNewMCtxDepth`, here is an example from `simp`: https://github.com/leanprover/lean4/blob/master/src/Lean/Meta/Tactic/Simp/Rewrite.lean#L100-L106

PR #1078, opened 2022-03-26
Render examples using LeanInk+Alectryon

Kha:
Rendered: https://kha.github.io/lean4/doc/examples.html

/cc @insightmind @cpitclaudel :)

### DIFF
--- a/doc/examples/interp.lean
+++ b/doc/examples/interp.lean
@@ -22,7 +22,7 @@ inductive Vector (α : Type u) : Nat → Type u
   | cons : α → Vector α n → Vector α (n+1)
 
 /-|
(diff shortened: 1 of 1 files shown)


cpitclaudel:
Wonderful!

cpitclaudel:
Careful with reST syntax: you have to leave a blank after a backtick, so you need to write `` `Vector`\ s``, not `` `Vector`s `` 

https://kha.github.io/lean4/doc/examples/interp.lean.html has the warning related to that.

Kha:
@cpitclaudel Thanks! We definitely want to switch to Markdown eventually, but this is good for now.

PR #1083, opened 2022-03-29
doc: explain `MonadControl`

EdAyers:
I found MonadControl quite confusing to understand and it took a bit of digging to discover its equivalent in haskell. So I've added a link and ported [this tutorial](https://lexi-lambda.github.io/blog/2019/09/07/demystifying-monadbasecontrol/) to Lean.

It might be better to place this in the documentation section of Lean rather than as a `/-!` comment.

### DIFF
--- a/src/Init/Control/Basic.lean
+++ b/src/Init/Control/Basic.lean
@@ -66,11 +66,100 @@ infixr:35 " <&&> " => andM
 @[macroInline] def notM {m : Type → Type v} [Applicative m] (x : m Bool) : m Bool :=
   not <$> x
 
+/-!
+
+# How `MonadControl` works
+
+There is a [tutorial by Alexis King](https://lexi-lambda.github.io/blog/2019/09/07/demystifying-monadbasecontrol/) that this docstring is based on.
+
+Suppose we have `foo : ∀ α, IO α → IO α` and `bar : StateT σ IO β` (ie, `bar : σ → IO (σ × β)`).
+We might want to 'map' `bar` by `foo`. Concretely we would write this as:
+
+```lean
+def mapped_foo : StateT σ m β → StateT σ m β := do
+  let s ← get
+  let (s', v) ← lift <| foo <| StateT.run bar s
+  set s'
+  pure v
+```
+
+This is fine but it's not going to generalise, what if we replace `StateT Nat IO` with a large tower of monad transformers?
+We would have to rewrite the above to handle each of the `run` functions for each transformer in the stack.
+
+Is there a way to generalise `run` as a kind of inverse of `lift`?
+We have `lift : m α → StateT σ m α` for all `m`, but we also need to 'unlift' the state.
+But `unlift : StateT σ IO α → IO α` can't be implemented. So we need something else.
+
+If we look at the definition of `mapped_foo`, we see that `lift <| foo <| StateT.run bar s`
+has the type `IO (σ × β)`. The key idea is that `σ × β` contains all of the information needed to reconstruct the state and the new value.
+
+Now lets define some values to generalise `mapped_foo`:
+- Write `IO (σ × β)` as `IO (stM β)`
+- Write `StateT.run . s` as `mapInBase : StateT σ IO α → IO (stM β)`
+- Define `restoreM : IO (stM α) → StateT σ IO α` as `fun x => do let (s', v) ← lift x; set s'; pure v`
+
+To get
+
+```lean
+def mapped_foo : StateT σ m β → StateT σ m β := do
+  let s ← get
+  let mapInBase := fun z => StateT.run z s
+  restoreM <| foo <| mapInBase bar
+```
+
+and finally define
+
+```lean
+control : {α : Type u} → (({β : Type u} → StateT σ IO β → IO (stM β)) → IO (stM α)) → StateT σ IO α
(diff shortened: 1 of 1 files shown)


src/Init/Control/Basic.lean
```diff
@@ -66,11 +66,100 @@ infixr:35 " <&&> " => andM
...
+  let s ← get
+  let mapInBase := fun z => StateT.run z s
+  restoreM <| foo <| mapInBase bar
+```
+
+and finally define
+
+```lean
+control : {α : Type u} → (({β : Type u} → StateT σ IO β → IO (stM β)) → IO (stM α)) → StateT σ IO α
+  | α, f => do
```
Kha, line 114: I would move these bindings to the signature

src/Init/Control/Basic.lean
```diff
@@ -66,11 +66,100 @@ infixr:35 " <&&> " => andM
...
+  control fun mapInBase => foo (mapInBase bar)
+```
+
+The core idea of `mapInBase` is that given any `β`, it runs an instance of
+`StateT σ IO β` and 'packages' the result and state as  `IO (stM β)` so that it can be piped through `foo`.
+Once it's been through `foo` we can then unpack the state again with `restoreM`.
+Hence we can apply `foo` to `bar` without losing track of the state.
+
+Here `stM β = σ × β` is the 'packaged result state', but we can generalise:
+if we have a tower `StateT σ₁ <| StateT σ₂ <| IO`, then we can get the
```
Kha, line 134: grammar: remove "we can get"

src/Init/Control/Basic.lean
```diff
@@ -66,11 +66,100 @@ infixr:35 " <&&> " => andM
...
+
+Suppose we have `foo : ∀ α, IO α → IO α` and `bar : StateT σ IO β` (ie, `bar : σ → IO (σ × β)`).
+We might want to 'map' `bar` by `foo`. Concretely we would write this as:
+
+```lean
+def mapped_foo : StateT σ m β → StateT σ m β := do
+  let s ← get
+  let (s', v) ← lift <| foo <| StateT.run bar s
+  set s'
+  pure v
```
Kha, line 83: Could you please make sure that all (non-pseudo) examples are runnable as is? Especially since we might eventually want to test them automatically. There seem to be multiple issues in this code block.

src/Init/Control/Basic.lean
```diff
@@ -66,11 +66,100 @@ infixr:35 " <&&> " => andM
...
+Once it's been through `foo` we can then unpack the state again with `restoreM`.
+Hence we can apply `foo` to `bar` without losing track of the state.
+
+Here `stM β = σ × β` is the 'packaged result state', but we can generalise:
+if we have a tower `StateT σ₁ <| StateT σ₂ <| IO`, then we can get the
+composite packaged state is going to be `stM₁₂ β := σ₁ × σ₂ × β` or `stM₁₂ := stM₁ ∘ stM₂`.
+
+Now we can define `MonadControl m n`. Call `m` the 'base monad', in the above example it was `IO`.
+`MonadControl m n` means that we can lift a monad function `foo : m α → m α` in the base monad
+to a function `foo' : n α → n α` in the `n` monad, without having to manually handle the state of `n`.
```
Kha, line 139: It is crucial that the lifted function is all-quantified over the monadic return value, as that is where the surrounding monad's state is stored. After that change, note that this is exactly `MonadFunctor`. We don't currently have an instance of `MonadControl` implying `MonadFunctor`. I also don't think we have a monad that can be a `MonadFunctor` but not a `MonadControl`.

Kha:
I forgot to say, thanks for doing this!

src/Init/Control/Basic.lean
Kha: An example of a function that can be implemented with `MonadControl` but not with `MonadFunctor` is https://github.com/leanprover/lean4/blob/6dfddbe2e791847ff812db37a098c82a9cf8a648/src/Lean/Meta/Basic.lean#L257-L261

src/Init/Control/Basic.lean
```diff
@@ -66,11 +66,100 @@ infixr:35 " <&&> " => andM
...
+We have `lift : m α → StateT σ m α` for all `m`, but we also need to 'unlift' the state.
+But `unlift : StateT σ IO α → IO α` can't be implemented. So we need something else.
+
+If we look at the definition of `mapped_foo`, we see that `lift <| foo <| StateT.run bar s`
+has the type `IO (σ × β)`. The key idea is that `σ × β` contains all of the information needed to reconstruct the state and the new value.
+
+Now lets define some values to generalise `mapped_foo`:
+- Write `IO (σ × β)` as `IO (stM β)`
+- Write `StateT.run . s` as `mapInBase : StateT σ IO α → IO (stM β)`
+- Define `restoreM : IO (stM α) → StateT σ IO α` as `fun x => do let (s', v) ← lift x; set s'; pure v`
```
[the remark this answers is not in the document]
EdAyers: yeah I wasn't sure how to do this. problem is we have `(s : σ) ⊢ mapInBase`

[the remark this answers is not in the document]
Kha: Ahh, makes sense

PR #1086, opened 2022-03-30
doc: fix issues with MonadControl docs

EdAyers:
#1083 was merged before some of @kha's issues had been fixed.

### DIFF
--- a/src/Init/Control/Basic.lean
+++ b/src/Init/Control/Basic.lean
@@ -146,15 +146,15 @@ Here `stM β = σ × β` is the 'packaged result state', but we can generalise:
 if we have a tower `StateT σ₁ <| StateT σ₂ <| IO`, then the
 composite packaged state is going to be `stM₁₂ β := σ₁ × σ₂ × β` or `stM₁₂ := stM₁ ∘ stM₂`.
 
-Now we can define `MonadControl m n`. Call `m` the 'base monad', in the above example it was `IO`.
-`MonadControl m n` means that we can lift a monad function `foo : m α → m α` in the base monad
-to a function `foo' : n α → n α` in the `n` monad, without having to manually handle the state of `n`.
-Futhermore, `MonadControl` is transitive, which is what `MonadControlT` is, it's the transitive closure of `MonadControl`.
-
+`MonadControl m n` means that when programming in the monad `n`,
+we can switch to a base monad `m` using `control`, just like with `liftM`.
+In contrast to `liftM`, however, we also get a function `runInBase` that
+allows us to "lower" actions in `n` into `m`.
 This is really useful when we have large towers of monad transformers, as we do in the metaprogramming library.
 
 For example there is a function `withNewMCtxDepthImp : MetaM α → MetaM α` that runs the input monad instance
-in a new nested metavariable context. We can lift this to `withNewMctxDepth : n α → n α` using `MonadControlT MetaM n`.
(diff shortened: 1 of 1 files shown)


src/Init/Control/Basic.lean
```diff
@@ -131,7 +143,7 @@ Once it's been through `foo` we can then unpack the state again with `restoreM`.
 Hence we can apply `foo` to `bar` without losing track of the state.
 
 Here `stM β = σ × β` is the 'packaged result state', but we can generalise:
-if we have a tower `StateT σ₁ <| StateT σ₂ <| IO`, then we can get the
+if we have a tower `StateT σ₁ <| StateT σ₂ <| IO`, then the
 composite packaged state is going to be `stM₁₂ β := σ₁ × σ₂ × β` or `stM₁₂ := stM₁ ∘ stM₂`.
 
 Now we can define `MonadControl m n`. Call `m` the 'base monad', in the above example it was `IO`.
```
Kha, line 149: How about something like the following instead the sentence below, to differentiate it from `MonadFunctor`:
> `MonadControl m n` means that when programming in the monad `n`, we can switch to the base monad `m` using `control`, just like with `liftM`. In contrast to `liftM`, however, we also get a function `runInBase` that allows us to "lower" actions in `n` into `m`.

This is still quite simplified, it doesn't address `stM` or `restoreM`.

src/Init/Control/Basic.lean
```diff
@@ -146,12 +158,33 @@ in a new nested metavariable context. We can lift this to `withNewMctxDepth : n
...
+ @[inline] def map1MetaM [MonadControlT MetaM m] [Monad m] (f : forall {α}, (β → MetaM α) → MetaM α) {α} (k : β → m α) : m α :=
+   control fun runInBase => f fun b => runInBase <| k b
+
+ @[inline] def map2MetaM [MonadControlT MetaM m] [Monad m] (f : forall {α}, (β → γ → MetaM α) → MetaM α) {α} (k : β → γ → m α) : m α :=
+   control fun runInBase => f fun b c => runInBase <| k b c
+```
+
+In these examples, `MonadControl` is needed because the lifted function
+needs to be all-quantified over the monadic return value,
+as that is where the surrounding monad's (`n`) state is stored.
```
Kha, line 177: No, that's not the issue here as `f` *is* all-quantified. The difference is that with `MonadFunctor`, we can "lower" (in the above sense) exactly one, fixed computation in `n` into `m`, the `n α` that is an input parameter to `monadMap`. But in these two cases, we want to lower an arbitrary number of values under a lambda, so `monadMap` is not applicable.

src/Init/Control/Basic.lean
```diff
@@ -146,12 +158,33 @@ in a new nested metavariable context. We can lift this to `withNewMctxDepth : n
...
+In the Lean monad-transformer stacks, there are no `MonadFunctor`s that
+are not also `MonadControl`s and so `MonadFunctor` is not used.
```
Kha, line 179: Oh, `monadMap` *is* used in plenty places! It could be argued whether we still need `MonadFunctor`, but let's keep that discussion out of this docstring.

Kha:
Great!

src/Init/Control/Basic.lean
EdAyers: Yeah there is a link to a more detailed tutorial so it doesn't need to be super detailed.

Issue #528, opened 2021-06-13
Proposal for Stdlib Contributions [RFC]


tydeu:
The current standard library in the Lean core is lacking many common utility functions. One way to go about fleshing it out is to allow the community to contributed useful utilities. However, such contributions would likely be small and thus not really suited for the current multi-step RFC contributing process. Instead, it would be nice to be able to fast track such small changes. To help make the process less unwieldy, we can provide examples of which contributions would be welcome and which would not. For example, we probably would want functions like `List.nth`, but may wish to relegate syntactic extensions to the formal RFC process.

To give an idea of what we want to evolve Lean's standard library into, we can take cues from, for instance, F# and Haskell:

* [ ] [F#'s Option](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-optionmodule.html)
* [ ] [Haskell's Maybe](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Maybe.html)
* [ ] [F#'s String](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-core-stringmodule.html)
* [ ] [Haskell's String](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-String.html)
* [ ] [F#'s List](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-listmodule.html)
* [ ] [Haskell's List](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-List.html)
* [ ] [F#'s Map](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-mapmodule.html)
* [ ] [Haskel's Map](https://hackage.haskell.org/package/containers-0.6.4.1/docs/Data-Map-Strict.html)
* [ ] [F#'s Set](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-setmodule.html)
* [ ] [Haskell's Set](https://hackage.haskell.org/package/containers-0.6.4.1/docs/Data-Set.html)
* [ ] [F#'s Array](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-arraymodule.html)

**Note:** For the Haskell references, it should be observed that Haskell is (usually) lazy and Lean is strict, so some utilities will not translate.

DanielFabian:
@leodemoura, @Kha please tag this as `help wanted`, if you're happy with it, so we can invite outside contributions on the lib.

leodemoura:
[...]

PR #1092, opened 2022-04-03
feat: add theorem exist_idx_of_elem (List/Basic.lean)

casavaca:
Still learning Lean. Hopefully my naming is good and it doesn't conflict with future merging of Mathlib4.

### DIFF
--- a/src/Init/Data/List/Basic.lean
+++ b/src/Init/Data/List/Basic.lean
@@ -217,6 +217,18 @@ def notElem [BEq α] (a : α) (as : List α) : Bool :=
 abbrev contains [BEq α] (as : List α) (a : α) : Bool :=
   elem a as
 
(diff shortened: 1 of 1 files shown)


leodemoura:
Thanks for the PR, but it is not a good time for merging this kind of theorem. It is not clear to me where it should go: Lean 4 main repo; Std library; or Mathlib 4.
The Std library plans are currently on hold (see https://github.com/leanprover/lean4/issues/528). We are currently focusing on porting Mathlib to Lean 4, and fixing issues and missing features. I think this theorem would be a good match for the Std library that will be maintained in a separate repository. 
Right now, we only want theorems in the main repo that fall in one of the following categories:
- They are required for bootstrapping the system and implementing built-in tactics.
- They are useful for writing tutorials and manuals.
- They state properties of built-in data structures, and users need them all the time when writing basic examples using Lean 4.

In the meantime, I suggest you contribute this theorem to the Mathlib 4 repo.

PR #1093, opened 2022-04-03
Embed examples into mdBook

Kha:
This PR switches Alectryon to my fork of Niklas' fork, which adds a hacky new pipeline for rendering code to HTML while leaving literate comments as is to be processed by mdBook instead. Thus literate comments now use the same Markdown as other parts of the docs.

While we're at it, I tried to get the Alectryon CSS a little closer to mdBook's, and also fixed some of the easier issues reported by @gebner:
* add delay before showing type hover (unsure if it should apply to goal hovers as well), always activate type hovers
* allow line breaks in type hovers, fix horizontal overflow

Rendered: https://kha.github.io/lean4/doc/examples/palindromes.lean.html

### DIFF
--- a/doc/alectryon.css
+++ b/doc/alectryon.css
@@ -68,7 +68,8 @@ SOFTWARE.
 .alectryon-mref,
 .alectryon-block, .alectryon-io,
 .alectryon-toggle-label, .alectryon-banner {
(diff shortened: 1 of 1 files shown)


Kha:
Notable regression from Alectryon/docutils: mdBook does not highlight inline code. But using two different highlighters on the same page might not be preferable anyway.

PR #1097, opened 2022-04-08
feat: add IO.monoNanosNow

### LINKED ISSUE #1095: Higher-resolution `monoMsNow`
I'd like to have a `monoNsNow` (with nanosecond resolution) in addition to the existing `monoMsNow` (with millisecond resolution). This is for profiling Aesop: I want to be sure that if I measure the cumulative elapsed time of different parts of the tactic, the results are not dominated by rounding errors.

If you agree this is desirable, I'd be happy to send a PR.

JLimperg:
Closes #1095

---

I chose this name because I thought `monoNsNow` is too close to `monoMsNow`.

Is `usize_to_nat` correct on 32-bit systems? Afaiu, a `nanoseconds` duration is represented by at least 64 bits.

### DIFF
--- a/src/Init/System/IO.lean
+++ b/src/Init/System/IO.lean
@@ -165,6 +165,9 @@ def lazyPure (fn : Unit → α) : IO α :=
 /-- Monotonically increasing time since an unspecified past point in milliseconds. No relation to wall clock time. -/
 @[extern "lean_io_mono_ms_now"] constant monoMsNow : BaseIO Nat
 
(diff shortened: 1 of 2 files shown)


Kha:
I think both functions should use `uint64_to_nat` to be safe. We could add a `static_assert(sizeof(std::chrono::nanoseconds::rep) <= sizeof(uint64))` if we want to be sure.

JLimperg:
Okay, I've made the corresponding changes.

PR #1103, opened 2022-04-10
fix: make `userRpcProcedures` use a persistent environment extension

EdAyers:
Currently, the `userRpcProcedures` environment extension is not persistent, which means that you can't use user rpcs from imported files. This is an attempt to fix that, although I am not sure that this is the best way to do it.

### DIFF
--- a/src/Lean/Server/Rpc/RequestHandling.lean
+++ b/src/Lean/Server/Rpc/RequestHandling.lean
@@ -98,8 +98,8 @@ def registerBuiltinRpcProcedure (method : Name) paramType respType
   let proc := wrapRpcProcedure method paramType respType handler
   builtinRpcProcedures.modify fun ps => ps.insert method proc
 
-open Lean.Elab.Command in
-def registerRpcProcedure (method : Name) : CommandElabM Unit := do
+open Lean Elab Command Term Meta in
+def registerRpcProcedure (method : Name) : CoreM Unit := do
   let env ← getEnv
   let errMsg := "Failed to register RPC call handler for '{method}'"
   if (←builtinRpcProcedures.get).contains method then
@@ -107,8 +107,18 @@ def registerRpcProcedure (method : Name) : CommandElabM Unit := do
   if userRpcProcedures.contains env method then
     throwError s!"{errMsg}: already registered"
   let wrappedName := method ++ `_rpc_wrapped
-  let d ← `(def $(mkIdent wrappedName) : RpcProcedure := wrapRpcProcedure $(quote method) _ _ $(mkIdent method))
-  elabCommand d
+  let procT := mkConst ``RpcProcedure
+  let proc ← MetaM.run' <| TermElabM.run' <| do
+     let c ← Lean.Elab.Term.elabTerm (← `(wrapRpcProcedure $(quote method) _ _ $(mkIdent method))) procT
+     return ← instantiateMVars c
+  addAndCompile <| Declaration.defnDecl {
+        name        := wrappedName
+        type        := procT
+        value       := proc
+        safety      := DefinitionSafety.safe
+        levelParams := []
+        hints := ReducibilityHints.opaque
+      }
   setEnv <| userRpcProcedures.insert (← getEnv) method wrappedName
 
 end Lean.Server


Vtec234:
I would like some feedback from the core team on this, if possible. We wish to have a method in `CommandElabM` which registers a user-defined method `foo` as an RPC procedure to be accessed from the `Environment` later, and which we will want to then execute. Before being executed, `foo` must at some point be transformed into a generic `RpcProcedure` closure. It seems like there are at least two ways to do this:

1. is implemented here and achieves this by making an auxiliary `def` that has the type `RpcProcedure`. We must do this as opposed to storing the `RpcProcedure` in env extension state because closures cannot be serialized into .oleans.
2. would be to just use a "tag" extension with a list of tagged `def`s in the state and then construct the `RpcProcedure` objects on import.

The former approach seems preferrable since it doesn't impose as much extra `import` time, but maybe there is a third way? Also I am just not sure which of the many `EnvExtension` variants to use here.

src/Lean/Server/Rpc/RequestHandling.lean
```diff
@@ -35,12 +36,19 @@ private def handleRpcCall (p : Lsp.RpcCallParams) : RequestM (RequestTask Json)
     fun snap => do
       if let some proc := (← builtinRpcProcedures.get).find? p.method then
         proc.wrapper p.sessionId p.params
-      else if let some proc := userRpcProcedures.getState snap.env |>.find? p.method then
-        proc.wrapper p.sessionId p.params
+      else if let some procName := userRpcProcedures.find? snap.env p.method then
+        let options := snap.cmdState.scopes.head!.opts
+        let proc : Except _ _ := Lean.Environment.evalConstCheck RpcProcedure snap.env options ``RpcProcedure procName
```
Vtec234, line 41: Could it be a problem that we `evalConstCheck` on every RPC call? Supposing the imported .olean has an associated shared library, the interpreter would jump to fast native code here, right?

src/Lean/Server/Rpc/RequestHandling.lean
```diff
@@ -87,21 +95,17 @@ def registerBuiltinRpcProcedure (method : Name) paramType respType
...
+open Lean.Elab.Command in
+def registerRpcProcedure (method : Name) : CommandElabM Unit := do
   let env ← getEnv
-
   let errMsg := "Failed to register RPC call handler for '{method}'"
   if (←builtinRpcProcedures.get).contains method then
     throwError s!"{errMsg}: already registered (builtin)"
-  if userRpcProcedures.getState env |>.contains method then
+  if userRpcProcedures.contains env method then
     throwError s!"{errMsg}: already registered"
```
Vtec234, line 108: This might cause trouble. When two imported modules register the same name, do both `registerRpcProcedure` calls succeed and store the same name in extension state? At what point do we see a conflict?

src/Lean/Server/Rpc/RequestHandling.lean
```diff
@@ -87,21 +98,17 @@ def registerBuiltinRpcProcedure (method : Name) paramType respType
   let proc := wrapRpcProcedure method paramType respType handler
   builtinRpcProcedures.modify fun ps => ps.insert method proc
 
-def registerRpcProcedure (method : Name) paramType respType
-    {paramLspType} [RpcEncoding paramType paramLspType] [FromJson paramLspType]
-    {respLspType} [RpcEncoding respType respLspType] [ToJson respLspType]
-    (handler : paramType → RequestM (RequestTask respType)) : CoreM Unit := do
+open Lean.Elab.Command in
+def registerRpcProcedure (method : Name) : CommandElabM Unit := do
```
Vtec234, line 102: We could previously register a function `foo` under an arbitrary `method : Name`. I think this approach is actually better because fully resolved constant names are already guaranteed unique (right?), so that should avoid any possibility of conflict here.

EdAyers:
> We wish to have a method in `CommandElabM` which registers a user-defined method `foo` as an RPC procedure to be accessed from the `Environment` later

Also some user RPC procedures will need to be called every time the user changes cursor position, so it's important to be fast or cached in some way. In practice I've found  that spamming the server with user RPCs can be slow enough to cause unresponsive behaviour. But I have not profiled this and maybe evalConst is not the bottleneck

src/Lean/Server/Rpc/RequestHandling.lean
leodemoura: Yes, fully resolved constant names are unique.

Vtec234 (approved):

PR #1110, opened 2022-04-15
Accept unregistered syntax kinds in stage 1

Kha:
...so that we can add a built-in parser and elaborator in the same stage. Also add the change that finally made me implement this as a proof of concept.

### DIFF
--- a/src/Lean/Elab/BuiltinNotation.lean
+++ b/src/Lean/Elab/BuiltinNotation.lean
@@ -131,7 +131,7 @@ At runtime, `msg` and the file position are printed to stderr unless the C
 function `lean_set_panic_messages(false)` has been executed before. If the C
 function `lean_set_exit_on_panic(true)` has been executed before, the process is
 then aborted. -/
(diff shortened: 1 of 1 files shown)


src/Lean/Elab/BuiltinNotation.lean
```diff
@@ -131,7 +131,7 @@ At runtime, `msg` and the file position are printed to stderr unless the C
 function `lean_set_panic_messages(false)` has been executed before. If the C
 function `lean_set_exit_on_panic(true)` has been executed before, the process is
 then aborted. -/
-@[builtinTermElab panic] def elabPanic : TermElab := fun stx expectedType? => do
+@[builtinTermElab Lean.Parser.Term.panic] def elabPanic : TermElab := fun stx expectedType? => do
```
Kha, line 134: @leodemoura minor regression where it accidentally resolved to `_root_.panic`, but since the change is purely internal, I don't think this is even worth a check whether the resolved constant is indeed a parser.

PR #1114, opened 2022-04-22
fix: preserve type of `let _ : t`

Kha:
This fixes a regression @larsk21 noticed when updating the Lean version of https://github.com/gebner/quote4. In https://github.com/gebner/quote4/blob/7bc102c32a5e44b37e918474aabb58c19484d6eb/Qq/ForLean/ReduceEval.lean#L32-L33, there is a `let _` binding a typeclass instance, but the inferred type of the instance and the specified type of the `let` are equal only  at default transparency mode. Thus if we don't preserve the specified type, later typeclass inference fails in this case.

There might be other possible fixes, I'm not sure if there is any meaningful difference between them.

### DIFF
--- a/src/Lean/Elab/Binders.lean
+++ b/src/Lean/Elab/Binders.lean
@@ -648,7 +648,7 @@ def elabLetDeclCore (stx : Syntax) (expectedType? : Option Expr) (useLetExpr : B
         `(match $val:term with | $pat => $body)
       else
         let type := optType[0][1]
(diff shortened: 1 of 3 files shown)


leodemoura:
I used a different fix. The fix above makes sure that `let _ := val` and `let _ : type := val` are interpreted as `letIdDecl` instead of `letPatDecl` as most users would expect. It also makes sure we don't introduce auxiliary let-declarations into the local context, nor rely on optimizations used in the `match` elaborator.

PR #1115, opened 2022-04-22
Minor documentation change: add amssymb package to latex example

vincentdehaan:
Although the example on this page works, it stops working once one uses some special characters like ℤ. With this fix it works for all characters that are supported by `lstlean.tex`.


Kha:
Thanks!

PR #1121, opened 2022-04-24
feat: Add nix build extraDrvArgs

Anderssorby:
This is useful in some cases to send extra env args.
In use for this https://github.com/yatima-inc/Unicode.lean


Kha:
This is a pretty simple extension, but extending *all* derivations unconditionally sounds wrong to me. I would rather like to see an (overlay?) system that allows you to override individual/all module derivations. Then you could even extend the `src` of modules with further files, obsoleting the need for these environment variables.

Anderssorby:
Sure if you want to write that, but this is simply a way to give users more control. I feel that shouldn't be controversial.

Kha:
Closing as per above

PR #1125, opened 2022-04-27
fix: use correct number of `none` patterns

### LINKED ISSUE #1124: syntax quotation produces malformed expression
### Prerequisites

### Description

```lean
syntax "foo" (ident ident)? : term

variable (x y : Option Syntax)
#check `(foo $[$x:ident $y:ident]?)
```
Not sure the best way to show this, but if you `set_option trace.Elab.step true` you will see the result of macro expansion is
```lean
  Bind.bind✝ MonadRef.mkInfoFromRefPos✝
    (fun info✝ =>
      Bind.bind✝ getCurrMacroScope✝
        (fun scp✝ =>
          Bind.bind✝ getMainModule✝
            (fun mainModule✝ =>
              Pure.pure✝
                (Syntax.node✝ SourceInfo.none✝ `Mathport.Translate.Tactic.termFoo____
                  (List.toArray✝
                    (List.cons✝ (Syntax.atom✝ info✝ "foo")
                      (List.cons✝
                        (Syntax.node✝ SourceInfo.none✝ `null
                          (Array.append✝ (List.toArray✝ List.nil✝)
                            (match x, y with
                            | some✝ x, some✝ y => List.toArray✝ (List.cons✝ x (List.cons✝ y List.nil✝))
                            | none✝ => Array.empty✝)))
                        List.nil✝)))))))
```
and in particular the last part
```
match x, y with
| some x, some y => #[x, y]
| none => #[]
```
has a different number of patterns in the two branches, leading to a type error when it tries to understand `#[]` as having type `Option Syntax -> Array Syntax`.

Observed on `Lean (version 4.0.0-nightly-2022-04-26, commit 3cf642688bcf, Release)`

digama0:
fixes #1124

### DIFF
--- a/tests/lean/run/1124.lean
+++ b/tests/lean/run/1124.lean
@@ -1,0 +1,5 @@
+open Lean
+syntax "foo" (ident ident)? : term
+
+variable (x y : Option Syntax)
+example : MacroM Syntax := `(foo $[$x:ident $y:ident]?)


Kha:
Thanks!

Issue #1119, opened 2022-04-24
goal state and `<;>` combinator


fpfu:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

```lean
example (as bs cs : List α) : (as ++ bs) ++ cs = as ++ (bs ++ cs) := by
induction as <;> skip <;> (try rename_i h; simp[h]) <;> rfl
```

**Actual behavior:** 

at the beginning and at the end of the first `<;>` i get:

```
α : Type ?u.7
as bs cs : List α
⊢ as ++ bs ++ cs = as ++ (bs ++ cs)
```

at the end of `skip` I get:

```
case cons
α : Type ?u.7
bs cs : List α
head✝ : α
tail✝ : List α
 : tail✝ ++ bs ++ cs = tail✝ ++ (bs ++ cs)
⊢ head✝ :: tail✝ ++ bs ++ cs = head✝ :: tail✝ ++ (bs ++ cs)
```

and at the end of `(try rename_i h; simp[h])` I get 

```Goals accomplished 🎉```

**Reproduces how often:** 100%

### Versions
`nightly-2022-04-24`

**Comment by digama0** [comment]
What's the expected behavior? This issue doesn't show anything other than current behavior.

**Comment by fpfu** [comment]
in the first and second case I expect to see the same state as in the end of `induction as`:

```
case nil
α : Type ?u.7
bs cs : List α
⊢ [] ++ bs ++ cs = [] ++ (bs ++ cs)
case cons
α : Type ?u.7
bs cs : List α
head✝ : α
tail✝ : List α
 : tail✝ ++ bs ++ cs = tail✝ ++ (bs ++ cs)
⊢ head✝ :: tail✝ ++ bs ++ cs = head✝ :: tail✝ ++ (bs ++ cs)
```

in the third case I expect to see the remaining goal (as in the beginning of last `<;>`):
```
case nil
α : Type ?u.7
bs cs : List α
⊢ [] ++ bs ++ cs = [] ++ (bs ++ cs)
```

**Comment by fpfu** [comment]
consider you are in the middle of writing such a proof

at that intermediate state case `nil` is hidden from you:

```lean
example (as bs cs : List α) : (as ++ bs) ++ cs = as ++ (bs ++ cs) := by
induction as <;> skip
```

[...]

PR #1128, opened 2022-04-28
Revise tactic info selection

Kha:
Lean's info tree contains the initial and final states of all tactics and tactic combinators executed in a tactic block, but there is some ambiguity as to which state we should show at which location. In particular, in contrast to Lean 3 we would like to show *all* intermediate states when hovering over a tactic that is run multiple times by a combinator such as `repeat`. I implemented a new heuristic in https://github.com/leanprover/lean4/commit/eb7bf2b272862940832ed63a5892a5c261fbe446 that forced the info view to show only the last such state *after* e.g. `repeat tac`, but this turned out to be less desirable inside combinator chains as in #1119.

This PR reverts the above commit, replacing it with a different but more general heuristic that also solves many other deficiencies: tactic info selection is now indentation-sensitive. Thus, while at the *end* of something like `next => repeat b` it is hopelessly ambiguous on what level the user wants to continue the proof, so we show the innermost infos, the repeated executions of `b` as before, if the user first inserts a line break, we select the info of `next`, `repeat`, or `b` depending on the indentation of the new line. Crucially, this also solves the issue of returning to the main goal after finishing a `have` sub-proof: as soon as you deindent back to the level of the `have` keyword, the main goal will be shown.

### DIFF
--- a/src/Lean/Data/Json/Parser.lean
+++ b/src/Lean/Data/Json/Parser.lean
@@ -50,17 +50,15 @@ partial def strCore (acc : String) : Parsec String := do
     return acc
   else
     let c ← anyChar
-    let ec ←
-      if c = '\\' then
-        escapedChar
-      -- as to whether c.val > 0xffff should be split up and encoded with multiple \u,
-      -- the JSON standard is not definite: both directly printing the character
-      -- and encoding it with multiple \u is allowed. we choose the former.
-      else if 0x0020 ≤ c.val ∧ c.val ≤ 0x10ffff then
-        pure c
-      else
-        fail "unexpected character in string"
-    strCore (acc.push ec)
+    if c = '\\' then
+      strCore (acc.push (← escapedChar))
+    -- as to whether c.val > 0xffff should be split up and encoded with multiple \u,
+    -- the JSON standard is not definite: both directly printing the character
+    -- and encoding it with multiple \u is allowed. we choose the former.
+    else if 0x0020 ≤ c.val ∧ c.val ≤ 0x10ffff then
+      strCore (acc.push c)
+    else
+      fail "unexpected character in string"
 
 def str : Parsec String := strCore ""
 


tests/lean/interactive/infoIssues.lean
```diff
@@ -1,28 +1,30 @@
...
   match x with
   | 0 => rfl
-  | y+1 => repeat rw [Nat.zero_add] /- Good: We get the correct behavior here: Goals accomplished. -/
-       --^ Bad: get the `match` tactic initial state here.
-  -- Bad: we get "Goals accomplished" here
+                                 -- Bad? We still get all states except the one closed by implicit `rfl`
+                                 --v $/lean/plainGoal
+  | y+1 => repeat rw [Nat.zero_add]
+       --^ $/lean/plainGoal
+       -- Bad: get the `match` tactic initial state here.
```
Kha, line 28: NB: this is an orthogonal issue in the info tree, not info selection

Kha: I have an in-progress solution to #1119 that should be applicable here as well

src/Lean/Server/InfoUtils.lean
```diff
@@ -25,23 +25,31 @@ def Lean.Syntax.getRange? (stx : Syntax) (originalOnly := false) : Option String
...
-        | some a => [a]
-        | _      => []
+    match ctx? with
+      | some ctx => p ctx i cs cs'
       | _        => []
   | _ => []
 
+/--
+  For every branch, find the deepest node in that branch matching `p`
+  with a surrounding context (the innermost one) and return all of them. -/
```
Vtec234, line 43: I found this a little difficult to understand.
```suggestion
  For every branch of the `InfoTree`, find the deepest node in that branch for which `p` returns
  `some _`  and return the union of all such nodes. The visitor `p` is given a node together with
  its innermost surrounding `ContextInfo`. -/
```

PR #1133, opened 2022-05-02
feat: support inductive and parametrised types in `RpcEncoding`

Vtec234:
Plus some assorted changes to the `Widget` module, and documentation of `Meta` stuff. This needs an `update-stage0` for the new `deriving` handler to be registered.

### DIFF
--- a/src/Lean/Server/Rpc/Deriving.lean
+++ b/src/Lean/Server/Rpc/Deriving.lean
@@ -204,7 +204,7 @@ private def deriveInductiveInstance (indVal : InductiveVal) (params : Array Expr
         let mkBody (tgt : Name) (func : Name) : TermElabM Syntax := do
           let items ← nms.mapM fun nm => `(← $(mkIdent func) $nm)
           let tm := Syntax.mkApp (mkIdent <| Name.mkStr tgt ctor.getString!) items
-          `(Parser.Term.termReturn| return $tm:term)
+          `(return $tm:term)
 
         let encArm ← `(matchF| | $(mkPattern indVal.name):term => $(← mkBody packetNm ``rpcEncode))
         let decArm ← `(matchF| | $(mkPattern packetNm):term => $(← mkBody indVal.name ``rpcDecode))


src/Lean/Server/Rpc/Deriving.lean
```diff
@@ -16,103 +17,260 @@ namespace Lean.Server.RpcEncoding
...
         rpcDecode := decode }
-    end $typeId:ident
+    end
   )
   elabCommand cmds
   return true
 
-/-- Creates an `RpcEncodingPacket` structure for `α` with all fields of `α` replaced
-by their `RpcEncoding` and uses that to instantiate `RpcEncoding α LspPacket`. -/
+-- TODO(WN): Move these to Meta somewhere?
```
Vtec234, line 47: Just bringing attention to this. These seem generally useful, and maybe they are worth moving to `Meta` (where?).

Vtec234:
(I expect some of the builds will time out, it seems 09e4c00c681a6e4fd822bd5a5c213af943a8c009 introduced this issue.)

src/Lean/Elab/Deriving/FromToJson.lean
```diff
@@ -28,7 +28,7 @@ def mkToJsonInstanceHandler (declNames : Array Name) : CommandElabM Bool := do
         let fields := getStructureFieldsFlattened (← getEnv) declNames[0] (includeSubobjectFields := false)
         let fields : Array Syntax ← fields.mapM fun field => do
           let (isOptField, nm) := mkJsonField field
-          if isOptField then `(opt $nm $(mkIdent <| header.targetNames[0] ++ field))
+          if isOptField then `($(mkIdent ``Json.opt) $nm (some $(mkIdent <| header.targetNames[0] ++ field)))
```
Kha, line 31: This looks wrong; if `opt` is in scope, this shouldn't be necessary

Kha: Correct lexical binding (i.e. "hygiene") requires that they are resolved at macro-definition site only

Vtec234: Ah, I remembered now why I prefer the `mkIdent`. It's true that the name is resolved using macro-definition scopes, but it is resolved at macro-usage *time* since Lean does not run elaboration or any non-syntactic checks when defining the macro. So I am able to write down something incorrect there like `oppt`, and it will define fine but will fail at usage time. On the other hand ```mkIdent ``opt``` would fail thanks to the ``` `` ```. But I don't strongly prefer checking, I know it's correct in this case.

Kha (approved):
I see that there are a few other ```$(mkIdent ``...)``` inside non-term quotations. Not sure what the best solution there is. We could at least on-hover/highlighting hints so that developers can be sure identifiers (may) resolve to global declarations in scope *at the time of writing*.

I also replaced explicit `mkNode` invocations with quotations, just to be sure I wasn't missing anything. The need for `Syntax.mkApp` is a bit annoying. I think it would be nicer to macro authors to guarantee that a `Term.app` node without arguments doesn't misbehave /cc @leodemoura.

PR #1140, opened 2022-05-03
Updated deprecated syntax category names

crabbo-rave:
Updated deprecated name `strLit ` and `numLit`  to `str`  and `num`


Kha:
Thanks, I pushed a different fix that removes the duplication between the snippets and the full code

PR #1141, opened 2022-05-03
feat: change "⊢" in conv goals to "|"

RobinBoehne:
In conv mode, we rewrite (and thus change) the original goal, the "|" signifies this. This is similar to Lean3.

Suggested by @javra.

### DIFF
--- a/src/Lean/Widget/InteractiveGoal.lean
+++ b/src/Lean/Widget/InteractiveGoal.lean
@@ -138,6 +138,7 @@ def goalToInteractive (mvarId : MVarId) : MetaM InteractiveGoal := do
       | Name.anonymous => none
       | name           => some <| toString name.eraseMacroScopes
     let mut pref := "⊢ "
+    -- use special prefix for `conv` goals
     if isLHSGoal? mvarDecl.type |>.isSome then pref := "| "
     return { hyps, type := goalFmt, userName?, goalPrefix := pref }
 


src/Lean/Widget/InteractiveGoal.lean
```diff
@@ -136,6 +137,8 @@ def goalToInteractive (mvarId : MVarId) : MetaM InteractiveGoal := do
     let userName? := match mvarDecl.userName with
       | Name.anonymous => none
       | name           => some <| toString name.eraseMacroScopes
-    return { hyps, type := goalFmt, userName? }
+    let mut pref := "⊢"
+    if isLHSGoal? mvarDecl.type |>.isSome then pref := "|"
```
Kha, line 141: It would be interesting to make this choice extensible in the future -- for example, for Iris, we may want to hide it completely so that we can append our own contexts (/cc @larsk21). But I am fine with this implementation for this PR.

src/Lean/Widget/InteractiveGoal.lean
```diff
@@ -136,6 +137,8 @@ def goalToInteractive (mvarId : MVarId) : MetaM InteractiveGoal := do
...
+    return { hyps, type := goalFmt, userName?, goalPrefix := pref}
```
Kha, line 142:
```suggestion
    return { hyps, type := goalFmt, userName?, goalPrefix := pref }
```

src/Lean/Widget/InteractiveGoal.lean
Kha: Given the above, however, it would be good to include the trailing space in `goalPrefix` so that we don't have to change the client again when we allow arbitrary and potentially empty prefixes.

src/Lean/Widget/InteractiveGoal.lean
```diff
@@ -136,6 +137,8 @@ def goalToInteractive (mvarId : MVarId) : MetaM InteractiveGoal := do
     let userName? := match mvarDecl.userName with
       | Name.anonymous => none
       | name           => some <| toString name.eraseMacroScopes
-    return { hyps, type := goalFmt, userName? }
+    let mut pref := "⊢ "
+    if isLHSGoal? mvarDecl.type |>.isSome then pref := "| "
```
Kha, line 141:
```suggestion
    -- use special prefix for `conv` goals
    if isLHSGoal? mvarDecl.type |>.isSome then pref := "| "
```
Let's add a quick comment here since it may not be immediately clear what `isLHSGoal?` is testing.

Kha (approved):
LGTM otherwise. If there are no concerns from the client side, let's merge this.

KaseQuark:
The vscode-lean4 devs asked for `goalPrefix` to be optional (leanprover/vscode-lean4#171). Should we make it optional on this side too?

gebner:
I wonder if we should also change `Lean.Meta.ppGoal`.  AFAICT this change only affects the interactive diagnostics in vscode/neovim, but emacs and the command-line output still use ⊢.

Kha:
@gebner Yes, we should; ff6537be1b638e746911a550e3c4b66630b5885f

PR #1146, opened 2022-05-08
Turn on info trees by default

Kha:
Makes sure that linters dependent on the info tree work in any context. Info trees were previously turned on by `--server` or `--ilean`, i.e. almost always in practice anyways. The configuration option is left in there in case there is any benefit to turning them off for external users.

/cc @larsk21


leodemoura:
@Kha Did you check whether it affects the build time?

Kha:
It shouldn't, we already activate info trees for .ilean output

PR #1151, opened 2022-05-12
Fix incorrect JSON encoding of optional fields

Vtec234:
Restore the correct JSON encoding for optional fields, and use the new `deriving RpcEncoding` handler in the server. /cc @leodemoura


leodemoura:
Thanks!

PR #1152, opened 2022-05-16
fix: info at pattern variables

Kha:
Adding info nodes in `collectPatternVars` is not necessary as patterns get elaborated later anyways, and it is imprecise as it would analyze pattern variables as references, not binders (see new test).

### DIFF
--- a/tests/lean/1018unknowMVarIssue.lean.expected.out
+++ b/tests/lean/1018unknowMVarIssue.lean.expected.out
@@ -36,18 +36,15 @@ a : α
     n : <failed-to-infer-type> @ ⟨8, 2⟩-⟨10, 19⟩ @ Lean.Elab.Term.elabMatch
     x : Fam2 α β @ ⟨8, 8⟩-⟨8, 9⟩
     [.] `Fam2.any : none @ ⟨9, 4⟩-⟨9, 12⟩
-    @Fam2.any : {α : Type} → Fam2 α α @ ⟨9, 4⟩-⟨9, 12⟩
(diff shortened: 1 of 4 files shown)


tests/lean/1018unknowMVarIssue.lean.expected.out
```diff
@@ -36,18 +36,15 @@ a : α
...
     @Fam2.any : {α : Type} → Fam2 α α @ ⟨9, 4⟩-⟨9, 12⟩
     [.] `Fam2.nat : none @ ⟨10, 4⟩-⟨10, 12⟩
     Fam2.nat : Nat → Fam2 Nat Nat @ ⟨10, 4⟩-⟨10, 12⟩
     [.] `n : none @ ⟨10, 13⟩-⟨10, 14⟩
     [.] `Fam2.any : none @ ⟨9, 4⟩-⟨9, 12⟩
-    @Fam2.any : {α : Type} → Fam2 α α @ ⟨9, 4⟩-⟨9, 12⟩
     [.] `Fam2.any : none @ ⟨9, 4⟩-⟨9, 12⟩
     @Fam2.any : {α : Type} → Fam2 α α @ ⟨9, 4⟩-⟨9, 12⟩
     [.] `a : none @ ⟨8, 2⟩†-⟨10, 19⟩†
-    a : α @ ⟨8, 2⟩†-⟨10, 19⟩†
```
Kha, line 50: Note that there is another info node for `a` with the correct `isBinder := true` below

PR #1154, opened 2022-05-17
feat: `@[serverRpcMethod]` attribute

EdAyers:
Functions can now be marked with the `@[rpc]` attribute, which
registers the function as a user-made RpcProcedure that can be used to
communicate with the code editor and infoview.


Kha:
Since attribute names are not namespaced, should this have a less ambiguous name that more explicitly ties it to the language server?

EdAyers:
`@[rpcMethod]` or maybe `@[serverRpcMethod]`?

Kha:
`serverRpcMethod` sounds good to me! I guess we won't have to type it all that often.

PR #1157, opened 2022-05-18
feat: add RPC call to retrieve defn/decl/type defn

Vtec234:
This is used in leanprover/vscode-lean4#180. The RPC signature is not specific to VSCode, so should also work in nvim, I think. Also adds subexpression positions to `CodeToken` to be used by interactive tactics which need to generate a path to a subexpression.

### DIFF
--- a/src/Lean/Server/FileWorker/RequestHandling.lean
+++ b/src/Lean/Server/FileWorker/RequestHandling.lean
@@ -13,6 +13,7 @@ import Lean.Server.FileWorker.Utils
 import Lean.Server.Requests
 import Lean.Server.Completion
 import Lean.Server.References
+import Lean.Server.GoTo
 
 import Lean.Widget.InteractiveGoal
 
@@ -57,10 +58,6 @@ def handleHover (p : HoverParams)
 
       return none
 
-inductive GoToKind
-  | declaration | definition | type
-  deriving BEq, ToJson, FromJson
-
 open Elab GoToKind in
 def locationLinksOfInfo (kind : GoToKind) (ci : Elab.ContextInfo) (i : Elab.Info)
     (infoTree? : Option InfoTree := none) : RequestM (Array LocationLink) := do
@@ -68,33 +65,8 @@ def locationLinksOfInfo (kind : GoToKind) (ci : Elab.ContextInfo) (i : Elab.Info
   let doc ← readDoc
   let text := doc.meta.text
 
-  let documentUriFromModule (modName : Name) : MetaM (Option DocumentUri) := do
-    let some modFname ← rc.srcSearchPath.findModuleWithExt "lean" modName
-      | pure none
-    -- resolve symlinks (such as `src` in the build dir) so that files are opened
-    -- in the right folder
-    let modFname ← IO.FS.realPath modFname
-    pure <| some <| Lsp.DocumentUri.ofPath modFname
-
-  let locationLinksFromDecl (i : Elab.Info) (n : Name) := do
-    let mod? ← findModuleOf? n
-    let modUri? ← match mod? with
-      | some modName => documentUriFromModule modName
-      | none         => pure <| some doc.meta.uri
-
-    let ranges? ← findDeclarationRanges? n
-    if let (some ranges, some modUri) := (ranges?, modUri?) then
-      let declRangeToLspRange (r : DeclarationRange) : Lsp.Range :=
-        { start := ⟨r.pos.line - 1, r.charUtf16⟩
-          «end» := ⟨r.endPos.line - 1, r.endCharUtf16⟩ }
-      let ll : LocationLink := {
-        originSelectionRange? := (·.toLspRange text) <$> i.range?
-        targetUri := modUri
-        targetRange := declRangeToLspRange ranges.range
-        targetSelectionRange := declRangeToLspRange ranges.selectionRange
-      }
-      return #[ll]
-    return #[]
+  let locationLinksFromDecl (i : Elab.Info) (n : Name) :=
+    locationLinksFromDecl rc.srcSearchPath doc.meta.uri n <| (·.toLspRange text) <$> i.range?
 
   let locationLinksFromBinder (i : Elab.Info) (id : FVarId) := do
     if let some i' := infoTree? >>= InfoTree.findInfo? fun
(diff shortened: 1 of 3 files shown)


src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -117,44 +116,55 @@ partial def handleDefinition (kind : GoToKind) (p : TextDocumentPositionParams)
     if let some modUri ← documentUriFromModule name then
       let range := { start := ⟨0, 0⟩, «end» := ⟨0, 0⟩ : Range }
       let ll : LocationLink := {
-        originSelectionRange? := none
+        originSelectionRange? := (·.toLspRange text) <$> i.stx[2].getRange?
```
gebner, line 119: We still don't make a distinction between positions in the source file and "positions" from the delaborator, so this will give very nonsensical results for delaborated syntax.

```suggestion
        originSelectionRange? := (·.toLspRange text) <$> i.stx[2].getRange? true
```

src/Lean/Server/FileWorker/WidgetRequests.lean
```diff
@@ -102,4 +102,12 @@ builtin_initialize
     (Array InteractiveDiagnostic)
     getInteractiveDiagnostics
 
+builtin_initialize
+  registerBuiltinRpcProcedure
+    `Lean.Widget.getGoToLocation
+    (FileWorker.GoToKind × WithRpcRef InfoWithCtx)
```
gebner, line 108:
```suggestion
    (FileWorker.GoToKind × WithRpcRef InfoWithCtx)
```
We should make this a structure so that we can add extra arguments in the future.

src/Lean/Data/Lsp/Capabilities.lean
```diff
@@ -29,8 +29,17 @@ structure TextDocumentClientCapabilities where
...
+  support : Bool
+  deriving ToJson, FromJson
+
+structure WindowClientCapabilities where
+  showDocument? : Option ShowDocumentClientCapabilities
+  deriving ToJson, FromJson
+
 structure ClientCapabilities where
   textDocument? : Option TextDocumentClientCapabilities
+  window? : Option WindowClientCapabilities
```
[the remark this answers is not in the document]
Vtec234: It is, I can remove it if you prefer. I thought it would be useful if RPC handlers could know whether the `showDocument` request is available (I used it in a prototype).

gebner:
> `CodeToken`

This name has become somewhat misleading, since it is no longer applied to tokens but to whole subexpressions.

Vtec234:
> This name has become somewhat misleading, since it is no longer applied to tokens but to whole subexpressions.

That was always the case, so it was always misleading :) Maybe `SubexpressionInfo`?

src/Lean/Data/Lsp/Capabilities.lean
lovettchris: I can't find where this new "window" capability is used anywhere in the code?

Vtec234: Yep, see the conversation above. It is however just part of the LSP spec, so adding it doesn't hurt.

gebner:
It doesn't work with numerals either (but neither does the source go-to-definition).  Opening the popup doesn't help, because you can't use go-to-definition on applications:
```lean
@OfNat.ofNat Nat 0 (instOfNatNat 0) :
  --^ can't go-to-definition here
Nat
```

gebner:
> It will jump to `elabNumLit`, but only if that is imported

Interestingly, only the source go-to-definition works if `elabNumLit` is imported, the widget go-to-definition still doesn't.

Vtec234:
> It doesn't work with numerals either (but neither does the source go-to-definition).

To follow up - are you suggesting we should resolve all these delaborator issues in this PR, or just in general?

Vtec234:
> Interestingly, only the source go-to-definition works if elabNumLit is imported, the widget go-to-definition still doesn't.

This is likely because during elaboration, the `InfoTree` gets nodes for which elaborator was used for a particular bit of syntax, but we don't add nodes for "which elaborator *would have* been used" during delaboration.

gebner:
> To follow up - are you suggesting we should resolve all these delaborator issues in this PR, or just in general?

I'm not completely convinced that these are bugs in the delaborator (or whether they should be fixed somewhere else).  We should at least fix go-to-definition on function applications before merging.  Otherwise the feature is not really usable.

Vtec234:
Okay, so we can turn on the tagging of all functions using `pp.tagAllFns` (see #911). This generally behaves well for text-function applications like `foo bar`, but has an issue with symbols in that e.g. in `foo = bar`, the entire `" = "` span is tagged, (also it is tagged with its type `Eq` instead of the actual appFn which in this case is `Eq.eq`, which looks like a bug). We don't want the entire span to be tagged since when it isn't, it's possible to hover over the spaces to see the entire expression. Ideally just "=" would be tagged, but that would require introducing a synthetic syntax node — right now there is just one `" = "` atom. Alternatively we could strip the whitespace in the client (e.g. in the infoview)? Perhaps the ideal solution is to have `ppTagSymbols false` which leaves the standard appFns but strips tags when the app unexpands into a notation.

gebner:
Frankly, the easiest solution would be to just call `locationLinksFromDecl` on the function symbol of the term info.  This is significantly more robust than the current approach, and the other cases (imports, fields, binders, ...) don't seem to be useful in the infoview.

(Another hacky solution would be to add a `.getAppFn` in the `ofTermInfo` case.  Not sure if that would break anything unrelated.)

There's also no guarantee that we'll keep using the constant node for go-to-definition.  At some point, we might want to use the whole function application and provide definitions for type class instances (i.e., go-to-definition on `toString 42` returns both the type class definition as well as the `ToString Nat` instance).

> `pp.tagAllFns` [...]

Right now we show a popup for every `SubexprInfo` in the editor; if we add more `SubexprInfo`s then we'll to indicate which `SubexprInfo`s should show a popup and which shouldn't.  There's a reason why there's no popup for `f` in `f a b c`---the popup for `f` is less useful than the one for `f a b c` because it shows fewer arguments, and the three additional choices makes selecting the right subexpression harder.

Another feature where we want a `SubexprInfo` but no popup is semantic highlighting (the commented-out `kind` attribute).

> which leaves the standard appFns but strips tags when the app unexpands into a notation.

Breaking go-to-definition on notation is not a good compromise, that's when you need it most.

Vtec234:
@gebner We now go to the head constant in function applications.

Vtec234:
I think this is good to go.

src/Lean/Server/GoTo.lean
```diff
@@ -0,0 +1,49 @@
+/-
+Copyright (c) 2022 Sebastian Ullrich. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+
+Authors: Sebastian Ullrich, larsk21, Wojciech Nawrocki
```
Kha, line 5:
```suggestion
Authors: Sebastian Ullrich, Lars König, Wojciech Nawrocki
```
:)

Kha (approved):

Kha:
Do you want to squash the last commit, or should I squash the entire PR?

Vtec234:
Squash'd

Issue #1160, opened 2022-05-20
Default to exclude 'LICENSE' files from 'make install'


asdasd1dsadsa:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

While **following the guide of building from source**, I found that the file `LICENSE` and `LICENSES` are also installed to the target directory. I need to use `CMAKE_INSTALL_PREFIX=/usr/local/lean4` instead of `/usr/local/` if I don't want to mess up those parrent directories, which additionally requires an update of `PATH` environment variable for lean4 binaries.

### Versions

```
Lean (version 4.0.0, commit d13fac6f458b, Release)
```

```
5.16.11-gentoo
```

### Additional Information

BTW, I need this when I try to write an ebuild file for the package manager Portage for Gentoo. As a workaround, I may set `-D CMAKE_INSTALL_PREFIX=/usr/local/lean4` or manually clean installed files in `/usr/local/`.

PR #1161, opened 2022-05-20
INSTALL_LICENSE

asdasd1dsadsa:
Add an option to exclude 'LICENSE' files from 'make install'

#1160

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -38,6 +38,8 @@ option(LLVM               "LLVM"               OFF)
 
 # When ON we include githash in the version string
 option(USE_GITHASH        "GIT_HASH"           ON)
+# When ON we install LICENSE files to CMAKE_INSTALL_PREFIX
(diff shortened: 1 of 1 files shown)


asdasd1dsadsa:
> Thanks for your contribution! Please make sure to follow our [Commit Convention](https://leanprover.github.io/lean4/doc/dev/commit_convention.html).

Sorry that I didn't follow that as I didn't know it. And now I'm in doubt what type this commit is of. Too weak to be a 'style' and too strong to be a 'feat'?

Kha:
I'm okay with this patch since it's so simple, but note that packaging a specific version of Lean is heavily discouraged in general. You should package elan instead.

asdasd1dsadsa:
> but note that packaging a specific version of Lean is heavily discouraged in general. You should package elan instead.

In Gentoo usually we only allow installations from source code, and the trouble about versions can be solved with functions of Gentoo ebuild. As a package manager, it can fetch lastest or specific commit from github and support multiple versions to work simultaneously.

PR #1162, opened 2022-05-20
feat: add tactic main mvar to infotree

EdAyers:
This is useful to recover the entire tactic state from the infotree.
This makes possible a feature where the infotree can be used to suggest tactics,
and we can run the tactics in an ephemeral tactic state first to see if the
suggested tactic is any good.

### DIFF
--- a/src/Lean/Elab/InfoTree.lean
+++ b/src/Lean/Elab/InfoTree.lean
@@ -97,6 +97,7 @@ structure FieldInfo where
     We also store the metavariable context at each time since we want metavariables
     unassigned at tactic execution time to be displayed as `?m...`. -/
 structure TacticInfo extends ElabInfo where
+  main : MVarId
   mctxBefore  : MetavarContext
   goalsBefore : List MVarId
   mctxAfter   : MetavarContext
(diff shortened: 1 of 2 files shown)


Kha:
Ah, this is the initial mvar of the entire tactic block, correct? At first I assumed it was the main goal at this point, which would be redundant with `goalsBefore`.
> This makes possible a feature where the infotree can be used to suggest tactics,
and we can run the tactics in an ephemeral tactic state first to see if the
suggested tactic is any good.

Meaning you would search upwards in the info tree for the highest node that has the same main mvar to recover the tactic proof syntax? That sounds sensible, yes.

EdAyers:
ahh, ok so I guess I can set `main` in the tactic context to be a dummy mvar and it should all still work.

EdAyers:
Ok cool I'll close this PR then

leodemoura:
@EdAyers Thanks. I just removed `Context.main`

gebner:
> Happy to remove `Context.main`.

In Lean 3 we used this `main` variable to implement the [`recover`](https://leanprover-community.github.io/mathlib_docs/tactic/interactive.html#tactic.interactive.recover) tactic.  `recover` turns any unassigned metavariables reachable from main into goals.  This is extremely useful to debug issues with "buggy" tactics that don't generate goals for every metavariable (which happens often enough):

```lean
import Lean

elab "buggy" : tactic =>
  Lean.Elab.Tactic.setGoals []

example : False := by
  buggy
```

The error message in Lean 3 was "result contains meta-variables" and the only way to debug this error was to call `recover` to see what goals went missing.  If we don't want to restore `main`, an alternative could be to improve the error message and print the "lost" goals here.

leodemoura:
@gebner Thanks for raising the issue.

> alternative could be to improve the error message and print the "lost" goals here.

I like this suggestion. It will also help with buggy elaborators.

PR #1167, opened 2022-05-25
Add extra note for non vscode users

RosieBaish:
The VSCode extension appears to run the right combination of commands itself to make this work, however if you don't use VSCode you have to run them manually.

From experimentation, in VSCode you have to add the parent directory which contains all your modules, however this doesn't appear to create another `lake` command so I didn't put that in the documentation

### DIFF
--- a/doc/quickstart.md
+++ b/doc/quickstart.md
@@ -41,6 +41,8 @@ On Linux/macOS, you first have to follow the instructions printed by the Lean in
 Note: Packages **have** to be opened using "File > Open Folder..." for imports to work.
 Saved changes are visible in other files after running "Lean 4: Refresh File Dependencies" (`Ctrl+Shift+X`).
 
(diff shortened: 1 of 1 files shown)


Kha:
What editor are you using? You should never have to run `lake serve` yourself.

RosieBaish:
I'm using Emacs which also presumably does the right things via lean4-mode
It was more of a note for debugging, because I just wasted an hour because I didn't have all the right combination of servers running.

Does that mean it ought to go to a different place in the docs?

Kha:
If it's a problem with lean4-mode, we should put/fix it there. On this page it's not clear what the average reader should do with that information.

RosieBaish:
Maybe make clear that it's only there for debugging purposes, e.g. 
"If you're still having issues confirm that you have a lake command in each of the root dirs"?
Or just paste a link to wherever this is actually documented and Reject the PR!

Kha:
I still don't know a better place for this (we don't have a page for "editor extension authors or users willing to fix faulty extensions"), but it should not concern pure users either, so I'll close this for now.

PR #1172, opened 2022-05-27
fix: go to definition/goal state at end of syntax

### LINKED ISSUE #767: Go to definition does not work at end of identifier
### Prerequisites

### Description

I have a little external math function called factorial and when I do goto definition with the cursor at the end of the word it does not work right.

image

### Steps to Reproduce

1. double click the word "factorial" to select the whole word
2. press F12 to goto definition

**Expected behavior:** [What you expect to happen]
Should work and take my to my factorial function, but  it only works if I move the cursor left 1 char so it is "inside" the word "factorial".

**Actual behavior:** [What actually happens]
Fails to find it.

**Reproduces how often:** [What percentage of the time does it reproduce?]
100%

### Versions

commit 89ab3cadf987c7ad005f34b0b744066e2e9c2402

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

Kha:
Fixes #767 as well as a related goal state issue we noticed in our lab course (see new `plainGoal` test)

### DIFF
--- a/src/Lean/Server/FileWorker/RequestHandling.lean
+++ b/src/Lean/Server/FileWorker/RequestHandling.lean
@@ -127,7 +127,7 @@ partial def handleDefinition (kind : GoToKind) (p : TextDocumentPositionParams)
 
   withWaitFindSnap doc (fun s => s.endPos > hoverPos)
     (notFoundX := pure #[]) fun snap => do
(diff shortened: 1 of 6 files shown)


Kha:
I'm not 100% sure this will solve all related issues, but it should be better than before at least

PR #1177, opened 2022-05-31
Relax InfoTree.visitM when visiting holes

larsk21:
Since https://github.com/leanprover/lean4/pull/1172 , `InfoTree.visitM` would cause a panic when the info tree contained holes, as in `tests/lean/forErrors.lean`.


Kha:
Thanks! It's... sub-optimal that the leantests do not catch panics.

Kha:
> It's... sub-optimal that the leantests do not catch panics.

Ah, no, I see now that it only failed in your lint PR

PR #1187, opened 2022-06-03
Unused variables linter fixes

Kha:
The second one was a bit more involved

/cc @larsk21

### DIFF
--- a/src/Lean/Linter/Basic.lean
+++ b/src/Lean/Linter/Basic.lean
@@ -111,7 +111,8 @@ def unusedVariables : Linter := fun stx => do
     if ignoredPatternFns.any (· declStx stack) then
       continue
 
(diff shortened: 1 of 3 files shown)


src/Lean/Linter/Basic.lean
```diff
@@ -66,7 +66,7 @@ def unusedVariables : Linter := fun stx => do
         match expr with
         | .fvar id _ => modify (·.insert id)
         | _          => pure ()
-        return false)
+        return true)
```
larsk21, line 69: Am I misunderstanding something or does this mean that the docstring of `Expr.forEach'` is wrong?

``[...] If `f t` return true, then t's children are not visited.``

Kha: Fixed, thanks!

PR #1188, opened 2022-06-03
refactor: move Lean.PrettyPrinter.Delaborator.SubExpr to Lean.SubExpr

### LINKED ISSUE #1183: rfc: moving `Lean.PrettyPrinter.Delaborator.SubExpr` to `Lean.SubExpr`
I have found metaprogramming uses for `Lean.SubExpr` outside of the delaborator. However it feels strange to write new methods for `SubExpr` in the `Lean.PrettyPrinter.Delaborator` namespace. Would you accept a PR to move it to the `Lean` namespace?

EdAyers:
This is because SubExpr has uses outside the Delaborator.

Closes #1183

### DIFF
--- a/src/Lean/SubExpr.lean
+++ b/src/Lean/SubExpr.lean
@@ -1,0 +1,38 @@
+/-
+Copyright (c) 2021 Microsoft Corporation. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Sebastian Ullrich, Daniel Selsam, Wojciech Nawrocki
+-/
+import Lean.Meta.Basic
+import Std.Data.RBMap
+
+namespace Lean
+
+/-- A position of a subexpression in an expression.
+
+See docstring of `SubExpr` for more detail.-/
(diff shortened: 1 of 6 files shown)


src/Lean/SubExpr.lean
```diff
@@ -0,0 +1,38 @@
...
+namespace Lean
+
+/-- A position of a subexpression in an expression.
+
+See docstring of `SubExpr` for more detail.-/
+abbrev SubExpr.Pos := Nat
+
+/-- An expression and the position of a subexpression within this expression.
+
+Subexpressions are encoded as the parent root expression `e` and a
```
Vtec234, line 18: In delaborator code, is it not the current expression rather than the root parent? E.g. in `Delaborator.SubExpr.descend` we set `expr := child`.

EdAyers: Oooooh. Yes you are right. I assumed it worked like `Substring` where it stores the whole `String`.

PR #1190, opened 2022-06-04
Unname some unused variables

Kha:
The following quick script takes as stdin a Lean build log and replaces reportedly unused variables with `_` in the following two cases where formatting isn't too much of an issue:
* single-letter variables
* something that looks like a top-level pattern variable, in which case we pad with an appropriate number of spaces after the `,` (if any)
```lean
def main : IO Unit := do
  let stdin ← IO.getStdin
  while !(← stdin.isEof) do
    match (← (← IO.getStdin).getLine).trim.splitOn with
    | [loc, "warning:", "unused", "variable", var] =>
      let [file, lineNo, col, ""] := loc.splitOn ":" | throw <| IO.userError loc
      let lineNo := lineNo.toNat! - 1
      let col := col.toNat!
      let var := var.extract ⟨1⟩ (var.prev var.endPos)  -- remove ``
      let lines := (← IO.FS.readFile file).splitOn "\n" |>.toArray
      let line := lines[lineNo]
      if (line.contains '|' && !line.contains '(') || var.length == 1 then  -- then
        IO.println loc
        IO.println line
        let mut line := line.take col ++ "_" ++ "".pushn ' ' (var.length - "_".length) ++ line.drop (col + var.length)
        -- adjust trailing comma
        if var.length > 1 && (line.mkIterator.nextn (col + var.length) |>.curr) == ',' then
          line := line.mkIterator.nextn (col + 1) |>.setCurr ',' |>.nextn (var.length - 1) |>.setCurr ' ' |>.toString
        IO.println line
        let lines := lines.set! lineNo line
        IO.FS.writeFile file (String.join <| lines.toList.intersperse "\n")
    | _ => continue
```
It looks like the linter sometimes reports the same variable multiple times, which confuses the script, so I filtered the input through `sort -u` first. Then I post-processed the result using `sed -i 's![_ : ![!g'` to fix instance implicits. And then I did some more manual cleanups as in the commits below.

### DIFF
--- a/tests/lean/rwEqThms.lean.expected.out
+++ b/tests/lean/rwEqThms.lean.expected.out
@@ -2,9 +2,9 @@
 a : α
 as bs : List α
 h : bs = a :: as
-⊢ List.length (?a :: as) = List.length bs
+⊢ List.length (?head :: as) = List.length bs
 
-case a
+case head
 α : Type ?u
 a : α
 as bs : List α


src/Init/Coe.lean
```diff
@@ -27,11 +27,11 @@ class CoeTail (α : Sort u) (β : Sort v) where
 class CoeHTCT (α : Sort u) (β : Sort v) where
   coe : α → β
 
-class CoeDep (α : Sort u) (_ : α) (β : Sort v) where
+class CoeDep (α : Sort u) (a : α) (β : Sort v) where
```
Kha, line 30: @larsk21 I wonder if we should exclude class signatures by default

src/Lean/Elab/Structure.lean
```diff
@@ -387,8 +387,8 @@ where
         let arg ← mkFreshExprMVar d
         go? (b.instantiate1 arg)
     | e =>
-      let _ := if e.isAppOfArity ``id 2 then e.appArg! else e
-      return some (← reduceProjs (← instantiateMVars e.appArg!) expandedStructNames)
+      let r := if e.isAppOfArity ``id 2 then e.appArg! else e
+      return some (← reduceProjs (← instantiateMVars r.appArg!) expandedStructNames)
```
Kha, line 391: @leodemoura

src/Lean/Meta/ExprDefEq.lean
```diff
@@ -1147,7 +1147,7 @@ private def unfoldBothDefEq (fn : Name) (t s : Expr) : MetaM LBool := do
 
 private def sameHeadSymbol (t s : Expr) : Bool :=
   match t.getAppFn, s.getAppFn with
-  | Expr.const c₁ _ _, Expr.const c₂ _ _ => true
+  | Expr.const _  _ _, Expr.const _  _ _ => true
```
Kha, line 1150: @leodemoura correct?

src/Init/Tactics.lean
```diff
@@ -384,7 +384,7 @@ syntax (name := dbgTrace) "dbg_trace " str : tactic
 
 /-- `stop` is a helper tactic for "discarding" the rest of a proof. It is useful when working on the middle of a complex proofs,
     and less messy than commenting the remainder of the proof. -/
-macro "stop" s:tacticSeq : tactic => `(repeat sorry)
+macro "stop" t:tacticSeq : tactic => `(repeat sorry)
```
gebner, line 387:
```suggestion
macro "stop" tacticSeq : tactic => `(repeat sorry)
```

src/Lean/Meta/Tactic/Simp/Main.lean
```diff
@@ -201,7 +201,7 @@ inductive SimpLetCase where
   | nondepDepVar -- `let x := v; b` is equivalent to `(fun x => b) v`, but result type depends on `x`
   | nondep -- `let x := v; b` is equivalent to `(fun x => b) v`, and result type does not depend on `x`
 
-def getSimpLetCase (n : Name) (t : Expr) (v : Expr) (b : Expr) : MetaM SimpLetCase := do
+def getSimpLetCase (n : Name) (t : Expr) (_ : Expr) (b : Expr) : MetaM SimpLetCase := do
```
gebner, line 204:
```suggestion
def getSimpLetCase (n : Name) (t : Expr) (b : Expr) : MetaM SimpLetCase := do
```

This renaming seems very counter-productive.. not only does it break `getSimpLetCase (v := ...)` and remove the only information about what the argument should be (the *v*alue), but it also hides the fact that the argument is actually unused.

src/Lean/Message.lean
```diff
@@ -318,7 +318,7 @@ def toMessageData (e : KernelException) (opts : Options) : MessageData :=
   match e with
   | unknownConstant env constName       => mkCtx env {} opts m!"(kernel) unknown constant '{constName}'"
   | alreadyDeclared env constName       => mkCtx env {} opts m!"(kernel) constant has already been declared '{constName}'"
-  | declTypeMismatch env decl givenType =>
+  | declTypeMismatch _   decl givenType =>
```
gebner, line 321:
```suggestion
  | declTypeMismatch env decl givenType =>
    mkCtx env {} opts <|
```
Maybe?

src/Lean/Data/JsonRpc.lean
```diff
@@ -27,7 +27,7 @@ instance : OfNat RequestID n := ⟨RequestID.num n⟩
 
 instance : ToString RequestID where
   toString
-  | RequestID.str s => s!"\"s\""
+  | RequestID.str s => s
```
gebner, line 30:
```suggestion
  | RequestID.str s => s!"\"{s}\""
```
I think the quotation marks are significant here.  Otherwise you can't distinguish `1` and `1`.

src/Lean/Compiler/IR/PushProj.lean
```diff
@@ -16,7 +16,7 @@ partial def pushProjs (bs : Array FnBody) (alts : Array Alt) (altsF : Array Inde
     let bs   := bs.pop
     let done (_ : Unit) := (bs.push b ++ ctx.reverse, alts)
     let skip (_ : Unit) := pushProjs bs alts altsF (ctx.push b) (b.collectFreeIndices ctxF)
-    let push (x : VarId) (t : IRType) (v : Expr) :=
+    let push (x : VarId) (_ : IRType) (_ : Expr) :=
```
gebner, line 19:
```suggestion
    let push (x : VarId) :=
```

src/Init/SimpLemmas.lean
```diff
@@ -23,10 +23,10 @@ theorem eq_false (h : ¬ p) : p = False :=
 theorem eq_false' (h : p → False) : p = False :=
   propext <| Iff.intro (fun h' => absurd h' h) (fun h' => False.elim h')
 
-theorem eq_true_of_decide {p : Prop} {s : Decidable p} (h : decide p = true) : p = True :=
-  propext <| Iff.intro (fun h => trivial) (fun _ => of_decide_eq_true h)
+theorem eq_true_of_decide {p : Prop} {_ : Decidable p} (h : decide p = true) : p = True :=
+  propext <| Iff.intro (fun _ => trivial) (fun _ => of_decide_eq_true h)
 
-theorem eq_false_of_decide {p : Prop} {s : Decidable p} (h : decide p = false) : p = False :=
+theorem eq_false_of_decide {p : Prop} {_ : Decidable p} (h : decide p = false) : p = False :=
```
gebner, line 29: I really dislike that this disables the `(s := ...)` syntax to specify implicit arguments.  (Not that `s` is a great name, `inst` or something would certainly be better.)

src/Lean/Meta/Tactic/Simp/Main.lean
Kha: A more clever script definitely would do something more sensible for parameter names

Kha:
@gebner Thanks, I definitely didn't look at every line in detail!

src/Init/SimpLemmas.lean
gebner: > implicit-as-inst-implicit pattern

Do you mean `{_ : Decidable p}` instead of `[Decidable p]`?  (Not sure I understood you correctly.  I would have named it the way around: inst-implicit-as-implicit pattern.)

tests/lean/rwEqThms.lean.expected.out
```diff
@@ -2,9 +2,9 @@
 a : α
 as bs : List α
 h : bs = a :: as
-⊢ List.length (?a :: as) = List.length bs
+⊢ List.length (?head :: as) = List.length bs
```
Kha, line 5: An unexpected consequence from renaming a pattern variable, but I can't say I've every done such a refold myself so far.

src/Lean/Elab/Structure.lean
Kha: Apparently not, the change breaks run/diamond{3,4}.lean

src/Init/SimpLemmas.lean
Kha: Yes, I was thinking of it the other way around - an implicit parameter that becomes an inst-implicit argument in an application

Kha: I just noticed that the `{_ :` style is prevalent in e.g. `Std.Data.PersistentHashMap`, so this really is a pre-existing style issue that should be discussed at a larger scope.

Kha: Just a quick idea, but a "nicer" way could be to allow selecting parameters by type as in `eq_false_of_decide (‹Decidable p› := ...)`. One slight downside would be a larger parser overlap.

src/Init/Coe.lean
larsk21: After seeing this use case, yes, probably. But the same issue occurs with `structure` and `inductive` signatures as well, doesn't it?

larsk21: see #1191

Kha: > But the same issue occurs with `structure` and `inductive` signatures as well, doesn't it?

That's a good question. It could be argued that it is likelier for classes where parameters that are exclusively for directing typeclass resolution sounds quite reasonable, whereas "faux-typed" structures such as the single `KeyedDeclsAttribute.Def` from this PR seem more edge case (and often may as well be `def`s, in which case we can't exclude them anyway).

src/Lean/Meta/ExprDefEq.lean
leodemoura: Thanks for catching this. I pushed a fix for it.

src/Lean/Elab/Structure.lean
leodemoura: Thanks for catching this. I pushed a fix for this.
```lean
      let r := if e.isAppOfArity ``id 2 then e.appArg! else e
      return some (← reduceProjs (← instantiateMVars r) expandedStructNames)
```

PR #1207, opened 2022-06-09
Check types at do pattern reassignment

### LINKED ISSUE #1206: Unused variable false positive on destructuring assignment
### Description

```lean
def boo : Id (HashSet Nat) := do
  let mut vs : HashSet Nat := HashSet.empty
  for i in List.range 10 do
    /- unused variable `vs` -/
    (_, vs) := (i, vs.insert i)
  return vs

/- [9, 8, 7, 6, 5, 4, 3, 2, 1, 0] -/
#eval boo |>.toList
```

**Expected behavior:** The variable `vs` is used as we can see from the `#eval` result, so it should not be marked.

**Actual behavior:** The linter produces a warning.

### Versions

`nightly-2022-06-08`

Kha:
I was mildly surprised that this works as is even with patterns such as ` (_, vs)`, but ` ensure_type_of%` seems lenient enough about that. Do you see any other problematic cases?

Fixes #1206 via the same mechanism as for non-pattern reassignments: the use of the pattern variables in `ensure_type_of%`.

### DIFF
--- a/src/Lean/Elab/Do.lean
+++ b/src/Lean/Elab/Do.lean
@@ -961,23 +961,10 @@ def declToTerm (decl : Syntax) (k : Syntax) : M Syntax := withRef decl <| withFr
     Macro.throwErrorAt decl "unexpected kind of 'do' declaration"
 
 def reassignToTerm (reassign : Syntax) (k : Syntax) : MacroM Syntax := withRef reassign <| withFreshMacroScope do
-  let kind := reassign.getKind
-  if kind == ``Lean.Parser.Term.doReassign then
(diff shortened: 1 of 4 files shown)


Kha:
Thinking a little more about this, it would probably be more robust to put separate `ensure_type_of%`s *after* the bindings, in the let body.

Kha:
> Thinking a little more about this, it would probably be more robust to put separate `ensure_type_of%`s _after_ the bindings, in the let body.

Which is not straightforward without extending `ensure_type_of%`. So we decided to keep the current version of the PR; just don't use inaccessible patterns or something in a reassignment...

PR #1208, opened 2022-06-09
feat: traversal functions for `Expr` and `SubExpr`

EdAyers:
I'm working on a 'contextual suggestion' API for the Lean infoview which lets the user click on subexpressions in the infoview and see suggested tactics for these. In order to create this I ended up with quite a few functions for working with `SubExpr.Pos` and traversing and lensing on expressions with these.

I am happy to break these up in to multiple PRs if the reviewers find that appropriate.
I am also happy to instead merge any/all of these features in mathlib instead.

## List of features in this PR:

- Helper functions for `SubExpr.Pos`: `push` `pop` `isRoot` etc.
- `Lean/Meta/ExprTraverse.lean` contains a set of traversal functions for expressions. These generalise the `visit` functions used in `transform` (it is possible to replace the implementation of `transform` with these traversal functions and all the tests still pass)
- `Lean/Meta/ExprLens.lean` is a set of methods for working on subexpressions of `e : Expr` indexed by `p : SubExpr.Pos`. The main functions are
  - `viewSubexpr (visit : (fvars : Array Expr) → (subexpr: Expr) → M α) (p : Pos) (root : Expr) : M α` which runs `visit` at the subexpression of `root` addressed by `p`. (with bound variables properly instantiated)
  - `replaceSubexpr (replace : (subexpr : Expr) → M Expr) (p : Pos) (root : Expr) : M Expr` replaces the expression at the given position using the `replace` function. Again making sure that the Meta context is correct.
- Some docstrings have been improved.

# Todo list

- [x] add some unit tests for `replaceSubExpr`
- [x] more documentation
- [x] code style

### DIFF
--- a/src/Lean/SubExpr.lean
+++ b/src/Lean/SubExpr.lean
@@ -47,20 +47,21 @@ def push (p : Pos) (c : Nat) : Pos :=
 variable {α : Type} [Inhabited α]
 
 /-- Fold over the position starting at the root and heading to the leaf-/
-def foldl  (f : α → Nat → α) : α → Pos → α :=
-  fix2 (fun r a p => if p.isRoot then a else f (r a p.tail) p.head)
+partial def foldl  (f : α → Nat → α) (a : α) (p : Pos) : α :=
+  if p.isRoot then a else f (foldl f a p.tail) p.head
 
 /-- Fold over the position starting at the leaf and heading to the root-/
-def foldr  (f : Nat → α → α) : Pos → α → α :=
-  fix2 (fun r p a => if p.isRoot then a else r p.tail (f p.head a))
+partial def foldr  (f : Nat → α → α) (p : Pos) (a : α) : α :=
+  if p.isRoot then a else foldr f p.tail (f p.head a)
 
-/-- monad-fold over the position starting at the root and heading to the leaf-/
-def foldlM  [Monad M] (f : α → Nat → M α) : α → Pos → M α :=
-  fix2 (fun r a p => if p.isRoot then pure a else do f (← r a p.tail) p.head)
+/-- monad-fold over the position starting at the root and heading to the leaf -/
+partial def foldlM  [Monad M] (f : α → Nat → M α) (a : α) (p : Pos) : M α :=
+  have : Inhabited (M α) := inferInstance
+  if p.isRoot then pure a else do foldlM f a p.tail >>= (f · p.head)
 
 /-- monad-fold over the position starting at the leaf and finishing at the root. -/
-def foldrM [Monad M] (f : Nat → α → M α) : Pos → α → M α :=
-  fix2 (fun r p a => if p.isRoot then pure a else f p.head a >>= r p.tail)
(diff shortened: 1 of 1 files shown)


Kha:
> I am happy to break these up in to multiple PRs if the reviewers find that appropriate. I am also happy to instead merge any/all of these features in mathlib instead.

@leodemoura and I briefly talked about this PR, we are fine with merging all this into core

src/Lean/SubExpr.lean
```diff
@@ -1,17 +1,102 @@
...
+  else (p.asNat - p.head) / maxChildren
+
+def push (p : Pos) (c : Nat) : Pos :=
+  if c >= maxChildren then panic! s!"invalid coordinate {c}"
+  else p.asNat * maxChildren + c
+
+variable {α : Type} [Inhabited α]
+
+/-- Fold over the position starting at the root and heading to the leaf-/
+def foldl  (f : α → Nat → α) : α → Pos → α :=
```
EdAyers, line 55: `foldup`? `folddown`? When you imagine an expression tree is the root at the top or the bottom?

src/Lean/Meta/ExprTraverse.lean
```diff
@@ -0,0 +1,96 @@
...
+
+variable {M} [Monad M] [MonadLiftT MetaM M] [MonadControlT MetaM M]
+
+/-- Convert a traversal function to a form without the `Pos` argument. -/
+private def forgetPos (t : (Pos → Expr → M Expr) → (Pos → Expr → M Expr)) (visit : Expr → M Expr) (e : Expr) : M Expr :=
+  t (fun _ => visit) Pos.root e
+
+/-- Similar to `traverseLambda` but with an additional pos argument to track position. -/
+def traverseLambdaWithPos
+  (f : Pos → Expr → M Expr) (p : Pos) (e : Expr) : M Expr := visit #[] p e
```
EdAyers, line 22: Alternative formulation: make

```
structure ExprTraverseContext with
  (pos : Pos)
  (fvars : Array Expr)
```

and then

```
traverseXXXWithPos (f : Expr → ReaderT ExprTraverseContext Expr) : Expr → ReaderT ExprTraverseContext Expr
```

Or perhaps instead include the variable instances `[MonadReader ExprTraverseContext M]` and `[MonadReaderWith ExprTraverseContext M]`?

leodemoura:
> > Why is Linux fsanitize crashing? https://github.com/leanprover/lean4/runs/6909389488?check_suite_focus=true#step:15:2360
> 
> @leodemoura It looks like we're leaking `base` here
> 
> https://github.com/leanprover/lean4/blob/7a3ee51d05b9ebbdbcc748a08be14b938df41b3e/src/Init/Fix.lean#L31-L32

Please do not use these functions. I meant to remove them a long time ago. They have runtime support, and were added to solve performance problems in the old parser engine we had. The problem that motivated these primitives does not even exist anymore, and I don't see a strong motivation for having them as primitives in our runtime.

leodemoura:
@EdAyers Could you please rewrite the code and avoid the `fix*` primitives? 

Thanks,
Leo

PR #1209, opened 2022-06-09
fix: add a better handling in case only worker crashes

marianaalanis93:
Add a better handling in case only worker crashes (and server stays available).

Context:

Based on this issue: https://github.com/leanprover/vscode-lean4/issues/175 there is some cases when user code causes a stack overflow and this can cause the worker crashes and the server remains available.
This error handling prevents the user from realizing it (stack overflow in their code) since the Infoview is disabled due to this error.

We can use this new code error to display the proper message in vscode and also put a message so the user can fix their code and restart the server properly.

### DIFF
--- a/src/Lean/Data/JsonRpc.lean
+++ b/src/Lean/Data/JsonRpc.lean
@@ -38,8 +38,6 @@ inductive ErrorCode where
   | methodNotFound
   | invalidParams
   | internalError
-  | workerCrashed
-  | stackOverflow
   | serverNotInitialized
   | unknownErrorCode
   -- LSP-specific codes below.
@@ -47,6 +45,8 @@ inductive ErrorCode where
   | requestCancelled
   -- Lean-specific codes below.
   | rpcNeedsReconnect
+  | workerExited
+  | workerCrashed
   deriving Inhabited, BEq
 
 instance : FromJson ErrorCode := ⟨fun
@@ -60,8 +60,8 @@ instance : FromJson ErrorCode := ⟨fun
   | num (-32801 : Int) => return ErrorCode.contentModified
   | num (-32800 : Int) => return ErrorCode.requestCancelled
   | num (-32900 : Int) => return ErrorCode.rpcNeedsReconnect
-  | num (-32901 : Int) => return ErrorCode.workerCrashed
-  | num (-32902 : Int) => return ErrorCode.stackOverflow
+  | num (-32901 : Int) => return ErrorCode.workerExited
+  | num (-32902 : Int) => return ErrorCode.workerCrashed
   | _  => throw "expected error code"⟩
 
 instance : ToJson ErrorCode := ⟨fun
@@ -75,8 +75,8 @@ instance : ToJson ErrorCode := ⟨fun
   | ErrorCode.contentModified      => (-32801 : Int)
   | ErrorCode.requestCancelled     => (-32800 : Int)
   | ErrorCode.rpcNeedsReconnect    => (-32900 : Int)
-  | ErrorCode.workerCrashed        => (-32901 : Int)
-  | ErrorCode.stackOverflow        => (-32902 : Int)⟩
+  | ErrorCode.workerExited        => (-32901 : Int)
+  | ErrorCode.workerCrashed        => (-32902 : Int)⟩
 
 /- Uses separate constructors for notifications and errors because client and server
 behavior is expected to be wildly different for both. -/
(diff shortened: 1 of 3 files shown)


src/Lean/Data/JsonRpc.lean
```diff
@@ -37,7 +37,8 @@ inductive ErrorCode where
   | invalidRequest
   | methodNotFound
   | invalidParams
-  | internalError
+  | workerCrashed
```
Vtec234, line 40: The `internalError` code is [a built-in](https://www.jsonrpc.org/specification#error_object) of JSON-RPC, so we should not change it. By convention we reserve the `326xy` range for those built-ins, so let's put the two new codes in the "Lean-specific" part, at `32901` and `32902`.

src/Lean/Server/FileWorker.lean
```diff
@@ -401,7 +401,7 @@ section MessageHandling
       catch e =>
         ctx.hOut.writeLspResponseError
           { id
-            code := ErrorCode.internalError
+            code := ErrorCode.workerCrashed
```
Vtec234, line 404: The worker does not crash at this point, it only signals that the given request has failed. We could have an RPC-specific code here, but that's probably orthogonal to the PR so let's leave it as `internalError`.

src/Lean/Server/Watchdog.lean
```diff
@@ -258,7 +258,7 @@ section ServerM
           return WorkerEvent.terminated
         else
           -- Worker crashed
-          fw.errorPendingRequests o ErrorCode.internalError
+          fw.errorPendingRequests o (if exitCode = 1 then ErrorCode.workerCrashed else ErrorCode.stackOverflow)
             s!"Server process for {fw.doc.meta.uri} crashed, {if exitCode = 1 then "see stderr for exception" else "likely due to a stack overflow in user code"}."
```
Vtec234, line 262: This message has historically been a bit misleading. The process crashes for all kinds of reasons, and in my experience most of them have not been stack overflows but rather bugs in the server. Since we're already changing things, maybe let's reword it as
```suggestion
            s!"Server process for {fw.doc.meta.uri} crashed, {if exitCode = 1 then "see stderr for exception" else "likely due to a stack overflow or a bug"}."
```
and rename the error codes `workerCrashed` -> `workerExited` (since in this case we do get a well-behaved exit and can look at the stderr outputs), `stackOverflow` -> `workerCrashed`. Does this seem reasonable?

Kha:
Note: the bot was complaining about the commit message, not the PR title :)

src/Lean/Server/Watchdog.lean
```diff
@@ -258,8 +258,8 @@ section ServerM
           return WorkerEvent.terminated
         else
           -- Worker crashed
-          fw.errorPendingRequests o ErrorCode.internalError
-            s!"Server process for {fw.doc.meta.uri} crashed, {if exitCode = 1 then "see stderr for exception" else "likely due to a stack overflow in user code"}."
+          fw.errorPendingRequests o (if exitCode = 1 then ErrorCode.workerCrashed else ErrorCode.stackOverflow)
+            s!"Server process for {fw.doc.meta.uri} crashed, {if exitCode = 1 then "see stderr for exception" else "likely due to a stack overflow or a bug in user code"}."
```
Vtec234, line 262: I intentionally suggested removing the "in user code" part because historically the bugs have been mostly in our (i.e. the developers') code, and I didn't want to needlessly blame the users :) But maybe as things stabilize this will become more true, so let's leave it.

src/Lean/Data/JsonRpc.lean
```diff
@@ -38,6 +38,8 @@ inductive ErrorCode where
   | methodNotFound
   | invalidParams
   | internalError
+  | workerCrashed
```
Vtec234, line 41: Let's put these error codes at the end, below "Lean-specific codes below.", and rename `workerCrashed` -> `workerExited`, `stackOverflow` -> `workerCrashed`.

src/Lean/Server/Requests.lean
```diff
@@ -38,7 +38,7 @@ def methodNotFound (method : String) : RequestError :=
     message := s!"No request handler found for '{method}'" }
 
 instance : Coe IO.Error RequestError where
-  coe e := { code := ErrorCode.internalError
+  coe e := { code := ErrorCode.workerCrashed
```
Vtec234, line 41: This is a bit non-obvious, but the error constructed here should *also* only be emitted without the worker crashing. It is returned from `RequestM` calls that fail for some `IO` reason like a missing file. The worker should be able to handle those failure graciously and without crashing (if it doesn't, that's a bug). So this should also be `internalError`.

src/Lean/Data/JsonRpc.lean
marianaalanis93: Ok :)

Vtec234 (approved):
Thanks!

Issue #1176, opened 2022-05-30
Should head function syntax be hoverable?


Vtec234:
Continuing discussion from [here](https://github.com/leanprover/lean4/issues/681#issuecomment-922544455) and #1157.

There is currently an inconsistency between whether the head symbol in a function application is hoverable/selectable in source code and in the infoview. I think of this as fundamentally an inconsistency between how we elaborate and delaborate `Elab.Info` nodes, although it could also be fixed at later stages (e.g. in the LSP request handlers).

Consider the context
```lean
/-- Fooifies the args. -/
constant foo : ℕ → ℕ → ℕ
/-- A bar. -/
constant bar : ℕ
```

Then in the editor we have the following LSP hovers
```lean
#check foo bar bar = id bar
/-     ------------0-------
       ---1-------   --2---
       -3- -4- -4-   5- -4-
0 => Prop
1 => Nat
2 => Nat
3 => foo : Nat → Nat → Nat | Fooify args.
4 => bar : Nat | A bar.
5 => id.{1} : {α : Type} → α → α
-/
```
Note that

- Since the smallest one is always selected, to see e.g. the full application `1`, you have to hover over the *space*. It seemed reasonable to me since space is actually the function application operator, however @gebner pointed out to me that this is not very discoverable.
- The head function syntaxes are selectable, e.g. in `3`, and we can see the docstrings for them. However this is not true of notation applications, i.e. `=` is not selectable.

Whereas in the infoview we have
```lean
/- foo bar bar = id bar
   ------------0-------
   ---1-------   --2---
       -3- -3-      -3-
0 => @Eq.eq Nat (foo bar bar) (id bar) : Prop
1 => foo bar bar : Nat
2 => @id Nat bar : Nat
3 => bar : Nat | A bar.
-/
```

And here
- We cannot select the head function, only the total application. This makes it easier to point at things since the span is wider.
- Since we only show the docstring for constants, the doc for `foo` is not shown at `1`. 
- The notation symbol `=` is similarly not selectable.

Now, should we make those entirely consistent? My intuition would be 'yes', although the infoview has extra capabilities — the contents of hover popups are themselves interactive — that might imply we should display less there unless the user requests it.

If we do make them consistent, which one to go for? In past discussions I preferred the editor variant (with hover-over-space), but I think I am in the minority on this. If we select the full-application variant, we should also show the docstring for the head constant on `Expr.app` nodes.

[...]

PR #1210, opened 2022-06-10
feat: prerequisites for widget components

EdAyers:
These are some preliminary changes that are needed for widgets and to make it easier for Lean to process where the infoview the user has clicked. There will be another PR with the widget and context suggestion code itself.

Also note that while this PR doesn't break the vscode-lean4 extension it does cause inaccessible hypotheses to be rendered as [anonymous] instead of as just an empty space. I have a fix for this that I will PR over there.

## changes in this pr

- `InteractiveHypothesis` now has `fvarIds?` field and `InteractiveGoal` now has a `mvarId?` field. This is so that goals and hypotheses in a goal can be uniquely referenced across RPC boundary.
- The way in which `InteractiveHypothesis` handles inaccessible names has changed. These are now represented as `Name.anonymous` instead of an empty array. This is so that the size of `fvarIds?` and `names` is always equal.
- refactor `goalToInteractive` to use for loop instead of fold
- rename `exprToInteractive` → `ppExprTagged`
- add an RPC `Lean.Widget.ppExprTagged`

### DIFF
--- a/src/Lean/Widget/InteractiveGoal.lean
+++ b/src/Lean/Widget/InteractiveGoal.lean
@@ -30,7 +30,7 @@ structure InteractiveGoal where
   goalPrefix : String
   /-- Identifies the goal (ie with the unique
   name of the MVar that it is a goal for.)
-  [todo] what should the key be for a term goal?-/
+  This is none when we are showing a term goal. -/
   mvarId? : Option MVarId := none
   deriving Inhabited, RpcEncoding
 


src/Lean/Widget/InteractiveGoal.lean
```diff
@@ -13,7 +13,10 @@ namespace Lean.Widget
 open Server
 
 structure InteractiveHypothesis where
-  names : Array String
+  /-- The user-friendly name for each hypothesis.
+  If anonymous then the name is inaccessible. -/
```
Kha, line 17: maybe
```suggestion
  If anonymous then the name is inaccessible and hidden. -/
```
since we have non-hidden inaccessibles

src/Lean/Widget/InteractiveGoal.lean
```diff
@@ -13,7 +13,10 @@ namespace Lean.Widget
 open Server
 
 structure InteractiveHypothesis where
```
Kha, line 15: Would something like `InteractiveHypothesisBundle` be a better name?

src/Lean/Widget/InteractiveGoal.lean
```diff
@@ -13,7 +13,10 @@ namespace Lean.Widget
...
-  names : Array String
+  /-- The user-friendly name for each hypothesis.
+  If anonymous then the name is inaccessible. -/
+  names : Array Name
+  fvarIds? : Option (Array FVarId)
```
Kha, line 19: When is this `none`?

src/Lean/Server/FileWorker/WidgetRequests.lean
```diff
@@ -17,6 +17,15 @@ import Lean.Server.FileWorker.RequestHandling
 namespace Lean.Widget
 open Server
 
+builtin_initialize
+  registerBuiltinRpcProcedure
+    `Lean.Widget.ppExprTagged
+    (WithRpcRef ExprWithCtx × Bool)
```
gebner, line 23: Please use a structure for future extensibility.

Kha (approved):

gebner:
The PR adds a new `ExprWithCtx` type without any way to construct it (from the RPC side).  What are your plans for the API with this type?  I'm a bit concerned that we're accumulating lots of incompatible single-purpose types, which apparently cannot be reused easily.

For example we now have `InfoWithCtx` and `ExprWithCtx` that contain essentially the same information (a subexpression).  And we already have a pretty-printing function for `InfoWithCtx`.  And I can get the type of an `InfoWithCtx` (only as text though), but not of an `ExprWithCtx`.

src/Lean/Widget/InteractiveGoal.lean
Vtec234: I think it should just be `Array FVarId`. We represent it as optional in the client so that the client can parse data received from older servers, but the server can rely on the client being up-to-date (and anyway extra JSON fields are no issue for JS).

EdAyers: yes the none was for backwards compat but you are right it can just give the extra field

src/Lean/Server/FileWorker/WidgetRequests.lean
EdAyers: @Vtec234 please could you check if I've put the WithRpcRef in the right place?

src/Lean/Widget/InteractiveGoal.lean
EdAyers: fixed

src/Lean/Widget/InteractiveGoal.lean
EdAyers: fixed

src/Lean/Server/FileWorker/WidgetRequests.lean
Vtec234: Yep! Maybe `PPExprTaggedParams` to follow the LSP conventions (not that I entirely followed them so far, but it's never too late!).

Vtec234:
> I'm a bit concerned that we're accumulating lots of incompatible single-purpose types, which apparently cannot be reused easily.

I also tried the generic
```lean
structure WithCtx (α) where
  lctx : LocalContext
  ctx : ContextInfo
  val : α
```
but it is not really better because
- `Elab.Info` already has a `LocalContext`, so we would duplicate it. 
- A blanket instance `WithRpcRef (WithCtx α)` is unsound. Any given `WithRpcRef` instance has an associated `typeName` which is checked when an RPC pointer is received, so a single such instance would accept any `WithCtx α` in place of any other `WithCtx β`. To avoid that we end up having to write `deriving instance RpcEncoding with {withRef := true} for WithCtx Expr`, and `deriving ... WithCtx Info`.

So I share your concern about duplication, but I was unfortunately not able to make the API better - do you have an idea of how to do so? 

> The PR adds a new `ExprWithCtx` type without any way to construct it (from the RPC side). What are your plans for the API with this type?

It is not hard to construct. [Here](https://github.com/Vtec234/npm-widget/blob/9500e632f3f2b196a7d196ce223158bec8087e99/UserWidget/Demos/CommDiag.lean#L126) is an example from a commutative diagram demo I am working on.

The plan is to make `InteractiveExpr` as defined [here](https://github.com/Vtec234/vscode-lean4/blob/beca0f75336f51feda139130441ed0d0e62acf1b/lean4-infoview/src/components.tsx#L17) work, for example in the JS part of the diagram demo [here](https://github.com/Vtec234/npm-widget/blob/9500e632f3f2b196a7d196ce223158bec8087e99/widget/src/squares.tsx#L106).

> For example we now have `InfoWithCtx` and `ExprWithCtx` that contain essentially the same information (a subexpression).

They are not really the same thing. An `Elab.Info` can be many things - `.ofTermInfo` is only one case. We could use `.ofTermInfo` to send expressions around but that seems hacky, in particular because every `.ofTermInfo` has [an associated](https://github.com/leanprover/lean4/blob/4ba7174c0c095a576d85f17b66a45cd26f00727e/src/Lean/Elab/InfoTree.lean#L52) elaborator and AST, whereas not every expression does.

The purpose of `InfoWithCtx` is to carry over information about delaborated `Info` nodes in a `CodeWithInfos`, and the associated pretty-printing functionality is purpose-specific to showing the contents of infoview popups.

The purpose of `ExprWithCtx` is so that widget writers can send an expression over and display it, as well as send it back in another RPC call for further processing. The former functionality can be achieved by sending `CodeWithInfos`, and even the latter could by inspecting the root tag for its `.ofTermInfo`. However `ExprWithCtx` gives a cleaner interface, and it gives the option to *not* pretty-print the expression, which can be useful if the widget code just needs it around as data, or wants to delay printing until the user requests it.

EdAyers:
Also note that while this PR doesn't break the vscode-lean4 extension it does cause inaccessible hypotheses to be rendered as `[anonymous]` instead of as just an empty space. I have a fix for this that I will PR over there.

gebner:
> It is not hard to construct. Here is an example from a commutative diagram demo I am working on.

I didn't doubt for a second that you can construct an `ExprWithCtx` from the Lean side.  (Though looking at the diagram code, a `ExprWithCtx.withCtx : Expr → MetaM ExprWithCtx` function might be a useful abstraction.)  What I referred to was that there is no *RPC* function which returns an `ExprWithCtx`.

So if I understand your response correctly: there are no immediate plans to use `ExprWithCtx` for any built-in functionality; the `ppExprTagged` function is only there as a helper function for external widgets?

> I also tried the generic [`WithCtx (α)`]

The only thing a generic wrapper would help with is the Lean side, which I think is fine right now.  My concern about duplication was about the RPC side: that functions like `getGoToDefinition`, `ppExprTagged`, etc., only work with one kind of subexpression, but not the other.  Maybe we're just missing `InfoWithCtx.getExpr?` and `ExprWithCtx.toInfo` functions though.

> [InfoWithCtx and ExprWithCtx] are not really the same thing.

I was being a bit hyperbolic: the implementation and underlying data structures are of course different.  But as far as the infoview is concerned, the `.ofTermInfo` case is the only one that is relevant.  (I don't think `.ofFieldInfo` is used at all, it has some non-interactive fallback code that I have never seen in action.)  So in practice, we could replace `InfoWithCtx` by `ExprWithCtx` without any loss in functionality.

Maybe part of my irritation is because `Elab.Info` is not at all the information we want to show as a popup.  Some cases like `.ofCompletionInfo` don't make sense at all to show in a popup.  And something like `.ofMacroExpansionInfo` would be shown together with the term information (so one popup would correspond to more than one `Elab.Info`).  On the UI side you select a subexpression, not an info node (because there can be many info nodes with the same range).

But that's unproductive rambling.  Perhaps it's best if I just think of `InfoWithCtx` as `InfoPopupThunk`.

EdAyers:
I can rename `InfoWithCtx` as `InfoPopupThunk`? I'll add a docstring explaining the difference anyway.

EdAyers:
@Vtec234 I couldn't find an example of using `InfoWithCtx` outside of `SubexprInfo`, and `InfoWithCtx` is an opaque reference to vscode. Maybe it could be replaced with `ExprWithCtx`?

Vtec234:
> I couldn't find an example of using `InfoWithCtx` outside of `SubexprInfo`, and `InfoWithCtx` is an opaque reference to vscode. Maybe it could be replaced with `ExprWithCtx`.

A priori, I am not opposed to this refactor if it makes the API cleaner. But it would be a breaking change to the RPC API, unless we want to keep calling the function `ExprWithCtx -> CodeWithInfos` `infoToInteractive`. One option would be to make a new function, bump the server version to indicate that new APIs exist, but keep the old name around for a while for backwards compatibility.

src/Lean/Widget/InteractiveCode.lean
```diff
@@ -46,10 +40,11 @@ where
   go (tt : TaggedText (Nat × Nat)) :=
     tt.rewrite fun (n, _) subTt =>
       match infos.find? n with
-      | none   => go subTt
-      | some i => TaggedText.tag ⟨WithRpcRef.mk { ctx, info := i }, n⟩ (go subTt)
+      | some (.ofTermInfo ti) => TaggedText.tag ⟨WithRpcRef.mk { ctx, lctx := ti.lctx, expr := ti.expr }, n⟩ (go subTt)
+      | some (.ofFieldInfo _fi) => go subTt -- [todo] what to do in this case?
```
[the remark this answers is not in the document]
Vtec234: Hm, so the move to `ExprWithCtx` only means we cannot distinguish terminfo from fieldinfo when constructing popups. In the editor hover code, we do
``````lean
    | Info.ofFieldInfo fi =>
      let tp ← Meta.inferType fi.val
      let tpFmt ← Meta.ppExpr tp
      return some f!"```lean
{fi.fieldName} : {tpFmt}
```"
``````
which is slightly different than just showing the projector. 

>  I don't think `.ofFieldInfo` is used at all, it has some non-interactive fallback code that I have never seen in action.

This looks like an oversight. The delaborator does produce these, and we should be showing them.

gebner:
I did not want to suggest any big refactorings for this PR.  If we merge `InfoWithCtx` and `ExprWithCtx` now, then we should stick with the `infoToInteractive` name for compatibility.  (There's also an `info` field name in `SubexprInfo`, BTW.)

Once we do a bigger refactoring and bump the server version, we should first a) decide on #1176 and b) figure out where to put the semantic highlighting information.  The latter might require some incompatible changes in the `CodeWithInfos` type as well.

EdAyers:
We are going to remove `ExprWithCtx` for now and we will figure out what to do with it. So this PR becomes just about contextual suggestions.

Vtec234 (approved):

PR #1211, opened 2022-06-10
Specialize `interpolatedStr` to `termParser`

Kha:
As other use cases seem unlikely


tydeu:
The current more general version seems like it could be useful in DSLs, no? Why change it?

Kha:
Yes, I decided against this change, just have to work around it a bit in other parts

PR #1213, opened 2022-06-12
Prove that `Nat.bitwise` terminates

Vtec234:
- Copy `strongInductionOn` over [from mathlib4](https://github.com/leanprover-community/mathlib4/blob/c56001bae38730606b855196acdd52d334971977/Mathlib/Init/Data/Nat/Lemmas.lean#L195). I am not sure if it belongs in `Init.WF`, but at least dependency-wise that seemed like the only possible place to put it.
- Prove `theorem div_lt_self {n k : Nat} (hLtN : 0 < n) (hLtK : 1 < k) : n / k < n` which may be generally useful for the WF tactics when `linarith` can discharge the inequalities.
- The redundant `Nat.lt_of_le_and_ne` is removed.

### DIFF
--- a/src/Init/Data/Nat/Bitwise.lean
+++ b/src/Init/Data/Nat/Bitwise.lean
@@ -10,7 +10,10 @@ import Init.Coe
 
 namespace Nat
 
-partial def bitwise (f : Bool → Bool → Bool) (n m : Nat) : Nat :=
+theorem bitwise_rec_lemma {n : Nat} (hNe : n ≠ 0) : n / 2 < n :=
+  Nat.div_lt_self (Nat.zero_lt_of_ne_zero hNe) (Nat.lt_succ_self _)
+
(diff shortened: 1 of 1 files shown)


src/Init/Data/Nat/Div.lean
```diff
@@ -39,6 +39,31 @@ theorem div.inductionOn.{u}
     base x y h
 decreasing_by apply div_rec_lemma; assumption
 
+theorem div_le_self (n k : Nat) : n / k ≤ n := by
+  induction n using Nat.strongInductionOn with
+  | ind n ih =>
+    rw [div_eq]
+    -- Note: manual split to avoid Classical.em which is not yet defined
+    cases (inferInstance : Decidable (0 < k ∧ k ≤ n)) with
```
Vtec234, line 47: `split` fails here since `Classical.em` is not in scope, so I did this slightly strange manual decidable split.

PR #1217, opened 2022-06-14
fix: tag idents so that ofFieldInfo nodes are preserved

Vtec234:
It was noted recently that structure field names in a `structInst` do not have associated `Info` nodes after formatting and hence are not highlightable. This should fix that, although I am not sure if it doesn't break something else - I still don't really understand how the formatter works.

### DIFF
--- a/src/Lean/PrettyPrinter/Formatter.lean
+++ b/src/Lean/PrettyPrinter/Formatter.lean
@@ -159,11 +159,16 @@ def group (x : Formatter) : Formatter := do
     isUngrouped := false
   }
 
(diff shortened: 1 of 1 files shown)


gebner:
Ah, that's where the bug came from.  Right now, we only propagate the tag for category parsers and symbols.  And idents are neither.

Do we now get duplicate `.tag`s for identifiers, since we tag them in both the ident and term formatter?

Vtec234:
> Do we now get duplicate `.tags` for identifiers, since we tag them in both the ident and term formatter?

We did, fixed.

PR #1222, opened 2022-06-15
fix: `LawfulBEq` class

fgdorais:
make arguments implicit for `LawfulBEq.eq_of_beq` and `LawfulBEq.rfl` and protect `LawfulBEq.rfl`


fgdorais:
This is to fix some shadowing issues I ran into when opening the `LawfulBEq` namespace.

leodemoura:
Merged manually.

PR #1226, opened 2022-06-16
Make grammar more strict regarding line breaks

### LINKED ISSUE #1218: Anomalous parsing of structure expressions
### Prerequisites

### Description

Per section 9.3 of the Manual, the following is allowed:
```
structure Point where
 x : Nat
 y : Nat
deriving Repr

def mkPoint (a : Nat) : Point where
  x := 1
  y := a
```
However, the following is also valid:
```
def mkPoint1 (a : Nat) : Point where
  x := 1y := a

#eval mkPoint1 2 -- { x := 1, y := 2 }
```
Note that `x := 1 y := a` and `x := ay := a` in the body of `mkPoint1` throw errors as expected.
### Steps to Reproduce

1. Copy the above code into VS Code.

**Expected behavior:** 
Syntax error

**Actual behavior:**
Parsing is successful

**Reproduces how often:**
100%

### Versions

leanprover/lean4:nightly-2022-06-13

### Additional Information

None

Kha:
Fixes #1218


tests/lean/Reformat/Input.lean
```diff
@@ -186,13 +186,9 @@ class Inhabited (α : Sort u) :=
 opaque arbitrary (α : Sort u) [s : Inhabited α] : α :=
   @Inhabited.default α s
 
-instance (α : Sort u) {β : Sort v} [Inhabited β] : Inhabited (α → β) := {
-  default := fun _ => arbitrary β
-}
+instance (α : Sort u) {β : Sort v} [Inhabited β] : Inhabited (α → β) := {default := fun _ => arbitrary β}
```
Kha, line 189: This is a quick hack to placate the current test's logic: if the pretty printer removed the line break, the test would complain about the syntax tree's structure being changed, as the line break is now represented by an explicit `[]` node in the tree. However, I am pretty sure that we do eventually want the formatter to change the syntax tree structure in order to e.g. enforce consistent use of trailing separators, so at that point we will have to change this test's logic anyway.

src/Lean/Parser/Extra.lean
```diff
@@ -57,6 +57,12 @@ attribute [runBuiltinParserAttributeHooks]
 @[runBuiltinParserAttributeHooks, inline] def manyIndent (p : Parser) : Parser :=
   withPosition $ many (checkColGe "irrelevant" >> p)
 
+@[runBuiltinParserAttributeHooks, inline] def sepByIndent (p : Parser) (sep : String) (psep : Parser := symbol sep) (allowTrailingSep : Bool := false) : Parser :=
+  withPosition $ sepBy (checkColGe "irrelevant" >> p) sep psep allowTrailingSep
```
gebner, line 61: I wonder if `sepByIndent` should automatically accept newlines as separator.  From what I can tell, this is the case everywhere it's used.

```suggestion
  withPosition $ sepBy (checkColGe "irrelevant" >> p) sep (psep <|> checkLinebreakBefore >> pushNone) allowTrailingSep
```

tests/lean/semicolonOrLinebreak.lean.expected.out
```diff
@@ -0,0 +1,5 @@
+building '/nix/store/kzswn86bwsdhxg5jb5d8vnfjkq36akl7-lean.drv'...
+semicolonOrLinebreak.lean:2:12: error: expected ';' or line break
+semicolonOrLinebreak.lean:9:31-10:8: error: field 'y' is missing
+semicolonOrLinebreak.lean:10:8: error: expected command
+semicolonOrLinebreak.lean:20:8: error: expected ':'
```
Kha, line 5: This message is "fun": the longest parse `{ x y` is not from the structure instance notation, but from the dep arrow `{ x y : _ } -> _`, thus the `:` reference. Not much we can do there, but I don't think it matters in practice.

tests/lean/semicolonOrLinebreak.lean.expected.out
```diff
@@ -0,0 +1,5 @@
+building '/nix/store/kzswn86bwsdhxg5jb5d8vnfjkq36akl7-lean.drv'...
```
gebner, line 1: This is probably a typo.

tests/lean/semicolonOrLinebreak.lean.expected.out
```diff
@@ -0,0 +1,5 @@
...
+semicolonOrLinebreak.lean:2:12: error: expected ';' or line break
+semicolonOrLinebreak.lean:9:31-10:8: error: field 'y' is missing
+semicolonOrLinebreak.lean:10:8: error: expected command
```
Kha, line 4: This error message was better in Leo's proposal, but again I don't think it matters in practice. The important thing is that there is an error at all.

src/Lean/Parser/Extra.lean
```diff
@@ -58,7 +58,7 @@ attribute [runBuiltinParserAttributeHooks]
   withPosition $ many (checkColGe "irrelevant" >> p)
 
 @[runBuiltinParserAttributeHooks, inline] def sepByIndent (p : Parser) (sep : String) (psep : Parser := symbol sep) (allowTrailingSep : Bool := false) : Parser :=
-  withPosition $ sepBy (checkColGe "irrelevant" >> p) sep psep allowTrailingSep
+  withPosition $ sepBy (checkColGe "irrelevant" >> p) sep (psep <|> checkLinebreakBefore >> pushNone) allowTrailingSep
```
gebner, line 61: The same suggestion also applies to `sepBy1Indent` below.

Kha: done

tests/lean/semicolonOrLinebreak.lean.expected.out
Kha: Thanks, I finally made sure this doesn't happen to me all the time...

src/Lean/Parser/Term.lean
```diff
@@ -92,7 +94,7 @@ def structInstField  := ppGroup $ leading_parser structInstLVal >> " := " >> ter
 def structInstFieldAbbrev := leading_parser atomic (ident >> notFollowedBy ("." <|> ":=" <|> symbol "[") "invalid field abbreviation") -- `x` is an abbreviation for `x := x`
 def optEllipsis      := leading_parser optional ".."
 @[builtinTermParser] def structInst := leading_parser "{" >> ppHardSpace >> optional (atomic (sepBy1 termParser ", " >> " with "))
-  >> manyIndent (group ((structInstFieldAbbrev <|> structInstField) >> optional ", "))
+  >> sepByIndent (structInstFieldAbbrev <|> structInstField) ", " (allowTrailingSep := true)
   >> optEllipsis
   >> optional (" : " >> termParser) >> " }"
```
Kha, line 99: Note: `{ x .. }` and `{ x, : Nat }` are accepted as before

src/Lean/Parser/Term.lean
```diff
@@ -45,6 +45,7 @@ def seq1 :=
 end Tactic
 
 def darrow : Parser := " => "
+def semicolonOrLinebreak := ";" <|> checkLinebreakBefore >> pushNone
```
Kha, line 48: Caveat: the current syntax pattern semantics make this case a bit peculiar: `` `(let $x:ident := $e; $f) `` will match regardless of whether `;` or a linebreak was used as tokens are ignored, but the same is not true when using a linebreak *in the pattern* as the `pushNone` will be checked structurally. We don't usually use linebreaks in syntax patterns at all though.

gebner: > We don't usually use linebreaks in syntax patterns at all though.

https://github.com/leanprover-community/mathport/blob/37b4fd1c21239c9bda220c19208f88ff12f13d01/Mathport/Syntax/Transform/Tactic.lean#L94-L107 :smile: 

This is quite the gotcha (because it is also asymmetric).  Could we insert an empty atom instead of the null node?

If I understand you correctly, `sepByIndent` is not affected, right?  The pattern
```
{
$a
$b
}
```
will match all of:
`{a,b,}`, `{a,b}`,
```
{a
 b}
```
and
```
{a,
 b
}
```

Kha: > This is quite the gotcha (because it is also asymmetric). Could we insert an empty atom instead of the null node?

I'm not sure I want to know what invariants an empty atom would break. Perhaps it's time to introduce a `matchIgnore` syntax kind or something we can put around the `<|>`.

> If I understand you correctly, sepByIndent is not affected, right?

It is affected in the same way; from the point of view of syntax `match`, there is no difference. In fact, ignoring the trailing separator presents an additional challenge. When matching a `nullNodeKind` pattern, syntax `match` currently eagerly checks that the number of elements in the pattern and in the discriminant is the same. We would have to change this such that a suffix of `matchIgnore` elements in the pattern may be missing from the discriminant :dizzy_face: .

gebner: Ah I see, the `sepBy` matching issues are already present.  I mixed it up with the `$[$a],*` patterns, which ignore the separators.

> We would have to change this such that a suffix of `matchIgnore` elements in the pattern may be missing from the discriminant

We could also always generate a trailing `matchIgnore` node (which is either empty or contains the separator).

Kha: @gebner Just to be sure, the mathport code already did not accept `;` before this PR, correct? If so, I will merge the PR for now before it bitrots and introduce `sepByIndent` later.

PR #1229, opened 2022-06-18
Auto-group `syntax` parsers where necessary

Kha:


src/Init/Conv.lean
```diff
@@ -54,7 +54,7 @@ macro "left" : conv => `(lhs)
 macro "right" : conv => `(rhs)
 macro "intro " xs:(colGt ident)* : conv => `(ext $xs*)
 
-syntax enterArg := ident <|> group("@"? num)
+syntax enterArg := ident <|> ("@"? num)
```
Kha, line 57: This is a separate pitfall: `<|>` is an operator on two `stx`, not `stx+`. I think we would need to change all occurrences of `stx+` to `sepBy1(stx+, " <|> ")` to fix this.

src/Lean/Elab/Syntax.lean
```diff
@@ -137,50 +147,63 @@ where
...
-    | [(c, false)]     => ensureNoPrec stx; `(ParserDescr.parser $(quote c))
+    | [(c, true)]      =>
+      ensureNoPrec stx
+      -- `syntax _ :=` at least enforces this
+      let stackSz := 1
+      return (mkIdentFrom stx c, stackSz)
+    | [(c, false)]     =>
+      ensureNoPrec stx
+      -- as usual, we assume that people using `Parser` know what they are doing
+      let stackSz := 1
```
Kha, line 181: Will be resolved when we parameterize `Parser` with `stackSz`... :)

leodemoura (approved):

PR #1231, opened 2022-06-18
Improve the notation pretty printer

hargoniX:
As the commit message says:

Instead of the previous constraints on the right hand side that only
allowed a permutation of variables as parameters to a function the
new heuristic allows anything to the right of a function as long as
each variable only appears at most once.

The constraint on each variable appearing at most once is because
this would not behave as the user might expect (with the current implementation), given:
```
notation "doubleRhsTest " x => Prod.mk x x
```
This: `Prod.mk 11 12` would delab to `doubleRhsTest 11` which is clearly not right.

I also updated stage0 in case it changes any pretty printers for built-in notation.

There is a single test case that behaves differently now, 255 where this:
```
  variable (α : Type)
  variable (x : α)
  notation "A" => id x
  #check A
```
Instead of not pretty printing (`id x`) will now pretty print into `A`. However as soon as the shadow cases from that test come up this heuristic will fail again, I'm not quite sure how to fix that/whether it is worth it to fix since it seems rather niche.

### DIFF
--- a/src/Lean/Elab/Notation.lean
+++ b/src/Lean/Elab/Notation.lean
@@ -5,6 +5,7 @@ Authors: Leonardo de Moura
 -/
 import Lean.Elab.Syntax
 import Lean.Elab.AuxDef
+import Lean.Elab.BuiltinNotation
 
 namespace Lean.Elab.Command
 open Lean.Syntax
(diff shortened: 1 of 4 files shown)


leodemoura:
I think this PR is an improvement. It is great to see notations such as `setOf` working.
Note that it is still sensitive to the output produced by the delaborator. Example:
```lean
def Set (α : Type u) := α → Prop
def setOf (x : α → Prop) : Set α := x
notation "{ " x " : " t " | " p " }" => setOf (fun (x : t) => p)
#check { x : Int | x < 5 }
-- setOf fun x => x < 5
set_option pp.funBinderTypes true
#check { x : Int | x < 5 }
-- { x : Int | < 5 } -- as expected 
```
I am also fine with a hardcoded `removeNonSemanticParenthesis` for now.
@Kha What do you think?
Happy to merge without the `update stage0` commit. We can do the "update stage0" later to avoid conflicts.

leodemoura:
Another similar example:
```lean
def Set (α : Type u) := α → Prop
def setOf (x : α → Prop) : Set α := x
notation "{ " x " | " p " }" => setOf (fun x => p)
set_option pp.funBinderTypes true
#check { x | x < 5 }
-- { (x : Nat) | x < 5 }  -- A bit weird, but the result is correct and can be reparsed.
```

src/Lean/Elab/Notation.lean
```diff
@@ -43,15 +44,70 @@ def expandNotationItemIntoPattern (stx : Syntax) : MacroM Syntax :=
...
+    match stx with
+    | .node info kind args => pure $ .node info kind (←args.mapM removeNonSemanticParenthesis)
+    | _ => pure stx
+
+partial def hasDuplicateAntiquot (stx : Array Syntax) : StateM NameSet Bool :=
+  go stx
+where
+  go (stx : Array Syntax) : StateM NameSet Bool := do
+    for s in stx do
+      if ←go' s then
```
Kha, line 73:
```suggestion
      if (← go' s) then
```
style

src/Lean/Elab/Notation.lean
```diff
@@ -43,15 +44,70 @@ def expandNotationItemIntoPattern (stx : Syntax) : MacroM Syntax :=
...
+    if let some stx ← Term.expandCDot? e then -- cdot notation doesn't have a delab, expand it
+      removeNonSemanticParenthesis stx
+    else
+      removeNonSemanticParenthesis e -- this parenthesis pair is non semantic, get rid off it
+  | stx =>
+    match stx with
+    | .node info kind args => pure $ .node info kind (←args.mapM removeNonSemanticParenthesis)
+    | _ => pure stx
+
+partial def hasDuplicateAntiquot (stx : Array Syntax) : StateM NameSet Bool :=
```
Kha, line 68:
```suggestion
partial def hasDuplicateAntiquot (stxs : Array Syntax) : StateM NameSet Bool :=
```
or `args`, just not `stx`

src/Lean/Elab/Notation.lean
```diff
@@ -43,15 +44,70 @@ def expandNotationItemIntoPattern (stx : Syntax) : MacroM Syntax :=
   else
     Macro.throwUnsupported
 
+partial def removeNonSemanticParenthesis (stx : Syntax) : MacroM Syntax := do
```
Kha, line 47:
```suggestion
partial def removeNonSemanticParentheses (stx : Syntax) : MacroM Syntax := do
```
I think it would be OK to drop the `NonSemantic` part. There are different kinds of parentheses, and then there are *the* parentheses.

Kha: As for implementation, see also https://github.com/leanprover/lean4/blob/b6f251bcd33e3941cd653777068bd0bb2919a6a4/tests/lean/run/Reparen.lean#L14-L20
(the style would be slightly different today)

src/Lean/Elab/Notation.lean
```diff
@@ -43,15 +44,70 @@ def expandNotationItemIntoPattern (stx : Syntax) : MacroM Syntax :=
...
+    else
+      match stx with
+      | .node _ _ args => go args
+      | _ => pure false
+
 /-- Try to derive a `SimpleDelab` from a notation.
-    The notation must be of the form `notation ... => c var_1 ... var_n`
-    where `c` is a declaration in the current scope and the `var_i` are a permutation of the LHS vars. -/
+    The notation must be of the form `notation ... => c body`
+    where `c` is a declaration in the current scope `body` any syntax
```
Kha, line 91: missing a word

src/Lean/Elab/Notation.lean
Kha: You could replace the recursion + state monad with a loop over `Syntax.topDown` (we probably want `firstChoiceOnly := true` even if we may not handle choice nodes correctly yet in other places) and a mutable variable.

Kha:
Yes, I agree that this is an improvement. Just for the larger picture, we were aware of these "brittle" delab output issues when implementing the delaborator. One solution I had in mind at one point is that in analogy to the expected type in the elaborator, the delaborator could pass down an "expected syntax" (with unresolved antiquotations as holes). That wouldn't magically fix anything by itself, but it would give delaborators sufficient information that they *could* implement special cases such as "I guess I'm supposed to print with `funBinderTypes` here".

leodemoura:
@Kha 

> That wouldn't magically fix anything by itself, but it would give delaborators sufficient information that they could implement special cases such as "I guess I'm supposed to print with funBinderTypes here".

I think this is a good idea.

I am merging the revised version since we all agree it is an improvement, and the issues raised above have been addressed.

Issue #1225, opened 2022-06-16
RFC: User widgets


Vtec234:
# RFC for the user widgets system

Authors: @EdAyers, @Vtec234

![Rubik's](https://user-images.githubusercontent.com/5064353/174113871-743ab334-6311-4239-a5a7-68418b6a75e3.png)

## Goals

The idea of user-widgets is to give Lean users the ability to write their own GUIs shown within the Lean infoview without having to muck around with custom versions of the infoview code.

The widgets in Lean 3 got some hype, but there were many limitations of the approach, the main one being that no user-written code ran 'client side' on the infoview.

With the Lean 4 extension, we want to instead let the infoview dynamically load React components from JavaScript served by the Lean server. This has the advantage of allowing users to bundle and use arbitrary JavaScript libraries within the infoview. For example, we can now do things like 3d rendering, diagram drawing, TeX rendering and so on without needing to modify the infoview. Moreover Lake integration makes building widget bundles from TypeScript + NPM modules nice and easy.

## Proposal

The current prototype widget protocol can be found [here](https://github.com/Vtec234/npm-widget/blob/9500e632f3f2b196a7d196ce223158bec8087e99/UserWidget/WidgetProtocol.lean).

### JavaScript widgets

A *widget source* is a valid JavaScript ESModule which exports a React component. The module must include `import * as React from 'react'` in the imports. The React component may accept a `props` argument whose value will be determined for each particular *widget instance* (below). Widget sources may import the `@lean4/infoview` package (which we intend to publish on NPM) in order to use components such as `InteractiveMessage` to display `MessageData` interactively. The precise API surface exposed in `@lean4/infoview` is orthogonal to this RFC.

NB: While only JavaScript is officially supported, we do not actually require that a widget source be JavaScript. As such, other formats can be stored as widget sources and fetched by custom infoview implementations written in other languages.

### Widget source storage

On the Lean side, we use the `@[widgetSource]` attribute on definitions of type `String` to register them as widget sources. A widget source registered with `@[widgetSource]` has a unique `widgetSourceId : Name` which is just the fully qualified definition name. Alternatively, instead of naming sources we could use content-addressing and associated each source with its hash.

We define the `Lean.Widget.getWidgetSource` RPC call with a `widgetSourceId` (or again, hash) argument to retrieve widget sources from the Lean server at a source position following the `@[widgetSource]` registration. On the infoview side, we treat the Lean server as a kind of content delivery network which can send us widget sources via `getWidgetSource`.

[...]

PR #1238, opened 2022-06-22
feat: user widgets

EdAyers:
image

See also
- RFC:  #1225
- vscode PR https://github.com/leanprover/vscode-lean4/pull/194
- https://github.com/Vtec234/npm-widget

Todo:
- [x] keep adding to documentation, including a full example of setting up npm etc with Lake
- [x] e2e interactive test for a simple widget
- [x] `ToJson Syntax`? Means that we can pass `UserWidgetInfo` directly to infoview. → removed
- [x] publish `@lean4/infoview` on npm https://github.com/leanprover/vscode-lean4/issues/203
- [x] vscode pr  https://github.com/leanprover/vscode-lean4/pull/194
- [x] content addressing of getWidgetSource
- [x] add rubiks cube, comm-diagrams, etc demos to lean-samples repository https://github.com/leanprover/lean4-samples/pull/3
- [x] fix hyperlink to lean4-samples (will fix when https://github.com/leanprover/lean4-samples/pull/3 is done)

### DIFF
--- a/tests/lean/interactive/run.lean
+++ b/tests/lean/interactive/run.lean
@@ -15,7 +15,6 @@ def ident : Parsec Name := do
 partial def main (args : List String) : IO Unit := do
   let uri := s!"file://{args.head!}"
   Ipc.runWith (←IO.appPath) #["--server"] do
-    let hIn ← Ipc.stdin
     let capabilities := {
       textDocument? := some {
         completion? := some {
@@ -85,20 +84,19 @@ partial def main (args : List String) : IO Unit := do
             let r ← Ipc.readResponseAs requestNo RpcConnected
             rpcSessionId := some r.result.sessionId
             requestNo := requestNo + 1
-          let tdpp : TextDocumentPositionParams := {textDocument := {uri := uri}, position := pos }
           let ps : RpcCallParams := {
             textDocument := {uri := uri},
             position := pos,
             sessionId := rpcSessionId.get!,
             method := `Lean.Widget.getWidgets,
-            params := toJson tdpp,
+            params := toJson pos,
           }
           Ipc.writeRequest ⟨requestNo, "$/lean/rpc/call", ps⟩
           let response ← Ipc.readResponseAs requestNo Lean.Widget.GetWidgetsResponse
           requestNo := requestNo + 1
           IO.eprintln (toJson response.result)
           for w in response.result.widgets do
-            let params : Lean.Widget.GetWidgetSourceParams := { pos := tdpp, hash := w.javascriptHash }
+            let params : Lean.Widget.GetWidgetSourceParams := { pos, hash := w.javascriptHash }
             let ps : RpcCallParams := {
               ps with
               method := `Lean.Widget.getWidgetSource,
@@ -113,13 +111,12 @@ partial def main (args : List String) : IO Unit := do
             | throw <| IO.userError s!"failed to parse {params}"
           let params := params.setObjVal! "textDocument" (toJson { uri := uri : TextDocumentIdentifier })
           -- TODO: correctly compute in presence of Unicode
-          let column := ws.endPos + "--"
           let params := params.setObjVal! "position" (toJson pos)
           IO.eprintln params
           Ipc.writeRequest ⟨requestNo, method, params⟩
           let rec readFirstResponse := do
             match ← Ipc.readMessage with
-            | resp@(Message.response id r) =>
+            | Message.response id r =>
               assert! id == requestNo
               return r
             | Message.notification .. => readFirstResponse


src/Lean/Widget/UserWidget.lean
```diff
@@ -0,0 +1,167 @@
...
+      return w
+    else
+      throw  (RequestError.mk JsonRpc.ErrorCode.invalidParams s!"No registered user-widget with name {args.widgetSourceId}")
+
+open Lean Elab
+
+/--
+  Try to retrieve the `UserWidgetInfo` at a particular position.
+-/
+partial def widgetInfoAt? (text : FileMap) (t : InfoTree) (hoverPos : String.Pos) : List UserWidgetInfo :=
```
Vtec234, line 90:
```suggestion
def widgetInfosAt? (text : FileMap) (t : InfoTree) (hoverPos : String.Pos) : List UserWidgetInfo :=
```
No recursive calls right?

src/Lean/Widget/UserWidget.lean
```diff
@@ -0,0 +1,167 @@
...
+The editor can use the `Lean.Widget.getWidgetSource` RPC method to
+get this object.
+
+See the [tutorial](doc/widgets.md) above this declaration for more information on
+how to use the widgets system.
+
+-/
+structure WidgetSource where
+  /-- Unique identifier for the widget. -/
+  widgetSourceId : Name
```
Vtec234, line 27: This is already a parameter, seems redundant to reply back with it.

src/Lean/Widget/UserWidget.lean
```diff
@@ -0,0 +1,183 @@
...
+    (Std.RBMap String WidgetSource compare)
+
+instance : Inhabited (Std.RBMap String WidgetSource compare) := ⟨∅⟩
+instance : Inhabited (WidgetSourceRegistry) := inferInstanceAs (Inhabited (PersistentEnvExtension _ _ (List _ × _)))
+
+-- Mapping widgetSourceId to hash
+builtin_initialize widgetHashes : MapDeclarationExtension String ← mkMapDeclarationExtension `widgetHashes
+
+-- Mapping hash to wWidgetComponent
+builtin_initialize widgetSourceRegistry : WidgetSourceRegistry ←
```
EdAyers, line 44: @leodemoura are there any performance implications of making an environment extension in this way?

leodemoura: How many entries are you expecting this table to have?

EdAyers: A low number, <20.

doc/widgets.md
```diff
@@ -0,0 +1,58 @@
...
+Here is an example lakefile snippet that sets this up.
+Your npm javascript project lives in a subfolder called `./widget` whose build process generates a single file `widget/dist/index.js`.
+
+```lean
+-- ./lakefile.lean
+
+def jsTarget (pkgDir : FilePath) : FileTarget :=
+  let jsFile := pkgDir / "widget/dist/index.js"
+  let srcFiles := inputFileTarget <| pkgDir / "widget/src/index.tsx"
+  fileTargetWithDep jsFile srcFiles fun _srcFile => do
```
gebner, line 40: Can we sneak in a dependency on `package.json`, etc., here?

EdAyers: @tydeu is there a way of globbing on non git ignored files in a directory?

doc/widgets.md
```diff
@@ -0,0 +1,58 @@
...
+    proc {
+      cmd := "npm"
+      args := #["install"]
```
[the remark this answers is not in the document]
Vtec234: I wasn't familiar with `npm ci`, but [the description](https://docs.npmjs.com/cli/v8/commands/npm-ci) notes that it removes `node_modules`. I'm not sure if we want that in the `jsTarget` as it would end up reinstalling all NPM dependencies every time one runs "Lean 4: Refresh file dependencies".

src/Lean/Widget/UserWidget.lean
```diff
@@ -0,0 +1,183 @@
...
+The editor can use the `Lean.Widget.getWidgetSource` RPC method to
+get this object.
+
+See the [tutorial](doc/widgets.md) above this declaration for more information on
+how to use the widgets system.
+
+-/
+structure WidgetSource where
+  /-- Sourcetext of the code to run.-/
+  sourcetext : String
```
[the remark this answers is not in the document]
gebner: Having a different widget id for every editor feels like the wrong approach to me, are we going to have three different copies of every widget then (in the extreme case)?  (Not that I expect this to happen. I don't think anybody is going to ship elisp or lua over the wire. But we seem to be future-proofing for this scenario.)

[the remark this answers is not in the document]
EdAyers: The way I was thinking about it is that each object pointed to by a `widgetSourceId` is just a sourcefile that the client editor can do whatever with. Maybe in the future a `UserWidget` will point to multiple `WidgetSource` objects, eg if it depends on a library it could depend on multiple ids. So maybe `UserWidget` should have a `javascript : Hash` field in stead?

doc/widgets.md
tydeu: @EdAyers Sadly, no, Lake does not currently provide such a feature.

src/Lean/Widget/UserWidget.lean
```diff
@@ -0,0 +1,183 @@
...
+  let some hash := widgetHashes.find? env widgetSourceId
+    | throw <| RequestError.mk .invalidParams s!"getHash: No source found for {widgetSourceId}."
+  return hash
+
+builtin_initialize registerBuiltinAttribute attributeImpl
+
+end WidgetSource
+
+structure GetWidgetSourceParams where
+  hash : String
```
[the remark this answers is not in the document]
EdAyers: I checked this and actually `ToJson UInt64` does already get sent to a json string. So this is fine and we don't need any special encoding logic.

https://github.com/leanprover/lean4/blob/920fa1109bbbdc6587de58ea5fdc18b30a05293f/src/Lean/Data/Json/FromToJson.lean#L93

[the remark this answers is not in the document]
gebner: Oh, wonderful!  Somebody must have thought ahead!  Then let's go with `UInt64`.

src/Lean/Widget/UserWidget.lean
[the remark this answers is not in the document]
EdAyers: ok I've done this

doc/widgets.md
EdAyers: I've removed this and put it in lean4-samples repository and added a link (currently broken but will fix)

tests/lean/interactive/userWidget.lean
```diff
@@ -0,0 +1,15 @@
...
+  name := "my fancy widget"
+  javascript:= "
+  import * as React from 'react';
+  export default function (props) {
+    return React.createElement('p', {}, 'hello')
+  }"
+}
+
+#widget widget1 (Json.mkObj [])
+  --^ widgets
```
Vtec234, line 15: Is the reason that we can't use the actual call name here that the widget API consists of RPC calls, which we don't generally support in the test script?

Vtec234 (approved):
🌈🌈🌈

tests/lean/interactive/run.lean
```diff
@@ -69,6 +80,58 @@ partial def main (args : List String) : IO Unit := do
...
+            let ps : RpcCallParams := {
+              ps with
+              method := `Lean.Widget.getWidgetSource,
+              params := toJson params,
+            }
+            Ipc.writeRequest ⟨requestNo, "$/lean/rpc/call", ps⟩
+            let resp ← Ipc.readResponseAs requestNo Lean.Widget.WidgetSource
+            IO.eprintln (toJson resp.result)
+            requestNo := requestNo + 1
+        | "RPC" =>
```
Vtec234, line 113: This call does not seem used in any test, unless I am missing something?

EdAyers: no it's not I did use it when I was writing the tests but then I switched to widget because it needs to do two rpc calls to get the widget info

EdAyers: I can remove if you like

tests/lean/interactive/userWidget.lean
EdAyers: the problem was that you have to perform two calls to rpc, and the input to the second call depends on hash so it's easier to write with it's own run handler

src/Lean/Widget/UserWidget.lean
```diff
@@ -0,0 +1,192 @@
...
+
+/-- Output of `getWidgets` RPC.-/
+structure GetWidgetsResponse where
+  widgets : Array UserWidgetInstance
+  deriving ToJson, FromJson
+
+open RequestM in
+/-- Get the `UserWidget`s present at a particular position. -/
+@[serverRpcMethod]
+def getWidgets (args : Lean.Lsp.TextDocumentPositionParams) : RequestM (RequestTask (GetWidgetsResponse)) := do
```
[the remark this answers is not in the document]
EdAyers: No because `UserWidget` doesn't have `javascript` property, only `UserWidgetDefinition`.

src/Lean/Widget/UserWidget.lean
```diff
@@ -0,0 +1,192 @@
...
+  /-- An ESmodule that exports a react component to render. -/
+  javascript: String
+  deriving Inhabited, ToJson, FromJson
+
+structure UserWidget where
+  id : Name
+  /-- Pretty name of widget to display to the user.-/
+  name : String
+  javascriptHash: UInt64
+  deriving Inhabited, ToJson, FromJson
```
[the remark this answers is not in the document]
EdAyers: this would cause sending full javascript every time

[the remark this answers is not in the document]
gebner: Ah didn't see it, that answers two of my questions!

tests/lean/interactive/run.lean
EdAyers: removed

src/Lean/Widget/UserWidget.lean
```diff
@@ -56,8 +56,8 @@ private def WidgetSourceRegistry := SimplePersistentEnvExtension
     (UInt64 × WidgetSource)
     (Std.RBMap UInt64 WidgetSource compare)
 
-instance : Inhabited (Std.RBMap UInt64 WidgetSource compare) := ⟨∅⟩
-instance : Inhabited (WidgetSourceRegistry) := inferInstanceAs (Inhabited (PersistentEnvExtension _ _ (List _ × _)))
+local instance : Inhabited (Std.RBMap UInt64 WidgetSource compare) := ⟨∅⟩
```
gebner, line 59: I think it would be better to add a global `instance : Inhabited (RBMap α β cmp) where default := .empty`.  This is useful more generally.

tests/common.sh
```diff
@@ -28,7 +28,7 @@ function compile_lean {
 function exec_capture {
     # mvar suffixes like in `?m.123` are deterministic but prone to change on minor changes, so strip them
     # also strip intermittent "building '/nix/store/..." messages
-    LEAN_BACKTRACE=0 "$@" 2>&1 | perl -pe 's/(\?\w)\.[0-9]+/\1/g;s!building '\''/nix/.*\n!!g' > "$f.produced.out"
+    LEAN_BACKTRACE=0 "$@" 2>&1 | perl -pe 's/(\?(\w|_\w+))\.[0-9]+/\1/g;s!building '\''/nix/.*\n!!g' > "$f.produced.out"
```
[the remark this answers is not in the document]
Kha: LGTM

Vtec234:
Ok, I think this is good to go!

PatrickMassot:
The PR description still has several TODOs. Should it be updated? Or are they meant as future work?

EdAyers:
Yeah the remaining todos are cyclic dependencies, I'll make issues for them.

doc/widgets.md
```diff
@@ -0,0 +1,37 @@
...
+This JavaScript text must include `import * as React from "react"` in the imports and may not use JSX.
+The default export of the sourcetext must be a React component whose props are an RPC encoding.
+The React component may accept a props argument whose value will be determined for each particular widget instance (below).
+Widget sources may import the `@lean4/infoview` package ([todo] publish on NPM) in order to use
+components such as `InteractiveMessage` to display `MessageData` interactively.
+
+## Using Lake to build your widgets
+
+For larger projects, you can use Lake to create files that will be used as `widgetSource`.
+To learn how to do this, please view the readme of the [WidgetsMinimal sample](https://github.com/leanprover/lean4-samples/tree/main/WidgetsMinimal) ([todo] merge sample).
```
Kha, line 37: You need to add this file to `SUMMARY.md` for it to be included in the mdbook. If we don't, we will surely forget about it :smile: .

src/Lean/Widget/UserWidget.lean
```diff
@@ -0,0 +1,189 @@
...
+      return w
+    else
+      throw <| RequestError.mk .invalidParams s!"No registered user-widget with hash {args.hash}"
+
+open Lean Elab
+
+/--
+  Try to retrieve the `UserWidgetInfo` at a particular position.
+-/
+def widgetInfoAt? (text : FileMap) (t : InfoTree) (hoverPos : String.Pos) : List UserWidgetInfo :=
```
Kha, line 110: Should rather be `widgetInfosAt`?

doc/widgets.md
```diff
@@ -0,0 +1,37 @@
+# The user-widgets system
+
+Proving is an inherently interactive task. Lots of mathematical objects that we use are visual in nature.
+The user-widget system lets users associate React components with the lean document which are then rendered in the Lean VSCode infoview.
```
Kha, line 4:
```suggestion
The user-widget system lets users associate React components with the Lean document which are then rendered in the Lean VSCode infoview.
```
nit, I assume it works on larger files as well

Kha (approved):

src/Lean/Widget/UserWidget.lean
Kha: @EdAyers Apologies for being so nitpicky, but there should not be a `?`. The docstring should be adjusted as well.

leodemoura:
@EdAyers Is it ready to be merged? @Vtec234 and @Kha have already approved the changes. They also look good to me.

EdAyers:
Happy to merge

PR #1242, opened 2022-06-23
Publish fatal progress after worker crash

Kha:


Vtec234 (approved):

src/Lean/Server/Watchdog.lean
```diff
@@ -260,6 +260,7 @@ section ServerM
           -- Worker crashed
           fw.errorPendingRequests o (if exitCode = 1 then ErrorCode.workerExited else ErrorCode.workerCrashed)
             s!"Server process for {fw.doc.meta.uri} crashed, {if exitCode = 1 then "see stderr for exception" else "likely due to a stack overflow or a bug"}."
+          publishProgressAtPos fw.doc.meta 0 o (kind := LeanFileProgressKind.fatalError)
```
gebner, line 263: There is almost certainly a race condition here, when the file worker crashes after being restarted.  Then we'd send the crashed progress notification while the new worker is already running.

I have been wondering if we should refactor the watchdog/file worker code at some point to use proper synchronization primitives (mutexes, queues, ..). There's a lot of essentially unsynchronized accesses (to separate `IO.Ref`s), busy-waiting loops (on contended accesses to `IO.Ref`s), and "accidental" synchronization by reusing the mutexes for stdin/stdout.

leodemoura: @gebner Thanks for raising this issue. Could you please create an issue for the proposed refactoring?

Kha: In this specific case, what about moving the line to `handleCrash`? That one is on the same thread as `startFileWorker` AFAICS.

gebner: To be honest, I don't think it's a very likely race condition.  Or at least not any more likely than all the other unsynchronized `publishDiagnostics` etc. calls.  The purpose of my comment was to build up some motivation for a `Mutex ServerState` type.

PR #1244, opened 2022-06-23
Show term goal at end of term as well

### LINKED ISSUE #1239: Show expected type if cursor is to the immediate right of a hole
Often I'll be in term mode, and I write something like `fun x => _` with my cursor to the right of the hole, and I have to back up a character to see the expected type at the hole. Please could we change the behaviour of the infotree to give the hole info even if we are immediately after the hole? 🙂

Kha:
Resolves #1239

### DIFF
--- a/src/Lean/Server/InfoUtils.lean
+++ b/src/Lean/Server/InfoUtils.lean
@@ -141,9 +141,11 @@ def InfoTree.smallestInfo? (p : Info → Bool) (t : InfoTree) : Option (ContextI
   infos.toArray.getMax? (fun a b => a.1 > b.1) |>.map fun (_, ci, i) => (ci, i)
 
 /-- Find an info node, if any, which should be shown on hover/cursor at position `hoverPos`. -/
-partial def InfoTree.hoverableInfoAt? (t : InfoTree) (hoverPos : String.Pos) (includeStop := false) : Option (ContextInfo × Info) := Id.run do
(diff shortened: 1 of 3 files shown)


src/Lean/Server/InfoUtils.lean
```diff
@@ -299,36 +301,9 @@ where
...
-  t.smallestInfo? fun i => Id.run do
-    if i.contains hoverPos then
-      if let Info.ofTermInfo ti := i then
-        return !ti.stx.isIdent || !headFns.contains i.pos?.get!
-    false
-  where
-    /- Returns the position of the head function symbol, if it is an identifier. -/
-    getHeadFnPos? (s : Syntax) (foundArgs := false) : Option String.Pos :=
-      match s with
-      | `(($s)) => getHeadFnPos? s foundArgs
```
Kha, line 329: I assume this was to make `(f a) b` behave like `f a b`, but is that really intuitive? I could even see myself temporarily replacing `f` with `(f)` just to see the instantiated by unapplied type.

gebner: Your assumption is correct.  It feels a bit counterintuitive that inserting parentheses changes the output (when they're usually no-ops only used to disambiguate precedence).  I'm fine with either version though.

gebner (approved):

src/Lean/Server/InfoUtils.lean
Kha: Let's wait and see then

PR #1245, opened 2022-06-23
doc: Fix overwide view when using Alectryon

ConnorBaker:
When including pages generated by Alectryon, LeanInk, and lean-mdbook we need to set `display: initial` or `display: contents` on the content `div` to prevent the page being scrollable horizontally when it shouldn't be.

I built and tested this against master using nix.

I'm not sure if this problem is specific to LeanInk/lean-mdbook, or if any Alectryon and mdbook combination have this problem. If it's the latter, this fix might be more appropriate upstream rather than here.

### Wide

Screenshot 2022-06-23 at 11 53 43 AM

### Narrow

Screenshot 2022-06-23 at 12 03 38 PM

### Wide

Screenshot 2022-06-23 at 11 56 48 AM

### Narrow

Screenshot 2022-06-23 at 12 03 46 PM

### DIFF
--- a/doc/alectryon-content-fix.css
+++ b/doc/alectryon-content-fix.css
@@ -1,8 +1,0 @@
-.content {
-    /*
-    Use `initial` instead of `contents` to avoid a browser bug which removes
-    the element from the accessibility tree.
-    https://developer.mozilla.org/en-US/docs/Web/CSS/display#display_contents
-     */
-    display: initial;
-}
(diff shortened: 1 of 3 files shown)


Kha:
Thanks! Since we've already cusstomized `alectryon.css`, I think it's fine to include it there. I'm also pretty sure we're the only ones putting Alectryon output in mdbooks.

ConnorBaker:
Awesome! I merged it into `alectryon.css`.

PR #1257, opened 2022-06-28
Fix resolution of executable names on Windows.

lovettchris:
This [lakefile.lean](https://github.com/EdAyers/lean4-samples/blob/rubiks/RubiksCube/lakefile.lean) fails to build on windows because the command 'npm' has to be resolved to `C:\Program Files\nodejs\npm.cmd` and there is a special Windows API that does that called [SearchPath](https://docs.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-searchpatha).  I've tested that this fix works and lake build is able to successfully complete the npm command using a locally patched version of lean.exe.

### DIFF
--- a/src/runtime/process.cpp
+++ b/src/runtime/process.cpp
@@ -176,8 +176,8 @@ static obj_res spawn(string_ref const & proc_name, array_ref<string_ref> const &
         command = buffer;
     }
 
-    if (command.find(' ') != std::string::npos) {
-        // path contains spaces so it needs quotes.
+    if (command.find(' ') != std::string::npos || command.find('\t') != std::string::npos) {
+        // path contains whitespace so it needs quotes.
         command.insert(command.begin(), '"');
         command += "\"";
     }


lovettchris:
> Could you provide references to other stdlib implementations showing that this is the correct approach for such an API?

Most unix-like tools that are using spawn are using the standard C library [_spawnlp](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/spawnlp-wspawnlp) for their Windows version and the version provided in the Microsoft C runtime does a FindExecutable call.  We are not using the std C library, instead we are using the lower level CreateProcess call which requires the file name be fully resolved already.   I'm not sure why we are using CreateProcess  instead of spawn, but probably for stdio redirection?  

Many unix-derrived tools also get this wrong and try and reverse engineer the algorithm in FindExecutable, for example, see [cmake implementation of spawn](https://github.com/Kitware/CMake/blob/10d250f780d324fca0c6ef53fd26422cf12cbdd9/Utilities/cmlibuv/src/win/process.c) which calls `search_path` to compute the `application_path` given to `CreateProcess` and search path is implemented there and is big and complex including the helper function `path_search_walk_ext` which looks for *.exe and *.cmd an *.com file exetensions and so on.  It's just easier to use FindExecutable.

Kha:
My question about other stdlibs stands btw. If [everyone else](https://github.com/search?q=npm.cmd&type=code) has to special-case `npm.cmd` on Windows, we may have to accept that.

Kha:
Another common option with process spawn API is an opt-in `shell` parameter. On Windows I assume that would use `cmd.exe` and would make `npm` work as is.

lovettchris:
>  "Use FindExecutable for documents." Surely we don't want Process.spawn "foo.docx" to open Word?

That sounds bad, let me do some more checking...

lovettchris:
@Kha - I think we still need to merge this so that Lake on Windows behaves better (See https://github.com/leanprover/lean4-samples/pull/3).

leodemoura:
@lovettchris Thanks for working on this issue. It seems there is one pending issue on this thread.

> Could you provide references to other stdlib implementations showing that this is the correct approach for such an API?

Could you please provide references? I think this is a valid concern.

lovettchris:
> Could you please provide references? I think this is a valid concern.

How about the llvm debugger which has a [ProcessLauncherWindows](https://github.com/llvm/llvm-project/blob/main/lldb/source/Host/windows/ProcessLauncherWindows.cpp) that calls the Win32 CreateProcess using a `ProcessLaunchInfo::GetExecutableFile` where this is checked (for example in [MonitoringProcessLauncher](https://github.com/llvm/llvm-project/blob/main/lldb/source/Host/common/MonitoringProcessLauncher.cpp#L39) and when the path does not exist it calls [ResolveExecutableLocation](https://github.com/llvm/llvm-project/blob/main/lldb/source/Host/common/FileSystem.cpp#L330) which calls the [Windows implementation of sys::findProgramByName](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Support/Windows/Program.inc#L38) which uses win32 `Searchpath`.

Kha:
@lovettchris Good to see that others use similar workarounds, but I hope you can appreciate the difference between doing so in a private API in some software with clear use cases versus in a public API in a language's standard library.
By the way, I learned that the batch story is even worse: that passing a .cmd file name to `CreateProcess` works at all is [undocumented behavior](https://github.com/rust-lang/rust/pull/95246#issuecomment-1076634982), with the documentation stating the opposite:

> To run a batch file, you must start the command interpreter; set lpApplicationName to cmd.exe and set lpCommandLine to the following arguments: /c plus the name of the batch file.

Vtec234:
Fwiw, I adjusted the originally mentioned `lakefile` to select `npm.cmd` on Windows, which is perhaps closer to the "select downstream but not in stdlib" approach.

lovettchris:
Thanks, meanwhile I'm trying to fix it with this: https://github.com/leanprover/lean4/pull/1257

leodemoura:
I am closing this PR. @Kha raised an important point, and @lovettchris found an alternative fix.

PR #1259, opened 2022-06-28
Fix empty snapshot list server errors

Kha:
It looks like I didn't properly review the `updateFinishedPrefix` uses

### DIFF
--- a/src/Lean/Server/FileWorker.lean
+++ b/src/Lean/Server/FileWorker.lean
@@ -273,21 +273,22 @@ section Updates
   def updateDocument (newMeta : DocumentMeta) : WorkerM Unit := do
     let ctx ← read
     let oldDoc := (←get).doc
-    let ⟨cmdSnaps, e?⟩ ← oldDoc.cmdSnaps.updateFinishedPrefix
-    match e? with
-    | some ElabTaskError.aborted =>
-      -- This case should not be possible. only the main task aborts tasks and ensures that aborted tasks
-      -- do not show up in `snapshots` of an EditableDocument.
-      throwServerError "Internal server error: elab task was aborted while still in use."
-    | some (ElabTaskError.ioError ioError) => throw ioError
-    | _ => -- No error or EOF
-      -- The watchdog only restarts the file worker when the semantic content of the header changes.
-      -- If e.g. a newline is deleted, it will not restart this file worker, but we still
-      -- need to reparse the header so that the offsets are correct.
-      let (newHeaderStx, newMpState, _) ← Parser.parseHeader newMeta.mkInputContext
-      let newHeaderSnap := { cmdSnaps.finishedPrefix.head! with stx := newHeaderStx, mpState := newMpState }
+    -- The watchdog only restarts the file worker when the semantic content of the header changes.
+    -- If e.g. a newline is deleted, it will not restart this file worker, but we still
+    -- need to reparse the header so that the offsets are correct.
+    let (newHeaderStx, newMpState, _) ← Parser.parseHeader newMeta.mkInputContext
+    let cancelTk ← CancelToken.new
+    -- Wait for at least one snapshot from the old doc, we don't want to unnecessarily re-run `print-paths`
+    let headSnapTask ← oldDoc.cmdSnaps.waitHead?
+    let newSnaps ← EIO.mapTask (ε := ElabTaskError) (t := headSnapTask) fun headSnap?? => do
+      let headSnap? ← MonadExcept.ofExcept headSnap??
+      -- There is always at least one snapshot absent exceptions
+      let headSnap := headSnap?.get!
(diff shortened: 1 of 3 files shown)


src/Lean/Server/FileWorker.lean
```diff
@@ -273,21 +273,22 @@ section Updates
...
+    let headSnapTask ← oldDoc.cmdSnaps.waitHead?
+    let newSnaps ← EIO.mapTask (ε := ElabTaskError) (t := headSnapTask) fun headSnap?? => do
+      let headSnap? ← MonadExcept.ofExcept headSnap??
+      -- There is always at least one snapshot absent exceptions
+      let headSnap := headSnap?.get!
+      let newHeaderSnap := { headSnap with stx := newHeaderStx, mpState := newMpState }
       oldDoc.cancelTk.set
       let changePos := oldDoc.meta.text.source.firstDiffPos newMeta.text.source
+      -- Ignore exceptions, we are only interested in the successful snapshots
+      let (cmdSnaps, _) ← oldDoc.cmdSnaps.updateFinishedPrefix
```
Kha, line 291: A coincidental simplification, but it seems reasonable to ignore errors from the old document. This subsumes the "Internal server error" from above.

leodemoura:
@Kha I am merging this PR as-is. My system became unusable today because of the panic error messages. It was happening every time I opened Emacs and then `src/Lean/Elab/Command.lean`. This PR seems to fix the problem after `update-stage0`.

src/Lean/Server/FileWorker.lean
Vtec234: Should we still have the check for `aborted` here? It's basically just an assertion to make sure we don't violate assumptions internally.

src/Lean/Server/FileWorker.lean
```diff
@@ -273,21 +273,22 @@ section Updates
   def updateDocument (newMeta : DocumentMeta) : WorkerM Unit := do
     let ctx ← read
     let oldDoc := (←get).doc
-    let ⟨cmdSnaps, e?⟩ ← oldDoc.cmdSnaps.updateFinishedPrefix
```
Vtec234, line 276: What was the bug? I am not seeing it.

src/Lean/Server/FileWorker.lean
Kha: I don't know, does the assertion have any bearing on the surrounding code? As the comment says, all we're here for is recycling snapshots before the old document gets dumped.

src/Lean/Server/FileWorker.lean
Kha: `updateFinishedPrefix` may find that not even the first (header) snapshot is finished yet, in which case `cmdSnaps.finishedPrefix.head!` panics

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -237,7 +237,7 @@ open Parser.Command in
 partial def handleDocumentSymbol (_ : DocumentSymbolParams)
     : RequestM (RequestTask DocumentSymbolResult) := do
   let doc ← readDoc
-  asTask do
+  mapTask (← doc.cmdSnaps.waitHead?) fun _ => do
```
Vtec234, line 240: Was this needed, given that we bind the `headerTask` before processing any request?

src/Lean/Server/FileWorker.lean
Vtec234: Thanks!

src/Lean/Server/FileWorker.lean
Vtec234: No, it is a "non-local" check which verifies that a totally different part of the code behaves correctly. Maybe this isn't great practice so makes sense to leave out.

src/Lean/Server/FileWorker/RequestHandling.lean
Kha: You are right, I had already forgotten about that. Should we revert it or keep it just to be safe?

Kha: For `RequestContext` specifically, we might also want to add a field for the header snapshot just to make this fact explicit, even if it is redundant with `doc.cmdSnaps`

Vtec234: Stating this in the type system as you suggest is a good idea as the handlers don't need to do this kind of defensive programming then. We should also document that the header snap is duplicated in the request context.

Kha: I didn't think this through, it can't replace `getLast!` in this case (and `cmdSnaps.getLastD headerSnap` would be plain weird IMO). I'd rather wait than introduce the field without a use case.

PR #1264, opened 2022-06-29
`AsyncList` refactorings

Kha:
See respective commit messages

### DIFF
--- a/src/Lean/Server/AsyncList.lean
+++ b/src/Lean/Server/AsyncList.lean
@@ -11,11 +11,11 @@ namespace IO
 universe u v
 
 /-- An async IO list is like a lazy list but instead of being *unevaluated* `Thunk`s,
-lazy tails are `Task`s *being evaluated asynchronously*. A tail can signal the end
+`delayed` suffixes are `Task`s *being evaluated asynchronously*. A delayed suffix can signal the end
 of computation (successful or due to a failure) with a terminating value of type `ε`. -/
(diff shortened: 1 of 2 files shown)


src/Lean/Server/AsyncList.lean
```diff
@@ -87,33 +87,23 @@ partial def waitFind? (p : α → Bool) : AsyncList ε α → BaseIO (Task $ Exc
...
-  | asyncTail tl => do
+  | delayed tl => do
     BaseIO.bindTask tl fun
       | Except.ok tl   => tl.waitFind? p
       | Except.error e => return Task.pure <| Except.error e
 
-/-- Extends the `finishedPrefix` as far as possible. If computation was ongoing
-and has finished, also returns the terminating value. -/
-partial def updateFinishedPrefix : AsyncList ε α → BaseIO (AsyncList ε α × Option ε)
+/-- Compute the already-computed prefix of the list. If computation has finished with an error, return it as well. -/
```
Vtec234, line 95:
```suggestion
/-- Return the already-computed prefix of the list. If computation has finished with a terminating value,
return it as well. -/
```
Since `ε` is the type of terminating values that can be either an error or a successful end-of-computation signal.

Kha: Right, I was wondering about that part in the previous docstring. Do you mean `ElabTaskError.aborted` by this? I don't really think of it as successful termination.

PR #1270, opened 2022-06-30
fix: only trigger TPIL build on nightly releases

lovettchris:
Related to https://github.com/leanprover/theorem_proving_in_lean4/pull/42


Kha:
What about using a corresponding `cron:` trigger in the TPIL repo instead? I think that would scale a little nicer if we start enabling nightly CI for many repos

lovettchris:
> What about using a corresponding `cron:` trigger in the TPIL repo instead? I think that would scale a little nicer if we start enabling nightly CI for many repos

Done.  See related PR https://github.com/leanprover/theorem_proving_in_lean4/pull/44

lovettchris:
@Kha - why did you close the PR instead of merging the change?

Kha:
Aah, I was confused by the PR title!

PR #1271, opened 2022-06-30
fix: Lean4 code snippets need to specify the lean4 language service

lovettchris:
fix: Lean4 code snippets need to specify the lean4 language service id which is `lean4` otherwise hover tips in vscode don't get colorized properly when lean3 extension is missing, and besides we want to colorize them using lean4 grammar, not the lean3 grammar.

### DIFF
--- a/src/Lean/Server/FileWorker/RequestHandling.lean
+++ b/src/Lean/Server/FileWorker/RequestHandling.lean
@@ -163,7 +163,7 @@ def handlePlainGoal (p : PlainGoalParams)
       { goals := #[], rendered := "no goals" }
     else
       let goalStrs := goals.map (toString ·.pretty)
-      let goalBlocks := goalStrs.map fun goal => s!"```lean
+      let goalBlocks := goalStrs.map fun goal => s!"```lean4
 {goal}
 ```"
       let md := String.intercalate "\n---\n" goalBlocks.toList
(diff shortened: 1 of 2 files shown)


lovettchris:
I see lots of [test baselines will need to be updated](https://github.com/leanprover/lean4/runs/7141500545?check_suite_focus=true), before I do that I want to check if this is the right change to make, also do you have a tool that just says "copy test output and promote it to be the new expected output" which might save me a lot of work...?

leodemoura:
Could you please try to find an alternative solution for the vscode issue?
We want to avoid
```` 
```lean4
````
because we all hope that in a near future the `4` in `Lean 4` will not be needed anymore.

lovettchris:
> Could you please try to find an alternative solution for the vscode issue?

@gebner might be able to confirm, but I think the only way to get lean4 syntax highlighting on lean4 code snippet using the string `lean` would require the lean4 VS code extension to register itself as language `lean` and thereby give up on side-by-side compatibility with lean3 extension.  To get lean3 people would have to disable lean4 extension and vice versa.  Are we ready to do that?

lovettchris:
I looked into how vscode is working some more and turns out is coloring code snippets with defined using triple back tick and no language using the lean4 grammar, so we can just do that and cancel this PR.

PR #1272, opened 2022-07-01
Relaxed antiquotation parsing

Kha:
This is an ~~incomplete~~ implementation of an idea I mentioned in the Typed Macros PR: https://github.com/leanprover/lean4/pull/1251#issuecomment-1166242297 (the first half). In short, when encountering an antiquotation `$x` while running a parser `p`, we do not immediately assume that the entire antiquotation is for `p` but check if we `p` accepts `$x` *and further tokens behind it*. Thus the example `` `(structInstField| $id := $val) `` is interpreted as `` `(structInstField| $id:ident := $val:term) ``, not `` `(structInstField| $id:structInstField <ERROR: expected ')'> ``.

Unfortunately, there are some regressions. One of them is defensible I think ~~, but the other would likely require us to adapt the Lean grammar; see below.~~

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,6 +1,8 @@
 Unreleased
 ---------
 
+* [Relaxed antiquotation parsing](https://github.com/leanprover/lean4/pull/1272) further reduces the need for explicit `$x:p` antiquotation kind annotations.
+
 * Add support for computed fields in inductives. Example:
   ```lean
   inductive Exp


src/Init/NotationExtra.lean
```diff
@@ -254,7 +254,7 @@ macro_rules
 syntax "repeat " doSeq " until " term : doElem
 
 macro_rules
-  | `(doElem| repeat $seq until $cond) => `(doElem| repeat do $seq; if $cond then break)
+  | `(doElem| repeat $seq until $cond) => `(doElem| repeat do $seq:doSeq; if $cond then break)
```
Kha, line 257: I think this one makes sense: how many people would read the original line and conclude that `$seq` is not the first of two elements of `do`, but the entire content of it? With the relaxed antiquotation parser, we do now get ``seq : TSyntax `doElem`` in the modified line.
The fix only makes this marginally clearer; I would have liked `(do $seq); ...` better, which does interpret `seq` as a `doSeq`. However, the parens terminate e.g. mutable variables, which is undesirable.

src/Lean/Elab/MutualDef.lean
```diff
@@ -169,7 +169,7 @@ private def expandWhereStructInst : Macro
       | `(letDecl|$decl:letIdDecl)   => pure decl
       | _                               => Macro.throwUnsupported
     let structInstFields ← letIdDecls.mapM fun
-      | stx@`(letIdDecl|$id:ident $binders* $[: $ty?]? := $val) => withRef stx do
+      | stx@`(letIdDecl|$id:ident $binders:letIdBinder* $[: $ty?]? := $val) => withRef stx do
```
Kha, line 172: This is the more problematic issue: a single `letIdBinder` also accepts a list of idents (or holes), so by the new semantics `$binders*` is interpreted as such because the "content parse" `$binders:ident*` consumes one more token than the "antiquotation parse" `$binders:letIdBinder` without the star. This is because the star is technically not part of the antiquotation but a separate modifier parsed afterwards. Note also that while this use site failed to compile, others like
```
  | `(have $x $bs* $[: $type]? := $val; $body)            => `(let_fun $x $bs* $[: $type]? := $val; $body)
```
are still type-correct, but now are more specific than should be, breaking many tests.

I think this points to a weakness in our grammar: we should avoid nesting sequences using the same separator (none, in this case):
```
letIdLhs ::= ident letIdBinder* ...
letIdBinder ::= simpleBinderWithoutType | ...
simpleBinderWithoutType ::= binderIdent*
```
I think we should replace `simpleBinderWithoutType` with a single `binderIdent` to avoid this inherent garmmar ambiguity. This would break compatibility with `simpleBinder` as noted in a comment above the former's definition, but hopefully these parts are already covered by `TSyntax` and thus simple to refactor.

Kha: `simpleBinder`, and perhaps others, should probably be changed as well, but I think this would be a change for the better: currently we accept the following example inspired by @gebner's cursed juxtaposition:
```
#check fun a : Type 0f => _  -- fun (a : Type 0) (f : _) => _
```

src/Init/NotationExtra.lean
gebner: To be fair, I can empathize with Lean here.  I've also found the zoo of do-notation nonterminals a bit confusing at times.

Indentation would be another way to disambiguate this example:
```lean
`(doElem|
  repeat
    do $seq
    if $cond then break)
```

Excessive use of braces helps too:
```lean
`(doElem| repeat do { do $seq }; if $cond then break)
```

We could also remove the slightly misleading syntax for nested do and replace it by braces.  (I think it's misleading because it's not a boundary for `return` like the normal `do`.)
```lean
syntax "{" doSeq "}" : doElem
```

A coercion from `doSeq` to `doSeqItem` could also help (but I guess you don't want a coercion to insert new tokens).

Kha: Yeah, it might be good to clean up the grammar there. And yes introducing tokens in a coercion probably isn't a good idea due to lack of access to `getRef` (though I even thought about a monadic coercion class that is automatically used by antiquotations).

Kha:
I'm going ahead and merging this as it should be a strict improvement now and I've been confused multiple times about why it didn't work on master yet.

PR #1273, opened 2022-07-01
chore: update Lake

tydeu:
Update to v3.2.0


tydeu:
@Kha / @leodemoura, this is ready to merge (the one failure also occurs on master and is not due to Lake).

PR #1276, opened 2022-07-02
fix: bound syntax kind at `v:(ppSpace ident)` etc.

### LINKED ISSUE #1274: Macro arguments should ignore `ppSpace` when determining syntax kind
In the following, `t` has the type `TSyntax Name.anonymous` instead of the expected `Ident` and therefore doesn't compile:
```lean
macro "👉" t:(ppSpace ident) : term => `($t)
```

### Versions

Nightly from July 2.

Kha:
With the information from #1229, we can now actually solve this in a general way

Fixes #1274

### DIFF
--- a/src/Lean/Elab/MacroArgUtil.lean
+++ b/src/Lean/Elab/MacroArgUtil.lean
@@ -45,8 +45,7 @@ where
   mkSplicePat (kind : SyntaxNodeKind) (stx : TSyntax `stx) (id : Term) (suffix : String) : CommandElabM Term :=
     return ⟨mkNullNode #[mkAntiquotSuffixSpliceNode kind (← mkAntiquotNode stx id) suffix]⟩
   mkAntiquotNode : TSyntax `stx → Term → CommandElabM Term
-    | `(stx| ($stx)), term => mkAntiquotNode stx term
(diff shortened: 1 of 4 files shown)


src/Lean/Elab/MacroArgUtil.lean
```diff
@@ -63,7 +62,18 @@ where
...
-    | stx, term =>
+    | stx, term => do
+      -- can't match against `` `(stx| ($stxs*)) `` as `*` is interpreted as the `stx` operator
+      if stx.raw.isOfKind ``Parser.Syntax.paren then
+        -- translate argument `v:(p1 ... pn)` where all but one `pi` produce zero nodes to
+        -- `v:pi` using that single `pi`
+        let nonNullaryNodes ← stx.raw[1].getArgs.filterM fun
+          | `(stx| $id:ident$[:$_]?) | `(stx| $id:ident($_)) => do
+            let info ← Parser.getParserAliasInfo id.getId
+            return info.stackSz? != some 0
```
Kha, line 73: This does not cover recursive cases such as `(group(ppSpace ppSpace) ident)`, but I couldn't see any practical use cases there...

Kha:
Hah, not sure we've had a broken stage 3 before.

PR #1284, opened 2022-07-05
fix unused variables linter around macros

### LINKED ISSUE #1266: Unused variables warning on macros/elaborators that expand to `opaque`
### Prerequisites

### Description

Macros and elaborators which expand syntax to opaque declarations  with named parameters produce unused variable warnings on the parameter names.

### Steps to Reproduce

```lean
macro "opaque' " id:declId sig:declSig : command => `(opaque $id $sig)
opaque' foo (x : UInt32) : UInt32 -- warning: unused variable `x`
```

**Expected behavior:**

No warning.

**Actual behavior:**

Unused variable warning.

**Reproduces how often:**

Always.

### Versions

Windows 20H2
`Lean (version 4.0.0-nightly-2022-06-17, commit f0e2696ec82b, Release)`

### Additional Information

I stumbled onto this because of [Alloy](https://github.com/tydeu/lean4-alloy) which expands its `extern` definitions to opaque declarations and thereby produces lots of unused variable warnings.

larsk21:
- fix reference search when multiple binders / non-binders occur in the same position (possible with macros)
- consider macro expansions when deciding whether to ignore an unused variable
- ~allow adding custom ignore functions~

fixes #1266

### DIFF
--- a/src/Lean/Linter/Basic.lean
+++ b/src/Lean/Linter/Basic.lean
@@ -129,7 +129,7 @@ def unusedVariables : Linter := fun cmdStx => do
     -- evaluate ignore functions on macro expansion outputs
     if ← infoTrees.anyM fun tree => do
       if let some macroExpansions ← collectMacroExpansions? range tree then
-        return macroExpansions.reverse.any fun expansion =>
+        return macroExpansions.any fun expansion =>
           if let some stack := findSyntaxStack? expansion.output declStx then
             ignoreFns.any (· declStx stack)
           else
--- a/src/Lean/Linter/Util.lean
+++ b/src/Lean/Linter/Util.lean
@@ -20,9 +20,16 @@ do
   let messages := (← get).messages |>.add (mkMessageCore ctx.fileName ctx.fileMap content severity range.start range.stop)
   modify ({ · with messages := messages })
 
+/-- Go upwards through the given `tree` starting from the smallest node that
+contains the given `range` and collect all `MacroExpansionInfo`s on the way up.
+The result is `some []` if no `MacroExpansionInfo` was found on the way and
+`none` if no `InfoTree` node was found that covers the given `range`.
+
+Return the result reversed, s.t. the macro expansion that would be applied to
+the original syntax first is the first element of the returned list. -/
 def collectMacroExpansions? {m} [Monad m] (range : String.Range) (tree : Elab.InfoTree) : m <| Option <| List Elab.MacroExpansionInfo := do
     if let .some <| .some result ← go then
-      return some result
+      return some result.reverse
     else
       return none
 where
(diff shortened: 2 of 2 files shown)


src/Lean/Linter/Basic.lean
```diff
@@ -25,6 +25,105 @@ def getLinterUnusedVariables (o : Options) : Bool := o.get linter.unusedVariable
...
+    pos == 0 &&
+    [``Lean.Parser.Command.optDeclSig, ``Lean.Parser.Command.declSig].any (stx.isOfKind ·)))
+
+-- in in constructor or structure binder
+builtin_initialize addUnusedVariablesIgnoreFn (fun _ stack _ =>
+  stackMatches stack [`null, none, `null, ``Lean.Parser.Command.optDeclSig, none] &&
+  (stack.get? 4 |>.any fun (stx, _) =>
+    [``Lean.Parser.Command.ctor, ``Lean.Parser.Command.structSimpleBinder].any (stx.isOfKind ·)))
+
+-- is in constant or axiom
```
Kha, line 63:
```suggestion
-- is in opaque or axiom
```

Kha:
!bench

src/Lean/Linter/Util.lean
```diff
@@ -20,6 +20,29 @@ do
   let messages := (← get).messages |>.add (mkMessageCore ctx.fileName ctx.fileMap content severity range.start range.stop)
   modify ({ · with messages := messages })
 
+def collectMacroExpansions? {m} [Monad m] (range : String.Range) (tree : Elab.InfoTree) : m <| Option <| List Elab.MacroExpansionInfo := do
```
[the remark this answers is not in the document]
larsk21: `go` returns `some _` if the targeted syntax range is in the given node or its children. If instead of `some []` it would return `none`, then nodes higher up in the InfoTree wouldn't know which list returned by their children is the one to prepend a `MacroExpansionInfo` to (`filterMap id` does this selection).

`some []` is returned to indicate that the correct InfoTree node was found, but there were no `MacroExpansionInfo`s on the way up yet.

Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
!bench

src/Lean/Linter/Basic.lean
```diff
@@ -76,50 +76,71 @@ def unusedVariables : Linter := fun stx => do
...
+      if let some macroExpansions ← collectMacroExpansions? range tree then
+        return macroExpansions.reverse.any fun expansion =>
+          if let some stack := findSyntaxStack? expansion.output declStx then
+            ignoreFns.any (· declStx stack)
+          else
+            false
+      else
+        return false
+    then
       continue
```
Kha, line 140: Something like this might be a little clearer?
```suggestion
    for tree in infoTrees do
      if let some macroExpansions ← collectMacroExpansions? range tree then
        for expansion in macroExpansions.reverse do
          if let some stack := findSyntaxStack? expansion.output declStx then
            if ignoreFns.any (· declStx stack) then
              continue
```

Kha: The `reverse` probably deserves a short comment

src/Lean/Linter/Util.lean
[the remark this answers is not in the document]
Kha: Could you add a docstring here?

src/Lean/Linter/Basic.lean
larsk21: Won't the `continue` target the innermost `for` loop instead of the one going through the found variables?

Kha: Ahh, you're right of course.

src/Lean/Linter/Util.lean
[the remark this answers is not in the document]
Kha: :+1:

PR #1290, opened 2022-07-07
Computed fields

gebner:
Add support for computed fields in inductives. Example:
```lean
inductive Exp
  | var (i : Nat)
  | app (a b : Exp)
with
  hash : Exp → Nat
    | .var i => i
    | .app a b => a.hash * b.hash + 1
```
The result of the `Exp.hash` function is then stored as an extra "computed" field in the `.var` and `.app` constructors; `Exp.hash` accesses this field and thus runs in constant time (even on dag-like values).

The implementation uses `implementedBy` to replace `Exp` by an inductive that has an extra `hash` field (overriding the functions `Exp.var`, `Exp.app`, `Exp.hash`, and `Exp.casesOn`):
```lean
inductive Exp._impl
  | var (hash : Nat) (i : Nat)
  | app (hash : Nat) (a b : Exp)
```
The fields are only added to constructors with arguments.  Argument-less constructors are represented as scalars in the runtime, and the computed field value is already stored as a closed value so we don't gain anything by allocating a heap cell for it.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,6 +1,19 @@
 Unreleased
 ---------
 
+* Add support for computed fields in inductives. Example:
+  ```lean
+  inductive Exp
+    | var (i : Nat)
+    | app (a b : Exp)
+  with
+    @[computedField] hash : Exp → Nat
+      | .var i => i
+      | .app a b => a.hash * b.hash + 1
+  ```
+  The result of the `Exp.hash` function is then stored as an extra "computed" field in the `.var` and `.app` constructors;
+  `Exp.hash` accesses this field and thus runs in constant time (even on dag-like values).
+
 * Update `a[i]` notation. It is now based on the typeclass
   ```lean
   class GetElem (Cont : Type u) (Idx : Type v) (Elem : outParam (Type w)) (Dom : outParam (Cont → Idx → Prop)) where
@@ -102,18 +115,6 @@ Unreleased
     | n+1 => (n+1) * Nat.fact n
   end Nat
   ```
-* Add support for computed fields in inductives. Example:
-  ```lean
-  inductive Exp
-    | var (i : Nat)
-    | app (a b : Exp)
-  with
-    hash : Exp → Nat
-      | .var i => i
-      | .app a b => a.hash * b.hash + 1
-  ```
-  The result of the `Exp.hash` function is then stored as an extra "computed" field in the `.var` and `.app` constructors;
-  `Exp.hash` accesses this field and thus runs in constant time (even on dag-like values).
 
 * Update Lake to v3.2.1. See the [v3.2.1 release notes](https://github.com/leanprover/lake/releases/tag/v3.2.1) for detailed changes.
 


leodemoura:
@gebner This is great. I will be merging the "easy" commits first.
I didn't review the PR yet, but I am wondering whether there are limitations. For example, does it support mutually inductive types too? What about nested inductive types?
Additional tests would be nice too.

gebner:
> For example, does it support mutually inductive types too?

Almost, the only limitation is that the inductive command cannot elaborate mutually recursive computed functions (see other comment). The following works just fine though:
```
mutual
  inductive A
    | a (b : B)
    | b (b : B)
  with
    f : A → Nat
      | .a c => 32
      | .b c => 42

  inductive B
    | c (a : A)
    | d
  with
    f : B → Nat
      | .c a => a.f
      | d => 0
end
```

It also supports parameters and indices and implicit arguments.

> What about nested inductive types?

Yes, they're supported too.  (The encoding doesn't care about recursive occurrences of the inductive type, in fact `Exp._impl` is always nonrecursive.)

> Additional tests would be nice too.

I'll do that.

gebner:
> I am wondering whether there are limitations.

Some noteworthy limitations:
 - It only works if every inductive in the mutual block has more than two constructors (otherwise we might need to override projections as well, I didn't check).
 - The types of the computed fields may not depend on the indices, or on the value of the inductive type.  (So `hash2 : (e : Exp) → Fin e.hash` is excluded.)
 - The universe of the computed fields must fit into the inductive type (i.e., adding the computed field as an extra field must type check).
 - The frontend cannot elaborate mutually recursive computed fields.

All of them are fixable with some effort, but none of these limitations affect the `Expr`/`Name`/`Level` use case.

gebner:
Ah, I've found a very problematic limitation.  The encoding disables an optimization in the compiler:
```lean
def getAppFn : Exp → Exp
  | .app f _ => getAppFn f
  | e => e
```

With computed fields, the second arm is no longer a no-op.  It generates the following code:
```
  Exp.var._impl →
    let x_2 : u32 := sproj[0, 8] x_1;
    let x_3 : obj := Exp.var._override x_2;
    ret x_3
```

src/Lean/Elab/Inductive.lean
```diff
@@ -798,6 +806,20 @@ private def applyDerivingHandlers (views : Array InductiveView) : CommandElabM U
...
         classView.applyHandlers declNames
 
+private def applyComputedFields (indViews : Array InductiveView) : CommandElabM Unit := do
+  if indViews.all (·.computedFields.isEmpty) then return
+
+  let mut computedFields := #[]
+  for indView@{declName, ..} in indViews do
+    for {ref, fieldId, type, matchAlts ..} in indView.computedFields do
+      elabCommand <|
+        ← `(noncomputable def%$ref $(mkIdent <| declName.updatePrefix .anonymous ++ fieldId):ident : $type $matchAlts:matchAlts)
```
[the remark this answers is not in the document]
tydeu: Would it be possible to use the `_root_` workaround @leodemoura mentioned in the ["conflicting namespaces in mutual declaration" Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/conflicting.20namespaces.20in.20mutual.20declaration/near/288877156) to get the naïve approach to work?

src/Lean/Parser/Command.lean
```diff
@@ -80,7 +80,10 @@ def «example»        := leading_parser "example" >> ppIndent declSig >> declVa
 def ctor             := leading_parser "\n| " >> ppIndent (declModifiers true >> ident >> optDeclSig)
 def derivingClasses  := sepBy1 (group (ident >> optional (" with " >> Term.structInst))) ", "
 def optDeriving      := leading_parser optional (ppLine >> atomic ("deriving " >> notSymbol "instance") >> derivingClasses)
-def «inductive»      := leading_parser "inductive " >> declId >> optDeclSig >> optional (symbol " :=" <|> " where") >> many ctor >> optDeriving
+def computedField    := leading_parser ident >> " : " >> termParser >> Term.matchAlts
```
tydeu, line 83: I noticed this syntax doesn't support any declaration modifiers (e.g., documentation, visibility, attributes, etc.). Is this for expediency or are there real limitations to what kind of modifiers these fields can support?

leodemoura:
I played with this PR today. I think it is problematic to replace `casesOn` and constructors before erasure using `implementedBy`. The first problem is that many `match`-related optimizations do not kick in. For example, the `match` in the following code is only eliminated much later at the IR level.
```lean
def Exp.f (e : Exp) : Exp :=
   match app e e with
   | var _   => e
   | app a _ => g a
```
The IR compiler input is
```lean
def Exp.f (x_1 : obj) : obj :=
  let x_2 : obj := Exp.app._override x_1 x_1;
  case x_2 : obj of
  Exp.var._impl →
    ret x_1
  Exp.app._impl →
    let x_3 : obj := proj[1] x_2;
    let x_4 : obj := Exp.g x_3;
    ret x_4
``` 
All the floating-match optimizations are affected by this issue.
This PR also exposes a bug in the compiler, `unsafeCast` is being inlined before erasure. 

We should perform the replacement during erasure (step: `erase_irrelevant`). We don't have a mechanism for doing that, but we could add a simple one that is good enough for this PR.

src/Lean/Parser/Command.lean
gebner: Thanks for the suggestion; docstrings, visibility, and attributes are supported now.

gebner:
> We should perform the replacement during erasure (step: `erase_irrelevant`).

Thanks for the suggestion.  The two match-optimizations mentioned in this thread now generate the expected code (see test).

Kha:
Syntactic nit: `inductive ... with` might make people think of inductive-recursive types first. Though a `mutual` block would be even more natural for those. I just wonder if we should come up with a slightly more specialized syntax (or refine the proposed one with an attribute?) in case we want to add *any* other kind of definition that needs to be declared alongside an inductive type later on.

gebner:
I'm open for concrete syntax suggestions.  I've also thought of using `mutual` blocks, but then the namespaces get wonky.

```lean
mutual
  inductive Exp
    | var (i : Nat)
    | app (a b : Exp)

  @[computedField Exp]
  def _root_.Exp.hash : Exp → Nat
    | .var i => i
    | .app a b => a.hash + b.hash
end
```

Kha:
I think I actually like `with` better as there is no actual mutuality. We could simply reject any `with` definition that is not tagged `[computedField]` for a start.

PR #1291, opened 2022-07-07
Use computed fields for `Expr`, `Level`, and `Name`

gebner:
Most of the changes are fairly mechanical:
 - Remove the underscores.
 - Replace `| .forallE n d b c => c.binderInfo` by `| .forallE n d b c => c`
 - stage0 updates

(I have added the `BinderInfo` argument at the *end*, so that the pattern `.forallE _ _ b _` doesn't quietly change its meaning.)

All other non-mechanical changes:
 - The first commit adds some extra material on the `LawfulBEq` class, which was useful to define the `DecidableEq Name` and `LawfulBEq Name` instances (`Init/Core.lean`, `Init/Meta.lean`, `Init/Data/List/Basic.lean`, `Init/Data/Nat/Linear.lean`, `Init/Prod.lean`).  The `Init/Meta.lean` file contains the lawful instances.
 - Name patterns no longer need to be special cases (`PatternVar.lean`)
 - `Init/Prelude.lean` contains the `Name` refactoring: `hash` is moved to a computed field, and the smart constructors are made abbreviations (because the RPC encoding deriver otherwise fails to synthesize an instance).
 - `Lean/Expr.lean` contains the `Expr` refactoring: `data` is moved to a computed field
 - `Lean/Level.lean` contains the `Level` refactoring: `data` is moved to a computed field
 - `kernel/level.h`: adapted, `Level.zero` is now a scalar
 - `ReduceEval.lean` is adapted to the new name constructor arity.
 - `AC/Main.lean` simplies a `@[pattern]` definition
 - `SizeOf.lean`: removed an argument from a simp lemma (after this PR, we can probably auto-derive the sizeof instance for names)
 - `lake` is updated to https://github.com/leanprover/lake/pull/96

If there was a way to hide the underscore removals from the diff, it would be really short.


Kha:
!bench

Kha:
Something is rotten in the state of Benchmark...

Kha:
!bench

gebner:
Now the diff is a net reduction of 26kLOC. :smile:  @Kha Could you please start a new benchmark run?

gebner:
Oops, forgot the borrow annotation. :-/  @Kha Can I bother you again?

Kha:
Actually you might have sufficient permissions to trigger it yourself?

gebner:
!bench

gebner:
!bench

gebner:
!bench

gebner:
!bench

gebner:
I'm still at a loss where the 5% stdlib runtime increase comes from, but I've found an exciting related bug.  It turns out that `Expr.updateApp!` and co. have been broken for some time, and *always* constructed a new expression.  As a result, the following printed false:
```lean
import Lean
open Lean

#eval
  let e := mkApp (mkSort levelZero) (mkSort levelZero)
  let e' := e.replace fun _ => none
  ptrAddrUnsafe e == ptrAddrUnsafe e'
```

The reason is a missed optimization:
```lean

inductive Exp
  | var (i : Nat)
  | app (a b : Exp)

def Exp.hash : Exp → Nat
    | .var i => i
    | .app a b => a.hash + b.hash

def Exp.complicatedHash : Exp → Nat
  | app a b =>
    -- This always constructs an `app a b`-object, even though we could reuse the input.
    (app a b).hash
  | e => e.hash
```

I've now reimplemented all the `update*` functions in Lean and they now behave as they should.

gebner:
Also somewhat concerningly, `Level.updateMax!` is not referentially transparent (since it applies the simplification only in one branch):
```lean
import Lean
open Lean

@[noinline] def noinline (a : α) := a

#eval --- false?!?!?!?
  let b := levelZero
  let a1 := mkLevelParam `a
  let a2 := mkLevelParam (noinline `a)
  let l := mkLevelMax a1 b
  (l.updateMax! a1 b) == (l.updateMax! a2 b)
```

gebner:
!bench

leodemoura:
>  It turns out that Expr.updateApp! and co. have been broken for some time

Thanks for reporting this issue. I pushed a fix for it.

leodemoura:
> Also somewhat concerningly, Level.updateMax! is not referentially transparent (since it applies the simplification only in one branch):

Pushed a fix for this too.

gebner:
!bench

gebner:
!bench

src/Init/Data/Prod.lean
```diff
@@ -7,5 +7,7 @@ prelude
 import Init.SimpLemmas
 
 instance [BEq α] [BEq β] [LawfulBEq α] [LawfulBEq β] : LawfulBEq (α × β) where
-  eq_of_beq {a b} h := by cases a; cases b; simp [BEq.beq] at h; have ⟨h₁, h₂⟩ := h; rw [eq_of_beq h₁, eq_of_beq h₂]
+  eq_of_beq {a b} (h : a.1 == b.1 && a.2 == b.2) := by
+    cases a; cases b
+    refine congr (congrArg _ (eq_of_beq ?_)) (eq_of_beq ?_) <;> simp_all
```
leodemoura, line 12: Same here.

gebner: Those are from the first commit, which added the `beq_iff_eq` simp lemma.  The previous proof used a "nonterminal simp", which are very likely to break when changing the library.

src/Lean/ToExpr.lean
```diff
@@ -35,9 +35,9 @@ instance : ToExpr Unit where
   toTypeExpr := mkConst ``Unit
 
 def Name.toExprAux : Name → Expr
-  | Name.anonymous  => mkConst `Lean.Name.anonymous
-  | Name.str p s _  => mkAppB (mkConst ``Lean.Name.mkStr) (toExprAux p) (toExpr s)
-  | Name.num p n _  => mkAppB (mkConst ``Lean.Name.mkNum) (toExprAux p) (toExpr n)
+  | .anonymous => mkConst ``Lean.Name.anonymous
+  | .str p s ..=> mkApp2 (mkConst ``Lean.Name.str) (toExprAux p) (toExpr s)
```
leodemoura, line 39: Unnecessary `..`

leodemoura:
I have only minor issues. We can address them after the merge. I just enabled the auto-merge.

PR #1296, opened 2022-07-08
Avoid `simpleBinder` nesting

Kha:
See https://github.com/leanprover/lean4/pull/1272#discussion_r911873614


Kha:
The code changes suggest that it would be nice to have `Term.binderIdent` be a proper syntax kind, in which case it could be unified with `Lean.binderIdent` from `Init.Tactics`, which I *might* do in a follow-up PR.

gebner:
> The code changes suggest that it would be nice to have Term.binderIdent be a proper syntax kind, in which case it could be unified with Lean.binderIdent from Init.Tactics, which I might do in a follow-up PR.

Yes, that's a very good idea.  Please also unify it with `ident <|> "_"` when you're at it.  See also: https://github.com/leanprover-community/mathport/pull/152#issuecomment-1173841553

src/Lean/Elab/Binders.lean
```diff
@@ -286,17 +296,13 @@ partial def expandFunBinders (binders : Array Syntax) (body : Syntax) : MacroM (
...
-        let ident ← mkFreshIdent binder
-        let type := binder
-        loop body (i+1) (newBinders.push <| mkExplicitBinder ident type)
-      | Syntax.node _ ``Lean.Parser.Term.paren _    =>
+      match binder.getKind with
+      | ``Lean.Parser.Term.implicitBinder
+      | ``Lean.Parser.Term.strictImplicitBinder
+      | ``Lean.Parser.Term.instBinder
+      | ``Lean.Parser.Term.explicitBinder
+      | ``Lean.Parser.Term.hole | `ident => loop body (i+1) (newBinders.push binder)
```
leodemoura, line 304: Nice. It is great to see new features making our code cleaner :)

src/Lean/Elab/Binders.lean
```diff
@@ -552,7 +558,7 @@ def expandMatchAltsWhereDecls (matchAltsWhereDecls : Syntax) : MacroM Syntax :=
 
 open Lean.Elab.Term.Quotation in
 @[builtinQuotPrecheck Lean.Parser.Term.fun] def precheckFun : Precheck
-  | `(fun $binders* => $body) => do
+  | `(fun $binders* $[: $ty?]? => $body) => do
```
leodemoura, line 561: Should we use `_ty` to avoid the warning from the unused variable linter?

src/Lean/Elab/MutualDef.lean
```diff
@@ -85,8 +85,8 @@ private def isMultiConstant? (views : Array DefView) : Option (List Name) :=
   if views.size == 1 &&
      views[0]!.kind == DefKind.opaque &&
      views[0]!.binders.getArgs.size > 0 &&
-     views[0]!.binders.getArgs.all (·.getKind == ``Parser.Term.simpleBinder) then
-    some <| (views[0]!.binders.getArgs.toList.map (fun stx => stx[0].getArgs.toList.map (·.getId))).join
+     views[0]!.binders.getArgs.all (·.isIdent) then
+    some (views[0]!.binders.getArgs.toList.map (·.getId))
```
leodemoura, line 89: Nice.

leodemoura:
> > The code changes suggest that it would be nice to have Term.binderIdent be a proper syntax kind, in which case it could be unified with Lean.binderIdent from Init.Tactics, which I might do in a follow-up PR.
> 
> Yes, that's a very good idea. Please also unify it with `ident <|> "_"` when you're at it. See also: [leanprover-community/mathport#152 (comment)](https://github.com/leanprover-community/mathport/pull/152#issuecomment-1173841553)

I think this is a good idea, but we can use another PR for this unification.
The current one is already a good improvement.

leodemoura (approved):

leodemoura:
@kha The failure has nothing to do with this PR. Please feel free to merge it as-is.

src/Lean/Elab/Binders.lean
Kha: Ah, not sure why I didn't see the warning. Will fix when I'm home.

PR #1303, opened 2022-07-13
Allow custom ignore functions for the unused variables linter

larsk21:
as discussed in #1266


larsk21:
This is probably because I moved the retrieval of the ignore functions out of the for loop going through the found variables.

tydeu:
A quick glance at this PR does lead me to concur with Sebastian, using a `SimplePersistentEnvExtension` to store the ignore functions seems like a preferable solution (assuming it isn't less efficient).

larsk21:
Strangely, `liftCoreM` seems to duplicate trace messages here...

PR #1310, opened 2022-07-15
feat: recover from errors in attributes

gebner:
In mathport, we still have lots of stubbed-out unimplemented attributes.  Right now, these errors are fatal and the definition is not declared, resulting in lots of follow-up errors.

This PR logs the error instead of throwing it, thus declaring the definition (see `nonfatalattrs.lean` test).  I've also changed the error range to the attribute (which is slightly more helpful than the whole definition).

### DIFF
--- a/src/Lean/Elab/Declaration.lean
+++ b/src/Lean/Elab/Declaration.lean
@@ -183,16 +183,19 @@ def elabDeclaration : CommandElab := fun stx => do
     let newStx ← `(namespace $ns:ident $newStx end $ns:ident)
     withMacroExpansion stx newStx <| elabCommand newStx
   | none => do
-    let modifiers ← elabModifiers stx[0]
     let decl     := stx[1]
     let declKind := decl.getKind
     if declKind == ``Lean.Parser.Command.«axiom» then
+      let modifiers ← elabModifiers stx[0]
       elabAxiom modifiers decl
     else if declKind == ``Lean.Parser.Command.«inductive» then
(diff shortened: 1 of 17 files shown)


tests/lean/625.lean
```diff
@@ -1,8 +1,8 @@
 import Lean
 open Lean Lean.PrettyPrinter
 
-def pfoo : PUnit → PUnit := id
-def px   : PUnit := ()
+def foo : PUnit → PUnit := id
+def x   : PUnit := ()
```
gebner, line 5: I believe this was the original intent behind this test (at least as far as I can tell from the comment).

src/Lean/Elab/Declaration.lean
```diff
@@ -183,16 +183,19 @@ def elabDeclaration : CommandElab := fun stx => do
...
+      let modifiers ← elabModifiers stx[0]
       elabAxiom modifiers decl
     else if declKind == ``Lean.Parser.Command.«inductive» then
+      let modifiers ← elabModifiers stx[0]
       elabInductive modifiers decl
     else if declKind == ``Lean.Parser.Command.classInductive then
+      let modifiers ← elabModifiers stx[0]
       elabClassInductive modifiers decl
     else if declKind == ``Lean.Parser.Command.«structure» then
+      let modifiers ← elabModifiers stx[0]
```
gebner, line 198: Without this duplication, `elabModifiers` would be called twice for definitions.

Issue #591, opened 2021-07-29
Better support for browsing traces


leodemoura:
This issue is a placeholder for collecting ideas and features for visualizing and filtering traces.
@dselsam @Kha @Vtec234

dselsam:
I wrote a proposal a few months back for organizing the trace messages which is germane to this discussion: https://github.com/leanprover/lean4/issues/400#issuecomment-819752356 I still think this proposal is basically right and that proper message organization and query support is more important than visualization. It would be great if we can dump the (non-delaborated) trace data to a file and have a convenient way of querying it in multiple ways with multiple different pp options. Example of trace usage:
```lean
set_option trace.dump "foo.trace.out" in
...
```
Open new `.lean` file and write something like:
```lean
set_option pp.myFavoriteSettings true

#trace.query "foo.trace.out"
  enable_pos (16, 25)
  set_class Meta.isDefEq                   enable
  set_class Meta.isLevelDefEq              disable
  set_prefix Meta.synthInstance.newSubgoal enable
```
I think even a textual dump of specific queries like this would get us 90% of the way there. Of course, it could be even better if the query returns an object that can itself be browsed interactively.

Vtec234:
Another observation in support of a "trace explorer" is that often the traces that come up are too large to reasonably print. That is, we can do it with optimized enough code, but it takes a few seconds because it is bottlenecked by delaboration, and then the resulting dump is hard to leaf through. Visualizing them interactively would solve the former issue as well since we could use the RPC mechanism to delay delaboration until it is actually needed and only invoke it on the subtraces that you expand.

dselsam:
Yes, this could be a big win. I hope the trace explorer will let us be less conservative in collecting trace messages in the first place. Thanks to filtering and dynamic expansion, there may not be much cost in tracing every module at once (so we don't need to re-run the program to toggle a trace class).

gebner:
As requested by Wojciech, I'm reposting here what I wrote about interactive traces on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Expandable.20trace.20API/near/257867137).

We should make all nested traces collapsible, and not just the nodes added using `traceCtx`.  Concretely, the Lean API should look something like this:
```
withNestedTraces (cls : Name) (msg : MessageData) (k : m α) (collapsed := false) : m α
```

[...]

PR #1312, opened 2022-07-17
Support nested inductives in `deriving RpcEncoding`

gebner:
While looking into my proposal from #591, I noticed that the derive handler for `RpcEncoding` only supported non-recursive inductive types, and produced unnecessarily complex instances for them:
```lean
inductive MsgEmbed where
  | expr : CodeWithInfos → MsgEmbed
  | goal : InteractiveGoal → MsgEmbed
  | lazyTrace : Nat → Name → WithRpcRef MessageData → MsgEmbed
  deriving RpcEncoding

-- generates the instance:
Lean.Widget.instRpcEncodingMsgEmbedRpcEncodingPacket : (_rpcEnc : Type) →
  [inst : Lean.Server.RpcEncoding Lean.Widget.CodeWithInfos _rpcEnc] →
    (_rpcEnc_1 : Type) →
      [inst : Lean.Server.RpcEncoding Lean.Widget.InteractiveGoal _rpcEnc_1] →
        (_rpcEnc_2 : Type) →
          [inst : Lean.Server.RpcEncoding Nat _rpcEnc_2] →
            (_rpcEnc_3 : Type) →
              [inst : Lean.Server.RpcEncoding Lean.Name _rpcEnc_3] →
                (_rpcEnc_4 : Type) →
                  [inst : Lean.Server.RpcEncoding (Lean.Server.WithRpcRef Lean.MessageData) _rpcEnc_4] →
                    Lean.Server.RpcEncoding Lean.Widget.MsgEmbed
                      (Lean.Widget.RpcEncodingPacket✝ _rpcEnc _rpcEnc_1 _rpcEnc_2 _rpcEnc_3 _rpcEnc_4)
```

With a bit of refactoring the derive handler now works for (nested) inductives, and produces much nicer instances (in stage2):
```lean
Lean.Widget.instRpcEncodingMsgEmbed : Lean.Server.RpcEncoding Lean.Widget.MsgEmbed
  Lean.Widget.RpcEncodingPacket✝
```

I've also extended to test to check the JSON serialization, and it is the same for some examples.  Running the editor on stage2 doesn't seem to produce any obvious errors either.

### DIFF
--- a/src/Lean/Server/Rpc/Deriving.lean
+++ b/src/Lean/Server/Rpc/Deriving.lean
@@ -104,17 +104,15 @@ private def deriveStructureInstance (indVal : InductiveVal) (params : Array Expr
     -- helpers for type name syntax
     let paramIds ← params.mapM fun p => return mkIdent (← getFVarLocalDecl p).userName
     let typeId := Syntax.mkApp (← `(@$(mkIdent indVal.name))) paramIds
-    let packetId := mkIdent <| indVal.name ++ `RpcEncodingPacket
-    let packetAppliedId ← `($packetId ..)
     let instId := mkIdent (`_root_ ++ indVal.name.appendBefore "instRpcEncoding")
 
     `(variable $packetParamBinders* in
-      protected structure $packetId:ident where
+      structure RpcEncodingPacket where
         $[($fieldIds : $fieldEncTypeStxs)]*
         deriving FromJson, ToJson
 
       variable $(paramBinders ++ packetParamBinders ++ encInstBinders)* in
-      @[instance] def $instId := show RpcEncoding $typeId $packetAppliedId from {
+      @[instance] def $instId := show RpcEncoding $typeId (RpcEncodingPacket ..) from {
         rpcEncode := fun a => return { $[$encInits],* }
         rpcDecode := fun a => return { $[$decInits],* }
       }
@@ -133,7 +131,7 @@ private def deriveInductiveInstance (indVal : InductiveVal) (params : Array Expr
     (paramBinders packetParamBinders encInstBinders : Array (TSyntax ``Parser.Term.bracketedBinder)) : TermElabM Command := do
   trace[Elab.Deriving.RpcEncoding] "for inductive {indVal.name} with params {params}"
   withoutModifyingEnv do
-  let packetNm := indVal.name ++ `RpcEncodingPacket
+  let packetNm := (← `(RpcEncodingPacket)).1.getId
   addDecl <| .axiomDecl {
     name := packetNm
     levelParams := []
@@ -150,7 +148,7 @@ private def deriveInductiveInstance (indVal : InductiveVal) (params : Array Expr
       let tyStx ← PrettyPrinter.delab packetTp
       let name := (← getFVarLocalDecl arg).userName
       `(bracketedBinder| ($(mkIdent name) : $tyStx))
-    let pktCtor ← `(Parser.Command.ctor| | $(mkIdent ctor.getString!):ident $[$fieldStxs]* : $(mkIdent packetNm))
+    let pktCtor ← `(Parser.Command.ctor| | $(mkIdent ctor.getString!):ident $[$fieldStxs]* : RpcEncodingPacket)
 
     -- create encoder and decoder match arms
     let nms ← argVars.mapM fun _ => mkIdent <$> mkFreshBinderName
@@ -170,23 +168,22 @@ private def deriveInductiveInstance (indVal : InductiveVal) (params : Array Expr
   -- helpers for type name syntax
   let paramIds ← params.mapM fun p => return mkIdent (← getFVarLocalDecl p).userName
   let typeId := Syntax.mkApp (← `(@$(mkIdent indVal.name))) paramIds
-  let packetId ← `($(mkIdent packetNm) ..)
   let instId := mkIdent (`_root_ ++ indVal.name.appendBefore "instRpcEncoding")
 
   `(variable $packetParamBinders:bracketedBinder* in
-    protected inductive $(mkIdent packetNm) where
+    inductive RpcEncodingPacket where
       $[$(st.ctors):ctor]*
       deriving FromJson, ToJson
 
     variable $(paramBinders ++ packetParamBinders ++ encInstBinders)* in
-    @[instance] partial def $instId := show RpcEncoding $typeId $packetId from
+    @[instance] partial def $instId := show RpcEncoding $typeId (RpcEncodingPacket ..) from
       { rpcEncode, rpcDecode }
     where
-      rpcEncode {m} [Monad m] [MonadRpcSession m] (x : $typeId) : ExceptT String m $packetId :=
-        have inst : RpcEncoding $typeId $packetId := { rpcEncode, rpcDecode }
+      rpcEncode {m} [Monad m] [MonadRpcSession m] (x : $typeId) : ExceptT String m (RpcEncodingPacket ..) :=
+        have inst : RpcEncoding $typeId (RpcEncodingPacket ..) := { rpcEncode, rpcDecode }
         match x with $[$(st.encodes):matchAlt]*
-      rpcDecode {m} [Monad m] [MonadRpcSession m] (x : $packetId) : ExceptT String m $typeId :=
-        have inst : RpcEncoding $typeId $packetId := { rpcEncode, rpcDecode }
+      rpcDecode {m} [Monad m] [MonadRpcSession m] (x : RpcEncodingPacket ..) : ExceptT String m $typeId :=
+        have inst : RpcEncoding $typeId (RpcEncodingPacket ..) := { rpcEncode, rpcDecode }
         match x with $[$(st.decodes):matchAlt]*
   )
 
--- a/tests/lean/derivingRpcEncoding.lean.expected.out
+++ b/tests/lean/derivingRpcEncoding.lean.expected.out
@@ -1,57 +1,29 @@
 instRpcEncodingWithRpcRefFooRefRpcRef : RpcEncoding (WithRpcRef FooRef) Lsp.RpcRef
 ok: {"p": "0"}
-protected inductive Bar.RpcEncodingPacket : Type
-number of parameters: 0
-constructors:
-Bar.RpcEncodingPacket.mk : Lsp.RpcRef → FooJson → Bar.RpcEncodingPacket
-instRpcEncodingBar : RpcEncoding Bar Bar.RpcEncodingPacket
+instRpcEncodingBar : RpcEncoding Bar RpcEncodingPacket✝
 ok: {"fooRef": {"p": "0"}, "fooJson": {"s": ""}}
-instRpcEncodingBarTrans : RpcEncoding BarTrans BarTrans.RpcEncodingPacket
+instRpcEncodingBarTrans : RpcEncoding BarTrans RpcEncodingPacket✝
 ok: {"bar": {"fooRef": {"p": "0"}, "fooJson": {"s": ""}}}
-instRpcEncodingBaz : RpcEncoding Baz Baz.RpcEncodingPacket
+instRpcEncodingBaz : RpcEncoding Baz RpcEncodingPacket✝
 ok: {"arr": []}
-protected inductive FooGeneric.RpcEncodingPacket : Type → Type
-number of parameters: 1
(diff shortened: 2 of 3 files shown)


src/Lean/Server/Rpc/Basic.lean
```diff
@@ -42,10 +42,13 @@ non-JSON-serializable fields can be auto-encoded in two ways:
 -- TODO(WN): for Lean.js, have third parameter defining the client-side structure;
 -- or, compile `WithRpcRef` to "opaque reference" on the client
 class RpcEncoding (α : Type) (β : outParam Type) where
-  rpcEncode {m : Type → Type} [Monad m] [MonadRpcSession m] : α → m β
+  rpcEncode {m : Type → Type} [Monad m] [MonadRpcSession m] : α → ExceptT String m β
```
gebner, line 45: This is a bit ugly, but unfortunately necessary to define the unconditional `Inhabited (RpcEncoding α β)` instance below.  The instance is in turn necessary to define `RpcEncoding` instances using `partial def` (which is the easiest approach to support nested inductives).

tests/lean/derivingRpcEncoding.lean.expected.out
```diff
@@ -0,0 +1,41 @@
...
+  {"fooRef": {"p": "1"}, "fooJson": {"s": ""}}]}
+inductive FooInductive.RpcEncodingPacket : Type → Type
+number of parameters: 1
+constructors:
+FooInductive.RpcEncodingPacket.a : {αPacket : Type} → αPacket → Lsp.RpcRef → FooInductive.RpcEncodingPacket αPacket
+FooInductive.RpcEncodingPacket.b : {αPacket : Type} → Nat → αPacket → Nat → FooInductive.RpcEncodingPacket αPacket
+instRpcEncodingFooInductiveRpcEncodingPacket : (α αPacket : Type) →
+  [inst : RpcEncoding α αPacket] → RpcEncoding (FooInductive α) (FooInductive.RpcEncodingPacket αPacket)
+ok: {"a": [{"baz": []}, {"p": "0"}]}
+ok: {"b": [42, {"baz": []}, 0]}
```
gebner, line 33: @Vtec234 Reading the comment in the derive handler correctly, the constructor `| b : (n : Nat) → (a : α) → (m : Nat) → FooInductive α` should produce a map instead of an array:
```suggestion
ok: {"b": {"n": 42, "a": {"baz": []}, "m": 0}}
```

However this has already been broken on master for quite some time I believe.  In any case this test will catch future changes in the serialization format.

src/Lean/Server/Rpc/Deriving.lean
```diff
@@ -134,115 +110,93 @@ private def deriveStructureInstance (indVal : InductiveVal) (params : Array Expr
...
-        deriving $(mkIdent ``FromJson), $(mkIdent ``ToJson)
-
-      instance : $(mkIdent ``RpcEncoding) $typeId $packetAppliedId where
-        rpcEncode a := return {
-          $[$encInits],*
-        }
-        rpcDecode a := return {
-          $[$decInits],*
-        }
+    let packetId := mkIdent <| indVal.name ++ `RpcEncodingPacket
```
Vtec234, line 113: The name was intentionally hygienic -- it's only supposed to be obtainable using typeclass synthesis, i.e. through the `RpcEncoding` mapping. Do you need it to be accessible?

tests/lean/derivingRpcEncoding.lean.expected.out
Vtec234: Indeed that is what the `deriving` was supposed to do, and indeed I just confirmed that it does not work. It is just `ToJson` which encodes `|  c (n m : Nat) : FooInductive α` as `{"c": {"n": 1, "m": 2}}`, but `b` as the test output. Could you add a test checking that the `c` variant works?

src/Lean/Server/Rpc/Basic.lean
Vtec234: (See comment on the unconditional instance).

src/Lean/Server/Rpc/Deriving.lean
```diff
@@ -134,115 +110,93 @@ private def deriveStructureInstance (indVal : InductiveVal) (params : Array Expr
...
+  -- helpers for type name syntax
+  let paramIds ← params.mapM fun p => return mkIdent (← getFVarLocalDecl p).userName
+  let typeId := Syntax.mkApp (← `(@$(mkIdent indVal.name))) paramIds
+  let packetId := Syntax.mkApp (mkIdent packetNm) packetParamNames
+
+  `(inductive $(mkIdent packetNm) $packetParamBinders:bracketedBinder* where
+      $[$(st.ctors):ctor]*
+      deriving FromJson, ToJson
+
+    variable $(paramBinders ++ packetParamBinders ++ encInstBinders)* in
```
Vtec234, line 189: The point of using `variable` was to ensure that we can have `RpcEncoding` without requiring a packet for ghost parameters. For example, we should have `RpcEncoding (TSyntax n) _` without `RpcEncoding n _` because the `variable` will not be introduced if it doesn't appear in the body. Could you please check that this works and add a ghost case to the test? I clearly overdid it with the `variable`s though, we don't need them for *all* the types appearing inside.

Technically, if I understand the code correctly without having read it in detail (`derive` handlers are gnarly :/) the semantics after this PR will be different - we need `RpcEncoding` for argument types *at definition time*, whereas before we did not. In practice this leads to delayed synthesis failures, so the eager approach is better.

src/Lean/Server/Rpc/Deriving.lean
gebner: I've found it very useful to be able to e.g. `#print TaggedText.RpcEncodingPacket` which is something that seems completely impossible with inaccessible names.  But I can revert this change if you prefer.

src/Lean/Server/Rpc/Deriving.lean
gebner: > For example, when `inductive Unused (α : Type)`, we should have `RpcEncoding (Unused α) _` without `RpcEncoding α _` because the `variable [RpcEncoding α _]` will not be introduced if `@rpcEncode α` doesn't appear in the body. Could you please check that this works and add a ghost case to the test?

(I've changed the name because we already have a different type called `PEmpty`.)

This doesn't work, but for a different reason.  Namely the type of the instance is `(α αPacket : Type) →
  RpcEncoding (Unused α) (Unused.RpcEncodingPacket αPacket)`.  The instance doesn't have a `RpcEncoding α _` argument, but it will never apply because TC synthesis can't infer the `αPacket` argument.

Another issue is that empty inductives don't work.

Do you have a concrete example where you need unused parameters?

> the semantics after this PR will be different - we need `RpcEncoding` for argument types at definition time, whereas before we did not.

Right.

> In practice this leads to delayed synthesis failures, so the eager approach is better.

Yes, I was very surprised when I tried do derive `RpcEncoding` for an inductive type and then type class synthesis couldn't find it.

gebner: Both issues are fixed now.

tests/lean/derivingRpcEncoding.lean.expected.out
gebner: Okay, fixed.  The test was already there (but obviously produces a different output now).

src/Lean/Server/Rpc/Deriving.lean
Vtec234: > Do you have a concrete example where you need unused parameters?

I was thinking of `TSyntax`, but there the index is not a type and we already skip non-Type indices anyway. Actually this technically makes the instance unsound, because `TSyntax n` and `TSyntax m` have the same encoding and we can do ``rpcDecode (α := TSyntax `abc) (rpcEncode (α := TSyntax `def) stx)``. Not sure whether this is actually an issue.

src/Lean/Server/Rpc/Deriving.lean
Kha: That would still be a pretty awkward workflow, no? I usually use `trace.Elab.command` for debugging such macros fwiw.

src/Lean/Server/Rpc/Basic.lean
```diff
@@ -42,10 +42,13 @@ non-JSON-serializable fields can be auto-encoded in two ways:
...
   rpcDecode {m : Type → Type} [Monad m] [MonadRpcSession m] : β → ExceptT String m α
 export RpcEncoding (rpcEncode rpcDecode)
 
+instance : Nonempty (RpcEncoding α β) :=
```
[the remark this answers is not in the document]
Vtec234: > There is no way for TC synthesis to derive `RpcEncoding α β` from `Nonempty (RpcEncoding α β)`.

No, I know that there is no way to misuse this in practice. It just seems conceptually weird, as most pairs in the function graph defined by `RpcEncoding` *are* supposed to be empty. Maybe I am fundamentally confused about something here?

> Note that you can prove `(∀ {α β} {m : Type → Type} [Monad m] [MonadRpcSession m], α → m β)`.

I see, thanks. With the `ExceptT` change, though, we can split out `partial rpcEncode/rpcDecode` without using a `partial` instance and thus dropping the `Nonempty (RpcEncoding α β)`, right?

src/Lean/Server/Rpc/Deriving.lean
gebner: Note that the `TSyntax` instance is only unsound in the sense that it can violate the (not particularly enforced) invariant that a `TSyntax ks` has one of the "kinds" `ks`.  But I don't think there's any realistic way to enforce that invariant, even if we serialized `ks` as well.

It is not memory unsafe though.

src/Lean/Server/Rpc/Deriving.lean
gebner: Yes, I'm using `trace.Elab.command` as well.  But it doesn't really help you to figure out what's going on with `variable (α β γ δ ε) [RE α α'] [RE β β'] ... in inductive Blah`.  (That is, what parameters the inductive actually gets.)  With definitions it's even worse because you only see the input syntax, but not the elaborated definition (which is simpler in inductives).

src/Lean/Server/Rpc/Basic.lean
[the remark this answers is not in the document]
gebner: > It just seems conceptually weird, as most pairs in the function graph defined by `RpcEncoding` are supposed to be empty. Maybe I am fundamentally confused about something here?

Right, and the function graph defined by the derivable `RpcEncoding`s is *still* a partial function and mostly empty.

I find it really hard to understand your point here.  Are you also concerned about `#check { rpcDecode := default, rpcEncode := default : RpcEncoding Nat Int }`, i.e. any term of type `RpcEncoding ..` anywhere?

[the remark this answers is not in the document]
Vtec234: Yes, or `noncomputable instance : RpcEncoding Nat Int := Classical.ofNonempty`. But I agree this is such a clearly mistaken instance that it's probably not a problem in practice, so if you prefer doing things this way let's leave it.

gebner:
Re: the `ExceptT String` issue.  Another approach to deal with it would be to drop the out param in `RpcEncoding`, because then `rpcEncode` would be trivially inhabited:
```lean
class RpcEncoding (α : Type) where
  rpcDecode {...} (j : Json) : ExceptT String m α
  rpcEncode {...} (a : α) : m Json
```

gebner:
> This sounds quite appealing and might even simplify some code. I imagine the whole deriving handler would have to be rewritten though

The changes are quite minimal though, the encoding packet would have constructors like `ctor : Json → Json → Json → RpcEncodingPacket` instead of `ctor : Nat → String → RpcEncodingPacket → RpcEncodingPacket`.  The change would even simplify the derive handler since the encoding packet would be parameterless.

> I'd say let's go with what's here for now if it makes your usecases work, and switch later if we have to.

Yes, I'd prefer to get this PR merged soon to unblock #591.

Vtec234:
> the encoding packet would have constructors like `ctor : Json → Json → Json → RpcEncodingPacket`

Oh, that's a neat idea (I had a worse one in mind)! If you would like to implement it in this PR then let's go with that.

gebner:
> If you would like to implement it in this PR then let's go with that.

No, I'd rather do it in a future PR.  I don't want to delay this PR.

PR #1314, opened 2022-07-17
feat: register/chain LSP handlers in non-builtin `initialize` blocks

tydeu:


tydeu:
cc @gebner, @Vtec234

Vtec234 (approved):

tydeu:
cc @leodemoura / @Kha

Kha:
I believe we have quite a few `IORef`s that are not guarded by any such condition. AFAIK it is critical only for environment extensions.

leodemoura:
> I believe we have quite a few `IORef`s that are not guarded by any such condition. AFAIK it is critical only for environment extensions.

If these `IO.Ref`s are being updated by multiple threads, we should investigate. 
The `IO.initializing` and `Lean.initializing` are great for ensuring we have only one thread running. It is a great solution for `IO.Ref`s that are updated only during initialization, and then are read-only for many threads.

PR #1316, opened 2022-07-18
doc: explain AttributeKind

EdAyers:
docstring for AttributeKind

[skip ci]

### DIFF
--- a/src/Lean/Attributes.lean
+++ b/src/Lean/Attributes.lean
@@ -28,7 +28,7 @@ structure AttributeImplCore where
 For example: `attribute [local myattr, scoped yourattr, theirattr]`.
 
 This is used to indicate how an attribute should be scoped.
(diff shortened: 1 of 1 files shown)


src/Lean/Attributes.lean
```diff
@@ -24,6 +24,17 @@ structure AttributeImplCore where
   applicationTime := AttributeApplicationTime.afterTypeChecking
   deriving Inhabited
 
+/-- You can tag attributes with the 'local' or 'scoped' kind.
+For example: `attribute [local myattr, scoped yourattr, theirattr]`.
+
+This is used to indicate how an attribute should be scoped.
+- local means that the attribute should only be applied in the current scope and forgotten once the namespace is closed.
```
Kha, line 31:
```suggestion
- local means that the attribute should only be applied in the current scope and forgotten once the current section, namespace, or file is closed.
```

PR #1317, opened 2022-07-18
fix: make document symbols request deterministic

Kha:
And maybe fix other bugs resulting from `parseAhead`. As we now have other requests like semantic highlighting that also have to wait for the entire file, this special case probably doesn't make any sense anymore.


Vtec234 (approved):
In the future we could start sending LSP incremental progress for this kind of request which updates every time the `AsyncList` grows by a new snapshot.

Kha:
VS Code still doesn't support it though, no?

PR #1318, opened 2022-07-18
fix: translate `macro/elab` parameters `x:c` to `$x:c` instead of `$x`

Kha:
This ensures `x` is guaranteed to be of kind `c` even in presence of partial syntax.

### DIFF
--- a/src/Lean/Data/Name.lean
+++ b/src/Lean/Data/Name.lean
@@ -21,6 +21,10 @@ def getPrefix : Name → Name
   | str p _   => p
   | num p _   => p
 
+def getString? : Name → Option String
+  | str _ s => some s
+  | _       => none
+
 def getString! : Name → String
   | str _ s => s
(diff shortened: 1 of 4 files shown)


Kha:
Oof, this breaks macros using e.g. `declVal` because that is not actually a kind. So maybe we should think about the general story first after all. Or we need to extend `register_parser_alias` with the information of whether an alias corresponds to a kind or not.

leodemoura:
@Kha Should we close this PR, or will you try to fix it later?

Kha:
I think it would be better to

> make `$x` check its kind as well

and check whether it would in fact "severely limit elaboration in presence of partial syntax". Certainly any partial syntax in e.g. a nested category parser like `term` would be safe, as there is no kind to check against. Though for that, too, we would "need to extend `register_parser_alias` with the information of whether an alias corresponds to a kind or not".

PR #1320, opened 2022-07-18
fix: disable auto implicits in structure field default values

Kha:
See https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/structure.20with.20unassigned.20variable.20as.20default.20variable. In general, `:=` disabling auto implicits seems like a nice mnemonic. Unfortunately it's harder to implement for binder opt/autoParams as they are macros. We might need a `withoutAutoBoundImplicit%` helper elaborator for them.


leodemoura:
> We might need a withoutAutoBoundImplicit% helper elaborator for them.

Yeah, this is a good idea.

PR #1323, opened 2022-07-19
feat: size of a LocalContext

kim-em:
Requested in https://github.com/leanprover-community/mathlib4/pull/313


semorrison:
If it would be better to accumulate things like this in mathlib4 for a while (or forever), rather than the overhead of PRs to core, we can do that instead.

PR #1326, opened 2022-07-19
doc: various docstrings

EdAyers:


src/Init/Prelude.lean
```diff
@@ -961,6 +970,12 @@ theorem usize_size_eq : Or (Eq USize.size 4294967296) (Eq USize.size 18446744073
   | _, Or.inl rfl => Or.inl (by decide)
   | _, Or.inr rfl => Or.inr (by decide)
 
+/-- A USize is an unsigned integer with the size of a word
+for the platform's arctitecture.
```
gebner, line 974:
```suggestion
for the platform's architecture.
```

src/Lean/Meta/MatchUtil.lean
```diff
@@ -19,6 +19,7 @@ namespace Lean.Meta
   | none => p? (← whnf e)
   | s    => return s
 
+/-- matches `e` with `lhs = rhs : α` and returns `(α, lhs, rhs)`. -/
```
gebner, line 22:
```suggestion
/-- Matches `e` with `lhs = (rhs : α)` and returns `(α, lhs, rhs)`. -/
```

src/Lean/MetavarContext.lean
```diff
@@ -420,6 +420,12 @@ def hasAssignableMVar [Monad m] [MonadMCtx m] : Expr → m Bool
 def assignLevelMVar [MonadMCtx m] (mvarId : MVarId) (val : Level) : m Unit :=
   modifyMCtx fun m => { m with lAssignment := m.lAssignment.insert mvarId val, usedAssignment := true }
 
+/--
+  Add `mvarId := u` to the metavariable assignment.
+  This method does not check whether `mvarId` is already assigned, nor it checks whether
+  a cycle is being introduced.
+  This is a low-level API, and it is safer to use `isDefEq (mkLevelMVar mvarId) u`.
```
gebner, line 427:
```suggestion
Add `mvarId := u` to the metavariable assignment.
This method does not check whether `mvarId` is already assigned, nor it checks whether
a cycle is being introduced, or whether the expression has the right type.
This is a low-level API, and it is safer to use `isDefEq (mkLevelMVar mvarId) u`.
```

src/Init/Prelude.lean
```diff
@@ -858,6 +864,7 @@ instance : Inhabited UInt8 where
   default := UInt8.ofNatCore 0 (by decide)
 
 def UInt16.size : Nat := 65536
+/-- Unsigned, 16-bit integer. -/
```
gebner, line 867:
```suggestion
/-- Unsigned 16-bit integer. -/
```
And similarly below.

src/Lean/Meta/AppBuilder.lean
```diff
@@ -102,6 +102,7 @@ def mkEqTrans (h₁ h₂ : Expr) : MetaM Expr := do
     | none, _ => throwAppBuilderException ``Eq.trans ("equality proof expected" ++ hasTypeMsg h₁ hType₁)
     | _, none => throwAppBuilderException ``Eq.trans ("equality proof expected" ++ hasTypeMsg h₂ hType₂)
 
+/-- Given `HEq a b`, makes `HEq b a`.  -/
```
Vtec234, line 105:
```suggestion
/-- Given `h : HEq a b`, return a proof of `HEq b a`.  -/
```
And similarly for the other ones, to distinguish the proving functions from ones which would manipulate types.

EdAyers: Should it be `$h : HEq a b`?

Kha: I don't think we use `$` in this sense in docstrings elsewhere

Issue #1324, opened 2022-07-19
doc strings on `initialize`


digama0:
Reported on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/326056-ICERM22-after-party/topic/List.20of.20tactic.20status.20descriptions/near/290042488). Doc strings are currently not accepted in `initialize` statements.
```lean
/-- This is a doc string -/
initialize foo : Unit ← pure ()
```

PR #1327, opened 2022-07-19
feat: support all sensible modifiers on (builtin_)initialize

Kha:
Resolves https://github.com/leanprover/lean4/issues/1324


Kha:
I ended up refactoring the macro a little

src/Lean/Elab/Declaration.lean
```diff
@@ -327,35 +326,21 @@ def elabMutual : CommandElab := fun stx => do
...
-    else if optVisibility[0].getKind == ``Parser.Command.private then
-      `(def initFn : IO $type := do $doSeq
-        @[$attrId:ident initFn] private opaque $id : $type)
-    else if optVisibility[0].getKind == ``Parser.Command.protected then
+@[builtinMacro Lean.Parser.Command.«initialize»] def expandInitialize : Macro
+  | stx@`($declModifiers:declModifiers $kw:initializeKeyword $[$id? : $type? ←]? $doSeq) => do
+    let attrId := mkIdentFrom stx <| if kw.raw.isToken "initialize" then `init else `builtinInit
+    if let (some id, some type) := (id?, type?) then
+      let `(Parser.Command.declModifiersT| $[$doc?:docComment]? $[@[$attrs?,*]]? $(vis?)? $[unsafe%$unsafe?]?) := stx[0]
+        | Macro.throwErrorAt declModifiers "invalid initialization command, unexpected modifiers"
```
Kha, line 334: This error message is pretty broad, but also trying to apply `noncomputable` or `partial` to `intialize` should be pretty rare

src/Lean/Elab/Declaration.lean
```diff
@@ -327,35 +326,21 @@ def elabMutual : CommandElab := fun stx => do
...
+@[builtinMacro Lean.Parser.Command.«initialize»] def expandInitialize : Macro
+  | stx@`($declModifiers:declModifiers $kw:initializeKeyword $[$id? : $type? ←]? $doSeq) => do
+    let attrId := mkIdentFrom stx <| if kw.raw[0].isToken "initialize" then `init else `builtinInit
+    if let (some id, some type) := (id?, type?) then
+      let `(Parser.Command.declModifiersT| $[$doc?:docComment]? $[@[$attrs?,*]]? $(vis?)? $[unsafe%$unsafe?]?) := stx[0]
+        | Macro.throwErrorAt declModifiers "invalid initialization command, unexpected modifiers"
+      let declModifiers ← `(Parser.Command.declModifiersT| $[$doc?:docComment]? @[$attrId:ident initFn, $(attrs?.getD ∅),*] $(vis?)? $[unsafe%$unsafe?]?)
       `(def initFn : IO $type := do $doSeq
-        @[$attrId:ident initFn] protected opaque $id : $type)
+        $(⟨declModifiers⟩):declModifiers opaque $id : $type)
```
tydeu, line 337: How is an `unsafe` from `declModifiers` meant to work here? Should it not also apply to the `initFn`?

Kha: Nice catch, fixed

Kha: Well, I applied it only to the initializer function now. In theory, these are orthogonal distinctions of "inner" and "outer" unsafety. In practice, `unsafe initialize` is probably always meant to be inner unsafety only. If we had mathlib4's `unsafe` term macro for inner unsafety, we could restore this branch's behavior.

PR #1328, opened 2022-07-19
chore: add sanity check for `SemanticTokenType/Modifier.names`

tydeu:
A follow-up to #1325. I realized today that I could write a theorem to verify the invariant that `names[v.toNat] = <string of ctor>`, so I did. In doing so, I realized I had typo'd the signature for `SemanticTokenModifier.toNat`. Sorry for the double PR!


tydeu:
cc @Kha

Kha:
I don't know what to think about this PR, it feels a bit wrong to derive `ToJson` just for this. But I suppose it's better than not doing anything.

PR #1330, opened 2022-07-20
doc: src/Lean/Declaration.lean

bollu:
Add details explained by Leo at the hackathon.

### DIFF
--- a/src/Lean/MetavarContext.lean
+++ b/src/Lean/MetavarContext.lean
@@ -256,13 +256,14 @@ structure MetavarDecl where
   userName       : Name := Name.anonymous
   lctx           : LocalContext
   type           : Expr
-  depth          : Nat /--
+  /--
     The nesting depth of this metavariable. We do not want
     unification subproblems to influence the results of parent
     problems. The depth keeps track of this information and ensures
     that unification subproblems cannot leak information out, by unifying
     based on depth.
   -/
+  depth          : Nat
   localInstances : LocalInstances
   kind           : MetavarKind
   numScopeArgs   : Nat := 0 -- See comment at `CheckAssignment` `Meta/ExprDefEq.lean`


src/Lean/Declaration.lean
```diff
@@ -213,6 +220,19 @@ structure InductiveVal extends ConstantVal where
   isRec : Bool
   /-- Whether the definition is flagged as unsafe. -/
   isUnsafe : Bool
+  /-- An of inductive type called reflexive if it has at least one constructor that takes as an argument a function returning the
```
semorrison, line 223:
```suggestion
  /-- An inductive type is called reflexive if it has at least one constructor that takes as an argument a function returning the
```

src/Lean/MetavarContext.lean
```diff
@@ -252,7 +256,13 @@ structure MetavarDecl where
   userName       : Name := Name.anonymous
   lctx           : LocalContext
   type           : Expr
-  depth          : Nat
+  depth          : Nat /--
+    The nesting depth of this metavariable. We do not want a
```
semorrison, line 260:
```suggestion
    The nesting depth of this metavariable. We do not want
```

src/Lean/MetavarContext.lean
```diff
@@ -252,7 +256,13 @@ structure MetavarDecl where
   userName       : Name := Name.anonymous
   lctx           : LocalContext
   type           : Expr
-  depth          : Nat
+  depth          : Nat /--
```
leodemoura, line 259: This docstring must be before `depth : Nat`

leodemoura:
@bollu Thanks for submitting the doc-string PRs. They require minor changes, are you going to have time to do it?

Thanks,
Leo

bollu:
Yes, I will. I'm currently travelling, and I'll be ready to fixup my
patches on Wednesday.


Best,
~Siddharth

src/Lean/MetavarContext.lean
bollu: Done!

PR #1331, opened 2022-07-20
doc: Add explanations to MetavarContext

bollu:
This file has a lot of documentation at the top. I would like to inline the documentation in a second pass of this file. For now, I add some notes that Leo mentioned.


src/Lean/MetavarContext.lean
```diff
@@ -218,6 +218,10 @@ very simple unification and/or non-nested TC. So, if the "app builder" becomes a
 we may solve the issue by implementing `isDefEqCheap` that never invokes TC and uses tmp metavars.
 -/
 
+/--
+The LocalInstance stores information that the kernel does not care about, but is still
+important to carry for user-facing reasons about local instances in the LocalContext.
```
Kha, line 223: This is not a good docstring. It does not explain what the structure is used for, only what it's not used for.

src/Lean/MetavarContext.lean
```diff
@@ -264,6 +268,8 @@ structure MetavarDecl where
   `mvarIdPending` is a `syntheticOpaque` metavariable that has not been synthesized yet. The delayed assignment becomes a real one
   as soon as `mvarIdPending` has been fully synthesized.
   `fvars` are variables in the `mvarIdPending` local context.
+
+  See `instantiateLevelMVars` for the rationale of delayed assignments.
```
Kha, line 272:
```suggestion
  See the comment below `assignDelayedMVar ` for the rationale of delayed assignments.
```
I think it would make sense to inline that comment here.

bollu: @Kha I am not so confident, because I would then want to add a comment at `assignDelayedMVar` saying:

> see the comment above `DelayedMetavarAssignment` for more details.

Perhaps we should adopt the [`[Note]` system of GHC](https://github.com/ghc/ghc/search?p=3&q=Note+%5B+%5D&type=) as a lightweight mechanism to reference these kinds of details?

src/Lean/MetavarContext.lean
bollu: I added an explanation, please do check if it's correct (I'm unsure my recollection of the explanation of how `LocalInstance` is used is accurate).

src/Lean/MetavarContext.lean
digama0: FYI, mathlib already uses that system, see https://leanprover-community.github.io/mathlib_docs/commands.html#library_note . If desired we could move the `library_note` command to core.

src/Lean/MetavarContext.lean
```diff
@@ -218,6 +218,12 @@ very simple unification and/or non-nested TC. So, if the "app builder" becomes a
 we may solve the issue by implementing `isDefEqCheap` that never invokes TC and uses tmp metavars.
 -/
 
+/--
+The `LocalInstance` stores information necessary for user facing interactions:
+it stores the user facing name of the typeclass in `className`, and the
```
Kha, line 223:
```suggestion
`LocalInstance` represents a local typeclass instance registered by and for
the elaborator. It stores the name of the typeclass in `className`, and the
```
Quite close!

src/Lean/MetavarContext.lean
```diff
@@ -218,6 +218,12 @@ very simple unification and/or non-nested TC. So, if the "app builder" becomes a
...
+concrete typeclass instance in `fvar`. Note that the kernel does not care about
+this information, since typeclasses are entirely eliminated during elaboration.
```
Kha, line 225: I'm actually not sure why this comment is here specifically. The kernel doesn't care about *anything* in this module.

Kha: I suppose it's fine to contrast it with `LocalContext`

src/Lean/MetavarContext.lean
Kha: Yeah, `library_note` seems nice. I think a short RFC in that regard would be great.

PR #1335, opened 2022-07-20
feat: make elabSimpArgs public

EdAyers:
This is used in mathlib4, and currently is clunkily reexported. I also found it useful working on norm_num.

### DIFF
--- a/src/Lean/Elab/Tactic/Simp.lean
+++ b/src/Lean/Elab/Tactic/Simp.lean
@@ -126,11 +126,11 @@ inductive ResolveSimpIdResult where
   | ext  (ext : SimpExtension)
 
 /--
-  Elaborate extra simp theorems provided to `simp`. `stx` is of the `simpTheorem,*`
+  Elaborate extra simp theorems provided to `simp`. `stx` is of the form `"[" simpTheorem,* "]"`
   If `eraseLocal == true`, then we consider local declarations when resolving names for erased theorems (`- id`),
   this option only makes sense for `simp_all`.
 -/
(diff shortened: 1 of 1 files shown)


src/Lean/Elab/Tactic/Simp.lean
```diff
@@ -126,11 +126,11 @@ inductive ResolveSimpIdResult where
   | ext  (ext : SimpExtension)
 
 /--
-  Elaborate extra simp theorems provided to `simp`. `stx` is of the `simpTheorem,*`
+  Elaborate extra simp theorems provided to `simp`. `stx` is of the form `"[" simpTheorem,* "]"`
```
Kha, line 129: Suggestion: let's introduce
```
syntax simpEntry := simpStar <|> simpErase <|> simpLemma
syntax simpEntries := "[" simpEntry,* "]"
```
for better reuse and proper `TSyntax` signature. Happy to implement that if others agree.

EdAyers: A few caveats: sometimes the grammar is just `"[" (simpErase <|> simpLemma),* "]"`

Kha: Good to know, so the function should accept ``TSyntaxArray `simpEntry``

EdAyers: I had a go at implementing but it got errors

leodemoura:
I am merging the PR as-is. We can have other PRs for the syntax modifications.

PR #1340, opened 2022-07-21
fix: make foApprox mdata-invariant

Kha:


Kha:
Test case by @larsk21. This came up in the context of https://github.com/larsk21/iris-lean.

src/Lean/Meta/ExprDefEq.lean
```diff
@@ -888,14 +888,15 @@ def checkAssignment (mvarId : MVarId) (fvars : Array Expr) (v : Expr) : MetaM (O
       let v ← instantiateMVars v
       CheckAssignment.checkAssignmentAux mvarId fvars hasCtxLocals v
 
-private def processAssignmentFOApproxAux (mvar : Expr) (args : Array Expr) (v : Expr) : MetaM Bool :=
+private partial def processAssignmentFOApproxAux (mvar : Expr) (args : Array Expr) (v : Expr) : MetaM Bool :=
   match v with
+  | .mdata _ e   => processAssignmentFOApproxAux mvar args v
```
leodemoura, line 893: There is typo here. `v` => `e`

leodemoura: We can also fix it using 
```lean
match v.cleanupAnnotations with
```

Kha: Ahh, that's what I get for not rechecking the test case after refactoring the code... as well as ignoring the linter warning

src/Lean/Meta/ExprDefEq.lean
```diff
@@ -888,14 +888,15 @@ def checkAssignment (mvarId : MVarId) (fvars : Array Expr) (v : Expr) : MetaM (O
       let v ← instantiateMVars v
       CheckAssignment.checkAssignmentAux mvarId fvars hasCtxLocals v
 
-private def processAssignmentFOApproxAux (mvar : Expr) (args : Array Expr) (v : Expr) : MetaM Bool :=
+private partial def processAssignmentFOApproxAux (mvar : Expr) (args : Array Expr) (v : Expr) : MetaM Bool :=
```
leodemoura, line 891: You should also be able to remove `partial` now that the typo was fixed.

Kha: Right, thanks... I started out with `consumeMData` but noticed neither of them was used in other parts of the file, perhaps for performance purposes I assumed

leodemoura: `consumeMData` was added after we wrote this file.

PR #1355, opened 2022-07-22
fix: do not use `native_decide`

gebner:
We should not use `native_decide`, `sorry`, or `axiom` (except for `propext`, `Quot.sound`, and `Classical.choice`) in (safe) declarations the core library.  These are less trustworthy and hard to avoid for other libraries.

### DIFF
--- a/src/Lean/Data/Lsp/LanguageFeatures.lean
+++ b/src/Lean/Data/Lsp/LanguageFeatures.lean
@@ -265,9 +265,9 @@ def SemanticTokenType.toNat (type : SemanticTokenType) : Nat :=
   type.toCtorIdx
 
 -- sanity check
-theorem SemanticTokenType.names_toNat_eq_toJson {v : SemanticTokenType} :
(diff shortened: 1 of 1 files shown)


tydeu:
@gebner The `dsimp` was necessary prior to an update-stage0 change because it was preferring the native definitions of `names` and `toNat`. Secondly, what is wrong with using `native_decide`? It does use an axiom, but you can check that with `#print axioms` if some end user cares and it only matters if someone else uses the theorem.

gebner:
Thanks for the clarification, didn't see the stage0 update.

> what is wrong with using native_decide?

It increases the trusted code base enormously, can't be verified by external checkers, it's extremely easy to derive false with `implementedBy`, etc.  We might at some point have better solutions for this, but for now it's best to avoid it.

Even if you don't use the lemma explicitly, automation can still dig it up, it breaks external checkers, etc.  For a nonessential use like here, it's just not worth it to put stumbling blocks in the core library.

PR #1362, opened 2022-07-22
Further refine `evalTactic`

Kha:
See individual commits

### DIFF
--- a/src/Lean/Elab/Tactic/Basic.lean
+++ b/src/Lean/Elab/Tactic/Basic.lean
@@ -172,6 +172,7 @@ where
     @[inline] handleEx (s : SavedState) (failures : Array EvalTacticFailure) (ex : Exception) (k : Array EvalTacticFailure → TacticM Unit) := do
       match ex with
       | .error .. =>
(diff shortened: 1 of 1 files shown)


src/Lean/Elab/Tactic/Basic.lean
```diff
@@ -167,29 +162,27 @@ partial def evalTactic (stx : Syntax) : TacticM Unit :=
     | _ => throwError m!"unexpected tactic{indentD stx}"
 where 
    throwExs (failures : Array EvalTacticFailure) : TacticM Unit := do
-     let exs := failures.filterMap fun | .abort _ => none | .exception s ex => some (s, ex)
-     if h : 0 < exs.size then
-       -- Recall that `exs[0]` is the highest priority evalFn/macro
-       exs[0].1.restore (restoreInfo := true)
-       throw exs[0].2 -- (*)
+     if let some fail := failures[0]? then
```
Kha, line 165: Note that the refactoring also ensures that we really use the first macro/elaborator's result regardless of the exception type

PR #1376, opened 2022-07-28
feat: accept keywords as constructor names

Kha:
This is something @leodemoura and I have talked about previously - as the new dot notation makes it easy to avoid escaping at use sites, there really is no reason to demand it at the definition site. Also it makes my thesis look nicer.

I left structure fields unchanged as they can be referenced inside the declaration, and the issue is not as acute for them I believe.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,6 +1,15 @@
 Unreleased
 ---------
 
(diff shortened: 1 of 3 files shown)


leodemoura:
> Also it makes my thesis look nicer.

Loved this part :)

PR #1378, opened 2022-07-28
feat: decidable equality instance for subsingleton

fgdorais:
Appears to be an omission. This instance is especially useful for deriving further instances of decidable equality.

### DIFF
--- a/src/Init/Core.lean
+++ b/src/Init/Core.lean
@@ -531,6 +531,9 @@ protected def Subsingleton.helim {α β : Sort u} [h₁ : Subsingleton α] (h₂
   apply heq_of_eq
   apply Subsingleton.elim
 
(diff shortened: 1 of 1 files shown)


digama0:
I think this was a performance issue in mathlib; it may need further review.

fgdorais:
Right. This instance could be applied anywhere. Perhaps it's better to patch the derive handler instead.

gebner:
It's also a diamond (`Fin 0 × Fin 0` would have 11 reasonable decidability instances then).

leodemoura:
Based on the comments above, I am planning to close this PR soon. Ok?

fgdorais:
I approve closing. It would be nice for the derive handler for decidable equality to handle all forms of subsingletons, but I don't have time to do this on my own.

PR #271, opened 2021-01-16
fix: pass (some) cmdline arguments to watchdog and workers

cpehle:
Capture some of the command line arguments in an argument
vector, that then gets passed to the watchdog and
subsequently the worker processes. Currently flags like
"-T" and "-j2" are not passed on to the watchdog and
worker processes.

Addresses #246

PR #1379, opened 2022-07-28
fix: forward memory and stack sizes to server

Vtec234:
This is to prevent file workers on resource-constrained machines from taking over all memory.


Kha:
Could you remind me as to what arguments (apart from `--server`) we did *not* want to forward?

Vtec234:
I'm not quite sure! This was #271 but there doesn't seem to be an explanation.

Kha:
Right you are. Feel free to forward everything but `--server` :smile: .

Vtec234:
I think all the args that make sense to forward should be forwarded now. The remaining ones are mostly of the form "print/don't print something on the command line".

PR #1381, opened 2022-07-28
Add 'suspicious unexpander patterns' linter

Kha:
Let's get the linting economy going


Kha:
Now with a guaranteed linter warning-free stdlib

src/Init/Prelude.lean
```diff
@@ -792,6 +792,7 @@ protected theorem Nat.lt_irrefl (n : Nat) : Not (LT.lt n n) :=
 protected theorem Nat.lt_of_le_of_lt {n m k : Nat} (h₁ : LE.le n m) (h₂ : LT.lt m k) : LT.lt n k :=
   Nat.le_trans (Nat.succ_le_succ h₁) h₂
 
+set_option linter.unusedVariables.funArgs false in  -- #1214
```
leodemoura, line 795: Any progress on this issue? I am happy to make changes to the `match`-elaborator.
I would love to see it fixed because I suspect users will hit this issue quite often.

PR #1382, opened 2022-07-28
feat: add a message if Lean 4 is called with the Lean 3 extension

EdAyers:
If the Lean server fails to parse a request header, before throwing it checks whether the string looks like a Lean 3 header (by seeing if it is valid json and has the 'command' and 'seq_num' fields present). If it does then it will emit a different message pointing users to the new extension.

This fixes https://github.com/leanprover/vscode-lean4/issues/224

### DIFF
--- a/src/Init/Control/Except.lean
+++ b/src/Init/Control/Except.lean
@@ -13,11 +13,6 @@ import Init.Coe
 namespace Except
 variable {ε : Type u}
 
-/-- Returns true if the value is `Except.ok`.-/
-def isOk : Except ε α → Bool
-  | Except.ok _ => true
-  | _ => false
-
 @[inline] protected def pure (a : α) : Except ε α :=
   Except.ok a
 
@@ -39,9 +34,12 @@ def isOk : Except ε α → Bool
   | Except.error err => Except.error err
   | Except.ok v      => f v
 
+/-- Returns true if the value is `Except.ok`, false otherwise. -/
 @[inline] protected def toBool : Except ε α → Bool
   | Except.ok _    => true
   | Except.error _ => false
+
+abbrev isOk : Except ε α → Bool := Except.toBool
 
 @[inline] protected def toOption : Except ε α → Option α
   | Except.ok a    => some a


src/Init/Control/Except.lean
```diff
@@ -13,6 +13,11 @@ import Init.Coe
 namespace Except
 variable {ε : Type u}
 
+/-- Returns true if the value is `Except.ok`.-/
+def isOk : Except ε α → Bool
```
Vtec234, line 17: This is `Except.toBool`, but we should still have this alias I think (you can deduplicate the implementation though).

src/Lean/Data/Lsp/Communication.lean
```diff
@@ -25,6 +25,18 @@ section
       let value := ": ".intercalate (value :: rest)
       some ⟨name, value⟩
 
+  /-- Returns true when the string is a Lean 3 request.
+  This means that the user is running a Lean 3 language server that
```
Vtec234, line 29:
```suggestion
  This means that the user is running a Lean 3 language client that
```

src/Lean/Data/Lsp/Communication.lean
```diff
@@ -36,7 +48,11 @@ section
       | some hf =>
         let tail ← readHeaderFields h
         pure (hf :: tail)
-      | none => throw $ userError s!"Invalid header field: {repr l}"
+      | none =>
+        if isLean3Request l then
+          throw $ userError s!"A Lean 3 request was recieved. Please ensure that your editor has a Lean 4 compatible extension installed. For VSCode, this is\n\n    https://github.com/leanprover/vscode-lean4 "
```
Vtec234, line 53:
```suggestion
          throw $ userError s!"A Lean 3 request was received. Please ensure that your editor has a Lean 4 compatible extension installed. For VSCode, this is\n\n    https://github.com/leanprover/vscode-lean4 "
```

Vtec234 (approved):

Kha:
I don't mind the implementation specifically, but is it clear now why the client-based solution was not sufficient?

Kha:
Oh, I guess the keyword is

> whenever the lean4 extension is not enabled

src/Init/Control/Except.lean
EdAyers: ah yes oops

EdAyers:
It happened to me a few times, because I was enabling and disabling extensions. I also think it's reasonable mistake that someone installs the `lean` instead of the `lean4` extension. Maybe instead of this PR the Lean 3 vscode extension should check the version of lean before sending a server request.

Vtec234:
> Maybe instead of this PR the Lean 3 vscode extension should check the version of lean before sending a server request.

I like this solution since we adapt one server rather than every editor extension out there

Issue #1214, opened 2022-06-13
Unused variables false positive with match


gebner:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

Both `H1` and `H2` are marked as unused even though they're directly referenced in the match arms:
```lean
def Nat.discriminate (n : Nat) (H1 : n = 0 → α) (H2 : ∀ m, n = succ m → α) : α :=
  match n with
  | 0 => H1 rfl
  | succ m => H2 m rfl
```

### Versions

Nightly from June 13.

Vtec234:
I came across possibly another instance of this bug:
```lean
theorem Fin.eq_of_val_eq {n : Nat} : ∀ {x y : Fin n}, x.val = y.val → x = y
  /- unused variable `val._@.lib.Crypto.GF2.GF2BV._hyg.631` -/
  /- unused variable `isLt._@.lib.Crypto.GF2.GF2BV._hyg.632` -/
  | ⟨_, _⟩, _, rfl => rfl
```
We need the destructuring for dependent elimination to succeed.

leodemoura:
@Vtec234 I have seen this bug too at `Prelude.lean`. I didn't analyze it yet, but it is not due to discriminant generalization like the one posted by Gabriel.

leodemoura:
@Kha Do you know what is triggering the bug in @Vtec234's example?

Kha:
This is the same class of bugs as 41dfd06e8cbcdfa0df3dd9daf89fdc802a0096e7 and 143b2b49c8. Originally I assumed that this should be fixed in the info tree as in those commits so that all info consumers ignore them automatically. But since the goal view *does* show these variables, it could be argued that e.g. hover should show them as well. So I made the logic local to the linter after all: https://github.com/leanprover/lean4/pull/1221. Does that make sense?

gebner:
Generally speaking, I don't think we should reduce the information available via the info tree in order to fix a particular linter.  The info tree should offer as much information as we can gather.

As for underscores specifically, I think it's clearer to treat it as a binder.  After all, as you say it does bind a variable and you can access it via `‹_›`.  (And ideally that smiley would be highlighted as a reference.)

Kha:
> As for underscores specifically, I think it's clearer to treat it as a binder

[...]

PR #1383, opened 2022-07-28
chore: suppress unused variable linter false positives

digama0:
and add a reference to #1214 so that it will be cleaned up when that gets fixed.


Kha:
I'd already addressed this in #1381. Thanks for the additional references though!

PR #1384, opened 2022-07-29
feat: add go-to-def for simple attributes

digama0:
* [x] Depends on: #1386 
* [x] Depends on: #1387
* `AttributeImpl` and related definitions are extended with a `ref : Name` field which gives the name of a declaration which will be the target for go-to-definition queries on the attribute.
* Top level attribute declarations use `decl_name%` to supply the `ref` field as the declaration under construction, so lookups will find the declaration.
* `elabAttr` is modified to use the `ref` field for the info tree if it's an `Attr.simple`. (The `ref` field is ignored for custom parsers in syntax category `attr`; in this case it jumps to the syntax definition.)


Kha:
The approach looks sensible, and `decl_name%` should be useful by itself. In that light, could you PR the two parts separately? We can do the stage 0 update after each merge, though I assume your second update-stage0 was necessary to (mostly) placate the CI?

digama0:
How do you do Lake updates? I think CI is breaking because it necessitates a change to one file from Lake. Should I make a simultaneous PR there?

Kha:
@digama0 I didn't think of this before, but couldn't we use `(ref : Name := by exact decl_name%)` at the `register*` functions to avoid the churn at call sites?

digama0:
Sure, I'll try that. But it would have to go at the end, and I worry that it would compete with other optional args of the function

src/Lean/Compiler/InitAttr.lean
```diff
@@ -70,7 +71,11 @@ unsafe def registerInitAttrUnsafe (attrName : Name) (runAfterImport : Bool) : IO
   }
 
 @[implementedBy registerInitAttrUnsafe]
-opaque registerInitAttr (attrName : Name) (runAfterImport : Bool) : IO (ParametricAttribute Name)
+private opaque registerInitAttrInner (attrName : Name) (runAfterImport : Bool) (ref : Name) : IO (ParametricAttribute Name)
```
digama0, line 74: This is needed because `implementedBy` does not seem to like auto params; I tried putting the auto param on both or just one or the other and they all get rejected. So I settled for another wrapper around the unsafe impl. (We should probably move term-`unsafe` to lean core to clean this kind of thing up.)

Kha:
Nice, much better I'd say!

Kha:
I pushed a commit that hides the somewhat confusing type hover and term goal. Actually putting the attribute `descr` in the hover is a follow-up issue I suppose.

Kha (approved):

src/Lean/Compiler/InitAttr.lean
Kha: Right, `implementedBy` currently uses structural equality, and the two tactics are not equal anymore after extraction into separate declarations.

> We should probably move term-unsafe to lean core to clean this kind of thing up.

Yes, I think that would be a good idea.

digama0:
LGTM. If this gets merged the next stage will be actually putting docstrings on the attributes and making sure it all shows up in hovers and such.

Kha:
But that's exactly what `AttributeImplCore.descr` is for, no? Especially since for `EnumAttributes` we have multiple `descr`s but only one potential docstring.

Kha:
Oh, I have a last question: any particular reason `ref` is not declared in `AttributeImplCore`?

digama0:
> But that's exactly what `AttributeImplCore.descr` is for, no? Especially since for `EnumAttributes` we have multiple `descr`s but only one potential docstring.

I'm not sure what `descr` is actually used for - AFAICT it's as yet unused. I am not a fan of doc strings in string literals, you get worse highlighting and it heavily discourages multi-line docs. So I think a reasonable next step would be to see whether it is reasonable to remove them entirely and replace them with doc comments.

For EnumAttributes, it's true that the way things are currently set up you can only have one doc comment for the whole bunch. That's not necessarily terrible - you probably will want to know what other options are available in any case - but we could also solve that syntactically with a custom command with an `inductive` like syntax with multiple internal doc strings.

> Oh, I have one last question: any particular reason ref is not declared in AttributeImplCore?

The place that actually has the go-to-definition logic seems to assume it's an `AttributeImpl`, so that seemed like the more conservative option. I'm not actually sure what else `AttributeImplCore` is used for, but looking at it now it seems it exists to be the common base class of `AttributeImpl` and `ParametricAttributeImpl`, so since I ended up putting a `ref` in both I may as well lift it to `AttributeImplCore`.

Kha:
Oops, merge was too slow :< ...

digama0:
@Kha should be ready for merge again

Kha:
> I am not a fan of doc strings in string literals, you get worse highlighting and it heavily discourages multi-line docs. So I think a reasonable next step would be to see whether it is possible to remove them entirely and replace them with doc comments.

I see your point, but we should make sure everyone is on board before embarking on such refactorings -- /cc @leodemoura 

> but we could also solve that syntactically with a custom command with an `inductive` like syntax with multiple internal doc strings.

I really hope that will not be necessary given that we have exactly 3 enum attributes

digama0:
After thinking about it some more, I think it might be better to have both `descr` and the doc string. The idea would be that `descr` is for the one-line description which is useful in completions and general overviews, and the doc string is the long form documentation for the option.

PR #1387, opened 2022-07-29
feat: use `with_decl_name%` in `initialize`

digama0:
* The `initialize` macro is modified to use `with_decl_name%` so that `decl_name%` will pick up the name of either the anonymous/hygienic `initFn` definition for an anonymous `initialize`, or the name `foo` in `initialize foo <- ...` declarations.

This is extracted from #1384. I think it needs a separate update-stage0 run, and it touches `initialize` which is quite sensitive so it deserves testing separately.

### DIFF
--- a/src/Lean/Elab/Declaration.lean
+++ b/src/Lean/Elab/Declaration.lean
@@ -335,9 +335,9 @@ def elabMutual : CommandElab := fun stx => do
       `($[unsafe%$unsafe?]? def initFn : IO $type := with_decl_name% ?$id do $doSeq
         $[$doc?:docComment]? @[$attrId:ident initFn, $(attrs?.getD ∅),*] $(vis?)? opaque $id : $type)
     else
-      let `(Parser.Command.declModifiersT| ) := declModifiers
+      let `(Parser.Command.declModifiersT| $[$doc?:docComment]? ) := declModifiers
         | Macro.throwErrorAt declModifiers "invalid initialization command, unexpected modifiers"
-      `(@[$attrId:ident] def initFn : IO Unit := do $doSeq)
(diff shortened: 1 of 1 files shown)


src/Lean/Elab/Declaration.lean
```diff
@@ -332,12 +332,12 @@ def elabMutual : CommandElab := fun stx => do
...
         | Macro.throwErrorAt declModifiers "invalid initialization command, unexpected modifiers"
-      `($[unsafe%$unsafe?]? def initFn : IO $type := do $doSeq
+      `($[unsafe%$unsafe?]? def initFn : IO $type := with_decl_name% ?$id do $doSeq
         $[$doc?:docComment]? @[$attrId:ident initFn, $(attrs?.getD ∅),*] $(vis?)? opaque $id : $type)
     else
-      let `(Parser.Command.declModifiersT| ) := declModifiers
+      let `(Parser.Command.declModifiersT| $[$doc?:docComment]? ) := declModifiers
         | Macro.throwErrorAt declModifiers "invalid initialization command, unexpected modifiers"
-      `(@[$attrId:ident] def initFn : IO Unit := do $doSeq)
+      `($[$doc?:docComment]? @[$attrId:ident] def initFn : IO Unit := do $doSeq)
```
Kha, line 340: My assumption so far was that it never makes sense to attach a docstring to a hygiened declaration

digama0: This is needed by #1384, because `decl_name%` inside the `initialize` will pick up the anonymous declaration's name and put it in the `ref` field, thus showing this docstring to anyone who hovers over the attribute.

Kha: Ahh, we do have attributes in anonymous `initialize`s... yeah

PR #1391, opened 2022-07-31
feat: ForIn instance for NameMap and PersistentHashMap

digama0:
Implements some missing `ForIn` instances. For `PersistentHashMap` I started to do a manual implementation but it turns out the auxiliary function needs to bubble up the `done` error so it's the same as just using `foldlM` directly on the `ExceptT σ m` monad. So this implementation just delegates to that.


src/Lean/Parser/Basic.lean
```diff
@@ -1610,7 +1612,8 @@ inductive LeadingIdentBehavior where
   The method `termParser prec` is equivalent to the method above.
 -/
 structure ParserCategory where
-  tables   : PrattParsingTables
+  kinds    : SyntaxNodeKindSet := {}
```
leodemoura, line 1615: This is change is not related to the PR.

digama0: oops

PR #1392, opened 2022-07-31
feat: track parser names by category

digama0:
This is required for some introspective tactics in mathlib. Specifically, we keep in the `ParserCategory` the set of syntax kinds that can produce values of that category. (This is distinct from the kind sets in the parser info, which IIUC collect all kinds that can recursively appear in the generated syntax tree. Here we only collect the kinds that can appear at the root.) I'm using this to generate the tactic list in mathlib.

### DIFF
--- a/src/Lean/Parser/Basic.lean
+++ b/src/Lean/Parser/Basic.lean
@@ -1567,7 +1567,7 @@ structure PrattParsingTables where
 instance : Inhabited PrattParsingTables := ⟨{}⟩
 
 /--
-  The type `leadingIdentBehavior` specifies how the parsing table
+  The type `LeadingIdentBehavior` specifies how the parsing table
   lookup function behaves for identifiers.  The function `prattParser`
   uses two tables `leadingTable` and `trailingTable`. They map tokens
   to parsers.
@@ -1610,9 +1610,23 @@ inductive LeadingIdentBehavior where
   The method `termParser prec` is equivalent to the method above.
 -/
 structure ParserCategory where
+  /-- The name of a declaration which will be used as the target of
+  go-to-definition queries and from which doc strings will be extracted.
+  This is usually a dummy declaration of type `Lean.Parser.Category`
+  created by `declare_syntax_cat`, but for builtin categories this will
+  be the `builtin_initialize` command containing the call to
+  `registerBuiltinParserAttribute`. -/
   declName : Name
+  /-- The list of syntax nodes that can parse into this category.
+  This can be used to list all syntaxes in the category. -/
   kinds    : SyntaxNodeKindSet := {}
+  /-- The parsing tables, which consist of a dynamic set of parser
+  functions based on the syntaxes that have been declared so far. -/
   tables   : PrattParsingTables := {}
+  /-- The `LeadingIdentBehavior`, which specifies how the parsing table
+  lookup function behaves for the first identifier to be parsed.
+  This is used by the `tactic` parser to avoid creating a reserved
+  symbol for each builtin tactic (e.g., `apply`, `assumption`, etc.). -/
   behavior : LeadingIdentBehavior
   deriving Inhabited
 


Kha (approved):
LGTM

We should also add the reverse mapping to find the category of macros in order to make them type-safe. I will look into that.

digama0:
I'm not sure what you mean. Currently, we still don't have a way to connect the `@[builtinTermParser]` and `@[termParser]` attributes to/from the `term : Category` definition. I'm also not sure whether it is safe to assume there is at most one attribute of each kind, maybe that's just paranoia. What macro are you talking about?

Kha:
Ah, I was talking about arbitrary macros. For `macro_rules`, if all the patterns' kinds are of a category `c`, the RHS should be of type `TSyntax c`. It's not quite true that every kind is part of only one category (e.g. `ident`), but I think we can definitely require that for macro-expanded syntax.

digama0:
> Ah, I was talking about arbitrary macros.

Wouldn't macro syntax also show up in the parser tables? I don't think the parser distinguishes between macros and elabs when it's building the syntax object from the input string.

> For `macro_rules`, if all the patterns' kinds are of a category `c`, the RHS should be of type `TSyntax c`. It's not quite true that every kind is part of only one category (e.g. `ident`), but I think we can definitely require that for macro-expanded syntax.

Oh I see, you are thinking about the well typing stuff. For what I need it suffices to be able to enumerate all syntax kinds that can appear in a given category, which does not exclude the possibility that the same syntax kind shows up in more than one syntax category. For well typed macros, it seems like it would suffice to take `c` to be the declared syntax category of the macro, but then you need coercions or something to turn ``TSyntax `Lean.Parser.Term.app`` into ``TSyntax `Lean.Parser.Category.term`` and so on, but I suppose you already have this handled?

src/Lean/Parser/Basic.lean
```diff
@@ -1610,7 +1610,9 @@ inductive LeadingIdentBehavior where
   The method `termParser prec` is equivalent to the method above.
 -/
 structure ParserCategory where
-  tables   : PrattParsingTables
+  ref      : Name
```
leodemoura, line 1613: Could you please add doc strings? :)

leodemoura: BTW, I didn't read the code carefully, but it seems `ref` is the `declName`, right?

digama0: Yes. Shall I rename it to that? (I used `ref` for the attribute since the reference itself was mostly a dummy to access the location and doc string information, and that reasoning also applies here, although for categories there is a clearer relationship to a dedicated dummy definition of type `Lean.Parser.Category`, so `declName` also makes sense as the name.)

Kha: For built-in parsers `ref` will be the `builtin_initialize` instead of the `Category` declaration, right? But I don't mind `declName` either.

src/Lean/Parser/Basic.lean
```diff
@@ -1610,9 +1610,23 @@ inductive LeadingIdentBehavior where
...
   kinds    : SyntaxNodeKindSet := {}
+  /-- The parsing tables, which consist of a dynamic set of parser
+  functions based on the syntaxes that have been declared so far. -/
   tables   : PrattParsingTables := {}
+  /-- The `LeadingIdentBehavior`, which specifies how the parsing table
+  lookup function behaves for the first identifier to be parsed.
+  This is used by the `tactic` parser to avoid creating a reserved
+  symbol for each builtin tactic (e.g., `apply`, `assumption`, etc.). -/
   behavior : LeadingIdentBehavior
   deriving Inhabited
```
Kha, line 1631: Nice.

Kha:
> then you need coercions or something to turn `` TSyntax `Lean.Parser.Term.app `` into `` TSyntax `Lean.Parser.Category.term ``

Not usually, no. `` `(f $x) `` has type `Term`, not ```TSyntax ``app```.

tydeu:
@digama0 Neat! 🎉 Siddharth Bhat and I had been discussing an RFC for this exact feature for an idea we had. I noticed that you keep track of parser declaration names as a set of syntax kinds. This seems problematic for parsers that do not contribute a syntax kind. For example, custom token parsers (following the pattern of `nonReservedSymbol`). Should there be a disclaimer that `kinds` are just parser names and do not necessary produce nodes of the kind name?

Kha:
The assumption that parsers registered in a category produce a homonymous kind (except for a few built-ins like `Parser.Term.ident`, which can be blamed on the absence of `TSyntax` at that time) is not new, the pretty printer makes it as well. Could you elaborate on your use case?

tydeu:
@Kha  Ah, did not realize that this was a common assumption. No use case, was mostly just a question. An example of the kind of parser I was thinking of would be something like my old [`intTypeLit`](https://github.com/tydeu/lean4-papyrus/blob/7f862af8075d86a2d25d1f32fdfd58a4f50251fc/Papyrus/Script/IntegerType.lean) in Papyrus, which parses some subset of another token in a new token. That one does create a distinct kind, but I have played around with similar parsers that do not. Good to know that I should avoid such things in the future.

PR #1396, opened 2022-08-01
fix: unused variables false positive with `match`

### LINKED ISSUE #1214: Unused variables false positive with match

Kha:
Fixes #1214

### DIFF
--- a/src/Lean/Server/InfoUtils.lean
+++ b/src/Lean/Server/InfoUtils.lean
@@ -39,6 +39,12 @@ where go
   | none, node .. => panic! "unexpected context-free info tree node"
   | _, hole .. => pure none
 
+/-- `InfoTree.visitM` specialized to `Unit` return type -/
+def InfoTree.visitM' [Monad m]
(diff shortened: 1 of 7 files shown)


src/Lean/Server/InfoUtils.lean
```diff
@@ -91,6 +97,7 @@ def Info.stx : Info → Syntax
   | ofCompletionInfo i     => i.stx
   | ofCustomInfo i         => i.stx
   | ofUserWidgetInfo i     => i.stx
+  | ofFVarAliasInfo _      => .missing
```
Kha, line 100: This made me briefly question the approach, but I don't see a better solution nor any problems resulting from this in practice

leodemoura (approved):

PR #1398, opened 2022-08-01
fix: allow `{}` as conv

digama0:
This didn't even require any implementation changes: `{ convs }` runs all the convs to produce subgoals, then closes all remaining subgoals with `rfl`, so `{ }` is the vacuous case of this where there are no convs to run, resulting in an alternative way to write `skip`.

PS: The `skip` conv seems to be poorly named. I would have guessed that it is the conv that does nothing, but it is actually the conv that trivially closes the goal. I would suggest renaming `skip` to `rfl` (or `{}`) and introducing a `skip` that does nothing. With the current names, an alternative name for the nonexistent do-nothing conv would be `id`?

### DIFF
--- a/src/Init/Conv.lean
+++ b/src/Init/Conv.lean
@@ -13,7 +13,7 @@ namespace Lean.Parser.Tactic.Conv
 declare_syntax_cat conv (behavior := both)
 
 syntax convSeq1Indented := withPosition((colGe conv ";"?)+)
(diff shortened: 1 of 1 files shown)


leodemoura:
> it is actually the conv that trivially closes the goal. I would suggest renaming skip to rfl (or {}) and introducing a skip that does nothing.

Good for me.

PR #1400, opened 2022-08-01
chore: use Category declarations for builtin cats too

digama0:
It turns out that `Init/Notation.lean` already has `Category` definitions for almost all the builtin categories, so I changed `registerBuiltinParserAttribute` to refer to it directly instead of doing the `decl_name%` thing and getting a random `builtin_initialize`.

PS: Is there a reason the `Category` definition can't go next to the `registerBuiltinParserAttribute` line? I don't see any explicit references to the dummy so it seems like they could just move down, but maybe the `ParserDescr` stuff uses it?

### DIFF
--- a/src/Init/Notation.lean
+++ b/src/Init/Notation.lean
@@ -22,7 +22,7 @@ def term : Category := {}
 def tactic : Category := {}
 /-- The syntax category for elements that can appear in the `do` notation. -/
 def doElem : Category := {}
-/-- The level syntax category. Declarations can be annotated with attributes using the `@[...]` notation. -/
+/-- The level syntax category. -/
 def level : Category := {}
 /-- The attribute syntax category. Declarations can be annotated with attributes using the `@[...]` notation. -/
 def attr : Category := {}


leodemoura:
> PS: Is there a reason the Category definition can't go next to the registerBuiltinParserAttribute line? I don't see any explicit references to the dummy so it seems like they could just move down, but maybe the ParserDescr stuff uses it?

When populating the info tree for `syntax` declarations we need these dummy declarations to be in the environment. Otherwise, we will not show their doc strings when we hover over the `syntax` declarations.

src/Init/Notation.lean
```diff
@@ -22,6 +22,8 @@ def term : Category := {}
 def tactic : Category := {}
 /-- The syntax category for elements that can appear in the `do` notation. -/
 def doElem : Category := {}
+/-- The level syntax category. Declarations can be annotated with attributes using the `@[...]` notation. -/
```
leodemoura, line 25: Could you please remove `Declarations can be annotated with attributes using the `@[...]` notation.` ?
It seems to be a copy&paste leftover.

digama0: this comment was completely rewritten in the other commit anyway

PR #1401, opened 2022-08-01
doc: document all the syntax categories

digama0:
* [x] depends on #1400 to avoid later merge conflicts


leodemoura:
Disabled the auto-merge. I missed the dependency above.

PR #1405, opened 2022-08-01
fix: token antiquotations should create synthetic positions

Kha:


tydeu:
So, to clarify, with this change, what is the proper approach to tag an element of an antiquote with an original source position (or, at the very least, the same leading and trailing space)?

Kha:
@tydeu What is your use case?

tydeu:
@Kha Alloy. For example: 
https://github.com/tydeu/lean4-alloy/blob/e07d9cd971920b0116803411f22541ba0c6787e2/Alloy/C/Command.lean#L60-L61
https://github.com/tydeu/lean4-alloy/blob/e07d9cd971920b0116803411f22541ba0c6787e2/Alloy/C/Syntax.lean#L69-L71

Kha:
That doesn't really tell me anything. Why does a macro need to preserve formatting?

Kha:
To be more explicit, I think there's an argument to be made that `synthetic` source infos should be able to contain leading and trailing whitespace information as well, in which case token antiquotations could continue to set that. Most macros won't care either way, but even with a perfect code formatter that could be important for preserving a sensible position for comments in the macro input. Though for that use case one might usually want to handle the preceding and the trailing whitespace separately, so I'm not sure if the current token antiquotations are even the right choice.

tydeu:
@Kha 

> That doesn't really tell me anything. Why does a macro need to preserve formatting? Are users expected to look at the output?

The spacing is important for a few reasons:

* It keeps the formatting nice so that it can be read by humans (e.g., for debugging purposes). 
* It preserves significant whitespace (e.g., newlines in preprocessor commands). While not as important in Alloy, because C is mostly whitespace insensitive, it would be important DSLs which take a similar approach to Alloy for embedding foreign language that are more whitespace sensitive (e.g., Python).
* It impacts Lean -> C and C -> Lean position conversions for the LSP.

> To be more explicit, I think there's an argument to be made that `synthetic` source infos should be able to contain leading and trailing whitespace information as well, in which case token antiquotations could continue to set that. 

I would quite like this solution. It feels more principled. It indicates the token in question is synthetic while at the same time enabling prettier reprints. It also makes it feel less like a hack should a constructor of such Syntax wish to transform the original whitespace in some way (e.g., strip comments like Alloy does). Another potential feature of this approach is that synthetic source information created via quotations could conceivably preserve the spacing used in the quotation itself.

> Though for that use case one might usually want to handle the preceding and the trailing whitespace separately, so I'm not sure if the current token antiquotations are even the right choice.

True, I could see utility in handling preceding and trailing whitespace separately. However, one could conceivable accomplish this with the current syntax by making the `SourceInfo` of the reference a combination of the relevant leading and trailing spacing.  My primary concern here is to preserve the ability to tag nodes into quotations with whitespace. Adding new features may be nice, but is much less of a priority.

gebner:
In my opinion, using `Syntax.reprint` to write C files is a big hack and abuse of the syntax infrastructure.  `Syntax` is first and foremost there for Lean syntax.  We shouldn't shoehorn any other applications into it.  Note that mandatory whitespace in Lean is already handled perfectly well using formatters, not leading/trailing whitespace.  `Syntax.reprint` is not expected to produce correct code (even for Lean).

> Most macros won't care either way, but even with a perfect code formatter that could be important for preserving a sensible position for comments in the macro input. Though for that use case one might usually want to handle the preceding and the trailing whitespace separately, so I'm not sure if the current token antiquotations are even the right choice.

I don't think token antiquotations help with tracking comments through macro expansion (or any other transformations that a code formatter should do).  For example, ``macro_rules | `(by exact $x) => x`` drops all comments on `by` and `exact` (and there's not even a single token to attach the antiquotations to).  Vice versa, macros might want to attach the same position to multiple tokens in the output for better error messages (which should not duplicate the comments).

The "comments are whitespace attached to tokens" approach is also painful when you want to modify or add comments, because that's when you need to parse and pretty-print the comments manually.  You also need to figure out which token to attach it to, which is a bit of an art.  (If we allow syntax categories to define their own comment syntax in the future, then you'll also need to figure out which category the token is in!)

FWIW, my preferred comments API would be an out-of-band `MultiMap String.Pos (CommentKind × String)`, which the formatter would put in at appropriate places.  This automatically works with macro expansion, `Syntax.replaceM`, etc., and makes it easy to modify/add comments.

Kha:
> Vice versa, macros might want to attach the same position to multiple tokens in the output for better error messages (which should not duplicate the comments).

That's true, though I expect that syntax transformations where really care about the lexical output (e.g. code actions) would be written in a somewhat different style from regular macros. So it's not directly an issue if they use token antiquotations for different purposes. Unless we have a generic "unfold macro at point" action of course.

> FWIW, my preferred comments API would be an out-of-band MultiMap String.Pos (CommentKind × String), which the formatter would put in at appropriate places.

An interesting idea! Might be a bit awkward to work with for syntax transformations that do want to move around comments, though that's probably hard in any case, perhaps so much so that we won't bother with it anyway. My thoughts so far on preserving comments when the transformation (e.g. the macro unfold action) fails to do so itself is to diff the set of comments in the input and output and add back missing ones at an appropriate position. In theory this could also detect and delete duplicated comments.

Btw, if we want comments to be more structured, I think we should just make them (still subordinated) `Syntax` as well.

tydeu:
@gebner 

> In my opinion, using Syntax.reprint to write C files is a big hack and abuse of the syntax infrastructure.  `Syntax` is first and foremost there for Lean syntax. We shouldn't shoehorn any other applications into it.

Syntax is also used for any DSLs defined in Lean. As the C code is embedded in Lean and makes use of metaprogramming utilities such as macros to produce blocks of C code (e.g., `alloy c include`) it makes sense it would be built off the `Syntax` infrastructure. Furthermore, `Syntax` represents a very generic concrete syntax tree, so I fail to see why it would not be a good baseline for DSLs of any complexity embedded in Lean. 

On the use of `reprint` in particular, I do think it is a bit of a hack and should probably be replaced with a pretty-printer, but I just did not wish to commit the work-hours to that.

gebner:
> That's true, though I expect that syntax transformations where really care about the lexical output (e.g. code actions) would be written in a somewhat different style from regular macros.

I'm not sure if that's desirable.  We already have good and effective syntax for syntax transformations (i.e., syntax matching and quotations as used in macros, etc.).  I don't think anybody wants to learn a second syntax transformation mechanism that is only used in one place.  Or practically speaking, people will just write code actions using syntax matchers because that's the easiest way to do it.

> Syntax is also used for any custom syntax defined in Lean. As the C code is embedded in Lean and makes use of metaprogramming utilities such as macros to produce blocks of C code (e.g., `alloy c include`) it makes sense it would be built off the `Syntax` infrastructure. Furthermore, `Syntax` represents a very generic concrete syntax tree, so I fail to see why it would not be a good baseline for DSLs of any complexity embedded in Lean.

You're presenting a false equivalence here.  Using `Syntax` for DSLs (which is a good idea, and heavily used) does not mean you should put semantically meaningful information in the whitespace fields (which we don't do anywhere else; it is either ignored or left out), or format it using `reprint` (which is only really intended for debugging).

As you mention, the proper solution here is to either (1) use the Lean pretty-printer (hoping that you can stay in the intersection of the lexical syntaxes of Lean and C; and you also need to delete the comments first), or preferably (2) write your own `Syntax → Format` function.  Expecting that `Syntax.reprint` produces valid C syntax is a losing battle.

PR #1407, opened 2022-08-02
Add user widgets tutorial

Vtec234:
Adds a tutorial for user widgets. It is placed in `doc/examples/` because that folder seems to get special treatment with Alectryon and I didn't want to otherwise mess with the build setup.

Also adds some helpers for `RequestM` operations.

Also adds the `include_str` macro as pretty much every user widget needs it, and it seems generally useful.

- [x] ~~Merge leanprover/lean4-samples#3~~ There is a cyclic dependency in that this PR provides `include_str` which the sample needs.
- [x] Rewrite as Lean file for Alectryon

### DIFF
--- a/src/Lean/Server/Requests.lean
+++ b/src/Lean/Server/Requests.lean
@@ -120,7 +120,6 @@ def withWaitFindSnap (doc : EditableDocument) (p : Snapshot → Bool)
     (notFoundX : RequestM β)
     (x : Snapshot → RequestM β)
     : RequestM (RequestTask β) := do
-  let doc ← readDoc
   let findTask ← doc.cmdSnaps.waitFind? p
   mapTask findTask <| waitFindSnapAux notFoundX x
 
@@ -129,7 +128,6 @@ def bindWaitFindSnap (doc : EditableDocument) (p : Snapshot → Bool)
     (notFoundX : RequestM (RequestTask β))
     (x : Snapshot → RequestM (RequestTask β))
     : RequestM (RequestTask β) := do
-  let doc ← readDoc
   let findTask ← doc.cmdSnaps.waitFind? p
   bindTask findTask <| waitFindSnapAux notFoundX x
 


Vtec234:
> [1354](https://github.com/leanprover/lean4/runs/7640359246?check_suite_focus=true#step:12:1355)
[00:49 (00.000514)] lean-doc-test> /build/mdbook-EX1bMe/widgets.md_3.lean:130:0: error: unterminated comment

It looks like triple-backtick blocks are possibly prevented within doc comments. However I need this one since it's JS, not Lean.

EdAyers (approved):
It's fantastic

EdAyers:
Maybe one thing we could sneak in to this PR is an `include_str` macro?

doc/widgets.lean
```diff
@@ -0,0 +1,203 @@
...
+open Lean Widget
+
+/-!
+# The user-widgets system
+
+Proving and programming are inherently interactive tasks. Lots of mathematical objects and data
+structures are visual in nature. *User widgets* let you associate custom interactive UIs with
+sections of a Lean document. User widgets are rendered in the Lean infoview.
+
+![Rubik's cube](images/widgets_rubiks.png)
```
utensil, line 11: The image seems to be still showing the old code with `@[staticJS]`

Vtec234: I think that's okay, it is there simply to give an idea of "type stuff here, UI appears there" rather than serve as an API reference.

EdAyers:
It looks like markdown code blocks within `/-!` comments are not being translated properly (regardless of `lean` vs `lean,ignore` vs `javascript` header). Running `mdbook build` on such a file produces messed up html. Can we delete the javascript code block for now (the javascript code is still present in the string); and then open a new issue to support code blocks?

doc/examples/widgets.lean
```diff
@@ -0,0 +1,183 @@
...
+/-!
+![`#check` as a service](images/widgets_caas.png)
+
+## Building widget sources
+
+While typing JavaScript inline is fine for a simple example, for real developments we want to use
+packages from NPM, a proper build system, and JSX. Thus, most actual widget sources are built with
+Lake and NPM. They consist of multiple files and may import libraries which don't work as ESModules
+by default. On the other hand a widget source is just a single string. Readers familiar with web
+development may already have guessed that to obtain such a string, we need a *bundler*. Two popular
```
Kha, line 174: suggestion: "self-contained string"

src/Lean/Elab/IncludeStr.lean
```diff
@@ -0,0 +1,29 @@
...
+    let path ← evalFilePath path
+    let ctx ← readThe Lean.Core.Context
+    let srcPath := System.FilePath.mk ctx.fileName
+    let some srcDir := srcPath.parent
+      | throwError "cannot compute parent directory of '{srcPath}'"
+    let path := srcDir / path
+    mkStrLit <$> IO.FS.readFile path
+  | _, _ => throwUnsupportedSyntax
+
+end Lean.Elab.Term
```
Kha, line 29: I think this can just go in `BuiltinTerm.lean`.

src/Lean/Server/Requests.lean
```diff
@@ -100,62 +109,71 @@ def waitFindSnapAux (notFoundX : RequestM α) (x : Snapshot → RequestM α)
...
 /-- Create a task which waits for the first snapshot matching `p`, handles various errors,
 and if a matching snapshot was found executes `x` with it. If not found, the task executes
 `notFoundX`. -/
 def withWaitFindSnap (doc : EditableDocument) (p : Snapshot → Bool)
-  (notFoundX : RequestM β)
-  (x : Snapshot → RequestM β)
+    (notFoundX : RequestM β)
+    (x : Snapshot → RequestM β)
     : RequestM (RequestTask β) := do
+  let doc ← readDoc
```
Kha, line 123: Hmm?

src/Lean/Server/Requests.lean
```diff
@@ -65,31 +63,42 @@ structure RequestContext where
   initParams    : Lsp.InitializeParams
 
 abbrev RequestTask α := Task (Except RequestError α)
+abbrev RequestT m := ReaderT RequestContext <| ExceptT RequestError m
```
Kha, line 66: Is this all just for being able to call `readDoc` inside of `run*M` instead of outside?

Kha (changes requested):
Great read!

src/Lean/Server/Requests.lean
Vtec234: It's mostly so that we can type the argument to `runM` as `RequestT M`. Consequently the code inside can still access the request handler context, and throw request errors rather than only a generic `IO.Error`.

src/Lean/Server/Requests.lean
Vtec234:
```suggestion
```

src/Lean/Server/Requests.lean
```diff
@@ -100,62 +109,71 @@ def waitFindSnapAux (notFoundX : RequestM α) (x : Snapshot → RequestM α)
...
   mapTask findTask <| waitFindSnapAux notFoundX x
 
 /-- See `withWaitFindSnap`. -/
 def bindWaitFindSnap (doc : EditableDocument) (p : Snapshot → Bool)
-  (notFoundX : RequestM (RequestTask β))
-  (x : Snapshot → RequestM (RequestTask β))
+    (notFoundX : RequestM (RequestTask β))
+    (x : Snapshot → RequestM (RequestTask β))
     : RequestM (RequestTask β) := do
+  let doc ← readDoc
```
Vtec234, line 132:
```suggestion
```

src/Lean/Server/Requests.lean
Kha: Oh I see, it's about exceptions. As you may know I'd like to avoid duplicate exception layers, but I guess that's not realistic here. What error codes other than `internalError` are you expecting widget handlers to use?

Vtec234: At least `invalidParams` for incorrect arguments.

Kha: I had hoped that would be covered by the deserialization, but I guess the param type will not necessarily describe all argument constraints. I'm still not sure if it's worth the (not terribly big) code complication. The RPC interface of a widget should probably be considered private to the widget, so it's unlikely the client code will expect and handle invalidParams specially, no?

Vtec234: I could definitely see people writing widget-support RPC libraries which can be used by multiple JS codebases, in which case some kind of error handling needs to happen. For example, our own [getWidgetSource](https://github.com/leanprover/lean4/pull/1407/files#diff-1a6eaee5d12ad418d2657e155b36b353081ee0bc8fefbe6628c6ea0779d1258eR108) is now a bit simpler since it can run in `CoreM` and use all the standard functions rather than accessing parts of the `Snapshot` data.

doc/examples/widgets.lean
```diff
@@ -0,0 +1,183 @@
...
+We can register a widget source with the `@[widget]` attribute, giving it a friendlier name
+in the `name` field. This is bundled together in a `UserWidgetDefinition`.
+
+A *widget instance* is then the identifier of a `UserWidgetDefinition` (so `` `helloWidget ``,
+not `"Hello"`) associated with a range of positions in the Lean source code. In our example,
+the `#widget` command places a widget instance with the entire line as its range. We can think
+of a widget instance as an instruction for the infoview: "when the user places their cursor here,
+please render the following widget".
+
+Every widget instance also contains a `props : Json` value. This value is passed as an argument
```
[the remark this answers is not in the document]
Vtec234: I left that in but clarified widget instances a little above, is it better? In this case `contains` just means that wherever a widget instance is stored, we also store props.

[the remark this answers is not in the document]
Kha: Hah, now I understand why it's a JSON

Kha (approved):

PR #1414, opened 2022-08-03
feat: make MissingDocs extensible

digama0:
This should also fix the compile time cost issue noticed in #1390. This adds an environment extension and attribute to add `MissingDocs.Handler`s for user-defined commands, so that they can also be docs linted. The only thing you have to do in a handler is identify if there is a docstring on the syntax, and if not then call the `lintNamed` function with the span of the new identifier.

You can implement either `Handler` or `SimpleHandler`:
* `SimpleHandler := Syntax -> CommandElabM Unit` gets the syntax of the command and possibly reports a lint warning. It is only called if the syntax has the syntax kind specified in the attribute, and if the missingDocs linter itself is enabled.
* `Handler := Bool -> Syntax -> CommandElabM Unit` is called even if the linter is disabled, and is passed the enabled state. This is needed for commands that contain other commands, because nested commands can have a `set_option` that locally re-enables the linter.

### DIFF
--- a/tests/lean/linterMissingDocs.lean
+++ b/tests/lean/linterMissingDocs.lean
@@ -85,9 +85,9 @@ my_command x
 
 open Lean.Linter.MissingDocs in
 @[missingDocsHandler myCmd]
-def handleMyCmd : SimpleHandler := fun stx => do
-  if stx[0].isNone then
-    lintNamed stx[2] "my_command"
+def handleMyCmd : SimpleHandler := fun
+  | `(my_command $x:ident) => lintNamed x "my_command"
+  | _ => pure ()
 
 /-- doc -/
 my_command y
--- a/tests/lean/linterMissingDocs.lean.expected.out
+++ b/tests/lean/linterMissingDocs.lean.expected.out
@@ -24,4 +24,4 @@ linterMissingDocs.lean:78:13-78:22: warning: missing doc string for class abbrev
 linterMissingDocs.lean:80:16-80:24: warning: missing doc string for option myOption [linter.missingDocs]
 linterMissingDocs.lean:82:14-82:19: warning: missing doc string for elab myCmd [linter.missingDocs]
 linterMissingDocs.lean:88:4-88:15: warning: missing doc string for public def handleMyCmd [linter.missingDocs]
-linterMissingDocs.lean:95:11-95:12: warning: missing doc string for my_command [anonymous] [linter.missingDocs]
+linterMissingDocs.lean:95:11-95:12: warning: missing doc string for my_command z [linter.missingDocs]


tests/lean/linterMissingDocs.lean
```diff
@@ -78,3 +78,18 @@ macro "my_macro" : term => `(my_elab)
...
+
+elab (name := myCmd) (docComment)? "my_command" ident : command => pure ()
+
+my_command x
+
+open Lean.Linter.MissingDocs in
+@[missingDocsHandler myCmd]
+def handleMyCmd : SimpleHandler := fun stx => do
+  if stx[0].isNone then
+    lintNamed stx[2] "my_command"
```
Kha, line 90: This wouldn't have happened with a syntax match :)

src/Lean/Linter/MissingDocs.lean
```diff
@@ -19,111 +17,176 @@ register_builtin_option linter.missingDocs : Bool := {
 
 def getLinterMissingDocs (o : Options) : Bool := getLinterValue linter.missingDocs o
 
+
+namespace MissingDocs
+
+abbrev SimpleHandler := Syntax → CommandElabM Unit
+abbrev Handler := Bool → SimpleHandler
```
gebner, line 24:
```suggestion
abbrev SimpleHandler ks := TSyntax ks → CommandElabM Unit
abbrev Handler ks := Bool → SimpleHandler ks
```
Then we could also drop the `foo` in `@[missingDocsHandler foo]`.

digama0: That makes the `evalConst` a lot trickier, I think?

tests/lean/linterMissingDocs.lean
digama0: I'm not sure what you mean by "this". The code is correct AFAICT. There are two reasons I used explicit syntax indexing instead of syntax match in most missing docs handlers:

1. The matches were a huge performance issue.
2. Making an exhaustive match of all arguments of a command is tricky to get right, and is not forward compatible with changes (extensions) to the existing syntax. (Of course using indices is also a forward compatibility hazard but there are fewer things that can go wrong.)

Neither of these apply to the test case (or the user code that will presumably be modeled on the test case), so I've changed this one to use syntax match.

Kha: Sorry for being oblique about the bug, but looks like you found it. Anyway, regarding 1., do you mean the performance issue we noticed in the PR (which would not get any better with syntax match, yeah), or syntax match compared to regular match?

digama0: Actually I wouldn't say I've found the bug. I noticed that the test case was wrong but the code still seems like it should be correct: the `stx` here should have the form `(docComment)? "my_command" ident` and hence `stx[2]` should be a `Syntax.ident`. The code did seem to find the correct span, but `getId` was returning `.anonymous` and I don't get that. Did we introduce another level of indirection on identifiers?

Regarding syntax match vs regular match, it's true we experienced the issues with regular match but there is no reason to believe syntax match will be better. I took a look at the generated code and maybe it has something to do with the matcher construction? It does a whole bunch of useless stuff like proving that previous cases don't apply. It also does a nested pattern match using `Name.rec` instead of an equality if-else chain like the naive code would, although it does use if-else chains for the underlying string equality checks. Honestly I'm still not totally sure what the cost was; it might be worthwhile to take the previous code and turn it into a test case since it seems to be really stressing something.

Kha: Oh my, I completely misread the indices! It's actually from `macro/elab`: https://github.com/leanprover/lean4/blob/fbef8a32e19f8d6c473216aaa7ffc0eca6ff20ab/src/Lean/Elab/MacroArgUtil.lean#L42-L43. But it occurs *only* when ignoring a parameter in the actual macro/elab and then later accessing it in a different context like your linter handler, so I'd like to claim that it should not happen for real use cases.

src/Lean/Linter/MissingDocs.lean
Kha: Yeah, this is an interesting general refactoring that we can apply to other places as well, especially after strengthening the macro typing story.

Kha (approved):

leodemoura:
Merged manually due to stage0 conflicts

PR #1416, opened 2022-08-04
feat: add include_str macro

EdAyers:
`include_str` is a macro that takes a path argument and elaborates to the string contained in that file. It is useful for creating user widgets so I placed it in the Widget directory, however it might be better placed in Util. Where should it live? It uses the `elab` keyword I'm not sure if that causes any bootstrapping issues.


EdAyers:
oops it's already been added to https://github.com/leanprover/lean4/pull/1407

PR #1418, opened 2022-08-04
fix: do not show inaccessible variable in hover

Kha:


tests/lean/interactive/hoverBinderUndescore.lean.expected.out
```diff
@@ -2,29 +2,29 @@
  "position": {"line": 1, "character": 5}}
 {"range":
  {"start": {"line": 1, "character": 5}, "end": {"line": 1, "character": 6}},
- "contents": {"value": "```lean\nx✝ : Nat\n```", "kind": "markdown"}}
+ "contents": {"value": "```lean\nNat\n```", "kind": "markdown"}}
```
Kha, line 5: Seems sensible? Test case is
```
def f : Nat → Bool → Nat :=
 fun _ _ => 5
   --^ textDocument/hover
     --^ textDocument/hover
```

PR #1424, opened 2022-08-05
doc: add more docs to Init.Prelude

digama0:
I'm not done yet, but seeing as I managed to set up a race condition with Leo's docs I thought I would put what I have so far and update it as I go.


leodemoura:
@digama0 Could you please fix the tests?

PR #1425, opened 2022-08-05
feat: add unused/deprecation diagnostic tags

digama0:
This makes the deprecation warning cause the text to be drawn in vscode with a strikethrough, and the unused variables linter warnings cause the text to be faded.

### DIFF
--- a/src/Lean/Widget/InteractiveDiagnostic.lean
+++ b/src/Lean/Widget/InteractiveDiagnostic.lean
@@ -4,16 +4,8 @@ Released under Apache 2.0 license as described in the file LICENSE.
 
 Authors: Wojciech Nawrocki
 -/
-import Lean.Data.Lsp
-import Lean.Message
-import Lean.Elab.InfoTree
-
+import Lean.Linter.UnusedVariables
 import Lean.Server.Utils
-import Lean.Server.Rpc.Basic
-
-import Lean.Widget.Basic
-import Lean.Widget.TaggedText
-import Lean.Widget.InteractiveCode
 import Lean.Widget.InteractiveGoal
 
 namespace Lean.Widget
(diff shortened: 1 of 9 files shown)


src/Init/Util.lean
```diff
@@ -38,6 +38,7 @@ def dbgSleep {α : Type u} (ms : UInt32) (f : Unit → α) : α := f ()
 @[neverExtract, inline] def panicWithPosWithDecl {α : Type u} [Inhabited α] (modName : String) (declName : String) (line col : Nat) (msg : String) : α :=
   panic (mkPanicMessageWithDecl modName declName line col msg)
 
+set_option linter.unusedVariables.funArgs false in
```
Kha, line 41: Mmh, why is this necessary now?

src/Lean/Widget/InteractiveDiagnostic.lean
```diff
@@ -160,21 +152,19 @@ def msgToInteractiveDiagnostic (text : FileMap) (m : Message) (hasWidgets : Bool
     | none        => low
   let range : Range := ⟨low, high⟩
   let fullRange : Range := ⟨low, fullHigh⟩
-  let severity := match m.severity with
+  let severity? := match m.severity with
```
Kha, line 155: Did the type change? If not, that's really confusing.

Kha (approved):
Would be great if you could put independent changes into separate commits or even better PRs.

src/Lean/Widget/InteractiveDiagnostic.lean
digama0: I just wanted it to have the same name as the field so I could use the abbreviated syntax for structure initialization at the end of the function

src/Init/Util.lean
digama0: I didn't look too deeply into it, but I was getting a warning here and there are similar `set_option` lines elsewhere in the file.

Kha: I can't reproduce that. In fact, we now compile with `warningAsError`, so the CI should have failed on master if it happened while building Lean.

src/Lean/Widget/InteractiveDiagnostic.lean
Kha: Apparently it's always been `Option`. An explicit `some` sounds good!

src/Init/Util.lean
digama0: I'll remove it. There shouldn't be any other unrelated changes in this PR, except for the spacing fix and the comment typo. The import reductions may not be the minimum required but I could move those to a separate PR if desired.

leodemoura: > I can't reproduce that. In fact, we now compile with `warningAsError`, so the CI should have failed on master if it happened while building Lean.

BTW, I am not sure whether the `warningAsError` is working on the CI. I have found "unused variable warnings" in the last few days, and I believe the CI did not report any problem. Could you please confirm whether this is working or not?

PR #1427, opened 2022-08-05
Fix `warningAsError`

Kha:


src/Lean/Log.lean
```diff
@@ -45,13 +45,20 @@ def getRefPosition : m Position := do
...
 /--
 Log the message `msgData` at the position provided by `ref` with the given `severity`.
 If `getRef` has position information but `ref` does not, we use `getRef`.
 We use the `fileMap` to find the line and column numbers for the error message.
 -/
-def logAt (ref : Syntax) (msgData : MessageData) (severity : MessageSeverity := MessageSeverity.error): m Unit :=
-  unless severity == MessageSeverity.error && msgData.hasSyntheticSorry do
+def logAt (ref : Syntax) (msgData : MessageData) (severity : MessageSeverity := MessageSeverity.error) : m Unit :=
+  unless severity == .error && msgData.hasSyntheticSorry do
+    let severity := if severity == .warning && warningAsError.get (← getOptions) then .error else severity
```
Kha, line 61: This is still not fail-safe if someone does not go through `MonadLog`, but hopefully good enough. The alternative would be to apply the transformation when reading the log.

PR #1431, opened 2022-08-06
feat: add builtin MissingDocs handler attr

digama0:
The previous change to use a `@[missingDocsHandler]` attribute broke usage inside the lean repo itself. I guess that's what `builtin*` attributes are for.


Kha:
> The previous change to use a `@[missingDocsHandler]` attribute broke usage inside the lean repo itself.

Hah, I didn't even know that was an objective. Not that I mind.

PR #1434, opened 2022-08-06
feat: go to def on parser aliases

digama0:
Parser aliases now keep track of the declaration they resolve to, and syntax will correctly annotate the spans so that e.g. `syntax "foo" ppSpace num : term` will give hovers for `ppSpace` and `num`.

### DIFF
--- a/src/Lean/Elab/Syntax.lean
+++ b/src/Lean/Elab/Syntax.lean
@@ -53,11 +53,17 @@ def ensureUnaryOutput (x : Term × Nat) : Term :=
 @[inline] private def withNestedParser (x : ToParserDescr) : ToParserDescr := do
   withReader (fun ctx => { ctx with leftRec := false, first := false }) x
 
-/-- (Try to) a term info for the category `catName` at `ref`. -/
+/-- (Try to) add a term info for the category `catName` at `ref`. -/
 def addCategoryInfo (ref : Syntax) (catName : Name) : TermElabM Unit := do
-    let declName := ``Lean.Parser.Category ++ catName
-    if (← getEnv).contains declName then
-      addTermInfo' ref (Lean.mkConst declName)
+  let declName := ``Lean.Parser.Category ++ catName
+  if (← getEnv).contains declName then
+    addTermInfo' ref (Lean.mkConst declName)
+
+/-- (Try to) add a term info for the alias with info `info` at `ref`. -/
+def addAliasInfo (ref : Syntax) (info : Parser.ParserAliasInfo) : TermElabM Unit := do
+  if (← getInfoState).enabled then
+    if (← getEnv).contains info.declName then
+      addTermInfo' ref (Lean.mkConst info.declName)
 
 def checkLeftRec (stx : Syntax) : ToParserDescrM Bool := do
   let ctx ← read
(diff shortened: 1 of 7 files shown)


src/Lean/Elab/Syntax.lean
```diff
@@ -201,18 +201,20 @@ where
         throwError "unknown parser declaration/category/alias '{id}'"
 
   processSepBy (stx : Syntax) := do
+    let sepBy := mkCIdentFrom stx[0] ``ParserDescr.sepBy
     let p ← ensureUnaryOutput <$> withNestedParser do process stx[1]
     let sep := stx[3]
     let psep ← if stx[4].isNone then `(ParserDescr.symbol $sep) else ensureUnaryOutput <$> withNestedParser do process stx[4][1]
     let allowTrailingSep := !stx[5].isNone
-    return (← `(ParserDescr.sepBy $p $sep $psep $(quote allowTrailingSep)), 1)
+    return (← `($sepBy:ident $p $sep $psep $(quote allowTrailingSep)), 1)
```
digama0, line 209: @Kha This doesn't work. (I mean, it doesn't hurt but it doesn't do anything.) I was hoping that by annotating the span on the `ParserDescr.sepBy` I would get hover info on `sepBy(` in syntax, but it doesn't seem to attach any info. Any idea why?

src/Lean/Elab/Syntax.lean
```diff
@@ -235,10 +237,25 @@ where
...
+@[builtinTermElab parserAlias] def elabParserAlias : TermElab
+  | `(parser_alias% $id), expectedType? => do
+    let aliasName := id.getId.eraseMacroScopes
+    if (← getInfoState).enabled then
+      let decl := (← Parser.getParserAliasInfo aliasName).declName
+      if let some info := (← getEnv).find? decl then
+        pushInfoLeaf <| Info.ofTermInfo {
+          elaborator := .anonymous
+          lctx := .empty
+          expr := ← mkConst decl (info.levelParams.map mkLevelParam)
```
Kha, line 253: That's `mkConstWithLevelParams`. Not that I'd ever expect levels to come up here.

digama0: it's an adaptation of it to avoid the error case

src/Lean/Elab/Syntax.lean
Kha: This should add synthetic positions, which are ignored by the language server

src/Lean/Elab/Syntax.lean
Kha: I don't see anything wrong with failing here, we'll fail below anyway in that case. But I guess it's fine this way.

digama0: If it does `throwError`, then bad things happen. I suppose I could catch the error here, but it's not supposed to prevent the syntax from doing what it does in the case that the alias wasn't found or it had a bogus `declName` attached.

src/Lean/Elab/Syntax.lean
Kha: Untested, `(with_annotate_term $(stx[0]) ParserDescr.sepBy) ...` might work

digama0: great, that worked. I pushed a simplification which avoids the need for `parser_alias%`, since it turns out that `toParserDescr` is actually in `TermElabM` rather than `IO` or `MacroM` like I originally thought, so you can just add terminfo directly.

src/Lean/Elab/Syntax.lean
Kha: Ah I guess that's true. Nice!

Kha (approved):

Kha:
Could you add a test as well, e.g. in `interactive/hover.lean`?

PR #1439, opened 2022-08-07
chore: remove List.redLength

digama0:
I think the need for this function no longer exists now that `List.length` and `List.lengthTR` exist, but this should get a benchmark run just in case.

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -1895,14 +1895,9 @@ def List.toArrayAux : List α → Array α → Array α
   | nil,       r => r
   | cons a as, r => toArrayAux as (r.push a)
 
-@[inlineIfReduce]
-def List.redLength : List α → Nat
-  | nil       => 0
-  | cons _ as => as.redLength.succ
-
 @[inline, matchPattern, export lean_list_to_array]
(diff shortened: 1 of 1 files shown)


leodemoura:
Note that `redLength` uses the attribute `[inlineIfReduce]`.

digama0:
Does that mean that it will be inlined only if it completely evaluates, e.g. `[a, b, c].redLength` is inlined to `3` but `(a :: b :: l).redLength` is not inlined at all? Does it work to have `redLength` be a `@[inlineIfReduce]` wrapper around `length` or `lengthTR`?

leodemoura:
> Does that mean that it will be inlined only if it completely evaluates, e.g. `[a, b, c].redLength` is inlined to `3` but `(a :: b :: l).redLength` is not inlined at all? Does it work to have `redLength` be a `@[inlineIfReduce]` wrapper around `length` or `lengthTR`?

Didn't check.

The plan is to make sure annotations such as `inline`, `specialize`, 
`inlineIfReduce` can be nested in declarations, and is correctly supported by the new code generator. We will be able to remove `redLength` after we delete the current code generator.

PR #1444, opened 2022-08-07
fix: annotate all syntax nodes produced by quotations as synthetic

Kha:
Since we've been talking about synthetic positions anyway, let's finally implement https://github.com/leanprover/lean4/pull/987#issuecomment-1073766025.

Regarding implementation of macros, the important change is that we should use `withRef` instead of token antiquotations for attaching positions to a whole subtree, see https://github.com/leanprover/lean4/commit/3f92092f2835cc5b3aee3b9b9d8b1e27ecfaef15. For single tokens such as in https://github.com/leanprover/lean4/blob/578adcd7f047c4879f144614a114fb212916bd40/src/Init/Conv.lean#L132 there is nothing wrong with continuing to use token antiquotations.

Resolves #987


src/Lean/Elab/Do.lean
```diff
@@ -693,11 +693,11 @@ private def expandDoIf? (stx : Syntax) : MacroM (Option Syntax) := match stx wit
     let mut eIsSeq := true
     for (i, cond, t) in Array.zip (is.reverse.push i) (Array.zip (conds.reverse.push cond) (ts.reverse.push t)) do
       e ← if eIsSeq then pure e else `(doSeq|$e:doElem)
-      e ← withRef cond <| match cond with
-        | `(doIfCond|let $pat := $d) => `(doElem| match%$i $d:term with | $pat:term => $t | _ => $e)
-        | `(doIfCond|let $pat ← $d)  => `(doElem| match%$i ← $d    with | $pat:term => $t | _ => $e)
-        | `(doIfCond|$cond:doIfProp) => `(doElem| if%$i $cond:doIfProp then $t else $e)
-        | _                          => `(doElem| if%$i $(Syntax.missing) then $t else $e)
+      e ← withRef (mkNullNode #[i, t]) <| match cond with
```
gebner, line 696: This could still produce a negative range if `i` and `t` were swapped, or am I missing something?

Kha: Yes, that's true. It looks like this really was working around a position loss in `if`, so I fixed that one instead

Kha: We do have a few more occurrences of this pattern, but there the chance of getting swapped is much lower than here I'd say.

Kha:
@leodemoura Do you have a test case for https://github.com/leanprover/lean4/commit/3e7d45aaba7e5e91ba72d0f0b4ecb697c23ed648? This is the only remaining use of token antiquotations outside of patterns in `Do.lean`.

leodemoura (approved):

leodemoura:
> @leodemoura Do you have a test case for [3e7d45a](https://github.com/leanprover/lean4/commit/3e7d45aaba7e5e91ba72d0f0b4ecb697c23ed648)? This is the only remaining use of token antiquotations outside of patterns in `Do.lean`.

Unfortunately, I don't. I will try to add some.

leodemoura:
Merged manually due to stage0 conflicts and linter issue (had to temporarily disable `warningAsError`).

PR #1445, opened 2022-08-07
doc: finish Init.Prelude docs

digama0:
* :heavy_check_mark: Init.Prelude docs :tada:
* :black_square_button: everything else...

I put `set_option linter.missingDocs true` at the top of the file, so that should hopefully keep the file documented when new things are added.

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -1088,39 +1088,73 @@ instance (r : α → γ → Sort u) : Trans Eq r r where
 instance (r : α → β → Sort u) : Trans r Eq r where
   trans h' heq := heq ▸ h'
 
-/-- The typeclass behind the notation `a + b : γ` where `a : α`, `b : β`. -/
+/--
+The notation typeclass for heterogeneous addition.
+This enables the notation `a + b : γ` where `a : α`, `b : β`.
+-/
 class HAdd (α : Type u) (β : Type v) (γ : outParam (Type w)) where
-  /-- The implementation of `a + b : γ`. -/
+  /-- `a + b` computes the sum of `a` and `b`.
+  The meaning of this notation is type-dependent. -/
   hAdd : α → β → γ
 
-/-- The typeclass behind the notation `a - b : γ` where `a : α`, `b : β`. -/
+/--
+The notation typeclass for heterogeneous subtraction.
+This enables the notation `a - b : γ` where `a : α`, `b : β`.
+-/
 class HSub (α : Type u) (β : Type v) (γ : outParam (Type w)) where
-  /-- The implementation of `a - b : γ`. -/
+  /-- `a - b` computes the difference of `a` and `b`.
+  The meaning of this notation is type-dependent.
+  * For natural numbers, this operator saturates at 0: `a - b = 0` when `a ≤ b`. -/
   hSub : α → β → γ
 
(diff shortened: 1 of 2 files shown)


leodemoura:
This is great. Thanks!
BTW, why is the first commit an `update-stage0`?

What do you mean by the "everything else" bullet? I am assuming it is a joke, and the PR is ready to be merged :)

digama0:
CI doesn't pass unless #1440 is available in stage0. If you run it on the master branch then it won't be needed anymore.

The everything else is a joke referring to all the other files that need documentation. This is file 1 of N for a depressingly large value of N.

src/Init/Prelude.lean
```diff
@@ -995,94 +1088,184 @@ instance (r : α → γ → Sort u) : Trans Eq r r where
...
+implementation so it can decide when to evaluate it.
+-/
 class HAndThen (α : Type u) (β : Type v) (γ : outParam (Type w)) where
+  /-- The implementation of `a >> b : γ`. `b` is passed as a thunk so it
+  can be forced only when needed. -/
   hAndThen : α → (Unit → β) → γ
 
+/-- The typeclass behind the notation `a &&& b : γ` where `a : α`, `b : β`. -/
 class HAnd (α : Type u) (β : Type v) (γ : outParam (Type w)) where
+  /-- The implementation of `a &&& b : γ`. -/
```
gebner, line 1148: Is this docstring shown when hovering over `a &&& b`?  If so, then I'd change the docstring to something more helpful including the intended meaning of the operation:
```suggestion
  /-- `a &&& b` computes the bitwise and of `a` and `b`. -/
```

Same goes for the other TC ops.

src/Init/Prelude.lean
```diff
@@ -941,50 +942,142 @@ be eagerly evaluated (see `ite`).
...
   | true  => false
   | false => true
 
-/-- The type of natural numbers. `0`, `1`, `2`, ...-/
+/--
+The type of natural numbers, starting at zero. It is defined as an
+inductive type freely generated by "zero is a natural number" and
+"the successor of a natural number is a natural number".
+
+You can prove a theorem `P n` about `n : Nat` by `induction n`, which will
```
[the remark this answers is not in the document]
digama0: The docs here are to some extent supposed to be tutorial-like; the intent is for them to function like Kevin's docs on "how do I make this? how do I use this?". Plus, Nat is pretty intrinsically connected to induction - people will want to know how to do induction on Nat specifically even though it's a general mechanism. The special casing information is later because it is more advanced.

src/Init/Prelude.lean
digama0: It's an overloadable operation, so it might not actually do that. Certainly we should put the "intended meaning" of the operation in the docstring, maybe a couple important instances, but it's ultimately just an overloadable symbol.

gebner: Yes, it is supposed to be bitwise and (and because it's an obscure hieroglyph it makes even more sense to explain that in the docstring).  All of the bitwise ops are triple characters to avoid clashes with `&&`, `||`, and `>>`.

> but it's ultimately just an overloadable symbol

Sure, nothing prevents you from providing weird instances, but the class is called `HAnd` (where the and stands for bitwise and).  And I don't think it's a good idea to overload the bitwise and symbol with other operations just because you can.  It's easy enough to define your own operators after all.

PR #1447, opened 2022-08-07
feat: generalize ReaderT instances

digama0:
There are no behavioral changes here, only generalizing things like "`ReaderT r m` is an applicative when `m` is" to all the defined semimonad types instead of assuming `Monad` for everything.

### DIFF
--- a/src/Init/Control/Lawful.lean
+++ b/src/Init/Control/Lawful.lean
@@ -183,7 +183,7 @@ end ExceptT
 
 namespace ReaderT
 
-theorem ext [Monad m] {x y : ReaderT ρ m α} (h : ∀ ctx, x.run ctx = y.run ctx) : x = y := by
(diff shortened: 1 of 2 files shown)


leodemoura:
bench!

digama0:
I think the bang is on the other side. Can I do that?

!bench

leodemoura:
!bench

leodemoura:
!bench

leodemoura:
@digama0 Did you find out what was triggering the stack overflow in debug mode?

digama0:
No, in the end I ran out of time / patience on debugging and just did black box testing to find the part of the PR that doesn't hit the error. That's also why it scaled back from the original intent (it's not really generalizing the instances anymore, only adding implementations for the missing functions). If we could find the culprit we might be able to do this better later.

leodemoura:
@digama0 Thanks for the explanation. I will try to take a look later at the StackOverflow issue.

Issue #1410, opened 2022-08-02
Lean.Core.Context fileName is mangled on Windows


lovettchris:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

```lean
elab (name := includeStr) "include_str " str:str : term => do
  let str := str.getString
  let ctx ← readThe Lean.Core.Context
  let srcPath := System.FilePath.mk ctx.fileName
  let some srcDir := srcPath.parent | throwError "{srcPath} not in a valid directory"
  let path := srcDir / str
  Lean.mkStrLit <$> IO.FS.readFile path
```

### Steps to Reproduce

1. try and build https://github.com/leanprover/lean4-samples/pull/3 on Windows using `lake build rubiksJs`

**Expected behavior:** 

Build should succeed.

**Actual behavior:** 

```
error: no such file or directory (error code: 2)
error: build failed
```

The problem appears to be that the beginning of the `ctx.fileName` is mangled like this:

`\d%3A\Temp\lean4-samples\RubiksCube\Rubiks.lean`

It should be as follows instead:

`d:\Temp\lean4-samples\RubiksCube\Rubiks.lean`

This could be related to https://github.com/leanprover/lean4/pull/1257

**Reproduces how often:** 100%

### Versions

`Lean (version 4.0.0-nightly-2022-08-01, commit c76fa0681651, Release)`

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue.

Kha:
I think this is `DocumentUri.toPath?` not supporting absolute Windows paths

lovettchris:
Is that where `ctx.fileName` is coming from?  I couldn't find the source code, can you include a pointer to the code so I can try and fix it on my machine?

lovettchris:
Note that https://github.com/leanprover/lean4/pull/1257 fixes the npm build problem, but the Rubiks cube sample still doesn't work due to this path problem, so I see this in the editor:

![image](https://user-images.githubusercontent.com/18707114/182481899-1417de59-8f03-4adb-8f1c-ae855c282a2e.png)

Kha:
> Is that where `ctx.fileName` is coming from? I couldn't find the source code, can you include a pointer to the code so I can try and fix it on my machine?

[...]

PR #1452, opened 2022-08-10
feat: simple uri escaping and unescaping

lovettchris:
```lean
import Std.System.Uri

open System.Uri

/- Uri character escaping includes UTF-8 encoding for the 😵 char! -/
#eval toFileUri "/temp/test.xml?😵=2022" -- "file:///temp/test.xml%3f%f0%9f%98%b5%3d2022"

/- round trip test -/
#eval fromFileUri? (toFileUri "/temp/test.xml?😵=2022") -- some "/temp/test.xml?😵=2022"

/- and to System.FilePath -/
#eval fileUriToPath? (toFileUri "/temp/test.xml?😵=2022") -- some (FilePath.mk "/temp/test.xml?😵=2022")

/- invalid file uri -/
#eval fileUriToPath? "invalid" -- none

/- invalid escape followed by valid escapes -/
#eval unescapeUri "file://test%xx/%3Fa%3D123" -- "file://test%xx/?a=123"

/- tilde is NOT escaped -/
#eval toFileUri "~/git/lean4" -- "file:///~/git/lean4"

/- trailing truncated escape ignored -/
#eval unescapeUri "lean%4" -- "lean%4"
```

Fixes https://github.com/leanprover/lean4/issues/1410

This fixes the RubiksCube demo on Windows:

image

### DIFF
--- a/src/Lean/Compiler/TerminalCases.lean
+++ b/src/Lean/Compiler/TerminalCases.lean
@@ -52,7 +52,11 @@ partial def visitLet (e : Expr) (fvars : Array Expr) : M Expr := do
         let body ← mkLetUsingScope body
         let bodyAbst := body.abstract #[x]
         return bodyAbst
-      let jp ← mkJpDecl (.lam binderName type bodyAbst .default)
+      let jp ← if (← isSimpleLCNF bodyAbst) then
+        -- Join point is too simple, we eagerly inline it.
+        pure <| .lam binderName type bodyAbst .default
+      else
+        mkJpDecl (.lam binderName type bodyAbst .default)
       withReader (fun _ => { jp? := some jp }) do
         visitCases casesInfo value
     else
@@ -71,19 +75,20 @@ partial def visitLet (e : Expr) (fvars : Array Expr) : M Expr := do
         else
           return e
       | some jp =>
-        let .forallE _ d _ _ ← inferType jp | unreachable!
+        let .forallE _ d b _ ← inferType jp | unreachable!
+        let mkJpApp (x : Expr) := mkApp jp x |>.headBeta
         if isLcUnreachable e then
-          mkLcUnreachable d
+          mkLcUnreachable b
         else if compatibleTypes (← inferType e) d then
           let x ← mkAuxLetDecl e
-          return mkApp jp x
+          return mkJpApp x
         else if let some x := isLcCast? e then
-          let e ← mkAuxLetDecl (← mkLcCast x d)
-          return mkApp jp e
+          let x ← mkAuxLetDecl (← mkLcCast x d)
+          return mkJpApp x
         else
           let x ← mkAuxLetDecl e
           let x ← mkAuxLetDecl (← mkLcCast x d)
-          return mkApp jp x
+          return mkJpApp x
 
 end
 
--- a/src/Lean/Compiler/Util.lean
+++ b/src/Lean/Compiler/Util.lean
@@ -110,6 +110,17 @@ def getCtorArity? (declName : Name) : CoreM (Option Nat) := do
   return val.numParams + val.numFields
 
 /--
+Return `true` if the `value` if not a `lambda`, `let` or cases-like expression.
+-/
+def isSimpleLCNF (value : Expr) : CoreM Bool := do
+  if value.isLet || value.isLambda then
+    return false
+  else if let some _ ← isCasesApp? value then
+    return false
+  else
+    return true
+
+/--
 List of types that have builtin runtime support
 -/
 def builtinRuntimeTypes : List Name := [
--- a/tests/lean/run/lcnf1.lean
+++ b/tests/lean/run/lcnf1.lean
@@ -41,8 +41,6 @@ def Vec.head : Vec α (n+1) → α
 
 #eval Compiler.compile #[``Vec.zip]
 
-#eval Compiler.compile #[``Vec.zip.match_1]
-
 #eval Compiler.compile #[``Lean.Elab.Term.reportStuckSyntheticMVar]
 
 #eval Compiler.compile #[``Lean.Elab.Term.synthesizeSyntheticMVars]
@@ -61,3 +59,5 @@ def foo (a b : Nat) :=
   Nat.add d e
 
 #eval Compiler.compile #[``foo]
+
+#eval Compiler.compile #[``Vec.zip.match_1]


Kha:
This should probably be in `Std`, not `Init`. We also try to avoid orphan modules (I believe), so you should create `Std.System` as well and recursively import them up to `Std`.

src/Init/System/Uri.lean
```diff
@@ -0,0 +1,105 @@
...
+  else if 'a' ≤ c ∧ c ≤ 'f' then some (c.toNat - 'a'.toNat + 10)
+  else if 'A' ≤ c ∧ c ≤ 'F' then some (c.toNat - 'A'.toNat + 10)
+  else none
+
+def decodeUri (uri : String) : String := Id.run do
+  let mut decoded : ByteArray := ByteArray.empty
+  let len := uri.data.length
+  let mut pos : Nat := 0
+  while pos < len do
+    let c := uri.data[pos]!
```
Vtec234, line 22: Someone should correct me if there's magic to prevent this but I think this may be quadratic since there's an access of `List Char` at increasing indices. `String.get` and `String.next` to advance the `pos` should be linear.

src/Init/System/Uri.lean
```diff
@@ -0,0 +1,105 @@
...
+           | some hd2 => (hd1 * 16 + hd2).toUInt8
+           | none => none
+         decoded := match hc with
+         | some b => decoded.push b
+         | none => (decoded.push 37) ++ h1.toString.toUTF8 ++ h2.toString.toUTF8
+         pos := pos + 3
+    else
+      decoded := decoded.append c.toString.toUTF8
+      pos := pos + 1
+   return String.fromUTF8Unchecked decoded
```
Vtec234, line 47: Instead of `ByteArray` and `fromUTF8Unchecked`, you could use `String.push` directly.

src/Lean/Server/Utils.lean
```diff
@@ -66,23 +67,11 @@ namespace Lean.Lsp.DocumentUri
 
 /-- Transform the given path to a file:// URI. -/
 def ofPath (fname : System.FilePath) : DocumentUri :=
```
Vtec234, line 69: We can probably just rm these two now.

Kha (commented):
Unhandled test case, I think: `"%%"`

src/Init/System/Uri.lean
lovettchris: String.push takes a Char (a decoded Unicode code point) not a byte, the % escaped values in a Uri are bytes, not Unicode code points.  For example:

```lean
def dump_string (s:String) : IO Unit := do
  for c in s.data do
    IO.println c.toNat

#eval dump_string "😵‍💫"
```
prints 3 Chars:
```
128565
8205
128171
```

Which are then converted to UTF-8 bytes and escaped in a Uri as `%3f%f0%9f%98%b5%e2%80%8d%f0%9f%92%ab%3d`

So to get this back into a string one has to create the 13 byte ByteArray from decoding these % escapes, then call fromUTF8Unchecked to decode that back into the 3 Chars Unicode codepoints.

src/Init/System/Uri.lean
lovettchris: rats string[pos] being a linear search sucks, this isn't optimized somewhere deep inside Lean compiler?  This would mean I have to use `for c in uri` and build a proper parsing state machine to remember if I've seen a '%' and so on... yuk.

lovettchris:
> This should probably be in `Std`, not `Init`. We also try to avoid orphan modules (I believe), so you should create `Std.System` as well and recursively import them up to `Std`.

Why is System.FilePath different and more special than System.Uri - these seem closely related to me... eventually I'd like to see a full Uri class there that does Uri component parsing, resolve_uri, make_relative and all that...

I have no idea what the second part of your sentence means, can explain more about what an orphan module is and share an example of "recursively import them up to `Std`." ?

Kha:
> Why is System.FilePath different and more special than System.Uri

The Lean runtime deals with FilePaths but not with Uris, so that's one profound difference. But would be good to hear from others here.

An orphan module (which I think I made up) is a module not imported by the package root. So `Std` should be importing `Std.System` should import `Std.System.Uri`.

src/Init/System/Uri.lean
Kha: You should use a `String.iterator` (created by `String.mkIterator`), then the code won't fundamentally have to change

Kha: Well, I guess it would have to change even less using Wojciech's suggestion. You just have to make sure to use `utf8ByteSize` instead of `length`.

lovettchris: Ah, I see how to do it now, I switched to String.mkIterator.  Pity the string iterator doesn't have an "eof" property.  If it had that I wouldn't need to track the byte position myself.  Please let me know if I can simplify how I've used the iterator, thanks.

lovettchris:
> An orphan module (which I think I made up) is a module not imported by the package root. So `Std` should be importing `Std.System` should import `Std.System.Uri`.

Ok, I get it now, I moved the code to `Std.System.Uri`.

src/Lean/Server/Utils.lean
Vtec234: I see, it looks like the previous version returns `none` when the argument is not actually a `file://` URI and yours returns the argument in that case. I would prefer to keep the `Option` to signal the error condition rather than silently hide it, what do you think @Kha?

lovettchris:
> Unhandled test case, I think: `"%%"`

I have this test case:

```lean
/- escaped percent -/
#eval unescapeUri "/temp/test%%.xml"  -- "/temp/test%.xml"
```

src/Init/System/Uri.lean
Vtec234: I see, I guess the alternative is redoing UTF-8 parsing in order to know when to call `push`. Not sure if we should do this in order to validate the input, or just trust it (the `Unchecked` call can definitely create invalid UTF-8 given a malformed URI-encoded string).

src/Std/System/Uri.lean
```diff
@@ -0,0 +1,109 @@
...
+  else none
+
+def decodeUri (uri : String) : String := Id.run do
+  let mut decoded : ByteArray := ByteArray.empty
+  let len := uri.utf8ByteSize
+  let mut pos : Nat := 0
+  let mut i := uri.mkIterator
+  while pos < len do
+    let c := i.curr
+    if c == '%' && pos + 2 < len then
```
Vtec234, line 24: I think this check might fail if you try just the string `"%%"`.

src/Init/System/Uri.lean
Vtec234: Does `Iterator.atEnd` work?

src/Std/System/Uri.lean
```diff
@@ -0,0 +1,109 @@
...
+         | none => none
+         | some hd1 =>
+           match hexDigitToNat h2 with
+           | some hd2 => (hd1 * 16 + hd2).toUInt8
+           | none => none
+         decoded := match hc with
+         | some b => decoded.push b
+         | none => (decoded.push 37) ++ h1.toString.toUTF8 ++ h2.toString.toUTF8
+         pos := pos + 3
+         i := i.next.next.next
```
Vtec234, line 46: You could use `Iterator.nextn 3`.

src/Lean/Server/Utils.lean
lovettchris: I can make it return Option string then, not a problem.  I think the future full Uri parser will do things differently, and we can change it then, for example "/tmp/test.lean" is a valid Uri, it just has the "scheme" of "".  Then your code would check the scheme is "file" and proceed from there with the Uri.path component, etc...

lovettchris: After all that I found this code in the LSP to be pretty funny `doc.uri.toPath?.getD doc.uri`

src/Std/System/Uri.lean
lovettchris: Oh, good catch, fixed, thanks.

src/Std/System/Uri.lean
lovettchris: Done, thanks.

src/Lean/Server/Utils.lean
lovettchris: and I did the "rm" operation to remove the redundant methods.

src/Init/System/Uri.lean
lovettchris: Ah, I see, unfortunately atEnd is also quadratic, as it uses utf8ByteSize which is recursively calling `c.utf8Size.toNat` on every char in the string, so it is better to snapshot the utf8ByteSize once outside the loop.

src/Init/System/Uri.lean
lovettchris: Yes, you can encode illegal utf-8 sequences in the %HH sequence and get back random junk, for example:
```
#eval unescapeUri "test%FF%FF"  -- "testAA"
```

src/Lean/Server/Utils.lean
Vtec234: Hah even in the other places if it's `none` we just ignore it rather than flag an error 😅

src/Init/System/Uri.lean
Vtec234: You had me worried, but actually it's [constant time](https://github.com/leanprover/lean4/blob/e9545a426f51cdadf447e4e05a677ae846446bf3/src/Init/Prelude.lean#L2301), so we should use it. To be fair Mario just added the docstring, it was hidden in the `extern` before.

src/Lean/Server/Utils.lean
lovettchris: indeed, but at least there we have a different control flow happening in the none case.

src/Init/System/Uri.lean
lovettchris: Ah, I see it's a call out to lean_string_utf8_byte_size.  Well, I can still snapshot this value in a local variable, doesn't hurt, unless you find `while i.2.byteIdx < len do` to be terribly ugly or something... turns out I'm using "i.2.byteIdx" in a few places so I'm now trying to do this, but it doesn't work...

```
while let pos := i.2.byteIdx; pos < len do
    let c := i.curr
    if c == '%' && pos + 1 < len then ...
```
the let pos in the while doesn't make pos available inside the while do block... rats...

lovettchris: Ok, I switched to this since it looks nicer:
```lean
  while !i.atEnd do
    let pos := i.2.byteIdx
    ...
```

lovettchris: and it seems pretty fast:

```lean
def doIt : IO Unit := do
  let x := fromFileUri? "file://a very long 12000 char Uri string"

def doNTimes (fn : IO Unit) (n : Nat) := do
  let mut i := 0
  while i < n do
    fn
    i := i + 1

#eval timeit "yes" (doNTimes doIt 1000000)   --  431ms
```

I think this means it is about 5 microseconds for the 12k Uri.  Not bad, assuming it is not compiling away the unused variable 'x'.

lovettchris:
@Kha - I'm done messing with this one now, I think.

src/Init/System/Uri.lean
lovettchris: Oh, rats, I checked the main.c file and it was dropping the call to fromFileUri, this version measures the time correctly:

```lean
import Test
import Std.System.Uri

open System.Uri

def doIt (uri: String): IO Nat := do
  let x := fromFileUri? uri
  pure (match x with
  | some y => y.utf8ByteSize
  | none   => 0)

def doNTimes (fn : String → IO Nat) (s: String) (n : Nat) := do
  let mut i := 0
  let mut total := 0
  while i < n do
    let len ← fn s
    total := total + len
    i := i + 1
  pure ()

def main : IO Unit :=
  timeit "Time for 1 thousand calls = " (doNTimes doIt longuri 1000)
```

and it comes out at `Time for 1 thousand calls =  633ms` which is about 0.633ms per call, so this implementation is pretty slow.  I probably should remove the `h2.toString.toUTF8` calls somehow and extract the raw bytes that are already encoded in the string, I can use the byteIdx to figure out how many there are per char....

Kha (changes requested):

src/Std/System/Uri.lean
```diff
@@ -0,0 +1,120 @@
...
+  let len := uri.utf8ByteSize
+  let mut i := uri.mkIterator
+  while !i.atEnd do
+    let pos := i.2.byteIdx
+    let c := i.curr
+    if c == '%' && pos + 1 < len then
+      let h1 : Char := i.next.curr
+      if h1 == '%' then
+        -- this is an escaped '%%' which should become a single percent.
+        decoded := decoded.push 37 -- %
```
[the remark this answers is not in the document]
lovettchris: fixed.

src/Std/System/Uri.lean
```diff
@@ -0,0 +1,120 @@
...
+    else
+      decoded := decoded.append c.toString.toUTF8
+      i := i.next
+   return String.fromUTF8Unchecked decoded
+
+/- https://datatracker.ietf.org/doc/html/rfc3986/#page-12 -/
+def rfc3986ReservedChars : List Char := [ ';', ':', '?', '#', '[', ']', '@', '&', '=', '+', '$', ',', '!', '\'', '(', ')', '*', '%' ]
+
+def uriEscapeAsciiChar (c : Char) : String :=
+  if (rfc3986ReservedChars.contains c) then
```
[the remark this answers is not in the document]
lovettchris: fixed

src/Std/System/Uri.lean
```diff
@@ -0,0 +1,120 @@
...
+    "%" ++ smallCharToHex c
+  else if ((Char.toNat c) < 127) then
+    c.toString
+  else
+    c.toString.toUTF8.foldl (fun s c => s ++ "%" ++ (smallCharToHex (Char.ofNat c.toNat))) ""
```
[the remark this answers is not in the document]
lovettchris: I renamed 'c' to 'b' since it is a UInt8 and there's no toChar on UInt8.

src/Std/System/Uri.lean
```diff
@@ -0,0 +1,120 @@
...
+  smallCharToHex (c : Char) : String :=
+    let n  := Char.toNat c;
+    let d2 := n / 16;
+    let d1 := n % 16;
+    hexDigitRepr d2 ++ hexDigitRepr d1
+end UriEscape
+
+/-- Convert the given file name to a "file:///encodedpath" Uri
+where the encoded path may contain %xx escaping when needed. -/
+def toFileUri (s : String) : String := Id.run do
```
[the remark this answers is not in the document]
lovettchris: The FilePath helpers are handy and are being used by the LSP Utils instead of LSP Utils defining it's own local FilePath version, plus the FilePath case has to be careful to use "normalized" strings in `pathToUri` and if we have `pathToUri` we have to have the counter part `fileUriToPath?`

src/Std/System/Uri.lean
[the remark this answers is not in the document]
Kha: Ahh, that makes it more clear, yes!

src/Std/System/Uri.lean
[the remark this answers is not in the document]
Kha: I should have been clearer: I think we should have only `pathToUri` and `fileUriToPath?`. If a parameter is representing a file path, it should always be a `FilePath` in public APIs.

src/Std/System/Uri.lean
```diff
@@ -0,0 +1,129 @@
...
+  else if 'a' ≤ c ∧ c ≤ 'f' then some (c.toNat - 'a'.toNat + 10)
+  else if 'A' ≤ c ∧ c ≤ 'F' then some (c.toNat - 'A'.toNat + 10)
+  else none
+
+/-- Decode %HH escapings in the given string. Note that sometimes a consecutive
+sequence of multiple escapings can represet a utf-8 encoded sequence for
+a single unicode code point and these will also be decoded correctly. -/
+def decodeUri (uri : String) : String := Id.run do
+  let mut decoded : ByteArray := ByteArray.empty
+  let raw_bytes := uri.toUTF8
```
Kha, line 22:
```suggestion
  let rawBytes := uri.toUTF8
```
and elsewhere

Kha: I'm afraid the mixing of raw byte access and `String.Iterator` has made the code quite messy. If you want to go that route, I'd suggest iterating over `rawBytes` directly, doing a byte-wise copy without involving any `String` functions at all. That should also be even faster.

src/Std/System/Uri.lean
```diff
@@ -0,0 +1,120 @@
...
+def rfc3986ReservedChars : List Char := [ ';', ':', '?', '#', '[', ']', '@', '&', '=', '+', '$', ',', '!', '\'', '(', ')', '*', '%' ]
+
+def uriEscapeAsciiChar (c : Char) : String :=
+  if (rfc3986ReservedChars.contains c) then
+    "%" ++ smallCharToHex c
+  else if ((Char.toNat c) < 127) then
+    c.toString
+  else
+    c.toString.toUTF8.foldl (fun s c => s ++ "%" ++ (smallCharToHex (Char.ofNat c.toNat))) ""
+  where
```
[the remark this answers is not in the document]
lovettchris: fixed

src/Std/System/Uri.lean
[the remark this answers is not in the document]
lovettchris: Ah, I see, done.

src/Std/System/Uri.lean
lovettchris: I guess it is safe to look for '%' since it can never be a valid 2nd, 3rd or 4th byte of a utf-8 encoding sequence according to https://en.wikipedia.org/wiki/UTF-8#Encoding.  But the byte right after '%' should be a hex digit and I think it is also safe that if it is a valid hex digit it cannot be the start of a new utf-8 encoding sequence, and same for the second hex digit... so I'll give it a try...

lovettchris: what is the fastest way to get `bytes[i]` array access to the bytes in a ByteArray, the data is stored in `Array UInt8` and the data in an Array is stored as a List, so doesn't array access on ByteArray result in a nasty quadratic `List.get`... ?

lovettchris: never mind perhaps `#eval ("😊".toUTF8)[2]!` is ok because it uses `uget` which is implemented externally in `lean_byte_array_uget`...

Kha: Correct, the `List` does not actually exist unless you explicitly access it

lovettchris: Ok, @Kha, it made no impact on performance, but I'm also running into what looks like a compiler bug, here's the repro:

```lean

def Zero : UInt8 := '0'.toNat.toUInt8
def Nine : UInt8 := '9'.toNat.toUInt8
def Lettera : UInt8 := 'a'.toNat.toUInt8
def Letterf : UInt8 := 'f'.toNat.toUInt8
def LetterA : UInt8 := 'A'.toNat.toUInt8
def LetterF : UInt8 := 'F'.toNat.toUInt8

/-- Decode %HH escapings in the given string. Note that sometimes a consecutive
sequence of multiple escapings can represet a utf-8 encoded sequence for
a single unicode code point and these will also be decoded correctly. -/
def decodeUri (uri : String) : String := Id.run do
  let mut decoded : ByteArray := ByteArray.empty
  let raw_bytes := uri.toUTF8
  let len := raw_bytes.size
  let mut i := 0
  let percent := '%'.toNat.toUInt8
  while i < len do
    let c := raw_bytes[i]!
    let (result, pos) := if c == percent && i + 1 < len then
      let h1 := raw_bytes[i + 1]!
      if let some hd1 := hexDigitToUInt8? h1 then
        if i + 2 < len then
          let h2 := raw_bytes[i + 2]!
          if let some hd2 := hexDigitToUInt8? h2 then
            -- decode the hex digits into a byte.
            (decoded.push (hd1 * 16 + hd2), i + 3)
          else
            -- not a valid second hex digit so keep the original bytes
            -- BUGBUG not working?
            (decoded.push c ++ decoded.push h1 ++ decoded.push h2, i + 3)
        else
          -- hit end of string, there is no h2.
          -- BUGBUG not working?
          (decoded.push c ++ decoded.push h1, i + 2)
      else
        -- not a valid hex digit so keep the original bytes
        (decoded.push c ++ decoded.push h1, i + 2)
    else
      (decoded.push c, i + 1)
    decoded := result
    i := pos
  return String.fromUTF8Unchecked decoded
where hexDigitToUInt8? (c : UInt8) : Option UInt8 :=
  if Zero ≤ c ∧ c ≤ Nine then some (c - Zero)
  else if Lettera ≤ c ∧ c ≤ Letterf then some (c - Lettera + 10)
  else if LetterA ≤ c ∧ c ≤ LetterF then some (c - LetterA + 10)
  else none

#eval decodeUri "%3a"            -- ":"    correct
#eval decodeUri "%3a%"         -- ":%" correct
#eval decodeUri "%3a%3"       -- ":%:3" weirdly broken, looks like went backwards
#eval decodeUri "%3a%3x"     -- ":%:3:x" weirdly broken
#eval decodeUri "%3a%3a%3a" -- ":::" correct

Is there a bug in my code here?
```

lovettchris: Ok, I switched to this `(((decoded.push c).push h1).push h2, i + 3)` and got a bit of a speed up, so I pushed these fixes.

lovettchris: @Kha - I realize it says unchecked, but how unsafe is it? the following seems to hang my VS code LSP server:

```lean
#eval String.fromUTF8Unchecked (ByteArray.mk #[0xab, 0xea])
```
did it walk into never never land looking for the rest of an incomplete UTF-8 sequence?

lovettchris: @Kha I found a python test library here https://github.com/python/cpython/blob/main/Lib/test/test_urllib.py and copied many of the tests and fixed a bunch of bugs.

lovettchris: I also fixed the snake casing problems.

Kha: > @Kha - I realize it says unchecked, but how unsafe is it? the following seems to hang my VS code LSP server:
> 
> ```lean
> #eval String.fromUTF8Unchecked (ByteArray.mk #[0xab, 0xea])
> ```
> 
> did it walk into never never land looking for the rest of an incomplete UTF-8 sequence?

Oh that's funny, it looks like it overshoots the following condition, trapping itself in an infinite loop: https://github.com/leanprover/lean4/blob/185829d089080b290c012adb380beddfd49a4712/src/Init/Data/String/Basic.lean#L312

src/Std/System/Uri.lean
```diff
@@ -0,0 +1,115 @@
...
+a single unicode code point and these will also be decoded correctly. -/
+def decodeUri (uri : String) : String := Id.run do
+  let mut decoded : ByteArray := ByteArray.empty
+  let rawBytes := uri.toUTF8
+  let len := rawBytes.size
+  let mut i := 0
+  let percent := '%'.toNat.toUInt8
+  while i < len do
+    let c := rawBytes[i]!
+    let (result, pos) := if c == percent && i + 1 < len then
```
Kha, line 30: Why are you using this pair instead of assigning to the mutable variables directly?

src/Std/System/Uri.lean
```diff
@@ -0,0 +1,115 @@
...
+namespace Uri
+namespace UriEscape
+
+/- https://www.ietf.org/rfc/rfc3986.txt -/
+def Zero : UInt8 := '0'.toNat.toUInt8
+def Nine : UInt8 := '9'.toNat.toUInt8
+def Lettera : UInt8 := 'a'.toNat.toUInt8
+def Letterf : UInt8 := 'f'.toNat.toUInt8
+def LetterA : UInt8 := 'A'.toNat.toUInt8
+def LetterF : UInt8 := 'F'.toNat.toUInt8
```
Kha, line 17: Should be lower-case and likely `private`, though I would probably just inline them

Kha (approved):
Last review round (from me)!

Vtec234 (approved):
Looks great! Awesome that it's well-tested now.

src/Std/System/Uri.lean
leodemoura: @lovettchris Could you please inline these functions?

leodemoura:
@lovettchris This PR is almost ready to be merged. Thanks! Could you please squash and resolve conflicts? GitHub is saying it cannot be rebased due to conflicts.

src/Std/System/Uri.lean
lovettchris: I thought imperative code was bad :-) I found it looked a bit cleaner this way and it doesn't seem to change the performance is the compiler somehow optimizing away this pair object?

src/Std/System/Uri.lean
lovettchris: Done.

lovettchris:
> squash

Done, weird my local git didn't think there were conflicts.  I assume we do a squash merge when PR is completed?

src/Std/System/Uri.lean
Kha: Personally I think assigning directly to descriptively-named variables is clearer than using anonymous pairs. But it's fine either way.

lovettchris: I see, then I think you might like the updated version that assigns directly to our existing variable `(decoded, i) := if c == percent && i + 1 < len then`  ...  as for anonymous pairs, I agree anonymous tuples can get out of hand but I think in this case it's pretty clear what's going on.  So do you mind if I leave it this way for now?  Were you expecting a performance hit from the pair, or does that being virtually free make sense to you?

src/Std/System/Uri.lean
lovettchris: should a file a separate bug on this, or do you want me to fix it in this PR?  If the latter, what is the fix exactly?

src/Std/System/Uri.lean
lovettchris: Ah, you said "But it's fine either way." so I'm resolving this conversation.

src/Std/System/Uri.lean
lovettchris: I filed a [new bug](https://github.com/leanprover/lean4/issues/1470) so we don't lose it.

PR #1458, opened 2022-08-11
fix: use saturating casts in `lean_float_to_uint8` to avoid UB

digama0:
The C Standard, 6.3.1.4, paragraph 1 [ISO/IEC 9899:2011], says,

> When a finite value of real floating type is converted to an integer type other than `_Bool`, the fractional part is discarded (i.e., the value is truncated toward zero). If the value of the integral part cannot be represented by the integer type, the behavior is undefined.

This means that the simple implementation `uint8_t f(double d) { return (uint8_t)d; }` is undefined behavior for large values of `d`. The correct version is to check that the value would truncate to a value in the range of the target type, that is, `0 <= d < 2^8`. We have to write out `2^8` as a double constant though since this doesn't appear in any standard headers.

This changes the observable behavior of the cast from wrapping modulo 2^8 to saturating, but I think this behavior is more defensible (and it's also what rust did when [they hit the same issue](https://github.com/rust-lang/rust/issues/10184)).

See also the [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/0.20.2F.200.20.3D.20-nan).


Kha:
This should probably documented in docstrings...? :)

Kha (approved):

digama0:
I assure you, the float functions will all be documented extensively. That will be another PR though.

PR #1460, opened 2022-08-11
feat: add more float functions

### LINKED ISSUE #1457: Feature request: More `Float` primitives.
Please could we have the following primitives available for `Float`. I think they can all be implemented in terms of the other primitives but they would be nice to have explicitly. The particular problem I am trying to solve is writing a JSON encoding for `Float` (ideally without going via a string). I had a go at adding these but I couldn't get the bootstrapping to work.

- `isinf`
- `isnan`
- `isfinite`
- [`frexp`](https://en.cppreference.com/w/cpp/numeric/math/frexp)
- I'm not sure what the appropriate primitive is, but getting the integer mantissa and exponent from a `Float` in a platform-safe way.

digama0:
* [x] Depends on: #1458
* [x] Depends on: #1459

fixes #1457

### DIFF
--- a/src/runtime/object.cpp
+++ b/src/runtime/object.cpp
@@ -32,6 +32,12 @@ Author: Leonardo de Moura
 #ifndef isnan
 #define isnan(x) std::isnan(x)
 #endif
+#ifndef isfinite
+#define isfinite(x) std::isfinite(x)
+#endif
+#ifndef isinf
+#define isinf(x) std::isinf(x)
+#endif
 
 // see `Task.Priority.max`
 #define LEAN_MAX_PRIO 8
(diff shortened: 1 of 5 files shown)


Kha (approved):

src/runtime/object.cpp
```diff
@@ -1467,6 +1472,17 @@ extern "C" LEAN_EXPORT double lean_float_scaleb(double a, b_lean_obj_arg b) {
    }
 }
 
+extern "C" LEAN_EXPORT uint8_t lean_float_isnan(double a) { return (std::isnan)(a); }
+extern "C" LEAN_EXPORT uint8_t lean_float_isfinite(double a) { return (std::isfinite)(a); }
+extern "C" LEAN_EXPORT uint8_t lean_float_isinf(double a) { return (std::isinf)(a); }
+extern "C" LEAN_EXPORT obj_res lean_float_frexp(double a) {
+    object* r = lean_alloc_ctor(0, 2, 0);
+    int exp;
+    lean_ctor_set(r, 0, lean_box_float((std::frexp)(a, &exp)));
```
pcpthm, line 1481: https://en.cppreference.com/w/cpp/numeric/math/frexp#Error_handling

> - If arg is ±∞, it is returned, and an unspecified value is stored in *exp.
> - If arg is NaN, NaN is returned, and an unspecified value is stored in *exp.

Probably it should check if `a` is finite.

PR #1464, opened 2022-08-12
feat: utilities for using scientific notation and floats in json

EdAyers:
- [x] waiting on https://github.com/leanprover/lean4/pull/1460 to include a `JsonNumber.fromFloat`


leodemoura:
@EdAyers We have merged #1460, is this one ready to be merged?

EdAyers:
not yet; got to write a `Float → JsonNumber` function.

EdAyers:
Ok I wrote a `Float → JsonNumber` function but it goes via `toString : Float → String` for now.

PR #1473, opened 2022-08-13
feat: doc comments on notation

digama0:
This makes it possible to write doc comments on `infixl` and `notation` declarations, and in fact makes them required when `linter.missingDocs` is enabled. (The reasoning is that these are just ways of writing a `syntax`, and we want to put documentation on `syntax` since that's what usually shows up in hovers.)


leodemoura:
Merged manually because of update-stage0, and added `update-stage0` after last commit to make sure we restore `stdlib_flags.h`

PR #1476, opened 2022-08-13
Join point checker

hargoniX:
Add an additional check to `Decl.check` that ensures join points are always in tail position and applied fully so they can still be optimized properly later on.

Also some doc-strings as I went through the code.

### DIFF
--- a/src/Lean/Compiler/CompilerM.lean
+++ b/src/Lean/Compiler/CompilerM.lean
@@ -8,7 +8,14 @@ import Lean.Compiler.InferType
 
 namespace Lean.Compiler
 
+/--
+The state managed by the `CompilerM` `Monad`.
+-/
 structure CompilerM.State where
+  /--
+  A `LocalContext` to store local declarations from let binders
+  and other constructs in as we move through `Expr`s.
+  -/
   lctx     : LocalContext := {}
   letFVars : Array Expr := #[]
   /-- Next auxiliary variable suffix -/
@@ -26,24 +33,43 @@ instance : AddMessageContext CompilerM where
 instance : MonadInferType CompilerM where
   inferType e := do InferType.inferType e { lctx := (← get).lctx }
 
+/--
+Add a new local declaration with the given arguments to the `LocalContext` of `CompilerM`.
+Returns the free variable representing the new declaration.
+-/
 def mkLocalDecl (binderName : Name) (type : Expr) (bi := BinderInfo.default) : CompilerM Expr := do
   let fvarId ← mkFreshFVarId
   modify fun s => { s with lctx := s.lctx.mkLocalDecl fvarId binderName type bi }
   return .fvar fvarId
 
+/--
+Add a new let declaration with the given arguments to the `LocalContext` of `CompilerM`.
+Returns the free variable representing the new declaration.
+-/
 def mkLetDecl (binderName : Name) (type : Expr) (value : Expr) (nonDep : Bool) : CompilerM Expr := do
   let fvarId ← mkFreshFVarId
   let x := .fvar fvarId
   modify fun s => { s with lctx := s.lctx.mkLetDecl fvarId binderName type value nonDep, letFVars := s.letFVars.push x }
   return x
 
+/--
+Look for a declaration with the given `FvarId` in the `LocalContext` of `CompilerM`.
+-/
 def findDecl? (fvarId : FVarId) : CompilerM (Option LocalDecl) := do
   let lctx := (← get).lctx
   return lctx.find? fvarId
 
+/--
+Whether a given local declaration is a join point.
+-/
 def _root_.Lean.LocalDecl.isJp (decl : LocalDecl) : Bool :=
   decl.userName.getPrefix == `_jp
 
+/--
+Create a new auxiliary let declaration with value `e` The name of the
+declaration is guaranteed to be unique.
+Returns the free variable representing the new declaration.
+-/
 def mkAuxLetDecl (e : Expr) (prefixName := `_x) : CompilerM Expr := do
   if e.isFVar then
     return e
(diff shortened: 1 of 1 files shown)


src/Lean/Compiler/Decl.lean
```diff
@@ -62,8 +76,34 @@ def toDecl (declName : Name) : CoreM Decl := do
...
+They always need to satisfy two conditions:
+1. Be fully applied
+2. Always be called in tail position
+in order to allow the optimizer to turn them into efficient machine code.
+This function ensures that inside the given declaration both of these
+conditions are satisfied and throws an exception otherwise.
+-/
+def Decl.checkJoinPoints (decl : Decl) : CompilerM Unit := do
+  let tails ← JoinPointChecker.getTails decl.value
+  discard <| tails.mapM JoinPointChecker.checkTail
```
leodemoura, line 92: `tails.forM`?

src/Lean/Compiler/JoinPoints.lean
```diff
@@ -0,0 +1,107 @@
...
+    if decl.userName.getPrefix ==`_jp then
+      throwError s!"Join point {decl.userName} in forbidden position"
+  | .sort .. | .forallE .. | .const .. | .lit .. => return ()
+  | .bvar .. | .mvar .. | .mdata .. => unreachable! 
+
+/--
+Obtain all the tail `Expr`s of `e`. Already checking whether non
+tail values contain a join point and throwing an exception if they do.
+-/
+partial def getTails (e : Expr) :  CompilerM (Array Expr) := do
```
leodemoura, line 45: This function creates a single local context containing all let-declarations and lambda binders `e`. 
This approach does work for the join point checker, but I found it unexpected. I was expecting the functions `getTails` and `checkTail` to be combined into a single `checkJoinPoints` function.

src/Lean/Compiler/JoinPoints.lean
```diff
@@ -0,0 +1,107 @@
...
+    let e := e.consumeMData
+    match e with
+    | .letE .. =>
+      let body ← visitLet e (fun value => do goLetValue value; pure value)
+      go body
+    | .app .. =>
+      if let some casesInfo ←isCasesApp? e then
+        let (motive, discrs, arms) ← visitMatch e casesInfo
+        containsNoJp motive
+        discard <| discrs.mapM (liftM ∘ containsNoJp)
```
leodemoura, line 70: We can use `forM` instead of `mapM` and avoid the `discard`.

src/Lean/Compiler/JoinPoints.lean
```diff
@@ -0,0 +1,107 @@
...
+        discard <| arms.mapM go
+      else
+        let tails ← get
+        set <| tails.push e
```
leodemoura, line 74: We can use `modify fun tails => tails.push e` instead of `get` + `set`

src/Lean/Compiler/Decl.lean
```diff
@@ -62,8 +76,34 @@ def toDecl (declName : Name) : CoreM Decl := do
...
+on them, if it does not throw an exception. This is meant as a sanity
+check for the code generator itself, not to find errors in the user code.
+
+These invariants are:
+- The ones enforced by `Compiler.check`
+- The stored and the inferred LCNF type of the declaration are compatible according to `compatibleTypes`
+-/
 def Decl.check (decl : Decl) (cfg : Check.Config := {}): CoreM Unit := do
   Compiler.check decl.value cfg { lctx := {} }
+  discard <| checkJoinPoints decl |>.run {}
```
leodemoura, line 106: You can use `run'` and avoid the `discard`

src/Lean/Compiler/JoinPoints.lean
```diff
@@ -0,0 +1,107 @@
...
+
+/--
+Checks that a tail is valid, that is either a fully applied join point
+or doesn't contain a join point at all.
+-/
+def checkTail (e : Expr) : CompilerM Unit := do
+  match e with
+  | .app (.fvar fvarId) arg =>
+    let some decl ← findDecl? fvarId | unreachable!
+    if decl.userName.getPrefix == `_jp then
```
leodemoura, line 88: Let's create an auxiliary function `isJpDecl` and hide how we encode join points.
```lean
def LocalDecl.isJoinPoint (decl : LocalDecl) :=
  decl.userName.getPrefix == `_jp
```
I should have already created this function.

leodemoura:
@hargoniX Thanks for sending the PR. I have some minor comments. 
BTW, I will merge commit 6d9a900 now to make sure it does not get out of sync.

src/Lean/Compiler/JoinPoints.lean
```diff
@@ -0,0 +1,89 @@
...
+-/
+partial def checkJoinPoints (e : Expr) :  CompilerM Unit := do
+  let (_, body) ← visitLambda e
+  go body
+where
+  goLetValue (value : Expr) : CompilerM Unit := do
+    let value := value.consumeMData
+    match value with
+    | .lam .. =>
+      let (_, body) ← visitLambda value
```
leodemoura, line 54: We can use `withNewScope do` to make sure the local context is restored after we visit the lambda.
```lean
withNewScope do
  let (_, body) ← visitLambda value
  go body
```
For this checker, this change has very little (or no) impact, but it makes sure our code looks uniform.

src/Lean/Compiler/JoinPoints.lean
```diff
@@ -0,0 +1,89 @@
...
+      go body
+    | _ =>
+      containsNoJp value
+      return ()
+
+  go (e : Expr) : CompilerM Unit := do
+    let e := e.consumeMData
+    match e with
+    | .letE .. =>
+      let body ← visitLet e (fun value => do goLetValue value; pure value)
```
leodemoura, line 64: We can use `withNewScope do` here to before using `visitLet`.

PR #1479, opened 2022-08-14
feat: blocking behavior on EOF for readLine

goens:
As discussed in [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.60lean.20--stdin.60.20does.20not.20respect.20Ctrl.2BD/near/292550058), this implements the blocking behavior seen by Rust. When a stream sees an EOF on a readLine, it will return an empty string once. When called again, it will read the input again, instead of immediately returning an empty string again.

Currently there's no test, as we need to control stdin for that. I have a simple test that works for this, but relies on [expect](https://en.wikipedia.org/wiki/Expect:
```
def main : IO Unit := do
  IO.println "input a line"
  let input <- (← IO.getStdin).getLine
  println! "input length: {input.length}"
  let input <- (← IO.getStdin).getLine
  println! "input length: {input.length}"
  let input <- (← IO.getStdin).getLine
  println! "input length: {input.length}"
```
and 
```
#!/opt/homebrew/bin/expect -f
set force_conservative 0  ;# set to 1 to force conservative mode even if
			  ;# script wasn't run conservatively originally
if {$force_conservative} {
	set send_slow {1 .1}
	proc send {ignore arg} {
		sleep .1
		exp_send -s -- $arg
	}
}

set timeout -1
spawn ./build/bin/test
match_max 100000
expect -exact "input a line\r
"
send -- ""
expect -exact "input length: 0\r
"
send -- "a\r"
expect -exact "a\r
input length: 2\r
"
send -- "a\r"
expect eof
```

 Happy to include it if wanted.

### DIFF
--- a/src/runtime/io.cpp
+++ b/src/runtime/io.cpp
@@ -4,7 +4,6 @@ Released under Apache 2.0 license as described in the file LICENSE.
 
 Authors: Leonardo de Moura, Sebastian Ullrich
 */
-#include <cstdio>
 #if defined(LEAN_WINDOWS)
 #include <windows.h>
 #include <io.h>


src/runtime/io.cpp
```diff
@@ -340,6 +337,8 @@ extern "C" LEAN_EXPORT obj_res lean_io_prim_handle_get_line(b_obj_arg h, obj_arg
             }
             result.append(out);
         } else if (std::feof(fp)) {
+            ungetc(' ',fp); // reset the eof state
+            getc(fp); // remove the character introduced above
```
Kha, line 341: Why not `clearerr`?

goens: simple answer: because I didn't know that one existed!

PR #1482, opened 2022-08-14
feat: add declId hover for syntax/notation/mixfix/elab/macro

digama0:
This makes the `foo` in `syntax (name := foo) ...`  act like a declaration name for the purpose of hovers (so by hovering on it you can see the rendered doc string and the full name of the declaration), and clicking on it goes to references of the syntax.

TODO: do something similar for anonymous syntax?


leodemoura:
@digama0 Could you please resolve the conflicts?

PR #1486, opened 2022-08-15
fix: `decodeScientificLitVal?` parses 1.0e+1 correctly

### LINKED ISSUE #1484: Unusual Float parsing
### Prerequisites

### Description

The parsing of Float literals in Lean is somewhat unusual compared to other popular languages which might lead to unexpected behaviour.

### Steps to Reproduce

```lean
#eval 1.0e   -- 1.000000
#eval 1.0e+1 -- 2.000000
```

**Expected behavior:**
1. I expect `1.0e` to be an error. `e` should always have to be followed by a number.
2. I expect `1.0e+1` to parse as `10.0` just like it does in JavaScript, Java, C++, and Python.

**Actual behavior:** `1.0e` is a valid Float literal and `1.0e+1` gets parsed as `(1.0e) + (1)` which is `2.0`.

### Versions

Lean v4.0.0-m5 (commit 7dbfaf9) on Ubuntu 20.04.

EdAyers:
fixes #1484

- [x] I added a case for handling `+` to `decodeScientificLitVal?`, but I think the tokeniser is parsing infix `+` in`1.0e+1` before this function is called.
- [x] should `#eval 1.0e+ 1` be `2.0` or a syntax error? → it should be an error.

### DIFF
--- a/src/Lean/Parser/Basic.lean
+++ b/src/Lean/Parser/Basic.lean
@@ -910,7 +910,7 @@ where
     let curr  := input.get i
     if curr == 'e' || curr == 'E' then
       let i    := input.next i
-      let i    := if input.get i == '-' then input.next i else i
+      let i    := if input.get i == '-' || input.get i == '+' then input.next i else i
       let curr := input.get i
       if curr.isDigit then
         takeWhileFn (fun c => c.isDigit) c (s.setPos i)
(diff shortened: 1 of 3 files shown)


EdAyers:
I couldn't find the code that causes 1.0e+1 to be tokenised as `(+ (1.0e) (1))`. Any hints?

leodemoura:
@EdAyers Thanks for investigating this issue. The code is here:
https://github.com/leanprover/lean4/blob/master/src/Lean/Parser/Basic.lean#L881

PR #1491, opened 2022-08-17
doc: documentation for Init.SizeOf

digama0:
Also adds some missing SizeOf instances for all the types that were presumably added to `Init.Prelude` after this file was written.

PS: Perhaps this file's contents should be added to `Init.Prelude` instead and placed as high as possible, probably right after `Nat`? That would make it easier to keep this list in sync.


leodemoura:
> PS: Perhaps this file's contents should be added to Init.Prelude instead and placed as high as possible, probably right after Nat? That would make it easier to keep this list in sync.

It might be possible, but it is not worth doing it now.

PR #1497, opened 2022-08-19
feat: add LLVM C API bindings

bollu:
This makes the following changes:
1. It adds a `src/Lean/Compiler/IR/LLVMBindings.lean` with the C API bindings required for the LLVM backend
2. It adds a stub `src/Lean/Compiler/IR/EmitLLVM.cpp`, to import `LLVMBindings` and setup the import structure
3. It adds `src/library/compiler/llvm.cpp` containing the wrappers around the LLVM C API.
4. It modifies the `CMakeLists.txt` to link against LLVM.

This PR is still a draft, since I suspect CI will fail on platforms without LLVM, so I need to get that fixed. Furthermore, the `src/library/compiler/llvm.cpp` needs some love; it's currently a bunch of copy-pasted code with debugging code thrown in.

### DIFF
--- a/src/Lean/Compiler/IR/LLVMBindings.lean
+++ b/src/Lean/Compiler/IR/LLVMBindings.lean
@@ -347,5 +347,4 @@ def constInt64 (ctx : Context) (value : UInt64) (signExtend : Bool := false) : B
 
 def constIntUnsigned (ctx : Context) (value : UInt64) (signExtend : Bool := false) : BaseIO (Value ctx) :=
   constInt' ctx 64 value signExtend
-
 end LLVM


src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -0,0 +1,305 @@
...
+opaque doubleTypeInContext (ctx: @&Ptr Context): IO (Ptr LLVMType)
+
+@[extern "lean_llvm_pointer_type"]
+opaque pointerType (elemty: @&Ptr LLVMType): IO (Ptr LLVMType)
+
+@[extern "lean_llvm_array_type"]
+opaque arrayType (elemty: @&Ptr LLVMType) (nelem: @&UInt64): IO (Ptr LLVMType)
+
+@[extern "lean_llvm_const_array"]
+opaque constArray (elemty: @&Ptr LLVMType) (vals: @&Array (Ptr Value)): IO (Ptr LLVMType)
```
tobiasgrosser, line 100: The casing of `elemTy` is inconsistent.

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -0,0 +1,305 @@
...
+def i8Type (ctx: LLVM.Ptr LLVM.Context): IO (LLVM.Ptr LLVM.LLVMType) :=
+  LLVM.intTypeInContext ctx 8
+
+def i16Type (ctx: LLVM.Ptr LLVM.Context): IO (LLVM.Ptr LLVM.LLVMType) :=
+  LLVM.intTypeInContext ctx 16
+
+def i32Type (ctx: LLVM.Ptr LLVM.Context): IO (LLVM.Ptr LLVM.LLVMType) :=
+  LLVM.intTypeInContext ctx 32
+
+
```
tobiasgrosser, line 269: Unnecessary whitespace change.

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -0,0 +1,305 @@
...
+def i64Type (ctx: LLVM.Ptr LLVM.Context): IO (LLVM.Ptr LLVM.LLVMType) :=
+  LLVM.intTypeInContext ctx 64
+
+def voidPtrType (ctx: LLVM.Ptr LLVM.Context): IO (LLVM.Ptr LLVM.LLVMType) :=
+  do LLVM.pointerType (← LLVM.intTypeInContext ctx 8)
+
+def i8PtrType (ctx: LLVM.Ptr LLVM.Context): IO (LLVM.Ptr LLVM.LLVMType) := voidPtrType ctx
+
+-- TODO: instantiate target triple and find out what size_t is.
+def size_tType (ctx: @&Ptr Context): IO (Ptr LLVMType) := i64Type ctx
```
tobiasgrosser, line 279: Why exactly is this hard-coded as 64 bit? Llvm does not even have a size t type. I am surprised the api offers such magic type interface. I would maybe drop this for now and add it when actually needed. I personally would probably put this type selection further out in the backend or maybe just call it differently. This functionality really is more about the c implementation on a given platform than about llvm.

src/library/compiler/llvm.cpp
```diff
@@ -0,0 +1,1242 @@
...
+#include <llvm-c/TargetMachine.h>
+
+#include "runtime/array_ref.h"
+#include "runtime/string_ref.h"
+
+#define LLVM_DEBUG 0
+
+static void donothing_finalize(void *obj) {
+  (void)obj;
+  // delete static_cast<S*>(obj);
```
tobiasgrosser, line 28: Why is this commented?

src/library/compiler/llvm.cpp
```diff
@@ -0,0 +1,1242 @@
...
+}
+
+static void donothing_foreach(void *, b_lean_obj_arg) {
+  // do nothing since `S` does not contain nested Lean objects
```
tobiasgrosser, line 32: `S` is not defined as a name.

src/library/compiler/llvm.cpp
```diff
@@ -0,0 +1,1242 @@
...
+
+// == Array Type <-> C array of types ==
+
+// TODO: there is redundancy here, but I am loath to clean it up with templates,
+// because we will somehow dispatch to the correct `lean_to_*` function. I can't
+// think of an immediate, clean way to achieve this (plenty of unclean ways. eg:
+// create a templated function that we partial template specialize that tells us
+// what the correct `lean_to_*` is.
+// Concretely: leanToFn<LLVMTypeRef> = lean_to_LLVMType; leanToFn<LLVMValueRef>
+// = lean_to_LLVMValue).
```
tobiasgrosser, line 190: Maybe just leave the redundancy? If it is obvious repetitive in a rarely changing file, maybe that's actually a clean approach.

src/library/compiler/llvm.cpp
```diff
@@ -0,0 +1,1242 @@
...
+      fprintf(stderr, "...%s ; error?: %d \n", __PRETTY_FUNCTION__, is_error);
+    }
+
+    if (LLVM_DEBUG && is_error) {
+      fprintf(stderr, "...%s ; err_msg: %s \n", __PRETTY_FUNCTION__, err_msg);
+    }
+
+    
+    return lean_io_result_mk_ok(lean_box(0));
+
```
tobiasgrosser, line 1241: Too many newlines?

src/library/compiler/llvm.cpp
```diff
@@ -0,0 +1,1242 @@
...
+    }
+    char *err_msg = NULL;
+    char *filepath_c_str = strdup(lean_string_cstr(filepath)); // TODO(bollu): do we need the `strdup`?
+    // LLVMTargetMachineEmitToFile(LLVMTargetMachineRef T, LLVMModuleRef M, char *Filename, LLVMCodeGenFileType codegen, char **ErrorMessage)
+    int is_error = LLVMTargetMachineEmitToFile(lean_to_TargetMachine(target_machine),
+				lean_to_Module(module),
+				filepath_c_str,
+				LLVMCodeGenFileType(codegenType),
+				&err_msg);
+
```
tobiasgrosser, line 1230: Too many newlines?

src/library/compiler/llvm.cpp
```diff
@@ -0,0 +1,1242 @@
...
+  if (LLVM_DEBUG) {
+    fprintf(stderr, "...%s ; error?: %d \n", __PRETTY_FUNCTION__, is_error);
+  }
+
+  assert(!is_error && "failed to link modules");
+  return lean_io_result_mk_ok(Module_to_lean(out_module));
+}
+
+
+
```
tobiasgrosser, line 1117: Why three newlines?

tobiasgrosser (commented):
Nice. I just added a couple of minor comments as I scrolled through.

tobiasgrosser:
I can not find src/Lean/Compiler/IR/EmitLLVM.cpp.

nix/bootstrap.nix
```diff
@@ -1,12 +1,12 @@
...
-  stdenv, lib, cmake, gmp, gnumake, bash, buildLeanPackage, writeShellScriptBin, runCommand, symlinkJoin, lndir, perl, gnused, darwin,
+  stdenv, lib, cmake, gmp, gnumake, bash, buildLeanPackage, llvmPackages_14, writeShellScriptBin, runCommand, symlinkJoin, lndir, perl, gnused, darwin,
   ... } @ args:
 with builtins;
 rec {
   inherit stdenv;
   buildCMake = args: stdenv.mkDerivation ({
     nativeBuildInputs = [ cmake ];
-    buildInputs = [ gmp ];
+    buildInputs = [ gmp llvmPackages_14.llvm llvmPackages_14.libllvm llvmPackages_14.libllvm.dev llvmPackages_14.bintools ];
```
Kha, line 9: It looks like `llvmPackages_14.libllvm` is sufficient, but I can try again when the build is passing

Kha:
There are no auto-built ARM releases for lack of ARM runners, and LLVM cross-compilation being too crazy for me :) . The current releases are built manually by @gebner for Linux and me for macOS.

src/CMakeLists.txt
```diff
@@ -313,6 +376,7 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
   string(APPEND LEAN_EXTRA_LINKER_FLAGS " -ldl")
 endif()
 
+
```
tobiasgrosser, line 379: This change is unnecessary?

src/CMakeLists.txt
```diff
@@ -282,6 +333,18 @@ endif()
...
+# TODO: hide this behind a flag
+# Is there a guarantee that the build environment for `leanprover/lean-llvm` will match that of
+# `leanprover/lean4`? Some of the system-libs as for hardcoded paths. For example:
+# 
+# string(APPEND LEANSHARED_LINKER_FLAGS " ${LLVM_CONFIG_LDFLAGS} ${LLVM_CONFIG_LIBS} ${LLVM_CONFIG_SYSTEM_LIBS}")
+# string(APPEND LEANC_STATIC_LINKER_FLAGS " ${LLVM_CONFIG_LDFLAGS} ${LLVM_CONFIG_LIBS} ${LLVM_CONFIG_SYSTEM_LIBS}")
+# for whatever reason, the system libs on macos ask for `-lm` and `-lz` which are supposedly no longer shipped
+string(APPEND LEANSHARED_LINKER_FLAGS " ${LLVM_CONFIG_LDFLAGS} ${LLVM_CONFIG_LIBS} ${LLVM_CONFIG_SYSTEM_LIBS}")
+set(APPEND LEAN_EXTRA_CXX_FLAGS  " ${LLVM_CONFIG_CXXFLAGS}")
+# string(APPEND LEANC_STATIC_LINKER_FLAGS " ${LLVM_CONFIG_LDFLAGS} ${LLVM_CONFIG_LIBS}")
```
tobiasgrosser, line 346: This is commented out, can we remove it?

src/CMakeLists.txt
```diff
@@ -282,6 +333,18 @@ endif()
...
 string(APPEND LEANSHARED_LINKER_FLAGS " ${LEAN_CXX_STDLIB}")
 
+# Link LLVM for code generation
+# TODO: hide this behind a flag
+# Is there a guarantee that the build environment for `leanprover/lean-llvm` will match that of
+# `leanprover/lean4`? Some of the system-libs as for hardcoded paths. For example:
+# 
+# string(APPEND LEANSHARED_LINKER_FLAGS " ${LLVM_CONFIG_LDFLAGS} ${LLVM_CONFIG_LIBS} ${LLVM_CONFIG_SYSTEM_LIBS}")
+# string(APPEND LEANC_STATIC_LINKER_FLAGS " ${LLVM_CONFIG_LDFLAGS} ${LLVM_CONFIG_LIBS} ${LLVM_CONFIG_SYSTEM_LIBS}")
+# for whatever reason, the system libs on macos ask for `-lm` and `-lz` which are supposedly no longer shipped
```
tobiasgrosser, line 343: This comment is outdated, no?

src/CMakeLists.txt
```diff
@@ -282,6 +333,18 @@ endif()
 string(APPEND LEANC_STATIC_LINKER_FLAGS " ${LEAN_CXX_STDLIB}")
 string(APPEND LEANSHARED_LINKER_FLAGS " ${LEAN_CXX_STDLIB}")
 
+# Link LLVM for code generation
+# TODO: hide this behind a flag
+# Is there a guarantee that the build environment for `leanprover/lean-llvm` will match that of
+# `leanprover/lean4`? Some of the system-libs as for hardcoded paths. For example:
+# 
+# string(APPEND LEANSHARED_LINKER_FLAGS " ${LLVM_CONFIG_LDFLAGS} ${LLVM_CONFIG_LIBS} ${LLVM_CONFIG_SYSTEM_LIBS}")
+# string(APPEND LEANC_STATIC_LINKER_FLAGS " ${LLVM_CONFIG_LDFLAGS} ${LLVM_CONFIG_LIBS} ${LLVM_CONFIG_SYSTEM_LIBS}")
```
tobiasgrosser, line 342: This commented out stuff can be removed, no?

src/CMakeLists.txt
```diff
@@ -263,10 +278,46 @@ find_package(PythonInterp)
...
 # libleancpp/Lean as well as libleanrt/Init are cyclically dependent. This works by default on macOS, which also doesn't like
+
 # the linker flags necessary on other platforms.
+# Furthermore, add LLVM flags into the list of flags we need when linking.
+execute_process(COMMAND ${LLVM_CONFIG}  --ldflags OUTPUT_VARIABLE LLVM_CONFIG_LDFLAGS OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process(COMMAND ${LLVM_CONFIG}  --libs OUTPUT_VARIABLE LLVM_CONFIG_LIBS OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process(COMMAND ${LLVM_CONFIG}  --libdir OUTPUT_VARIABLE LLVM_CONFIG_LIBDIR OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process(COMMAND ${LLVM_CONFIG}  --includedir OUTPUT_VARIABLE LLVM_CONFIG_INCLUDEDIR OUTPUT_STRIP_TRAILING_WHITESPACE)
+execute_process(COMMAND ${LLVM_CONFIG}  --cxxflags OUTPUT_VARIABLE LLVM_CONFIG_CXXFLAGS OUTPUT_STRIP_TRAILING_WHITESPACE)
+# if we build LLVM by source (as @bollu does)<with -DBUILD_SHARED_LIBS=ON, LLVM dynamically loads system libraries
```
tobiasgrosser, line 306: Update?

src/CMakeLists.txt
```diff
@@ -158,9 +160,22 @@ if(AUTO_THREAD_FINALIZATION AND NOT MSVC)
 endif()
 
 if(LLVM)
-  if(NOT (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_C_COMPILER_ID STREQUAL "Clang"))
-    message(FATAL_ERROR "LLVM=ON must be used with clang. Set `CMAKE_CXX_COMPILER` and `CMAKE_C_COMPILER` accordingly.")
-  endif()
+    # if(NOT (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_C_COMPILER_ID STREQUAL "Clang"))
+    # message(FATAL_ERROR "LLVM=ON must be used with clang. Set `CMAKE_CXX_COMPILER=Clang` and `CMAKE_C_COMPILER=Clang` accordingly.")
+    # endif()
```
tobiasgrosser, line 165: Dead code?

src/CMakeLists.txt
```diff
@@ -158,9 +160,22 @@ if(AUTO_THREAD_FINALIZATION AND NOT MSVC)
...
+  message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
+  message(STATUS "Using LLVMConfig.cmake in: ${LLVM_DIR}")
+
+  include_directories(${LLVM_INCLUDE_DIRS})
+  string(APPEND LEAN_EXTRA_CXX_FLAGS " -D LEAN_LLVM")
+else()
+  #message(WARNING "Disabling LLVM support.  JIT compilation will not be available")
```
tobiasgrosser, line 172: Why JIT compilation?

src/CMakeLists.txt
```diff
@@ -158,9 +160,22 @@ if(AUTO_THREAD_FINALIZATION AND NOT MSVC)
...
+endif()
+
+if(LLVM_BACKEND)
+    # if(NOT (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_C_COMPILER_ID STREQUAL "Clang"))
+    #     message(FATAL_ERROR "LLVM_BACKEND=ON must be used with clang. Set `CMAKE_CXX_COMPILER=clang++` and `CMAKE_C_COMPILER=clang` accordingly. Found ${CMAKE_CXX_COMPILER_ID}, ${CMAKE_C_COMPILER_ID}")
+    # endif()
```
tobiasgrosser, line 178: Update?

tobiasgrosser (commented):
There is still some old code in this PR. I guess we need a cleanup pass.

tobiasgrosser:
@bollu, this build went through: https://github.com/tobiasgrosser/lean4/pull/6

tobiasgrosser:
In particular, you need sth like:

```
if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
  find_package(ZLIB REQUIRED)
  message("ZLIB_LIBRARY: ${ZLIB_LIBRARY}")
  cmake_path(GET ZLIB_LIBRARY PARENT_PATH ZLIB_LIBRARY_PARENT_PATH)
  string(APPEND LEAN_EXTRA_LINKER_FLAGS " -L ${ZLIB_LIBRARY_PARENT_PATH}")

  string(APPEND LEANC_STATIC_LINKER_FLAGS " -lleancpp -lInit -lLean -lleanrt ${LLVM_CONFIG_LDFLAGS} ${LLVM_CONFIG_LIBS}")
```

script/prepare-llvm-linux.sh
```diff
@@ -22,6 +22,8 @@ cp -L llvm/bin/clang stage1/bin/
 cp -L llvm/bin/ld.lld stage1/bin/
 # a static archiver!
 cp -L llvm/bin/llvm-ar stage1/bin/
+# a LLVM configurator for PATH & LD_LIBRARY_PATH setup!
+cp -L llvm/bin/llvm-config stage1/bin/
```
Kha, line 26: Why do we need to ship this?

bollu: Agreed, I'll change the script to not ship `llvm-config`.

script/prepare-llvm-linux.sh
```diff
@@ -28,6 +28,8 @@ $CP $ZLIB/lib/libz.so* stage1/lib/
 find stage1 -type f -exec strip --strip-unneeded '{}' \; 2> /dev/null
 # lean.h dependencies
 $CP llvm/lib/clang/*/include/{std*,__std*,limits}.h stage1/include/clang
+# LLVM backend includes
+$CP -r llvm/include/* stage1/include/
```
Kha, line 32: Why do we need to ship this?

Kha: That is true, but I'm not sure if that warrants shipping 21MB of headers (counting only `llvm/`) to everyone!

CMakeLists.txt
```diff
@@ -11,7 +11,8 @@ foreach(var ${vars})
     list(APPEND STAGE0_ARGS "-D${CMAKE_MATCH_1}=${${var}}")
   elseif("${currentHelpString}" MATCHES "No help, variable specified on the command line." OR "${currentHelpString}" STREQUAL "")
     list(APPEND CL_ARGS "-D${var}=${${var}}")
-    if("${var}" STREQUAL "USE_GMP")
+    set(FORWARDED_OPTIONS "USE_GMP" "LLVM_CONFIG" "LLVM")
+    if("${var}" IN_LIST FORWARDED_OPTIONS)
```
Kha, line 15: This should be reverted again now, right? Not having to worry about linking LLVM twice for the cross builds is actually a good reason not to link LLVM in stage 0.

bollu: @Kha when exactly does `FORWARDED_OPTIONS` kick in? Suppose we perform:
```
lean4/$ mkdir build && cd build && cmake ../ (1)
lean4/build$ make -j # (2)
```

Is it correct that:

- The `FORWARDED_OPTIONS` in `CmakeLists.txt` in configured in step `(1)`.
- The `src/CMakeLists.txt`,  runs next, which sets up the `stage1` build. 
- The `stage0/src/CMakelists.txt` runs at step `(2)`, which re-runs its private copy of `CMake`? It is at this step that `FORWARDED_OPTIONS` is used to forward the options from the "outer" `Cmake` in step `(1)` to the "inner" `Cmake` in step (2)?

stage0/src/CMakeLists.txt
```diff
@@ -282,6 +307,15 @@ endif()
 string(APPEND LEANC_STATIC_LINKER_FLAGS " ${LEAN_CXX_STDLIB}")
 string(APPEND LEANSHARED_LINKER_FLAGS " ${LEAN_CXX_STDLIB}")
 
+if(LLVM)
+    string(APPEND LEANSHARED_LINKER_FLAGS " -L${LLVM_CONFIG_LIBDIR} ${LLVM_CONFIG_LDFLAGS} ${LLVM_CONFIG_LIBS} ${LLVM_CONFIG_SYSTEM_LIBS}")
+    string(REPLACE "llvm-host" "llvm" LEANSHARED_LINKER_FLAGS ${LEANSHARED_LINKER_FLAGS})
```
bollu, line 312: Note that this is incredibly janky, as we have the user pass us an `LLVM_CONFIG` path, after which we perform string replacement of `llvm-host` to `llvm`. We do this because:

- `llvm-config` prints paths relative to itself, and we're using the host's `llvm-config`, so `llvm-config` must live inside `llvm-host/`
- We need the paths to point to `llvm-target`, because that's where the correct version of the libs live.

I feel the right solution is to add two cmake flags, `-DLLVM_LINK_FLAGS`. `-DLLVM_CXX_FLAGS`. Then, in `script/prepare-llvm-*.sh`, we invoke `llvm-host/bin/llvm-config` to find the build flags, along with the required `sed` shenanigans to find-and-replace `llvm-host` with `llvm`

bollu: @Kha : does the above strategy sound sensible to you?

CMakeLists.txt
Kha: AFAIK all stages are configured in (2), as they are all set up using `ExternalProject_add`. But it doesn't really matter in practice when that happens.

bollu: I'm trying to build a mental model of the build system's moving parts, and this is very helpful.  Did I get the logical flow of information correct, from `FORWARDED_OPTIONS` in `CMakeLists.txt` to `src/stage0/CMakeLists.txt`?

stage0/src/CMakeLists.txt
Kha: The current solution looks fine to me to be honest. Ultimately I feel it should be `llvm-config` itself that provides a solution for this, but it does not.

CMakeLists.txt
Kha: Yes, that is how `ExternalProject` works

stage0/src/CMakeLists.txt
bollu: @Kha The reason I am a tad uncomfortable with the current strategy is that on the off-chance someone downstream tries to build Lean4 with LLVM support, and they have `/path/to/llvm-config`  look like: `/path/to/llvm-host/path/to/llvm-config`, this path will be silently changed to `/path/to/llvm/path/to/llvm-config` (note the replacement of `llvm-host` -> `llvm`). I can imagine the resulting errors would be very confusing. 

I would suggest, as a happy middle ground, keeping the `string(REPLACE "llvm-host" "llvm" ...)` hack for this patch, and creating separate CMake flags `-DLLVM_{CXX,LINK}_FLAGS` in another patch?

Kha: > they have `/path/to/llvm-config` look like: `/path/to/llvm-host/path/to/llvm-config`

I will gladly blame users for such a weird setup

src/CMakeLists.txt
```diff
@@ -453,6 +465,7 @@ else()
   endif()
 endif()
 
+
```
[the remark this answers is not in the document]
bollu: Fixed.

src/CMakeLists.txt
bollu: Fixed.

src/CMakeLists.txt
bollu: Fixed.

src/CMakeLists.txt
bollu: Fixed.

src/CMakeLists.txt
bollu: Fixed.

src/CMakeLists.txt
bollu: Fixed.

src/CMakeLists.txt
bollu: Fixed.

script/prepare-llvm-linux.sh
bollu: I've removed copying the `include`/ subdirectory in the latest version.

src/library/compiler/llvm.cpp
bollu: Very well. I've removed the `TODO`, but left the comment in-place to point out the redundancy.

src/library/compiler/llvm.cpp
bollu: Resolved by running `clang-format` on the file.

nix/bootstrap.nix
```diff
@@ -114,6 +114,7 @@ rec {
         LEAN_CC=${stdenv.cc}/bin/cc ${lean-bin-tools-unwrapped}/bin/leanc -shared ${lib.optionalString stdenv.isLinux "-Bsymbolic"} \
           ${if stdenv.isDarwin then "-Wl,-force_load,${Init.staticLib}/libInit.a -Wl,-force_load,${Lean.staticLib}/libLean.a -Wl,-force_load,${leancpp}/lib/lean/libleancpp.a ${leancpp}/lib/libleanrt_initial-exec.a -lc++"
             else "-Wl,--whole-archive -lInit -lLean -lleancpp ${leancpp}/lib/libleanrt_initial-exec.a -Wl,--no-whole-archive -lstdc++"} -lm ${stdlibLinkFlags} \
+          $(${llvmPackages_14.libllvm.dev}/bin/llvm-config --ldflags) $(${llvmPackages_14.libllvm.dev}/bin/llvm-config --libs) \
```
Kha, line 117:
```suggestion
          $(${llvmPackages_14.libllvm.dev}/bin/llvm-config --ldflags --libs) \
```
works as well I believe

Kha: Do you still have my Nix commit? I think I avoided the explicit reference to `llvmPackages_14`. Otherwise it looks good.

script/prepare-llvm-macos.sh
```diff
@@ -36,6 +36,7 @@ $CP llvm/lib/clang/*/include/{std*,__std*,limits}.h stage1/include/clang
 cp $SDK/usr/lib/libSystem.tbd stage1/lib/libc
 # use for linking, use system libs for running
 gcp llvm/lib/lib{c++,c++abi,unwind}.dylib stage1/lib/libc
+echo -n " -DLLVM=ON -DCMAKE_CXX_COMPILER_ID=Clang -DCMAKE_C_COMPILER_ID=Clang -DLLVM_CONFIG=$PWD/llvm/bin/llvm-config" # manually point to `llvm-config` location
```
Kha, line 39:
```suggestion
echo -n " -DLLVM=ON -DCMAKE_CXX_COMPILER_ID=Clang -DCMAKE_C_COMPILER_ID=Clang -DLLVM_CONFIG=$PWD/llvm-host/bin/llvm-config" # manually point to `llvm-config` location
```
Should be as in the Linux script, I assume :)

nix/bootstrap.nix
bollu: I do not have a reference: I rebased many times, and appear to have lost it in the confusion.

script/prepare-llvm-mingw.sh
```diff
@@ -17,8 +17,12 @@ cp llvm/bin/clang stage1/bin/
 cp llvm/bin/{ld.lld,lld} stage1/bin/
 # a static archiver!
 cp llvm/bin/llvm-ar stage1/bin/
+# a LLVM configurator for PATH & LD_LIBRARY_PATH setup!
+cp  llvm/bin/llvm-config stage1/bin/
 # dependencies of the above
 cp $(ldd llvm/bin/{clang,lld,llvm-ar}.exe | cut -f3 -d' ' --only-delimited | grep -E 'llvm|clang64') stage1/bin
+# LLVM libraries
+cp -a llvm/lib/lib{clang-cpp,LLVM}*.a stage1/lib/
```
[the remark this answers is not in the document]
bollu: The script no longer copies these over.

src/Lean/Compiler/IR/LLVMBindings.lean
bollu: Fixed, this is removed from `LLVMBindings.lean`.

Kha:
Btw, why did you give up using the LLVM cmake files instead of llvm-config?

bollu:
Two reasons:

1. LLVM's CMake config assumes that the binary/library one is building is configured using CMake's targets. However, Lean's `CMakeLists.txt` does not seem to use targets to build things like `leanc` or `libleanshared`. I was unable to get the library and include commands right. I tried [a fair bit, the history in this branch tells the story](https://github.com/bollu/lean4/pull/12).  

1. This is minor, but I LLVM's CMake config did not give me all LLVM components with `llvm_map_libnames_to_components(all)`. I was forced to specify precisely the LLVM components I needed. It got tedious to figure out the precise collection of LLVM components I cared about by filtering from:
```
$ llvm-config --components
aarch64 aarch64asmparser aarch64codegen aarch64desc aarch64disassembler
aarch64info aarch64utils aggressiveinstcombine all all-targets amdgpu
amdgpuasmparser amdgpucodegen amdgpudesc amdgpudisassembler amdgpuinfo
amdgputargetmca amdgpuutils analysis arm armasmparser armcodegen armdesc
armdisassembler arminfo armutils asmparser asmprinter avr avrasmparser
avrcodegen avrdesc avrdisassembler avrinfo binaryformat bitreader
bitstreamreader bitwriter bpf bpfasmparser bpfcodegen bpfdesc bpfdisassembler
bpfinfo cfguard codegen core coroutines coverage debuginfocodeview
debuginfodwarf debuginfogsym debuginfomsf debuginfopdb demangle dlltooldriver
dwarflinker dwp engine executionengine extensions filecheck frontendopenacc
frontendopenmp fuzzmutate globalisel hexagon hexagonasmparser hexagoncodegen
hexagondesc hexagondisassembler hexagoninfo instcombine instrumentation
interfacestub interpreter ipo irreader jitlink lanai lanaiasmparser
lanaicodegen lanaidesc lanaidisassembler lanaiinfo libdriver lineeditor linker
lto mc mca mcdisassembler mcjit mcparser mips mipsasmparser mipscodegen
mipsdesc mipsdisassembler mipsinfo mirparser msp430 msp430asmparser
msp430codegen msp430desc msp430disassembler msp430info native nativecodegen
nvptx nvptxcodegen nvptxdesc nvptxinfo objcarcopts object objectyaml option
orcjit orcshared orctargetprocess passes perfjitevents powerpc powerpcasmparser
powerpccodegen powerpcdesc powerpcdisassembler powerpcinfo profiledata remarks
riscv riscvasmparser riscvcodegen riscvdesc riscvdisassembler riscvinfo
runtimedyld scalaropts selectiondag sparc sparcasmparser sparccodegen sparcdesc
sparcdisassembler sparcinfo support symbolize systemz systemzasmparser
systemzcodegen systemzdesc systemzdisassembler systemzinfo tablegen target
textapi transformutils ve veasmparser vecodegen vectorize vedesc vedisassembler
veinfo webassembly webassemblyasmparser webassemblycodegen webassemblydesc
webassemblydisassembler webassemblyinfo webassemblyutils windowsmanifest x86
x86asmparser x86codegen x86desc x86disassembler x86info x86targetmca xcore
xcorecodegen xcoredesc xcoredisassembler xcoreinfo xray
```

bollu:
TL;DR: I knew I could make `llvm-config` work, and using LLVM's `CMake` config seemed to require invasive changes to Lean's `CMake` config, which I was reluctant to perform.

nix/bootstrap.nix
Kha: I found it

nix/bootstrap.nix
```diff
@@ -1,12 +1,12 @@
 { debug ? false, stage0debug ? false, extraCMakeFlags ? [],
-  stdenv, lib, cmake, gmp, gnumake, bash, buildLeanPackage, writeShellScriptBin, runCommand, symlinkJoin, lndir, perl, gnused, darwin, llvmPackages_14,
+  stdenv, lib, cmake, gmp, gnumake, bash, buildLeanPackage, writeShellScriptBin, runCommand, symlinkJoin, lndir, perl, gnused, darwin, llvmPackages,
```
bollu, line 2: Where is the version of LLVM setup? Does it pick it up from:

```
# nix/packages.nix
llvmPackages = if stdenv.isDarwin then llvmPackages_11 else llvmPackages_14;

```

[`nix/packages.nix:8](https://github.com/leanprover/lean4/blob/3134e5c55d714151927fd16a6c7e61fe8dab8d63/nix/packages.nix#L8?)

Kha: yes

stage0/src/CMakeLists.txt
```diff
@@ -313,6 +347,7 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
   string(APPEND LEAN_EXTRA_LINKER_FLAGS " -ldl")
 endif()
 
+
 if(NOT(${CMAKE_SYSTEM_NAME} MATCHES "Windows"))
```
tobiasgrosser, line 351: There seems to be an unrelated whitespace change.

bollu:
@Kha would you like me to run a full `update-stage0`, since I do change the `src/stage0/CMakeLists.txt`? 

Running `update-stage0` would make the diff much larger. Perhaps we should update stage0 as a chore commit after comitting this one?

Kha:
> since I do change the `src/stage0/CMakeLists.txt`

Why do we need that?

stage0/src/CMakeLists.txt
```diff
@@ -35,7 +35,6 @@ option(CCACHE             "use ccache"         ON)
 option(SPLIT_STACK        "SPLIT_STACK"        OFF)
 # When OFF we disable LLVM support
 option(LLVM               "LLVM"               OFF)
-
 # When ON we include githash in the version string
```
tobiasgrosser, line 38: Another unrelated whitespace change.

bollu:
@Kha  I think you're right, now that we do not forward `-DLLVM, -DLLVM_CONFIG` to `stage0`, we don't need a bumped up `stage0/src/CmakeLists.txt` 

I'm still unclear on the CMake option forwarding semantics. Is `-DLLVM=ON` when configuring `cmake` passed along to `stage0`? If so, doesn't the stale `stage0/src/CMakeLists.txt` do something _very different_ with the option `-DLLVM=ON`? Thus, I would prefer to be lazy and bump up `stage0` rather than having to carefully think through the interactions between the old and new build scripts. 

However, AFAIU, since we no longer forward options to `stage0`, the point is moot.

stage0/src/CMakeLists.txt
bollu: Should be resolved, post `stage0` revert.

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -0,0 +1,301 @@
...
+
+structure BasicBlock where
+structure Context where
+structure Module where
+structure Builder where
+structure LLVMType where
+structure Value where
+structure MemoryBuffer where
+structure Target where
+structure TargetMachine where
```
Kha, line 47: If these types exist merely for building type-safe interfaces, they should probably be opaque as well

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -0,0 +1,301 @@
...
+
+-- A raw pointer to a C object, whose Lean representation
+-- is given by α
+opaque Ptr (α: Type): Type := Unit
```
Kha, line 51:
```suggestion
opaque Ptr (α : Type) : Type
```
Any reason for this approach instead of using `Context` etc. directly?

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -0,0 +1,301 @@
...
+opaque setInitializer (glbl: @&Ptr Value) (val: @&Ptr Value): IO Unit
+
+@[extern "lean_llvm_function_type"]
+opaque functionType (retty: @&Ptr LLVMType) (args: @&Array (Ptr LLVMType)) (isVarArg: Bool := false): IO (Ptr LLVMType)
+
+@[extern "lean_llvm_void_type_in_context"]
+opaque voidType (ctx: @&Ptr Context): IO (Ptr LLVMType)
+
+@[extern "lean_llvm_int_type_in_context"]
+opaque intTypeInContext (ctx: @&Ptr Context) (width: UInt64): IO (Ptr LLVMType)
```
Kha, line 90: Could functions like this be considered pure? I assume that would simplify the code a little. Could also be a function in namespace `Context`.

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -0,0 +1,301 @@
...
+def i64Type (ctx: LLVM.Ptr LLVM.Context): IO (LLVM.Ptr LLVM.LLVMType) :=
+  LLVM.intTypeInContext ctx 64
+
+def voidPtrType (ctx: LLVM.Ptr LLVM.Context): IO (LLVM.Ptr LLVM.LLVMType) :=
+  do LLVM.pointerType (← LLVM.intTypeInContext ctx 8)
+
+def i8PtrType (ctx: LLVM.Ptr LLVM.Context): IO (LLVM.Ptr LLVM.LLVMType) := 
+  voidPtrType ctx
+
+def True (ctx: Ptr Context): IO (Ptr Value) :=
```
Kha, line 277: Why uppercase?

src/library/compiler/llvm.cpp
```diff
@@ -0,0 +1,1612 @@
...
+// think of an immediate, clean way to achieve this (plenty of unclean ways. eg:
+// create a templated function that we partial template specialize that tells us
+// what the correct `lean_to_*` is.
+// Concretely:
+//  leanToFn<LLVMTypeRef> = lean_to_LLVMType;
+//  leanToFn<LLVMValueRef> = lean_to_LLVMValue).
+LLVMTypeRef *array_ref_to_ArrayLLVMType(
+    const lean::array_ref<lean_object *> &arr) {
+#ifndef LEAN_LLVM
+    assert(false && ("Please build a version of Lean4 with -DLLVM=ON to invoke "
```
Kha, line 204: Should be `always_assert`

src/library/compiler/llvm.cpp
```diff
@@ -0,0 +1,1612 @@
...
+                     "the LLVM backend function."));
+#else
+    LLVMModuleRef mod = LLVMModuleCreateWithNameInContext(lean_string_cstr(str),
+                                                          lean_to_Context(ctx));
+    if (LLVM_DEBUG) {
+        fprintf(stderr, "%s ; mod: %p\n", __PRETTY_FUNCTION__, mod);
+    }
+    // LLVMWriteBitcodeToFile(mod, "/home/bollu/temp/module.bc");
```
Kha, line 271: stray comment

Kha (changes requested):
I'm not sure I understand the allocation management on LLVM's side yet. Is the idea that the context takes care of everything and we just use raw pointers (`*Ref`s) into it (where is the context freed btw)? If yes, shouldn't we use raw `USize`s on the Lean side as well for them instead of wrapping them into actual Lean objects?

bollu:
@Kha I confess, I don't know the FFI well. If I understand the suggestion correctly, since we use raw pointers, we ought to cast the pointers to `usize_t` when crossing the FFI from `C -> Lean`, and cast back to the correct pointer type when crossing back from `Lean -> C`? 

How do we encapsulate the fact that a type like `Context` is al alias for `USize`? Does `opaque Context := USize` do the trick?

I imagine that will work, and I'm happy to refactor it to be thus.

Kha:
> we ought to cast the pointers to `usize_t` when crossing the FFI from `C -> Lean`, and cast back to the correct pointer type when crossing back from `Lean -> C`?

Yes, that should do it.



> Does `opaque Context := USize` do the trick?

No, that doesn't work currently.

```lean
opaque Context : Type := USize
def f : Context → Context := id
```
You can either use `abbrev Context := USize` or
```
structure Context where
  rawPtr : USize
```
They both have the same ABI representation.

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -0,0 +1,301 @@
...
+end CodeGenFileType
+
+namespace IntPredicate
+-- https://github.com/llvm/llvm-project/blob/c3e073bcbdc523b0f758d44a89a6333e38bff863/llvm/include/llvm-c/Core.h#L290
+def EQ : UInt64 := 32
+def NE : UInt64 := EQ + 1
+def UGT : UInt64 := NE + 1
+end IntPredicate
+
+inductive Ty where
```
Kha, line 29: Where is this type used?

Kha:
This doesn't directly help with the arrays as we don't currently have unboxed `USize` arrays. It looks like these arrays are relatively small, but if you are concerned about the additional allocations, you should probably introduce a new abstract Lean type for them, though you would need to re-expose the necessary `Array` functions (see `ByteArray`).

bollu:
@Kha I've updated the API to use `(Ptr α)` as a synonym for `USize`, and kept the opaque types (`BasicBlock`, `Value`, etc) around to index the `α` .

Kha:
Ok. As I said above I'm not sure about the value added by `Ptr`, but it seems fine now.

src/library/compiler/llvm.cpp
```diff
@@ -13,11 +13,11 @@ Lean's IR.
 
 #include <cassert>
 
-#include "runtime/debug.h"
 #include "runtime/array_ref.h"
+#include "runtime/debug.h"
 #include "runtime/string_ref.h"
 
-#define LLVM_DEBUG 0
+#define LLVM_DEBUG 1
```
tobiasgrosser, line 20: Is this change deliberate?

bollu:
@Kha , I've fixed the API, and the warts left over in `CMakeLists.txt`.

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -0,0 +1,292 @@
...
+opaque LLVMType : Type
+opaque Value : Type
+opaque MemoryBuffer : Type
+opaque Target : Type
+opaque TargetMachine : Type
+
+-- A raw pointer to a C object, whose Lean representation
+-- is given by α
+structure Ptr (α: Type): Type where
+  ptr: USize
```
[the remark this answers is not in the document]
Kha: The current design is a result of https://github.com/leanprover/lean4/pull/1497#pullrequestreview-1168624977, i.e. you should think of these objects as mere indices into the `Context`, apparently. But we should at least make constructors of `ptr` structures private, right.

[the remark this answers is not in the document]
Kha: If we wanted to make the API actually safe, perhaps we should parameterize all these types by their `Context` instance to make sure that they cannot outlive it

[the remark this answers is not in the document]
Kha: I already forgot about dot notation again, that would be nice too, yes.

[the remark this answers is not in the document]
Kha: So to summarize, how about
```
structure BasicBlock (ctx : Context) : Type where
  private mk ::
  ptr : USize
```
?

bollu:
Done once again, I've made changes as requested by @Kha , @gebner .

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -11,282 +14,303 @@ are to:
...
+structure IntPredicate where
+  private mk :: val : UInt64
+
+def IntPredicate.EQ : IntPredicate := { val := 32 }
+def IntPredicate.NE : IntPredicate := { val := IntPredicate.EQ.val + 1 }
+def IntPredicate.UGT : IntPredicate := { val := IntPredicate.NE.val + 1 }
+
+structure BasicBlock (ctx : Context)  where
+  private mk :: ptr : USize
+  deriving Inhabited
```
Kha, line 34: Hmm, using this instance is unsafe I suppose, but I don't have a better solution except for transforming affected `opaque`s into `axiom`s

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -11,282 +14,303 @@ are to:
...
-opaque printModuletoString (mod: @&Ptr Module): IO (String)
+opaque printModuletoString (mod : @&Module ctx) : BaseIO (String)
 
 @[extern "lean_llvm_print_module_to_file"]
-opaque printModuletoFile (mod: @&Ptr Module) (file: @&String): IO Unit
+opaque printModuletoFile (mod : @&Module ctx) (file : @&String) : BaseIO Unit
 
 @[extern "llvm_count_params"]
-opaque countParams (fn: @&Ptr Function): UInt64 -- will this cause problems..?
+opaque countParams (fn : @&Value ctx) : UInt64 -- will this cause problems..?
```
Kha, line 232: What is this comment referring to?

Kha (approved):

src/Lean/Compiler/IR/LLVMBindings.lean
gebner: You can turn it into a `Nonempty (BasicBlock ctx)` instance, which is safe.

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -0,0 +1,316 @@
...
+opaque createModule (ctx : @&Context) (name : @&String) : BaseIO (Module ctx)
+
+@[extern "lean_llvm_module_to_string"]
+opaque moduleToString (m : @&Module ctx) : BaseIO String
+
+@[extern "lean_llvm_write_bitcode_to_file"]
+opaque writeBitcodeToFile (m : @&Module ctx) (path : @&String) : BaseIO Unit
+
+@[extern "lean_llvm_add_function"]
+opaque addFunction (m : @&Module ctx) (name : @&String) (type : @&LLVMType ctx) : BaseIO (Value ctx)
```
gebner, line 81:
```suggestion
opaque addFunction (m : Module ctx) (name : @& String) (type : LLVMType ctx) : BaseIO (Value ctx)
```
All of the borrow annotations here are moot btw, since `Module` etc. are all value types (`USize`).

src/library/compiler/llvm.cpp
```diff
@@ -0,0 +1,1639 @@
...
+        fprintf(stderr, "...%s ; ix: %lu\n", __PRETTY_FUNCTION__, ix);
+    }
+    LLVMValueRef out = LLVMGetParam(lean_to_Value(f), ix);
+    fprintf(stderr, "%s ; out: %s\n", __PRETTY_FUNCTION__,
+            LLVMPrintValueToString(out));
+    return lean_io_result_mk_ok(lean_box_usize(Value_to_lean(out)));
+#endif  // LEAN_LLVM
+}
+
+extern "C" LEAN_EXPORT uint64_t llvm_count_params(size_t f,
```
gebner, line 1410:
```suggestion
extern "C" LEAN_EXPORT uint64_t llvm_count_params(size_t /* ctx */, size_t f,
```
I'm pretty sure the implicit argument is passed as an argument to the C function.

You can check the C function signature the compiler expects by reading the `LLVMBindings.c` file.

bollu:
@Kha , if we could merge this, I can make progress on https://github.com/leanprover/lean4/pull/1837, since there's code there I'd like to split across PRs, but the diffs would be too big without this one. I'd like to:

- Split up the code that changes building `libleanrt` into a separate PR
-

Kha:
Let's merge then!

PR #1500, opened 2022-08-22
chore: remove unsafe withPtrEq

gebner:
A couple of years ago, there was the idea floating around of adding a safe wrapper around `ptrEq`.  There are many possible variants of such a wrapper, maybe the most elegant one is the following:
```lean
@[implementedBy ptrEq]
opaque safePtrEq {α : Sort u} (a b : α) : Squash { e : Bool // e → a = b }
```
Unfortunately, I just realized that all of these variants are unsafe, including the `withPtrEq` version we currently have in core.  And there is no obvious or easy fix.

The problem is that it is very easy to construct provably different Lean terms that evaluate to identical VM objects.  In hindsight, this is an obvious effect of erasure and common subexpression elimination.  The following are some concrete examples:
```lean
@[simp] theorem Bool_ne_Empty : Bool ≠ Empty :=
  (Empty.rec <| · ▸ true)

-- 1) Types are erased, so ptrEq will always claim that they are equal:
def problem1 := withPtrEq Bool Empty (fun _ => false) (by simp)
#reduce problem1 -- false
#eval problem1 -- true

-- 2) This works even if we're not directly comparing types, but also when
-- they're nested in a heap value (due to CSE).
def problem2 := withPtrEq (Bool, 1) (Empty, 1) (fun _ => false) (by simp)
#reduce problem2 -- false
#eval problem2 -- true
```

Potential fixes:
1. Cripple the compiler to never do CSE after erasure, and have `withPtrEq` return false if one of the arguments is `box(0)`.  (Only mentioned for completeness, I don't think this is a good idea.)
2. Add a type class `class InjectiveVMRepr (α : Sort u) : Prop where unsafe mk ::`, auto-generate instances for inductives and structures that are safe to compare with `ptrEq`, and add an additional `[InjectiveVMRepr α]` argument to `withPtrEq`.

I propose to remove `withPtrEq` in the meantime until we have decided on a solution.


gebner:
> 2\. Add a type class `class InjectiveVMRepr (α : Sort u) : Prop where unsafe mk ::`,

Unfortunately using `unsafe` for "tag" classes doesn't work either.  That's because you need to prove `Nonempty (InjectiveVMRepr Foo)` to add a safe declaration of type `InjectiveVMRepr Foo`, but we already have `Nonempty (InjectiveVMRepr Foo) → InjectiveVMRepr Foo` in safe code.  The best we could do is `class InjectiveVMRepr (α : Sort u) : Prop where private mk ::`, but private is trivial to circumvent with `⟨⟩` or `by constructor`.

digama0:
One option is to define `InjectiveVMRepr` directly as
```
class SafePtrEq (α) where
  safePtrEq (a b : α) : Squash { e : Bool // e → a = b }
```
where every implementation is just `safePtrEq := unsafe ptrEq`. There isn't any way to fake that without using `unsafe`, since it's not just a tag but actually the function itself.

gebner:
The `SafePtrEq` class is unfortunately not strong enough to deal with polymorphic types.  I'd like to be able to have instances like `[InjectiveVMRepr α] → InjectiveVMRepr (List α)`.  (`[SafePtrEq α] → SafePtrEq (List α)` is obviously unsafe.)

leodemoura:
@gebner Thanks for reporting the problem. Could you please create an issue for it? I want to fix this issue, but I don't have time now, and I will forget this issue exists if we don't have a proper issue for it.

gebner:
#1502

leodemoura:
I will leave it broken for now. We are working on the new code generator, and we have a plan to fix this issue instead of removing the feature. See discussion #1502.
I am marking this PR with the "depends on the new code generator". If our plan fails, I will merge it.

PR #1503, opened 2022-08-23
Document, fix EOF behavior

Kha:
#1479 broke `isEof` in combination with `getLine`, so merely deprecating would be too weak IMO

### DIFF
--- a/src/Init/System/IO.lean
+++ b/src/Init/System/IO.lean
@@ -221,8 +221,18 @@ opaque FS.Handle : Type := Unit
   capture output of `#eval` commands into memory. -/
 structure FS.Stream where
   flush   : IO Unit
+  /--
+Read up to the given number of bytes from the stream.
+If the returned array is empty, an end-of-file marker has been reached.
+Note that EOF does not actually close a stream, so further reads may block and return more data.
+  -/
   read    : USize → IO ByteArray
   write   : ByteArray → IO Unit
+  /--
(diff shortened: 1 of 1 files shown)


Kha:
Note that this effectively moves us closer to `read` file descriptor behavior and away from `fread` C stream behavior. The only remaining behavior inherited from C streams visible to Lean code I think is buffering, as evidenced by the existence of `Handle.flush`. I believe most languages other than C have their own stream buffering implementation, so we might eventually want to completely forgo C streams in favor of the respective unbuffered platform primitive.

src/Init/System/IO.lean
```diff
@@ -273,9 +273,19 @@ Once `h.isEof` is true, reading `h` will always return an empty array.
...
+Note that EOF does not actually close a handle, so further reads may block and return more data.
+-/
+@[extern "lean_io_prim_handle_read"] opaque read (h : @& Handle) (bytes : USize) : IO ByteArray
 @[extern "lean_io_prim_handle_write"] opaque write (h : @& Handle) (buffer : @& ByteArray) : IO Unit
 
+/--
+Read text up to (including) the next line break from the handle.
+If the returned string is empty, an end-of-file marker has been reached.
+Note that EOF does not actually close a handle, so further reads may block and return more data.
+-/
```
gebner, line 281: I think we should put the same comment on `Stream.getLine` as well.

PR #1504, opened 2022-08-23
feat: generalise forEachExpr

EdAyers:
Lean.Meta.forEachExpr should be general over monads rather than restricted to the MetaM monad.
This is similar to the generalisation of `Lean.Meta.transform`.

- [x] Please can someone double check that I have implemented the MonadCacheT stuff correctly? I don't completely understand what it is doing.

### DIFF
--- a/src/Lean/Meta/ForEachExpr.lean
+++ b/src/Lean/Meta/ForEachExpr.lean
@@ -11,6 +11,41 @@ namespace Lean.Meta
 
 variable {m} [Monad m] [MonadLiftT MetaM m] [MonadControlT MetaM m]
 
+/-- Given an expression `e = fun (x₁ : α₁) .. (xₙ : αₙ) => b`, runs `f` on each `αᵢ` and `b`. -/
+def visitLambda (f : Expr → m Unit) (e : Expr) : m Unit := visit #[] e
+  where visit (fvars : Array Expr) : Expr → m Unit
+    | Expr.lam n d b c => do
+      let d := d.instantiateRev fvars
+      f d
+      withLocalDecl n c d fun x =>
+        visit (fvars.push x) b
+    | e => do
+      f <| e.instantiateRev fvars
+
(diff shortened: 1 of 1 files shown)


src/Lean/Meta/ForEachExpr.lean
```diff
@@ -8,51 +8,47 @@ import Lean.Util.MonadCache
...
+/-- Similar to `Expr.forEach'`, but creates free variables whenever going inside of a binder.
+If the inner function returns `false`, deeper subexpressions will not be visited.
+ -/
+partial def forEachExpr' {m} [Monad m] [MonadLiftT MetaM m] [MonadControlT MetaM m]
+  (input : Expr)
+  (fn : Expr → m Bool)
+  : m Unit := do
+  let _ : STWorld IO.RealWorld m := ⟨⟩
+  let _ : MonadLiftT (ST IO.RealWorld) m := { monadLift := fun x => liftM (m := MetaM) (liftM (m := ST IO.RealWorld) x) }
+  let rec visit (e : Expr) : MonadCacheT Expr Unit m Unit :=
```
EdAyers, line 21: In Transform.lean this is `MonadCacheT ExprStructEq Expr m Expr`. Should I be using ExprStructEq?

leodemoura:
!bench

EdAyers:
I extracted `visitLet`, `visitLambda`, `visitForall`. Please could you run another benchmark?

EdAyers:
!bench

leodemoura:
!bench

src/Lean/Meta/ForEachExpr.lean
leodemoura: Good question. To be on the safe side it is better to use `ExprStructEq`. 
The difference is that `ExprStructEq` takes binder names and annotations into account. 
That is, `forEachExpr'` may skip `fun {a} => a` because `fun a => a` was already visited. 
I didn't check all uses of `forEachExp'`, but even if this distinction is irrelevant in every single use in this code base, external users may not expect it. What do you think?

leodemoura: I am going to merge as-is. We can make change this in another PR.

PR #1505, opened 2022-08-24
doc: add documentation on monads

lovettchris:
Following the structure of https://mmhaskell.com/monads

You can see the result here: https://lovettsoftware.com/lean4/monads/intro.html


Kha:
If you want to iteratively extend an example, it sounds like you'll want to make it a literate LeanInk file like in examples/ instead of further extending mdBook.

lovettchris:
Hmm, yes, I can see the overlap, but that's a lot more for me to learn right now, and I don't really like the "markdown editing" experience of something like `palindromes.lean`.  I don't need the full power of Lean for these chapters, just `# ++ ` accumulation of snippets... I can see why it's handy for widgets.lean.

Unless you are saying the entire reference manual is moving to leanink and I probably should bite the bullet... but it wasn't what I thought I was signing up for when I started these chapters...

lovettchris:
@leodemoura this has received a good review from Henrik (see [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/monads.20documentation)) and I have implemented all his suggestions.

One thing I wasn't sure about in the original Haskell language is the use of the term "structures", I find this a bit confusing in Lean because lean has a special "structure" type. Is there something I should use instead? "type classes" perhaps?

image

doc/monads/monads.lean
```diff
@@ -0,0 +1,307 @@
...
+  | none,   _ => none
+  | some a, f => f a
+```
+
+> **Side note**: this function definition is using a special shorthand syntax in Lean where the `:=
+match a, b with` code can be collapsed away. To make this more clear consider the following simpler
+example, where `Option.bind` is using the second form like `bar`:
+
+-/
+def foo (x : Option Nat) (y: Nat) : Option Nat :=
```
leodemoura, line 88: Please add space between `y` and `:`. There are many other instances of this issue.

doc/monads/monads.lean
```diff
@@ -0,0 +1,307 @@
...
+But let's bring in the definition of a monad. What does it mean to describe `Option` as a
+computational context?
+
+The `Option` monad encapsulates the context of failure. Essentially, the `Option` monad lets us
+abort a series of operations whenever one of them fails. This allows future operations to assume
+that all previous operations have succeeded. Here's some code to motivate this idea:
+
+-/
+def optionFunc1 : String -> Option Nat
+| "" => none
```
leodemoura, line 116: Please indent the `|`s.

doc/monads/functors.lean
```diff
@@ -0,0 +1,215 @@
...
+  { totalSize := 1800, numBedrooms := 4, masterBedroomSize := 500,
+    livingRoomSize := 900, kitchenSize := 400 }
+/-!
+
+Now, suppose you want anyone to be able to map a `LivingSpace` from one type of measurement unit to
+another.  Then you would provide a `Functor` instance as follows:
+
+-/
+def LivingSpace.map (f : α → β) (s : LivingSpace α) : LivingSpace β :=
+  { totalSize := f s.totalSize,
```
leodemoura, line 133: We can avoid the `,`s.

doc/monads/laws.lean
```diff
@@ -0,0 +1,319 @@
...
+  deriving Repr, BEq
+
+def Point.map (f : α → β) (s : Point α) : Point β :=
+  { x := f s.y,  -- an example of something weird
+    y := f s.x }
+
+instance : Functor Point where
+  map := Point.map
+
+#eval (.+2) <$> (Point.mk 1 2) -- { x := 4, y := 3 }
```
leodemoura, line 38: Please use `\cdot` instead of `.`. You are using `\cdot` in other places.

doc/monads/intro.md
```diff
@@ -0,0 +1,61 @@
...
+first before fully understanding monads:
+
+![image](../images/monads.png)
+
+This chapter is organized to give you a bottom up introduction to monads, starting with functors and
+applicative functors to get an intuition for how these abstract structures work in Lean. Then we’ll
+tackle monads and look at some of the most common ones.
+
+## [Functor](functors.lean.md)
+A functor is a type class that provides a `map` function and the map function is something many
```
leodemoura, line 20: Missing backticks in the second `map`?

doc/monads/functors.lean
```diff
@@ -0,0 +1,215 @@
+/-!
+# Functor
+
+A `Functor` is any type that can act as a generic container that allows you to transform the
+underlying values inside the container using a function, so that the values are all updated, but the
+structure of the container is the same. This is called "mapping".
+
+A List is the most basic example of a `Functor`.
```
leodemoura, line 8: add "one of"?

doc/monads/functors.lean
```diff
@@ -0,0 +1,215 @@
...
+infixr:100 " doodle " => Functor.map
+
+#eval (· * 5) doodle [1, 2, 3]  -- [5, 10, 15]
+/-!
+
+Let's look at some more examples:
+
+-/
+-- List String → List Nat
+#eval ["elephant", "tiger", "giraffe"].map (fun s => s.length)
```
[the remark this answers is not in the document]
leodemoura: We can say that we are using `List.map`, and remind the reader that we are using the dot notation. Not sure what the best approach is.

doc/monads/monads.lean
```diff
@@ -0,0 +1,307 @@
...
+The `bind` operator also has infix notation `>>=` where `x >>= g` represents the result of
+executing `x` to get a value of type `f α` then unwrapping the value `α` from that and
+passing it to function `g` of type `α → f β` returning the result of type `f β` where
+`f` is the target structure type (like `Option` or List)
+
+This `bind` operation looks similar to the other ones you've seen so far, if you put
+them all together `Monad` has the following operations:
+
+```lean,ignore
+class Monad (f : Type u → Type v) extends Applicative f, Bind f where
```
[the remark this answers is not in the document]
lovettchris: right, thanks for the reminder, I captured this here: https://github.com/leanprover/LeanInk/issues/30

doc/monads/monads.lean
lovettchris: Fixed: https://github.com/leanprover/lean4/pull/1505/commits/ed117c1d35e45fadd510a7c2223d3e93b46749c7

doc/monads/functors.lean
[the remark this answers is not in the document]
lovettchris: That's fine the earlier snippet was not trying to redefine List.map it was just showing the reader what List.map is.

doc/monads/functors.lean
lovettchris: fixed in https://github.com/leanprover/lean4/pull/1505/commits/e53237c5a9afcfcfd2536283350598f393081ab3

doc/monads/intro.md
lovettchris: fixed in https://github.com/leanprover/lean4/pull/1505/commits/e53237c5a9afcfcfd2536283350598f393081ab3

doc/monads/laws.lean
lovettchris: fixed in https://github.com/leanprover/lean4/pull/1505/commits/e53237c5a9afcfcfd2536283350598f393081ab3, actually it is not a \cdot, it is a \centerdot or a \.  The \cdot seems to be unicode value 2b1d (very small square)  whereas this dot is unicode 00b7 (middle dot)...

doc/monads/functors.lean
lovettchris: fixed in https://github.com/leanprover/lean4/pull/1505/commits/e53237c5a9afcfcfd2536283350598f393081ab3

doc/monads/monads.lean
lovettchris: fixed in https://github.com/leanprover/lean4/pull/1505/commits/e53237c5a9afcfcfd2536283350598f393081ab3

doc/monads/functors.lean
```diff
@@ -0,0 +1,215 @@
...
+  kitchenSize := 37.161216 }
+  -/
+/-!
+
+Wow, this is pretty powerful.  By providing a functor instance on `LivingSpace` with an
+implementation of the `map` function it is now super easy for anyone to come along and
+transform the units of a `LivingSpace` using very simple functions like `squareFeetToMeters`. Notice
+that squareFeetToMeters knows nothing about `LivingSpace`.
+
+You can implement the `Functor` pattern in other languages, people have even done it in C++, but it
```
leodemoura, line 186: This sentence is a bit out of place. Could you please remove it?

lovettchris: Done.

doc/monads/applicatives.lean
```diff
@@ -0,0 +1,329 @@
...
+An [applicative functor](https://en.wikipedia.org/wiki/Applicative_functor) is an intermediate
+structure between `Functor` and `Monad`. It mainly consists of two operations:
+
+* `pure : α → F α`
+* `seq : F (α → β) → F α → F β` (written as `<*>`)
+
+The `pure` operator specifies how you can wrap a normal object `α` into an instance of this structure `F α`.
+This is the "default" mechanism mentioned above.
+
+The `seq` operator gives a notion of evaluation order to the effects, where the first argument is
```
leodemoura, line 26: The word "effects" was introduced here without any explanation. I think this is the first occurrence.

doc/monads/applicatives.lean
```diff
@@ -0,0 +1,329 @@
...
+
+The `Seq` type class is also a simple type class that provides the `seq` operator which can
+also be written using the special syntax `<*>`.
+
+```lean,ignore
+class Seq (f : Type u → Type v) : Type (max (u+1) v) where
+  seq : {α β : Type u} → f (α → β) → (Unit → f α) → f β
+```
+
+The `seq` operator allows you to chain operations by wrapping a function in a structure. The name
```
leodemoura, line 73: There are two sentences starting with "The `seq` operator ...", but no connection between them.

doc/monads/applicatives.lean
```diff
@@ -0,0 +1,329 @@
...
+"applicative" comes from the fact that you "apply" functions from within the structure, rather than
+simply from outside the structure, as was the case with `Functor.map`.
+
+## Basic Applicative Examples
+
+Many of the basic functors also have instances of `Applicative` inherited from their Monad instances.
```
leodemoura, line 79: `Monad`s have not been introduced yet.

doc/monads/applicatives.lean
```diff
@@ -0,0 +1,329 @@
...
+#eval (.*.) <$> some 4 <*> none  -- none
+/-!
+
+For a more interesting example, let's make `List` an applicative by adding the following
+definition:
+
+-/
+instance : Applicative List where
+  pure := List.pure
+  seq f x := List.bind f fun y => Functor.map y (x ())
```
leodemoura, line 137: You are mixing `\lambda` and `fun`. It would be great to use one of them everywhere. I prefer `fun`.

doc/monads/applicatives.lean
leodemoura: We should also explain why we have `Unit -> ...` here.

doc/monads/applicatives.lean
leodemoura: You are also skipping the parentheses around the `fun y ...`. This is the style we use in our Lean code. I like it, but it is not used in other places in this PR.

doc/monads/applicatives.lean
```diff
@@ -0,0 +1,329 @@
...
+-/
+#eval [(· +2), (· *3)] <*> [4, 6] -- [6, 8, 12, 18]
+/-!
+
+Why did this produce 4 values?  The reason is because `<*>` applies _every_ function to _every_
+value in a pairwise manner.  This makes sequence really convenient for solving certain problems. For
+example, how do you get the pairwise combinations of all values from two lists?
+
+-/
+#eval Prod.mk <$> [1, 2, 3] <*> [4, 5, 6]
```
leodemoura, line 160: Is it worth reminding the reader that this is parsed as `(Prod.mk <$> [1, 2, 3]) <*> [4, 5, 6]`.

doc/monads/applicatives.lean
```diff
@@ -0,0 +1,329 @@
...
+master bedroom size _and_ the living room size? That wouldn't really make sense. And what would the
+numBedrooms value be for the default? What would it mean to "chain" two of these objects together?
+
+If you can't answer these questions very well, then it suggests this type isn't really an
+Applicative functor.
+
+## SeqLeft and SeqRight
+
+You may remember seeing the `SeqLeft` and `SeqRight` base types on `class Applicative` earlier.
+These provide the `seqLeft` and `seqRight` operatoions which also have some handy notation
```
leodemoura, line 253: Typo: "operatoions"

doc/monads/applicatives.lean
leodemoura: Perhaps, you could add a link to the Lazy Evaluation section, or just mention you will discuss it later.

doc/monads/applicatives.lean
```diff
@@ -0,0 +1,329 @@
...
+
+## Lazy Evaluation
+
+Diving a bit deeper, (you can skip this and jump to the [Applicative
+Laws](laws.lean.md#what-are-the-applicative-laws) if don't want to dive into this implementation detail right
+now). But, if you write a simple `Option` example `(.*.) <$> some 4 <*> some 5` that produces `some 20`
+using `Seq.seq` you will see somthing interesting:
+
+-/
+#eval Seq.seq ((.*.) <$> (some 4)) (fun (_ : Unit) => (some 5)) -- some 20
```
leodemoura, line 287: You don't need parentheses around `some 5`

doc/monads/applicatives.lean
```diff
@@ -0,0 +1,329 @@
...
+`fun (_ : Unit) => (some 5)`?
+
+Well if you take a close look at the type class definition:
+```lean
+class Seq (f : Type u → Type v) where
+  seq : {α β : Type u} → f (α → β) → (Unit → f α) → f β
+```
+
+You will see this function defined here: `(Unit → f α)`, this is a function that takes `Unit` as input
+and produces the output of type `f α` where `f` is the container type `Type u`, in this example `Option`
```
leodemoura, line 300: `f` has type `Type u -> Type v` instead of `Type u`.

doc/monads/functors.lean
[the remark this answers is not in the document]
leodemoura: It does not look fine to me. We should at least say it. Otherwise, the reader may assume that after you define a function called `map`, they can write `[1,2,3,4,5].map ...`

doc/monads/monads.lean
```diff
@@ -0,0 +1,307 @@
...
+  ...
+```
+
+Notice `Monad` also contains `pure` it must also have a "default" way to wrap a value in the
+structure.
+
+The `bind` operator is similar to the applicative `seq` operator in that it chains two operations,
+with one of them being function related.
+
+Notice that `bind`, `seq` and `map` all take a function of some kind, and then a structure wrapped
```
leodemoura, line 57: This paragraph looks weird to me. Not sure what is the point you are trying to make here. Does it come from the source material?

doc/monads/monads.lean
```diff
@@ -0,0 +1,307 @@
...
+#eval "apple".toList  -- ['a', 'p', 'p', 'l', 'e']
+
+#eval ["apple", "orange"] >>= String.toList
+-- ['a', 'p', 'p', 'l', 'e', 'o', 'r', 'a', 'n', 'g', 'e']
+/-!
+List also provides some functions that are designed to operate in the context of a monad.
+These methods end in upper case M like `anyM` below:
+-/
+
+def hasSomeItemGreaterThan (x : List Nat) (n : Nat): Option Bool := do
```
leodemoura, line 220: This example is strange. `Option` is not necessary here, and one can write more compact pure code.
```lean
def hasSomeItemGreaterThan' (x : List Nat) (n : Nat) : Bool :=
  x.any λ a => if a > n then true else false 
```
BTW, note that a novice reader may also get confused by how this example even type checks in Lean. Lean is injecting a coercion (`some` in this case) at `x`.
Could you please find a better example?

doc/monads/readers.lean
```diff
@@ -0,0 +1,196 @@
...
+-/
+
+structure Environment where
+  path : String
+  home : String
+  user : String
+  deriving Repr
+
+def getEnvDefault (name : String): IO String := do
+  pure <| match (← IO.getEnv name) with
```
leodemoura, line 27: I think you did not introduce the nested monadic action notation `(<- ...)` yet.

leodemoura: Perhaps, write it as follows at this point.
```
def getEnvDefault (name : String) : IO String := do
 let val? ← IO.getEnv name
 match val? with
 | none => pure ""
 | some s => pure s
```

doc/monads/readers.lean
```diff
@@ -0,0 +1,196 @@
...
+operations happening here.  The `readerFunc3` function uses the  `do` notation you learned about in
+[Monads](monads.lean.md) which hides that bind operation and can make the code look cleaner.
+
+The `do` notation with `let x ← readerFunc2` is also calling the `run` function to actually execute
+the monadic action, then it unwraps the value x, so it is really doing `let x ← readerFunc2.run`.
+
+The important difference here to the earlier code is that `readerFunc3` and `readerFunc2` no longer
+have an **explicit** Environment input parameter that needs to be passed along all the way to
+`readerFunc1`.  Instead, the `ReaderM` monad is taking care of that for you, which gives you the
+illusion of something like global state where the state is now available to all functions that use
```
leodemoura, line 135: "global context" instead of "global state".

doc/monads/readers.lean
```diff
@@ -0,0 +1,196 @@
...
+the `ReaderM` monad.
+
+The above code also introduces an important idea. Whenever you learn about a monad "X", there's
+often (but not always) a `run` function to execute that monad, and sometimes some additional
+functions like `read` that interact with the monad state.
+
+You might be wondering, how does the state actually move through the `ReaderM` monad? How can you
```
leodemoura, line 142: "state" => "context"

doc/monads/functors.lean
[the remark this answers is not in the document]
lovettchris: I'm still not sure I agree, notice I use the list `.map` function BEFORE I show the map implementation on this line: 
```lean
#eval [1,2,3].map (λ x => toString x) -- ["1", "2", "3"]
```
And it clearly introduces the code as "List has a specialized version of `map` defined as follows:" this is pretty clearly saying this is what map does, and it is not saying something like "Let's define a new map function like this:" then sure, we'd have to show the full Functor redefinition, but that would be jumping the gun because this is the chapter on Functors which we haven't introduced yet...

So I move to the "dot" notation because of your comments on List and Array docs should use dot notation.  Would it be better if I moved these early snippets back to using `List.map` to make it clear which version of map we are using here?

doc/monads/readers.lean
```diff
@@ -0,0 +1,196 @@
...
+add an input argument to a function by modifying it's return type?  There is a special command in
+the Lean interpreter that will show you the reduced Types:
+-/
+#reduce ReaderM Environment String   -- Environment → String
+/-!
+And you can see here that this type is actually a function!  It's a function that takes an
+`Environment` as input and returns a `String`.  So, remember in Lean that a function that takes
+argument a and returns a string: `def f (a : Nat) → String` is the same as the function that takes no
```
leodemoura, line 150: This paragraph is incorrect. We cannot say that `f : Nat -> String` is a function that takes no arguments and returns another function of type `Nat -> String`.

doc/monads/readers.lean
```diff
@@ -0,0 +1,196 @@
...
+ReaderM Monad to add an input argument to all the functions that use the `ReaderM` monad and this is why
+`main` is able to start things off by simply passing that new input argument in `readerFunc3 env`.
+So now that you know the implementation details of the `ReaderM` monad you can see that what it is
+doing looks very much like the original code we wrote at the beginning of this section, only it's
+taking a lot of the tedious work off your plate and it is creating a nice clean separation between
+what your pure functions are doing, and the global state idea that the `ReaderM` adds.
+
+## withReader
+
+One `ReaderM` function can call another with a modified version of the `ReaderM` state.
```
leodemoura, line 161: This section needs to be improved. The `withReader` is poorly motivated. We should say that `withReader f act` executes `act` in context modified by `f`. We should emphasize that the modification affects only `act`. Adding an example would help too.

doc/monads/applicatives.lean
lovettchris: I carefully crafted the doc to avoid that complication until the Lazy Evaluation section because I don't want to side track the reader from what is important about seq, this is why I jump to using the `<*>` notation instead of showing how to call `seq` directly.  I think a forward reference would only complicate things unnecessarily.

doc/monads/states.lean
```diff
@@ -0,0 +1,235 @@
...
+`StateM` monad such that the initial `GameState` is passed in as a new input argument, and the output
+value has been changed to the pair `Bool × GameState` so that it can return the pure `Bool` and the
+updated `GameState`.  This is why the call to `nextTurn` looks like this: `let (_, g) := nextTurn gs`.
+This expression `(_, g)` conveniently breaks the pair up into 2 values, it doesn't care what the first
+value is (hence the underscore `_`), but it does need the updated state `g` which you can then assign
+back to the mutable `gs` variable to use next time around this loop.
+
+## StateM vs ReaderM
+
+While `ReaderM` functions can use `withReader` to modify the state before calling another function,
```
leodemoura, line 193: I have never seen `withReader f act` as modifying the `state`. For me, it is modifying the **context** for the monadic action `act`. Remark, some people use environment instead of context.

doc/monads/except.lean
```diff
@@ -0,0 +1,135 @@
...
+-/
+def testCatch :=
+  try
+    let r ← divide 8 0  -- 'r' is type Float
+    pure (toString r)
+  catch e =>
+    pure s!"Caught exception: {e}"
+/-!
+
+Note that the type inferred by Lean for this function is `Except String String` so the `Except
```
leodemoura, line 91: I am not sure what you are trying to say here. What transformation are you referring to? 
There is no magic happening here, you have used `toString` to convert the `Float` into a `String`.

doc/monads/applicatives.lean
lovettchris: fixed in https://github.com/leanprover/lean4/pull/1505/commits/753cc2765f26d4b111fd6c4b0b0ee0d183d741cb

doc/monads/applicatives.lean
lovettchris: fixed in https://github.com/leanprover/lean4/pull/1505/commits/753cc2765f26d4b111fd6c4b0b0ee0d183d741cb

doc/monads/functors.lean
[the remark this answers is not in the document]
leodemoura: Yes, there are four different `map` functions in this section: `List.map`, `Option.map`, `Functor.map`, and `map` (the version of `List.map` you have defined here). The notation for `<$>` is for `Functor.map`.

doc/monads/transformers.lean
```diff
@@ -0,0 +1,316 @@
...
+
+The StateT monad transformer defines an instance of `MonadLift` like this:
+
+```lean
+@[inline] protected def lift {α : Type u} (t : m α) : StateT σ m α :=
+  fun s => do let a ← t; pure (a, s)
+
+instance : MonadLift m (StateT σ m) := ⟨StateT.lift⟩
+```
+This means that any monad `m` can be wrapped in a `StateT` monad by inventing the trivial function
```
leodemoura, line 183: Suggestion: "by inventing the trivial function" => "by using the function"

doc/monads/transformers.lean
```diff
@@ -0,0 +1,316 @@
...
+`fun s => do let a ← t; pure (a, s)` that takes state `s`, runs the inner monad action `t`, and
+returns the result and the new state in a pair `(a, s)` without making any changes to `s`.
+
+Because `MonadLift` is a type class, type inference can automatically find the required `monadLift`
```
leodemoura, line 187: "type inference can automatically find" 
=>
"Lean can automatically find"

doc/monads/applicatives.lean
lovettchris: Fixed, and `(some 4)` also...

doc/monads/applicatives.lean
leodemoura: > Interesting, I think it is educational to use both so readers become familiar and comfortable with both notations.

Then you should probably at least mention it. Otherwise, they will keep scratching their head about why sometimes you use it, and sometimes you do not.

doc/monads/applicatives.lean
lovettchris: Fixed I added it to Functors here: https://github.com/leanprover/lean4/pull/1505/commits/ae2bafb5f640f7b778990aadf8cc217cc934cc35

doc/monads/applicatives.lean
leodemoura: It looks confusing to me. There are two `seq`s in the same section with different types and no explanation why.

doc/monads/laws.lean
```diff
@@ -0,0 +1,319 @@
...
+Transformers](transformers.lean.md).
+
+So far, you've learned the concrete details you need in order to _use_ monads in your Lean programs.
+But there's still one more important concept you need if you want to _create_ new functors,
+applicatives and monads. Namely, the notion of _structural "laws"_ -- rules that these type
+classes should follow in order to meet other programmers' expectations about your code.
+
+## Life without Laws
+
+Remember Lean represents each of these abstract classes by a type class. Each of these type classes
```
leodemoura, line 17: "abstract classes" => "abstract structures"

doc/monads/monads.lean
lovettchris: Yes this is the original, I only added type information so you can map the prose back to the Monad type class definition shown a few lines above:

> All these operators take a function of some kind, and then a structure wrapped over an a type, and then produce a structure wrapping a b type. They just vary in what the function looks like. For the functor, the function is a normal pure function. For applicatives, the function is still pure, but wrapped in the structure. Now with monads, our function argument takes a "pure" input but produces an output in the structure.

lovettchris: I personally found it helpful to look at the 3 different functions for map, seq, bind and understand their differences, which this paragraph does, and it helped me get familiar with the language "structure wrapped over a type" that is being performed by the `f α` and  `f β` terms in the Monad type class definition.

doc/monads/laws.lean
```diff
@@ -0,0 +1,319 @@
...
+should be the same as "composing" the functions and simply mapping that one super-function over the
+functor.  In Lean you can compose two functions using `Function.comp f g` (or the syntax `f ∘ g`,
+which you can type in VS code using `\o `) and you will get the same results from both of these
+showing that the composition law holds for `List Nat`:
+
+-/
+def double (x : Nat) := x + x
+def square (x : Nat) := x * x
+
+#eval (double <$> (square <$> list1)) -- [2, 8, 18]
```
leodemoura, line 80: You don't need the outer parentheses here.
```
#eval double <$> (square <$> list1)
```

doc/monads/monads.lean
lovettchris: Can you suggest a better example of `anyM` or any other monadic function provided by List ?

doc/monads/monads.lean
leodemoura: The paragraph seems to be stating in English the information that is in their types. I don't see it providing new information. The rest of the text is already showing how to use them, the laws they obey, and examples.

doc/monads/monads.lean
leodemoura: If I remember correctly, you have defined only two monads in this section `Option` and `List`. If you want to keep the example with `anyM`, I think it is better to move it to the `Reader` section. Example
```
structure Env where 
  min : Nat
  max : Nat

def hasSomeItemOutsideRange (x : List Nat) : ReaderM Env Bool :=
  x.anyM fun a => do let env ← read; pure (a < env.min || a > env.max)
```

doc/monads/readers.lean
```diff
@@ -0,0 +1,196 @@
...
+
+The `let env ← read` expression in `readerFunc1` unwraps the environment from the `ReaderM` so we
+can use it. Each type of monad might provide one or more extra functions like this, functions that
+become available only when you are in the context of that monad.
+
+Here the `readerFunc2` function uses the `bind` operator `>>=` just to show you that there are bind
+operations happening here.  The `readerFunc3` function uses the  `do` notation you learned about in
+[Monads](monads.lean.md) which hides that bind operation and can make the code look cleaner.
+
+The `do` notation with `let x ← readerFunc2` is also calling the `run` function to actually execute
```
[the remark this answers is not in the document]
lovettchris: Ok I was confused about that, can you tell me more? The reason I was confused is that I can replace that line with `let x ← readerFunc2.run` and I get the exact same result.  The unwrapping of the readerFunc2 and assignment to `x` surely implies executing the monad to get that value somehow?

doc/monads/readers.lean
lovettchris: Ok, what about in states.lean where I also use "global state", perhaps that is ok because it is talking about `StateM` ?

leodemoura: Yes, exactly.

doc/monads/readers.lean
lovettchris: I was trying to say  `def f (a : Nat) → String` can be rewritten as `def f : Nat → String` to help make the connection to the user that this is how ReaderM can add an input argument even though ReaderM is used as a return type.  It is building on Lean's support for "currying" ?  What is a better way to say all this?

doc/monads/readers.lean
lovettchris: Fixed in https://github.com/leanprover/lean4/pull/1505/commits/6fed17209a17e523b02d2dd6b0cc773ec551738f

doc/monads/readers.lean
lovettchris: Fixed in https://github.com/leanprover/lean4/pull/1505/commits/6fed17209a17e523b02d2dd6b0cc773ec551738f

doc/monads/readers.lean
lovettchris: Fixed in https://github.com/leanprover/lean4/pull/1505/commits/6fed17209a17e523b02d2dd6b0cc773ec551738f

doc/monads/readers.lean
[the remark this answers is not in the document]
leodemoura: Getting the same result does not imply that the `do` notation calls the function `run`. This is just not true. You can check by yourself by printing how the `do` notation has been elaborated by Lean.
```lean
set_option pp.notation false
#print readerFunc1
```

doc/monads/monads.lean
```diff
@@ -0,0 +1,307 @@
...
+
+The state of the world outside your program can change at virtually any moment, and so this IO
+context is particularly special. So these IO operations are "side effects" which means you cannot
+perform them from "pure" Lean functions.
+
+Now, the most important job of pretty much any computer program is precisely to perform this
+interaction with the outside world. For this reason, the root of all executable Lean code is a
+function called main, with the type `IO Unit`. So every program starts in the IO monad!
+
+When you function is `IO` monadic, you can get any input you need, call into "pure" code with the
```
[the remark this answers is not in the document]
lovettchris: Thanks, fixed in https://github.com/leanprover/lean4/pull/1505/commits/6fed17209a17e523b02d2dd6b0cc773ec551738f

doc/monads/applicatives.lean
lovettchris: Good catch, fixed in https://github.com/leanprover/lean4/pull/1505/commits/6fed17209a17e523b02d2dd6b0cc773ec551738f

doc/monads/readers.lean
lovettchris: How's this instead?

## withReader

One `ReaderM` function can call another with a modified version of the `ReaderM` context.
You can use the `withReader` function from the `MonadWithReader` typeclass to do this:

-/
def readerFunc3WithReader : ReaderM Environment String := do
  let f := (λ env => { env with user := "new user" })
  let x ← withReader f readerFunc2
  return "Result: " ++ (toString x)

/-!

So `withReader f m` executes monad `m` in the `ReaderM` context modified by `f`.

doc/monads/states.lean
lovettchris: Yes, great example of how using the term "context" instead of state here helps disambiguate.  I changed it.

doc/monads/except.lean
lovettchris: Yeah it is subtle and I ran into a ton of confusion in this area which is why I'm calling this out.  If I wrote this:

```lean
def test := divide 5 0
```
lean would have inferred the type for `test` to be `Except String Float`.  But if I change the pure return type to a string as in testCatch, then lean magically figures out that I must have wanted the type for `testCatch` to be `Except String String`, without me having to state that explicitly.  

We had a zulip discussion about this, or was it in the last wed call, where I was wondering why `testCatch` didn't infer the return type of `String` since I caught the exception and explicitly made the catch block convert the error to a string instead of rethrowing.  You said lean chose to continue to propagate the Except type in this case because someone "might" choose to rethrow in a case like this... anyway this is what I'm trying to explain.  Any thoughts on how to improve it?

doc/monads/laws.lean
lovettchris: fixed in https://github.com/leanprover/lean4/pull/1505/commits/8dd24bc9dae40eb19b4625bf0b94cfb3b15a2f07

doc/monads/laws.lean
lovettchris: Fixed in https://github.com/leanprover/lean4/pull/1505/commits/8dd24bc9dae40eb19b4625bf0b94cfb3b15a2f07

doc/monads/transformers.lean
lovettchris: Fixed in https://github.com/leanprover/lean4/pull/1505/commits/8dd24bc9dae40eb19b4625bf0b94cfb3b15a2f07

doc/monads/transformers.lean
lovettchris: Fixed in https://github.com/leanprover/lean4/pull/1505/commits/8dd24bc9dae40eb19b4625bf0b94cfb3b15a2f07

doc/monads/readers.lean
leodemoura: This is not related to currying. If I remember correctly, David's book explains currying. You can also check the page on Wikipedia. 
The types `ReaderM Environment String` and `Environment -> String` are definitionally equal in Lean. The following definition is type correct in Lean
```lean
def f : ReaderM Environment String :=
  fun (_ : Environment) => "hello world"
``` 
The expression `fun (_ : Environment) => "hello world"` has type `Environment -> String`.
```lean
#check fun (_ : Environment) => "hello world" 
```
Lean accepts the definition because the types are definitionally equal. TPIL mentions definitionally equality several times. I think it is worth keeping re-reading TPIL until the Lean type system is clear to you.

doc/monads/states.lean
lovettchris: Fixed in https://github.com/leanprover/lean4/pull/1505/commits/8dd24bc9dae40eb19b4625bf0b94cfb3b15a2f07

doc/monads/readers.lean
lovettchris: Fixed in https://github.com/leanprover/lean4/pull/1505/commits/8dd24bc9dae40eb19b4625bf0b94cfb3b15a2f07

doc/monads/except.lean
lovettchris: but perhaps "transformation" is the wrong language, perhaps I should just say it is another kind of "type inference".

doc/monads/readers.lean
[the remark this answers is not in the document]
lovettchris: Thanks for the handy tool - it helps clear things up, when I use this on readerFunc3 without any run I see 
```
bind readerFunc2 fun x => pure (HAppend.hAppend "Result: " (toString x))
```
and with the run I see
```
bind (ReaderT.run readerFunc2) fun x => pure (HAppend.hAppend "Result: " (toString x))
```

So I rewrote it like this:

> The `do` notation with `let x ← readerFunc2` is also calling the `bind` function under the covers,
so that you can access the unwrapped value `x` needed for the `toString x` conversion.

doc/monads/readers.lean
lovettchris: Ok, so building on that I can see that these are definitially equal, so I will use that terminology:
```lean
def f (a : Nat) : String := "abc"

#print f -- fun a => "abc"

def g : Nat → String := λ a => "abc"

#print g -- fun a => "abc"

example : f = g := by rfl  -- Goals accomplished 🎉
```
I will not call this `currying` even though that's exactly what David called it here:
https://github.com/leanprover/fp-lean/blob/master/functional-programming-lean/src/getting-to-know/functions-and-definitions.md?plain=1#L70

lovettchris: PS: I had a typo in the original text that may explain all your confusion here, I had `def f (a : Nat) → String` instead of `def f (a : Nat) : String`.

I have updated this text in https://github.com/leanprover/lean4/pull/1505/commits/9b24f8de09a70d3793d040aff7d217bed8be8acc, so please me know if the new version is better.

doc/monads/readers.lean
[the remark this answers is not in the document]
lovettchris: fixed in https://github.com/leanprover/lean4/pull/1505/commits/9b24f8de09a70d3793d040aff7d217bed8be8acc

doc/monads/readers.lean
leodemoura: > I will not call this currying even though that's exactly what David called it here:
https://github.com/leanprover/fp-lean/blob/master/functional-programming-lean/src/getting-to-know/functions-and-definitions.md?plain=1#L70

Read the text again. You will see it is different. He has a binary function.

doc/monads/monads.lean
lovettchris: Thanks, I like the idea of requiring a more interesting monad context like ReaderM, or Except perhaps?  Problem with hasSomeItemOutsideRange is I would normally just lift `let env ← read` outside and call x.any.

How about this one:

```lean
def validateList (x : List Nat) (n : Nat): Except String Bool := do
  x.anyM (λ a => if a > n then do throw "illegal value found in list" else pure true)

#eval validateList [1, 2, 5, 3, 8] 10 -- some true
```

doc/monads/except.lean
leodemoura: > But if I change the pure return type to a string as in testCatch, then lean magically figures out that I must have wanted the type for testCatch to be Except String String

Let's focus on the `try` block you wrote in this example. It is
```lean
    let r ← divide 8 0  -- 'r' is type Float
    pure (toString r)
```
Which is notation for
```lean
bind (divide 8 0) (fun (r : Float) => pure (toString r))
```
We have 
```lean
divide 8 0 : Except String Float
```
and 
```lean
fun (r : Float) => pure (toString r) : Fload -> Except String String
```
Recall that `toString r` has type `String`, and `pure (toString r) : Except String String`.
Now, the `bind` puts together these two "pieces" together that have types `Except String Float` and `Float -> Except String String`, and produces a value of type `Except String String`. 
There is no magic.

doc/monads/monads.lean
lovettchris: I'll move this to except.lean.

doc/monads/except.lean
leodemoura: > but perhaps "transformation" is the wrong language, perhaps I should just say it is another kind of compositional "type inference".

There is no such thing in Lean.

doc/monads/monads.lean
leodemoura: > Thanks, I like the idea of requiring a more interesting monad context like ReaderM, or Except perhaps? Problem with hasSomeItemOutsideRange is I would normally just lift let env ← read outside and call x.any.

Yes, you could do that. We can create a realistic example by using 
```lean
def inRange (x : Nat) :  ReaderM Env Bool := do
  let env ← read
  pure (a < env.min || a > env.max)

def hasSomeItemOutsideRange (x : List Nat) : ReaderM Env Bool :=
  x.anyM inRange
```
You will find this kind of pattern in many places in our code base.

leodemoura: The following example is not very realistic since the nested function fails or returns true. That is, it never returns false. It feels weird to use `anyM`.
```lean
def validateList (x : List Nat) (n : Nat): Except String Bool := do
  x.anyM (λ a => if a > n then do throw "illegal value found in list" else pure true)
```
A more realistic option is to use
```lean
def validateList (x : List Nat) (n : Nat): Except String Unit :=
  x.forM fun a => do 
    if a > n then throw "illegal value found in list"
```

doc/monads/except.lean
lovettchris: Fixed in https://github.com/leanprover/lean4/pull/1505/commits/9e2398bf897e5bd656ec1ad48b83d32896079e6a

doc/monads/monads.lean
lovettchris: ha ha but Leo that's because you think and breath Lean.  The rest of us need it spelled out so we can understand what all those symbols mean!

leodemoura: In my point of view, you are just spelling out the types of these functions. Are you assuming that people reading this section cannot read Lean types?

lovettchris: Are you saying you don't like the original paragraph either, or are you saying you don't like the type annotations I added?

doc/monads/monads.lean
lovettchris: Ok, that's cool, I'll use this one, thanks.

doc/monads/monads.lean
leodemoura: I checked the original. Note that they flipped the `bind` to make their point. I also think their sentence is a bit weird, but it reads better.

doc/monads/monads.lean
lovettchris: Fixed in https://github.com/leanprover/lean4/pull/1505/commits/40da0e10b21be41da5e54fdf709a14ff447abde6

doc/monads/applicatives.lean
lovettchris: Fixed in Functors where it is first used: https://github.com/leanprover/lean4/pull/1505/commits/6fd6e59c7e0c82966c591ca70768ad0a39663897

doc/monads/monads.lean
lovettchris: I couldn't find =<< in Lean?  So what's your conclusion?  Delete the whole paragraph, leave it, remove the types, rewrite it?  I'm happy with either decision.

lovettchris: How about this instead:


Notice that `bind`, `seq` and `map` all take a function of some kind.  Let's examine those function types:

- map: `(α → β)`
- seq: `f (α → β)`   -- ignoring `(Unit → f α)` for now.
- bind: `(α → f β)`

So `map` is a pure function, `seq` is a pure function wrapped in the structure, and `bind` takes
a "pure" input but produces an output wrapped in the structure.

lovettchris: Fixed in https://github.com/leanprover/lean4/pull/1505/commits/11ceaf23f12fda8daf3ae64f991011672ad8409c (sorry about the reflow, I'm using a vscode reflow line wrapping plugin and I didn't expect it to roll forwards like that).

digama0: > In my point of view, you are just spelling out the types of these functions. Are you assuming that people reading this section cannot read Lean types?

I agree with @lovettchris 's point on this. The purpose of documentation is to help people understand, so restating information that is "already contained in the types" is not redundant information if it helps people understand how to gloss lean types in general. In particular, `bind` has a famously complex type signature that has birthed many tutorials and articles, even though it's quite trivial once you are used to this kind of thing.

doc/monads/except.lean
Kha: Why leave this to inference anyway? We always give explicit types of top-level declarations, is there any reason not to do that here?

doc/monads/readers.lean
```diff
@@ -0,0 +1,196 @@
...
+/-!
+`main2` loads the environment as before, and estabilishes the `ReaderM` state by passing `env` to
+readerFunc3.  The `do` notation you learned about in [Monads](monads.lean.md) is hiding some things
+here.  Technically `readerFunc3` is a monadic action and in order to "run" that action the `ReaderM`
+monad provides a `run` method and it is the `ReaderM` run method that takes the initial
+`Environment` state.  So you can actually write the complete code by doing this: `let str :=
+readerFunc3.run env`.
+
+> **Side note 1**: The `return` statement used above also needs some explanation.  The `return`
+statement in Lean can only be used inside a `do` block and is closely related to `pure`, but a
```
Kha, line 103: This isn't true, `return` alone acts as if it had an implicit `do` block around it (i.e. the only difference to `pure` is the parsing precedence). Though I understand that you might not want to lead with this extensions. Perhaps you could remove the "only" and mention return-without-do at the end of the paragraph or elsewhere.

doc/monads/readers.lean
```diff
@@ -0,0 +1,196 @@
...
+`readerFunc1`.  Instead, the `ReaderM` monad is taking care of that for you, which gives you the
+illusion of something like global state where the state is now available to all functions that use
+the `ReaderM` monad.
+
+The above code also introduces an important idea. Whenever you learn about a monad "X", there's
+often (but not always) a `run` function to execute that monad, and sometimes some additional
+functions like `read` that interact with the monad state.
+
+You might be wondering, how does the state actually move through the `ReaderM` monad? How can you
+add an input argument to a function by modifying it's return type?  There is a special command in
```
Kha, line 144:
```suggestion
add an input argument to a function by modifying its return type?  There is a special command in
```

doc/monads/readers.lean
```diff
@@ -0,0 +1,196 @@
...
+the Lean interpreter that will show you the reduced Types:
```
Kha, line 145:
```suggestion
Lean that will show you the reduced types:
```

doc/monads/readers.lean
```diff
@@ -0,0 +1,199 @@
...
+You might be wondering, how does the context actually move through the `ReaderM` monad? How can you
+add an input argument to a function by modifying it's return type?  There is a special command in
+the Lean interpreter that will show you the reduced Types:
+-/
+#reduce ReaderM Environment String   -- Environment → String
+/-!
+And you can see here that this type is actually a function!  It's a function that takes an
+`Environment` as input and returns a `String`.
+
+Now, remember in Lean that a function that takes argument of type `Nat` and returns a `String` like
```
Kha, line 152:
```suggestion
Now, remember in Lean that a function that takes an argument of type `Nat` and returns a `String` like
```

doc/monads/readers.lean
```diff
@@ -0,0 +1,199 @@
...
+`def f (a : Nat) : String` is the same as this function `def f : Nat → String`.  These are
+definitially equal.  Well this is being used by the `ReaderM` Monad to add an input argument to
```
Kha, line 154: You could even say "exactly equal [as types]" here. I didn't check if you discuss defeq in other places.

doc/monads/readers.lean
```diff
@@ -0,0 +1,199 @@
...
+  IO.println str
+
+#eval main2 -- Result: 7538
+/-!
+`main2` loads the environment as before, and estabilishes the `ReaderM` context by passing `env` to
+readerFunc3.  The `do` notation you learned about in [Monads](monads.lean.md) is hiding some things
+here.  Technically `readerFunc3` is a monadic action and in order to "run" that action the `ReaderM`
+monad provides a `run` method and it is the `ReaderM` run method that takes the initial
+`Environment` context.  So you can actually write the complete code by doing this: `let str :=
+readerFunc3.run env`.
```
Kha, line 100: I think I would like it better if `readerFunc3.run env` was presented as the "canonical" code, and then later in the `ReaderM` implementation section you can state that this is equivalent to `readerFunc3 env` (unless we enact https://github.com/leanprover/lean4/issues/1547#issuecomment-1231838157 :) ).

doc/monads/readers.lean
```diff
@@ -0,0 +1,199 @@
...
+global context idea that the `ReaderM` adds.
+
+## withReader
+
+One `ReaderM` function can call another with a modified version of the `ReaderM` context. You can
+use the `withReader` function from the `MonadWithReader` typeclass to do this:
+
+-/
+def readerFunc3WithReader : ReaderM Environment String := do
+  let f := (λ env => { env with user := "new user" })
```
Kha, line 169:
```suggestion
  let f env := { env with user := "new user" }
```
Unless you want to avoid this syntax extension

doc/monads/readers.lean
```diff
@@ -0,0 +1,199 @@
...
+  let x ← withReader f readerFunc2
+  return "Result: " ++ (toString x)
```
Kha, line 171:
```suggestion
  return "Result: " ++ toString x
```

doc/monads/readers.lean
lovettchris: Thanks, I removed the comment about only in `do` blocks.  The rest I got from Henrik who said on the Zulip discussion this:

> Note that unlike in haskell there is actually a semantic difference between return and pure in Lean, first return is a special keyword so I can write `return f x` and it will be parsed as `return (f x)` while `pure f x` gets parsed as `(pure f) x` so you usually have to put in a `pure <| f x`, furthermore return will also cause an early return in a monadic function like in an imperative language while pure will not.

doc/monads/readers.lean
lovettchris: fixed.

doc/monads/readers.lean
lovettchris: Fixed in https://github.com/lovettchris/lean4/commit/aa8a2586f27ea31f3c8d4a32893322ed5536d43f

doc/monads/readers.lean
lovettchris: There is definitely a very real conceptual overload that happens with Lean very quickly if we throw everything in the kitchen sink at the reader all at once.  I'm struggling with this myself.  For example, I didn't even know about this syntax extension.  So, I think it might be better to move that to a more advanced chapter on really cool syntax extensions that can make your code even more beautiful, or something.

doc/monads/readers.lean
lovettchris: Fixed in [e876200bae755f0a6a2e0e535e28136fe0178c3b ](https://github.com/lovettchris/lean4/commit/e876200bae755f0a6a2e0e535e28136fe0178c3b) - there were 2 cases.

doc/monads/readers.lean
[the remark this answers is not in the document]
lovettchris: Leo, related to the above discussion, I added this paragraph in [Stat](https://lovettsoftware.com/lean4/monads/states.lean.html#implementation), love to get your feedback on it:

> It is also interesting to see how much work the `do` and `←` notation are doing for you.  To
implement the `nextTurn` function without these you would have to write this, manually plumbing
the state all the way through:
```lean4
def nextTurnManually : StateM GameState Bool
| state =>
  let (i, gs) := chooseRandomMove |>.run state
  let (_, gs') := applyMove i  |>.run gs
  let (result, gs'') := isGameDone |>.run gs'
  (result, gs'')
```

doc/monads/readers.lean
lovettchris: Fixed, I wrote this instead: 
> These are exactly equal as types, which means they are "definitionally equal"

doc/monads/readers.lean
leodemoura: Let's go with Sebastian's suggestion.

doc/monads/functors.lean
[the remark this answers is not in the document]
lovettchris: Ok, I added `List.map` and moved the reference to `<$>` down and introduced it like this:

> Lean also defines custom infix operator `<$>` for `Functor.map` which allows you to write this:

See https://github.com/lovettchris/lean4/commit/2ef1e40fc5eae9a452a5d73af6cd065c9e352a69

doc/monads/readers.lean
lovettchris: Hmm, no comment on my point about conceptual overload.  So I will go with this insead which sidesteps the whole issue:

```lean
def readerFunc3WithReader : ReaderM Environment String := do
  let x ← withReader (λ env => { env with user := "new user" }) readerFunc2
  return "Result: " ++ toString x
```

doc/monads/states.lean
```diff
@@ -0,0 +1,252 @@
+import Bootstrap.Data.HashMap
```
lovettchris, line 1: By the way @leodemoura I just want to call out I had to change this import recently from `import Std.Data.HashMap`, so was this breaking change intentional?

digama0: As I mentioned on Zulip, this is a temporary state of affairs. If you want to keep it compiling, you need to change it to `import Lean.Data.HashMap`, but we will want to guide people to use the Std library, which will locate this at `import Std.Data.HashMap`. (Not sure whether adding Std to the docs will cause boostrapping issues.)

lovettchris: image

doc/monads/readers.lean
Kha: That makes it sound like that's the definition of definitional equality, which is misleading at best. I don't know why you want to reference that here, it doesn't seem relevant at all.

lovettchris: I see, well I was just using the language Leo gave me earlier, but I dropped it since it seems to be confusing things: https://github.com/lovettchris/lean4/commit/4b9d73814a625e08135719672cbf611134a8e472

doc/monads/states.lean
digama0: You need to change it to `import Lean.Data.HashMap` [tomorrow](https://github.com/leanprover/lean4/commit/ebb5b97d73e3ad). Like I said, things are in flux.

lovettchris: I see, will do, thanks for the link.

doc/monads/readers.lean
leodemoura: @lovettchris I used definitionally equality in the `ReaderM Environment String` vs `Environment -> String` issue.

doc/monads/states.lean
lovettchris: Ok, I tested it with my own local build of master and it works great so I checked in the change to `import Lean.Data.HashMap`.

doc/monads/readers.lean
lovettchris: Right, but I agree with Sebastian it's an advanced topic better left to TPIL book, so probably doesn't belong here.

PR #1512, opened 2022-08-25
fix: `Core.transform` API and uses

Kha:
`Core.transform`'s docs for `pre` currently say
> If the result is `TransformStep.visit sNew`, then `sNew` is traversed by transform.

This description is ambiguous: is `sNew` passed to `pre` again, or do we only visit its children (which is the correct thing to do for the default implementation of returning `.visit s`)? The current implementation does the latter, but *multiple* use sites assume it is the former (I stopped further investigating after isolating the first two such bugs).

This PR separates the two interpretations into two distinct `TransformStep` constructors, fixing the two isolated bugs and possibly more.

### DIFF
--- a/src/Lean/Expr.lean
+++ b/src/Lean/Expr.lean
@@ -1297,7 +1297,7 @@ If `useZeta = true`, then `let`-expressions are visited. That is, it assumes
 that zeta-reduction (aka let-expansion) is going to be used.
 -/
 def isHeadBetaTarget (e : Expr) (useZeta := false) : Bool :=
-  e.getAppFn.isHeadBetaTargetFn useZeta
+  e.isApp && e.getAppFn.isHeadBetaTargetFn useZeta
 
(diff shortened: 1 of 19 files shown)


src/Lean/Expr.lean
```diff
@@ -1297,7 +1297,7 @@ If `useZeta = true`, then `let`-expressions are visited. That is, it assumes
 that zeta-reduction (aka let-expansion) is going to be used.
 -/
 def isHeadBetaTarget (e : Expr) (useZeta := false) : Bool :=
-  e.getAppFn.isHeadBetaTargetFn useZeta
+  e.isApp && e.getAppFn.isHeadBetaTargetFn useZeta
```
Kha, line 1300: Necessary change to ensure termination of `if isHeadBetaTarget e then .visit e.headBeta ...` uses

PR #1513, opened 2022-08-25
fix: collision between implicit and auto-bound level names

Kha:
As reported in https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.E2.9C.94.20universe.20coherence.20in.20.60forall_3_true_iff.60/near/295263725

### DIFF
--- a/src/Lean/Elab/MutualDef.lean
+++ b/src/Lean/Elab/MutualDef.lean
@@ -708,8 +708,9 @@ private def levelMVarToParamHeaders (views : Array DefView) (headers : Array Def
     let mut newHeaders := #[]
     for view in views, header in headers do
       if view.kind.isTheorem then
-        newHeaders := newHeaders.push { header with type :=
-          (← monadMap (@withLevelNames (levelNames := header.levelNames)) (levelMVarToParam' header.type)) }
+        newHeaders ←
+          withLevelNames header.levelNames do
+            return newHeaders.push { header with type := (← levelMVarToParam header.type), levelNames := (← getLevelNames) }
       else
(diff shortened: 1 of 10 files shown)


src/Lean/Elab/MutualDef.lean
```diff
@@ -708,7 +708,8 @@ private def levelMVarToParamHeaders (views : Array DefView) (headers : Array Def
     let mut newHeaders := #[]
     for view in views, header in headers do
       if view.kind.isTheorem then
-        newHeaders := newHeaders.push { header with type := (← levelMVarToParam' header.type) }
+        newHeaders := newHeaders.push { header with type :=
+          (← monadMap (@withLevelNames (levelNames := header.levelNames)) (levelMVarToParam' header.type)) }
```
Kha, line 712: Not entirely sure why the implicit lambda didn't work

leodemoura: It is worth investigating, but let's avoid `monadMap` here. We should use the idiom we have in the rest of the code.
```lean
def withLevelNamesImp (levelNames : List Name) (x : TermElabM α) : TermElabM α := do
...

def withLevelNames [MonadFunctorT TermElabM m] [Monad m] (levelNames : List Name) (x : m α) : m α := 
...
```

Kha: Okay, I had to restructure this part anyway to fix the last reported bug, which was due to not storing the new level names between independent invocations of `levelMVarToParam`. Fixing this removed the need for the state transformer, simplifying the code.

Issue #1485, opened 2022-08-15
Feature request: even more float functions


EdAyers:
I have more requests:

- `Float.abs : Float → Float` seems to be missing
- `Float.toInt : Float → Int`, where it returns a best-effort GMP integer if the float value wouldn't fit in to a scalar value.
- `Float.mantissa: Float → Int` and `Float.exponent : Float → Int` to return the mantissa and exponent parts of the float such that `mantissa * 2^exponent` is equal to the original float.

PR #1514, opened 2022-08-25
feat: Float.abs function

fynnsu:
This is an implementation of a Float.abs function as mentioned in #1485 .

### DIFF
--- a/stage0/src/Init/Data/Float.lean
+++ b/stage0/src/Init/Data/Float.lean
@@ -129,7 +129,6 @@ instance : ReprAtom Float  := ⟨⟩
 @[extern "ceil"] opaque Float.ceil : Float → Float
 @[extern "floor"] opaque Float.floor : Float → Float
 @[extern "round"] opaque Float.round : Float → Float
-@[extern "fabs"] opaque Float.abs : Float → Float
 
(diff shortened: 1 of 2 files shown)


tests/lean/floatAbs.lean
```diff
@@ -0,0 +1,7 @@
+def main : IO Unit := do
+IO.println ((3.14159 : Float).abs);
+IO.println ((-3.14159 : Float).abs);
+IO.println ((-2 : Float).abs);
+IO.println ((0 : Float).abs);
```
digama0, line 5: This should probably go in the compiler tests, so it is tested as both compiled and interpreted. I recommend adding it to `float.lean` next to the other float ops. Also, float ops should always be tested on infinities, NaN, and signed zero.

leodemoura:
Thanks for the PR. 
Could you please remove the `stage0` files from the PR? They are not needed.

tests/lean/floatAbs.lean
FynnSu: I moved the tests as requested. 

I wasn't sure if there is a way to explicitly input -inf or NaN as a float, so I used 1 / 0, and 0 / 0 respectively.

PR #1515, opened 2022-08-26
feat: `dbgStackTrace`

Kha:


Kha:
For usability, I think it would be a good idea to make the `Unit` parameter implicit in this kind of debug functions and rely on implicit lambdas. Any downsides?

leodemoura:
> For usability, I think it would be a good idea to make the `Unit` parameter implicit in this kind of debug functions and rely on implicit lambda insertion. Any downsides?

I will merge as-is. We can try some experiments with implicit lambdas and create another PR.

PR #1516, opened 2022-08-26
doc: `ptrAdr`, JSON-RPC

EdAyers:
- Docstrings for some basic char and string functions
- ~~A note on `ptrAddr`, including a link to the paper on it.~~
- Copy some docstrings and links from JSON-RPC and LSP specs.

### DIFF
--- a/src/Init/Util.lean
+++ b/src/Init/Util.lean
@@ -38,41 +38,6 @@ def dbgSleep {α : Type u} (ms : UInt32) (f : Unit → α) : α := f ()
 @[neverExtract, inline] def panicWithPosWithDecl {α : Type u} [Inhabited α] (modName : String) (declName : String) (line col : Nat) (msg : String) : α :=
   panic (mkPanicMessageWithDecl modName declName line col msg)
 
-/-!
-
-# The unsoundness of `ptrAddr`
-
-Reference: [Sealing Pointer-Based Optimizations Behind Pure Functions](https://arxiv.org/pdf/2003.01685.pdf)
-
-Note that we can not allow bare reference-equality in trusted Lean code.
-To see why, let `α` be `List String` and define `x` and `y` as follows:
-
-```lean
-def x := ["hello"]
-def y := x.map id
-
-#eval ptrEq x y -- false
-#eval x = y     -- true
-```
-
-Suppose that `ptrEq` was a safe definition (call it `evilEq`), then using definition of equality in Lean we can derive:
-
-```
-@[implementedBy ptrEq]
-opaque evilEq {α : Type u} (a b : α) : Bool
-
-theorem evil {α} : ∀ {a b : α}, (a = b) → (evilEq a b) = (evilEq a a)
-  | _, _, rfl => rfl
-
-#eval (evilEq x y) = (evilEq x x) -- false!
-```
-
-But now the evaluation of `(evilEq x y) = (evilEq x x)` contradicts the `evil` theorem!
-
-To create the sound version of `ptrEq` called `withPtrEq`, we have to provide the hypothesis `h : a = b → k() = true`,
-since now if the values are not pointer-equal, it will run the decision procedure `k` instead.
--/
-
 @[extern "lean_ptr_addr"]
 unsafe opaque ptrAddrUnsafe {α : Type u} (a : @& α) : USize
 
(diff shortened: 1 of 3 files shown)


src/Init/Util.lean
```diff
@@ -38,6 +38,41 @@ def dbgSleep {α : Type u} (ms : UInt32) (f : Unit → α) : α := f ()
 @[neverExtract, inline] def panicWithPosWithDecl {α : Type u} [Inhabited α] (modName : String) (declName : String) (line col : Nat) (msg : String) : α :=
   panic (mkPanicMessageWithDecl modName declName line col msg)
 
+/-!
+
+# The unsoundness of `ptrAddr`
+
+Reference: [Sealing Pointer-Based Optimizations Behind Pure Functions](https://arxiv.org/pdf/2003.01685.pdf)
```
gebner, line 45: I'm not sure we should cite this preprint in the light of #1500.  And we should certainly avoid any claims of safety here.

src/Lean/Data/JsonRpc.lean
```diff
@@ -107,6 +146,17 @@ structure Notification (α : Type u) where
 instance [ToJson α] : Coe (Notification α) Message :=
   ⟨fun r => Message.notification r.method (toStructured? r.param).toOption⟩
 
+/-- Generic version of `Message.response`.
+
+A Response Message sent as a result of a request. If a request doesn’t provide a
+ result value the receiver of a request still needs to return a response message
+ to conform to the JSON-RPC specification. The result property of the ResponseMessage
+ should be set to null in this case to signal a successful request.
```
gebner, line 154:
```suggestion
result value the receiver of a request still needs to return a response message
to conform to the JSON-RPC specification. The result property of the ResponseMessage
should be set to null in this case to signal a successful request.
```

src/Lean/Data/JsonRpc.lean
```diff
@@ -115,10 +165,19 @@ structure Response (α : Type u) where
...
+- [JSON-RPC](https://www.jsonrpc.org/specification#error_object)
+- [LSP](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#responseError).
+-/
 structure ResponseError (α : Type u) where
   id      : RequestID
   code    : ErrorCode
+  /-- A string providing a short description of the error. -/
   message : String
+  /-- A primitive or structured value that contains additional
+	    information about the error. Can be omitted.-/
```
gebner, line 180:
```suggestion
  /-- A primitive or structured value that contains additional
  information about the error. Can be omitted. -/
```
Please avoid tabs.

src/Init/Util.lean
EdAyers: ah sorry I missed the memo on this

src/Lean/Data/JsonRpc.lean
```diff
@@ -31,18 +33,36 @@ instance : ToString RequestID where
...
-  | contentModified
-  | requestCancelled
+  | /-- The server detected that the content of a document got
+	      modified outside normal conditions. A server should
+	      NOT send this error code if it detects a content change
+	      in it unprocessed messages. The result even computed
+	      on an older state might still be useful for the client.
+
+	      If a client decides that a result is not of any use anymore
+	      the client should cancel the request. -/
```
gebner, line 62:
```suggestion
  | /-- The server detected that the content of a document got
    modified outside normal conditions. A server should
    NOT send this error code if it detects a content change
    in it unprocessed messages. The result even computed
    on an older state might still be useful for the client.

    If a client decides that a result is not of any use anymore
    the client should cancel the request. -/
```

src/Lean/Data/JsonRpc.lean
```diff
@@ -31,18 +33,36 @@ instance : ToString RequestID where
...
+  | /-- The JSON sent is not a valid Request object. -/
+    invalidRequest
+  | /-- The method does not exist / is not available. -/
+    methodNotFound
+  | /-- Invalid method parameter(s). -/
+    invalidParams
+  | /-- Internal JSON-RPC error. -/
+    internalError
+  | /-- Error code indicating that a server received a notification or
+	      request before the server has received the `initialize` request. -/
```
gebner, line 51:
```suggestion
  | /-- Error code indicating that a server received a notification or
    request before the server has received the `initialize` request. -/
```

PR #1517, opened 2022-08-26
Show parser docstrings independent of info tree

Kha:
This change finally lets us document arbitrary syntax. More specifically, if there is a syntax node of a documented parser that's strictly inside the info node currently used for hover, we show that docstring instead.

TODO:
- [ ] Should go-to-definition do the same? Or should it jump to the innermost parser regardless of presence of docstrings?


Kha:
> * Should go-to-definition do the same? Or should it jump to the innermost parser regardless of presence of docstrings?

Probably the best way to answer details like this is to merge the PR as is and have people play around with it. There shouldn't be any real regressions in behavior I think.

gebner:
> More specifically, if there is a syntax node of a documented parser that's strictly inside the info node currently used for hover, we show that docstring instead.

What happens if we add an info node to `·` (as we should imho, in order to show the type of the `·`)?  Would the docstring be gone again?

Kha:
I assume the info node would be for the `cdot` syntax kind as well, so we would still show the same docstring

PR #1518, opened 2022-08-26
Fix parameter order for some functions

iacore:
This patch enables Elixir-style chaining for patched functions.

```
let x := Option.some 1 |> Option.getD 3
open Substring
let s := "hello world".toSubstring |> drop 6 |> toString
```

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -2110,15 +2110,20 @@ instance {α} : Inhabited (Option α) where
   default := none
 
 /--
+Get with default. Same semantic with `Option.getD`, but the caller need to pass in a lazy function as default value.
+-/
+def Option.getDLazy : (dflt : Unit → α) → Option α → α
+| _, some x => x
+| e, none   => e ()
+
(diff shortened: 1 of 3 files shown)


Kha:
This is not a problem using the more common style
```
  let x := some 1 |>.getD 3
  let s := "hello world".toSubstring.drop 6 |> toString
```

locriacyber:
If I chain more, the syntax becomes weird.

`"((hello world".toSubstring.drop 6).take 4).toString`
vs
`"hello world".toSubstring |> drop 6 |> take 4 |> toString`

Also, it's easier to have curried version of those functions.

Kha:
No, it doesn't become weird using the notation I suggested:
```lean
"hello world".toSubstring |>.drop 6 |>.take 4 |>.toString
```
We basically never open type namespaces like `Substring`, and we don't really use currying with them either - usually, when the expected type is available, `(·.take 4)` works as a shorthand of what would be `Substring.take 4` with the adjusted parameter order.

PR #1519, opened 2022-08-26
fix: use withContext at ac_rfl

digama0:
Fixes an issue reported on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.E2.9C.94.20ac_refl.20and.20intro).

### DIFF
--- a/src/Lean/Meta/Tactic/AC/Main.lean
+++ b/src/Lean/Meta/Tactic/AC/Main.lean
@@ -171,7 +171,7 @@ where
 
 @[builtinTactic acRfl] def acRflTactic : Lean.Elab.Tactic.Tactic := fun _ => do
   let goal ← getMainGoal
-  rewriteUnnormalized goal
+  goal.withContext <| rewriteUnnormalized goal
 
 builtin_initialize
   registerTraceClass `Meta.AC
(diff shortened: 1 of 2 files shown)


Kha:
Is there anything we can do to avoid this common footgun? Can we move it into a custom `MonadLiftT MetaM/... TacticM` instance, without incurring too much overhead?

digama0:
Is there a reason that we can't call this right before dispatching to a `@[tactic] def` plugin? Granted, some tactics don't work on the current goal, but it seems like it would be a good default behavior.

Kha:
If we want to do it eagerly, I think it would be best at `setGoals/...` for 100% consistency

Kha:
For the sake of completeness, a drastically different, explicit design would be not to provide an implicit lift from `MetaM` at all but to always require an explicit lift in respect to a given goal mvar. But since we are almost always intersted in the main goal only, that might be too onerous.

digama0:
We can't, because `withContext` is a wrapper-style tactic, and necessarily so since the relevant state is in the `Context` of the `MetaM` monad. So it would have to be done by whatever infrastructure dispatches to the user tactic. Any other solution would have the user explicitly writing the wrapper themselves, and we already have a fairly complete selection of such wrappers. To decrease the footgun propensity we would thus have to avoid functions like `getMainGoal` in favor of callback variations.

Kha:
Right, we can't do that building on top of `MetaM`. But basing `TacticM` on top of `MetaM` even though its information partially overlaps with the goal mvars is the fundamental issue we're discussing. It should not be too hard to change `TacticM` to carry `Meta.Context` as *state* instead and then reestablish the current behavior (modulo fixing the issue at hand) via explicit lift instances.

PR #1521, opened 2022-08-27
doc: update the mdbook instructions

lovettchris:
add info about leanink and alectryon

### DIFF
--- a/doc/dev/mdbook.md
+++ b/doc/dev/mdbook.md
@@ -57,21 +57,11 @@ Then run the following:
 executable to your `$HOME/.elan/bin` folder or `%USERPROFILE%\.elan\bin` so Alectryon can find it
 there.
 
-1. Create a Python 3.10 environment, if you use anaconda then that would be:
-    ```
-    conda create -n leanink python=3.10
-    ```
+1. Create a Python 3.10 environment.
 
-1. Activate this environment and install the following packages:
+1. Install the following packages:
     ```
-    pip install pygments dominate beautifulsoup4 docutils
-    ```
-
-1. Clone https://github.com/Kha/alectryon.git and run the following install the
-Alectryon python package:
-    ```
-    git checkout typeid
-    python setup.py install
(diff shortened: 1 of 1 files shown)


doc/dev/mdbook.md
```diff
@@ -1,3 +1,12 @@
+# Documentation
+
+The Lean `doc` folder contains the [Lean Manual](https://leanprover.github.io/lean4/doc/) and is
+authored in a combination of markdown (*.md) files and special lean programs.  The .lean files are
```
Kha, line 4:
```suggestion
authored in a combination of markdown (*.md) files and literate Lean files.  The .lean files are
```

doc/dev/mdbook.md
```diff
@@ -14,30 +23,94 @@ We are using the following settings while editing the markdown docs.
...
+    ```
+    pip install pygments dominate beautifulsoup4 docutils
+    ```
+
+1. Clone https://github.com/Kha/alectryon.git and run the following install the
+Alectryon python package:
+    ```
+    git checkout typeid
+    python setup.py install
+    ```
```
Kha, line 64: This is pretty complicated, what's wrong with the `pip` one-liner from the Alectryon docs?

lovettchris: What's complicated? The anaconda info?  The pip line is one line there?  It is important to mention it requires python 3.10 and that is usually easiest to do in a virtual environment, since it is not the default version on Ubuntu yet is it? (and what if they are not on Ubuntu?)  The git clone of and install alectryon is fine right?

Kha: Hard-coding the dependencies of a third-party project in our docs is just wrong. You make a good point about the environment (though I'd prefer to see that point in Alectryon's own docs), but even then would
```
python3 -m pip install git+https://github.com/Kha/alectryon.git@typeid
```
not work?

lovettchris: Much better, thanks, I updated the instructions.

PR #1522, opened 2022-08-27
add list docs

lovettchris:
similar to array.

### DIFF
--- a/doc/list.md
+++ b/doc/list.md
@@ -174,13 +174,13 @@ You can pop some elements off the front of a list
 #eval [1,2,3].take 2   -- [1, 2]
 ```
 
-And you can drop some times off the front and get the remainder:
+And you can drop some items off the front and get the remainder:
 
 ```lean
 #eval [1,2,3].drop 1   -- [2, 3]
 ```
 
-You can also erase a the first matching item:
+You can also erase the first matching item:
 
 ```lean
 #eval ["a", "b", "c", "b"].erase "b" -- ["a", "c", "b"]


doc/array.md
```diff
@@ -35,6 +35,13 @@ To create an array of size `n` in which all the elements are initialized to some
 -- #['a', 'a', 'a', 'a', 'a']
 ```
 
+To create an array from a range of numbers use `List.range` and convert the result to an Array:
```
leodemoura, line 38: This is not a good example. We should eventually have `Array.range` in the stdlib which avoids the overhead of creating a temporary list.

doc/array.md
```diff
@@ -73,5 +84,73 @@ produces a panic error message if the index `i` is out of bounds. The latter ret
...
 #eval #['a', 'b', 'c'][1]!
--- 'b!
+-- 'b'
+```
+
+## Concatenation
+
+You can push a new item to the end of an array:
+```lean
+#eval Array.push #[1, 2, 3] 7
```
leodemoura, line 94: I think we should emphasize the dot-notation in the manual. We rarely write `Array.push`. In the example above, we would write `#[1, 2, 3].push 7`.

doc/list.md
```diff
@@ -1 +1,195 @@
 # List
+
+A `List` looks similar to an [Array](Array.md) but it is a pure Lean inductive type that does not
```
leodemoura, line 3: Lists are a basic type in functional languages. It feels very weird to introduce them by making a connection with arrays.
I prefer you use the more standard style used by F#.
https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/lists

doc/list.md
```diff
@@ -1 +1,195 @@
...
+
+#check #[1, 2, 3] -- Array Nat - arrays have a # prefix.
+
+#check [] -- List ?m  -- unresolved type.
+#check List.nil (α := String) -- List String - specify the type parameter
+
+#eval List.cons "a" ["b", "c", "d"]     -- ["a", "b", "c", "d"]
+```
+
+Notice that Lean is able to infer the list element type in these cases, createing `List nat` and `List String`.
```
leodemoura, line 33: Typo: `List Nat`.

doc/list.md
```diff
@@ -1 +1,195 @@
...
+
+Notice also that `[]` is short hand for `List.nil` and that it is missing any list element type. To
```
leodemoura, line 35: The type is an implicit argument.

doc/list.md
```diff
@@ -1 +1,195 @@
...
+resolve this you can write the following:
+
+There is special syntax for `cons`, namely, `::` as follows:
+
+```lean
+#eval 0 :: [1, 2, 3]        -- [0, 1, 2, 3]
+```
+
+The type of the list elements is inferred from the literals used and must be consistent.
```
leodemoura, line 44: There is nothing special about `List.cons`. We should explain the type system in a different part of the manual once and for all.

doc/list.md
```diff
@@ -1 +1,195 @@
...
+```lean
+#check ["hello", "world"] -- List String
+
+-- The following is not valid, the List element type must be homogeneous
+#check_failure [10, "hello"]
+```
+Recall that the command `#check_failure <term>` only succeeds when the given term is not type correct.
+
+You can also invoke type inference on empty lists using parens to type the expression:
```
leodemoura, line 53: Type ascription should also be explained in a different part of the manual.

doc/list.md
```diff
@@ -1 +1,195 @@
...
+
+To create an list from a range of numbers use `List.range`:
+
+```lean
+#eval List.range 5      -- [0, 1, 2, 3, 4]
+```
+
+## Accessing elements
+
+You can access list elements by using square brackets `[` and `]`.
```
leodemoura, line 80: We rarely access list elements this way.
It also feels odd to explain the `a[i]` notation and its variants in the `List` section.

doc/list.md
```diff
@@ -1 +1,195 @@
...
+#eval f [1, 2, 3] -- This is a function application
+
+#eval as[2] -- This is an array access
+
+#eval [1, 2, 3][2]  -- 3
+```
+
+## Concatenation
+
+You can add a new item to the front of a list by constructing a new List using the `cons`
```
leodemoura, line 123: We usually use `7 :: [1, 2, 3]`.

leodemoura:
What about using F# documentation as a reference: https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/lists

doc/array.md
lovettchris: When that happens we can update this comment.  But right now folks will puzzle on how to do it and this is the best way for now.

doc/list.md
lovettchris: fixed

doc/list.md
lovettchris: I added a big warning note about dang erase quadratic performance from using these square brackets.  But we need to document them or remove them.

doc/list.md
lovettchris: Perhaps I found this information particularly missing on lists because there is a concept of empty container.  Non-container types don't really have this problem, except 'Option.none'.

doc/list.md
lovettchris: I don't understand this comment.  `List.cons` is clearly specific to lists, what is general type system info here?

doc/list.md
lovettchris: I reworded this as you suggested

doc/list.md
lovettchris: Fixed

doc/list.md
lovettchris: Got it, I've changed the intro accordingly.

doc/array.md
lovettchris: Done -- both here and in list.md

lovettchris:
I like the look of the F# documentation.  I think we can work towards that incrementally, method level stuff can probably come from docgen... interesting point here is the F# page combines high level conceptual stuff with docgen level api stuff.  So do we do the same and bulk up the list.lean code with tons of big paragraphs, or do we keep them separate, but merge them somehow with some kind of new leanink instruction like `{{insert docgen api here}}`...

doc/list.md
leodemoura: I was referring to the sentence 

> The type of the list elements is inferred from the literals used and must be consistent.

If one understands how the type system works, the sentence is redundant.
The second problem is that it uses non-standard terminology, we don't say the types are consistent.
I think all discussions regarding typing should go into a different session.

leodemoura:
> I like the look of the F# documentation. I think we can work towards that incrementally, method level stuff can probably come from docgen... interesting point here is the F# page combines high level conceptual stuff with docgen level api stuff. So do we do the same and bulk up the list.lean code with tons of big paragraphs, or do we keep them separate, but merge them somehow with some kind of new leanink instruction like `{{insert docgen api here}}`...

I am confused by this PR. In our last weekly meeting, I think I said we should use docstrings and module docstrings to document APIs. Then, I got the PR request review yesterday, and I assumed you wanted to do it using mdbook. I suggested the F# documentation because it is high quality, and I was concerned about the content and explanations in this PR would put Lean users in the wrong direction.

lovettchris:
> I am confused by this PR.

Well I still think there's a need for a quick conceptual overview on lists which is what this is trying to be.  It is the information I gleaned from various code snippets and samples along the way that I wish I had in this document from the beginning.  

The full doc generated pages from doc comments would look more like this: https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-listmodule.html

But if you think the content I added is wrong or leading people in the wrong direction then please tell me where.  If you think the whole thing is hopeless and need to go in the trash can then that's fine I'll cancel the PR.

I was just trying to help fill in some of the worst holes in the reference manual in time for the Jan class and I figured this would be an easy one since lists are pretty simple and I was copying the patten I saw in the existing document on arrays.md.

lovettchris:
Perhaps this could be a temporary stop gap measure - filling the hole we have in the lean4 manual on [Lists](https://leanprover.github.io/lean4/doc/list.html) until a real Lean Expert gets the time to rewrite it.

leodemoura:
@lovettchris Mario is making a lot of progress on the doc strings. We will soon have great doc strings for Lists. 
I will close this PR.

PR #1523, opened 2022-08-27
fix: @[inheritDoc] on notation

digama0:
This fixes the error when you use `@[inheritDoc] notation "foo" => bar`, but I notice that you get inheritDoc behavior even without the attribute, and indeed the docs on the notation are completely ignored in the hover for reasons I haven't figured out. @Kha do you know where notation hovers come from?


Kha:
If a term elaborates to a single constant, we always show its docstring https://github.com/leanprover/lean4/blob/0925051c51e183a0b8c504ae53f5cab0525d6cac/src/Lean/Server/InfoUtils.lean#L189-L191

Kha (approved):

PR #1526, opened 2022-08-27
feat: add list.indexOf?

lovettchris:
I found this helper method handy while writing the section on monads and every list class I've used in other languages has this.

### DIFF
--- a/src/Init/Data/List/Basic.lean
+++ b/src/Init/Data/List/Basic.lean
@@ -227,11 +227,14 @@ def findSome? (f : α → Option β) : List α → Option β
     | some b => some b
     | none   => findSome? f as
 
-def indexOf? [BEq α] (s : α) (i := 0): List α → Option Nat
+private def indexOfAux? [BEq α] (s : α) (i := 0): List α → Option Nat
   | [] => none
   | a :: tail => match a == s with
     | true => some i
-    | false => indexOf? s (i+1) tail
+    | false => indexOfAux? s (i+1) tail
+
+def indexOf? [BEq α] (as : List α) (s : α) : Option Nat :=
+  as.indexOfAux? s 0
 
 def replace [BEq α] : List α → α → α → List α
   | [],    _, _ => []


src/Init/Data/List/Basic.lean
```diff
@@ -227,6 +227,12 @@ def findSome? (f : α → Option β) : List α → Option β
     | some b => some b
     | none   => findSome? f as
 
+def indexOf? [BEq α] (s : α) (i := 0): List α → Option Nat
```
digama0, line 230: I think this parameter `(i := 0)` needs to be in an auxiliary, otherwise it will need to be passed explicitly when you use the function without dot notation.

Also there should be a space after the binder.

lovettchris: I added indexOfAux?, but didn't understand the second comment "binding" ? what binder?

lovettchris: And how do you hide the Aux methods from intellisense in VS code ?  I tried making it private but that didn't seem to help...

image

digama0: There's no need to hide the method, it's legal for people to call. If you use an inner function it will get a name like `indexOf?.go`, not sure how that shows up in the dropdown.

Also note that you are getting text suggestions there (as indicated by the "abc"); this is some vscode magic sauce that involves things you have written recently and words in the file so I wouldn't put too much stock in it. The better question is what the lean suggestions look like (and why you aren't getting them in that case).

The binder `(i := 0)` needs a space after it, i.e. `(i := 0) :`. It probably doesn't matter after the other change anyway.

leodemoura:
As discussed in the last dev update meeting, we are creating a separate repo for the stdlib. Mario volunteered to be the lead maintainer. See: https://docs.google.com/presentation/d/1YdVJbdZg8JESBzME8lb9NPUNHFInLrpXyAn32EahJBM/edit#slide=id.g134547f0559_0_277

This PR should go to the new repo that will be created soon.

leodemoura:
> The better question is what the lean suggestions look like (and why you aren't getting them in that case).

@digama0 I hit this problem all the time when I am using VS Code. BTW, it seems to happen only on VS Code. It seems to work more reliably on Emacs. @Kha also has a better experience with Emacs. 
See this thread: https://leanprover.zulipchat.com/#narrow/stream/147302-lean4-dev/topic/Completion/near/292348639

leodemoura:
We now have https://github.com/leanprover/std4
@lovettchris Could you please submit this PR to this new repo?

> I found this helper method handy while writing the section on monads and every list class I've used in other languages has this.

Regarding the section on Monads, please use a different function.

PR #1528, opened 2022-08-28
chore: make NonScalar opaque

digama0:
This makes the type a bit less error-prone, by not exposing the constructor or eliminator for the type (which is almost certainly an error given that elements of this type are generally constructed by `unsafeCast` and so the eliminator is essentially a lie).

It also defines `NonScalar` as `PNonScalar.{0}`, similar to other universe polymorphic types.

### DIFF
--- a/src/Init/Core.lean
+++ b/src/Init/Core.lean
@@ -167,13 +167,24 @@ protected def bind {α : Type u} {β : Type v} (x : Task α) (f : α → Task β
 
 end Task
 
-/-- Some type that is not a scalar value in our runtime. -/
-structure NonScalar where
-  val : Nat
+private inductive PNonScalarSpec : Type u where
+  | mk (v : Nat)
+
+opaque nonScalarSpec : NonemptyType.{u} := ⟨PNonScalarSpec, ⟨⟨0⟩⟩⟩
 
 /-- Some type that is not a scalar value in our runtime and is universe polymorphic. -/
-inductive PNonScalar : Type u where
-  | mk (v : Nat) : PNonScalar
+def PNonScalar : Type u := nonScalarSpec.type
+
+instance : Nonempty PNonScalar := nonScalarSpec.property
+
+@[inline] unsafe def PNonScalar.null : PNonScalar := unsafeCast (0 : Nat)
+
+/-- Some type that is not a scalar value in our runtime. -/
+def NonScalar : Type := PNonScalar.{0}
+
+instance : Nonempty NonScalar := nonScalarSpec.property
+
+@[inline] unsafe def NonScalar.null : NonScalar := PNonScalar.null
 
 @[simp] theorem Nat.add_zero (n : Nat) : n + 0 = n := rfl
 
@@ -514,7 +525,7 @@ abbrev noConfusionEnum {α : Sort u} {β : Sort v} [inst : DecidableEq β] (f : 
 instance : Inhabited Prop where
   default := True
 
-deriving instance Inhabited for NonScalar, PNonScalar, True, ForInStep
+deriving instance Inhabited for True, ForInStep
 
 theorem nonempty_of_exists {α : Sort u} {p : α → Prop} : Exists (fun x => p x) → Nonempty α
   | ⟨w, _⟩ => ⟨w⟩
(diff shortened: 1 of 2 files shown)


src/Init/Core.lean
```diff
@@ -167,13 +167,14 @@ protected def bind {α : Type u} {β : Type v} (x : Task α) (f : α → Task β
...
-/-- Some type that is not a scalar value in our runtime. -/
-structure NonScalar where
-  val : Nat
+private inductive PNonScalarSpec : Type u where
+  | mk (v : Nat)
 
 /-- Some type that is not a scalar value in our runtime and is universe polymorphic. -/
-inductive PNonScalar : Type u where
-  | mk (v : Nat) : PNonScalar
+opaque PNonScalar : Type u := PNonScalarSpec
```
gebner, line 174: We should probably have a `Nonempty PNonScalar.{u}` instance.

digama0: I wanted to see if we could avoid that, but the build errors are saying no. I've introduced a `Nonempty PNonScalar` as well as `unsafe def PNonScalar.null := unsafeCast (0 : Nat)` for use in the array functions.

src/Init/Core.lean
```diff
@@ -167,13 +167,24 @@ protected def bind {α : Type u} {β : Type v} (x : Task α) (f : α → Task β
 
 end Task
 
-/-- Some type that is not a scalar value in our runtime. -/
-structure NonScalar where
-  val : Nat
+private inductive PNonScalarSpec : Type u where
+  | mk (v : Nat)
+
+opaque nonScalarSpec : NonemptyType.{u} := ⟨PNonScalarSpec, ⟨⟨0⟩⟩⟩
```
gebner, line 173:
```suggestion
opaque nonScalarSpec : NonemptyType.{u}
```

digama0: That doesn't look like a behavior-preserving change, since it will implement `NonScalar` as `Unit` instead of `Nat`. Is `Unit` actually a non-scalar type?

gebner: At least currently, this doesn't make any difference since the compiler doesn't unfold opaque types.  We use this pattern quite extensively (just grep for `opaque.*Nonempty`).

With the new compiler, we might want to change that and better take advantage of the types.  But then `⟨PNonScalarSpec, ⟨⟨0⟩⟩⟩` would be just as wrong (because `PNonScalarSpec` can only contain scalar and mpz values, but `PNonScalar` should obviously contain more).

Come to think of it, the name `PNonScalar` is extremely misleading.  We use it not just for non-scalar values, but for *potentially* non-scalar values (i.e. including scalars).

Kha: I believe *scalar* here means "unboxed integer type", i.e. not `Nat`

digama0: > At least currently, this doesn't make any difference since the compiler doesn't unfold opaque types. We use this pattern quite extensively (just grep for `opaque.*Nonempty`).

I'm aware that we use `opaque.*Nonempty` a lot, but my understanding was that this was for blocking the kernel from seeing through the type alias while still letting the compiler use the same runtime representation. That is, I would have expected an opaque wrapper around `UInt8` to still have a `uint8_t` ABI.

digama0: > Come to think of it, the name `PNonScalar` is extremely misleading. We use it not just for non-scalar values, but for _potentially_ non-scalar values (i.e. including scalars).

Everything should be "boxed" in this type, including scalars. That may not mean that there is actually a pointer indirection because of the nat optimization, but even if the value is "actually" a `UInt8` it should still be passed as a `lean_object*`. So I don't think the name is that misleading, although it is subtle and perhaps another name is more suitable (what's a scalar anyway? we don't use this terminology anywhere else and we use it incompatibly in mathlib).

digama0: I think a better name might be something like `OpaqueBox` or `OpaquePointer`. Something more evocative of `void*`.

gebner: I got the scalar name from `lean_is_scalar`.  It is indeed confusing if the terminology is used for unboxed values as well.

I would suggest to rename `NonScalar` to `Object` for consistency with `ObjectMap`.

Kha: We also have `tobject/object` in https://leanprover-community.github.io/mathlib4_docs/Lean/Compiler/IR/Basic.html#Lean.IR.IRType

digama0: I'm not a fan of grabbing such prime real estate as `_root_.Object` for this niche thing. Can we at least namespace it? Or use `LCObject` to go with the other `lcFunction`s.

gebner: Fwiw it's `Std.ObjectMap` so `Std.Object` would be consistent.

digama0: `Std.ObjectMap` is *in* Std though. `NonScalar` is in `Init` because it is used by some array ops.

gebner: `Std.Format` is also in init. `Lean.Syntax` too.

PR #1535, opened 2022-08-29
chore: remove BinomialHeap, DList, Stack, Queue

digama0:
* [x] depends on #1534

This removes the types `Std.Data.{BinomialHeap, DList, Stack, Queue}` which are not used in Lean itself. They will be moved to the new std4 repository.


leodemoura:
Merged manually.

PR #1541, opened 2022-08-29
chore: move Bootstrap.Data -> Lean.Data

digama0:
* [x] depends on #1537
* [x] depends on #1539

PR #1542, opened 2022-08-30
chore: move Bootstrap.Dynamic -> Init.Dynamic

digama0:
* [x] depends on #1539
* [x] depends on #1541

This moves `Bootstrap.Dynamic` to `Init` and moves `Std.TypeName` and `Std.Dynamic` to the root namespace. This is the last file remaining in `Bootstrap`, and a subsequent PR will remove it entirely.


leodemoura:
Forcing the merge. The failure is not related to the PR.

PR #1544, opened 2022-08-30
refactor: Init.SimpLemmas proof golf / cleanup

digama0:
This file may need another pass later when these lemmas are merged with the other ones from logic.basic or the equivalent. In particular, many of these are stated as equalities, which IIUC is not required by `simp`; we will want the theorems with names like `iff_false` to be iff statements, so these should either be renamed to `iff_false_eq` or we should just make them iff statements and rely on `simp` to add the `propext`.

### DIFF
--- a/src/Init/SimpLemmas.lean
+++ b/src/Init/SimpLemmas.lean
@@ -7,130 +7,112 @@ notation, basic datatypes and type classes
 -/
 prelude
 import Init.Core
+set_option linter.missingDocs true -- keep it documented
 
-@[simp] theorem eq_self (a : α) : (a = a) = True :=
-  propext <| Iff.intro (fun _ => trivial) (fun _ => rfl)
-
-theorem of_eq_true (h : p = True) : p :=
-  h ▸ trivial
+theorem of_eq_true (h : p = True) : p := h ▸ trivial
 
 theorem eq_true (h : p) : p = True :=
-  propext <| Iff.intro (fun _ => trivial) (fun _ => h)
+  propext ⟨fun _ => trivial, fun _ => h⟩
 
 theorem eq_false (h : ¬ p) : p = False :=
-  propext <| Iff.intro (fun h' => absurd h' h) (fun h' => False.elim h')
+  propext ⟨fun h' => absurd h' h, fun h' => False.elim h'⟩
 
-theorem eq_false' (h : p → False) : p = False :=
-  propext <| Iff.intro (fun h' => absurd h' h) (fun h' => False.elim h')
+theorem eq_false' (h : p → False) : p = False := eq_false h
 
 theorem eq_true_of_decide {p : Prop} {_ : Decidable p} (h : decide p = true) : p = True :=
-  propext <| Iff.intro (fun _ => trivial) (fun _ => of_decide_eq_true h)
+  eq_true (of_decide_eq_true h)
 
 theorem eq_false_of_decide {p : Prop} {_ : Decidable p} (h : decide p = false) : p = False :=
-  propext <| Iff.intro (fun h' => absurd h' (of_decide_eq_false h)) (fun h => False.elim h)
+  eq_false (of_decide_eq_false h)
+
+@[simp] theorem eq_self (a : α) : (a = a) = True := eq_true rfl
 
 theorem implies_congr {p₁ p₂ : Sort u} {q₁ q₂ : Sort v} (h₁ : p₁ = p₂) (h₂ : q₁ = q₂) : (p₁ → q₁) = (p₂ → q₂) :=
   h₁ ▸ h₂ ▸ rfl
 
-theorem implies_congr_ctx {p₁ p₂ q₁ q₂ : Prop} (h₁ : p₁ = p₂) (h₂ : p₂ → q₁ = q₂) : (p₁ → q₁) = (p₂ → q₂) :=
-  propext <| Iff.intro
-    (fun h hp₂ =>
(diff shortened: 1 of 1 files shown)


leodemoura:
>  In particular, many of these are stated as equalities, which IIUC is not required by simp

Yes, this is true, `simp` supports `<->` theorems, but it creates an auxiliary theorem internally. This file is adding theorems for `simp` not for humans.

src/Init/SimpLemmas.lean
```diff
@@ -7,142 +7,126 @@ notation, basic datatypes and type classes
...
+set_option linter.missingDocs true -- keep it documented
 
-@[simp] theorem eq_self (a : α) : (a = a) = True :=
-  propext <| Iff.intro (fun _ => trivial) (fun _ => rfl)
-
-theorem of_eq_true (h : p = True) : p :=
-  h ▸ trivial
+theorem of_eq_true (h : p = True) : p := h ▸ trivial
 
 theorem eq_true (h : p) : p = True :=
```
leodemoura, line 14: What is the motivation for this kind of change?

leodemoura: For the one above, it is at least avoiding the `propext`, which is a bit irrelevant if we consider the fact the theorem is used only by `simp`, and if you use `simp`, you are admitting you have no problem with axioms such as `propext`.

digama0: I'm not sure what "this kind of change" is referring to here. I don't think I changed any of the proofs in any essential way - they all still have the same statements and use `propext` in all the same places.

gebner:
> Yes, this is true, `simp` supports `<->` theorems, but it creates an auxiliary theorem internally. This file is adding theorems for `simp` not for humans.

All of the theorem names in this file are in the root namespace, and use the most prominent names.

If a theorem is named `and_self` then it is the first choice for (human) users, and should have the user-friendly type `a ∧ a ↔ a`.  We typically don't use equality for propositions.

One option could be to put these theorems into a separate namespace.  But I think that's a bad idea.  You would expect `squeeze_simp` to return `simp only [and_self]` and not `simp only [SimpTheorems.and_self]`.

leodemoura:
> > Yes, this is true, `simp` supports `<->` theorems, but it creates an auxiliary theorem internally. This file is adding theorems for `simp` not for humans.
> 
> All of the theorem names in this file are in the root namespace, and use the most prominent names.
> 
> If a theorem is named `and_self` then it is the first choice for (human) users, and should have the user-friendly type `a ∧ a ↔ a`. We typically don't use equality for propositions.
> 
> Another option could be to put these theorems into a separate namespace. But I think that's a bad idea. You would expect `squeeze_simp` to return `simp only [and_self]` and not `simp only [SimpTheorems.and_self]`.

Note that if we write the `simp` theorem `a ∧ a ↔ a`, Lean internally creates an auxiliary theorem using `=`, and this is the one that is actually used in the proof. It will not be very `squeeze_simp` friendly. 
I see the following options
- Use the suffix `_eq`. The theorem would be called `and_self_eq`. So, it is clear it is not a theorem using `<->`, and `squeeze_simp` would use a reasonable name.
- Add a mechanism for collecting the original theorem name for `squeeze_simp`. I don't like this one. It is extra work, and `simp only [and_self]` would generate the auxiliary theorem again.
- Add support at `simp` for rewriting using `<->`. I dislike this option too. It is just extra work and more bookkeeping. One can claim that constructive users would benefit from it, but I think this ship has already sailed. Lean is not a system for people that cares about constructivism.

src/Init/SimpLemmas.lean
leodemoura: I was referring to style changes.

digama0:
> Add a mechanism for collecting the original theorem name for squeeze_simp. I don't like this one. It is extra work, and simp only [and_self] would generate the auxiliary theorem again.

My understanding was that `simp` would apply `propext` on the spot, rather than constructing an auxiliary theorem. Maybe it's being inlined? In any case, it is not something `squeeze_simp` has trouble handling. So as far as I am concerned, "Add support at simp for rewriting using <->." already exists, using `propext`. (I think that being able to rewrite with relations *is* useful, not just for `<->` but more for custom relations like `modeq` or `equiv`. But it's fine to put this off for a while.)

gebner:
> Lean internally creates an auxiliary theorem using `=`, and this is the one that is actually used in the proof. It will not be very `squeeze_simp` friendly.

I know.  And in mathlib4 we already have code to revert this mapping, because we need it for the simp linter. :smile: https://github.com/leanprover-community/mathlib4/blob/90004fda28ba90e80c9930d1bacf2aeaf28bc3a5/Mathlib/Tactic/Lint/Simp.lean#L81-L111

> Add a mechanism for collecting the original theorem name for `squeeze_simp`. I don't like this one. It is extra work,

Even if we consistently use `=` in this file, users want to and will put `@[simp]` on theorems with `↔`.  We need some way to get the original theorem name.  We already have an (ugly and slow) way to get it.  Storing the original theorem name in an environment extension is merely a performance optimization.

leodemoura:
> I know. And in mathlib4 we already have code to revert this mapping, because we need it for the simp linter. 😄 https://github.com/leanprover-community/mathlib4/blob/90004fda28ba90e80c9930d1bacf2aeaf28bc3a5/Mathlib/Tactic/Lint/Simp.lean#L81-L111

Oh wow, I didn't know you were reverse engineering the process. Please, tell me about this in the future. 
I just talked with @digama0 about this. I will create an issue for addressing this problem.

PR #1545, opened 2022-08-30
feat: add TR versions of Nat.{fold, any, all, repeat}

digama0:
This changes the definitions of `Nat.{fold, any, all, repeat}` to be the "naive" versions and renames the existing ones with a `TR` (tail-recursive) suffix and a `[csimp]` lemma proving the equality.

I suspect these definitions are not actually being used in lean core, since `Prod.allI` was off by a negation, in which case they may be candidates for moving to std.

### DIFF
--- a/src/Init/Data/List/Basic.lean
+++ b/src/Init/Data/List/Basic.lean
@@ -143,7 +143,7 @@ theorem reverseAux_eq_append (as bs : List α) : reverseAux as bs = reverseAux a
   | nil => simp
   | cons a as ih => simp [ih]; rw [append_assoc]
 
-theorem mapTRAux_eq (f : α → β) (as : List α) (bs : List β) : mapTRAux f as bs =  bs.reverse ++ map f as := by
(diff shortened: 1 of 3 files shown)


leodemoura:
> I suspect these definitions are not actually being used in lean core, since Prod.allI was off by a negation, in which case they may be candidates for moving to std.

I agree. Could you please confirm?

digama0:
Of the three Prod definitions at the end, the only one in use is a single use of `Prod.anyI` inside `Lean.IR.EmitC.overwriteParam`. It's a bit of a weird function, so I'm inclined to just inline it and delete all three.

Of the other definitions touched here:
* `Nat.fold` has 12 uses
* `Nat.foldRev` has 2 uses
* `Nat.any` has 5 uses
* `Nat.all` has 1 use
* `Nat.repeat` has 2 uses

leodemoura:
@digama0 Thanks for checking. There are many uses, let's keep the definitions here for now. I will merge the PR as-is.

PR #1548, opened 2022-08-30
feat: use RPC method from first snapshot

gebner:
Fixes leanprover/vscode-lean4#239.  As suggested by @Vtec234 in https://github.com/leanprover/vscode-lean4/pull/244#issuecomment-1228292682

The linked issue happens because the RPC call request tries to get the snapshot at the cursor position, in order to find RPC methods defined by `@[serverRpcMethod]` in the current file.  However there are no snapshots after `import` (if it fails), or after `#exit`.

This PR uses the first snapshot that contains the RPC method.  If the RPC method is built-in, it is called directly without waiting for snapshots.  A similar fix is applied to the `getWidgets` requests, which also failed if there were no snapshots under the cursor.

While going through the code, I also noticed that many `AsyncList` methods were in `BaseIO` unnecessarily.  I've made them pure.

### DIFF
--- a/src/Lean/Widget/UserWidget.lean
+++ b/src/Lean/Widget/UserWidget.lean
@@ -95,17 +95,20 @@ structure GetWidgetSourceParams where
   pos : Lean.Lsp.Position
   deriving ToJson, FromJson
 
-open Server in
+open Server RequestM in
 @[serverRpcMethod]
-def getWidgetSource (args : GetWidgetSourceParams) : RequestM (RequestTask WidgetSource) :=
-  RequestM.withWaitFindSnapAtPos args.pos fun snap => do
-    RequestM.runCoreM snap do
-      let env ← getEnv
-      if let some id := widgetSourceRegistry.getState env |>.find? args.hash then
-        let d ← getUserWidgetDefinition id
-        return {sourcetext := d.javascript}
+def getWidgetSource (args : GetWidgetSourceParams) : RequestM (RequestTask WidgetSource) := do
+  let doc ← readDoc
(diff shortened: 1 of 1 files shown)


src/Lean/Widget/UserWidget.lean
```diff
@@ -145,7 +145,7 @@ def getWidgets (args : Lean.Lsp.Position) : RequestM (RequestTask (GetWidgetsRes
   let doc ← readDoc
   let filemap := doc.meta.text
   let pos := filemap.lspPosToUtf8Pos args
-  withWaitFindSnapAtPos args fun snap => do
+  withWaitFindSnap doc (·.endPos >= pos) (notFoundX := return ⟨∅⟩) fun snap => do
```
Vtec234, line 148: Could we also do an early lookup at getWidgetSource?

gebner: We could, but it is not as critical here.  1) We do the getWidgetSource request right after the getWidgets requests has finished.  At that point the snapshot is available.  2) The client caches the javascript on first use.

Vtec234: What about the following situation: the getWidgets request resolves and comes back with a source id to the editor. In the meantime the user makes some changes, so the snapshot is gone. Then getWidgetSource is called and has to wait. If it's just the same simple change to the predicate, I'd say let's do it in this PR.

gebner: It's of course a possible behavior.  I have changed getWidgetSource in the same way.

Vtec234 (approved):

PR #1550, opened 2022-08-30
Basic Pass Manager

hargoniX:
Make the code generation procedure extensible by adding a basic pass manager,
this will allow users to write custom compiler passes and enable us to write
tests for our own compiler passes in a nice way.

### DIFF
--- a/src/Lean/Compiler/LCNF/Main.lean
+++ b/src/Lean/Compiler/LCNF/Main.lean
@@ -5,6 +5,7 @@ Authors: Leonardo de Moura
 -/
 import Lean.Compiler.Options
 import Lean.Compiler.LCNF.PassManager
+import Lean.Compiler.LCNF.Passes
 import Lean.Compiler.LCNF.PrettyPrinter
 import Lean.Compiler.LCNF.ToDecl
 import Lean.Compiler.LCNF.Check
--- a/src/Lean/Compiler/LCNF/Passes.lean
+++ b/src/Lean/Compiler/LCNF/Passes.lean
@@ -1,0 +1,20 @@
+/-
+Copyright (c) 2022 Henrik Böving. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Henrik Böving
+-/
+import Lean.Compiler.LCNF.PassManager
+import Lean.Compiler.LCNF.PullLetDecls
+import Lean.Compiler.LCNF.CSE
+
+namespace Lean.Compiler.LCNF
+
+@[cpass]
+def pullInstancesInstaller : PassInstaller :=
+  .installAfter `init (.mkPerDeclaration `pullInstances Decl.pullInstances)
+
+@[cpass]
+def cseInstaller : PassInstaller :=
+  .installAfter `pullInstances (.mkPerDeclaration `cse Decl.cse)
+
+end Lean.Compiler.LCNF


src/Lean/Compiler/LCNF/Main.lean
```diff
@@ -47,20 +49,29 @@ def checkpoint (stepName : Name) (decls : Array Decl) : CompilerM Unit := do
...
+namespace PassManager
+
+def run (declNames : Array Name) : CompilerM (Array Decl) := do
+  let declNames ← declNames.filterM (shouldGenerateCode ·)
+  if declNames.isEmpty then return #[]
+  let mut decls ← declNames.mapM toDecl
+  let mut manager := { passes := #[{ name := `init, run := pure }] }
+  let installers := PassInstaller.passInstallerExt.getState (←getEnv)
+  manager ← installers.foldlM (init := manager) PassInstaller.runFromDecl
+  trace[Compiler] s!"Running the following passes: {manager.passes.map (·.name)}"
```
Kha, line 61: Probably better to add a separate message for each pass just before running it

src/Lean/Compiler/LCNF/PassManager.lean
```diff
@@ -0,0 +1,103 @@
...
+  install : Array Pass → CompilerM (Array Pass)
+  deriving Inhabited
+
+structure PassManager where
+  passes : Array Pass
+  deriving Inhabited
+
+namespace Pass
+
+def perDeclaration (name : Name) (run : Decl → CompilerM Decl) : Pass :=
```
Kha, line 29: style nit: you could use `def ... where` for these

src/Lean/Compiler/LCNF/Passes.lean
```diff
@@ -0,0 +1,20 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Henrik Böving
+-/
+import Lean.Compiler.LCNF.PassManager
+import Lean.Compiler.LCNF.PullLetDecls
+import Lean.Compiler.LCNF.CSE
+
+namespace Lean.Compiler.LCNF
+
+@[cpass]
```
leodemoura, line 12: I was expecting this line to compile only after the `update-stage0`.
How did you manage to compile this commit before the `update-stage0`?

hargoniX: I didn't I just already had some parts of the code staged so i unstaged the stage0 stuff and did the full commit of the code first, I can flip them if you want.

src/Lean/Compiler/LCNF/PassManager.lean
```diff
@@ -0,0 +1,101 @@
...
+  let info ← getConstInfo declName
+  match info.type with
+  | .const `Lean.Compiler.LCNF.PassInstaller .. =>
+    modifyEnv fun env => passInstallerExt.addEntry env declName
+  | _ =>
+    throwError "invalid 'cpass' only 'PassInstaller's can be added via the 'cpass' attribute: {info.type}"
+
+builtin_initialize
+  registerBuiltinAttribute {
+    name  := `cpass
```
bollu, line 76: Quick question: does `cpass` stand for "compiler pass"? If so, can we name the attribute `compilerPass`, favoring descriptiveness over brevity? :)

hargoniX: Yes, I mostly named it this way because we already have `csimp` which stands for compiler simplification (rule) so I wanted to stay in order with that but we can change it of course.

src/Lean/Compiler/LCNF/Main.lean
```diff
@@ -47,20 +49,29 @@ def checkpoint (stepName : Name) (decls : Array Decl) : CompilerM Unit := do
...
+  for pass in manager.passes do
+    decls ← pass.run decls
+    checkpoint pass.name decls
+  saveStage1Decls decls
+  decls.forM fun decl => do trace[Compiler.stat] "{decl.name} : {decl.size}"
```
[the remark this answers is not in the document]
hargoniX: Moved it into the checkpoint.

src/Lean/Compiler/LCNF/PassManager.lean
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+end Pass
+
+namespace PassInstaller
+
+def installAtEnd (p : Pass) : PassInstaller :=
+  { install := fun ps => return ps.push p }
+
+def installAfter (target : Name) (p : Pass) : PassInstaller :=
```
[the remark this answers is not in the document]
hargoniX: Leo and I initially discussed this and had the idea to register multiple passes of the same type under different names if necessary, e.g. simp-1 simp-2 ..., what do you think about this proposal @leodemoura ?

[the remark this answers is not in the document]
leodemoura: I think this is a possibility. We can do it in another PR.

PR #1551, opened 2022-08-31
feat: LSP Code actions [WIP]

### LINKED ISSUE #1494: RFC: Code actions
Hi I've collected some thoughts on how to proceed with making a good code completions / actions interface for Lean:

# Scope

This RFC intends to cover an API for the protocol between Lean and an editor (initially vscode-lean4) for handling code actions. By code action, I mean the process of Lean presenting suggested semantic code changes to the user via the editor or infoview interface. Examples are ([source](https://github.com/leanprover/lean4/issues/1223#issuecomment-1156836435)): 

- Suggested expressions to fill `_` holes with
- More advanced suggestions with `{! ... !}`-style holes as described by @david-christiansen [here](https://github.com/leanprover/lean4/issues/1223#issuecomment-1159050274).
- Suggestions to fix syntax errors and elaboration errors.  
- From the infoview; contextual suggestions where the user clicks on a certain part of the goal state in the infoview, causing the tactic state to update. Examples are clicking on a subexpression in the infoview and generating a `conv at` or suggesting potential rewrite rules for a particular position.
- Replace a tactic by another tactic. For example: `squeeze_simp [...]`.
-  Insert a tactic. For example: `library_search`.
- Replace an expression by another expression. For example: `by show_term { exact ... }`.
- Insert an expression. For example: insert structure instance stub, `by library_search`.
- Replace (non-expression) syntax by another syntax. For example: unfold macro.

Out of scope: code formatting, snippets.

# Existing work and discussion

- [LSP code actions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction),
- [ `CodeActionProvider` in VSCode extension API](https://code.visualstudio.com/api/references/vscode-api#CodeActionProvider%3CT%3E)
- Lean 3 VSCode extension: [Hole support](src/holes.ts), "Try this" feature.
- Previous discussions: https://github.com/leanprover/vscode-lean4/issues/45, https://github.com/leanprover/lean4/issues/1223
- @kasequark (Robin Bohne ) is [working on a Conv-zoom command](https://github.com/KaseQuark/lean4).

## Summary of existing discussion

- It's important that CPU-intensive suggestions for code actions (eg finding rewrites, calling a hammer) are only triggered upon a user action explicitly for suggesting code actions.
- There is a worry that there will be cases where there are too many suggestions, in particular for unbounded cases where you are suggesting potential terms to fill a hole or lemmas to close a goal.
- @david-christiansen [notes](https://github.com/leanprover/lean4/issues/1223#issuecomment-1157305262) that there is some potential to include the more innovative Agda-style holes.
- @digama0 was sceptical about `{! !}` holes at least in terms of notation and calls for a retrospective. Leo says this is because it was underdeveloped
- @Kha [likes the snippets in rust-analyser](https://github.com/leanprover/lean4/issues/1223#issuecomment-1158899320), where you get the action using a snippet-style workflow when you type `match`. @gebner [is less keen](https://github.com/leanprover/lean4/issues/1223#issuecomment-1159003971).

# Design proposal

## Available mechanisms for implemention of code actions

- The [LSP Code Actions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction) system. That is, the 'lightbulbs'.
- The [LSP completions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) system. That is, they appear in the 'intellisense' completion tooltip.
- A custom API, eg something like the [contextual suggestions RPC](https://github.com/leanprover/lean4/issues/1223). The Lean server then modifies the source document directly.
- When a tactic is elaborated, they can dump a user-widget to the infotree. Users can interact with this custom UI and eventualy cause a call an RPC method that the Lean server runs.

I think there are cases where you want to use all of the above. I think the code actions give the best solution to the hinting/replacing style tactics such as `squeeze_simp`. The right way to implement these I think is to get the tactic to emit a code action diagnostic that is stored on the infotree. A similar approach can be used when holes are elaborated. In addition to the editor-specific code actions menu, these actions could also be presented in the infoview to make them more promenant.

 
The difference between completions and code actions seems to be that code actions are present as solutions to diagnostics ('squigglies'), whereas completions appear as you type new code. These completions can make arbitrary text edits, so we can do things like get `match x with` to suggest a completion where it autofills all of the cases. I think the rule should be that if you are _inserting_ a new command such as while typing a term or a tactic script, it should give completions, otherwise if it is _replacing_ it should use code actions.

## Modifying the document.

Currently, tactics can also put a user-widget on the infotree. User-widgets are able to send a command to the editor requesting a text edit, but it is currently clunky and isn't syntax aware.

I think that the best approach here is for Lean to provide a function `diff : Syntax → Syntax → TextEdit` or something similar, where the first argument is the existing syntax (stored on the infotree) and the second argument is the new syntax, the `TextEdit` object is then sent to the client editor. This is better than the Lean server editing the file directly because the resulting edit will not appear in the user's undo stack, and there may be some complications with buffers getting out of sync with the file on disk.

In the case of code actions and code completions, I propose that the `TextEdit` is computed at elaboration time, rather than asynchronously. This is acceptable because code actions that might take a long time to compute (like `squeeze_simp`) will be explicitly invoked by the user.

By storing these completions on the infotree at elaboration time, we don't have to decide here on an API for working with holes or deciding exactly which completions should be supported for a different PR. If we want any custom UI for working with holes, they can use the user-widget system.

## Changes proposed:

- [ ] Introduce a new `InfoTree` object called `CodeActionInfo`. Interactive tactics such as `squeeze_simp` will put one of these on the infotree.
- [ ] Implement `textDocument/codeAction` LSP request.
- [ ] Add support to `CompletionInfo` for sending `TextEdit` objects.
- [ ] Write a `diff : Syntax → Syntax → TextEdit` function to generate LSP text edits from a syntax change.
- [ ] Write a Lean → client  RPC call commanding a `TextEdit`, this can then be used by the infoview ( calls from contextual suggestions and user-widgets) to control the editor.

EdAyers:
Support for [LSP code actions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction).

resolves #1494

## todo:

- [ ] unit tests
- [x] implement `codeAction/resolve`
- [ ] getting code actions from providers should be async
- [ ] implement some example code actions that are useful
- [x] break off core changes to own pr

### DIFF
--- a/src/Lean/Data/Lsp/Basic.lean
+++ b/src/Lean/Data/Lsp/Basic.lean
@@ -244,6 +244,9 @@ instance : Append WorkspaceEdit where
 def ofTextDocumentEdit (e : TextDocumentEdit) : WorkspaceEdit :=
   { documentChanges := #[DocumentChange.edit e]}
 
+def ofTextEdit (uri : DocumentUri) (te : TextEdit) : WorkspaceEdit :=
+  { changes := RBMap.empty.insert uri #[te]}
+
 end WorkspaceEdit
 
 /-- An item to transfer a text document from the client to the server.
--- a/src/Lean/Data/Position.lean
+++ b/src/Lean/Data/Position.lean
@@ -29,6 +29,7 @@ instance : ToExpr Position where
 
 end Position
 
+/-- `FileMap` is a helper structure for mapping `String.Pos` in the input string to line/column information. -/
 structure FileMap where
   source    : String
   positions : Array String.Pos
--- a/src/Lean/Elab/Deriving/FromToJson.lean
+++ b/src/Lean/Elab/Deriving/FromToJson.lean
@@ -103,24 +103,29 @@ where
 
 def mkFromJsonInstanceHandler (declNames : Array Name) : CommandElabM Bool := do
   if declNames.size == 1 then
-    if isStructure (← getEnv) declNames[0]! then
+    let declName := declNames[0]!
+    if isStructure (← getEnv) declName then
       let cmds ← liftTermElabM do
-        let ctx ← mkContext "fromJson" declNames[0]!
+        let ctx ← mkContext "fromJson" declName
         let header ← mkHeader ``FromJson 0 ctx.typeInfos[0]!
-        let fields := getStructureFieldsFlattened (← getEnv) declNames[0]! (includeSubobjectFields := false)
-        let jsonFields := fields.map (Prod.snd ∘ mkJsonField)
+        let fields := getStructureFieldsFlattened (← getEnv) declName (includeSubobjectFields := false)
+        let getters ← fields.mapM (fun field => do
+          let getter ← `(getObjValAs? j _ $(Prod.snd <| mkJsonField field))
+          let getter ← `(doElem| Except.mapError (fun s => (toString $(quote declName)) ++ "." ++ (toString $(quote field)) ++ ": " ++ s) <| $getter)
+          return getter
+        )
         let fields := fields.map mkIdent
         let cmd ← `(private def $(mkIdent ctx.auxFunNames[0]!):ident $header.binders:bracketedBinder* (j : Json)
           : Except String $(← mkInductiveApp ctx.typeInfos[0]! header.argNames) := do
-          $[let $fields:ident ← getObjValAs? j _ $jsonFields]*
+          $[let $fields:ident ← $getters]*
           return { $[$fields:ident := $(id fields)],* })
         return #[cmd] ++ (← mkInstanceCmds ctx ``FromJson declNames)
       cmds.forM elabCommand
       return true
     else
-      let indVal ← getConstInfoInduct declNames[0]!
+      let indVal ← getConstInfoInduct declName
       let cmds ← liftTermElabM do
-        let ctx ← mkContext "fromJson" declNames[0]!
+        let ctx ← mkContext "fromJson" declName
         let header ← mkHeader ``FromJson 0 ctx.typeInfos[0]!
         let fromJsonFuncId := mkIdent ctx.auxFunNames[0]!
         let alts ← mkAlts indVal fromJsonFuncId
--- a/src/Lean/Elab/Deriving/TypeName.lean
+++ b/src/Lean/Elab/Deriving/TypeName.lean
@@ -23,3 +23,6 @@ private def deriveTypeNameInstance (declNames : Array Name) : CommandElabM Bool 
 
 initialize
   registerDerivingHandler ``TypeName deriveTypeNameInstance
+
+deriving instance TypeName for String
+instance : Inhabited Dynamic := ⟨Dynamic.mk ""⟩
--- a/src/Lean/Elab/InfoTree/Main.lean
+++ b/src/Lean/Elab/InfoTree/Main.lean
@@ -37,7 +37,7 @@ def CompletionInfo.stx : CompletionInfo → Syntax
   | tactic stx .. => stx
 
 def CustomInfo.format : CustomInfo → Format
-  | i => Std.ToFormat.format i.json
+  | i => f!"CustomInfo({i.value.typeName})"
 
 instance : ToFormat CustomInfo := ⟨CustomInfo.format⟩
 
--- a/src/Lean/Elab/InfoTree/Types.lean
+++ b/src/Lean/Elab/InfoTree/Types.lean
@@ -6,6 +6,7 @@ Authors: Wojciech Nawrocki, Leonardo de Moura, Sebastian Ullrich
 -/
 import Lean.Message
 import Lean.Data.Json
+import Lean.Data.Lsp.CodeActions
 
 namespace Lean.Elab
 
@@ -22,9 +23,12 @@ structure ContextInfo where
   ngen          : NameGenerator -- We must save the name generator to implement `ContextInfo.runMetaM` and making we not create `MVarId`s used in `mctx`.
   deriving Inhabited
 
-/-- An elaboration step -/
+/-- Base structure for `TermInfo`, `CommandInfo` and `TacticInfo`. -/
 structure ElabInfo where
+  /-- The name of the elaborator that created this info. -/
   elaborator : Name
+  /-- The piece of syntax that the elaborator created this info for.
+  Note that this also implicitly stores the code position in the syntax's SourceInfo. -/
   stx : Syntax
   deriving Inhabited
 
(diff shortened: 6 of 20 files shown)


src/Lean/Data/Lsp/CodeActions.lean
```diff
@@ -0,0 +1,162 @@
...
+structure ResolveSupport where
+  properties: Array String
+  deriving FromJson, ToJson
+
+def CodeActionLiteralSupport := Array CodeActionKind
+
+instance : FromJson CodeActionLiteralSupport where
+  fromJson? j := do
+    let j ← j.getObjVal? "codeActionKind"
+    let j ← j.getObjVal? "valueSet"
```
gebner, line 139: I'd just make two extra structures, like we have everywhere else.

src/Lean/Data/Lsp/Basic.lean
```diff
@@ -91,20 +123,138 @@ structure VersionedTextDocumentIdentifier where
...
+  uri: DocumentUri
+  options? : Option DeleteFile.Options := none
+  annotationId?: Option String := none
+  deriving ToJson, FromJson
+
+/-- A change to a file resource.
+
+[reference](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#resourceChanges)-/
+inductive DocumentChange where
+| create : CreateFile → DocumentChange
```
gebner, line 182: Please indent inductive cases.

src/Lean/Data/Lsp/Capabilities.lean
```diff
@@ -37,9 +39,28 @@ structure WindowClientCapabilities where
...
+  groupsOnLabel := false
+  deriving ToJson, FromJson
+
+structure WorkspaceEditClientCapabilities where
+  /-- The client supports versioned document changes in `WorkspaceEdit`s. -/
+  documentChanges : Bool := false
+  /--  Whether the client in general supports change annotations on text edits. -/
+  changeAnnotationSupport? : Option ChangeAnnotationSupport := none
+  /-- The resource operations the client supports. Clients should at least support 'create', 'rename' and 'delete' files and folders.-/
+  resourceOperations?: Option (Array String) := none
```
gebner, line 52: Please put a space before the colons.

src/Lean/Data/Lsp/Basic.lean
```diff
@@ -91,20 +123,138 @@ structure VersionedTextDocumentIdentifier where
...
+      `AnnotatedTextEdit`s or create, rename and delete file / folder
+      operations.
+
+      Whether clients honor this property depends on the client capability
+      `workspace.changeAnnotationSupport`. -/
+  changeAnnotations: Std.RBMap String ChangeAnnotation compare := ∅
+  deriving ToJson, FromJson
+
+namespace WorkspaceEdit
+  instance : EmptyCollection WorkspaceEdit := ⟨{}⟩
```
gebner, line 234: Please no indentation inside namespaces.

src/Lean/Data/Lsp/CodeActions.lean
```diff
@@ -0,0 +1,162 @@
...
+/-- [Reference](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionClientCapabilities)-/
+structure CodeActionClientCapabilities where
+  dynamicRegistration := false
+  codeActionLiteralSupport? : Option CodeActionLiteralSupport := none
+  isPreferredSupport := false
+  disabledSupport := false
+  dataSupport := false
+  resolveSupport? : Option ResolveSupport := none
+  honorsChangeAnnotations := false
+
```
gebner, line 158: Please no empty line before `deriving`.

src/Lean/Data/Lsp/CodeActions.lean
```diff
@@ -0,0 +1,162 @@
...
+    let xs : Array CodeActionKind ← fromJson? j
+    return xs
+
+instance : ToJson CodeActionLiteralSupport where
+  toJson (x : Array CodeActionKind) :=
+    toJson x
+    |> Json.mkSingleton "valueSet"
+    |> Json.mkSingleton "codeActionKind"
+
+/-- [Reference](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeActionClientCapabilities)-/
```
gebner, line 149: Please space before end of comment.

src/Lean/Server/FileWorker/CodeActions.lean
```diff
@@ -0,0 +1,87 @@
...
+builtin_initialize registerBuiltinAttribute {
+  name := `codeActionProvider
+  descr := "Use to decorate methods for suggesting code actions."
+  add := fun src _stx _kind => do
+    -- [todo] assert src is a CodeActionProvider
+    modifyEnv (codeActionProviderExt.addEntry · src)
+}
+
+private unsafe def evalCodeActionProviderUnsafe [MonadEnv M] [MonadOptions M] [MonadError M] [Monad M] (declName : Name) : M CodeActionProvider := do
+  evalConstCheck CodeActionProvider `CodeActionProvider declName
```
gebner, line 49:
```suggestion
  evalConstCheck CodeActionProvider ``CodeActionProvider declName
```

src/Lean/Server/FileWorker/CodeActions.lean
```diff
@@ -0,0 +1,87 @@
...
+private opaque evalCodeActionProvider [MonadEnv M] [MonadOptions M] [MonadError M] [Monad M] (declName : Name) : M CodeActionProvider
+
+/-- Handles a `textDocument/codeAction` request.
+
+[reference](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction).-/
+def handleCodeAction (params : CodeActionParams) : RequestM (RequestTask (Array CodeAction)) := do
+  let doc ← readDoc
+  let pos := doc.meta.text.lspPosToUtf8Pos params.range.end
+  bindWaitFindSnap doc (fun s => s.endPos ≥ pos)
+    (notFoundX := throw ⟨.invalidParams, s!"no snapshot found at {pos}"⟩)
```
gebner, line 61:
```suggestion
    (notFoundX := return ⟨∅⟩)
```
See #1548

EdAyers:
Moved to #1654, #1661 and #1665

PR #1553, opened 2022-08-31
feat: add Hashable for Subtype

marcusrossel:
[Related Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Hashable.20Subtype)


src/Init/Prelude.lean
```diff
@@ -3244,6 +3244,9 @@ def USize.toUInt64 (u : USize) : UInt64 where
 @[extern "lean_uint64_mix_hash"]
 opaque mixHash (u₁ u₂ : UInt64) : UInt64
 
+instance [Hashable α] {p : α → Prop} : Hashable { a : α // p a } where
```
leodemoura, line 3247: The notation `{ ... // ... }` is not available in the `Prelude.lean` file.

Issue #1280, opened 2022-07-04
Additional concurrency primitives


gebner:
Lean only has a single concurrency primitive right now, `Task`, which is very nice when it fits your use case (and also hard to misuse) but you can't easily use it to achieve mutual exclusion.  The most complex concurrent program (to my knowledge) in Lean so far is the LSP server, and it already casually mentions possible race conditions in its comments:
```lean
  private def nextCmdSnap (ctx : WorkerContext) (m : DocumentMeta) (cancelTk : CancelToken)
      : AsyncElabM (Option Snapshot) := do
    cancelTk.check
    let s ← get
    let lastSnap := s.snaps.back
    if lastSnap.isAtEnd then  -- ...
      return none
    publishProgressAtPos m lastSnap.endPos ctx.hOut
    -- ...
    cancelTk.check
    /- [...] Specifically, it may happen that
      we interrupted this elaboration task right at this point and a newer elaboration task
      emits diagnostics, after which we emit old diagnostics because we did not yet detect
      the interrupt. Explicitly clearing diagnostics is difficult for a similar reason,
      because we cannot guarantee that no further diagnostics are emitted after clearing
      them. [...] -/
    publishDiagnostics m snap.diagnostics.toArray ctx.hOut
    publishIleanInfoUpdate m ctx.hOut #[snap]
    return some snap
```
To be clear, the concrete issue here is that something can happen between `cancelTk.check` and `publishDiagnostics`. (Of course, it's not just the `publishDiagnostics` call that is a problem.  Every other `publish*` call falls prey to the same race condition.)

The obvious fix for this race condition is to add a mutex primitive and then just acquire the lock to execute `cancelTk.check` and `publishDiagnostics` atomically.  A concrete API proposal could look like this:
```lean
BaseMutex : Type
BaseMutex.mk : BaseIO BaseMutex
BaseMutex.lock (mutex : BaseMutex) : BaseIO Unit
BaseMutex.unlock (mutex : BaseMutex) : BaseIO Unit

-- rust-style guarded state
structure Mutex (α : Type) where
  ref : IO.Ref α
  mutex : BaseMutex

def atomically [Monad M] [MonadLift BaseIO M] [MonadFinally M] (mutex : Mutex α) (k : StateT α M β) : M β := ...
```

[...]

PR #1555, opened 2022-09-01
Additional concurrency primitives

gebner:
This PR adds several additional concurrency primitives, as described in RFC #1280.

- `BaseMutex` is a mutex
- `Mutex α` is a mutex with guarded state `α` (like in Rust)
- `Condvar` is a condition variable
- `Promise` creates a task whose value will be set later

There's also two very unoptimized queue-like primitives building on top of these:
- `Channel` with `ch.recv? : BaseIO (Option α)`
- `AsyncChannel` with `ch.recvAsync : BaseIO (Task (Option α))`

The channels don't perform too badly, based on some extremly quick benchmarking results.  `Channel` transports messages at  400kHz, `AsyncChannel` at 700kHz on a slow machine.  This seems in the same ballpark as Rust's MPSC channels, which seem to be roughly at [900kHz](https://users.rust-lang.org/t/poor-performance-of-sending-items-over-an-mpsc-channel/50559) (I didn't run the benchmark locally).

Maybe it's worth removing `Channel` and only keep the `AsyncChannel` as it is faster (and more general).

### DIFF
--- a/src/Init/Data/Channel.lean
+++ b/src/Init/Data/Channel.lean
@@ -22,10 +22,10 @@ structure Channel.State (α : Type) where
   deriving Inhabited
 
 /--
-FIFO channel with a receive operation returning a `Task`.
+FIFO channel with unbounded buffer, where `recv?` returns a `Task`.
 
 A channel can be closed.  Once it is closed, all `send`s are ignored, and
-`recvAsync` returns `none` once the queue is empty.
+`recv?` returns `none` once the queue is empty.
 -/
 def Channel (α : Type) : Type := Mutex (Channel.State α)
 
@@ -61,11 +61,13 @@ def Channel.close (ch : Channel α) : BaseIO Unit :=
     set { st with closed := true, consumers := ∅ }
 
 /--
-Receives a message.
+Receives a message, without blocking.
+The returned task waits for the message.
+Every message is only received once.
 
 Returns `none` if the channel is closed and the queue is empty.
 -/
-def Channel.recvAsync (ch : Channel α) : BaseIO (Task (Option α)) :=
+def Channel.recv? (ch : Channel α) : BaseIO (Task (Option α)) :=
   ch.atomically do
     let st ← get
     if let some (a, values) := st.values.dequeue? then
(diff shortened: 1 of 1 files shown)


src/Init/Data/Queue.lean
```diff
@@ -0,0 +1,50 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Daniel Selsam
+
+Simple queue implemented using two lists.
+Note: this is only a temporary placeholder.
+-/
+prelude
+import Init.Data.List
+
+namespace Std
```
gebner, line 12: @digama0 I'll defer to your judgement on the naming here.

src/runtime/object.cpp
```diff
@@ -771,6 +776,15 @@ class task_manager {
         enqueue_core(t);
     }
 
+    void resolve(lean_task_object * t, object * v) {
+        unique_lock<mutex> lock(m_mutex);
+        if (t->m_value || !t->m_imp) {
```
Kha, line 781: When is the second alternative reachable?

src/Init/System/Promise.lean
```diff
@@ -0,0 +1,57 @@
...
+
+Typical usage is as follows:
+1. `let promise ← Promise.new` creates a promise
+2. `promise.result : Task α` can now be passed around
+3. `promise.result.get` blocks until the promise is resolved
+4. `promise.resolve a` resolves the promise
+5. `promise.result.get` now returns `a`
+
+Every promise must eventually be resolved.
+Otherwise the memory used for the promise will be leaked,
```
Kha, line 26: Couldn't we skip the `Deactivated` state in this case? We would need more data in the `lean_task_impl` so that `deactivate_task` can identify this case I suppose.

src/Init/System/Mutex.lean
```diff
@@ -0,0 +1,123 @@
...
+
+/--
+`AtomicT α M` is the monad that can be atomically executed inside a `Mutex α`,
+with outside monad `M`.
+The action has access to the state `α` of the mutex (via `get` and `set`).
+-/
+abbrev AtomicT := StateRefT' IO.RealWorld
+
+/-- `mutex.atomically k` runs `k` with access to the mutex's state while locking the mutex. -/
+def Mutex.atomically [Monad M] [MonadLiftT BaseIO M] [MonadFinally M]
```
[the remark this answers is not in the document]
leodemoura: Please let's keep the current naming convention for now.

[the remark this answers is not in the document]
gebner: I think I wrote this a couple weeks back when `Monad M` was just gaining traction.  It's uppercase because it's a type.  I've changed it to `m` in this PR.  But there's plenty of other occurrences, check out `rg 'Monad M'`.

src/runtime/object.cpp
gebner: It's not (I hope you mean the `!t->m_imp`).

src/Init/System/Promise.lean
gebner: Indeed, we'd need more data.  Right now, a Promised task looks the same as a Running task.  And Running tasks are owned by the worker (and thus can't be freed).

src/Init/System/Mutex.lean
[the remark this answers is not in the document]
leodemoura: Yes, it seems they all come from @EdAyers :)

PR #1557, opened 2022-09-02
doc: update lean 3 -> lean 4 in declarations.md

digama0:
I didn't fully rewrite this file, but I updated the syntax and naming conventions of the examples in this file, and removed a few comments that were out of date for lean 4. (I didn't look deeply into how or where this md file is used; it might just need to be removed.)

### DIFF
--- a/doc/declarations.md
+++ b/doc/declarations.md
@@ -165,70 +165,70 @@ Below are some common examples of inductive types, many of which are defined in 
 
 .. code-block:: lean
 
-  namespace hide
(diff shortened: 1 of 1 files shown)


leodemoura:
@digama0 Note that this file is currently not reachable. We were trying to move the Lean 3 reference manual to Lean 4, but did not manage to finish. So, you are not going to see the effects of your changes on the website manual.
I am going to merge the PR because one day we may want to put back this page in the manual.

PR #1559, opened 2022-09-03
A basic compiler pass test framework

hargoniX:
I played around with this for a while and am now sort of content but I'm very sure there is still lots of space for further improvement. The concept of a `PassInstaller` does in general allow us to perform arbitrarily complex tests, in theory even across multiple compiler passes. If someone has different ideas for APIs, types of test etc. please do suggest them.

### DIFF
--- a/tests/lean/CompilerCSE.lean
+++ b/tests/lean/CompilerCSE.lean
@@ -1,0 +1,19 @@
+import Lean.Compiler.Main
+import Lean.Compiler.LCNF.Testing
+import Lean.Elab.Do
+
+open Lean
+open Lean.Compiler.LCNF
+
+-- Run compilation twice to avoid the output caused by the inliner
+#eval Compiler.compile #[``Lean.Meta.synthInstance, ``Lean.Elab.Term.Do.elabDo]
+
+@[cpass]
+def cseFixTest : PassInstaller := Testing.assertIsAtFixPoint `cse
+
+@[cpass]
+def cseSizeTest : PassInstaller :=
+  Testing.assertReducesOrPreservesSize `cse `cseSizeLeq "CSE increased size of declaration"
+
+set_option trace.Compiler.test true in
+#eval Compiler.compile #[``Lean.Meta.synthInstance, ``Lean.Elab.Term.Do.elabDo]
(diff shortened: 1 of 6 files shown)


src/Lean/Compiler/LCNF/Testing.lean
```diff
@@ -0,0 +1,154 @@
...
+
+namespace Lean.Compiler.LCNF
+
+partial def Code.containsConst (constName : Name) (code : Code) : Bool :=
+  match code with
+  | .let decl k => goExpr decl.value || containsConst constName k
+  | .fun decl k => containsConst constName decl.value || containsConst constName k
+  | .jp decl k => containsConst constName decl.value || containsConst constName k
+  | .jmp _ args => args.any goExpr
+  | .cases cs => cs.alts.any goAlt
```
leodemoura, line 17: We can use `cs.alts.any fun alt => containsConst alt.getCode` and delete `goAlt`.

src/Lean/Compiler/LCNF/Testing.lean
```diff
@@ -0,0 +1,154 @@
...
+where
+  goAlt (alt : Alt) : Bool :=
+    match alt with
+    | .alt _ _ body | .default body => containsConst constName body
+  goExpr (e : Expr) : Bool :=
+    match e with
+    | .const name .. => name == constName
+    | .app fn arg .. => goExpr fn || goExpr arg
+    | .lam _ _ body .. => goExpr body
+    | .letE _ _ value body .. => goExpr value || goExpr body
```
leodemoura, line 28: LCNF should not contain `.letE`. We can put `unreachable!` here. 
BTW, I think I have not done that consistently in the rest of the code :(

src/Lean/Compiler/LCNF/Testing.lean
```diff
@@ -0,0 +1,154 @@
...
+
+def assertForEachDeclAfter (passUnderTestName : Name) (testName : Name) (assertion : Decl → Bool) (msg : String) : PassInstaller :=
+  assertConditionAfter passUnderTestName testName (Array.forM (fun decl => assert (assertion decl) msg))
+
+/--
+Replace a pass with a wrapper one that allows the user to provide an assertion which
+takes into account both the declarations that were sent to and produced by the pass
+under test.
+-/
+def assertConditionAround (passUnderTestName : Name) (testName : Name) (assertion : Array Decl → Array Decl → CompilerM Unit) : PassInstaller :=
```
leodemoura, line 66: This is a nice combinator.

leodemoura:
I have only minor comments. I am going to merge as-is.

PR #1560, opened 2022-09-04
feat: syntax declarations in `mutual` inductive block

Kha:
This finally allows us to write code such as
```lean
mutual
notation:60 "⟨" c ", " σ "⟩"  " ⇓ " σ' => Bigstep c σ σ'

inductive Bigstep : Com → State → State → Prop where
  | skip :
    ⟨skip, σ⟩ ⇓ σ
  ...
end
```
which is a significant improvement over [this hack](https://github.com/IPDSnelting/tba-all/blob/6e618b8f0097b6a0b90ef588a6295a435ea13522/TBA/While/Semantics.lean#L41-L64) @javra and I used for the last two iterations of our course (the specific occasion that reminded me of this missing feature though was the PLMW tutorial @davidchristiansen and I are preparing).

The implementation is far from ideal, see comments below. The observable behavior however looks good, so for me this feels important enough to include as is and refine from there on.

### DIFF
--- a/src/Lean/Elab/Declaration.lean
+++ b/src/Lean/Elab/Declaration.lean
@@ -230,13 +230,32 @@ private def isMutualInductive (stx : Syntax) : Bool :=
   stx[1].getArgs.all fun elem =>
     let decl     := elem[1]
     let declKind := decl.getKind
-    declKind == `Lean.Parser.Command.inductive
+    declKind == ``Parser.Command.inductive || isValidSyntaxCmdInMutual elem
 
 private def elabMutualInductive (elems : Array Syntax) : CommandElabM Unit := do
-  let views ← elems.mapM fun stx => do
+  let (inds, stxCmds) := elems.partition (·.isOfKind ``Parser.Command.declaration)
+
+  let views ← inds.mapM fun stx => do
      let modifiers ← elabModifiers stx[0]
      inductiveSyntaxToView modifiers stx[1]
+
+  -- first, approximate syntax commands elaboration pass to collect macro state
+  let env ← getEnv
+  -- HACK: While we know all the inductive declaration names of the mutual block at this point,
+  -- we currently lack a nice API to forward that knowledge to the precheck/quotation code.
+  -- And in any case, the syntax cannot be hygienic as inside an inductive declaration its own
+  -- name is in the local context instead of in the environment.
+  withScope (fun scope => { scope with opts := scope.opts.setBool `quotPrecheck false |>.setBool `hygiene false }) do
+    stxCmds.forM elabCommand
+  let macroStateWithSyntax := macroAttribute.ext.getState (← getEnv)
+
+  -- now elaborate inductive views in original environment, but updated macro state
+  setEnv (macroAttribute.ext.modifyState env (fun _ => macroStateWithSyntax))
   elabInductiveViews views
+
+  -- finally, reset macro state and elaborate syntax commands *properly*, with inductive types in scope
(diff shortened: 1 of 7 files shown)


src/Lean/Elab/Declaration.lean
```diff
@@ -230,14 +230,33 @@ private def isMutualInductive (stx : Syntax) : Bool :=
...
+
+  -- first, approximate syntax commands elaboration pass to collect macro state
+  let env ← getEnv
+  -- HACK: While we know all the inductive declaration names of the mutual block at this point,
+  -- we currently lack a nice API to forward that knowledge to the precheck/quotation code.
+  -- And in any case, the syntax cannot be hygienic as inside an inductive declaration its own
+  -- name is in the local context instead of in the environment.
+  withScope (fun scope => { scope with opts := scope.opts.setBool `quotPrecheck false |>.setBool `hygiene false }) do
+    stxCmds.forM elabCommand
+  let macroStateWithSyntax := macroAttribute.ext.getState (← getEnv)
```
Kha, line 250: @leodemoura This is related to previous discussions we had about whether there are better solutions for `isAuxDecl` binders. I have some new thoughts on that.

src/Lean/Elab/Syntax.lean
```diff
@@ -412,11 +412,30 @@ def expandNoKindMacroRulesAux (alts : Array (TSyntax ``matchAlt)) (cmdName : Str
...
+      try
+        for stx in s.stxStack do
+          if isValidSyntaxCmdInMutual stx then
+            elabCommand stx
+      catch e => logException e
+    let Except.ok (_, s') ← go { fileName := c.fileName, fileMap := c.fileMap, tacticCache? := none, cmdPos := s.pos }
+      |>.run { env := c.env, maxRecDepth := maxRecDepth.get c.options, scopes := [{ header := "", currNamespace := c.currNamespace }] }
+      -- safety: the default elaborators of the accepted commands should not have any observable side-effects.
+      -- In theory this may not be true if they are overridden.
+      |> unsafeEIO
```
Kha, line 435: We haven't documented the actual caveats of `unsafeIO` so far, but I'm relatively sure this should be fine

src/Lean/Parser/Command.lean
```diff
@@ -147,7 +147,7 @@ def openDecl         := withAntiquot (mkAntiquot "openDecl" `Lean.Parser.Command
   openHiding <|> openRenaming <|> openOnly <|> openSimple <|> openScoped
 @[builtinCommandParser] def «open»    := leading_parser withPosition ("open " >> openDecl)
 
-@[builtinCommandParser] def «mutual» := leading_parser "mutual " >> many1 (ppLine >> notSymbol "end" >> commandParser) >> ppDedent (ppLine >> "end") >> terminationSuffix
+@[builtinCommandParser] def «mutual» := leading_parser "mutual " >> many1 (ppLine >> notSymbol "end" >> withSyntaxOfStack commandParser) >> ppDedent (ppLine >> "end") >> terminationSuffix
```
Kha, line 150: Repeatedly elaborating the notations in each iteration could be solved by a fusion of `withSyntaxOfStack` and `many1`, but it is hard to imagine this being relevant in practice.

Kha:
One semantics caveat that should not be surprising given how parsing in Lean works: while notations may make forward references to inductive types, inductive types may not make forward references to new syntax.

leodemoura:
We have decided to support this feature by modifying how we process `mutual` blocks.
The idea is to process them as we process `section`s.

PR #1565, opened 2022-09-07
Test: parser combinators without `[inline]`

leodemoura:


leodemoura:
!bench

leodemoura:
@Kha @gebner What do you think? I think it is worth merging. The generated code is smaller, and it is faster to build.

Kha:
I'm surprised parsing didn't get slower, but looks good!

gebner:
I'm not extremely surprised; the `@[inline]` attribute only controls one of the inlining phases, some of the functions might still be inlined.  This inlining phase is also pre-erasure, and thus increases the stage1 IR size quite dramatically.  So most of the wins here are probably compilation-speed and code-size related.

This doesn't explain the parsing gains, which I agree are indeed surprising.

gebner:
@Kha The "update stage0" commit here is not necessary for benchmarking since it runs the stage2 build, right?

Kha:
> This doesn't explain the parsing gains, which I agree are indeed surprising.

One shouldn't read too much into the specific stdlib measures though as there are only two runs (so the stddev calculation is a bit questionable) and each run is parallelized (using 3 jobs on a 4-core machine IIRC), so there might not be quite independent.

> The "update stage0" commit here is not necessary for benchmarking since it runs the stage2 build, right?

Yes, that's true.

Kha:
> BTW, is there a way to download the raw data from velcom?

I don't think so other than intercepting and customizing the API requests, but I'm happy to provide a database dump.

leodemoura:
Merged [7bc9c12](https://github.com/leanprover/lean4/pull/1565/commits/7bc9c12462f8af8858b720df7d1959106ac2e6ed)

PR #1566, opened 2022-09-07
feat: remove description argument from register_simp_attr

gebner:


src/Lean/DocString.lean
```diff
@@ -92,4 +92,9 @@ def getDocStringText [Monad m] [MonadError m] [MonadRef m] (stx : TSyntax `Lean.
   | Syntax.atom _ val => return val.extract 0 (val.endPos - ⟨2⟩)
   | _                 => throwErrorAt stx "unexpected doc string{indentD stx.raw[1]}"
 
+def TSyntax.getDocString (stx : TSyntax `Lean.Parser.Command.docComment) : String :=
```
gebner, line 95: I couldn't use `getDocStringText` because `MacroM` doesn't have `MonadError`.

PR #1567, opened 2022-09-07
find join points pass

hargoniX:
Adds a pass for identifying and replacing functions that might instead be join point declarations.


leodemoura:
Merged manually

PR #1568, opened 2022-09-08
fix doc typos and do some polish on wording

lovettchris:
Having not looked at it for a few days, now re-reading it I found a few things to fix and improve.  Now it's perfect :-)


doc/monads/applicatives.lean.leanInk
```diff
@@ -0,0 +1,3940 @@
+[{"contents":
```
leodemoura, line 1: This is an autogenerated file.

lovettchris: woops - fixed.

PR #1572, opened 2022-09-08
fix(PrettyPrinter/Delaborator): unify default behaviour of pp.proof with false

alexjbest:
Previously the behaviour of `set_option pp.proofs false` was different from not setting the option at all, this seems surprising and possibly unintended to me.
This can be seen in the following example:
```lean
def List.nthLe (l : List α) (n) (h : n < l.length) : α := sorry

set_option pp.proofs false -- comment and uncomment this line

example : List.nthLe [1, 2] 1 (by simp) = 2 := by
  -- cursor here, and click on (_ : 1 < List.length [1, 2]) to display the widget
  -- with this change commenting and uncommenting line 3 does nothing
  -- without change the widget display changes from
  -- @of_eq_true (1 < List.length [1, 2]) (eq_true_of_decide (Eq.refl true)) : 1 < List.length [1, 2]
  -- to @of_eq_true (1 < List.length [1, 2]) (_ : (1 < List.length [1, 2]) = True) : 1 < List.length [1, 2]
  sorry
```

### DIFF
--- a/src/Lean/PrettyPrinter/Delaborator/Basic.lean
+++ b/src/Lean/PrettyPrinter/Delaborator/Basic.lean
@@ -283,7 +283,7 @@ def delabCore (e : Expr) (optionsPerPos : OptionsPerPos := {}) (delab := Delabor
   trace[PrettyPrinter.delab.input] "{Std.format e}"
   let mut opts ← getOptions
   -- default `pp.proofs` to `true` if `e` is a proof
-  if pp.proofs.get? opts == none then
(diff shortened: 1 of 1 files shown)


gebner:
It is intentional.  This is the same as the Lean 3 behavior (minus the type of the proposition, I think we only showed the underscore).  By default, proofs are hidden unless you are printing a proof.

Note that there's also `pp.proofs.withType` to customize this even further.

PR #1578, opened 2022-09-10
PassManager and compiler test framework updates

hargoniX:
- adds the discussed phase and occurence features to the `PassManager`
- refactors the test framework into a monadic style one that is actually reasonable to use compared to the old one.

### DIFF
--- a/src/Lean/Compiler/LCNF/PassManager.lean
+++ b/src/Lean/Compiler/LCNF/PassManager.lean
@@ -18,10 +18,10 @@ inductive Phase where
 | /-- Here we still carry most of the original type information, most
       of the dependent portion is already (partially) erased though. -/
   base
-| /-- Here all types have been turned into `Any`. -/
-  untyped
-| /-- The lambda RC representation use for reference counting optimizations. -/
-  rc
+| /-- In this phase polymorphism has been eliminated. -/
+  mono
+| /-- In this phase impure stuff such as RC or efficient BaseIO transformations happen. -/
+  impure
 deriving Inhabited
 
 /--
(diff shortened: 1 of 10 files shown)


src/Lean/Compiler/LCNF/PassManager.lean
```diff
@@ -11,27 +11,118 @@ import Lean.Compiler.LCNF.CompilerM
...
 
+/--
+The pipeline phase a certain `Pass` is supposed to happen in.
+-/
+inductive Phase where
+| /-- Here we still carry most of the original type information, most
+      of the dependent portion is already (partially) erased though. -/
+  base
+| /-- Here all types have been turned into `Any`. -/
+  untyped
```
leodemoura, line 22: In the second phase, the plan is to eliminate polymorphism, but we will preserve the types. We can call it `mono`.
For example, the declaration 
```lean
def f (α : Type u) (p : List (α × α)) : List (α × α)
```
becomes 
```lean
def f (α : ◾) (p : List (⊤ × ⊤) : List (⊤ × ⊤)
```
where ` ◾` represents erased information and `⊤` is the any type. 
The type of 
```lean
def g (as : List (Nat × Bool) : Nat
```
is not altered.

src/Lean/Compiler/LCNF/PassManager.lean
```diff
@@ -11,27 +11,118 @@ import Lean.Compiler.LCNF.CompilerM
...
+| /-- The lambda RC representation use for reference counting optimizations. -/
+  rc
```
leodemoura, line 24: Yes, in the third phase we want to inject RC operations. We also want to eliminate the overhead of `BaseIO`. The general theme here is impurity. I think we should call this phase `impure`.

leodemoura:
Merged manually

PR #1579, opened 2022-09-10
fix: use `Type u` universes in List.foldl

digama0:
Although the definition of `List.foldl` makes sense even when `α` is a Prop, none of the other functions on lists work on Prop arguments, including `List.foldr`. In particular, `List.foldlM` doesn't take a Prop for `α` so this results in either universe-constrained lemmas or having to reprove all the theorems for `List.foldlM` on `List.foldl`.

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -2169,7 +2169,7 @@ Folds a function over a list from the left:
 `foldl f z [a, b, c] = f (f (f z a) b) c`
 -/
 @[specialize]
-def List.foldl {α β} (f : α → β → α) : (init : α) → List β → α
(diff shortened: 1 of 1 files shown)


src/Init/Prelude.lean
```diff
@@ -2169,7 +2169,7 @@ Folds a function over a list from the left:
 `foldl f z [a, b, c] = f (f (f z a) b) c`
 -/
 @[specialize]
-def List.foldl {α β} (f : α → β → α) : (init : α) → List β → α
+def List.foldl {α β : Type _} (f : α → β → α) : (init : α) → List β → α
```
leodemoura, line 2172: Could you please use `Type u` and `Type v` instead of `Type _`? We don't use `Type _` anywhere in the code base.

PR #1581, opened 2022-09-11
feat: tail-recursive List.{mapM, foldrM}

digama0:
I've been constructing tail-recursive versions of all the lean core list definitions using `@[csimp]`, but I got stuck on these. The issue is that `@[csimp]` lemmas only work for direct constant replacement (fully unapplied), and that means that we can't assume the monad is lawful in this case, which means that we can only make the most trivial rewrites to the expression, and `mapM` is decidedly *not* tail recursive no matter how you write it. It's not safe to use `implementedBy` either, since you could use that on a non-lawful monad instance to violate memory safety.

The remaining option is what is implemented in this PR, namely to change the definition itself to be naturally tail recursive. That means we lose the nice equation lemmas for it, but those can be recovered over a lawful monad so it shouldn't be so bad.

### DIFF
--- a/src/Init/Data/List/Control.lean
+++ b/src/Init/Data/List/Control.lean
@@ -40,10 +40,12 @@ Finally, we rarely use `mapM` with something that is not a `Monad`.
 Users that want to use `mapM` with `Applicative` should use `mapA` instead.
 -/
 
-@[specialize]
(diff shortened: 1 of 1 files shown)


leodemoura (approved):

leodemoura:
@gebner @Kha Any concerns? Otherwise, I will merge it soon.

gebner:
lgtm

PR #1582, opened 2022-09-12
feat: support `{s with ..}`

digama0:
This came up in the last mathlib mentor meeting. Explicit and implicit sources are not mutually exclusive, so change the type to reflect that and remove the ``invalid structure instance `with` and `..` cannot be used together`` error.

### DIFF
--- a/src/Lean/Elab/StructInst.lean
+++ b/src/Lean/Elab/StructInst.lean
@@ -193,15 +193,11 @@ private def elabModifyOp (stx modifyOp : Syntax) (sources : Array ExplicitSource
 private def getStructName (expectedType? : Option Expr) (sourceView : Source) : TermElabM Name := do
   tryPostponeIfNoneOrMVar expectedType?
   let useSource : Unit → TermElabM Name := fun _ => do
-    match sourceView.explicit.size with
-    | 0 =>
-      match expectedType? with
-      | some expectedType => throwUnexpectedExpectedType expectedType
-      | none => throwUnknownExpectedType
-    | 1 => return sourceView.explicit[0]!.structName
-    | _ =>
-      throwErrorAt sourceView.explicit[1]!.stx
-        "invalid \{...} notation, expected type is not known, using the type of the first source, extra sources are not needed"
+    unless sourceView.explicit.isEmpty do
+      return sourceView.explicit[0]!.structName
+    match expectedType? with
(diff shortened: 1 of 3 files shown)


src/Lean/Elab/StructInst.lean
```diff
@@ -197,13 +193,15 @@ private def elabModifyOp (stx modifyOp : Syntax) (sources : Array ExplicitSource
...
-    | _, none              => throwUnknownExpectedType
+    match sourceView.explicit.size with
+    | 0 =>
+      match expectedType? with
+      | some expectedType => throwUnexpectedExpectedType expectedType
+      | none => throwUnknownExpectedType
+    | 1 => return sourceView.explicit[0]!.structName
+    | _ =>
+      throwErrorAt sourceView.explicit[1]!.stx
+        "invalid \{...} notation, expected type is not known, using the type of the first source, extra sources are not needed"
```
digama0, line 204: This error message seems to be incorrect by the way. It says "using the type of the first source" but it's using `throwErrorAt` which is a fatal error. Should it be using a non-fatal error here?

leodemoura:
> This came up in the last mathlib mentor meeting. Explicit and implicit sources are not mutually exclusive, so change the type to reflect that and remove the invalid structure instance `with` and `..` cannot be used together error.

Correct. They are not mutually exclusive anymore in Lean 4.

src/Lean/Elab/StructInst.lean
leodemoura: I agree. We could just use the type of the first source. Did you try it?

leodemoura: > Should it still be a non-fatal error? Or should we just not report anything

I am not sure. If we don't report anything I feat it may produce counterintuitive behavior and confuse users. What do you think?

leodemoura (approved):

src/Lean/Elab/StructInst.lean
leodemoura: BTW, I am fine with merging as-is, and addressing this issue in another PR or just keeping it as-is.

digama0: I'm currently compiling a fix to just report nothing and use the first source. I think it's teachable enough, and also fairly easy to override using the expected type syntax.

PR #1584, opened 2022-09-12
doc: fix an example in the Macro Overview

eyelash:


eyelash:
For some reason the CI failed when trying to download something from Cachix. Not sure what to do here.

PR #1585, opened 2022-09-12
feat: put getElem? in the GetElem typeclass

digama0:
This makes `getElem?` a typeclass function instead of a definition, so individual instances can customize the definitional equalities that result. (The specific reason this came up is that `List.get? list i` and `list[i]?` are not defeq and this prevents using the latter as the simp normal form.)

I opted against a theorem that says that `getElem?` is equal to its default definition, both because it should be usable "lawlessly" and also because I can imagine cases where you might not want them to agree: for example `Syntax` has a getElem instance with domain `True` but one might (I didn't) choose the `getElem?` to use the underlying array bounds instead of just always returning `some (getElem ..)`.

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -2223,6 +2223,15 @@ def List.get {α : Type u} : (as : List α) → Fin as.length → α
   | cons _ as, ⟨Nat.succ i, h⟩ => get as ⟨i, Nat.le_of_succ_le_succ h⟩
 
 /--
+`as.get? i` returns the `i`'th element of the list `as`, or `none` if the index
+is out of bounds.
+-/
+def List.get? : List α → Nat → Option α
+  | cons a _,  0          => some a
+  | cons _ as, Nat.succ n => get? as n
+  | _,         _          => none
+
+/--
 `String` is the type of (UTF-8 encoded) strings.
 
 The compiler overrides the data representation of this type to a byte sequence,
@@ -2435,8 +2444,14 @@ class GetElem (cont : Type u) (idx : Type v) (elem : outParam (Type w)) (dom : o
   * `arr[i]'h`: uses `h` to prove the side goal
   -/
   getElem (xs : cont) (i : idx) (h : dom xs i) : elem
+  /--
+  The `getElem? arr i` function (notation `arr[i]?`) gets the nth element of `arr`,
+  or `none` if the index is out of bounds.
+  -/
+  getElem? (xs : cont) (i : idx) [Decidable (dom xs i)] : Option elem :=
+    dite _ (fun h => some (getElem xs i h)) (fun _ => none)
 
-export GetElem (getElem)
+export GetElem (getElem getElem?)
 
 /--
 `Array α` is the type of [dynamic arrays](https://en.wikipedia.org/wiki/Dynamic_array)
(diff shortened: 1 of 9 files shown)


src/Init/Prelude.lean
```diff
@@ -2435,8 +2444,14 @@ class GetElem (cont : Type u) (idx : Type v) (elem : outParam (Type w)) (dom : o
   * `arr[i]'h`: uses `h` to prove the side goal
   -/
   getElem (xs : cont) (i : idx) (h : dom xs i) : elem
+  /--
+  The `getElem? arr i` function (notation `arr[i]?`) gets the nth element of `arr`,
+  or `none` if the index is out of bounds.
+  -/
+  getElem? (xs : cont) (i : idx) [Decidable (dom xs i)] : Option elem :=
```
gebner, line 2451:
```suggestion
  getElem? (xs : cont) (i : idx) : Option elem :=
```
I wonder if it makes sense to bundle decidability here.

digama0: I would love to, but what would the default implementation be in that case?

gebner:
> I opted against a theorem that says that `getElem?` is equal to its default definition, both because it should be usable "lawlessly" and also because I can imagine cases where you might not want them to agree:

My preference would be to add the theorem.  We should avoid cases where `getElem` and `getElem?` disagree; those just cause confusion.

src/Init/Prelude.lean
gebner:
```lean
local macro "get_elem?_default" : tactic =>
  `(exact dite .....)

-- ...
  getElem? (xs : cont) (i : idx) : Option elem := by get_elem?_default
```

Unfortunately Sebastian has built in all kinds of presumably helpful checks, so you can't just write `:= by exact ...` and need to do this dance with the separate macro.

digama0: Ironically, we can't use that syntax yet, because this is in `Init.Prelude` and `macro` isn't legal. I think we can just move `GetElem` out of the prelude though...

digama0: @gebner Well, even after the big refactor necessary to get GetElem out of Init.Prelude, it still doesn't work. Here's a MWE:
```lean
local macro "get_elem?_default"  : tactic =>
  `(trace_state; exact fun xs i => if h : _ then some ($(Lean.mkIdent `getElemₓ) xs i h) else none)

class GetElemₓ (cont : Type u) (idx : Type v) (elem : outParam (Type w)) (dom : outParam (cont → idx → Prop)) where
  getElemₓ (xs : cont) (i : idx) (h : dom xs i) : elem
  getElemₓ? : (xs : cont) → (i : idx) → Option elem := by get_elem?_default 

instance : GetElemₓ (Array α) USize α fun xs i => i.toNat < xs.size where
  getElemₓ xs i h := xs.uget i h
```
You can see in the printout that the auto param exists in the target local context (containing only `α`), which in particular lacks the `getElem` local variable that is available in regular default arguments. So there is no way for the auto param to know what was passed as the `getElem` argument to the structure.

gebner: :sob:

leodemoura:
@digama0 @gebner Are you on the same page about this PR, or should we discuss it at the next dev meeting? I did not read it yet.

digama0:
Even with the fix here, I think there are still some aspects of `getElem?` that make it unsuitable as a simp normal form, so I'm okay with just dropping this PR.

An alternative strategy would be to have a separate `GetElem?` typeclass (extending `GetElem`) which provides the `getElem?` function and the `getElem!` definition. All existing `GetElem` implementations would want to implement this extension.

Re: procedural concerns, it is something we could discuss at the meeting.

gebner:
I think we should discuss this at the next meeting.

> I think there are still some aspects of getElem? that make it unsuitable as a simp normal form

FWIW, I've used `getElem` as the simp-normal form in many proofs in mathlib4.  (There are no lemmas in the default simp set rewriting `getElem?` to `getElem` though.)  The proof argument is already annoying enough, I really don't want to deal with `Option.bind` and co.

digama0:
I'm happy with the notational aspects of the `x[i]` notation, but for it to be suitable as a simp normal form it shouldn't have any extra arguments like a spurious decidability hypothesis that will cause problems in rewriting. Also note that I said `getElem?`, not `getElem`. `getElem` already has the best type signature we can give it, with an out param and everything to make it look nice. `getElem?` ends up with a lot of cruft though as a result of being dependent on a proof argument for `getElem` that is no longer relevant.

PR #1586, opened 2022-09-12
feat: add `withIsShared` utility function

digama0:
This adds two combinators for working directly with uniqueness:
```lean
@[extern] def withSquashIsShared (a : α) (f : α → Squash Bool → β) : β

def withIsShared (a : α) (f : (a' : α) → a' = a → Bool → β) (h : f a rfl false = f a rfl true) : β
```
The second one is defined using the first. They allow writing code that can branch on the RC count, as long as it does not affect the result. This is useful for functions which use a different strategy when they have unique access to the object. As an application, `dbgTraceIfShared` can now be defined in pure lean code.

I tried to create a version with a simpler type signature, namely
```lean
@[extern] unsafe def isSharedUnsafe : @& α → Bool
```
but this seems to always return `true` in the interpreter, even on values for which `dbgTraceIfShared` just reported are unshared.

### DIFF
--- a/src/Init/Util.lean
+++ b/src/Init/Util.lean
@@ -18,23 +18,26 @@ def dbgTrace {α : Type u} (s : String) (f : Unit → α) : α := f ()
 def dbgTraceVal {α : Type u} [ToString α] (a : α) : α :=
   dbgTrace (toString a) (fun _ => a)
 
-@[neverExtract, extern "lean_with_is_shared"]
-def withSquashIsShared {α : Type u} {β : Type v} (a : α) (f : α → Squash Bool → β) : β := f a (Squash.mk true)
+set_option linter.unusedVariables.funArgs false in
+@[neverExtract, extern "lean_is_shared_extern"]
+unsafe def isSharedUnsafe {α : Type u} (a : α) : Bool := true
 
-def withIsShared {α : Type u} {β : Type v} (a : α)
-    (f : (a' : α) → a' = a → Bool → β) (h : f a rfl false = f a rfl true) : β :=
-  withSquashIsShared (α := {x // x = a}) ⟨a, rfl⟩ fun ⟨a', eq⟩ b =>
-    Quot.liftOn b (f a' eq) fun x y _ => by
-      cases eq.symm
-      exact match x, y with
-      | true, true | false, false => rfl
-      | false, true => h
-      | true, false => h.symm
+@[inline] private unsafe def isSharedImpl {α : Type u} (a : α) : Squash Bool := Squash.mk (isSharedUnsafe a)
+
+@[implementedBy isSharedImpl]
+def isShared {α : Type u} (a : α) : Squash Bool := Squash.mk true
+
+@[inline] def withIsShared {α : Type u} {β : Type v} (a : α) (f : Bool → β) (h : f false = f true) : β :=
+  Quot.liftOn (isShared a) f fun x y _ =>
+    match x, y with
+    | true, true | false, false => rfl
+    | false, true => h
+    | true, false => h.symm
 
 /-- Display the given message if `a` is shared, that is, RC(a) > 1 -/
 def dbgTraceIfShared {α : Type u} (s : String) (a : α) : α :=
   withIsShared a
-    (fun a _ val => if val then dbgTrace ("shared RC " ++ s) (fun _ => a) else a)
+    (fun val => if val then dbgTrace ("shared RC " ++ s) (fun _ => a) else a)
     (by simp [dbgTrace])
 
 /-- Print stack trace to stderr before evaluating given closure. Currently supported on Linux only. -/
--- a/src/include/lean/lean.h
+++ b/src/include/lean/lean.h
@@ -494,6 +494,14 @@ static inline bool lean_is_shared(lean_object * o) {
     }
 }
 
+static inline bool lean_is_exclusive_after_dec(lean_object * o) {
+    if (LEAN_LIKELY(lean_is_st(o))) {
+        return o->m_rc <= 2;
+    } else {
+        return false;
+    }
+}
+
 LEAN_SHARED void lean_mark_mt(lean_object * o);
 LEAN_SHARED void lean_mark_persistent(lean_object * o);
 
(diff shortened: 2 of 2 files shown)


gebner:
> I tried to create a version with a simpler type signature, namely
> 
> ```lean
> @[extern] unsafe def isSharedUnsafe : @& α → Bool
> ```

This is unfortunate.  I think `isShared : α → Squash Bool` would really be the best interface here.  Did you figure out why this breaks?  Do you have a concrete test case which shows the issue?

I tried to reproduce for a minute but the refcounting instructions seem to be okay here:
```
@[extern] unsafe opaque isSharedUnsafe : @& α → Bool

unsafe def a (n : Nat) : List Nat :=
  let xs := [1,n]
  if isSharedUnsafe xs then
    xs
  else
    []
```

digama0:
The refcounting instructions were also okay in my test case, and it worked as expected when compiled, but when running it with `#eval` it would act just as if the RC count was always greater than 1 (or maybe it was tripping the other check, rc = 0 = multi-threaded?). I'll post a test case once my current build finishes.

digama0:
Here's a MWE:

`lean.h`:
```c
static inline uint8_t lean_is_shared_extern(b_lean_obj_arg a) {
    return !lean_is_scalar(a) && lean_is_shared(a);
}
```
`Init/Util.lean`:
```
@[neverExtract, extern "lean_is_shared_extern"]
unsafe opaque isSharedUnsafe {α : Type u} (a : @& α) : Bool := true
```
`Test.lean`:
```lean
unsafe def foo (s : Array Nat) : Bool :=
  isSharedUnsafe (dbgTraceIfShared "bla" s)

unsafe def main := do
  let n ← liftM IO.getNumHeartbeats
  let x := #[n]
  IO.println (foo x)

#eval main -- true, no "bla" printout
```
```
$ lake build
$ build/bin/Test
false
```

gebner:
That's the expected output, right?  `x` is not shared, so it returns false.

digama0:
The result of `build/bin/Test` is expected, but the result of `#eval main` is incorrect.

gebner:
Ah, this is interesting.  Apparently the interpreter can only call native functions using the "boxed" wrapper.  And from what I can tell, the "boxed' wrapper *always* takes owned arguments.  (I'm really shocked actually, isn't this a huge linearity performance footgun for functions that take borrowed arguments?)

So what you could do is take the argument owned, and then do a `dec` first before checking for sharedness.

digama0:
> So what you could do is take the argument owned, and then do a `dec` first before checking for sharedness.

Isn't that dangerous? What if the user called it on the last reference? Calling `dec` would invalidate the object and make the access to `m_rc` invalid.

gebner:
> take the argument **owned**

I mean this:
```
@[extern "lean_is_shared_extern"]
unsafe opaque isSharedUnsafe (a : α) : Bool
```
Note the missing `@&`.

> Isn't that dangerous?

No, just unnecessary inc/dec. :shrug:

digama0:
What I mean is, when you use owned ABI then you may or may not be holding the last reference, which means that a `lean_dec` on the object might deallocate it. Accessing the object after it is deallocated is UB, unless I am misunderstanding the data representation here.

digama0:
I just realized that `is_shared` and `is_exclusive` are not actually negations of each other - they both return `false` on multithreaded objects. I think what we actually want here is `!is_exclusive()`, not `is_shared()`, at least as far as avoiding array copies and such.

gebner:
Ah yeah, you're right of course.  I guess you could change the check to `rc > 2` then.

digama0:
I added this function, and indeed it generates some nice looking IR:
```lean
def dbgTraceIfShared {α : Type u} (s : String) (a : α) : α :=
  withIsShared a
    (fun val => if val then dbgTrace ("shared RC " ++ s) (fun _ => a) else a)
    (by simp [dbgTrace])

-->

def dbgTraceIfShared._rarg._closed_1 : obj :=
  let x_1 : obj := "shared RC ";
  ret x_1
def dbgTraceIfShared._rarg (x_1 : @& obj) (x_2 : obj) : obj :=
  inc x_2;
  let x_3 : u8 := isSharedUnsafe ◾ x_2;
  case x_3 : u8 of
  Bool.false →
    ret x_2
  Bool.true →
    let x_4 : obj := dbgTraceIfShared._rarg._closed_1;
    let x_5 : obj := String.append x_4 x_1;
    let x_6 : obj := pap dbgTraceVal._rarg._lambda_1._boxed x_2;
    let x_7 : obj := dbgTrace ◾ x_5 x_6;
    ret x_7
```
It also definitely makes things easier to not have to deal with a new binding and a proof that the new value is equal to the old one. 

I'm a bit worried about this interface however, because if we don't thread the object through the extern function then nothing stops the compiler from lifting code over the `isSharedUnsafe` call, potentially breaking the check.

leodemoura:
Let's discuss this PR in the next dev meeting. 
We are currently working on the new compiler and will have to modify the interpreter too. 
For now, I will just mark this PR as a low priority.

digama0:
Ah drat, this is actually blocking some algorithms in Std that want to depend on this info. Do you have a sense of which kind of interface we can or can't support reliably here? That is, can we solve the API design question and just do something that works for now on the current compiler? I'm fairly happy with the current version after gabriel's suggested changes, although I will need to play with it more in some real use cases to see if any gotchas show up. Since it's only a performance optimization it shouldn't be a huge deal even if it returns the wrong result sometimes.

leodemoura:
> Ah drat, this is actually blocking some algorithms in Std that want to depend on this info.

Could you postpone working on these algorithms? We don't have a shortage of things to work on right now :)

> Do you have a sense of which kind of interface we can or can't support reliably here? 

Let's talk during the dev meeting next Monday. It will be great to get feedback from you, Gabriel, and Sebastian.

> I will need to play with it more in some real use cases to see if any gotchas show up.

Collecting gotchas would be very useful.

> Since it's only a performance optimization it shouldn't be a huge deal even if it returns the wrong result sometimes.

Not sure others will agree.

leodemoura:
We discussed this PR in the last dev meeting, and decided to put it on hold until we get to the ref-count support in the new code generator. This is an important primitive and we want to support it without having to perform an inc/dec to test whether an object is shared or not.

PR #1588, opened 2022-09-12
Show error message for empty instances

gebner:
This has been bugging me for a while.  When you write an instance, Lean didn't tell you what fields you need to fill in.  At least not until you correctly guessed the first one.
```lean
instance : CoeSort BundledGroup (Type u) where
  -- was it `coeSort`?  or was it `coe`?
```
Before: `error: unexpected end of input; expected identifier` (syntax error)
After: `error: missing fields: 'coe'`

### DIFF
--- a/tests/lean/instance1.lean
+++ b/tests/lean/instance1.lean
@@ -1,0 +1,9 @@
+def Mine := Id
+
+class MyMonad (m) extends Monad m where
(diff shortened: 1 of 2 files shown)


gebner:
I also agree with Mario that we should generally replace `many1` by `many` wherever possible, as with the `whereStructInst` here.

PR #1589, opened 2022-09-12
doc: some `do` extensions

Kha:
As captured in a screenshot in my slides :) . I can do the rest some other time.


digama0:
> There seem to be some bootstrap troubles with inheritDoc and built-in parsers, so I reverted that single line for now.

What's the issue?

Kha:
See https://github.com/leanprover/lean4/actions/runs/3038941090/jobs/4893276077, it's conflicting with the docstring from the previous stage

digama0:
This should work as long as you do an update stage0 between adding the doc and adding the inheritDoc, no?

Issue #404, opened 2021-04-14
Linters


leodemoura:
We need many different linters.
- Style violations.
- Common mistakes.
- Bad idioms.
- Potential performance problems.

The different linters can be developed independently of each other, and they are self-contained projects.
Core developers will help with the API and integration.

jack-ullery:
Do you want linters for the **Lean** or **C++** code (or both)?

For the C++ code, it wouldn't be very difficult for me to add tools like _clang-tidy_ and _cppcheck_ to the build. 
They could be triggered to only run when `CMAKE_BUILD_TYPE=DEBUG`, for example.

It has been a while since this issue was posted. If you're aware of any Lean targeted linters, I'd like to take a look.

digama0:
mathlib has a linter framework: https://leanprover-community.github.io/mathlib_docs/commands.html#linting%20commands . Some of this should probably be incorporated into the core lean 4 distribution, but the details have not been worked out. (My impression is that the issue is primarily about Lean linters, as the C++ code is foundational and both user code and the elaborator itself are mostly written in lean.)

digama0:
Let's close this issue now that the linter framework has been implemented, and reopen issues for specific linters.

PR #1591, opened 2022-09-12
fix: Clear Diagnostics when file is closed

lovettchris:
fix: LSP when file is closed clear any diagnostics so the "problems" list is cleared in VS code.

Note: if we want "project" level errors we should implement the full PublishDiagnostics LSP protocol, specifically the part about if a language has a project system (for example C#) diagnostics are not cleared when a file closes. When a project is opened all diagnostics for all files are recomputed (or read from a cache), etc.  I think we should implement this for lean lake projects, but it's a much bigger change.


src/Lean/Server/FileWorker.lean
```diff
@@ -310,6 +310,16 @@ section Updates
           validSnaps := validSnaps.dropLast
       unfoldCmdSnaps newMeta validSnaps.toArray cancelTk ctx
     modify fun st => { st with doc := ⟨newMeta, AsyncList.delayed newSnaps, cancelTk⟩ }
+
+  /-- called when document is closed so we can clear any diagnostics and cancel pending work -/
+  def exitWorker : WorkerM Unit := do
```
Vtec234, line 315: Nit: I would put this right at the beginning of the `MainLoop` section as it is not an update handler.

Vtec234:
My three cents: there are two unrelated changes here.

The change that clears diagnostics when the file worker exits seems for the better. Leftover diagnostics are often noisy and out-of-date. Indeed computing them for an entire Lake project as you write would be ideal, but short of that we should do the clearing.

On the other hand the information -> hint change is likely to be controversial (many users including myself are quite used to the info squiggles) — I am not sure it is the right move. I'd happily approve the PR if it were just the first of the two changes.

src/Lean/Server/FileWorker.lean
lovettchris: Done.

lovettchris:
Ok, I split the more controversial change out to separate PR:
https://github.com/leanprover/lean4/pull/1600

Vtec234 (approved):

Kha:
Should this be moved into the watchdog instead? Sending a response after receiving `exit` looks weird and may explain the supposed deadlocks in the tests.

lovettchris:
I saw the "exit" event as an ideal place to do it, just before the process terminates it sends the message. I don't see why that would create a deadlock, at worst it would mean some test doesn't see the messages get cleared ?  But it seemed very reliable in VS code when I was opening and closing files manually.  I suppose I could design a torture test to see if I can force it to break...

Vtec234:
> Sending a response after receiving `exit` looks weird and may explain the supposed deadlocks in the tests.

Hm, I missed the test failures. @lovettchris could you try clearing the diagnostics in the watchdog after sending the `exit` notification to the worker? Perhaps in `terminateFileWorker` could work.

Vtec234:
> , I fixed the hang by changing the order of these two lines in response to "shutdown":

I am not sure this is the right thing to do here. Note that from the watchdog's perspective, sending a response to the shutdown request means "we are ready to exit (i.e. kill the process)". But we are *not* ready to exit before the `shutdown` function is done, because that function ensures that the worker state is cleaned up. So I think the two statements should stay in the order they are currently at.

lovettchris:
> I am not sure this is the right thing to do here. Note that from the watchdog's perspective, sending a response to the shutdown request means "we are ready to exit (i.e. kill the process)". But we are _not_ ready to exit before the `shutdown` function is done, because that function ensures that the worker state is cleaned up. So I think the two statements should stay in the order they are currently at.

So it's not just an 'ack' that shutdown was received?   I could fix it by changing the test to expect the new diagnostics message before the ack on the shutdown...

lovettchris:
I pushed this change, so `run.lean` now loops until the shutdown response is found.

src/Lean/Data/Lsp/Communication.lean
```diff
@@ -117,6 +117,24 @@ section
 
   def writeLspResponseErrorWithData (h : FS.Stream) (e : ResponseError α) : IO Unit :=
     h.writeLspMessage e
+
+  partial def lspShutdown (hOut : FS.Stream) (hIn : FS.Stream) (requestNo : Nat) : IO Unit := do
```
Vtec234, line 121: Let's put this in `Ipc.lean` as it is only used for testing the server, and let's put a docstring explaining that it performs the LSP shutdown sequence.

tests/lean/server/edits.lean
```diff
@@ -30,8 +30,5 @@ def main : IO Unit := do
       else
         throw $ userError "Failed parsing test file."
 
-      Ipc.writeRequest ⟨3, "shutdown", Json.null⟩
-      let shutResp ← Ipc.readResponseAs 3 Json
-      assert! shutResp.result.isNull
-      Ipc.writeNotification ⟨"exit", Json.null⟩
+      Ipc.shutdown 2
```
Vtec234, line 33:
```suggestion
      Ipc.shutdown 3
```

Vtec234:
@lovettchris the changes to the tests look good. It is legitimate for the server to send more messages after receiving the `shutdown` request, so the simulated client should handle those.

On the other hand I am still not sure about the clearing of diagnostics after receiving `exit` in the worker. Is there any reason not to do it in the watchdog (e.g. in `terminateFileWorker`) before replying to `shutdown` like @Kha suggested?

src/Lean/Data/Lsp/Communication.lean
lovettchris: done.

tests/lean/server/edits.lean
lovettchris: fixed, thanks.

lovettchris:
> On the other hand I am still not sure about the clearing of diagnostics after receiving `exit` in the worker. Is there any reason not to do it in the watchdog (e.g. in `terminateFileWorker`) before replying to `shutdown` like @Kha suggested?

Ok, I moved it.

Vtec234 (approved):
Thanks, lgtm! We'll see in practice if all editors are happy with this change.

src/Lean/Server/FileWorker.lean
```diff
@@ -450,8 +457,7 @@ section MainLoop
       handleRequest id method (toJson params)
       mainLoop
     | Message.notification "exit" none =>
-      let doc := (←get).doc
-      doc.cancelTk.set
+      exitWorker
```
Kha, line 460: please revert

src/Lean/Server/Watchdog.lean
```diff
@@ -307,6 +307,9 @@ section ServerM
   def terminateFileWorker (uri : DocumentUri) : ServerM Unit := do
     let fw ← findFileWorker! uri
     try
+      let ctx ← read
+      let meta := fw.doc.meta
+      publishDiagnostics meta #[] ctx.hOut
```
Kha, line 312: I think this belongs in `forwardMessages`, where we know the file worker has actually exited and won't send out diagnostics of its own anymore. Though as @gebner mentioned afair, we still have other races such as not waiting for the old file worker to exit before starting the new one.

tests/lean/server/diags.lean
```diff
@@ -1,5 +1,8 @@
 import Lean.Data.Lsp
 open IO Lean Lsp
+open Lean
+open Lean.Lsp
+open Lean.JsonRpc
```
Kha, line 5: ?

src/Lean/Data/Lsp/Ipc.lean
```diff
@@ -39,6 +39,23 @@ def writeRequest (r : Request α) : IpcM Unit := do
 def writeNotification (n : Notification α) : IpcM Unit := do
   (←stdin).writeLspNotification n
 
+def shutdown (requestNo : Nat) : IpcM Unit := do
+  let hIn ← stdout
+  let hOut ← stdin
+  hOut.writeLspRequest ⟨requestNo, "shutdown", Json.null⟩
+  while True do
```
Kha, line 46:
```suggestion
  repeat
```

Kha (changes requested):

src/Lean/Server/Watchdog.lean
Kha: @Vtec234 Thoughts? Maybe it's okay to leave it like this for now, at least we won't race with diagnostics from the new file worker process then...

Vtec234: Hm I have unfortunately lost any mental model of what order the watchdog does things in a while ago, but I think you are right, we should wait for the worker to exit. I am not sure if we should also clear the diagnostics when the worker crashes. Maybe only doing it on termination is better, since when it crashes the "leftover" diagnostics could be helpful in diagnosing the issue.

gebner: *Ideally*, I think the diagnostics should only be cleared if the file is closed.  If the worker crashes, we should preserve the current diagnostics, show in the progress bar which part of the file was not processed due to crashing, and maybe even add an error message about the crash to the diagnostics.

The race conditions require a larger refactoring anyhow, so I'm not going to think about the best location for the bandaid.

gebner: Although this particular one here is an even larger race condition than usual...

tests/lean/server/diags.lean
lovettchris: yeah, it was a left over from before Ipc.shutdown refactor.  Fixed.

src/Lean/Server/FileWorker.lean
lovettchris: done

src/Lean/Data/Lsp/Ipc.lean
lovettchris: done, thanks.

src/Lean/Server/Watchdog.lean
lovettchris: Are you saying where it was before in FileWorker was better because that way we can ensure it is the last thing the `mainLoop` does before exiting?

Kha: The corresponding watchdog location should be before https://github.com/leanprover/lean4/blob/9f29967fb0464542cf524bbef56ae598db71fc54/src/Lean/Server/Watchdog.lean#L259

Kha: > Although this particular one here is an even larger race condition than usual...

It would be relatively simple to fix though - we'd have to move the `startFileWorker` call from `handleEdit` to the location I just mentioned above

lovettchris: Ok, folks, I did some work, along the lines outlined by Sebastian above, to ensure publish diagnostics is always the last thing sent when a document is closed, by adding a new WorkerState.Exiting (changing state to an IO.Ref) so forwardMessages can see this state change and ensure that the very last thing we do is clear the diagnostics).  

I tested it with some large Lean files that take a while to parse, quickly closing them before all the messages have been sent, and I could not get any left-over messages to remain in the output window - trying different timing on when I closed the documents.  I also tested opening multiple lean source files and closing them all at once.  Do you guys have any suggestions on what would be the trickiest test case we can try for this?

lovettchris:
All tests are passing except on Linux Debug which got a very strange error:

image

src/Lean/Server/FileWorker.lean
```diff
@@ -428,6 +428,7 @@ section MainLoop
     let state ← get
     let doc := state.doc
     doc.cancelTk.set
+    -- this is acting like an acknowledgement back to the Watchdog that exit was in fact received.
     publishDiagnostics doc.meta #[] ctx.hOut
```
Kha, line 432: This still looks wrong, there is no guarantee that all asynchronously executing, possibly diagnostics-producing tasks have finished at this point - after all, we have flagged them for cancellation only in the line above.

lovettchris: It works because the WorkerState in the watch dog has been set to “Exiting” but we just need this one more guaranteed msg from the worker to ensure we actually enter the forwardMessages loop one last time and it is there that we actually send the clear diagnostics as the very last thing and I’ve verified this is working as designed.  The problem is not when too many messages are sent from from the worker, the problem was when the worker sends none at all.

Kha: Ok, but why introduce a fake message when we already have the worker process exit "message" at the location I linked to above?

lovettchris: Oh, I see what you mean now, keep running the loop until the exception occurs with exit code 0 means normal exit, so that's a good time to clear the diagnostics, no need for a final message.  Ok, I changed it to this simpler version.

src/Lean/Server/Watchdog.lean
lovettchris: never mind, I moved the clear diagnostics to [lean4/src/Lean/Server/Watchdog.lean](https://github.com/leanprover/lean4/blob/9f29967fb0464542cf524bbef56ae598db71fc54/src/Lean/Server/Watchdog.lean#L259)

Kha (approved):
Please revert the stray changes

src/Lean/Server/Watchdog.lean
```diff
@@ -336,7 +338,7 @@ section ServerM
       | none    => (false, none)
     if pendingEdit then
       return
-    match fw.state with
+    match (fw.state) with
```
leodemoura, line 341: Is this a workaround for some bug in Lean? We should not need parentheses here.

lovettchris: no just a left over from a previous edit were I was using a `←` statement, I'm reverting the change.

digama0: by the way you don't need parentheses even when you use `<-` in match. Speaking of which @leodemoura , I notice that it is fairly common to see `match (<- foo) with`  in the lean sources, is this a deliberate style choice or did the precedence change at some point? I usually leave the parentheses off in this situation.

src/Lean/Server/FileWorker.lean
lovettchris: Here's the updated diagram:

image

lovettchris:
> Please revert the stray changes

Done.

src/Lean/Server/Watchdog.lean
leodemoura: Yes, this is a deliberate style choice. I also like to avoid `(<-foo)` without a space between `<-` and `foo`. Sometimes I regret not enforcing the parentheses around `<- foo`. I would like the code in this repo to use a consistent style.

digama0: Good to know, I'll keep that in mind.

lovettchris: I notice we have some linting stuff happening, can this be added to the linter?

digama0: I was just about to say the same thing. Could you open an issue to track this @lovettchris ? There are plenty of places to use a missing / confusing parentheses linter.

leodemoura: Yes, it can be done with a linter, but please no linter PR unless it is almost perfect. This linter is very low priority.

digama0: Just the issue for now, I think, so we don't forget. Style linters generally will need some design work.

leodemoura: Yes, an issue is perfect.

lovettchris: There's already this one, is this enough?
https://github.com/leanprover/lean4/issues/404

PR #1594, opened 2022-09-13
doc: fix pipeline doc and add info on abbreviations.

lovettchris:
See [Zulip Thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Lean.20doc.20suggestion)

### DIFF
--- a/doc/functions.md
+++ b/doc/functions.md
@@ -70,7 +70,7 @@ If we were able to partially define `loop?`, we could prove `False` with it.
 A lambda expression is an unnamed function.
 You define lambda expressions by using the `fun` keyword. A lambda expression resembles a function definition, except that instead of the `:=` token,
 the `=>` token is used to separate the argument list from the function body. As in a regular function definition,
-the argument types can be inferred or specified explicitly, and the return type of the lambda expression is inferred from the type of the
+the argument types can be inferred or specified explicitly, and the return type of the lambda expression is always inferred from the type of the
 last expression in the body.
 
 ```lean
@@ -89,9 +89,10 @@ def twice (f : Nat -> Nat) (x : Nat) : Nat :=
 
 Note: you can also use the Unicode symbol `λ` instead of `fun` which you can type into VS Code using the abbreviation `\la `.
 
-While you cannot specify an explicit return type in a lambda expression you can use the `(e : T)` in the body to
-declare the type of the expression `e`.  For example: `λ x => (x : Int)` would be able to convert a given value of
-type `Nat` to an `Int` using coercion.
+While you cannot specify an explicit return type in a lambda expression you can use the type ascription
+expression `(e : T)` in the body to declare a new type for the expression `e`.
+For example: `λ (x : Nat) => (x : Int)` is able to convert a given value of type `Nat` to an `Int` using coercion.
+
 
 
 # Syntax sugar for simple lambda expressions


doc/quickstart.md
```diff
@@ -43,6 +43,11 @@ On Linux/macOS, you first have to follow the instructions printed by the Lean in
 Note: Packages **have** to be opened using "File > Open Folder..." for imports to work.
 Saved changes are visible in other files after running "Lean 4: Refresh File Dependencies" (`Ctrl+Shift+X`).
 
+## Abbreviations
+
+The VS Code editor helps you type nice Unicode characters supported by Lean,
```
leodemoura, line 48: I don't want to have `abbreviations.json` in this repo. We can say users can use the VS Code command "Lean 4: Show all abbreviations" to list all abbreviations. We can also remind them that when hovering over Unicode symbols, VS Code will show how to input them.

lovettchris: Ok, done.  The intro to abbreviations in the quickstart now says this:

image

doc/functions.md
```diff
@@ -1,7 +1,7 @@
 # Functions
 
 Functions are the fundamental unit of program execution in any programming language.
-As in other languages, a Lean function has a name, can have parameters and take arguments, and has a body.
+As in other languages, a Lean function has a name, can have parameters and take arguments, have a body and return a value.
```
leodemoura, line 4: Please revert this.

doc/functions.md
```diff
@@ -69,7 +69,8 @@ If we were able to partially define `loop?`, we could prove `False` with it.
 
 A lambda expression is an unnamed function.
 You define lambda expressions by using the `fun` keyword. A lambda expression resembles a function definition, except that instead of the `:=` token,
-the `=>` token is used to separate the argument list from the function body. As in a regular function definition,
+the `=>` token is used to separate the argument list from the function body and you cannot specify an explicit return type, it is always inferred.
```
leodemoura, line 72: This is a bit misleading. We have type ascriptions. Example:
```lean
#check fun x : Nat => (x : Int)
-- Nat -> Int
```

lovettchris: Ok, wow, I've never seen that one before, it is extremely weird to say "x : Int" it sounds like you are saying x is an Int which directly contradicts the earlier statement that x is a Nat.  I'm talking about this error:

image

digama0: That's just syntactically invalid, there is no slot to put a type there in the syntax. I would be concerned about syntactic ambiguity if you could.

lovettchris: I know, and I'm not proposing we add it, I'm proposing we document this very thing "there is no slot to put a type there in the syntax."

leodemoura: > Ok, wow, I've never seen that one before, it is extremely weird to say "x : Int" it sounds like you are saying x is an Int which directly contradicts the earlier statement that x is a Nat.

@lovettchris I strongly recommend you read "Theorem Proving in Lean 4" again. This notation is addressed at the beginning of the book. Here is the quote from the book: "One can always specify the type T of an expression e by writing (e : T). This instructs Lean's elaborator to use the value T as the type of e when trying to resolve implicit arguments."

lovettchris: I think you misunderstood my comment.  I'm not saying I've never seen "name colon type", I've just never seen type ascription used this particular way:
```lean
#check fun x : Nat => (x : Int)
```

What is => pointing at here?  It doesn't look like a function body.  What do the parens mean here?  Is that a tuple ?  No it's not a tuple.  The "x: Int" is clear, it is saying x is type Int.  But that contradicts the earlier statement "x : Nat" which says x is a Nat.  So I'm still very confused...

lovettchris: Perhaps this is a "casting operation", or a type of "explicit coercion" where you are "forcing x to become an Int" -- in C++ I would have written `(Int)x`.  Is that what this is?  Lean has "casting" ? Where is casting documented?  Perhaps your term for casting is "type ascription", but that term doesn't mean "casting" to me.  I see this also works:

```lean
def natToInt (p : Nat) : Int := (p : Int)
```

But this does not work:
```lean
def natToFloat (p : Nat) : Float := (p : Float)
/-
type mismatch
  p
has type
  Nat : Type
but is expected to have type
  Float : Type
-/
```

leodemoura: >  I'm not saying I've never seen "name colon type", I've just never seen type ascription used this particular way:

@lovettchris Type ascription is explained in "Theorem Proving in Lean 4," I quoted the exact passage that explains it. Chapter 2 has more information. I understand you think this book is not very accessible, but I am sure it will be much easier to understand the book now that you have more experience with Lean.

lovettchris: where exactly?

image

leodemoura: @lovettchris Did you read my message? I gave the exact quote.

lovettchris: If you are talking about the paragraph starting with `One can always specify the type T of an expression e by writing (e : T). This instructs Lean's elaborator to use the value T as the type of e when trying to resolve implicit arguments` then sure, this is the general definition of `e : T`, but it says nothing here about using that as a method body in a type of very limited type coercion operation...

leodemoura: @lovettchris This is not the general definition of `e : T`. Please read the book again. It is introducing the notation here, but it is not the only occurrence. The notion of coercion is introduced later. The type ascription `(e : T)` is used to specify the expected type. Everything will make more sense after you read the book again.

lovettchris: Unless Lean has some other general behavior that says any expression can be used anywhere, so of course you can ascribe `(e : T)` as the entirety of a function body... or something.  In playing with this I see:

```lean
def NatToInt (p: Nat) := 
  let x := p
  (x : Int)
```
works too, but x is not Nat in `let x := p`, so this means `(x : Int)` is indeed a coercion.  It's saying the expression "x" is the return value, but I want you to change the type of this expression to `Int` using the special syntax `(x : Nat)` which requires parentheses.  Perhaps this is what the Theorem Proving paragraph was trying to say, but I didn't read it that way, I read it as just explaining the generally "name colon type" syntax that is used everywhere in Lean...

leodemoura: @lovettchris If you want to help me, please read the book. It will save us time, and you will have a much better understanding of what is going on.

lovettchris: I see now it works more generally too:
```lean4
def NatToIntPair : Nat × Nat →  Int × Int 
  | (x, y) => ((x : Int), (y : Int))
```
so it's starting to make sense.  Perhaps this is just a case of "the TPIL book is very dense" and I missed nuance here that you can use this type ascription syntax to force some simple type coercions to happen, a bit like casting, but cannot cast Float to Int or Int to Float or Int to Nat, it can only do Nat to Int...

leodemoura: > Perhaps this is just a case of "the TPIL book is very dense" and I missed nuance here that you can use this type ascription syntax to force some simple type coercions to happen, a bit like casting, but cannot cast Float to Int or Int to Float or Int to Nat, it can only do Nat to Int...

TPIL will make much more sense in a second read. TPIL has exercises, I suggest you try to solve all of them. It will help for use. Many people in the community managed to learn Lean this way and succeeded.

doc/functions.md
lovettchris: It just seemed weird introducing functions without mentioning a return value, which is arguably the most important thing about a function.  How about this instead `have a body which returns a value` ?

leodemoura: @lovettchris The body is the value.

lovettchris: Great, can we say that then?

leodemoura: @lovettchris What about learning more about Lean before trying to edit the existing documentation?

doc/functions.md
lovettchris: Well, I recompiled Lean without `instance : Coe Nat Int := ⟨Int.ofNat⟩` and indeed the `#check fun x : Nat => (x : Int)` no longer compiles, so (e : T) is not only specifying the type of the expression it is triggering coercion as described in [Coercions using Type Classes](https://leanprover.github.io/theorem_proving_in_lean4/type_classes.html#coercions-using-type-classes).  The problem with that section is it glosses over this easy case never showing this example, and dives right into much more complicated stuff involving _class of sorts_ and _class of function types_ and so on, and not once in the entire section does it use `(e : T)` syntax as a forced coercion.  It is shown in an #eval in https://leanprover.github.io/lean4/doc/nat.html but that is making the literal 5 be interpreted as an Int, and this is not coercion because that still compiles after I made that change to Lean. PS: There are also no example exercises under `Coercions using Type Classes`... 

But I now understand that (e : T) works for any coercion, for example, I see there's an `instance : Coe String IO.Error := ⟨IO.userError⟩` and so I can write `#check ("test" : IO.Error)` and get an `-- IO.userError "test" : IO.Error`.  Previously when reading the Coercions section and reading lots of Lean code I had only gleaned that Coercions allowed you to do this sort of thing where coercion is transparently happening to match specified arg types or return types.  

```lean
def foo (e : IO.Error) :=  e
#eval foo "test"    
```
In fact, in my change to Lean removing `instance : Coe Nat Int := ⟨Int.ofNat⟩` I found only 3 instances of `(e : Int)` in the entire code base, (2 in JSon Basic.lean and 1 in Congr.lean) so I think it's not too surprising I somehow missed it.

doc/functions.md
lovettchris: I updated the text to say `As in other languages, a Lean function has a name, can have parameters and take arguments, declare a return type and have a body.` 

perhaps that satisfies both needs, that a "return type" is mentioned somewhere, but no mention of "return value" since body is the return value.

doc/functions.md
lovettchris: So to close this thread, I removed the contentious statement about no declared return type and added this paragraph which adds the missing information and link between type ascription and coercion:

> While you cannot specify an explicit return type in a lambda expression you can use the type ascription
expression `(e : T)` in the body to declare a new type for the expression `e`.
For example: `λ (x : Nat) => (x : Int)` is able to convert a given value of type `Nat` to an `Int` using coercion.

doc/functions.md
```diff
@@ -121,9 +129,22 @@ def times2 x := x * 2
 #eval 100 |> add1 |> times2
 #eval times2 <| add1 <| 100
 ```
-The result of the previous `#eval` commands is 202.
-The forward pipeline `|>` operator takes a function and an argument and return a value.
-In contrast, the backward pipeline `<|` operator takes an argument and a function and returns a value.
+The result of all three previous `#eval` commands is 202.
+
+So `x |> f` means the same as the same as `f x`, and it chains such that `x |> f |> g` is
```
TomasPuverle, line 134: suggestion:

So `x |> f` means the same as the same as `f x`, and ~it chains such that~ `x |> f |> g` is

doc/functions.md
```diff
@@ -121,9 +129,22 @@ def times2 x := x * 2
...
+interpreted as `g (f x)`.
+
+Similarly `f <| x` means the same as the same as `f x`, except that it parses x with lower precedence,
+which means that `f <| g <| x` is interpreted as `f (g x)` rather than `(f g) x`.
```
TomasPuverle, line 138: Similarly `f <| x` means the same as the same as `f x`, ~except that it parses x with lower precedence,
which means that~, and `f <| g <| x` is interpreted as `f (g x)` ~rather than `(f g) x`.~

doc/functions.md
lovettchris: I got this directly from `Notation.lean` - is there a problem with saying "it chains" ?

doc/functions.md
TomasPuverle: No issue, just suggestions from reading it. 
Also, I think that ‘parses with lower precedence’ here is really intending to say ‘is right associative’ because the lower precedence is just an implementation detail of a precedence climbing/Pratt expression parser.

Thanks!

digama0: The precedence is important for more than the associativity of `<|` itself, it also comes up when comparing `<|` to other operators like `+` - `f <| x + y` means `f (x + y)` and not `f x + y`.

leodemoura:
I am closing this PR, the hovers for `|>` and `<|` have a nice explanation now.

PR #1595, opened 2022-09-13
fix: make all syntax accessors non-panicking

gebner:
See https://leanprover.zulipchat.com/#narrow/stream/341532-lean4-dev/topic/Panics.20in.20syntax.20accessors/near/298554338.  It is very easy to hit the panic cases in these accessors, since they are only triggered with invalid syntax from parsing errors.  For example:
```lean
macro "same" &"issue" n:num : term =>
  open Lean in pure (quote n.getNat : Term)

#check same issue -- PANIC
```

Some syntax functions are already non-panicking: `stx[i]` and `stx.getIdent`.  This PR adopts the non-panicking behavior uniformly, and makes the other functions non-panicking as well.

### DIFF
--- a/src/Init/Meta.lean
+++ b/src/Init/Meta.lean
@@ -901,22 +901,22 @@ end Syntax
 namespace TSyntax
 
 def getNat (s : NumLit) : Nat :=
-  s.raw.isNatLit?.get!
+  s.raw.isNatLit?.getD 0
 
 def getId (s : Ident) : Name :=
   s.raw.getId
 
 def getScientific (s : ScientificLit) : Nat × Bool × Nat :=
-  s.raw.isScientificLit?.get!
+  s.raw.isScientificLit?.getD (0, false, 0)
 
 def getString (s : StrLit) : String :=
-  s.raw.isStrLit?.get!
+  s.raw.isStrLit?.getD ""
 
 def getChar (s : CharLit) : Char :=
-  s.raw.isCharLit?.get!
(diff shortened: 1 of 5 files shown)


src/Init/Meta.lean
```diff
@@ -901,22 +901,22 @@ end Syntax
...
 def getNat (s : NumLit) : Nat :=
-  s.raw.isNatLit?.get!
+  s.raw.isNatLit?.getD 0
 
 def getId (s : Ident) : Name :=
   s.raw.getId
 
 def getScientific (s : ScientificLit) : Nat × Bool × Nat :=
-  s.raw.isScientificLit?.get!
+  s.raw.isScientificLit?.getD (0, false, 0)
```
digama0, line 910: can we have an `Option.getD default` that doesn't panic? This looks a little silly.

gebner: In this case it actually makes some sense to specify the default value (since the tuple has a special interpretation).  I hope that I picked the equivalent of `0e0` here.

Ideally, we'd have a custom type here though:
```lean
structure ScientificLitDecoded where
  mantissa : Nat
  isExponentNegative : Bool
  exponent : Nat

instance : Inhabited ScientificLitDecoded where
  default := { mantissa := 0, exponent := 0, isExponentNegative := false }
```

leodemoura: @gebner I am happy to merge as-is. Do you want to make the change above before merging?

PR #1597, opened 2022-09-13
chore: remove getElem', use a custom delab

digama0:
Otherwise you have to `simp` with it, which makes it inadvisable to use in the statements of theorems, which partially defeats the purpose of the notation in the first place.

### DIFF
--- a/src/Init/NotationExtra.lean
+++ b/src/Init/NotationExtra.lean
@@ -198,6 +198,10 @@ syntax (name := calcTactic) "calc" ppLine withPosition(calcStep) ppLine withPosi
   | `($f [$k] $sep)  => `($f $k $sep)
   | _                => throw ()
 
+@[appUnexpander GetElem.getElem] def unexpandGetElem : Lean.PrettyPrinter.Unexpander
+  | `($_ $array $index $_) => `($array[$index])
+  | _ => throw ()
+
 @[appUnexpander getElem!] def unexpandGetElem! : Lean.PrettyPrinter.Unexpander
   | `($_ $array $index) => `($array[$index]!)
   | _ => throw ()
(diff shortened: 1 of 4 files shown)


leodemoura (approved):

gebner:
Could we just remove it instead?  I've already put the following in mathlib4:
```lean
macro_rules | `($x[$i]'$h) => `(getElem $x $i $h)
```
(One reason is the simp issue that you've mentioned.  The other is that it is annoying to see the proof terms.  If I want to see the proof term I can just look at the infoview.  Also I don't think it is a good idea to couple pretty-printing options to the input syntax, whether you want to see the proof term doesn't really depend on whether you wrote it manually in the first place.)

digama0:
Ah, that's a fair point. I'd actually copied that into Std without realizing, but it wasn't imported in the file I was working on. That suits me just fine, I prefer not to add extra junk in the terms for notational reasons.

leodemoura:
> Ah, that's a fair point. I'd actually copied that into Std without realizing, but it wasn't imported in the file I was working on. That suits me just fine, I prefer not to add extra junk in the terms for notational reasons.

Are you going to update the PR?

digama0:
I started working on this, and I ran into a hiccup with the unexpander monad. (cc: @Kha) I was thinking of making the unexpander check the pp.proofs option to determine whether to use the `x[i]'h` or `x[i]` form, but it seems like `UnexpandM` is very impoverished, it is literally just `Syntax -> Option Syntax`, so checking options is out of the question. What is the reason for using this kind of monad for the unexpander?

Kha:
`UnexpandM` is declared as early as possible, so the available types are limited

gebner:
It's an interesting choice to tie this to `pp.proofs`.  Note that `pp.proofs` is defaulted to true when you're pretty-printing a proof term (i.e. something which has a proposition as type).  (Personally I think the infoview is good enough and we don't need the fancy delab.)

PR #1598, opened 2022-09-13
feat: move docstring before | in ctors

gebner:
As discussed on Zulip: https://leanprover.zulipchat.com/#narrow/stream/341532-lean4-dev/topic/.5BRFC.5D.20Syntax.20of.20documentation.20comments.20for.20constructors/near/298579241

### DIFF
--- a/src/Lean/Elab/Inductive.lean
+++ b/src/Lean/Elab/Inductive.lean
@@ -787,7 +787,7 @@ private def mkInductiveDecl (vars : Array Expr) (views : Array InductiveView) : 
       for view in views do
         Term.addTermInfo' view.ref[1] (← mkConstWithLevelParams view.declName) (isBinder := true)
         for ctor in view.ctors do
-          Term.addTermInfo' ctor.ref[2] (← mkConstWithLevelParams ctor.declName) (isBinder := true)
+          Term.addTermInfo' ctor.ref[3] (← mkConstWithLevelParams ctor.declName) (isBinder := true)
         -- We need to invoke `applyAttributes` because `class` is implemented as an attribute.
         Term.applyAttributesAt view.declName view.modifiers.attrs .afterTypeChecking
 


src/Lean/Parser/Command.lean
```diff
@@ -77,7 +77,7 @@ def «instance»       := leading_parser Term.attrKind >> "instance" >> optNamed
 def «axiom»          := leading_parser "axiom " >> declId >> ppIndent declSig
 /- As `declSig` starts with a space, "example" does not need a trailing space. -/
 def «example»        := leading_parser "example" >> ppIndent optDeclSig >> declVal
-def ctor             := leading_parser "\n| " >> ppIndent (declModifiers true >> rawIdent >> optDeclSig)
+def ctor             := leading_parser atomic (declModifiers true >> "\n| ") >> ppGroup (rawIdent >> optDeclSig)
```
gebner, line 80: I'm not sure if using `atomic` here has any undesirable performance consequences.  Without the atomic, the inductive would eat the docstring of the following definition (and turn it into a constructor with a parse error).

leodemoura (approved):

src/Init/Prelude.lean
```diff
@@ -3361,22 +3361,20 @@ end Name
...
+  -/
+  | original (leading : Substring) (pos : String.Pos) (trailing : Substring) (endPos : String.Pos)
+  /--
+  Synthesized syntax (e.g. from a quotation) annotated with a span from the original source.
+  In the delaborator, we "misuse" this constructor to store synthetic positions identifying
+  subterms.
+  -/
+  | synthetic (pos : String.Pos) (endPos : String.Pos)
+  /-- Synthesized token without position information. -/
+  protected | none
```
gebner, line 3377: This is maybe the only stylistic downside of moving all modifiers before the pipe.  We could split the doc comment and the other modifiers, but that requires modifying the `elabModifiers` code which is written in the old `stx[5][0]` style that I don't really want to touch unless necessary.

Moving the `protected` after the pipe again is a much smaller change that we can also do in a follow-up PR.

leodemoura: AFAIR, this was the motivation for using the weird `| /-- ... -/ ctor` syntax: we avoided the weird `protected | none`

src/Lean/Parser/Command.lean
leodemoura: This is fine.

src/Init/Prelude.lean
leodemoura: What about the following hack?
- Leave `declModifiers` where it is.
- Add a `docComment` before `|`
- Return error at elaboration time if `declModifiers` contains a `docComment`.

digama0: That sounds good to me, that's pretty much the same as the hack I suggested to handle `let rec` which has a similar issue

leodemoura: > That sounds good to me, that's pretty much the same as the hack I suggested to handle `let rec` which has a similar issue

Yeah. Sorry, missed your comment at Zulip.

gebner: Okay, I've implemented the "hacky" solution now.

src/Lean/Elab/Inductive.lean
```diff
@@ -787,7 +787,7 @@ private def mkInductiveDecl (vars : Array Expr) (views : Array InductiveView) :
       for view in views do
         Term.addTermInfo' view.ref[1] (← mkConstWithLevelParams view.declName) (isBinder := true)
         for ctor in view.ctors do
-          Term.addTermInfo' ctor.ref[2] (← mkConstWithLevelParams ctor.declName) (isBinder := true)
+          Term.addTermInfo' ctor.ref[3] (← mkConstWithLevelParams ctor.declName) (isBinder := true)
```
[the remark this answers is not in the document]
leodemoura: I can empathize. I had similar unpleasant experiences in the past. However, I am reluctant to fully move to the match-syntax style since I had even greater unpleasant experiences using it. The match-syntax has improved a lot over time, and many of these horrible experiences from the past would not happen today, but I want to move slowly to the match-syntax style.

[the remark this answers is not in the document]
digama0: I'm a bit concerned about the match syntax not being zero-cost compared to using `ref[2]`. I would like to have a version of `ref[2]` that plays better with "find all references" and such (i.e. being actually tied to the syntax type involved) but isn't actually more code than this at runtime and preferably not too expensive at compile time either (I've noticed some compile time performance issues with syntax match in the past).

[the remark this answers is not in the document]
leodemoura: @digama0 Another topic for the next dev meeting :)

Issue #1590, opened 2022-09-12
Make blue sqiggles go away


lovettchris:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

See [Zulip Thread](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/blue.20squiggly.20lines).

The LSP reports output messages (e.g. from #eval) as DiagnosticSeverity.information which results in them showing up as blue squiggles and as "Problems" in the VS Code Output tab.  If we switch them to DiagnosticSeverity.hint then the squiggles are replaced with a nice little triple dot you can hover over and the Problems output tab is cleaned up:

![image](https://user-images.githubusercontent.com/18707114/189755290-7cd700da-db71-4ad2-9928-6090dbfeb81c.png)

### Steps to Reproduce

```lean
import Foo

def main : IO Unit :=
  IO.println s!"Hello, {hello}!"

#eval Lean.versionString

#eval main
```
**Expected behavior:** no blue squiggles or problems shown in Problems output tab

**Actual behavior:** blue squiggles and problems shown in Problems output tab

**Reproduces how often:** 100%

### Versions

Lean 4 latest

PatrickMassot:
From a semantic point of view, I think that "Information" fits better than "Hint" here. Maybe the proper solution is some theme customization?

lovettchris:
The user doesn't really see whether it is "information" or "hint" though right?  It's just how we choose to map things to the LSP protocol.

digama0:
The user does see "information" vs "hint" quite clearly, as you've observed: "information" shows up with the info icon in the Problems view. They are eye catching and bring your attention to them from anywhere in the project, which is exactly the intended behavior for something that is the equivalent of a write to stdout. "Hint" can only be seen when you happen to be looking in the area (there is no global list of them AFAIK) and plays a role more similar to code lens or other local information gathering.

That is all to say that I agree with Patrick: lean is using the correct mechanism to signal the output of output-producing operations in the IDE. If you want to make the blue squiggles go away, stop writing `#eval` and `#check`.

leodemoura:
Closing this issue. See feedback above.

lovettchris:
[...]

PR #1600, opened 2022-09-13
fix: change information to hint so #eval output doesn't show up in "problems"

lovettchris:
Fixes https://github.com/leanprover/lean4/issues/1590.

In LSP Change 'info' to 'hint' to eliminate noisy blue squiggles  and stop them from showing up as "problems" in VS code.

The "hints" now show up with a more subtle ellipsis UI with hover tips containing the message and the problem list remains empty (unless you have an actual error in your lean program):

image

Without this change all blue squiggles show up in the "Problems" output window which is a bit weird and could cause confusion with other languages that use blue squiggles to signify things that you need to work on - like spell checkers often use blue squiggles.

image


leodemoura:
Closing this PR. See feedback at issue #1591

PR #1601, opened 2022-09-14
chore: import reductions

digama0:
In an attempt to decrease rebuild time, I took a look at the import graph to determine if there are any unnecessarily linearized import chains. The result was that the longest import chain (entirely contained in the `Lean` package) decreased from 71 files down to 50, and the median number of dependencies of a random file (i.e. the things that have to be rebuilt) decreased from 37 to 20.

That said, this was just done by hack and slash on the imports of some critical files as long as removing them didn't break the build. I tried not to gratuitously remove transitive imports since this can make moving things around more difficult, but it's hard to tell which imports are actually important so I may have removed some by accident.

Besides the import changes, there were two code moves:

* `NameMap` and `NameSet` were moved out of `Data.Lean.Name` and into a new file `Data.Lean.NameMap`. That way you don't need hash sets to talk about `Name`.
* `Lean.Elab.MutualDef` now imports `Lean.Elab.Deriving.Basic` instead of the other way around. The common code they require, `processDefDeriving`, was moved from `Lean.Elab.MutualDef` to `Lean.Elab.Deriving.Basic`. This avoids deriving handlers from needing to import the somewhat heavy `PreDefinition` module.

### DIFF
--- a/tests/lean/1021.lean.expected.out
+++ b/tests/lean/1021.lean.expected.out
@@ -1,8 +1,8 @@
-some { range := { pos := { line := 136, column := 41 },
+some { range := { pos := { line := 134, column := 41 },
     charUtf16 := 41,
-    endPos := { line := 142, column := 31 },
+    endPos := { line := 140, column := 31 },
     endCharUtf16 := 31 },
-  selectionRange := { pos := { line := 136, column := 45 },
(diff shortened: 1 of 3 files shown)


gebner (approved):
You should also file a lake PR with the changes.

leodemoura (approved):

Issue #1469, opened 2022-08-12
Highlight changes in goal state


Kha:
tl;dr: https://twitter.com/famontesi/status/1557346899610443782

We already have both the state before and after a tactic in the info tree, we should use them at the same time! A full diff is probably too noisy since the old state is usually less interesting than the new one, as long as you can see where it changed at all. We could probably do this by diffing the `TaggedText`s, introducing a new kind of `MsgEmbed`. The Coq implementation is documented at https://coq.inria.fr/distrib/current/refman/proofs/writing-proofs/proof-mode.html#how-diffs-are-calculated.

EdAyers:
I am keen to work on this

EdAyers:
I can think of two ways of implementing this:

1. VSCode's goal view component intercepts rendering of tagged text and highlights subexpressions that have changed since the last render. React provides some mechanisms for remembering what the last state was, so this could be done entirely on the client.
2. Lean computes the diffs in goal state, each fileworker remembers the last goal that was requested and alongside the InteractiveGoal object, we also pass an InteractiveGoalDiff, which in turn includes an ExprDiff which the vscode extension can then render.

Kha:
I don't think it should depend on the navigation history but always, deterministically provide the diff between the selected tactic's pre and post state - which would be convenient because that's exactly the information we already have in every `TacticInfo`.
I didn't check Coq's behavior myself, but from the description I'm assuming they do this as well.

PR #1607, opened 2022-09-15
feat: preliminary definitions for goal-diffs

EdAyers:
While working on #1469 I have some additions to core that are not directly related to goal-diffs. So I've added them here.
Happy to break up into separate PRs.

Changes:
- `Option.merge`
- `withPos` versions for  `visitLambda`, `forEachExpr`
- non-variable-instantiating version of `visitChildren`
- `refactor Expr.find?` to use a common `findChildren` recursor.
- Introduce monadic `Expr.find?`
- prefix methods for `SubExpr.Pos`
- some docs for `MetaVarContext`

### DIFF
--- a/src/Lean/SubExpr.lean
+++ b/src/Lean/SubExpr.lean
@@ -304,7 +304,6 @@ def visitLetWithPos (f : Pos → Expr → m Unit) (p : Pos) (e : Expr) : m Unit 
 Includes a `Pos` parameter that is used to track position in the subexpression.
 
 Instantiates bound variables if the subexpression is below a binder.
-For that, use `Lean.Meta.visitChildrenWithPos`.
 -/
 def visitChildrenWithPos (visit : Pos → Expr → m Unit) (p : Pos) (e : Expr) :=
   match e with


src/Lean/MetavarContext.lean
```diff
@@ -240,10 +240,12 @@ def LocalInstances.erase (localInsts : LocalInstances) (fvarId : FVarId) : Local
   | some idx => localInsts.eraseIdx idx
   | _        => localInsts
 
+/-- A kind for the metavariable that determines its unification behaviour.
+For more information see the large comment at the beginning of this file. -/
 inductive MetavarKind where
-  | natural
-  | synthetic
-  | syntheticOpaque
+  | /-- Normal unification behaviour -/ natural
```
leodemoura, line 246: We are putting the docstrings before the `|` now.
See recent PR from @gebner

src/Lean/SubExpr.lean
```diff
@@ -150,17 +167,76 @@ def mkRoot (e : Expr) : SubExpr := ⟨e, Pos.root⟩
...
+  match e with
+  | .app f a =>
+    visitAppWithPos visit p.pushAppFn f
+    *> visit p.pushAppArg a
+  | e => visit p e
+
+end Visit
+
+section Find
+variable {m} [Alternative m]
```
leodemoura, line 202: You don't need `{m}`.

src/Lean/MetavarContext.lean
```diff
@@ -240,10 +240,12 @@ def LocalInstances.erase (localInsts : LocalInstances) (fvarId : FVarId) : Local
   | some idx => localInsts.eraseIdx idx
   | _        => localInsts
```
leodemoura, line 242: It would be great to get the docstrings for `MetavarContext.lean` in a separate PR.

src/Lean/Util/FindExpr.lean
```diff
@@ -53,19 +61,13 @@ unsafe def findUnsafe? (p : Expr → Bool) (e : Expr) : Option Expr :=
 
 end FindImpl
 
+/-- Depth-first explores the subexpressions returning the first subexpression such that the predicate `p` returns true.-/
 @[implementedBy FindImpl.findUnsafe?]
-def find? (p : Expr → Bool) (e : Expr) : Option Expr :=
+partial def find? (p : Expr → Bool) (e : Expr) : Option Expr :=
```
leodemoura, line 66: Let's mark it as `opaque`, and delete the body. I don't see an application for a `partial` reference implementation. We can't use it for reasoning.

src/Lean/MetavarContext.lean
digama0: It works in both places now, right? As a style point, I would suggest that we put it after the `|` if the comment is on the same line as the `|`, because otherwise the `|` will appear not column aligned which I think will be a bad idea following #1612. (We may or may not additionally recommend to *always* put docs on a separate line from the constructor name; it makes alignment in general kind of messy when you do this.)

gebner: The column alignment will apply to the whole constructor node.  So if the docstring comes first, the docstring needs to be at the right column.  If the pipe comes first, then the pipe needs to be aligned.  Both styles will be possible.

gebner: In general, I would rather tend to disallow docstring-after-the-pipe entirely (at elaboration time).

src/Lean/MetavarContext.lean
EdAyers: #1613

EdAyers:
After review there are a few things in this PR that are not actually needed to get a working version of the goal diff.  Maybe #1610 should be reviewed first.

src/Lean/Util/FindExpr.lean
```diff
@@ -8,6 +8,21 @@ import Lean.Expr
 namespace Lean
 namespace Expr
 
+/-- Run the given `visit` function on each _immediate_ child subexpression of the given expression.
+It will fail if there are no subexpressions.
+
+Does not instantiate bound variables if the subexpression is below a binder.
+Usage with monads deriving from `MetaM` is not recommended.
+-/
+def findChildren [Alternative m] (visit : Expr → m α) : Expr → m α
```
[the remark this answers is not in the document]
EdAyers: possibly silly question -- where can I see the generated code?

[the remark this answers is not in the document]
leodemoura: If you want to see the IR for a function without ref-count instructions. You can add
```
set_option trace.compiler.ir.init true
```
The output is a bit hard to read since it contains new functions that have been lambda lifted, ground constants that have been extracted, and code that has been specialized.

leodemoura:
@EdAyers Could you also please rebase? I have already merged some related PRs.

src/Lean/Meta/ForEachExpr.lean
```diff
@@ -4,47 +4,124 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+      f p.pushLetValue v
       withLetDecl n d v fun x =>
-        visit (fvars.push x) b
+        visit (fvars.push x) p.pushLetBody b
     | e => do
-      f <| e.instantiateRev fvars
+      f p <| e.instantiateRev fvars
+
+/-- Given an expression `e = fun (x₁ : α₁) .. (xₙ : αₙ) => b`, runs `f` on each `αᵢ` and `b`. -/
+def visitLambda (f : Expr → m Unit) (e : Expr) := forgetPos visitLambdaWithPos f e
```
leodemoura, line 90: I am concerned about the performance implications of changes like this one. 
It feels like `forEachExpr'` is getting more abstract and slower.
I am not sure the current code generator is eliminating all the overhead introduced here. I would have to inspect the generated code.
BTW, I think the transition to `visitLambda (f : Expr -> m Unit) ...` had already introduced overhead, and this one is adding even more overhead.

EdAyers: I see, I'll revert the changes for now.

EdAyers:
Closing this because not needed anymore

PR #1609, opened 2022-09-15
feat: remove bounds checks in Array.{reverse, insertAt}

digama0:
`reverse` is slightly simpler with two indexes instead of one, and I think it should be less to execute too. `insertAt` looks like it should be using a C-implemented primitive for doing bulk moves in an array, but this implementation is okay otherwise. I also factored out a non-panicking version of `insertAt`, and the original has been renamed to `insertAt!`.

### DIFF
--- a/src/Init/Data/Array/Basic.lean
+++ b/src/Init/Data/Array/Basic.lean
@@ -717,14 +717,13 @@ def erase [BEq α] (as : Array α) (a : α) : Array α :=
   let j := as.size
   let as := as.push a
   loop as ⟨j, size_push .. ▸ j.lt_succ_self⟩
-termination_by _ _ _ j => j.1
+termination_by loop j => j.1
 
 /-- Insert element `a` at position `i`. Panics if `i` is not `i ≤ as.size`. -/
 def insertAt! (as : Array α) (i : Nat) (a : α) : Array α :=
   if h : i ≤ as.size then
     insertAt as ⟨i, Nat.lt_succ_of_le h⟩ a
   else panic! "invalid index"
-termination_by _ _ _ j => j.1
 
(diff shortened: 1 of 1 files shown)


src/Init/Data/Array/Basic.lean
```diff
@@ -696,21 +705,26 @@ def erase [BEq α] (as : Array α) (a : α) : Array α :=
...
+    if i.1 < j then
+      let j' := ⟨j-1, Nat.lt_of_le_of_lt (Nat.pred_le _) j.2⟩
+      let as := as.swap j' j
+      loop as ⟨j', by rw [size_swap]; exact j'.2⟩
+    else
+      as
+  let j := as.size
+  let as := as.push a
+  loop as ⟨j, size_push .. ▸ j.lt_succ_self⟩
+termination_by _ _ _ j => j.1
```
leodemoura, line 720: You don't need to "fill" with `_`s. The given names are processed from right to left. It is not a bug but an intentional feature. Otherwise, we would have to calculate the free variables in the nested function.
That is, you can write just 
```lean
termination_by _ j => j.1
```
or 
```lean
termination_by loop j => j.1
```

digama0: Do you have a style preference on using `termination_by loop` vs `termination_by _`? I almost always use `termination_by _` since it's rare for more than one definition in a mutual block or def + where to require a nontrivial termination proof, but arguably it's less clear what it's applying to.

leodemoura: > Do you have a style preference on using termination_by loop vs termination_by _?

I (try to) use `termination_by loop` whenever more than one function is defined in the command.
However, I cannot claim I used this style consistently in the code base :)

PR #1610, opened 2022-09-16
feat: goal-diffs

### LINKED ISSUE #1469: Highlight changes in goal state

EdAyers:
goal-diff-final

`InteractiveGoals` now has information that a client infoview can use to show what parts of the goal have changed after applying a tactic.

There are lots of exciting ways to extend this, but I think this PR should be just limited to a first version and getting a sensible and extensible server/client API in place.

## API changes

All API changes are conservative so old versions of vscode will still be able to render the goals.

- `InteractiveGoals`, `InteractiveGoal`, `InteractiveHypothesisBundle` now have a `message? : Option String` field that can be used to emit debugging messages.
- `InteractiveGoal` and `InteractiveHypothesisBundle` now have a `isInserted? : Option Bool` which is true when it is a new goal or hypothesis and should be rendered with a green highlight.
- `SubexprInfo` has a new `tags?: Option String` field. Currently the infoview is highlighting green anything that has the tag `inserted` on it. I would like some comments on whether this is a good way to implement the diff api. The tag approach could cause issues later because it depends on both client and server using the same magic string. Another problem is that it might be causing bloat of `SubexprInfo`, perhaps instead the diff info should live on its own map from position to tags.

## Todos

- [x] tidy up
- [x] ~~#1607~~ no longer depends on this
- [ ] vscode extension PR https://github.com/leanprover/vscode-lean4/pull/251
- [x] unit tests
- [x] explain the api here
- [x] write a docstring explaining diffing algorithm used

## Todos/ future work

- [x] We use `goalsBefore` and `goalsAfter` on the `TacticInfo` object to get the diff. Currently I only show the diff on `useAfter = true`, perhaps I can also write a version for `useAfter = false` so that you see the change that is _about_ to happen.
- [ ] If a hypothesis is deleted, render it with a red triangle at the point that the hypothesis was deleted.

Resolves #1469

### DIFF
--- a/src/Lean/Widget/InteractiveCode.lean
+++ b/src/Lean/Widget/InteractiveCode.lean
@@ -21,29 +21,7 @@ inductive HighlightColor where
   | red
   | yellow
   | purple
-
-def HighlightColor.toString : HighlightColor → String
-  | .green  => "green"
-  | .blue   => "blue"
-  | .red    => "red"
-  | .yellow => "yellow"
-  | .purple => "purple"
-
-def HighlightColor.fromString : String → Except String HighlightColor
-  | "green"  => Except.ok HighlightColor.green
-  | "blue"   => Except.ok HighlightColor.blue
-  | "red"    => Except.ok HighlightColor.red
-  | "yellow" => Except.ok HighlightColor.yellow
-  | "purple" => Except.ok HighlightColor.purple
-  | s        => Except.error s!"expected an HighlightColor ctor string but got {s}"
-
-instance : ToString HighlightColor := ⟨HighlightColor.toString⟩
-
-instance : FromJson HighlightColor where
-  fromJson? j := j.getStr? >>= HighlightColor.fromString
-
-instance : ToJson HighlightColor where
-  toJson x := x.toString |> Json.str
+  deriving ToJson, FromJson
 
 /-- Information about a subexpression within delaborated code. -/
 structure SubexprInfo where


src/Lean/Util/FindExpr.lean
```diff
@@ -8,6 +8,22 @@ import Lean.Expr
 namespace Lean
 namespace Expr
 
+/-- Run the given `visit` function on each _immediate_ child subexpression of the given expression.
```
Kha, line 11: What does "immediate" mean here?

src/Lean/SubExpr.lean
```diff
@@ -150,17 +167,76 @@ def mkRoot (e : Expr) : SubExpr := ⟨e, Pos.root⟩
...
+-/
+def findChildrenWithPos [Alternative m] (visit : Pos → Expr → m α) (p : Pos) (e : Expr) : m α :=
+  match e with
+  | .app ..     => findAppWithPos visit p e
+  | .forallE .. => findForallWithPos visit p e
+  | .lam ..     => findLambdaWithPos visit p e
+  | .letE ..    => findLetWithPos visit p e
+  | .mdata _ b  => findChildrenWithPos visit p b
+  | .proj _ _ b => visit p.pushProj b
+  | _           => failure
```
Kha, line 237: Are any of these functions actually used?

src/Lean/Widget/Diff.lean
```diff
@@ -0,0 +1,224 @@
...
+ -/
+partial def exprDiff (before after : Expr) (p : Pos := Pos.root) : MetaM ExprDiff := do
+  if before == after then
+    return ∅
+  if !(← sameHead before after) then
+    return ExprDiff.singleton p ExprDiffPoint.inserted
+  let afterArgs := after.getAppArgs
+  let beforeArgs := before.getAppArgs
+  if afterArgs.size == 0 then
+    return ExprDiff.singleton p ExprDiffPoint.inserted
```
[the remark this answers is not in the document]
EdAyers: yes still working on that case

src/Lean/Widget/Diff.lean
```diff
@@ -0,0 +1,224 @@
...
+      if !e.hasMVar then return FindStep.done else
+      match e with
+      | .mvar m =>
+        if m == m₂ then
+          return FindStep.found
+        let m' ← getDelayedMVarRoot m
+        if m' == m₂ then
+          return FindStep.found
+        if m' != m ∧ (← m'.isAssigned) ∧ fuel > 0 then
+          if ← core m' (fuel - 1) then -- [todo] is this wf without fuel?
```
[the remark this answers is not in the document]
EdAyers: I don't really understand how the delayed mvars work in lean4. I was worried that there would be a way of getting an infinite descent with `getDelayedMVarRoot m` so I added a fuel param to the recursion.

The question is:

Let `m₁ ⊏ m₂` when `m₁ ≠ m₂` and `m₂ = getDelayedMVarRoot m₁` or `m₂` occurs in `m₁.assignment`. If I take the transitive closure of `⊏`, is it still irreflexive? It feels like it ought to be but I wasn't sure.

src/Lean/Widget/Diff.lean
```diff
@@ -0,0 +1,224 @@
...
+      | _ => return FindStep.visit
+    )
+    return r.isSome
+
+/-- A marker for a point in the expression where a subexpression has been inserted.
+
+NOTE: in the future we may add other tags such as `deleted` or `end_insert`.
+  -/
+inductive ExprDiffPoint where
+  | inserted
```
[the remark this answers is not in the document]
EdAyers: Yes I'm not sure on what names to use for these yet. I'm interpreting `inserted` as having the semantics of "paint it green".

src/Lean/Widget/Diff.lean
```diff
@@ -0,0 +1,224 @@
...
+  let beforeArgs := before.getAppArgs
+  if afterArgs.size == 0 then
+    return ExprDiff.singleton p ExprDiffPoint.inserted
+  let n := afterArgs.size
+  let bas := Array.zip beforeArgs afterArgs
+  let numDifferent := bas.filter (fun (a,b) => a != b) |>.size
+  if numDifferent > 1 ∧ p != Pos.root then
+    /- heuristic: if we are not at the root expression and more than one
+     argument is different, we say that the change has occurred at the head term.
+     eg if comm `x + y ↝ y + x`, we want to highlight all of `y + x`.
```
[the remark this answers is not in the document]
EdAyers: Yeah I just thought it looked a little better but very subjective

src/Lean/Util/FindExpr.lean
EdAyers: added some clarification

src/Lean/SubExpr.lean
EdAyers: removed, an earlier more complicated version of diff used them

src/Lean/Widget/Diff.lean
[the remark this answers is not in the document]
EdAyers: I'll rename to `change`

src/Lean/Widget/Diff.lean
```diff
@@ -0,0 +1,224 @@
...
+        try
+          return ← withTypeDiff t₀ h₁
+        catch e =>
+          let f ← e.toMessageData.format
+          return {h₁ with message? := s!"{f}"}
+      else
+        return {h₁ with isInserted? := true}
+  -- all fvids are present on original so we can assume no change.
+  return h₁
+  where withTypeDiff (t₀ : Expr) (h₁ : InteractiveHypothesisBundle) : MetaM InteractiveHypothesisBundle := do
```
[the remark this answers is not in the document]
Kha: yes

src/Lean/Widget/Diff.lean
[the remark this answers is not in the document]
Kha: I don't know whether searching through mvar assignments is the right thing to do here. Note that tactics could produce an arbitrarily long chain between an input and output mvar, but we definitely don't want to do an arbitrarily deep search through the proof term.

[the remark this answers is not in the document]
EdAyers: Ok I've removed the recursion and it seems to work ok on my set of examples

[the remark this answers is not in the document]
EdAyers: Ah no it doesn't work on `rw [h₁] at h₂`, you need to recurse to get the parent relationship. I am a bit worried about using a heuristic where you match up goals in zip order because it seems to me it's going to be easy to produce misleading output, particularly if the tactic is being applied to many goals with a combinator.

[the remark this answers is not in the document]
EdAyers: I'll add a cache so at least it doesn't check the same mvar over and over.

src/Lean/Widget/Diff.lean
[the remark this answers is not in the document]
EdAyers: I've removed this for now.

src/Lean/Widget/Diff.lean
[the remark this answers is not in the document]
EdAyers: This should now diff lambdas and pis correctly.

src/Lean/Widget/InteractiveGoal.lean
```diff
@@ -33,6 +45,12 @@ structure InteractiveGoal where
   name of the MVar that it is a goal for.)
   This is none when we are showing a term goal. -/
   mvarId? : Option MVarId := none
+  /-- If true, the goal was not present on the previous tactic state. -/
+  isInserted?: Option Bool := none
+  /-- If true, the goal will be removed on the next tactic state. -/
+  isRemoved? : Option Bool := none
+  /-- An optional message string to show the user after the goal. This is useful for debugging.-/
```
Vtec234, line 52: If the `message?` fields are only used for development, do we need to keep them around? What's the advantage of sending back a `message` rather than throwing an exception (which would also show up in the infoview)?

EdAyers: They are just for development. It's harder to debug if I used exceptions because I wouldn't be able to look at the tactic state.

EdAyers: I will remove at the end

src/Init/Data/List/Basic.lean
```diff
@@ -456,11 +456,22 @@ def isPrefixOf [BEq α] : List α → List α → Bool
   | _,     []    => false
   | a::as, b::bs => a == b && isPrefixOf as bs
 
+/-- `getSuffix? l₁ l₂` returns `some t` when `l₂ == l₁ ++ t`. -/
+def getSuffix? [BEq α] : List α → List α → Option (List α)
```
Vtec234, line 460: `isSuffix?` might be a little more consistent with the rest of the library (e.g. `Syntax.isNatLit?`).

src/Init/Data/Option/Basic.lean
```diff
@@ -76,6 +76,14 @@ instance (r : α → α → Prop) [s : DecidableRel r] : DecidableRel (Option.lt
   | some _, none   => isFalse not_false
   | none,   none   => isFalse not_false
 
+/-- Take a pair of options and if they are both `some`, apply the given fn to produce an output.
+Otherwise pass the value through. -/
```
Vtec234, line 80:
```suggestion
Otherwise act like `orElse`. -/
```

src/Lean/Widget/InteractiveGoal.lean
```diff
@@ -13,15 +13,27 @@ import Lean.Data.Lsp.Extra
...
+/-- In the infoview, if multiple hypotheses `h₁`, `h₂` have the same type `α`, they are rendered as `h₁ h₂ : α`.
+We call this a 'hypothesis bundle'. -/
 structure InteractiveHypothesisBundle where
   /-- The user-friendly name for each hypothesis.
   If anonymous then the name is inaccessible and hidden. -/
   names : Array Name
+  /-- The ids for each variable. Should have the same length as `names`. -/
   fvarIds : Array FVarId
   type : CodeWithInfos
+  /-- The value, in the case that the hypothesis is assigned. -/
```
Vtec234, line 25:
```suggestion
  /-- The value, in the case that the hypothesis is a `let`-binder. -/
```

src/Lean/Widget/InteractiveGoal.lean
```diff
@@ -13,15 +13,27 @@ import Lean.Data.Lsp.Extra
...
   val? : Option CodeWithInfos := none
+  /-- The hypothesis is a typeclass instance. -/
   isInstance : Bool
+  /-- The hypothesis is a type. -/
   isType : Bool
+  /-- If true, the hypothesis was not present on the previous tactic state. -/
+  isInserted? : Option Bool := none
```
Vtec234, line 32: This has three possible values. What is the semantics of `some false`? If the `Option` is just to avoid sending extra data when unnecessary, let's explain in the comment that `some false` should never be used.

src/Lean/Widget/InteractiveGoal.lean
```diff
@@ -66,6 +84,7 @@ def pretty (g : InteractiveGoal) : Format := Id.run do
 
 end InteractiveGoal
 
+/-- This is everything needed to render an interactive goal in the infoview. -/
```
Vtec234, line 87:
```suggestion
/-- This is everything needed to render an interactive term goal in the infoview. -/
```

src/Lean/Widget/InteractiveGoal.lean
```diff
@@ -98,16 +126,22 @@ def addInteractiveHypothesisBundle (hyps : Array InteractiveHypothesisBundle) (i
     isType     := (← instantiateMVars type).isSort
   }
 
+open Meta in
+variable [MonadControlT MetaM n] [Monad n] [MonadError n] [MonadOptions n] [MonadMCtx n] in
+def withGoalCtx (goal : MVarId) (action : LocalContext → MetavarDecl → n α) : n α := do
```
Vtec234, line 131: Should this be here or somewhere in `Meta`?

src/Lean/Widget/InteractiveCode.lean
```diff
@@ -24,24 +24,48 @@ structure SubexprInfo where
...
   -- kind : Lsp.SymbolKind
+  /-- Additional tags for rendering subexpresions. This is used for the goal-diff system. -/
+  tags?: Option (Array String) := none
   deriving Inhabited, RpcEncodable
 
 /-- Pretty-printed syntax (usually but not necessarily an `Expr`) with embedded `Info`s. -/
 abbrev CodeWithInfos := TaggedText SubexprInfo
 
+/-- Map the given `tt : CodeWithInfos`, if its pos appears in `pm`
+then it will modify the info at that position using `merger`. -/
```
Vtec234, line 35:
```suggestion
```
This function is short enough and a little difficult to explain (I don't understand the docstring at all). I would just remove the docstring.

src/Lean/Widget/Diff.lean
```diff
@@ -0,0 +1,326 @@
...
+
+/-- A description of the differences between a pair of expressions `before`, `after : Expr`.
+The information can be used to display a 'visual diff' for
+either `before`, showing the parts of the expression that are about to change,
+or `after` showing which parts of the expression have changed. -/
+structure ExprDiff where
+  /-- Map from subexpr positions in `e₀` to diff points.-/
+  changesBefore : PosMap ExprDiffTag := ∅
+  /-- A map from subexpr positions in `e₁` to 'diff points' which are tags
+  describing how the expression is changed at the given position.-/
```
Vtec234, line 75:
```suggestion
  describing how the expression has changed relative to `before` at the given position.-/
```
Do I understand correctly that this is what the map describes?

src/Init/Data/List/Basic.lean
EdAyers: on the other hand, this would mean that `isPrefix` and `isPrefix?` would be doing different things which might be confusing. I was hoping there would be a fn in mathlib3 I could steal the name from.

src/Lean/Widget/InteractiveGoal.lean
EdAyers: If I do `isInserted? : Bool`, will the json deriver encode that properly?

src/Lean/Widget/InteractiveGoal.lean
EdAyers: I wasn't sure, it's an extraction of functionality of the method below, and it appeared to me to be doing an InteractiveGoal-specific thing of `sanitizeNames` which I didn't really understand so I kept it here.

src/Lean/Widget/InteractiveGoal.lean
Vtec234: No, we'd be sending `"isInserted": false` then which takes up space, so the `Option` seems like the right choice; but it seems worth noting that this is the reason for using `Option`.

Vtec234:
> * `SubexprInfo` has a new `tags?: Option String` field. Currently the infoview is highlighting green anything that has the tag `inserted` on it. I would like some comments on whether this is a good way to implement the diff api. The tag approach could cause issues later because it depends on both client and server using the same magic string. Another problem is that it might be causing bloat of `SubexprInfo`, perhaps instead the diff info should live on its own map from position to tags.

From the perspective of the size of sent packets this is not an issue because we serialize `tags? := none` as nothing (0 bytes) in JSON. Earlier on I was thinking we may still prefer to use an external map `SubexprPos -> Tag` alongside the current `TaggedText<SubexprInfo>` rather than including `Tag`s in the `SubexprInfo`, basically because the current set of tags (used for goal diffs) is only relevant to the goal view; so when we use `CodeWithInfos` aka `TaggedText<SubexprInfo>` elswhere, such as in error messages, we get the "extra stuff" of tags which could be considered bloat.

*However* I am now thinking that tags could be useful more generally. For example, some error messages may want to bring the user's attention to a subexpression at which a tactic failed, or whatever, while still displaying the full top-level expression. For this though, it might be better to think of the infoview as a "dumb" display which doesn't know the semantics of "inserted/deleted/relevant" subexpression but simply gives it a particular colour or style. So i'd suggest renaming the tags to something style-y rather than semantic-y. E.g. "green"/"read". What do you think?

src/Lean/Widget/Diff.lean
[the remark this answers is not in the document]
EdAyers: Fixed: I should have been using `Lean.Meta.getMVars`.

src/Init/Data/List/Basic.lean
EdAyers: @digama0 what should it be called?

EdAyers:
Yes I'm always in favour of functional styling

src/Init/Data/List/Basic.lean
digama0: What do you mean by them doing different things? We do have a naming convention that `?` means Option-returning, and it is true that `(l1.isPrefix? l2).isSome == l1.isPrefix l2` so it seems okay to me.

...Oh, are prefix and suffix switched in this definition? I would recommend using `l1.isPrefix? l2` for the function that returns t such that `l2 ++ t == l1`, and similarly for `isSuffix?`.

src/Lean/Widget/InteractiveGoal.lean
EdAyers: removed

src/Lean/Widget/InteractiveCode.lean
```diff
@@ -15,6 +15,36 @@ with environment and subexpression information. -/
...
+  | "green"  => Except.ok HighlightColor.green
+  | "blue"   => Except.ok HighlightColor.blue
+  | "red"    => Except.ok HighlightColor.red
+  | "yellow" => Except.ok HighlightColor.yellow
+  | "purple" => Except.ok HighlightColor.purple
+  | s        => Except.error s!"expected an HighlightColor ctor string but got {s}"
+
+instance : ToString HighlightColor := ⟨HighlightColor.toString⟩
+
+instance : FromJson HighlightColor where
```
Vtec234, line 42: I think `deriving FromJson, ToJson` should have the same effect, although you'd still need the manual `HighlightColor.toString`.

Vtec234 (approved):
I didn't review `Diff.lean` in detail but everything else looks good!

src/Lean/Widget/InteractiveCode.lean
EdAyers: ah I didn't realise deriving worked on enums now :)

EdAyers: I've just deleted the toJson fromJson becuase it's not sent to the client anyway

src/Lean/Expr.lean
```diff
@@ -852,6 +852,33 @@ def getForallBody : Expr → Expr
...
+/-- A binder is the `(a : α)` that appears in `(a : α) → β` or `fun (a : α) => B`.
+It consists of a user-friendly name, a type expression and the BinderInfo. -/
+structure Binder where
+  name : Name
+  type : Expr
+  info : BinderInfo
+
+/-- Given a sequence of nested foralls `(a₁ : α₁) → ... → (aₙ : αₙ) → _`,
+returns `[(a₁ : α₁), ..., (aₙ : αₙ)]`. -/
+def getForallBinders : Expr → List Binder
```
Kha, line 869: This is a somewhat dangerous operation as it can expose loose bvars. I would suggest to merge it into `getForallBinderNames`, and perhaps to move all of these to the use site as they are quite usage specific.

Kha (commented):
I didn't think about the new heuristics too hard yet. I'd say let's simply give them a try.

src/Lean/Widget/InteractiveCode.lean
Vtec234: We still have the manual instances for `HighlightColor`

EdAyers: oops

Kha:
@EdAyers Feel free to PR a changelog entry for RELEASES.md once the extension change has landed!

gebner:
Sorry for only getting to this PR now.

> "green background" may have different meanings, it might be better to think of the infoview as a "dumb" display which doesn't know the semantics of "inserted/deleted/relevant" subexpression but simply gives it a particular colour or style. So i'd suggest renaming the tags to something style-y rather than semantic-y. E.g. "green"/"red". What do you think?

For the record, I would have preferred "inserted" and "removed" instead of "green" and "red," resp.  There are several reasons:
 1. It is consistent with the `isInserted?` and `isRemoved?` fields.
 2. We could maybe show an "inserted" tooltip to explain what the color means.  (I didn't think  about this a lot though, the tooltip probably doesn't mix well with the popup we already have.)
 3. The "blue" color clashes with the highlighting.

digama0:
Also, it's generally a bad idea to talk about colors directly because people have themes that can make the colors different, and then it's really confusing when `isGreen` means it's blue and `isRed` means it's gray in the theme. (Also colorblind people exist, and they usually switch the themes to something appropriate.)

Vtec234:
I agree, but it seems none of the points raised here apply to the style-predicates vs semantic-predicates choice, it's just that `green/red/blue` is maybe not the best choice of style-predicates. To be fair, we could think of `inserted/removed` as style-predicates which indicate "draw this in the same way `git diff` would".

EdAyers:
I am happy to open a PR reverting to semantic tags instead of colours. In response to the theming and colour-blindness, the actual colours that are displayed at the moment are a function of whether the theme is light or dark. It should be possible to get the colours to be completely theme-dependent by using [these](https://code.visualstudio.com/api/references/theme-color#integrated-terminal-colors).

Kha:
> In the case of using a semantic tags model, are you ok with having a `tags?: string[]` field or would you prefer something more type-safe like `diffTag?: 'inserted' | 'removed'`?

Let me highlight this question since people may have missed the edit. Personally I don't have any strong opinions on it.
I'm assuming we're waiting on this change before the extension PR is merged.

Issue #1606, opened 2022-09-15
Enforce uniform indentation in tactic blocks / do blocks / `sepByIndent`


gebner:
The use of `colGe` to implement indentation-sensitive syntax can lead to extremely surprising parses:
```
example : false := by
  conv =>
    lhs
    dsimp
    simp
```
Namely if `dsimp` is a conv-tactic then this parses as three conv-tactics (like the indentation suggests).  But if `dsimp` is not a conv-tactic, then this parses as `(conv => lhs); dsimp; simp`.

We should try out whether a `colEq` combinator (that enforces uniform indentation) causes any unexpected issues.

Another option (that fixes (?) the particular issue here) would be to implement a fallback `elab ident : conv => throwError "unknown conv tactic"` like we have for `tactic`.

leodemoura:
> We should try out whether a colEq combinator (that enforces uniform indentation) causes any unexpected issues.

I like this.

PR #1612, opened 2022-09-16
feat: use sepBy1Indent for tactic and conv blocks

gebner:
This is the first step towards using `colEq` in tactic/conv blocks, addressing #1606.

The main user-visible change is that `by simp rfl` (without newline or semicolon) is no longer supported.

### DIFF
--- a/src/Lean/Elab/Tactic/Conv/Basic.lean
+++ b/src/Lean/Elab/Tactic/Conv/Basic.lean
@@ -85,17 +85,23 @@ def changeLhs (lhs' : Expr) : TacticM Unit := do
    withMainContext do
      changeLhs (← zetaReduce (← getLhs))
 
+/-- Evaluate `sepByIndent conv "; " -/
+def evalSepByIndentConv (stx : Syntax) : TacticM Unit := do
+  for arg in stx.getArgs, i in [:stx.getArgs.size] do
+    if i % 2 == 0 then
+      evalTactic arg
+    else
+      saveTacticInfoForToken arg
+
 @[builtinTactic Lean.Parser.Tactic.Conv.convSeq1Indented] def evalConvSeq1Indented : Tactic := fun stx => do
-  for seqElem in stx[0].getSepArgs do
-    evalTactic seqElem
+  evalSepByIndentConv stx[0]
 
 @[builtinTactic Lean.Parser.Tactic.Conv.convSeqBracketed] def evalConvSeqBracketed : Tactic := fun stx => do
   let initInfo ← mkInitialTacticInfo stx[0]
   withRef stx[2] <| closeUsingOrAdmit do
     -- save state before/after entering focus on `{`
     withInfoContext (pure ()) initInfo
-    for seqElem in stx[1].getSepArgs do
-      evalTactic seqElem
+    evalSepByIndentConv stx[1]
     evalTactic (← `(tactic| all_goals (try rfl)))
 
 @[builtinTactic Lean.Parser.Tactic.Conv.nestedConv] def evalNestedConv : Tactic := fun stx => do


gebner:
I had to do a slightly ugly bootstrapping hack in 4657b744e00f645d0618ea539c02498acccd8ea7.

The issue is that parser aliases (such as aliases) are always taken from the previous stage (because they're defined using `builtin_initialize`).  With other syntax, you can use `parseQuotWithCurrentStage` to use the current stage.  But from what I can tell this does not affect parser aliases.

tests/lean/traceTacticSteps.lean.expected.out
```diff
@@ -5,30 +5,30 @@
...
   [Elab.step.result] ?m
 [Elab.step] skip
 [Elab.step] 
       skip
       trivial
+      
 [Elab.step] 
       skip
       trivial
+
```
digama0, line 18: It seems the formatter regressed here?

gebner: This is fun.  It turns out that `tacticSeq` now parses the trailing newline (after `trivial`) as a separator.  And the formatter then prints this final newline-separator.  We might want to change the formatter to drop the final separator if it's a newline.

gebner:
After this PR, it will be a very small change to enforce uniform start columns in tactic blocks (and `sepByIndent` generally): https://github.com/gebner/lean4/compare/sepbyindenttac...gebner:lean4:coleq?expand=1

src/Lean/Elab/Tactic/BuiltinTactic.lean
```diff
@@ -35,55 +35,52 @@ open Parser.Tactic
 @[builtinTactic paren] def evalParen : Tactic := fun stx =>
   evalTactic stx[1]
 
-def isCheckpointableTactic (arg : Syntax) : TacticM Bool := do
+def isCheckpointableTactic (arg : Syntax.Tactic) : TacticM Bool := do
```
leodemoura, line 38: Why did you have to change to `Syntax.Tactic`?

src/Lean/Elab/Tactic/BuiltinTactic.lean
```diff
@@ -35,55 +35,52 @@ open Parser.Tactic
...
 
-  go (i : Nat) (acc : Array Syntax) (result : Array Syntax) : TacticM (Array Syntax) := do
+  go (i : Nat) (acc : Array Syntax.Tactic) (result : Array Syntax.Tactic) : TacticM (Array Syntax.Tactic) := do
     if h : i < args.size then
       let arg := args.get ⟨i, h⟩
-      if (← isCheckpointableTactic arg[0]) then
+      if (← isCheckpointableTactic arg) then
         -- `argNew` is `arg` as singleton sequence. The idea is to make sure it does not satisfy `isCheckpointableTactic` anymore
-        let argNew := arg.setArg 0 (mkNullNode #[arg[0]])
+        let argNew := ⟨mkNullNode #[arg]⟩ -- HACK
```
leodemoura, line 62: Could you please explain the "HACK"?

src/Lean/Elab/Tactic/BuiltinTactic.lean
```diff
@@ -35,55 +35,52 @@ open Parser.Tactic
...
-  push (acc : Array Syntax) (result : Array Syntax) : Array Syntax :=
+  push (acc : Array Syntax.Tactic) (result : Array Syntax.Tactic) : Array Syntax.Tactic :=
     if acc.isEmpty then
       result
     else
-      let ref := acc.back
-      let accSeq := mkNode ``Lean.Parser.Tactic.tacticSeq1Indented #[mkNullNode acc]
-      let checkpoint := mkNode ``Lean.Parser.Tactic.checkpoint #[mkAtomFrom ref "checkpoint", accSeq]
-      result.push (mkNode groupKind #[checkpoint, mkNullNode])
+      result.push <| Unhygienic.run do withRef acc.back do
```
leodemoura, line 54: It is not clear to me why this change is needed.

src/Lean/Elab/Tactic/BuiltinTactic.lean
```diff
@@ -35,55 +35,52 @@ open Parser.Tactic
...
         let acc := acc.push argNew
         go (i+1) #[] (push acc result)
       else
         go (i+1) (acc.push arg) result
     else
       return result ++ acc
 
-/-- Evaluate `many (group (tactic >> optional ";")) -/
-def evalManyTacticOptSemi (stx : Syntax) : TacticM Unit := do
-  for seqElem in (← addCheckpoints stx.getArgs) do
```
leodemoura, line 74: You are converting here to `Array Syntax.Tactic`. It is not clear to me why.

src/Lean/Elab/Tactic/BuiltinTactic.lean
```diff
@@ -35,55 +35,52 @@ open Parser.Tactic
...
-    evalTactic seqElem[0]
-    saveTacticInfoForToken seqElem[1] -- add TacticInfo node for `;`
```
leodemoura, line 76: It is not clear to me why we don't need this anymore.

leodemoura: "this" here is the line `saveTacticInfoForToken seqElem[1]`

gebner: Yes, I understand.  I thought this was subsumed by Sebastian's other tactic position tweaks.  And the tests passed (including the ones checking tactic states at semicolons), so this was plausible.

But it turns out this was indeed necessary, but only for the space after the semicolon...  I'm adding a test.
```
  tac1; tac2
     --^
```

src/Lean/Elab/Tactic/BuiltinTactic.lean
gebner: It's a hack because it creates invalid syntax (see pretty-printer issues in the Zulip thread):

```
import Lean open Lean Elab Command PrettyPrinter

elab "run" : command => do
  let tac ← `(tactic| simp)
  let tac : TSyntax `tactic := ⟨mkNullNode #[tac]⟩ -- HACK: produces invalid tactic
  let term ← `(by $tac:tactic)
  logInfo (← liftCoreM do ppTerm term)

run -- unknown constant 'null'
```

src/Lean/Elab/Tactic/BuiltinTactic.lean
leodemoura: Adding more tests is great. I am frequently worried about changes that affect the infoview because of the lack of tests.

src/Lean/Elab/Tactic/BuiltinTactic.lean
gebner: Reverted.

tests/lean/traceTacticSteps.lean.expected.out
gebner: Final newline tweak implemented.

src/Lean/Elab/Tactic/BuiltinTactic.lean
gebner: (Note: I have rewritten this function one more time since your comment.)

The general reason is that the AST of tactic sequences has changed from `many1 (group (tactic >> optional ";"))` to `sepBy1Indent tactic`, i.e. from `[(group (skip) [])]` to `[(skip) []]`.  Having explicit types available (`Array Syntax.Tactic` instead of `Syntax`) makes the refactoring easier because the type system tells you what is a list of tactics, what is a tactic, etc., and you don't have to guess yourself what the `Syntax` means.

At first it seemed like it was not necessary to keep the separators, making `Array Syntax.Tactic` a natural choice.

I have now changed it back to untyped `Syntax` because we need to preserve the separators.  Ideally we'd have a `TSepIndentArray` for this because the manual index calculations are not particularly pretty.  (However that doesn't solve all problems either.  Even the `TSepArray` API is pretty bare-bones---and it doesn't preserve the separators either.)

gebner: > `TSepIndentArray`

Oh I completely forgot we had this already.  It's called ``TSepArray `tactic ""``.

gebner:
Re: saving tactic info at separators.

This prevents the regression with `tac1; tac2` noted above, but it is extremely fragile because the separators can be easily lost:
```lean
syntax "mytac" tacticSeq : tactic
macro_rules
  | `(tactic| mytac $[$tacs]*) => `(tactic| { $tacs* })

example : True ∨ True := by
  mytac constructor; constructor
-- still has the issue with the space after the semicolon
```

We probably need to fix this somewhere else.

leodemoura:
Thanks for the changes. I will go over the code tomorrow, and try to merge it. If I am still confused, we can discuss during the dev meeting.

PR #1620, opened 2022-09-18
feat: add `(canonical := true)` option in Syntax

digama0:
This adds a `canonical` option in synthetic syntax to allow it to be treated "as if" it was original for the purpose of hovers and otherwise transmitting error messages and information about the synthetic syntax to the original span it came from, without actually being `.original` since this messes up things that actually want pre-macro expansion tokens. In most cases, checks for `originalOnly` are changed to `canonicalOnly`, meaning that canonical synthetic syntax gets treated like original syntax. The main user-facing change is that `%$tk` will produce a canonical syntax now, on the reasoning that if you are going to associate a particular token with a given one then you probably are doing so in order to transmit information attached to that token. Regular tokens in a syntax quotation get their span from the `MonadRef` and are not canonical.

I reviewed all existing uses of functions which produce synthetic tokens to determine which ones should be "synthetic" and which should be "original" for the purpose of conveying hover info from expanded macros back to the original source that produced them.

### DIFF
--- a/src/Lean/Linter/UnusedVariables.lean
+++ b/src/Lean/Linter/UnusedVariables.lean
@@ -102,12 +102,6 @@ builtin_initialize addBuiltinUnusedVariablesIgnoreFn (fun _ stack opts =>
     (stx.isOfKind ``Lean.Parser.Term.matchAlt && pos == 1) ||
     (stx.isOfKind ``Lean.Parser.Tactic.inductionAltLHS && pos == 2))
 
--- is anonymous have variable
-builtin_initialize addBuiltinUnusedVariablesIgnoreFn (fun stx stack _ =>
-    stx.getId == `this &&
-    (stack.matches [none, ``Lean.Parser.Term.haveIdDecl] ||
-    stack.matches [none, ``Lean.Parser.Term.haveEqnsDecl]))
-
 builtin_initialize unusedVariablesIgnoreFnsExt : SimplePersistentEnvExtension Name Unit ←
   registerSimplePersistentEnvExtension {
     name          := `unusedVariablesIgnoreFns
--- a/src/Lean/Server/References.lean
+++ b/src/Lean/Server/References.lean
@@ -136,7 +136,8 @@ def findReferences (text : FileMap) (trees : Array InfoTree) : Array Reference :
     tree.visitM' (postNode := fun ci info _ => do
       if let some (ident, isBinder) := identOf info then
         if let some range := info.range? then
-          modify (·.push { ident, range := range.toLspRange text, stx := info.stx, ci, info, isBinder }))
+          if info.stx.getHeadInfo matches .original .. then  -- we are not interested in canonical syntax here
+            modify (·.push { ident, range := range.toLspRange text, stx := info.stx, ci, info, isBinder }))
   get
 
 /--


src/Lean/Elab/Quotation.lean
```diff
@@ -194,7 +194,6 @@ def addNamedQuotInfo (stx : Syntax) (k : SyntaxNodeKind) : TermElabM SyntaxNodeK
     if s.length > 3 then
       if let (some l, some r) := (stx[0].getPos? true, stx[0].getTailPos? true) then
         -- HACK: The atom is the string "`(foo|", so chop off the edges.
-        -- HACK: We have to use .original here or the hover won't show up
```
digama0, line 197: The code hasn't changed here, but it's no longer considered a hack but a supported use case now.

Kha:
Mmh, the linter failure suggests that we may need some combination of `original` and `synthetic` after all

Kha:
Another issue is that synthetic input may be turned into original output, no?

digama0:
Note that it was a linter failure on `unusedVariables` in a mathlib tactic implementation that lead to this yak shave in the first place. I don't really challenge that both `original` and `synthetic` need to exist, but rather I think we need to draw the line more accurately. Previous to this PR, all %$tk were `.synthetic`, now they are all `.original`; if we need to add per-macro control we can add something like `%%$tk`.

digama0:
I noticed the thing about synthetic being turned into original, and I wasn't exactly sure how to solve it, because `getTailPos?` takes `originalOnly` only as an argument, and it does different things depending on whether it is searching for original nodes or not so it's not like just returning whether the found source info was original or not will preserve behavior.

leodemoura:
We discussed this issue in the last dev meeting and decided that we need more tests and documentation for `canonical`.

src/Lean/Linter/UnusedVariables.lean
```diff
@@ -102,6 +102,11 @@ builtin_initialize addBuiltinUnusedVariablesIgnoreFn (fun _ stack opts =>
     (stx.isOfKind ``Lean.Parser.Term.matchAlt && pos == 1) ||
     (stx.isOfKind ``Lean.Parser.Tactic.inductionAltLHS && pos == 2))
 
+-- is anonymous have variable
+builtin_initialize addBuiltinUnusedVariablesIgnoreFn (fun stx stack _ =>
+    stx.getId == `this &&
+    (stack.matches [none, ``Lean.Parser.Term.haveIdDecl] ||
+    stack.matches [none, ``Lean.Parser.Term.haveEqnsDecl]))
```
Kha, line 109: I still think that the unused variables linter should only look at original identifiers. More specifically, it uses `findReferences`, which was definitely not intended to return information not from the source text. If we ensure that, this workaround should not be necessary anymore, no?

src/Lean/Elab/BuiltinNotation.lean
```diff
@@ -74,19 +74,18 @@ open Meta
   | _                        => Macro.throwUnsupported
 
 @[builtinMacro Lean.Parser.Term.have] def expandHave : Macro := fun stx =>
-  let thisId := mkIdentFrom stx `this
   match stx with
   | `(have $x $bs* $[: $type]? := $val; $body)            => `(let_fun $x $bs* $[: $type]? := $val; $body)
-  | `(have $[: $type]? := $val; $body)                    => `(have $thisId $[: $type]? := $val; $body)
+  | `(have%$tk $[: $type]? := $val; $body)                => `(have $(mkIdentFrom tk `this (canonical := true)) $[: $type]? := $val; $body)
```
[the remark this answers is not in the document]
digama0: That would produce a hygienic `this`, we want a bare one. It might be possible to wrap it in a `Unhygienic.run` instead

src/Init/Prelude.lean
```diff
@@ -3371,8 +3371,26 @@ inductive SourceInfo where
...
+  for the purpose of hovers and error messages. This is usually used on identifiers,
+  to connect the binding site to the user's original syntax even if the name of the identifier
+  changes during expansion, as well as on tokens where we will attach targeted messages.
+
+  The syntax `token%$stx` in a syntax quotation will annotate the token `token` with the span
+  from `stx` and also mark it as canonical.
+
+  As a rough guide, a macro expansion should only use a given piece of input syntax in
+  a single canonical token, although this is sometimes violated when the same identifier
+  is used to declare two binders, as in the macro expansion for dependent if:
```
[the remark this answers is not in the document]
digama0: True, these `h`'s are actually original syntax. We'd have to for example change the identifier name to need synthetic syntax. I don't really have another familiar example though.

src/Lean/Linter/UnusedVariables.lean
digama0: That sounds like a separate PR. I think the implementation of `unusedVariables` is weird in a bunch of ways - I remember noticing while straightening out the import graph that this file has an unexpected dependency on the LSP files, because it's using a function that was designed for a different purpose inside the server.

src/Lean/Elab/BuiltinNotation.lean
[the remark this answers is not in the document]
Kha: No, it produces the `this` keyword. But that's not accepted in the second case, ahh...

src/Lean/Linter/UnusedVariables.lean
Kha: I've restored the behavior of `findReferences`, please take a look. Btw, I was not able to reproduce the `have` issue.

Kha: `findReferences` should probably be moved, yeah, but it is the right function to use here; recall e.g. https://github.com/leanprover/lean4/pull/1396

src/Lean/Elab/BuiltinNotation.lean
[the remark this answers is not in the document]
digama0: I'm going to make a PR to make `this` an ident again some day soon... This causes so many issues in tactics that weren't thinking about `this` and just use the `ident` parser.

digama0:
LGTM

Kha (approved):

Kha:
@digama0 Could you update the PR title & description for posterity?

PR #1622, opened 2022-09-19
feat: use colEq in sepByIndent

### LINKED ISSUE #1606: Enforce uniform indentation in tactic blocks / do blocks / `sepByIndent`

gebner:
Fixes #1606.

There is a single syntax error when compiling Lean with this change, in `Lean/Elab/Tactic/Basic.lean`.  The indentation there is really crazy.  I did the minimum change to make it build again.

The main downside is that the error messages aren't great.  If you have a tactic with the wrong indentation, then it won't be parsed as part of the tactic block.  Usually this means it becomes part of the next command, so you end up with an `expected command` error if you have the wrong indentation.
Note that other indendation-related error messages are just as bad; if you don't indent enough you also get the `expected command` error.

### DIFF
--- a/src/Lean/Elab/Tactic/Basic.lean
+++ b/src/Lean/Elab/Tactic/Basic.lean
@@ -150,7 +150,7 @@ partial def evalTactic (stx : Syntax) : TacticM Unit :=
     | .missing => pure ()
     | _ => throwError m!"unexpected tactic{indentD stx}"
 where
-   throwExs (failures : Array EvalTacticFailure) : TacticM Unit := do
+    throwExs (failures : Array EvalTacticFailure) : TacticM Unit := do
      if let some fail := failures[0]? then
        -- Recall that `failures[0]` is the highest priority evalFn/macro
        fail.state.restore (restoreInfo := true)


leodemoura:
> There is a single syntax error when compiling Lean with this change, in Lean/Elab/Tactic/Basic.lean. The indentation there is really crazy. I did the minimum change to make it build again.

I only noticed one extra space starting at `@[inline] handleEx`. Is there something else?

digama0:
I think `throwExs` is indented one space too many, and `expandEval` and `eval` two spaces too many.

gebner:
No as I wrote, that's the only change.

> I think throwExs is indented one space too many, and handleEx, expandEval and eval two spaces too many.

Before, `throwExs` was indendented by three spaces, and the other functions by four spaces.  I've changed `handleEx` to four spaces as well.  But yeah, usually we do two spaces I think.

leodemoura:
> No as I wrote, that's the only change.
> 
> > I think throwExs is indented one space too many, and handleEx, expandEval and eval two spaces too many.
> 
> Before, `throwExs` was indendented by three spaces, and the other functions by four spaces. I've changed `handleEx` to four spaces as well. But yeah, usually we do two spaces I think.

When I read "The indentation there is really crazy.", I assumed there was a major issue there, and you were putting a temporary workaround there.

Issue #1449, opened 2022-08-08
Term goals only sometimes show type class instances


gebner:
```lean
example [Inhabited Empty] : α := (default : Empty).rec
                               --^ term goal (instance missing)

example [Inhabited α] : α := default
                           --^ term goal (instance shown)

example [Inhabited α] : β := _
                           --^ term goal (instance shown)
```

I would expect type-class instances to be shown always.  This is all quite confusing since they show up just fine if you write `by exact`..

### Versions

Since 484e5102216badf51e3738242c45c982895f8d6c

See also #1345

leodemoura:
We have been trying to minimize the local context in error messages and the expected type info view. It may not be perfect, but we don't want to give it up because it is very helpful in many other cases. Note that, in tactic mode, we don't perform the aggressive local context filtering. This is why `by exact ...` produces a different behavior.

Are you proposing we never filter out local instances? This is a simple change.

gebner:
> We have been trying to minimize the local context in error messages and the expected type info view.

I know, that's why I'm trying to bring up particularly surprising or confusing cases that I run into.  Also so that we have some papertrail about what we want to be hidden or not.

> It may not be perfect, but we don't want to give it up because it is very helpful in many other cases.

I certainly agree that there is room for improvement.  Do we have a list of things that we absolutely *should* hide?  I tried to find out concrete motivating examples for hiding inaccessible assumptions but couldn't find any discussion on Zulip or github, so this is a recollection of issues that came up recently.
 - Implementation details of macros.  Like the `discr` in syntax matches or `do_jp` in do-notation.  (Note that these should probably also be hidden from `assumption` etc.)
 - Let-values in programming contexts.  (See also Mario's `let dep` proposal in #1345)
 - Auxiliary declarations.  (Note that these definitely need to be hidden from tactics as well.)

Did I miss anything?

> Are you proposing we never filter out local instances? This is a simple change.

That would be an easy fix for this particular issue, yes.

digama0:
> > Let-values in programming contexts. (See also Mario's let dep proposal in #1345)
> 
> I like the `let dep` proposal, but we should probably get support from users first. BTW, we can even use the `nonDep` flag to implement this feature :) We would also be able to remove `let_fun`. BTW, should we call it `let dep` or `let_dep`?

I was also thinking about `let!`, although I think this kind of "try harder" naming convention is more common in mathlib than in core so maybe users with a different naming sense would prefer `let dep`.

[...]

PR #1624, opened 2022-09-19
chore: few updates to Expr docstrings

ydewit:
@leodemoura, I noticed the following comment in the letE constructor:
```
  Let-expressions. The field `nonDep` is not currently used, but will be used in the future
  by the code generator (and possibly `simp`) to track whether a let-expression is non-dependent
  or not.
```
Now that we have LCNF, is this outdated and can go away?

### DIFF
--- a/src/Lean/Expr.lean
+++ b/src/Lean/Expr.lean
@@ -364,9 +364,11 @@ inductive Expr where
 
   /--
   A function application.
+
   For example, the natural number one, i.e. `Nat.succ Nat.zero` is represented as
   `Expr.app (.const `Nat.succ []) (.const .zero [])`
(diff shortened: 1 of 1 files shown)


leodemoura:
> Now that we have LCNF, is this outdated and can go away?

Yes, you are correct. This is should go away.

ydewit:
I will remove this part of the docstring.

Should I refactor Expr and remove `(nonDep : Bool)` or not a good time?

leodemoura:
> Should I refactor Expr and remove (nonDep : Bool) or not a good time?

Not yet. We have a pending issue where we are considering a `let dep` that will be enforced by the kernel, and used by `simp`.
https://github.com/leanprover/lean4/issues/1449

PR #1626, opened 2022-09-20
feat: add tactic.simp.trace option

digama0:
This adds the equivalent of `simp?` functionality. When you enable the `tactic.simp.trace` option, every `simp`, `simp_all` and `dsimp` call will print an equivalent `simp only` call that shows all the statements that were used in the proof.

The `simp?` tactic abbreviation has not been implemented, and I consider the display itself using `Try this: simp only [foo, bar]` to be provisional until the LSP support for code replacement actions is up to scratch.

The `update-stage0` is needed to make the tests pass, but it can be done independently of this PR.

### DIFF
--- a/tests/lean/simp_trace.lean
+++ b/tests/lean/simp_trace.lean
@@ -1,15 +1,16 @@
 set_option tactic.simp.trace true
+set_option trace.Meta.Tactic.simp.rewrite true
 
 def f (x : α) := x
 
-theorem ex1 (a : α) (b : List α) : f (a::b = []) = False :=
+example (a : α) (b : List α) : f (a::b = []) = False :=
   by simp [f]
 
 def length : List α → Nat
   | []    => 0
   | a::as => length as + 1
 
-theorem ex2 (a b c : α) (as : List α) : length (a :: b :: as) > length as := by
+example (a b c : α) (as : List α) : length (a :: b :: as) > length as := by
   simp [length]
   apply Nat.lt.step
   apply Nat.lt_succ_self
@@ -31,12 +32,12 @@ def head [Inhabited α] : List α → α
   | []   => default
   | a::_ => a
 
-theorem ex4 [Inhabited α] (a : α) (as : List α) : head (a::as) = a :=
+example [Inhabited α] (a : α) (as : List α) : head (a::as) = a :=
   by simp [head]
 
 def foo := 10
 
-theorem ex5 (x : Nat) : foo + x = 10 + x := by
+example (x : Nat) : foo + x = 10 + x := by
   simp [foo]
   done
 
@@ -44,7 +45,7 @@ def g (x : Nat) : Nat := Id.run <| do
   let x := x
   return x
 
-theorem ex6 : g x = x := by
+example : g x = x := by
   simp [g, bind, pure]
   rfl
 
@@ -55,7 +56,7 @@ def f2 : StateM Nat Unit := do
   let s ← get
   set <| g s
 
-theorem ex7 : f1 = f2 := by
+example : f1 = f2 := by
   simp [f1, f2, bind, StateT.bind, get, getThe, MonadStateOf.get, StateT.get, pure, set, StateT.set, modify, modifyGet, MonadStateOf.modifyGet, StateT.modifyGet]
 
 def h (x : Nat) : Sum (Nat × Nat) Nat := Sum.inl (x, x)
@@ -65,5 +66,18 @@ def bla (x : Nat) :=
   | Sum.inl (y, z) => y + z
   | Sum.inr _ => 0
 
-theorem ex8 (x : Nat) : bla x = x + x := by
+example (x : Nat) : bla x = x + x := by
   simp [bla, h]
+
+example (x : Nat) (h : 1 ≤ x) : x - 1 + 1 + 2 = x + 2 := by
+  simp [h, Nat.sub_add_cancel]
+
+example (x : Nat) : (if h : 1 ≤ x then x - 1 + 1 else 0) = (if _h : 1 ≤ x then x else 0) := by
+  simp (config := {contextual := true}) [h, Nat.sub_add_cancel]
+
+theorem my_thm : a ∧ a ↔ a := ⟨fun h => h.1, fun h => ⟨h, h⟩⟩
+
+example : a ∧ (b ∧ b) ↔ a ∧ b := by simp [my_thm]
+example : (a ∧ (b ∧ b)) = (a ∧ b) := by simp only [my_thm]
+
+example : x - 1 + 1 = x := by simp (discharger := sorry) [Nat.sub_add_cancel]
--- a/tests/lean/simp_trace.lean.expected.out
+++ b/tests/lean/simp_trace.lean.expected.out
@@ -1,8 +1,61 @@
 Try this: simp only [f]
+[Meta.Tactic.simp.rewrite] unfold f, f (a :: b = []) ==> a :: b = []
+[Meta.Tactic.simp.rewrite] @eq_self:1000, False = False ==> True
 Try this: simp only [length]
+[Meta.Tactic.simp.rewrite] unfold length, length (a :: b :: as) ==> length (b :: as) + 1
+[Meta.Tactic.simp.rewrite] unfold length, length (b :: as) ==> length as + 1
 Try this: simp only [fact]
+[Meta.Tactic.simp.rewrite] unfold fact, fact (Nat.succ x) ==> (x + 1) * fact x
 Try this: simp only [head]
+[Meta.Tactic.simp.rewrite] unfold head, head (a :: as) ==> match a :: as with
+    | [] => default
+    | a :: tail => a
+[Meta.Tactic.simp.rewrite] @eq_self:1000, a = a ==> True
 Try this: simp only [foo]
+[Meta.Tactic.simp.rewrite] unfold foo, foo ==> 10
+[Meta.Tactic.simp.rewrite] @eq_self:1000, 10 + x = 10 + x ==> True
 Try this: simp only [g, Id.pure_eq]
+[Meta.Tactic.simp.rewrite] unfold g, g x ==> Id.run
+      (let x := x;
+      pure x)
+[Meta.Tactic.simp.rewrite] @Id.pure_eq:1000, pure x ==> x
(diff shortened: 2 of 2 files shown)


gebner:
I didn't look at the implementation.

1. Does this work with dischargers?
2. Does it at least work with conditional simp lemmas? → add a test please

I am not particularly happy about the `tactic.simp.trace` option.  Can't we just wait with that until code actions are in?

digama0:
> Does this work with dischargers?

Yes, for most fancy options and dischargers it just regurgitates the original config arg in the output. I thought about tracking whether options and/or dischargers are "used", but there is no guarantee that the discharger has similar tracing capability so we can only track it at a relatively coarse level and it didn't seem very helpful.

> Does it at least work with conditional simp lemmas? → add a test please

yes → ok

> I am not particularly happy about the tactic.simp.trace option. Can't we just wait with that until code actions are in?

The work on `simp?` implementation is largely independent of code actions, and I'd rather not serialize the work. I can remove the words `Try this:` or use another display method if you have one I can use today. But it's useful to have something that users can play with, because without it no one will notice bugs in the implementation and I can't have test cases for it.

gebner:
> But it's useful to have something that users can play with

Fair enough.  I just want to avoid that this interface becomes official in any way.

tests/lean/1079.lean.expected.out
```diff
@@ -1,5 +1,5 @@
 1079.lean:3:2-6:12: error: alternative 'isFalse' has not been provided
-[Meta.Tactic.simp.rewrite] h:1000, m ==> n
+[Meta.Tactic.simp.rewrite] _uniq.82:1000, m ==> n
```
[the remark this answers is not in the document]
digama0: What's happening is that the `ToFormat` instance for `SimpLemma` doesn't have the context to look fvars up in the local context. Previously the name of the simp lemma was the user name of the hypothesis, but this can be ambiguous. It seemed relatively unimportant to me so I let it be.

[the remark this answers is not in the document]
gebner: This is a trace message; it would be better to do `{mkFVar name}` or `{mkConst name}` anyhow (depending on what it is) because then you can hover over it, get go-to-definition, etc.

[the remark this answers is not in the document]
digama0: Is this trace stuff run even when the trace is disabled? The intention here is for the simp lemmas to carry the minimum necessary for reconstruction so as not to slow `simp` down, but if it does all this computation in trace messages it could potentially be a performance issue.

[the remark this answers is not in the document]
gebner: The code looks like this:
```lean
trace[Meta.Tactic.simp.rewrite] "{thm}, {e} ==> {rhs}"
```
The `trace[cls] ...` macro expands to `if ← isTracingEnabledForClass cls then ...` so the message is only evaluated if traces are enabled.

src/Lean/Elab/Tactic/Simp.lean
```diff
@@ -236,12 +244,52 @@ def mkSimpContext (stx : Syntax) (eraseLocal : Bool) (kind := SimpKind.simp) (ig
       unless simpTheorems.isErased localDecl.userName do
         let fvarId := localDecl.fvarId
         let proof  := localDecl.toExpr
-        let id     ← mkFreshUserName `h
-        fvarIdToLemmaId := fvarIdToLemmaId.insert fvarId id
-        simpTheorems ← simpTheorems.addTheorem proof (name? := id)
+        fvarIdToLemmaId := fvarIdToLemmaId.insert fvarId fvarId.name
```
leodemoura, line 247: Why is this change necessary? You are using the internal free variable identifier as the theorem name. This name may appear in the trace. I prefer to have a `h???` than the unique free variable identifier.

src/Lean/Meta/Tactic/Simp/Types.lean
```diff
@@ -22,11 +22,12 @@ abbrev Cache := ExprMap Result
 abbrev CongrCache := ExprMap (Option CongrTheorem)
 
 structure Context where
-  config         : Config      := {}
-  simpTheorems   : SimpTheoremsArray  := {}
+  config         : Config := {}
+  simpTheorems   : SimpTheoremsArray := {}
   congrTheorems  : SimpCongrTheorems := {}
+  namedStx       : NameMap Syntax := {}
```
leodemoura, line 28: We need a docstring at `namedStx`.

src/Lean/Meta/Tactic/Simp/SimpTheorems.lean
```diff
@@ -38,16 +38,11 @@ structure SimpTheorem where
     `name?` is mainly relevant for producing trace messages.
```
leodemoura, line 38: Docstring still contains `name?`. Note that it says that we use the name for producing trace messages, and it seems you are using free variable unique identifiers as names.

src/Lean/Meta/Tactic/Simp/Types.lean
```diff
@@ -97,6 +99,19 @@ def getSimpCongrTheorems : M SimpCongrTheorems :=
...
     modify fun s => { s with cache := cacheSaved }
 
+def recordSimpTheorem (n : Name) : SimpM Unit :=
+  modify fun s => { s with usedTheorems := s.usedTheorems.insert n }
+
+def ppSimpTheorem (s : SimpTheorem) : SimpM MessageData :=
+  s.toMessageDataM fun n => do
+    if (← getEnv).contains n then
+      return ← mkConstWithLevelParams n
+    if (← getLCtx).contains ⟨n⟩ then
```
leodemoura, line 109: Good. I found the explanation for the change: free variable ids as theorem identifiers. We need to document this. Could you please document the field `name` and explain the new approach?

src/Lean/Meta/Tactic/Simp/Types.lean
```diff
@@ -97,6 +99,19 @@ def getSimpCongrTheorems : M SimpCongrTheorems :=
...
+      return mkFVar ⟨n⟩
+    if let some stx := (← read).namedStx.find? n then
+      return stx
+    return n
```
leodemoura, line 113: This `n` here is the weird part. The user may see the free variable identifiers. I suspect it can happen if we are using `contextual := true`, and we have something like

```
if x = 0 then 
 ... f x ... -- `x` is going to be rewritten here, and the theorem used will not be in the local context.
else
  ....
```

tests/lean/simp_trace.lean
```diff
@@ -1,3 +1,5 @@
+set_option tactic.simp.trace true
+
```
leodemoura, line 2: It would be great to have tests that use `contextual := true`.

src/Lean/Elab/Tactic/Simp.lean
```diff
@@ -236,12 +244,52 @@ def mkSimpContext (stx : Syntax) (eraseLocal : Bool) (kind := SimpKind.simp) (ig
...
+  descr    := "When tracing is enabled, calls to `simp` or `dsimp` will print an equivalent `simp only` call."
+}
+
+def traceSimpCall (stx : Syntax) (ctx : Simp.Context) (usedSimps : NameSet) : MetaM Unit := do
+  let mut stx := stx
+  if stx[3].isNone then
+    stx := stx.setArg 3 (mkNullNode #[mkAtom "only"])
+  let mut args := #[]
+  let mut localsOrStar := some #[]
+  let lc ← getLCtx
```
leodemoura, line 263: We usually use `lctx` instead of `lc`

src/Lean/Elab/Tactic/Simp.lean
```diff
@@ -236,12 +244,52 @@ def mkSimpContext (stx : Syntax) (eraseLocal : Bool) (kind := SimpKind.simp) (ig
...
+          some (locals.push ldecl.userName)
+        else
+          none
+    -- global definitions in the environment
+    else if env.contains thm then
+      if thm != ``eq_self then -- this one is implicitly available
+        args := args.push (← `(Parser.Tactic.simpLemma| $(mkIdent (← unresolveNameGlobal thm)):ident))
+    -- Ignore everything else.
+    -- Note: this is possible for `simp (config := {contextual := true})` when
+    -- rewriting with a variable that was introduced in a scope.
```
leodemoura, line 283: Good. Happy you commented this.

src/Lean/Elab/Tactic/Simp.lean
leodemoura: I found the explanation for this change, but we need to document it.

src/Lean/Meta/Tactic/Simp/Types.lean
leodemoura: I found your comment at `traceSimpCall`. So, you are aware of this. I am now assuming this code will never be invoked to trace a theorem if `n` is the free variable id of a theorem that is not in scope anymore.

leodemoura (changes requested):
Summary:

- Document the field `name` and how the use of free variable ids works here. BTW, it feels like `name` is a bad name here. This name can be a declaration name, a free variable id, something else?. What about creating a new type that makes this different possibilities clear to us. I think it will help maintaining the code in the future.

- I didn't find support for reverting auxiliary theorem names. Please add a test that uses a simp theorem of the form `a <-> b`. @gebner does something like that in Mathlib 4.

- Please add tests with `contextual := true`

digama0:
> I didn't find support for reverting auxiliary theorem names. Please add a test that uses a simp theorem of the form a <-> b. @gebner does something like that in Mathlib 4.

Without testing, my understanding of the solution implemented here is that the `name` of a simp theorem encodes the "origin" of the theorem, the reason why it is in the simp set. Multiple simp theorems can have the same `name` if they were decomposed from a common declaration, and the `a = b` simp theorem derived from `theorem bla : a <-> b` should have name `bla`.

I'm happy to implement a more structured type for `name` instead of overloading fvars, syntax, and declaration names like this. We could also put options and the discharger in there, like gabriel mentioned, although displaying that information will be interesting.

leodemoura:
> I'm happy to implement a more structured type for name instead of overloading fvars, syntax, and declaration names like this. We could also put options and the discharger in there, like gabriel mentioned, although displaying that information will be interesting.

Great.

leodemoura (approved):

leodemoura:
Happy with this PR, and ready to merge as soon as stage0 conflicts are resolved.

PR #1630, opened 2022-09-20
feat: multiple `delta`

digama0:
This is used by mathport. `delta foo, bar` is like `delta foo` but it simultaneously unfolds any of the declarations in the list in the chosen targets.

### DIFF
--- a/tests/lean/run/unfoldMany.lean
+++ b/tests/lean/run/unfoldMany.lean
@@ -3,5 +3,5 @@ def f (x : Nat) := x + 1
 def g (x : Nat) := f x + f x
 
 example : g x > 0 := by
-  unfold g, f
+  unfold g f
   simp_arith


Kha:
Do we have a top-level comma-separate list anywhere else in the tactic syntax? Why not space-separated?

Kha (changes requested):
You should also update the docstrings

digama0:
I added docstrings and used space-separated lists for `unfold` / `delta` tactic and conv.

Kha:
> It's also used in `unfold`. We could go back to space separated though.

Oh wow, that *is* unexpected. Don't we need `colGt` now though?
@leodemoura Do you agree that we should usually separate identifiers by spaces in tactics?

digama0:
oops yes, it needs `colGt`

leodemoura:
> @leodemoura Do you agree that we should usually separate identifiers by spaces in tactics?

Yes, we have been using this convention in other tactics.
Happy to talk about whether we should change this convention in the next dev meeting.

PR #1632, opened 2022-09-21
feat: `pattern (occs := ...)` conv

digama0:
Used in mathport. The conv tactic `pattern (occs := 1 2 3) pat` is like `pattern pat`, but instead of just producing a subgoal for the first occurrence of `pat` it produces 3 subgoals for the first three occurrences of `pat`. `(occs := *)` produces subgoals for all occurrences of `pat`.

### DIFF
--- a/src/Init/Conv.lean
+++ b/src/Init/Conv.lean
@@ -23,6 +23,18 @@ syntax convSeqBracketed := "{" sepByIndentSemicolon(conv) "}"
 -- automatically closing goals
 syntax convSeq := convSeqBracketed <|> convSeq1Indented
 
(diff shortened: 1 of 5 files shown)


gebner:
Are we even using this in mathlib?  (I couldn't find any occurrence with grep, but maybe I didn't look too hard.)

leodemoura:
> Are we even using this in mathlib? (I couldn't find any occurrence with grep, but maybe I didn't look too hard.)

If it is not used, let's skip it.

digama0:
I was going to use it to implement the `for` conv and `nth_rewrite`.

leodemoura (approved):

leodemoura:
Could you please fix the conflict?

PR #1634, opened 2022-09-21
feat: `if _ : cond then t else e` syntax

digama0:
This allows you to use `_` as the binder inside `dite` notation. This differs from `if cond then ...`  in that it desugars to `dite` instead of `ite`, and so in the body of the if statement you can make use of the hypothesis without naming it. This is useful when the `unusedVariables` linter wants to make the variable anonymous but it is important for the expression to be a `dite`.

### DIFF
--- a/src/Init/Notation.lean
+++ b/src/Init/Notation.lean
@@ -343,14 +343,13 @@ end Lean
   ppRealGroup(ppRealFill(ppIndent("if " Lean.binderIdent " : " term " then") ppSpace term)
     ppDedent(ppSpace) ppRealFill("else " term)) : term
 
-@[macro termDepIfThenElse] private def macroRules.termDepIfThenElse : Lean.Macro := fun
+macro_rules
   | `(if $h:ident : $c then $t else $e) => do
     let mvar ← Lean.withRef c `(?m)
     `(let_mvar% ?m := $c; wait_if_type_mvar% ?m; dite $mvar (fun $h:ident => $t) (fun $h:ident => $e))
   | `(if _%$h : $c then $t else $e) => do
     let mvar ← Lean.withRef c `(?m)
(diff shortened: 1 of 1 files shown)


leodemoura:
I have hit the problem this PR is solving in the past (e.g., `h : c` is needed to discharge the condition at `a[i]` automatically). @Kha and I briefly discussed the possibility of elaborating all `if-then-else` using `dite`, and having `simp` theorems to convert the non-dependent ones back into `ite`.

digama0:
As long as `ite` and `dite` both exist, I think the most predictable desugaring just uses `if _ : cond` as `dite` and `if cond` as `ite`.
One way to make your version "just work" would be to have `if cond then t else e` elaborate `t` and `e` with the appropriate condition variables in scope, and then if they ended up being used then construct a `dite` term, else `ite`. I worry about cases where you have metavariables in there though and want to reason about the expression anyway.

For proving purposes I think we do want a way to create `ite` terms directly, for similar reasons as why we don't always want `match h : cond with` dependent match style. But the main reason I see to support `if _ : cond` independent of whether `if cond` is "clever" is syntactic uniformity. It's a binder, so it should allow `_`.

leodemoura:
> One way to make your version "just work" would be to have if cond then t else e elaborate t and e with the appropriate condition variables in scope, and then if they ended up being used then construct a dite term, else ite. I worry about cases where you have metavariables in there though and want to reason about the expression anyway.

We considered this option. It is not reliable because metavariables. If `t` or `e` contain a metavariable, we don't really know whether we need the variable or not.

leodemoura:
> For proving purposes I think we do want a way to create ite terms directly, for similar reasons as why we don't always want match h : cond with dependent match style.

Yes, I agree.

leodemoura:
Could you please rebase? We have conflicts at stage0.

leodemoura (approved):

PR #1636, opened 2022-09-22
feat: more conv goal structuring tactics

digama0:
This adds a few tactics that are useful in both tactic mode and conv mode, by macro expansion to the tactic versions.

Also fixes a bug in `tactic' => $seq` which causes even `tactic' => skip` to turn a `| a` goal into `|- a` because the underlying equality is hidden behind an assigned metavariable.


leodemoura:
Could you please add new tests for all new tactics?

PR #1638, opened 2022-09-22
feat: show decl module in hover

digama0:
This adds a *in Foo.Bar.Module* remark at the end of the hover display. This is helpful for knowing what to import to get access to a certain declaration you can see being used somewhere.

hover_module

I was originally hoping to put a link there along the lines of rust-analyzer's "Go to [MyType](#)", but it turns out this is implemented via an LSP extension for "hover commands" and some client support, so I will leave it to the folks who are more involved in the frontend if they want to implement that.

### DIFF
--- a/src/Lean/Server/InfoUtils.lean
+++ b/src/Lean/Server/InfoUtils.lean
@@ -186,33 +186,39 @@ def Info.docString? (i : Info) : MetaM (Option String) := do
 def Info.fmtHover? (ci : ContextInfo) (i : Info) : IO (Option Format) := do
   ci.runMetaM i.lctx do
     let mut fmts := #[]
-    try
-      if let some f ← fmtTerm? then
+    let modFmt ← try
+      let (termFmt, modFmt) ← fmtTermAndModule?
+      if let some f := termFmt then
         fmts := fmts.push f
-    catch _ => pure ()
+      pure modFmt
+    catch _ => pure none
     if let some m ← i.docString? then
       fmts := fmts.push m
+    if let some f := modFmt then
+      fmts := fmts.push f
     if fmts.isEmpty then
       return none
     else
       return f!"\n***\n".joinSep fmts.toList
 
 where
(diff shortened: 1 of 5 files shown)


Kha:
The namespace/module ambiguity strikes back :sweat_smile: . How about "from module `Lean.Elab.Term`", or even just `import Lean.Elab.Term`?

src/Lean/Server/InfoUtils.lean
```diff
@@ -186,33 +186,36 @@ def Info.docString? (i : Info) : MetaM (Option String) := do
...
+      if let some f ← fmtModule? decl then
+        fmts := fmts.push f
     if fmts.isEmpty then
       return none
     else
       return f!"\n***\n".joinSep fmts.toList
 
 where
-  fmtTerm? : MetaM (Option Format) := do
+  fmtTerm? : MetaM (Option Format × Option Name) := do
```
Kha, line 206: This looks a bit obscure to me, I would suggest repeating the `e.isConst` branch in `fmtModule?` instead

digama0: Well it was the `whnf` that I wanted to avoid calling a second time.

digama0:
I suppose that example wasn't the best to use regarding namespace vs module. The namespace is already visible in the first line though, so I think it should be clear after a few examples?

src/Lean/Server/InfoUtils.lean
Kha: What `whnf`?

digama0: ...I must have hallucinated that. I mean `instantiateMVars`

leodemoura (approved):

src/Lean/Server/InfoUtils.lean
Kha: Ah. `docString?` seems to show that it's not really necessary for constants, but I suppose it can't hurt.

PR #1639, opened 2022-09-22
fix: underscore intellisense

lovettchris:
This fixes https://github.com/leanprover/vscode-lean4/issues/14 
This is a short term hack fix.

What is happening is that `_` by itself is dropped by the parser, which results in Completion.find? finding idCompletion on the term "main" and you can see the same result when you type Ctrl+Space at the end of main:

image

This explains why `_` after main yields `map_eq_pure_bind` because this symbol contains the string `main` :-) 
You can see the same thing happening below when you type `a_` it finds symbols starting with `a` and containing `_` somewhere.

image

Notice that it highlights the matching chars, including the underscores.  
Now the interesting thing is that without this fix the following does not work properly:

image

Instead, you will see a lawful function listed, instead of my `_foo` function.  But as you can see this now works correctly because VS code is doing it's own thing, probably using the `textDocument/semanticTokens/full` info returned by Lean.

And if I remove my "_foo" function then "main _" shows no intellisense, which is the fix we wanted here.  

Note: there is a related bug: https://github.com/leanprover/lean4/issues/1640

### DIFF
--- a/src/Lean/Server/Completion.lean
+++ b/src/Lean/Server/Completion.lean
@@ -28,7 +28,6 @@ def addToBlackList (env : Environment) (declName : Name) : Environment :=
 
 private def isBlackListed (declName : Name) : MetaM Bool := do
   let env ← getEnv
-  -- todo would be nice if there was a better way to detect built in symbols.
   (pure (declName.isInternal && !isPrivateName declName))
   <||> (pure <| isAuxRecursor env declName)
   <||> (pure <| isNoConfusion env declName)
@@ -249,7 +248,7 @@ private def idCompletionCore (ctx : ContextInfo) (id : Name) (hoverInfo : HoverI
     unless (← isBlackListed declName) do
       let matchUsingNamespace (ns : Name): M Bool := do
         if let some (label, score) ← matchDecl? ns id danglingDot declName then
-          -- dbg_trace "matched with {id}, {declName}, {label}, {isLocal}"
+          -- dbg_trace "matched with {id}, {declName}, {label}"
           addCompletionItem label c.type expectedType? declName (← getCompletionKindForDecl c) score
           return true
         else
@@ -366,8 +365,7 @@ private def dotCompletion (ctx : ContextInfo) (info : TermInfo) (hoverInfo : Hov
       else
         failure
     else
-      let env ← getEnv
-      env.constants.forM fun declName c => do
+      (← getEnv).constants.forM fun declName c => do
         let typeName := (← normPrivateName declName).getPrefix
         if nameSet.contains typeName then
           unless (← isBlackListed c.name) do


Kha:
> What is happening is that `_` by itself is dropped by the parser

This is certainly not the case, so I think we need a more thorough analysis of what's going wrong in the `main _` case

lovettchris:
> This is certainly not the case, so I think we need a more thorough analysis of what's going wrong in the `main _` case

What I meant to say is that it is not returned in the infoTree, when I print the info tree found at the end of the line `#eval main _` I get this:

```
Info: ?m.120 : List String @ ⟨10, 11⟩-⟨10, 12⟩ @ Lean.Elab.Term.elabHole
Info: main : List String → IO UInt32 @ ⟨10, 6⟩-⟨10, 10⟩
Info: [.] `main : none @ ⟨10, 6⟩-⟨10, 10⟩
Info: main ?m.120 : IO UInt32 @ ⟨10, 6⟩-⟨10, 12⟩ @ Lean.Elab.Term.elabApp
Info: command @ ⟨10, 0⟩-⟨10, 12⟩ @ Lean.Elab.Command.elabEval
```

I see the info `main ?m.120` which does go all the way to position 12, which includes the underscore, but the Completion.find? method somehow misses the presence of `?m.120` so it only returns these 3 Info's:

```
main: List String → IO UInt32
map_eq_pure_bind: ∀ [inst : Monad m] [inst_1 : LawfulMonad m] (f : α → β) (x : m α),
  f <$> x = do
    let a ← x
    pure (f a)
MonadFinally: (Type u → Type v) → Type (max (u + 1) v)
```

In debugging I see that the `choose` method chooses the Info at `Info: [.] `main : none @ ⟨10, 6⟩-⟨10, 10⟩` which does not include the position 12 underscore metavariable.  I think it is because that term is type `Lean.Elab.Term.elabHole` and not what it is looking for which is `info.isCompletion`.

Kha:
It's not clear to me why `choose` would consider infos that are not even adjacent to the cursor position at all. Unfortunately I don't know the completion code very well.

lovettchris:
Well if I was just typing spaces after `main` then perhaps intellisense on `main` is still interesting, although I tend to agree with you I'd rather that move into method parameter intellisense (even though I know that is hard to do).  

See my alternate solution which might help explain this a bit better: https://github.com/leanprover/lean4/pull/1644

Kha:
I guess considering completion that is not adjacent makes sense for the `set_option |` case, but in all other completion cases (which unlike `optionCompletion` do not even look at `stx`!) this looks wrong. I'm quite sure that is the root issue.

Kha:
Case in point: completion at `|` in `#check f |` should not be limited to completions starting with `f`. The underscore was a red herring.

lovettchris:
Please continue the discussion here: https://github.com/leanprover/lean4/pull/1644
and I will close this PR since we all seem to agree this hack is no good.

lovettchris:
Closing in favor of https://github.com/leanprover/lean4/pull/1644

lovettchris:
Reopening because the proper fix https://github.com/leanprover/lean4/pull/1644 ran into tricky InfoTree problems.

leodemoura:
I am closing this PR because it is not hackish.

lovettchris:
Actually did you see Sebastian say this in the #1644 discussion?
> I'd rather prefer a simple temp workaround like https://github.com/leanprover/lean4/pull/1639.

leodemoura:
> Actually did you see Sebastian say this in the #1644 discussion?
> 
> > I'd rather prefer a simple temp workaround like #1639.

Yes, I saw, but I don't like piling up hacks. This module is already too messy.

PR #1643, opened 2022-09-23
RB tree refinements

Kha:
For reasons I wanted to make sure I actually understand the red-black tree code in the Immutable Beans paper, which led me to cleaning up the rbmap benchmarks a little. Then I wanted to make sure I understand the relation between Okasaki's original implementation and ours, which led me to simplify the actual `Std.RBNode` - I'll leave the benchmark as is as I don't want to change every single version in every language.

I double-checked that the implementation is functionally unchanged, while the IR code should have become a little simpler.

### DIFF
--- a/src/Lean/Data/RBMap.lean
+++ b/src/Lean/Data/RBMap.lean
@@ -77,18 +77,16 @@ def singleton (k : α) (v : β k) : RBNode α β :=
   node red leaf k v leaf
 
 -- the first half of Okasaki's `balance`, concerning red-red sequences in the left child
--- precondition: the first tree is red
 @[inline] def balance1 : RBNode α β → (a : α) → β a → RBNode α β → RBNode α β
-  | node _ (node red a kx vx b) ky vy c, kz, vz, d
-  | node _ a kx vx (node red b ky vy c), kz, vz, d => node red (node black a kx vx b) ky vy (node black c kz vz d)
-  | a,                                   kx, vx, b => node black a kx vx b
+  | node red (node red a kx vx b) ky vy c, kz, vz, d
+  | node red a kx vx (node red b ky vy c), kz, vz, d => node red (node black a kx vx b) ky vy (node black c kz vz d)
+  | a,                                     kx, vx, b => node black a kx vx b
 
 -- the second half, concerning red-red sequences in the right child
--- precondition: the second tree is red
 @[inline] def balance2 : RBNode α β → (a : α) → β a → RBNode α β → RBNode α β
-  | a, kx, vx, node _ (node red b ky vy c) kz vz d
-  | a, kx, vx, node _  b ky vy (node red c kz vz d) => node red (node black a kx vx b) ky vy (node black c kz vz d)
-  | a, kx, vx, b                                    => node black a kx vx b
+  | a, kx, vx, node red (node red b ky vy c) kz vz d
+  | a, kx, vx, node red b ky vy (node red c kz vz d) => node red (node black a kx vx b) ky vy (node black c kz vz d)
+  | a, kx, vx, b                                     => node black a kx vx b
 
 def isRed : RBNode α β → Bool
   | node red .. => true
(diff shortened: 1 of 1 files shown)


Kha:
!bench

Kha:
I don't want to necessarily burden Leo with this, so maybe someone else wants to re-check that the `RBNode` code is still doing the same thing?:)

tests/bench/rbmap4.lean
```diff
@@ -1,76 +0,0 @@
-/-
```
Kha, line 1: This file was functionally identical to `rbmap.lean` except for bit rot

Kha:
Fewer instructions and fewer lines of C output make me relatively confident that there is no regression in performance for `RBNode` either

leodemoura:
> I don't want to necessarily burden Leo with this, so maybe someone else wants to re-check that the `RBNode` code is still doing the same thing? :)

Thanks. I am currently overwhelmed and will not have time to check.
A random remark: if I remember correctly our current implementation is based on https://www.cs.princeton.edu/~appel/papers/redblack.pdf, and Appel's mentions that he did not manage to certify Okasaki's version. Don't remember the details, but it is worth taking a look.

Kha:
Fun historical fact: Okasaki uses pseudo code for or-patterns to avoid the repeated right-hand sides, which Haskell does not have. But we do!

digama0:
I have verified a sketch of the formal proof of equivalence of these two functions.

Kha:
!bench

Kha:
> I have verified that these two functions are equivalent (not quite a formal proof but enough to convince me that the formal proof would go through). I didn't check the changes to the tests. But the code looks simpler so +1 from me.

Thanks!

Kha:
Interestingly this one further simplification towards Okasaki does slow things down
```diff
modified   src/Lean/Data/RBMap.lean
@@ -77,16 +77,16 @@ def singleton (k : α) (v : β k) : RBNode α β :=
   node red leaf k v leaf
 
 -- the first half of Okasaki's `balance`, concerning red-red sequences in the left child
-@[inline] def balance1 : RBNode α β → (a : α) → β a → RBNode α β → RBNode α β
-  | node red (node red a kx vx b) ky vy c, kz, vz, d
-  | node red a kx vx (node red b ky vy c), kz, vz, d => node red (node black a kx vx b) ky vy (node black c kz vz d)
-  | a,                                     kx, vx, b => node black a kx vx b
+@[inline] def balance1 : Rbcolor → RBNode α β → (a : α) → β a → RBNode α β → RBNode α β
+  | black, node red (node red a kx vx b) ky vy c, kz, vz, d
+  | black, node red a kx vx (node red b ky vy c), kz, vz, d => node red (node black a kx vx b) ky vy (node black c kz vz d)
+  | c,     a,                                     kx, vx, b => node c a kx vx b
 
 -- the second half, concerning red-red sequences in the right child
-@[inline] def balance2 : RBNode α β → (a : α) → β a → RBNode α β → RBNode α β
-  | a, kx, vx, node red (node red b ky vy c) kz vz d
-  | a, kx, vx, node red b ky vy (node red c kz vz d) => node red (node black a kx vx b) ky vy (node black c kz vz d)
-  | a, kx, vx, b                                     => node black a kx vx b
+@[inline] def balance2 : Rbcolor → RBNode α β → (a : α) → β a → RBNode α β → RBNode α β
+  | black, a, kx, vx, node red (node red b ky vy c) kz vz d
+  | black, a, kx, vx, node red b ky vy (node red c kz vz d) => node red (node black a kx vx b) ky vy (node black c kz vz d)
+  | c,     a, kx, vx, b                                     => node c a kx vx b
 
 def isRed : RBNode α β → Bool
   | node red .. => true
@@ -102,16 +102,11 @@ variable (cmp : α → α → Ordering)
 
 @[specialize] def ins : RBNode α β → (k : α) → β k → RBNode α β
   | leaf,               kx, vx => node red leaf kx vx leaf
-  | node red a ky vy b, kx, vx =>
+  | node c a ky vy b, kx, vx =>
     match cmp kx ky with
-    | Ordering.lt => node red (ins a kx vx) ky vy b
-    | Ordering.gt => node red a ky vy (ins b kx vx)
-    | Ordering.eq => node red a kx vx b
-  | node black a ky vy b, kx, vx =>
-    match cmp kx ky with
-    | Ordering.lt => balance1 (ins a kx vx) ky vy b
-    | Ordering.gt => balance2 a ky vy (ins b kx vx)
-    | Ordering.eq => node black a kx vx b
+    | Ordering.lt => balance1 c (ins a kx vx) ky vy b
+    | Ordering.gt => balance2 c a ky vy (ins b kx vx)
+    | Ordering.eq => node c a kx vx b
 
 def setBlack : RBNode α β → RBNode α β
   | node _ l k v r => node black l k v r
```

Kha:
Also note that the `rbmap_liibrary` benchmark I added today is still slower than `rbmap` from the standalone benchmark, which I can only assume is due to the `Bool` map value being unboxed in the latter.

PR #1644, opened 2022-09-23
fix underscore completion (alternative solution)

lovettchris:
This fixes https://github.com/leanprover/vscode-lean4/issues/14 

What is happening is that `_` by itself is not represented by an `ofCompletionInfo` Info in the InfoTree, so `Completion.find?` was ignoring it (even though it was a better match for the current position), and going on to find things so it found "main" instead (in the example below):

image

This explains why `_` after main yields `map_eq_pure_bind` because this symbol contains the string `main` :-) 
You can see the same thing happening below when you type `a_` it finds symbols starting with `a` and containing `_` somewhere.

image

Notice that it highlights the matching chars, including the underscores.  

With this fix we allow the `Completion.find? choose` function to return the `Lean.Elab.Term.elabHole` so that `Completion.find?` can use that information to return an empty completion list, which is what we want in this case.


lovettchris:
Well after messing with this for a while, I'm finding the code to be extremely complex, I cannot easily to find a way to recognize the elabHole case in Completion.find? that doesn't also break a bunch of other unit tests...

So now I'm back to being in favor of the "hack" solution in https://github.com/leanprover/lean4/pull/1639.

lovettchris:
The reason it fails is that we find these Info objects:

```
Completion.find? looking for completions at 4,29
Info: _example (isBinder := true) : Foo → sorryAx (Sort u_1) true @ ⟨4, 0⟩-⟨4, 7⟩ : isCompletion=false and inside=false
Info: sorryAx (sorryAx (Sort u_1) true)
  true : sorryAx (Sort u_1) true @ ⟨4, 0⟩†-⟨4, 29⟩† @ Lean.Elab.Term.StructInst.elabStructInst : isCompletion=false and inside=false
Info: x (isBinder := true) : Foo @ ⟨4, 9⟩-⟨4, 10⟩ : isCompletion=false and inside=false
Info: _example (isBinder := true) : Foo → sorryAx (Sort u_1) true @ ⟨4, 0⟩†-⟨4, 29⟩† : isCompletion=false and inside=false
Info: [.] id x : Foo @ ⟨4, 20⟩-⟨4, 24⟩ : some Sort.{?_uniq.195} : isCompletion=true and inside=false
Info: x : Foo @ ⟨4, 23⟩-⟨4, 24⟩ : isCompletion=false and inside=false
Info: [.] `x : some ?_uniq.197 @ ⟨4, 23⟩-⟨4, 24⟩ : isCompletion=true and inside=false
Info: x : Foo @ ⟨4, 23⟩-⟨4, 24⟩ @ Lean.Elab.Term.elabIdent : isCompletion=false and inside=false
Info: @id : {α : Type} → α → α @ ⟨4, 20⟩-⟨4, 22⟩ : isCompletion=false and inside=false
Info: [.] `id : none @ ⟨4, 20⟩-⟨4, 22⟩ : isCompletion=true and inside=false
Info: id x : Foo @ ⟨4, 20⟩-⟨4, 24⟩ @ Lean.Elab.Term.elabApp : isCompletion=false and inside=false
Info: sorryAx (Sort u_1) true : Sort u_1 @ ⟨4, 20⟩-⟨4, 29⟩ @ Lean.Elab.Term.elabPipeProj : isCompletion=false and inside=false
Info: x (isBinder := true) : Foo @ ⟨4, 9⟩-⟨4, 10⟩ : isCompletion=false and inside=false
Info: Foo : Type @ ⟨4, 13⟩-⟨4, 16⟩ : isCompletion=false and inside=false
Info: [.] `Foo : some Sort.{?_uniq.193} @ ⟨4, 13⟩-⟨4, 16⟩ : isCompletion=true and inside=false
Info: Foo : Type @ ⟨4, 13⟩-⟨4, 16⟩ @ Lean.Elab.Term.elabIdent : isCompletion=false and inside=false
Info: command @ ⟨4, 0⟩-⟨4, 29⟩ @ Lean.Elab.Command.elabDeclaration : isCompletion=false and inside=false
Chosen Info: sorryAx (Sort u_1) true : Sort u_1 @ ⟨4, 20⟩-⟨4, 29⟩ @ Lean.Elab.Term.elabPipeProj : isCompletion=false and inside=false
```

Notice that the `Lean.Elab.Term.elabPipeProj` is an exact match, ending at the desired column 29.  But the Info we need to be using in this case is this one:

```
Chosen Info: [.] id x : Foo @ ⟨4, 20⟩-⟨4, 24⟩ : some Sort.{?_uniq.195} : isCompletion=true and inside=false
```
Notice this info is for the expression `id f`.  So the intellisense this Info provides somhow knows that x is of type Foo that it's a completion on that so it shows the members of structure `Foo` yielding the desired result:

image

But this seems like a hack.  Why isn't there an info for `|>.f` that also has isCompletion set to true finishing at position 29 ?

It turns out the infoTree DOES have the correct info at span 4,20 to 4,28 if you remove the final `f` on this line, so intellisense works in this case:

```lean
example (x : Foo) : id x |>.
```

Why does typing an `f` remove that good one?

Kha:
> Why isn't there an info for `|>.f` that also has isCompletion set to true finishing at position 29 ?

I agree, that looks problematic as well.

lovettchris:
> I agree, that looks problematic as well.

Should I file that as a separate but, change the `completion.lean` test so it passes so we can complete this PR, or do you want to include a fix for this problem in this PR ?  If the latter I will need help with that...

Kha:
Yeah, I think really solving this needs a larger refactor that would go in a different direction from this PR. If we decide that the underscore issue is too annoying to wait for that, I'd rather prefer a simple temp workaround like https://github.com/leanprover/lean4/pull/1639.

lovettchris:
Well I can resurrect #1639, but I'm curious, can you say a little more about the larger refactor that would go in a different direction from this PR?  I'd like to ponder that new information before going back to #1639.

Kha:
I guess it's not all that large:
1. make sure we have completion nodes for `|>.f` and perhaps more missing cases
1. make sure `HoverInfo.after` only triggers completion where it makes sense (e.g. `set_option`)

What I was really thinking of is a refactoring I had mentioned to @leodemoura before:
* eliminate completion nodes wherever the regular syntax/info tree already provides all necessary information, to make completion more easily extensible inside and outside lean4

This would change how we implement 1. from above, but otherwise it is probably orthogonal

leodemoura:
I am closing this PR since it is stuck. It is not a high priority now, requires refactoring in the code base, etc.

PR #1645, opened 2022-09-24
chore: move Std.* data structures to Lean.*

digama0:
As previously discussed, even though the Std data structures are likely to be very similar to the Lean data structures we would like to keep them separate so that they don't have to evolve together. The files were already moved from the Std package to the Lean package, but they still had the `Std` namespace. This PR moves them all to the `Lean` namespace, with the exception of `Std.Range` and `Std.Format`, which we want to be "vocabulary types".

### DIFF
--- a/src/Lean/Data/PersistentArray.lean
+++ b/src/Lean/Data/PersistentArray.lean
@@ -5,7 +5,7 @@ Authors: Leonardo de Moura
 -/
 universe u v w
 
-namespace Std
+namespace Lean
 
 inductive PersistentArrayNode (α : Type u) where
   | node (cs : Array (PersistentArrayNode α)) : PersistentArrayNode α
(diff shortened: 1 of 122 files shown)


src/Lean/Data/PersistentArray.lean
```diff
@@ -370,19 +370,15 @@ def mkPersistentArray {α : Type u} (n : Nat) (v : α) : PArray α :=
...
   mkPersistentArray n v
 
-end Std
+end Lean
 
-open Std (PersistentArray PersistentArray.empty)
+open Lean (PersistentArray)
 
-def List.toPersistentArrayAux {α : Type u} : List α → PersistentArray α → PersistentArray α
+def List.toPArray' {α : Type u} (xs : List α) : PersistentArray α :=
```
digama0, line 377: The use of `'` here is a bit awkward. Since we are adding things outside the Lean namespace here we have to be careful. We would like the name `List.toPArray` to stay open for Std, but we also don't want to lose dot notation for code inside the Lean package, so the compromise is to use `'` on these functions.

leodemoura (approved):

leodemoura:
Happy to merge as soon as conflicts are resolved.

Issue #407, opened 2021-04-14
String interpolation by default


leodemoura:
The notation `" ..."` would now behave like `s!"..."`, and we would have a notation for raw strings.

leodemoura:
@Kha you were more active in the Zulip thread about this issue. Do you have further remarks, details, etc?
BTW, what would be the notation for raw strings? Would we still have escaped characters for raw strings?

Kha:
I don't think there was any consensus on raw strings, but personally my favorite syntax is Rust's where you can just add more `#`s if you *still* have a collision with the string contents: https://doc.rust-lang.org/reference/tokens.html#characters-and-strings. Not trivial to implement though.
I think raw strings should also turn off escapes. I would hope that there is not sufficient need for a syntax that allows escapes but no interpolation.

DanielFabian:
What makes the rust style raw strings more difficult to implement? It's not a regular grammar, but any grammar that can handle parentheses should be able to express it no? `s(#^n).*?(#^n)`

Kha:
My concern was over the `FirstTokens` set, but it should probably be a new token parser (activated by `r[#"]`), in which case it can do whatever it wants.

eddyb:
> What makes the rust style raw strings more difficult to implement? It's not a regular grammar, but any grammar that can handle parentheses should be able to express it no? `s(#^n).*?(#^n)`

Note that your `?` is a context-sensitive disambiguator: `.*` can include `#^k` for `k` both smaller and greater than `n` and it's simply impossible for a CFG to tell them apart. Even without the `?` there's arguably still an implicit disambiguator in that most regex engines will only give you the "the longest match".

At best you would have an ambiguity-preserving parse forest (à la SPPF from some Earley/GLR/GLL/etc. general CFG parser), and then disambiguate raw strings *on that*. <sub>(amusingly, the disambiguation can be a noop if `#^n` doesn't repeat *in the entire input*, or if all the other choices cause parse errors *after* their presumed end of the raw string - not so funny if the correct choice has its own parse errors though, since you have to disambiguate to know what to report)</sub>

But if you already have more flexible tokenization (as seems to be the case here), you don't have to go through all that complication, and you can implement the counting directly. <sub>(there could be benefits to parse forests if you didn't already have a non-trivial recursive descent parser, but generally it's a messy tradeoff IME)</sub>

For more background on Rust's own raw strings context-sensitivity:
* https://github.com/rust-lang/reference/pull/1185
[...]

PR #1647, opened 2022-09-26
feat: track callers in panicking functions

digama0:
This implements a feature similar to the `#[track_caller]` attribute in Rust or the `[HasCallStack]` typeclass in GHC Haskell.

* The panicking primitive `panicWithPosWithDecl` and `panicWithPos` are changed to `StackInfo.panic` which bundles together the module/decl/position information into a struct `StackInfo`.
* `panicWithInfo` is a wrapper over `StackInfo.panic` which accepts the `StackInfo` by auto param, using the `stack_info` tactic.
* The `stack_info` tactic is a macro for `first | assumption | exact stack_info_here`, a poor man's typeclass inference with a tactic fallback.
* The `stack_info_here` term macro expands to a literal containing the declaration, module and line information for the application in which it is called.
* `panic!` used to synthesize the decl/module/line info and call `panicWithPosWithDecl` and `panicWithPos`, now it is just a trivial wrapper over `panicWithInfo`, whose auto param does the equivalent thing.

The upshot is that when you call `panic!`, it now either synthesizes the position on the spot, or gets it from an argument of type `StackInfo` if it is in the context. Even more importantly, functions with a `(info : StackInfo := by stack_info)` auto param will *also* do the same thing, meaning that it's possible to bubble up panic blame.

Generally speaking, we want to put these `(info : StackInfo := by stack_info)` auto params on any function which panics due to a precondition violation rather than a bug in the code itself. Helpfully, these are all marked with a `!` in the code (like `Option.get!`), so they are easy to find.

I would have liked to use a typeclass like GHC does instead of an auto param. The inference behavior of typeclasses is much closer to what we want here - auto params have to go at the end of the declaration, which is awkward, and they are also a lot more verbose than a `[HasCallStack]` argument would be. But currently there is no way to trigger a tactic from a typeclass search, or even a typeclass search failure as in this case.

### DIFF
--- a/src/Lean/Compiler/LCNF/CompilerM.lean
+++ b/src/Lean/Compiler/LCNF/CompilerM.lean
@@ -196,7 +196,7 @@ private partial def normFVarImp (s : FVarSubst) (fvarId : FVarId) (translator : 
       fvarId'
     else
       normFVarImp s fvarId' translator
-  | some e => panic! s!"invalid LCNF substitution of free variable with expression {e}"
+  | some e => panic! "invalid LCNF substitution of free variable with expression {e}"
   | none => fvarId
 
 /--
--- a/src/Lean/Data/Json/Basic.lean
+++ b/src/Lean/Data/Json/Basic.lean
@@ -146,7 +146,7 @@ def toFloat : JsonNumber → Float
 should switch to using those. -/
 private def fromPositiveFloat! (x : Float) (info : CallerInfo := by caller_info) : JsonNumber :=
   match Lean.Syntax.decodeScientificLitVal? (toString x) with
-  | none => panic! s!"Failed to parse {toString x}"
+  | none => panic! "Failed to parse {x}"
   | some (m, sign, e) => OfScientific.ofScientific m sign e
 
 /-- Attempts to convert a float to a JsonNumber, if the number isn't finite returns
@@ -266,7 +266,7 @@ def getObjValD (j : Json) (k : String) : Json :=
 
 def setObjVal! : Json → String → Json → (info : CallerInfo := by caller_info) → Json
   | obj kvs, k, v, _ => obj <| kvs.insert compare k v
-  | _      , _, _, _ => panic! "Json.setObjVal!: not an object: {j}"
+  | _      , _, _, _ => panic! "Json.setObjVal!: not an object"
 
 inductive Structured where
   | arr (elems : Array Json)
--- a/src/Lean/SubExpr.lean
+++ b/src/Lean/SubExpr.lean
@@ -41,7 +41,7 @@ def tail (p : Pos) : Pos :=
   else (p.asNat - p.head) / maxChildren
 
 def push (p : Pos) (c : Nat) : Pos :=
-  if c >= maxChildren then panic! s!"invalid coordinate {c}"
+  if c >= maxChildren then panic! "invalid coordinate {c}"
   else p.asNat * maxChildren + c
 
 variable {α : Type} [Inhabited α]


src/Init/Util.lean
```diff
@@ -30,17 +30,22 @@ def dbgStackTrace {α : Type u} (f : Unit → α) : α := f ()
 @[extern "lean_dbg_sleep"]
 def dbgSleep {α : Type u} (ms : UInt32) (f : Unit → α) : α := f ()
 
-@[noinline] private def mkPanicMessage (modName : String) (line col : Nat) (msg : String) : String :=
-  "PANIC at " ++ modName ++ ":" ++ toString line ++ ":" ++ toString col ++ ": " ++ msg
+structure StackInfo where
+  modAndDecl : String
+  line : Nat
+  col : Nat
```
gebner, line 36:
```suggestion
structure CallerInfo where
  module : Name
  declName : Name
  position : Lean.Position
```

This isn't a stack trace (which the name stack info suggests, at least to me).

Having the module/declaration name available directly could make this more generally useful.

src/Init/Util.lean
```diff
@@ -30,17 +30,22 @@ def dbgStackTrace {α : Type u} (f : Unit → α) : α := f ()
...
+  panic (info.mkPanicMessage msg)
 
-@[noinline] private def mkPanicMessageWithDecl (modName : String) (declName : String) (line col : Nat) (msg : String) : String :=
-  "PANIC at " ++ declName ++ " " ++ modName ++ ":" ++ toString line ++ ":" ++ toString col ++ ": " ++ msg
+@[neverExtract, inline] def panicWithPos {α : Type u} [Inhabited α] (modName : String) (line col : Nat) (msg : String) : α :=
+  StackInfo.panic ⟨modName, line, col⟩ msg
 
 @[neverExtract, inline] def panicWithPosWithDecl {α : Type u} [Inhabited α] (modName : String) (declName : String) (line col : Nat) (msg : String) : α :=
-  panic (mkPanicMessageWithDecl modName declName line col msg)
+  StackInfo.panic ⟨declName ++ " " ++ modName, line, col⟩ msg
```
gebner, line 48: I think these can be removed now.

digama0:
Question: Do we still need the `panic!` macro? In this PR it is completely trivial: `panic! $arg` expands to `panicWithInfo $arg` which is just a regular definition. You might think it is needed because of interpolation but actually it doesn't even do that: you have to write `panic! s!"..."` if you want interpolation. Should I make it do interpolation? It would be more consistent that way. (Maybe it had something to do with staging, since `panic` is available quite early, but it was refactored in this PR to come a bit later, so that is probably not an issue anymore.)

gebner:
We want to make string interpolation the default anyhow: #407

src/Init/Util.lean
gebner: `Lean.Name` is in the prelude so there is no issue.

I also think it would be better to move the `Lean.Position` structure to `init` if that means we can use the type consistently everywhere.

digama0:
This PR now implements `panic!` interpolation as well. I cleaned up the commits and it has a whopping 3 update-stage0's in it, so it's likely to rot quickly.

leodemoura:
I am very concerned about this PR. It is increasing the size of the generated code quite a bit (see some of the stage0 .c files). I didn't carefully investigate the issue (nor have time to do it), but the problem seems to be that emit extra code at every call site using this new approach. 
We also currently have a mechanism for controlling code generation for debug vs release builds. 
BTW, I am very happy with the call stack dump at panic that has recently been added. I think we can survive with this feature until we add better support for debugging.
I marking this PR to be discussed at the next dev meeting.

digama0:
Can you run the benchmark? I did anticipate that code size growth could be an issue, since there is now a lot more `_closed_n` terms corresponding to all the new places that errors can be thrown. AFAICT that's a fundamental problem here: panics used to only occur in a few places like `Option.get!` itself, now all the *uses* of `get!` get their own panic location information and that info has to be stored in the binary. If we ever get debug info this problem will be much worse.

An alternative solution would be to just flag how many call frames to drop from the call stack when reporting the panic, but this would (1) require us to have the complete call stack and (2) not have optimized out the function of interest. I don't think we can reliably guarantee either one (we can do (1) only on linux).

leodemoura:
!bench

digama0:
One way to decrease the cost of this in lean itself is to have an option `set_option callerInfo false` to be set inside lean, which will make the `caller_info` macro use the caller info of the function being called (i.e. `get!` or `find!`) instead of a newly synthesized caller info. Also the whole thing can be behind a thunk and some auxiliary functions to avoid doing anything other than passing a pointer on the hot path. I'd really like some way to have the default user experience get this debug info though, even if it doesn't outweigh the performance implications inside the lean implementation.

PR #1653, opened 2022-09-27
doc: add documentation on ST monad.

lovettchris:
See discussion in this [Zulip Thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/modifiable.20state.20inside.20a.20ReaderM.20struct).

### DIFF
--- a/doc/monads/st.lean
+++ b/doc/monads/st.lean
@@ -10,9 +10,9 @@ modifying the old state in place.
 This is based on the work of John Launchbury and Simon L Peyton Jones in
 [Lazy Functional State Threads](https://www.microsoft.com/en-us/research/wp-content/uploads/1994/06/lazy-functional-state-threads.pdf).
 
-A state transfomer can be used as a first-class value which means it can be passed to a funciton,
+A reference can be used as a first-class value which means it can be passed to a function,
 returned as a result and stored in a data structure, which means you can store it in a `ReaderM`
-context to make part of that context updatable.
+context to make the value being referenced updatable.
 
 ## Motivating example: A simple Job Server
 


doc/monads/st.lean
```diff
@@ -0,0 +1,117 @@
...
+called `ST` which stands for "State Transfomer" which is a _stateful computation_: the computation
+is seen as transforming one state into another, where the new state will actually be constructed by
+modifying the old state in place.
+
+This is based on the work of John Launchbury and Simon L Peyton Jones in
+[Lazy Functional State Threads](https://www.microsoft.com/en-us/research/wp-content/uploads/1994/06/lazy-functional-state-threads.pdf).
+
+A state transfomer can be used as a first-class value which means it can be passed to a funciton,
+returned as a result and stored in a data structure, which means you can store it in a `ReaderM`
+context to make part of that context updatable.
```
digama0, line 15:
```suggestion
A reference can be used as a first-class value which means it can be passed to a function,
returned as a result and stored in a data structure, which means you can store it in a `ReaderM`
context to make the value being referenced updatable.
```

doc/monads/st.lean
```diff
@@ -0,0 +1,117 @@
...
+
+## Motivating example: A simple Job Server
+
+This simple server has a `ReaderT ServerContext` that contains a bunch of useful state needed by our
+server function, but it also has a reference to an updateable `List Nat` which is typed as `IO.Ref
```
digama0, line 20:
```suggestion
server function, but it also has a reference to an updatable `List Nat` which is typed as `IO.Ref
```
Wiktionary says both are in use but this one is more common. But you also used the other one in the previous paragraph, so be consistent.

doc/monads/st.lean
```diff
@@ -0,0 +1,117 @@
+/-!
+# ST monad
```
digama0, line 2:
```suggestion
# References and the ST monad
```
It makes sense to explain IO.Ref in the same section, and I would try to use this section to motivate the concept of references and explain ST toward the end as a way to use references in pure code.

doc/monads/st.lean
```diff
@@ -0,0 +1,117 @@
...
+  for task in tasks do
+    ofExcept task.get
+
+def server (args : List Nat) : IO (List Nat) := do
+  let hIn ← IO.getStdin
+  let hOut ← IO.getStdout
+  let processedRef : IO.Ref (List Nat) ← IO.mkRef []
+  let context : ServerContext := { hIn, hOut, args, processed := processedRef}
+  processTasks args |>.run context
+  context.processed.get
```
digama0, line 51:
```suggestion
def server (args : Array Nat) : IO (Array Nat) := do
  -- make an array for the completed results
  let processed : IO.Ref (Array Nat) ← IO.mkRef #[]
  -- each task will run in parallel
  let tasks ← args.mapM fun i => do
    IO.asTask do
      IO.sleep 1 -- do some work
      processed.modify (·.push i)
  -- block on the completion of each task
  for task in tasks do
    ofExcept task.get
  -- now we can read the result
  processed.get
```
The example is overly complicated. It should be reduced to focus on the reference operations.

doc/monads/st.lean
lovettchris: well the problem is this is the chapter on "Monads" not the chapter on "References".  It seemed to me that using the ST monad implies using ST.Ref ?  Or am I conflating 2 different things here?

doc/monads/st.lean
lovettchris: Actually, I don't think so.  The reader just learned about ReaderT and StateT and is now seeing that an updatable ST.Ref can be used in a ReaderT context whereas your version would be great for a chapter purely focused on "references" and doesn't care at all about monads.  Plus your version could also be in a section that talks about the limitations of `let mut` in a mutithreaded app and how an IO.Ref can overcome that limitation.

doc/monads/intro.md
```diff
@@ -42,6 +42,12 @@ This section introduces the `StateM` monad. This monad allows you to access a pa
 both read from and write to. It opens the door to fully stateful programming, allowing you to do many
 of the things a function programming language supposedly "can't" do.
 
+## [ST Monad](st.lean.md)
+This section introduces the `ST` monad. This monad allows you to use updatable state in a
+ReaderM context. This is an alternative to `StateM` that is more convenient in some scenarios,
+specifically in multi-threading scenarios because this monad ensures single threaded access to
+the mutable state.
```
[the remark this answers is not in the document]
lovettchris: This is the chapter on Monads where the reader just learned about ReaderM and StateM and is now seeing that an updatable ST.Ref can be used in a ReaderM context which provides a kind of hybrid model somewhere in between ReaderM and StateM.  This is why I have this kind of language and at least this made sense to me when I try to fit all this info in my brain.

I think your comments here may indicate we need a separate doc on "References" that drops all the connection to Monads?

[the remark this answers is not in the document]
digama0: Not necessarily. References and the ST monad are pretty tightly related, and IO references make sense to document in the same section since they are basically the same thing, and it would be way too much to have a section on the IO monad similar to the others. You can't talk about references without talking about monads because the functions are all monadic and you need the do notation to present the examples. Talking about ST *is* talking about references. The stuff with asynchronous programming on the other hand is a distraction, ST doesn't have anything to do with tasks or threads and indeed it doesn't even mix very well.

leodemoura:
I am closing this PR too. Mario made many corrections, but the PR is still does not look good to me. The topic is not high priority for any of the next milestones.
David is a pro at explaining this kind of stuff. Let's him do it.

PR #1654, opened 2022-09-27
feat: datatypes for LSP code actions

EdAyers:
A subset of https://github.com/leanprover/lean4/pull/1551 with just the new datatypes for the LSP protocol. I have also added documentation (copied from spec) and reference links to the LSP specification.

### DIFF
--- a/src/Lean/Data/Json/FromToJson.lean
+++ b/src/Lean/Data/Json/FromToJson.lean
@@ -133,7 +133,7 @@ instance : FromJson Float where
     | (Json.num jn) => Except.ok jn.toFloat
     | _ => Except.error "Expected a number or a string 'Infinity', '-Infinity', 'NaN'."
 
-instance {cmp} [ToJson α]: ToJson (RBMap String α cmp) where
+instance [ToJson α] : ToJson (RBMap String α cmp) where
   toJson m := Json.obj <| RBNode.map (fun _ => toJson) <| m.val
 
 instance {cmp} [FromJson α] : FromJson (RBMap String α cmp) where
--- a/src/Lean/Data/Lsp/Basic.lean
+++ b/src/Lean/Data/Lsp/Basic.lean
@@ -109,7 +109,7 @@ instance : ToJson TextEditBatch :=
 instance : EmptyCollection TextEditBatch := ⟨#[]⟩
 
 instance : Append TextEditBatch :=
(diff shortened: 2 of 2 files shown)


src/Lean/Data/Lsp/Basic.lean
```diff
@@ -91,20 +123,141 @@ structure VersionedTextDocumentIdentifier where
...
+  uri           : DocumentUri
+  options?      : Option DeleteFile.Options := none
+  annotationId? : Option String := none
+  deriving ToJson, FromJson
 
+/-- A change to a file resource.
+
+[reference](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#resourceChanges) -/
+inductive DocumentChange where
+  | create : CreateFile       → DocumentChange
```
EdAyers, line 182: Can I get this to work with `derive ToJson`?

src/Lean/Data/Json/FromToJson.lean
```diff
@@ -133,6 +133,14 @@ instance : FromJson Float where
     | (Json.num jn) => Except.ok jn.toFloat
     | _ => Except.error "Expected a number or a string 'Infinity', '-Infinity', 'NaN'."
 
+instance {cmp} [ToJson α]: ToJson (RBMap String α cmp) where
```
Kha, line 136:
```suggestion
instance [ToJson α] : ToJson (RBMap String α cmp) where
```
`{cmp}` shouldn't do anything here, does it?

src/Lean/Data/Lsp/Basic.lean
```diff
@@ -82,6 +106,14 @@ instance : FromJson TextEditBatch :=
 instance : ToJson TextEditBatch :=
   ⟨@toJson (Array TextEdit) _⟩
 
+instance : EmptyCollection TextEditBatch := ⟨#[]⟩
+
+instance : Append TextEditBatch :=
+  show Append (Array TextEdit) by infer_instance
```
Kha, line 112:
```suggestion
  inferInstanceAs (Append (Array _))
```

Kha (approved):

src/Lean/Data/Lsp/Basic.lean
Kha: Not right now I guess

PR #1656, opened 2022-09-27
feat: fromJson? derivations now say where a parsing error occurred

EdAyers:
Consider the following structures

```
structure Test1 where
  hello : String
  cheese? : Option Nat
  deriving ToJson, FromJson, Repr

structure Test2 where
  jam: Test1
  deriving ToJson, FromJson, Repr

#eval fromJson? (α := Test2) <| Json.mkObj [("jam", Json.mkObj [("hello", 4)])]
```

Previously, the expression in `#eval` would return an error `"String expected"`. However often these parsing errors can occur in very deeply nested structures, making it hard to discover where exactly a string was inspected. 

This PR causes the `fromJson?` method to prepend the field where the failure occurred if parsing fails. So now the above example will yield `"Test2.jam: Test1.hello: String expected"`, making it clear that the badly formed sub-json was in `jam.hello`.

### DIFF
--- a/src/Lean/Elab/Deriving/FromToJson.lean
+++ b/src/Lean/Elab/Deriving/FromToJson.lean
@@ -110,10 +110,8 @@ def mkFromJsonInstanceHandler (declNames : Array Name) : CommandElabM Bool := do
         let header ← mkHeader ``FromJson 0 ctx.typeInfos[0]!
         let fields := getStructureFieldsFlattened (← getEnv) declName (includeSubobjectFields := false)
         let getters ← fields.mapM (fun field => do
-          let jsonField := Prod.snd <| mkJsonField field
-          let quotedField : Term := quote field
-          let getter ← `(getObjValAs? j _ $jsonField)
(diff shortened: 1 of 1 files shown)


src/Lean/Elab/Deriving/FromToJson.lean
```diff
@@ -103,24 +103,31 @@ where
...
       let cmds ← liftTermElabM do
-        let ctx ← mkContext "fromJson" declNames[0]!
+        let ctx ← mkContext "fromJson" declName
         let header ← mkHeader ``FromJson 0 ctx.typeInfos[0]!
-        let fields := getStructureFieldsFlattened (← getEnv) declNames[0]! (includeSubobjectFields := false)
-        let jsonFields := fields.map (Prod.snd ∘ mkJsonField)
+        let fields := getStructureFieldsFlattened (← getEnv) declName (includeSubobjectFields := false)
+        let getters ← fields.mapM (fun field => do
+          let jsonField := Prod.snd <| mkJsonField field
+          let quotedField : Term := quote field
```
Kha, line 114: Did it not work without the `let` and type annotation?

Kha (approved):

PR #1658, opened 2022-09-27
fix: prefer longer parse even if unsuccessful

Kha:
As mentioned in https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.22have.22.20with.20tactic-mode.20proof/near/300188387

### DIFF
--- a/src/Init/Data/Ord.lean
+++ b/src/Init/Data/Ord.lean
@@ -59,7 +59,8 @@ instance : Ord USize where
 instance : Ord Char where
   compare x y := compareOfLessAndEq x y
 
-instance lexOrd [Ord α] [Ord β] : Ord (α × β) where
+/-- The lexicographic order on pairs. -/
+def lexOrd [Ord α] [Ord β] : Ord (α × β) where
   compare p1 p2 := match compare p1.1 p2.1 with
     | .eq => compare p1.2 p2.2
     | o   => o
--- a/src/Lean/Parser/Basic.lean
+++ b/src/Lean/Parser/Basic.lean
@@ -1424,7 +1424,7 @@ def longestMatchStep (left? : Option Syntax) (startSize startLhsPrec : Nat) (sta
   let prevLhsPrec   := s.lhsPrec
   let s             := s.restore prevSize startPos
   let s             := runLongestMatchParser left? startLhsPrec p c s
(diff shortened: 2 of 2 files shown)


src/Init/Data/Ord.lean
```diff
@@ -59,6 +59,10 @@ instance : Ord USize where
 instance : Ord Char where
   compare x y := compareOfLessAndEq x y
 
+instance lexOrd [Ord α] [Ord β] : Ord (α × β) where
```
gebner, line 62: I'd prefer if we didn't make this a global instance.  It conflicts with the general pointwise order instance which works more generally for preorders as well.

Kha: Yeah, I thought we had discussed this instance before and was quite confused that it did not exist at all (but I may have confused it with the general Ord/LT diamond discussion).

src/Lean/Parser/Basic.lean
```diff
@@ -1424,7 +1424,7 @@ def longestMatchStep (left? : Option Syntax) (startSize startLhsPrec : Nat) (sta
   let prevLhsPrec   := s.lhsPrec
   let s             := s.restore prevSize startPos
   let s             := runLongestMatchParser left? startLhsPrec p c s
-  match compare previousScore (score s prio) with
+  match (let _ := @lexOrd; compare previousScore (score s prio)) with
```
Kha, line 1427: Is this the correct idiom then :) ?

digama0: You can also do `local attribute [instance] lexOrd` before the function, or have a type alias `Lex A B` on which to unconditionally put the instance. But this way is fine too, or `(inst := lexOrd)` in `compare`. If this program were subject to proofs I would generally prefer `haveI` here to avoid creating the additional let binding in the generated term, but for this code that's not really a concern and the compiler will normalize the two variants the same anyway.

leodemoura: I will merge as-is. I think the `haveI`/`letI` remark is important, we should discuss it in the next dev-meeting. We also get back to the `let` vs `let dep` discussion. It would be great to use the same style for `letI` (`let inl`?), and have it as part of the `let` family: `let`, `let rec`, `let dep`, `let mut`, ...

PR #1660, opened 2022-09-28
feat: hover for `cases`/`induction` case names

digama0:
This makes the `zero` and `succ` in
```lean
  induction x with
  | zero => ...
  | succ x => ...
```
into clickable links to the inductive constructors. It also does its best with custom inductives but most likely will not work unless the custom inductive is added manually. (Fundamentally this is because the case names correspond to local binding names in the recursor, not directly to definitions.)

### DIFF
--- a/src/Lean/Elab/Tactic/Induction.lean
+++ b/src/Lean/Elab/Tactic/Induction.lean
@@ -321,6 +321,8 @@ where
           if altVars.size > numFieldsToName then
             logError m!"too many variable names provided at alternative '{altName}', #{altVars.size} provided, but #{numFieldsToName} expected"
           let mut (fvarIds, altMVarId) ← altMVarId.introN numFields (altVars.toList.map getNameOfIdent') (useNamesForExplicitOnly := !altHasExplicitModifier altStx)
+          -- Delay adding the infos for the pattern LHS because we want them to nest
+          -- inside tacticInfo for the current alternative (in `evalAlt`)
           let addInfo := do
             if (← getInfoState).enabled then
               if let some declName := declName? then


tests/lean/interactive/hover.lean
```diff
@@ -215,3 +215,24 @@ example : α → α := by my_intro x; assumption
...
+example : Nat → True := by
+  intro x
+      --^ textDocument/hover
+  induction x with
+  | zero => trivial
+  --^ textDocument/hover
+       -- TODO: the next two are broken
+       --v textDocument/hover
+  | succ _ ih => exact ih
+         --^ textDocument/hover
```
digama0, line 238: @Kha Setting aside the `_` binder, do you have any ideas why the hover on `ih` doesn't work? Go-to-def will *target* the correct span of the `ih` binder (for example if you click on the `ih` in `exact ih`), but directly hovering the binder does not reveal the type and instead does the fallback thing of tellling you about the induction alt LHS syntax `| @foo x y z`. But it seems very similar to `intro x`, which does work.

Kha: Ahh, it's the old "how tree-structured should the info tree actually be" discussion - note that in the info tree, the pattern variable infos are outside of the tactic info for that case. Perhaps we should just resurrect https://github.com/Kha/lean4/commit/0bd864deca82ee865a2e5134d5c1f6105f5a701c#diff-83bd7ca17a546fdf495692949c7412fc3a059f292c59e667898804da030dba22L157-L158 again, it's not exactly a terrible complication.

Vtec234:
>  It also does its best with custom inductives but most likely will not work unless the custom inductive is added manually. 

Do you mean "with custom induction principles", or actually that `Nat` specifically somehow has special support?

digama0:
I mean things that aren't inductives but try to act like one by having a bunch of theorems with names similar to the constructors and recursor of an inductive type. Because the implementation uses `isInductive` on the parent of the recursor, that means that if you have a `Nat.myRecOn` (where `Nat` could be any `inductive`) with minor premises named `zero` and `succ`, then the hovers will work, and will target `Nat.zero` and `Nat.succ`. If you name the minor premises `foo` and `bar` then nothing will be linked. And if you have `MyNat.recOn` where `def MyNat := Nat`, then even if `MyNat.zero` and `MyNat.succ` exist they will not be linked.

tests/lean/interactive/hover.lean
digama0: What exactly is the nesting relation we need here? The tactic info for `induction` should be huge here, covering the `induction` all the way until the end of the proof, so I'm not sure how it could have migrated out.

Kha: The current server code assumes that info tree nodes' ranges are contained within their parent range and that sibling ranges are disjoint (apart from duplicates). But as I said, there is a tactic info for each case that spans over all pattern variables yet does not contain their infos.

digama0: oh I see, is this the `| foo x y z =>` tactic info that we use to show the initial and final states for the case?

Kha: Yes, exactly

digama0: I fixed the info tree handling, so now these hovers work as expected. (There was also a similar bug where the `x` in `induction x with ...` was being blocked by the inner tactic info we use to show the original goal state outside the alternatives, which has also been fixed.)

src/Lean/Elab/Tactic/Induction.lean
```diff
@@ -295,17 +316,22 @@ where
             altMVarIds.forM fun mvarId => admitGoal mvarId
       | some altStx =>
         (subgoals, usedWildcard) ← withRef altStx do
-          let unusedAlt :=
+          let altVars := getAltVars altStx
+          let numFieldsToName ← if altHasExplicitModifier altStx then pure numFields else getNumExplicitFields altMVarId numFields
+          if altVars.size > numFieldsToName then
+            logError m!"too many variable names provided at alternative '{altName}', #{altVars.size} provided, but #{numFieldsToName} expected"
+          let mut (fvarIds, altMVarId) ← altMVarId.introN numFields (altVars.toList.map getNameOfIdent') (useNamesForExplicitOnly := !altHasExplicitModifier altStx)
+          let addInfo := do
```
Kha, line 324: Should probably have a comment about the info nesting

Kha (approved):

PR #1661, opened 2022-09-28
feat: CodeActionProvider

EdAyers:
This is a low-level system for registering LSP code actions. LSP code actions are editor integrated code actions. In VScode, you can trigger the `textDocument/codeAction` handler by hitting `Ctrl+.` (`Cmd+.` on macs). If a code action is of the kind `"quickfix"` this will also cause a little lightbulb to appear next to the cursor.

Rather than having a hard-coded handler, users can register their own code actions by creating a `CodeActionProvider : CodeActionParams → RequestM (RequestTask (Array CodeAction))` and registering with the `@[codeActionProvider]` attribute.
In future commits I am going to add features on top of this. In particular I have implemented a special `TacticSuggestion` with some infoview compatibility so users can see the tactic state after the suggestion is applied.

This interface can be used to implement the following applications:
- refactoring code: adding underscores to unused variables,
- suggesting imports
- document-level refactoring: removing unused imports, sorting imports, formatting.
- Helper suggestions for working with type holes (`_`)
- Helper suggestions for tactics.

Related issue: https://github.com/leanprover/lean4/issues/1494

### DIFF
--- a/tests/lean/interactive/codeaction.lean.expected.out
+++ b/tests/lean/interactive/codeaction.lean.expected.out
@@ -9,15 +9,7 @@
        "end": {"line": 29, "character": 4}},
       "newText": "hello!!!"}]}],
   "changes": {},
-  "changeAnnotations": {}},
- "data":
- {"providerResultIndex": 0,
-  "providerName": "helloProvider",
-  "params":
-  {"textDocument": {"uri": "file://codeaction.lean"},
-   "range":
-   {"start": {"line": 29, "character": 4}, "end": {"line": 29, "character": 4}},
-   "context": {"diagnostics": []}}}}
+  "changeAnnotations": {}}}
 {"title": "a long-running action",
  "kind": "refactor",
  "data":


src/Lean/Server/CodeActions.lean
```diff
@@ -0,0 +1,88 @@
...
+import Lean.Server.FileWorker.RequestHandling
+import Lean.Server.InfoUtils
+
+namespace Lean.Server
+
+open Lsp
+open RequestM
+open Snapshots
+
+/-- A code action provider is a function for getting LSP code actions for vscode. You can register them with the `@[codeActionProvider]` attribute.
```
Kha, line 17:
```suggestion
/-- A code action provider is a function for providing LSP code actions in an editor. You can register them with the `@[codeActionProvider]` attribute.
```
:)

src/Lean/Server/CodeActions.lean
```diff
@@ -0,0 +1,88 @@
...
+This is a low-level interface for making LSP code actions.
+This interface can be used to implement the following applications:
+- refactoring code: adding underscores to unused variables,
+- suggesting imports
+- document-level refactoring: removing unused imports, sorting imports, formatting.
+- Helper suggestions for working with type holes (`_`)
+- Helper suggestions for tactics.
+
+When implementing your own `CodeActionProvider`, we assume that no long-running computations are allowed.
+If you need to create a code-action with a long-running computation, we suggest the following approach:
```
Kha, line 28: This feels a bit disappointing to be honest. Would it be too complicated to support `codeAction/resolve` from the beginning? Tbh I think this would have been the more reasonable LSP API from the beginning.

src/Lean/Server/CodeActions.lean
```diff
@@ -0,0 +1,88 @@
...
+private unsafe def evalCodeActionProviderUnsafe [MonadEnv M] [MonadOptions M] [MonadError M] [Monad M] (declName : Name) : M CodeActionProvider := do
+  evalConstCheck CodeActionProvider ``CodeActionProvider declName
+
+/-- Get a `CodeActionProvider` from a declaration name. -/
+@[implementedBy evalCodeActionProviderUnsafe]
+private opaque evalCodeActionProvider [MonadEnv M] [MonadOptions M] [MonadError M] [Monad M] (declName : Name) : M CodeActionProvider
+
+/-- Handles a `textDocument/codeAction` request.
+
+This is implemented by polling all of the registered CodeActionProvider functions.
```
Kha, line 65: "Polling" to me implies repeated sampling of something

src/Lean/Server/CodeActions.lean
Kha: Ah, I missed your edit regarding future work. Perhaps it would be enough to point that out in the docstring then. I just want to avoid the macro approach becoming dominant where an async edit would be the more user-friendly choice.

Vtec234: The `codeAction/resolve` API looks a bit strange since it is a client->server request. I would expect a server->client notification or request that we could use to just push the result to the client once computed, but as specified it seems like the client has to poll for the result. Is that what's supposed to happen @EdAyers ? I am guessing there could be a `codeActionResults` table or something like that in the server which the code action providers store their results into once finished, and `codeAction/resolve` looks up values in that table.

EdAyers: I can add `codeAction/resolve` to this PR. I wasn't sure how to support this though. I guess `CodeActionProvider`'s return type should support a lazy variant of `CodeAction`?

src/Lean/Server/CodeActions.lean
```diff
@@ -0,0 +1,88 @@
...
+  let doc ← readDoc
+  let pos := doc.meta.text.lspPosToUtf8Pos params.range.end
+  bindWaitFindSnap doc (fun s => s.endPos ≥ pos)
+    (notFoundX := return RequestTask.pure #[])
+    fun snap => do
+      let caps ← RequestM.runCoreM snap (do
+        let env ← getEnv
+        let names := codeActionProviderExt.getState env
+        names.foldM (fun b n => b.push <$> evalCodeActionProvider n ) #[]
+      )
```
gebner, line 79:
```suggestion
      let caps ← RequestM.runCoreM snap do
        let env ← getEnv
        let names := codeActionProviderExt.getState env
        names.mapM evalCodeActionProvider
```

src/Lean/Server/CodeActions.lean
gebner: @Vtec234 The whole idea of `codeAction/resolve` is that the code action is only computed if the user selects it.  It works like this:
1. LSP server sends list of possible code actions.
2. User selects a code action.
3. Editor requests the edit for the code action via `codeAction/resolve`.

EdAyers: I mean, what should the Lean-side API be? We don't want code action writers to handle both `textDocument/codeAction` and `codeAction/resolve`, so there should be some system for matching up the `codeAction/resolve` requests with the original `textDocument/codeAction` requests.

gebner:
> Rather than having a hard-coded handler, users can register their own code actions

I think it is a good approach to make the `codeAction` extensible in this way, and we should do the same with all the other requests as well.

But I don't think we should corral users into writing low-level handlers for every code action.  I think you're already planning to do this based on the `TacticSuggestion` reference.  The documentation should reflect this, and make clear that this is a low-level API and users should not use it directly.

src/Lean/Server/CodeActions.lean
EdAyers: If taking the strategy of getting the server to store the array, I'm not sure how the data should be stored. I think the options are:
- `RequestM` is modified to have a state
- Store the state as an `IO.ref`

gebner: I'm not concerned about the implementation, but about storing the *last* code action request specifically.  Does vscode never send a codeAction request while the popup menu is open?  Do other editors behave in the same way?

It's not particularly bad if we get this wrong now, this doesn't affect the editor API and users are supposed to use higher-level APIs instead.

EdAyers: I'm making the assumption that `codeAction/resolve` will only be sent for the most recent result of `textDocument/codeAction`. [The spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeAction_resolve) doesn't say that this is necessary, but I don't think there are common cases where you would want to resolve something that wasn't from the most recent request.

gebner: Code action requests are not only sent due to explicit user actions.  I believe VS Code queries the code actions for the whole file so that it knows where to show the lightbulbs (or maybe just the currently visible part of the file, I never looked too much into it).

So something like the following does not seem impossible:
1. User does ctrl+dot (does this send a separate code action request?)
2. VS Code updates the lightbulbs with a code action request.
3. After a few seconds of decision making, the user selects the second code action, returned by (1), which is no longer the last request.

Kha: @gebner That's really good to know! Then I agree that `resolve` should be a separate function of a `CodeActionProvider`, with no state being stored between their invocations.

src/Lean/Server/CodeActions.lean
```diff
@@ -0,0 +1,155 @@
...
+
+/-- A code action optionally supporting a lazy code action computation that is only run when the user clicks on
+the code action in the editor.
+
+If you want to use the lazy feature, make sure that the `edit?` field on the `eager` code action result is `none`.
+ -/
+structure LazyCodeAction where
+  /-- This is the initial code action that is sent to the server, to implement -/
+  eager : CodeAction
+  lazy? : Option (IO CodeAction) := none
```
gebner, line 43:
```suggestion

The edit should always be computed lazily.
 -/
structure LazyCodeAction where
  /-- Initial code action that is sent to the client, it should not include the edit. -/
  eager : CodeAction
  /-- Final code action, including the edit. -/
  lazy? : Option (IO CodeAction)
```

The more I think about it, I believe we should make lazy edits the default.  We're probably going to have many code actions (particularly in "work-in-progress" files where there are still errors).  I'm still not entirely sure when VS Code requests code actions, but it does request them when you're just moving the cursor.  If we compute them upfront, that would be a lot of code actions to compute and ship.

gebner (commented):
I'm still really concerned about https://github.com/leanprover/lean4/pull/1661#discussion_r982560717

Options:
 1. Skip the `resolve` support for now.  This is probably a bad idea, since editors query code actions automatically and we'd need to compute, pretty-print, and send all the edits every time the editor wants to update them.
 2. Double evaluation.  Store the `CodeActionParams`+index as `data`, and then have `resolve` run the original request again (so that it can resolve the index).
 3. Pass the buck up to the next abstraction layer, and make `CodeActionProvider` a structure with two fields `codeAction : Snapshot → CodeActionParams → RequestM (Array CodeAction)` and `resolve : Snapshot → Json → RequestM CodeAction` (where the request handler injects an `proverId` field for routing).
 4. Store the code actions in the snapshots.  Every snapshot would store only the code actions for the last command (to prevent quadratic explosion).  That is, have a field `codeActions : IO.Ref (Option (Array (CodeAction × IO.Ref (Sum (IO CodeAction) CodeAction))))`.  Bonus: we'd be able to cache code actions.

(2) is the easiest out of the bunch.  The (3) API that came up in the other comment thread looks too error-prone to me, and just shifts around the possibility.  Number (4) is more work, probably improves performance, but doesn't nicely allow code actions spanning more than one command.

My suggestion would be: keep the current `LazyCodeAction` API, but implement it using (2) for now.

Kha:
> My suggestion would be: keep the current LazyCodeAction API, but implement it using (2) for now.

SGTM

gebner (approved):

PR #1663, opened 2022-09-28
doc: add a ninja build system for the docs.

lovettchris:
Now that the doc contains LeanInk sources it is handy to have a fast incremental build using Ninja.

### DIFF
--- a/doc/ninja/configure.py
+++ b/doc/ninja/configure.py
@@ -4,33 +4,44 @@ import ninja_syntax
 
 script_path = os.path.dirname(__file__)
 
+
+def ignore(d):
+    return d.startswith("build") or d.startswith("lean_packages") or \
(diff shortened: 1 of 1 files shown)


Kha:
It makes sense to have a build system other than Nix. But to be honest my first reaction to this PR was "couldn't this be done in a few lines of `make`?". Am I thinking too naively?

Kha:
> I just don't want to write that make file by hand.

I can't say that I enjoy that either, but the result still seems preferable to me.
```
.PHONY: book  # `mdbook` is fast, why duplicate its dependency resolution logic?

LITERATE = $(shell find examples monads -maxdepth 1 -name '*.lean')

book: $(LITERATE:%=%.md)
	mdbook build

%.lean.md: %.lean
	alectryon --frontend lean4+markup $< --backend webpage -o $@
```

leodemoura:
@Kha Good point.

I am closing this PR.

PR #1664, opened 2022-09-29
fix: `.ident` hover in patterns

digama0:
This is the result of many hours of debugging. When you use `| .succ x =>` in a pattern, it was not giving the hover for constant `Nat.succ`, only the whole term `Nat.succ x`. If you write `| Nat.succ x =>` however you do get the hover for the constant. The reason is because the termInfo for `Nat.succ` is added during `collectPatternVars`, but this is too early for dotted identifiers and they are only resolved once we enter the pattern. But once `isPattern` is true, we discard generated termInfos and replace them with a `patternRef` expr annotation.

Normally that would be okay because these annotations are later unpacked after the dependent match eliminator has done its job, but it turns out that for the constructor head itself it will clobber `patternRef` annotations. Plus, even if it did preserve them we aren't really interested in the post-match-elimination expression here, we want to record the data early like is done for plain identifiers.

The solution is simple: override the `isPattern` suppression for this particular termInfo, since we want it to act like it was done before we entered the pattern.

### DIFF
--- a/src/Lean/Elab/Term.lean
+++ b/src/Lean/Elab/Term.lean
@@ -1092,7 +1092,24 @@ def mkTermInfo (elaborator : Name) (stx : Syntax) (e : Expr) (expectedType? : Op
     let e := removeSaveInfoAnnotation e
     return Sum.inl <| Info.ofTermInfo { elaborator, lctx := lctx?.getD (← getLCtx), expr := e, stx, expectedType?, isBinder }
 
-def addTermInfo (stx : Syntax) (e : Expr) (expectedType? : Option Expr := none) (lctx? : Option LocalContext := none) (elaborator := Name.anonymous) (isBinder := false) (force := false) : TermElabM Expr := do
+/--
+Pushes a new leaf node to the info tree associating the expression `e` to the syntax `stx`.
+As a result, when the user hovers over `stx` they will see the type of `e`, and if `e`
+is a constant they will see the constant's doc string.
+
+* `expectedType?`: the expected type of `e` at the point of elaboration, if available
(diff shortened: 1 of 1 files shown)


Kha (approved):
I don't have the pattern elaboration in my working memory, but going from your description the change sounds reasonable

leodemoura:
I want to review this PR before merging.
I am assigning myself as a reviewer.

src/Lean/Elab/Term.lean
```diff
@@ -1092,8 +1092,8 @@ def mkTermInfo (elaborator : Name) (stx : Syntax) (e : Expr) (expectedType? : Op
     let e := removeSaveInfoAnnotation e
     return Sum.inl <| Info.ofTermInfo { elaborator, lctx := lctx?.getD (← getLCtx), expr := e, stx, expectedType?, isBinder }
 
-def addTermInfo (stx : Syntax) (e : Expr) (expectedType? : Option Expr := none) (lctx? : Option LocalContext := none) (elaborator := Name.anonymous) (isBinder := false) : TermElabM Expr := do
-  if (← read).inPattern then
+def addTermInfo (stx : Syntax) (e : Expr) (expectedType? : Option Expr := none) (lctx? : Option LocalContext := none) (elaborator := Name.anonymous) (isBinder := false) (force := false) : TermElabM Expr := do
```
leodemoura, line 1095: Could you please add a docstring and document `force`?

leodemoura (changes requested):
Document `force` in docstring at `addTermInfo`. Could you please also add note describing why it was added and a reference to this PR?

leodemoura:
Thanks for the changes. I will merge as soon as the conflicts above have been addressed.

leodemoura (approved):

PR #1666, opened 2022-09-30
fix: replace highlight with diffTag

EdAyers:
Addendum to #1610 to replace `highlightColor` with something more semantic re @gebner's comment https://github.com/leanprover/lean4/pull/1610#issuecomment-1258231789

### DIFF
--- a/src/Lean/Widget/InteractiveCode.lean
+++ b/src/Lean/Widget/InteractiveCode.lean
@@ -35,7 +35,7 @@ structure SubexprInfo where
   -- TODO(WN): add fields for semantic highlighting
   -- kind : Lsp.SymbolKind
   /-- Ask the renderer to highlight this node in the given color. -/
-  diffTag? : Option DiffTag := none
+  diffStatus? : Option DiffTag := none
   deriving Inhabited, RpcEncodable
 
 /-- Pretty-printed syntax (usually but not necessarily an `Expr`) with embedded `Info`s. -/
@@ -53,7 +53,7 @@ def CodeWithInfos.pretty (tt : CodeWithInfos) :=
   tt.stripTags
 
 def SubexprInfo.withDiffTag (tag : DiffTag) (c : SubexprInfo) : SubexprInfo :=
-  {c with diffTag? := some tag }
+  {c with diffStatus? := some tag }
 
 /-- Tags a pretty-printed `Expr` with infos from the delaborator. -/
 partial def tagExprInfos (ctx : Elab.ContextInfo) (infos : SubExpr.PosMap Elab.Info) (tt : TaggedText (Nat × Nat))
--- a/tests/lean/interactive/Diff.lean.expected.out
+++ b/tests/lean/interactive/Diff.lean.expected.out
@@ -1,265 +1,265 @@
 { goals := #[{ type := Lean.Widget.TaggedText.tag
-                         { subexprPos := "/", diffTag? := none }
+                         { subexprPos := "/", diffStatus? := none }
                          (Lean.Widget.TaggedText.append
                            #[Lean.Widget.TaggedText.tag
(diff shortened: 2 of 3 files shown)


leodemoura (approved):

src/Lean/Widget/InteractiveCode.lean
```diff
@@ -33,7 +35,7 @@ structure SubexprInfo where
   -- TODO(WN): add fields for semantic highlighting
   -- kind : Lsp.SymbolKind
   /-- Ask the renderer to highlight this node in the given color. -/
-  highlightColor? : Option HighlightColor := none
+  diffTag? : Option DiffTag := none
```
gebner, line 38:
```suggestion
  diffStatus? : Option DiffTag := none
```

gebner (approved):

Vtec234:
For reasons explained [here](https://github.com/leanprover/lean4/pull/1610#issuecomment-1254135576) I am still somewhat opposed to putting a `DiffTag` in a `SubexprInfo`. In short, that's because `SubexprInfo` is the data that is rendered in every `InteractiveCode` component within the infoview, but not every `InteractiveCode` is a diff; on the other hand, it does make sense to make every subexpression *stylable*, which in particular allows styling diffs, but not *only* diffs, which is what this PR is doing.

Suppose that in the future we are writing a tactic which wants to display a `MessageLog` info within which one subexpression should stand out visually, for example by being highlighted in a contrasting colour. With the current, styling names, the tactic can just mark it as `green`. After this change, the tactic would have to mark the subexpr as `wasInserted`, which makes very little sense. That, or we'd have to introduce another `Status` field to `SubexprInfo`.

I agree that `green/red/etc` is not great because themes differ. And in fact, we can think of `DiffStatus` as a *styling* instruction which means exactly "draw this like `git diff` would". So how about the following: make `diffStatus?` into a generic `highlightStyle?` field, and change `DiffStatus` to `HighlightStyle`, but keep all the current constructor. That way we keep the "draw as a diff" stuff, but the `HighlightStyle` is extensible to different styling instructions.

gebner:
> but not every `InteractiveCode` is a diff

In general I completely agree with you, because this is an issue we already have: `SubexprInfo` is used for different and conflicting ends.  In the beginning, it was only used to mark highlightable subexpressions.  Then we also used it to mark go-to-definition spans.  Here we want to use it for diff highlighting.  In the future we might want to use it for keyword highlighting as well.  We should make it clear that not every `SubexprInfo` means all of these things.

Come to think of it, what happens if the diff marks a partially applied function as changed?  Do we get an unwanted popup?  (That is, say `f a b` is the expression, and `f a` is highlighted as inserted.  Then `f a` should *not* get a popup.)

I don't have a strong opinion on how to name the field, it can be `diffStatus?` or `highlighting?` or `keywordKind?` or everything in  a single field.  But the values should be semantic; it should be `inserted` and `highlighted` and `variable`.  We want to leverage the editor theme for these different styles so that it looks consistent with the editor, and allow the user to customize the styles for accessibility (color blindness) and personal preference reasons.

EdAyers:
> @Vtec234  I am still somewhat opposed to putting a DiffTag in a SubexprInfo.

I've come around to @gebner's point of view on this.
Some solace may be derived from the fact that at least the `diffTags` field can be set to None and so may be ignored when using SubExprInfo for other tasks.
I can change `diffTag?` back to `tags?`but keep semantic tag names if that is preferred.

> Come to think of it, what happens if the diff marks a partially applied function as changed?

At the moment, the diff algorithm doesn't check partially applied exprs (the recursion is on each arg from `Expr.getArgs`), so `f a` changing would be interpreted as the whole expression `f a b` changing.

EdAyers:
We had a network fail can we rerun the CI/macOS task?

Issue #1461, opened 2022-08-11
docstring style guide


PatrickMassot:
Leo asked for help from the community to write docstrings in the Lean 4 codebase. It would be nice to have a docstring style guide to help in this effort and ensure some consistency. Below I'll list some questions but feel free to ignore any item that you think is irrelevant.

Potential questions includes:
* Should docstring refer to variable names? Right now this is confusing since tooltips don't show those names, only the signature, but this will hopefully improve. Mario points out that variable names are user-facing since we can use named argument in function calls. 
* How to start the docstring of a function? We don't have to impose a style but if you have preferences then it's better to tell before people start writing. Existing styles include
    * Using a sample function call. For instance the docstring of `ReaderT.adapt` starts with "`adapt (f : ρ' → ρ)` precomposes function `f`"
    * An infinive verb (or is it conjuguated using imperative?) like "Execute an `EStateM` on initial state `s` to get a `Result`." for `EStateM.run`
    * A conjuguated verb like "Converts a `UInt64` to a `USize` by reducing modulo `USize.size`." for UInt64.toUSize
* Do we want a rigid way of "repeating the signature" such as the ones for python listed at http://daouzli.com/blog/docstring.html?
* How should we describe possible failures? Some existing docstrings don't mention failure at all. Some say "fails if..." without anything specific, some mention throwing exceptions or panicking. 
* Is it ok to have docstring that refer to other declarations in an essential way? For instance the docstring of `ReaderT.readThe` is "Like `read`, but with `ρ` explicit. This is useful if a monad supports `MonadReaderOf` for multiple different types `ρ`."
  

We also need an answer to the syntax vs elab discussion from https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.5BRFC.5D.20Hover.20doc.20strings.20.28lean4.231443.29.

On a slightly different topic, I think it would be nice to have some list of files to prioritize since there are so many files to document.

leodemoura:
@digama0 You made great suggestions during the last dev meeting. It would be great if you could document them here.

david-christiansen:
I believe all of these questions are now answered in [the style guide](https://github.com/leanprover/lean4/blob/master/doc/style.md). Please let me know if it's missing answers to anything and I'll get it updated!

PR #1667, opened 2022-09-30
doc: normalize Init.Conv docs

Kha:
This commit was co-written by OpenAI's davinci model, using the following prompt:
```
Change first sentence inside comment markers from imperative mood to indicative mood and make sure it ends with a period.
Input: /-- Restore backtrackable parts of the state. -/
Output: /-- Restores backtrackable parts of the state. -/
Input: /-- Creates the expression `d → b` -/
Output: /-- Creates the expression `d → b`. -/
Input: /-- Remove local instance with the given `fvarId`. Do nothing if `localInsts` does not contain any free variable with id `fvarId`. -/
Output: /-- Removes local instance with the given `fvarId`. Does nothing if `localInsts` does not contain any free variable with id `fvarId`. -/
Input: <docstring>
Output:
```
This cost around $0.30 of free credits to generate. I did have to retract about half of the generated suggestions because they didn't make sense.

Why indicative mood? While @leodemoura and I seemed to use imperative mood in a lot of places, I noticed that many newer contributors defaulted to indicative mood, which indeed seems to be the default in other stdlibs such as Rust's or Haskell's. There is also the stylistic argument that it better fits the alternative "`f x` does ..." style that we also use in a lot of places.

### DIFF
--- a/src/Init/Conv.lean
+++ b/src/Init/Conv.lean
@@ -59,10 +59,10 @@ in "head position" until a constructor is exposed. For example, `List.map f [a, 
 weak head normalizes to `f a :: List.map f [b, c]`. -/
 syntax (name := whnf) "whnf" : conv
 
-/-- Expand let-declarations and let-variables. -/
+/-- Expands let-declarations and let-variables. -/
 syntax (name := zeta) "zeta" : conv
 
-/-- Put term in normal form, this tactic is ment for debugging purposes only -/
+/-- Puts term in normal form, this tactic is meant for debugging purposes only. -/
 syntax (name := reduce) "reduce" : conv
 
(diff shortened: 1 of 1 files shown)


leodemoura:
> Why indicative mood? While @leodemoura and I seemed to use imperative mood in a lot of places, I noticed that many newer contributors defaulted to indicative mood, which indeed seems to be the default in other stdlibs such as Rust's or Haskell's. There is also the stylistic argument that it better fits the alternative "f x does ..." style that we also use in a lot of places.

I am happy to use the same style used in Rust and Haskell.

PatrickMassot:
Maybe this is a good opportunity to revive #1461?

PR #1671, opened 2022-10-01
fix: derive Repr missing a comma

### LINKED ISSUE #1668: Inconsistent deriving of Repr for polymorphic structures
### Prerequisites

### Description

Ever since the lean4 nightly from 16 September, `deriving Repr` is missing a comma between fields on a polymorphic version of a `structure`, even though it has it on the monomorphic version. This results in invalid syntax for the polymorphic version. The problem cannot be reproduced in the 15 September nightly.

### Steps to Reproduce

First, open a new Lean file in a project with the following `lean-toolchain` file: `leanprover/lean4:nightly-2022-09-16`

Next, enter the following code:
```
structure Point where
  x : Nat
  y : Nat
deriving Repr

structure PPoint (α : Type) where
  x : α
  y : α
deriving Repr

def origin : Point :=
  { x := Nat.zero, y := Nat.zero }

def natOrigin : PPoint Nat :=
  { x := Nat.zero, y := Nat.zero }

#eval origin
#eval natOrigin
```

**Expected behavior:** I expect that both `#eval` lines produce identical output, namely `{ x := 0, y := 0 }`

**Actual behavior:** The second eval line produces the output `{ x := 0y := 0 }`

On the 15th, I can see that there was a change to `deriving Repr` that plausibly affects the placement of commas, but I haven't investigated in any depth: https://github.com/leanprover/lean4/commit/bbc70c4cf0418a8ed5bc80bc32033f9f31c88fe5

### Versions
```
$ lean --version
Lean (version 4.0.0-nightly-2022-09-16, commit abe1f7f6f981, Release)
$ uname -a
Linux localhost.localdomain 5.19.11-200.fc36.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Sep 23 15:07:44 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

The problem can also be reproduced in the September 30 nightly, as well as a few of the intervening nightlies.

pcpthm:
Fix #1668


david-christiansen:
Thanks!

PR #1680, opened 2022-10-03
fix: handle multi namespace/section in foldingRange and documentSymbol

digama0:
In lean 4, unlike lean 3, `namespace`/`section` and `end` are not strictly paired -- the following nesting pattern is valid:
```lean
namespace O.A
namespace B
namespace C.D
end B.C.D
section E.F
end F
end A.E
```
This PR fixes the `documentSymbol` and `foldingRange` request handlers to handle this situation correctly, by breaking up start or end commands as necessary to create a well nested set of scopes, as if it were written:

```lean
namespace O namespace A
namespace B
namespace C.D
end C.D end B
section E section F
end F
end E end A

end O
```
For folding ranges, if you have multiple folding ranges starting on the same line only the first one is presented and the others are ignored in vscode, so we suppress all but the largest folding range starting at a given `namespace` or `section` command.

The `toDocumentSymbols` function is also refactored in this PR to be tail recursive and not make so many intermediate lists, making it resemble the `addRanges` function more closely.

### DIFF
--- a/src/Lean/Server/FileWorker/RequestHandling.lean
+++ b/src/Lean/Server/FileWorker/RequestHandling.lean
@@ -265,6 +265,8 @@ partial def handleDocumentHighlight (p : DocumentHighlightParams)
       return #[]
 
 structure NamespaceEntry where
+  /-- The list of the name components introduced by this namespace command,
+  in reverse order so that `end` will peel them off from the front. -/
   name : List Name
   stx : Syntax
   selection : Syntax
@@ -304,10 +306,10 @@ where
     | [] => stack.foldl (fun syms entry => entry.finish text syms none) syms
     | stx::stxs => match stx with
       | `(namespace $id)  =>
-        let entry := { name := id.getId.components', stx, selection := id, prevSiblings := syms }
+        let entry := { name := id.getId.componentsRev, stx, selection := id, prevSiblings := syms }
         toDocumentSymbols text stxs #[] (entry :: stack)
       | `(section $(id)?) =>
-        let name := id.map (·.getId.components') |>.getD [`«»]
+        let name := id.map (·.getId.componentsRev) |>.getD [`«»]
         let entry := { name, stx, selection := id.map (·.raw) |>.getD stx, prevSiblings := syms }
         toDocumentSymbols text stxs #[] (entry :: stack)
       | `(end $(id)?) =>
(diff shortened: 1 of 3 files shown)


src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -440,16 +465,33 @@ partial def handleFoldingRange (_ : FoldingRangeParams)
     isImport stx := stx.isOfKind ``Lean.Parser.Module.header || stx.isOfKind ``Lean.Parser.Command.open
 
     addRanges (text : FileMap) sections
-    | [] => return
+    | [] => do
+      if let (_, start)::rest := sections then
+        addRange text FoldingRangeKind.region start text.source.endPos
+        addRanges text rest []
+      else
+        return
```
Kha, line 473: Can be elided, no?

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -272,20 +295,38 @@ partial def handleDocumentSymbol (_ : DocumentSymbolParams)
...
+      Array DocumentSymbol :=
     match stxs with
-    | [] => ([], [])
+    | [] => stack.foldl (fun syms entry => entry.finish text syms none) syms
     | stx::stxs => match stx with
-      | `(namespace $id)  => sectionLikeToDocumentSymbols text stx stxs (id.getId.toString) SymbolKind.namespace id
-      | `(section $(id)?) => sectionLikeToDocumentSymbols text stx stxs ((·.getId.toString) <$> id |>.getD "<section>") SymbolKind.namespace (id.map (·.raw) |>.getD stx)
-      | `(end $(_id)?) => ([], stx::stxs)
+      | `(namespace $id)  =>
+        let entry := { name := id.getId.components', stx, selection := id, prevSiblings := syms }
```
Kha, line 307: OMG, we should definitely rename `components'` to `componentsRev`. Your `foldr` was a complete mystery to me before I realized that, there should be a comment on *why* we store `name` like that.

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -272,20 +295,38 @@ partial def handleDocumentSymbol (_ : DocumentSymbolParams)
...
+        toDocumentSymbols text stxs #[] (entry :: stack)
+      | `(section $(id)?) =>
+        let name := id.map (·.getId.components') |>.getD [`«»]
```
Kha, line 310: This is to differentiate it from `.anonymous` from a partial syntax tree I assume?

Kha (approved):
LGTM, but more comments and docstrings for the tricky parts would be appreciated!

src/Lean/Server/FileWorker/RequestHandling.lean
digama0: This isn't an arbitrary choice: anonymous `section` is literally equivalent to `section «»`, and 
```lean
section A
section 
namespace B
end A.«».B
```
is legal, cursed though it is.

PR #1685, opened 2022-10-04
chore: benchmark new compiler

Kha:


Kha:
!bench

Kha:
Looks good 
image

gebner:
Does this measure the same thing as the old metric?  From what I can tell the old one includes IR compilation as well.

Kha:
Correct, it only measures the parts that are already there, so the two numbers should not be compared directly without keeping that in mind

leodemoura:
Remarks:

- The new code generator is re-type-checking the code after each pass. This is useful for debugging, but we will not enable these checks by default in the future.

- The most expensive passes are already in the new code generator: `simp` and `specialize`.

PR #1691, opened 2022-10-05
Bump up to LLVM 15

bollu:
Let us refactor the patch that adds LLVM support to Lean4 into another component: We first bump up LLVM to LLVM 15, and then add the bindings.

We see that the windows build breaks at this stage in itself. Help debugging this would be great, @Kha !

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
             os: ubuntu-latest
             release: true
             shell: nix-shell --arg pkgsDist "import (fetchTarball \"channel:nixos-19.03\") {{}}" --run "bash -euxo pipefail {0}"
-            llvm-url: https://github.com/leanprover/lean-llvm/releases/download/14.0.0/lean-llvm-x86_64-linux-gnu.tar.zst
+            llvm-url: https://github.com/leanprover/lean-llvm/releases/download/15.0.1/lean-llvm-x86_64-linux-gnu.tar.zst
             prepare-llvm: ../script/prepare-llvm-linux.sh lean-llvm*
             binary-check: ldd -v
             # foreign code may be linked against more recent glibc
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
The failure on windows is surprising to me. Here a couple of things I already looked into: 

https://releases.llvm.org/15.0.0/projects/libcxx/docs/ReleaseNotes.html mentions:
> The _LIBCPP_ABI_USE_CXX03_NULLPTR_EMULATION macro controlling whether we use an emulation for std::nullptr_t in C++03 mode has been removed. After this change, _LIBCPP_ABI_USE_CXX03_NULLPTR_EMULATION will not be honoured anymore and there will be no way to opt back into the C++03 emulation of std::nullptr_t.

which is further discussed at https://reviews.llvm.org/D109459 and https://reviews.llvm.org/D114786. I had tried to build lean4 stage 1 with C++11, but could not get it working.

There are also a small number of reports that point to wrong include directories (mostly for apple builds). I am not sure what is breaking here.

script/prepare-llvm-linux.sh
```diff
@@ -29,8 +29,8 @@ find stage1 -type f -exec strip --strip-unneeded '{}' \; 2> /dev/null
 # lean.h dependencies
 $CP llvm/lib/clang/*/include/{std*,__std*,limits}.h stage1/include/clang
 # ELF dependencies, must be put there for `--sysroot`
-$CP $GLIBC/lib/crt* llvm/lib/
-$CP $GLIBC/lib/crt* stage1/lib/
+$CP $GLIBC/lib/*crt* llvm/lib/
```
bollu, line 32: Ah, that's what you meant. Thanks!

bollu, line 32: Ah, that's what you meant. Thanks!

bollu, line 32: Ah, that's what you meant by adding them from `$GLIBC`. Thanks!

PR #1693, opened 2022-10-06
feat: automatic extension names

digama0:
The name of an environment extension affects very little; it has to be globally unique but otherwise only shows up in some error messages. The various extensions were making up random names for this, without namespacing for the most part, and when reading the code it's not always obvious whether this name is important / will come up later (it won't, by contrast to attribute names which are user-facing).

This PR simplifies the interface by making extension names be declaration names. This solves the uniqueness issue, and also makes them slightly more useful since now you can look up a declaration from an extension name for hovers and whatnot (although this is not implemented in this PR). By using the `decl_name%` macro we can automatically pick up a useful declaration name, or inherit one from the caller. As a result, all the `registerEnvExtension` calls are simplified to no longer give a `name` (although it is still legal to do so).


src/Lean/Environment.lean
```diff
@@ -612,13 +612,9 @@ unsafe def Environment.freeRegions (env : Environment) : IO Unit :=
...
-  let entries : Array (Name × Array EnvExtensionEntry) := pExts.size.fold
-    (fun i result =>
-      let state  := (pExts.get! i).getState env
-      let exportEntriesFn := (pExts.get! i).exportEntriesFn
-      let extName    := (pExts.get! i).name
-      result.push (extName, exportEntriesFn state))
-    #[]
+  let entries := pExts.map fun pExt =>
+    let state := pExt.getState env
+    (pExt.name, pExt.exportEntriesFn state)
```
digama0, line 617: The original code was a bit weird here; this is just a peephole simplification since I was in the area.

leodemoura:
Merged manually because of conflicts.

PR #1701, opened 2022-10-07
fix: path normalization should not case-normalize entire path

Kha:
This follows Rust's approach: https://doc.rust-lang.org/std/path/index.html#case-sensitivity

### DIFF
--- a/src/Init/System/Uri.lean
+++ b/src/Init/System/Uri.lean
@@ -83,33 +83,32 @@ a single unicode code point and these will also be decoded correctly. -/
 def unescapeUri (s: String) : String :=
   UriEscape.decodeUri s
 
-/-- Convert the given FilePath to a "file:///encodedpath" Uri
-where the encoded path may contain %xx escaping when needed. -/
+/-- Convert the given FilePath to a "file:///encodedpath" Uri. -/
 def pathToUri (fname : System.FilePath) : String := Id.run do
   let mut uri := fname.normalize.toString
   if System.Platform.isWindows then
+    -- normalize drive letter
+    -- lower-case drive letters seem to be preferred in URIs
+    if uri.length >= 2 && (uri.get 0).isUpper && uri.get ⟨1⟩ == ':' then
+      uri := uri.set 0 (uri.get 0).toLower
     uri := uri.map (fun c => if c == '\\' then '/' else c)
   uri := uri.foldl (fun s c => s ++ UriEscape.uriEscapeAsciiChar c) ""
   let result := if uri.startsWith "/" then "file://" ++ uri else "file:///" ++ uri
   result
 
 /-- Convert the given uri to a FilePath stripping the 'file://' prefix,
-ignoring the optional host name and unescaping any %HH escaped chars.
-It is also careful to create correct paths on Windows that start with a drive letter. -/
+ignoring the optional host name. -/
 def fileUriToPath? (uri : String) : Option System.FilePath := Id.run do
   if !uri.startsWith "file://" then
     none
   else
     let mut p := (unescapeUri uri).drop "file://".length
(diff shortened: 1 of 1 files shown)


lovettchris (approved):

lovettchris:
But now I do worry that the `structure References` which has some `HashMap`s on these paths that it looks up using `references.findAt module`

I'm not sure if this function will still do the right thing, and the comment is no longer valid:
```lean
@[export lean_module_name_of_file]
def moduleNameOfFileName (fname : FilePath) (rootDir : Option FilePath) : IO Name := do
  let fname ← IO.FS.realPath fname
  let rootDir ← match rootDir with
    | some rootDir => pure rootDir
    | none         => IO.currentDir
  let mut rootDir ← realPathNormalized rootDir
  if !rootDir.toString.endsWith System.FilePath.pathSeparator.toString then
    rootDir := ⟨rootDir.toString ++ System.FilePath.pathSeparator.toString⟩
  if !rootDir.toString.isPrefixOf fname.normalize.toString then
    throw $ IO.userError s!"input file '{fname}' must be contained in root directory ({rootDir})"
  -- NOTE: use `fname` instead of `fname.normalize` to preserve casing on all platforms
  let fnameSuffix := fname.toString.drop rootDir.toString.length
  let modNameStr := FilePath.mk fnameSuffix |>.withExtension ""
  let modName    := modNameStr.components.foldl Name.mkStr Name.anonymous
  pure modName
```
And I believe we will need some "case insensitive" file name comparisons in LSP as a result of this change, for example the `if !rootDir.toString.isPrefixOf fname.normalize.toString then` comparison here will need to be case insensitive comparison on Windows.  We should encapsulate all that logic in `FilePath` and only use FilePath methods to compare file names, and subsets of the path thereof.

lovettchris:
> That problem already existed before this PR, and in fact should improve from trying harder to use only one case. I dearly hope that with this, the language server does not really have to worry about casing in practice.

I am testing it and so far it looks fine, I renamed a subfolder to lower case, but referenced it in the lakefile.lean using upper case, and goto-definition still works across these files.  The VS code file tab shows upper case "D:\Temp\lean_examples\Foo\Foo\Hello.lean" when the real file on disk is "D:\Temp\lean_examples\Foo\foo\hello.lean".  So somehow this still works through the LSP and VS code doesn't care about the case differences.  So I think part is good for now.

lovettchris:
> @lovettchris What shall we do about the broken `Uri` test? Should `pathToUri` do the reverse transformation?

I guess I'm curious why drive letter case folding is even needed?  it was only needed in rollup because they had nodejs regular expressions looking for uppercase char colon regex patterns which we don't have.  

So I'm now testing to see what happens when I comment out lines 45,46 of normalize. What bug was that fixing?

Kha:
> So I'm now testing to see what happens when I comment out lines 45,46 of normalize. What bug was that fixing?

I couldn't find the reason, it was added very early: https://github.com/rust-lang/rust/commit/4a9dd3f8403f84ad518b0119fa2312c379df5dd7#diff-2fddd6220cb2006f2e025a3a682366ec85198a4c192e4e8c479bf72be6f2aa5aR240. I'm okay with just reverting it for now and seeing what, if anything, breaks.

leodemoura:
> I'm okay with just reverting it for now and seeing what, if anything, breaks.

Please don't revert. Let's go with the Rust approach.

This PR is high priority. Could you please disable the broken test for now and merge?

Kha:
Since a conversion roundtrip failure presumably is the source of the bug, I decided to be explicit and do the reverse conversions in the URI functions. With that the drive letter normalization in `FilePath.normalize` is not strictly necessary anymore to fix the bug, but I'm leaving it in there to follow Rust's behavior.

PR #1702, opened 2022-10-07
doc: fix link to initialization section in ffi section

dwrensha:
The current link goes to doc/dev#init, where there is nothing about initialization. This PR fixes the link so that it points to the initialization section lower down on the ffi page.


dwrensha:
I tested this by building the docs locally, and it worked as expected.

PR #1706, opened 2022-10-08
FloatLetIn

hargoniX:
Float let statements into `cases` arms if possible in order to avoid unnecessary computation.

Notably all type class instances and projections on type class instances are excluded from this since they are for free anyways.

### DIFF
--- a/src/Lean/Compiler/LCNF/FloatLetIn.lean
+++ b/src/Lean/Compiler/LCNF/FloatLetIn.lean
@@ -1,0 +1,311 @@
+/-
+Copyright (c) 2022 Henrik Böving. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Henrik Böving
+-/
+import Lean.Compiler.LCNF.CompilerM
+import Lean.Compiler.LCNF.FVarUtil
+import Lean.Compiler.LCNF.PassManager
+import Lean.Compiler.LCNF.Types
+
+namespace Lean.Compiler.LCNF
+
+namespace FloatLetIn
+
+/--
+The decision of the float mechanism.
+-/
+inductive Decision where
+|
+  /--
+  Push into the arm with name `name`.
+  -/
+  arm (name : Name)
+| /--
+  Push into the default arm.
+  -/
+  default
+|
+  /--
+  Dont move this declaration it is needed where it is right now.
+  -/
+  dont
+|
+  /--
+  No decision has been made yet.
+  -/
+  unknown
+deriving Hashable, BEq, Inhabited, Repr
+
+def Decision.ofAlt : Alt → Decision
+| .alt name _ _ => .arm name
+| .default _ => .default
+
+/--
+The context for `BaseFloatM`.
+-/
+structure BaseFloatContext where
+  /--
+  All the declarations that were collected in the current LCNF basic
+  block up to the current statement (in reverse order for efficiency).
+  -/
+  decls : List CodeDecl := []
+
+/--
+The state for `FloatM`
+-/
+structure FloatState where
+  /--
+  A map from identifiers of declarations to their current decision.
+  -/
+  decision : HashMap FVarId Decision
+  /--
+  A map from decisions (excluding `unknown`) to the declarations with
+  these decisions (in correct order). Basically:
+  - Which declarations do we not move
+  - Which declarations do we move into a certain arm
+  - Which declarations do we move into the default arm
+  -/
+  newArms : HashMap Decision (List CodeDecl)
+
+/--
+Use to collect relevant declarations for the floating mechanism.
+-/
+abbrev BaseFloatM :=  ReaderT BaseFloatContext CompilerM
+
+/--
+Use to compute the actual floating.
+-/
+abbrev FloatM := StateRefT FloatState BaseFloatM
+
+/--
+Add `decl` to the list of declarations and run `x` with that updated context.
+-/
+def withNewCandidate (decl : CodeDecl) (x : BaseFloatM α) : BaseFloatM α :=
+  withReader (fun r => { r with decls := decl :: r.decls }) do
+    x
+
+/--
+Run `x` with an empty list of declarations.
+-/
+def withNewScope (x : BaseFloatM α) : BaseFloatM α := do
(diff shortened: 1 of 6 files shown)


src/Lean/Compiler/LCNF/FVarUtil.lean
```diff
@@ -0,0 +1,156 @@
...
+  mapFVarM := LetDecl.mapFVarM
+  forFVarM := LetDecl.forFVarM
+
+partial def Code.mapFVarM [MonadLiftT CompilerM m] [Monad m] (f : FVarId → m FVarId) (c : Code) : m Code := do
+  match c with
+  | .let decl k =>
+    let decl ← LetDecl.mapFVarM f decl
+    return Code.updateLet! c decl (← mapFVarM f k)
+  | .fun decl k =>
+    -- XXX: We do not visit params...is this okay/intuitive?
```
leodemoura, line 75: We should visit params.

hargoniX: To explain my thought process here. I do also not visit e.g. the FVarId's that are names of say the let declarations but only the ones that are actually mentioned in the code so my thought process was to treat the parameters similar since they should occur in values or types if they are being used.

If I am supposed to visit parameters, should I also visit the FVarIds of the declarations?

leodemoura: I meant visiting the parameter types, they may contain free variable ids.

src/Lean/Compiler/LCNF/FloatLetIn.lean
```diff
@@ -0,0 +1,263 @@
...
+
+/--
+The context for `BaseFloatM`.
+-/
+structure BaseFloatContext where
+  /--
+  All the declarations that were collected in the current LCNF basic
+  block up to the current statement.
+  -/
+  decls : Array CodeDecl := #[]
```
leodemoura, line 52: I think it is better using `List CodeDecl` or `PArray CodeDecl`. Types used with `ReaderT` are seldom used linearly. For example,  the function `withNewCandidate` will force the array to be copied since it is not an unique reference. If we had just a few calls to `withNewCandidate`, it would be fine to use `Array CodeDecl`, but this is not the case here.

src/Lean/Compiler/LCNF/FloatLetIn.lean
```diff
@@ -0,0 +1,263 @@
...
+  -/
+  decision : HashMap FVarId Decision
+  /--
+  A map from decisions (excluding `unknown`) to the declarations with
+  these decisions. Basically:
+  - Which declarations do we not move
+  - Which declarations do we move into a certain arm
+  - Which declarations do we move into the default arm
+  -/
+  newArms : HashMap Decision (Array CodeDecl)
```
leodemoura, line 69: `HashMap`s of arrays are often problematic because they require discipline to update the arrays linearly. 
For example, in the following code
```
s.newArms.insert .dont (s.newArms.find! .dont |>.push decl)
```
we first retrieve the array from the hashmap. Now, we have two references to it: one inside of the hashmap, and another that we are passing to `Array.push`. Thus, the array is not updated destructively. 
To ensure it is, we need to remove the entry from the hashmap, then update the array, and finally reinsert it.
In the future, we hope the have nice combinators for performing this kind of operation.
Right now, I think it is easier to just use `HashMap Decision (List CodeDecl)`.

src/Lean/Compiler/LCNF/FloatLetIn.lean
```diff
@@ -0,0 +1,263 @@
...
+  withReader (fun _ => {}) do
+    x
+
+/--
+Whether to ignore `decl` for the floating mechanism. We want to do this if:
+- `decl`' is storing a typeclass instance
+- `decl` is a projection from a variable that is storing a typeclass instance
+-/
+def ignore? (decl : LetDecl) : BaseFloatM Bool :=  do
+   if (← isClass? decl.type).isSome then
```
leodemoura, line 101: Let's use `isArrowClass?` instead. It subsumes `isClass?`.

src/Lean/Compiler/LCNF/FloatLetIn.lean
```diff
@@ -0,0 +1,263 @@
...
+    cs.alts.forM goAlt
+
+/--
+Compute the initial new arms. This will just set up a map from all arms of
+`cs` to empty `Array`s, plus one additional entry for `dont`.
+-/
+def initialNewArms (cs : Cases) : HashMap Decision (Array CodeDecl) := Id.run do
+  let mut map := mkHashMap (cs.alts.size + 1)
+  map := map.insert .dont #[]
+  cs.alts.foldr (init := map) (fun val acc => acc.insert (.ofAlt val) #[])
```
leodemoura, line 151: Suggestion
```lean
cs.alts.foldr (init := map) fun val acc => acc.insert (.ofAlt val) #[]
```

leodemoura (commented):
The code is nicely organized. I have only a few comments regarding destructive updates on arrays. The arrays are being copied is a few places. I think the simplest solution is to use `List CodeDecl` in both places.

src/Lean/Compiler/LCNF/FloatLetIn.lean
```diff
@@ -0,0 +1,311 @@
...
+    if (← get).decision.contains fvar then
+      modify fun s => { s with decision := s.decision.insert fvar .dont }
+
+/--
+Will:
+- put `decl` into the arm it is marked to be moved into
+- for any variables that might occur in `decl` and are of interest:
+  - if they are already meant to be floated into the same arm or not at all leave them untouched:
+    ```
+    let x := ...
```
leodemoura, line 184: Nice docstring :)

leodemoura (approved):

leodemoura:
Could you please rebase? There is a tiny conflict at `Passes.lean`.

hargoniX:
The version that is pushed now is failing a single new test although I'm not yet sure why that is the case but commenting out the FloatLetIn pass does help so its definitely this PRs fault.

leodemoura:
It required an `update-stage0`. Happy to explain why next time we meet :)

PR #1710, opened 2022-10-09
fix: improve fuzzy-matching heuristics

### LINKED ISSUE #1546: Unexpected workspace symbol search results
Gonna use this issue to collect a few unexpected results I've encountered
* searching for `smap` or `SMap` does not find the actual type

rish987:
Closes #1546. There were no bugs with the current implementation as far as I could tell, but I noticed three areas that could be improved here:
1. Penalties should really only apply to the beginning of consecutive runs, and they shouldn't accumulate across namespaces. For example, if we had `Lean.AVeryLongNamespace.AnotherVeryLongNamespace.SMap` every miss would incur an additional penalty that would overwhelmingly negate the merit of a perfect match with the main identifier and give it no chance to show up in the top 100 results. So, rather than accumulating penalties for every miss, we apply a penalty once at the start of every consecutive run to indicate how "awkward" of a place to start a match this is. This awkwardness factor increases within a namespace but is reset at the beginning of every new namespace. However, we still do add a constant amount of penalty for every namespace so that we prefer less nested namespaces.
2. Matches in the main identifier (after all of the namespaces) should get a bonus.
3. Consecutive runs should be valued a lot more. What I've done is make a bonus for every match that is proportional to the length of the current run. So longer matches will get exponentially higher scores than ones that are more split up, which I think is better than a constant difference that could more easily have been negated.

LMK what you think of the above heuristics. After some (very brief) testing it looks like it's working great, but if someone could test-drive this and give me feedback it would be much appreciated!

### DIFF
--- a/src/Lean/Data/FuzzyMatching.lean
+++ b/src/Lean/Data/FuzzyMatching.lean
@@ -93,7 +93,7 @@ algorithm uses different scores for the last operation (miss/match). This is
 necessary to give consecutive character matches a bonus. -/
 private def fuzzyMatchCore (pattern word : String) (patternRoles wordRoles : Array CharRole) : Option Int := Id.run do
   /- Flattened array where the value at index (i, j, k) represents the best possible score of a fuzzy match
-  between the substrings pattern[:i+1] and [:j+1] assuming that pattern[i] misses at index word[j] (k = 0, i.e.
+  between the substrings pattern[:i+1] and word[:j+1] assuming that pattern[i] misses at word[j] (k = 0, i.e.
   it was matched earlier), or matches at word[j] (k = 1). A value of `none` corresponds to a score of -∞, and is used
   where no such match/miss is possible or for unneeded parts of the table. -/
   let mut result : Array (Option Int) := Array.mkArray (pattern.length * word.length * 2) none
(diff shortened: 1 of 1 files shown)


src/Lean/Data/FuzzyMatching.lean
```diff
@@ -92,64 +92,90 @@ In addition to the current characters in the pattern and the word, the
...
+      penaltyNs := penaltyNs + 1
+      lastSepIdx := wordIdx
+    penaltySkip := penaltySkip + skipPenalty (wordRoles.get! wordIdx) (wordIdx == 0)
+    startPenalties := startPenalties.set! wordIdx $ penaltySkip + penaltyNs
 
   -- TODO: the following code is assuming all characters are ASCII
   for patternIdx in [:pattern.length] do
-    let patternComplete := patternIdx == pattern.length - 1
-
+    -- for the IH, it's only necessary to populate a range of length `word.length - pattern.length` at each index
```
Kha, line 120: What's the IH?

Kha:
@rish987 Could you rebase against master for benchmarking? Locally it looks like there is some overhead up to 25%, unfortunately.
```
13:0: 
160678 decls
104155 matches

13:0: 
interpretation of _eval took 1.46s

14:0: 
160678 decls
143691 matches

14:0: 
interpretation of _eval took 3.18s

15:0: 
160678 decls
9890 matches

15:0: 
interpretation of _eval took 1.05s

16:0: 
160678 decls
334 matches

16:0: 
interpretation of _eval took 546ms
```

Kha:
!bench

Kha:
On the other hand, this is still fast enough that we can test-drive it and then think about further behavior/performance refinements. The improvements should be worth it.

rish987:
At first glance I'm a bit puzzled that it's slower, since I actually made the `result` table that we have to populate with scores *smaller*, so there might be some expensive operation somewhere that I've introduced. Or maybe it's just the overhead of the new `runLengths` table that we're populating in parallel to `result`? I'll investigate later today.

PR #1713, opened 2022-10-10
doc: add question mark to LEAN_IDENT_RE

lovettchris:
See https://leanprover.github.io/lean4/doc/functions.html

Without this fix:
image

The question mark is highlighted in a way that sends a confusing message to the user that the question mark is NOT part of the name - but it is just part of the name.

With this fix:

image

### DIFF
--- a/doc/highlight.js
+++ b/doc/highlight.js
@@ -1143,7 +1143,7 @@ hljs.registerLanguage("lean", function(hljs) {
       'sorry admit',
   };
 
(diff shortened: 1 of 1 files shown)


digama0:
You added `?` as an *initial* identifier character though, which is not correct: e.g. `?a` is not an identifier (it is a named metavariable). `?` is legal as a trailing identifier character.

lovettchris:
Woops good catch - I pushed a fix.

PR #1715, opened 2022-10-11
fix: watch out for heartbeat wraparound

lovettchris:
Fix for possible heartbeat wrap-around in very long running processes (e.g LSP perhaps?)  See discussion: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/heartbeats.2C.20GMP.20vs.20no.20GMP

Computing this with the following program the wrap around in this case with single core program would happen in about 4146 years on my particular computer, but I don't know what type of Lean code causes the heartbeat to increase at the highest speed, seems to be incremented in memory allocation, so perhaps something allocating large data structures over and over would hit the wraparound more quickly.  But this loop allocates a Nat each time around, and since it is a very tight loop, perhaps this is also close to the worst case? Not sure.  Then of course, there is future hardware and multi-core processing to worry about. Given multi-core apps, it is a bit of a worry that `g_heap->m_heartbeat++;` is not atomic, but perhaps the numbers won't get mangled, they'll only get clobbered... except on a 32bit machine of course, but we probably don't care about those.
```lean

def watchHeartbeats : IO Unit := do
  let mut p ← IO.getNumHeartbeats
  let mut c := 0
  repeat
    let i ← IO.getNumHeartbeats
    if i < p then
      IO.println "found wrap around!"
      return
    p := i
    c := c + 1
    if c == 1000000 then
      IO.println s!"Heartbeats={i}"
      c := 0

def main (args: List String): IO Unit :=
  timeit "watchHeartbeats: " watchHeartbeats

```

### DIFF
--- a/src/Init/System/IO.lean
+++ b/src/Init/System/IO.lean
@@ -215,6 +215,7 @@ local macro "nonempty_list" : tactic =>
 
 /-- Helper method for implementing "deterministic" timeouts. It is the number of "small" memory allocations performed by the current execution thread. -/
 @[extern "lean_io_get_num_heartbeats"] opaque getNumHeartbeats : BaseIO Nat
+@[extern "lean_io_get_heartbeat_max"] opaque getHeartbeatMax : BaseIO Nat
 
 inductive FS.Mode where
   | read | write | readWrite | append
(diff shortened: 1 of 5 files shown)


leodemoura:
There is no wrap-around issue being reported in the linked thread.

leodemoura:
`g_heap` is thread local. There is no issue here. I am closing the PR

PR #1716, opened 2022-10-11
feat: `import` command stub

digama0:
This implements `import` at command position to give an error saying that imports should only go at the top of the file (they are not real commands). It also has an `#exit`-like effect: elaboration of the rest of the file is halted because it's likely to be nonsense.

To make this work, I also had to change the semantics of exit commands such that elaboration halts *after* running the command elab, rather than before it. So now the `#exit` command runs as well, and I implemented it to give a warning `using 'exit' to interrupt Lean` similar to lean 3, which is a change of behavior. I can change it back to the lean 4 behavior of silently exiting if people prefer that.

### DIFF
--- a/src/Lean/Parser/Module.lean
+++ b/src/Lean/Parser/Module.lean
@@ -59,7 +59,7 @@ private def mkEOI (pos : String.Pos) : Syntax :=
 def isEOI (s : Syntax) : Bool :=
   s.isOfKind `Lean.Parser.Module.eoi
 
-def isExitCommand (s : Syntax) : Bool :=
+def isTerminalCommand (s : Syntax) : Bool :=
   s.isOfKind ``Command.exit || s.isOfKind ``Command.import
 
 private def consumeInput (c : ParserContext) (pos : String.Pos) : String.Pos :=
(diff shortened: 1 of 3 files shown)


Kha:
> So now the `#exit` command runs as well, and I implemented it to give a warning `using 'exit' to interrupt Lean` similar to lean 3, which is a change of behavior.

Yes! :tada:

src/Lean/Parser/Module.lean
```diff
@@ -60,7 +60,7 @@ def isEOI (s : Syntax) : Bool :=
   s.isOfKind `Lean.Parser.Module.eoi
 
 def isExitCommand (s : Syntax) : Bool :=
```
Kha, line 62: Should we rename it to e.g. `isTerminalCommand`?

Kha (approved):

gebner:
> This implements `import` at command position to give an error saying that imports should only go at the top of the file (they are not real commands). It also has an `#exit`-like effect: elaboration of the rest of the file is halted because it's likely to be nonsense.

Should we also do the same if the initial `import` fails?  Right now we just continue elaborating the file, but with *no* modules imported (not even `sorry` works because `sorryAx` isn't defined).

Kha:
Oh, that is an unexpected divergence between cmdline and server mode. We should definitely do that.

PR #1717, opened 2022-10-11
fix highlight of `deriving instance`

lovettchris:
with this fix:

image

without this fix the `deriving instance` bleeds all the way to the closing `where`.

image

### DIFF
--- a/doc/highlight.js
+++ b/doc/highlight.js
@@ -1167,7 +1167,7 @@ hljs.registerLanguage("lean", function(hljs) {
 
   var LEAN_DEFINITION =	{
     className: 'theorem',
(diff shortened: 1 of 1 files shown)


Kha:
Unclean history again

lovettchris:
are you talking about the 4 commits?  doesn't squash merge take care of squashing the 4 commits you see there?

Kha:
Oh I see, the conflict message is specific to rebase... I wasn't aware of that because most other contributors rebase their branches

Kha:
And because of that I now merged it using the wrong commit message...

leodemoura:
I will make a destructive update.

PR #1720, opened 2022-10-12
chore: replace let by have in stx matches

gebner:
Right now we still have `x : Syntax := { raw := __discr }` in the local context.  This PR fixes that by using `have` instead of `let` in the syntax matcher.  I have checked the generated code and it seems to be identical to the `let`-version.

See also the discussion at https://github.com/leanprover/lean4/pull/1692#issuecomment-1275802847


gebner:
!bench

Kha (approved):

gebner:
> What was the nature of the regression? Timeout/crash/type error/...

Mainly compilation time I believe (see benchmark results).  I did check a small example and verified that we generate identical code with `let`, `have`, and with `have` plus the dedicated `let_fun` support.

PR #1721, opened 2022-10-12
perf: improve `Unhygienic.run` code

gebner:
We sometimes write ``Unhygienic.run `(this + $x)`` for syntax quotations where (macro scope) hygiene is not necessary or desired.  However, I've just noticed that the code we currently generate for this is decidedly suboptimal: it allocates a closure, and a temporary pair.

Not sure if there are any takeaways for the current compiler rewrite:
 - Always specialize HO functions?  (would help with `Unhygienic.run` but not with `mkInfoFromRefPos`)
 - Inline functions that apply arguments which are statically known to be closures (i.e., where we can eliminate a closure call after inlining)?  (would help with `Unhygienic.run`)
 - The (specialized) `mkInfoFromRefPos` function has the form `fun a b => (..., b)` (that is, the argument is always returned as the second field).  We probably have a lot of monadic functions that are "read-only" operations.  Should we do some static analysis to detect these situations?  This could also enable further optimization opportunities at the call site: common subexpression elimination, omitting reference counting operations, dead code elimination (if the function doesn't even read the state), etc.

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -3980,6 +3980,7 @@ class MonadQuotation (m : Type → Type) extends MonadRef m where
 export MonadQuotation (getCurrMacroScope getMainModule withFreshMacroScope)
 
 /-- Construct a synthetic `SourceInfo` from the `ref` in the monad state. -/
+@[inline]
 def MonadRef.mkInfoFromRefPos [Monad m] [MonadRef m] : m SourceInfo :=
   return SourceInfo.fromRef (← getRef)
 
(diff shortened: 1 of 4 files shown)


gebner:
!bench

gebner:
The qsort/binary_trees/unionfind results must be completely bogus; I don't think there's any way this change could affect them.

Maybe interestingly, the C code lines went down by 0.9%.  The libleanshared size also went down by 0.8%.

leodemoura (approved):

PR #1723, opened 2022-10-13
test: benchmark workspace symbols search

Kha:
This is dependent on the number of constants in core, but at least we rarely have a *significant* change in that number

### DIFF
--- a/tests/bench/speedcenter.exec.velcom.yaml
+++ b/tests/bench/speedcenter.exec.velcom.yaml
@@ -160,3 +160,9 @@
     cmd: ./unionfind.lean.out 3000000
   build_config:
     cmd: ./compile.sh unionfind.lean
+- attributes:
+    description: workspaceSymbols
+    tags: [fast, suite]
+  run_config:
+    <<: *time
+    cmd: lean workspaceSymbols.lean
--- a/tests/bench/workspaceSymbols.lean
+++ b/tests/bench/workspaceSymbols.lean
@@ -1,0 +1,16 @@
+import Lean
+open Lean Elab
+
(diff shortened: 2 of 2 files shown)


Kha:
!bench

Kha:
!bench

gebner:
From what I rememember, performance was pretty variable depending on the query length: https://github.com/leanprover/lean4/pull/1023#issuecomment-1055837461  Should we run a few more queries (of different length) here?

Kha:
Inside the same benchmark? Unfortunately it's a bit of a pain to set up separate benchmarks, at least without just copying the whole file... I already resorted to the interpreter so I don't have to worry about `-rdynamic`

gebner:
> Inside the same benchmark? 

Yes, I think it's more representative even if we only measure the sum.

BTW, I think the actual symbol search does a lot of preprocessing and filters out many declarations.  But that seems to be spread out over the request handler / ilean code.

Kha:
```
#eval bench "L"
#eval bench "Lean."
#eval bench "Lean.Expr"
#eval bench "Lean.Expr.const"
```
```
13:0: 
160670 decls
104159 matches

13:0: 
interpretation of _eval took 1.21s

14:0: 
160670 decls
143590 matches

14:0: 
interpretation of _eval took 2.43s

15:0: 
160670 decls
11988 matches

15:0: 
interpretation of _eval took 884ms

16:0: 
160670 decls
522 matches

16:0: 
interpretation of _eval took 529ms
```

Kha:
I was confused at first how a longer query can result in more matches, but I guess it's more matches *above the threshold*

PR #1729, opened 2022-10-14
feat: function coercions with unification

### LINKED ISSUE #1725: Function coercion does not trigger if types require unification
As reported on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Regression.20in.20function.20coercion.20inference/near/303901188).  This used to be supported in Lean 3 (community edition), and is used pretty heavily in mathlib:

```lean
structure Equiv (α : Sort _) (β : Sort _) :=
(toFun    : α → β)
(invFun   : β → α)

local infix:25 " ≃ " => Equiv

variable {α β γ : Sort _}

instance : CoeFun (α ≃ β) (λ _ => α → β) := ⟨Equiv.toFun⟩

def Equiv.symm (e : α ≃ β) : β ≃ α := ⟨e.invFun, e.toFun⟩

protected def Equiv.trans (e₁ : α ≃ β) (e₂ : β ≃ γ) : α ≃ γ :=
  ⟨e₂ ∘ e₁, e₁.symm ∘ e₂.symm⟩ -- both fields raise an error
```

Slightly simpler repro:
```lean
structure Fun (α β : Type) : Type where
  toFun : α → β

instance : CoeFun (Fun α β) (fun _ => α → β) where
  coe := Fun.toFun

example (f : Fun α β) : (γ δ : Type) × (γ → δ) :=
  ⟨_, _, f⟩
```

The issue is that we try to synthesize the instance `CoeT (Fun α β) f (?m → ?n)`, but that does not find anything since we can't assign `?m` or `?n` during TC search (and the last argument of `CoeT` is not an out param).  In Lean 3, we used to check for `has_coe_to_fn` instances separately.

gebner:
This PR moves all coercion functions into `Lean.Meta.Coe`.  The delayed coercions in the elaborator now call the `coerce? : Expr → Expr → MetaM (LOption Expr)` function directly.  In particular this direct call can now unify the out-param of the function coercions.

Fixes #1725.

### DIFF
--- a/src/Lean/Meta/Coe.lean
+++ b/src/Lean/Meta/Coe.lean
@@ -162,7 +162,7 @@ def coerceMonadLift? (e expectedType : Expr) : MetaM (Option Expr) := do
     try
       -- Construct lift from `m` to `n`
       let monadLiftType ← mkAppM ``MonadLiftT #[m, n]
-      let .some monadLiftVal ← trySynthInstanceForCoe monadLiftType | return none
+      let .some monadLiftVal ← trySynthInstance monadLiftType | return none
       let u_1 ← getDecLevel α
       let u_2 ← getDecLevel eType
       let u_3 ← getDecLevel expectedType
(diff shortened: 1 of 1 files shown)


leodemoura:
I didn't review in detail, but it looks like a very good change to me. Let's merge it.

gebner:
It looks like I broke an edge case in monad coercions that only occurs in two lake files. Let me investigate.

gebner:
What happened was the following: we have a nice `maxCoeSize` option ostensibly applying to all coercions.  However, we completely ignored that option for monad coercions before this PR.

I've now changed it back so that we ignore `maxCoeSize` for monad coercions.  Generally, I feel that `maxCoeSize` could lead to very confusing behavior and I'd rather remove it entirely.  The failure mode of `maxCoeSize`---reporting a type mismatch error instead of complaining about too complicated coercions---is in my opinion not helpful for debugging.  I wouldn't have made the connection between the error and the option myself if I hadn't seen the option yesterday.

gebner:
A stage0 update is not strictly necessary, but it will prevent surprising build errors in stage2 due to potential changes in coercion handling.

leodemoura:
> Generally, I feel that maxCoeSize could lead to very confusing behavior and I'd rather remove it entirely.

The suggestion makes sense. It is also related to another issue you raised recently: the mechanism used to perform coercion resolution. Let's discuss this issue in the next dev-meeting. 

> The failure mode of maxCoeSize---reporting a type mismatch error instead of complaining about too complicated coercions---is in my opinion not helpful for debugging. 

I agree. I think improving the error message is the least intrusive improvement at this point.

PR #1732, opened 2022-10-14
chore: CI: exclude "full program" tests from debug/sanitized builds

Kha:
These take ~1min *each*, and so are responsible for a big chunk of the used time. In fact, the only reason the sanitized build is faster than debug is that we already exlcluded the laketests from it because it (used to?) triggers a segfault in them that we never debugged.


gebner:
Maybe exclude `leanpkgtest` as well?

PR #1736, opened 2022-10-15
fix: 'AND' escaping for 'if(LLVM) ...' in 'src/shell'

bollu:
Recall that CMake semantics are defined entirely
via string expansion. Without the string escaping, CMake tries to evaluate:

```
if(${STATIC} AND NOT ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin"))
→ if( AND NOT ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin"))
→ CMake Error at shell/CMakeLists.txt:19 (if):
    if given arguments:
      "AND" "(" "NOT" "(" "Linux" "MATCHES" "Darwin" ")" ")"
```

With the escaping, CMake evaluates as:

```
if("${STATIC}" AND NOT ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin"))
→ if("" AND NOT ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin"))
→ if(False AND NOT ("${CMAKE_SYSTEM_NAME}" MATCHES "Darwin"))
→ if(False)
→ ...
```

The key difference is '${STATIC}' evaluates to something like
'null', while '"${STATIC}"' evaluates to 'empty string/false'.


bollu:
This was discovered when cleaning up #1687 to combine the `LLVM` and `LLVM_BACKEND` flags.

Kha:
`STATIC` was removed in 5f4b1b1d44a55c542754a2e5e95cc8651b9c0ff2

Issue #1737, opened 2022-10-15
Cannot derive Hashable for (indirect) mutually recursive inductives


ydewit:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

The Hashable deriving handler is not able to derive the instance for inductives that are mutually recursive when the recursion is not direct.

For instance, consider the following example where Term is indirectly mutually-recursive through the List type former:
```
inductive T
| t (args: List T)
deriving Hashable  -- ERROR: failed to synthesize instance Hashable (List T)
```
Here the Hashable deriving handler will fail with the following error: `failed to synthesize instance Hashable (List Term)`

Note that when the recursion is direct, deriving Hashable works fine.
```
inductive T
| t (arg : T)
deriving Hashable  -- Works fine
```

### Steps to Reproduce

Try to compile the above example.

**Expected behavior:** Hashable should be able to derive without any errors.

**Actual behavior:** Compilation error: `failed to synthesize instance Hashable (List Term)`

**Reproduces how often:** always.

### Versions

Lean (version 4.0.0, commit 1eda0fd7349d, Release)
Linux PopOS

ydewit:
I am preparing a PR for this.

PR #1738, opened 2022-10-15
fix: add local Hashable instance

ydewit:
When inductives are (indirectly) mutually recursive, say `inductive T | t (args: List T)` instead of `inductive T | t (arg : T)`, Lean would fail to find an instance of Hashable for `T` because it was not yet defined.

This commit makes sure that the deriving handler adds a needed Hashable T instance to the local scope so that Hashable.hash can be resolved recursively.


leodemoura:
Could you please add tests for the issue being fixed?
It would also be nice to have comments explaining the change.

digama0:
you should also put "fixes #1737" in the PR description or commit comment if it is intended to address/close an issue.

ydewit:
> you should also put "fixes #1737" in the PR description or commit comment if it is intended to address/close an issue.

Sure, will do! I actually took a quick peek at the recent commits and followed what I saw there: `fix: .. (#????)`.

Kha:
The reference in the commit title is a PR number from squashing, not the issue number

PR #1739, opened 2022-10-16
fix: environment leak on Windows

### LINKED ISSUE #1282: Environment leaks between spawned processes (on Windows)
### Prerequisites

### Description

On Windows, spawning a process using `IO.Process.spawn` with one environment in one thread will affect the environment of another spawned process in another thread. This is a major problem for Lake which often wishes to spawn processes in different build threads with different environments.

This is due to this section of code (which specifically states that its manner of setting environment variables is thread-unsafe):

https://github.com/leanprover/lean4/blob/2061c9bbea847afd6cd301381c5fb239fb21062f/src/runtime/process.cpp#L163-L176

Based on the comment, I assume this may be an issue for @gebner?

### Steps to Reproduce

```lean
def printenv (var : String) (env : Array (String × Option String) := #[]) : IO String :=
  IO.Process.output {cmd := "printenv", args :=  #[var], env} <&> (·.stdout)

def singleTest : IO PUnit := do
  IO.println s!"Initial: {repr (← printenv "LEAN_CC").trim}"

def jointSyncTest : IO PUnit := do
  let a ← printenv "LEAN_CC" #[("LEAN_CC", "cc")]
  let b ← printenv "LEAN_CC" #[]
  IO.print s!"1st: {repr a.trim}; "
  IO.println s!"2nd: {repr b.trim}"

def jointAsyncTest : IO PUnit := do
  let t1 ← IO.asTask <| printenv "LEAN_CC" #[("LEAN_CC", "cc")]
  let t2 ← IO.asTask <| printenv "LEAN_CC" #[]
  IO.print s!"Task 1: {repr (← IO.ofExcept (← IO.wait t1)).trim}; "
  IO.println s!"Task 2: {repr (← IO.ofExcept (← IO.wait t2)).trim}"

#eval singleTest -- Initial: ""
#eval jointSyncTest -- 1st: "cc"; 2nd: ""
#eval jointAsyncTest -- Task 1: "cc"; Task 2: "cc"
```

**Expected behavior:**

Environment does not leak between separately spawned process (even those spawned in parallel).

**Actual behavior:**

Environment leaks between processes spawned in parallel.

**Reproduces how often:**

Always (on Windows).

### Versions

Windows 20H2
`Lean (version 4.0.0-nightly-2022-07-03, commit 7326c817d22e, Release)`

### Additional Information

This bug can exacerbate the issue in #1281. If one process has `LEAN_CC` set and `leanc` is run in separate, parallel process, `leanc` breaks. I discovered this when attempting to get [Alloy](https://github.com/tydeu/lean4-alloy) to work with the new build features of Lake. The environment set for the `alloy` executable (which included a set `LEAN_CC`) would break parallel runs of `leanc`.

awson:
This fixes an old todo, where spawning a child process on Windows modified the environment variables of the parent.

Fixes #1282

### DIFF
--- a/src/runtime/process.cpp
+++ b/src/runtime/process.cpp
@@ -160,19 +160,40 @@ static obj_res spawn(string_ref const & proc_name, array_ref<string_ref> const &
     siStartInfo.hStdError = child_stderr;
     siStartInfo.dwFlags |= STARTF_USESTDHANDLES;
 
-    // TODO(gabriel): this is thread-unsafe
-    std::unordered_map<std::string, optional<std::string>> old_env_vars;
-    for (auto & entry : env) {
-        optional<std::string> old;
-        if (auto old_val = getenv(entry.fst().data()))
-            old = std::string(old_val);
-        old_env_vars[entry.fst().to_std_string()] = old;
+    std::unique_ptr<char[]> new_env(nullptr);
 
-        if (entry.snd()) {
-            SetEnvironmentVariable(entry.fst().data(), entry.snd().get()->data());
-        } else {
-            SetEnvironmentVariable(entry.fst().data(), nullptr);
+    if (env.size()) {
+        auto * esp = GetEnvironmentStrings();
+
+        std::unordered_map<std::string, std::string> new_env_vars; // C++17 gives us no-copy std::string_view for this, much better!
+        for (auto & entry : env) {
+            new_env_vars[entry.fst().data()] = entry.snd() ? entry.snd().get()->data() : std::string{};
         }
+        static constexpr auto env_buf_size = 0x7fff; // according to MS docs 0x7fff is the max total size of env block
+        new_env = std::make_unique<char[]>(env_buf_size);
+        // First copy old evars not in new evars.
+        char *new_envp = new_env.get(), *key_begin = esp;
+        while (*key_begin) {
+            char *key_end = strchr(key_begin, '=');
+            char *entry_end = key_end + strlen(key_end);
+            if (!new_env_vars.count({key_begin, key_end})) {
+                new_envp = std::copy(key_begin, entry_end + 1, new_envp);
+            }
+            key_begin = entry_end + 1;
+        }
+        // Then copy new evars if nonempty
+        for(const auto & ev : new_env_vars) {
+          if (ev.second.empty()) continue;
+          // Check if the destination buffer has enough room.
+          if (new_envp + ev.first.length() + 1 + ev.second.length() + 1 > new_env.get() + env_buf_size - 1) break;
+          new_envp = std::copy(ev.first.cbegin(), ev.first.cend(), new_envp);
+          *new_envp++ = '=';
+          new_envp = std::copy(ev.second.cbegin(), ev.second.cend(), new_envp);
+          *new_envp++ = '\0';
+        }
+        *new_envp = '\0';
+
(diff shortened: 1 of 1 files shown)


src/runtime/process.cpp
```diff
@@ -160,19 +160,41 @@ static obj_res spawn(string_ref const & proc_name, array_ref<string_ref> const &
...
-        if (auto old_val = getenv(entry.fst().data()))
-            old = std::string(old_val);
-        old_env_vars[entry.fst().to_std_string()] = old;
-
-        if (entry.snd()) {
-            SetEnvironmentVariable(entry.fst().data(), entry.snd().get()->data());
-        } else {
-            SetEnvironmentVariable(entry.fst().data(), nullptr);
+    char * new_env = nullptr, * esp = nullptr;
+    std::unique_ptr<char> new_env_buf;
```
gebner, line 164:
```suggestion
    std::vector<char> new_env_buf;
```
And then pass `new_env_buf.data()` to `CreateProcess`.

src/runtime/process.cpp
```diff
@@ -160,19 +160,41 @@ static obj_res spawn(string_ref const & proc_name, array_ref<string_ref> const &
...
+        while (*ep) {
+            char *eqp = ep = strchr(ep, '=');
+            ep += strlen(ep);
+            if (!new_env_vars.count({oep, static_cast<size_t>(eqp - oep)})) {
+                nep = std::copy(oep, ep, nep);
+                *nep++ = '\0';
+            }
+            ep++;
+            oep = ep;
+        }
```
gebner, line 186:
```suggestion
        for (char *entryp = esp; *ep;) {
            char *equalsp = strchr(entryp, '=');
            size_t entry_len = strlen(entryp);
            if (!new_env_vars.count({entryp, static_cast<size_t>(equalsp - entryp)})) {
                std::copy(entryp, entryp + entry_len + 1, std::back_inserter(new_env));
            }
            entryp += entry_len;
        }
```
Please try to use (easily) decipherable variable names, scope them as locally as possible, and avoid mutating them during the loop body.

gebner: Please also declare the `esp` inside the if-block (and free it here as well).

gebner (commented):
Thanks for fixing this really old todo!

Please squash the commits!

Please remove the `(#1282)`  from the commit message and PR title.  Instead, please write `Fixes #1282` in the PR message!

gebner:
@awson Is there a reason you decided to stick with the statically-sized buffer instead of the dynamically-growing `std::vector<char>` version?

Is `0x8000` some magic limit on Windows?  If so, we should document that!

awson:
> @awson Is there a reason you decided to stick with the statically-sized buffer instead of the dynamically-growing `std::vector<char>` version?
> 
> Is `0x8000` some magic limit on Windows? If so, we should document that!

Exactly. MS documentation says "The total size of the environment block for a process may not exceed 32,767 characters".

gebner:
> Exactly. MS documentation says "The total size of the environment block for a process may not exceed 32,767 characters".

Good to know!  Can you put that in a comment?  Then the PR is good to go!

awson:
That said, my interpretation have always been that 32767 is the size of the block *without* the extra null char, but after reading MS docs here and there I see them using "total size" wording, hence we, perhaps, need to use 0x7fff here.

PR #1744, opened 2022-10-17
Make tokens in `<|>` relevant to syntax match

### LINKED ISSUE #1275: Antiquotation types silently ignore tokens
```lean
import Lean open Lean
@[defaultInstance] instance : MonadQuotation Elab.TermElabM := inferInstance

syntax "👉" (ident <|> "_") : term
#eval `(👉 $_)
```

The last underscore has type ``TSyntax `ident``, which suggests that only identifiers are valid to insert here.

This just majorly confused me in leanprover-community/mathport#152.

To clarify: I would expect a type like ``TSyntax [`ident, `token.«_»]``.

### Versions

Nightly from July 2, as well as the relaxed antiquotations branch.

Kha:
Implements https://github.com/leanprover/lean4/issues/1275#issuecomment-1172939175

Fixes #1275

### DIFF
--- a/src/Init/NotationExtra.lean
+++ b/src/Init/NotationExtra.lean
@@ -61,10 +61,10 @@ def expandBrackedBinders (combinatorDeclName : Name) (bracketedExplicitBinders :
   let combinator := mkIdentFrom (← getRef) combinatorDeclName
   expandBrackedBindersAux combinator #[bracketedExplicitBinders] body
 
-syntax unifConstraint := term (" =?= " <|> " ≟ ") term
+syntax unifConstraint := term patternIgnore(" =?= " <|> " ≟ ") term
 syntax unifConstraintElem := colGe unifConstraint ", "?
 
-syntax (docComment)? attrKind "unif_hint " (ident)? bracketedBinder* " where " withPosition(unifConstraintElem*) ("|-" <|> "⊢ ") unifConstraint : command
+syntax (docComment)? attrKind "unif_hint " (ident)? bracketedBinder* " where " withPosition(unifConstraintElem*) patternIgnore("|-" <|> "⊢ ") unifConstraint : command
 
 macro_rules
   | `($[$doc?:docComment]? $kind:attrKind unif_hint $(n)? $bs* where $[$cs₁ ≟ $cs₂]* |- $t₁ ≟ $t₂) => do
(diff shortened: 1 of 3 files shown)


Kha:
I suppose `&"foo" <|> ...` should trigger a token match just as well. Which begs the question of how to handle
```
syntax discharger := atomic(" (" (&"discharger" <|> &"disch")) " := " tacticSeq ")"
```

Kha:
The simplest solution would be to define
```
syntax dischTk := &"discharger" <|> &"disch"
```
and then use `$_:dischTk`. It would basically be the reverse of the previous situation: introduce intermediary `syntax` when you do *not* want to match the token.

gebner:
I think for discharger, we can just remove the `disch` variant.  Generally, adding an extra syntax node sounds good.

Kha:
I'm still unsure what to do about something like
```
syntax unifConstraint := term (" =?= " <|> " ≟ ") term
```
In built-in syntax this would be `unicodeSymbol`, but we can't make that a parser alias as it takes two strings, not parsers. By the above we could do
```
syntax unifTk := " =?= " <|> " ≟ "
```
but then we would have to write `$_:unifTk` in patterns instead. Do we want to keep the pattern irrelevance in these cases? Then we could also introduce a generic parser combinator, as I think was proposed before, like
```
syntax unifTk := nomatch (" =?= " <|> " ≟ ")
```
except maybe with a better name.

gebner:
The one-size-fits-all solution for when ParserDescr is not expressive enough is to make a parser alias.  I think this also depends on (y)our long-term plans: do we want to get rid of `ParserDescr` or get rid of `Parser`?  As it stands, we're running into `ParserDescr` limitations left and right (e.g. we can't expose `sepByIndent` either because it takes a string and a parser, and I'm not even getting into optional arguments).

We've discussed the `nomatch` before, such a feature would also be great for `sepByIndent`, etc.  Here the `nomatch` should be outside of the `unifTk` though, we don't want to check the kind of the `unifTk` node.

src/Init/NotationExtra.lean
```diff
@@ -337,14 +337,15 @@ macro_rules
 
 section
 open Lean.Parser.Tactic
+syntax cdotTk := patternIgnore("·" <|> ".")
```
Kha, line 340: This is the only case where I had to introduce a new syntax abbrev, since a token quotation inside `patternIgnore` obviously can't work anymore. And yes, I went with a less "cute" name than `nomatch` after all.

Issue #1745, opened 2022-10-17
`change e at h` and `change e1 with e2` exist but are unimplemented


digama0:
```lean
example (h : True) : True := by
  change id True at h -- tactic 'Lean.Parser.Tactic.change' has not been implemented
  change id True with True -- tactic 'Lean.Parser.Tactic.changeWith' has not been implemented
  with_reducible exact h
```
For the regular form `change e`, `show e` exists as an alternative, but `change e at h` and `change e1 with e2` have no alternatives. It may also be appropriate to remove these and move them to std4/mathlib4 instead of implementing them here.

digama0:
I did a census of other syntax missing macro/elab. Results:

* `change e`
* `change e1 with e2`
* `dbg_trace "foo"` (the tactic)
* `#resolve_name` (the command)

 I also found some anomalies where the tactic seems to work despite not having a macro/elab:
* `register_simp_attr` (this seems to be a bug in the `macro` command: although the [`macro` exists](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=Lean.Parser.Command.registerSimpAttr#src), when you use a `_root_` prefix it ends up macro expanding to `@[builtinMacro Lean.Meta._root_.Lean.Parser.Command.registerSimpAttr]` on the elab declaration. It still seems to work though...
* char/ident/num/scientific/str: these are builtins so I think they are handled differently
* namedPattern: I guess this one is deliberate, but you still get a weird error message when you `#check a@b`
* `Lean.Parser.Tactic.Conv.conv.quot`, I think this is an issue in syntax categories declared by `declare_syntax_cat` since the actual syntax generated by `` `(conv| skip)`` uses `conv.quot` instead. `rawStx.quot` seems to have a similar issue but I couldn't find a legal syntax to put in `` `(rawStx| ...)``.

leodemoura:
closed by https://github.com/leanprover/lean4/pull/3308

dylanede:
Is this actually fixed though? Seems like `change e1 with e2` (syntax rule `changeWith`) is still unimplemented.

PR #1746, opened 2022-10-17
chore: remove `#resolve_name`

digama0:
This appears to be a long-obsolete command whose elaborator was removed in c595eea.

Addresses part of #1745.


digama0:
Failure is due to network error, can someone restart it?

PR #1748, opened 2022-10-18
Add dynamic quotations for syntax categories

gebner:
```lean
declare_syntax_cat mycat
syntax "mystx" : mycat
#check `(mycat| mystx)
```

 - Collateral change: ``macro "foo" : cat => `(bar)`` no longer defaults the rhs quotation parser to `cat`.  (I think that's good because the magic easily broke if you added a `do` e.g.)
 - `macro "foo" : cat => return blah` now enforces that `blah` has type ``TSyntax `cat``.
 - Most of the specialized category quotation parsers (think `` `(attr| ...) ``) are gone.
 - I've also unified all the different parser name resolution functions.

See also https://leanprover.zulipchat.com/#narrow/stream/147302-lean4-maintainers/topic/Macro.20syntax/near/304524822

Future work:
 - Remove quotation parsers for `command` and `tactic`.  This is a bit tricky since the existing parsers are abused to parse multiple commands/tactics, resulting in invalid syntax (that breaks the pretty-printer).  I'd like to have a plan to handle that before doing further refactorings there.
 - Support parser aliases.  This is blocked by not being able to do IO in parsers at the moment.  (The aliases are stored in a global reference.)

### DIFF
--- a/doc/macro_overview.md
+++ b/doc/macro_overview.md
@@ -302,7 +302,7 @@ To see the desugared definition of the actual expansion, we can again use
 ```lean
 
 set_option trace.Elab.definition true in
-macro "exfalso" : tactic => `(apply False.elim)
+macro "exfalso" : tactic => `(tactic| apply False.elim)
 
 /-
 Results in the expansion:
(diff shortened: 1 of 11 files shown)


Kha (approved):

gebner:
There was an exciting stage3 failure here!  To fix it, I added another sanity check:
 - `macro "foo" : cat => return blah` now enforces that `blah` has type ``TSyntax `cat``.

cc @Kha

Kha:
Nice. I didn't want to do it before we had a solution for `macro_rules` as well, but I certainly don't mind.

Kha:
Feel free to remove the `changelog` label if you want tow write a more detailed entry btw

PR #1749, opened 2022-10-18
feat: `show` tactic (full version)

digama0:
This implements the TODO in the `show` tactic: `show e` will first try to change the goal to `e`, and if that doesn't work it will try to change one of the other open goals to `e` and select it. Note: This implementation allows the goal to change in a non-defeq way and will insert coercions. I plan to follow this up with an implementation of `change` which is only for strictly defeq changes, similar to `dsimp only`.

A thought: should the goal-selecting version of `show` be spelled `show e => tac` instead? It is a bit awkward that it acts like a move-to-front tactic but has no `case =>` form so you have to do `show e; . tac` to actually work on the selected goal.

### DIFF
--- a/src/Lean/Elab/Tactic/BuiltinTactic.lean
+++ b/src/Lean/Elab/Tactic/BuiltinTactic.lean
@@ -379,6 +379,31 @@ private def getCaseGoals (tag : TSyntax ``binderIdent) : TacticM (MVarId × List
     setGoals (acc.toList ++ (← getGoals))
   | _ => throwUnsupportedSyntax
 
+@[builtinTactic «show»] def evalShow : Tactic
+  | `(tactic| show $e) => do
+    let mvarId :: mvarIds ← getUnsolvedGoals | throwNoGoalsToBeSolved
+    let s ← saveState
+    try
+      mvarId.withContext do
+        let mvarIdNew ← mkFreshExprMVar (← Term.elabType e)
+        mvarId.assign (← Term.ensureHasType (← mvarId.getType) mvarIdNew "'show' tactic failed")
+        setGoals (mvarIdNew.mvarId! :: mvarIds)
+      return
+    catch err =>
+      let errState ← saveState
+      s.restore
+      for mvarId' in mvarIds do
+        try
+          mvarId'.withContext do
+            let mvarIdNew ← mkFreshExprMVar (← Term.elabType e)
+            mvarId'.assign (← Term.ensureHasType (← mvarId'.getType) mvarIdNew)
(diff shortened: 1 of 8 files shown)


src/Lean/Elab/Tactic/BuiltinTactic.lean
```diff
@@ -379,6 +379,31 @@ private def getCaseGoals (tag : TSyntax ``binderIdent) : TacticM (MVarId × List
...
   | _ => throwUnsupportedSyntax
 
+@[builtinTactic «show»] def evalShow : Tactic
+  | `(tactic| show $e) => do
+    let mvarId :: mvarIds ← getUnsolvedGoals | throwNoGoalsToBeSolved
+    let s ← saveState
+    try
+      mvarId.withContext do
+        let mvarIdNew ← mkFreshExprMVar (← Term.elabType e)
+        mvarId.assign (← Term.ensureHasType (← mvarId.getType) mvarIdNew "'show' tactic failed")
```
gebner, line 389: As you mentioned in the PR description, this will insert coercions if the types don't match.

I'm not sure if this such a good fit for a tactic that selects goals by type (as it is quite likely that it will just pick the first goal and insert a coercion).

gebner:
> should the goal-selecting version of show be spelled show e => tac instead?

👍

src/Lean/Elab/Tactic/BuiltinTactic.lean
digama0: The coercion is not a subgoal, so I don't see how this would be a real concern unless both the source and target types are sufficiently known that the coercion can be fully synthesized. At least, it did not appear to be an issue in my tests.

digama0: In that case, it sounds like it would suffice to call `synthesizeSyntheticMVarsNoPostponing` after the assignment, right?

gebner: That might be a good idea anyhow, just to resolve things like `show ⟨1,2⟩ = _`.

I'm still not convinced by the automatic coercions.  Do you have a use case for that?

digama0: What do you mean by "resolve" there? Will `show ⟨1,2⟩ = _` be rejected if we do that? (In particular, the `_` should not be an error nor should it produce a subgoal.)

As for automatic coercions, no it's not a big deal to lose them, but it would be more consistent with term mode `show`, which will perform coercions. (I would like tactics to behave the same as their term mode counterparts where reasonable.)

kim-em:

PR #1750, opened 2022-10-18
fix: prevent blacklisted symbols from appearing in `.ilean` files

### LINKED ISSUE #1510: Workspace symbol search returns internal definitions
The workspace symbol search (`ctrl-p #` in vscode) returns all kinds of internal definitions.  Example queries:
 - `_sunfold` returns smart unfolding lemmas
 - `_unsafe_rec` returns helper definitions used to compile recursive definitions

These should be filtered out.

Nightly from August 25.

rish987:
Closes #1510. I'm still trying to figure out how to test this properly, I'll ask about that at the meeting tomorrow. But at least when I run `lean file.lean -i file.ilean` it looks like the undesirable definitions are indeed being filtered out.

### DIFF
--- a/src/Lean/Server/Completion.lean
+++ b/src/Lean/Server/Completion.lean
@@ -26,14 +26,13 @@ builtin_initialize completionBlackListExt : TagDeclarationExtension ← mkTagDec
 def addToBlackList (env : Environment) (declName : Name) : Environment :=
   completionBlackListExt.tag env declName
 
-private def isBlackListed (declName : Name) : MetaM Bool := do
-  let env ← getEnv
-  (pure (declName.isInternal && !isPrivateName declName))
-  <||> (pure <| isAuxRecursor env declName)
-  <||> (pure <| isNoConfusion env declName)
-  <||> isRec declName
-  <||> (pure <| completionBlackListExt.isTagged env declName)
-  <||> isMatcher declName
+def isBlackListed (env : Environment) (declName : Name) : Bool :=
+  (declName.isInternal && !isPrivateName declName)
+  || isAuxRecursor env declName
+  || isNoConfusion env declName
+  || isRecCore env declName
+  || completionBlackListExt.isTagged env declName
+  || isMatcherCore env declName
 
 private partial def consumeImplicitPrefix (e : Expr) (k : Expr → MetaM α) : MetaM α := do
   match e with
@@ -245,7 +244,7 @@ private def idCompletionCore (ctx : ContextInfo) (id : Name) (hoverInfo : HoverI
   -- search for matches in the environment
   let env ← getEnv
   env.constants.forM fun declName c => do
-    unless (← isBlackListed declName) do
+    unless isBlackListed (← getEnv) declName do
       let matchUsingNamespace (ns : Name): M Bool := do
         if let some (label, score) ← matchDecl? ns id danglingDot declName then
           -- dbg_trace "matched with {id}, {declName}, {label}"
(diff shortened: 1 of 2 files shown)


src/Lean/Server/Completion.lean
```diff
@@ -26,14 +26,13 @@ builtin_initialize completionBlackListExt : TagDeclarationExtension ← mkTagDec
...
-  <||> isRec declName
-  <||> (pure <| completionBlackListExt.isTagged env declName)
-  <||> isMatcher declName
+def isBlackListed (env : Environment) (declName : Name) : Bool :=
+  (declName.isInternal && !isPrivateName declName)
+  || isAuxRecursor env declName
+  || isNoConfusion env declName
+  || isRecCore env declName
+  || completionBlackListExt.isTagged env declName
+  || isMatcherCore env declName
```
rish987, line 35: I was quite unsure about this change. I needed this to be outside of `MetaM` but I'm not certain about the extent to which Lean is now eagerly evaluating the components of this big chained or -- ideally it should stop after the first one that returns true (which I assume was certainly the case in the previous `MetaM` version).

src/Lean/Server/References.lean
```diff
@@ -130,13 +131,18 @@ def identOf : Info → Option (RefIdent × Bool)
   | Info.ofFieldInfo fi => some (RefIdent.const fi.projName, false)
   | _ => none
 
+open Completion in
```
rish987, line 134: Having to pull in `Completion` here felt a bit hackish to me, is there a good candidate parent file that I can migrate the `isBlacklisted` related functions to?

src/Lean/Server/Completion.lean
Kha: That's fine, `||` is lazy

src/Lean/Server/References.lean
```diff
@@ -130,13 +131,18 @@ def identOf : Info → Option (RefIdent × Bool)
...
 def findReferences (text : FileMap) (trees : Array InfoTree) : Array Reference := Id.run <| StateT.run' (s := #[]) do
   for tree in trees do
     tree.visitM' (postNode := fun ci info _ => do
       if let some (ident, isBinder) := identOf info then
-        if let some range := info.range? then
-          if info.stx.getHeadInfo matches .original .. then  -- we are not interested in canonical syntax here
-            modify (·.push { ident, range := range.toLspRange text, stx := info.stx, ci, info, isBinder }))
+        let blacklisted? := match ident with
+        | .const name => isBlackListed ci.env name
+        | _ => false
```
Kha, line 141: I would indent these lines

src/Lean/Server/References.lean
```diff
@@ -130,13 +131,18 @@ def identOf : Info → Option (RefIdent × Bool)
...
 
+open Completion in
 def findReferences (text : FileMap) (trees : Array InfoTree) : Array Reference := Id.run <| StateT.run' (s := #[]) do
   for tree in trees do
     tree.visitM' (postNode := fun ci info _ => do
       if let some (ident, isBinder) := identOf info then
-        if let some range := info.range? then
-          if info.stx.getHeadInfo matches .original .. then  -- we are not interested in canonical syntax here
-            modify (·.push { ident, range := range.toLspRange text, stx := info.stx, ci, info, isBinder }))
+        let blacklisted? := match ident with
```
Kha, line 139:
```suggestion
        let blacklisted := match ident with
```
It's not an `Option` or `Except`

src/Lean/Server/References.lean
Kha: When in doubt, use `Lean.Server.Utils` :)

Kha (changes requested):

Kha:
I'm not sure if this is the best way to solve the issue - it's not clear to me why the `_sunfold`s etc. are in the info tree to begin with, and with an original source span at that.
Note that the issue seems to be limited to `let rec`. @leodemoura Do you see where these info nodes could come from? Perhaps we can investigate in today's meeting.

leodemoura:
In the frontend meeting, I think we all agreed to close this PR. If there are no objections, let's close it.

PR #1753, opened 2022-10-19
fix: `hoverableInfoAt?` in presence of canonical syntax

Kha:


src/Lean/Server/InfoUtils.lean
```diff
@@ -171,6 +171,16 @@ partial def InfoTree.hoverableInfoAt? (t : InfoTree) (hoverPos : String.Pos) (in
       if ti.expr.isSyntheticSorry then
         return none
   return res?
+where containsHoverPos info :=
+  if info.contains hoverPos includeStop then
+    info.range?
+  -- `(· + ·)` expands to a `+` node that has a synthetic range because its first and last token are synthetic.
```
gebner, line 177: Is this an argument to store parser ranges for nodes as well?

The heuristic here seems a bit fragile to be honest.

Kha: Mmh, you mean we should store the entire span as original info in the `+` node? That could work, yeah.

Kha:
I've retracted the PR to the less questionable fix for now

src/Lean/Server/InfoUtils.lean
Kha: This would however be a pretty fundamental change for a pretty rare use case: modifying a node's children without recreating the node as part of a quotation. I think in most cases we're trying to move away from this pattern. `expandCDot?` really is the only case that comes to mind where we do a generic traversal over the syntax tree.

gebner: I agree that it's a bigger change.  It also has the advantage that it is consistent with how we store synthetic positions from macro expansions (which are put on the nodes).

> generic traversal over the syntax tree

We also need to do this for `(← ..)` and for `$(f a)`, and really for any "two-part" syntax.

> I think in most cases we're trying to move away from this pattern.

From what I can tell, there are quite a few places where we explicitly use this pattern to preserve separators, atom source infos, syntax variants (unicode/non-unicode, semicolon/newline), etc.

Kha: > We also need to do this for (← ..) and for $(f a), and really for any "two-part" syntax.

Ah, this is fun: `x + (← y)` is fine since we do not touch the parentheses. And infos around antiquotations shouldn't matter in practice anyway, no?

> From what I can tell, there are quite a few places where we explicitly use this pattern to preserve separators, atom source infos, syntax variants (unicode/non-unicode, semicolon/newline), etc.

I'm not directly aware of that, it might be incidental.

tests/lean/interactive/hover.lean
```diff
@@ -244,3 +244,7 @@ example : Nat → Nat
     --v textDocument/hover
   | .succ x => .succ x
                --^ textDocument/hover
+
+example : Inhabited Nat := ⟨Nat.zero⟩
+                         --^ textDocument/hover
+                          --^ textDocument/hover
```
digama0, line 250: Can you add an example for hovering `(. + .)`? I think that was the example that was reported.

Kha: That's the separate issue about original ranges becoming (non-canonical) synthetic, whose fix I removed again

Kha: Actually would it help if `expandCDot?` turned the original ranges into canonical ones? That *sounds* sensible.

Kha: Ok, looks pretty good! We lost the elaborator docstring on `·`, but that seems fair in exchange for the added type information. Not quite sure how to achieve both at the same time, except by applying the docstring to a copy of `elabIdent`...

Kha:
Whew, another post-merge breakage. Fortunately this time it was from the temporary merge the CI does on checkout.

src/Lean/Server/InfoUtils.lean
```diff
@@ -150,13 +150,21 @@ partial def InfoTree.hoverableInfoAt? (t : InfoTree) (hoverPos : String.Pos) (in
...
+      -- i.e. there are two info tree siblings whose spans are not disjoint and we should choose the smaller node
+      -- surrounding the cursor
+      Int.negOfNat (r.stop - r.start).byteIdx,
+      -- prefer results for constants over variables (which overlap at declaration names)
+      if info matches .ofTermInfo { expr := .fvar .., .. } then 0 else 1)
     [(priority, ctx, info)]) |>.getD []
+  -- sort results by lexicographical priority
+  let _ := @lexOrd
+  let _ := @leOfOrd.{0}
+  let _ := @maxOfLe
```
Kha, line 167: This code is getting a bit creative after introduction of `Max`. Note the fun universe annotation: the let binding is not actually used in the final term, presumably because it was unfolded in `Decidable`.
Should these definitions be scoped instances instead?

PR #1767, opened 2022-10-22
Make go-to-definition on a typeclass projection application go to the instance(s)

### LINKED ISSUE #1722: Go-to-definition on typeclass method should go to instance
...while go-to-declaration should go to the typeclass as it does right now. I hope that's not too controversial (and may even have been suggested before)? Note that you can still move from the instance to the class by using go-to-definition on the method name again.

rish987:
Closes #1722, pending #1903.


tests/lean/interactive/jumpMutual.lean
```diff
@@ -3,7 +3,6 @@ mutual
 def h (x : Nat) : Nat :=
   match x with
   | 0 => 1
-                     --v textDocument/definition
```
rish987, line 6: I'm not sure if there was a particular reason this was kept here but it definitely doesn't have the effect of testing go-to-definition on `h` (in this PR it produces `-32603: unknown constant 'Lean.Elab.Term.elabMatch'` which breaks the test).

rish987:
Hmm the CI test failures are puzzling me (it looks like they're still not getting the instance locations for some reason), because everything passes locally when I do:
```
$ ./test_single.sh {goTo,goto2}.lean
```

Kha:
The interactive tests are not designed for referencing core files in their output. Either you'll need to change the existing tests not to reference core instances, or we implement something in the vein of https://github.com/leanprover/lean4/issues/1722#issuecomment-1287787981 that would prevent these tests from showing instances.

leodemoura:
@rish987 Today we were discussing this PR in the dev meeting. Could you please clarify the semantics you are implementing?

rish987:
Sure, I was trying to follow what Sebastian described [here](https://github.com/leanprover/lean4/issues/1722#issuecomment-1286597236), basically, on go-to-definition or go-to-type if we get an application of a typeclass projection, rather than going to the syntax or elaborator, we extract all of the relevant instances and present that in a list on go-to-definition. I've also included the elaborator in this list because you were originally only able to get to that through go-to-definition (the syntax is still accessible as before through go-to-declaration).

I think I may need some help with the tests though (I blindly copied over expected outputs that I just noticed include URIs on my local filesystem 😰 so of course it won't work). Perhaps we need some additional filtering to preserve existing behavior (and some new tests as well to specifically test the new functionality), I was discussing this with @Kha yesterday and it sounded like he had an idea regarding that?

Kha:
Right, the fundamental issue is that when looking at the typeclass method call `Expr`, we don't know if *this* info node was responsible for the TC synthesis. What I proposed to Rish is to simply whitelist syntax that we know does this, e.g. ```i.stx.isOfKind identKind || i.stx.isOfKind ``Parser.Term.app``` and probably others as well. A different, probably more general heuristic would be to check if no *nested* info node contains the same `Expr` (making use of the tree structure of the info tree, heh).

rish987:
Okay this should finally be ready!

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -126,6 +126,44 @@ def locationLinksOfInfo (kind : GoToKind) (ci : Elab.ContextInfo) (i : Elab.Info
     | Expr.const n .. => return ← ci.runMetaM i.lctx <| locationLinksFromDecl i n
     | Expr.fvar id .. => return ← ci.runMetaM i.lctx <| locationLinksFromBinder i id
     | _ => pure ()
+
+    let isExprGenerator := children.all fun tree => match tree with
+        | .node (Info.ofTermInfo info) _ =>
+          info.expr != expr
+        | .node (Info.ofMacroExpansionInfo _) _ =>
+          false
+        | _ => true
```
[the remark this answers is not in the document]
Kha: This should have a short explanatory comment

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -126,6 +126,44 @@ def locationLinksOfInfo (kind : GoToKind) (ci : Elab.ContextInfo) (i : Elab.Info
...
+      -- should return all instances generated by TC
+      expr ← ci.runMetaM i.lctx do instantiateMVars expr
+      match expr.getAppFn with
+      | .const n _ =>
+        -- also include constant along with instance results
+        let mut results ← ci.runMetaM i.lctx <| locationLinksFromDecl i n
+        if let some info := ci.env.getProjectionFnInfo? n then
+          let mut result ←
+            if let some ei := i.toElabInfo? then do
+              if ei.elaborator != `Delab && ei.elaborator != `Lean.Elab.Term.elabApp && ei.elaborator != `Lean.Elab.Term.elabIdent then do
```
Kha, line 149: needs comment

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -126,6 +126,44 @@ def locationLinksOfInfo (kind : GoToKind) (ci : Elab.ContextInfo) (i : Elab.Info
...
+                -- also include elaborator along with instance results
+                ci.runMetaM i.lctx <| locationLinksFromDecl i ei.elaborator
+              else pure default
+            else pure default
```
Kha, line 153: You could initialize `result` to this and then use a conditional assignment

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -126,6 +126,44 @@ def locationLinksOfInfo (kind : GoToKind) (ci : Elab.ContextInfo) (i : Elab.Info
...
+          let instIdx := info.numParams
+          let appArgs := expr.getAppArgs
+          let rec extractInstances : Expr → RequestM (Array Name)
+            | .const declName _ => do
+              if ← ci.runMetaM i.lctx do Lean.Meta.isInstance declName then pure #[declName] else pure #[]
+            | .app fn arg => do pure $ (← extractInstances fn).append (← extractInstances arg)
+            | _ => pure #[]
+          if appArgs.size > instIdx then
+            let instArg := appArgs.get! instIdx
```
Kha, line 162:
```suggestion
          if let some instArg := appArgs.get? instIdx then
```

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -126,6 +126,44 @@ def locationLinksOfInfo (kind : GoToKind) (ci : Elab.ContextInfo) (i : Elab.Info
...
+            results := results.append <| ← (← extractInstances instArg).foldlM (init := result) fun acc n => do
+              pure $ acc.append (← ci.runMetaM i.lctx <| locationLinksFromDecl i n)
```
Kha, line 164: I think this would be nicer with a loop, unless I'm missing something?

src/Lean/Server/InfoUtils.lean
```diff
@@ -135,11 +135,18 @@ def InfoTree.smallestInfo? (p : Info → Bool) (t : InfoTree) : Option (ContextI
...
-partial def InfoTree.hoverableInfoAt? (t : InfoTree) (hoverPos : String.Pos) (includeStop := false) (omitAppFns := false) : Option (ContextInfo × Info) := Id.run do
-  let results := t.visitM (m := Id) (postNode := fun ctx info _ results => do
+partial def InfoTree.hoverableInfoAt? (t : InfoTree) (hoverPos : String.Pos) (includeStop := false) (omitAppFns := false) (omitIdentApps := false) : Option (ContextInfo × Info × (PersistentArray InfoTree)) := Id.run do
+  let results := t.visitM (m := Id) (postNode := fun ctx info c results => do
     let mut results := results.bind (·.getD [])
-    if omitAppFns && info.stx.isOfKind ``Parser.Term.app && info.stx[0].isIdent then
-      results := results.filter (·.2.2.stx != info.stx[0])
+    if omitAppFns then
+      if info.stx.isOfKind ``Parser.Term.app && info.stx[0].isIdent then
+        results := results.filter (·.2.2.1.stx != info.stx[0])
```
Kha, line 143: No functional change (apart from `.1`)?

src/Lean/Server/InfoUtils.lean
```diff
@@ -135,11 +135,18 @@ def InfoTree.smallestInfo? (p : Info → Bool) (t : InfoTree) : Option (ContextI
...
+    if omitIdentApps then
+      -- if an identifier stands for an application (e.g. in the case of a typeclass projection), prefer the application
+      if info.stx.isIdent then
+        if let .ofTermInfo ti := info then
+          if ti.expr matches .app _ _ then
```
Kha, line 148:
```suggestion
          if ti.expr.isApp then
```

src/Lean/Widget/Basic.lean
```diff
@@ -18,6 +18,7 @@ functionality is purpose-specific to showing the contents of infoview popups.
 structure InfoWithCtx where
   ctx  : Elab.ContextInfo
   info : Elab.Info
+  children : PersistentArray InfoTree
   deriving Inhabited, TypeName
```
Kha, line 22: Let's move this to `InfoUtils` and use it in `hoverableInfoAt?` and `locationLinksOfInfo` etc. The `deriving TypeName` should probably be left here though.

Kha (approved):
Looks great all in all!

src/Lean/Server/InfoUtils.lean
rish987: Nope (I've reverted the conditional)

src/Lean/Widget/Basic.lean
rish987: Done, let me know what you think of the refactor.

Vtec234: Hi, sorry to only comment on this now but `InfoWithCtx` is only really intended for use as a way of bundling data together in the RPC protocol so that interactive expressions and go-to-def in the infoview work. As Gabriel put it, it's a kind of "type popup + go-to-def thunk". Unless you need to do something with RPC here, it would be better not to move it and to maybe add another type with whatever semantics you need.

Kha: Duplicating it wouldn't be hard, but since the type is opaque in the RPC protocol I'm not really seeing the motivation. It's a sensible bundle of data to have in the server in general, and the name fits as well. Even if we don't necessarily need `children` in the widgets.

Vtec234: The type seems like a better fit for `Elab.InfoTree`, if anything because it was already hacky as far as the RPC interface is concerned — the outputs of delaboration are `Elab.Info`s but they should really be a much more restricted `Delab.Info`. But okay, since `InfoWithCtx` is opaque we can always keep it around for uses such as in this PR but replace it with something else in the RPC interface. Btw @rish987 you probably want to also make the output of `hoverableInfoAt?` and `termGoalAt?` be `InfoWithCtx`, document what the `children` field is, and change the docstring to note that it's used for more than just RPC.

rish987: Done, thanks for the suggestions.

tests/lean/interactive/goTo.lean
```diff
@@ -61,6 +61,53 @@ macro_rules | `(test) => `(3)
...
+instance : Foo2 := .mk id 0
+instance : Foo3 := .mk 0
+instance : Foo4 0 := .mk 0
+instance [foo2 : Foo2] : Foo5 := .mk foo2
+
+-- should go-to instance
+              --v textDocument/definition
+#check Foo2.foo  2
+          --^ textDocument/definition
+#check Foo2.foo
```
rish987, line 90: Reporting back after a merge with master -- It seems that this syntax is now elaborated to have the type `[self : Foo2] → Nat → Nat`, which is why the tests are failing now. Did something change recently for the `self` TC parameter to be now be interpreted as strict implicit?

Kha: That is https://github.com/leanprover/lean4/pull/1943

PR #1773, opened 2022-10-23
feat: use `withoutPosition` consistently

digama0:
* [ ] depends on #1772 (to avoid merge conflicts)

Although the parser for `(e)` used `withoutPosition`, there are many other bracketing notations in lean, and most of them did not use `withoutPosition`, meaning that you can have some weird/surprising parse errors like
```lean
example : String := by exact s!"{
  toString 2 -- expected }
}"
```
where an application fails to parse because it comes to the left of a withPosition parser like `exact` in this case. This PR adds `withoutPosition` on any notation that uses explicit bracketing characters like `(...)`, `{...}`, `[...]`. There are potentially other places which might have a bracketing effect but use keywords for delimiting like `unless ... do`, but these commonly also interact with the whitespace sensitivity behavior so it might be better to leave them as is (and have an unfinished `unless` be marked as such rather than trying to parse the next command as an expression).

### DIFF
--- a/src/Init/Conv.lean
+++ b/src/Init/Conv.lean
@@ -18,7 +18,7 @@ only on the left side of an equality. -/
 declare_syntax_cat conv (behavior := both)
 
 syntax convSeq1Indented := sepBy1IndentSemicolon(conv)
-syntax convSeqBracketed := "{" sepByIndentSemicolon(conv) "}"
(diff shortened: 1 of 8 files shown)


gebner:
As per the dev meeting: this is good to go once the merge conflicts are resolved.

PR #1778, opened 2022-10-25
fix: revert LLVM 15 bump on MacOS aarch64

digama0:
As reported by @AdrienChampion and @Kha , the uploaded artifact for LLVM 15 is for x86, not aarch64. I have not tested this fix, but this should hopefully be enough to stop us building bad nightlies until @Kha has more time to investigate.


Kha:
Thanks!

PR #1781, opened 2022-10-26
fix: bug in level normalization (soundness bug)

digama0:
As [reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Bug.20in.20kernel.20level.20normalization/near/304823172). The bug causes expressions of the form `imax u v + N` to be normalized to `imax u v` regardless of the value of `N`, leading to an inconsistency in the logic. The bug does not exist in lean 3, it was introduced in fb5440a.


leodemoura:
@digama0 Thanks for investigating and fixing this issue.

PR #1783, opened 2022-10-27
Add hover/go-to-def/refs for options

digama0:
Now, clicking on `foo` in `set_option foo true` takes you to the `register_option` declaration for `foo`, if one exists; hover shows the docstring for the declaration, or the option's description; and find references on the declaration takes you to uses of the `set_option`.

### DIFF
--- a/tests/lean/interactive/hover.lean
+++ b/tests/lean/interactive/hover.lean
@@ -175,6 +175,9 @@ example : Nat → Nat → Nat :=
     x
   --^ textDocument/hover
 
+           -- textDocument/definition -- removed because the result is platform-dependent
+set_option linter.unusedVariables false in
+          --^ textDocument/hover
 example : Nat → Nat → Nat := by
   intro x y
       --^ textDocument/hover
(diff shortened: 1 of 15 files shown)


tests/lean/interactive/hover.lean
```diff
@@ -175,6 +175,9 @@ example : Nat → Nat → Nat :=
     x
   --^ textDocument/hover
 
+           -- textDocument/definition -- removed because the result is platform-dependent
```
digama0, line 178: Unfortunately this is another of those tests that can't be written: This test jumps into the location of the lean 4 sources, which is different for the different CI runners and also for me locally. The goTo.lean tests fix this by only doing intra-file jumps, but options must be declared in a separate file so we have the usual issue about tests being single-file only.

gebner: Can we just replace the lean4 source directory by the empty string like we do for the tests?

digama0:
I don't know what is being asked of me here. I think the test scripts need to be rewritten in some way?

tests/lean/interactive/hover.lean
Kha: I don't know if this is desirable. We try to avoid source location references into src/ from tests since they are far from stable. It's really annoying when supposedly unrelated tests suddenly fail.

gebner (approved):

PR #1786, opened 2022-10-27
feat: improve import duplicate definition error

gebner:


gebner:
Not sure how to test this, I couldn't find any way to import local files in the tests.

leodemoura:
It is not very convenient to add this kind of test. We usually create a small package here
https://github.com/leanprover/lean4/tree/master/tests/pkg

gebner:
Right, I've seen those tests.  But there you can't easily test the error message.  (All it checks is that it builds, which it of course won't.)

leodemoura:
Some of them test for errors. It is not nice. 
https://github.com/leanprover/lean4/blob/master/tests/pkg/prv/test.sh

gebner:
Ugh, that's both ugly and a lot of effort.  If you want, I can add such a test for the error message here, but I'd rather avoid it.

PR #1794, opened 2022-11-01
fix: bug in replaceLocalDeclDefEq

### LINKED ISSUE #1615: Bug in replaceLocalDeclDefEq
### Prerequisites

### Description

Lean.MVarId.replaceLocalDeclDefEq doesn't do anything if the new type for the FVar happens to equal the type of the goal

### Steps to Reproduce

Here is an example that illustrates the problem:

elab "test" h:ident : tactic => do
  let g ← Lean.Elab.Tactic.getMainGoal
  g.withContext do
    let ldec ← Lean.Meta.getLocalDeclFromUserName h.getId
    let f := ldec.fvarId
    let e ← Lean.Meta.whnf (← Lean.instantiateMVars ldec.type)
    let newgoal ← g.replaceLocalDeclDefEq f e
    Lean.Elab.Tactic.replaceMainGoal [newgoal]

example (n : Nat) (h : 2 ∣ n) : ∃ c, n = c * 2 := by
  test h

example (n : Nat) (h : 2 ∣ n) : ∃ c, n = 2 * c := by
  test h

**Expected behavior:** In both cases, the type of h should change to ∃ c, n = 2 * c

**Actual behavior:** It changes in the first example, but not the second

### Versions

Lean (version 4.0.0-nightly-2022-08-25, commit d4799eaf3a74, Release)
MacOS 12.5.1

### Additional Information

It is clear what the bug is:  The test "if typeNew == mvarDecl.type then" is the wrong test.  Should be comparing typeNew to the type of the fvar, not the type of the mvar.

digama0:
fixes #1615

### DIFF
--- a/src/Lean/Meta/Tactic/Replace.lean
+++ b/src/Lean/Meta/Tactic/Replace.lean
@@ -99,10 +99,10 @@ Remark: this method assumes that `typeNew` is definitionally equal to the curren
 -/
 def _root_.Lean.MVarId.replaceLocalDeclDefEq (mvarId : MVarId) (fvarId : FVarId) (typeNew : Expr) : MetaM MVarId := do
   mvarId.withContext do
-    let mvarDecl ← mvarId.getDecl
-    if typeNew == mvarDecl.type then
+    if typeNew == (← fvarId.getType) then
       return mvarId
     else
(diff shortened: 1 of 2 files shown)


digama0:
CI job needs a restart

PR #1797, opened 2022-11-03
Add empty type ascription syntax `(e :)`

digama0:
This adds the syntax `(e :)` which can be used to completely suppress the expected type when elaborating `e`. The lean 3 idiom for this was `(e : _)`, but this no longer works because the elaboration of `e` is postponed, `_` is unified with the expected type, and then `e` is re-elaborated with the expected type. To fix this, `(e :)` uses a checkpoint to avoid postponing across it. We could special case the `(e : _)` syntax to have this behavior, but it seems more consistent to have a separate syntax for the special behavior. This syntax is needed by mathport to translate the lean 3 `(e : _)` idiom.

In order to avoid some of the same issues that happen with `[]` syntax, I think it is better to have this notation built in and merged with the other `(...)` parsers. It also means that the notation is treated uniformly with other type ascriptions in most cases.

### DIFF
--- a/src/Lean/Elab/Do.lean
+++ b/src/Lean/Elab/Do.lean
@@ -12,9 +12,6 @@ import Lean.Parser.Do
 -- HACK: avoid code explosion until heuristics are improved
 set_option compiler.reuse false
 
--- remove after stage0 update
-set_option quotPrecheck false
-
 namespace Lean.Elab.Term
 open Lean.Parser.Term
 open Meta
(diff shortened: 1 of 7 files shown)


leodemoura:
The new functionality makes sense to me, and we should support it.
Not sure about the syntax. @digama0 Are people happy with this new syntax on Zulip?
@Kha @gebner What do you think?

PR #1799, opened 2022-11-03
Introduce parser memoization to avoid exponentional behavior

Kha:
Cache invocations of category, `leading_parser` and `syntax :=` parsers using a mapping from parser name and current context & state to parse result. Abstract parser context and syntax stack to enforce correct use.

PR #1800, opened 2022-11-04
refactor: split paren parser

Kha:
Experiment based on #1799


Kha:
!bench

Kha:
Wait... it got faster?

src/Lean/Parser/Term.lean
```diff
@@ -137,8 +135,8 @@ Parentheses, used for
   - `(f · a b)` is shorthand for `fun x => f x a b`
   - `(h (· + 1) ·)` is shorthand for `fun x => h (fun y => y + 1) x`
 -/
-@[builtin_term_parser] def paren := leading_parser
-  "(" >> (withoutPosition (withoutForbidden (optional (ppDedentIfGrouped termParser >> parenSpecial)))) >> ")"
+@[builtin_term_parser] def paren := (leading_parser
```
gebner: If we're already splitting this parser, we should actually split it in three declarations.

gebner: Nevermind, I didn't see part 2.

Kha: Ah, there is a fun development story hidden here: given that the paren-like syntaxes are used in quotations all over the place, it was very important to make them use the new parser from the same stage that their elaborators were adapted in. But `parseQuotWithCurrentStage` does not affect *newly* added category parsers, thus the intermediate solution above. All in all, the code manually building syntax trees of the affected kinds was definitely the bigger issue as you have to locate it in the first place before you can fix it.

PR #1801, opened 2022-11-04
chore: ensure LCNF pretty printer set up LocalContext

ydewit:
I noticed that the LCNF pretty printer was not adding params, let and fun decls to the local context making it difficult to correlated variables.


leodemoura:
LCNF has an internalization process where all local declarations are added to the local context. The pretty printer assumes the declarations have been internalized.
We can chat about it during the compiler meeting. I am closing the PR.

PR #1805, opened 2022-11-07
perf: lean_apply_n: release closure before call, use tail call

Kha:
Not only should this, hopefully, be faster, it also dramatically cleans up the call stack and ensures we can capture more actually useful frames.


Kha:
!bench

Kha:
!bench

Kha:
Postponed; the impact in practice seems to be small and we can apply it locally if we want clean perf traces

Kha:
!bench

leanprover-radar:
[Benchmark results](https://radar.lean-lang.org/repos/lean4/commits/d5ada4049d3822ba600921be342a53a470c0756a?reference=cec9758c2d2b34f586e896777d08f1a1edc8dbd5) for d5ada4049d3822ba600921be342a53a470c0756a against cec9758c2d2b34f586e896777d08f1a1edc8dbd5 are in! @Kha
<details open>
<summary>Minor changes (1)</summary>

- 🟥 `Init.Data.BitVec.Lemmas re-elab//instructions`: **+5.3G** (**+0.6%**)
</details>

Kha:
I don't remember why I thought/how I managed this to actually do tail calls :thinking:

PR #1807, opened 2022-11-07
fix: improved error span for `inherit_doc`

digama0:
This moves the error span for `"foo" already has a docstring` in `/-- doc -/ @[inherit_doc bar] def foo` from `bar` to `inherit_doc`.

### DIFF
--- a/src/Lean/Elab/InheritDoc.lean
+++ b/src/Lean/Elab/InheritDoc.lean
@@ -16,8 +16,7 @@ builtin_initialize
       unless kind == AttributeKind.global do
         throwError "invalid `[inherit_doc]` attribute, must be global"
       match stx with
-      | `(attr| inherit_doc $[$id?:ident]?) =>
-        withRef stx[0] do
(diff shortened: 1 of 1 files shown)


src/Lean/Elab/InheritDoc.lean
```diff
@@ -17,11 +17,12 @@ builtin_initialize
         throwError "invalid `[inherit_doc]` attribute, must be global"
       match stx with
       | `(attr| inherit_doc $[$id?:ident]?) =>
+        withRef stx[0] do
```
Kha, line 20: You can move it to the line above if you don't want to reindent :)

Kha (approved):

src/Lean/Elab/InheritDoc.lean
digama0, line 20:
```suggestion
      | `(attr| inherit_doc $[$id?:ident]?) => withRef stx[0] do
```

PR #1809, opened 2022-11-07
feat: allow `doSeq` in `let x <- e | seq`

### LINKED ISSUE #1804: Counter intuitive behavior at `do` notation
```lean
def inc : StateM Nat Nat := do
  let s ← get
  modify (· + 1)
  return s

def f (x : Bool) : StateM Nat Nat := do
  let .true := x | return (← inc)
  get

def g (x : Bool) : StateM Nat Nat := do
  let .true := x | do return (← inc)
  get

#eval g true |>.run' 0 -- `0` as expected 
#eval f true |>.run' 0 -- `1` :(
/-
`f` is being incorrectly compiled as
``
def f (x : Bool) : StateM Nat Nat := do
  let aux ← inc
  let .true := x | return aux
  get
``
-/
```

digama0:
By changing `let x <- e | seq` to allow a sequence of `doElem`s at `seq` instead of just one, it fixes #1804.  The syntax extension is also convenient in some cases.


tests/lean/1804.lean
```diff
@@ -0,0 +1,24 @@
...
+def f (x : Bool) : StateM Nat Nat := do
+  let .true := x | return (← inc)
+  get
+
+def g (x : Bool) : StateM Nat Nat := do
+  let .true := x | do return (← inc)
+  get
+
+#eval g true |>.run' 0 -- `0` as expected
+#eval f true |>.run' 0 -- `1` :(
```
digama0, line 15: Opinions: should I change this test to show the new (fixed) behavior or keep the commentary that describes a bug that no longer exists but is being protected against regressions?

Kha: I usually change them to something like "should not return `1`"

Kha:
Nice generalization. I assume it's not trivial to trigger a sequence unintentionally unless you do something like
```
let x <- a | b
             c
```
? Seem fine to me then.

Kha (approved):

Kha:
Yes, `colEq` would be appropriate

gebner:
Ideally, do would use sepByIndent like by etc.  The reason I didn't do that myself is because I didn't want to touch the do notation expander, but you could give it a try.

PR #1811, opened 2022-11-08
Require strict indentation in nested `by`-s

AdrienChampion:
This makes the non-bracketed tactic sequence parser following a *nested* `by` require **strict**
indentation.

Mentored by @digama0, this PR introduces `Tactic.tacticSeqIndentGt` which is the same as
`Tactic.tacticSeq` (`by`'s current tactic sequence parser) but with `checkColGt "strict
indentation"` right before the non-bracketed tactic sequence.

It does **not** use `leading_parser` but instead builds a `Tactic.tacticSeq` node, the idea being
that antiquotations should not have to care whether the tactic sequence allows non-strict
indentation or not. It also makes the change relatively shallow and local.

## Motivation

Currently, the first element of a non-bracketed tactic sequences after a `by` can be indented
arbitrarily. The parser will accept the sequence as long as its tail has the same indentation as the
first element:

```lean
theorem easy : True := by
skip
skip
sorry
```

No problem for top-level `by`s, but things can get confusing (especially for newcomers) for nested
`by`s:

```lean
theorem easy : True := by
  have : True := by
  sorry -- recognized as the first element of the tactic sequence for `have`

  -- cannot write anything legal here
  -- - same indentation as `have` ⇒ part of the `have` tactic sequence ⇒ "no goals to be solved"
  -- - smaller indentation ⇒ not recognized as a tactic for top-level `by` ⇒ syntax error
```

This PR turns it into an error:

```lean
theorem easy : True := by
  have : True := by
  sorry -- expected '{' or strict indentation
```

to force

```lean
theorem easy : True := by
  have : True := by
    sorry
  sorry
```

### DIFF
--- a/src/Init/WF.lean
+++ b/src/Init/WF.lean
@@ -151,20 +151,20 @@ namespace Nat
 def lt_wfRel : WellFoundedRelation Nat where
   rel := Nat.lt
   wf  := by
-  apply WellFounded.intro
-  intro n
-  induction n with
-  | zero      =>
-    apply Acc.intro 0
-    intro _ h
-    apply absurd h (Nat.not_lt_zero _)
-  | succ n ih =>
-    apply Acc.intro (Nat.succ n)
-    intro m h
-    have : m = n ∨ m < n := Nat.eq_or_lt_of_le (Nat.le_of_succ_le_succ h)
-    match this with
-    | Or.inl e => subst e; assumption
-    | Or.inr e => exact Acc.inv ih e
+    apply WellFounded.intro
+    intro n
+    induction n with
+    | zero      =>
+      apply Acc.intro 0
+      intro _ h
+      apply absurd h (Nat.not_lt_zero _)
+    | succ n ih =>
+      apply Acc.intro (Nat.succ n)
+      intro m h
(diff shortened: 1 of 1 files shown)


digama0 (approved):

Kha:
The "expected command" error message is not great, but we have that issue in many other places. I suppose we could add a `colGt`-like parser combinator that unconditionally runs the inner parser, then complains about the indentation afterwards (but still at the initial position).

Kha (approved):

AdrienChampion:
Fixed `stage2`: `where` structure fields set the position, causing `by` to require strict indentation, which two structure definitions did not respect.

Which leads me to clarify that the following is not allowed anymore:

```lean
structure MyStruct where
  field1 := by
  long
  unindented
  proof
  field2 := by
  still
  no
  indentation
```

I would say it's good for this to be banned as it's perfectly unreadable to me. But reasonable minds might disagree, so another approach would be for field definitions to be parsed with `withoutPosition termParser` (thanks @digama0 for pointing this out).

Kha:
Looks fine to me

PR #1816, opened 2022-11-11
chore: abort build on panic

Kha:


Kha:
Tested locally that it in fact works

PR #1817, opened 2022-11-13
fix: `protected` on `Nat.add_zero`

digama0:


gebner (approved):

Kha:
TPiL4 was affected: https://github.com/leanprover/theorem_proving_in_lean4/actions/runs/3460438308/jobs/5776971025#step:6:645

Kha:
@digama0 Do you want to fix TPiL? I could do it of course since it shouldn't be difficult at all, but I don't want to singularly be responsible for fixing these breakages...

PR #1821, opened 2022-11-14
fix: issue with negative integers on Mac binaries 

alexjbest:
~~I believe this is the correct fix based on the default C++ values of INT_MIN and INT_MAX (one is even one odd), but to be honest I don't fully understand how these bounds are meant to be used and checked.~~

This PR tries to fix the following behaviour 
```lean
theorem boom : -2147483648 + 2147483648 = 0 := by rfl
theorem boom' : -2147483648 + 2147483648 = -18446744069414584320 := by native_decide
theorem boom'' : False := by
  have := boom'
  simp [boom] at this
```
which occurs at least on a 64-bit intel mac

### DIFF
--- a/tests/lean/1821.lean.expected.out
+++ b/tests/lean/1821.lean.expected.out
@@ -1,9 +1,9 @@
-tests/lean/1821.lean:2:8-2:13: error: application type mismatch
+1821.lean:2:8-2:13: error: application type mismatch
   Lean.ofReduceBool boom'._nativeDecide_1 true (Eq.refl true)
 argument has type
   true = true
 but function has type
   Lean.reduceBool boom'._nativeDecide_1 = true → boom'._nativeDecide_1 = true
-tests/lean/1821.lean:4:10-4:15: error: unknown identifier 'boom''
-tests/lean/1821.lean:3:26-5:21: error: unsolved goals
+1821.lean:4:10-4:15: error: unknown identifier 'boom''
+1821.lean:3:26-5:21: error: unsolved goals
 ⊢ False


leodemoura:
@Kha This issue affects 32-bit machines. Is there an easy way to test Lean in a 32-bit VM in our CI?

alexjbest:
@leodemoura this is very curious actually, my machine is a recent intel mac, so it should indeed be 64 bit, and CMake detects it as such, so clearly something very odd is happening that I even hit this bug at all. I'll look into this more.

alexjbest:
What's most likely at this point is that simply the act of building on my own machine fixed this issue for me, not the fix proposed here. The original issue appears to be a real issue affecting my 64 bit mac at least.

leodemoura:
> What's most likely at this point is that simply the act of building lean on my own machine fixed this issue for me, not the fix proposed here. The original issue appears to be a real issue affecting my 64 bit mac at least.

@alexjbest Could you please try to build without the change?

@kha Could you please try to reproduce the bug using your Mac mini?

alexjbest:
@leodemoura I just did this and indeed simply building master on my machine rather than using the latest Mac nightly fixes the issue for me

leodemoura:
> @leodemoura I just did this and indeed simply building master on my machine rather than using the latest Mac nightly fixes the issue for me

Same for me.

@alexjbest Is this a new issue?

alexjbest:
The test case I reduced this from was one I hadn't run until a few days ago so I'm not sure how far back this issue goes,

leodemoura:
@alexjbest Could you please close the PR and create an issue?

alexjbest:
Done, sorry for the confusion!

PR #1824, opened 2022-11-14
feat: go to definition for `KeyedDeclsAttribute`s

digama0:
Adds ctrl-click support for the identifier `foo` in attributes powered by `KeyedDeclsAttribute.init`:
* `[(builtin_)delab app.foo]`
* `[(builtin_)combinator_parenthesizer foo]`
* `[(builtin_)category_parenthesizer foo]`
* `[(builtin_)parenthesizer foo]`
* `[(builtin_)formatter foo]`
* `[(builtin_)quot_precheck foo]`

### DIFF
--- a/src/Lean/Elab/Util.lean
+++ b/src/Lean/Elab/Util.lean
@@ -102,13 +102,7 @@ unsafe def mkElabAttribute (γ) (attrBuiltinName attrName : Name) (parserNamespa
       let kind ← syntaxNodeKindOfAttrParam parserNamespace stx
       /- Recall that a `SyntaxNodeKind` is often the name of the parser, but this is not always true, and we must check it. -/
       if (← getEnv).contains kind && (← getInfoState).enabled then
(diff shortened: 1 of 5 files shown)


digama0:
I didn't change name resolution because many of these are using syntax kinds and stuff like `ident` needs not to be accidentally qualified if there happens to be an `ident` definition in scope. I noticed that one of them is using a custom name resolution procedure for syntax kinds, maybe we can use that.

Or, maybe we can just get rid of syntax kinds that aren't declarations. They are a common source of errors since you can't use e.g.  ``` TSyntax ``tactic ``` with them. We already have dummy declarations for syntax categories; we should just extend that to the builtin syntax node kinds like `ident` and `num`.

Kha:
Note that the `combinator` attributes do in fact take declaration names, not syntax kinds.

> We already have dummy declarations for syntax categories; we should just extend that to the builtin syntax node kinds like ident and num.

That would be nice.

PR #1830, opened 2022-11-14
perf: remove unnecessary, cache-defeating `withPosition` in `doReassignArrow`

Kha:
As noticed by @digama0. This was introduced in https://github.com/leanprover/lean4/commit/2909313475588cc20ac0436e55548a4502050d0a, but the `checkColGt "|"` part should be sufficient since `doSeqIndent` (and `termReturn`) already uses `withPosition`.


Kha:
!bench

PR #1837, opened 2022-11-16
[WIP] LLVM Backend

bollu:
This is a WIP PR, to be merged after #1497 . This PR shows the LLVM backend implementation, and is a stacked PR, kept open here to ensure that changes suggested to #1497 are kept in sync with the actual backend.

We want CI to run on this branch. Since CI is disabled on a draft PRs, this PR is not sent as a github draft PR


src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,2425 @@
...
+-- ***lean_box***
+def getOrCreateLeanBoxFn: M (LLVM.Ptr LLVM.Value) := do
+  let ctx ← getLLVMContext
+  getOrCreateFunctionPrototype ctx (← getLLVMModule)
+    (← LLVM.voidPtrType ctx) "lean_box"  #[(← LLVM.size_tType ctx)]
+
+def callLeanBox (builder: LLVM.Ptr LLVM.Builder) (arg: LLVM.Ptr LLVM.Value) (name: String := ""): M (LLVM.Ptr LLVM.Value) := do
+  LLVM.buildCall builder (← getOrCreateLeanBoxFn) #[arg] name
+
+-- ***void lean_mark_persistent (void *) ***--
```
Kha, line 130: I think you've upset GH's code highlighter 
image

Kha:
It's interesting that even programs importing a select few modules from `Lean` try to link against LLVM: https://github.com/leanprover/lean4/actions/runs/3482585390/jobs/5825073061#step:11:4385. We should probably check what chain of symbols leads to that. But this also begs the larger question, should it be possible to compile `import Lean` programs without linking against LLVM? Perhaps the backend should be moved to its own package at some point, but a simple workaround would be to *not* import `EmitLLVM` from `Compiler.IR`.

bollu:
> ... a simple workaround would be to not import EmitLLVM from Compiler.IR.

@Kha I imagine that in this scenario, the compiler frontend (`src/util/shell.cpp`) will invoke  `initialize_Lean_Compiler_IR_EmitLLVM` directly when the user asks to emit LLVM bitcode?

Kha:
Yes, exactly!

bollu:
Does github have a concept of a "stacked PR" by any chance? I keep rebasing over https://github.com/leanprover/lean4/pull/1497, which needs me to force-push every time.

bollu:
@Kha It's quite unfortunate that we bumped from LLVM 14 to LLVM 15, [right when the opaque pointers breaking changes were added](https://releases.llvm.org/15.0.0/docs/ReleaseNotes.html#changes-to-the-llvm-ir). I've currently added a check to the CMake to ensure that we only run enable LLVM when `llvm-config` confirms that we have LLVM 15. 

Unfortunately, I don't think we can build successfully on Nix, since Nix does not have support for LLVM 15.

Kha:
Let's just set LLVM=OFF for Nix for now, support is around the corner https://github.com/NixOS/nixpkgs/pull/194634

Kha:
@bollu Oh I guess we should use `CMAKE_ARCHIVE_OUTPUT_DIRECTORY` for `lean.h.bc` since `LIBRARY` points to `bin/` on Windows

src/util/shell.cpp
```diff
@@ -727,6 +746,26 @@ extern "C" LEAN_EXPORT int lean_main(int argc, char ** argv) {
             out.close();
         }
 
+        if (llvm_output && ok) {
+          initialize_Lean_Compiler_IR_EmitLLVM(/*builtin*/ false,
+                                               lean_io_mk_world());
+
+	  //initialize_llvm_backend();
```
tobiasgrosser, line 753: This dead comment can be removed.

tests/bench/.gitignore
```diff
@@ -1,5 +1,10 @@
...
+*.lean.c.ll
+*.lean.c.o2.ll
+*.bc
+*.bc.ll
+*.bc.o2.ll
 *.cmi
 *.cmx
-*.o
\ No newline at end of file
+*.o
```
tobiasgrosser, line 10: Newline at the end of the file.

tobiasgrosser (commented):
Some minor comments on the code.

src/CMakeLists.txt
```diff
@@ -261,10 +261,15 @@ if(LLVM)
...
-  message(STATUS "Found 'llvm-config' as ${LLVM_CONFIG}")
+  execute_process(COMMAND ${LLVM_CONFIG} --version COMMAND_ERROR_IS_FATAL ANY OUTPUT_VARIABLE LLVM_CONFIG_VERSION ECHO_OUTPUT_VARIABLE OUTPUT_STRIP_TRAILING_WHITESPACE)
+  string(REGEX MATCH "^[0-9]*" LLVM_CONFIG_MAJOR_VERSION ${LLVM_CONFIG_VERSION})
+  message(STATUS "Found 'llvm-config' at '${LLVM_CONFIG}' with version '${LLVM_CONFIG_VERSION}', major version '${LLVM_CONFIG_MAJOR_VERSION}'")
+  if (NOT LLVM_CONFIG_MAJOR_VERSION STREQUAL "15")
+    message(FATAL_ERROR "Unable to find llvm-config version 15. Found invalid version '${LLVM_CONFIG_MAJOR_VERSION}'")
+  endif()
   # -DLEAN_LLVM is used to conditionally compile Lean features that depend on LLVM
-  string(APPEND LEAN_EXTRA_CXX_FLAGS " -D LEAN_LLVM")
+  # string(APPEND LEAN_EXTRA_CXX_FLAGS " -DLEAN_EXTRA_CXX_FLAGS -D LEAN_LLVM")
```
tobiasgrosser, line 271: This is unused dead code as well?

src/CMakeLists.txt
```diff
@@ -320,7 +325,9 @@ if(LLVM)
     #   we need the path to be `-Lllvm/lib/libLLVM`. Thus, we perform this replacement here.
     string(APPEND LEANSHARED_LINKER_FLAGS " -L${LLVM_CONFIG_LIBDIR} ${LLVM_CONFIG_LDFLAGS} ${LLVM_CONFIG_LIBS} ${LLVM_CONFIG_SYSTEM_LIBS}")
     string(REPLACE "llvm-host" "llvm" LEANSHARED_LINKER_FLAGS ${LEANSHARED_LINKER_FLAGS})
-    set(APPEND LEAN_EXTRA_CXX_FLAGS  " -I${LLVM_CONFIG_INCLUDEDIR}")
+    # string(APPEND LEAN_EXTRA_CXX_FLAGS  " -DLEAN_EXTRA_CXX_FLAGS -I${LLVM_CONFIG_INCLUDEDIR}")
+    string(APPEND CMAKE_CXX_FLAGS  " -I${LLVM_CONFIG_INCLUDEDIR}")
+    # string(APPEND LEANC_EXTRA_FLAGS " -I${LLVM_CONFIG_INCLUDEDIR}")
```
tobiasgrosser, line 330: What about this unused dead code?

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -37,10 +41,6 @@ structure Builder (ctx : Context) where
   private mk :: ptr : USize
 instance : Nonempty (Builder ctx) := ⟨{ ptr := default }⟩
 
-structure Context where
-  private mk :: ptr : USize
-instance : Nonempty Context := ⟨{ ptr := default }⟩
-
 structure LLVMType (ctx : Context) where
```
tobiasgrosser, line 44: Not sure moving this code is needed.

src/util/shell.cpp
```diff
@@ -727,6 +744,24 @@ extern "C" LEAN_EXPORT int lean_main(int argc, char ** argv) {
...
+          time_task _("LLVM code generation", opts);
+          object *r = lean_ir_emit_llvm(
+              env.to_obj_arg(), (*main_module_name).to_obj_arg(),
+              lean::string_ref(*llvm_output).to_obj_arg(), lean_io_mk_world());
+          if (lean_io_result_is_ok(r)) {
+            dec_ref(r);
+          } else {
+            string_ref s(lean_io_result_get_error(r), true);
+            dec_ref(r);  // TODO: This is correct reference count handling,
+                         // correct?
```
tobiasgrosser, line 760: Is this TODO still needed?

tests/common.sh
```diff
@@ -46,7 +60,7 @@ function exec_check {
 function diff_produced {
     if test -f "$f.expected.out"; then
         if $DIFF -au --strip-trailing-cr -I "executing external script" "$f.expected.out" "$f.produced.out"; then
-            exit 0
+            :
```
tobiasgrosser, line 63: Just to double-check, is this intentional?

tobiasgrosser (commented):
Some more minor comments.

Kha:
Just to be clear, this is still WIP and not ready for reviews yet, correct?

bollu:
@Kha Indeed, I'm still polishing up the patch. I'll change the status and ping you when it's ready for review.

Kha:
> CI is disabled on a draft PRs

Just wanted to say, this is not true afaik. We have other draft PRs with CI enabled.

bollu:
@Kha , ready for review, with concerns I have tagged as comments!

src/Lean/Data/HashMap.lean
```diff
@@ -31,7 +31,6 @@ namespace HashMapImp
 variable {α : Type u} {β : Type v}
 
 /- Remark: we use a C implementation because this function is performance critical. -/
-@[extern c inline "(size_t)(#2) & (lean_unbox(#1) - 1)"]
```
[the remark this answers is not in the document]
Kha: Yes, we should probably introduce a new header... but given the small number of affected functions, I think it's fine to just append them to `lean.h` for now

tests/compiler/computedFieldsExtern.lean
```diff
@@ -2,7 +2,7 @@ inductive Exp
   | var (i : Nat)
   | app (a b : Exp)
 with
-  @[computed_field, extern c inline "(lean_ctor_get_uint64(#1, lean_ctor_num_objs(#1)*sizeof(void*)) + 40)"]
+  @[computed_field]
```
Kha, line 5: Might as well delete the test case, haha

tests/compiler/test_single.sh
```diff
@@ -1,6 +1,18 @@
...
 diff_produced
+
+# Then check the LLVM version
+if lean_has_llvm_support; then
+    echo "running LLVM program..."
+    rm "./$f.out" || true
+    compile_lean_llvm_backend
+    exec_check "./$f.out"
+    diff_produced
+fi
```
Kha, line 18: This change is different from the other `.sh`s

Kha:
@bollu Could you create an issue on what's left to do (e.g. `size_t` TODO) before the backend is "ready" and/or the default?

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,1532 @@
...
+     emitSet builder x i y; emitBlock builder b
+  | FnBody.uset _x _i _y _b    =>
+       -- TODO(bollu): implement `uset`.
+       --  NOTE(bollu): It is disturbing that we pass the Lean CI without this.
+       throw (Error.unimplemented "uset")
+  | FnBody.sset x i o y t b    => emitSSet builder x i o y t; emitBlock builder b
+  | FnBody.mdata _ _b          =>
+       -- TODO(bollu): implement `mdata`.
+       --  NOTE(bollu): It is disturbing that we pass the Lean CI without this.
+       throw (Error.unimplemented "metadata")
```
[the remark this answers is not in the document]
Kha: I think it's unused /cc @leodemoura

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,1532 @@
...
+  | FnBody.del x b             =>  emitDel builder x; emitBlock builder b
+
+  | FnBody.setTag x i b        =>  emitSetTag builder x i; emitBlock builder b
+
+  | FnBody.set x i y b         =>
+     emitSet builder x i y; emitBlock builder b
+  | FnBody.uset _x _i _y _b    =>
+       -- TODO(bollu): implement `uset`.
+       --  NOTE(bollu): It is disturbing that we pass the Lean CI without this.
+       throw (Error.unimplemented "uset")
```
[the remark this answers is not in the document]
Kha: If you want to create one, something like this should work
```
structure Point where
  x : USize
  y : UInt32

@[noinline] def Point.right (p : Point) : Point :=
  { p with x := p.x + 1 }

def main : IO Unit :=
  IO.println (Point.right ⟨0, 0⟩).x
```

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,1521 @@
...
+
+def constIntUnsigned (n : Nat) : M llvmctx (LLVM.Value llvmctx) :=  do
+    LLVM.constIntUnsigned llvmctx (UInt64.ofNat n)
+
+def getOrCreateFunctionPrototype (mod : LLVM.Module llvmctx)
+  (retty : LLVM.LLVMType llvmctx) (name : String) (args : Array (LLVM.LLVMType llvmctx)) : M llvmctx  (LLVM.Value llvmctx) := do
+  LLVM.getOrAddFunction mod name $
+     (← LLVM.functionType retty args (isVarArg := false))
+
+def callLeanBox (builder : LLVM.Builder llvmctx) (arg : LLVM.Value llvmctx) (name : String := "") : M llvmctx (LLVM.Value llvmctx) := do
```
hargoniX, line 97: It would be pretty cool if we had some sort of general framework for this type of stuff. That is, making a Lean function of this style that automatically generates a call to an LLVM function with a given signature but I guess that's far out of scope here.

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,1521 @@
...
+  | _ => throw (Error.compileError s!"failed to emit extern application '{f}'")
+
+def getFunIdTy (f : FunId) : M llvmctx (LLVM.LLVMType llvmctx) := do
+  let decl ← getDecl f
+  let retty ← toLLVMType decl.resultType
+  let argtys ← decl.params.mapM (fun p => do toLLVMType p.ty)
+  LLVM.functionType retty argtys
+
+/-
+Create a function declaration and return a pointer to the function.
```
hargoniX, line 610: Is there a reason for switching between multiline comment style? Further above you have multiple `--`. If you inteded for this to be a doc string that gets rendered by e.g. doc-gen4 you'll have to put `/--`

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,1521 @@
...
+
+def emitFullApp (builder : LLVM.Builder llvmctx)
+  (z : VarId) (f : FunId) (ys : Array Arg) : M llvmctx Unit := do
+  let (__zty, zslot) ← emitLhsSlot_ z
+  let decl ← getDecl f
+  match decl with
+  | Decl.extern _ ps retty extData =>
+     let zv ← emitExternCall builder f ps extData ys retty
+     LLVM.buildStore builder zv zslot
+  | _ =>
```
hargoniX, line 679: Is there a particular reason for the `_` here? In general I feel like adding wildcards when you don't need them can be dangerous, if we end up having a new `Decl` variant in the future for some reason we might overlook something here because the compiler did not tell us.

hargoniX (commented):
Just my 5 cent, I don't really know much about LLVM to comment about the code itself.

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,1521 @@
...
+
+structure State (llvmctx : LLVM.Context) where
+  var2val : HashMap VarId (LLVM.LLVMType llvmctx × LLVM.Value llvmctx)
+  jp2bb   : HashMap JoinPointId (LLVM.BasicBlock llvmctx)
+
+inductive Error where
+| unknownDeclaration : Name → Error
+| invalidExportName : Name → Error
+| unimplemented : String → Error
+| compileError : String → Error -- TODO: change this into an exhaustive Inductive enumeration.
```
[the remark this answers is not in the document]
bollu: @Kha I wanted to have an exhaustive enumeration, so we could eventually produce richer messages. For now, I think I'll remove it and switch to `String`.

src/Lean/Compiler/IR/EmitLLVM.lean
bollu: Fixed here. In some other places, I use `_` where there are quite a few variants to be handled (eg. `IR.isIntegerType`). Lean does not have [or-patterns](https://v2.ocaml.org/manual/patterns.html#sss:pat-or), does it? If it does, then I could write the code as:

```lean
match scrutinee with
| useful => ...
| useless1 | useless2 | ... | uselessN => throw ... 
```

Without this construct, it would lead to a bunch of duplication of the RHS:

```lean
match scrutinee with
| useful => ...
| useless1 => throw ... 
| useless2 => throw ...  -- duplicated
| ...  -- duplicated
| uselessN => throw ... -- duplicated 
```

src/Lean/Compiler/IR/EmitLLVM.lean
bollu: Thanks! I did intend for it to be a docstring.

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,1521 @@
...
+
+instance : ToString Error where
+  toString e := match e with
+   | .unknownDeclaration n => s!"unknown declaration '{n}'"
+   | .invalidExportName n => s!"invalid export name '{n}'"
+   | .unimplemented s => s!"unimplemented '{s}'"
+   | .compileError s => s!"compile error '{s}'"
+
+abbrev M (llvmctx : LLVM.Context) :=
+  StateT (State llvmctx) (ReaderT (Context llvmctx) (ExceptT Error IO))
```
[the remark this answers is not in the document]
bollu: Old Haskell habits die hard :) Fixed to be `StateRefT`.

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,1521 @@
...
+
+def getEnv : M llvmctx Environment := Context.env <$> read
+
+def getModName : M llvmctx  Name := Context.modName <$> read
+
+def getDecl (n : Name) : M llvmctx Decl := do
+  let env ← getEnv
+  match findEnvDecl env n with
+  | some d => pure d
+  | none   => IO.eprintln "getDecl failed!"; throw (Error.unknownDeclaration n)
```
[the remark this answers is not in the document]
bollu: Resolved!

src/CMakeLists.txt
```diff
@@ -261,10 +261,14 @@ if(LLVM)
...
-  message(STATUS "Found 'llvm-config' as ${LLVM_CONFIG}")
+  execute_process(COMMAND ${LLVM_CONFIG} --version COMMAND_ERROR_IS_FATAL ANY OUTPUT_VARIABLE LLVM_CONFIG_VERSION ECHO_OUTPUT_VARIABLE OUTPUT_STRIP_TRAILING_WHITESPACE)
+  string(REGEX MATCH "^[0-9]*" LLVM_CONFIG_MAJOR_VERSION ${LLVM_CONFIG_VERSION})
+  message(STATUS "Found 'llvm-config' at '${LLVM_CONFIG}' with version '${LLVM_CONFIG_VERSION}', major version '${LLVM_CONFIG_MAJOR_VERSION}'")
+  if (NOT LLVM_CONFIG_MAJOR_VERSION STREQUAL "15")
+    message(FATAL_ERROR "Unable to find llvm-config version 15. Found invalid version '${LLVM_CONFIG_MAJOR_VERSION}'")
+  endif()
   # -DLEAN_LLVM is used to conditionally compile Lean features that depend on LLVM
-  string(APPEND LEAN_EXTRA_CXX_FLAGS " -D LEAN_LLVM")
+  string(APPEND CMAKE_CXX_FLAGS " -DCMAKE_CXX_FLAGS -D LEAN_LLVM")
```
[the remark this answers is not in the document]
bollu: fixed, removed the incorrect flag.

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,1521 @@
...
+  LLVM.buildCall2 builder fnty fn #[strPtr] name
+
+def callLeanCStrToNatFn (builder : LLVM.Builder llvmctx) (n : Nat) (name : String := "") : M llvmctx (LLVM.Value llvmctx) := do
+  let fnName :=  "lean_cstr_to_nat"
+  let retty ← LLVM.voidPtrType llvmctx
+  let argtys :=  #[← LLVM.voidPtrType llvmctx]
+  let fn ← getOrCreateFunctionPrototype (← getLLVMModule) retty fnName argtys
+  let fnty ← LLVM.functionType retty argtys
+  let s ← LLVM.buildGlobalString builder (value := toString n)
+  let s ← LLVM.buildPointerCast builder s (← LLVM.i8PtrType llvmctx)
```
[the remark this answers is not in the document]
bollu: You're right, this is old code from LLVM 14. Removed pointer casts from the generated code and the API.

src/Lean/Compiler/IR/EmitLLVM.lean
Kha: > Lean does not have [or-patterns](https://v2.ocaml.org/manual/patterns.html#sss:pat-or), does it?

It does, using that exact syntax

tests/common.sh
tobiasgrosser: I am confused. This change is still present.

bollu:
Yes, the `:` is necessary. It ensures that if the test passes, we continue
execution.

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,1521 @@
...
+| invalidExportName : Name → Error
+| unimplemented : String → Error
+| compileError : String → Error -- TODO: change this into an exhaustive Inductive enumeration.
+
+instance : ToString Error where
+  toString e := match e with
+   | .unknownDeclaration n => s!"unknown declaration '{n}'"
+   | .invalidExportName n => s!"invalid export name '{n}'"
+   | .unimplemented s => s!"unimplemented '{s}'"
+   | .compileError s => s!"compile error '{s}'"
```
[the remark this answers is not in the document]
bollu: Inductive has been removed.

src/Lean/Compiler/IR/EmitLLVM.lean
bollu: Agreed, it would be nice!

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,1521 @@
...
+def getOrCreateFunctionPrototype (mod : LLVM.Module llvmctx)
+  (retty : LLVM.LLVMType llvmctx) (name : String) (args : Array (LLVM.LLVMType llvmctx)) : M llvmctx  (LLVM.Value llvmctx) := do
+  LLVM.getOrAddFunction mod name $
+     (← LLVM.functionType retty args (isVarArg := false))
+
+def callLeanBox (builder : LLVM.Builder llvmctx) (arg : LLVM.Value llvmctx) (name : String := "") : M llvmctx (LLVM.Value llvmctx) := do
+  let fnName :=  "lean_box"
+  let retty ← LLVM.voidPtrType llvmctx
+  let argtys := #[ ← LLVM.size_tType llvmctx ]
+  let fn ← getOrCreateFunctionPrototype (← getLLVMModule) retty fnName argtys
```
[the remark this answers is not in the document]
bollu: It's not as bad as it looks, as [LLVM `Type`s are uniqued](https://llvm.org/doxygen/classllvm_1_1Type.html#details), so most of the time is spent in the fast path, looking up caches. 

Here's [the implementation of `FunctionType::get`](https://llvm.org/doxygen/Type_8cpp_source.html#l00360
) for example:
```cpp
// This is the factory function for the FunctionType class.
FunctionType *FunctionType::get(Type *ReturnType,
                                 ArrayRef<Type*> Params, bool isVarArg) {
   ...
   // Since we only want to allocate a fresh function type in case none is found
   // and we don't want to perform two lookups (one for checking if existent and
   // one for inserting the newly allocated one), here we instead lookup based on
   // Key and update the reference to the function type in-place to a newly
   // allocated one if not found.
   auto Insertion = pImpl->FunctionTypes.insert_as(nullptr, Key);
   if (Insertion.second) {
     // The function type was not found. Allocate one and update FunctionTypes
     // in-place.
     ...
  } else {
     // The function type was found. Just return it.
     FT = *Insertion.first;
   }
   return FT;
 }
```

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,1521 @@
...
+  let voidptr ← LLVM.voidPtrType llvmctx
+  let argtys := if argv?.isSome then #[ voidptr, voidptr ] else #[ voidptr ]
+  let fn ← getOrCreateFunctionPrototype (← getLLVMModule) retty leanMainFn argtys
+  let fnty ← LLVM.functionType retty argtys
+  let args := match argv? with 
+              | .some argv => #[argv, world]
+              | .none => #[world]
+  LLVM.buildCall2 builder fnty fn args name
+
+def emitMainFn (mod : LLVM.Module llvmctx) (builder : LLVM.Builder llvmctx) : M llvmctx Unit := do
```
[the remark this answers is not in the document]
bollu: Agreed. I opened an issue to track this refactor: https://github.com/leanprover/lean4/issues/1973

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -0,0 +1,1521 @@
...
+         LLVM.writeBitcodeToFile emitLLVMCtx.llvmmodule filepath
+         let tripleStr ← LLVM.getDefaultTargetTriple
+         let target ← LLVM.getTargetFromTriple tripleStr
+         let cpu := "generic"
+         let features := ""
+         let targetMachine ← LLVM.createTargetMachine target tripleStr cpu features
+         let codegenType := LLVM.CodegenFileType.ObjectFile
+         LLVM.targetMachineEmitToFile targetMachine emitLLVMCtx.llvmmodule (filepath ++ ".o") codegenType
+         LLVM.disposeModule emitLLVMCtx.llvmmodule
+         LLVM.disposeTargetMachine targetMachine
```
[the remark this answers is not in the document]
bollu: Yes, I agree it is unlikely to go wrong, since they are allocated once and freed once. I did make an issue to track this: https://github.com/leanprover/lean4/issues/1974

src/Lean/Data/HashMap.lean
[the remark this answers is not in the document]
bollu: Done, this is now moved to `lean.h`.

src/util/shell.cpp
```diff
@@ -727,6 +744,23 @@ extern "C" LEAN_EXPORT int lean_main(int argc, char ** argv) {
...
+          object *r = lean_ir_emit_llvm(
+              env.to_obj_arg(), (*main_module_name).to_obj_arg(),
+              lean::string_ref(*llvm_output).to_obj_arg(), lean_io_mk_world());
+          if (lean_io_result_is_ok(r)) {
+            dec_ref(r);
+          } else {
+            string_ref s(lean_io_result_get_error(r), true);
+            dec_ref(r);
+            throw exception(s.to_std_string());
+          }
```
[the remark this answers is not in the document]
bollu: Changed the `if-then-else` to use `consume_io_result`.

tests/bench/.gitignore
```diff
@@ -1,5 +1,10 @@
 *.out
 *.lean.c
+*.lean.c.ll
+*.lean.c.o2.ll
+*.bc
+*.bc.ll
+*.bc.o2.ll
```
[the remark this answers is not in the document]
bollu: Fixed, the `gitignore` is minimised now.

tests/compiler/test_single.sh
bollu: Fixed, the shell scripts ought to be consistent now.

tests/compiler/computedFieldsExtern.lean
bollu: Deleted test case.

tests/common.sh
bollu: Yes, I am sorry. The `:` is necessary, and ensures that the rest of the test cases are run ([`:` is a no-op in shell](https://stackoverflow.com/questions/3224878/what-is-the-purpose-of-the-colon-gnu-bash-builtin)).

bollu (commented):
@Kha I've addressed this round of comments. Open issues:

1. Merging  the output of `--features` with the header (which might mess with extension sniffing code? /cc @gebner)

src/runtime/runtmp/lean_inlines.cpp
```diff
@@ -0,0 +1,1887 @@
+/*
```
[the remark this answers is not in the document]
Kha: This is still here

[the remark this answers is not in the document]
bollu: This has now been removed as of 10afa9b417cd266435cf3bedcc62b341802e672d

Kha:
> @Kha I've addressed this round of comments. Open issues:
> 
>     1. Merging  the output of `--features` with the header (which might mess with extension sniffing code? /cc @gebner)

If we expect that we add more such configuration-dependent features in the future, and perhaps more than should reasonably be printed in the `lean -v` header, then your current setup might make more sense. For example, `GMP` should probably be a feature (but also be part of another descriptor of the .olean format, which LLVM does not affect). @leanprover/dev Thoughts?

Kha:
Alright, let's merge this then as a reasonable foundation. Great work, @bollu!

PR #1839, opened 2022-11-16
doc: update widget docs

EdAyers:
- Fix the rpc code example
- Additional section about how to insert text from a widget.

### DIFF
--- a/doc/examples/widgets.lean
+++ b/doc/examples/widgets.lean
@@ -180,4 +180,38 @@ the infoview we need to:
 
 In the RubiksCube sample, we provide a working `rollup.js` build configuration in
 [rollup.config.js](https://github.com/leanprover/lean4-samples/blob/main/RubiksCube/widget/rollup.config.js).
+
+## Inserting text
+
+We can also instruct the editor to insert text, copy text to the clipboard, or
+reveal a certain location in the document.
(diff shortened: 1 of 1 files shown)


doc/examples/widgets.lean
```diff
@@ -137,24 +137,21 @@ def checkWidget : UserWidgetDefinition where
...
+  const [value, setValue] = React.useState(undefined)
 
-  const [status, val, err] = useAsync(() =>
-    rs.call('getType', { name, pos: props.pos }), [name, rs, props.pos])
-
-  const type = status === 'fulfilled' ? val && e(InteractiveCode, {fmt: val})
-    : status === 'rejected' ? e('p', null, mapRpcError(err).message)
-      : e('p', null, 'Loading..')
+  function run() {
+    rs.call('getType', { name, pos: props.pos }).then(setValue)
```
Vtec234, line 148: Why discourage the use of `useAsync`? It has better atomicity properties than a raw `rs.call`.

Vtec234: Oh, I see we made a release of `vscode-lean4` which changed the signature of `useAsync`. I will make a new widget API release and reintroduce it here.

EdAyers: Yes also `useAsync` wasn't available on the root ES module.

PR #1840, opened 2022-11-16
perf: pick cache size coprime to pointer alignment

gebner:
Expression pointers are aligned to at least 8 bytes, so `ptrAddrUnsafe (e : Expr) % 8192` can only use 1024 out of the 8192 cache entries.  Experimental evaluation suggests that most expressions pointers are congruent to 24 modulo 32, so the effective cache utilization is likely far below the 1024 entries.

This PR changes the cache size to 8192-1, which is coprime to any power-of-two alignment and should avoid this unfortunate alignment of alignments.

### DIFF
--- a/src/Lean/Util/FindExpr.lean
+++ b/src/Lean/Util/FindExpr.lean
@@ -10,7 +10,7 @@ namespace Expr
 
 namespace FindImpl
 
-abbrev cacheSize : USize := 8192
+abbrev cacheSize : USize := 8192 - 1
 
 structure State where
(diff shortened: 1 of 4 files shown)


gebner:
!bench

Kha:
Would it be unreasonable for me to hope that this will resolve the benchmark jumps as well? 😄

leodemoura:
!bench

Kha:
!bench

Kha:
!bench
:( ?

Kha:
!bench

gebner:
!bench

PR #1843, opened 2022-11-17
fix: inconsistent use of precompiled symbols from interpreter on Windows

### LINKED ISSUE #1758: Precompilation breakage on Windows
See https://github.com/JLimperg/aesop/issues/28
May or may not have something to do with this line :thinking: 
https://github.com/leanprover/lean4/blob/9fd433785b35a4a309022db05ec6074da64da209/src/library/compiler/ir_interpreter.cpp#L300

Kha:
Fixes #1758

### DIFF
--- a/src/library/compiler/ir_interpreter.cpp
+++ b/src/library/compiler/ir_interpreter.cpp
@@ -297,11 +297,18 @@ void print_value(tout const & ios, value const & v, type t) {
 
 void * lookup_symbol_in_cur_exe(char const * sym) {
 #ifdef LEAN_WINDOWS
-    HMODULE hmods[128];  // 128 modules should be enough for everyone
+    std::vector<HMODULE> hmods(128);
     DWORD bytes_needed;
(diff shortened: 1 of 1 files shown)


Kha:
That `lookup_symbol_in_cur_exe` is linear in the number of modules is of course an issue in itself, though I believe this could actually be true for Linux/macOS as well? If we knew which dynlib belonged to which module, we could use the environment to look up a declaration's module and then do a more targeted `dlsym` for that specific dynlib (if any). But let's maybe wait until this function actually turns up in benchmarks.

Kha:
And also Aesop should definitely be precompiled into a single dynlib, which would bring down the number of loaded libs from 135 to 2.

JLimperg:
> And also Aesop should definitely be precompiled into a single dynlib, which would bring down the number of loaded libs from 135 to 2.

Is this something I can do?

Kha:
No, this is a missing Lake feature. I've opened https://github.com/leanprover/lake/issues/136 to track it.

JLimperg:
All right, thanks!

PR #1844, opened 2022-11-17
fix unintuitive `calc` indentation

AdrienChampion:
This improves the way `calc` handles the indentation of its sequence of `<relation> := <proof>`.

After #1811, it was pointed out [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/indenting.20after.20.22by.22) that `calc` was not very intuitive regarding indentation, in particular *w.r.t.* `by`.

Mentored by @digama0, this PR removes the `withPosition` of the `calcStep` sub-parser, along with adding more tests for `calc` including tests that would have failed and revealed the issues highlighted in the zulip thread.

### DIFF
--- a/src/Init/NotationExtra.lean
+++ b/src/Init/NotationExtra.lean
@@ -87,9 +87,11 @@ macro:35 xs:bracketedExplicitBinders " × " b:term:35  : term => expandBrackedBi
 macro:35 xs:bracketedExplicitBinders " ×' " b:term:35 : term => expandBrackedBinders ``PSigma xs b
 end
 
+-- first step of a `calc` block
+syntax calcFirstStep := ppIndent(colGe term (" := " term)?)
 -- enforce indentation of calc steps so we know when to stop parsing them
-syntax calcStep := ppIndent(colGe term " := " withPosition(term))
-syntax calcSteps := ppLine withPosition(calcStep) ppLine withPosition((calcStep ppLine)*)
+syntax calcStep := ppIndent(colGe term " := " term)
+syntax calcSteps := ppLine withPosition(calcFirstStep) ppLine withPosition((calcStep ppLine)*)
 
 /-- Step-wise reasoning over transitive relations.
 ```
@@ -102,6 +104,23 @@ calc
 proves `a = z` from the given step-wise proofs. `=` can be replaced with any
 relation implementing the typeclass `Trans`. Instead of repeating the right-
 hand sides, subsequent left-hand sides can be replaced with `_`.
+```
+calc
+  a = b := pab
+  _ = c := pbc
+  ...
+  _ = z := pyz
+```
+It is also possible to write the *first* relation as `<lhs>\n  _ = <rhs> :=
+<proof>`. This is useful for aligning relation symbols, especially on longer:
+identifiers:
+```
+calc abc
+  _ = bce := pabce
+  _ = cef := pbcef
+  ...
+  _ = xyz := pwxyz
+```
 
 `calc` has term mode and tactic mode variants. This is the term mode variant.
 
@@ -122,6 +141,22 @@ calc
 proves `a = z` from the given step-wise proofs. `=` can be replaced with any
 relation implementing the typeclass `Trans`. Instead of repeating the right-
 hand sides, subsequent left-hand sides can be replaced with `_`.
+```
+calc
+  a = b := pab
+  _ = c := pbc
+  ...
+  _ = z := pyz
+```
+It is also possible to write the *first* relation as `<lhs>\n  _ = <rhs> :=
+<proof>`. This is useful for aligning relation symbols:
+```
+calc abc
+  _ = bce := pabce
+  _ = cef := pbcef
+  ...
+  _ = xyz := pwxyz
+```
 
 `calc` has term mode and tactic mode variants. This is the tactic mode variant,
 which supports an additional feature: it works even if the goal is `a = z'`
--- a/src/Lean/Elab/Calc.lean
+++ b/src/Lean/Elab/Calc.lean
@@ -67,15 +67,25 @@ where
     | .node i k as => return .node i k (← as.mapM go)
     | _ => set false; return t
 
-def getCalcSteps (steps : TSyntax ``calcSteps) : Array (TSyntax ``calcStep) :=
+def getCalcFirstStep (step0 : TSyntax ``calcFirstStep) : TermElabM (TSyntax ``calcStep) :=
+  match step0  with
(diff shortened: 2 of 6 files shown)


leodemoura:
> Mentored by @digama0, this PR removes the withPosition of the calcStep sub-parser, along with adding more tests for calc including tests that would have failed and revealed the issues highlighted in the zulip thread.

The new style seems unintuitive too. I found a few places where users use the following style in Mathlib (3)
```
calc <term> = <term>
        ... = <term>
        ... = <term>
```
It feels natural to convert these proofs to Lean 4 as
```
calc <term> = <term>
          _ = <term>
          _ = <term>
```

AdrienChampion:
I agree, but it seemed to us like rewriting

```
calc <term> = <term>
          _ = <term>
          _ = <term>
```

as

```
calc <term> = <term>
     _      = <term>
     _      = <term>
```

sounded like a fair compromise to avoid having to indent proofs with the `:=`, which feels like severe indentation drifting.

AdrienChampion:
Just added a test for `calc` + `by` showing that this version avoids the indentation drifting

leodemoura:
> sounded like a fair compromise to avoid having to indent proofs with the :=, which feels like severe indentation drifting.

The severe indentation drifting was created by #1811, correct?

AdrienChampion:
If you're talking about `by` in particular, then yes. But the `withPosition` in `calcStep` was already there.

So function application for instance requires indenting to the `:=` (to the token right after, actually):

```lean
variable (t1 t2 t3 t4 : Nat)
variable (pf12 : t1 = t2) (pf23 : t2 = t3) (pf34 : t3 = t4)

theorem foo : t1 = t4 :=
  calc
    t1 = t2 := id
      pf12
    _ = t3 := pf23
    _ = t4 := pf34
```

which fails with

```text
❯ lake build
Building CalcBug
error: > LEAN_PATH=./build/lib DYLD_LIBRARY_PATH=/Users/adrien/.elan/toolchains/leanprover--lean4---nightly-2022-10-28/lib:./build/lib /Users/adrien/.elan/toolchains/leanprover--lean4---nightly-2022-10-28/bin/lean ./././CalcBug.lean -R ././. -o ./build/lib/CalcBug.olean -i ./build/lib/CalcBug.ilean -c ./build/ir/CalcBug.c
error: stdout:
./././CalcBug.lean:8:4: error: expected ':='
error: external command `/Users/adrien/.elan/toolchains/leanprover--lean4---nightly-2022-10-28/bin/lean` exited with code 1
```

AdrienChampion:
But with this being said, I agree that if you want to keep the `withPosition` in `calcStep` then we probably need to either revert #1811 or have some special handling of `by` in `calc` as @Kha mentioned in the zulip thread.

AdrienChampion:
> > sounded like a fair compromise to avoid having to indent proofs with the :=, which feels like severe indentation drifting.
> 
> The severe indentation drifting was created by #1811, correct?

In case that's helpful, I compiled a few examples that show different styles and problems and labeled them with whether they compile on

- current nightly
- this PR
- pre #1811

leodemoura:
> In case that's helpful, I compiled gathered a few examples that show different styles and problems and labeled them with whether they compile on

Thanks, this is very useful.

Kha:
I believe we said that we should document the pros and cons of the three styles (master, PR, yours) and ask for feedback from the community. An implementation is not absolutely necessary for that.

AdrienChampion:
I created a zulip thread to try to gather people's opinions:

- https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.60calc.60.20indentation

AdrienChampion:
> I created a zulip thread to try to gather people's opinions:
> 
>     * https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.60calc.60.20indentation

I'm not sure how productive the discussion has been so far. There seems to be interest in "Mario's Crazy Idea", *i.e.* this kind of syntax:

```lean
calc longId t1
  _ = t2 := pf12
  _ < t3 := pf23
  _ = t4 := pf34
  _ < t5 := pf45
```

AdrienChampion:
I ended up implementing support for this syntax [in a separate branch](https://github.com/AdrienChampion/lean4/commit/ba391209f9ae1362393c76f1c992570b5d6968ff). I'm trying to get more feedback [in the zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.60calc.60.20indentation/near/319560721) and will update accordingly

AdrienChampion:
We collected feedback, and allowing underscores in the first `calc` relation seems to be a good solution to most people who answered. Which is not a lot of people, to be clear.

I implemented the alternative syntax [here](https://github.com/leanprover/lean4/commit/df2cede4af657c27984cfb8f12f79a2e5da73168) (term and tactic mode) and added basic tests [here](https://github.com/leanprover/lean4/commit/673c07c2166018db4cd961f5b7588186236b9323). The idea is to merge these commits into this PR, if that's fine with everyone.

gebner:
> We collected feedback, and allowing underscores in the first `calc` relation seems to be a good solution to most people who answered. Which is not a lot of people, to be clear.

I like to think that's a positive sign.  People are not railing against the new syntax, which means they like it. 😄 

> I implemented the alternative syntax [here](https://github.com/leanprover/lean4/commit/df2cede4af657c27984cfb8f12f79a2e5da73168) (term and tactic mode) and added basic tests [here](https://github.com/leanprover/lean4/commit/673c07c2166018db4cd961f5b7588186236b9323). The idea is to merge these commits into this PR, if that's fine with everyone.

Yes, please update the PR!  I think it's been stuck in the design phase for long enough now, let's see how it does in the wild.

AdrienChampion:
@gebner done!

gebner:
LGTM.  Please squash everything into a single commit!

AdrienChampion:
> LGTM. Please squash everything into a single commit!

Done!

Kha:
There are some expected regressions in TPiL4 such as at https://github.com/leanprover/theorem_proving_in_lean4/blob/master/inductive_types.md?plain=1#L842-L843 ([full log](https://github.com/leanprover/theorem_proving_in_lean4/actions/runs/4261389609/jobs/7415655186)). Do we want to use and explain the Mario underscore form there? @AdrienChampion Would you like to take this on?

Kha:
We should also have a `RELEASES.md` entry for this change

AdrienChampion:
> We should also have a `RELEASES.md` entry for this change

#2124

PR #1849, opened 2022-11-18
chore: remove remnants of C++ `format`

Kha:


Kha:
We can remove `pp_expr` as well after removing the old compiler

PR #1855, opened 2022-11-18
Add lake to the nix package of lean

HamelinDavid:
This commits add lake to the nix version of lean.

This fixes `TODO: Add Lake` in `nix/bootstrap.nix`. 

While lake isn't used to build lean packages on nix, it is still used by the lean4 extension of vscode when opening projects with multiple files.

This mean vscode on nix currently doesn't work with multi-file projects. This isn't a big deal on most distributions because one can just install lean with lake included through elan, but ~elan doesn't work on NixOS~ (edit: it does).

I added a new flake input because currently git submodules don't work well with nix flakes.

I created a [demo lean 4 project repository](https://github.com/HamelinDavid/lean4-nix-demo) to test this patch; you can clone it to check that it does indeed work.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,6 +8,10 @@
     url = github:leanprover/lean4-mode;
     flake = false;
   };
+  inputs.lake = {
+    url = github:leanprover/lake;
+    flake = false;
+  };
   # used *only* by `stage0-from-input` below
   #inputs.lean-stage0 = {
   #  url = github:leanprover/lean4;
@@ -24,7 +28,7 @@
         # for `vscode-with-extensions`
         config.allowUnfree = true;
       };
-      lean-packages = pkgs.callPackage (./nix/packages.nix) { inherit nix lean4-mode; };
+      lean-packages = pkgs.callPackage (./nix/packages.nix) { inherit nix lean4-mode lake; };
     in {
       packages = lean-packages // rec {
         debug = lean-packages.override { debug = true; };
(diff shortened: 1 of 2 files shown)


Kha:
This is supposed to work out of the box, and it does for me using `nix shell .#lean-dev -c code .`. I'm inclined to drop `vscode-dev` in favor of that as it's always out of date and the inability to install additional extensions means it's mostly just a toy demo.

> but elan doesn't work on NixOS

It certainly does, it's in Nixpkgs.

HamelinDavid:
> > but elan doesn't work on NixOS
> 
> It certainly does, it's in Nixpkgs.

My bad, I wrongly assumed it wouldn't work since most package manager don't work on NixOS.

This makes this patch less important, but it's still nice having pure nix development environments (and it seems to have been planned anyway, since there was a `TODO: add Lake` in `bootstrap.nix`).

Kha:
The TODO will be resolved when Nix gains proper support for submodules. We should not have independent pinned commits of Lake in both .gitmodules and flake.lock. The only real advantage of adding Lake is that the CI can test it. For users of the Nix setup, the Lake adapter included in `lean-dev` is sufficient.

HamelinDavid:
> The TODO will be resolved when Nix gains proper support for submodules. We should not have independent pinned commits of Lake in both .gitmodules and flake.lock. The only real advantage of adding Lake is that the CI can test it. For users of the Nix setup, the Lake adapter included in `lean-dev` is sufficient.

Got it, closing this then

PR #1862, opened 2022-11-21
fix: traceSimpCall: exclude hyps from simp_all args 

### LINKED ISSUE #1853: simp_all [h], where h is a hypothesis, removes h
It seems like the `simp_all` logic that prevents hypotheses from self-simplifying to `True` does not fire if the hypothesis is given explicitly:

```lean
example (h : n = m) : n = 0 := by
  simp_all
  -- n m : ℕ
  -- h : n = m
  -- ⊢ m = 0

example (h : n = m) : n = 0 := by
  simp_all [h]
  -- n m : ℕ
  -- ⊢ m = 0
```

I discovered this because `simp_all?` suggests calls of the form `simp_all only [h]`:

```lean
import Mathlib

example (h : n = m) : n = 0 := by
  simp_all?
  -- Try this: simp_all only [h]
```

JLimperg:
fixes #1853

---

I've also taken the liberty to split `traceSimpCall` into two functions; this way I don't have to copy-paste it for Aesop.

### DIFF
--- a/src/Lean/Elab/Tactic/Simp.lean
+++ b/src/Lean/Elab/Tactic/Simp.lean
@@ -249,7 +249,14 @@ register_builtin_option tactic.simp.trace : Bool := {
   descr    := "When tracing is enabled, calls to `simp` or `dsimp` will print an equivalent `simp only` call."
 }
 
-def traceSimpCall (stx : Syntax) (usedSimps : UsedSimps) : MetaM Unit := do
+/--
+If `stx` is the syntax of a `simp`, `simp_all` or `dsimp` tactic invocation, and
+`usedSimps` is the set of simp lemmas used by this invocation, then `mkSimpOnly`
+creates the syntax of an equivalent `simp only`, `simp_all only` or `dsimp only`
+invocation.
+-/
+def mkSimpOnly (stx : Syntax) (usedSimps : UsedSimps) : MetaM Syntax.Tactic := do
+  let isSimpAll := stx[0].getAtomVal == "simp_all"
   let mut stx := stx
   if stx[3].isNone then
     stx := stx.setArg 3 (mkNullNode #[mkAtom "only"])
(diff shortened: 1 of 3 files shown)


src/Lean/Elab/Tactic/Simp.lean
```diff
@@ -249,7 +249,14 @@ register_builtin_option tactic.simp.trace : Bool := {
...
 
-def traceSimpCall (stx : Syntax) (usedSimps : UsedSimps) : MetaM Unit := do
+/--
+If `stx` is the syntax of a `simp`, `simp_all` or `dsimp` tactic invocation, and
+`usedSimps` is the set of simp lemmas used by this invocation, then `mkSimpOnly`
+creates the syntax of an equivalent `simp only`, `simp_all only` or `dsimp only`
+invocation.
+-/
+def mkSimpOnly (stx : Syntax) (usedSimps : UsedSimps) : MetaM Syntax.Tactic := do
+  let isSimpAll := stx[0].getAtomVal == "simp_all"
```
leodemoura, line 259: We usually check the `SyntaxNodeKind`.

src/Lean/Elab/Tactic/Simp.lean
```diff
@@ -263,6 +270,14 @@ def traceSimpCall (stx : Syntax) (usedSimps : UsedSimps) : MetaM Unit := do
       if env.contains declName && !simpOnlyBuiltins.contains declName then
         args := args.push (← `(Parser.Tactic.simpLemma| $(mkIdent (← unresolveNameGlobal declName)):ident))
     | .fvar fvarId => -- local hypotheses in the context
+      if isSimpAll then
+        continue
```
leodemoura, line 274: `simp_all` only uses nondependent hypotheses.

JLimperg: If I read `initEntries` right, it uses all hypotheses but only adds nondependent ones as entries to simplify. However, trying to test this, I'm getting slightly mysterious (to me) behaviour:

```lean
example {P : Nat → Type} {Q : (n m : Nat) → n = m → Type} (h₁ : n = m) (h₂ : P n) : Q n m h₁ := by
  simp_all
  /-
  ...
  h₁ : n = m
  h₂ : P n
  ⊢ Q m m (_ : m = m)
  -/
```

So it used the dependent `h₁` to rewrite in the target, but not in the nondependent `h₂`. Is this intended? Maybe it's just getting too late for me...

semorrison:
@JLimperg, what is the status of this PR? Sorry, it languished for a long time, and there have been changes to simp_all and to simp tracing in the meantime, so I would like to get a status update before looking at it more.

JLimperg:
The original motivation for this PR is not relevant any more since `simp_all` no longer self-simplifies hyps. But I still think hyps should be excluded from the `simp_all only` list generated by `simp_all?`. See #2953. If you agree with the analysis there, I'd be happy to resurrect this PR.

JLimperg:
Superseded by #2969

PR #1866, opened 2022-11-21
feat: infer def/theorem DefKind for `let rec`

digama0:
Since `let rec` has no place to put the `def`/`theorem` annotation, we have to infer it. Currently uses of `let rec` inside instances or defs to construct proofs will cause the `defLemma` mathlib linter to trigger (this occurs in practice in mathlib, because as it turns out we were actually doing `let rec` in lean 3 using the `_match` hack).

### DIFF
--- a/src/Lean/Elab/MutualDef.lean
+++ b/src/Lean/Elab/MutualDef.lean
@@ -563,7 +563,7 @@ private def mkLetRecClosureFor (toLift : LetRecToLift) (freeVars : Array FVarId)
       xs.mapM fun x => do
         let localDecl ← x.fvarId!.getDecl
         return (localDecl.userName, localDecl.binderInfo)
-    /- Auxiliary map for preserving binder user-facind names and `BinderInfo` for types. -/
+    /- Auxiliary map for preserving binder user-facing names and `BinderInfo` for types. -/
     let mut userNameBinderInfoMap : FVarIdMap (Name × BinderInfo) := {}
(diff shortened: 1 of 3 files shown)


src/Lean/Elab/MutualDef.lean
```diff
@@ -639,11 +639,17 @@ def pushMain (preDefs : Array PreDefinition) (sectionVars : Array Expr) (mainHea
...
-def pushLetRecs (preDefs : Array PreDefinition) (letRecClosures : List LetRecClosure) (kind : DefKind) (modifiers : Modifiers) : Array PreDefinition :=
-  letRecClosures.foldl (init := preDefs) fun preDefs c =>
+def pushLetRecs (preDefs : Array PreDefinition) (letRecClosures : List LetRecClosure) (kind : DefKind) (modifiers : Modifiers) : MetaM (Array PreDefinition) :=
+  letRecClosures.foldlM (init := preDefs) fun preDefs c => do
     let type  := Closure.mkForall c.localDecls c.toLift.type
     let value := Closure.mkLambda c.localDecls c.toLift.val
-    preDefs.push {
+    -- Convert any proof let recs inside a `def` to `theorem` kind
+    let kind ← if kind.isDefOrAbbrevOrOpaque then
+      withReader (fun ctx => { ctx with lctx := c.toLift.lctx, localInstances := c.toLift.localInstances }) do
```
leodemoura, line 648: Could you please use `withLCtx` instead of `withReader`?

PR #1871, opened 2022-11-22
feat: build libleanrt.bc with -O3 alwaysinline

bollu:
This adds `libleanrt.bc`, a LLVM bitcode file of the Lean
runtime that is to be inlined. This is programatically generated.

1. This differs from the previous `libleanrt.ll`, since it produces an
   LLVM bitcode file, versus a textual IR file. The bitcode file
   is faster to parse and build an in-memory LLVM module from.
2. We build `libleanrt.bc` by adding it as a target to `shell`,
   which ensures that it is always built.

3. We eschew the need for:

```cpp
#define static extern
```

which causes breakage in the build, since changing the meaning of
`static` messes with definitions in the C++ headers.

Instead, we build `libleanrt.bc` by copying everything in
`src/include/lean/lean.h`, renaming `inline` to
`__attribute__(alwaysinline)` [which forces LLVM to generate
`alwaysinline` annotations], then running the `-O3` pass pipeline
to get reasonably optimised IR, and will be perfectly inlined
when linked into the generated LLVM code by
`src/Lean/Compiler/IR/EmitLLVM.lean`.

### DIFF
--- a/nix/bootstrap.nix
+++ b/nix/bootstrap.nix
@@ -44,12 +44,12 @@ rec {
   leancpp = buildCMake {
     name = "leancpp";
     src = ../src;
-    buildFlags = [ "leancpp" "leanrt" "leanrt_initial-exec" "shell" ];
+    buildFlags = [ "leancpp" "leanrt" "leanrt_initial-exec" "shell" "runtime_bc" ];
     installPhase = ''
       mkdir -p $out
       mv lib/ $out/
       mv shell/CMakeFiles/shell.dir/lean.cpp.o $out/lib
-      mv runtime/libleanrt_initial-exec.a $out/lib
+      mv runtime/libleanrt_initial-exec.a runtime/lean.h.bc $out/lib
     '';
   };
   # rename derivation so `nix run` uses the right executable name but we still see the stage in the build log
--- a/src/runtime/CMakeLists.txt
+++ b/src/runtime/CMakeLists.txt
@@ -18,11 +18,17 @@ target_compile_options(leanrt PRIVATE -ftls-model=local-exec)
 endif()
 
 if(LLVM)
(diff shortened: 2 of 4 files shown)


bollu:
The `macos` build fails with:
```
2022-11-22T05:10:37.2898360Z sed: -I or -i may not be used with stdin
2022-11-22T05:10:37.2914830Z make[5]: *** [runtime/libleanrt.bc] Error 1
```

I imagine this is because of the `sed` being non-GNU sed. 


The `Linux aarch64` build fails with:

```
2022-11-22T05:09:02.5125283Z In file included from runtmp/lean_inlines.cpp:8:
2022-11-22T05:09:02.5128732Z /home/runner/work/lean4/lean4/build/llvm-host/bin/../include/c++/v1/stddef.h:45:15: fatal error: 'stddef.h' file not found
2022-11-22T05:09:02.5131481Z #include_next <stddef.h>
2022-11-22T05:09:02.5133956Z               ^~~~~~~~~~
```

I need to dig down and see what's going wrong here...

bollu:
Perhaps the most cross-platform way of invoking `sed` is to use [CMake's `FILE` command ?](https://cmake.org/pipermail/cmake/2007-August/015936.html)

bollu:
Great,  I make progress on the MacOS build:

```
2022-11-22T06:47:36.1756050Z /Applications/Xcode_13.2.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/ctime:49:10: fatal error: 'time.h' file not found
2022-11-22T06:47:36.1857180Z #include <time.h>
2022-11-22T06:47:36.1958120Z          ^~~~~~~~
```

Issue #1488, opened 2022-08-16
[RFC] Pretty Printer / Code Formatting


larsk21:
I'm currently looking into improving the pretty printer and would like to hear what you think about how we should format Lean code. This is important for many potential editor features, like automatic code formatting, but also features that want to make changes to a Lean file, like automatically generating `match` cases or adding unimplemented functions of a class to an instance. Here the code formatting would allow the implementation of the feature to work on the `Syntax` level rather than on the textual level.

Pretty printing in Lean is implemented in three steps: The first step is called *delaboration* and transforms fully elaborated `Expr` terms from the Lean core into surface-level `Syntax`. After that, in the second step, the `Syntax` is turned into an output independent `Format` via *formatters*. Both of these steps are automatically created for new notation / macros (when possible) and can be customized. The resulting `Format` contains the string contents of the `Expr`/`Syntax` representation, but is not fixed to a specific formatting with a fixed target width. This is done in the last step where the *(feel free to suggest a name for this step)* decides where line breaks should be inserted to get a compact `String` representation that fits in the available width (if possible). In addition there is the *parenthesizer* between the delaborator and the formatter that inserts parantheses into a `Syntax` object where necessary.

## Delaborator: `Expr` to `Syntax`

From what I've seen so far, I don't think the delaborators require much work, but feel free to correct me here. This step is also not required for the code formatting use case.

## Formatter: `Syntax` to `Format`

This is probably the most controversial part, since each of us has different preferences on how to format (Lean) code. But this step needs work nonetheless, because some formatters are broken or output code that we would not want to have in our Lean projects. I hope we can find some clean formatting rules that most people can work with. Feel free to make suggestions on how to format specific Lean syntax and I will also post examples here when I encounter them.

Examples of broken formatters:
```lean
def foo (xs : Array Nat) :=
  Id.run do
    -- do
      -- something
    for (i, x) in xs.mapIdx (·, ·)do
      bar i x where
  bar := fun _ _ => ()
```

Formatters are either implemented as `Formatter`s or (most of the time) automatically generated. To control the automatic formatter, there are formatting primitives that can be used when defining a new notation, such as `ppLine`, which emits a hard line break, or `ppSpace`, which emits a soft line break that can be printed as either a space or a line break. I'd also like to change some of the names of these formatting primitives to make it more obvious what their purpose is.

## ?: `Format` to `String`

[...]

PR #488, opened 2021-05-25
feat: IO.FS.readDir

Kha:

Issue #369, opened 2021-03-23
Make the pretty printer usable for code formatting


Kha:
The formatter in the pretty printer already produces pretty good output for goals and also makes a reasonable attempt at preserving comments and command-level layout: https://github.com/leanprover/lean4/blob/master/tests/lean/Reformat.lean.expected.out. But it is not yet at a point where we would reformat our code with it despite the advantages of doing so:
* consistent style between different collaborators
* no more formatting discussions in PRs
* refactorings would not have to worry about preserving/producing reasonable formatting (we should definitely preserve comments though...)

Kha:
An example of a missing feature is "hanging" `do/by/fun`

leodemoura:
See https://github.com/leanprover/lean4/pull/867

leodemoura:
@Kha Should we close this issue?

Kha:
I fear I could only close it in good conscience if we used it to reformat our own codebase :) .

gebner:
From my point of view there are still several challenges ahead.  The applications I have in mind are mostly mathport / code action related, so this list might a bit biased.

#### Formatting rules for special-case formatting

The `ppLine`/etc. helpers are okay for the odd minor formatting tweak, but they can't do any complicated operations.  (And many syntax definitions are already more than 50% pp* combinators, which is not so great for readability.)

A complicated (yet common and important) syntax is the structure instance syntax.  This is used for different applications, with different formatting requirements.
1. `{}` (empty collection) should be formatted without a space.
2. `{ x := 1, y := 2 }` should be formatted on a single line.
3. `{ x := 1 y := 2 }` should get newlines and be ungrouped (i.e., the first `{` should be "hanging"):
```lean
let s := {
  x := 1
  y := 2
}
```

Particularly for 1 it would be nice if you could just add a "formatting rule" that says "format `{}` as `"{}"`".

See also `()`.  The inline/multi-line distinction comes up in many other syntaxes as well, e.g. `by simp` vs `by rw [foo] simp`.  (Unfortunately there's also things like `by conv => rw [a] rw [b] rw [c]` which are a single tactic but should maybe be multi-line.)

#### Combinators for indentation-sensitive syntax

[...]

PR #1872, opened 2022-11-22
feat: intra-line withPosition formatting

digama0:
Now that people are starting to use mathport more heavily, I think it is more important that we get the formatting right so that people don't blindly copy current formatting mistakes as though they are recommended practice.

This PR is primarily aimed at fixing the formatting of the `· tac` combinator that is ubiquitous in mathported proofs.

1. There was an extra newline after each use of `· tac`. So for example you would get:
```lean
by
  · skip
    done
  
  · · skip
      done
    
    
  · done
  
```
2. The cause of (1) is that the `·` parser uses  `cdotTk many1Indent(tactic ";"? ppLine)`, and here we can see a second issue: since it's not using `sepBy1IndentSemicolon` we get the same issues as used to be happening in `by`, e.g. `· skip skip` works, and the tactics don't have to be strictly aligned resulting in weird parses.

If we just switch out the `many1Indent` for `sepBy1IndentSemicolon` we get some improvements and some regressions:
```lean
by
  ·
    skip
    done
  · ·
      skip
      done
  · done
```
The dot now uses the same "hanging" indentation style as `by`,`try` and all the other combinators. But we'd really like it to fit on the *same* line if possible.

After investigating how to do this in `Formatter` for a while I came to the conclusion that it would be very hackish and special cased to the `·` tactic, when what we really want is a way to get `withPosition`-safe line wrapping. So instead, this implements it inside the `Format` algorithm itself.

* We add a new `Format` constructor, `Format.align (force : Bool)`. The semantics are that if the current position is less than the indentation level (such as would happen immediately after a `nest 2`), `align` will pad with spaces until the column matches the indent. If the current column is greater than the indent, then it adds a newline and indents as normal.
* When the `force` flag is disabled, `align` becomes `nil` when the enclosing group is flattened; otherwise it always pads or line-breaks to the indent. (Currently only `force := true` is being used, see below.)
* The `sepByIndent` combinator used to conservatively put a `Format.line` before the list when there is at least one newline in the list (that's the spurious newlines we are seeing above) to prevent it from breaking `withPosition` parsing. Now it can use `.align true` instead.

The results look pretty good to me, see the `formatTerm` test. Because the `align` is used in `sepByIndent` the same behavior is inherited by other parsers as well. It doesn't make too much difference with lean's default 2-space indent but if you increase the indent to 4 then you will see things like `by`, `try`, `do` also getting the same nesting behavior when they appear at the start of a line.

I had to remove the `HACK` which forcibly sets `sepByIndent` to ungrouped, because this causes there to be no indent on the last line, as in:
```
· skip
  done
·
skip
done
```
(which still parses correctly, but is clearly not desirable). This was being done in order to allow structure `{` to hang on the previous line, as a way of recovering from the forced newline that used to be there. Now the `{` is at the beginning of the line, which also looks fine and matches with the `}` that hangs at the end of the last line. (I think we have the tools to do hanging `{` properly now if we need to.)

One issue that existed before and has not been fixed in this PR is that the check for whether to insert a newline, now an `align`, at the start of `sepByIndent` is imprecise: it looks for the existence of forced newlines in the sepBy list, e.g. `skip\n skip` instead of `skip; skip`. The problem with this is that even if every tactic has a `;` after it, if the sequence is long you can get line wrapping caused by the `Format` because it is using the `fill` combinator. After the line wrapping, you might get a tactic at the indent line even though we did not use an `align` at the start, as in:
```
try pos; tac; tac; tac; tac; tac; tac; tac; tac; tac; tac; tac; 
  ohno; tac; tac; tac;
```
The `ohno` tactic is to the left of the `pos` tactic, so it is either a parse error or will be parsed as part of the surrounding block. It is not possible to fix this in the `Formatter` though because it doesn't know whether the `Format` will be line broken or not. We would either need some more `Format` primitives or a general mechanism that I have seen in some implementations of the Wadler-Leijen pretty printer to allow `choice (useIfFlattened useIfNormal : Format)` nodes.

### DIFF
--- a/src/Init/Data/Format/Basic.lean
+++ b/src/Init/Data/Format/Basic.lean
@@ -41,6 +41,19 @@ inductive Format where
   /-- A position where a newline may be inserted
   if the current group does not fit within the allotted column width. -/
   | line                : Format
+  /-- `align` tells the formatter to pad with spaces to the current indent,
+  or else add a newline if we are already past the indent. For example:
+  ```
+  nest 2 <| "." ++ align ++ "a" ++ line ++ "b"
+  ```
+  results in:
+  ```
+  . a
+    b
+  ```
+  If `force` is true, then it will pad to the indent even if it is in a flattened group.
+  -/
+  | align (force : Bool) : Format
   /-- A node containing a plain string. -/
   | text                : String → Format
   /-- `nest n f` tells the formatter that `f` is nested inside something with length `n`
@@ -70,6 +83,7 @@ namespace Format
 def isEmpty : Format → Bool
   | nil          => true
   | line         => false
+  | align _      => true
   | text msg     => msg == ""
   | nest _ f     => f.isEmpty
   | append f₁ f₂ => f₁.isEmpty && f₂.isEmpty
@@ -103,17 +117,23 @@ private structure SpaceResult where
     let r₂ := r₂ (w - r₁.space);
     { r₂ with space := r₁.space + r₂.space }
 
-private def spaceUptoLine : Format → Bool → Nat → SpaceResult
-  | nil,          _,       _ => {}
-  | line,         flatten, _ => if flatten then { space := 1 } else { foundLine := true }
-  | text s,       flatten, _ =>
-    let p := s.posOf '\n';
-    let off := s.offsetOfPos p;
+private def spaceUptoLine : Format → Bool → Int → Nat → SpaceResult
+  | nil,          _,       _, _ => {}
+  | line,         flatten, _, _ => if flatten then { space := 1 } else { foundLine := true }
+  | align force,  flatten, m, w =>
+    if flatten && !force then {}
+    else if w ≤ m then
+      { space := (m - w).toNat }
+    else
+      { foundLine := true }
+  | text s,       flatten, _, _ =>
+    let p := s.posOf '\n'
+    let off := s.offsetOfPos p
     { foundLine := p != s.endPos, foundFlattenedHardLine := flatten && p != s.endPos, space := off }
-  | append f₁ f₂, flatten, w => merge w (spaceUptoLine f₁ flatten w) (spaceUptoLine f₂ flatten)
-  | nest _ f,     flatten, w => spaceUptoLine f flatten w
-  | group f _,    _,       w => spaceUptoLine f true w
-  | tag _ f,      flatten, w => spaceUptoLine f flatten w
+  | append f₁ f₂, flatten, m, w => merge w (spaceUptoLine f₁ flatten m w) (spaceUptoLine f₂ flatten m)
+  | nest n f,     flatten, m, w => spaceUptoLine f flatten (m - n) w
+  | group f _,    _,       m, w => spaceUptoLine f true m w
+  | tag _ f,      flatten, m, w => spaceUptoLine f flatten m w
 
 private structure WorkItem where
   f : Format
@@ -125,10 +145,13 @@ private structure WorkGroup where
   flb     : FlattenBehavior
   items   : List WorkItem
 
-private partial def spaceUptoLine' : List WorkGroup → Nat → SpaceResult
-  |   [],                         _ => {}
-  |   { items := [],    .. }::gs, w => spaceUptoLine' gs w
-  | g@{ items := i::is, .. }::gs, w => merge w (spaceUptoLine i.f g.flatten w) (spaceUptoLine' ({ g with items := is }::gs))
+private partial def spaceUptoLine' : List WorkGroup → Nat → Nat → SpaceResult
+  |   [],                         _,   _ => {}
+  |   { items := [],    .. }::gs, col, w => spaceUptoLine' gs col w
+  | g@{ items := i::is, .. }::gs, col, w =>
+    merge w
+      (spaceUptoLine i.f g.flatten (w + col - i.indent) w)
+      (spaceUptoLine' ({ g with items := is }::gs) col)
 
 /-- A monad in which we can pretty-print `Format` objects. -/
 class MonadPrettyFormat (m : Type → Type) where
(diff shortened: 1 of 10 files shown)


Kha:
Mmh, this does look a bit complicated for a single tactic. Don't we have basically all the tools to emulate `sepBy1Indent` with correct formatting in `syntax`?
```
syntax cdotTk ppHardSpace withPosition(sepBy1(colGe tactic, "", "; " <|> (colEq linebreak ppLine), allowTrailingSep)) : tactic
```
It looks like the `ppLine` is ignored, but I'm not quite sure why...

gebner:
> Mmh, this does look a bit complicated for a single tactic.

I had hoped that we could use this combinator in a lot of cases where we have to resort to `ppLine` now (and still have bugs!).  See e.g. https://github.com/leanprover/lean4/issues/1488#issuecomment-1219680415 for the class of issues this would fix:
```lean
[let x := 42
 id x]
```
The above doesn't round-trip atm.

Top-level structure instances would also look a lot nicer, right now they're printed like this:
```lean
{
  a := 42
  b := 8 }
```

gebner:
I've wanted such an `align` combinator for a long time btw: https://github.com/leanprover/lean4/issues/369#issuecomment-1017837443

Kha:
I hate top-level parentheses :smile: . It's so awkward that they introduce non-standard single-character indentation. But it's not like I have a better formatting in mind.

The hanging initial token looks like the core issue to me. If we only had top-level parentheses, I assume we could simply indent by 1 for them. I'm also concerned about Mario's last example, I can't tell yet whether this `align` is still the way to go for that issue as well.

Kha:
I should also mention that I'm working on integrating Lars' refactored `Format.pretty` mentioned in #488 into Lean. The current implementation is unmaintainable to me, and we do need to extend it going forward.

digama0:
> The hanging initial token looks like the core issue to me. If we only had top-level parentheses, I assume we could simply indent by 1 for them. Thus why I'm saying we shouldn't need any new tools for cdot, which never hangs.

The issue is that you have to think about how many characters there will be in the previous token, as well as what the indent size is, to implement an equivalent of `.align` for things like the formatter implementation of `· tac`. And `by tac`. And `(tac`. And `try tac`. Maybe we shouldn't have made indent size an option if we were planning to hard-code it... And if the previous syntax has a newline then you have to scrap the whole approach and do something else. That's why `.align` is better, it just transparently does the right thing in all cases, and the `Formatter` can just focus on whether it wants to output aligned code or not.

digama0:
> I'm also concerned about Mario's last example, I can't tell yet whether this `align` is still the way to go for that issue as well.

I forgot to mention: there is a simple fix for that issue with this PR, namely to remove buggy newline reasoning and just always add an align at the start of a `sepByIndent`. The downside is that
```lean
try
  tac
```
will always be line-broken like this, as well as other one-line tactic combinators like `case =>` etc, even though in the majority of cases they don't need it. For the immediate (mathport) concern I think that's a worthwhile tradeoff, especially considering that the formatter already does line wrapping incorrectly for mathlib, which has a 100 character limit not 120.

Kha:
> it just transparently does the right thing in all cases

That's great if you primarily care about correct output, and maybe we should have it just for that, but evidently it is not sufficient to get *good* output: https://github.com/leanprover/lean4/pull/1872/files#diff-6bde8d5882f415e03ae56f70a39d8d0bed774302649baa5934c1962f024fdd54R40.

My current thinking is:
* For the rare cases like
```
. tac1
  tac2
```
and
```
(tac1
 tac2)
```
where we really want to "right-append" a block to a token, we need to indent by the token(+whitespace) length (which we can have a helper combinator do), which should be sufficient as long as we're not in a hanging context. And I really don't ever want to see something like
```
arbitrarilyLongPrefix (tac1
                       tac2)
```
If `align` (used *in front* of the token) is the most convenient way to avoid a hanging context, great, let's use it.
* In the more common case like `try` where we want to insert a newline in between token and tactic block iff there is a newline in the *formatted* tactic block, `"try" ppRealGroup(ppSpace tac1 "; " tac2)` really should be sufficient, that's exactly what it's supposed to do. Am I missing something here?
* Finally, the structure instances... I'm getting convinced that having two different formatting styles dependent on context is just weird, so I'm inclined to say we should just use
```
{
  f1
  f2
}
```
everywhere where a single line is not sufficient, even in non-hanging contexts.

Kha:
> Maybe we shouldn't have made indent size an option if we were planning to hard-code it

The indentation option is far older than any plans to use Format for actual code formatting. But I agree that we may as well remove it to make our lives easier.

> especially considering that the formatter already does line wrapping incorrectly for mathlib, which has a 100 character limit not 120

Why is that, is the option getting lost somewhere?

gebner:
This PR has some pretty ugly regressions on mathport:
```lean
-def lsmul : A →ₐ[R] Module.EndCat R M where
+def lsmul :
+    A →ₐ[R] Module.EndCat R
+        M where 
```
(Also note the trailing space on the last line now.)

digama0:
The trailing `where` is I think a separate bug, it was already triggering before under different conditions. I was aware of the potential for trailing spaces but I didn't think it would be too important since most users are set up to strip trailing whitespace (we can even do that in mathport postprocessing if necessary).

I'm not sure what causes the crazy indentation in your example though; everything should be well below the line wrap threshold. What's the generated `Format` object?

gebner:
The trailing spaces are because `.align` isn't guaranteed to insert a space.  The crazy indentation is probably because the align combinator is not resetting the `isUngrouped` flag.

I think it would be best to make align behave more like the whitespace combinators.  It always resets the lead word, and always results in some kind of whitespace (separate line or space).

digama0:
Although I think the no-space version of align is useful for some things, it's not important for our immediate goals so I'm fine with replacing it with a version that always uses at least one space for now.

Testing the change is going to be really annoying though. Do you have an idea for how to turn that example into a MWE that can go in `formatTerm.lean`?

PR #1877, opened 2022-11-23
chore: bundle libatomic in releases

gebner:
This should make `lean +gebner/lean4:bundlelibatomic` work again (once CI finishes after lunch).

### DIFF
--- a/script/prepare-llvm-linux.sh
+++ b/script/prepare-llvm-linux.sh
@@ -25,6 +25,9 @@ cp -L llvm/bin/llvm-ar stage1/bin/
 # dependencies of the above
 $CP llvm/lib/lib{clang-cpp,LLVM}*.so* stage1/lib/
 $CP $ZLIB/lib/libz.so* stage1/lib/
+# bundle libatomic (referenced by LLVM >= 15, and required by the lean executable to run)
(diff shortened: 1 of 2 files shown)


Kha:
LGTM modulo first commit

gebner:
> modulo first commit

This is a CI feature that I've wanted for a long time, I'm really tired of naming all tags like `vrpcrace` etc.

gebner:
Okay, I've validated the fix.  The downloaded binaries now work on both NixOS and Ubuntu.

gebner:
Ah I see.  It's unfortunate that uploading binaries is tied to appearing on the releases tab. :disappointed:  Setting it as prerelease or draft doesn't help either.

Note that we can always remove a release manually if we don't want it to show up in the list (but then you can't use elan on it either :sad:).

> We don't want https://github.com/leanprover/lean4/releases/tag/NFM2022 etc. to appear as a release

We probably don't want it to show up in the list, but still available via elan.

Kha:
> We probably don't want it to show up in the list, but still available via elan.

So far we only really used them in links to code, in the actual supplements we use regular nightly releases as all those tags are on `master`.

PR #1879, opened 2022-11-24
Update Lake to latest prerelease

gebner:
Several file locations have changed, requiring adjustments to `.gitignore` files in projects using Lake.  The manifest is now in `lake-manifest.json` (instead of `lean_packages/manifest.json`), and the dependencies are fetched into the `lake-packages` directory (instead of `lean_packages`).


gebner:
This also fixes the issue that broke mathport CI when we added aesop to the mathlib4 dependencies.

PR #1885, opened 2022-11-26
Work around VS Code completion bug

### LINKED ISSUE #1858: No autocompletion when typing even number of characters in VSCode
### Prerequisites

### Description

With VSCode, typing an even number of characters does not trigger autocompletion (the Lean 4 server produces no results). If the user has "Word Based Suggestions" in VSCode enabled, the editor will fall back to such suggestions, preventing any further attempts at intelligent autocompletion from the Lean 4 server.

### Steps to Reproduce

````lean
#check B -- completion results on typing 'B'
#check Ba -- if typed together, no results; if typed as 'B' then after results, 'a', then results
#check Bas --  if typed together, results; if typed as 'B' then 'as', no results
#check Base -- if typed together or as 'Ba' then 'se', no results; if typed as 'B' then 'ase' or  'Bas' then 'e', results
````

**Expected behavior:**

Autocompletion regardless of the number of characters typed.

**Actual behavior:**

No autocompletion when typing an even number of characters.

**Reproduces how often:**

Always.

### Versions

VSCode 1.73.1
Windows 10 20H2
`Lean (version 4.0.0-nightly-2022-11-10, commit 2386c401d231, Release)`

### Additional Information

Bug first reported on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Bug.3A.20No.20autocomplete.20on.20even.20chars) and confirmed by @Kha and Kevin Buzzard.

### PR DESCRIPTION
**Pull request by Kha**
Builds on top of #1884
Fixes #1858

I consider this a clear bug on VS Code's part: https://github.com/microsoft/vscode/issues/155738

### DIFF
--- a/src/Lean/Server/FileWorker/RequestHandling.lean
+++ b/src/Lean/Server/FileWorker/RequestHandling.lean
@@ -34,7 +34,10 @@ def handleCompletion (p : CompletionParams)
   -- such as a trailing dot after an option name. This shouldn't be a problem since any subsequent
   -- command starts with a keyword that (currently?) does not participate in completion.
   withWaitFindSnap doc (·.endPos + ' ' >= pos)
-    (notFoundX := pure { items := #[], isIncomplete := true }) fun snap => do
+    (notFoundX := pure { items := #[], isIncomplete := true })
+    (abortedX :=
+      -- work around https://github.com/microsoft/vscode/issues/155738
+      pure { items := #[{label := "-"}], isIncomplete := true }) fun snap => do
       if let some r ← Completion.find? doc.meta.text pos snap.infoTree caps then
         return r
       return { items := #[ ], isIncomplete := true }
--- a/src/Lean/Server/Requests.lean
+++ b/src/Lean/Server/Requests.lean
@@ -107,13 +107,12 @@ def bindTask (t : Task α) (f : α → RequestM (RequestTask β)) : RequestM (Re
   let rc ← readThe RequestContext
   EIO.bindTask t (f · rc)
 
-def waitFindSnapAux (notFoundX : RequestM α) (x : Snapshot → RequestM α)
+def waitFindSnapAux (notFoundX abortedX : RequestM α) (x : Snapshot → RequestM α)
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/Lean/Server/FileWorker/RequestHandling.lean [hunk 1]
```diff
@@ -34,7 +34,9 @@ def handleCompletion (p : CompletionParams)
   -- such as a trailing dot after an option name. This shouldn't be a problem since any subsequent
   -- command starts with a keyword that (currently?) does not participate in completion.
   withWaitFindSnap doc (·.endPos + ' ' >= pos)
-    (notFoundX := pure { items := #[], isIncomplete := true }) fun snap => do
+    (notFoundX :=
+      -- work around https://github.com/microsoft/vscode/issues/155738
+      pure { items := #[{label := "-"}], isIncomplete := true }) fun snap => do
```
Kha, line 39: Note that the the completion item should not be observable in practice as a follow-up request must already be received to end up in this branch (or when our snapshots are screwed up).

src/Lean/Server/Requests.lean
```diff
@@ -110,10 +110,9 @@ def bindTask (t : Task α) (f : α → RequestM (RequestTask β)) : RequestM (Re
 def waitFindSnapAux (notFoundX : RequestM α) (x : Snapshot → RequestM α)
     : Except ElabTaskError (Option Snapshot) → RequestM α
   /- The elaboration task that we're waiting for may be aborted if the file contents change.
-  In that case, we reply with the `fileChanged` error. Thanks to this, the server doesn't
+  In that case, we say there are no snapshots. Thanks to this, the server doesn't
   get bogged down in requests for an old state of the document. -/
-  | Except.error FileWorker.ElabTaskError.aborted =>
-    throwThe RequestError RequestError.fileChanged
+  | Except.error FileWorker.ElabTaskError.aborted => notFoundX
```
gebner, line 115: We should only change the behavior here for the completion.  For example, the infoview relies on the file-changed error to retry.

gebner (changes requested):

PR #1889, opened 2022-11-27
feat: add HashSet.insertMany

kim-em:
This had been added separately in aesop and in mathlib4, and once we imported both sources in the same place there was a name clash. This suggests it's time to move it to the correct place.


digama0:
Note: if lean4 doesn't want to take this definition, std can take it in `Std.Lean.HashMap` (where additions to Lean's `HashMap` type will be stored).

PR #1890, opened 2022-11-27
Optimizing multiplication and division operations of powers of 2 into shift operations.

hargoniX:


hargoniX:
I don't see why the Linux CI is failing? It is complaining that a function is missing which is very clearly there.

Kha:
No, `shiftRight` does not exist yet at `Init/Data/Nat/Bitwise.lean:16`. You're trying to optimize the implementation of shifting with itself :) . The optimization should do nothing if the constant is not in the environment yet.

Kha:
Btw, any reason not to leave these optimizations to LLVM? Not sure whether for `Nat` GMP implements the optimization dynamically, but I'd expect so.

hargoniX:
Yeah we can leave this optimization to LLVM as well I guess. In that case do you think the log2 for UInt's is still useful in core? Or should I move it over into std4 instead?

Kha:
In the maintainers meeting we decided there is no real downside to this PR, so happy to merge after it's fixed!

PR #1895, opened 2022-11-29
chore: protect Prod.Lex

kim-em:
mathlib4 also defines a `Lex`, and in porting people have encountered collisions with this one when `Prod` is open. Protecting `Prod.Lex` in core is not essential, but would be convenient for mathlib4.

### DIFF
--- a/src/Init/WF.lean
+++ b/src/Init/WF.lean
@@ -208,9 +208,9 @@ variable  (ra  : α → α → Prop)
 variable  (rb  : β → β → Prop)
 
 -- Lexicographical order based on ra and rb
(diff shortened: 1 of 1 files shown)


gebner (approved):
For context, in mathlib `Lex α` is a type wrapper.  `Lex (α × β)` is the type of lexicographically ordered pairs (with a *global* `LT` instance, etc.).

gebner:
Zulip thread: https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.60Prod.2ELex.60.20vs.20.60Lex.60/near/312872860

PR #1896, opened 2022-11-30
Support dashes in field names

gebner:
The motivation is JSON serialization obviously.  The Lean field names need to match the JSON field names, and some people use dashes there. :shrug:

### DIFF
--- a/src/Lean/Elab/Deriving/FromToJson.lean
+++ b/src/Lean/Elab/Deriving/FromToJson.lean
@@ -14,10 +14,10 @@ open Lean.Json
 open Lean.Parser.Term
 open Lean.Meta
 
-def mkJsonField (n : Name) : Bool × Term :=
-  let s  := n.toString
+def mkJsonField (n : Name) : CoreM (Bool × Term) := do
+  let .str .anonymous s := n | throwError "invalid json field name {n}"
   let s₁ := s.dropRightWhile (· == '?')
-  (s != s₁, Syntax.mkStrLit s₁)
+  return (s != s₁, Syntax.mkStrLit s₁)
 
 def mkToJsonInstanceHandler (declNames : Array Name) : CommandElabM Bool := do
   if declNames.size == 1 then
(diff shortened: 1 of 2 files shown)


src/Lean/Elab/Deriving/FromToJson.lean
```diff
@@ -27,9 +27,10 @@ def mkToJsonInstanceHandler (declNames : Array Name) : CommandElabM Bool := do
         let header ← mkHeader ``ToJson 1 ctx.typeInfos[0]!
         let fields := getStructureFieldsFlattened (← getEnv) declNames[0]! (includeSubobjectFields := false)
         let fields ← fields.mapM fun field => do
-          let (isOptField, nm) := mkJsonField field
-          if isOptField then ``(opt $nm $(mkIdent <| header.targetNames[0]! ++ field))
-          else ``([($nm, toJson $(mkIdent <| header.targetNames[0]! ++ field))])
```
gebner, line 32: This was appending a name to a hygienic name.  I'm surprised that it worked at all.

leodemoura: I gave up the "private names as macro scopes" branch. But, it may be useful to merge
https://github.com/leanprover/lean4/commit/0665473629afd34e15edb2d5626da2631e904e3c#diff-52ef0c67eea613acf6c0b6284063fd7112cdd40750de38365c214abaef246db4R4219-R4225
What do you think?

gebner: The `Append Name` change is the one I'm most lukewarm about.  We're switching from one set of ugly corner cases to another set of ugly corner cases.  Right now, hygienic structure name ++ field name is broken, but namespace ++ hygienic definition name works.  With the change, the breakage is simply reversed.

I'm not particularly motivated to put effort into such a halfhearted fix.  If we're refactoring names, then I'd like at least some compile-time help from the type system.  Right now it is a huge amount of guesswork to figure out if a Name is a simply string literal (field name), has scopes or doesn't, etc., and we're all getting it constantly wrong.

So ideally we'd use different types, say `String` (for name components like field names), `CoreName` (without scopes), `Name` (with scopes).  Then we can safely add a `HAppend Name String` or `HAppend Name CoreName` instance that always does the right thing.

leodemoura: > We're switching from one set of ugly corner cases to another set of ugly corner cases. Right now, hygienic structure name ++ field name is broken, but namespace ++ hygienic definition name works. With the change, the breakage is simply reversed.

We can support both cases. The only problematic case is `hygienic ++ hygienic`

> I'm not particularly motivated to put effort into such a halfhearted fix.

I was volunteering to merge the `Append Name` in the commit above. I can also add support for `name ++ hygienic`.

> If we're refactoring names, then I'd like at least some compile-time help from the type system.

I am not proposing any big refactoring. I just want to make sure that `Append Name` handles `hygienic ++ name` and `nam e ++ hygienic` correctly.

> So ideally we'd use different types, say String (for name components like field names), CoreName (without scopes), Name (with scopes). Then we can safely add a HAppend Name String or HAppend Name CoreName instance that always does the right thing.

This would be nice, but it is a big refactor. I am not motivated to do it or review a PR.

gebner: > I was volunteering to merge the `Append Name` in the commit above. I can also add support for `name ++ hygienic`.

If you're doing it, then I'm not going to complain.  I got the idea that there was still a lot of work to do there (from the talk about snowballing today).  It would be good to do a `!bench` first though.

leodemoura: > I got the idea that there was still a lot of work to do there (from the talk about snowballing today). It would be good to do a !bench first though.

Sorry for the misunderstanding. I suggested merging only the new `Append Name` instance in the commit above.

Kha: > This would be nice, but it is a big refactor. I am not motivated to do it or review a PR.

We might be able to do this incrementally by introducing new `TSyntax` kinds with appropriate extraction methods without changing `Syntax.ident` itself

gebner: At least *I* was talking about `Name`, not `Ident`.

PR #1904, opened 2022-12-01
feat: option to prettyprint fvarid instead of binder name in LCNF

ydewit:
For debugging, I often need to pretty print LCNF declarations and these declarations are not really internalized:
```
set_option trace.Compiler.result true in
#eval runOnDecls [``String.toSubstring] <|
  Probe.ppDecl
```
And here is an example of what I get:
```
def Array.getD α a i v₀ : _uniq.1 :=
  let _x.1 := @Array.size _ _uniq.2;
  let _x.2 := Nat.decLt _uniq.3 _uniq.5;
  cases _uniq.6 : _uniq.1
  | Decidable.isFalse x.3 =>
    return _uniq.4
  | Decidable.isTrue h =>
    let _x.4 := @Fin.mk _uniq.5 _uniq.3 ◾;
    let _x.5 := @Array.get _ _uniq.2 _uniq.9;
    return _uniq.10
```

One possibility is to explicitly internalize before pretty printing, but this changes the compiler state.

This PR proposes an alternative one: a new option, namely `pp.letShowFVarId`, to use fvarid's instead of binderNames when pretty printing (we always have fvarId's). We lose the pretty binder names, but we now are able to link let terms in LCNF when debugging. And without changing the compiler state.

For instance:
```
set_option pp.letShowFVarId true in
set_option trace.Compiler.result true in
def add1 (x : Nat) : Nat := x + 1
```
And the corresponding pretty-printed LCNF:
```
 def add1 _uniq.3652 : Nat :=
      let _uniq.3653 := 1;
      let _uniq.3654 := Nat.add _uniq.3652 _uniq.3653;
      return _uniq.3654
```

### DIFF
--- a/src/Lean/Compiler/LCNF/PrettyPrinter.lean
+++ b/src/Lean/Compiler/LCNF/PrettyPrinter.lean
@@ -30,11 +30,14 @@ private def prefixJoin (pre : Format) (as : Array α) (f : α → M Format) : M 
     result := f!"{result}{pre}{← f a}"
   return result
 
-def ppFVar (fvarId : FVarId) : M Format :=
-  try
-    return format (← getBinderName fvarId)
-  catch _ =>
-    return format fvarId.name
+def ppFVar (fvarId : FVarId) : M Format := do
+    if pp.letShowFVarId.get (← getOptions) then
+      return format fvarId.name
+    else
+      try
+        return format (← getBinderName fvarId)
+      catch _ =>
+        return format fvarId.name
 
 def ppExpr (e : Expr) : M Format := do
   Meta.ppExpr e |>.run' { lctx := (← read) }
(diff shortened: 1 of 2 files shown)


Kha:
> One possibility is to explicitly internalize before pretty printing, but this changes the compiler state.

Could you elaborate on this? What's wrong with internalizing purely for pretty-printing as in `ppDecls'`?

ydewit:
I may have misunderstood this the last time we spoke about this in the dev meeting: I thought it was an open issue.

```
def ppDecl' (decl : Decl) : CoreM Format := do
  runCompilerWithoutModifyingState do
    ppDecl (← decl.internalize)
```

If I can assume that ppDecl' will not modify the compiler state, as the code suggests, then I can get rid of this PR.

ydewit:
Sorry for the noise. Indeed it addresses my pain:
```
def ReaderT.read (ρ : Type u) (m : Type u → Type v) (inst.1 : Monad m) (a.2 : ρ) : m ρ :=
  let _x.3 : Applicative m := inst.1 # 0;
  let _x.4 : Pure m := _x.3 # 1;
  let _x.5 : {α : Type u} → α → m α := _x.4 # 0;
  let _x.6 : m ρ := _x.5 _ a.2;
  return _x.6
```

ydewit:
Closing since this PR is unnecessary.

PR #1913, opened 2022-12-02
chore: CI: enable emulated tests on Linux aarch64

gebner:


gebner:
Just putting this up as a PR here because CI is disabled on branches.

zwarich:
We have actual AArch64 HW available in major cloud providers. If we want to run Linux on AArch64 in addition to macOS, we probably shouldn't be using qemu at this point.

PR #1917, opened 2022-12-06
fix: fuzzy-find bonus for matching last characters of pattern and symbol

### LINKED ISSUE #1546: Unexpected workspace symbol search results

rish987:
Closes #1546. I thought it best if we simply give a constant bonus when the last character of the pattern matches the last character of the symbol, but let me know if you can think of a better heuristic.

Also something I noticed is that we don't have any tests for the fuzzy-matching algorithm directly, rather we're testing this indirectly through the completion tests. I wanted to add a test for the `aext` example but wasn't sure where to put that, will be happy to make a new fixture for it if there isn't already somewhere I can sneak it into.


Kha:
> will be happy to make a new fixture for it if there isn't already somewhere I can sneak it into.

Yeah, feel free to just `#eval` underlying functions in a new test case!

src/Lean/Data/FuzzyMatching.lean
```diff
@@ -198,13 +196,16 @@ private def fuzzyMatchCore (pattern word : String) (patternRoles wordRoles : Arr
...
-      if patternChar == wordChar || (patternRole matches CharRole.head && wordRole matches CharRole.head) then
+      if (pattern.get ⟨patternIdx⟩) == (word.get ⟨wordIdx⟩) || (patternRole matches CharRole.head && wordRole matches CharRole.head) then
         score := score + 1
-      /- Beginning of the word. -/
-      if wordStart then
+      /- Matched end of word with end of pattern -/
+      if (wordIdx == (word.length - 1)) && (patternIdx == (pattern.length - 1)) then
+        score := score + 2
+      /- Matched beginning of the word. -/
+      if (wordIdx == 0) then
```
Kha, line 208: All those parentheses can be dropped :)

Kha (approved):

rish987:
Okay, so it turns out that it wasn't as simple as calling an underlying function since the logic to get & sort the matching symbols is implemented in the watchdog itself. I could perhaps extract that to an external function but for now I've taken the easy route of modifying the test runner. LMK if this approach is acceptable -- it also uses a hack to replace URI references to within the local file system with the placeholder `file://__root__` after the test output has been produced.

rish987:
Oh, haha it looks like CI will actually need some additional tools for this hack to work... I'm not yet familiar with how things are set up for the different test environments, but all that I did was use `git` and `sed` in `run_single.sh`, perhaps we just need to install those two programs?

Kha:
> Okay, so it turns out that testing this wasn't as simple as calling an underlying function since the logic to get & sort the matching symbols is implemented in the watchdog itself

There isn't really any logic there apart from a `qsort` call and an `extract` that we can ignore for the tests, no?
```
#eval #[`Array.extract, `Lean.extractMainModule].filterMap (fun n => (n, ·) <$> fuzzyMatchScoreWithThreshold? "extract" n.toString) |>.qsort (·.2 < ·.2) |>.map (·.1)
```
Something like this would be much preferable over being reliant on actual Environment data that's constantly changing.

tests/lean/interactive/run.lean
```diff
@@ -107,6 +107,9 @@ partial def main (args : List String) : IO Unit := do
           for diag in diags do
             IO.eprintln (toJson diag.param)
           requestNo := requestNo + 1
+        | "collectInitDiagnostics" =>
+          for diag in initDiags do
+            IO.eprintln (toJson diag.param)
```
rish987, line 112: Is this okay? It looks like we were discarding the initial diagnostics (that is, we were expecting some change to the file before collecting diagnostics), so I added this command so we can print them out anyways if we want to. But perhaps the interactive tests is not the best place to put this workspace symbol search test?

Kha: > But perhaps the interactive tests is not the best place to put this workspace symbol search test?

Exactly, you can put it in `tests/lean` with the other output tests

tests/lean/interactive/completion7.lean.expected.out
```diff
@@ -15,6 +15,10 @@
     "`and x y`, or `x && y`, is the boolean \"and\" operation (not to be confused\nwith `And : Prop → Prop → Prop`, which is the propositional connective).\nIt is `@[macro_inline]` because it has C-like short-circuiting behavior:\nif `x` is false then `y` is not evaluated.\n",
     "kind": "markdown"},
    "detail": "Bool → Bool → Bool"},
+  {"label": "false_and",
+   "kind": 3,
+   "detail": "∀ (p : Prop), (False ∧ p) = False"},
+  {"label": "true_and", "kind": 3, "detail": "∀ (p : Prop), (True ∧ p) = p"},
   {"label": "AndOp",
```
Kha, line 22: @rish987 Sorry for noticing only now, but does this order suggest that the end-of-word bonus should maybe be smaller than beginning-of-word?

rish987: Yes, thanks for catching this. Fixed!

Issue #375, opened 2021-04-02
Signature pretty printer


Kha:
In hover text etc., we should not naively print the signature as a term to the right of the colon `:` but move parameters to the left of it, ensuring that parameter names are printed and that we can use special syntax for opt/autoParams.

Kha:
This is even more important after #941. Instead of
```
@Option.none.{?u.1345} : {α : Type ?u.1345} → Option α
```
we really should be printing
```
Option.none.{u} {α : Type u} : Option α
```

Kha:
@Vtec234 We probably want to make sure that the output is still usable as an interactive message for widgets, correct? That is, even if the LSP hover is currently non-interactive, if we use the signature pretty printer in e.g. `#check`, we would expect to be able to inspect any nested terms, in either parameters or the result type. So the output `Option.none.{u} {α : Type u} : Option α` should be represented by a `MessageData` values roughly as follows
```
MessageData.withContext ... <|  -- save environment
  MessageData.ofFormat f!"Option.none.{u} {α : " ++  -- insert more formatting structure here... do we want metadata on `α` as well?
  MessageData.ofExpr `(Type u) ++  -- pretend this is an `Expr` quotation
  MessageData.ofFormat f!"} : " ++
  MessageData.withContext ... <|  -- add `α` to lctx
    MessageData.ofExpr `(Option α)
```
Additional parameters would introduce additional contexts.

Vtec234:
That would be one way to do it, but maybe we just want to change the delaborator, perhaps adding a config option to select the current/new behaviour? As long as `delabCore` [here](https://github.com/leanprover/lean4/blob/d2c626e15876b33f8a309a7ab1442673252a010e/src/Lean/Widget/InteractiveCode.lean#L82) produces the right `Syntax`, I think we need not worry about how it gets turned into `MessagaData` later.

Kha:
A signature is not a valid term syntax, so this can't be part of the delaborator unconditionally. In theory, we could introduce a purely internal delaborator option for formatting a function type as a signature suffix and then apply that option to the root when build a signature message. But I don't think we can currently embed `OptionsPerPos` into `Message` for that.

Kha:
OTOH, I'm not even sure we want to use the signature printer on `#check none` since `#check` takes terms in general, not single idents. So this might be a mostly theoretical issue at this point if we only want to use it in the LSP hover.

gebner:
[...]

PR #1919, opened 2022-12-06
refactor: consolidate `MessageData` constructors into lazy formatting with infos

Kha:
In preparation of #375

### DIFF
--- a/src/Lean/Widget/InteractiveDiagnostic.lean
+++ b/src/Lean/Widget/InteractiveDiagnostic.lean
@@ -22,15 +22,20 @@ structure LazyTraceChildren where
   deriving TypeName
 
 inductive MsgEmbed where
-  | code : CodeWithInfos → MsgEmbed
+  /-- A piece of Lean code with elaboration/typing data.
+  Note: does not necessarily correspond to an `Expr`, the name is for RPC API compatibility. -/
+  | expr : CodeWithInfos → MsgEmbed
+  /-- An interactive goal display. -/
   | goal : InteractiveGoal → MsgEmbed
+  /-- Some messages (in particular, traces) are too costly to print eagerly. Instead, we allow
+  the user to expand sub-traces interactively. -/
   | trace (indent : Nat) (cls : Name) (msg : TaggedText MsgEmbed) (collapsed : Bool)
       (children : StrictOrLazy (Array (TaggedText MsgEmbed)) (WithRpcRef LazyTraceChildren))
   deriving Inhabited, RpcEncodable
 
-/-- We embed objects in LSP diagnostics by storing them in the tag of an empty subtree (`text ""`).
-In other words, we terminate the `MsgEmbed`-tagged tree at embedded objects and instead store
-the pretty-printed embed (which can itself be a `TaggedText`) in the tag. -/
+/-- The `message` field is the text of a message possibly containing interactive *embeds* of type
+`MsgEmbed`. We maintain the invariant that embeds are stored in `.tag`s with empty `.text` subtrees,
+i.e. `.tag embed (.text "")`, because a `MsgEmbed` display involve more than just text. -/
 abbrev InteractiveDiagnostic := Lsp.DiagnosticWith (TaggedText MsgEmbed)
 
 deriving instance RpcEncodable for Lsp.DiagnosticWith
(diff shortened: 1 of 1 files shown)


src/Lean/Message.lean
```diff
@@ -37,16 +38,22 @@ structure NamingContext where
   currNamespace : Name
   openDecls : List OpenDecl
 
+/-- Lazily formatted text to be used in `MessageData`. -/
+structure PPFormat where
+  /-- Pretty-prints text using surrounding context, if any. -/
+  pp : Option PPContext → IO FormatWithInfos
+  /-- Searches for synthetic sorries in original input. Used to filter out certain messages. -/
+  hasSyntheticSorry : MetavarContext → Bool := fun _ => false
```
Kha, line 46: We could probably get rid of this field in favor of "check if `pp` returns a `TermInfo` at the root and use its `expr`" in exchange for formatting terms we will not display

gebner (approved):
This is the exactly the design we discussed in #375, so I've got nothing to complain about.  Should we do a stage0 update?

src/Lean/Widget/InteractiveDiagnostic.lean
```diff
@@ -22,7 +22,7 @@ structure LazyTraceChildren where
   deriving TypeName
 
 inductive MsgEmbed where
-  | expr : CodeWithInfos → MsgEmbed
+  | code : CodeWithInfos → MsgEmbed
```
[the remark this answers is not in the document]
gebner:
```suggestion
  | expr : CodeWithInfos → MsgEmbed
```

Yeah, Wojciech is completely right.  This should really stay `expr`.

src/Lean/Widget/InteractiveDiagnostic.lean
```diff
@@ -96,15 +96,10 @@ where
 
   go : NamingContext → Option MessageDataContext → MessageData → MsgFmtM Format
   | _,    _,         ofFormat fmt             => withIgnoreTags (pure fmt)
-  | _,    _,         ofLevel u                => return format u
-  | _,    _,         ofName n                 => return format n
-  | nCtx, some ctx,  ofSyntax s               => withIgnoreTags (ppTerm (mkPPContext nCtx ctx) ⟨s⟩) -- HACK: might not be a term
```
Vtec234, line 101: This change looks a bit fishy since `withIgnoreTags` is no longer used.

Vtec234: Okay, I convinced myself that this is fine, but man this code is awfully complicated lol. I wrote a docstring trying to explain what's going on and pushed it directly here. Hope that's okay, but do let me know if you'd rather I not push directly.

src/Lean/Widget/InteractiveDiagnostic.lean
```diff
@@ -55,31 +60,45 @@ private def mkPPContext (nCtx : NamingContext) (ctx : MessageDataContext) : PPCo
...
+`Format` corresponds to an `EmbedFmt`. The kind of this tag determines how the nested `Format`
+object (possibly including further `.tag`s), is processed. For example, if the output is
+`.tag (.expr ctx infos) fmt` then tags in the nested `fmt` object refer to elements of `infos`.
+
+In the second stage, we recursively transform such a `Format` into `TaggedText MsgEmbed` according
+to the rule above by first pretty-printing it and then grabbing data referenced by the tags from
+all the nested arrays (such as the `infos` array in the example above).
+
+We cannot easily do the translation in a single `MessageData → TaggedText MsgEmbed` step because
+that would effectively require reimplementing the (stateful, to keep track of indentation)
```
Kha, line 76: @Vtec234 The pretty printer does not track indentation. Only `Format.pretty` does, but I guess that's not what you mean here.

src/Lean/Widget/InteractiveDiagnostic.lean
```diff
@@ -94,55 +113,46 @@ where
...
         let children := children.map fun child =>
           MessageData.withNamingContext nCtx <|
             match ctx with
             | some ctx => MessageData.withContext ctx child
             | none     => child
-        pure (.lazy children)
+        .lazy children
       else
-        pure (.strict (← children.mapM (go nCtx ctx)))
+        .strict (← children.mapM (go nCtx ctx))
```
gebner, line 143: This change had a subtle performance impact: #1987

PR #1920, opened 2022-12-06
Remove/restrict dangerous coercion instances

gebner:
During TC search it is absolutely critical to ensure subgoals like `Add ?α` are never generated.  That is, subgoals where there is an assignable expression metavariable in a non-out-param argument.  Such goals cause Lean to enumerate all types with (in this example) an addition operation (and that's of course infinitely many).

In mathlib4 we run the dangerousInstance linter to statically check that instances don't violate this rule.  The linter found several problematic instances in core.  This PR fixes the easy ones, where we only need to make small changes, like replacing `Coe` by `CoeHead`, and everything still compiles.

There a three other instances that the linter flags, which would be a bit harder to fix:
 1. `Coe (Fin ?n) Nat` and `Coe (Mutex ?α) BaseMutex`
 2. The `[Quote α k] [CoeHTCT (TSyntax k) (TSyntax [k'])] : Quote α k'` instance.

Ad 1. For example the `Coe (Fin ?n) Nat` conflicts with a potential `Coe (Fin ?n) (Fin (?n+1))` instance.  The reason I couldn't easily downgrade these two to `CoeHead` is because `CoeHead` is not chained.  Recall that we chain coercions like `CoeHead? Coe* CoeTail?` (and chain the `Coe` from the right).  We could change this to `CoeHead* Coe* CoeTail?`.

### DIFF
--- a/src/Init/Meta.lean
+++ b/src/Init/Meta.lean
@@ -1100,22 +1100,20 @@ instance : EmptyCollection (SepArray sep) where
 instance : EmptyCollection (TSepArray sep k) where
   emptyCollection := ⟨∅⟩
 
-/-
-We use `CoeTail` here instead of `Coe` to avoid a "loop" when computing `CoeTC`.
-The "loop" is interrupted using the maximum instance size threshold, but it is a performance bottleneck.
-The loop occurs because the predicate `isNewAnswer` is too imprecise.
--/
-instance : CoeTail (SepArray sep) (Array Syntax) where
+instance : CoeHead (SepArray sep) (Array Syntax) where
   coe := SepArray.getElems
 
-instance : Coe (TSepArray k sep) (TSyntaxArray k) where
+instance : CoeHead (TSepArray k sep) (TSyntaxArray k) where
   coe := TSepArray.getElems
 
 instance [Coe (TSyntax k) (TSyntax k')] : Coe (TSyntaxArray k) (TSyntaxArray k') where
   coe a := a.map Coe.coe
 
-instance : Coe (TSyntaxArray k) (Array Syntax) where
+instance : CoeHead (TSyntaxArray k) (Array Syntax) where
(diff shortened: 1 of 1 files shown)


digama0:
> Recall that we chain coercions like `CoeHead? Coe* CoeTail?` (and chain the `Coe` from the right). We could change this to `CoeHead* Coe* CoeTail?`.

Regarding this, I was wondering whether we could have `CoeHead? CoeOut* CoeIn* CoeTail?` where `CoeIn` chains from the right and `CoeOut` chains from the left. We currently have a bit of combinatorial explosion of typeclass instances though since we don't want the empty sequence to be a coercion (not sure how important this is given that the instance is post-processed by the coe framework).

gebner:
Practically speaking, I only know of a single `CoeTail` or `CoeHead` instance (family) that couldn't be `CoeOut` or `Coe`.  Namely `[Semiring R] → CoeTail Nat R` (and co., such as the unbundled-to-bundled morphism coercions).  And that's `CoeTail`.  So I don't think we necessarily need a `CoeHead`/`CoeOut` distinction.

> We currently have a bit of combinatorial explosion of typeclass instances though since we don't want the empty sequence to be a coercion (not sure how important this is given that the instance is post-processed by the coe framework).

I tend to agree, we only apply coercions when the types are not defeq.  Except for `↑`, but there we could check if the coercion is the identity (if we really want).

Note that we already have a `coeId` instance anyhow. 🤷

gebner:
> Practically speaking, I only know of a single `CoeTail` or `CoeHead` instance (family) that couldn't be `CoeOut` or `Coe`.

Today I've learned of another such instance: `[SetLike A B] → CoeHead A (Set B)`.  So maybe we do need the `CoeOut` after all.

gebner:
Closed in favor of #1923.

PR #1923, opened 2022-12-07
Chain `CoeOut` instances during coercion resolution

gebner:
As mentioned in #1920.  The `Coe.lean` file is restructured like Mario suggested to avoid the combinatorial explosion.  I also added a separate `CoeOut` (:bike: :house:) class.

Builds on #1920.

### DIFF
--- a/src/Init/Coe.lean
+++ b/src/Init/Coe.lean
@@ -34,6 +34,7 @@ which is not a type error anymore.
 
 You can also use the `↑` operator to explicitly indicate a coercion. Using `↑x`
 instead of `x` in the example will result in the same output.
+
 Because there are many polymorphic functions in Lean, it is often ambiguous where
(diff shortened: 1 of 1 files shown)


digama0:
Do you know why this PR touches so many .c files? I would have thought that since the coercion paths are erased during elaboration, the resulting C files are unchanged. I can't make any sense out of the diff, it looks like the header was just reordered.

src/Init/Coe.lean
```diff
@@ -103,25 +103,66 @@ class Coe (α : Sort u) (β : Sort v) where
   /-- Coerces a value of type `α` to type `β`. Accessible by the notation `↑x`,
```
gebner, line 103: Explain chaining, when to use Coe vs. CoeOut, variable conditions above in module docstring!

PR #1931, opened 2022-12-08
fix: ignore empty character literals

### LINKED ISSUE #1922: Can't use '' as infix operator
### Prerequisites

### Description

We'd like to use `''` as an infix operator in mathlib; we use it for the image of set. However it generates a `missing end of character literal` whenever used:

### Steps to Reproduce

```lean
def Set (α : Type u) := α → Prop

def image (f : α → β) (s : Set α) : Set β :=
  fun b => ∃ a, s a ∧ f a = b

infixl:80 " '' " => image

example {f : α → β} {s : Set α} {y : β} : f '' s = f '' s := rfl -- Errors on `f '' s` with `missing end of character literal`.
```

**Expected behavior:** 

No error.

**Actual behavior:**

`missing end of character literal`

### Versions

Lean (version 4.0.0-nightly-2022-12-05, commit 0b243f0ca33e, Release)

### Additional Information

As reported in [zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/data.2Eset.2Ebasic/near/314346400).

Kha:
Fixes #1922

As a side-effect disallows the literal `'''` (unused in core), which looks like an unintended feature and also horrible compared to the more sensible and equivalent `'\''`.

### DIFF
--- a/src/Lean/Parser/Basic.lean
+++ b/src/Lean/Parser/Basic.lean
@@ -786,7 +786,7 @@ private def tokenFnAux : ParserFn := fun c s =>
   let curr  := input.get i
   if curr == '\"' then
     strLitFnAux i c (s.next input i)
-  else if curr == '\'' then
+  else if curr == '\'' && getNext input i != '\'' then
     charLitFnAux i c (s.next input i)
   else if curr.isDigit then
     numberFnAux c s


gebner:
Let's merge this to unblock mathlib.

PR #1935, opened 2022-12-08
feat: give exprToSyntax an optional type argument

thorimur:
Currently, `exprToSyntax` always infers the type of the given expression to find the type to use for the created metavariable. This PR allows the type of the expression to be given directly as an optional second argument if known.

Note: I'm aware this PR doesn't strictly adhere to the contribution guidelines; however, I was encouraged to make it directly (i.e. without making an issue first) since it's small.


Kha:
Isn't this covered by `Term.ensureHasType`?

Kha:
I'm going to close this PR as I do think it's moving in the wrong direction. In general, PRs should come with sufficient motivation and ideally a test case.

PR #1945, opened 2022-12-13
Whitespace fixes in pretty-printer

gebner:
Addresses the issues in https://github.com/leanprover/lean4/pull/1872#issuecomment-1331546401

This PR makes `.align` always print whitespace, and fixes the issue with the `where`.

It turns out that the `where` issue is unrelated to the `.align` change (as far as I can tell).  It was probably just made worse by decreasing the line width from 120 to 100.

The actual reason was that a `" where"` token prints an actual hard space before the `where`.  (Which is very much what we want in most cases, like with `" + "`.)  It is good to remember that leading spaces have a completely different semantics than trailing spaces here!  I've just changed the leading space to a `ppIndent ppSpace`.

### DIFF
--- a/src/Lean/Parser/Command.lean
+++ b/src/Lean/Parser/Command.lean
@@ -90,7 +90,7 @@ def declValEqns      := leading_parser
 def whereStructField := leading_parser
   Term.letDecl
 def whereStructInst  := leading_parser
-  " where" >> sepByIndent (ppGroup whereStructField) "; " (allowTrailingSep := true) >>
+  ppIndent ppSpace >> "where" >> sepByIndent (ppGroup whereStructField) "; " (allowTrailingSep := true) >>
   optional Term.whereDecls
 /-
   Remark: we should not use `Term.whereDecls` at `declVal`
--- a/src/Lean/Parser/Term.lean
+++ b/src/Lean/Parser/Term.lean
@@ -495,7 +495,7 @@ def «letrec» := leading_parser:leadPrec
 
 @[run_builtin_parser_attribute_hooks]
 def whereDecls := leading_parser
-  " where" >> sepBy1Indent (ppGroup letRecDecl) "; " (allowTrailingSep := true)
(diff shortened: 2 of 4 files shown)


tests/lean/1275.lean.expected.out
```diff
@@ -13,7 +14,8 @@ fun x => do
   let scp ← getCurrMacroScope
   let mainModule ← getMainModule
   pure
-      { raw :=
+      {
+        raw :=
```
gebner, line 18: This used to be formatted differently because we unconditionally printed a hard space after the opening brace.  Hard spaces should only be used if we know that they are not followed by other whitespace (like sepByIndent after the align change!), to prevent double spaces.

gebner:
OMG this is so amazing!  The tests segfault!  ❤️

tests/lean/1275.lean.expected.out
digama0: why is this introducing an extra newline though? Shouldn't the align be sufficient in this case, so that no space needs to be introduced after `{`?

Kha:
Same function prologue with `-fstack-clash-protection`:
```
0000000002915ca0 <l_Lean_Expr_withAppAux___at_Lean_instantiateMVarsCore___spec__38___rarg>:
 2915ca0:       55                      push   %rbp
 2915ca1:       48 89 e5                mov    %rsp,%rbp
 2915ca4:       48 81 ec 00 10 00 00    sub    $0x1000,%rsp
 2915cab:       48 c7 04 24 00 00 00    movq   $0x0,(%rsp)  --!!!
 2915cb2:       00
 2915cb3:       48 81 ec 40 02 00 00    sub    $0x240,%rsp
...
```

tests/lean/1275.lean.expected.out
gebner: That's the thing: there's no `.align` here since the structure instance doesn't have newline separators!  What we have here is a `.line` and apparently the formatter decided to put the whole `.fill` containing the multi-line field on a new line.

Kha (approved):

PR #1946, opened 2022-12-13
chore: remove Div Int and Mod Int instances, so mathlib can use different ones

kim-em:
Investigating one solution to Mario's RFC about the `Int Div` convention. [zulip](https://leanprover.zulipchat.com/#narrow/stream/341532-lean4-dev/topic/.5BRFC.5D.20Int.2Ediv.20convention).

Just seeing what happens in CI for now.

### DIFF
--- a/doc/int.md
+++ b/doc/int.md
@@ -11,9 +11,11 @@ Recall that nonnegative numerals are considered to be a `Nat` if there are no ty
 #check (1:Int) -- Int
 ```
 
-The operator `/` for `Int` implements integer division.
+The function `Int.tdiv` implements integer division.
+(As there are several rounding conventions for integer division,
+Lean 4 itself does not provide a default instance for the `/` notation.)
 ```lean
-#eval -10 / 4 -- -2
+#eval Int.tdiv (-10) 4 -- -2
 ```
 
 Similar to `Nat`, the internal representation of `Int` is optimized. Small integers are


src/Init/Data/Int/Basic.lean
```diff
@@ -134,26 +134,51 @@ def natAbs (m : @& Int) : Nat :=
   | ofNat m   => m
   | negSucc m => m.succ
 
-@[extern "lean_int_div"]
-def div : (@& Int) → (@& Int) → Int
+/--
```
gebner, line 137:
```suggestion
/-!
```

src/Init/Data/Int/Basic.lean
```diff
@@ -134,26 +134,51 @@ def natAbs (m : @& Int) : Nat :=
...
+* `Int.tdiv`, `Int.tmod`: T-rounding (truncating division):
+  satisfies `tdiv x y = round_to_zero (x / y)`
+
+In each case, the pair of functions unconditionally satisfies
+`x % y + (x / y) * y = x`
+which is used to fix the value of one function based on the other.
+All versions also satisfy `x / 0 = 0` and `x % 0 = x`.
+
+Lean 3 used `Int.ediv` and `Int.emod` for the typeclass instances.
+Lean 4 is now using `Int.tdiv` and `Int.tmod`,
```
gebner, line 154: This doesn't seem applicable since the PR removes these instances.

src/Init/Data/Int/Basic.lean
```diff
@@ -134,26 +134,51 @@ def natAbs (m : @& Int) : Nat :=
...
+for compatibility with C conventions.
+See https://blog.vero.site/post/modulo for the history of these conventions.
+
+Mathlib4 intends to continue using `Int.ediv` and `Int.emod`,
+and to enable that without being prescriptive,
+we do not provide instances `Div Int` or `Mod Int` here,
+preferring to directly call the functions in the few places they are needed
+in Lean 4.
```
gebner, line 162: I don't think this kind of comment should go in a docstring.

gebner (commented):
Please revert all changes to the `stage0/` directory and squash the commits.

digama0:
I think if we are going to do this in core, we should move all three definitions to core as well and give them all a native implementation.

gebner:
I think we've decided at the dev meeting last year that we're going to change `/` to ediv (like Python).

> we should move all three definitions to core as well and give them all a native implementation.

I would already be happy if we just harmonized the division definition so that std4 doesn't redefine it.  Adding the two other ones can be a follow-up PR if that speeds things up.

PR #1948, opened 2022-12-13
Set explicit size for lean_object.m_rc

iacore:
I'm not sure if `int`'s size is well-defined in C, so I made this change.


Kha:
Sorry, changes to such a central file should come with strong motivation. There is no need for any change here in our opinion.

locriacyber:
ABI stability is the motivation.

PR #1949, opened 2022-12-13
fix: protect against jumping over the stack guard page

Kha:
Ensures our stack overflow detector always works and avoids potential UB. Performance impact should, presumably, be immeasurable.

See https://github.com/leanprover/lean4/pull/1945#issuecomment-1348535571 ff for context


Kha:
!bench

gebner:
We should enable this on GCC as well.

Kha:
Ah, I missed that it's the same option for both compilers

Issue #1829, opened 2022-11-14
type error in example, sensitive to simp attribute


semorrison:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

A type error that comes and goes depending on whether we use `simp [attribute] X in ...` on an earlier declaration.

### Steps to Reproduce

```
theorem eq_iff_true_of_subsingleton [Subsingleton α] (x y : α) : x = y ↔ True :=
  ⟨fun _ => ⟨⟩, fun _ => (Subsingleton.elim ..)⟩

/- comment out following `attribute [simp]` to make the example below work -/
attribute [simp] eq_iff_true_of_subsingleton in
example : True := trivial

structure Func' (α : Sort _) (β : Sort _) :=
(toFun    : α → β)

def r : Func' α α := ⟨id⟩

@[simp] theorem r_toFun {α : Sort u_1} (a : α) : Func'.toFun r a = id a := rfl

example (x y : α) (h : x = y) : r.toFun x = y := by simp <;> rw [h]
/-
in the presence of `attribute [simp] eq_iff_true_of_subsingleton in`, one gets

application type mismatch
  Eq
argument has type
  Sort u_1
but function has type
  {α : Prop} → α → α → Prop
-/

-- If we change `example` to `theorem`, it doesn't care about the `attribute [simp]`.
theorem noissue (x y : α) (h : x = y) : r.toFun x = y := by simp <;> rw [h]
```

With the `attribute [simp] eq_iff_true_of_subsingleton in` line, we get an error on the later example. Without the attribute, the error goes away.

**Expected behavior:** [What you expect to happen]

No errors!

### Versions

leanprover/lean4:nightly-2022-11-10

### Other information

(Minimised slightly from Yakov Pechersky's minimisation at https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/application.20type.20mismatch.20in.20example/near/309533447. We're seeing this in the mathlib4 port, e.g. https://github.com/leanprover-community/mathlib4/actions/runs/3460347374/jobs/5776794830.)

kbuzzard:
#1814 also mentioned the same simp lemma

gebner:
The `in example ..` is a red herring.  `attribute [simp] foo in ...` does exactly the same as `attribute [simp] foo`.

[...]

PR #1952, opened 2022-12-13
refactor: remove ignoreLevelMVarDepth option

### LINKED ISSUE #1951: Universe parameters assigned by out-params during simp
The following issue popped up on Zulip today: https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/simp.20regression.20with.20MonoidHomClass

```lean
class FunLike (F : Type _) (A : outParam (Type _)) (B : outParam (A → Type _)) where toFun : F → ∀ a, B a
instance [FunLike F A B] : CoeFun F fun _ => ∀ a, B a where coe := FunLike.toFun
class One (M) where one : M
instance [One M] : OfNat M (nat_lit 1) where ofNat := One.one
class OneHomClass (F) (A B : outParam _) [One A] [One B] extends FunLike F A fun _ => B where
  map_one (f : F) : f 1 = 1
class MulHomClass (F) (A B : outParam _) [Mul A] [Mul B] extends FunLike F A fun _ => B
class Monoid (M) extends Mul M, One M
class MonoidHomClass (F) (A B : outParam _) [Monoid A] [Monoid B] extends MulHomClass F A B, OneHomClass F A B

theorem map_one [Monoid A] [Monoid B] [OneHomClass F A B] (f : F) : f 1 = 1 := OneHomClass.map_one f
example [Monoid M] [Monoid N] [MonoidHomClass F M N] (f : F) : f 1 = 1 := by
  simp only [map_one]
```

What's going on here is that simp disables ignoreMVarLevelDepth, but the simp lemma `map_one` needs to assign the level mvar of the out-param.

See also #1829, #1901.

gebner:
Instead, we generalize the type class universe parameters together with the out-params.

See comment at https://github.com/leanprover/lean4/issues/1829#issuecomment-1315611785

Fixes #1951.

(Mutually exclusive) alternative to #1953.

### DIFF
--- a/src/Lean/Data/HashSet.lean
+++ b/src/Lean/Data/HashSet.lean
@@ -52,10 +52,10 @@ private def mkIdx {sz : Nat} (hash : UInt64) (h : sz.isPowerOfTwo) : { u : USize
 @[inline] def fold {δ : Type w} (f : δ → α → δ) (d : δ) (m : HashSetImp α) : δ :=
   foldBuckets m.buckets d f
 
-@[inline] def forBucketsM {m : Type w → Type w} [Monad m] (data : HashSetBucket α) (f : α → m PUnit) : m PUnit :=
+@[inline] def forBucketsM {m : Type v → Type w} [Monad m] (data : HashSetBucket α) (f : α → m PUnit) : m PUnit :=
   data.val.forM fun as => as.forM f
 
-@[inline] def forM {m : Type w → Type w} [Monad m] (f : α → m PUnit) (h : HashSetImp α) : m PUnit :=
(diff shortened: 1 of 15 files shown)


gebner:
One unexpected thing that breaks here:
```lean
def foo (α) : Except String α := throw "foo"
```
It's not that finding an instance of `MonadExcept String (Except String)` would require us to assign a universe metavariable.  But during TC search we try to apply the instance `MonadError.toMonadExceptOf` which would force `α : Type`.  So TC search aborts right there since the instance may apply after assigning the level of `α`.

gebner:
Closed in favor of #1953.

PR #1954, opened 2022-12-14
Update Leanc.lean to correct typo in help output

intoverflow:
# The issue

The help output from `leanc` has a typo:

```console
$ leanc
Lean C compiler

A simple wrapper around a C compiler. Defaults to `ROOT/bin/clang`,
which can be overridden with the environment variable `LEAN_CC`. All parameters are passed
as-is to the wrapped compiler.

Interesting options:
* `--print-cflags`: print C compiler flags necessary for building against the Lean runtime and exit
* `--print-ldlags`: print C compiler flags necessary for statically linking against the Lean library and exit
$ leanc --print-ldlags
clang: error: unsupported option '--print-ldlags'
```

# The fix

Update the help text

### DIFF
--- a/src/Leanc.lean
+++ b/src/Leanc.lean
@@ -12,7 +12,7 @@ as-is to the wrapped compiler.
 
 Interesting options:
 * `--print-cflags`: print C compiler flags necessary for building against the Lean runtime and exit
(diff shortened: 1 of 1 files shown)


Kha:
Duplicate of #1947

PR #1955, opened 2022-12-14
fix: macro scopes in Conv.congr

digama0:
As [reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/server.20error/near/315845493).

### DIFF
--- a/src/Lean/Elab/Tactic/Conv/Congr.lean
+++ b/src/Lean/Elab/Tactic/Conv/Congr.lean
@@ -50,7 +50,7 @@ def congr (mvarId : MVarId) (addImplicitArgs := false) (nameSubgoals := true) :
       | .eq    =>
         if addImplicitArgs || argInfo.isExplicit then
           let tag ← if nameSubgoals then
-            pure (origTag ++ (← whnf (← inferType proof)).bindingName!)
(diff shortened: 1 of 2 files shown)


Kha:
Missing test output?

digama0:
it takes me 20 minutes to run the test anyway... I believe the test should result in no change to the output so I'm just optimistically running CI.

Kha:
Ah, wouldn't it be nicer though to also test that the goal tag looks as expected?

digama0:
well it's a hygienic name anyway, so it's hard to test that it has the right scopes. This is an edge case anyway, so I'm happy as long as it doesn't panic.

Kha:
It's not hygiened anymore after `appendTag`!

digama0:
it should still be hygienic, with the same macro scopes as the original case tag. The bug occurs when you have both a hygienic base tag and want to append a hygienic field name to it. With the fix, the field name is no longer hygienic but the base tag still is, so you get a hygienic result.

Kha:
No, I don't think goal tags are supposed to ever be hygienic, that's why `appendTag` is defined that way. 
image

Issue #1957, opened 2022-12-15
panic loop with (double) external add_decl_doc


PatrickMassot:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

Lean goes into an infinite panic loop when trying to add documentation to an external declaration twice.

### Steps to Reproduce

In a new Lean file, write:

```lean
/-- Test -/
add_decl_doc Nat

/-- Test -/
add_decl_doc Nat
```

**Expected behavior:** [What you expect to happen]

No error.

**Actual behavior:** [What actually happens]

stderr is filled with a never-ending stream of panic messages.

**Reproduces how often:** [What percentage of the time does it reproduce?]

### Versions

You can get this information from copy and pasting the output of `lean --version`,
please include the OS and what version of the OS you're running.

leanprover/lean4:nightly-2022-12-13  on linux

### Additional Information

Zulip thread: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/add_decl_doc.20.20panic.20loop

Sebastian says one `add_decl_doc` is enough on his machine.

semorrison:
A related but not identical panic, that only requires a single `add_decl_doc`, is:

in `1957_1.lean`:
```
import Lean

open Lean Meta

/-- The environment extension to track `@[ext]` lemmas. -/
initialize extExtension :
    SimpleScopedEnvExtension (Name × Array (DiscrTree.Key true)) (DiscrTree Name true) ←
  registerSimpleScopedEnvExtension {
    addEntry := fun dt (n, ks) => dt.insertCore ks n
    initial := {}
  }

initialize registerBuiltinAttribute {
  name := `ext
  descr := "Marks a lemma as extensionality lemma"
  add := fun _ _ _ => pure ()
}
```
and in `1957_2.lean`:
```
import 1957_1

@[ext]
class HasSup (α : Type u) where
  sup : α → α → α

/-- Supremum of a set -/
add_decl_doc HasSup.sup
-- PANIC at Lean.MapDeclarationExtension.insert Lean.Environment:587:2: assertion violation: env.getModuleIdxFor? declName |>.isNone -- See comment at `MapDeclarationExtension`
-- PANIC at Lean.EnvExtensionInterfaceUnsafe.getState Lean.Environment:319:4: invalid environment extension has been accessed
```

[...]

Issue #1965, opened 2022-12-18
Closed-term extraction creates out-of-bound array accesses


gebner:
```lean
set_option trace.compiler.ir.result true in
def f : Nat :=
  let arr := #[]
  if h : 0 < arr.size then
    arr[0]
  else
    42
```
We lift the "closed" term `#[][0]` here, which causes the program to crash when compiled.  (I don't think this can affect the interpreter since it only evaluates closed terms on-demand.)

gebner:
We just hit a variant of this bug in the wild in https://github.com/leanprover-community/mathlib4/pull/3509  There it's not arrays, but `Option.get none ◾`.

hargoniX:
Fixed by https://github.com/leanprover/lean4/pull/12044

PR #1966, opened 2022-12-18
fix: remove `Inhabited Environment` instance

gebner:
See #1957.

### DIFF
--- a/src/Init/Data/Array/QSort.lean
+++ b/src/Init/Data/Array/QSort.lean
@@ -8,9 +8,9 @@ import Init.Data.Array.Basic
 
 namespace Array
 -- TODO: remove the [Inhabited α] parameters as soon as we have the tactic framework for automating proof generation and using Array.fget
--- TODO: remove `partial` using well-founded recursion
 
-@[inline] def qpartition {α : Type} [Inhabited α] (as : Array α) (lt : α → α → Bool) (lo hi : Nat) : Nat × Array α :=
+def qpartition (as : Array α) (lt : α → α → Bool) (lo hi : Nat) : Nat × Array α :=
+  if h : as.size = 0 then (0, as) else have : Inhabited α := ⟨as[0]'(by revert h; cases as.size <;> simp [Nat.zero_lt_succ])⟩ -- TODO: remove
   let mid := (lo + hi) / 2
   let as  := if lt (as.get! mid) (as.get! lo) then as.swap! lo mid else as
   let as  := if lt (as.get! hi)  (as.get! lo) then as.swap! lo hi  else as
@@ -29,7 +29,7 @@ namespace Array
   loop as lo lo
 termination_by _ => hi - j
 
-@[inline] partial def qsort {α : Type} [Inhabited α] (as : Array α) (lt : α → α → Bool) (low := 0) (high := as.size - 1) : Array α :=
+@[inline] partial def qsort (as : Array α) (lt : α → α → Bool) (low := 0) (high := as.size - 1) : Array α :=
   let rec @[specialize] sort (as : Array α) (low high : Nat) :=
     if low < high then
       let p := qpartition as lt low high;
--- a/src/Lean/Server/FileWorker.lean
+++ b/src/Lean/Server/FileWorker.lean
@@ -90,7 +90,7 @@ section Elab
       : AsyncElabM (Option Snapshot) := do
     cancelTk.check
     let s ← get
-    let lastSnap := s.snaps.back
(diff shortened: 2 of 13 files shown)


src/Init/Data/Array/QSort.lean
```diff
@@ -8,9 +8,9 @@ import Init.Data.Array.Basic
 
 namespace Array
 -- TODO: remove the [Inhabited α] parameters as soon as we have the tactic framework for automating proof generation and using Array.fget
--- TODO: remove `partial` using well-founded recursion
 
-@[inline] def qpartition {α : Type} [Inhabited α] (as : Array α) (lt : α → α → Bool) (lo hi : Nat) : Nat × Array α :=
+def qpartition (as : Array α) (lt : α → α → Bool) (lo hi : Nat) : Nat × Array α :=
+  if h : as.size = 0 then (0, as) else have : Inhabited α := ⟨as[0]'(by revert h; cases as.size <;> simp [Nat.zero_lt_succ])⟩ -- TODO: remove
```
gebner, line 13: See #1965.  I've put the `if` into `qpartition` instead of `qsort` and removed the `inline` attribute as a workaround.

src/Lean/Server/FileWorker.lean
```diff
@@ -289,28 +289,29 @@ section Updates
...
+      let some headSnap ← MonadExcept.ofExcept headSnap?? | panic! "empty snapshots"
       let newHeaderSnap := { headSnap with stx := newHeaderStx, mpState := newMpState }
       let changePos := oldDoc.meta.text.source.firstDiffPos newMeta.text.source
       -- Ignore exceptions, we are only interested in the successful snapshots
       let (cmdSnaps, _) ← oldDoc.cmdSnaps.getFinishedPrefix
       -- NOTE(WN): we invalidate eagerly as `endPos` consumes input greedily. To re-elaborate only
       -- when really necessary, we could do a whitespace-aware `Syntax` comparison instead.
-      let mut validSnaps := cmdSnaps.takeWhile (fun s => s.endPos < changePos)
-      if validSnaps.length ≤ 1 then
+      let mut validSnaps ← pure (cmdSnaps.takeWhile (fun s => s.endPos < changePos))
```
gebner, line 300: Workaround to create non-dependent let-binding; otherwise `subst` can't eliminate it.  Another reason for `let dep`.

src/Lean/Server/FileWorker.lean
```diff
@@ -289,28 +289,29 @@ section Updates
...
+      if h : validSnaps.length ≤ 1 then
         validSnaps := [newHeaderSnap]
       else
         /- When at least one valid non-header snap exists, it may happen that a change does not fall
            within the syntactic range of that last snap but still modifies it by appending tokens.
            We check for this here. We do not currently handle crazy grammars in which an appended
            token can merge two or more previous commands into one. To do so would require reparsing
            the entire file. -/
-        let mut lastSnap := validSnaps.getLast!
-        let preLastSnap := if validSnaps.length ≥ 2
```
gebner, line 311: This if-condition was actually redundant---now we just prove it!

src/Lean/Server/FileWorker.lean
```diff
@@ -90,7 +90,7 @@ section Elab
       : AsyncElabM (Option Snapshot) := do
     cancelTk.check
     let s ← get
-    let lastSnap := s.snaps.back
+    let .some lastSnap := s.snaps.back? | panic! "empty snapshots"
```
Vtec234, line 93: `AsyncElabM` is an exception monad, why not `throw`?

gebner: Throwing a string is a bit cumbersome here, you'd need `throw (.ioError (.userError "empty snapshots"))`.  Ultimately, I think the effect is the same: you see an error in the editor.

Kha (approved):
Nice.

src/Lean/Server/FileWorker.lean
Vtec234: You can write `throwThe ElabTaskError "empty snapshots"`, I think. The thing that I find weird about the `panic!` is that in case it happens, we go on to execute the `Inhabited` instance, right? If I'm reading `#synth` right that ends up being `pure none` which is a lot of implicit magic. I would agree with Mario that `panic!` is best avoided, although would say more weakly that avoiding is not always worth it when there isn't already an exception effect available (but there is here).

Kha: Let's not change panic usage policy in this PR. I would say the current policy is to use panics for internal logic errors, which is true for this instance.

PR #1967, opened 2022-12-18
feat: support MessageData in #eval

gebner:
This adds two features to `#eval` that I've wanted for a long time.
 - Interactive expressions and traces.
 - Doesn't refuse to evaluate terms that have missing `Repr`/etc. instances.  `#eval (Except.ok (42, Bar.mk) : Except Foo (Nat × Bar))` now works and prints 42 even if there are no instances for `Foo` and `Bar`.

### DIFF
--- a/tests/lean/byCasesMetaM.lean.expected.out
+++ b/tests/lean/byCasesMetaM.lean.expected.out
@@ -1,4 +1,14 @@
 [Meta.debug] p q : Prop
     h : p ∧ q
     ⊢ q ∧ q
-[Meta.debug] case inl p q : Prop h : p ∧ q hAux : p ⊢ q ∧ q ------ case inr p q : Prop h : p ∧ q hAux : ¬p ⊢ q ∧ q
+[Meta.debug] case inl
+    p q : Prop
+    h : p ∧ q
+    hAux : p
+    ⊢ q ∧ q
+    ------
+    case inr
+    p q : Prop
+    h : p ∧ q
+    hAux : ¬p
+    ⊢ q ∧ q
--- a/src/Lean/Meta/PPGoal.lean
+++ b/src/Lean/Meta/PPGoal.lean
@@ -32,7 +32,7 @@ register_builtin_option pp.showLetValues : Bool := {
 }
 
 private def addLine (fmt : Format) : Format :=
-  if fmt.isNil then fmt else fmt ++ Format.line
+  if fmt.isNil then fmt else fmt ++ "\n"
 
 def getGoalPrefix (mvarDecl : MetavarDecl) : String :=
   if isLHSGoal? mvarDecl.type |>.isSome then
@@ -98,6 +98,6 @@ def ppGoal (mvarId : MVarId) : MetaM Format := do
       let fmt := fmt ++ getGoalPrefix mvarDecl ++ Format.nest indent typeFmt
       match mvarDecl.userName with
       | Name.anonymous => return fmt
-      | name           => return "case " ++ format name.eraseMacroScopes ++ Format.line ++ fmt
+      | name           => return "case " ++ format name.eraseMacroScopes ++ "\n" ++ fmt
 
 end Lean.Meta
--- a/tests/lean/simpDisch.lean.expected.out
+++ b/tests/lean/simpDisch.lean.expected.out
@@ -8,7 +8,12 @@ h1 : x ≠ 0
 h2 : y ≠ 0
 h3 : x = y
 ⊢ x ≠ 0
-case simp.discharger x y : Nat h1 : x ≠ 0 h2 : y ≠ 0 h3 : x = y ⊢ y ≠ 0
+case simp.discharger
+x y : Nat
+h1 : x ≠ 0
+h2 : y ≠ 0
+h3 : x = y
+⊢ y ≠ 0
 x y : Nat
 h1 : x ≠ 0
 h2 : y ≠ 0


tests/lean/byCasesMetaM.lean.expected.out
```diff
@@ -1,14 +1,4 @@
...
-    h : p ∧ q
-    hAux : p
-    ⊢ q ∧ q
-    ------
-    case inr
-    p q : Prop
-    h : p ∧ q
-    hAux : ¬p
-    ⊢ q ∧ q
+[Meta.debug] case inl p q : Prop h : p ∧ q hAux : p ⊢ q ∧ q ------ case inr p q : Prop h : p ∧ q hAux : ¬p ⊢ q ∧ q
```
Vtec234, line 4: Regression?

tests/lean/defInst.lean.expected.out
```diff
@@ -1,4 +1,4 @@
-[4, 5, 6]
+<missing Repr instance for Foo>
```
Vtec234, line 1: Regression? Maybe okay since `Foo` is a non-reducible type alias?

src/Lean/Eval.lean
```diff
@@ -3,24 +3,62 @@ Copyright (c) 2019 Microsoft Corporation. All rights reserved.
...
+  eval a _ := return toMessageData a
+
+instance : MetaEval Unit where
+  eval _ hideUnit := return if hideUnit then none else some "()"
+
+instance [MetaEval α] : MetaEval (BaseIO α) where
+  eval a _ := do MetaEval.eval (hideUnit := true) (← liftCoreM a)
+
+instance [MetaEval α] : MetaEval (IO α) where
+  eval a _ := do MetaEval.eval (hideUnit := true) (← liftCoreM a)
```
Vtec234, line 33: Not that it really matters but I think this could be lifted directly
```suggestion
  eval a _ := do MetaEval.eval (hideUnit := true) (← liftIO a)

instance [MetaEval α] : MetaEval (IO α) where
  eval a _ := do MetaEval.eval (hideUnit := true) (← liftIO a)
```

tests/lean/679.lean.expected.out
```diff
@@ -1 +1 @@
-Except.ok Lean.Name.anonymous
+Except.ok ([anonymous])
```
Vtec234, line 1: Something that's consistently confusing is whether the output of `#eval` should be valid code via `Repr`, or if it can be any string representation whatsoever via `ToString` too. In any case the output here is a bit weird because it is mixing the valid code `Except.ok (..)` representation with `ToString Name`. Both the previous version and `ok: [anonymous]` seem to make more sense. I thought the new `scoped instance (priority := low - 10) [ToString α] : Repr α where` was causing this but it still appears after removing the instance so I am not sure about the reason.

tests/lean/defInst.lean.expected.out
gebner: I guess you could call it a regression.  We used to reduce the types if we couldn't find an `Eval` instance.  Now there's always an `Eval` instance so we can't do that anymore (easily).

Personally I think it's the right tradeoff.  Getting type class instances for the reduced type is completely at odds with every other part of Lean.  You're supposed to be able to use `def` to create type wrappers with different instances, like a `def WithZero (A : Type u) := Option A` where `none` should always be printed as `0`.

tests/lean/679.lean.expected.out
gebner: Indeed, it's not completely what the output of `#eval` should be.  It used to take either a `ToString` or a `Repr` instance, so you already weren't guaranteed to get a "valid code" output.

The "culprit" here is the new instance:
```lean
scoped instance [ToMessageData ε] [ToMessageData α] : ToMessageData (Except ε α)
```
Which raises another related question: should `MessageData` be more like `Repr` or like `ToString`?

gebner: As for what it *should* be: for me the most important part is to get *some* output, and that expressions are clickable.  None of the `ToMessageData`/`Repr`/`ToString` classes have instances for everything, so we'll need to mix them for some time.

tests/lean/byCasesMetaM.lean.expected.out
gebner: I don't really know why this PR changed the output here, but none of the spaces here were hard linebreaks (and everything fits on one line).  `ppGoal` uses `Format.line`, and `trace[foo] "bar\n"` *also* uses `Format.line` (because the String→MessageData coercion replaces newlines by `Format.line`).

I've changed `ppGoal` to use hard newlines.

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -296,58 +298,46 @@ def failIfSucceeds (x : CommandElabM Unit) : CommandElabM Unit := do
...
+  let cls := mkApp (Lean.mkConst evalClassName [u]) α
+  let .some inst ← trySynthInstance cls | do
+    if ← MonadLog.hasErrors then throwAbortCommand
+    else throwError "failed to synthesize instance {cls}"
+  let inst ← Meta.transform inst fun e => do
+    if let .app (.const ``EvalInstances.typeInfo [v]) α := e then
+      let ppE := (← ppExpr α).pretty
+      return .done (.app (.app (.const ``EvalInstances.TypeInfo.mk [v]) α) (toExpr ppE))
+    else
+      return .continue
```
Kha, line 323: I think this block deserves a short comment

tests/lean/679.lean.expected.out
Kha: > Which raises another related question: should MessageData be more like Repr or like ToString?

I'm afraid the answer is that we want both. We use the class for many specific types like Expr and Syntax where it is like `ToString`, but then there are instance like for `List`, and `Except` above, that should definitely be `Repr`-like (which would also properly handle the parentheses).

Kha (commented):
I very much like the goals of this PR, just not sure yet about the current output

tests/lean/679.lean.expected.out
gebner: > that should definitely be Repr-like (which would also properly handle the parentheses).

Are you suggesting that `toMessageData` should have a precedence argument?

Kha: I think we would a new, separate typeclass `ReprToMessageData` (or whatever), which should have a precedence argument, yeah

tests/lean/updateExprIssue.lean.expected.out
```diff
@@ -42,4 +42,4 @@ def sefFn (x_1 : obj) (x_2 : obj) : obj :=
   Lean.Expr.mdata._impl →
     ret x_1
   Lean.Expr.proj._impl →
-    ret x_1
+    ret x_1
```
tydeu, line 45: Why did this test lose a EOF newline?

Kha:
Closing for now due to bitrot

PR #1968, opened 2022-12-18
fix: implement assertAfter using revert

### LINKED ISSUE #1963: MVarId.assertAfter incorrectly upgrades recursive auxDecl to kind "default".
```lean
import Lean

open Lean Elab.Tactic Lean.Meta

syntax (name := assertTrival) "assert_trivial ": tactic

elab_rules : tactic
  | `(tactic| assert_trivial) => withMainContext do
  let fvarIds := (← getLCtx).getFVarIds
  guard (fvarIds.size > 0)
  let fvarId := fvarIds[0]!
  let ty ← elabTerm (← `(True)) none
  let e ← elabTerm (← `(True.intro)) none
  let _ ← liftMetaTactic fun mvarId =>  do
     let result ← mvarId.assertAfter fvarId (Name.mkSimple "inserted_by_assert_trivial") ty e
     pure [result.mvarId]
  pure ()

variable {p : Prop}

theorem foo : p :=
by assert_trivial
   sorry

/-
  at the `sorry`, the context is:

    p : Prop
    inserted_by_assert_trivial : True
    foo : p
    ⊢ p

That `foo` hypothesis should not be there.
-/
```

```
$ lean --version
Lean (version 4.0.0-nightly-2022-12-13, commit 88c8cd5cf29a, Release)
```

gebner:
Fixes #1963.

### DIFF
--- a/src/Lean/Meta/Tactic/Assert.lean
+++ b/src/Lean/Meta/Tactic/Assert.lean
@@ -5,6 +5,7 @@ Authors: Leonardo de Moura
 -/
 import Lean.Meta.Tactic.FVarSubst
 import Lean.Meta.Tactic.Intro
+import Lean.Meta.Tactic.Revert
 
 namespace Lean.Meta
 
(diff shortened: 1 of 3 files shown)


src/Lean/Meta/Tactic/Revert.lean
```diff
@@ -43,6 +45,13 @@ def _root_.Lean.MVarId.revert (mvarId : MVarId) (fvarIds : Array FVarId) (preser
     mvar.mvarId!.setTag tag
     return (toRevert.map Expr.fvarId!, mvar.mvarId!)
 
+/-- Reverts all local declarations after `fvarId`. -/
+def _root_.Lean.MVarId.revertAfter (mvarId : MVarId) (fvarId : FVarId) : MetaM (Array FVarId × MVarId) :=
+  mvarId.withContext do
+    let localDecl ← fvarId.getDecl
+    let fvarIds := (← getLCtx).foldl (init := #[]) (start := localDecl.index+1) fun fvarIds decl => fvarIds.push decl.fvarId
+    mvarId.revert fvarIds (preserveOrder := true) (clearAuxDeclsInsteadOfRevert := true)
```
gebner, line 53: Looks scary, but it's already what we do in `subst`.

src/Lean/Meta/Tactic/Assert.lean
```diff
@@ -72,27 +73,15 @@ structure AssertAfterResult where
...
-    return { fvarId := fvarIdNew, mvarId := mvarIdNew, subst := subst }
+  mvarId.checkNotAssigned `assertAfter
+  let (fvarIds, mvarId) ← mvarId.revertAfter fvarId
+  let mvarId ← mvarId.assert userName type val
+  let (fvarIdNew, mvarId) ← mvarId.intro1P
+  let (fvarIdsNew, mvarId) ← mvarId.introNP fvarIds.size
+  let mut subst := {}
+  for f in fvarIds, fNew in fvarIdsNew do
+    subst := subst.insert f (mkFVar fNew)
+  return { fvarId := fvarIdNew, mvarId, subst }
```
gebner, line 84: Alternatively, we could also reuse the existing local decls here (and not revert anything).  But I believe Leo was a bit uneasy about messing with the local context like that.

gebner:
Merging as this fixes an issue in mathlib.

PR #1969, opened 2022-12-18
Provide context info in interactive goals

Vtec234:
This is a companion PR to leanprover/vscode-lean4#275. Please see that for motivation. This PR makes the following changes:
- Abstract the common parts of term-mode and tactic-mode goals as `InteractiveGoalCore`. Add a `WithRpcRef Elab.ContextInfo` field there which RPC handlers can use to run `MetaM` commands without looking for a snapshot at the given position (see motivation).
- Add an `InteractiveTermGoal` type (which already exists in vscode-lean4 but for some reason didn't exist in core) and add a `WithRpcRef Elab.TermInfo` field there. This is also so that RPC handlers can run `MetaM` computations with the term's local context.
- Consistently use `Option Bool` instead of `Bool` in the RPC API (`undefined` is falsish so this is backwards-compatible in JS clients; I hope it is for Lua, too).
- Add `GoalLocation`/`GoalsLocation` which are Lean variants of the types added to the user widget API. They are not used in core but are useful for user-widget RPC handlers.
- Bump server version.

### DIFF
--- a/src/Lean/Widget/InteractiveGoal.lean
+++ b/src/Lean/Widget/InteractiveGoal.lean
@@ -43,7 +43,6 @@ structure InteractiveGoalCore where
   type : CodeWithInfos
   /-- Metavariable context that the goal is well-typed in. -/
   ctx : WithRpcRef Elab.ContextInfo
-  deriving Inhabited
 
 /-- An interactive tactic-mode goal. -/
 structure InteractiveGoal extends InteractiveGoalCore where
@@ -58,7 +57,7 @@ structure InteractiveGoal extends InteractiveGoalCore where
   isInserted? : Option Bool := none
   /-- If true, the goal will be removed on the next tactic state. -/
   isRemoved? : Option Bool := none
-  deriving Inhabited, RpcEncodable
+  deriving RpcEncodable
 
 /-- An interactive term-mode goal. -/
 structure InteractiveTermGoal extends InteractiveGoalCore where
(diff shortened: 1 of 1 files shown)


src/Lean/Widget/InteractiveGoal.lean
```diff
@@ -8,59 +8,70 @@ import Lean.Meta.PPGoal
...
-  name of the MVar that it is a goal for.)
-  This is none when we are showing a term goal. -/
-  mvarId? : Option MVarId := none
-  /-- If true, the goal was not present on the previous tactic state.
-      Uses `none` instead of `some false` to save space in the json encoding.  -/
-  isInserted?: Option Bool := none
-  /-- If true, the goal will be removed on the next tactic state.
-      Uses `none` instead of `some false` to save space in the json encoding. -/
+  /-- Identifies the goal (ie with the unique name of the MVar that it is a goal for.) -/
+  mvarId : MVarId
```
Vtec234, line 56: It is worth discussing what data, which eventually gets passed to user widgets, should be included here. It might also be worthwhile to put the entire `TacticInfo` in `InteractiveGoals` since that contains the `Syntax` which self-replacing tactics that come with associated widgets could use to accurately edit the source. Note that this is independent of code actions which use a different mechanism.

gebner: Passing more infos to the editor seems like a good idea in general.

gebner:
Can you please summarize what is new in this PR?  There's a lot of purely cosmetic changes, docstring improvements, etc.  Are there any changes to the RPC API?  Is the `GoalsLocation` type really unused?

src/Lean/Widget/InteractiveGoal.lean
Vtec234: Ok! After all I am not sure it is needed yet but we can always add it later.

Vtec234:
@gebner edited.

gebner:
LGTM.  I'll trust you that you didn't sneak in any other changes while reformatting the code.

PR #1976, opened 2022-12-20
feat: introduce custom `leanSorryLike` semantic token type for `sorry`, `admit`, `stop`

### LINKED ISSUE #1975: Feature request: don't semantically highlight tactic sorry as a keyword
Compare
````lean
example : True := sorry
example : True := by sorry
````
If I am reading correctly the output of "Developer: Inspect Editor Tokens and Scopes", in both cases the token `sorry` is identified for regular highlighting as "textmate scopes: invalid.illegal.lean4".  However, when `sorry` is a tactic, it is also given semantic highlighting as "semantic token type: keyword" and this takes precedence.

In any case, the upshot in my standard-theme VS Code (can provide more configuration details if needed) is that when semantic highlighting is turned on, the *term* `sorry` is highlighted in red, whereas the *tactic* `sorry` is highlighted in blue.  This has been [noted by others](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Dropping.20'admit'/near/290461489) in Zulip also.

I don't know how hard it would be to semantically highlight the tactic `sorry` differently from all other tactics.  But it would definitely be very useful!

The same would be desirable for the `admit` and `stop` tactics.

### PR DESCRIPTION
**Pull request by Kha**
Fixes #1975, server-side at least

Note that as editors seem to ignore unknown token types, this automatically falls back to their existing lexical highlighting until we add support for the new type.

### DIFF
--- a/src/Lean/Server/FileWorker/RequestHandling.lean
+++ b/src/Lean/Server/FileWorker/RequestHandling.lean
@@ -382,6 +382,7 @@ def keywordSemanticTokenMap : RBMap String SemanticTokenType compare :=
     |>.insert "sorry" .leanSorryLike
     |>.insert "admit" .leanSorryLike
     |>.insert "stop" .leanSorryLike
+    |>.insert "#exit" .leanSorryLike
 
 partial def handleSemanticTokens (beginPos endPos : String.Pos)
     : RequestM (RequestTask SemanticTokens) := do

### CODE REVIEW

File: src/Lean/Data/Lsp/LanguageFeatures.lean [hunk 1]
```diff
@@ -252,22 +252,25 @@ inductive SemanticTokenType where
...
   type.toCtorIdx
 
 -- sanity check
-example {v : SemanticTokenType} : open SemanticTokenType in
-    names[v.toNat]?.map (toString <| toJson ·) = some (toString <| toJson v) := by
-  cases v <;> native_decide
+-- TODO: restore after update-stage0
+--example {v : SemanticTokenType} : open SemanticTokenType in
+--    names[v.toNat]?.map (toString <| toJson ·) = some (toString <| toJson v) := by
+--  cases v <;> native_decide
```
Kha, line 273: Since the kernel does not take options, `interpreter.prefer_native` does not help here :scream:

leodemoura (approved):

Kha:
Any takers for the VS Code part?

Issue #1438, opened 2022-08-07
Better import errors


digama0:
This is not something I expect to get fixed overnight, but this is an issue to track improvements in import errors.

Currently, any errors in file `A` will cause file `B` that imports `A` to get an error at `B.lean:1:1` which contains the dump of the lake invocations. This means that the error is both not localized to the actual failing import, and also the error itself is buried in a stream of other data.

If we play "what would Rust do", we shouldn't report import errors in file `B` at all, but rather we should put an error *inside file `A`, while elaborating `B`*, assuming we haven't already done so. LSP does allow you to do this. If `A` happens to be outside the project, then this may not be helpful, so in that case I would suggest putting the error on the import line for the first workspace file to import `A`.

gebner:
See also the discussion at https://leanprover.zulipchat.com/#narrow/stream/147302-lean4-dev/topic/LeanPaths/near/289386295

Kha:
Which itself may be obsoleted by more recent discussion around integrating the language server into Lake.
It should be noted though that in contrast to rust-analyzer et al, the Lean 4 language server intentionally does not take a global, coherent view of the project. So there's no guarantee that import errors for `A` occurring in `B` are still actual errors in `A` and not from some previous version of the file.

gebner:
I thought that was the last discussion on integrating the two, did a miss another one?

You're right about the need to show conflicting information though, when `A` has different errors from the worker and from the compilation attempt as a dependency.  (This doesn't even need to be different versions of the file, it's also possible that the imported modules in the worker are out-of-date.)

Kha:
Oops, without checking i assumed you linked to the previous discussion about extending the `print-paths` protocol https://leanprover.zulipchat.com/#narrow/stream/147302-lean4-dev/topic/.5BRFC.5D.20Server.2FLake.20Import.20Interop

Btw, for `--deps-json` I did implement error reporting as I proposed in the above thread: https://github.com/leanprover/lean4/blob/7b7e2f54dab95825a78a0d6b4c9420e794c5931d/src/Lean/Elab/Import.lean#L47. It's a trivial extension.

PR #1978, opened 2022-12-21
fix: lsp change debouncing

gebner:
This moves the sleep to after we've reproduced the existing valid snapshots.

### DIFF
--- a/src/Lean/Server/FileWorker.lean
+++ b/src/Lean/Server/FileWorker.lean
@@ -120,7 +120,7 @@ section Elab
     return some snap
 
   /-- Elaborates all commands after the last snap (at least the header snap is assumed to exist), emitting the diagnostics into `hOut`. -/
-  def unfoldCmdSnaps (m : DocumentMeta) (snaps : Array Snapshot) (cancelTk : CancelToken)
+  def unfoldCmdSnaps (m : DocumentMeta) (snaps : Array Snapshot) (cancelTk : CancelToken) (startAfterMs : UInt32)
       : ReaderT WorkerContext IO (AsyncList ElabTaskError Snapshot) := do
     let ctx ← read
     let headerSnap := snaps[0]!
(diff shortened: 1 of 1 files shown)


src/Lean/Server/FileWorker.lean
```diff
@@ -281,7 +283,6 @@ section Updates
     let initHeaderStx := (← get).initHeaderStx
     let (newHeaderStx, newMpState, _) ← Parser.parseHeader newMeta.mkInputContext
     let cancelTk ← CancelToken.new
-    -- Wait for at least one snapshot from the old doc, we don't want to unnecessarily re-run `print-paths`
```
gebner, line 284: This comment was misleading, we only call `compileHeader` once in `initializeWorker`.

To be fair, I'm not sure how this even works at all.  If I quickly change `import Lean` to `import Lea` and back to `import Lean`, then `cmdSnaps` should always be `AsyncList.delayed <error>` from then on.  Even if we get back to the original header syntax, we'd just propagate the exception when unwrapping the head task.

But I couldn't turn this into an actual bug during (my limited) testing.

Kha: > This comment was misleading, we only call `compileHeader` once in `initializeWorker`.

In a single worker process, yes, but thanks to worker restart we could quickly end up with parallel `print-paths` processes... which I apparently forgot about when moving the `exit` into a new task independent of `headSnapTask`. Which also means that in your example the worker is restarted each time, avoiding the bug. I only vaguely remember that I went through multiple revisions of where to put the `exit`.

src/Lean/Server/FileWorker.lean
```diff
@@ -134,7 +134,9 @@ section Elab
       -- This will overwrite existing ilean info for the file since this has a
       -- higher version number.
       publishIleanInfoUpdate m ctx.hOut snaps
```
Kha, line 136: We might want to move this behind the sleep as well, though it doesn't look like it currently compares document versions to deal with racing updates.

Kha (approved):

src/Lean/Server/FileWorker.lean
gebner: Thank you for the clarification!  That Lake continues to run even after the worker has stopped is a bigger issue, see also https://github.com/leanprover/lean4/issues/1438#issuecomment-1207354919

src/Lean/Server/FileWorker.lean
gebner: Hmm, I don't think the current version is incorrect though.  It's merely redundant notifications.

PR #1953, opened 2022-12-14
refactor: replace `ignoreLevelMVarDepth` by `levelAssignDepth`

gebner:
This removes the "global" `ignoreLevelMVarDepth` option that allows assignment of level mvars at *any* depth, by a much more restricted `levelAssignDepth` field in the metavariable context, which only allows assignment of level mvars at depth greater or equal to `levelAssignDepth`.  The `withNewMCtxDepth` is function gets an extra argument to opt-in to allowing level mvars from the outer scope to be assigned.  This API is enough for TC synthesis to work just like it does now.

Fixes #1951.

PR #1981, opened 2022-12-21
Miscellaneous coercion fixes

gebner:
These are various fixes to get mathlib to compile again.

### DIFF
--- a/tests/lean/syntheticHolesAsPatterns.lean.expected.out
+++ b/tests/lean/syntheticHolesAsPatterns.lean.expected.out
@@ -13,6 +13,12 @@ x : Fam2 α✝ β
 α : Type
 a : α
 ⊢ α
+syntheticHolesAsPatterns.lean:13:18-13:19: error: don't know how to synthesize placeholder
+context:
+α β : Type
+x : Fam2 α β
+n a : Nat
+⊢ Nat
 syntheticHolesAsPatterns.lean:12:18-12:19: error: don't know how to synthesize placeholder
 context:
 α✝ β : Type
@@ -20,9 +26,3 @@ x : Fam2 α✝ β
 α : Type
 a : α
 ⊢ α
-syntheticHolesAsPatterns.lean:13:18-13:19: error: don't know how to synthesize placeholder
-context:
-α β : Type
-x : Fam2 α β
-n a : Nat
-⊢ Nat


src/Init/Coe.lean
```diff
@@ -163,6 +163,7 @@ attribute [coe_decl] CoeOTC.coe
 
 instance [CoeOut α β] [CoeOTC β γ] : CoeOTC α γ where coe a := CoeOTC.coe (CoeOut.coe a : β)
 instance [CoeTC α β] : CoeOTC α β where coe a := CoeTC.coe a
+instance : CoeOTC α α where coe a := a
```
gebner, line 166: We already had the reflexivity instance for `CoeTC`, this just adds them for the other auxiliary classes as well.

src/Lean/Meta/Coe.lean
```diff
@@ -32,26 +32,18 @@ partial def expandCoe (e : Expr) : MetaM Expr :=
             return .visit e.headBeta
       return .continue
 
-register_builtin_option maxCoeSize : Nat := {
```
gebner, line 35: I'm so done with these arbitrary limits.  I spent an hour yesterday trying to figure out what broke with the latest update.  Why CoeHTCT worked but CoeT didn't, etc.

There's no reason to have a limit here, and I'm not willing to waste any more time debugging the random breakage this causes.

src/Lean/Meta/SynthInstance.lean
```diff
@@ -102,7 +102,7 @@ partial def normLevel (u : Level) : M Level := do
     | Level.max v w     => return u.updateMax! (← normLevel v) (← normLevel w)
     | Level.imax v w    => return u.updateIMax! (← normLevel v) (← normLevel w)
     | Level.mvar mvarId =>
-      if !(← isLevelMVarAssignable mvarId) then
+      if (← getMCtx).getLevelDepth mvarId != (← getMCtx).depth then
```
gebner, line 105: This was an oversight in #1953.  The `isLevelMVarAssignable` used to check for depth equality, irrespective of whether `ignoreLevelMVarDepth` was set or not.  (While other functions like `isReadOnly` took it into account.)  Now these functions all use the same logic.

gebner:
Merging to unblock the mathlib bump.

PR #1982, opened 2022-12-22
Problem: can't compile LCNF to alternative backends

cognivore:
Solution: compilers for alternative backends would like to know the bodies of functions that are exposed to the world. One example of such function is Float.ofScientific.


cognivore:
Sorry, silly Github defaults!

PR #1984, opened 2022-12-22
feat: add `fun x ↦ y` syntax

### LINKED ISSUE #1979: Support for ↦ as an alternative to => in lambdas.
### Prerequisites

### Description

This is a proposal to:
1. Allow `↦` as an alternative to `=>` when writing lambdas.
2. Have a delaborator which will print lambdas using `↦`.
3. Add a `pp` option, defaulting to false, which controls whether this delaborator is used.

### Background

Several users of the mathlib4, in particular several intending to use it in teaching, would very much like to be able to use the `↦` arrow when writing lambdas.

The mathlib4 maintainers agree that that this would be an acceptable default in mathlib4. (Noting that while not everyone prefers `↦`, all or nearly all would be content with it.)

We have attempted to provide support for this [directly in mathlib4](https://github.com/leanprover-community/mathlib4/blob/master/Mathlib/Util/MapsTo.lean). However this solution results in `Exists` (as well as all its relatives, e.g. `finset.sum`) not pretty printing correctly. We considered [reverting this change](https://github.com/leanprover-community/mathlib4/pull/1032), but did not do so after protests from the pro `↦` crowd! :-)

The hope is that the proposal here would
* make the alternative input syntax available across the Lean 4 ecosystem
* while leaving `=>` as the default input and output form for lambdas,
* but allowing mathlib4 to globally enable the `pp` option that would cause lambdas to print using `↦`, 
* without breaking pretty printing of `Exists` and related notations.

gebner:
Fixes #1979.

### DIFF
--- a/tests/lean/ppUnicode.lean
+++ b/tests/lean/ppUnicode.lean
@@ -1,0 +1,5 @@
+#check fun a b c ↦  a + b + c + 1
+set_option pp.unicode.fun true in
+#check fun a b c => a + b + c + 1
+set_option pp.unicode.fun true in
+#check ∃ n : Nat, n ≤ 0
--- a/src/Lean/Parser/Term.lean
+++ b/src/Lean/Parser/Term.lean
@@ -363,7 +363,7 @@ def funBinder : Parser :=
 -- NOTE: we disable anonymous antiquotations to ensure that `fun $b => ...`
 -- remains a `term` antiquotation
 def basicFun : Parser := leading_parser (withAnonymousAntiquot := false)
-  ppGroup (many1 (ppSpace >> funBinder) >> optType >> " =>") >> ppSpace >> termParser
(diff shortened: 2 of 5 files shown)


Kha (approved):

tests/lean/ppUnicode.lean
```diff
@@ -0,0 +1,3 @@
+#check fun a b c ↦  a + b + c + 1
+set_option pp.unicode.fun true in
+#check fun a b c => a + b + c + 1
```
hrmacbeth, line 3: Would it be worth adding
```lean
set_option pp.unicode.fun true in
#check ∃ n : Nat, n ≤ 0
```
with expected output
```text
∃ n, n ≤ 0 : Prop
```
to make sure that the [issue with the previous implementation](https://github.com/leanprover/lean4/issues/1979) is never reintroduced?

gebner: Done.

PR #1987, opened 2022-12-23
perf: avoid lifting ← over an if

gebner:
Expanding traces should be a bit faster now.

### DIFF
--- a/src/Lean/Widget/InteractiveDiagnostic.lean
+++ b/src/Lean/Widget/InteractiveDiagnostic.lean
@@ -131,16 +131,16 @@ where
   | ctx,       group d                  => Format.group <$> go nCtx ctx d
   | ctx,       .trace cls header children collapsed => do
     let header := (← go nCtx ctx header).nest 4
-    let nodes :=
+    let nodes ←
       if collapsed && !children.isEmpty then
         let children := children.map fun child =>
           MessageData.withNamingContext nCtx <|
             match ctx with
             | some ctx => MessageData.withContext ctx child
             | none     => child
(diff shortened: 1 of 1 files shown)


src/Lean/Widget/InteractiveDiagnostic.lean
```diff
@@ -131,16 +131,16 @@ where
   | ctx,       group d                  => Format.group <$> go nCtx ctx d
   | ctx,       .trace cls header children collapsed => do
     let header := (← go nCtx ctx header).nest 4
-    let nodes :=
+    let nodes ←
```
Vtec234, line 134: My apologies for introducing the regression. I had to stare at this change for a bit before I could guess what's going on. Is the point that if we use `←` instead of `:=` here then the whole expression is parsed as a `doSeq` instead of a `term` and the `if` becomes and imperative `if` so that `← children.mapM` is not lifted over it? Ideally we could have a comment here to avoid future confusion.

gebner: Yes, it's the same class of gotcha as `if ← a then if ← b then foo` vs. `if (← a) && (← b) then foo`, or `trace[foo] "{← a}"` vs. ``addTrace `foo m!"{← a}"``.

An option would be to disallow `←` lifting over `if` (like we already do for `match` and `fun`).

Another even better option would be to make the lifting customizable.  That is, `if a then b else f (← c)` is lifted to `← if a then pure b else let x ← c; pure (f x)`.  It would also be cool to lift `a && ← b` to `← if a then b else false`.

PR #1988, opened 2022-12-23
doc: Add more detail to the `split` tactic docs

jeremysalwen:
Adds additional docs to inform users
1. That an additional unnamed hypothesis is introduced for each branch.
2. That the unnamed hypothesis is commonly renamed using `cases` or `next`
3. That it can also be applied to hypotheses using the `split at h` syntax.

I am still uncertain what the correct terminology is for "goal" vs "target", so I am open to suggestions for how to reword it.

### DIFF
--- a/src/Init/Tactics.lean
+++ b/src/Init/Tactics.lean
@@ -625,13 +625,13 @@ macro_rules | `(tactic| trivial) => `(tactic| simp)
 syntax "trivial" : tactic
 
 /--
-The `split` tactic is useful for breaking nested if-then-else and match expressions in cases.
-For a `match` expression with `n` cases, the `split` tactic generates at most `n` subgoals
+The `split` tactic is useful for breaking nested if-then-else and `match` expressions into separate cases.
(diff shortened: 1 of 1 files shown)


src/Init/Tactics.lean
```diff
@@ -627,6 +627,14 @@ syntax "trivial" : tactic
...
 The `split` tactic is useful for breaking nested if-then-else and match expressions in cases.
 For a `match` expression with `n` cases, the `split` tactic generates at most `n` subgoals
+
+For example, given `n : Nat`, and a target `if n = 0 then Q else R`, `split` will generate
+one goal with hypothesis `n = 0` and target `Q`, and a second goal with hypothesis
+`¬n = 0` and target `R`.  Note that the introduced hypothesis is unnamed, and is commonly
+renamed used the `cases` or `next` tactics.
+
+- `simp` will split the goal (target).
+- `simp at h` will split the hypothesis `h`.
```
tydeu, line 637:
```suggestion
- `split` will split the goal (target).
- `split at h` will split the hypothesis `h`.
```

😉

Issue #1991, opened 2022-12-23
Use git lfs for stage0 to save space


Anderssorby:
### Suggestion

The lean4 repository takes up a large amount of space ~1.3GiB and that makes the experience less than ideal.
Since the stage0 is just generated and treated as binary data I suggest putting it into tracking by git-lfs to speed up downloads and reduce space needed. It tracks content hashes and registers and stores the data separately avoiding download of a large amount of data.
https://git-lfs.com/

```
git lfs track stage0
```

When downloading a fresh repo remember to `git lfs checkout`.

What do you think?

Kha:
No plans to implement this

PR #1992, opened 2022-12-23
feat: Add stage0 to git-lfs

Anderssorby:
Proposed in #1991. Probably some updates to docs and build scripts are necessary.


Kha:
See https://github.com/leanprover/lean4/issues/1991#issuecomment-1375629037

PR #1995, opened 2022-12-28
Include timings in trace messages when `profiler` is true

Kha:
Rudimentary implementation. Would be interesting to have an "expand hot path" context menu item.

image

### DIFF
--- a/src/Lean/Util/Trace.lean
+++ b/src/Lean/Util/Trace.lean
@@ -150,18 +150,33 @@ private def addTraceNode (oldTraces : PersistentArray TraceElem)
   let msg ← addMessageContext msg
   addTraceNodeCore oldTraces cls ref msg collapsed
 
-def withTraceNode [MonadExcept ε m] (cls : Name) (msg : Except ε α → m MessageData) (k : m α)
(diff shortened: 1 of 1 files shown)


Kha:
!bench

src/Lean/Util/Trace.lean
```diff
@@ -150,18 +150,27 @@ private def addTraceNode (oldTraces : PersistentArray TraceElem)
...
+def withTraceNode [MonadExcept ε m] [MonadLiftT BaseIO m] (cls : Name) (msg : Except ε α → m MessageData) (k : m α)
     (collapsed := true) : m α := do
   if !(← isTracingEnabledFor cls) then
     k
   else
     let ref ← getRef
     let oldTraces ← getResetTraces
-    let res ← observing k
-    addTraceNode oldTraces cls ref (← msg res) collapsed
+    let (res, secs) ← withSeconds <| observing k
```
leodemoura, line 166: `withSeconds` is being executed even when `profiler` is set to false. Have you checked the impact on performance?

leodemoura: I see you tried `!bench!` last week, but it seems it did not work.

Kha:
!bench

Kha:
!bench

Kha:
!bench

PR #1999, opened 2022-12-29
feat: add `LawfulAlternative`

fgdorais:
Adding `LawfulAlternative` is mostly for completeness at the moment. The obvious use case is `Option` but that isn't even a `LawfulMonad` yet (it is in Std). I want to keep this a zero-cost feature: neither Std nor Mathlib have a `LawfulAlternative` class.

I didn't include any "distributive laws" with `Applicative` since these are somewhat questionable. Haskell has `Alternative` and `MonadPlus` for that kind of distinction. (Note that Haskell has no kind of lawfulness enforcement, the distinction is all about intended use.)

### DIFF
--- a/src/Init/Control/Lawful.lean
+++ b/src/Init/Control/Lawful.lean
@@ -109,6 +109,106 @@ instance : LawfulMonad Id := by
 
 end Id
 
+/-! # Option -/
+
+namespace Option
+
+instance : LawfulMonad Option where
+  id_map x           := by cases x <;> rfl
+  map_const          := by intros; rfl
+  seqLeft_eq x y     := by cases x <;> cases y <;> rfl
+  seqRight_eq x y    := by cases x <;> cases y <;> rfl
+  pure_seq           := by intros; rfl
+  bind_pure_comp _ x := by cases x <;> rfl
+  bind_map           := by intros; rfl
+  pure_bind          := by intros; rfl
+  bind_assoc x       := by intros; cases x <;> rfl
+
+instance : LawfulAlternative Option where
+  map_failure        := by intros; rfl
+  map_orElse _ x _   := by cases x <;> rfl
+  failure_orElse     := by intros; rfl
+  orElse_failure x   := by cases x <;> rfl
+  orElse_assoc x _ _ := by cases x <;> rfl
+  pure_orElse        := by intros; rfl
+
+end Option
+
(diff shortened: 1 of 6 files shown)


digama0:
Given the fact that the choice of laws seems to be shaky and it doesn't have any actual use inside core, I'm inclined to put this in std instead of lean. In fact, I think even `LawfulMonad` can move to std - it doesn't seem to be doing a lot of good in core.

fgdorais:
I'm in favor of moving this entire file to Std. I only decided to PR here because  it fits very cleanly in this single file. (Though it now seems that some test needs to be fixed.)

As for the laws. The three monoid laws are clear, and they are the only three laws for a stand-alone `Alternative`. I don't remember why I did not add the functorial laws:

* `f <$> failure = failure`
* `f <$> (a <|> b) = (f <$> a) <|> (f <$> b)`

Since `Alternative` actually extends `Applicative`, the following laws also sensible: 

* `pure x <|> _ = pure x`
* `failure <*> _ = failure`
* `_ <*> failure = failure`

The following distributive laws (and analogous laws for `*>` and `<*`) are more debatable:

* `a <*> (b <|> c) = (a <*> b) <|> (a <*> c)`
* `(a <|> b) <*> c = (a <*> c) <|> (b <*> c)`

Finally, for the `Monad`  + `Alternative` combo, there is the very debatable:

* `(a <|> b) >>= f = (a >>= f) <|> (b >>= f)`

The issue with these additional laws is that they are not necessarily "linear": the lhs and rhs can disagree on how many times certain things are evaluated. The functorial laws and the "left catch law" `pure x <|> _ = pure x` are always linear and could be added at essentially no cost.

fgdorais:
I just added `map_failure`, `map_orElse` and `pure_orElse`. Mario is right, they were omitted because I wanted to keep this PR simple.

The reason these were omitted is because the combination of `Monad m` and `Alternative m` runs into the issue that `Monad.toApplicative` and `Alternative.toApplicative` don't necessarily match. The fix is to add a `MonadAlternative` class. (I did not fish to find other places to use this class.) Unfortunately, this PR is no longer simple...

Kha:
> I'm inclined to put this in std instead of lean

Agreed, std4 is a better target. Moving the entire file to std4 is something we could consider in the future, but that shouldn't stop moving just this new class there.

PR #2002, opened 2023-01-03
perf: do not backtrack after eta-defeq

gebner:
While looking into https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/type.20class.20inference.20looping/near/319109651, I noticed that we duplicate a lot of reduction steps during defeq in that example.

Namely, we start with ` fun x x_1 => x • x_1 =?= SMul.smul` and then try *both* eta-expansion and delta-reduction.  If we do eta-expansion first, we can still do the delta-reduction afterwards.  And vice versa: we can also do delta-reduction first, and then eta-expansion.  There's a couple of reduction steps, each of which does a separate eta-expansion (with a different fvar of course).  It gets even better: `SMul.smul x y` reduces to `fun i => x • y i`, so we can play the same game again after reduction!

I don't think there is any scenario where `a x =?= b x` fails to unify, but `a =?= fun x => b x` succeeds.  So this PR shouldn't cause any difference in behavior.

### DIFF
--- a/src/Lean/Meta/ExprDefEq.lean
+++ b/src/Lean/Meta/ExprDefEq.lean
@@ -58,25 +58,33 @@ where
         return false
 
 /--
-  Try to solve `a := (fun x => t) =?= b` by eta-expanding `b`.
+  Try to solve `a := (fun x => t) =?= b` by eta-expanding `b`,
+  resulting in `t =?= b x` (with a fresh free variable `x`).
 
   Remark: eta-reduction is not a good alternative even in a system without universe cumulativity like Lean.
   Example:
     ```
     (fun x : A => f ?m) =?= f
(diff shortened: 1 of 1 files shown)


gebner:
!bench

Kha:
!bench

src/Lean/Meta/ExprDefEq.lean
```diff
@@ -66,17 +66,17 @@ where
     (fun x : A => f ?m) =?= f
     ```
     The left-hand side of the constraint above it not eta-reduced because `?m` is a metavariable. -/
```
gebner, line 68: Add comment here justifying that we don't backtrack, and that we used to.

PR #2014, opened 2023-01-05
fix: redefine Nat.mod to match leanprover-community/lean

eric-wieser:
In mathlib, our definition of `0 : fin n` exploits the fact that `0 % n = 0` is true by definition.
This is desirable because it means that `⟨0, zero_pos⟩ = ofNat 0` is true by definition.

On the assumption that Lean4 core doesn't care about the underlying definition, this adjusts it to match the one used in Lean3, with the expectation that this will make porting things about `fin` easier.

The proofs are loosely inspired by https://github.com/leanprover-community/lean3port/blob/master/Leanbin/Init/Data/Nat/Lemmas.lean#L1318-L1345; althought these used syntax that is not yet available.

See the [Zulip discussion here](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Nat.2Emod.200.20n.20.3D.200.20no.20longer.20true.20by.20rfl/near/319683008).
This is not the only solution to the problem, but is certainly the one that requires the least change in mathlib.

### DIFF
--- a/src/Init/Data/Nat/Gcd.lean
+++ b/src/Init/Data/Nat/Gcd.lean
@@ -28,7 +28,11 @@ theorem gcd_succ (x y : Nat) : gcd (succ x) y = gcd (y % succ x) (succ x) :=
   rfl
 
 @[simp] theorem gcd_zero_right (n : Nat) : gcd n 0 = n := by
-  cases n <;> simp [gcd_succ]
+  cases n with
+  | zero => simp [gcd_succ]
+  | succ n =>
+    rw [gcd_succ]
+    exact gcd_zero_left _
 
 @[simp] theorem gcd_self (n : Nat) : gcd n n = n := by
   cases n <;> simp [gcd_succ]


gebner:
FWIW, if we're messing around with defeqs then I think we should be explicit about it.

```lean
protected def modCore (x y : Nat) : Nat :=
  if 0 < y ∧ y ≤ x then
    Nat.modCore (x - y) y
  else
    x
decreasing_by apply div_rec_lemma; assumption

@[extern "lean_nat_mod"]
protected def mod : @& Nat → @& Nat → Nat
  | 0, _ => 0 -- special case so that `0 % n` reduces to `0`
  | x, y => Nat.modCore x y
```

eric-wieser:
Oh, I guess that's a much smaller fix!

Either way, the current spelling seems to have exposed [a problem with `simp`](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Nat.2Emod.200.20n.20.3D.200.20no.20longer.20true.20by.20rfl/near/319698001), so trying it wasn't a total waste of time!

gebner:
Alternatively, we could also define `OfNat (Fin (no_index (n+1))) (nat_lit 0)` and `OfNat (Fin (no_index (n+1))) (no_index (i+1))` separately.

eric-wieser:
Yes, that is the approach I describe [in this message](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Data.2EFin.2EBasic.20.28mathlib4.231084.29/near/319679847)

gebner:
Superceded by #2020

PR #2027, opened 2023-01-10
feat: `funext` no arg tactic

JamesGallicchio:
Description of funext tactic includes behavior that is not implemented. This implements the behavior.

### DIFF
--- a/src/Init/NotationExtra.lean
+++ b/src/Init/NotationExtra.lean
@@ -299,7 +299,7 @@ Patterns can be used like in the `intro` tactic. Example, given a goal
 syntax "funext" (ppSpace colGt term:max)* : tactic
 
 macro_rules
-  | `(tactic|funext) => `(tactic| apply funext; intro; try funext)
+  | `(tactic|funext) => `(tactic| repeat (apply funext; intro))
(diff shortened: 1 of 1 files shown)


src/Init/NotationExtra.lean
```diff
@@ -296,9 +296,10 @@ Patterns can be used like in the `intro` tactic. Example, given a goal
 ```
 `funext (a, b)` applies `funext` once and performs pattern matching on the newly introduced pair.
 -/
-syntax "funext " (colGt term:max)+ : tactic
+syntax "funext " (colGt term:max)* : tactic
```
digama0, line 299:
```suggestion
syntax "funext" (ppSpace colGt term:max)* : tactic
```

src/Init/NotationExtra.lean
```diff
@@ -296,9 +296,10 @@ Patterns can be used like in the `intro` tactic. Example, given a goal
...
 
 macro_rules
+  | `(tactic|funext) => `(tactic| apply funext; intro; funext <|> skip)
```
digama0, line 302:
```suggestion
  | `(tactic|funext) => `(tactic| repeat apply funext; intro)
```

PR #2028, opened 2023-01-11
perf: avoid `instantiateLevelParams` in `isProofQuickApp`

Kha:
I noticed that in both `Do.lean` and a mathlib4 slice, `instantiateLevelParams` is by far the most frequent caller of `lean_dec_ref_cold` with 2.5% of all cycles. That function in turn derives about half of its call from `inferType` and the other half from `isProofQuickApp`. Given the low prevalence of `Sort u`, it seems reasonable to make the latter approximation universe-insensitive.

Opening this PR for preliminary benchmarking.

### DIFF
--- a/src/Lean/Data/LBool.lean
+++ b/src/Lean/Data/LBool.lean
@@ -19,7 +19,7 @@ def neg : LBool → LBool
   | false => true
   | undef => undef
 
-def and : LBool → LBool → LBool
+@[macro_inline] def and : LBool → LBool → LBool
   | true,  b  => b
   | a,     _  => a
 


Kha:
3% speedup can also be replicated on `Mathlib/Algebra/Group/Defs.lean`

src/Lean/Meta/InferType.lean
```diff
@@ -203,25 +203,33 @@ private def isAlwaysZero : Level → Bool
...
+  | .imax _ u   => isAlwaysZeroQuick u
+
 /--
   `isArrowProp type n` is an "approximate" predicate which returns `LBool.true`
    if `type` is of the form `A_1 -> ... -> A_n -> Prop`.
    Remark: `type` can be a dependent arrow. -/
-private partial def isArrowProp : Expr → Nat → MetaM LBool
-  | .sort u,          0   => return isAlwaysZero (← instantiateLevelMVars u) |>.toLBool
-  | .forallE ..,      0   => return LBool.false
+private partial def isArrowProp : Expr → Nat → LBool
```
Kha, line 218: This should probably be renamed `isArrowPropQuick`

Kha (commented):
Happy to add documentation on what (old and new) approximations are applied

src/Lean/Meta/InferType.lean
```diff
@@ -203,25 +203,33 @@ private def isAlwaysZero : Level → Bool
   | .max u v    => isAlwaysZero u && isAlwaysZero v
   | .imax _ u   => isAlwaysZero u
 
+private def isAlwaysZeroQuick : Level → LBool
```
gebner, line 206: This isn't a "quick" approximation, it's a actually a decision procedure, or am I missing something?

src/Lean/Meta/InferType.lean
```diff
@@ -241,9 +249,9 @@ partial def isPropQuick : Expr → MetaM LBool
   | .proj ..          => return LBool.undef
   | .forallE _ _ b _  => isPropQuick b
   | .mdata _ e        => isPropQuick e
-  | .const c lvls     => do let constType ← inferConstType c lvls; isArrowProp constType 0
```
gebner, line 244: The level instantiation here shouldn't be so bad since it's cached.

gebner (approved):

src/Lean/Meta/InferType.lean
Kha: That's true. I don't know if we have a naming scheme here, `isAlwayZeroNoInst`?

src/Lean/Meta/InferType.lean
Kha: Do you mean there is a qualitative difference to the one in `isProofQuickApp`?
In general the cache seems extremely effective and efficient (otherwise we would be completely screwed), but there are still misses from backtracking and fresh level mvars.

src/Lean/Meta/InferType.lean
gebner: I'd go for `isAlwaysZero`.  I think `Always` already implies `NoInst`.  We have a very similar function called `isNeverZero` already.

src/Lean/Meta/InferType.lean
gebner: Yes, the one in `isProofQuickApp` applies instantiation to a random universe level term, which is not cached at all.

But the one here goes through the instantiation cache for constant infos, and should almost always be a hit since we've probably inferred the type of the constant before.

src/Lean/Meta/InferType.lean
Kha: We still have the original `isAlwaysZero` above though. But since that one is now used only in the non-quick `isProp` branch, we could probably replace it with this version+`== .true` for insignificant overhead.

src/Lean/Meta/InferType.lean
Kha: I feel like I'm missing something :smile: . But since we often have constants of the form `c.{us} : ... -> C.{us}`, it seems to me that both now-eliminated `inferConstType` calls were in danger of cache misses on level mvars.

src/Lean/Meta/InferType.lean
gebner: Yeah, I think replacing `isAlwaysZero` by the `LBool` version is probably the cleanest approach.  We should also mark `LBool.and` as `@[macro_inline]`.

Unfortunately `Level.isZero` is already taken, or we could define `isZero : Level → LBool` and `isAlwaysZero l := l.isZero == .true` and `isNeverZero l := l.isZeo == .false`.

src/Lean/Meta/InferType.lean
gebner: I don't think we can avoid the instantiation here.  Even if we avoid it in `isProp`, we usually do `inferType` to unify the types during unification and there we certainly need to instantiate.  (That is, we can choose between a cache miss here or elsewhere but the total number of cache misses won't go down.)

Kha: Alright, that makes sense

Kha:
!bench

src/Lean/Meta/InferType.lean
Kha: I think it's fine to leave it as a single `private` here as long as we don't need it elsewhere

gebner (approved):

src/Lean/Meta/InferType.lean
```diff
@@ -279,7 +281,7 @@ mutual
   `isProofQuickApp f n` is an "approximate" predicate which returns `LBool.true`
    if `f` applied to `n` arguments is a proof. -/
 private partial def isProofQuickApp : Expr → Nat → MetaM LBool
-  | .const c lvls,   arity   => do let constType ← inferConstType c lvls; isArrowProposition constType arity
+  | .const c _,      arity   => do let cinfo ← getConstInfo c; isArrowProposition cinfo.type arity
```
leodemoura, line 284: This change is making the function less precise. I am also bothered that the universes are ignored here but taken into account in other functions above. 
I am also concerned that the change may improve performance in some examples (e.g., core lean), but affect the performance in others.

Kha:
Unclear how to quantify or make more precise

PR #2034, opened 2023-01-12
feat: add --target flag for LLVM backend to build objects of a different architecture

bollu:
The command line flag option `target` was chosen over `triple` to [maintain consistency with `clang`](https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-target).

This option is necessary for eventual cross-compilation plans with Lean, and as a fun side effect, also enables building wasm objects with `--target=wasm`.

### DIFF
--- a/src/Lean/Compiler/IR/EmitLLVM.lean
+++ b/src/Lean/Compiler/IR/EmitLLVM.lean
@@ -1527,7 +1527,7 @@ def emitLLVM (env : Environment) (modName : Name) (filepath : String) (tripleStr
          LLVM.linkModules (dest := emitLLVMCtx.llvmmodule) (src := modruntime)
          optimizeLLVMModule emitLLVMCtx.llvmmodule
          LLVM.writeBitcodeToFile emitLLVMCtx.llvmmodule filepath
-         let tripleStr := tripleStr?.getD (← LLVM.getDefaultTargetTriple) 
+         let tripleStr := tripleStr?.getD (← LLVM.getDefaultTargetTriple)
          let target ← LLVM.getTargetFromTriple tripleStr
          let cpu := "generic"
          let features := ""


src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -1509,9 +1509,16 @@ def optimizeLLVMModule (mod : LLVM.Module ctx) : IO Unit := do
   LLVM.disposePassManager pm
   LLVM.disposePassManagerBuilder pmb
 
+/--
+`emitLLVM` is the entrypoint for the lean shell to code generate LLVM.
+TODO(bollu): tripleStr? has type `String` instead of `Option String`
+             because there didn't seem to be a convenient way to convert a
+             `lean::optional<string>` to the correct `object_ref` which represents
```
Kha, line 1516: https://github.com/leanprover/lean4/blob/67a58467429c9c47e40aef3d04aa50c71eacd2b5/src/runtime/object.h#L305-L306

Kha: Ah, but also
https://github.com/leanprover/lean4/blob/67a58467429c9c47e40aef3d04aa50c71eacd2b5/src/runtime/object_ref.h#L171-L179
which should work with `optional<string_ref>`

bollu: Hum, but I start with an `optional<std::string>`, not an `optional<string_ref>`. I elected to use the low technology solution and build the `lean_object` by hand.

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -1509,9 +1509,12 @@ def optimizeLLVMModule (mod : LLVM.Module ctx) : IO Unit := do
   LLVM.disposePassManager pm
   LLVM.disposePassManagerBuilder pmb
 
+/--
+`emitLLVM` is the entrypoint for the lean shell to code generate LLVM.
+-/
 @[export lean_ir_emit_llvm]
-def emitLLVM (env : Environment) (modName : Name) (filepath : String) : IO Unit := do
-  LLVM.llvmInitialize
+def emitLLVM (env : Environment) (modName : Name) (filepath : String) (tripleStr? : Option String): IO Unit := do
```
Kha, line 1516:
```suggestion
def emitLLVM (env : Environment) (modName : Name) (filepath : String) (tripleStr? : Option String) : IO Unit := do
```

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -1524,7 +1527,7 @@ def emitLLVM (env : Environment) (modName : Name) (filepath : String) : IO Unit
          LLVM.linkModules (dest := emitLLVMCtx.llvmmodule) (src := modruntime)
          optimizeLLVMModule emitLLVMCtx.llvmmodule
          LLVM.writeBitcodeToFile emitLLVMCtx.llvmmodule filepath
-         let tripleStr ← LLVM.getDefaultTargetTriple
+         let tripleStr := tripleStr?.getD (← LLVM.getDefaultTargetTriple)
```
Kha, line 1530: trailing space

src/util/shell.cpp
```diff
@@ -744,13 +752,25 @@ extern "C" LEAN_EXPORT int lean_main(int argc, char ** argv) {
...
+        // a great deal of pain by erroring out if they misuse flags.
+        if (target_triple && !llvm_output) {
+            std::cerr << "ERROR: '--target' must be used with '--bc' to enable LLVM backend. Quitting code generation.\n";
+            return 1;
+        }
+
         if (llvm_output && ok) {
+	        // marshal 'optional<string>' to 'lean_object*'
+	        lean_object* const target_triple_lean =
+	        target_triple ? mk_option_some(lean::string_ref(*target_triple).to_obj_arg()) : mk_option_none();
```
Kha, line 765: indentation

src/Lean/Compiler/IR/EmitLLVM.lean
Kha: > Hum, but I start with an `optional<std::string>`, not an `optional<string_ref>`.

There's no reason you couldn't do the latter though.

Kha (approved):

PR #1364, opened 2022-07-22
connect maxHeartbeats in CoreM to kernel variable max_heartbeats

bollu:

PR #2036, opened 2023-01-14
fix: catch missing exceptions in kernel

gebner:
Without this change, the kernel throws C++ exceptions (which are not catchable in Lean code).  For example, run the following in server mode:
```lean
import Lean
open Lean

def slowlyEvalsToTrue : Prop :=
  go 1000000
where
  go | 0 => True | n + 1 => go n

#eval addDecl <| .thmDecl {
  name := `test
  levelParams := []
  type := .const ``slowlyEvalsToTrue []
  value := .const ``True.intro []
}
```

The snippet above crashes the server (due to an uncaught `lean::stack_space_exception`).  But with this PR, `addDecl` converts the C++ exception into a Lean exception and you get a nice error message.

This is blocking mathport.  One of the declarations we add to the environment takes forever to check in the kernel, breaking mathport.  If we could set a timeout, then this failure could be handled gracefully.

### DIFF
--- a/src/runtime/interrupt.cpp
+++ b/src/runtime/interrupt.cpp
@@ -27,10 +27,15 @@ void set_max_heartbeat_thousands(unsigned max) { g_max_heartbeat = static_cast<s
 scope_heartbeat::scope_heartbeat(size_t max):flet<size_t>(g_heartbeat, max) {}
 scope_max_heartbeat::scope_max_heartbeat(size_t max):flet<size_t>(g_max_heartbeat, max) {}
 
+// separate definition to allow breakpoint in debugger
+void throw_heartbeat_exception() {
+    throw heartbeat_exception();
+}
+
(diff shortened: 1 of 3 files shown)


gebner:
https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/mathport.20fails.20after.20mathlib4.20update/near/321295828

gebner:
Unfortunately, this doesn't help at all since there's lots of "pure" functions (like `Expr.instantiateRevRange`) which can throw C++ exceptions but don't go through the wrapper. :sob:

gebner:
The lack of exception catching also affects #1364, if we ever want to revive it.

gebner:
- Disable heartbeats for instantiate_rev_range
- Alternative: add heartbeat-less version of replace.

PR #2037, opened 2023-01-14
fix: IsExtern handling of @[extern ...] variants

bollu:
The `@[extern ...]` atttribute has a bunch of difference encodings: https://leanprover-community.github.io/mathlib4_docs/Lean/Compiler/ExternAttr.html#Lean.ExternAttrData

```
- @[extern] 
    encoding: .entries = [adhoc `all]
- @[extern "level_hash"]
    encoding: .entries = [standard `all "levelHash"]
- @[extern cpp "lean::string_size" llvm "lean_str_size"]
    encoding: .entries = [standard `cpp "lean::string_size", standard `llvm "leanStrSize"]
- @[extern cpp inline "#1 + #2"]
    encoding: .entries = [inline `cpp "#1 + #2"]
- @[extern cpp "foo" llvm adhoc]
    encoding: .entries = [standard `cpp "foo", adhoc `llvm]
- @[extern 2 cpp "io_prim_println"]
    encoding: .arity? = 2, .entries = [standard `cpp "ioPrimPrintln"]
```

We now check for any of:

```
'standard `all'
'standard `cpp'
'standard `c'
```

to conclude that the definition is an `extern` definition.

cc @digama0 : Did I get the check right? Are there more cases to worry about?


Kha:
This looks like patching over deficiencies in the design, especially since the patch doesn't match the docstring at all (and also the commit message mentions an unrelated function). I can't conceive of a situation where we wouldn't want to elide irrelevant parameters of an extern function, so my vote is for replacing `isExternC` with `isExtern` in order to clarify the design. /cc @leodemoura
(For further clarification we also might to remove the unused `ExternEntry` variants... which after avoiding `inline` will be all but `standard`)

leodemoura:
I am going to get back to the new code generator soon. This kind of patch is too distracting while transitioning to the new code generator. Please document any issues as bugs.

leodemoura:
See comment above.

PR #2041, opened 2023-01-17
fix: replace DecidableEq with BEq/LawfulBEq in List mem theorems

cppio:
Replaces the `[DecidableEq α]` requirement with the more general `[BEq α] [LawfulBEq α]` as `List.elem` uses `BEq`.


semorrison:
I've just rebased this onto `master`, so that CI will test Std and Mathlib against this.

semorrison (approved):

semorrison:
Should we change
```
instance [DecidableEq α] : Membership α (Array α) where
  mem a as := as.contains a
```
from `DecidableEq` to `BEq` at the same time?

PR #2046, opened 2023-01-19
fix: notation unexpander on overapplication of non-nullary notation

### LINKED ISSUE #2045: Pretty printing applied function composition
### Prerequisites

### Description

The delaborator doesn't use the function composition notation when the composition is applied.

### Steps to Reproduce

```lean
variable (f : α → β) (g : β → γ) (x : α)

#check (g ∘ f) x
```

**Expected behavior:** [What you expect to happen]

See `(g ∘ f) x  : γ`.

**Actual behavior:** [What actually happens]

See `Function.comp g f x : γ`.

**Reproduces how often:** [What percentage of the time does it reproduce?]

100%

### Versions

`leanprover/lean4:nightly-2023-01-17` under Linux.

See the discussion at https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Function.2Ecomp.20pretty.20printing

Kha:
Fixes #2045

### DIFF
--- a/src/Lean/Elab/Quotation.lean
+++ b/src/Lean/Elab/Quotation.lean
@@ -458,9 +458,9 @@ private partial def getHeadInfo (alt : Alt) : TermElabM HeadInfo :=
           | _ => undecided
         doMatch := fun yes no => do
           let prefixDiscrs ← (List.range idx).mapM (`(Syntax.getArg __discr $(quote ·)))
-          let sliceDiscr ← `(mkNullNode (__discr.getArgs.extract $(quote idx) (__discr.getNumArgs - $(quote numSuffix))))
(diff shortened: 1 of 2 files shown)


Kha:
Should be uncontroversial, the code movement into Prelude really is a general fix to allow more quotation features early.

djvelleman:
On leanprover/lean4:nightly-2023-02-03, I'm still having the problem reported by @PatrickMassot in [#2045](https://github.com/leanprover/lean4/issues/2045):
```
variable (f : α → β) (g : β → γ) (x : α)

#check (g ∘ f) x           --Function.comp g f x : γ
```
Wasn't this supposed to fix it?

Kha:
Thanks for the reminder, should be fixed in the next nightly

PR #2053, opened 2023-01-21
fix: missing withContext in simp trace

digama0:
As [reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/simp.3F.20.5B*.5D/near/322724789).

### DIFF
--- a/src/Lean/Elab/Tactic/Simp.lean
+++ b/src/Lean/Elab/Tactic/Simp.lean
@@ -283,7 +283,7 @@ def traceSimpCall (stx : Syntax) (usedSimps : UsedSimps) : MetaM Unit := do
     | .fvar fvarId => -- local hypotheses in the context
       if let some ldecl := lctx.find? fvarId then
         localsOrStar := localsOrStar.bind fun locals =>
-          if !ldecl.userName.isInaccessibleUserName &&
+          if !ldecl.userName.isInaccessibleUserName && !ldecl.userName.hasMacroScopes &&
               (lctx.findFromUserName? ldecl.userName).get!.fvarId == ldecl.fvarId then
             some (locals.push ldecl.userName)
           else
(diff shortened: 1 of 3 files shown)


src/Lean/Elab/Tactic/Simp.lean
```diff
@@ -265,7 +265,7 @@ def traceSimpCall (stx : Syntax) (usedSimps : UsedSimps) : MetaM Unit := do
     | .fvar fvarId => -- local hypotheses in the context
       if let some ldecl := lctx.find? fvarId then
         localsOrStar := localsOrStar.bind fun locals =>
-          if !ldecl.userName.isInaccessibleUserName &&
+          if !ldecl.userName.isInaccessibleUserName && !ldecl.userName.hasMacroScopes &&
```
Kha, line 268: This seems like an unrelated change, does it affect the same test case? Should `isInaccessibleUserName` even be relevant here?

digama0: I think it is indeed an unrelated bugfix. I believe this code is chosen to match `Name.toString`, since this is what is used to ultimately print the local names and it should be re-parseable, but I agree that it would have to be pretty weird code for the `isInaccessibleUserName` conjunct to trigger.

Kha: Ok, if you could add a first test with a named hypothesis and then a second commit replacing `isInaccessibleUserName` with `hasMacroScopes` with the unnamed test case, that would be great.

Kha: I.e. separating the context and inaccessible fix

semorrison: I've separated this into two commits, and verified both changes are needed.

PR #2060, opened 2023-01-24
Use `Bool.asProp` for coercions

### LINKED ISSUE #2043: [RFC] Implement Boolean-to-Prop coercion using `Bool.asProp`
A while ago, Sebastian suggested[1] that we should redefine the Boolean-to-Prop coercion using a dedicated function.  This recently came up again[3] because we're porting material about `List.filter`, which is now defined on Booleans and thus exercises the coercion heavily.
```lean
def Bool.asProp (b : Bool) : Prop :=
  b = true

instance : Coe Bool Prop where
  coe := Bool.asProp
```

Pros:
  1. No accidental simplification of `¬ b = true` to `b = false` (which requires duplication of simp lemmas).
  2. `decide b = b` is defeq (after #2038)
  3. Pretty-printing and `norm_cast` work out-of-the-box.

Cons:
  1. `simp only [h]` no longer works as well for a hypothesis `h : b` with `b : Bool`.

Original proposal: [1] https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/rfc.3A.20theorem.20names/near/240161960
Other threads:
[2] https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/BEq.20Subtype/near/296650727
[3] https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Prop.20-.3E.20Bool.20regression/near/321996465

gebner:
Fixes #2043.


src/library/compiler/csimp.cpp
```diff
@@ -1606,10 +1606,8 @@ class csimp_fn {
 
         if (is_constructor_app(env(), major)) {
             return reduce_cases_cnstr(args, I_val, major, is_let_val);
-        } else if (!is_let_val) {
-            return visit_cases_default(e);
         } else {
-            return e;
+            return visit_cases_default(e);
```
gebner, line 1610: Without this change, we would not inline anything inside
```
let x_42 := Foo.casesOn x_3
  fun .. => ... here ...
```

gebner:
Required mathlib changes are manageable, as expected: https://github.com/leanprover/std4/compare/pr2060?expand=1 https://github.com/leanprover-community/mathlib4/compare/pr2060?expand=1

gebner:
!bench

fgdorais:
I really like this but please post something on Zulip when merging since this looks like some significant update changes. A few quick hints might help too. Please remember that mathlib is not the only thing downstream 😄

kim-em:

PR #2067, opened 2023-01-27
fix: disable gmp on windows

gebner:
The msys2 gmp package does not support static linking at the moment. https://github.com/msys2/MINGW-packages/commit/f31bdf893a9a2c13fe924e58d6cb0d94891109d9


gebner:
Merging to unblock nightlies.

PR #2068, opened 2023-01-28
change default value of decide in simp config

hrmacbeth:
In #1888, @semorrison changed the default value of `decide` in the dsimp config from `true` to `false`.  This is a proposal to do the same for simp.

Perhaps this is a controversial change, so one anecdote: this is currently interacting with another issue (which I think @digama0 will record separately) to make the following `simp` take 3 seconds, more than 10x as long as `simp (config := { decide := false })`.
```lean
import Std.Data.Rat.Defs

example : (4:Rat) * 1 * 1 + 4 = 0 := by simp ; sorry
```
Of course, that other issue should also be fixed, and/or `Rat` should be designed differently to avoid it, but it seems like users will write inefficient decidability instances often enough that the same point of confusion will occur again.

### DIFF
--- a/src/Init/Meta.lean
+++ b/src/Init/Meta.lean
@@ -1229,7 +1229,7 @@ structure Config where
   etaStruct         : EtaStructMode := .all
   iota              : Bool := true
   proj              : Bool := true
-  decide            : Bool := true
+  decide            : Bool := false
   arith             : Bool := false
   autoUnfold        : Bool := false
   /--


gebner:
Related gotcha: https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.60simp.20only.60.20does.20decidable.20checks.3F/near/321408648

digama0:
For future reference: the issue with `4 * 1 * 1 + 4` is #2161 (addressed in #2162), I think it is essentially independent.

hrmacbeth:
Closing now that an alternative PR for this has been merged.

PR #2075, opened 2023-01-30
feat: add compiler.enableNew for the new compiler

hargoniX:


src/Lean/CoreM.lean
```diff
@@ -294,13 +294,21 @@ private def checkUnsupported [Monad m] [MonadEnv m] [MonadError m] (decl : Decla
     | some (Expr.const declName ..) => throwError "code generator does not support recursor '{declName}' yet, consider using 'match ... with' and/or structural recursion"
     | _ => pure ()
 
+register_builtin_option compiler.enableNew : Bool := {
+  defValue := true
+  group    := "compiler"
+  descr    := "(compiler) enable the new code generator, right now this has no practical effect apart from testing it on your code."
```
fgdorais, line 300: Is the description intended to mean:
* "setting or unsetting this option will have (practically) no effect" (this is what it currently says but I don't think that's what's intended)
* "setting this option has no practical purpose other than testing the new compiler on your code; it is currently unset by default" (I think this is right, but please correct or adjust if I'm wrong)

digama0: "this" could be interpreted to mean either "the new code generator" or "this option that enables the new code generator"; the sentence has basically the same meaning either way, closer to your first bullet. Also "it is currently unset by default" is clearly wrong, as you can see `defValue := true`.

fgdorais: I just noticed that! I guess it's bedtime for me 😴  

So "(compiler) enable the new code generator, unset to use the old code generator" would be better.

fgdorais: BTW "no practical effect" is the part that's really confusing. Why have an option with no practical effect?

fgdorais: I see! But the new code generator could still throw new errors, right? Otherwise, why is this enabled by default?

digama0: It has the effect of disabling a module that currently is not doing anything user-facing. The reason this flag was introduced (which should probably be in the PR description) is because the new code generator adds data to the olean files, which significantly increases their size, as well as the compile time of the file. The new code generator was enabled in the first place to find bugs in the compiler by testing it on user code, but some users might not want to pay the price for this, hence the opt-out.

fgdorais: Now I'm really confused. Please correct me if I'm wrong:

* This option is enabled by default
* This option has no visible effects to users
* One "invisible effect" is longer compilation time because everything is compiled twice
* One "invisible effect" is larger olean files with additional data from the new compiler

I don't understand why this option is set by default. It seems to me this should be opt-in rather than opt-out. What am I missing?

digama0: It could indeed throw new errors. It shouldn't though, to first approximation they should accept the same code. I'm not sure whether it impacts the `noncomputable` marker - I would guess that only the old code generator is responsible for this right now.

I'd rather not get into what the best choice of default here is. Enabling it by default matches the current behavior, and making it an option allows users to choose for themselves what to do about it.

fgdorais: OK. Sounds like I should't need to worry about this setting all that much. A clearer description would be:

"(compiler) enable the new code generator, this should have no significant effect on your code but it does help to test the new code generator; unset to disable this feature."

I also don't care much about the default here (or any such things in general) but, unless Lean is somehow gathering info from my local oleans (is this happening?!), then I really don't understand why it's opt-out rather than opt-in.

hargoniX: Is that a text we can agree on then?

hargoniX: Also to answer the question regarding the default. The idea is that users will come to the Zulip with their issues by themselves.

digama0: The default needs a dev meeting discussion. I will put this PR on the agenda for next week.

PR #2087, opened 2023-02-03
doc: document how to set lean's build options with CMake

### LINKED ISSUE #2086: [cmake] feature flags not declared public
### Description

The feature flags are not visible in `cmake-gui` or `make edit_cache`.

They are visible in `build_dir/stage1` though, just not in `build_dir`.

### Steps to Reproduce

1. `cmake .....`
2. In build directory, `make edit_cache`

**Expected behavior:** Should see options specific to lean4

**Actual behavior:** No special options

### Versions
Lean (version 4.0.0, commit 4b974fd60b9f, Release)

iacore:
Closes #2086

### DIFF
--- a/doc/make/index.md
+++ b/doc/make/index.md
@@ -25,10 +25,14 @@ Setting up a basic release build:
 ```bash
 git clone https://github.com/leanprover/lean4 --recurse-submodules
 cd lean4
-mkdir -p build/release
-cd build/release
-cmake ../..
-make
+cmake -B build
+cd build
+
+pushd stage1
+make edit_cache # Enable/Disable optional features
+popd
+
+make -j
 ```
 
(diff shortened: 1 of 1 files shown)


Kha:
The only supported way of setting CMake variables is as part of the initial `cmake` call as documented below in that file, otherwise we'll get into trouble with forwarding them to the correct stages. Lean very much is not a standard CMake project.

locriacyber:
There is no documentation of how to do that

hargoniX:
It's written right below in that file how to set CMake options isnt it? And on top of that even a couple of interesting ones specific to the Lean project.

locriacyber:
I mean like `LLVM=ON`. Those options are available with `make edit_cache` and in the CMake GUI.

PR #2089, opened 2023-02-04
doc: add missing clause to `Init.Coe`

chabulhwi:
Add a missing clause to the documentation of `Init.Coe`.

### DIFF
--- a/src/Init/Coe.lean
+++ b/src/Init/Coe.lean
@@ -97,13 +97,13 @@ These classes should be implemented for coercions:
 
 * `CoeFun α (γ : α → Sort v)` is a coercion to a function. `γ a` should be a
   (coercion-to-)function type, and this is triggered whenever an element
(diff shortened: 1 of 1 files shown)


src/Init/Coe.lean
```diff
@@ -102,7 +102,8 @@ These classes should be implemented for coercions:
   `CoeFun` instances apply to `CoeOut` as well.
 
 * `CoeSort α β` is a coercion to a sort. `β` must be a universe, and if
-  `a : α` appears in a place where a type is expected, like `(x : a)` or `a → a`.
+  `a : α` appears in a place where a type is expected, like `(x : a)` or `a → a`,
+  then it will be turned into `(x : CoeSort.coe a)`.
```
gebner, line 106: This is misleading, `CoeSort.coe` never appears in the elaborated term.

PR #2091, opened 2023-02-04
fix: type error in lean_io_condvar_wait

iacore:


src/runtime/mutex.cpp
```diff
@@ -51,9 +51,9 @@ extern "C" LEAN_EXPORT obj_res lean_io_condvar_new(obj_arg) {
 }
 
 extern "C" LEAN_EXPORT obj_res lean_io_condvar_wait(b_obj_arg condvar, b_obj_arg mtx, obj_arg) {
-    unique_lock<mutex> lock(*basemutex_get(mtx), std::adopt_lock_t());
+    unique_lock<mutex> lock(*basemutex_get(mtx));
```
gebner, line 54: Can you explain why `adopt_lock_t` is wrong here?  (And how this is a "type error"?)

`Condvar.wait` expects the caller to have locked the mutex already (which is the typical way to use a condition variable).  As far as I understand it, `unique_lock` would lock the mutex again without the `adopt_lock_t` flag.  Note that calling `lock` twice is UB.

locriacyber: It doesn't compile for me

locriacyber:
This was a fluke

locriacyber:
I was trying to build with -DLEAN_MULTI_THREAD=OFF. That's when the problem happened

PR #2094, opened 2023-02-05
fix: #include <mutex> in mutex.cpp

iacore:
succeeds #2091

Before this patch `unique_lock` refers to `lean::unique_lock` under certain conditions ( -DLEAN_MULTI_THREAD=OFF).

### DIFF
--- a/src/runtime/mutex.cpp
+++ b/src/runtime/mutex.cpp
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 
 Authors: Gabriel Ebner
 */
+#include <mutex>
 #include <lean/lean.h>
 #include "runtime/mutex.h"
 #include "runtime/io.h"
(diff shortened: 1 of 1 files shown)


gebner:
> Before this patch `unique_lock` refers to `lean::unique_lock` under certain conditions ( -DLEAN_MULTI_THREAD=OFF).

That's the intended behavior.  The `LEAN_MULTI_THREAD` flag should disable all multi-threading.  That flag will also enable optimizations that are only safe if there's only a single thread.  So you should never run `LEAN_MULTI_THREAD=OFF` build with multiple threads.

The actual issue is that Lean no longer builds with `LEAN_MULTI_THREAD=OFF` because `adopt_lock_t` doesn't exist, right?  If that's the case, then we should add `adopt_lock_t` to the `mutex.h` stub.  (With have no-op stubs for all threading/mutex operation when multi-threading is disabled.)

gebner (changes requested):
see comment

gebner (changes requested):
see comment

iacore:
> The actual issue is that Lean no longer builds with `LEAN_MULTI_THREAD=OFF` because `adopt_lock_t` doesn't exist, right?

Correct

> If that's the case, then we should add `adopt_lock_t` to the `mutex.h` stub. (With have no-op stubs for all threading/mutex operation when multi-threading is disabled.)

I don't know about Lean enough to say either way.


This is the error I got when trying to build stage0 with MULTI_THREAD=OFF

```
.../lean4/stage0/src/runtime/mutex.cpp: In function ‘lean::object* lean::lean_io_con
dvar_wait(b_obj_arg, b_obj_arg, obj_arg)’:
.../lean4/stage0/src/runtime/mutex.cpp:54:55: error: ‘adopt_lock_t’ is not a member 
of ‘std’
   54 |     unique_lock<mutex> lock(*basemutex_get(mtx), std::adopt_lock_t());
      |     
```

zwarich:
Closing because this is a few years old. I think that @gebner identified the underlying issue here. If someone wants Lean to build without `LEAN_MULTI_THREAD`, they can file a new issue and we can make an `adapt_lock_t` wrapper.

PR #2097, opened 2023-02-07
Profiler fixes

Kha:
The "interpretation" metric is now more accurate by subtracting run times of invoked native code. As a side effect, an interpreted closure is now reported as a separate task (regardless of whether the caller is the interpreter or native code), but I think the cumulative metric is more important here.

### DIFF
--- a/src/library/profiling.cpp
+++ b/src/library/profiling.cpp
@@ -24,8 +24,8 @@ bool get_profiler(options const & opts) {
     return opts.get_bool(*g_profiler, LEAN_DEFAULT_PROFILER);
 }
 
-second_duration get_profiling_threshold(options const & opts) {
-    return second_duration(static_cast<double>(opts.get_unsigned(*g_profiler_threshold, LEAN_DEFAULT_PROFILER_THRESHOLD))/1000.0);
+prof_clock::duration get_profiling_threshold(options const & opts) {
+    return prof_clock::duration(std::chrono::milliseconds(opts.get_unsigned(*g_profiler_threshold, LEAN_DEFAULT_PROFILER_THRESHOLD)));
 }
 
 void initialize_profiling() {
--- a/src/library/profiling.h
+++ b/src/library/profiling.h
@@ -10,10 +10,10 @@ Author: Gabriel Ebner
 
 namespace lean {
 
-using second_duration = std::chrono::duration<double>;
+using prof_clock = std::chrono::steady_clock;
 
 bool get_profiler(options const &);
-second_duration get_profiling_threshold(options const &);
+prof_clock::duration get_profiling_threshold(options const &);
 
 void initialize_profiling();
 void finalize_profiling();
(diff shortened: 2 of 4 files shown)


Kha:
!bench

Kha:
I guess the profiling mutex is not quite insignificant!

gebner:
IIUC this adds a KVMap-lookup to every native function call (even `Array.size`?) and every `pap`, even if the profiler is turned off.  That feels like a very high cost for typical interpreted code.

I don't think we're going to get representative results from benchmarking core (as virtually everything is compiled).  IIRC you said that mathlib didn't have a high percentage of interpreted runtime either; could you benchmark mathlib with this change (https://github.com/leanprover-community/mathlib4/pull/2113 fixes the panic btw)?

Kha:
!bench

Kha:
Curious, apparently the expensive part is neither the mutex (which I just realized isn't actually invoked) or `options` but `steady_clock::now`

Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
Ohh, it's caching the failure...

Kha:
!bench

Kha:
Interesting, so even if the class doesn't do much at all when the profiler is disabled, there is a very measurable slowdown. So it seems we would either have to accept this overhead or live with wildly inaccurate `interpretation` metrics.

I'll do a benchmark run of mathlib4 after the 2003 changes are ready.

Kha:
!bench

Kha:
> IIUC this adds a KVMap-lookup to every native function call (even `Array.size`?) and every `pap`, even if the profiler is turned off.

I missed this the first time, we only pay the cost when entering the interpreter (though calling an interpreted closure from the interpreter via `ap` does count as entering it, yes; in theory, we could cache the value of `profiler` in the closure).

I'm tending towards merging this as it's better to make the interpreter a little slower than to have no idea how much time it actually takes. I can do another mathlib4 run with the optimization I just pushed though.

gebner:
> I'm tending towards merging this as it's better to make the interpreter a little slower than to have no idea how much time it actually takes. I can do another mathlib4 run with the optimization I just pushed though.

Yes, please do that.  The last version of this PR was a pretty significant regression, and I think we should only merge this if the impact is much smaller.  IIUC https://github.com/leanprover-community/mathlib4/pull/3048#issuecomment-1481504601 correctly, it was a ~10% slowdown across the board.  And tactic execution was even 50% slower.

Kha:
I'm officially declaring this PR *cursed*. I noticed today that on my laptop, I don't get the ~10% slowdown from above for `workspaceSymbols` but instead **450%**. I can only surmise that `clock_gettime` overhead very much depends on the hardware. `CLOCK_MONOTONIC_COARSE` (~4ms precision?) could help with that but is, as usual, Linux-only.

Kha:
> IIUC [leanprover-community/mathlib4#3048 (comment)](https://github.com/leanprover-community/mathlib4/pull/3048#issuecomment-1481504601) correctly, it was a ~10% slowdown across the board

2-3% on average judging from instructions/task-clock, at least when `--profile`ing

> And tactic execution was even 50% slower

Only superficially, it's time previously incorrectly detected as time spent in the interpreter

Kha:
> interestingly the slowest file, `Mathlib.GroupTheory.MonoidLocalization`, leads with +25%

(it's probably `[to_additive]`)

PR #2101, opened 2023-02-10
feat: profile tactic execution

Kha:
I believe we had this category in Lean 3. I thought a bit about where to place the call, this way it will report single tactics within a block if they are over the threshold, but not tactic blocks that are slow in sum.

### DIFF
--- a/src/Lean/Elab/Tactic/Basic.lean
+++ b/src/Lean/Elab/Tactic/Basic.lean
@@ -139,7 +139,8 @@ structure EvalTacticFailure where
   exception : Exception
   state : SavedState
 
-partial def evalTactic (stx : Syntax) : TacticM Unit :=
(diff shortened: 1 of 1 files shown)


Kha:
!bench

Kha:
Might be worth double-checking the overhead on more tactic-heavy code with and without profiling enabled. Profiling continues to be a surprising source of overhead (to me), which is a little counterproductive.

Kha:
!bench

Kha:
!bench

Kha:
I moved `profileitM` to the main entry points, but apart from the standard deviation nothing really changed. Needs more benchmarking indeed.

Kha:
No discernible impact on mathlib4 using the granular option

PR #2103, opened 2023-02-10
fix: reenable structure eta during tc search

gebner:


gebner:
TODO: extract failing test case from Mathlib.Data.Rat.Defs  https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/bump.20to.202023-02-06/near/326209589

leodemoura:
Subsumed by #2210

PR #2106, opened 2023-02-11
Profiler fixes without regressions

Kha:
#2097 without interpreter changes


Kha:
!bench

Kha:
Perhaps stdlib task/wall-clock changes by ~1% are best ignored

PR #2110, opened 2023-02-15
fix: remove extra 'backtrace:' when backtrace support isn't compiled in

iacore:


src/runtime/object.cpp
```diff
@@ -100,13 +101,10 @@ extern "C" LEAN_EXPORT object * lean_panic_fn(object * default_val, object * msg
     // TODO(Leo, Kha): add thread local buffer for interpreter.
     if (g_panic_messages) {
         std::cerr << lean_string_cstr(msg) << "\n";
-#ifdef __GLIBC__
         char * bt_env = getenv("LEAN_BACKTRACE");
         if (!bt_env || strcmp(bt_env, "0") != 0) {
-            std::cerr << "backtrace:\n";
```
Kha, line 106: This line clearly is not executed for `LEAN_BACKTRACE=0`...?

Kha:
How is this supposed to change anything,

src/runtime/object.cpp
locriacyber: sorry, wrong description

Kha: This line also clearly is not executed when `__GLIBC__` is not defined...

Issue #1901, opened 2022-11-30
Type class parameters in type classes


gebner:
This issue came up again today on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/instance.20loop.3F/near/313125082):
```lean
class Funny (F : Type _) (A B : outParam (Type _)) where toFun : F → A → B
instance [Funny F A B] : CoeFun F fun _ => A → B where coe := Funny.toFun
class MulHomClass (F) (A B : outParam _) [Mul A] [Mul B] extends Funny F A B
class Monoid (M) extends Mul M
instance [Mul A] : Mul (Id A) := ‹_›
set_option trace.Meta.synthInstance true
example [Monoid A] [Monoid B] [MulHomClass F A B] (f : F) (a : A) : f a = f a := rfl -- infinite loop
```

The core issue is the instance:
```
MulHomClass.toFunny [Mul A] [Mul B] [MulHomClass F A B] : Funny F A B
```
When this instance is applied to `Funny F ?A ?B` (which is perfectly reasonable), we get a subgoal of the form `Mul ?A` (which is completely evil, as it causes us to enumerate all multiplicative structures).

We have tried various things in the past to address this issue, see 3f497b8d8effd97b2c8c793f99c5fa9e1473c215 and #1852.  But there doesn't seem to be a good description of the intended behavior we want in the end, and as a result these bandaid fixes rot over time.

For example 3f497b8d8effd97b2c8c793f99c5fa9e1473c215 changes the arguments to `{inst : Mul A} {inst : Mul B}` because they're outParam-instances, in the projection construction.  This was silently broken by several changes: 1) #1852 which removed the outParam-marker, and 2) by multiple inheritance for structures, where the non-subobject projections are generated differently.

gebner:
The quickest short-term fix is probably to write a function `makeOutParamArgsImplicit : Expr → CoreM Expr`, which uses the same logic as `checkOutParam` and then call it from both `mkCoercionToCopiedParent` and `mk_projections`.

eric-wieser:
Was @digama0's concern in https://github.com/leanprover/lean4/commit/51a29098aba4fedb665e11d0063e7b2d06b55a46#commitcomment-90468951 addressed?

gebner:
Mario's question is unrelated to today's bug.

gebner:
Reopening the issue.  The initial bug was only intended to be a discussion starter so that we can converge on a solid design here.

[...]

PR #2117, opened 2023-02-22
Out-param inprocessing and subgoal reordering in TC search

gebner:
See #1901.  This PR implements several changes:
 1. Out-parameters are replaced by metavariables during TC search (and not just at the beginning).  This reduces the amount of badly-conditioned nested TC problems.
 2. The order in which subgoals are synthesized is automatically determined when adding an instance to the environment.  If you have subgoals like `[Monoid A] [Monoid B] [MulHomClass F A B]`, then they are processed in the order 2, 0, 1.  This removes the need to mark arguments to instances as implicit instead of instance-implicit.
 3. To make it possible to compute this order, we add an additional `semiOutParam` gadget.  One way to think of it is "like an out-param, but with multiple results."  It means that all instances of the class need to provide a concrete value for this parameter.  For example, the first argument of `Coe` is a semi-out-param, so `instance : Coe α Nat where coe _ := 42` is not a valid instance because it does not assign the `α` (the instance needs to be a `CoeHead` instead).

### DIFF
--- a/src/Init/Meta.lean
+++ b/src/Init/Meta.lean
@@ -928,6 +928,7 @@ class Quote (α : Type) (k : SyntaxNodeKind := `term) where
 
 export Quote (quote)
 
+set_option synthInstance.checkSynthOrder false in
 instance [Quote α k] [CoeHTCT (TSyntax k) (TSyntax [k'])] : Quote α k' := ⟨fun a => quote (k := k) a⟩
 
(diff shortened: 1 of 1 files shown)


gebner:
!bench

gebner:
!bench

gebner:
!bench

gebner:
Closing in favor of #2174.

PR #2124, opened 2023-02-24
chore: discuss alternative `calc` syntax in `RELEASES.md`

AdrienChampion:
In answer to @Kha's [comment](https://github.com/leanprover/lean4/pull/1844#issuecomment-1443415158) on #1844

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,10 +1,10 @@
 Unreleased
 ---------
 
-* [make `calc` require the sequence of relation/proof-s to have the same indentation](https://github.com/leanprover/lean4/pull/1844),
-  and [add `calc` alternative syntax allowing underscores `_` in the first relation](https://github.com/leanprover/lean4/pull/1844)
+* [Make `calc` require the sequence of relation/proof-s to have the same indentation](https://github.com/leanprover/lean4/pull/1844),
+  and [add `calc` alternative syntax allowing underscores `_` in the first relation](https://github.com/leanprover/lean4/pull/1844).
 
-  the flexible indentation in `calc` was often used to align relations/proofs
+  The flexible indentation in `calc` was often used to align the relation symbols:
   ```lean
   example (x y : Nat) : (x + y) * (x + y) = x * x + y * x + x * y + y * y :=
     calc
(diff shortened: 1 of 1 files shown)


RELEASES.md
```diff
@@ -1,6 +1,31 @@
 Unreleased
 ---------
 
+* [make `calc` require the sequence of relation/proof-s to have the same indentation](https://github.com/leanprover/lean4/pull/1844),
+  and [add `calc` alternative syntax allowing underscores `_` in the first relation](https://github.com/leanprover/lean4/pull/1844)
```
gebner, line 5:
```suggestion
  and [add `calc` alternative syntax allowing underscores `_` in the first relation](https://github.com/leanprover/lean4/pull/1844).
```

RELEASES.md
```diff
@@ -1,6 +1,31 @@
...
+
+  the flexible indentation in `calc` was often used to align relations/proofs
```
gebner, line 7:
```suggestion
  The flexible indentation in `calc` was often used to align the relation symbols:
```
Nothing really changed about the proofs.

RELEASES.md
```diff
@@ -1,6 +1,31 @@
...
+  example (x y : Nat) : (x + y) * (x + y) = x * x + y * x + x * y + y * y :=
+    calc
+        (x + y) * (x + y) = (x + y) * x + (x + y) * y       := by rw [Nat.mul_add]
+                        -- improper indentation
+                        _ = x * x + y * x + (x + y) * y     := by rw [Nat.add_mul]
+                        _ = x * x + y * x + (x * y + y * y) := by rw [Nat.add_mul]
+                        _ = x * x + y * x + x * y + y * y   := by rw [←Nat.add_assoc]
+  ```
+
+  this is no longer legal, but the alternative syntax accomplishes essentially the same goal with proper indentation:
```
gebner, line 18:
```suggestion
  This is no longer legal.  The new syntax puts the first term right after the `calc` and each step has the same indentation:
```

AdrienChampion: The first term does not need to be right after the `calc`, at least not in the same line. For instance:

```
calc
  <expr>_0 
  '_' 'op_1' <expr>_1 ':=' <proof>_1
  '_' 'op_2' <expr>_2 ':=' <proof>_2
  ...
  '_' 'op_n' <expr>_n ':=' <proof>_n
```

AdrienChampion: Yeah ideally we would have a formatter standardizing the whole thing. I would prefer having the expression right after the `calc` but if it's a long expression it might make sense to newline it.

In any case I'm fine with the change you propose, it's just that it sounds like it is mandatory

PR #2127, opened 2023-02-28
fix: reject occurrences of inductive type in index

### LINKED ISSUE #2125: Unsoundness with self-referential indices in inductives
```lean
open Classical

noncomputable def f (α : Type) : Bool :=
  Nonempty α

inductive C : Bool → Type
  | c : C (f (C false))

theorem unsound : False :=
  if h : Nonempty (C false) then by
    let ⟨c⟩ := h
    generalize hf : false = b at c
    let ⟨⟩ := c
    simp [f, h] at hf
  else
    have : f (C false) = false := by simp [f, h]
    h ⟨this ▸ .c⟩
```

Thanks to @nikswamy for the pointer!

gebner:
Fixes #2125

### DIFF
--- a/src/kernel/inductive.cpp
+++ b/src/kernel/inductive.cpp
@@ -330,8 +330,9 @@ public:
     }
 
     /** \brief Return true iff `t` is a term of the form `I As t`
-        where `I` is the inductive datatype at position `i` being declared and
-        `As` are the global parameters of this declaration. */
+        where `I` is the inductive datatype at position `i` being declared,
+        `As` are the global parameters of this declaration,
+        and `t` does not contain any inductive datatype being declared. */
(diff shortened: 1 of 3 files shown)


digama0 (approved):

nomeata:
The unsoundness example in #2125 has an occurrence of the inductive type in the indices of the constructor’s return type:
```
-- The following inductive is unsound:
inductive C : Bool → Type
  -- Must be rejected because `C` occurs inside an index.
  | c : C (f (C false))
```

But the fix in this PR also forbids such occurrences in the *arguments* of the constructor:
```
inductive D : Bool → Type
  | c : D (f (D false)) → D true
```

Was that intentional? Or would such occurrences there be sound (maybe useful even)?

PR #2138, opened 2023-03-07
Fix leaking of file descriptors into spawned processes

### LINKED ISSUE #2137: deadlock with piped process IO + tasks
### Prerequisites

### Description

A (nondeterministic) deadlock can be reached by spawning processes with piped stdin/out concurrently, and reading from their stdouts.

### Steps to Reproduce

```
def makeProc : IO Unit := do
  let child ← IO.Process.spawn {
    cmd := "cat"
    args := #[]
    stdin := .piped
    stdout := .piped
  }
  IO.print (← child.stdout.getLine)

def main (_ : List String) : IO UInt32 := do
  IO.println "test"
  makeProc
  IO.println "done test"
  
  for _ in [0:6] do
    let _ ← IO.asTask makeProc

  return 0
```

**Expected behavior:** return successfully

**Actual behavior:** almost always deadlocks

**Reproduces how often:** > 50%. deadlock probability goes up as number of spawned processes goes up.

### Versions

Lean: 4.0.0-nightly-2023-01-29
OS: KDE neon 5.26 (Ubuntu 22.04)

Kha:
Clean up some technical debt on the way

Fixes #2137


gebner:
Oh wow that's an amazingly awful part of Linux.  AFAIUI this PR only marks the file descriptors created using the Lean API as `O_CLOEXEC`.  Sockets, files opened by FFI libraries, etc., would all still cause similar issues, right?

Alternatively, [we could also close all fds before `exec`](https://github.com/GNOME/glib/blob/bebf13d5d62fd8e962a7b55523305b70c60155c0/glib/gspawn.c#L1612-L1668).

Kha:
No worries, as far as I can tell all three operating systems are equally awful in this respect! Though it looks like in the standard Windows API it is opt-in, so I guess we should blame POSIX?

> Sockets, files opened by FFI libraries, etc., would all still cause similar issues, right?

Yes, as long they don't also use `O_CLOEXEC`, though at least Rust does so always and hopefully other nice stdlibs do as well. In the end, I guess leaking a few fds into child processes is not the end of the world as long as you don't depend on implicit file closure on fd drop as we perhaps unwisely do (nobody tell me `takeStdin` was a mistake, it has such a nice depend signature...).

> Alternatively, [we could also close all fds before exec](https://github.com/GNOME/glib/blob/bebf13d5d62fd8e962a7b55523305b70c60155c0/glib/gspawn.c#L1612-L1668).

Uhh I think I'm in preference of the current PR

gebner:
Closing fds isn't so bad, glib (the first example I could dig up) is just being very very defensive.  For us a simple `close_range` + fallback `fdwalk` should suffice.  (Modern glibc versions apparently have a `closefrom` function doing exactly that.)

> though at least Rust does so always and hopefully other nice stdlibs do as well.

I didn't check too extensively, but e.g. not even curl sets the flag (by itself).

> as long as you don't depend on implicit file closure on fd drop as we perhaps unwisely do

Can you explain how implicit file closure figures into this bug?  I thought we had completely deterministic finalization, and we'd see exactly the same behavior with an explicit `stdin.close`.

> nobody tell me takeStdin was a mistake, it has such a nice depend signature...

An extra `Handle.close` (with error reporting) certainly wouldn't hurt for cases like this where the program semantics depend on it.

Kha:
> or us a simple `close_range` + fallback `fdwalk` should suffice.

Now we're deep into non-POSIX, what about macOS and Windows? Given the complexity, I would argue this PR is "quite good enough" for now (as apparently it is for Rust, though https://docs.rs/close_fds/latest/close_fds).

> and we'd see exactly the same behavior with an explicit stdin.close

Ah, you are right of course.

> An extra Handle.close (with error reporting) certainly wouldn't hurt for cases like this where the program semantics depend on it.

Yes. The error handling is a good point in any case, `withFile` should probably use `close` for that reason.

gebner:
> good enough" for now

I tend to agree, at least it unblocks James.

I hope I understand the original issue though.  Is this what happened?
 1. We create 2x2 pipes for 2 child processes in the parent process.
 2. We start the two child processes, each inheriting all 2x2x2 pipe fds.
 3. In the parent, we close the opposite 2x2 ends of the pipes.
 4. Each child closes the opposite 2 pipe ends for its corresponding pipes.

But each child then still has access to the pipe fds for the other child, which are kept open indefinitely.

Kha:
Yes, exactly. I find it a bit surprising that the timing works out to consistently cause a deadlock, but I guess there is not much delay between the threads compared to the critical windows between `pipe()` and `execve()`.

Kha:
I should mention that the removal of I/O text mode means that `println` will now produce `\n`, not `\r\n` on Windows. Which again is just what Rust does as well, is supported by virtually all Windows applications nowadays, and arguably is the only right choice for better cross-platform consistency.

PR #2146, opened 2023-03-11
fix: offset unification with a+a+1

### LINKED ISSUE #2136: Equality of  `+ 1` and `Nat.succ` in the presence of extra `+`
### Prerequisites

### Description

Syntactic (?) equality is inconsistent in whether to identify `Nat.succ a` with `a + 1`.  This seems to happen when there are other `+` around.

### Steps to Reproduce

```lean
opaque zero : Nat → Nat
theorem foo : zero (Nat.succ a) = 0 := sorry
example : zero (a + 1) = 0 := by with_reducible exact foo -- ok
example : zero (a + a + 1) = 0 := by with_reducible exact foo -- fail
```

**Expected behavior:**

Both examples should succeed.

**Actual behavior:**

Failure in the second example, 
```text
type mismatch
  foo
has type
  zero (Nat.succ ?m.267) = 0 : Prop
but is expected to have type
  zero (a + a + 1) = 0 : Prop
```

### Versions

Nightly 2023-02-24

### Additional Information

[Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Failed.20rewrite.20by.20equation.20lemmas)

Another example, in which this causes a failure to rewrite by an equation lemma for a recursive definition:
```lean
def factorial : Nat → Nat
  | 0 => 1
  | n + 1 => (n + 1) * factorial n

example (a : Nat) : factorial (a + 1) = (a + 1) * factorial a := by
  rw [factorial] -- works

example (a b : Nat) : factorial ((a + b) + 1) = ((a + b) + 1) * factorial (a + b) := by
  rw [factorial] -- fails
```

gebner:
Fixes #2136

### DIFF
--- a/src/Lean/Meta/Offset.lean
+++ b/src/Lean/Meta/Offset.lean
@@ -60,47 +60,42 @@ where
         failure
     | _ => failure
 
-/-- Quick function for converting `e` into `s + k` s.t. `e` is definitionally equal to `Nat.add s k`. -/
-private partial def getOffsetAux (e : Expr) (top : Bool) : OptionT MetaM (Expr × Nat) := do
+mutual
+
+/--
+Quick function for converting `e` into `s + k` s.t. `e` is definitionally equal to `Nat.add s k`.
+This function always succeeds in finding such `s` and `k`
+(as a last resort it returns `e` and `0`).
+-/
+private partial def getOffset (e : Expr) : MetaM (Expr × Nat) :=
+  return (← isOffset? e).getD (e, 0)
+
+/--
+Similar to `getOffset` but returns `none` if the expression is not syntactically an offset.
+-/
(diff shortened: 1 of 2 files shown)


Kha:
Looks good to me from my limited understanding of the code. It is certainly more straight-forward.

semorrison:
@gebner just noting this caused two mild regressions in mathlib4, at
* https://github.com/leanprover-community/mathlib4/pull/3414/commits/0143ada019d67cc896075cf76bb26f2ed293f009
* https://github.com/leanprover-community/mathlib4/pull/3414/commits/ea334e44c4ca827fbf45c58fb8d6c4933d1ccb10
in case you want to look. I don't think either are serious.

PR #2151, opened 2023-03-14
perf: scale `Expr.replace` cache with input size

Kha:
When the cache is much bigger than the traversed term, we're spending most of the time with allocating and deallocating it. This change speeds up mathlib4 cumulative CPU time by ~20% and TC inference time by ~32%. In particular, `instantiateLevelParams` falls in one arbitrarily selected test case (`Mathlib.Algebra.Order.Ring.Defs`) from 12.7% cumulative time share to 3.4%.

Other approaches are conceivable such as doing no caching at all below a certain term depth, but this one does not introduce any new magic numbers and the potential for further savings in this part is unclear.


Kha:
!bench

Kha:
In particular, this should mostly obsolete the need for #2028

gebner (approved):

Issue #2055, opened 2023-01-23
typeclass inference can go haywire


kbuzzard:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

Lean's type class inference system sometimes tries and fails to solve the same problem 384 or more times consecutively in the middle of a proof.

### Steps to Reproduce

**NOTE:** Far simpler examples are further down the thread.

```lean
class Zero (α : Type u) where
  zero : α

instance Zero.toOfNat0 {α} [Zero α] : OfNat α (nat_lit 0) where
  ofNat := ‹Zero α›.1

instance Zero.ofOfNat0 {α} [OfNat α (nat_lit 0)] : Zero α where
  zero := 0

class One (α : Type u) where
  one : α

instance One.toOfNat1 {α} [One α] : OfNat α (nat_lit 1) where
  ofNat := ‹One α›.1

instance One.ofOfNat1 {α} [OfNat α (nat_lit 1)] : One α where
  one := 1

class MulZeroClass (M₀ : Type _) extends Mul M₀, Zero M₀ where
  zero_mul : ∀ a : M₀, 0 * a = 0
  mul_zero : ∀ a : M₀, a * 0 = 0

class AddSemigroup (G : Type u) extends Add G where
  add_assoc : ∀ a b c : G, a + b + c = a + (b + c)

class Semigroup (G : Type u) extends Mul G where
  mul_assoc : ∀ a b c : G, a * b * c = a * (b * c)

class CommSemigroup (G : Type u) extends Semigroup G where
  mul_comm : ∀ a b : G, a * b = b * a

class AddCommSemigroup (G : Type u) extends AddSemigroup G where
  add_comm : ∀ a b : G, a + b = b + a

class SemigroupWithZero (S₀ : Type _) extends Semigroup S₀, MulZeroClass S₀

class MulOneClass (M : Type u) extends One M, Mul M where
  one_mul : ∀ a : M, 1 * a = a
  mul_one : ∀ a : M, a * 1 = a

class AddZeroClass (M : Type u) extends Zero M, Add M where
  zero_add : ∀ a : M, 0 + a = a
  add_zero : ∀ a : M, a + 0 = a

class MulZeroOneClass (M₀ : Type u) extends MulOneClass M₀, MulZeroClass M₀

[...]

PR #2152, opened 2023-03-14
Cache more type class queries

gebner:
This PR makes two changes to the TC synthesis cache:
 1. Instead of resetting the cache when adding local instances, we add the list of local instances to the cache key.
 2. We also cache results when they contain metavariables.

Both of these changes address issues in #2055.  There we have lots of nested TC problems that are not cached at all.  Like a lot of code in mathlib, the universes of the types have level metavariables.  But even using `Type` instead of `Type _` doesn't help, because `mvarId.withContext` (used in the term elaborator when resuming) resets the instance cache when the local instances change.  (Presumably because we run the top-level TermElabM in the empty local context, but I haven't investigated this too deeply.)

The two changes combined result in a 10% speedup on mathlib.


gebner:
!bench

PR #2162, opened 2023-03-22
fix: `match` discriminant reduction should not unfold irreducible defs

### LINKED ISSUE #2161: `decide` looks inside definitions marked `@[irreducible]`
### Prerequisites

### Description

The tactic `decide` seems to be looking inside definitions marked `@[irreducible]`.  False statements involving `@[irreducible]` definitions fail slowly (presumably because `decide` is unfolding the definitions), and true statements involving `@[irreducible]` definitions can be proved.

### Steps to Reproduce

```lean
structure Foo where num : Nat deriving DecidableEq

namespace Foo

instance : OfNat Foo n := ⟨⟨n⟩⟩

/-! # Example 1 -/

@[irreducible] def mul (a b : Foo) : Foo :=
  let d := Nat.gcd a.num 1
  ⟨(a.num.div d) * (b.num.div d)⟩

example : ((Foo.mul 4 1).mul 1).mul 1 = 4 := by decide

/-! # Example 2 -/

@[irreducible] def add (a b : Foo) : Foo := ⟨a.num * b.num⟩

example : ((Foo.add 4 1).add 1).add 1 = 4 := by decide
```

**Expected behavior:** Expect both examples to fail quickly.

**Actual behavior:** Example 1 fails slowly (16s for me), and Example 2 succeeds.

### Versions

nightly-2023-03-15

### Additional Information

@digama0 and I encountered this while investigating the following slow example in Std:

```lean
import Std.Data.Rat.Basic

example : (4:Rat) * 1 * 1 + 4 = 0 := by decide -- fails, takes 3.17s
```

Kha:
This seems like a simple oversight to me: none of the functions special-cased in `canUnfoldAtMatcher` are irreducible (and it would be quite weird if they were). Likewise, Mixing `[match_pattern]` and `[irreducible]` seems inadvisable.

Fixes #2161

### DIFF
--- a/tests/lean/2161.lean
+++ b/tests/lean/2161.lean
@@ -1,0 +1,22 @@
+structure Foo where num : Nat deriving DecidableEq
+
+namespace Foo
+
+instance : OfNat Foo n := ⟨⟨n⟩⟩
+
+/-! # Example 1 -/
+
+@[irreducible] def mul (a b : Foo) : Foo :=
+  let d := Nat.gcd a.num 1
+  ⟨(a.num.div d) * (b.num.div d)⟩
+
+-- should fail fast; exact heartbeat count at time of writing is 31
+set_option maxHeartbeats 310
+example : ((Foo.mul 4 1).mul 1).mul 1 = 4 := by decide
+
+/-! # Example 2 -/
+
+@[irreducible] def add (a b : Foo) : Foo := ⟨a.num * b.num⟩
+
+-- should not succeed (and fail fast); exact heartbeat count at time of writing is 21
+example : ((Foo.add 4 1).add 1).add 1 = 4 := by decide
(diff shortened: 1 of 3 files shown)


Kha:
Perhaps a cleaner approach would be to first call `canUnfoldDefault` and on failure apply the special casing

tests/lean/2161.lean
```diff
@@ -0,0 +1,19 @@
...
+
+instance : OfNat Foo n := ⟨⟨n⟩⟩
+
+/-! # Example 1 -/
+
+@[irreducible] def mul (a b : Foo) : Foo :=
+  let d := Nat.gcd a.num 1
+  ⟨(a.num.div d) * (b.num.div d)⟩
+
+example : ((Foo.mul 4 1).mul 1).mul 1 = 4 := by decide
```
hrmacbeth, line 13: This fails both before and after this PR, but it goes from failing slowly to failing quickly.  Is there a way to put a heartbeat limit on the test?

Kha: That's a good idea, thanks

hrmacbeth: I think you put the heartbeat limit on the wrong test!

Kha: Yikes, thanks again. Now it's on both.

gebner (approved):

PR #2166, opened 2023-03-24
chore: change simp default to decide := false

kim-em:
In #1888 we turned off `decide := true` as the default behaviour in `dsimp`.

I think at the time my intent had been to turn this off in `simp` as well, but that's not what the PR did.

Since then issue #2068 (see also related [zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.60simp.20only.60.20does.20decidable.20checks.3F/near/321408648) thread) requested that we turn it off in `simp` as well.

I will compile mathlib against this PR in case of regressions there.

### DIFF
--- a/doc/flake.nix
+++ b/doc/flake.nix
@@ -4,15 +4,15 @@
   inputs.lean.url = path:../.;
   inputs.flake-utils.follows = "lean/flake-utils";
   inputs.mdBook = {
-    url = github:leanprover/mdBook;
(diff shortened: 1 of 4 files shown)


semorrison:
Definitely not working out of the box. :-) I'll try again later.

gebner:
You're not the first one to notice that this requires more than changing the default. 😄  #2068

FWIW I think false is the right default.

PR #2168, opened 2023-03-26
chore: fix typos

int-y1:
The main typo fix is `implict` -> `implicit`

### DIFF
--- a/src/Lean/Elab/Term.lean
+++ b/src/Lean/Elab/Term.lean
@@ -196,9 +196,9 @@ structure Context where
   sectionFVars       : NameMap Expr    := {}
   /-- Enable/disable implicit lambdas feature. -/
   implicitLambda     : Bool            := true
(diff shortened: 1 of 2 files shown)


src/Lean/Elab/Term.lean
```diff
@@ -1155,7 +1155,7 @@ def resolveLocalName (n : Name) : TermElabM (Option (Expr × List String)) := do
           go ... := ...
        ```
        The current namespace is `Foo.Bla`, and the full name for `go` is `Foo.Bla.f.g`, but we want to
-       refer to it using just `go`. It is also accepted to refer to it using `f.go`, `Bla.f.go`, etc.
+       refer to it using just `go`. It is also acceptable to refer to it using `f.go`, `Bla.f.go`, etc.
```
Kha, line 1158: I think this one is fine, it is literally accepted by Lean.

PR #2172, opened 2023-03-27
fix: update nix files to use strings instead of URL literals

ConnorBaker:
@Kha

This PR changes URLs to strings by double-quoting them to move away from URL literals, which were deprecated by https://github.com/NixOS/rfcs/pull/45.

Without this change, users with `no-url-literals` set to true are unable to create a flake from the template, or otherwise interact with the nix component of this repository.

As always, thank you for the nix support :)


Kha:
Thanks!

PR #2176, opened 2023-03-30
fix: respect pp.raw in interactive .ofGoal

### LINKED ISSUE #2175: Interactive diagnostics ignore `pp.raw`
```lean
set_option pp.raw true
example : 0 = 0 := no_implicit_lambda% _
```

On the command-line, this prints:
```
error: don't know how to synthesize placeholder
context:
⊢ [mdata noImplicitLambda:1 Eq.{1} Nat (OfNat.ofNat.{0} Nat 0 (instOfNatNat 0)) (OfNat.ofNat.{0} Nat 0 (instOfNatNat 0))]
```

But in VS Code, I get:
```
don't know how to synthesize placeholder
context:
⊢ 0 = 0
```

gebner:
Fixes #2175

### DIFF
--- a/src/Lean/Widget/InteractiveCode.lean
+++ b/src/Lean/Widget/InteractiveCode.lean
@@ -74,6 +74,8 @@ where
         TaggedText.tag t (go subTt)
 
 def ppExprTagged (e : Expr) (explicit : Bool := false) : MetaM CodeWithInfos := do
+  if pp.raw.get (← getOptions) then
+    return .text (toString e)
   let delab := open PrettyPrinter.Delaborator in
     if explicit then
       withOptionAtCurrPos pp.tagAppFns.name true do


src/Lean/Widget/InteractiveCode.lean
```diff
@@ -74,6 +74,8 @@ where
         TaggedText.tag t (go subTt)
 
 def ppExprTagged (e : Expr) (explicit : Bool := false) : MetaM CodeWithInfos := do
+  if pp.raw.get (← getOptions) then
```
Vtec234, line 77: Does `withOptionAtCurrPos pp.raw true` on the `delab` call not work?

gebner: `pp.raw` completely bypasses the delaborator.  That's why `pp.raw` is the only that doesn't work.

PR #2177, opened 2023-03-30
fix: simp: strip mdata when testing for True/False

### LINKED ISSUE #2173: `mdata` bug affecting interaction of `simp` and `match`
### Prerequisites

### Description

Running `simp (config := {decide := false})` inside a branch of a `match` statement will often leave an unsolved goal `True`.
This behaviour has been hidden by the default `decide := true`, which subsequently closes the `True`.

### Steps to Reproduce

```lean
example : 0 = 0 := by
  simp (config := {decide := false}) -- Works fine.

example (n : Nat) : 0 = 0 := by
  match n with
  | 0 => simp (config := {decide := false}) -- Fails: unsolved goal `True`
  | _ => simp

example (n : Nat) : 0 = 0 := by
  match n with
  | 0 => simp (config := {decide := true}) -- this is the default, so masks the problem?
  | _ => simp
```

**Expected behavior:** 
All three examples to work.

**Actual behavior:** 
The second example fails in the `0` branch, with unsolved goal `True`.

@Kha says "this looks like an mdata bug. `pp.raw` + `trace_state` uncovers the difference."
```lean
⊢ Eq.{1} Nat (OfNat.ofNat.{0} Nat 0 (instOfNatNat 0)) (OfNat.ofNat.{0} Nat 0 (instOfNatNat 0))

n : Nat
⊢ [mdata noImplicitLambda:1 Eq.{1} Nat (OfNat.ofNat.{0} Nat 0 (instOfNatNat 0)) (OfNat.ofNat.{0} Nat 0 (instOfNatNat 0))]
```

**Reproduces how often:**

Always.

### Versions

Lean (version 4.0.0-nightly-2023-03-17, commit 8650804b025c, Release)

### Additional Information

Reported on [zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/simp.20fails.20to.20finish.20after.20reaching.20.60True.60.2C.20inside.20a.20.60match.60).

gebner:
Fixes #2173

### DIFF
--- a/src/Lean/Meta/Tactic/Simp/Main.lean
+++ b/src/Lean/Meta/Tactic/Simp/Main.lean
@@ -769,7 +769,7 @@ where
   go (e : Expr) : Bool :=
     match e with
     | .forallE _ d b _ => (d.isEq || d.isHEq || b.hasLooseBVar 0) && go b
-    | _ => e.isConstOf ``False
+    | _ => e.consumeMData.isConstOf ``False
 
 abbrev Discharge := Expr → SimpM (Option Expr)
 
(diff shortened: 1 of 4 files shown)


Kha:
I think I would be mildly in favor of just using `.consumeMData.isConstOf`, the new name doesn't exactly roll off the tongue and my first association with `mod` in these contexts is "module".

PR #2184, opened 2023-04-04
chore: modify misleading doc-string for repeat tactic

kim-em:
This doc-string has tripped people up multiple times [1](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/repeat.20issues) [2](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Semantics.20of.20repeat/near/307770628) etc.

(`Std` provides a `repeat'` with "the expected" behaviour.)

### DIFF
--- a/src/Init/Tactics.lean
+++ b/src/Init/Tactics.lean
@@ -607,10 +607,9 @@ syntax (name := cases) "cases " casesTarget,+ (" using " ident)? (inductionAlts)
 syntax (name := renameI) "rename_i " (colGt binderIdent)+ : tactic
 
 /--
-`repeat tac` applies `tac` to main goal.
-If the application succeeds, the tactic is applied recursively
-to the first of the generated subgoals until it eventually fails.
-Failure on one subgoal will result in `tac` not running on any later subgoals.
+`repeat tac` repeatedly applies `tac` to the main goal until it fails.
+That is, if `tac` produces multiple subgoals, only subgoals up to the first failure will be visited.
+The `Std` library provides `repeat'` which repeats separately in each subgoal.
 -/
 syntax "repeat " tacticSeq : tactic
 macro_rules


src/Init/Tactics.lean
```diff
@@ -607,8 +607,10 @@ syntax (name := cases) "cases " casesTarget,+ (" using " ident)? (inductionAlts)
 syntax (name := renameI) "rename_i " (colGt binderIdent)+ : tactic
 
 /--
-`repeat tac` applies `tac` to main goal. If the application succeeds,
-the tactic is applied recursively to the generated subgoals until it eventually fails.
+`repeat tac` applies `tac` to main goal.
+If the application succeeds, the tactic is applied recursively
+to the first of the generated subgoals until it eventually fails.
+Failure on one subgoal will result in `tac` not running on any later subgoals.
```
Kha, line 613: I would suggest a more direct/iterative description, with the not-quite-recursive behavior as a caveat. Do you agree?
```suggestion
`repeat tac` repeatedly applies `tac` to the main goal until it fails.
That is, if `tac` produces multiple subgoals, only subgoals up to the first failure will be visited.
```

digama0: it should probably also reference the `repeat'` tactic for people who want the other behavior.

semorrison:
```suggestion
`repeat tac` repeatedly applies `tac` to the main goal until it fails.
That is, if `tac` produces multiple subgoals, only subgoals up to the first failure will be visited.
The `Std` library provides `repeat'` which repeats separately in each subgoal.
```

semorrison:
I switched to @Kha's suggestion, and incorporated @digama0's suggestion to reference `repeat'` from `Std`.

PR #2185, opened 2023-04-05
fix: advance pointer into array when generating random bytes

olitb:
otherwise if we have more than one chunk then the first one is overwritten over and over, and the end of the array is not really random


Vtec234 (approved):
Well spotted, thanks!

PR #2187, opened 2023-04-09
feat: add timing profiling to the new compiler

hargoniX:


Kha:
I would suggest keeping the trace class as `Compiler` for now, the new output doesn't seem very helpful when enabling specific classes

Kha:
Speaking from a profiling PoV, we should also have a trace node for the entire compiler invocation

hargoniX:
I fixed the first part, for the second we have:
```
@[export lean_lcnf_compile_decls]
def main (declNames : List Name) : CoreM Unit := do
  profileitM Exception "compilation new" (← getOptions) do
    CompilerM.run <| discard <| PassManager.run declNames.toArray
```
right now. What is our policy when the thing already has a `profileitM` call attach? Do I just trace node it anyways? Do I remove profileit?

Kha:
> Do I just trace node it anyways?

Yes, the two profilers are completely separate currently

Kha:
Then I'll merge this as an improvement to profiling that should not otherwise significantly affect compiler dev experience

PR #2196, opened 2023-04-18
feat: profiling of linters

hargoniX:
This is a minimally breaking change in order to get nice and usable names. An example trace from https://proofassistants.stackexchange.com/questions/2101/speeding-up-large-data-structures-in-lean would be:
```
...
[Elab.info] [5.388586s] running linters
  [Elab.info] [5.388538s] running linter: Lean.Linter.unusedVariables
```
so it'll certainly be helpful in noticing when a linter is at fault.

### DIFF
--- a/src/Lean/Elab/Command.lean
+++ b/src/Lean/Elab/Command.lean
@@ -47,7 +47,9 @@ structure Context where
 abbrev CommandElabCoreM (ε) := ReaderT Context $ StateRefT State $ EIO ε
 abbrev CommandElabM := CommandElabCoreM Exception
 abbrev CommandElab  := Syntax → CommandElabM Unit
-abbrev Linter := Syntax → CommandElabM Unit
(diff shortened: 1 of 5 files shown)


src/Lean/Elab/Command.lean
```diff
@@ -195,17 +197,20 @@ instance : MonadLog CommandElabM where
...
-      let savedState ← get
-      try
-        linter stx
-      catch ex =>
-        logException ex
-      finally
-        modify fun s => { savedState with messages := s.messages }
+def runLinters (stx : Syntax) : CommandElabM Unit := do
+  profileitM Exception "linting" (← getOptions) do
+    withTraceNode `Elab.info (fun _ => return m!"running linters") do
```
Kha, line 202: Let's use a new trace class `Elab.lint`, `info` is for inspecting the info tree

PR #2200, opened 2023-04-21
fix: instantiate metavariables in isClassQuick? and isClassExpensive?

### LINKED ISSUE #2199: introN doesn't always correctly update instance cache
### Description

Consider 
```lean
theorem exists_foo : ∃ T : Type, Nonempty T := ⟨Unit, ⟨()⟩⟩

example : True := by
  cases exists_foo
  rename_i T hT
  have : Nonempty T := hT -- delete this line to get error in next line
  have : Nonempty T := inferInstance
  trivial
```
The problem is that the local instance cache is not updated correctly by cases.

* `cases` calls [Lean.MVarId.introN](https://leanprover-community.github.io/mathlib_docs/find/Lean.MVarId.introN)
* `introN` (via `introNImp`) calls [Lean.Meta.withNewLocalInstances](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=Lean.Meta.withNewLocalInstances#doc) that checks whether certain fvars are classes using two methods `isClassQuick?` and `isClassExpensive?`. 
* These two functions both have some support dealing with metavariables, but it makes the wrong decision in certain cases, not recognizing that (the metavariable assigned to) `Nonempty T` is a class.
* [this Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/!4.232894) has more discussion.

fpvandoorn:
* Fixes #2199
* It is possible that this is not the most performant way to fix issue #2199. It might be useful to benchmark this PR before merging it.

### DIFF
--- a/tests/lean/CompilerElimDeadBranches.lean.expected.out
+++ b/tests/lean/CompilerElimDeadBranches.lean.expected.out
@@ -1,17 +1,17 @@
-[Compiler.elimDeadBranches] Eliminating addSomeVal._redArg with #[("val.16", Lean.Compiler.LCNF.UnreachableBranches.Value.ctor `Nat.zero #[]),
+[Compiler.elimDeadBranches] Eliminating addSomeVal._redArg with #[("_x.35", Lean.Compiler.LCNF.UnreachableBranches.Value.top),
+      ("val.16", Lean.Compiler.LCNF.UnreachableBranches.Value.ctor `Nat.zero #[]),
+      ("val.20", Lean.Compiler.LCNF.UnreachableBranches.Value.ctor `Nat.zero #[]),
       ("_x.36",
        Lean.Compiler.LCNF.UnreachableBranches.Value.ctor `Option.some #[Lean.Compiler.LCNF.UnreachableBranches.Value.top]),
-      ("_x.35", Lean.Compiler.LCNF.UnreachableBranches.Value.top),
       ("_x.37",
        Lean.Compiler.LCNF.UnreachableBranches.Value.ctor
          `Option.some
-         #[Lean.Compiler.LCNF.UnreachableBranches.Value.ctor `Nat.zero #[]]),
-      ("val.20", Lean.Compiler.LCNF.UnreachableBranches.Value.ctor `Nat.zero #[])]
+         #[Lean.Compiler.LCNF.UnreachableBranches.Value.ctor `Nat.zero #[]])]
 [Compiler.elimDeadBranches] Threw away cases _x.37 branch Option.none
 [Compiler.elimDeadBranches] Threw away cases _x.37 branch Option.none
-[Compiler.elimDeadBranches] Eliminating addSomeVal with #[("_x.38",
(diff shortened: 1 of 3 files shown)


src/Lean/Meta/Basic.lean
```diff
@@ -1007,7 +1016,7 @@ mutual
 
   private partial def isClassExpensive? (type : Expr) : MetaM (Option Name) :=
     withReducible do -- when testing whether a type is a type class, we only unfold reducible constants.
-      forallTelescopeReducingAux type none fun _ type => do
+      forallTelescopeReducingAux (← instantiateMVars type) none fun _ type => do
```
gebner, line 1019: Is this even necessary?  There's a `whnf` below.

src/Lean/Meta/Basic.lean
```diff
@@ -864,14 +864,23 @@ private partial def isClassQuick? : Expr → MetaM (LOption Name)
   | .mdata _ e       => isClassQuick? e
   | .const n _       => isClassQuickConst? n
   | .mvar mvarId     => do
-    match (← getExprMVarAssignment? mvarId) with
+    match ← getExprMVarAssignment? mvarId with
```
fpvandoorn, line 867:
```suggestion
    match (← getExprMVarAssignment? mvarId) with
```
sorry

src/Lean/Meta/Basic.lean
```diff
@@ -864,14 +864,23 @@ private partial def isClassQuick? : Expr → MetaM (LOption Name)
...
     | some val => isClassQuick? val
     | none     => return .none
-  | .app f _         =>
-    match f.getAppFn with
+  | .app f _         => do
+    match ← getAppFnAssigningMVars f with
     | .const n .. => isClassQuickConst? n
     | .lam ..     => return .undef
```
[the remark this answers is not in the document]
fpvandoorn: We also considered this. We expected that this was more expensive overall, but I'm happy to go with the simpler logic.

src/Lean/Meta/Basic.lean
fpvandoorn: I do not remember whether this change was necessary on this test in particular. 
I think it is a problem when we call this and the argument `type` of `isClassExpensive?` is itself a metavariable, then we might not have telescoped inside all pi-types.
I will investigate, and either add a test where this change is necessary (or confirm it's necessary on the already added test) or I will revert this. (this will take a few days, since I only have built Lean from source on my office computer.)

fpvandoorn: It is necessary, even for this test. I moved it 2 lines lower to a (hopefully) more intuitive location. The reason why it is needed is to make the `type.getAppFn` work: we should instantiate enough metavariables to make sure we get the correct `type.getAppFn`.

src/Lean/Meta/Basic.lean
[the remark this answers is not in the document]
fpvandoorn: I used the simpler logic. 
I had to change the `expected.out` of three tests with this logic.

leodemoura:
Thanks for investigating this issue. It was very useful, but I used a slightly different fix.

PR #2206, opened 2023-05-01
chore: lower monad for Term.reportUnsolvedGoals

kim-em:
No need for this to be in `TermElabM`, and @thorimur [wanted](https://github.com/leanprover-community/mathlib4/pull/3758#discussion_r1181937785) the `MetaM` version.

### DIFF
--- a/src/Lean/Elab/Tactic/Basic.lean
+++ b/src/Lean/Elab/Tactic/Basic.lean
@@ -17,7 +17,7 @@ def admitGoal (mvarId : MVarId) : MetaM Unit :=
 def goalsToMessageData (goals : List MVarId) : MessageData :=
   MessageData.joinSep (goals.map MessageData.ofGoal) m!"\n\n"
 
-def Term.reportUnsolvedGoals (goals : List MVarId) : TermElabM Unit := do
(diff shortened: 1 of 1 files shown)


thorimur:
I notice that this uses `admitGoal` on every unsolved goal! Even though we backtrack the state, if we were to use this for every unsuccessful refl application in every `rfl`, would it be expensive? Is it worth having some `reportUnsolvedGoals'` which doesn't admit the goals?

digama0:
(This is probably better discussed on the mathlib PR) It's not an issue because `admitGoals` is only called on the list of subgoals generated by `apply`. If the refl lemma doesn't apply, then it will fail and we won't get there, and if it does apply then unless it is a malformed refl lemma or there is some funny business with metavariables in the input `apply` will not generate any subgoals to be admitted.

The reason the `admitGoals` is there is so that the remainder of the tactic block can still more or less work and report errors on all the other subgoals. This is the major advantage of tactics that have a statically known set of subgoals (here 0): we can continue processing even if the tactic fails because we know what goals the later tactics would have received. Tactics that do arbitrary things to the goal list completely stop processing on failure because we don't know what would happen next.

gebner (approved):

PR #2211, opened 2023-05-11
feat: trace nodes for kernel type checking

hargoniX:
This is just yet another profileit that gets added to the trace node
feature so we can see it nicely in the flame graphs. I am not quite
happy with the trace node text it provides though. I am also not quite
sure how to improve it in all cases, certain declarations have an obvious
as string representation such as axioms or definitions but it is not
at all obvious to me how to deal with e.g. `mutual` do we just print
`mutual name1 name2 name3 end`?

It would also be cool if our kernel hat tracing so maybe prolonged type
checking could be attributed more easily and people can design their code
to work around it more easily. But that would require to pull the trace node
feature both into C++ and then also into the kernel which I suspect is not
something we want to do, at least not at the moment.

And lastly I think since this adds a trace class it needs an update-stage0?
I was not sure what other class to put it in, it is strictly speaking not part
of elaboration, meta or compilation right?

### DIFF
--- a/src/Lean/CoreM.lean
+++ b/src/Lean/CoreM.lean
@@ -19,6 +19,8 @@ register_builtin_option maxHeartbeats : Nat := {
   descr := "maximum amount of heartbeats per command. A heartbeat is number of (small) memory allocations (in thousands), 0 means no limit"
 }
 
+builtin_initialize registerTraceClass `Kernel
+
 def getMaxHeartbeats (opts : Options) : Nat :=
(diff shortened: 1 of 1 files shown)


Kha:
This is a fine start, the declaration is usually clear from the surrounding `Elab.command` node. But for consistency I think the trace class should be `Kernel`, even if there is no such namespace yet.

Kha (approved):

PR #2238, opened 2023-05-27
fix: panic in Match.SimpH.substRHS

digama0:
As [reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.60simp.60.20panic/near/361446418).


leodemoura:
Merged manually after putting assertion back.

PR #2239, opened 2023-05-28
fix: resolve `_root_` in `macro`, `syntax`, `elab`

digama0:
This is not just a convenience feature: Because `_root_` is resolved in `def`, if we do not also resolve it in `macro` there will be a mismatch between the syntax kind used for the definition and the one used for the parenthesizer / formatter. This comes up in practice for `register_simp_attr`, which is defined like:
```lean
namespace Lean.Meta
...

macro (name := _root_.Lean.Parser.Command.registerSimpAttr) doc:(docComment)?
  "register_simp_attr" id:ident : command => ...
```
leading to a parser that puts the syntax kind `Lean.Meta._root_.Lean.Parser.Command.registerSimpAttr` on generated nodes, and a parenthesizer that is expecting to find syntax kind `Lean.Parser.Command.registerSimpAttr`.


digama0:
Ping on this one, it is the cause of a bug in mathport. How should I address the issue about name joining?

Kha:
> there should definitely be a name function to do this rather than reinventing it all over the place.

Do you want to add such a function (possibly #2341)? Then I think this is good to go. Actually, why do we need the stage 0 update?

digama0:
@Kha Should this PR wait on #2341? I'm fine with landing that and updating this to use it.

digama0:
@Kha perhaps the difference between the two CI runs answers your question about why the update-stage0 is needed.

Kha:
> @Kha Should this PR wait on #2341? I'm fine with landing that and updating this to use it.

Could you cherry-pick it into this PR (and adjust if desired)? I would prefer it to be part of a PR with clear impact

PR #2241, opened 2023-05-29
chore: update Lake

Kha:
Contains https://github.com/leanprover/lake/pull/170

/cc @tydeu


tydeu:
LGTM (I imagine the test failure is just an artifact of the test failing on master already.)

PR #2246, opened 2023-05-30
feat: `hygieneInfo` parser (aka `this` 2.0)

digama0:
As [discussed on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/let.20this/near/362130762), this implements a `hygieneInfo` parser, which captures the hygiene information like `ident`, but without actually parsing anything. This can be used to implement call-site hygienic identifiers like `have this := ...`. The reimplementation of `have this` on top of this parser is done in #2247.

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -3722,6 +3722,16 @@ abbrev nameLitKind : SyntaxNodeKind := `name
 abbrev fieldIdxKind : SyntaxNodeKind := `fieldIdx
 
 /--
+`hygieneInfo` is the node kind of the `hygieneInfo` parser, which is an
+"invisible token" which captures the hygiene information at the current point
+without parsing anything.
+
+They can be used to generate identifiers (with `Lean.HygieneInfo.mkIdent`)
+as if they were introduced by the calling context, not the called macro.
+-/
+abbrev hygieneInfoKind : SyntaxNodeKind := `hygieneInfo
+
+/--
 `interpolatedStrLitKind` is the node kind of interpolated string literal
 fragments like `"value = {` and `}"` in `s!"value = {x}"`.
 -/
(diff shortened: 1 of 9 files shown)


Kha (approved):
Let's wait for #2245 to be merged/resolved first

src/Init/Prelude.lean
```diff
@@ -3721,6 +3721,12 @@ abbrev nameLitKind : SyntaxNodeKind := `name
 /-- `fieldIdx` is the node kind of projection indices like the `2` in `x.2`. -/
 abbrev fieldIdxKind : SyntaxNodeKind := `fieldIdx
 
+/--
+`macroScope` is the node kind of macro scopes, which are "invisible tokens"
+which capture the hygiene information at the current point without parsing anything.
+-/
```
Kha, line 3727:
```suggestion
They can be used to generate identifiers (with `Lean.TSyntax.mkIdent`) as if they were
introduced by the calling context, not the called macro. -/
```
or something to that effect

Kha: Perhaps `TSyntax` is not the ideal namespace for this function even if it is convenient. Or maybe it needs a more explicit name like `mkIdentFromMacroScope`

digama0: I was a bit surprised to see that `getNat` etc are in the `TSyntax` namespace instead of `NatLit`, but I guess it doesn't work if you use that namespace?

Kha:
After thinking a bit more about this change, I'm undecided on the name `macroScope`. It makes sense in that it represents the "scope of a macro" but of course we already have a term of that name, with `macroScope` encoding a *list* of macro scopes.

src/Init/Prelude.lean
Kha: It only works if you have an explicit type annotation `(stx : NumLit)`, but not with `$stx:num`. It's not ideal, no.

digama0: I went with `HygieneInfo.mkIdent`, which is a compromise which sometimes works with dot notation and reads ok even if you can't use dot notation.

Kha:
Merged as part of #2247

PR #2247, opened 2023-05-30
feat: implement `have this`

digama0:
This builds on #2246, to make use of the `hygieneInfo` parser in `have` and `suffices`. It also allows `_` binders for `have` and `let` (rather than treating them as patterns), superseding part of #2105.

### DIFF
--- a/src/Lean/Elab/Deriving/DecEq.lean
+++ b/src/Lean/Elab/Deriving/DecEq.lean
@@ -166,7 +166,7 @@ def mkDecEqEnum (declName : Name) : CommandElabM Unit := do
       fun x y =>
         if h : x.toCtorIdx = y.toCtorIdx then
           -- We use `rfl` in the following proof because the first script fails for unit-like datatypes due to etaStruct.
-          isTrue (by first | refine_lift let_fun aux := congrArg $ofNatIdent h; ?_; rw [$auxThmIdent:ident, $auxThmIdent:ident] at aux; assumption | rfl)
+          isTrue (by first | have aux := congrArg $ofNatIdent h; rw [$auxThmIdent:ident, $auxThmIdent:ident] at aux; assumption | rfl)
         else
           isFalse fun h => by subst h; contradiction
   )
--- a/src/Lean/Elab/Do.lean
+++ b/src/Lean/Elab/Do.lean
@@ -624,11 +624,11 @@ def getDoLetVars (doLet : Syntax) : TermElabM (Array Var) :=
   -- leading_parser "let " >> optional "mut " >> letDecl
   getLetDeclVars doLet[2]
 
-def getHaveIdLhsVar (optIdent : Syntax) : TermElabM Var :=
-  if optIdent.isNone then
-    `(this)
+def getHaveIdLhsVar (optIdent : Syntax) : Var :=
+  if optIdent.getKind == hygieneInfoKind then
+    HygieneInfo.mkIdent optIdent[0] `this
   else
-    pure optIdent[0]
+    optIdent
 
 def getDoHaveVars (doHave : Syntax) : TermElabM (Array Var) := do
   -- doHave := leading_parser "have " >> Term.haveDecl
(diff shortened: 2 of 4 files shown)


leodemoura:
> @leodemoura Are you on board with consistently moving towards binderIdent for binders where it makes sense? I'm willing to review https://github.com/leanprover/lean4/pull/2105 if Mario is interested in updating and finishing it.

I am onboard if it does *not* create a "snowball" of new issues.

digama0:
Hm, I am not sure what I can do to determine how likely it is to create a cascade of other issues. What I can do concretely is fix it up so it passes all the tests, and try bumping std and mathlib to make sure they also don't break. (I'll need to wait until just after an actual nightly bump to do that test.) Hopefully the changes to the code and any test changes made as a result will help answer the question of what the concrete fallout will be.

I am optimistic that will have essentially no effects on user code, although it will involve changing some tactics to handle the `hole` case.

Kha:
> the reason I added it is because without it `have _ := foo` is interpreted as `have this _ := foo` instead of `have _x := foo`

Hmm, don't we have the same issue with any other kind of pattern then? `have \<x, y\>` should not mean `have this \<x, y\>` either

Kha:
If we do rebootstrap the PR, one remark and one question:
* We should probably use `mkAntiquotation (anonymous := false)`, we really don't want to capture an invisible piece of syntax by default
* Is it clear why we need two update-stage0s? Did you try using `parseQuotWithCurrentStage` like described in bootstrap.md?

digama0:
> Hmm, don't we have the same issue with any other kind of pattern then? `have \<x, y\>` should not mean `have this \<x, y\>` either

No, because `\<x, y\>` is not a binder. You can't write `have foo \<x, y\> := ...`, it needs to be `have foo | \<x, y\> => ...` or something.

digama0:
> If we do rebootstrap the PR,

I'm not sure exactly what you mean by this

>     * We should probably use `mkAntiquotation (anonymous := false)`, we really don't want to capture an invisible piece of syntax by default

>     * Is it clear why we need two update-stage0s? Did you try using `parseQuotWithCurrentStage` like described in bootstrap.md?

I tried my best to optimize it, and I am using `parseQuotWithCurrentStage` in the middle stage. The current staging looks like this:

* Part 1: Prepare `have` to not break when seeing `hygieneInfo`. This is useless in the first stage but it prevents breakage in stage 2.
* Update stage 0: now the builtin `have` understands `hygieneInfo`
* Part 2: Main implementation. `parseQuotWithCurrentStage` is enabled, the new grammar productions are added, and the new `have` implementation uses the new grammar productions. Because of `parseQuotWithCurrentStage`, various things are using the new syntax and passing it through the builtin `have` implementation, which is why we needed part 1.
  * There is a use of `have` inside the DecidableEq handler, which is parsed using the old grammar and will cause errors in stage 3. The easiest way to get around this is to just swap out the implementation for something else temporarily.
* Update stage 0: now the new notation is fully installed
* Part 3: cleanup staging effects, fix tests

Kha:
> No, because `\<x, y\>` is not a binder.

Ah, of course. Then I don't have any remaining reservations on this PR anymore.



> I'm not sure exactly what you mean by this

Just regenerating the stage0 updates.



> via a script

Does it use `git rebase --exec`? Would be nice to contribute and document!

Kha:
I just checked out the PR, it builds fine without the stage0 updates except for the `deriving DecidableEq` issue. Changing a built-in syntax like term `have` that is depended on by non-builtin syntax like tactic `have` is an unfortunate known issue.

digama0:
> I just checked out the PR, it builds fine without the stage0 updates except for the `deriving DecidableEq` issue. Changing a built-in syntax like term `have` that is depended on by non-builtin syntax like tactic `have` is an unfortunate known issue.

are you saying you can build it if you merge part 1 & 2? I'm pretty sure I tried that, it fails right at the start of part 3.

digama0:
right, if you just don't do any staging then you won't be able to boostrap, that's what a failure in stage 2 is. If you run `update-stage0` then the `Init` package can't be parsed

Kha:
I don't see why parsing would be affected. I get a very late error due to the deriving issue after rebootstrapping:
```
Lean/Meta/Tactic/LinearArith/Solver.lean:130:22: error: tactic 'rfl' failed, equality lhs
  x✝
is not definitionally equal to rhs
  y✝
x✝ y✝ : CnstrKind
h✝ : CnstrKind.toCtorIdx x✝ = CnstrKind.toCtorIdx y✝
⊢ x✝ = y✝
```

Kha:
Just to be perfectly clear, this is my history: 
image

digama0:
what happens if you put the update stage0 between part 2 and part 3? The main thing that part 3 does is to take out the `have -> let_fun` hack which was done to get around the issue that you are reporting.

Kha:
> what happens if you put the update stage0 between part 2 and part 3?

Right, that makes more sense. But it builds and tests fine.

digama0:
Okay, I updated it with only two stages and added the `(anonymous := false)` (which is actually part of the other PR), and rebased both PRs on master.

Kha:
Thanks!

Kha:
@digama0 Ah, do you want to add a release note as well describing the user-level changes?

digama0:
is there a special syntax that I put in the PR description?

Kha:
No, feel free to open a new PR adjusting RELEASES.md

PR #2253, opened 2023-06-02
fix: prefer resolving parser alias over declaration

### LINKED ISSUE #2249: incorrect arity calculation for `Parser` parsers in `syntax`
```lean
import Lean

macro (name := fooParser) "foo" x:(ppSpace ident)* : term => 
  `([$(x.map (Lean.mkIdent ·.getId)),*])

namespace Lean.Parser.Term
macro (name := barParser) "bar" x:(ppSpace ident)* : term => 
  `([$(x.map (Lean.mkIdent ·.getId)),*])

variable (a : Nat)
#check foo a -- ok
#check bar a -- unknown identifier '[anonymous]'
```

The reason for the difference is that when `Lean.Parser.Term` is open, `ppSpace` refers to `Lean.Parser.Term.ppSpace` (the actual definition) rather than `ppSpace` (the parser alias), and the arity calculation for parsers is incorrect:

https://github.com/leanprover/lean4/blob/28538fc74889ea2dedbc30e236ce26492e5823ca/src/Lean/Elab/Syntax.lean#L200-L204

this assumes that all `Parser`s have arity 1, which is not true for `ppSpace` and several other parsers. The result is that an extra `group` node is inserted in one of these parsers:
```lean
#print fooParser
-- def fooParser : ParserDescr :=
-- ParserDescr.node `fooParser 1022
--   (ParserDescr.binary `andthen (ParserDescr.symbol "foo")
--     (ParserDescr.unary `many
--       (ParserDescr.binary `andthen (ParserDescr.const `ppSpace)
--         (ParserDescr.const `ident))))

#print barParser
-- def Lean.Parser.Term.barParser : ParserDescr :=
-- ParserDescr.node `Lean.Parser.Term.barParser 1022
--   (ParserDescr.binary `andthen (ParserDescr.symbol "bar")
--     (ParserDescr.unary `many
--       (ParserDescr.unary `group
--         (ParserDescr.binary `andthen (ParserDescr.parser `Lean.Parser.ppSpace)
--           (ParserDescr.parser `Lean.Parser.Term.ident)))))
```
 
My suggestion to fix this is to move the arity declaration from the alias to the `ParserInfo`.

cc: @Kha

Kha:
Resolves #2249. Aliases have more metadata than the corresponding parser decl, and there is no syntax to disambiguate them otherwise.

### DIFF
--- a/src/Lean/Elab/Syntax.lean
+++ b/src/Lean/Elab/Syntax.lean
@@ -191,26 +191,31 @@ where
     return (stx, stackSz)
 
   processNullaryOrCat (stx : Syntax) := do
-    match (← elabParserName? stx[0]) with
+    let ident := stx[0]
+    let id := ident.getId.eraseMacroScopes
+    -- run when parser is neither a decl nor a cat
+    let default := do
+      if (← Parser.isParserAlias id) then
+        ensureNoPrec stx
+        return (← processAlias ident #[])
+      throwError "unknown parser declaration/category/alias '{id}'"
+    match (← elabParserName? ident) with
     | some (.parser c (isDescr := true)) =>
       ensureNoPrec stx
       -- `syntax _ :=` at least enforces this
       let stackSz := 1
       return (mkIdentFrom stx c, stackSz)
     | some (.parser c (isDescr := false)) =>
+      if (← Parser.getParserAliasInfo id).declName == c then
+        -- prefer parser alias over base declaration because it has more metadata, #2249
+        return (← default)
       ensureNoPrec stx
       -- as usual, we assume that people using `Parser` know what they are doing
       let stackSz := 1
       return (← `(ParserDescr.parser $(quote c)), stackSz)
     | some (.category _) =>
       processParserCategory stx
-    | none =>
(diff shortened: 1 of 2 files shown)


src/Lean/Elab/Syntax.lean
```diff
@@ -191,26 +191,31 @@ where
...
+      throwError "unknown parser declaration/category/alias '{id}'"
+    match (← elabParserName? ident) with
     | some (.parser c (isDescr := true)) =>
       ensureNoPrec stx
       -- `syntax _ :=` at least enforces this
       let stackSz := 1
       return (mkIdentFrom stx c, stackSz)
     | some (.parser c (isDescr := false)) =>
+      if (← Parser.getParserAliasInfo id).declName == c then
+        -- prefer parser alias over base declaration because it has more metadata, #2249
```
Kha, line 210: I decided to restrict the change to the specific issue at hand where preferring the declaration is never useful. Looking at https://github.com/leanprover/lean4/blob/master/tests/lean/run/parserAliasShadow.lean, it didn't appear to be the correct choice in general anymore.

PR #2254, opened 2023-06-02
chore: update changelog

digama0:
Add an entry for #2247. I elected not to mention either of the two main features actually implemented (`have _` and the `hygieneInfo` matcher) because they are of a more technical nature. The more user-visible effect is the unkeywordification of `this`.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,6 +1,19 @@
 Unreleased
 ---------
 
+* [`hygieneInfo` parser, new `have this` implementation](https://github.com/leanprover/lean4/pull/2247).
+
+  Previously, `this` was a keyword which macro-expands to an identifier, but this
+  was a leaky abstraction resulting in several places where the name `this` has
+  divergent behavior compared to other variable names:
+  ```lean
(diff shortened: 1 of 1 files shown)


Kha:
My bad, presented like this it does sound like a straight bugfix, which we don't usually document. I was thinking about the changed scoping behavior of `this`, but it seems quite unlikely that anyone made use of it across macro scopes in a way that would now break. What do you think?

Kha:
Or perhaps more succinctly: "`this` is now an identifier again that is implicitly introduced by anonymous `have :=` with regular scoping. It used to be a keyword that was visible in all scopes and led to unexpected behavior when explicitly used as a binder name.

digama0:
> I was thinking about the changed scoping behavior of `this`, but it seems quite unlikely that anyone made use of it across macro scopes in a way that would now break. What do you think?

I don't think anyone has used `this` in a macro, at least not that I have seen in mathlib and certainly not in the two or more nested macro-macros case that would actually be affected by this.

I am fine with your rewording. I was probably too hard on the original version so it is probably best for you to phrase it.

PR #2255, opened 2023-06-05
feat: report section name in invalid end msg

digama0:
As [requested on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/section.20name.20mismatch/near/363562505).


leodemoura:
Could you please add a new test?

PR #2259, opened 2023-06-06
fix: `save` when used as last tactic

Kha:
@tobiasgrosser noticed that `save` doesn't do anything when it is the last tactic in the block before or after an edit. It turns out it is a simple off-by-one error.

### DIFF
--- a/src/Lean/Elab/Tactic/BuiltinTactic.lean
+++ b/src/Lean/Elab/Tactic/BuiltinTactic.lean
@@ -69,11 +69,14 @@ Output:
 -/
 -- Note that we need to preserve the separators to show the right goals after semicolons.
 def addCheckpoints (stx : Syntax) : TacticM Syntax := do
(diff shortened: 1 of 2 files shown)


src/Lean/Elab/Tactic/BuiltinTactic.lean
```diff
@@ -83,7 +86,7 @@ def addCheckpoints (stx : Syntax) : TacticM Syntax := do
           mkNode ``tacticSeq #[
             mkNode ``tacticSeq1Indented #[
               -- HACK: null node is not a valid tactic, but prevents infinite loop
-              mkNullNode (currentCheckpointBlock.push tac)
+              mkNullNode (currentCheckpointBlock.push (mkNullNode #[tac]))
```
Kha, line 89: The outer `mkNullNode` is actually necessary, the hack was not necessary before because it was masked by the off-by-one bug

PR #2264, opened 2023-06-09
doc: fix contradictory docstring

Vtec234:
@PatrickMassot pointed out to me that the docstring on `SubExpr` was false.


digama0:
What is the contradiction? All I can see is that you moved the docstring.

PR #2271, opened 2023-06-13
refactor: redefine `String.splitOnAux`

chabulhwi:
Remove a redundant part of the definition of `String.splitOnAux`.


Kha:
Please revert the second change, it is not semantics-preserving

PR #2275, opened 2023-06-15
 fix: delete Measure and SizeOfRef

fpvandoorn:
Fixes an overload with Mathlib's `MeasureTheory.Measure` that causes performance issues, as described here: https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Measure.20is.20overloaded

`Measure` and `SizeOfRef` were unused.

### DIFF
--- a/src/Init/WF.lean
+++ b/src/Init/WF.lean
@@ -184,14 +184,8 @@ protected theorem caseStrongInductionOn
 
 end Nat
 
-def Nat.Measure {α : Sort u} : (α → Nat) → α → α → Prop :=
-  InvImage (fun a b => a < b)
-
 abbrev measure {α : Sort u} (f : α → Nat) : WellFoundedRelation α :=
   invImage f Nat.lt_wfRel
-
-def SizeOfRef (α : Sort u) [SizeOf α] : α → α → Prop :=
-  Nat.Measure sizeOf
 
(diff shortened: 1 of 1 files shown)


semorrison:
Simply inlining the definition at `SizeOfRef` would also work.

gebner:
Historically speaking, these definitions date from a time when you had to specify the well-founded relation manually.  Nowadays this all happens automatically via the HasWellFounded type class, and you should never need to use this measure function here.

With that in mind (and considering the name clashes), I think it would be best to just remove these definitions here.

digama0:
I'm pretty sure there are occurrences of `measure` in std4 / mathlib4, I would recommend testing before removing that one.

semorrison:
I've (started) creating a release. Once [this](https://github.com/semorrison/lean4/actions/runs/5290881325) is finished, you'll be able to use `semorrison/lean4:release-20230616-measure` in your `lean-toolchain` to try out this PR. I can have a look tomorrow.

semorrison:
This [build](https://github.com/leanprover-community/mathlib4/actions/runs/5296103839/jobs/9586976108) of mathlib4 verifies that everything compiles unchanged.

leodemoura:
> [leanprover-community/mathlib4#5259](https://github.com/leanprover-community/mathlib4/pull/5259) contains the minor adaptations required for mathlib4 to compile against this PR.

@semorrison I will merge this PR tonight. Please say something if you disagree.

PR #2276, opened 2023-06-19
fix: respect type reducibility when generating injectivity lemmas

eric-wieser:
The relevant test case is:
```lean
def WrappedNat (_n : Nat) := Nat

structure WithWrapped : Type :=
(n : Nat)
(m : WrappedNat n)

#check WithWrapped.mk.injEq
#print WithWrapped.noConfusionType
```
The diff due to this PR is intended to be
```diff
 WithWrapped.mk.injEq (n : Nat) (m : WrappedNat n) (n : Nat) (m : WrappedNat n) :
-  ({ n := n, m := m } = { n := n, m := m }) = (n = n ∧ m = m)
+  ({ n := n, m := m } = { n := n, m := m }) = (n = n ∧ HEq m m)
@[reducible] protected def WithWrapped.noConfusionType.{u} : Sort u → WithWrapped → WithWrapped → Sort u :=
-fun P v1 v2 => WithWrapped.casesOn v1 fun n m => WithWrapped.casesOn v2 fun n_1 m_1 => (n = n_1 → m = m_1 → P) → P
+fun P v1 v2 => WithWrapped.casesOn v1 fun n m => WithWrapped.casesOn v2 fun n_1 m_1 => (n = n_1 → HEq m m_1 → P) → P
```

This forward-ports https://github.com/leanprover-community/lean/pull/812.

For now, I am just creating this to see if it passes CI, since CI appears not to run on my branch in a fork.

### DIFF
--- a/src/library/constructions/no_confusion.cpp
+++ b/src/library/constructions/no_confusion.cpp
@@ -101,6 +101,7 @@ static optional<environment> mk_no_confusion_type(environment const & env, name 
                         expr rhs_type = lctx.get_type(rhs);
                         level l       = sort_level(type_checker(env, lctx).ensure_type(lhs_type));
                         expr h_type;
+                        // TODO: this should respect reducibility
                         if (type_checker(env, lctx).is_def_eq(lhs_type, rhs_type)) {
                             h_type = mk_app(mk_constant(get_eq_name(), levels(l)), lhs_type, lhs, rhs);
                         } else {


Kha:
The CI failure seems to be a simple matter of reordered messages; do we agree on the general change?

leodemoura:
> The CI failure seems to be a simple matter of reordered messages.

I don't look at PRs until it is green. We should make it clear to all PR submitters that this is a requirement.

> do we agree on the general change?

The change is not properly motivated. A link to another thread is not an acceptable. We should also make it clear on the contribution guidelines that a clear and short motivation for unsolicited PRs is a requirement.

src/library/constructions/no_confusion.cpp
```diff
@@ -101,6 +101,7 @@ static optional<environment> mk_no_confusion_type(environment const & env, name
                         expr rhs_type = lctx.get_type(rhs);
                         level l       = sort_level(type_checker(env, lctx).ensure_type(lhs_type));
                         expr h_type;
+                        // TODO: this should respect reducibility
```
eric-wieser, line 104: In lean 3, this was possible with `type_context_old ctx(env, transparency_mode::Reducible);`.

I think the issue here is that `mk_no_confusion_type` is using the kernel, when it should be using the elaborator; but fixing that is a large refactor.

PR #2286, opened 2023-06-24
chore: compact objects in post-order

digama0:
This changes the order of the .olean compactor to use post-order traversal, rather than reverse pre-order. (That is, given an object graph like `a(b(c, d), e)` it used to produce an olean file with the layout `e d c b a` and now it produces `c d b e a`.) While not strictly required, it is a relatively simple fix which makes it possible to get bit-for-bit identical olean files from https://github.com/digama0/leangz, which has many other applications.

### DIFF
--- a/src/runtime/compact.cpp
+++ b/src/runtime/compact.cpp
@@ -168,14 +168,16 @@ void object_compactor::insert_string(object * o) {
 
 bool object_compactor::insert_constructor(object * o) {
     std::vector<object_offset> & offsets = m_tmp;
-    offsets.clear();
     bool missing_children = false;
     unsigned num_objs     = lean_ctor_num_objs(o);
-    for (unsigned i = 0; i < num_objs; i++) {
+    offsets.resize(num_objs);
+    unsigned i = num_objs;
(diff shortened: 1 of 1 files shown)


src/runtime/compact.cpp
```diff
@@ -168,14 +168,14 @@ void object_compactor::insert_string(object * o) {
 
 bool object_compactor::insert_constructor(object * o) {
     std::vector<object_offset> & offsets = m_tmp;
-    offsets.clear();
     bool missing_children = false;
     unsigned num_objs     = lean_ctor_num_objs(o);
-    for (unsigned i = 0; i < num_objs; i++) {
+    offsets.resize(num_objs);
+    for (unsigned i = num_objs; i-- > 0;) {
```
leodemoura, line 174: Could you please use the style we use in the rest of the code base?
```cpp
        unsigned i = num_objs;
        while (i > 0) {
            i--;
            // ....
```
Same for the other for-loop in the commit.

leodemoura:
Could you please rename the PR since it is not fixing but changing the .olean format?

PR #2293, opened 2023-06-27
Make token parser customizable

Kha:
In theory, users can implement arbitrary lexical grammars as the built-in token parser is merely another combinator whose use can be avoided; in practice, avoiding it also means forgoing the existing infrastructure around token caching and further parsers built on top of `tokenFn`, such as `symbol` implicitly used by `syntax`.

There are other nice-to-haves we have talked about before like per-category token tables, but this PR implements the most simple change to address the issues mentioned above: it makes the implementation of `tokenFn` customizable such that people can implement their own tokenization in it without losing access to all the infrastructure built on top of it.

Implementation-wise, the change mostly consist of further splitting `ParserContextCore/ParserFn` to prevent non-wellfounded type recursion and moving the Lean language's tokenizer into a new module `Lean.Parser.LeanToken` to distinguish it from language-generic parsers in `Basic` (which is too big anyway). Making it easier to adapt these Lean token functions for implementing a custom token parser is left as future work.

The change is validated by a simple test that changes the whitespace/comment syntax of the existing Lean language as well as an incomplete implementation of the venerable Whitespace language.

### DIFF
--- a/src/Lean/Parser/Types.lean
+++ b/src/Lean/Parser/Types.lean
@@ -283,17 +283,26 @@ def mkUnexpectedErrorAt (s : ParserState) (msg : String) (pos : String.Pos) : Pa
 
 end ParserState
 
+/--
+  Parser context available to token parsers. It does not give access to the current tokenizer function
+  in order to prevent cyclic dependencies. -/
 structure TokenParserContext extends InputContext, ParserModuleContext, CacheableParserContext where
-  tokens  : TokenTable
+  /-- Current token table, available to token parsers. -/
+  tokens : TokenTable
 
+/-- A token parser. See `TokenParserContext`. -/
 def TokenParserFn := TokenParserContext → ParserState → ParserState
 
 instance : Inhabited TokenParserFn where
   default := fun _ s => s
 
-/-- Parser context updateable in `adaptUncacheableContextFn`. -/
+/--
+  Parser context updateable in `adaptUncacheableContextFn`.
+  On top of `CacheableParserContext`, it includes fields such as `tokens` and `tokenFn` that cannot easily
+  be included in the cache key. -/
 structure ParserContextCore extends TokenParserContext where
-  tokenFn      : TokenParserFn
+  /-- Current tokenizer function invoked by `Lean.Parser.tokenFn` on cache miss. -/
+  tokenFn : TokenParserFn
 
 /-- Opaque parser context updateable using `adaptCacheableContextFn` and `adaptUncacheableContextFn`. -/
 structure ParserContext extends ParserContextCore where private mk ::


Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
I have now benchmarked the PR on https://github.com/Kha/C-parsing-for-Lean4/tree/custom-tokens as well, before applying any changes to the actual parsing. Performance slightly increases, which doesn't really make sense and as such is probably a linker artifact, but it is at least weak evidence that performance of parsing is not negatively affected
```
$ lake build; nix run nixpkgs#hyperfine ./build/bin/cParser -- --warmup 1
Benchmark 1: ./build/bin/cParser
  Time (mean ± σ):      1.233 s ±  0.008 s    [User: 1.139 s, System: 0.095 s]
  Range (min … max):    1.224 s …  1.251 s    10 runs
$ git co custom-tokens; lake build; nix run nixpkgs#hyperfine ./build/bin/cParser -- --warmup 1
Benchmark 1: ./build/bin/cParser
  Time (mean ± σ):      1.187 s ±  0.007 s    [User: 1.102 s, System: 0.087 s]
  Range (min … max):    1.178 s …  1.198 s    10 runs
```

Kha:
Just to note some alternatives: instead of making the token parser function dynamically replaceable, we could also "simply" demand embedded languages to call their own token parser function directly. This would involve duplication around all those little functions that build on top of the token parser though, and `syntax` as well would somehow have to be taught to use a different `symbol` function. Making the switch to a different token parser explicit via the parser context also means we can reliable enforce that the token cache is cleared at that point like done in this PR.
Furthermore, we could tie token parsers to syntax categories such that they are automatically activated when the category is entered (much like we might want to do with per-category token tables?). This is something that can be done on top of this PR however.

src/Lean/Parser/Types.lean
```diff
@@ -290,11 +283,29 @@ def mkUnexpectedErrorAt (s : ParserState) (msg : String) (pos : String.Pos) : Pa
 
 end ParserState
 
+structure TokenParserContext extends InputContext, ParserModuleContext, CacheableParserContext where
```
leodemoura, line 286: Please add comment.

src/Lean/Parser/Types.lean
```diff
@@ -290,11 +283,29 @@ def mkUnexpectedErrorAt (s : ParserState) (msg : String) (pos : String.Pos) : Pa
...
+  tokens  : TokenTable
+
+def TokenParserFn := TokenParserContext → ParserState → ParserState
+
+instance : Inhabited TokenParserFn where
+  default := fun _ s => s
+
+/-- Parser context updateable in `adaptUncacheableContextFn`. -/
+structure ParserContextCore extends TokenParserContext where
```
leodemoura, line 295: Please expand comment and describe the purpose of the new structure.

src/Lean/Parser/LeanToken.lean
```diff
@@ -0,0 +1,326 @@
+import Lean.Parser.Basic
+
+/-! Tokenization functions used for the implementation of the Lean language. -/
```
leodemoura, line 3: Have you just moved the code? Or, have you made modifications too?

leodemoura:
Let's discuss this PR in the next meeting.

src/Lean/Parser/Basic.lean
```diff
@@ -422,379 +433,63 @@ def withResultOf (p : Parser) (f : Syntax → Syntax) : Parser := {
...
-
-def quotedCharFn : ParserFn :=
-  quotedCharCoreFn isQuotableCharDefault
-
-/-- Push `(Syntax.node tk <new-atom>)` onto syntax stack if parse was successful. -/
-def mkNodeToken (n : SyntaxNodeKind) (startPos : String.Pos) : ParserFn := fun c s => Id.run do
+/-- Function to be called directly after parsing a token. When not in an error state, parses following whitespace,
+    sets up `SourceInfo`, and pushes result of calling `f` with token substring and info onto stack. -/
+@[specialize]
+def pushToken (f : Substring → SourceInfo → Syntax) (startPos : String.Pos) (whitespaceFn : TokenParserFn) :
```
leodemoura, line 461: I want to discuss this one. I see the motivations for the change, but the parser is starting to accumulate "patchwork."

src/Lean/Parser/LeanToken.lean
Kha: Only changes for moving functions from `ParserFn` to `TokenParserFn`, plus https://github.com/Kha/lean4/commit/bd44113a58f385e687a5beb12f20c6ff29257a24

Kha:
!bench

tydeu:
@Kha I noticed that you removed the whitespace customization. While it may be really used in the Lean core, I think it will be very useful for DSLs. Particularly, for those that are meant to function in both an embedded context and standalone. For example, in Alloy, when parsing C code in a Lean file, using Lean-style comments is ideal. However, when parsing a C file itself using C-style comments is necessary. Being able to use the same grammar for both would be amazing and would be enabled by the whitespace configuration.

Kha:
Existing token functions can simply be parameterized over the whitespace parser if desired, the parser framework doesn't have to be changed for that

tydeu:
@Kha 

> Existing token functions can simply be parameterized over the whitespace parser if desired,

In my case, that would still require a separate grammar for C w/ Lean comments and C w/ C comments. Right? (Since the whitespace function has to be determine a syntax level in such a setup rather than at parser runtime. The only other solution I can see would be to parameterize the entire grammar which would prohibit the use of `syntax`.)

Issue #2297, opened 2023-06-28
rfc: allow universe unification to solve `max u v =?= max u ?v`


semorrison:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

Currently universe unification will not solve `max u v =?= max u ?v` or `max u v =?= max v ?u`. (Lean 3 would do this.)
This is causing some problems in mathlib4. 

### Steps to Reproduce

```
universe u v

-- This is a mock-up of the `HasLimitsOfSize` typeclass in mathlib4.
class HLOS.{a,b} (C : Type b) where
  P : Type a

-- We only have an instance when there is a "universe inequality".
instance HLOS_max : HLOS.{a} (Type max a b) := sorry

example : HLOS.{a} (Type max a b) := HLOS_max.{a, max a b} -- Success
example : HLOS.{a} (Type max a b) := inferInstance -- Fails: stuck at max a b =?= max a ?u

-- In mathlib4 we currently make use of the following workaround:
abbrev TypeMax := Type max u v

instance (priority := high) HLOS_max' : HLOS.{a} (TypeMax.{a, b}) := sorry

example : HLOS.{a} (TypeMax.{a, b}) := HLOS_max'.{a} -- Success
example : HLOS.{a} (TypeMax.{a, b}) := inferInstance -- Success

-- However, these still fail
example : Type max v u = TypeMax.{v} := rfl -- Fails: `max u v =?= max v ?u`
example : Type max v u = TypeMax.{u} := rfl -- Fails: `max u v =?= max u ?u`

-- and there are situations in which, however much we try to use `TypeMax`,
-- we find failing syntheses of: `HLOS.{a} (Type max a b)` 
```

**Expected behavior:**

It would be nice if this all succeeded!

**Actual behavior:**

Failures with `stuck at max a b =?= max a ?u`.

**Reproduces how often:**

Consistently.

### Versions

```
Lean (version 4.0.0-nightly-2023-06-20, commit a44dd71ad62a, Release)
```

### Additional Information

I have a small patch to Lean 4, which resolves this issue in a fairly conservative manner:
```
diff --git a/src/Lean/Meta/LevelDefEq.lean b/src/Lean/Meta/LevelDefEq.lean
index d8bdfef123..53ad40ddd7 100644
[...]

PR #2298, opened 2023-06-28
rfc: allow universe unification to solve `max u v =?= max u ?v`

kim-em:
See issue description at #2297.

### DIFF
--- a/src/Lean/Meta/LevelDefEq.lean
+++ b/src/Lean/Meta/LevelDefEq.lean
@@ -81,6 +81,11 @@ mutual
         match (← Meta.decLevel? v) with
         | some v => Bool.toLBool <$> isLevelDefEqAux u v
         | none   => return LBool.undef
+    -- Handle `max u ?_ =?= max u v` and `max u ?_ =?= max v u` for `u` a parameter.
+    | Level.max p@(.param _) m@(.mvar _), Level.max u v =>
+      if u == p then solve m v
+      else if v == p then solve m u
+      else return LBool.undef
     | _, _ => return LBool.undef
 
   @[export lean_is_level_def_eq]
--- a/tests/lean/run/2297.lean
+++ b/tests/lean/run/2297.lean
@@ -1,0 +1,22 @@
+universe u v
+
+-- This is a mock-up of the `HasLimitsOfSize` typeclass in mathlib4.
+class HLOS.{a,b} (C : Type b) where
+  P : Type a
+
+-- We only have an instance when there is a "universe inequality".
+instance HLOS_max : HLOS.{a} (Type max a b) := sorry
+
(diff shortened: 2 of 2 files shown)


tydeu:
@semorrison Should not this be symmetric? That is, if Lean solves `max u v =?= max u ?v` should it not also solve `max u v =?= max ?u v`?

semorrison:
@tydeu, my understanding is that `max`s have already been sorted by the time they arrive here, in a way that ensures the mvars don't appear in the first slot.

tydeu:
@semorrison Ah, okay, makes sense. 👍

src/Lean/Meta/LevelDefEq.lean
```diff
@@ -81,6 +81,11 @@ mutual
         match (← Meta.decLevel? v) with
         | some v => Bool.toLBool <$> isLevelDefEqAux u v
         | none   => return LBool.undef
+    -- Handle `max u ?_ =?= max u v` and `max u ?_ =?= max v u` for `u` a parameter.
+    | Level.max p@(.param _) m@(.mvar _), Level.max u v =>
+      if u == p then solve m v
+      else if v == p then solve m u
+      else return LBool.undef
```
semorrison, line 88:
```suggestion
    | Level.max p@(.param _) (.mvar m), Level.max u v =>
      if u == p then
        assignLevelMVar m v
        return LBool.true
      else if v == p then
        assignLevelMVar m u
        return LBool.true
      else return LBool.undef
```
also works here, and is perhaps more conservative as there's no recursive call to `solve`, but I think should have the same behaviour. It also compiles mathlib4 fine, and fixes our problems there.

leodemoura:
> @tydeu, my understanding is that `max`s have already been sorted by the time they arrive here, in a way that ensures the mvars don't appear in the first slot.

@semorrison You must confirm that. You also need to properly document `LevelDefEq.lean` and add a Note (in this module) about the  change that has been made.

semorrison:
I'm going to label this as low priority for now, in the context of the `UnivLE` workaround I developed in Mathlib seeming to work for now. (See https://github.com/leanprover-community/mathlib4/pull/5723, https://github.com/leanprover-community/mathlib4/pull/5724, https://github.com/leanprover-community/mathlib4/pull/5726.) I don't think `UnivLE` is a complete solution, so this may be worth revisiting later.

One unfortunate aspect of the current set up is that universe unification issues as in the title often appear in error messages that obscure the "actual" error that the user needs to address (usually an instance not being available). We might think about how to improve that situation without changing the unification algorithm, although at present I don't have ideas in that direction. Anyone coming to this issue in the meantime who has been tripped up by the error messages: creating a separate issue about the error message situation would be a great intermediate step.

semorrison:
Superseded by https://github.com/leanprover/lean4/pull/3981

PR #2299, opened 2023-06-28
chore: write `"|-"` as `"|" noWs "-"`

fpvandoorn:
* The old way made `|-` a token, which conflicts with the parsing of `|-a|`.
* The [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.7C-) has more discussion.

### DIFF
--- a/tests/lean/run/2299.lean
+++ b/tests/lean/run/2299.lean
@@ -1,0 +1,7 @@
+class Abs (α : Type _) where
+  abs : α → α
+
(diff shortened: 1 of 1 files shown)


leodemoura:
@fpvandoorn It is fine to remove `|-`, but could you please add some ASCII alternative for `⊢` ? See Sebastian's comment in the Zulip thread.

fpvandoorn:
Ok, I'll do that.

fpvandoorn:
I think this is good to go. I added a test that fails on current `master`.

PR #2303, opened 2023-06-30
chore: update Lake

tydeu:
Update Lake to the latest prerelease. Fixes an [issue](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/lake.20build.20all/near/370788618) reported on Zulip.


tydeu:
Updated due to adding some last minute helpers for [mathlib4#5593](https://github.com/leanprover-community/mathlib4/pull/5593#issuecomment-1613888248).

PR #2307, opened 2023-07-04
Kill descendant processes on worker exit

Kha:
This fixes the Linux/macOS issue of the language server leaking build processes, which unnecessarily consume CPU and race with other builds on the output directory, making users revert to frequent `killall lean`


semorrison:
Just tested locally (macos), and it works great:

* run `lake clean` in mathlib4
* open VSCode, open a file deep in the hierarchy
* observe that many lake/lean processes have started up
* close VSCode

Before this PR, all the processes continue, after this PR, they all promptly disappear!

Kha:
Happy for feedback on the new APIs, but otherwise I think this should be ready now that it's tested. Will merge later if there are no objections.

PR #2310, opened 2023-07-06
feat: use nat_pow in the kernel

kim-em:
Enable using the existing `lean_nat_pow` function in the kernel.

This enables mathlib4's Mersenne prime checker to get beyond `2^23209 - 1` (from 1979). It's now done `2^216091 - 1` (from 1985), and I suspect that the current implementation, plus this tweak, would be sufficient to get (slowly) to the modern era if someone wanted to spend the CPU time.

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -1471,7 +1471,7 @@ set_option bootstrap.genMatcherCode false in
 /--
 The power operation on natural numbers.
 
-This definition is overridden in the compiler to efficiently
+This definition is overridden in both the kernel and the compiler to efficiently
 evaluate using the "bignum" representation (see `Nat`). The definition provided
 here is the logical model.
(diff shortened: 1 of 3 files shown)


digama0:
@leodemoura Just so you know, after this PR I posted a fix to norm_num (linked above) which solves the motivating issue for this change (it was not supposed to be as slow as it was). This modification is still a ~2x performance boost over building proof terms as in the linked issue, but it is no longer the 1000x improvement quoted in the PR description. I am okay with either keeping the new function or reverting it, but it is not essential to us.

digama0:
There is also a missing `Nat.pow` around [here](https://github.com/leanprover/lean4/blob/a3ebfe29ea516e45dba97dfd303cfc11cc47ff04/src/Lean/Meta/WHNF.lean#L820).

PR #2316, opened 2023-07-10
Compile against glibc 2.26

Kha:


Kha:
Yikes, the ARM CI jobs were not registered as mandatory checks yet...

digama0:
The ARM jobs are cross-compiles, so I don't think they can be easily tested.

Kha:
They should at least build though

Kha:
Reverted in https://github.com/leanprover/lean4/commit/a3ebfe29ea516e45dba97dfd303cfc11cc47ff04

gebner:
> The ARM jobs are cross-compiles, so I don't think they can be easily tested.

@digama0 It's fairly easy to run them under qemu:
https://github.com/leanprover/lean4/pull/1913

Unfortunately emulation is too slow to run all tests, but we should probably still run a subset.  Feel free to take over that PR if you want.

PR #2319, opened 2023-07-12
Do not use GMP on ARM Linux

Kha:
This should fix the linking issue reported at https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/libgmp.20linking.20error

Interestingly, `zlib` is the one other library referencing the symbol, but there as part of glibc:
```
$ nm Downloads/lean-4.0.0-linux_aarch64/lib/libz.so.1.2.11 | grep chk_guard
                 U __stack_chk_guard@@GLIBC_2.17
$ nm Downloads/lean-4.0.0-linux_aarch64/lib/libgmp.a | grep chk_guard
nm: mp_clz_tab.o: no symbols
                 U __stack_chk_guard
                 U __stack_chk_guard
...
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,7 +105,7 @@ jobs:
             binary-check: ldd
           - name: Linux aarch64
             os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


Kha:
This should be good to go, we do the same for macOS ARM

PR #2333, opened 2023-07-19
fix: pass usedSimps argument

kim-em:


digama0 (approved):

semorrison:
It's slightly awkward to write a test for this: `simp?` is in `Std`, which is how this should really be tested. I can add a test that uses the plumbing functions if desired.

digama0:
There is an alternate syntax for `simp?` which is to set `set_option tactic.simp.trace true`. I think there is a test file for this?

semorrison:
> tactic.simp.trace

I've added a test to `tests/lean/simp_trace.lean`.

leodemoura:
CI is not green.

PR #2334, opened 2023-07-19
feat: add failIfUnchanged flag to simp

kim-em:
This adds a flag `failIfUnchanged` to the configuration options of `dsimp` / `simp` / `simp_all`.

The followup PR https://github.com/leanprover/lean4/pull/2336 turns the flag on by default. See there for links to builds for Std and Mathlib with the flag enabled.

### DIFF
--- a/src/Init/Meta.lean
+++ b/src/Init/Meta.lean
@@ -1227,6 +1227,9 @@ structure Config where
   proj              : Bool := true
   decide            : Bool := false
   autoUnfold        : Bool := false
+  /-- If `failIfUnchanged := true`, then calls to `simp`, `dsimp`, or `simp_all`
+  will fail if they do not make progress. -/
+  failIfUnchanged   : Bool := false
   deriving Inhabited, BEq, Repr
 
 end DSimp
@@ -1255,6 +1258,9 @@ structure Config where
     `simp` to visit them. If `dsimp := false`, then argument is not visited.
   -/
   dsimp             : Bool := true
+  /-- If `failIfUnchanged := true`, then calls to `simp`, `dsimp`, or `simp_all`
+  will fail if they do not make progress. -/
+  failIfUnchanged   : Bool := false
   deriving Inhabited, BEq, Repr
 
 -- Configuration object for `simp_all`
--- a/src/Lean/Meta/Tactic/Simp/SimpAll.lean
+++ b/src/Lean/Meta/Tactic/Simp/SimpAll.lean
@@ -17,6 +17,7 @@ structure Entry where
   fvarId   : FVarId -- original fvarId
   userName : Name
   id       : Origin -- id of the theorem at `SimpTheorems`
+  origType : Expr
   type     : Expr
   proof    : Expr
   deriving Inhabited
@@ -42,7 +43,8 @@ private def initEntries : M Unit := do
       modify fun s => { s with ctx.simpTheorems := simpThms }
       if hsNonDeps.contains h then
         -- We only simplify nondependent hypotheses
-        let entry : Entry := { fvarId := h, userName := localDecl.userName, id := .fvar h, type := (← instantiateMVars localDecl.type), proof := proof }
+        let type ← instantiateMVars localDecl.type
+        let entry : Entry := { fvarId := h, userName := localDecl.userName, id := .fvar h, origType := type, type, proof }
         modify fun s => { s with entries := s.entries.push entry }
 
 private abbrev getSimpTheorems : M SimpTheoremsArray :=
@@ -118,17 +120,32 @@ def main : M (Option MVarId) := do
     return none -- close the goal
   else
     let mvarId := (← get).mvarId
-    let entries := (← get).entries
-    let (_, mvarId) ← mvarId.assertHypotheses <| entries.filterMap fun e =>
-      -- Do not assert `True` hypotheses
-      if e.type.consumeMData.isConstOf ``True then none else some { userName := e.userName, type := e.type, value := e.proof }
-    mvarId.tryClearMany (entries.map fun e => e.fvarId)
+    -- Prior to #2334, the logic here was to re-assert all hypotheses and call `tryClearMany` on them all.
+    -- This had the effect that the order of hypotheses was sometimes modified, whether or not any where simplified.
+    -- Now we only re-assert the first modified hypothesis,
+    -- along with all subsequent hypotheses, so as to preserve the order of hypotheses.
+    let mut toAssert := #[]
+    let mut toClear := #[]
+    let mut modified := false
+    for e in (← get).entries do
+      if e.type.consumeMData.isConstOf ``True then
+        -- Do not assert `True` hypotheses
+        toClear := toClear.push e.fvarId
+      else if modified || e.type != e.origType then
+        toClear := toClear.push e.fvarId
+        toAssert := toAssert.push { userName := e.userName, type := e.type, value := e.proof }
+        modified := true
+    let (_, mvarId) ← mvarId.assertHypotheses toAssert
+    mvarId.tryClearMany toClear
 
 end SimpAll
 
 def simpAll (mvarId : MVarId) (ctx : Simp.Context) (usedSimps : UsedSimps := {}) : MetaM (Option MVarId × UsedSimps) := do
   mvarId.withContext do
     let (r, s) ← SimpAll.main.run { mvarId, ctx, usedSimps }
+    if let .some mvarIdNew := r then
+      if ctx.config.failIfUnchanged && mvarId == mvarIdNew then
+        throwError "simp_all made no progress"
     return (r, s.usedSimps)
 
 end Lean.Meta
--- a/tests/lean/run/constProp.lean
+++ b/tests/lean/run/constProp.lean
@@ -237,7 +237,7 @@ notation:60 "(" σ ", " s ")"  " ⇓ " σ':60 => Bigstep σ s σ'
 theorem Bigstem.det (h₁ : (σ, s) ⇓ σ₁) (h₂ : (σ, s) ⇓ σ₂) : σ₁ = σ₂ := by
   induction h₁ generalizing σ₂ <;> cases h₂ <;> simp_all
   -- The rest of this proof should be automatic with congruence closure and a bit of forward reasoning
-  case seq ih₁ ih₂ h₁ h₂ =>
+  case seq ih₁ ih₂ _ h₁ h₂ =>
     simp [ih₁ h₁] at ih₂
     simp [ih₂ h₂]
   case ifTrue ih h =>
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,6 +1,20 @@
 Unreleased
 ---------
 
+* [`simp_all` now preserves order of hypotheses](https://github.com/leanprover/lean4/pull/2334).
+
+  In order to support the `failIfUnchanged` configuration option for `dsimp` / `simp` / `simp_all`
+  the way `simp_all` replaces hypotheses has changed.
+  In particular it is now more likely to preserve the order of hypotheses.
+  See [`simp_all` reorders hypotheses unnecessarily](https://github.com/leanprover/lean4/pull/2334).
+  (Previously all non-dependent propositional hypotheses were reverted and reintroduced.
+  Now only such hypotheses which were changed, or which come after a changed hypothesis,
+  are reverted and reintroduced.
+  This has the effect of preserving the ordering amongst the non-dependent propositional hypotheses,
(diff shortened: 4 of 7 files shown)


src/Init/Meta.lean
```diff
@@ -1255,6 +1255,8 @@ structure Config where
     `simp` to visit them. If `dsimp := false`, then argument is not visited.
   -/
   dsimp             : Bool := true
+  /-- This check is implemented in `Lean.Meta.Tactic.Simp.Main`, in `simpGoal`. -/
```
digama0, line 1258: the comment here should describe what the flag does, not where it is implemented

semorrison: Changed. My concern was that with so many entry points into the `simp` machinery, it might be part of the specification to say which entry points were affected...

digama0 (approved):

src/Init/WFTactics.lean
```diff
@@ -22,7 +22,7 @@ macro_rules | `(tactic| decreasing_trivial) => `(tactic| linarith)
 -/
 syntax "decreasing_trivial" : tactic
 
-macro_rules | `(tactic| decreasing_trivial) => `(tactic| simp (config := { arith := true }); done)
+macro_rules | `(tactic| decreasing_trivial) => `(tactic| (try simp (config := { arith := true })); done)
```
eric-wieser, line 25: Should this instead be
```suggestion
macro_rules | `(tactic| decreasing_trivial) => `(tactic| (simp (config := { arith := true, failIfUnchanged := false })); done)
```
Presumably you don't want to catch any and all failures

src/Init/WFTactics.lean
```diff
@@ -33,7 +33,7 @@ lexicographic order lemmas and using `ts` to solve the base case. If it fails,
 it prints a message to help the user diagnose an ill-founded recursive definition. -/
 macro "decreasing_with " ts:tacticSeq : tactic =>
  `(tactic|
-   (simp_wf
+   (try simp_wf
```
eric-wieser, line 36: Same remark as [here](https://github.com/leanprover/lean4/pull/2334/files#r1269219137)

tests/lean/run/constProp.lean
```diff
@@ -237,14 +237,14 @@ notation:60 "(" σ ", " s ")"  " ⇓ " σ':60 => Bigstep σ s σ'
 theorem Bigstem.det (h₁ : (σ, s) ⇓ σ₁) (h₂ : (σ, s) ⇓ σ₂) : σ₁ = σ₂ := by
   induction h₁ generalizing σ₂ <;> cases h₂ <;> simp_all
   -- The rest of this proof should be automatic with congruence closure and a bit of forward reasoning
-  case seq ih₁ ih₂ h₁ h₂ =>
+  case seq ih₁ ih₂ _ h₁ h₂ =>
```
leodemoura, line 240: What is going on here? Did you investigate why you had to add these extra `_`s?

leodemoura:
Waiting on author addressing questions above.

tests/lean/run/constProp.lean
semorrison: This is a consequence of the change to `simp_all` in https://github.com/leanprover/lean4/pull/2334/files#diff-003c74b01f7b82f8997b033dcd3d35690025c53797c5f806c6db9182952281e8.

The problem we had to address here was that `simp_all` was changing the goal unnecessarily by asserting everything and then try to clear. The solution Mario came up with only asserts the hypotheses we actually want to assert, and as a consequence of this change the order of hypotheses after `simp_all` can sometimes change a bit.

I agree this isn't ideal, but two things to note:
1) This problem didn't reappear when fixing up Mathlib to adapt to this change.
2) Arguably if you call tactics (here `simp_all`) in between `cases ...` and `case ... =>`, the onus is probably on the proof author to double check the order of hypotheses, as it's playing with fire to do this at all.

src/Lean/Meta/Tactic/Simp/SimpAll.lean
```diff
@@ -118,17 +120,26 @@ def main : M (Option MVarId) := do
...
-    mvarId.tryClearMany (entries.map fun e => e.fvarId)
+    let mut toAssert := #[]
+    let mut toClear := #[]
+    for e in (← get).entries do
+      if e.type.consumeMData.isConstOf ``True then
+        -- Do not assert `True` hypotheses
+        toClear := toClear.push e.fvarId
+      else if e.type != e.origType then
+        toClear := toClear.push e.fvarId
+        toAssert := toAssert.push { userName := e.userName, type := e.type, value := e.proof }
```
digama0, line 131: > I'm sorry about this, I didn't think the hypothesis reordering would be important. One thing we could do to preserve hypothesis ordering but also preserve the property that if no hypotheses need to change then nothing is done, is to have a flag

> This change must be documented in the change log: https://github.com/leanprover/lean4/blob/master/RELEASES.md We must also document why it has been made, and add links to zulip threads where the discussion happened, and how the consensus has been reached. Reason: this change will affect (non mathlib) users.

Sorry, this is partially my fault, I didn't think the hypothesis reordering would be important. One thing we could do to preserve hypothesis ordering but also preserve the property that if no hypotheses need to change then nothing is done, is to have a flag which is set when a prior hypothesis has been rewritten and then just rewrite all subsequent hypotheses, like so:

```suggestion
    let mut modified := false
    for e in (← get).entries do
      if e.type.consumeMData.isConstOf ``True then
        -- Do not assert `True` hypotheses
        toClear := toClear.push e.fvarId
      else if modified || e.type != e.origType then
        toClear := toClear.push e.fvarId
        toAssert := toAssert.push { userName := e.userName, type := e.type, value := e.proof }
        modified := true
```
Here's a patch that will

semorrison: Okay, I've pushed this and cleaned up.

src/Init/WFTactics.lean
```diff
@@ -22,7 +22,7 @@ macro_rules | `(tactic| decreasing_trivial) => `(tactic| linarith)
 -/
 syntax "decreasing_trivial" : tactic
 
-macro_rules | `(tactic| decreasing_trivial) => `(tactic| simp (config := { arith := true }); done)
+macro_rules | `(tactic| decreasing_trivial) => `(tactic| (simp (config := { arith := true })); done)
```
leodemoura, line 25: Why do we need the extra parentheses? Is this some kind of a workaround?

tests/lean/run/constProp.lean
leodemoura: @semorrison Notice that you are already tracking whether things changed or not. You don't need to rely on whether the `mvarId` changed or nor.

src/Init/WFTactics.lean
digama0: It should be unnecessary. I believe this line was modified as part of a stage0 update (adding the flag to the config caused compiled `simp` not to match the new VM layout), and it was incompletely reverted.

tests/lean/run/constProp.lean
digama0: @leodemoura Are you referring to the `modified` flag here? That flag is not sufficient to determine whether the mvarId changed, because it only tracks changes to the hypotheses, not the goal. The main purpose of it is to ensure that if the hypotheses are not simplified then there is no needless mvarId changing, so that the mvarId check is accurate.

src/Init/WFTactics.lean
semorrison:
```suggestion
macro_rules | `(tactic| decreasing_trivial) => `(tactic| simp (config := { arith := true }); done)
```

tests/lean/run/constProp.lean
semorrison: (Just noting that the extra `_`s are much reduced relative to @leodemoura's first query of this issue. Now there's just the one, instead of one in each of the cases here. There's still one, of course! Also noting that this issue does not seem to have arisen anywhere besides this file, including when we look at the [mathlib4 PR](https://github.com/leanprover-community/mathlib4/pull/6019) adapting to this change.)

leodemoura: @semorrison It is great the issue is not affecting mathlib4, but mathlib is not the only user. Please modify the logic and ensure the order does not depend on whether a hypothesis was simplified or not. It feels like a very small change.

digama0: @semorrison  this is strange, there should not be any `_` needed at all with the modification I suggested. "ensure the order does not depend on whether a hypothesis was simplified or not" should already be the case, and if not then there is a bug somewhere.

semorrison: @leodemoura, @digama0, I think there's some confusion here arising from the fact that it is the *old* `simp_all` which is reordering the hypotheses, while the *new* `simp_all` preserves the order exactly.

Thus the question becomes:
* do we prefer conserving the old behaviour (which was a consequence of re-asserting everything and then calling `tryClearMany` on everything) for backwards compatibility sake, or
* do we prefer the new behaviour, which strictly preserves the order of hypotheses when possible, including in this case.

Here are the relevant fragments of the goal:
* The `seq.seq` case immediately before `simp_all`:
```
case seq.seq
σ: State
s: Stmt
σ₁σ₁✝: State
s₁✝: Stmt
σ₂✝¹: State
s₂✝: Stmt
σ₃✝: State
a✝³: (σ₁✝, s₁✝) ⇓ σ₂✝¹
a✝²: (σ₂✝¹, s₂✝) ⇓ σ₃✝
a_ih✝¹: ∀ {σ₂ : State}, (σ₁✝, s₁✝) ⇓ σ₂ → σ₂✝¹ = σ₂
a_ih✝: ∀ {σ₂ : State}, (σ₂✝¹, s₂✝) ⇓ σ₂ → σ₃✝ = σ₂
σ₂σ₂✝: State
a✝¹: (σ₁✝, s₁✝) ⇓ σ₂✝
a✝: (σ₂✝, s₂✝) ⇓ σ₂
⊢ σ₃✝ = σ₂
```

Then after the new `simp_all` (note no changes):
```
case seq.seq
σ: State
s: Stmt
σ₁σ₁✝: State
s₁✝: Stmt
σ₂✝¹: State
s₂✝: Stmt
σ₃✝: State
a✝³: (σ₁✝, s₁✝) ⇓ σ₂✝¹
a✝²: (σ₂✝¹, s₂✝) ⇓ σ₃✝
a_ih✝¹: ∀ {σ₂ : State}, (σ₁✝, s₁✝) ⇓ σ₂ → σ₂✝¹ = σ₂
a_ih✝: ∀ {σ₂ : State}, (σ₂✝¹, s₂✝) ⇓ σ₂ → σ₃✝ = σ₂
σ₂σ₂✝: State
a✝¹: (σ₁✝, s₁✝) ⇓ σ₂✝
a✝: (σ₂✝, s₂✝) ⇓ σ₂
⊢ σ₃✝ = σ₂
```

Alternatively, the behaviour of `simp_all` on master (note that `σ₂σ₂✝: State` have migrated upwards, and hence out of the range where we need to specify them as arguments to `case seq`).
```
case seq.seq
σ: State
s: Stmt
σ₁σ₁✝: State
s₁✝: Stmt
σ₂✝¹: State
s₂✝: Stmt
σ₃✝σ₂σ₂✝: State
a✝³: (σ₁✝, s₁✝) ⇓ σ₂✝¹
a✝²: (σ₂✝¹, s₂✝) ⇓ σ₃✝
a_ih✝¹: ∀ {σ₂ : State}, (σ₁✝, s₁✝) ⇓ σ₂ → σ₂✝¹ = σ₂
a_ih✝: ∀ {σ₂ : State}, (σ₂✝¹, s₂✝) ⇓ σ₂ → σ₃✝ = σ₂
a✝¹: (σ₁✝, s₁✝) ⇓ σ₂✝
a✝: (σ₂✝, s₂✝) ⇓ σ₂
⊢ σ₃✝ = σ₂
```

semorrison: In particular, the request:

> Please modify the logic and ensure the order does not depend on whether a hypothesis was simplified or not.

is I think actually satisfied better by the new logic (which preserves order whether or not a hypothesis was simplified) than the old logic (which would sometimes change the order, whether or not a hypothesis was simplified).

semorrison: I will add some comments in the code explaining all this.

semorrison: I have added a simple test checking that `simp_all` now preserves order of hypothesis (which used to fail), and updated `RELEASE.md` to properly describe the new behaviour.

leodemoura:
I am going to merge it as-is. From now on, I expect way more tests when making this kind of change. It is hard to believe 3 new `simp_all` tests covers all interesting cases.

PR #2335, opened 2023-07-19
Sort messages by position when printing

Kha:
As exhibited in #2276 but already encountered many times before I believe, so it is high time to make the message ordering more deterministic. Unfortunately there is some perhaps unwanted reordering for messages at the same position as shown in the diff below. If we agree to not reorder messages at the same position, is it time to implement a stable sort on arrays?


leodemoura:
> is it time to implement a stable sort on arrays?

+1

leodemoura:
CI is not green

semorrison:
Does this just need all the tests cleaned up to match the new order? Or is this waiting on something else? I can't quite decipher the intention from the above.

Kha:
It's waiting on a stable sort implementation. I haven't encountered the original issue since so I haven't made this PR a priority.

PR #2336, opened 2023-07-19
feat: turning on failIfUnchanged by default in simp

kim-em:
- [x] depends on #2334

See 
* https://github.com/leanprover/std4/pull/190 for the corresponding bump to Std4
* https://github.com/leanprover-community/mathlib4/pull/6019 for the corresponding bump to Mathlib

### DIFF
--- a/src/Init/WFTactics.lean
+++ b/src/Init/WFTactics.lean
@@ -10,8 +10,7 @@ import Init.WF
 /-- Unfold definitions commonly used in well founded relation definitions.
 This is primarily intended for internal use in `decreasing_tactic`. -/
 macro "simp_wf" : tactic =>
-  `(tactic| simp [invImage, InvImage, Prod.lex, sizeOfWFRel,
-          measure, Nat.lt_wfRel, WellFoundedRelation.rel])
+  `(tactic| try simp [invImage, InvImage, Prod.lex, sizeOfWFRel, measure, Nat.lt_wfRel, WellFoundedRelation.rel])
 
 /-- Extensible helper tactic for `decreasing_tactic`. This handles the "base case"
 reasoning after applying lexicographic order lemmas.
@@ -22,7 +21,7 @@ macro_rules | `(tactic| decreasing_trivial) => `(tactic| linarith)
 -/
 syntax "decreasing_trivial" : tactic
 
-macro_rules | `(tactic| decreasing_trivial) => `(tactic| simp (config := { arith := true }); done)
+macro_rules | `(tactic| decreasing_trivial) => `(tactic| (simp (config := { arith := true, failIfUnchanged := false })); done)
 macro_rules | `(tactic| decreasing_trivial) => `(tactic| assumption)
 macro_rules | `(tactic| decreasing_trivial) => `(tactic| apply Nat.sub_succ_lt_self; assumption) -- a - (i+1) < a - i if i < a
 macro_rules | `(tactic| decreasing_trivial) => `(tactic| apply Nat.pred_lt'; assumption) -- i-1 < i if j < i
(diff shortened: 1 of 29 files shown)


tests/lean/run/discrTreeSimp.lean
```diff
@@ -8,7 +8,7 @@ theorem map_map (f : α → β) (g : β → γ) (xs : List α) : (xs.map f |>.ma
   sorry
 
 theorem ex1 (f : Nat → Nat) (xs : List Nat) : (xs.map f |>.map f) = xs.map (f ∘ f) := by
-  simp
+  fail_if_success simp
```
eric-wieser, line 11: I think these could do with a comment explaining what is being tested

src/Init/WFTactics.lean
```diff
@@ -10,7 +10,7 @@ import Init.WF
 /-- Unfold definitions commonly used in well founded relation definitions.
 This is primarily intended for internal use in `decreasing_tactic`. -/
 macro "simp_wf" : tactic =>
-  `(tactic| simp [invImage, InvImage, Prod.lex, sizeOfWFRel,
+  `(tactic| try simp [invImage, InvImage, Prod.lex, sizeOfWFRel,
```
eric-wieser, line 13: I'd argue this one should be using `{ failIfUnchanged := false }`

tests/lean/run/discrTreeSimp.lean
semorrison: Documenting these tests is orthogonal to this PR, no?

src/Init/WFTactics.lean
semorrison: First, why?

Second, it's not possible. This is used very early on at a stage where the `config` argument is apparently discarded. With this change the build would fail in stage1 at Init/Data/Array/Basic.lean:663:6.

digama0: Oh fun, it's a update-stage0 PR. It is possible, it's just annoying. The config argument isn't being discarded, it's being reinterpreted at the wrong type and probably the next boolean over has the actual value (this sometimes causes segfaults). I think you have to set `interpreter.prefer_native` to `true` in `stage0/src/stdlib_flags.h` to get it to build, and then run an update-stage0 and build lean again.

eric-wieser: > First, why?

It means that if `simp` fails in some other unpredictable way we don't just discard the error, Using `try` feels like the anti-pattern of Python's
```
try:
   v = d[k]
except:
   print("key was missing")
```
which actually catches all sorts of error that it would be best to propagate

> Second, it's not possible.

Ah, I had no idea that this touched stage0. Is `simp_wf` re-defined in a later stage anyway using the new syntax?

digama0: Actually I think this PR doesn't need a update-stage0, you just need one between the previous PR and this one or else you can't use `failIfUnchanged` and expect `simp` to do something about it since it's still stage0 simp from before the previous PR.

leodemoura:
This change must be documented in the change log: https://github.com/leanprover/lean4/blob/master/RELEASES.md
We must also document why it has been made, and add links to zulip threads where the discussion happened, and how the consensus has been reached.
Reason: this change will affect (non mathlib) users.

src/Lean/Meta/Tactic/Simp/SimpAll.lean
```diff
@@ -118,17 +120,26 @@ def main : M (Option MVarId) := do
...
+    let mut toAssert := #[]
+    let mut toClear := #[]
+    for e in (← get).entries do
+      if e.type.consumeMData.isConstOf ``True then
+        -- Do not assert `True` hypotheses
+        toClear := toClear.push e.fvarId
+      else if e.type != e.origType then
+        toClear := toClear.push e.fvarId
+        toAssert := toAssert.push { userName := e.userName, type := e.type, value := e.proof }
+    let (_, mvarId) ← mvarId.assertHypotheses toAssert
```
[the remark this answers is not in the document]
semorrison: I've documented this in the change log now, back in the initial PR #2334.

src/Init/WFTactics.lean
semorrison: Okay, I will replace the `try` here after #2334 is merged.

PR #2337, opened 2023-07-19
Allow linking against Lake

Kha:
And other cleanups

### DIFF
--- a/src/lake/test/116/test.sh
+++ b/src/lake/test/116/test.sh
@@ -14,14 +14,7 @@ INITD_NOT=$'Content-Length: 40\r\n\r\n{"jsonrpc":"2.0","method":"initialized"}'
 OPEN_REQ=$'Content-Length: 145\r\n\r\n{"jsonrpc":"2.0","method":"textDocument/didOpen","params":{"textDocument":{"uri":"file://Test.lean","languageId":"lean4","version":0,"text":""}}}'
 MSGS="$INIT_REQ$INITD_NOT$OPEN_REQ"
 
-set -o pipefail
-
-cp /dev/null lake.in
-tail -f lake.in | $LAKE serve 2>&1 > lake.out &
-pid=$!
-echo -n "$MSGS" >> lake.in
-sleep 3
-kill $pid
+(echo -n "$MSGS"; sleep 3) | $LAKE serve 2>&1 > lake.out || true
 
 echo
 grep -m1 Lake lake.out
--- a/src/shell/CMakeLists.txt
+++ b/src/shell/CMakeLists.txt
@@ -177,11 +177,10 @@ FOREACH(T ${LEANTESTS})
   endif()
 ENDFOREACH(T)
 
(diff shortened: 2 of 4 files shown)


src/shell/CMakeLists.txt
```diff
@@ -177,11 +177,10 @@ FOREACH(T ${LEANTESTS})
   endif()
 ENDFOREACH(T)
 
-# Create a lake test for each subdirectory of `lake/examples` which contains a `test.sh` file.
-file(GLOB_RECURSE LEANLAKETESTS "${LEAN_SOURCE_DIR}/lake/examples/test.sh")
+# Create a lake test for each subdirectory of `lake` which contains a `test.sh` file.
+file(GLOB_RECURSE LEANLAKETESTS "${LEAN_SOURCE_DIR}/lake/test.sh")
```
Kha, line 181: @tydeu I noticed we didn't even run most of the Lake test suite, which would considerably weaken my claim that Lean CI might be sufficient for Lake :)

src/lake/test/116/test.sh
```diff
@@ -14,14 +14,7 @@ INITD_NOT=$'Content-Length: 40\r\n\r\n{"jsonrpc":"2.0","method":"initialized"}'
...
 
-set -o pipefail
-
-cp /dev/null lake.in
-tail -f lake.in | $LAKE serve 2>&1 > lake.out &
-pid=$!
-echo -n "$MSGS" >> lake.in
-sleep 3
-kill $pid
+(echo -n "$MSGS"; sleep 3) | $LAKE serve 2>&1 > lake.out || true
```
Kha, line 17: Curiously this broke under Nix, perhaps because Nix' isolation interferes with sending the signal to the process group. A more proper solution would be to send the correct shutdown notification to the process

src/shell/CMakeLists.txt
```diff
@@ -177,11 +177,10 @@ FOREACH(T ${LEANTESTS})
...
 FOREACH(T ${LEANLAKETESTS})
-  if(NOT T MATCHES ".*lake-packages.*")
-  if(NOT T MATCHES ".*bootstrap.*")
+  if(NOT T MATCHES ".*(lake-packages|bootstrap|104|git).*")
```
Kha, line 183: The latter two test cases still reference the old Lake repo root, should I update them?

src/lake/test/116/test.sh
tydeu: I am pretty sure sending that, for some reason, the shutdown notification was insufficient when I was writing this test. I should have made a note of why, though, because I have since forgotten.

src/shell/CMakeLists.txt
Kha: I'll leave that to you then :)

src/lake/test/116/test.sh
Kha: Happy with my version for now :smile:

Kha:
So, ready to merge this part?

tydeu:
@Kha LGTM.

PR #2338, opened 2023-07-19
chore: better error message for loose bvar in whnf

JLimperg:
~~`whnf` used to panic when given an `Expr` with loose bvars. With this PR, it throws an error instead.~~

This PR clarifies the error message when `whnf` panics upon encountering a loose `bvar`.

### DIFF
--- a/src/Lean/Meta/WHNF.lean
+++ b/src/Lean/Meta/WHNF.lean
@@ -314,7 +314,7 @@ end
   | .lam ..        => return e
   | .sort ..       => return e
   | .lit ..        => return e
-  | .bvar ..       => unreachable!
+  | .bvar ..       => panic! "loose bvar in expression"
   | .letE ..       => k e
   | .const ..      => k e
   | .app ..        => k e


JLimperg:
Sebastian explained to me that this is not a good idea, so I'll refactor the PR to leave the panic in place, but with a better error message.

src/Lean/Meta/WHNF.lean
```diff
@@ -314,7 +314,7 @@ end
   | .lam ..        => return e
   | .sort ..       => return e
   | .lit ..        => return e
-  | .bvar ..       => unreachable!
+  | .bvar ..       => panic! "whnf: loose bvar in expression"
```
kmill, line 317: Note that including "whnf: " here might be seen as being redundant. This message will display as
```
PANIC at Lean.Meta.whnfEasyCases Lean.Meta.WHNF:317:22: whnf: loose bvar in expression
```
I'm not sure that's a bad thing, but I just wanted to point it out.

leodemoura:
Could you please apply Kyle's suggestion? Then, we are ready to merge

JLimperg:
Done, thanks!

Issue #2085, opened 2023-02-03
LLVM backend generates incorrect `main` function for wasm32


locriacyber:
### Steps to Reproduce

`hello.lean`
```
def main : IO Unit := do
  IO.println s!"Hello!"
```
1. `lean --bc=Hello.bc --target=wasm32 hello.lean`
2. `wasm2wat Hello.bc.o | grep main`
```
  (func $_lean_main (type 0) (param i32) (result i32)
  (func $_init_l_main___closed__1 (type 2) (result i32)
        call $_init_l_main___closed__1
  (func $main (type 6) (param i64 i32) (result i64)
      call $_lean_main
  (data $l_main___closed__1 (i32.const 0) "\00\00\00\00")
```

**Expected behavior:** main's signature should be `(i32, i32) -> i32`

**Actual behavior:** is `(i64, i32) -> i64`

### Versions
Lean (version 4.0.0, commit 4b974fd60b9f, Release)

PR #2090, opened 2023-02-04
fix: LLVM backend: detect size_t from target triple

iacore:
I added the binding API to get pointer size from target. I hope I got the lifetime right.

A lot of plumbing need to be done in Lean

@bollu

PR #2340, opened 2023-07-20
Allow building stage 1+ with the LLVM backend

bollu:
This PR changes the build system and outlines uses of `extern inline` to enable LLVM stage2 builds. Draft PR till this branch actually succeeds.


src/CMakeLists.txt
```diff
@@ -526,6 +526,12 @@ add_custom_target(make_stdlib ALL
   COMMAND $(MAKE) -f ${CMAKE_BINARY_DIR}/stdlib.make Init Lean
   VERBATIM)
 
+# if we have LLVM enabled, then build `lean.h.bc` which has the LLVM bitcode
```
bollu, line 529: @Kha , could you explain to me why moving this code to `src/CMakeLists.txt` fixes the build on `Linux Release`? To summarize the change, before this commit, [the build breaks with](https://github.com/leanprover/lean4/actions/runs/5608926936/jobs/10261983565#step:9:1343)

```
lean_llvm_create_memory_buffer_with_contents_of_file 
file:/home/runner/work/lean4/lean4/build/stage0/../stage1/lib/lean/lean.h.bc
ERROR:      No such file or directory
```

and after this commit, the linux release build succeeds! I don't understand the build-time control flow, and why moving the file here helps.

Kha: The important part is not the code move but replacing `lean` with `make_stdlib`. Which is the target that otherwise generates the error you quoted.

src/Lean/Expr.lean
```diff
@@ -459,7 +459,7 @@ inductive Expr where
   -/
   | proj (typeName : Name) (idx : Nat) (struct : Expr)
 with
-  @[computed_field, extern c inline "lean_ctor_get_uint64(#1, lean_ctor_num_objs(#1)*sizeof(void*))"]
+  @[computed_field, extern "lean_expr_data"]
```
leodemoura, line 462: @Kha I am assuming you are reviewing this PR. If I am correct, could you please ensure that every "c inline" that was converted into a function has been properly marked as an "static inline"? This is the case for this one, but I didn't check all of them.

Kha: Yes, they are neatly lined up here: https://github.com/leanprover/lean4/pull/2340/files#diff-210fa1e51e911735f3210d99d02d03b44aa715dcfb93f428885b5b830882b29c
When the PR is functioning, we should do a performance run for each backend to check for regressions

tests/compiler/foreign/Makefile
```diff
@@ -1,7 +1,7 @@
 CXX ?= c++
 CPPFLAGS = -O3
 # TODO: use Lake instead
-include lean.mk
+# include lean.mk
```
Kha, line 4: @bollu I've already fixed the test

Kha:
The, hopefully, last remaining error is quite puzzling: this works
```
"D:/a/lean4/lean4/build/stage1/leanc.sh" D:/a/lean4/lean4/build/stage1/shell/CMakeFiles/shell.dir/lean.cpp.obj D:/a/lean4/lean4/build/stage1/shell/CMakeFiles/shell.dir/manifest.rc.obj -lleanshared    -O3 -DNDEBUG -o D:/a/lean4/lean4/build/stage1/bin/lean.exe
```
but this fails:
```
D:/a/lean4/lean4/build/stage1/leanc.sh -o "D:/a/lean4/lean4/build/stage1/bin/leanc.exe" D:/a/lean4/lean4/build/stage1/temp/Leanc.o -O3 -DNDEBUG -lleanshared  -v
 "D:/a/lean4/lean4/build/stage1/bin/ld.lld" --sysroot=D:/a/lean4/lean4/build/stage1 -m i386pep -Bdynamic -o D:/a/lean4/lean4/build/stage1/bin/leanc.exe D:/a/lean4/lean4/build/stage1/lib/crt2.o D:/a/lean4/lean4/build/stage1/lib/crtbegin.o -LD:/a/lean4/lean4/build/stage1/lib -LD:/a/lean4/lean4/build/stage1/lib/lean -LD:/a/lean4/lean4/build/stage1/x86_64-w64-mingw32/lib -LD:/a/lean4/lean4/build/stage1/x86_64-w64-mingw32/mingw/lib -LD:/a/lean4/lean4/build/stage1/lib -LD:/a/lean4/lean4/build/stage1/x86_64-w64-mingw32/sys-root/mingw/lib -LD:/a/lean4/lean4/build/stage1/lib/clang/15.0.1/lib/windows D:/a/lean4/lean4/build/stage1/temp/Leanc.o -lleanshared -Bstatic -lgmp -lunwind -Bdynamic -lgmp -lucrtbase --stack 104857600 -lm -lbcrypt -lmingw32 D:/a/lean4/lean4/build/stage1/lib/clang/15.0.1/lib/windows/libclang_rt.builtins-x86_64.a -lmoldname -lmingwex -lpthread -ladvapi32 -lshell32 -luser32 -lkernel32 -lmingw32 D:/a/lean4/lean4/build/stage1/lib/clang/15.0.1/lib/windows/libclang_rt.builtins-x86_64.a -lmoldname -lmingwex -lkernel32 D:/a/lean4/lean4/build/stage1/lib/crtend.o
ld.lld: error: undefined symbol: l_Lean_Compiler_FFI_getCFlags
>>> referenced by D:/a/lean4/lean4/build/stage1/temp/Leanc.o:(l_main___lambda__4)
```
...but the command lines at least should be identical on `master`

Kha:
!bench

Kha:
Due to new commits on `master`, we have a new test failure because of new warnings:
```
[06:57 (00.000005)] lean-test-stage1> > /nix/store/h1sl8hi6wp2i435d06cwpz9y3dhmdiil-lean-bin-tools/include/lean/lean.h:1942:79: warning: omitting the parameter name in a function definition is a C2x extension [-Wc2x-extensions]
[06:57 (00.000006)] lean-test-stage1> > static inline lean_object* lean_expr_checker_get_max_ctor_fields(lean_object *) {
[06:57 (00.000005)] lean-test-stage1> >                                                                               ^
...
```
We could work on filtering out these warnings but we should probably instead make sure the header is valid C99(?) /cc @bollu

Kha:
Let's please also use the correct `lean_object *` `typedef`s in these functions

Kha:
Interesting, this is the diff with LLVM disabled. But since especially `rbmap` should not use any of the outlined functions, it is hard to see any causal relationship with this PR. I will turn LLVM back on for now and do a run with it.

Kha:
!bench

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -294,6 +294,29 @@ opaque disposeTargetMachine (tm : TargetMachine ctx) : BaseIO Unit
 opaque disposeModule (m : Module ctx) : BaseIO Unit
 
 
+-- https://github.com/llvm/llvm-project/blob/c3e073bcbdc523b0f758d44a89a6333e38bff863/llvm/include/llvm-c/Core.h#L198
+structure Visibility where
+  private mk :: val : UInt64
+
+def Visibility.default   : Visibility := { val := 0 }
+def Visibility.hidden    : Visibility := { val := 1 }
+def Visibility.protected : Visibility := { val := 2}
```
tobiasgrosser, line 303: `2}` -> `2 }`

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -415,25 +415,37 @@ def emitFnDeclAux (mod : LLVM.Module llvmctx)
...
-        argtys := #[← LLVM.pointerType (← LLVM.voidPtrType llvmctx)]
-      let fnty ← LLVM.functionType retty argtys (isVarArg := false)
-      LLVM.getOrAddFunction mod cppBaseName fnty
-
+    if isClosedTermName env decl.name then LLVM.setVisibility global LLVM.Visibility.hidden -- emit "static "
+    else if isExternal then pure () -- emit "extern "
+    else LLVM.setDLLStorageClass global LLVM.DLLStorageClass.export 
+  else if !isExternal 
+    -- @Kha, @grosser: can you explain what's going on here? why do we set the
+    -- DLL class to expor -- if it is *not* external?
```
Kha, line 444: Because an extern function is not necessarily linked into the same DLL as the current module, so it would be wrong to mark it `dllexport`

Kha (commented):
Ah, I suppose Windows will not be fixed without an additional update-stage0 since you touched EmitLLVM. And now that I think of it, we can't actually run update-stage0 with this PR because there are no C files emitted. I believe we should simply produce both .bc and .c with LLVM=ON, it should introduce insignificant overhead.

src/Lean/Compiler/IR/EmitLLVM.lean
Kha: By the way, GitHub does not alert on mentions in source code

Kha:
Nice, looks like it's only missing visibility modifiers on initializers (https://github.com/leanprover/lean4/actions/runs/5642004685/job/15281100057?pr=2340). @bollu Could you fix that and then put the Emit changes in a separate PR so we can do an update-stage0 on master in between?

bollu:
@Kha I assume I need an update-stage0 for the windows build? I’m not sure what’s up with the Linux release lake failure.

Kha:
Hmm, the test might be flaky but twice in a row and never before is a bit concerning. Anyways, if Windows passes, please go ahead with the Emit PR

hargoniX:
I cannot reproduce the failures on my local setup :(

bollu:
@Kha I asked @hargoniX to rerun the linux tests on their machine, where it did succeed. Could we rerun the linux release CI? I would be troubled if `leanlaketest_116` fails once more.

bollu:
@Kha I squash-merged this branch against the latest `upstream/master`. I resolved the `stage0` changes by using the versions of the files in `upstream/master` . This build fails on `aarch64`: https://github.com/bollu/lean4/actions/runs/5660012113/job/15334806801. 

I guess we need a second `stage0` update with our `CMake` changes to propagate the `CMake` changes which correctly setup `--target`?

bollu:
(To be clear, I would like to understand how to structure this PR so we can start reviewing it!)

src/CMakeLists.txt
```diff
@@ -254,7 +254,7 @@ find_package(PythonInterp)
 include_directories(${CMAKE_BINARY_DIR}/include)
 
 # Pick up `llvm-config` to setup LLVM flags.
-if(LLVM)
+if(LLVM MATCHES "ON")
```
Kha, line 257: revert

Kha:
> I squash-merged this branch against the latest `upstream/master`

I would prefer independent parts to stay as separate commits; it is fine to squash away incremental work internal to the PR. You can manually copy `src/CMakeLists.txt` to `stage0/` which I believe should fix CI, it is only the standard library that we always copy with the script.

bollu:
I don't have a good enough mental model of the CMake build to successfully disambuguate all the changes we've made :D I can split this up into two unrelated commits: (1) nix changes, (2) outlining, (3) "build system changes" [all the changes to `CMake` plus the backporting into `stage0`]

tobiasgrosser:
Maybe the easiest is to see if merging master (while kipping the stage0 from master), plus copying src/CMakeLists.txt to stage0 already reduces the size of this PR quite a bit?

bollu:
@Kha can we have a bench to make sure we do not regress performance? 

I am hoping that https://github.com/leanprover/lean4/pull/2358 will alleviate the flakiness of the Linux release build.

src/CMakeLists.txt
```diff
@@ -432,7 +435,9 @@ else()
 endif()
 configure_file("${LEAN_SOURCE_DIR}/config.h.in" "${LEAN_BINARY_DIR}/include/lean/config.h")
 install(DIRECTORY ${LEAN_BINARY_DIR}/include/ DESTINATION include)
+
 configure_file(${LEAN_SOURCE_DIR}/lean.mk.in ${LEAN_BINARY_DIR}/share/lean/lean.mk)
+
```
tobiasgrosser, line 440: This whitespace change seems unnecessary

tobiasgrosser (commented):
Some minor comments.

Kha:
!bench

Kha:
Oh jeez, the larger timeout actually fixed it... that points to a significant performance regression compared to the C backend then though

src/Lean/Expr.lean
```diff
@@ -152,7 +152,7 @@ def Expr.Data.hasLevelMVar (c : Expr.Data) : Bool :=
 def Expr.Data.hasLevelParam (c : Expr.Data) : Bool :=
   ((c.shiftRight 43).land 1) == 1
 
-@[extern c inline "(uint64_t)#1"]
+@[extern "lean_expr_binderinfo_to_uint64"]
```
tydeu, line 155:
```suggestion
@[extern "lean_uint8_to_uint64"]
```

Less code duplication.

src/include/lean/lean.h
```diff
@@ -1927,6 +1927,70 @@ static inline size_t lean_data_hashmap_mk_idx(lean_object* sz, uint64_t hash) {
...
+    return LEAN_VERSION_IS_RELEASE;
+}
+
+static inline lean_object *lean_init_meta_version_get_special_desc(lean_obj_arg _unit) {
+    return lean_mk_string(LEAN_SPECIAL_VERSION_DESC);
+}
+
+static inline uint8_t lean_init_meta_internal_is_stage0(lean_obj_arg _unit) {
+    return LEAN_IS_STAGE0;
+}
```
tydeu, line 1988: I do not think the prefixes (`expr_checker`, `init_core`. `init_meta`, `init_meta_internal`) are useful for these. The constants they are wrapping do not have the prefix and such prefixing is not the style used in other C externs.

src/include/lean/lean.h
```diff
@@ -1927,6 +1927,70 @@ static inline size_t lean_data_hashmap_mk_idx(lean_object* sz, uint64_t hash) {
     return (size_t)(hash & (lean_unbox(sz) - 1));
 }
 
+static inline size_t lean_data_hashset_mk_idx(lean_object* sz, uint64_t hash) {
```
tydeu, line 1930:
```suggestion
static inline size_t lean_hashset_mk_idx(lean_object* sz, uint64_t hash) {
```

Similarly, I would just removing the `data` prefix from this name.

src/Init/Meta.lean
```diff
@@ -11,28 +11,28 @@ import Init.Data.Option.BasicAux
...
+@[extern "lean_init_meta_version_get_patch"]
 private opaque version.getPatch (u : Unit) : Nat
 def version.patch : Nat := version.getPatch ()
 
 @[extern "lean_get_githash"]
 opaque getGithash (u : Unit) : String
 def githash : String := getGithash ()
 
-@[extern c inline "LEAN_VERSION_IS_RELEASE"]
+@[extern "lean_init_meta_version_is_release"]
```
tydeu, line 30:
```suggestion
@[extern "lean_version_get_is_release"]
```

Also, this should probably have a `get` in it to parallel the others.

tydeu:
@Kha / @bollu 

> Oh jeez, the larger timeout actually fixed it... that points to a significant performance regression compared to the C backend then though

Hopefully, the timeout only really needed a 1 second bump and, if so, the performance impact is not so extreme.

src/include/lean/lean.h
```diff
@@ -1927,6 +1927,70 @@ static inline size_t lean_data_hashmap_mk_idx(lean_object* sz, uint64_t hash) {
...
+
+static inline lean_object *lean_init_meta_version_get_special_desc(lean_obj_arg _unit) {
+    return lean_mk_string(LEAN_SPECIAL_VERSION_DESC);
+}
+
+static inline uint8_t lean_init_meta_internal_is_stage0(lean_obj_arg _unit) {
+    return LEAN_IS_STAGE0;
+}
+
+static inline lean_object *lean_nat_pred(lean_object *n) {
```
tydeu, line 1990:
```suggestion
static inline lean_object *lean_nat_pred(lean_b_obj_arg n) {
```

This records that the object is borrowed in the C signature (and fits with the usually extern function style).

tydeu:
Also, for the return values of the extern functions. you may wish to use `lean_obj_res` rather than `lean_object*` to follow the usual style.

Kha:
Before we merge, we will have to switch back to the C backend anyway. I would suggest creating a copy of the "Linux release" CI job and set LLVM=ON for it and only it for now. We can disable the test for that specific CI job then and consider it part of future performance investigations of the LLVM backend.

Kha:
But also we should fix that test to not require a timeout :)

bollu:
@Kha I would like to cover the following triples:
- Vanilla LLVM build on linux release.
- One cross compile build (`MacOS aarch64`).
- One windows build for windows linker things (`Windows`). 

Let us haggle on which of these we should have a `LLVM=ON` build copy 😄

src/include/lean/lean.h
bollu: @tydeu note that the fully qualified name `lean_data_hashmap_mk_idx` already exists (https://github.com/leanprover/lean4/blob/master/src/include/lean/lean.h#L1926-L1928). I followed suit and decided to fully qualify the new ones I have outlines. 

I am happy to remove the prefixes.

tydeu: @bollu Hmm. I think that case is an outlier (and probably should be fixed), other externs do not have module prefixes. See, for instance, other externs in [`Init.Core`](https://github.com/leanprover/lean4/blob/53477089fe255280d7a816371ec315c6e2e94220/src/Init/Core.lean#L370-L371) and [`Init.IO`](https://github.com/leanprover/lean4/blob/53477089fe255280d7a816371ec315c6e2e94220/src/Init/System/IO.lean#L169). Afaik, the standard style is for C symbols to only include *namespace* prefixes, not *module* ones.

Kha:
We already have too many CI jobs that we run on each PR. If you want to see the backend be tested for each configuration, let's work on making it the default :) .

tobiasgrosser:
That's a good incentive!

bollu:
>  We already have too many CI jobs that we run on each PR. If you want to see the backend be tested for each configuration, let's work on making it the default :) .

@Kha that sounds good to me. May I know what would be necessary to flip on LLVM by default across all backends? From the milestone list (https://github.com/leanprover/lean4/milestone/3) , I presume that what is missing is:

(1) pulling in the patch that correctly initializes `size_t`. We can test this on a non 64-bit target in a VM after pulling in https://github.com/leanprover/lean4/pull/2090. This should also fix https://github.com/leanprover/lean4/issues/2085. 

(2) making the LLVM backend have performance parity across all tests. I would like to know how to run the benchmark suite locally, and how I could gather a list of tests where we see a performance drop? From the current `!bench`, I can see that `tests/bench/` and `unionfind` need to be investigated:



> Here are the [benchmark results](http://speedcenter.informatik.kit.edu/velcom/run-detail/0f348c9c-f6e1-4e9f-8c36-bbb2099758c2) for commit [c8a169a](https://github.com/leanprover/lean4/commit/c8a169a4df683b6f47bd0e10d26b15f3944b9a16). There were [significant changes](http://speedcenter.informatik.kit.edu/velcom/compare/d09cb4b9-3392-4a26-b70f-28425a482e1a/to/0f348c9c-f6e1-4e9f-8c36-bbb2099758c2) against commit [5347708](https://github.com/leanprover/lean4/commit/53477089fe255280d7a816371ec315c6e2e94220):
> 
> ```diff
>   Benchmark                  Metric             Change
>   ================================================================
> - tests/bench/ interpreted   instructions         2.9%  (5962.0 σ)
> - unionfind                  task-clock           3.9%    (78.2 σ)
> - unionfind                  wall-clock           3.8%    (75.1 σ)
> ```

Is that an exhaustive list of milestones before we can flip on LLVM by default?

Kha:
> May I know what would be necessary to flip on LLVM by default across all backends?

Ideally it should be strictly better than the C backend! I.e. no known regressions and at least some notable benefits. One minor benefit would be that we shouldn't have to distribute libclang anymore, which is about 7% of the unpacked toolchain. Unless, that is, the ecosystem started depending on there being a C compiler, which I haven't checked.

> I would like to know how to run the benchmark suite locally, and how I could gather a list of tests where we see a performance drop?

Yes, running the Speedcenter suite locally with and without LLVM is the way to go for now I believe. Please tell me if there is anything missing from https://github.com/leanprover/lean4/blob/master/tests/bench/README.md. Note that, as I said, `!bench` is not yet set up to use LLVM at all.

Kha:
> Is that an exhaustive list of milestones before we can flip on LLVM by default?

I added everything that directly came to mind to the milestone

.github/workflows/ci.yml
```diff
@@ -50,7 +50,17 @@ jobs:
       matrix:
         include:
           # portable release build: use channel with older glibc (2.27)
-          - name: Linux release
+          - name: Linux release [LLVM]
```
Kha, line 53: Let's call it "Linux LLVM" for now, it's not a release job

.github/workflows/ci.yml
```diff
@@ -90,7 +100,7 @@ jobs:
             shell: bash -euxo pipefail {0}
             CMAKE_OPTIONS: -DCMAKE_OSX_DEPLOYMENT_TARGET=10.15 -DUSE_GMP=OFF -DLEAN_INSTALL_SUFFIX=-darwin_aarch64
             llvm-url: https://github.com/leanprover/lean-llvm/releases/download/15.0.1/lean-llvm-aarch64-apple-darwin.tar.zst https://github.com/leanprover/lean-llvm/releases/download/15.0.1/lean-llvm-x86_64-apple-darwin.tar.zst
-            prepare-llvm: EXTRA_FLAGS=--target=aarch64-apple-darwin ../script/prepare-llvm-macos.sh lean-llvm-aarch64-* lean-llvm-x86_64-*
```
Kha, line 93: Revert, this flag is used by clang as well

Kha (changes requested):

hargoniX:
The thing that is making the CI fail on aarch64 is fixed in the current `shell.cpp`, I guess we need an update-stage0 for this?

script/prepare-llvm-linux.sh
```diff
@@ -48,10 +48,10 @@ $CP -r llvm/include/*-*-* llvm-host/include/
 $CP $GLIBC/lib/libc_nonshared.a stage1/lib/glibc
 for f in $GLIBC/lib/lib{c,dl,m,rt,pthread}-*; do b=$(basename $f); cp $f stage1/lib/glibc/${b%-*}.so; done
 OPTIONS=()
-echo -n " -DLLVM=ON -DLLVM_CONFIG=$PWD/llvm-host/bin/llvm-config" # manually point to `llvm-config` location
+echo -n " -DLLVM_CONFIG=$PWD/llvm-host/bin/llvm-config" # manually point to `llvm-config` location
 echo -n " -DLEAN_STANDALONE=ON"
 echo -n " -DCMAKE_CXX_COMPILER=$PWD/llvm-host/bin/clang++ -DLEAN_CXX_STDLIB='-Wl,-Bstatic -lc++ -lc++abi -Wl,-Bdynamic'"
-echo -n " -DLEAN_EXTRA_CXX_FLAGS='--sysroot $PWD/llvm -idirafter $GLIBC_DEV/include ${EXTRA_FLAGS:-}'"
+echo -n " -DLEAN_EXTRA_CXX_FLAGS='--sysroot $PWD/llvm -idirafter $GLIBC_DEV/include ${EXTRA_FLAGS:-}' -DLEAN_EXTRA_MAKE_OPTS='${EXTRA_FLAGS:-}'"
```
Kha, line 54: We can revert this to avoid the shell.cpp change

Kha: Let's move it to the `-DLLVM` line and then comment that line out?

bollu: @Kha I do not follow what you mean here :)

Kha: Easier to do than to explain :)

Kha:
The `weak` workaround has yet to be reinstated. While doing that, we should document what it does and why it is necessary. I would also welcome opinions on whether there are alternative approaches to it, it is quite a departure from the C backend after all and the repercussions are not clear to me.

tobiasgrosser:
> > We removed the `static` from these functions in lean.h.bc, but in order to sucessfully link together resulting object files, we must not expose the functions as global symbols (i.e. make sure they are eliminated or make them hidden or weak)
> 
> ```
> /nix/store/fx1h3mmzl2hjci3nn15n26r57mc3cs3z-binutils-2.40/bin/ld: /home/sebastian/lean/lean/build/release/stage2/lib/lean/libLean.a(Widget.o): in function `lean_set_st_header':
> Lean.Widget:(.text+0x4b0): multiple definition of `lean_set_st_header'; /home/sebastian/lean/lean/build/release/stage2/lib/lean/libInit.a(EState.o):Init.Control.EState:(.text+0xdf0): first defined here
> ```
I guess we want to set the function linkage type to `internal` then:

https://llvm.org/docs/LangRef.html#linkage-types
> Similar to private, but the value shows as a local symbol (STB_LOCAL in the case of ELF) in the object file. This corresponds to the notion of the ‘static’ keyword in C.

Kha:
Sounds good. At which step of the pipeline should we do that, after loading the clang-generated bitcode but before linking it with the Lean backend output? I assume the overhead of redoing this (in memory) for every Lean file should be negligible.

tobiasgrosser:
> Sounds good. At which step of the pipeline should we do that, after loading the clang-generated bitcode but before linking it with the Lean backend output? I assume the overhead of redoing this (in memory) for every Lean file should be negligible.

I am not sure I have enough context to answer this. Why did we remove the static in the first place?

Kha:
The unfortunate problem is that clang will not generate any bitcode at all if we give it lean.h as is with `static inline`s. So we remove the `static` for generating lean.h.bc and now just have to revert that change later somewhere on the bitcode level.

Kha:
That sounds handy, but the docs seem to imply it doesn't do anything if there is no `main`? https://llvm.org/docs/Passes.html#internalize-internalize-global-symbols

tobiasgrosser:
> That sounds handy, but the docs seem to imply it doesn't do anything if there is no `main`? https://llvm.org/docs/Passes.html#internalize-internalize-global-symbols

I teste it and it works even if there is no main function.

Kha:
Awesome, let me try that right now

Kha:
Heh, perhaps not unexpectedly that already breaks linking in the header with the backend output. So we would have to selectively internalize these symbols after this linking step.

tobiasgrosser:
There is a clang option named `-femit-all-decls` that should force clang to also emit `static inline` functions. Is this maybe what we want?

Kha:
That's also good to know but it probably won't solve the issue of `LLVMLinkModules2` ignoring internal symbols, no?

tobiasgrosser:
I thought about this a little more. `LLVMLinkModules2` does not expose functionality to internalize definitions while linking and the high-level C++ interface used in llvm-linker also requires a callback that does the actual internalization:

```
L.linkInModule(
          std::move(M), ApplicableFlags, [](Module &M, const StringSet<> &GVS) {
            internalizeModule(M, [&GVS](const GlobalValue &GV) {
              return !GV.hasName() || (GVS.count(GV.getName()) == 0);
            });
          });
```

Hence, our best bet would probably after calling LLVMLinkModule2 to iterate over the resulting module and call `void LLVMSetLinkage(LLVMValueRef Global, LLVMLinkage Linkage);` with `LLVMExternalLinkage` on any function that is also contained in the llvm.h.bc file?

tobiasgrosser:
@bollu is on it, so let's see if that works.

bollu:
Here is the `lean.h.bc` file: https://gist.github.com/bollu/5700b2b693bfe1de2addaa2957699f6d

I must admit, I find the failure quite puzzling. My best guess is that I'm using the API wrong, so I'm going to write a small test C++/python script to check that we are indeed using the API correctly to grab functions in the module.

tobiasgrosser:
Hey @Siddharth, I feel that we first need to revert @Kha's patch that internalizes the functions before linking. We meanwhile learned that this will prevent any of them to be linked into the module. Second, `.str` is private so likely won't be linked. Did you try to just drop the exception and continue even if some symbols are not found?

tobiasgrosser:
Also, do we need to rebuild stage0 after the latest changes?

bollu:
[New error](https://github.com/leanprover/lean4/actions/runs/5775636172/job/15653888341#step:14:13), now after successfully compiling: 

```
2023-08-06T09:17:51.2248970Z ##[endgroup]
2023-08-06T09:17:52.5048863Z + build/stage1/bin/lean --stats src/Lean.lean
2023-08-06T09:17:52.5248665Z /home/runner/work/_temp/5f638e92-27e3-4d32-b1ae-79fdae16a66f: line 1: 33726 Bus error               build/stage1/bin/lean --stats src/Lean.lean
2023-08-06T09:17:52.5284898Z ##[error]Process completed with exit code 135.
```

It seems like `lean` the executable is borked. I can reproduce the segfault locally. 

```
➜  lean4 gdb --args ./build/stage1/bin/lean stats src/Lean.lean
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7a2aeb7 in lean_mark_persistent () from /home/t-sibhat/projects/lean-llvm/lean4/build/stage1/bin/../lib/lean/libleanshared.so
(gdb) bt
#0  0x00007ffff7a2aeb7 in lean_mark_persistent () from /home/t-sibhat/projects/lean-llvm/lean4/build/stage1/bin/../lib/lean/libleanshared.so
#1  0x00007ffff60cc49a in initialize_Init_Prelude () from /home/t-sibhat/projects/lean-llvm/lean4/build/stage1/bin/../lib/lean/libleanshared.so
#2  0x00007ffff620f73e in initialize_Init () from /home/t-sibhat/projects/lean-llvm/lean4/build/stage1/bin/../lib/lean/libleanshared.so
#3  0x00007ffff7a26e9f in lean::initializer::initializer() ()
   from /home/t-sibhat/projects/lean-llvm/lean4/build/stage1/bin/../lib/lean/libleanshared.so
#4  0x00007ffff7920dc9 in lean_main () from /home/t-sibhat/projects/lean-llvm/lean4/build/stage1/bin/../lib/lean/libleanshared.so
#5  0x00007ffff578dd90 in __libc_start_call_main (main=main@entry=0x555555555140 <main>, argc=argc@entry=3, argv=argv@entry=0x7fffffffdce8)
    at ../sysdeps/nptl/libc_start_call_main.h:58
#6  0x00007ffff578de40 in __libc_start_main_impl (main=0x555555555140 <main>, argc=3, argv=0x7fffffffdce8, init=<optimized out>,
    fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffdcd8) at ../csu/libc-start.c:392
#7  0x0000555555555075 in _start ()
(gdb)
```

bollu:
```
➜  lean4 opt -S  build/stage1/lib/temp/Init/Prelude.bc > ~/temp/prelude.ll
Global is external, but doesn't have external or weak linkage!
ptr @lean_alloc_small
Global is external, but doesn't have external or weak linkage!
ptr @lean_inc_ref_cold
Global is external, but doesn't have external or weak linkage!
ptr @lean_dec_ref_cold
Global is external, but doesn't have external or weak linkage!
ptr @lean_alloc_object
Global is external, but doesn't have external or weak linkage!
ptr @lean_internal_panic_out_of_memory
Global is external, but doesn't have external or weak linkage!
ptr @lean_array_get_panic
Global is external, but doesn't have external or weak linkage!
ptr @lean_copy_expand_array
Global is external, but doesn't have external or weak linkage!
ptr @lean_array_set_panic
...
```


Note that in [lean.h.ll](https://gist.github.com/bollu/5700b2b693bfe1de2addaa2957699f6d#file-lean-h-ll-L211), declarations are not marked with `internal`:
```
declare i8* @lean_alloc_small(i32 noundef, i32 noundef) local_unnamed_addr #2
declare void @lean_inc_ref_cold(%struct.lean_object* noundef) local_unnamed_addr #2
```

while definitions are marked `internal`:

```
define internal %struct.lean_object* @lean_alloc_ctor_memory(i32 noundef %0) local_unnamed_addr #1 {
```

We should do the same.

bollu:
OK, now the failures are:

```
libleanshared.so: undefined reference to `LLVMInitializeMSP430AsmPrinter@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeLanaiAsmPrinter@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeMipsTargetMC@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeRISCVTargetInfo@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeSparcTargetMC@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeVETarget@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeVETargetMC@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeXCoreAsmPrinter@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeARMTargetInfo@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeLanaiTargetMC@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeBPFAsmParser@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeXCoreTargetInfo@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeARMAsmParser@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializePowerPCTarget@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeMipsAsmParser@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeAVRTargetInfo@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeRISCVTargetMC@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeNVPTXAsmPrinter@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializePowerPCTargetMC@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeMSP430TargetInfo@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeBPFAsmPrinter@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeLanaiTargetInfo@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeNVPTXTarget@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeVEAsmPrinter@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeSystemZAsmPrinter@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeSparcTarget@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeRISCVAsmParser@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeMipsTarget@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeAMDGPUAsmPrinter@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeMSP430AsmParser@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeAMDGPUTargetInfo@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeMipsTargetInfo@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeARMTargetMC@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeLanaiTarget@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeHexagonAsmParser@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeHexagonTargetMC@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeNVPTXTargetMC@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeRISCVTarget@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeNVPTXTargetInfo@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeXCoreTarget@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeMSP430Target@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeMipsAsmPrinter@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializePowerPCTargetInfo@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeVETargetInfo@LLVM_15'
libleanshared.so: undefined reference to `LLVMInitializeHexagonAsmPrinter@LLVM_15'
```

I have seen this before with @hargoniX , when we accidentally include an *incorrect* `llvm/Config/Targets.def`, due to multiple copies of the LLVM headers floating around (system LLVM and our LLVM). The system LLVM is usually built with support for all targets, while our flavour of LLVM is built with support for X86 and aarch64. 

If we accidentally include the system LLVM headers, this causes the code in [`llvm-c/Target.h`](https://llvm.org/doxygen/llvm-c_2Target_8h_source.html) to invoke `LLVMInitialize##TargetName##Target` on more targets than we have built support for in our copy of LLVM. This fails at link time, since we call the target initialization functions of targets we have not built. 

The offending lines are:

```cpp
// https://llvm.org/doxygen/llvm-c_2Target_8h_source.html
#define LLVM_TARGET(TargetName) void LLVMInitialize##TargetName##Target(void);
#include "llvm/Config/Targets.def"
#undef LLVM_TARGET  /* Explicit undef to make SWIG happier */
```

I believe this theory is correct, since in the list of `undefined reference LLVMInitialize##TargetName##Target`, we **do not miss** `X86` and `AArch64`. 

@Kha , did some of the CMake changes accidentally mess up our include paths?

src/CMakeLists.txt
```diff
@@ -312,23 +312,26 @@ endif()
...
     # - However, the target's `llvm/bin/llvm-config` has a different target
     #   triple from the host, and thus cannot be run on the host.
     # - So, we run the host `llvm-host/bin/llvm-config` from which we pick up
     #   compiler options, and change the output of the host to point to the target.
-    # - In particular, `host/bin/llvm-config` produces flags like `-Lllvm-host/lib/libLLVM`, while 
+    # - In particular, `host/bin/llvm-config` produces flags like `-Lllvm-host/lib/libLLVM`, while
     #   we need the path to be `-Lllvm/lib/libLLVM`. Thus, we perform this replacement here.
-    string(APPEND LEANSHARED_LINKER_FLAGS " -L${LLVM_CONFIG_LIBDIR} ${LLVM_CONFIG_LDFLAGS} ${LLVM_CONFIG_LIBS} ${LLVM_CONFIG_SYSTEM_LIBS}")
     string(REPLACE "llvm-host" "llvm" LEANSHARED_LINKER_FLAGS ${LEANSHARED_LINKER_FLAGS})
-    string(APPEND CMAKE_CXX_FLAGS  " -I${LLVM_CONFIG_INCLUDEDIR}")
```
bollu, line 331: @Kha I suspect that this was (miraculously) setting the include paths correctly to pick up the right `Targets.def`. I shall test this now.

bollu: EDIT: Ah, I see, we merely moved them up...

bollu:
In the build, we appear to be picking up `llvm-config` from `nix`, and not from our copy of the LLVM toolchain: 

```
2023-08-06T23:11:02.3213334Z -- '-DLLVM_CONFIG=<path/to/llvm-config>' not passed as CMake flag, finding `llvm-config` via CMake...
2023-08-06T23:11:02.3219855Z -- Executing 'llvm-config --version' at '/nix/store/qqj4bp95wf9nks5fpb9413zhxv0wnbpk-llvm-15.0.7-dev/bin/llvm-config' to check configuration.
2023-08-06T23:11:02.3249929Z 15.0.7
2023-08-06T23:11:02.3255852Z -- Found 'llvm-config' at '/nix/store/qqj4bp95wf9nks5fpb9413zhxv0wnbpk-llvm-15.0.7-dev/bin/llvm-config' with version '15.0.7', major version '15'
```

I now suspect the diff at `script/prepare-llvm-linux.sh`.

script/prepare-llvm-linux.sh
```diff
@@ -48,7 +48,8 @@ $CP -r llvm/include/*-*-* llvm-host/include/
 $CP $GLIBC/lib/libc_nonshared.a stage1/lib/glibc
 for f in $GLIBC/lib/lib{c,dl,m,rt,pthread}-*; do b=$(basename $f); cp $f stage1/lib/glibc/${b%-*}.so; done
 OPTIONS=()
-echo -n " -DLLVM=ON -DLLVM_CONFIG=$PWD/llvm-host/bin/llvm-config" # manually point to `llvm-config` location
```
bollu, line 51: @Kha do we not need to point our build to the correct `llvm-config` which we download from [`leanprover/lean-llvm`](https://github.com/leanprover/lean-llvm)?

Kha: Oh yes, Linux LLVM should use this file as well and this line should become conditional. I can do the change of desired.

bollu: Yes please, that would be very super helpful!

tobiasgrosser: Yes, this seems to now all go through, modulo the one remaining change to make this line conditional. @Kha, if you could take care of this that would be much appreciated.

hargoniX:
The CI is passing with the latest update-stage0 now. One last question from Sid and me: we did see that the LLVM build in its current configuration does not build stage2/stage3 because the `check-stage3` flag is missing in the Github Action. However it is currently used mutually exclusive with the `release` flag in the action. Do we want to set only `check-stage3` (we definitely want this because the entire point of the PR is to basically get this working) or `check-stage3` + `release`?

script/prepare-llvm-linux.sh
Kha: Thanks @hargoniX!

bollu:
@Kha I shall repeat the process peeling the new LLVM bindings + `update-stage0` into a separate commit, followed by a final change to `EmitLLVM.lean`, CMake, CI, and build scripts.

tobiasgrosser:
I pulled out some whitespace changes in https://github.com/leanprover/lean4/pull/2417/files and also merged master to see what diff is still remaining. I feel we should pull the EmitLLVM changes as a separate PR (together with the necessary stage0 updates), before we push the CI changes. My hope is that this will make this very PR small, such that we can start cleaning up the CI changes, starting off from a working PR.

tobiasgrosser:
This PR has been upstreamed in small individual pieces in:
- https://github.com/leanprover/lean4/pull/2418
- https://github.com/leanprover/lean4/pull/2417
- https://github.com/leanprover/lean4/pull/2416
- https://github.com/leanprover/lean4/pull/2410
- https://github.com/leanprover/lean4/pull/2384
- https://github.com/leanprover/lean4/pull/2375
- https://github.com/leanprover/lean4/pull/2365
- https://github.com/leanprover/lean4/pull/2363
- https://github.com/leanprover/lean4/pull/2358

It can now be closed. I do not have permission to close PRs. So maybe someone can help out.

PR #2341, opened 2023-07-21
feat: add namespace utility function

fgdorais:
It was pointed out in [std4#184](https://github.com/leanprover/std4/pull/184) that this function is frequently reimplemented and that it may find a better home here.

### DIFF
--- a/src/Lean/Data/OpenDecl.lean
+++ b/src/Lean/Data/OpenDecl.lean
@@ -28,13 +28,15 @@ def rootNamespace := `_root_
 def removeRoot (n : Name) : Name :=
   n.replacePrefix rootNamespace Name.anonymous
 
-/-- Adds namespace prefix unless in root namespace, when it just removes `_root_`.
+/-- `addNamespaceUnlessRoot ns n` adds namespace `ns` prefix unless `n` has a `_root_` prefix.
+If `n` has a `_root_` prefix then it is removed and `ns` is not added. The end result is a fully
+qualified name assuming that `ns` is a fully qualified namespace prefix.
 ```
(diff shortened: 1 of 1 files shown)


digama0:
Functions I found implementing this function or a variant:
* `Lean.Elab.mkDeclName`
* `Lean.Elab.Command.elabSyntax` doesn't use the function, but should (see #2239). That PR introduces 4 reimplementations of the function
* `Lean.Elab.Term.elabWithDeclName` should also be using the function (basically every use of `(← getCurrNamespace) ++ id` is suspect)

fgdorais:
I'm wondering if removing the `_root_` prefix from the rhs is a bit excessive. Maybe the user should decide whether to remove it afterwards.

digama0:
I think the function should be specced to always remove the _root_ prefix and return a fully qualified name. The use code can put `_root_` on the front if needed.

fgdorais:
Should it then also remove `_root_` from the namespace prefix, or just assume that it is a proper namespace?

digama0:
Namespaces can't have `_root_`.

fgdorais:
So it is fine as is, with a small edit to the docs?

fgdorais:
Fixed!

fgdorais:
[std4#184](https://github.com/leanprover/std4/pull/184) has evolved into [std4#200](https://github.com/leanprover/std4/pull/200), which doesn't currently use this function. I'm not closing this since this does not imply that this PR is not useful for other purposes.

Kha:
Thanks for the PR. At this point I think it's fair to say it will not be merged by itself but might still be included in other PRs where it makes sense.

PR #2342, opened 2023-07-21
doc: fix type signature of `Coe`

chabulhwi:
Change `Type` to `Sort`.


Kha:
What is the motivation for this PR?

chabulhwi:
`Coe` is defined this way: https://github.com/leanprover/lean4/blob/6e904421304c362a146883baa360f451376dd9b1/src/Init/Coe.lean#L126-L129.

Kha:
Hah my bad, I missed that

PR #2345, opened 2023-07-22
feat: add `IO.FS.rename`

digama0:


fgdorais (approved):

src/runtime/io.cpp
```diff
@@ -657,6 +657,14 @@ extern "C" LEAN_EXPORT obj_res lean_io_remove_dir(b_obj_arg p, obj_arg) {
     }
 }
 
+extern "C" LEAN_EXPORT obj_res lean_io_rename(b_obj_arg from, b_obj_arg to) {
+    if (std::rename(string_cstr(from), string_cstr(to)) == 0) {
+        return io_result_mk_ok(box(0));
+    } else {
+        return io_result_mk_error(decode_io_error(errno, from));
```
Kha, line 664: The arbitrary choice of `from` is certainly not optimal, I would at least have expected a comment here. Perhaps it wouldn't be too bad to include both file names in the string, it is designed for display purposes only.

Kha (changes requested):

src/runtime/io.cpp
digama0: The function doesn't seem to lend itself well to that... I would have to inline the whole thing and customize all the errors.

Kha: I'm not sure what "that" is. I was thinking of literally passing `"{from} and/or {to}"` as `fname`. It's a bit silly but at least not misleading.

Kha:
Thanks!

PR #2348, opened 2023-07-23
feat: golf proof of `Quotient.exact`

chabulhwi:


Kha:
I'm not sure this kind of minimal-improvement PR is worthwhile. Hopefully at some point we will have code formatters and linters that can do cleanup like this automatically.

Issue #1659, opened 2022-09-28
Completion should show results in namespaces


digama0:
```lean
import Lean

#check elabTermEnsuringType
                         --^ textDocument/completion
```
returns no result, even though the document symbols (ctrl-T) request does reveal `Lean.Elab.Term.elabTermEnsuringType` and `Lean.Elab.Tactic.elabTermEnsuringType`. This is a regression compared to lean 3 and it's quite painful to not be able to complete things in other namespaces.

PR #2351, opened 2023-07-24
Auto-complete declaration names in arbitrary namespaces

Kha:
This fixes exactly #1659. It does not (yet) attempt to do fuzzy-matching across the entire declaration name including namespaces, as that is potentially slow and can yield unhelpful results if the namespace is the *only* thing that matched; on the other hand, some examples in https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/autocomplete.20woes are of this kind, so this is something we should revisit in the future.

### DIFF
--- a/src/Lean/Server/Completion.lean
+++ b/src/Lean/Server/Completion.lean
@@ -69,7 +69,7 @@ private def sortCompletionItems (items : Array (CompletionItem × Float)) : Arra
 private def mkCompletionItem (label : Name) (type : Expr) (docString? : Option String) (kind : CompletionItemKind) : MetaM CompletionItem := do
   let doc? := docString?.map fun docString => { value := docString, kind := MarkupKind.markdown : MarkupContent }
   let detail ← consumeImplicitPrefix type fun type => return toString (← Meta.ppExpr type)
-  return { label := label.getString!, detail? := detail, documentation? := doc?, kind? := kind }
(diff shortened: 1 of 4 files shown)


eric-wieser:
> It does not (yet) attempt to do fuzzy-matching across the entire declaration name including namespaces,

One thing that's quite common is to not know whether the lemma you're looking for is of the form `SomeProp.someLemma` or `someProp_someLemma` (`_` vs `.` and `s` vs `S`). Is this something that autocomplete can help with as of this PR?

Kha:
> Is this something that autocomplete can help with as of this PR?

Only if matching against `someLemma` is sufficient for finding it

leodemoura:
We need more autocompletion tests at `tests/lean/interactive`.

Kha:
> > It does not (yet) attempt to do fuzzy-matching across the entire declaration name including namespaces,
> 
> One thing that's quite common is to not know whether the lemma you're looking for is of the form `SomeProp.someLemma` or `someProp_someLemma` (`_` vs `.` and `s` vs `S`). Is this something that autocomplete can help with as of this PR?

I'm wondering whether including the last *two* segments, i.e. usually the type/prop and function/theorem would yield a sensible balance of exhaustiveness, relevance, and performance. But that feels a bit arbitrary, let's go with this more limited and clear-cut extension for now.

Kha:
I think this is ready to go as a relatively conservative extension of the current behavior, unless someone can think of additional interesting tests to add

PR #2352, opened 2023-07-24
fix: missing `mkCIdent`s in Lean.Elab.Deriving.Util

### LINKED ISSUE #2344: `deriving DecidableEq` bug
```lean
namespace Value

inductive Primitive where
  | Bool (b : Bool)
  | Int (i : Int)
  | String (s : String)

deriving instance DecidableEq for Primitive -- Works

inductive Value where
  | Primitive (p : Primitive)

deriving instance DecidableEq for Primitive -- Fails
deriving instance DecidableEq for Value.Primitive -- Fails 
deriving instance DecidableEq for _root_.Value.Primitive -- Fails
```

Kha:
Fixes #2344 

Whenever we move names from `Expr` world to `Syntax` world, we need to think about whether we want `mkCIdent` or `mkIdent`. If we wanted to have the type system help with this, we could have a `ConstName` newtype that uses `mkCIdent` when coerced to `Syntax`.

### DIFF
--- a/src/Lean/Elab/Deriving/Util.lean
+++ b/src/Lean/Elab/Deriving/Util.lean
@@ -27,7 +27,7 @@ def mkInductArgNames (indVal : InductiveVal) : TermElabM (Array Name) := do
 
 /-- Return the inductive declaration's type applied to the arguments in `argNames`. -/
 def mkInductiveApp (indVal : InductiveVal) (argNames : Array Name) : TermElabM Term :=
-  let f    := mkIdent indVal.name
+  let f    := mkCIdent indVal.name
   let args := argNames.map mkIdent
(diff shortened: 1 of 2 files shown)


Kha:
An alternative would be to rename `mkIdent` to `mkLIdent` (for "local") and deprecate the old name to make the distinction more clear.

Kha:
> we could have a `ConstName` newtype

FWIW, https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.E2.9C.94.20Pretty-Printing.20Name.20in.20InfoView brought up a use case for embedding `ConstName`s in `MessageData`. It would work with `Name` as well of course, but one would have to be more careful.

PR #2355, opened 2023-07-25
fix: bug in gen_injective_theorems%

digama0:
In the example
```lean
inductive WF : Nat → Type 1
  | mk (α : Type) (fn : α → Nat) (arg : α) : WF (fn arg)
```
the implementation was treating `fn` and `arg` as fixed because they appear in the target type, but not `α` because it is not in the target type, leading to a type error because the type of `fn` depends on `α`. Now we take the transitive closure of the variables in the target type under the backward dependency relation.

Surprisingly there was not a function which does this backward dependency calculation (it is implemented in bits and pieces scattered around, and forward dependency has a function for it but not backward dependency) so I exposed a function to do it as part of `CollectFVars`.

### DIFF
--- a/src/Lean/Util/CollectFVars.lean
+++ b/src/Lean/Util/CollectFVars.lean
@@ -35,9 +35,24 @@ mutual
     | _                    => id
 end
 
+partial def collectBackwardDeps (lctx : LocalContext) (i : Nat) : Visitor := fun s =>
+  if h : i < s.fvarIds.size then
+    let s := if let some ldecl := lctx.find? s.fvarIds[i] then
+      visit ldecl.type s
+    else
+      s
+    collectBackwardDeps lctx (i + 1) s
+  else
+    s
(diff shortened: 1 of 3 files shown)


src/Lean/Util/CollectFVars.lean
```diff
@@ -35,9 +35,24 @@ mutual
     | _                    => id
 end
 
+partial def collectBackwardDeps (lctx : LocalContext) (i : Nat) : Visitor := fun s =>
+  if h : i < s.fvarIds.size then
+    let s := if let some ldecl := lctx.find? s.fvarIds[i] then
+      visit ldecl.type s
```
leodemoura, line 41: Missing `ldecl.value?`. 
It will not affect the current PR since we don't use let-decls in inductive declarations, but it may affect future uses of `collectBackwardDeps`.

PR #2358, opened 2023-07-27
chore: disable lake 116 test.

bollu:
The test is flaky, and thus we disable it to improve CI reliability. Ping @tydeu

### DIFF
--- a/src/shell/CMakeLists.txt
+++ b/src/shell/CMakeLists.txt
@@ -178,8 +178,9 @@ FOREACH(T ${LEANTESTS})
 ENDFOREACH(T)
 
 # Create a lake test for each subdirectory of `lake` which contains a `test.sh` file.
-# Disable tests that are prone to flake by construction
-# (e.g. laketest_116 has a 'sleep()' for a fixed duration).
+# bootstrap: too slow
+# 104, git: not adapted to living in the lean4 repo yet
+# 116: non-deterministic (`sleep`)
 file(GLOB_RECURSE LEANLAKETESTS "${LEAN_SOURCE_DIR}/lake/test.sh")
 FOREACH(T ${LEANLAKETESTS})
   if(NOT T MATCHES ".*(lake-packages|bootstrap|104|116|git).*")


Kha:
Where is the evidence for flakiness? Didn't it fail every single time on your PR on Linux Release after a certain point?

bollu:
Writing down for future documentation: The test passes upon increasing the timeout at #2358 (https://github.com/leanprover/lean4/pull/2340#issuecomment-1653538059). @Kha , would you rather I increase the timeout (to 30s? 10s?) rather than disable the test altogether?

tydeu:
@Kha Regarding https://github.com/leanprover/lean4/issues/2340#issuecomment-1653998617, since my https://github.com/leanprover/lean4/pull/2337#discussion_r1268011791 I have remembered why a shutdown notification was insufficient. The reason this has a timeout is that it wants to wait until the build completes to test that the it prints the proper error. Without a timeout, the process will be killed immediately and the build will not complete, preventing the proper output from being produced. Thus, to avoid a timeout, the test needs some way to tell the server to die only after the build completes.

Kha:
I see, we use `Ipc.collectDiagnostics` for that in the server tests

tydeu:
@bollu I vote for (1): Just do the bump,

Kha:
After some more consideration, I am not at all comfortable with having an arbitrary timeout in our test suite. The offending part of the test should be disabled pending a rewrite using `Ipc`.

tydeu:
@Kha In that case, we should probably just disable the test completely like this PR originally did. The test doesn't do anything else.

tydeu:
@Kha Oh, honestly, I forgot about that part. However, I do not think worth keeping the test enabled. Without the test to check that serve actually properly sets the environment variable, testing the environment variable itself is minimally useful. Furthermore, commenting it out would just make it more annoying to test manually as it would require uncommenting during and then re-commenting for commit.

bollu:
@Kha I've reverted the patch back to its original behavior of disabling `116`.

src/shell/CMakeLists.txt
```diff
@@ -178,9 +178,11 @@ FOREACH(T ${LEANTESTS})
 ENDFOREACH(T)
 
 # Create a lake test for each subdirectory of `lake` which contains a `test.sh` file.
+# Disable tests that are prone to flake by construction
+# (e.g. laketest_116 has a 'sleep()' for a fixed duration).
```
Kha, line 182:
```suggestion
# bootstrap: too slow
# 104, git: not adapted to living in the lean4 repo yet
# 116: non-deterministic (`sleep`)
```

tydeu: @Kha Oh, one thing I had not noticed before  - why is bootstrap considered "too slow"?

Kha: I don't know where the cut-off for "too slow" should be, but it is certainly much slower than all other tests
```
1692/1706 Test #1686: leaninteractivetest_stdOutput.lean ........................   Passed    5.02 sec
1695/1706 Test #1690: leanlaketest_deps .........................................   Passed    5.49 sec
1700/1706 Test #1693: leanlaketest_init .........................................   Passed    5.52 sec
 737/1706 Test  #417: leantest_notationDelab.lean ...............................   Passed    6.42 sec
1636/1706 Test #1607: leanbenchtest_liasolver.lean ..............................   Passed    6.61 sec
1326/1706 Test #1158: leanruntest_lcnf4.lean ....................................   Passed    6.78 sec
1156/1706 Test  #907: leanruntest_constProp.lean ................................   Passed    7.56 sec
1074/1706 Test  #823: leanruntest_aStructPerfIssue.lean .........................   Passed    8.52 sec
1530/1706 Test #1293: leanruntest_printDecls.lean ...............................   Passed    9.03 sec
1703/1706 Test #1691: leanlaketest_ffi ..........................................   Passed    9.17 sec
 552/1706 Test  #164: leantest_PPRoundtrip.lean .................................   Passed    9.81 sec
1490/1706 Test #1216: leanruntest_meta2.lean ....................................   Passed   10.38 sec
1704/1706 Test #1696: leanlaketest_targets ......................................   Passed   12.47 sec
1705/1706 Test #1689: leanlaketest_bootstrap ....................................   Passed   22.74 sec
```

tydeu: @Kha Ah, yeah. Admittedly, that is expected as it is rebuilding all of Lake. I think of it kind of like a Lean stage2 build but for Lake.

PR #2363, opened 2023-07-27
chore: get rid of all inline C annotations for LLVM

bollu:
the LLVM backend cannot code generate `inline c` annotations, so we outline these into `lean.h`. 

Reference for `extern` versus `extern c` semantics: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Difference.20between.20extern.20and.20extern.20c.3F/near/321191966

Peeled from https://github.com/leanprover/lean4/pull/2340 to enable LLVM for stage1+ builds.

### DIFF
--- a/src/runtime/CMakeLists.txt
+++ b/src/runtime/CMakeLists.txt
@@ -23,7 +23,7 @@ if(LLVM)
   endif()
   FILE(READ "${CMAKE_CURRENT_SOURCE_DIR}/../include/lean/lean.h" LEAN_H)
   # generate LLVM IR for `static inline` definitions in lean.h for the LLVM backend
-  string(REPLACE "static inline" "__attribute__((always_inline, weak))" LEAN_H "${LEAN_H}")
+  string(REPLACE "static inline" "__attribute__((always_inline))" LEAN_H "${LEAN_H}")
   # drop '#pragma once' in .c file to avoid warning
   string(REPLACE "#pragma once" "" LEAN_H "${LEAN_H}")
   file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/lean_inlines.c" "${LEAN_H}")


bollu:
@tydeu I've incorporated your suggested changes here.

tydeu:
@bollu I would have left `version` (and possibly `internal`) in the name for those extern symbols as that is part of the C version constant's name and the definitions are within a `Version` (or `Internal`) namespace in Lean. Otherwise, LGTM! 😄

src/runtime/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ if(LLVM)
   endif()
   FILE(READ "${CMAKE_CURRENT_SOURCE_DIR}/../include/lean/lean.h" LEAN_H)
   # generate LLVM IR for `static inline` definitions in lean.h for the LLVM backend
-  string(REPLACE "static inline" "__attribute__((always_inline))" LEAN_H "${LEAN_H}")
+  string(REPLACE "static inline" "__attribute__((always_inline, weak))" LEAN_H "${LEAN_H}")
```
Kha, line 26: Wrong PR?

Kha (changes requested):

src/runtime/CMakeLists.txt
bollu: I'm unsure where this change should live. 

I was wondering if it ought to be here, since it sets up the outlining for the backend, or in the "main PR" which sets up the build system.

I'll move it over to the other PR, since I guess that you feel it's more appropriate there.

bollu:
@Kha I revert the changes in `src/runtime/CMakeLists.txt`

bollu:
@Kha can we run a `!bench` on this PR to make sure that just this change does not lead to a change in perf?

tydeu:
!bench

tydeu:
Interesting, while the bot acknowledge my `!bench` with a rocket, it does not appear to have actually succeeded.  I will leave it @Kha then.

tydeu:
Oh, so the `!bench` did work, it just took much longer than I expected. Interesting. 🤔

Kha:
Yes, because the queue is shared with mathlib4

bollu:
@Kha I'm happpy to merge the PR from my end.

Issue #2064, opened 2023-01-26
RFC: Widget messages


Vtec234:
# RFC: Widget messages

The current [user widgets](https://leanprover.github.io/lean4/doc/examples/widgets.lean.html) system supports exactly one kind of user widget - a *widget panel*, that is a piece of UI which can appear as a top-level panel next to the tactic state, messages, etc, in the infoview. Since we already have all the infrastructure for dynamic code loading in the infoview set up, there is no reason to limit ourselves to panels. Here I would like to propose *widget messages*. Implementation-wise, these would need one new constructor of `MessageData`, `MessageData.ofWidget (msg : WidgetMsg)`. I have not implemented this, but I am happy to do it if there is interest/approval. Widget messages are primarily motivated by the use-cases they would enable:

## Example 1: "Try this"

We can write a generic "Try this" UI component resulting in `tryThis : Syntax → Syntax → WidgetMsg` where `tryThis stxRef newStx` is the data of a "Try this" display that replaces `stxRef` with `newStx` upon a click. We would then include it in messages output by tactic code using something like `m!"<some tactic output>{.tryThis stxRef newStx}"`.

## Example 2: Contextual traces in failure messages

It is currently not always easy to view extra detail associated with a failure message. For instance, suppose the system prints a `failed to synthesize instance SomeClass α` message. In cases where it is not obvious why the synthesis failed (e.g. when debugging typeclass hierarchies), we can debug it by setting `trace.Meta.synthInstance true`. However we must either do so on the entire enclosing command which produces lots of irrelevant output, or extract a self-contained `trace.Meta.synthInstance true in #synth SomeClass α` which may be non-trivial if `α` depends on local context. And in any case we have to type the option in every time! It would be convenient if the failure message directly included a `[Trace]` button below it which would show this output when clicked. This could be achieved straightforwardly with a widget message. The idea extends to other failures, e.g. `trace.Meta.isDefEq` traces for `expected .. but got ..`. Note that this is *not* supposed to collect traces eagerly as these can use up substantial memory. Rather, the data of a `[Trace]` button is just the inputs to the failed computation, and clicking it would rerun the computation with tracing enabled.

## Pitfalls and alternatives

[...]

PR #2364, opened 2023-07-27
fix: handle error in withTraceNode message action

### LINKED ISSUE #2362: Trace messages disappear when `msg` fails in `withTraceNode`
### Description

In the following snippet, the trace message action that throws `"oh no"` in `abc` makes the message `"pre abc"` in `ghi` disappear.

```lean
open Lean Elab Term

def abc : TermElabM Unit :=
  withTraceNode `abc (fun _ => throwError "oh no") do
    trace[abc] "a message"

def ghi : TermElabM Unit :=
  withTraceNode `ghi (fun 
      | .error e => pure m!"fail: {e.toMessageData}"
      | .ok _ => pure m!"success") do
    trace[ghi] "pre abc"
    abc
    trace[ghi] "post abc"

elab "ts" : term => do
  withOptions (fun opts => opts.setBool `trace.abc true |>.setBool `trace.ghi true)
    ghi
  elabTerm (← `(37)) none

/-
[ghi] fail: oh no
  [abc] a message
-/
#check ts
```

### Steps to Reproduce

1. Run the above snippet.

**Expected behavior:** We either see something like the following trace, where the failing message handler does not prevent further execution
```
[ghi] success ▼
  [] pre abc 
  [abc] <exception thrown while producing trace node message> ▼
    [] a message 
  [] post abc 
```

or we see something like the following, where the failing message handler is treated as if `abc` failed at the end
```
[ghi] fail: oh no ▼
  [] pre abc
  [abc] a message
```

or we see something like below, where the message handler failure is treated as a failure, but the trace node is still shown
```
[ghi] fail: oh no ▼
  [] pre abc
  [abc] <exception thrown while producing trace node message> ▼
    [] a message 
```

In either case, `"pre abc"` should not disappear.

**Actual behavior:** We see
```
[ghi] fail: oh no
  [abc] a message
```

### Versions

`4.0.0-nightly-2023-07-12`

Vtec234:
Closes #2362 using the first proposed solution. We take the stance that failures in the message-producing action should not impact control flow, so in case one happens we just print a dummy message and continue.

### DIFF
--- a/src/Lean/Util/Trace.lean
+++ b/src/Lean/Util/Trace.lean
@@ -193,7 +193,7 @@ def withTraceNode [MonadExcept ε m] [MonadLiftT BaseIO m] (cls : Name) (msg : E
     modifyTraces (oldTraces ++ ·)
     return (← MonadExcept.ofExcept res)
   let ref ← getRef
-  let mut m ← try msg res catch _ => m!"<exception thrown while producing trace node message>"
+  let mut m ← try msg res catch _ => pure m!"<exception thrown while producing trace node message>"
   if profiler.get opts || aboveThresh then
     m := m!"[{secs}s] {m}"
   addTraceNode oldTraces cls ref m collapsed


src/Lean/Util/Trace.lean
```diff
@@ -193,7 +193,9 @@ def withTraceNode [MonadExcept ε m] [MonadLiftT BaseIO m] (cls : Name) (msg : E
     modifyTraces (oldTraces ++ ·)
     return (← MonadExcept.ofExcept res)
   let ref ← getRef
-  let mut m ← msg res
+  let mut m := match ← observing <| msg res with
+    | .ok m => m
+    | .error _ => m!"<exception thrown while producing trace node message>"
```
gebner, line 198:
```suggestion
  let mut m ← try msg res catch _ => m!"<exception thrown while producing trace node message>"
```

Do you see an easy way to make the `<exception ...>` part interactive?  (I've often wished that I could hover over the 💣 emoji and see what went wrong.)

Vtec234: What kind of interaction do you have in mind? You can write your message action like `fun | .error e => m!"💣 {e.toMessageData}" | ...`, and this will be interactive if the error is. Do you mean something else?

gebner: Oh, I meant interactive in the sense that it only shows the emoji by default and you need to hover to see the exception message.

But maybe we should just always print the exception.

Vtec234: Ah, we could probably do that with #2064 :)

PR #2366, opened 2023-07-28
chore: revert #2317

kim-em:
This requires too many downstream changes: the majority of rewrites at multiple locations now fail.

See https://github.com/leanprover/std4/pull/202 for the changes required in std if we keep #2317 as is, and https://github.com/leanprover-community/mathlib4/pull/6201 for a tiny fraction of those required in mathlib4.

I propose simply reverting the PR for now, and then restoring a re-implementation that just resets the metavariables for each location.


Kha (approved):
Thanks, a semantic change was likely unintended

alexjbest:
I also agree this was an unfortunate (unintended) breakage and should be reverted and a refined implementation can come later.

PR #2369, opened 2023-07-29
perf: faster `replace "\r\n" "\n"`

digama0:
This is an issue which was flagged [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/time.20to.20execute.20.60lake.20exe.20cache.60/near/379423687). It was reported in the context of `lake exe cache`, but `lake` does the same computation and they may as well share this code.

Here's a synthetic benchmark:
```lean
#eval show IO Unit from do
  let mut s ← IO.FS.readFile "src/lake/Lake/Build/Trace.lean"
  -- s := s.replace "\n" "\r\n"
  for _ in [0:7] do
    s := s ++ s
  let r₁ ← timeit "new" (fun w => (pure (replaceCRLFToLF s) : IO _) w)
  println! r₁.length
  let r₂ ← timeit "old" (fun w => (pure (s.replace "\r\n" "\n") : IO _) w)
  println! r₂.length
```
The initial bit generates a 1MB file containing either `\n` newlines or `\r\n` depending on the commented out bit (assuming the host uses `\n` newlines). Results:
```lean
-- with \n line endings
new 187ms
1144576
old 779ms
1144576
```
```lean
-- with \r\n line endings
new 242ms
1144576
old 748ms
1144576
```

---

This is my first Lake PR performed via the lean4 repo. LMK if there are additional procedural matters to attend to.

### DIFF
--- a/src/lake/Lake/Build/Trace.lean
+++ b/src/lake/Lake/Build/Trace.lean
@@ -137,10 +137,27 @@ instance : ComputeHash FilePath IO := ⟨computeFileHash⟩
 structure TextFilePath where
   path : FilePath
 
+/-- This is the same as `String.replace text "\r\n" "\n"`, but more efficient. -/
+partial def crlf2lf (text : String) : String :=
+  go "" 0 0
+where
+  go (acc : String) (accStop pos : String.Pos) : String :=
+    if h : text.atEnd pos then
+      -- note: if accStop = 0 then acc is empty
+      if accStop = 0 then text else acc ++ text.extract accStop pos
+    else
+      let c := text.get' pos h
+      let pos' := text.next' pos h
+      if c == '\r' && text.get pos' == '\n' then
+        let acc := acc ++ text.extract accStop pos
+        go acc pos' (text.next pos')
+      else
+        go acc accStop pos'
+
 instance : ComputeHash TextFilePath IO where
   computeHash file := do
     let text ← IO.FS.readFile file.path
-    let text := text.replace "\r\n" "\n"
+    let text := crlf2lf text
     return Hash.ofString text
 
 instance [ComputeHash α m] [Monad m] : ComputeHash (Array α) m where


tydeu:
@digama0: My only suggestion would be to name the function `crlf2lf`. Otherwise, LGTM!

@Kha: Would you like me to copy you on Lake-only PRs? Or should I  usually just review & merge them myself?

Kha:
> Or should I usually just review & merge them myself?

Yes, go ahead.

My only suggestion would be to use the `perf:` commit message prefix :smile:

src/lake/Lake/Build/Trace.lean
```diff
@@ -137,10 +137,38 @@ instance : ComputeHash FilePath IO := ⟨computeFileHash⟩
 structure TextFilePath where
   path : FilePath
 
+/-- This is the same as `String.replace text "\r\n" "\n"`, but more efficient. -/
+partial def replaceCRLFToLF (text : String) : String :=
+  start 0
+where
+  start (pos : String.Pos) : String :=
```
Kha, line 144: Just wondering, did you measure the impact of this duplication? I assume the final append is insignificant

digama0: It's less about the append than the fact that a new allocation is needed. The case where there are no `\r\n` in the file is common (e.g. all files on linux/mac) and so it is important to have a fast path for this case.

Kha: The empty string allocation by `extract`?

digama0: The effect is small (about 5-10%) but measurable. Adding `if accStop = 0 then text else acc ++ ...` into the base case of `go` avoids the duplication but is about as fast as just calling `go "" 0 0` directly and lacks the 5% boost mentioned. (Note: benchmarks were run on the interpreted code, I did not try the compiled version.)

Kha: It doesn't really make sense to benchmark the interpreted code if the goal is to have the code be compiled. Even just the overhead of passing around the two additional arguments is likely incomparable between interpreted and compiled.

digama0: > The empty string allocation by extract?

no, in the case where there is nothing to replace in the string the returned result is effectively `"" ++ text.extract 0 text.endPos` which is a full copy of `text`. I checked, it does not produce the same (pointer-equal) string, and I don't see a fast path for this in either `lean_string_utf8_extract` or `lean_string_append` so I think it might even make two copies of `text`.

digama0: Okay, I did the compiled benchmark, and it looks even more competitive, 9x faster rather than 4x:
```
new 46.9ms
41590784
old 403ms
41590784
```
Calling `go "" 0 0` instead of `start 0` results in `101ms` instead so the difference is much more easily measurable here. Adding the `if` statement to `go` makes it go back down to about `46ms` +- 2ms so I'll use that version instead (which avoids having to copy the main logic).

JasonGross:
A random thought: this could be a fast-path added to `String.replace` instead at almost no cost, right?  Assuming the primary bottleneck is string allocation, most of this algorithm could be reused in `String.replace` (if it does not already use this algorithm), and there could be an extra fast-path that avoids allocating to append the replacement whenever the replacement is a suffix of the search string.

Or, even more generally and faster, there could be a three-place extract-append method that takes three strings, start and end positions in all of them, and concatenates the results, all with a single allocation.

Or perhaps even faster, you can do the entire method with a single allocation: first iterate over the string to determine the length of the new string; then allocate a buffer, then iterate again to copy.

Alternatively, with just one iteration but more allocation, you could allocate a buffer that is the length of the entire string (+ overhead that would come from it just being repetitions of the search string, whenever the replacement text is longer) the first time you find the search text, always copy into that buffer, and once you're done, allocate a new buffer of the right length.  (Or if you use null-terminated strings you can keep the longer buffer and avoid the extra allocation.)

tydeu:
@JasonGross I agree that this could (and ideally should) be optimized more generally in core, but I think that should be a separate PR.

digama0:
> A random thought: this could be a fast-path added to `String.replace` instead at almost no cost, right? Assuming the primary bottleneck is string allocation, most of this algorithm could be reused in `String.replace` (if it does not already use this algorithm), and there could be an extra fast-path that avoids allocating to append the replacement whenever the replacement is a suffix of the search string.

My profiling results [above](https://github.com/leanprover/lean4/pull/2369#discussion_r1278294392) indicate that the string allocation is not the *primary* bottleneck: there is a 4x performance boost from inlining and specializing `substrEq "\r\n"`, and a further 2x boost from the string allocation. My guess is that the boost is much less than 2x in the general case of `replace` because `substrEq` is so much heavier. But certainly it could help.

> Or, even more generally and faster, there could be a three-place extract-append method that takes three strings, start and end positions in all of them, and concatenates the results, all with a single allocation.

I think you want something more like `s1.extract 0 a ++ s2.extract b c` (where `s1` is unshared) for the general case in which `s1` can be updated in place. If you aren't taking a prefix of `s1` then you have to do an in-place memcpy and I'm not sure it's worth it.

> Or perhaps even faster, you can do the entire method with a single allocation: first iterate over the string to determine the length of the new string; then allocate a buffer, then iterate again to copy.

Note that there is currently no way to request a string buffer with a specified length, unlike with arrays.

But also, I suspect this won't actually be all that great because of the overhead of the iteration itself. (It would be great if this could be an iteration over the bytes instead of going characters at a time, since there is no need to do UTF-8 decoding to recognize `\r\n` sequences in the text. If you could do that and also use `USize` indexes then I think this would optimize to an actual C loop and maybe if you are lucky the compiler might even turn it into `memchr` and get those sweet auto-vectorization benefits.)

PR #2377, opened 2023-08-02
chore: Lake touchups

tydeu:
Better acclimates Lake to the Lean 4 repo and applies other quick fixes and tweaks.


tydeu:
@Kha Question: In general, should I squash a set of Lake commits into single monolith, or it fine to rebase and merge them as is to keep commits more focused?

Kha:
This kind of history looks perfectly fine to me, I don't see a reason to squash it

Kha:
Ah, something that came to my mind too late: I do make use of prefixes like `server: ` for changes specific to a (large) subcomponent if the component is not otherwise apparent from the message

tydeu:
@Kha Good point! I will make sure to add a `lake:`  prefix to such commits in the future. Unfortunately, in addition to these all the old merged-in Lake commits also have no prefix and thus could be a bit confusing as well. On flip side, it does mean these don't stand out too much. 😆

PR #2378, opened 2023-08-02
chore: correct doc-string for elabTerm

kim-em:


tydeu:
Sorry for the back-and-forth. Upon further reflection, I am not 100% sure (based on the code and docstring) if `implicitLambda` is being used to toggle the feature in the subterms or in the top-level syntax. Thus, I will leave this to others with better knowledge of this code.

semorrison:
This option is consumed in [`Lean/Elab/Term.lean`](https://github.com/leanprover/lean4/blob/master/src/Lean/Elab/Term.lean#L1367) in this line:
```
let useImplicitResult ← if implicitLambda && (← read).implicitLambda then useImplicitLambda stx expectedType? else pure .no
```

I think this matches the new docstring -- if `Context.implicitLambda == false`, then the `implicitLambda` argument is irrelevant. Otherwise, if it is true then we use implicit lambdas, and if it is false we don't.

tydeu:
@semorrison Yeah,  that is my reading as well. I am just not 100% confident in it. so I will let Sebastian or Leo merge.

tydeu:
Oh, and I see this issue was discussed on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/disabling.20implicit.20lambda).

PR #2381, opened 2023-08-03
test: reverse FFI from C with Lake

Kha:
This a simple example of how a Lake library (`lib/`) can be used from a foreign language and build system (`main.c` and `Makefile`).


Kha:
Wow, only one revision to get it working on all platforms? This almost felt too easy.

PR #2384, opened 2023-08-04
test: lake: make 116 deterministic & related tweaks

tydeu:
I also made the glob use only take `test.sh` scripts from `examples` and `test`, and I removed a leftover `-E leanlaketest_git` in the CI I missed in 86f1131.

PR #2385, opened 2023-08-04
feat: `lake.lock` file for builds

tydeu:
A feature frequently requested on Zulip (c.f. [1](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/how.20to.20run.20.60lake.20exe.20cache.20get.60.20while.20in.20VSCode/near/341935974), [2](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/infoview.20failing.20on.20codespaces/near/364484400), [3](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Multiple.20tabs.20in.20VS.20Code.2C.20shared.20work.3F/near/381226982)).

### DIFF
--- a/src/lake/Lake/Build/Monad.lean
+++ b/src/lake/Lake/Build/Monad.lean
@@ -1,7 +1,7 @@
 /-
 Copyright (c) 2021 Mac Malone. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Mac Malone
+Authors: Mac Malone, Gabriel Ebner, Sebastian Ullrich
 -/
 import Lake.Config.Monad
 import Lake.Build.Context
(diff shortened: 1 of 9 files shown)


tydeu:
I am currently stumped. I have no clue why this or #2384 is failing. The do not fail on my machine  (in MSY?S2 or WSL), so it likely has to be some difference between how the CI and my local machine handle background processes.

digama0:
You made an ABI breaking change to a type in Init, you most likely need to set `interpreter.prefer_native` in `stage0/src/stdflib_flags.h`  and possibly do a stage0 update

tydeu:
@digama0 That does not appear to be the bug. Looking, at the logs, lake is working fine (the lock file is generated and waiting on properly). The test is simply not terminating once the script finishes for some reason.

src/lake/Lake/Build/Monad.lean
```diff
@@ -47,10 +47,46 @@ If a cycle is encountered, log it and then fail.
...
+@[inline] partial def busyAcquireLockFile (lockFile : FilePath) : IO PUnit := do
+  busyLoop true
+where
+  busyLoop firstTime :=
+    try
+      -- Remark: fail if already exists
+      -- (not part of POSIX, but supported on all our platforms)
+      createParentDirs lockFile
+      let h ← IO.FS.Handle.mk lockFile .writeNew
+      h.putStrLn <| toString <| ← IO.Process.getPID
```
Kha, line 63: So the PID doesn't do anything yet?

Kha (commented):
This looks good but I'd also like to hear about potential future plans. Do we believe we'll want to end up with granular locking, e.g. per-module? What would that look like?

Kha:
> The test is simply not terminating once the script finishes for some reason.

Are we sure it's the end of the script and not `fg` that's blocking? Could the background process for some reason interfere?

src/lake/Lake/Build/Monad.lean
tydeu: Not yet, Lake does not use it. However, it useful for users to know what process is holding the lock (or was, if `lake` died ungracefully).

tydeu:
@Kha 

> Are we sure it's the end of the script and not `fg` that's blocking? Could the background process for some reason interfere?

I am not yet, but that is what I plan on testing today. However, my reason for believing it is the end of the script is that Lake background job does successfully resume, acquire the lock and begin building (as can be seen from the `Building Test` output).

tydeu:
@Kha The latest run confirms that the problem is on the script finish, not the `fg` (and `pkill` does not work).

tydeu:
@Kha Update: It appears the problem is the process substitution. CTest is waiting on it finish before completing the test.

src/Init/System/IO.lean
```diff
@@ -220,7 +220,7 @@ local macro "nonempty_list" : tactic =>
 @[extern "lean_io_get_num_heartbeats"] opaque getNumHeartbeats : BaseIO Nat
 
 inductive FS.Mode where
-  | read | write | readWrite | append
+  | read | write | writeNew | readWrite | append
```
Kha, line 223: This needs a clarifying docstring, in which case we probably should clarify all of them.

src/Init/System/IO.lean
```diff
@@ -517,6 +517,10 @@ partial def FS.removeDirAll (p : FilePath) : IO Unit := do
   removeDir p
 
 namespace Process
+
+/-- Get the process ID of the current process. -/
```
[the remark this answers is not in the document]
tydeu: #2400

PR #2386, opened 2023-08-04
chore: make Lean build shell configurable

tydeu:
This is very useful on Windows were VSCode automation (e.g., CMake Tools) uses a CMD environment where bash is from WSL rather than MSYS2.

### DIFF
--- a/src/stdlib.make.in
+++ b/src/stdlib.make.in
@@ -1,4 +1,8 @@
-SHELL := /usr/bin/env bash -euo pipefail
+# LEAN_BASH: Makes the Bash shell used by the Lean build configurable.
+# On Windows, when CMake/Make is spawned directly (e.g., VSCode's CMake Tools),
+# it lacks a proper shell environment, so we need to manually point it to Bash.
+LEAN_BASH ?= /usr/bin/env bash
+SHELL := $(LEAN_BASH) -eo pipefail
 
 # any absolute path to the stdlib breaks the Makefile
 export LEAN_PATH=
--- a/stage0/src/stdlib.make.in
+++ b/stage0/src/stdlib.make.in
@@ -1,4 +1,8 @@
-SHELL := /usr/bin/env bash -euo pipefail
+# LEAN_BASH: Makes the Bash shell used by the Lean build configurable.
+# On Windows, when CMake/Make is spawned directly (e.g., VSCode's CMake Tools),
+# it lacks a proper shell environment, so we need to manually point it to Bash.
+LEAN_BASH ?= /usr/bin/env bash
+SHELL := $(LEAN_BASH) -eo pipefail
 
 # any absolute path to the stdlib breaks the Makefile
 export LEAN_PATH=


tydeu:
@Kha This updates `stdlib.make.in` in both `src` and `stage0/src`. I know the `stage0` update would be done with `chore: update stage0` commit, but that felt too heavy duty for this change. Should even small changes like this follow that pattern, or is update both like this fine?

Kha:
The change itself is fine but aren't there tons of other places where we depend on `env bash`?

tydeu:
@Kha `env bash` works great *within* a MSYS2 context. The problem is that this make file is not run within one due to oddities of CMake automation in VSCode. So once the make file is set to use the MSYS2 shell via `LEAN_BUILD_SHELL`, things work normally. 

It is also possible that I have simply not found the magic VSCode setting for CMake Tools to use MSYS2, but I have tried the usual settings (e.g., `terminal.integrated.automationProfile.windows`), and they did not work.

Kha:
It sounds like there should be some form of MSYS2 support? https://github.com/microsoft/vscode-cmake-tools/pull/3056

By the way, have you re-evaluated using WSL for development on Windows? Considering the terrible performance of CMake on Windows I assume that would be quite a bit more comfortable and that it doesn't really matter for local development which platform the result is for - you can still test the other platform in another terminal.

tydeu:
@Kha 

> It sounds like there should be some form of MSYS2 support? [microsoft/vscode-cmake-tools#3056](https://github.com/microsoft/vscode-cmake-tools/pull/3056)

Yes, however that only effects the compiler toolchain used, it does not seem to configure the environment tools like `make` operate in.

> By the way, have you re-evaluated using WSL for development on Windows? Considering the terrible performance of CMake on Windows I assume that would be quite a bit more comfortable 

After having spent some time configuring the proper VSCode extensions for CMake, the build workflow has actually become quite comfortable and the builds are actually not that slow. Also, I am still rather convinced that is good to have at least one core developer who is regularly using standard Windows to accumulate experience to help guide any external users / potential contributors who are also using Windows.

> that it doesn't really matter for local development which platform the result is for - you can still test the other platform in another terminal.

I am not sure I understand this assertion. If I want to test Windows and I am using WSL for development, I would still have to rebuild the Lean core for Windows before testing it. So, if I am rebuilding infrequently during development, using WSL and testing on Windows would not provide much benefit as far as I can tell.

tydeu:
@Kha 
> Do you think adding general documentation on how to set up CMake Tools would be worthwhile as well?

Yep, and I think generally enriching the Development section of the manual would be good for contributors. However, I think that is best covered in a separate PR, especially since we may wish to debug some of the current foibles there such as why the CMake configuration arguments due are not carried over from the root project to subprojects on Windows.

tydeu:
@Kha Sorry, I ended having to remove `-u`  because it broke the MSYS2 built-in profile script. Also, `-u` doesn't really make much sense for a makefile shell anyway (as it will not catch unbound makefile variables).

PR #2390, opened 2023-08-07
fix: don't try to generate `below` for nested predicates.

### LINKED ISSUE #2389: Error generating `below` for nested predicates
### Prerequisites

### Description

First reported [here](https://proofassistants.stackexchange.com/questions/2045/unexpected-inductive-type-occurrence-error-for-recursive-inductive-type). After reducing the problem to a MWE, it turns out any nested predicate triggers an error `only trivial inductive applications supported in premises` when trying to generate the `below` auxiliary lemma.

### Steps to Reproduce

```lean
inductive Foo : Prop → Prop

inductive Bar : Prop
  | bar : Foo Bar → Bar
/-only trivial inductive applications supported in premises:
  Foo Bar-/
```

**Expected behavior:** Should not trigger an error

**Actual behavior:** The `mkBelow` function tries to generate the `below` helper function, even though it's not able to.

### Versions

4.0.0-nightly-2023-07-12

arthur-adjedj:
Closes #2389

### DIFF
--- a/src/Lean/Meta/IndPredBelow.lean
+++ b/src/Lean/Meta/IndPredBelow.lean
@@ -571,7 +571,7 @@ def findBelowIdx (xs : Array Expr) (motive : Expr) : MetaM $ Option (Expr × Nat
     else pure none
   | _, _ => pure none
 
-/--Generates the auxiliarry lemmas `below` and `brecOn` for a recursive inductive predicate.
+/-- Generates the auxiliary lemmas `below` and `brecOn` for a recursive inductive predicate.
 The current generator doesn't support nested predicates, but pattern-matching on them still works
 thanks to well-founded recursion. -/
 def mkBelow (declName : Name) : MetaM Unit := do
--- a/tests/lean/run/2389.lean
+++ b/tests/lean/run/2389.lean
@@ -1,7 +1,7 @@
 /-!
-# Test verifying that nested predicates don't trigger the generation of `below` lemmas
-Since the case of nested predicates is not currently handled by `mkBelow`,
-trying to generate `Bar.below` triggers an error upon defining the inductive type.
+# Verify that nested predicates don't trigger the generation of `below` lemmas
+Since the case of nested predicates is not currently handled by `mkBelow` (in `src/Lean/Meta/IndPredBelow.lean`),
+trying to generate `OnlyZeros.below` triggers an error upon defining the inductive type.
 -/
 
 inductive Forall (P : α → Prop) : List α → Prop
@@ -12,21 +12,21 @@ inductive Tree : Type :=
   | leaf : Nat → Tree
   | node : List Tree → Tree
 
-set_option trace.Meta.IndPredBelow true
+set_option trace.Meta.IndPredBelow true in
 
-/--Despite not having `.below` and `.brecOn`,
-the type is still usable thanks to well-founded recursion-/
+/-- Despite not having `.below` and `.brecOn`,
+the type is still usable thanks to well-founded recursion. -/
 inductive OnlyZeros : Tree → Prop :=
   | leaf : OnlyZeros (.leaf 0)
   | node (l : List Tree): Forall OnlyZeros l → OnlyZeros (.node l)
 
-/- Fixpoint equivalent of `OnlyZeros`-/
+/-- Equivalent definition of `OnlyZeros`, defined by a function instead of an inductive type. -/
 def onlyZeros : Tree → Prop
-  | .leaf n => n=0
+  | .leaf n => n = 0
   | .node [] => True
   | .node (x::s) => onlyZeros x ∧ onlyZeros (.node s)
 
-/- Pattern-matching on `OnlyZeros` works despite `below` and `brecOn` not being generated-/
+/-- Pattern-matching on `OnlyZeros` works despite `below` and `brecOn` not being generated. -/
 def toFixPoint : OnlyZeros t → onlyZeros t
   | .leaf => rfl
   | .node [] _ => True.intro


src/Lean/Meta/IndPredBelow.lean
```diff
@@ -574,7 +574,7 @@ def findBelowIdx (xs : Array Expr) (motive : Expr) : MetaM $ Option (Expr × Nat
 def mkBelow (declName : Name) : MetaM Unit := do
   if (← isInductivePredicate declName) then
     let x ← getConstInfoInduct declName
-    if x.isRec then
+    if x.isRec && !x.isNested then
```
semorrison, line 577: Could you please add a comment here that explains this condition?

Ideally this comment will make clear why we want to allow some inductives for which we can not generate `.below`.

It may be best to write this comment in the form of a doc-string on `mkBelow`, explaining when it is and isn't meant to work.

arthur-adjedj: I am not absolutely certain that it's not possible to generate `.below` for nested predicates, but the current implementation doesn't support them, and the modifications necessary to support them looks to be non-trivial. These nested predicates are still sound and accepted at a kernel level. As for why we want to allow them, some users have stumbled upon the error [in the past](https://proofassistants.stackexchange.com/questions/2045/unexpected-inductive-type-occurrence-error-for-recursive-inductive-type). Also, note that Lean already "accepts" nested predicates, in the sense that nested predicates still get silently added to the environment despite Lean not being able to generate their `.below` lemma, this fix simply avoids triggering the error in the first place. I'll nonetheless add documentation, and make a more complete/relevant test.

tests/lean/run/2389.lean
```diff
@@ -0,0 +1,10 @@
+/-!
+# Test verifying that nested predicates don't trigger the generation of `below` lemmas
+Since the case of nested predicates is not currently handled by `mkBelow`,
+trying to generate `Bar.below` triggers an error upon defining the inductive type.
+-/
+
+inductive Foo : Prop → Prop
+
+inductive Bar : Prop
+  | bar : Foo Bar → Bar
```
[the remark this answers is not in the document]
arthur-adjedj: Done. The type is still usable without `below` thanks to well-founded recursion, so there isn't much that's missing because of this.

src/Lean/Meta/IndPredBelow.lean
arthur-adjedj: Done.

src/Lean/Meta/IndPredBelow.lean
```diff
@@ -571,10 +571,13 @@ def findBelowIdx (xs : Array Expr) (motive : Expr) : MetaM $ Option (Expr × Nat
     else pure none
   | _, _ => pure none
 
+/--Generates the auxiliarry lemmas `below` and `brecOn` for a recursive inductive predicate.
```
semorrison, line 574:
```suggestion
/-- Generates the auxiliary lemmas `below` and `brecOn` for a recursive inductive predicate.
```

tests/lean/run/2389.lean
```diff
@@ -0,0 +1,35 @@
...
+inductive Forall (P : α → Prop) : List α → Prop
+  | nil : Forall P []
+  | cons : {x : α} → P x → Forall P l → Forall P (x::l)
+
+inductive Tree : Type :=
+  | leaf : Nat → Tree
+  | node : List Tree → Tree
+
+set_option trace.Meta.IndPredBelow true
+
```
semorrison, line 16:
```suggestion
set_option trace.Meta.IndPredBelow true in
```

tests/lean/run/2389.lean
```diff
@@ -0,0 +1,35 @@
...
+/--Despite not having `.below` and `.brecOn`,
+the type is still usable thanks to well-founded recursion-/
```
semorrison, line 18:
```suggestion
/-- Despite not having `.below` and `.brecOn`,
the type is still usable thanks to well-founded recursion. -/
```

tests/lean/run/2389.lean
```diff
@@ -0,0 +1,35 @@
...
+inductive OnlyZeros : Tree → Prop :=
+  | leaf : OnlyZeros (.leaf 0)
+  | node (l : List Tree): Forall OnlyZeros l → OnlyZeros (.node l)
+
+/- Fixpoint equivalent of `OnlyZeros`-/
```
semorrison, line 23:
```suggestion
/-- Fixpoint equivalent of `OnlyZeros`. -/
```
Could you rephrase this doc-string? I don't understand it.

tests/lean/run/2389.lean
```diff
@@ -0,0 +1,35 @@
...
+def onlyZeros : Tree → Prop
+  | .leaf n => n=0
```
semorrison, line 25:
```suggestion
  | .leaf n => n = 0
```

tests/lean/run/2389.lean
```diff
@@ -0,0 +1,35 @@
...
+  | .node [] => True
+  | .node (x::s) => onlyZeros x ∧ onlyZeros (.node s)
+
+/- Pattern-matching on `OnlyZeros` works despite `below` and `brecOn` not being generated-/
```
semorrison, line 29:
```suggestion
/-- Pattern-matching on `OnlyZeros` works despite `below` and `brecOn` not being generated. -/
```

tests/lean/run/2389.lean
semorrison: (This makes the expected output easier to understand, as it makes clear where the trace message should be coming from.)

tests/lean/run/2389.lean
```diff
@@ -0,0 +1,35 @@
+/-!
+# Test verifying that nested predicates don't trigger the generation of `below` lemmas
```
semorrison, line 2:
```suggestion
# Verify that nested predicates don't trigger the generation of `below` lemmas

```

tests/lean/run/2389.lean
```diff
@@ -0,0 +1,35 @@
...
+Since the case of nested predicates is not currently handled by `mkBelow`,
+trying to generate `Bar.below` triggers an error upon defining the inductive type.
```
semorrison, line 4: This needs updating, no `Bar` anymore.

semorrison, line 4: Also, reference to `mkBelow` should name the file `src/Lean/Meta/IndPredBelow.lean` to help someone find the relevant code.

src/Lean/Meta/IndPredBelow.lean
semorrison: Back in the bug report issue could you:
1. Quote the actual error message that occurs, as a breadcrumb for the future.
2. Link to that `proofassistants` question as explanation of why this is needed.

arthur-adjedj: Sorry for the style problems, and thank you for spending time on this pull request. I've added the stack-exchange post and the error message in the issue, and fixed the code according to your suggestions.

PR #2391, opened 2023-08-08
experiment: make releases for PRs, too

kim-em:
If we want to try compiling Mathlib against PRs, we need to generate a release for the `lean-toolchain` in Mathlib.


semorrison:
More complicated than initially hoped, off to try again as a PR to my fork so I can do more dangerous things. :-)

PR #2392, opened 2023-08-08
test: update `parser` benchmark, add to speedcenter suite

Kha:
@leodemoura We had this benchmark in the Beans paper but it never made it into the speedcenter. It depends on the contents of `Init.Prelude` but I think that file is quite stable enough.


Kha:
!bench

Issue #1971, opened 2022-12-19
RFC: Move the error message earlier on incomplete-tactic-followed-by-deindentation


hrmacbeth:
This is intended as a proposal for discussion.  When there is an incomplete tactic call, Lean considers the error as occurring in the first place afterwards where new text appears to make it malformed.  This is done even in examples like the following,  when the incomplete tactic call is followed by de-indentation, so the user's intention was very likely that the new (de-indented) text be unrelated to that tactic call.
````lean
theorem lem1 : True := by
  exact

theorem lem2 : True := sorry -- red underline ("expected term") on the word `theorem`
````

This is the same behaviour as in Lean 3, but the situation arose much more rarely there because usually there was a `begin`/`end` block or a comma to contain the error to a single line.

Proposal: if there is text following an incomplete tactic call which is the same or a lower indentation level, then Lean could treat the error as being located on the `exact` itself or in the space after it.  (So in this case we are assuming that that later text wasn't supposed to be part of the tactic call.)

This proposal comes from [Mario](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/premature.20.22goals.20accomplished.22/near/316465519) in a Zulip discussion; he [notes](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/premature.20.22goals.20accomplished.22/near/316657831) there that this would seem to require a change within the core parser infrastructure / somewhere around the `withPosition` parser combinator.

**Comment by PatrickMassot** [comment]
This also goes wrong when you go in the other direction, supplying too many arguments to tactics. The simplest case could arise from mixing up `exact` and `assumption`:
```lean
example (P : Prop) (h : P) : P := by
  assumption P
```
which gives the error `expected command` since the proof is over after `assumption` and Lean tries to process whatever comes after this proof. 

A nasty variation if you add an extraneous argument in a calc block:
```lean
example : 1 + 1 = 2 := by
  calc 1 + 1 = 1 + 1 := by rfl
  _ = 2 := by rfl [h]

example (P : Prop) (h : P) : P := by
  exact h
```
Here the error comes on line 5 and is `expected :=`.

PatrickMassot:
Note that in my example everything would be much clearer if we simply went back to delimiting tactic proofs by `begin`/`end`.

kbuzzard:
[...]

PR #2393, opened 2023-08-08
Adjust message range on unexpected token error

Kha:
This PR proposes a quite simple and general solution to #1971: when the parser hits an unexpected token, it will include all preceding whitespace in the error message range as the expected token could be inserted at any of these places to fix the error. I'll let the test file speak for itself: 
image

---

Implementation notes: In order to implement this strategy, I first had to teach the parser to report ranges at all - so far we only reported single positions and left it to the editor to including what it thinks is the next token, which is not ideal and completely breaks when we want to include preceding whitespace like this. For the purposes of this PR I only changed the token error messages to report ranges, which they do via a new helper function `mkUnexpectedTokenError`. True to its name, I also used this refactoring opportunity to actually report what the unexpected token is; see changed test outputs below. Because of this preparatory work, I recommend looking at the changes commit by commit.

### DIFF
--- a/src/Lean/Parser/Module.lean
+++ b/src/Lean/Parser/Module.lean
@@ -36,12 +36,12 @@ private def mkErrorMessage (c : InputContext) (s : ParserState) (e : Parser.Erro
   let mut pos := s.pos
   let mut endPos? := none
   let mut e := e
-  if let some tk := e.unexpectedTk? then
+  unless e.unexpectedTk.isMissing do
     -- calculate error parts too costly to do eagerly
-    if let some r := tk.getRange? then
+    if let some r := e.unexpectedTk.getRange? then
       pos := r.start
       endPos? := some r.stop
-    let unexpected := match tk with
+    let unexpected := match e.unexpectedTk with
       | .ident .. => "unexpected identifier"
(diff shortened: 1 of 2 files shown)


Kha:
!bench

Kha:
!bench

leodemoura:
> Interesting, that's a little more overhead than I would have assumed

I agree. Could you please investigate?

Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
I don't think I will get closer than that without addressing the underlying problem that we seem to apparently too often from token mismatches, apparently from code like `"tk" ... <|> ...`

Kha:
I believe this is ready to merge but as it is a subjective change, it would be good to hear about some feedback: can the original problem be considered solved with this? Is the space before `def f4` consistent or "weird"? Please give it a try e.g. via the PR release when it is released.

kbuzzard:
My instinct is that this change makes things less confusing for beginners.

Kha:
Let's give it a try in practice

PR #2397, opened 2023-08-08
test: basic tests exercising rw

kim-em:


tests/lean/run/rewrite.lean
```diff
@@ -0,0 +1,64 @@
...
+
+-- These are often solved by the implicit `rfl` at the end of `rw`.
+example : f 2 = f 3 := by
+  rw [s]
+
+-- We can rewrite at multiple locations,
+-- specializing the arguments differently at each location.
+-- This behaviour was accidentally broken by
+-- https://github.com/leanprover/lean4/pull/2317
+-- and requiring a revert.
```
tydeu, line 43: Stylistically, should probably be a `/- ... -/` comment. It's a bit to long for line comments.

tydeu:
Also, the commit should probably have the `test:` prefix instead of `chore:`. Otherwise, LGTM, but I am not a  expert on `rw`.

tests/lean/run/rewrite.lean
```diff
@@ -0,0 +1,64 @@
...
+example : [f 1, f 2, f 1] = [f 0, f 0, f 0] := by
+  rw [t]
+  rw [t 2]
+
+-- Rewriting by lemmas with undetermined arguments creates additional goals.
+example : f 2 = f 3 := by
+  rewrite [s]
+  rfl
+
+-- These are often solved by the implicit `rfl` at the end of `rw`.
```
Kha, line 35: Potential additional test case: the `rfl` should only unfold `reducible`

Kha (commented):
Potential additional test cases:
* sequence of rewrites
* empty rewrite (i.e. cheap rfl)
* rewriting by `Iff`
* rewriting touches main goal only
* rewriting with config
* should we have negative test cases like "motive is not type-correct"? Not sure.

semorrison:
I've added these (but not one for "motive not type-correct").

I would like to expose the `occs` parameter through the config in a separate PR, and I will add more tests along with that.

Kha (approved):

PR #2400, opened 2023-08-09
doc: `IO.Process.getPID` tweak + `IO.FS.Mode`

tydeu:
Follow-up to #2385 addressing @Kha's post-merge documentation notes.

### DIFF
--- a/src/Init/System/IO.lean
+++ b/src/Init/System/IO.lean
@@ -1,7 +1,7 @@
 /-
 Copyright (c) 2017 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Luke Nelson, Jared Roesch, Leonardo de Moura, Sebastian Ullrich
+Authors: Luke Nelson, Jared Roesch, Leonardo de Moura, Sebastian Ullrich, Mac Malone
 -/
 prelude
 import Init.Control.EState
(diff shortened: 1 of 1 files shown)


Kha:
@tydeu Thank you very much for the elaborate documentation! My only question is whether we want to describe platform details (considering POSIX as a platform as well) to this degree or whether the IO docs to present an abstract, platform-independent interface (as much as reasonable and possible). Did you consult with the documentation of any existing stdlib on the style?

tydeu:
@Kha 

> whether we want to describe platform details (considering POSIX as a platform as well) to this degree or whether the IO docs should present an abstract, platform-independent interface (as much as reasonable and possible)

My belief is that it should be both. There should be a abstract description and then details of the specific implementation, which is what I tried to aim for in these docs.

> Did you consult with the documentation of any existing stdlib on the style?

I went more with what I would like to see in standard library docs. Some languages' standard library documentation is quite lacking. For example, Haskell has none for its [`openFile`/`IOMode`](https://hackage.haskell.org/package/base-4.18.0.0/docs/System-IO.html#v:openFile), Conversely, Ruby's [`File`](https://ruby-doc.org/3.2.2/File.html) and Node.js' [`fs.open`](https://nodejs.org/api/fs.html#fsopenpath-flags-mode-callback) documentation are quite detailed and contain similar descriptions of platform-specific [implementation details](https://nodejs.org/api/fs.html#file-system-flags). Python's [`open`](https://docs.python.org/3/library/functions.html#open) docs are more platform-independent but still very detailed about the implementation. And, like Ruby, Rust's [`OpenOptions`](https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.open) interface closes mirrors the platform details. In fact, I would love to adapt some of these more detailed interfaces in Lean in a future PR.

Kha:
Great, I'm happy with that. Let's see if there are other opinions since there is no need to rush here but otherwise this should be good to go.

PR #2401, opened 2023-08-09
feat: lake: lean lib `extraDepTargets` & related tweaks

tydeu:

PR #2404, opened 2023-08-10
fix: lake: lock test timeout + README typos

### LINKED ISSUE #2403: `leanlaketest_lock` can time out
See https://github.com/leanprover/lean4/actions/runs/5815446364/job/15767018271

tydeu:
Fixes #2403 and @Kha's follow-up comment in #2401.

### DIFF
--- a/src/lake/test/lock/test.sh
+++ b/src/lake/test/lock/test.sh
@@ -12,6 +12,7 @@ fi
 ./clean.sh
 
 # Test lock file creation on build
+touch test.log
 $LAKE build Loop 2>&1 > test.log &
 grep -q "Building" < <($TAIL --pid=$$ -f test.log)
 kill %%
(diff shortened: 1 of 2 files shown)


src/lake/test/lock/test.sh
```diff
@@ -12,6 +12,7 @@ fi
 ./clean.sh
 
 # Test lock file creation on build
+touch test.log
 $LAKE build Loop 2>&1 > test.log &
 grep -q "Building" < <($TAIL --pid=$$ -f test.log)
```
Kha, line 17: Will `tail` not immediately exit if it sees the file from `touch` before the `lake` line opened it for the second time?

Kha: Ah, of course not with `-f`! Clever.

PR #2405, opened 2023-08-11
feat: `lake update <pkg>` & related tweaks

tydeu:
Closes leanprover/lake#148. 

~~A bare `lake update` now updates the manifest without updating packages (i.e., just downloads new ones). To update all packages,  `lake update @` is now used instead. `lake update deleted` (or `lake update @`) will remove a package whose `require` statement was deleted. (I choose this approach for now because it is the simplest to implement.)~~

**EDIT:** `lake update` now still upgrades all dependencies (and `lake update @` has been removed).

### DIFF
--- a/src/lake/Lake/CLI/Help.lean
+++ b/src/lake/Lake/CLI/Help.lean
@@ -29,7 +29,7 @@ COMMANDS:
   new <name> [<temp>]   create a Lean package in a new directory
   init <name> [<temp>]  create a Lean package in the current directory
   build [<targets>...]  build targets
-  update                update dependencies
+  update                update dependencies and save them to the manifest
   upload <tag>          upload build artifacts to a GitHub release
   clean                 remove build outputs
   script                manage and run workspace scripts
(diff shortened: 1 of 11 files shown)


Kha:
> (I choose this approach for now because it is the simplest to implement.)

CLI design and stability is important, we shouldn't change it just because it's convenient in the short term. If `lake update <pkg>` is the one we all agree on, let's add just this one for now and leave all other changes to a documented redesign?

tydeu:
@Kha 

>  (I choose this approach for now because it is the simplest to implement.)
> CLI design and stability is important,  we shouldn't change it just because it's convenient in the short term.

Sorry, to clarify, "the simplest to implement" was just about the `lake update deleted` which is not breaking change. Also, I would expect it to still work in the future. The part that I imagine changing have `lake update` also clear removed dependencies from the manifest. However, that is likely not desirable until Lake has a better story for leanprover/lake#149 because automatically removing conditional dependencies from the manifest for which the conditions are not set in a given `lake update` seems problematic at the moment.

> If `lake update <pkg>` is the one we all agree on, let's add just this one for now.

The other parts were intentional design decisions and by-products of this feature addition. If one can specify a single package to update, it provides consistency to be able to specify none, many, or all. It also mirrors, e.g., how [`bundle update`](https://bundler.io/v2.4/man/bundle-update.1.html#UPDATING-ALL-GEMS) and [`composer update`](https://getcomposer.org/doc/03-cli.md#update-u-upgrade) work, so I plan to keep that part of CLI stable regardless going forward.

tydeu:
@Kha I pared down the PR. Does it look good now to auto-merge?

PR #2407, opened 2023-08-11
test: compiling multiple files using the Frontend

kim-em:
This test is currently expected to fail. The RFC #2408 explains the full context. The PR https://github.com/leanprover/lean4/pull/2398 also contains a copy of this test, adapted to use the mechanism proposed there.


semorrison:
This test has been incorporated into the proposed fix at https://github.com/leanprover/lean4/pull/2423.

PR #2410, opened 2023-08-11
 LLVMBindings.lean & llvm.cpp updates for Linkage

hargoniX:
Peeled from https://github.com/leanprover/lean4/pull/2340.

This splits the easy to review review `LLVMBindings.lean` from the second patch that edits the CI, build system, and EmitLLVM.

### DIFF
--- a/src/Lean/Compiler/IR/LLVMBindings.lean
+++ b/src/Lean/Compiler/IR/LLVMBindings.lean
@@ -73,6 +73,12 @@ structure Value (ctx : Context) where
   private mk :: ptr : USize
 instance : Nonempty (Value ctx) := ⟨{ ptr := default }⟩
 
+/-- Check if the value is a null pointer. --/
+def Value.isNull (v : Value ctx) : Bool := v.ptr == 0
+
+@[extern "lean_llvm_get_value_name2"]
+opaque Value.getName {ctx : Context} (value : Value ctx) : BaseIO String
+
 @[extern "lean_llvm_initialize_target_info"]
 opaque llvmInitializeTargetInfo : BaseIO (Unit)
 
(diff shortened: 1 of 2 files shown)


src/util/shell.cpp
```diff
@@ -752,13 +752,6 @@ extern "C" LEAN_EXPORT int lean_main(int argc, char ** argv) {
             out.close();
         }
 
-        // target triple is only used by the LLVM backend. Save users
-        // a great deal of pain by erroring out if they misuse flags.
-        if (target_triple && !llvm_output) {
-            std::cerr << "ERROR: '--target' must be used with '--bc' to enable LLVM backend. Quitting code generation.\n";
-            return 1;
-        }
```
Kha, line 760: Can you explain/recap the motivation for this change? We're not forwarding `--target` to `lean` right now, are we?

hargoniX: We did notice that not doing this does trigger the warning in test runs, e.g. here: https://github.com/leanprover/lean4/actions/runs/5789866313/job/15691817010 so it does seem that it gets forwarded?

Kha: Yes because it's reintroduced on that branch: https://github.com/leanprover/lean4/compare/master...bollu:lean4:kha-llvm-stage2-build#diff-0d112c28beff241f98979d3807483fed193e94d8e0e9db9df7d2f5bd878a6cb8R42-R43

src/Lean/Compiler/IR/LLVMBindings.lean
```diff
@@ -316,6 +337,50 @@ def DLLStorageClass.export  : DLLStorageClass := { val := 2 }
...
+/-- Like Private, but linker removes. -/
+def Linkage.linkerPrivate : Linkage := { val := 15 }
+/-- Like LinkerPrivate, but is weak. -/
+def Linkage.linkerPrivateWeak : Linkage := { val := 16 }
+
+
+@[extern "lean_llvm_set_linkage"]
+opaque setLinkage {ctx : Context} (value : Value ctx) (linkage : Linkage) : BaseIO Unit
+
+
```
Kha, line 382: Please be mindful of extraneous whitespace

Kha:
We shouldn't need the stage 0 update here, should we? There is no semantic change.

hargoniX:
So we only ever want the update-stage0 in the PR that introduces the other LLVM changes then I guess?

Kha:
Only where in fact necessary, yes :)

PR #2416, opened 2023-08-12
LLVM CI

hargoniX:
The second half of #2340


src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -1544,7 +1564,19 @@ def emitLLVM (env : Environment) (modName : Name) (filepath : String) (tripleStr
...
+              | throw <| IO.Error.userError s!"ERROR: linked module must have global from runtime module: '{name}'"
+           LLVM.setLinkage global LLVM.Linkage.internal
+         for name in runtime_function_names do
+           let some fn ← LLVM.getNamedFunction emitLLVMCtx.llvmmodule name
+              | throw <| IO.Error.userError s!"ERROR: linked module must have function from runtime module: '{name}'"
+           LLVM.setLinkage fn LLVM.Linkage.internal
+
          optimizeLLVMModule emitLLVMCtx.llvmmodule
          LLVM.writeBitcodeToFile emitLLVMCtx.llvmmodule filepath
          let tripleStr := tripleStr?.getD (← LLVM.getDefaultTargetTriple)
```
tobiasgrosser, line 1582: I am curious, but was wondering if these changes + the stage0 update could be factored out into a separate PR?

Kha: Yes, these probably should have been part of #2410

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -1529,6 +1529,26 @@ def optimizeLLVMModule (mod : LLVM.Module ctx) : IO Unit := do
...
 
+/-- Get the names of all global symbols in the module -/
+partial def getModuleGlobals (mod : LLVM.Module llvmctx) : IO (HashSet String) := do
+  let rec go (v : LLVM.Value llvmctx) (acc : HashSet String) : IO (HashSet String) := do
+    if v.isNull then return acc
+    else go (← LLVM.getNextGlobal v) (acc.insert (← LLVM.Value.getName v))
+  go (← LLVM.getFirstGlobal mod) {}
+
+/-- Get the names of all global functions in the module -/
+partial def getModuleFunctions (mod : LLVM.Module llvmctx) : IO (HashSet String) := do
```
Kha, line 1540: Nit before someone makes an update-stage0 for this: why a HashSet anyway?

src/Lean/Compiler/IR/EmitLLVM.lean
hargoniX: Before we pull this out I would be curious to understand as to why this PR is failing know unlike #2340 :(

hargoniX: From diffing this against the newly updated #2340 it seems to me as if there is only diff in the update stage0 due to the different times that these were done. I don't understand how this ends us up with a failing linker in the LLVM CI though.

Kha: I'm not sure what you mwean by "different times", this branch does not have an update-stage0 at all. And it definitely needs one exactly for the linker fix it contains and itself needs.

Kha:
My only other nit is random CMake whitespace changes, I think we can merge as soon as those are addressed and CI passes

src/Lean/Compiler/IR/EmitLLVM.lean
tobiasgrosser: Right. I chatted about this with @bollu, I feel trying to debug these differences is a time-sink that we can avoid by pulling stuff out of #2340 and keeping it updated. I am trying to do this at the moment.

tobiasgrosser:
Nice. Chances are that this update-stage0 might be enough to get the PR through. :crossed_fingers:

src/Lean/Compiler/IR/EmitLLVM.lean
hargoniX: Changed to Array then re-update stage0

Kha:
Oh my gosh, I just looked over the changes one more time in Emacs and noticed that it introduces two *new* instances of trailing whitespace.

I understand this kind of incremental PR with many different authors is not easy but when it is marked ready for review, I expect
* the formatting to be double-checked. There are still two random newline insertions as well. We tell VS Code to remove trailing whitespace, if you use a different editor, it is your obligation to ensure that!
* the commit messages to be double-checked. The first commit mentions the same step twice (and we should make sure the information is included in the source code as well). The second commit talks about stage 2, which is not correct (anymore). The third commit talks about cleaning up the CI while it introduces it. And by the way, the assertion in the fourth commit message is not true as a) we do not cross-compile LLVM and b) we (necessarily) use the host compiler for cross-compilation.

Kha:
But apart from these nits, let me emphasize what an important milestone successfully building Lean with the backend is! I assume the grand majority of work is now behind us.

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -1529,6 +1529,26 @@ def optimizeLLVMModule (mod : LLVM.Module ctx) : IO Unit := do
...
+/-- Get the names of all global functions in the module -/
+partial def getModuleFunctions (mod : LLVM.Module llvmctx) : IO (Array String) := do
+  let rec go (v : LLVM.Value llvmctx) (acc : Array String) : IO (Array String) := do
+    if v.isNull then return acc
+    else 
+      let name ← LLVM.Value.getName v
+      -- | Do not insert internal linkage for 
+      -- intrinsics such as `@llvm.umul.with.overflow.i64` which clang generates, and also
+      -- for declarations such as `lean_inc_ref_cold` which are externally defined.
+      let acc := if (← LLVM.isDeclaration v) then acc else acc.push name
```
Kha, line 1548: This seems to be the wrong place for that code, it should be in `emitLLVM`

hargoniX:
Okay I hope I got all of your remarks down this time? If not please specify more precisely what you want us to change.

tobiasgrosser:
Lovely. Thank you @Kha

PR #2417, opened 2023-08-13
chore: remove trailing whitespace in EmitLLVM

tobiasgrosser:
This patch should not result in any functional changes, but will reduce the diff of an upcoming PR.


tobiasgrosser:
Thank you. That was fast.

PR #2421, opened 2023-08-15
test: compiling from the interpreter, with common imports

kim-em:
This is a (currently failing) test case to help explain the issue reported at https://github.com/leanprover/lean4/issues/2422.

If this problem is solvable by a change in Lean this PR can become a test case.


semorrison:
This test has been incorporated into the proposed fix at https://github.com/leanprover/lean4/pull/2423.

Issue #2422, opened 2023-08-15
Difference in `Frontend` behaviour in compiled vs interpreted code.


semorrison:
Consider the following situation:

* `A.lean` uses an initializer (e.g. `registerOption`)
* `B1.lean` and `B2.lean` both import `A.lean`.
* `C.lean` imports `B1`, and implements a `main` function which compiles a target source file.

Now:
* Running `C` to compile `B2` works fine.
* Using `#eval main ["B2.lean"]` in the interpreter in `C.lean` results in an error message while compiling `B2.lean`:
```
<input>:1:0: error: invalid option declaration 'option', option already exists
```

There is a Lean 4 test (currently failing) in https://github.com/leanprover/lean4/pull/2421 demonstrating this issue. The same files are available as a standalone repo https://github.com/semorrison/frontend-issue-2422 in case that's useful to experiment with.

Questions:
* Can someone explain to me why there is this difference?
* Are there workarounds, besides just always running the compiled version?

PR #2423, opened 2023-08-15
fix: interpret module initializer at most once

### LINKED ISSUE #2408: RFC: Using the Lean `Frontend` to compile multiple modules
Both @adamtopaz and I have recently been trying to use functions from `Lean.Elab.Frontend` to compile files from `Mathlib`.

In my case it is to extract data from `InfoTree`s (e.g. with to export all goal/tactic pairs in the library, or answer questions like "of goals currently closed by `aesop`, how many are solvable by `my_other_tactic`).

In a recent [zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/difficulties.20with.20.60Frontend.60) we've both reported similar difficulties. This issue summarises the problems, with a reproducible test case.

## To reproduce

A minimized version of this is set up as a test in #2407, which of course [currently fails](https://github.com/leanprover/lean4/actions/runs/5830058402/job/15810685857?pr=2407#step:11:4153).

You can put the following file into either `Mathlib`, or any project depending on `Mathlib`.

```lean
-- Save as "Frontend.lean"
import Lean.Elab.Frontend

open Lean Elab

unsafe def processInput (input : String) (initializers := false)  :
    IO (Environment × List Message) := do
  let fileName   := "<input>"
  let inputCtx   := Parser.mkInputContext input fileName
  if initializers then enableInitializersExecution
  let (header, parserState, messages) ← Parser.parseHeader inputCtx
  let (env, messages) ← processHeader header {} messages inputCtx
  let s ← IO.processCommands inputCtx parserState (Command.mkState env messages {}) <&> Frontend.State.commandState
  pure (s.env, s.messages.msgs.toList)

open System in
def findLean (mod : Name) : IO FilePath := do
  return FilePath.mk ((← findOLean mod).toString.replace "build/lib/" "") |>.withExtension "lean"

/-- Read the source code of the named module. -/
def moduleSource (mod : Name) : IO String := do
  IO.FS.readFile (← findLean mod)

unsafe def compileModule (mod : Name) (initializers := false) :
    IO (Environment × List Message) := do
  processInput (← moduleSource mod) initializers

unsafe def main (args : List String) : IO Unit := do
  initSearchPath (← findSysroot)
  let mut initializers := false
  match ← IO.getEnv "LEAN_INITIALIZERS" with
  | some "ONCE" =>
    IO.println "Invoking `enableInitializersExecution` once at the beginning."
    enableInitializersExecution
  | some "EACH" =>
    IO.println "Invoking `enableInitializersExecution` at each compilation."
    initializers := true
  | _ =>
    IO.println "Not invoking `enableInitializersExecution`"
  for mod in args do
    IO.println s!"Compiling {mod}"
    let (_, msgs) ← compileModule mod.toName initializers
    for m in msgs do IO.println (← m.toString)
```

and add it to your `lakefile` with the entry:
```
lean_exe «frontend» {
  root := `Frontend
}
```

## Usage

Now the command `lake exe frontend module1 ... modulen` will try to compile the named modules, one after the other.
The goal is to be able to run this on an arbitrary subset of modules from Mathlib. (Although having a limitation such as requiring a topological sort of the import graph would not be serious obstacle.) However at the moment I can only get this to work on a single module at a time.

Further, we can control when `enableInitializersExecution` is run, using an environment variable `LEAN_INITIALIZERS`, set to `ONCE`, `EACH`, or `NONE`.

* For `ONCE`, we execute `enableInitializersExecution` once, before starting to compile anything.
* For `EACH`, we execute `enableInitializersExecution` before compiling each file.
* For `NONE`, we don't run it at all.

## Results

* `NONE` simply doesn't work at all:

```
% LEAN_INITIALIZERS=NONE lake exe frontend Mathlib.Algebra.Abs
Not invoking `enableInitializersExecution`
Compiling Mathlib.Algebra.Abs
libc++abi: terminating with uncaught exception of type lean::exception: cannot evaluate `[init]` declaration 'Mathlib.Prelude.Rename.linter.uppercaseLean3' in the same module
```

* `ONCE` works on any single file, but not on any two files:

```
% LEAN_INITIALIZERS=ONCE lake exe frontend Mathlib.Algebra.Abs
Invoking `enableInitializersExecution` once at the beginning.
Compiling Mathlib.Algebra.Abs

% LEAN_INITIALIZERS=ONCE lake exe frontend Mathlib.Algebra.AddTorsor
Invoking `enableInitializersExecution` once at the beginning.
Compiling Mathlib.Algebra.AddTorsor

% LEAN_INITIALIZERS=ONCE lake exe frontend Mathlib.Algebra.Abs Mathlib.Algebra.AddTorsor
Invoking `enableInitializersExecution` at each compilation.
Compiling Mathlib.Algebra.Abs
Compiling Mathlib.Algebra.AddTorsor
libc++abi: terminating with uncaught exception of type lean::exception: cannot evaluate `[init]` declaration 'Mathlib.Tactic.reflExt' in the same module
```

* `EACH` works on any single file (because it is equivalent then to `ONCE`!) and fails with a different error message when run on any two files:

```
% LEAN_INITIALIZERS=EACH lake exe frontend Mathlib.Algebra.Abs Mathlib.Algebra.AddTorsor
Invoking `enableInitializersExecution` at each compilation.
Compiling Mathlib.Algebra.Abs
Compiling Mathlib.Algebra.AddTorsor
PANIC at Lean.PersistentHashMap.find! Lean.Data.PersistentHashMap:160:14: key is not in the map
...
<input>:6:0: error: invalid option declaration 'linter.uppercaseLean3', option already exists
...
```
(Many copies of the PANIC, and many subsequent errors, have been snipped.)

## Minimization

A minimal set of files to observe this behaviour is:

```
-- Save as "Frontend/RegisterOption.lean"
import Lean

register_option option : Bool := {
  defValue := true
  descr := ""
}
```

```
-- Save as "Frontend/Import1.Lean"
import Frontend.RegisterOption
```

```
-- Save as "Frontend/Import2.Lean"
import Frontend.RegisterOption
```

and then run
```
% LEAN_INITIALIZERS=EACH lake exe frontend Frontend.Import1 Frontend.Import2
Invoking `enableInitializersExecution` at each compilation.
Compiling Frontend.Import1
Compiling Frontend.Import2
<input>:1:0: error: invalid option declaration 'option', option already exists
```

## Discussion

Up for discussion:
* Do we really need this? Is working one file at a time manageable?
* Are there ways to achieve this goal with the existing code?
* @digama0 has suggested a change that addresses this issue https://github.com/leanprover/lean4/pull/2398, but requires a change to Lean 4. With that we have:

```
-- % LEAN_INITIALIZERS=REINIT lake exe frontend Mathlib.Algebra.Abs Mathlib.Algebra.AddTorsor
-- Invoking `enableInitializersExecution` and `enableReinitialization` at each compilation.
-- Compiling Mathlib.Algebra.Abs
-- Compiling Mathlib.Algebra.AddTorsor
```

I've grafted the failing test from #2407 onto that PR, and modified it to use `enableReinitialization`.

Kha:
Fixes #2408, #2422 

This fixes the inconsistency between interpreter and compiler at the heart of these issues. This was frankly an oversight in the initial implementation so I believe this is a good change independently of whether we think importing stuff multiple times in a single process is a good solution for the use cases.

@semorrison I tested it on both tests separately but could not trivially cherry-pick both of them. Assuming we want both of them, could you work on renaming or unifying them?

### DIFF
--- a/tests/pkg/frontend/Frontend.lean
+++ b/tests/pkg/frontend/Frontend.lean
@@ -1,0 +1,3 @@
+import Frontend.RegisterOption
+import Frontend.Import1
+import Frontend.Import2
--- a/tests/pkg/frontend/Frontend/Compile.lean
+++ b/tests/pkg/frontend/Frontend/Compile.lean
@@ -1,0 +1,28 @@
+import Lean.Elab.Frontend
+
+open Lean Elab
+
+unsafe def processInput (input : String) (initializers := false)  :
+    IO (Environment × List Message) := do
+  let fileName   := "<input>"
+  let inputCtx   := Parser.mkInputContext input fileName
+  if initializers then enableInitializersExecution
+  let (header, parserState, messages) ← Parser.parseHeader inputCtx
+  let (env, messages) ← processHeader header {} messages inputCtx
+  let s ← IO.processCommands inputCtx parserState (Command.mkState env messages {}) <&> Frontend.State.commandState
+  pure (s.env, s.messages.msgs.toList)
+
+open System in
+def findLean (mod : Name) : IO FilePath := do
+  let olean ← findOLean mod
+  -- Remove a "build/lib/" substring from the path.
+  let lean := olean.toString.replace (toString (FilePath.mk "build" / "lib") ++ FilePath.pathSeparator.toString) ""
+  return FilePath.mk lean |>.withExtension "lean"
+
+/-- Read the source code of the named module. -/
+def moduleSource (mod : Name) : IO String := do
+  IO.FS.readFile (← findLean mod)
+
+unsafe def compileModule (mod : Name) (initializers := false) :
+    IO (Environment × List Message) := do
+  processInput (← moduleSource mod) initializers
--- a/tests/pkg/frontend/Frontend/Import1.lean
+++ b/tests/pkg/frontend/Frontend/Import1.lean
@@ -1,0 +1,1 @@
+import Frontend.RegisterOption
--- a/tests/pkg/frontend/Frontend/Import2.lean
+++ b/tests/pkg/frontend/Frontend/Import2.lean
@@ -1,0 +1,1 @@
+import Frontend.RegisterOption
--- a/tests/pkg/frontend/Frontend/Main.lean
+++ b/tests/pkg/frontend/Frontend/Main.lean
@@ -1,0 +1,13 @@
+import Frontend.Compile
+
+open Lean
+
+unsafe def main (args : List String) : IO UInt32 := do
+  initSearchPath (← findSysroot)
+  let mut count : UInt32 := 0
+  for mod in args do
+    IO.println s!"Compiling {mod}"
+    let (_env, msgs) ← compileModule mod.toName true
+    for m in msgs do IO.println (← m.toString)
+    if msgs.length > 0 then count := 1
+  return count
--- a/tests/pkg/frontend/Frontend/Main_with_Import2.lean
+++ b/tests/pkg/frontend/Frontend/Main_with_Import2.lean
@@ -1,0 +1,2 @@
+import Frontend.Main
+import Frontend.Import2
--- a/tests/pkg/frontend/Frontend/Main_with_Import2_and_eval.lean
+++ b/tests/pkg/frontend/Frontend/Main_with_Import2_and_eval.lean
@@ -1,0 +1,6 @@
+import Frontend.Main
+import Frontend.Import2
+
+#eval show IO _ from do
+  let r ← main ["Frontend.Import1"]
+  if r ≠ 0 then throw <| IO.userError "Messages were generated!"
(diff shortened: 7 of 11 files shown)


digama0:
:+1: from me, seeing as it is apparently isomorphic to #2398 (the only reason I didn't write it like this is because I didn't know if it was okay to incur the overhead of keeping a map when the standard use case doesn't need it for correct behavior).

semorrison:
I've incorporated the test from #2407, and added a hack to the `findSource` function so it works on Windows too.

Kha:
Thanks Scott, also for the helpful test description in `test.sh`. @leodemoura I'm happy with the PR now, do you want to review as well?

leodemoura:
Happy with it too. Merged :)

PR #2424, opened 2023-08-15
fix: lake: reverse-ffi, manifest, and 62 tests

tydeu:
Fixes broken aspects of `reverse-ffi` on Windows, `manifest` when the default git branch is not `master`, and directly calls `elan` in 62. 

Also, per @Kha's request, test 62 is now excluded from the CTest suite because it downloads a toolchain. `reverse-ffi` is also excluded from the Linux LLVM CI because the test does not account for the new shared libraries that need linking (e.g., `libLLVM-15` and its dependencies).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,8 @@ jobs:
             prepare-llvm: ../script/prepare-llvm-linux.sh lean-llvm*
             binary-check: ldd -v
             # foreign code may be linked against more recent glibc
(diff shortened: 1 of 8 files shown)


Kha:
Out of interest, what was the Windows problem? There's a lot going on in that diff so I can't really tell

tydeu:
@Kha Also, apparently `make run-local` does not work in the Nix CI. It wasn't tested before, so it may have never worked. Any idea why it breaks? (If not, I will probably just not test `run-local` on Nix.)

PR #2428, opened 2023-08-17
feat: `lake env` w/o configuration + more

tydeu:
Closes leanprover/lake#179.

Improves `lake env` in a number of ways:
* Allows it to be run without a `lakefile.lean` configuration
* Runing it without a command will display the environment variables set in NAME=VALUE form a la POSIX `env`
* Now adds Lean/Lake and worskapce `bin` directories to `PATH`
* Adds Lean's shared library directories to the shared library path 
* Adds a test

### DIFF
--- a/src/lake/test/env/test.sh
+++ b/src/lake/test/env/test.sh
@@ -29,8 +29,11 @@ test `LEAN_CC=foo $LAKE env printenv LEAN_CC` = foo
 if [ "$OS" = Windows_NT ]; then
 $LAKE env which libleanshared.dll # DLL in `bin` directory is in `PATH`
 elif [ "`uname`" = Darwin ]; then
-$LAKE env printenv DYLD_LIBRARY_PATH | grep lib/lean
-$LAKE -d ../../examples/hello env printenv DYLD_LIBRARY_PATH | grep examples/hello
+# MacOS's System Integrity Protection does not permit
+# us to spawn a `printenv` process with `DYLD_LIBRARY_PATH` set
+# https://apple.stackexchange.com/questions/212945/unable-to-set-dyld-fallback-library-path-in-shell-on-osx-10-11-1
+$LAKE env | grep DYLD_LIBRARY_PATH | grep lib/lean
+$LAKE -d ../../examples/hello env | grep DYLD_LIBRARY_PATH | grep examples/hello
 else
 $LAKE env printenv LD_LIBRARY_PATH  | grep lib/lean
 $LAKE -d ../../examples/hello env printenv LD_LIBRARY_PATH | grep examples/hello


src/lake/Lake/CLI/Help.lean
```diff
@@ -36,8 +36,8 @@ COMMANDS:
   scripts               shorthand for `lake script list`
   run <script>          shorthand for `lake script run`
   serve                 start the Lean language server
-  env <cmd> [<args>...] execute a command in the workspace's environment
-  exe <exe> [<args>...] build an exe and run it in the workspace's environment
+  env <cmd> [<args>...] execute a command in Lake's environment
```
semorrison, line 39:
```suggestion
  env                   print the value of environment variables that would be set by `env <cmd>`
  env <cmd> [<args>...] execute a command in Lake's environment
```

semorrison: Oh, scratch that, it is in the detailed help.

src/lake/Lake/CLI/Help.lean
```diff
@@ -188,38 +188,42 @@ with the package configuration's `moreServerArgs` field and `args`.
...
-  LD_LIBRARY_PATH       adds the workspace's library directories (other Unix)"
+  LEAN_CC               set to the detected `cc` (if not using the bundled one)
+  LEAN_PATH             adds Lake's and the workspace's Lean library dirs
+  LEAN_SRC_PATH         adds Lake's and the workspace's source dirs
+  PATH                  adds Lean's, Lake's, and the workspace's binary dirs
+  PATH                  adds Lean's and the workspace's library dirs (Windows)
+  DYLD_LIBRARY_PATH     adds Lean's and the workspace's library dirs (MacOS)
+  LD_LIBRARY_PATH       adds Lean's and the workspace's library dirs (other)
+
+If run with no command, Lake will print out the variables set and their values,
```
semorrison, line 214:
```suggestion
If run with no command, `lake env` will print out the variables set and their values,
```

src/lake/test/env/test.sh
```diff
@@ -0,0 +1,36 @@
...
+$LAKE env printenv LAKE_HOME
+$LAKE env printenv LEAN_SYSROOT
+$LAKE env printenv LEAN_AR | grep -m1 ar
+$LAKE env printenv LEAN_PATH
+$LAKE -d ../../examples/hello env printenv LEAN_PATH | grep -m1 examples/hello
+$LAKE env printenv LEAN_SRC_PATH | grep -m1 lake
+$LAKE -d ../../examples/hello env printenv LEAN_SRC_PATH | grep -m1 examples/hello
+$LAKE -d ../../examples/hello env printenv PATH | grep -m1 examples/hello
+
+# Test that `env` perserves environment for certain variables
```
semorrison, line 23:
```suggestion
# Test that `env` preserves environment for certain variables
```

src/lake/test/env/test.sh
```diff
@@ -0,0 +1,36 @@
...
+
+LAKE=${LAKE:-../../build/bin/lake}
+
+# Test the functionality of `lake env`
+# Also test https://github.com/leanprover/lake/issues/179
+
+# Test `env` with no command
+$LAKE env
+
+# Test installation variables are set
```
semorrison, line 12:
```suggestion
# Test installation variables are set
# Note: `set -euxo pipefail` above ensures these fail if the variable is not set.
```
Maybe not necessary, but it took me a moment to understand how this was actually testing anything.

src/lake/test/env/test.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+LAKE=${LAKE:-../../build/bin/lake}
+
+# Test the functionality of `lake env`
+# Also test https://github.com/leanprover/lake/issues/179
+
+# Test `env` with no command
+$LAKE env
```
semorrison, line 10: Test this actually returns something? Just ` | grep ...` would do.

tydeu:
@Kha, you removed your review request a while back. Should I assume that means this is fine to merge as-is?

Kha:
It means that if you (and the other reviewers) are happy with it, I'm happy with it :) . I didn't review it.

PR #2432, opened 2023-08-17
feat: warn rather than error if `lake.lock` disappears

tydeu:
As [discussed](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/lake.20lock.20file/near/385655928) on Zulip.

### DIFF
--- a/src/lake/Lake/Build/Monad.lean
+++ b/src/lake/Lake/Build/Monad.lean
@@ -73,7 +73,15 @@ where
 
 /-- Busy wait to acquire the lock of `lockFile`, run `act`, and then release the lock. -/
 @[inline] def withLockFile [Monad m] [MonadFinally m] [MonadLiftT IO m] (lockFile : FilePath) (act : m α) : m α := do
-  try busyAcquireLockFile lockFile; act finally IO.FS.removeFile lockFile
+  try
+    busyAcquireLockFile lockFile; act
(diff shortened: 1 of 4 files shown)


src/lake/Lake/Build/Monad.lean
```diff
@@ -73,7 +73,15 @@ where
 
 /-- Busy wait to acquire the lock of `lockFile`, run `act`, and then release the lock. -/
 @[inline] def withLockFile [Monad m] [MonadFinally m] [MonadLiftT IO m] (lockFile : FilePath) (act : m α) : m α := do
-  try busyAcquireLockFile lockFile; act finally IO.FS.removeFile lockFile
+  try
+    busyAcquireLockFile lockFile; act
```
semorrison, line 77:
```suggestion
    busyAcquireLockFile lockFile
    act
```
Since you're no longer trying to save on new lines. :-) Feel free to ignore of course.

semorrison (approved):
LGTM

src/lake/Lake/Build/Monad.lean
tydeu: My hand was forced by the length of the line. 😆

tydeu: I am not opposed, but I do not think another CI build is worth it. ⏳

PR #2433, opened 2023-08-18
fix: FFI signature mismatches

Kha:
As reported in https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/wasm.20build.20signature.20mismatch

### DIFF
--- a/src/library/compiler/llvm.cpp
+++ b/src/library/compiler/llvm.cpp
@@ -1022,7 +1022,7 @@ extern "C" LEAN_EXPORT lean_object * llvm_count_params(size_t ctx, size_t f,
         false && ("Please build a version of Lean4 with -DLLVM=ON to invoke "
                   "the LLVM backend function."));
 #else
-    int n = LLVMCountParams(lean_to_Value(f));
+    unsigned n = LLVMCountParams(lean_to_Value(f));
     return lean_io_result_mk_ok(lean_box_uint64(n));
 #endif  // LEAN_LLVM
 }


Kha:
@bollu Could you please review the LLVM changes?

src/Lean/LocalContext.lean
```diff
@@ -177,9 +177,10 @@ def mkLocalDecl (lctx : LocalContext) (fvarId : FVarId) (userName : Name) (type
     let decl := LocalDecl.cdecl idx fvarId userName type bi kind
     { fvarIdToDecl := map.insert fvarId decl, decls := decls.push decl }
 
+-- `mkLocalDecl` without `kind`
 @[export lean_local_ctx_mk_local_decl]
-private def mkLocalDeclExported (lctx : LocalContext) (fvarId : FVarId) (userName : Name) (type : Expr) (bi : BinderInfo) (kind : LocalDeclKind := .default) : LocalContext :=
-  mkLocalDecl lctx fvarId userName type bi kind
+private def mkLocalDeclExported (lctx : LocalContext) (fvarId : FVarId) (userName : Name) (type : Expr) (bi : BinderInfo) : LocalContext :=
+  mkLocalDecl lctx fvarId userName type bi
```
Kha, line 183: @gebner I assume this was the original intention?

src/library/compiler/llvm.cpp
```diff
@@ -1023,7 +1023,7 @@ extern "C" LEAN_EXPORT uint64_t llvm_count_params(size_t ctx, size_t f,
                   "the LLVM backend function."));
 #else
     int n = LLVMCountParams(lean_to_Value(f));
```
Kha, line 1025: @bollu Shouldn't this be `unsigned`?

abentkamp:
I forgot to mention: `lean_get_githash` has the same issue.

hargoniX:
I wonder, is it possible for us to have some sort of linter that catches this type of stuff? So something that computes the C type (this should be the easy part as we already kind of do it) and checks if it matches with the actual extern type (this is the part that idk how to do)

src/library/compiler/llvm.cpp
bollu: Yes indeed, it should.

bollu (commented):
The LLVM changes look good to me.

src/Lean/LocalContext.lean
gebner: Ah yes, must have been a copy & paste error.  The `mkLetDeclExported` below does the right thing.

PR #2435, opened 2023-08-19
fix: make sure `refine` preserves pre-existing natural mvars

### LINKED ISSUE #2434: bug: `refine` loses track of natural goals
### Prerequisites

### Description

`refine` currently fails to include pre-existing natural metavariables from the refined term in the resulting goal list.

### Steps to Reproduce

1. Create natural goals during a tactic proof. For example, `constructor` creates natural metavariables.
2. `refine` a term which explicitly references these natural metavariables by name (e.g. `?left`).

**Expected behavior:** The resulting goal list should include all `?`-style goals appearing in the term.

**Actual behavior:** Natural pre-existing metavariables referred to in the term do not appear in the resulting goal list, and can therefore be closed erroneously.
```lean
example : True ∧ True := by -- (kernel) declaration has metavariables '_example'
  constructor
  · refine ?left
  · refine ?right 
```

### Versions

`4.0.0-nightly-2023-08-07`, on MacOS Ventura (Intel-based).

#### Explanation & possible fix

I believe this is a bug in `withCollectingNewGoalsFrom` in `elabTermWithHoles`. Note that `refine'` does not suffer from this issue. `refine` and `refine'` differ only in the value of the argument `allowNewMVars`, which guides behavior in the following block:
```lean
    ...
    let newMVarIds ← if allowNaturalHoles then
      pure newMVarIds.toList
    else
      let naturalMVarIds ← newMVarIds.filterM fun mvarId => return (← mvarId.getKind).isNatural
      let syntheticMVarIds ← newMVarIds.filterM fun mvarId => return !(← mvarId.getKind).isNatural
      let naturalMVarIds ← filterOldMVars naturalMVarIds mvarCounterSaved
      logUnassignedAndAbort naturalMVarIds
      pure syntheticMVarIds.toList
    ...
```
This properly distinguishes between new and old natural mvars for the purposes of error reporting, but fails to include the old natural mvars in the returned goals (and only returns `syntheticMVarIds`).

(Note that `newMVarIds` is simply what `withCollectingNewGoalsFrom` calls the collected `MVarId`s; `mvarCounterSaved` is what actually distinguishes "new" from "old", and this is not reflected in the variable names.)

Instead of separating synthetic mvars, we could simply look for ones which are both natural and new, and log and abort if we find them. If we make it through without aborting, then the MVarIds we've collected so far (`newMVarIds`) don't contain any erroneous mvars, and we can simply use them.
```lean
    unless allowNaturalHoles do
      let naturalMVarIds ← newMVarIds.filterM fun mvarId => return (← mvarId.getKind).isNatural
      let naturalMVarIds ← filterOldMVars naturalMVarIds mvarCounterSaved
      logUnassignedAndAbort naturalMVarIds
    let newMVarIds := newMVarIds.toList
```
#### More examples

Note that the behavior of `refine'` differs from `refine`. Contrast the above with
```lean
example : True ∧ True := by
  constructor
  · refine' ?right
    trivial
    -- note that this now occurs in the same goal focus;
    -- as expected, the current goal (`?left`) was replaced with `?right`
```
In general, consider
```lean
open Lean Meta Elab Tactic Term

elab "add_natural_goal" s:ident " : " t:term : tactic => do
  let g ← mkFreshExprMVar (← elabType t) .natural s.getId
  appendGoals [g.mvarId!]

elab "add_synthetic_goal" s:ident " : " t:term : tactic => do
  let g ← mkFreshExprMVar (← elabType t) .syntheticOpaque s.getId
  appendGoals [g.mvarId!]
```
We can see that it is indeed throwing away all natural goals; it's not just refining the main goal which causes the issue:
```lean
-- Buggy behavior:
-- Note that each focus ends with "no goals".
example : Bool × Nat := by -- (kernel) declaration has metavariables '_example'
  add_natural_goal d : Bool
  add_natural_goal e : Nat
  · refine (?d,?e)
  · refine ?d
  · refine ?e

example : Bool × Bool := by -- (kernel) declaration has metavariables '_example'
  add_natural_goal d : Bool
  add_natural_goal e : Bool
  · refine (?d,?e)
  · case d => refine ?e
  · refine ?e

-- Expected behavior:
example : Bool × Nat := by
  add_synthetic_goal d : Bool
  add_synthetic_goal e : Nat
  · refine (?d,?e) -- unsolved goals case d ⊢ Bool case e ⊢ Nat
  · refine ?d -- no goals to be solved
  · refine ?e
```
This was discussed briefly [on zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Bug.20in.20.60refine.60).

The possibility of opening a PR was mentioned in that thread; I'm happy to do so, but wanted to double-check here that it would be acceptable, given the contribution guidelines. :) In the meantime I've opened a draft PR at #2435.

thorimur:
Fixes #2434. `refine` currently fails to include pre-existing natural metavariables from the refined term in the resulting goal list. For example, `constructor` creates natural metavariables, and we can close the goals with `refine` alone:
```
example : True ∧ True := by -- (kernel) declaration has metavariables '_example'
  constructor
  · refine ?left
  · refine ?right
```
The source of this issue is a bug in `withCollectingNewGoalsFrom` in `elabTermWithHoles` when `allowNaturalHoles` is `false`. Metavariables are separated into natural and non-natural, and then any new unassigned natural mvars (as determined by their `index`) cause the tactic to be aborted. If the tactic isn't aborted here, the non-natural mvars are passed through to constitute the new goal list. However, this orphans any pre-existing "old" natural mvars.

This PR fixes this by changing the process of logging and aborting upon encountering new natural mvars into an `unless allowedNaturalHoles do ...`-style guard. If we pass that guard, then we don't need to filter the original list of mvars at all; it's guaranteed to be free of new natural mvars, as required.

### DIFF
--- a/src/Lean/Elab/Tactic/ElabTerm.lean
+++ b/src/Lean/Elab/Tactic/ElabTerm.lean
@@ -124,9 +124,12 @@ where
     let newMVarIds ← getMVarsNoDelayed val
     /- ignore let-rec auxiliary variables, they are synthesized automatically later -/
     let newMVarIds ← newMVarIds.filterM fun mvarId => return !(← Term.isLetRecAuxMVar mvarId)
-    /- if we pass this check while `allowNaturalHoles` is `false`, then all mvars in `newMVarIds`
-    are either non-natural or pre-existing natural mvars (and thus should be included in the
-    resulting goal list). See #2434. -/
+    /- The following `unless … do` block guards against unassigned natural mvarids created during
+    `k` in the case that `allowNaturalHoles := false`. If we pass this block without aborting, we
+    can be assured that `newMVarIds` does not contain unassigned natural mvars created during `k`.
+    Note that in all cases we must allow `newMVarIds` to contain unassigned natural mvars which
+    were created *before* `k`; this is the purpose of `mvarCounterSaved`, which lets us distinguish
+    mvars created before `k` from those created during and after. See issue #2434. -/
     unless allowNaturalHoles do
       let naturalMVarIds ← newMVarIds.filterM fun mvarId => return (← mvarId.getKind).isNatural
       let naturalMVarIds ← filterOldMVars naturalMVarIds mvarCounterSaved


src/Lean/Elab/Tactic/ElabTerm.lean
```diff
@@ -124,21 +124,20 @@ where
     let newMVarIds ← getMVarsNoDelayed val
     /- ignore let-rec auxiliary variables, they are synthesized automatically later -/
     let newMVarIds ← newMVarIds.filterM fun mvarId => return !(← Term.isLetRecAuxMVar mvarId)
-    let newMVarIds ← if allowNaturalHoles then
-      pure newMVarIds.toList
-    else
+    /- if we pass this check while `allowNaturalHoles` is `false`, then all mvars in `newMVarIds`
```
semorrison, line 127: Could you rephrase this comment? What is "this check" referring to?

tests/lean/refinePreservesNaturalMVars.lean
```diff
@@ -0,0 +1,53 @@
...
+`no goals` error. Instead, this should succeed.
+
+-/
+
+example : Bool × Bool := by
+  add_natural_goal d : Bool
+  add_natural_goal e : Bool
+  · refine (?d,?e)
+    refine ?d
+    refine ?e
```
semorrison, line 52:
```suggestion
    refine ?e   -- This unifies `?d` and `?e`, so only one goal remains.
```
Took me a moment to understand why this was working!

src/Lean/Elab/Tactic/ElabTerm.lean
thorimur: How's this as a rewrite of the comment? (Too verbose?)
```
    /- The following `unless … do` block guards against unassigned natural mvarids created during 
    `k` in the case that `allowNaturalMVars := false`. If we pass this block without aborting, we 
    can be assured that `newMVarIds` does not contain unassigned natural mvars created during `k`. 
    Note that in all cases we must allow `newMVarIds` to contain unassigned natural mvars which 
    were created *before* `k`; this is the purpose of `mvarCounterSaved`, which lets us distinguish 
    mvars created before `k` from those created during and after. See issue #2434. -/
```

thorimur: Tentatively committed, as I might be away from my computer for a while soon. :)

semorrison (approved):

thorimur:
Commenting in an attempt to cause github to reopen this PR.

PR #2439, opened 2023-08-21
doc: fix typo in Lake's Require DSL

ConnorBaker:
While this did bring me some amount of joy, I believe this is a typo.


tydeu:
Thank you!

PR #2440, opened 2023-08-21
doc: fix mid priority doc comment

digama0:


Kha:
It's not `med` though :smile:

Issue #2441, opened 2023-08-21
RFC: releases for PRs


semorrison:
This RFC proposes modifying CI so that every PR generates its own release. 

These releases would live on the `leanprover/lean4-pr-releases` repository (i.e. just as nightly releases live on `leanprover/lean4-nightly`) rather than the main repo.

Why?
* We would like to automatically test Mathlib and Std against PRs, and to do so we need an available `lean-toolchain`. (This plan is not to have Lean 4 CI fail if there are failures in Mathlib or Std. Instead if there is a failure, a bot will comment, providing a link to a branch on the relevant repo and the CI log.)
* Currently it is possible to generate your own releases, by tagging a commit on your own fork (making sure to satisfy some rules: no underscores, don't start with a number) and enabling CI actions on your fork. But users struggle with this, and it would be nice to make it easier.

I'll put up a PR shortly that implements this (modulo creating the `leanprover/lean4-pr-releases` repo, and creating an appropriate token that can push to this from lean4 CI).

When this is enabled, once CI is complete you will be able to run against your PR by putting `leanprover/lean4-pr-releases:pr-release-NNNN` in your `lean-toolchain`.

semorrison:
This is done now.

PR #2442, opened 2023-08-21
feat: create release at lean4-pr-releases for each PR

kim-em:
This creates a release for each PR. When the PR is updated the release is simply deleted and recreated. See https://github.com/leanprover/lean4/issues/2441 for the motivation.

After CI has run, every PR will be usable by putting `leanprover/lean4-pr-releases:pr-release-NNNN` in your `lean-toolchain`.

~~This PR is not quite ready to go: we need to actually create the `leanprover/lean4-pr-releases` repo, and install an appropriate token.~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -302,7 +302,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
         with:
-          token: ${{ secrets.PR_RELEASE_TOKEN }}
+          token: ${{ secrets.PR_RELEASES_TOKEN }}
           # needed for tagging
           fetch-depth: 0
       - uses: actions/download-artifact@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -294,6 +294,42 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+  pr-release:
+    if: ${{ ! startsWith(github.ref, 'refs/tags/') && ! needs.set-nightly.outputs.nightly }}
```
tydeu, line 298: Will this not also trigger on directly pushes to branches? It seems like it would be best to simply check that `github.event_name == 'pull_request'`,

semorrison: Hmm, but then that will only trigger when the PR is opened or its metadata modified in some way: just pushing to the branch would not trigger updating the release.

tydeu: @semorrison That is not my understanding. From Github's documentation on the [`pull_request`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) event, all forms of pull request modifications (including pushing to branch, which falls under `synchronize`) have `pull_request` event name.

semorrison: Instead I've added an additional check that the `github.event.number` is nonempty.

semorrison: Ah, I hate the Github docs. That page gives no definition for `synchronize`!

I'll try your suggestion.

semorrison:
This needs rethinking. The relevant secret is not available for PRs from forks. I will switch to using a second action based on [`workflow-run`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run).

PR #2444, opened 2023-08-22
feat: cache built file hashes

tydeu:
After building a file, Lake saves the file's hash to a `.hash` which can be reused by future builds to avoid the performance cost of rehashing the file.

A new CLI option `--rehash`/`-H` can be used to prevent Lake from using `.hash` files.

### DIFF
--- a/src/lake/Lake/Build/Common.lean
+++ b/src/lake/Lake/Build/Common.lean
@@ -14,24 +14,49 @@ namespace Lake
 def inputFile (path : FilePath) : SchedulerM (BuildJob FilePath) :=
   Job.async <| (path, ·) <$> computeTrace path
 
-@[specialize] def buildUnlessUpToDate [CheckExists ι] [GetMTime ι] (info : ι)
+/-- Check if the `info` is up-to-date by comparing `depTrace` with `traceFile`. -/
+@[specialize] def BuildTrace.checkUpToDate [CheckExists ι] [GetMTime ι]
+(info : ι) (depTrace : BuildTrace) (traceFile : FilePath) : JobM Bool := do
+  if (← getIsOldMode) then
+    depTrace.checkAgainstTime info
+  else
+    depTrace.checkAgainstFile info traceFile
+
+@[inline] def buildUnlessUpToDate [CheckExists ι] [GetMTime ι] (info : ι)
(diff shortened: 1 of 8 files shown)


Vtec234:
Is the performance cost of hashing source code a significant fraction of build times / Lake overhead? I would find that quite surprising!

tydeu:
@Vtec234 This does not in fact cache source hashes, just build outputs (e.g., `.olean`, `.c`, `.o`, `.so`, etc.).

Kha:
Any measurements of the impact? :)

digama0:
@Vtec234 Yes, this was a major factor in reducing the runtime of lean-cache, see https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Lake.20server.20uses.20options.20from.20the.20wrong.20package/near/376858786

tydeu:
!bench

tydeu:
Comparing this run on this PR with the run with the base master commit + benchmark (i.e., https://github.com/leanprover/lean4/pull/2457#issuecomment-1699782763) gives the [expected significant improvement](http://speed.lean-fro.org/lean4/compare/275405ae-de4b-4046-9bd6-9968c8a7b6f3/to/bab50d33-13a1-40c4-85a0-fda5f93f8b22?hash2=b56cb22b1e3bacfe9a47682997102840c61347fc) in no-op build performance. From the comparison:

```diff
  Benchmark         Metric         Change
  =================================================
+ lake build no-op  branches       -32.7% (-63 σ)
+ lake build no-op  instructions   -44.5% (-96.2 σ)
+ lake build no-op  maxrss         -54.8% (-48.3 σ)
+ lake build no-op  task-clock     -57.6% (-25.1 σ)
```

tydeu:
!bench

tydeu:
Comparing [the two runs](http://speed.lean-fro.org/lean4/compare/7ef02c1c-37bd-4a78-b8fb-bc3ab93f4670/to/bab50d33-13a1-40c4-85a0-fda5f93f8b22?hash2=143e265d24ce3ce0ad543e6dbb18683b533eaeb3) with a split between the Lean and dummy no-op builds, only the Lean build displays a significant improvement. Thus, this the change is clearly dependent on the complexity of the OLean code.

PR #2445, opened 2023-08-22
chore: disable `lake.lock` (for now)

tydeu:
@semorrison merge this if you feel it is best.

### DIFF
--- a/src/lake/Lake/Build/Monad.lean
+++ b/src/lake/Lake/Build/Monad.lean
@@ -94,7 +94,13 @@ where
 /-- Run the given build function in the Workspace's context. -/
 @[inline] def Workspace.runBuild (ws : Workspace) (build : BuildM α) (oldMode := false) : LogIO α := do
   let ctx ← mkBuildContext ws oldMode
-  --withLockFile ws.lockFile do
+  /-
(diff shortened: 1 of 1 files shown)


semorrison:
Okay, a quick survey of Mathlib maintainers suggests that they think that this is likely annoying to regular users, so let's revert the lockfile feature for now, and restore it once we have `ctrl-c` handling.

src/lake/Lake/Build/Monad.lean
```diff
@@ -94,7 +94,8 @@ where
 /-- Run the given build function in the Workspace's context. -/
 @[inline] def Workspace.runBuild (ws : Workspace) (build : BuildM α) (oldMode := false) : LogIO α := do
   let ctx ← mkBuildContext ws oldMode
-  withLockFile ws.lockFile do build.run ctx
+  --withLockFile ws.lockFile do
```
semorrison, line 97:
```suggestion
  -- Use of the lock file has been temporarily disabled (in lean4#2445)
  -- until we have API for catching a `ctrl-c` during the build.
  -- Without this the lock file is too disruptive for users.
  -- withLockFile ws.lockFile do
```

PR #2457, opened 2023-08-24
test: add lake benchmarks

tydeu:
@Kha Hopefully I configured this right.

### DIFF
--- a/tests/bench/speedcenter.exec.velcom.yaml
+++ b/tests/bench/speedcenter.exec.velcom.yaml
@@ -124,26 +124,26 @@
       lake -dinundation run mk
       lake -dinundation build
       "
-- attributes:
-    description: lake build lean
-    tags: [slow]
-  run_config:
-    <<: *time
-    cmd: |
-      bash -c "
-      set -ex
-      ulimit -s unlimited
-      lake -Ktest=Clean clean
-      lake -Ktest=Clean build
-      "
-    max_runs: 2
-  build_config:
-    cmd: |
-      bash -c "
-      set -ex
-      ulimit -s unlimited
-      lake -Ktest=Clean build
-      "
+#- attributes:
+#    description: lake build lean
+#    tags: [slow]
+#  run_config:
+#    <<: *time
+#    cmd: |
+#      bash -c "
+#      set -ex
+#      ulimit -s unlimited
+#      lake -Ktest=Clean clean
+#      lake -Ktest=Clean build
+#      "
+#    max_runs: 2
+#  build_config:
+#    cmd: |
+#      bash -c "
+#      set -ex
+#      ulimit -s unlimited
+#      lake -Ktest=Clean build
+#      "
 - attributes:
     description: lake build no-op
     tags: [fast]


tydeu:
!bench

tydeu:
!bench

tydeu:
!bench

tydeu:
!bench

tydeu:
@Kha Any suggestion for improvements? Two potential uncovered test cases I thought of:

1. Multi-package elaboration
2. Modules with actual code

Case (2) could be reasonably addressed by benchmarking a build of the Lean core itself. However, I am not sure either of these cases would provide significantly different benchmarks to be worth the added complexity.

tydeu:
I liked the benchmarks from #2463, so I went ahead and merged them into here.

tydeu:
!bench

tydeu:
Interesting, it appears a no-op build of Lean is faster than I no-op build of the inundation dummy project. So I guess number of files to hash is more significant than the contents of them?

Kha (changes requested):
> Modules with actual code

Why is that interesting? We want to benchmark Lake, not Lean. A synthetic benchmark is great, it ensures that the input stays constant and cannot influence the timings.

Kha:
> Modules with actual code

Ah, I suppose it is interesting e.g. in the context of #2444. I think I would still value isolated tests over that. It's okay to do benchmarks like this one manually if we don't expect output size to matter anymore after this change.

tydeu:
@Kha Sime other interesting points of the Lean builds and real code builds:

 * Provides a time comparison with another build system (CMake/Make) -- we can officially say that, in parallel, Lake is faster than Make even for a complex project.

 * Enables measuring of code-level finer-grain dependency specification (like the forthcoming module system or leanprover/lake#86).

 * Similarly, things like asynchronous proof checking could also be captured by such a benchmark (if these are integrated with Lake enough to enable it to move onto building downstream files while the proofs are still being checked).

However, I am happy to remove the benchmark if you are confident it is not worth it. I was also initially unsure whether it was good idea.

tydeu:
!bench

tydeu:
!bench

tydeu:
!bench

tydeu:
!bench

tests/bench/inundation/README.md
```diff
@@ -0,0 +1,37 @@
+# Stress Test for Lake
+
+This folder generates a deeply nested import tree of Lean modules
+that can then be used as a stress test of large Lake builds (e.g., Mathlib).
+The the modules lack any code.
```
Kha, line 5:
```suggestion
The modules lack any code.

```

tydeu:
> In any case, I'm still not convinced this is worth increasing the benchmark run time by 50%, no.

It may be worth noting that the Lean build itself is only contributing 4-5 minutes (so about 25% of the overall time), the other parts contribute about 1-3 minutes and benchmark without these changes is already usually 13-14 minutes. However, I presume that means you prefer me to take it out, so I will do that soon.

Kha:
> It may be worth noting that the Lean build itself is only contributing 4-5 minutes (so about 25% of the overall time), the other parts contribute about 1-3 minutes and benchmark without these changes is already usually 13-14 minutes

I'm not sure what you are trying to tell me with that. All I see is a total increase from 13 minutes to 20 minutes.

PR #2459, opened 2023-08-25
fix: deprecation span should only cover the ident

digama0:
Previously, if `foo` is deprecated then the entire `simp [foo]` call would be highlighted. This causes issues for a linter I am writing which requires that the warning just cover the identifier, as well as just being generally confusing for the user if there are many identifiers in the `simp` call. 

`simp` is not the only affected tactic, many functions call `resolveId?` without setting the ref first. An alternative solution would be to have `Linter.checkDeprecated` take a syntax directly instead of using the ref, but there are also some other errors that identifier resolution can throw which are subject to the same issue.


Kha:
LGTM but could you add a brief comment as to what the new test is testing?

PR #2460, opened 2023-08-25
feat: add labels from comments

kim-em:
This PR allows all users to manage the `awaiting-review`, `awaiting-author`, and `WIP` labels, simply by writing one of those labels as a comment (with no other content). The other labels are removed if present.

I've added a bullet point to the PR template to let people know about this.

I think it will be okay to allow all users access to these labels. If it's a problem, we can change the behaviour of this workflow so it only listens to the author of the PR.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,3 +9,6 @@
 
 * Please make sure the PR has excellent documentation and tests.
   If we label it `missing documentation` or `missing tests` then it needs fixing!
+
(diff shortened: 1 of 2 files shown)


semorrison:
(This same workflow has just been deployed on `std4`, and it works!)

Kha:
LGTM but could we weaken the condition to "on a line of its own"? Then we could save ourselves from some double-comments. I believe `!bench` has this logic.

semorrison:
Okay, should be good to go now.

semorrison:
Let's do some testing-in-production:

awaiting-review

semorrison:
Looks good.

PR #2463, opened 2023-08-26
test: add lake benchmarks (builds Init/Lean/Lake)

tydeu:
An alternative to #2457 that builds Init/Lean/Lake rather than a dummy package.


tydeu:
!bench

tydeu:
!bench

tydeu:
Merged into #2457.

PR #2464, opened 2023-08-26
feat: FFI thread initialization

digama0:
* #2490
* [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Expose.20thread.20management.20routines.20to.20FFI.3F/near/387373357)

This exposes the following functions to FFI:
```c
void lean_finalize_runtime_module();
void lean_finalize();
void lean_thread_initialize();
void lean_thread_finalize();
bool lean_in_thread_finalization();
void lean_reset_thread_local();
void lean_register_thread_finalizer(void (*fn)(void *), void * p);
void lean_register_post_thread_finalizer(void (*fn)(void *), void * p);
void lean_run_thread_finalizers();
void lean_run_post_thread_finalizers();
```

Rationale:
* `lean_finalize_runtime_module`, `lean_finalize`: these are counterparts to `lean_initialize(_runtime_module)` which are already exposed. They are useful in particular for running multiple lean sessions in the same process.
* `lean_thread_initialize`, `lean_thread_finalize`: These are essential for running lean code in multithreaded programs (where lean was not responsible for starting the thread).
* `lean_in_thread_finalization`: A useful utility when working with finalizers
* `lean_reset_thread_local`: Useful for resetting lean state if we want to run lean code from an unknown context
* `lean_register_thread_finalizer`, `lean_register_post_thread_finalizer`: Allows FFI code to register itself for cleanup along with lean code, especially important when `lean_external` objects in the lean heap depend on some global state not managed by lean.
* `lean_run_thread_finalizers`, `lean_run_post_thread_finalizers`: a more itemized version of `lean_thread_finalize` in case you want to run code between the two (as lean itself does).

### DIFF
--- a/doc/dev/ffi.md
+++ b/doc/dev/ffi.md
@@ -109,6 +109,30 @@ if (lean_io_result_is_ok(res)) {
 
 //lean_init_task_manager();  // necessary if you (indirectly) use `Task`
 lean_io_mark_end_initialization();
+
+// Do work using the Lean runtime, e.g. lean_alloc_object(sz);
+
+// Finalization is not required since the OS will normally clean up,
+// but this will avoid memory leak reports in valgrind
+void lean_finalize_runtime_module();
+void lean_finalize();
+lean_finalize_runtime_module();
+//lean_finalize();
+```
+
+When using Lean from a multithreaded program, any threads created externally to Lean must also be initialized before any Lean code can be run on them.
+```c
+void lean_thread_initialize();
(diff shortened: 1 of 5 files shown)


SchrodingerZhu:
Any update on this?

digama0:
you should submit your RFC as an RFC issue

SchrodingerZhu:
Will do.

PR #2466, opened 2023-08-27
doc: use {b_}obj_{arg,res} consistently in lean.h

digama0:
* [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/memory.20model.20of.20natural.20numbers/near/387485542)

Documentation regarding the borrowing convention of arguments and return values for C functions in `lean.h` and related files is very important for FFI writers to be able to use `lean_inc` and `lean_dec` correctly. This PR replaces all uses of `object *` in arguments and returns with the appropriate typedef from `b_lean_obj_arg`, `u_lean_obj_arg`, `lean_obj_arg` , `b_lean_obj_res`, `lean_obj_res`. I have reviewed all annotations added here. Functions that use lean objects in "weird" states like alloc/free have been left as `lean_object *`.

One aspect which I could use some guidance on is whether to use `lean_obj_arg` or `obj_arg` in `object.cpp` and other cpp files which have both spellings available. The existing code seems to use them inconsistently. In this PR I opted to keep the signature spelled the same between the header file and the cpp file, meaning that functions would use the `lean_` spelling when the declaration is in `lean.h` and the `lean::` one for things in `object.h`.

### DIFF
--- a/src/runtime/object.h
+++ b/src/runtime/object.h
@@ -38,13 +38,13 @@ inline external_object_class * register_external_object_class(external_object_fi
     return lean_register_external_class(p1, p2);
 }
 
-inline bool is_scalar(object * o) { return lean_is_scalar(o); }
+inline bool is_scalar(b_obj_arg o) { return lean_is_scalar(o); }
 inline object * box(size_t n) { return lean_box(n); }
 inline size_t unbox(object * o) { return lean_unbox(o); }
 
-inline bool is_mt_heap_obj(object * o) { return lean_is_mt(o); }
-inline bool is_st_heap_obj(object * o) { return lean_is_st(o); }
-inline bool is_heap_obj(object * o) { return is_st_heap_obj(o) || is_mt_heap_obj(o); }
+inline bool is_mt_heap_obj(b_obj_arg o) { return lean_is_mt(o); }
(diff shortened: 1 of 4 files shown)


src/runtime/object.h
```diff
@@ -99,126 +99,126 @@ inline void cnstr_set_float(b_obj_arg o, unsigned offset, double v) { lean_ctor_
 // =======================================
 // Closures
 
-void free_closure_obj(object * o);
```
digama0, line 102: The declaration `free_closure_obj` was removed because the function doesn't exist.

src/runtime/object.h
```diff
@@ -340,7 +340,8 @@ inline obj_res nat_lxor(b_obj_arg a1, b_obj_arg a2) { return lean_nat_lxor(a1, a
 
 // =======================================
 // Integers
-object * mk_int_obj_core(mpz const & m);
+obj_res mpz_to_int_core(mpz const & m);
+inline obj_res mk_int_obj_core(mpz const & m) { return mpz_to_int_core(m); }
```
digama0, line 344: The function `mk_int_obj_core` did not exist, but `mpz_to_int_core` does. I just copied the pattern used in `mk_nat_obj_core` here.

PR #2471, opened 2023-08-28
perf: Don't hash the key when searching in empty hash tables

hargoniX:
Based on: https://github.com/rust-lang/hashbrown/pull/305

Judging from the rustc benchmarks this can save significant amounts of instructions and seems like quite a simple change. It is however unclear to me whether we do have the usage pattern of HashMaps that they do somewhere so it might possibly be useless for us as well. I don't have a local benchmark setup so we have to run performance tests here if we are interested in this. I'm guessing I don't have the privileges for that?

### DIFF
--- a/src/Lean/Data/HashMap.lean
+++ b/src/Lean/Data/HashMap.lean
@@ -64,18 +64,21 @@ private def mkIdx {sz : Nat} (hash : UInt64) (h : sz.isPowerOfTwo) : { u : USize
 
 def findEntry? [BEq α] [Hashable α] (m : HashMapImp α β) (a : α) : Option (α × β) :=
   match m with
+  | ⟨0, _⟩ => none
   | ⟨_, buckets⟩ =>
     let ⟨i, h⟩ := mkIdx (hash a) buckets.property
     buckets.val[i].findEntry? a
 
 def find? [beq : BEq α] [Hashable α] (m : HashMapImp α β) (a : α) : Option β :=
   match m with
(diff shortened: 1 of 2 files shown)


semorrison:
@hargoniX, I think anyone is allowed to run

!bench

semorrison:
Apparently not! @Kha, I'm presuming you need to sprinkle some magic permissions bits here?

leodemoura:
!bench

semorrison:
- ✅ Mathlib has successfully built against this PR. (2023-08-31 04:53:58)

semorrison:
- ✅ Mathlib branch [lean-pr-testing-2471](https://github.com/leanprover-community/mathlib4/compare/master...lean-pr-testing-2471) has successfully built against this PR. (2023-08-31 05:34:45)

semorrison:
Here are the bench results on Mathlib: http://speed.lean-fro.org/mathlib4/compare/93695581-e592-41ad-9a6a-53fe2cce9e75/to/e760a006-736d-49ba-b9b4-d6d79e4f697f.

I think without any interesting performance gain we can put this one to rest?

hargoniX:
@digama0 would you like to have this optimization on the std4 HashMap?

semorrison:
- ✅ Mathlib branch [lean-pr-testing-2471](https://github.com/leanprover-community/mathlib4/compare/master...lean-pr-testing-2471) has successfully built against this PR. (2023-08-31 10:59:42)

PR #2472, opened 2023-08-29
chore: update CI to create official releases

kim-em:
@tydeu and I realised, thanks to a bug report from Michael Stoll, that creating a release candidate simply by tagging is not correct, as this creates a source release without `Lean.toolchain` set.

This PR adjusts CI so that:
* When a tag is being pushed, we check if the tag is "v" followed by a valid SemVer string
* If so, use high compression and set `LEAN_SPECIAL_VERSION_DESC`
* and then release as usual for a tag release.

We do not attempt to write the release description, or fill in a changelog, as we expect to want to do this by hand, at least at first.

I will use this to create `v4.0.0-rc3` before we continue with `v4.0.0`. Unfortunately discovering this only now means that we probably should wait for `v4.0.0` until next week.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,8 +39,41 @@ jobs:
             fi
           fi
 
+  # This job determines if this CI build is for a tagged release.
+  # It only runs when a tag is pushed to the `leanprover` repository.
+  # It sets `set-release.outputs.RELEASE_TAG` to the tag, if the tag is "v" followed by a valid semver,
+  # and sets `set-release.outputs.{LEAN_VERSION_MAJOR,LEAN_VERSION_MINOR,LEAN_VERSION_PATCH,LEAN_SPECIAL_VERSION_DESC}`
+  # to the semver components parsed via regex.
+  set-release:
+    runs-on: ubuntu-latest
+    outputs:
+      LEAN_VERSION_MAJOR: ${{ steps.set.outputs.LEAN_VERSION_MAJOR }}
+      LEAN_VERSION_MINOR: ${{ steps.set.outputs.LEAN_VERSION_MINOR }}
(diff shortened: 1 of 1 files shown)


tydeu:
@semorrison The whole tag should not be in `LEAN_SPECIAL_VERSION_DESC`. Instead, it needs to be split into parts and multiple defines set:

* `LEAN_VERSION_MAJOR` ,  `LEAN_VERSION_MINOR`,  `LEAN_VERSION_PATCH`: The three numerical parts of the semver.
* `LEAN_VERSION_IS_RELEASE`: Set to true (`1`).
* `LEAN_SPECIAL_VERSION_DESC`: The part after the `-`.

digama0:
Can we just fix the `rc2` release? Seeing as it is not the code that was broken but rather the release process.

tydeu:
@digama0 We can retag `rc2` and get a new build, but it does need to be rebuilt because this data is stored in the compiled objects.

semorrison:
@tydeu, thanks, I now parse the SemVer string and set these options separately. Could you look again?

semorrison:
I'd prefer to be conservative here and take the attitude that we're testing out the release process itself, and can take as many release candidates as we need to get it right.

tydeu (approved):
LGTM!

semorrison:
Okay, let's try it in production then.

.github/workflows/ci.yml
```diff
@@ -39,8 +39,38 @@ jobs:
             fi
           fi
 
+  # This job determines if this CI build is for a tagged release.
+  # It only runs when a tag is pushed to the `leanprover` repository.
+  # It sets `steps.set.outputs.release` to the tag, if the tag is "v" followed by a valid semver.
```
tydeu, line 44: Oh, I think you may have forgotten to update this doc? It does not set `steps.set.outputs.release` now but the various variables, correct? Or are you using `release` as a stand-in for all the different variables?

PR #2478, opened 2023-08-29
feat: use supplied structure fields left to right and eta reduce terms in structure instance elaboration

### LINKED ISSUE #2451: RFC: tweak structure instance elaboration to avoid un-needed eta expansion
Currently, we have the following behavior for structure instance elaboration 
```lean 
structure A where 

structure B extends A where 

def a : A := sorry 

def b : B := { a with } 
#print b 
-- def b' : B :=
-- let src := a;
-- { toA := { } } <- There is an extra eta expansion for the `toA` field: 
```

I believe it would be advisable to change this to the following behavior
```lean 
structure A where 

structure B extends A where 

def a : A := sorry 

def b : B := { a with } 
#print b 
-- def b' : B :=
-- let src := a;
-- { toA := src } <- There is no eta expansion for the `toA` field: 
```

The extra eta expansion is due to the following: the function [`mkProjStx?`](https://github.com/leanprover/lean4/blob/d29b8e542236020f195faad5d7a4af566342f9ff/src/Lean/Elab/StructInst.lean#L453) will not apply when the arguments are `` _ `A `toA`` as `` `toA`` is not a field of `` `A.``  

The change in behavior can be achieved by adding two lines to the function `addMissingFields` [here](https://github.com/leanprover/lean4/blob/d29b8e542236020f195faad5d7a4af566342f9ff/src/Lean/Elab/StructInst.lean#L508): 
```lean4
if let some stx := s.source.explicit.find? (·.structName == substructName)|>.map (·.stx) then
  addField (FieldVal.term stx)
```
This accounts for the case where the projection is exactly to the type of an explicit source. 

> User Experience: How does this feature improve the user experience?

There is no extra eta expansion when constructing fields which are also fields of the explicit sources. This change makes the behavior more uniform and therefore reduces mental overhead for the user. 

Furthermore, there is a performance penalty for the extra eta expansion. While small for a single instance, things can snowball when the pattern is nested. The expanded term forces unfolding during unification when otherwise it would not be necessary. 

> Beneficiaries: Which Lean users and projects do benefit most from this feature/change?

The largest benefit is for the libraries with deeply nested type class hierarchies, so Mathlib. 

In leanprover-community/mathlib4#6759 I benchmarked Mathlib with a toolchain containing this change plus some complementary changes directly to Mathlib.  The [changes](http://speed.lean-fro.org/mathlib4/compare/e8adfe39-bc50-4ffb-bcfc-d119b2b5e7cb/to/ace9c3f2-3d04-4f20-811a-6e7ca9be167f) were overall very positive including a 4% reduction to total CPU instructions and dramatic improvements in many files.

Furthermore, at least one example of a serious timeout in algebraic hierarchy presented on Zulip has been completely eliminated. 

> Community Feedback: Have you sought feedback or insights from other Lean users?

This arose from multiple discussions on Zulip about performance problems in Mathlib, especially in the algebraic hierarchy. It bubbled up already in a separate issue #2387. 

> Maintainability: Will this change streamline code maintenance or simplify its structure?

Given the simplicity of the change, I would not expect additional maintenance burden. 

I have a [branch](https://github.com/leanprover/lean4/compare/master...mattrobball:lean4:reduce_etas) on my fork with this change plus a simple test (as above) to document the behavior and am ready to submit a PR. Looking forward to hearing feedback!

mattrobball:
Modifies the structure instance elaborator to
1. Fill in missing fields from sources in strict left-to-right order. In `{a, b with}`, sometimes the elaborator
  would ignore `a` even if both `a` and `b` provided the same field, depending on what subobject fields they had.
2. Use the sources, or subobjects of the sources, to fill in entire subobjects of the target structure as much as possible.
  Currently, a field cannot be filled directly by a source itself resulting in the term being eta expanded.
  This change avoids this unnecessary and surprisingly costly extra eta expansion.

Adds two new tests to illustrate the performance benefit (one courtesy @semorrison). These are currently failing on master and succeed on this branch.

There is one additional test to exercise the changes to the elaboration of structure instances.

Changes to make mathlib build are in leanprover-community/mathlib4#9843

Closes #2451

### DIFF
--- a/tests/lean/run/463.lean
+++ b/tests/lean/run/463.lean
@@ -4,4 +4,6 @@ structure C := (a b c : Nat)
 structure D := (toA : A) (c : Nat)
 
 def foo (s : C) : B := {s with} -- works in lean 4, works in lean 3
--- def bar (s : D) : B := {s with} -- works in lean 4, fails in lean 3
+-- def bar (s : D) : B := {s with} -- no longer works after #2478
+-- Should fields of non-subobject fields be treated as fields 
+-- of the original structure?


mattrobball:
awaiting-review

tests/lean/run/structWith_scott.lean
```diff
@@ -0,0 +1,1322 @@
...
+  RingQuot.instCommRingRingQuotToSemiringToCommSemiring _
+
+instance instAlgebra
+    {R A M} [CommSemiring R] [CommRing A] [Algebra R A] :
+    Algebra R (Quot_r A M) :=
+  RingQuot.instAlgebraRingQuot _
+
+--------------
+
+/- Typeclass synthesis should remain fast when multiple `with` patterns are nested -/
```
semorrison, line 1319: Include in comment that this took over 30000 heartbeats prior to #2478.

tests/lean/run/structWith_scott.lean
```diff
@@ -0,0 +1,1322 @@
+/- This file reproduces in pure Lean a Mathlib timeout occuring in typeclass synthesis when the
```
semorrison, line 1: Please rename to something without my name. :-)

tests/lean/run/structWith_scott.lean
```diff
@@ -0,0 +1,1322 @@
...
+structure elaboration uses an extra eta expansion at the top level lean4#2451 -/
```
semorrison, line 2: Hard to understand the intent when the comment just ends with a link. "As reported in"?

tests/lean/run/structInstFast.lean
```diff
@@ -0,0 +1,68 @@
...
+macro "deep_wide_struct_inst_with" type:ident val:ident depth:num width:num : command =>
+  mkStructAndInstStx type val width.getNat depth.getNat
+
+def foo : String := "foo"
+
+deep_wide_struct_inst_with String foo 50 20
+
+/- Structure instances using the `with` pattern should remain fast -/
+set_option maxHeartbeats 200 in
+example : a0 = a'0 := rfl
```
semorrison, line 68: Shouldn't this be `a19 = a'19`?

tests/lean/run/structInstFast.lean
```diff
@@ -0,0 +1,68 @@
...
+/-- We build a structure with `depth` nested structures and `width` fields at each level.
+    The structure is initialized with `val` at every field using the `with` construction pattern. -/
+macro "deep_wide_struct_inst_with" type:ident val:ident depth:num width:num : command =>
+  mkStructAndInstStx type val width.getNat depth.getNat
+
+def foo : String := "foo"
+
+deep_wide_struct_inst_with String foo 50 20
+
+/- Structure instances using the `with` pattern should remain fast -/
```
semorrison, line 66: Please clarify exactly what this is testing, in particular what "remain" means (are you testing that this PR doesn't make something worse, or guarding against future regressions).

semorrison: If it is guarding against future regressions, did this benchmark change in this PR?

tests/lean/structInstExtraEta.lean
```diff
@@ -0,0 +1,8 @@
+structure A where
```
semorrison, line 1: Please include a link to the issue and/or PR in this file.

src/Lean/Elab/StructInst.lean
```diff
@@ -505,8 +505,12 @@ mutual
             return { ref, lhs := [FieldLHS.fieldName ref fieldName], val := val } :: fields
           match Lean.isSubobjectField? env s.structName fieldName with
           | some substructName =>
-            -- If one of the sources has the subobject field, use it
-            if let some val ← s.source.explicit.findSomeM? fun source => mkProjStx? source.stx source.structName fieldName then
+            -- If src is a term for a parent field and the field is that parent projection, use it
+            -- This is necessary as findField? `A `toA returns none hence so will mkProjStx?
```
semorrison, line 509: backticks broken in this comment?

semorrison, line 509: Please add comments here noting the performance issue and explaining how this fix resolves it. Reading the comments here you wouldn't know that there was a performance issue to be addressed!

src/Lean/Elab/StructInst.lean
```diff
@@ -505,8 +505,12 @@ mutual
...
+            if let some stx := s.source.explicit.find? (·.structName == substructName)|>.map (·.stx) then
```
semorrison, line 510:
```suggestion
            if let some stx := s.source.explicit.find? (·.structName == substructName) |>.map (·.stx) then
```

semorrison:
awaiting-author

tests/lean/run/structInstFast.lean
mattrobball: It improves and guards against future regression. Though it only applies to the `with` pattern. It might be good to write another test separately for other patterns. 

Documentation improved.

mattrobball:
Thanks for the comments. Hopefully, they are addressed with these changes. 

awaiting-review

mattrobball:
awaiting author

mattrobball:
awaiting-author

tests/lean/run/structInstFast.lean
```diff
@@ -0,0 +1,95 @@
+/- This test uses a macro to generate a structure with `depth` nested structures and `width` fields
```
semorrison, line 1:
```suggestion
/-! This test uses a macro to generate a structure with `depth` nested structures and `width` fields
```

tests/lean/structInstExtraEta.lean
```diff
@@ -0,0 +1,11 @@
+/- This test assures that that `toA` field is not eta expanded using the `with` pattern as it was
```
semorrison, line 1:
```suggestion
/-! This test assures that that `toA` field is not eta expanded using the `with` pattern as it was
```

tests/lean/run/structInstFast.lean
semorrison: Thanks, new documentation is good.

semorrison:
Can you please rebase on master so we get a mathlib-compatible release to try out?

semorrison:
I'm waiting on this until we have the `builds-mathlib` label (ideally with a diff in the `lean-pr-testing-2478` branch that I can understand and explain).

mattrobball:
I am not sure whether the CI benchmarks the mathlib performance so I did [here](http://speed.lean-fro.org/mathlib4/compare/17111f65-044f-42f6-8b88-a9c44a8039f8/to/1275d3de-97e6-4f02-95e6-1f380bf3f946). 

The top line is an overall reduction of ~4.5% on total CPU instructions.

mattrobball:
Comments on the diff/performance: 
- Regressions are for the most part localized to `AlgebraicGeometry`. This is due to the poor API boundaries in that area. In particular, `erw` are ubiquitous. AG has to be fixed regardless of this PR. 
- The other main source of changes to build mathlib are failure of `Pi.smul_apply` to fire in `simp`. `rw [Pi.smul_apply]` still works fine. This difference in behavior between `simp` and `rw` is confusing (to me) in itself but is common in mathlib now. 
- For some unknown reason, compilation jumps up in the `Analysis` in some files leanprover-community/mathlib4#7103 generally. This is the cause in increase in overall compilation and for `Analysis.NormedSpace.OperatorNorm`.

mattrobball:
awaiting-review

semorrison:
I've left some comments on the matching Mathlib bump https://github.com/leanprover-community/mathlib4/pull/7257. That needs further cleanup before we can merge this:
* every change that can be made separately (i.e. all the minor speedups in AlgebraicGeometry) need to happen in a separate PR
* changes that can't be made separately need further explanation/investigation.

mattrobball:
I am not sure what happened but I did not close this PR.

mattrobball:
Reopening

leodemoura:
I see `.github/workflows/ci.yml` and `.github/workflows/nix-ci.yml` in the list of modified files.

mattrobball:
Cleaned up.

tydeu:
@mattrobball You appear to have accidently committed `src/lake/lakefile.olean` to this branch.

semorrison:
There's no associated mathlib PR for this at the moment, right?

tests/lean/run/structInstFast.lean
```diff
@@ -0,0 +1,95 @@
...
+      def base : Base := ⟨val, ..., val⟩
+    ```
+  The test is then
+  ```
+    set_option maxHeartbeats 200 in
+    example : a0 = a'0 := rfl
+  ```
+  As this type of unification occurs over and over and over when type checking declarations, it needs
+  to be and stay fast. Before #2478, this test took over 700 heartbeats.
+-/
```
eric-wieser, line 26: I think the weird indents here are confusing the syntax highlighter; can you unindent everything?

tests/lean/structInstExtraEta.lean
```diff
@@ -0,0 +1,11 @@
+/-! This test assures that that `toA` field is not eta expanded using the `with` pattern as it was
+  before #2478 -/
+
+structure A where
+
+structure B extends A
```
eric-wieser, line 6: Can you add a test with `structure C extends B` using `a with` too?

Whether the behavior has changed is less important than recording what the behavior actually is through the test.

tests/lean/structInstExtraEta.lean
```diff
@@ -0,0 +1,16 @@
...
+
+def a : A := sorry
+
+def b : B := { a with }
+#print b
+
+structure C extends B
+
+def c : C := { b with }
+#print c
```
eric-wieser, line 16: What I specifically want to see is `def c : C := { a with }`, though there's certainly no harm in having both tests

tests/lean/structInstExtraEta.lean
```diff
@@ -0,0 +1,16 @@
+/-! This test assures that that `toA` field is not eta expanded using the `with` pattern as it was
+  before #2478 -/
+
+structure A where
+
+structure B extends A
+
+def a : A := sorry
```
eric-wieser, line 8:
```suggestion
def a : A := {}
```
would eliminate the sorry

tests/lean/structInstExtraEta.lean.expected.out
```diff
@@ -2,3 +2,6 @@ structInstExtraEta.lean:8:4-8:5: warning: declaration uses 'sorry'
 def b : B :=
 let src := a;
 { toA := src }
+def c : C :=
+let src := a;
+{ toB := { toA := src } }
```
eric-wieser, line 7: Oh nice, I thought you said you hadn't implemented this!

tests/lean/structInstExtraEta.lean
```diff
@@ -0,0 +1,16 @@
...
+
+structure C extends B
+
+def a : A := sorry
+
+def b : B := { a with }
+#print b
+
+def c : C := { a with }
+#print c
```
eric-wieser, line 16: One more test I think would be useful is
```lean
structure D extends B

def d : D  := { c with }
#print d
```

mattrobball: That’s what you get.

eric-wieser (commented):
Whoops, I never submitted these suggestions:

src/Lean/Elab/StructInst.lean
```diff
@@ -505,8 +505,16 @@ mutual
             return { ref, lhs := [FieldLHS.fieldName ref fieldName], val := val } :: fields
           match Lean.isSubobjectField? env s.structName fieldName with
           | some substructName =>
-            -- If one of the sources has the subobject field, use it
-            if let some val ← s.source.explicit.findSomeM? fun source => mkProjStx? source.stx source.structName fieldName then
+            -- If src is a term for a parent field and the field is that parent projection, use it
+            -- This is necessary as `findField? `A `toA` returns none hence so will `mkProjStx?`
```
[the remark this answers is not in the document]
mattrobball: Done

src/Lean/Elab/StructInst.lean
```diff
@@ -505,8 +505,16 @@ mutual
...
+            -- The resulting extra eta expansion has a surprising effect on performance. Without
+            -- this change, the terms resulting from elaboration are larger. Worse, unification
+            -- must pull them apart when often, without the eta expansion, that would not be
+            -- necessary. This change occurred in #2478.
+            if let some stx := s.source.explicit.find? (·.structName == substructName) |>.map (·.stx) then
+              addField (FieldVal.term stx)
```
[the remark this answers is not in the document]
mattrobball: Done

tests/lean/run/structInstFast.lean
mattrobball: Done

eric-wieser (approved):

mattrobball:
awaiting-review

kbuzzard:
Note the [substantial speedup](https://github.com/leanprover-community/mathlib4/pull/9007#issuecomment-1856575019) to mathlib which comes with this PR.

tests/lean/run/463.lean
```diff
@@ -4,4 +4,4 @@ structure C := (a b c : Nat)
 structure D := (toA : A) (c : Nat)
 
 def foo (s : C) : B := {s with} -- works in lean 4, works in lean 3
-def bar (s : D) : B := {s with} -- works in lean 4, fails in lean 3
+-- def bar (s : D) : B := {s with} -- works in lean 4, fails in lean 3
```
kmill, line 7: Correct me if I'm wrong, you don't need to comment this out, right?

You can add a comment that this is not necessarily meant to succeed.

kmill (approved):
Thanks for all the work getting this to this point! I'm looking forward to seeing this merged.

It looks good to me. I see it's not rebased onto the most current `nightly`, so please do that to be sure the CI goes through.

tests/lean/run/463.lean
mattrobball: Doesn’t it fail a test because this errors?

PR #2480, opened 2023-08-29
feat: lake: save elaborated config as an olean

tydeu:
Local benchmarks indicate that this does not have much effect on simple configurations, but has a major impact on complex ones. For instance, using an prebuilt OLean gave a ~5x configuration speedup for the FFI example on my machine.  Essentially, it makes configure time near constant, rather than directly proportional to the complexity of the configuration.

The limiting factor for simple configurations appears to be that the initial import of `Lake` is still expensive.


tydeu:
While I would like a review, merging should wait until this is tested against benchmarks like #2457.

src/lake/Lake/Load/Elab.lean
```diff
@@ -65,3 +76,30 @@ def elabConfigFile (pkgDir : FilePath) (configOpts : NameMap String)
...
+    let s ← importModulesUsingCache mod.imports
+    let s := {s with
+      moduleData  := s.moduleData.push mod
+      regions     := s.regions.push region
+      moduleNames := s.moduleNames.push configModuleName
+    }
+    finalizeImport s #[configModuleName] leanOpts 1024
+  else
+    let env ← elabConfigFile pkgDir lakeOpts leanOpts configFile
+    Lean.writeModule env olean
```
digama0, line 104: I don't like that this is putting `lakefile.olean`, a build file, outside the `build` directory. My suggestion is to put it in `build/lakefile.olean`.

digama0 (approved):

src/lake/Lake/Load/Elab.lean
tydeu: @digama0 I am incline to think of `lakefile.olean` as more of static lock file like `lake-manifest.json` than a build file. The only major difference is that `lakefile.olean` is a platform/configuration-specific lock, hence its exclusion from Git. Also, a technical problem with putting it in the build directory is that Lake currently does not know *what the build directory is* until we load the configuration file.

tydeu: @digama0 When you say it should be supplied "by the user", do you mean a CLI option? Note that, in addition tot he environment variable, Cargo also allows you to specify the build directory via the configuration file with the [`build.target-dir`](https://doc.rust-lang.org/cargo/reference/config.html#buildtarget-dir) setting.

digama0: A CLI option would be fine, as would a config file (but that would be a significant new feature, here "config file" means a global or project-local setting which would generally not be committed). The environment variable is the lowest effort version of this which still gives the basic feature without all the issues involved in tying it up in lean elaboration.

tydeu: @digama0 For Lake, the elaborated configuration *is* the project-local "config file". I do not envision the creation of any other. If it could be reasonable stuck in a such a thing, it should be part of the lakefile then (as it currently is). A standard CLI override does sound like a good idea for a future feature though!

tydeu: One caveat to my statement about the creation of another static config file: I can imagine creating declarative config file after an elaboration that contains some subset of the olean data that can be declaratively represented.

digama0: The "config file" here is for things specific to the user's machine. It would normally just be global and in the user's home directory, but cargo also allows you to customize this on a per-subfolder (including per-project) basis. This is a very different thing from the lakefile / Cargo.toml, which is about properties of the project itself, and lake currently has almost no need for such a thing.

tydeu:
!bench

tydeu:
Comparing this benchmark with the plain one in https://github.com/leanprover/lean4/pull/2457#issuecomment-1699782763 displays the [expected significant improvement](http://speed.lean-fro.org/lean4/compare/bab50d33-13a1-40c4-85a0-fda5f93f8b22/to/df988ac1-4b34-47f8-8285-051ef5d51448?hash1=8b5a0230052e4afdcff85c965741a4380c782219) to configuration time. From the comparison:

```diff
  Benchmark         Metric         Change
  =================================================
+ lake build no-op  branch-misses  -12.1% (-16.4 σ)
+ lake build no-op  branches       -9.3% (-35.4 σ)
+ lake build no-op  instructions   -7.5% (-28.2 σ)
+ lake configure    branch-misses  -79.5% (-217.3 σ)
+ lake configure    branches       -70.7% (-990.4 σ)
+ lake configure    instructions   -70.9% (-783.7 σ)
+ lake configure    maxrss         -7.5% (-16.5 σ)
+ lake configure    task-clock     -52.4% (-34.7 σ)
+ lake configure    wall-clock     -52.6% (-35σ)
```

semorrison:
On Mathlib master, `lake env` currently takes `1.72s`. With `leanprover/lean4-pr-releases/pr-release-2480`, without a `lakefile.olean` present `lake env` takes `1.05s`, or `0.91s` with the `lakefile.lean`.

tydeu:
@semorrison

> On Mathlib master, `lake env` currently takes `1.72s`.
> without a `lakefile.olean` present `lake env` takes `1.05s`
> with a lakefile.olean present `lake env` takes `0.91s`

This is surprising. I am not sure why the configuration without the olean would have gotten significantly faster. Similarly, I am surprised the olean only gave an ~10% boost.

When you ran the test without the olean, did you run it in fresh repository or did you just delete the `lakeifle.olean`? If the latter, the difference may be due to the `lakefile.olean`  files still remaining in depedencies (e.g., `proofwidgets`).

semorrison:
That was it. Now I get (averaging 10 runs each):

* pre #2480, `lake env` takes `1.72s`
* post #2480, `lake env -R` takes `2.15s`
* post #2480, `lake env` takes `0.89s`

Is the slowdown with `-R` expected?

tydeu:
@semorrison Yes, Lake now just memoizes the import state, not the full environment (because that environment cannot be used by the cached oleans), so there is some expected overhead from recreating it for each dependency.

tydeu:
@semorrison One thing I am curious about from your numbers. How fast is `lake self-check` on the machine you were running these benchmarks on? That gives a good startup baseline for Lake to see how much room there is to improve.

src/Lean/Environment.lean
```diff
@@ -731,73 +733,77 @@ def throwAlreadyImported (s : ImportState) (const2ModIdx : HashMap Name ModuleId
...
-      }
-      importMods is
+    let (_, s) ← importModulesCore imports |>.run
+    finalizeImport s imports opts trustLevel
 
 /--
   Create environment object from imports and free compacted regions after calling `act`. No live references to the
   environment object or imported objects may exist after `act` finishes. -/
-unsafe def withImportModules {α : Type} (imports : List Import) (opts : Options) (trustLevel : UInt32 := 0) (x : Environment → IO α) : IO α := do
+unsafe def withImportModules {α : Type} (imports : Array Import) (opts : Options) (trustLevel : UInt32 := 0) (x : Environment → IO α) : IO α := do
```
semorrison, line 806: This is used in downstream projects, including Mathlib, so probably this signature change warrants recording in `RELEASES.md`.

semorrison: (The existence of `lakefile.olean` should be noted too, I guess, as projects will need to .gitignore it.)

semorrison:
`lake self-check` is taking `0.08s`.

src/Lean/Environment.lean
tydeu: @semorrison Are we going back to updating `RELEASE.md` for most changes? That seemed to have been abandoned earlier this year. (There are a lot of missing breaking changes from the last few months in `v4.0.0` for instance.) If so, it should also be more generally updated with some of key changes to Lake (i.e., the bump to v5.0.0 and merge with core along with this suite of performance fixes and the `lake update` change). As such, it may be better to have a separate PR with a general update.

tydeu:
@semorrison 

> `lake self-check` is taking `0.08s`.

Wow! That is also a much greater difference between startup and configure times than [the speed center benchmark](http://speed.lean-fro.org/lean4/run-detail/df988ac1-4b34-47f8-8285-051ef5d51448). As such, I would curious in a more fine-grained profile of the Lake's performance in mathlib after merging all these performance improvements. That is, using something like [Hotspot](https://github.com/KDAB/hotspot) (like [Sebastian's previous analysis](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/running.20.60cache.20get.60.20as.20part.20of.20every.20.60lake.20build.60/near/385841846)).

src/Lean/Environment.lean
semorrison: Let's talk about this on zulip first, and sort out a plan.

tydeu: @semorrison Do you think the PR should wait until that is decided or would the resolution come in a follow-up PR?

PR #2485, opened 2023-08-30
fix: revert #2435

kim-em:
Reverts leanprover/lean4#2435, which thoroughly broke Mathlib, because `change` statements are now duplicating goals. See the build output at https://github.com/leanprover-community/mathlib4/actions/runs/6021210761/job/16333708649?pr=6863.

I am still working on automatically posting reports on Mathlib builds back to Lean PRs, which would have caught this, but not there yet. What we do have is automatically generated toolchains for releases, e.g. as `lean/leanprover-pr-releases:pr-release-XXXX`, so at least it is already possible to manually test.

No fault of @thorimur's, but this does demonstrate how our testing library still falls far short. This scenario of `change` breaking so badly should have been caught well before Mathlib.

@thorimur, I am going to revert your PR for now, and leave it to you to reconsider how to do this. It may be best that we write some proper tests for `change` first!


digama0:
But `change` isn't implemented...? Are you talking about the `change` conv?

thorimur:
Oh dear, very sorry to have caused this nonetheless!

If I PR core or std4 in the future, I'll be sure to build Mathlib to make sure that only expected breakages occur. I look forward to the automatically posted build reports, however, as a full Mathlib build is quite a bit for my laptop to handle! :) (Though I suppose I could convince CI to do it on a branch/PR.)

I'll start investigating this tomorrow. Will the PR release for #2435 still work? (I've briefly tried `elan install leanprover/lean4-pr-releases:pr-release-2435`, but am met with an error.)

Also: can/should #2435 be reopened, or should I make a new one (possibly with a new branch)?

thorimur:
Couldn't resist taking a peek: what's happening is that `change` makes an intermediate fresh natural mvar to simulate a goal when elaborating a local hypothesis. Nothing actually depends on this mvar, so if you simply throw it away—as you might when throwing away all natural mvars!—you don't get any kernel error. As of 2435, we no longer throw it away, and it appears as a spurious unsolved goal. This shouldn't be difficult to fix. I'll try to open a PR in 12 hours or so. :)

semorrison:
> But `change` isn't implemented...? Are you talking about the `change` conv? The implementation of the `change` tactic is in `Mathlib.Tactic.Basic`.

Ah, okay, indeed that makes it harder to catch in Lean itself, and I feel less bad about this. With this toolchain many of the uses of `change` in mathlib are duplicating goals.

PR #2487, opened 2023-08-30
fix: simpGoal reports incomplete UsedSimps

### LINKED ISSUE #2486: simpGoal reports incomplete UsedSimps
### Prerequisites

### Description

`simpGoal` does not pass the `UsedSimps` from the hypotheses to the method that simplifies the target. As a result, `simp? at *` does not report simp theorems that were only used to simplify the hypotheses.

### Steps to Reproduce

Sort of minimal working example (based on the `simp_trace` test):

```lean
def h (x : Nat) : Sum (Nat × Nat) Nat := Sum.inl (x, x)

def bla (x : Nat) :=
  match h x with
  | Sum.inl (y, z) => y + z
  | Sum.inr _ => 0

example (h' : bla x = x) : x + x = x := by
  set_option tactic.simp.trace true in
  simp [bla, h] at *
  -- Expected: Try this: simp only [bla, h] at *
  -- Actual:   Try this: simp only at *
  exact h'
```

### Versions

Current master.

JLimperg:

Fixes #2486

### DIFF
--- a/tests/lean/simp_trace.lean
+++ b/tests/lean/simp_trace.lean
@@ -81,3 +81,35 @@ example : a ∧ (b ∧ b) ↔ a ∧ b := by simp [my_thm]
 example : (a ∧ (b ∧ b)) = (a ∧ b) := by simp only [my_thm]
 
 example : x - 1 + 1 = x := by simp (discharger := sorry) [Nat.sub_add_cancel]
+
+-- The following examples test simplification at hypotheses.
+
+-- Two simp lemmas applied to one hypothesis.
+example (h' : bla x = x) : x + x = x := by
+  simp [bla, h] at *
+  exact h'
+
+-- Ditto, but simplifying the hypothesis explicitly.
+example (h' : bla x = x) : x + x = x := by
+  simp [bla, h] at h'
+  exact h'
+
+-- Various simp lemmas applied to different hypotheses, but each lemma is
+-- applied to exactly one hypothesis.
+example {α : Type} (xs ys : List α) (h₁ : bla x = y) (h₂ : (xs ++ ys).length = y) : x = length xs := by
+  simp [bla, h, List.length_append] at *
+
+-- Ditto, but with an additional unused lemma.
+example {α : Type} (xs ys : List α) (h₁ : bla x = y) (h₂ : (xs ++ ys).length = y) : x = length xs := by
+  simp [bla, h, List.length_append, Nat.add_one] at *
+
+-- Two simp lemmas applied to two hypotheses, with each lemma applied to both
+-- hypotheses.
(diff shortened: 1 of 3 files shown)


semorrison:
I'd earlier made #2333, but didn't finish writing tests. I'll close that in favour of this.

tests/lean/simp_trace.lean
```diff
@@ -81,3 +81,7 @@ example : a ∧ (b ∧ b) ↔ a ∧ b := by simp [my_thm]
 example : (a ∧ (b ∧ b)) = (a ∧ b) := by simp only [my_thm]
 
 example : x - 1 + 1 = x := by simp (discharger := sorry) [Nat.sub_add_cancel]
+
+example (h' : bla x = x) : x + x = x := by
+  simp [bla, h] at *
+  exact h'
```
semorrison, line 87: To be thorough, could we also have the tests here at use `at h'`, and another that uses `at *` in which there are actually multiple hypotheses that get simplified, and different simp lemmas are used at each hypothesis?

JLimperg: Done. Thanks for the review!

JLimperg:
awaiting-review

semorrison:
Two more requests for tests:
* a `simp at *` where a lemma is used at two hypotheses, and check it is only reported once
* similarly where is used at a hypothesis and a goal
* perhaps already covered, but one where a provided simp lemma in the `[...]` arguments is not used at all.

JLimperg:
Also done. Btw, I believe we could use your mass-testing code to check whether there are any remaining issues. I have code that can check whether two tactics generate the same goals, so we could run every `simp` invocation in mathlib as `simp?` and check whether the generated `Try this` suggestion generates the same goals.

JLimperg:
awaiting-review

semorrison (approved):

tydeu:
@semorrison The mathlib bot appears to have built this. I presume the absence of a bot commit on the status is due to this occur between changes to the CI?

semorrison:
Yes, the "send the result back to the Lean4 PR" commit was only merged to Mathlib master after this.

semorrison:
I had not realised that the "added a commit" message was going to be automatically generated (because I put the URL for this PR in the commit message!). It occurs to me now that this may be sufficient feedback, and there is no need to write comments or use labels. Let's see how this pans out over the next few days.

tydeu:
@semorrison 

> It occurs to me now that this may be sufficient feedback, and there is no need to write comments or use labels.

The labels are still useful regardless when looking through the issues on the issue list. While the comment is less useful,, it does importantly produce a notification after the build succeeds or fails whereas the commit reference produces none.

PR #2489, opened 2023-08-30
fix: simp: include class projections in UsedSimps

### LINKED ISSUE #2488: Incorrect simp tracing of class methods
### Prerequisites

### Description

It appears that when `simp` unfolds a class method, this is not recorded in the `UsedSimps`. As a result, the `simp only` call reported by `simp?` does not behave like the original `simp` call.

### Steps to Reproduce

MWE:

```lean
set_option tactic.simp.trace true

class HasProp (A) where
  toProp : A → Prop

export HasProp (toProp)

instance : HasProp Nat where
  toProp _ := True

example : toProp 0 := by
  simp [toProp]
  -- Expected: Try this: simp only [toProp]
  -- Actual:   Try this: simp only

-- # Working as intended

-- ## Structure methods

structure HasProp₂ (A) where
  toProp₂ : A → Prop

export HasProp₂ (toProp₂)

def hasProp₂_Nat : HasProp₂ Nat where
  toProp₂ _ := True

example : toProp₂ hasProp₂_Nat 0 := by
  simp only [hasProp₂_Nat]
  -- Try this: simp only [hasProp₂_Nat]

-- ## Regular defs

def toProp₃ (_ : Nat) : Prop := True

example : toProp₃ 0 := by
  simp [toProp₃]
  -- Try this: simp only [toProp₃]

-- ## Theorems about class methods

theorem toProp_def (n : Nat) : toProp n = True := rfl

example : toProp 0 := by
  simp [toProp_def]
  -- Try this: simp only [toProp_def]
```

### Versions

Current master.

JLimperg:

Fixes #2488

The changed test case in `simp_trace.lean` was also affected by this issue: the given `simp` call solves the goal, but the `Try this` suggestion did not. With this PR, it does.


tests/lean/simp_trace.lean.expected.out
```diff
@@ -19,7 +19,8 @@ Try this: simp only [g, Id.pure_eq]
       (let x := x;
       pure x)
 [Meta.Tactic.simp.rewrite] @Id.pure_eq:1000, pure x ==> x
-Try this: simp only [f1, modify, StateT.modifyGet, f2, StateT.bind, getThe, StateT.get, StateT.set]
+Try this: simp only [f1, modify, modifyGet, MonadStateOf.modifyGet, StateT.modifyGet, pure, f2, bind, StateT.bind, get,
+  getThe, MonadStateOf.get, StateT.get, set, StateT.set]
```
semorrison, line 23: Let's put a comment in the source for this one noting that prior to #2489 the `simp only` suggestion was incomplete, just in case there is ever a regression here and someone is tempted to revert this change in `.expected.out`.

Otherwise I think this is good to go!

JLimperg: Done. Thanks again!

JLimperg:
This PR will generate a merge conflict with #2486 because the expected test output for contains line numbers.

semorrison (approved):

JLimperg:
Merge conflict resolved.

PR #2497, opened 2023-08-31
fix: 0-arg functions in C need f(void)

digama0:
The syntax `T foo();` in C to this day still means "foo takes an arbitrary number of arguments" rather than "foo takes no arguments". This is a minor fix but it should be easy to review. I noticed this when reviewing the results of automatic translation into Zig.


Kha:
Oh my gosh, I had no idea. While there should not be any effect to Lean itself, this LGTM.

PR #2499, opened 2023-08-31
doc: document all parser aliases

digama0:
Our grammar is still very under-documented, even though we have a nice hover mechanism that could be showing grammar docs. This is the first step in hopefully a series of docs to improve this situation, starting with parser aliases. (This isn't a complete list of things that can show up in a `syntax` declaration, but it's harder to prioritize the arbitrary parsers that people could be using or encountering.)

### DIFF
--- a/src/Lean/Parser/Term.lean
+++ b/src/Lean/Parser/Term.lean
@@ -19,7 +19,7 @@ def commentBody.parenthesizer := PrettyPrinter.Parenthesizer.visitToken
 def commentBody.formatter := PrettyPrinter.Formatter.visitAtom Name.anonymous
 
 /-- A `docComment` parses a "documentation comment" like `/-- foo -/`. This is not treated like
-a regular comment (that is, as whitespace); it is parsed and forms part of the abstract syntax tree.
+a regular comment (that is, as whitespace); it is parsed and forms part of the syntax tree structure.
 
 A `docComment` node contains a `/--` atom and then the remainder of the comment, `foo -/` in this
 example. Use `TSyntax.getDocString` to extract the body text from a doc string syntax node. -/


tydeu:
LGTM! @Kha, are you fine with it?

src/Lean/Parser/Basic.lean
```diff
@@ -1246,8 +1282,15 @@ def checkColGeFn (errorMsg : String) : ParserFn := fun c s =>
     if pos.column ≥ savedPos.column then s
     else s.mkError errorMsg
 
-def checkColGe (errorMsg : String := "checkColGe") : Parser :=
-  { fn := checkColGeFn errorMsg }
+/-- The `colGe` parser requires that the next token starts at least the column of the saved
```
semorrison, line 1285:
```suggestion
/-- The `colGe` parser requires that the next token starts at at least the column of the saved
```

src/Lean/Parser/Term.lean
```diff
@@ -251,6 +278,12 @@ and solved by typeclass inference of the specified class.
 -/
 def instBinder := ppGroup <| leading_parser
   "[" >> withoutPosition (optIdent >> termParser) >> "]"
+/-- A `bracketedBinder` matches any kind of binder group with an explicit bracketing syntax:
```
semorrison, line 281: What work is the word "explicit" doing here? It is slightly confusing because on the next line "explicit" of course means something else.

semorrison (approved):

src/Lean/Parser/Basic.lean
tydeu: To avoid the double `at`:

```suggestion
/-- The `colGe` parser requires that the next token starts from at least the column of the saved
```

src/Lean/Parser/Term.lean
digama0: It is being contrasted against one of the other binder syntax categories that also allows plain identifiers (e.g. in `fun x => ...`), which is an unbracketed binder group of size 1. I think of it as similar to the `rw [e1, e2]` vs `rw e1` situation in lean 3, the brackets are optional if there is only one thing in the group.

tydeu, line 281: @digama0 

In that case, what do you think of this wording?

```suggestion
/-- A `bracketedBinder` matches any kind of binder group that uses brackets:
```

tydeu: @digama0 LGTM! However, I am little surprised @semorrison (an Australian) agreed with that default because, according to my knowledge and the [Wikipedia page](https://en.wikipedia.org/wiki/Bracket), defaulting the general word bracket to square brackets (i.e., `[]`) is only standard in the US.

src/Lean/Parser/Term.lean
```diff
@@ -18,6 +18,11 @@ def commentBody.parenthesizer := PrettyPrinter.Parenthesizer.visitToken
 @[combinator_formatter commentBody]
 def commentBody.formatter := PrettyPrinter.Formatter.visitAtom Name.anonymous
 
+/-- A `docComment` parses a "documentation comment" like `/-- foo -/`. This is not treated like
+a regular comment (that is, as whitespace); it is parsed and forms part of the abstract syntax tree.
```
Kha, line 22:
```suggestion
a regular comment (that is, as whitespace); it is parsed and forms part of the syntax tree structure.
```

PR #2502, opened 2023-09-01
fix: only return new mvars from `refine`, `elabTermWithHoles`, and `withCollectingNewGoalsFrom`

### LINKED ISSUE #2495: bug: `refine` duplicates goals in the infoview
### Prerequisites

### Description

If `refine` is used on an expression which includes an existing goal, then that goal will appear twice in the resulting tactic state. Example:
```lean
example : True := by
  have : True := ?a
  refine ?a -- two `case a`s in the infoview
```

### Steps to Reproduce

1. Create a side goal.
2. `refine` a term which elaborates to include that goal.

**Expected behavior:** The goal appears only once in the infoview.

**Actual behavior:** The goal appears twice in the infoview.

### Versions

4.0.0-rc4, macOS Ventura 13.4.1 (Intel-based)

### Additional Information

This occurs because `withCollectingNewGoalsFrom` returns all mvars encountered in the elaborated expression, and `refineCore` simply replaces the main goal with a list of all such goals, even if they exist on the goal list already. There are multiple ways to resolve this; some options are discussed [on zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Bug.20in.20.60refine.60/near/388277583).

thorimur:
Fixes #2495 and #2434 by making sure that `withCollectingNewGoals` only returns mvarIds created during elaboration. This changes the behavior of `elabTermWithHoles` and `refine` accordingly: `refine e` now replaces the main goal only with goals that were created during elaboration of `e` (and does not include goals that appear in `e` but existed prior to its elaboration).

Consequences of this change:

* This prevents duplication of goals in the infoview by `refine` (#2495): pre-existing goals are now excluded from the resulting tactic state, and thus don't appear twice.
* It also prevents "goal tunneling", wherein a goal that was created before a given `refine` statement manages to be included in its resulting goals *unexpectedly*, due to being present in e.g. an implicit argument.
* Previously, `refine` was inconsistent in *which* pre-existing goals were returned; this is the content of issue #2434, which points out that pre-existing natural mvars are not returned, while synthetic mvars are. `refine`'s occurs check relied on all mvars appearing in the expression being returned, and so goals were erroneously closed. This renders that issue moot by uniformly not returning any old mvars, and updating the occurs check accordingly.
* This PR breaks mathlib, and requires fixes to a few proofs that use `refine` and `convert` (which relies on `withCollectingNewGoalsFrom`). All breakages are instances of goal tunneling, and are easily fixed. These fixes are done in leanprover-community/mathlib4#6928. Note, however, that it breaks mathlib *less* than an incremental fix of #2434 alone, which required us to filter out old mvars anyway in `change`.
* Finally, metaprogrammers likely expect `elabTermWithHoles` and `withCollectingNewGoalsFrom` to behave in this way already—see [zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Bug.20in.20.60refine.60/near/388422423) for an expression of this sentiment.

Note also that this streamlines the code of `withCollectingNewGoalsFrom`, replacing
```lean
    let newMVarIds ← if allowNaturalHoles then
      pure newMVarIds.toList
    else
      let naturalMVarIds ← newMVarIds.filterM fun mvarId => return (← mvarId.getKind).isNatural
      let syntheticMVarIds ← newMVarIds.filterM fun mvarId => return !(← mvarId.getKind).isNatural
      let naturalMVarIds ← filterOldMVars naturalMVarIds mvarCounterSaved
      logUnassignedAndAbort naturalMVarIds
      pure syntheticMVarIds.toList
```
with simply
```lean
    let newMVarIds ← filterOldMVars newMVarIds mvarCounterSaved
    unless allowNaturalHoles do
      let naturalMVarIds ← newMVarIds.filterM fun mvarId => return (← mvarId.getKind).isNatural
      logUnassignedAndAbort naturalMVarIds
```
(plus comments).

### Changes made by this PR

* prevents `withCollectingNewGoalsFrom`, and therefore `elabTermWithHoles`, `refine`, and `specialize` from returning old goals

* updates the occurs check in `refine` to compensate for this change (note that `specialize` does not have an occurs check; this is a separate issue)

* adds a missing docstring to `elabTermWithHoles`

* adds tests for #2495, goal tunneling, and #2434

* changes test `run/492.lean`, which made incidental use of goal tunneling

-----


thorimur:
WIP

thorimur:
awaiting-review

semorrison:
Could you merge your Mathlib fixes to the `lean-pr-testing-2502` branch, so we can hopefully get an automatically generated `builds-mathlib` label on this PR?

tests/lean/refineFiltersOldMVars.lean
```diff
@@ -0,0 +1,97 @@
...
+  x := _ + 1
+  y := by
+    refine' neq3 -- captures the underscore above
+    exact 0 -- solves it
+}
+
+/-!
+# Issue #2434
+
+Resolving this issue also resolves issue #2434, which identified an inconsistency in *which* pre-existing mvars were being preserved—namely, `refine` would include old pre-existing syntheticOpaque mvars, but not old natural mvars. This would erroneously close goals. Here, we demonstrate that uniformly filtering out old mvars also resolves that issue.
```
semorrison, line 63: Very long line!

semorrison, line 63:
```suggestion
Resolving this issue also resolves issue #2434, which identified an inconsistency in *which* pre-existing mvars were being preserved --- namely, `refine` would include old pre-existing syntheticOpaque mvars, but not old natural mvars. This would erroneously close goals. Here, we demonstrate that uniformly filtering out old mvars also resolves that issue.
```

semorrison: (I see you used a unicode em-dash, but in practice it's unreadable. :-)

tests/lean/refineFiltersOldMVars.lean
```diff
@@ -0,0 +1,97 @@
...
+open Lean Meta Elab Tactic Term in
+/-- `add_natural_goal a : A` adds a new natural goal of type `A` named `a` to the end of the tactic
+state. -/
+elab "add_natural_goal" s:ident " : " t:term : tactic => do
+  let g ← mkFreshExprMVar (← elabType t) .natural s.getId
+  appendGoals [g.mvarId!]
+
+/-!
+In the following, `refine` would erroneously close each focused goal, leading to a
+`(kernel) declaration has metavariables '_example'` error.
```
semorrison, line 75:
```suggestion
In the following, prior to lean4#2502 `refine` would erroneously close each focused goal, leading to a
`(kernel) declaration has metavariables '_example'` error.
```

tests/lean/refineFiltersOldMVars.lean
thorimur: Oops—the perils of VS code word wrap! :)

tests/lean/refineFiltersOldMVars.lean
thorimur: Fixed in 3d65dad.

semorrison (approved):

PR #2516, opened 2023-09-06
fix: allow setting version flags for cmake

kim-em:
Previously options such as `cmake -DLEAN_VERSION_IS_RELEASE=1` were being dropped because of the way `CMakeLists.txt` was set up.


tydeu:
@semorrison  I suspect the build failure is due to the problem I noticed on Windows where top-level CMake arguments do not propagate to the various stages' configurations.

semorrison:
Not working, closing this while I get help on driving `cmake`.

PR #2521, opened 2023-09-08
chore: assorted lake touchups

### LINKED ISSUE #2517: `lake new hello` should create a `xyz/Hello/` as well as a `xyz/Hello.lean`
Users don't know where they are meant to put their files, and are not going to read the documentation. `lake new` should set up a new project using the standard conventions.

Ideally there would even be a `xyz/Hello/Example.lean`, and `xyz/Hello.lean` would both import it, *and* have a comment explaining that you should import all new files there.

### LINKED ISSUE #2518: `lake +xyz new foo` should use `xyz` in `foo/lean-toolchain`
Currently if `xyz` here is a PR toolchain (i.e. of the form `leanprover/lean4-pr-releases:pr-release-NNNN`) then `lake` doesn't produce any `lean-toolchain` file.

1. Any condition that results in`lake new` not producing a `lean-toolchain` should prominently warn the user that it has produced an unusable project, explaining why.

2. This failure is because `lake` relies on `Lean.toolchain` being set, but setting `Lean.toolchain` in PR releases would be complicated (we can do it if necessary, but it requires adding extra epicycles to the CI). It would be easier if we could just take the user's intention indicated through `+xyz` and give them that toolchain directly.

### LINKED ISSUE #2525: If `lake` detects the manifest is out of date, it should suggest `lake update X` rather than `lake update`.
Now that there is a `lake update X` command to just update one dependency, the automatic suggestion to run `lake update` should specify the relevant dependency.

This will lessen the frequency of users breaking their project by unintentionally updating other dependencies.

### LINKED ISSUE #2540: `lake init Foo` (note initial capital) produces broken `lakefile.lean`
### Prerequisites

### Description

`lake init Foo` (note the initial capital) generates a `lakefile.file`

```
import Lake
open Lake DSL

package «Foo» {
  -- add package configuration options here
}

lean_lib «Foo» {
  -- add library configuration options here
}

@[default_target]
lean_exe «Foo» {
  root := `Main
}
```

which fails at `lean_exe` with `'Foo' has already been declared`.

Apparently `lake` is expecting that the user will know to type `lake init foo` instead. (I didn't.)

Hopefully this can be fixed quickly, either by `lake init` erroring if given a capitalised name, or automatically lower-casing it internally.

Silently producing an unusable `lakefile.lean` in 50% of new user interactions is not a good plan. :-)

### Versions

```
% elan --version
elan 3.0.0 (cdb40bff5 2023-09-08)
% lake --version
Lake version 5.0.0-ec94173 (Lean version 4.0.0)
% lean --version
Lean (version 4.0.0, commit ec941735c80d, Release)
```

tydeu:
Closes #2517. Closes #2518. Closes #2525. Closes #2540.

### DIFF
--- a/src/lake/Lake/Build/Monad.lean
+++ b/src/lake/Lake/Build/Monad.lean
@@ -78,7 +78,7 @@ where
       | e => throw e
 
 /-- The name of the Lake build lock file name (i.e., `lake.lock`). -/
-@[noinline] def lockFileName : String :=
+def lockFileName : String :=
   "lake.lock"
 
 /-- The workspace's build lock file. -/
--- a/src/lake/Lake/Build/Trace.lean
+++ b/src/lake/Lake/Build/Trace.lean
@@ -205,10 +205,6 @@ instance [GetMTime α] : ComputeTrace α IO MTime := ⟨getMTime⟩
 instance : GetMTime FilePath := ⟨getFileMTime⟩
 instance : GetMTime TextFilePath := ⟨(getFileMTime ·.path)⟩
 
-/-- Check if the info's `MTIme` is at least `depMTime`. -/
-@[inline] def checkIfNewer [GetMTime i] (info : i) (depMTime : MTime) : BaseIO Bool :=
-  (return (depMTime < (← getMTime info) : Bool)).catchExceptions fun _ => pure false
-
 --------------------------------------------------------------------------------
 /-! # Lake Build Trace (Hash + MTIme) -/
 --------------------------------------------------------------------------------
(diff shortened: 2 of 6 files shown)


src/lake/Lake/Load/Main.lean
```diff
@@ -148,33 +148,44 @@ downloading and/or updating them as necessary.
...
+        else if topLevel then
+          error <|
+            s!"dependency '{dep.name}' not in manifest; " ++
+            s!"use `lake update {dep.name}` to add it"
+        else
+          error <|
+            s!"dependency '{dep.name}' of '{pkg.name}' not in manifest; " ++
+            s!"check with the developer of {pkg.name} to see if they forgot to update their manifest file," ++
+            s!"or use `lake update` to generate a new, complete file (warning: this will update ALL workspace dependencies)"
+      return {pkg with opaqueDeps := ← depPkgs.mapM (.mk <$> resolve ·)}
```
tydeu, line 188: @semorrison I am curious what you think of my changes of the manifest-related error messages.

PR #2529, opened 2023-09-11
perf: Use flat ByteArrays in Trie

nomeata:
The existing implementation of `Lean.Parser.Trie`, which implements a `String`-valued map,  branches on the _characters_ on the string, and uses a `RBMap Char` at each node.

This is likely unnecessary expensive: For small maps , a compact linear scan tends to be faster than a complicated, nested data structure. And in these kind of tries, the branching factor is typically very small, often even 1.

Therefore, this PR changes the implementation in the following ways:

 * The branching happens on bytes, not chars, for more a compact map
 * Instead of a `RBMap`, it uses a `ByteArray` for the keys (nicely compact), and a `Array a` of the same size for the value. This array is linearly scanned.
 * Such tries often have many nodes of degree 1, so special-casing that case is well worth it.
 * Within `#find` and loogle, I want to use the same data structure, and there I need an “update or insert” operation, so I add `Trie.upsert` here. I believe it allows for in-place updates (including of the stored value).
 * Initially I converted the key to a `ByteArray` and traversed that, but copying the string data seems wasteful, so this adds a way to index into the utf8-encoded bytes of a `String` directly.
 * Because I hope to use this data structure in other projects, this moves its namespace from `Lean.Parser.Trie` to `Lean.Data.Trie`, also matching the module name.

Possible further tweaks:
 * Add `ByteString.findByte` or so that looks for a specific bytes, possibly wrapping the C library’s `memchr`.
 * See if path compression would be an improvement (storing a `ByteArray` instead of nested single-`UInt8`-nodes). Maybe as a follow-up.


nomeata:
WIP

leodemoura:
This is great. Same comment I made for the other PR is valid here too :)
Could you please add tests for the `Trie.lean` module?

Kha:
!bench

nomeata:
Of course, tests and docstrings will come, that’s why it’s still marked WIP :-)

nomeata:
Nice, the benchmarks are in. 3% improvement in parsing – do you think that is worth the extra complexity in the Trie module?
If yes, I’ll polish it (docstrings, tests, maybe some more performance tweaks etc.).
If not, I can move the code into mathlib or loogle.

leodemoura:
> do you think that is worth the extra complexity in the Trie module?

Yes. Thanks for doing it.

nomeata:
awaiting-review

See updated PR description for an overview.

Two questions from my side so far:

 * [ ] Is `upsert` a good name for the “insert or update” with parameter `(Option α → α)`?
 * [ ] Is it ok for `findPrefix (t : Trie α) (pre : String) : Array α` to return the entries in arbitrary order, or must I sort them (either on insertion or in `findPrefix`?) It is currently used to pass completions to LSP, and I don’t know if they will be sorted afterwards.

src/Init/Data/String/Extra.lean
```diff
@@ -33,6 +33,10 @@ opaque fromUTF8Unchecked (a : @& ByteArray) : String
 @[extern "lean_string_to_utf8"]
 opaque toUTF8 (a : @& String) : ByteArray
 
+/-- Accesses a byte in the [UTF-8](https://en.wikipedia.org/wiki/UTF-8) encoding of the `Strig` -/
```
Kha, line 36:
```suggestion
/-- Accesses a byte in the [UTF-8](https://en.wikipedia.org/wiki/UTF-8) encoding of the `String`. O(1). -/
```
Not sure we need the UTF-8 reference here :) . It should probably go in the docstring of `String` instead.

src/Lean/Data/Trie.lean
```diff
@@ -1,110 +1,202 @@
 /-
 Copyright (c) 2018 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Author: Sebastian Ullrich, Leonardo de Moura
+Author: Sebastian Ullrich, Leonardo de Moura, Joachim Breitner
 
-Trie for tokenizing the Lean language
+A string trie data strucuture, used for tokenizing the Lean language
```
Kha, line 6:
```suggestion
A string trie data structure, used for tokenizing the Lean language
```

src/Lean/Data/Trie.lean
```diff
@@ -1,110 +1,202 @@
...
+through the (compact) `ByteArray` is faster than using binary search or
+search trees like `RBTree`.
+
+Moreover, many nodes have degree 1, which justifies the special case `Node1`
+constructor.
+
+The code would be a bit less repetitive if we used something like the following
+```
+mutual
+def Trie α := (Option α, ByteAssoc α)
```
Kha, line 26:
```suggestion
def Trie α := Option α × ByteAssoc α
```
:)

src/Lean/Data/Trie.lean
```diff
@@ -1,110 +1,202 @@
...
+but that would come at the cost of extra indirections.
+-/
+
+/-- A Trie is a key-value store where the keys are of type `String`,
+and the internal structure is a tree that branches on the bytes of the string.  -/
 inductive Trie (α : Type) where
-  | Node : Option α → RBNode Char (fun _ => Trie α) → Trie α
+  | Leaf : Option α → Trie α
+  | Node1 : Option α → UInt8 → Trie α → Trie α
+  | Node : Option α → ByteArray -> Array (Trie α) → Trie α
```
Kha, line 42: Oops, all these constructors should be/have been lower-cased. Also there's a stray ASCII arrow.

Kha (changes requested):

Kha:
>     * [ ]  Is `upsert` a good name for the “insert or update” with parameter `(Option α → α)`, or would you want something else?

I like it. I don't think we have any similar function so far.

>     * [ ]  Is it ok for `findPrefix (t : Trie α) (pre : String) : Array α` to return the entries in arbitrary order, or must I sort them (either on insertion or in `findPrefix`?) It is currently used to pass completions to LSP, and I don’t know if they will be sorted afterwards.

They are sorted in the end of the completion handler.

Kha:
I did not review the implementation, I'm assuming it is sufficiently covered by the tests :)

src/Init/Data/String/Extra.lean
nomeata: I only cargo-culted it from `String.toUtf8` (where it is somewhat justifiable). Not sure I
I’d mention it in `String` – it’s an implementation detail, isn’t it?

nomeata:
awaiting-review

> I did not review the implementation, I'm assuming it is sufficiently covered by the tests :)

Does that mean you reviewed the tests ;-)

Kha:
Tests looked reasonable to me! Of course it's always hard to guess whether they cover all the interesting cases

src/Lean/Data/Trie.lean
```diff
@@ -1,110 +1,202 @@
...
 
 instance {α : Type} : ToString (Trie α) :=
   ⟨fun t => (flip Format.joinSep Format.line $ toStringAux t).pretty⟩
+
 end Trie
 
-end Parser
-end Lean
+end Data
+end Lean
```
Kha, line 202: Our VS Code settings should prevent the final newline from being stripped, are you using a different editor?

Kha (approved):

src/Lean/Data/Trie.lean
nomeata: I am using VS Code, but am no expert in it. Maybe because I used `code src/` rather than `code .`, (wrongly?) assuming I have to do that to benefit from the `elan` override in `src/` that’s described in `doc/dev/index.md`.

nomeata:
awaiting-review

tests/compiler/trie.lean
```diff
@@ -0,0 +1,88 @@
...
+    #["hello", "hella", "hellooo", "h", "hö", "hü", "💩", "", "hü"],
+    #["", "helooooo"]
+  ] do
+    IO.println "Resetting trie"
+    let mut t : T := T.empty
+    t.check
+    for s in seq do
+      IO.println s!"Inserting {s}"
+      t := t.insert s
+      t.check
```
Kha, line 88: Ah, another missing EOL

src/Lean/Data/Trie.lean
Kha: Good question! I believe you are right /cc @mhuisi

mhuisi: VS Code will use the `.vscode/settings.json` of the workspace that you open. If you opened `src`, then it will not use it.

I'll see if VS Code can be more helpful when opening the wrong folder in Lean 4 itself, too.

PR #2530, opened 2023-09-11
fix: rename parameter of withImportModules to match doc string

dwrensha:
Renames the `x` parameter in the code to match the `act` parameter in the docstring.
The mismatch was introduced in https://github.com/leanprover/lean4/commit/fb02fbb86704c11fdfeaecceb02d0e5bdb51f931.


digama0:
@semorrison The "view log" link is broken above, it just links to `https://github.com/leanprover-community/mathlib4/actions/runs/`

leodemoura (approved):

PR #2531, opened 2023-09-11
refactor: Trie.matchPrefix Do not return first field

nomeata:
It seems the `String.Pos` returned is not used anywhere, thus it seems this code can be simplified.

### DIFF
--- a/src/Lean/Data/Trie.lean
+++ b/src/Lean/Data/Trie.lean
@@ -79,24 +79,20 @@ where
       modify (·.push a)
     cs.forM fun _ c => collect c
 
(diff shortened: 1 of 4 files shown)


nomeata:
awaiting-review

leodemoura:
This is great.
Could you please add tests for the `Trie.lean` module?
We currently don't have many unit tests, but we are committed to fixing that.

nomeata:
I am rewriting the trie in #2529, and will write docstrings and tests as part of that (or, if that doesn’t get in, PR just the tests) if that’s ok.

leodemoura:
@nomeata Yes, it is ok. Thanks!

nomeata:
Subsumed by #2529

PR #2533, opened 2023-09-12
feat: use nat_gcd in the kernel

kim-em:
Followup to #2310.

### DIFF
--- a/tests/lean/run/lean_nat_gcd.lean
+++ b/tests/lean/run/lean_nat_gcd.lean
@@ -15,7 +15,7 @@ example : Nat.gcd 2  4  = 2  := rfl
 example : Nat.gcd 9  6  = 3  := rfl
 
 /-!
-We check that `Nat.gcd` is evaluated using GMP functions in the kernel.
+We check that `Nat.gcd` is evaluated using bignum functions in the kernel.
 
 Because of variations in run time on different operating systems during CI,
 for the larger calculations we don't attempt to do any timing.


leodemoura:
Could you please add more tests?

semorrison:
Added some tests. Not much experience with GMP, so not sure if this is the relevant sort of coverage.

I've hacked together the `#fast` command as the only guarantee that it is actually evaluating via GMP in the kernel.

semorrison:
The failing test suggests something may actually be wrong on windows!

```
1175/1724 Test #1171: leanruntest_lean_nat_gcd.lean .............................***Failed    7.19 sec
Unexpected return code 1 executing 'lean -j 0 -Dlinter.all=false lean_nat_gcd.lean'; expected 0. Output:
lean_nat_gcd.lean:32:0: error: Too slow! 5837ms
```

The test on line 32 is:
```
/- GCD with two prime factors on both sides, including one in common. -/
#fast example : Nat.gcd (m p_31 * m p_33) (m p_32 * m p_33) - m p_33 = 0 := rfl
```
and it takes only 33ms on my machine with this patch. (Prior to the patch if would fail with max recursion depth, which I guess it is not doing now even on Windows.)

leodemoura:
@semorrison Thanks for adding the new tests.

semorrison:
I think this was just variation in CI, and slow CI machines. The fact that we're not seeing max recursion depth errors is the real proof that we're successfully using GMP in these `rfl` proofs.

(I tried and failed to compile this PR in a local Windows VM, unfortunately. Lake wouldn't compile?)

Kha:
That's why I try to avoid these kinds of tests :) . The run time difference between Windows and the other platforms is still surprising though. Note that Windows is the only platform we run tests for that we do not build with GMP so this may be a significant difference between our bignum implementations.

Kha (approved):

tests/lean/run/lean_nat_gcd.lean
```diff
@@ -0,0 +1,67 @@
...
+example : Nat.gcd 1  1  = 1  := rfl
+example : Nat.gcd 1  17 = 1  := rfl
+example : Nat.gcd 17 1  = 1  := rfl
+example : Nat.gcd 2  2  = 2  := rfl
+example : Nat.gcd 2  3  = 1  := rfl
+example : Nat.gcd 2  4  = 2  := rfl
+example : Nat.gcd 9  6  = 3  := rfl
+
+/-!
+We check that `Nat.gcd` is evaluated using GMP functions in the kernel.
```
Kha, line 18:
```suggestion
We check that `Nat.gcd` is evaluated using bignum functions in the kernel.
```

Kha (approved):
I'm not a bignum expert but *looks* like good coverage to me!

Kha:
For full coverage, perhaps we should also test:
* with `GMP=ON/OFF`
* with #2612 -- I suspect this is the main reason for the previous high numbers

semorrison:
> For full coverage, perhaps we should also test:
> 
> * with `GMP=ON/OFF`
> * with [perf: do not inhibit caching of default-level `match` reduction #2612](https://github.com/leanprover/lean4/pull/2612) -- I suspect this is the main reason for the previous high numbers

Do I understand right that you are not suggesting changing ongoing testing in CI, but rather to:

* verify once off that with GMP=OFF at least some of these tests fail (i.e. to be sure we are really using the bignum algorithsm)
* find out what performance of these tests is like with GMP=OFF but with #2612?

semorrison:
> For full coverage, perhaps we should also test:
> 
> * with `GMP=ON/OFF`
> * with [perf: do not inhibit caching of default-level `match` reduction #2612](https://github.com/leanprover/lean4/pull/2612) -- I suspect this is the main reason for the previous high numbers

Isn't `USEGMP=ON/OFF` already tested in the matrix build, because different architecture have different values here?

For #2612 --- am I right in understanding you are suggesting that I try these tests in #2612 but without enabling bignum support for Nat.gcd, and see what the performance is like?

semorrison:
The test file fails on #2612 with `maximum recursion depth` in the same places it fails on `master` (i.e. every "big" calculation in the test file).

Kha:
Oh, I guess it makes sense that eliminating the exponential behavior does not actually reduce the stack depth. For GMP on/off I was mainly interested in the performance difference on the same machine but if the alternative is "stack overflow" then I suppose this is a good change with either bignum implementation!

PR #2539, opened 2023-09-13
feat: do not instantiate metavariables in kabstract/rw for disallowed occurrences

### LINKED ISSUE #2538: metavariables in `rw` are instantiated at the first match, even if this is disallowed by `occs`.
### Prerequisites

### Description

When using `rw` or `kabstract` with the `occs` parameter, metavariables in the lemma that we are rewriting by (or the expression we are abstracting) are instantiated by the first match found, even if this match is disallowed by occurrences.

### Steps to Reproduce

```
variable (f : Nat → Nat) (w : ∀ n, f n = 0)

example : [f 1, f 2, f 1, f 2] = [0, 0, 0, 0] := by
  rw (config := {occs := .pos [2]}) [w]
  -- Oh no! This currently rewrites the second `f 1`, rather than the first `f 2`,
  -- because the argument of `w` has already be fixed to `1` when we visit `f 1`.
```

Note this behaviour makes it impossible to achieve some rewrites. (Of course there are usually workarounds with `conv`, or by specifying arguments, but it is more difficult if automation wants to know about all possible rewrites.)

**Expected behavior:**

```
variable (f : Nat → Nat) (w : ∀ n, f n = 0)

example : [f 1, f 2, f 1, f 2] = [0, 0, 0, 0] := by
  rw (config := {occs := .pos [2]}) [w]
  trace_state -- ⊢ [f 1, 0, f 1, f 2] = [0, 0, 0, 0]
  rw [w, w]
```

**Actual behavior:**

```
variable (f : Nat → Nat) (w : ∀ n, f n = 0)

example : [f 1, f 2, f 1, f 2] = [0, 0, 0, 0] := by
  rw (config := {occs := .pos [2]}) [w]
  trace_state -- ⊢ [f 1, f 2, 0, f 2] = [0, 0, 0, 0]
  rw [w, w]
```

### Versions

The `config := {occs := ...}` option has only just landed in Lean, so you will need to either test this on master or on nightly-2023-09-14 when it arrives or later.

kim-em:
Fixes #2538.


src/Lean/Meta/KAbstract.lean
```diff
@@ -41,15 +43,22 @@ def kabstract (e : Expr) (p : Expr) (occs : Occurrences := .all) : MetaM Expr :=
...
-      else if (← isDefEq e p) then
-        let i ← get
-        set (i+1)
-        if occs.contains i then
-          return mkBVar offset
+      else
+        -- We save the metavariable context here,
+        -- so that it can be rolled back unless `occs.contains i`.
+        let mctx ← getMCtx
+        if (← isDefEq e p) then
```
leodemoura, line 50: What about exceptions? We should handle them here, or add a comment saying why it is ok to **not** rollback the mctx in this case.

semorrison: Sorry, I don't understand how exception handling is relevant here.

* Previously, if there was an exception at `isDefEq` then that would propagate to the caller of `kabstract`. That is still the case. Why would we try handling the exception here now?

* We are now rolling back the mctx **if** the occurrence has been rejected by the caller. Why would we want to rollback the mctx on an exception? (Again, previously an exception in `isDefEq` would result in the caller of `kabstract` receiving whatever mctx resulted.)

semorrison:
Someone just ran into this on [zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/How.20to.20use.20nth_rewrite.20in.20NNG4.3F).

MichaelStollBayreuth:
Fixing `nth_rw`/`nth_rewrite` would be good. The current behavior is confusing especially for beginners/students.

src/Lean/Meta/KAbstract.lean
semorrison: @leodemoura, could I ping you on this query?

kbuzzard:
Another confused beginner [here](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/rw.20tactic.20in.20NNG4.20behaving.20unexpectedly/near/409469383).

MichaelStollBayreuth:
[Here](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/nth_rw.20broken.3F/near/398897850) is me stumbling over unexpected `nth_rw` behavior.

src/Lean/Meta/KAbstract.lean
```diff
@@ -41,15 +43,22 @@ def kabstract (e : Expr) (p : Expr) (occs : Occurrences := .all) : MetaM Expr :=
...
         visitChildren ()
-      else if (← isDefEq e p) then
-        let i ← get
-        set (i+1)
-        if occs.contains i then
-          return mkBVar offset
+      else
+        -- We save the metavariable context here,
+        -- so that it can be rolled back unless `occs.contains i`.
+        let mctx ← getMCtx
```
JovanGerb, line 49: I noticed that in the definition of `checkPointDefEq`, it uses the `MonadBacktrack SavedState MetaM` instance for restoring the state if `isDefEq` fails or returns false. Is there any reason that you instead use the `MonadMCtx MetaM` instance?

PR #2544, opened 2023-09-14
fix: `stdin := .null` in `IO.Process.output`

tydeu:
Without this, if `IO.Process.output` is used from the LSP, it can accidently capture input from stdin, crashing the server.

There is an argument that this should be set by the user, because they may want tto receive input from stdin. Thus, `.inherit` may still be reasonable, but `.piped` certainly is not. Maybe `output` should have a flag `acceptsInput` to determine whether to chooise `.null` or `.inherit` that defaults to false?

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -10,6 +10,7 @@ Please check the [releases](https://github.com/leanprover/lean4/releases) page f
 v4.3.0 (development in progress)
 ---------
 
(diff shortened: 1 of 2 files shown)


Kha:
This is fine and the same choice as Rust's but could you please document the choice?

tydeu:
@Kha Does the docstring look good?

PR #2545, opened 2023-09-15
chore: begin development cycle for 4.2.0

kim-em:


RELEASES.md
```diff
@@ -7,6 +7,23 @@ only an expectation that breaking changes will be documented in this file.
 This file contains work-in-progress notes for the upcoming release, as well as previous stable releases.
 Please check the [releases](https://github.com/leanprover/lean4/releases) page for the current status of each version.
 
+v4.2.0 (development in progress)
+---------
+
+
+v4.1.0
+---------
+
```
tydeu, line 16:
```suggestion

* After elaborating a configuration file, Lake will now cache the configuration to a `lakefile.olean`. Subsequent runs of Lake will import this OLean instead of elaborating the configuration file. This provides a significant performance improvement (benchmarks indicate that using the OLean cuts Lake's startup time in half), but there are some important details to keep in mind:
  + Lake will regenerate this OLean after each modification to the `lakefile.lean` or `lean-toolchain`. You can also force a reconfigure by passing the new `--reconfigure` / `-R` option to `lake`.
  + Lake configuration options (i.e., `-K`) will be fixed at the moment of elaboration. Setting these options when `lake` is using the cached configuration will have no effect. To change options, run `lake` with `-R` / `--reconfigure`.
  + **The `lakefile.olean` is a local configuration and should not be committed to Git. Therefore, existing Lake packages need to add it to their `.gitignore`.**

* The signature of `Lake.buildO` has changed, `args` has been split into `weakArgs` and `traceArgs`. `traceArgs` are included in the input trace and `weakArgs` are not. See Lake's [FFI example](src/lake/examples/ffi/lib/lakefile.lean) for a demonstration of how to adapt to this change.

```

PR #2546, opened 2023-09-15
fix: use MoveFileEx for rename on win

digama0:
As [discovered on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/lake.20exe.20cache.20get!.20failure/near/391083452), the C++ standard `std::rename` function does not overwrite the target on windows, unlike on linux and macos. https://doc.rust-lang.org/std/fs/fn.rename.html mentions that it uses `MoveFileExW(from, to, MOVEFILE_REPLACE_EXISTING)` on windows to address this issue, so we implement the same fix here.

### DIFF
--- a/src/runtime/io.cpp
+++ b/src/runtime/io.cpp
@@ -660,7 +660,15 @@ extern "C" LEAN_EXPORT obj_res lean_io_remove_dir(b_obj_arg p, obj_arg) {
 }
 
 extern "C" LEAN_EXPORT obj_res lean_io_rename(b_obj_arg from, b_obj_arg to, lean_object * /* w */) {
-    if (std::rename(string_cstr(from), string_cstr(to)) == 0) {
+#ifdef LEAN_WINDOWS
+    // Note: On windows, std::rename gives an error if the `to` file already exists,
+    // so we have to call the underlying windows API directly to get behavior consistent
+    // with the unix-like OSs
+    bool ok = MoveFileEx(string_cstr(from), string_cstr(to), MOVEFILE_REPLACE_EXISTING) != 0;
+#else
+    bool ok = std::rename(string_cstr(from), string_cstr(to)) == 0;
+#endif
+    if (ok) {
         return io_result_mk_ok(box(0));
     } else {
         std::ostringstream s;
--- a/tests/lean/run/IO_test.lean
+++ b/tests/lean/run/IO_test.lean
@@ -35,6 +35,7 @@ pure ()
 #eval test
 
 def test2 : IO Unit := do
+let fn1 := "bar2.txt";
 let fn2 := "foo2.txt";
 let xs₀ : String := "⟨[₂,α]⟩";
 let xs₁ := "⟨[6,8,@]⟩";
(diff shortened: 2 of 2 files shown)


eyelash:
Does this work correctly for paths that contain (non-ASCII) Unicode characters?

semorrison:
Could we add a test for the basic functionality? Might as well answer @eyelash's question at the same time.

Probably we should also copy the "platform specific behaviour" section from https://doc.rust-lang.org/std/fs/fn.rename.html, and/or include that link.

digama0:
It is supposed to support non-ASCII, I don't see why it wouldn't, but it needs testing on a windows machine and I don't have one.

eyelash:
> It is supposed to support non-ASCII, I don't see why it wouldn't

The reason I was asking was because of these lines:

```cpp
std::wstring wfrom(string_cstr(from), string_cstr(from) + string_size(from));
std::wstring wto(string_cstr(to), string_cstr(to) + string_size(to));
```

I'm assuming that `string_cstr` returns a UTF-8 string and `MoveFileExW` takes a UTF-16 string and I don't see any conversion from UTF-8 to UTF-16 here.

By the way just calling the standard library with UTF-8 strings on Windows often does not work for non-ASCII strings so there might be more problems in this file related to Unicode on Windows.

digama0:
> I'm assuming that string_cstr returns a UTF-8 string and MoveFileExW takes a UTF-16 string and I don't see any conversion from UTF-8 to UTF-16 here.

The conversion is the wstring constructor, or at least it's supposed to... Like I said I have no ability to test this so someone else will have to tell me if it works.

semorrison:
A test file would run on Windows during CI! :-)

digama0:
I'm not sure I want to add tests which actually hit the filesystem...

Kha:
Please use the `MoveFileEx` base function like we do everywhere else. We do use the UTF-8 codepage... sometimes #2554.

> I'm not sure I want to add tests which actually hit the filesystem...

We already have `IO_test.lean`

digama0:
> Please use the `MoveFileEx` base function like we do everywhere else. We do use the UTF-8 codepage... sometimes #2554.

I'd be happy to use what is used elsewhere, I'm 100% cargo culting this code and don't do windows programming normally. But a grep for `MoveFileEx` comes up empty, do you have other examples?

src/runtime/io.cpp
```diff
@@ -660,7 +660,12 @@ extern "C" LEAN_EXPORT obj_res lean_io_remove_dir(b_obj_arg p, obj_arg) {
 }
 
 extern "C" LEAN_EXPORT obj_res lean_io_rename(b_obj_arg from, b_obj_arg to, lean_object * /* w */) {
-    if (std::rename(string_cstr(from), string_cstr(to)) == 0) {
+#if defined(LEAN_WINDOWS)
+    bool ok = MoveFileEx(wfrom.c_str(), wto.c_str(), MOVEFILE_REPLACE_EXISTING) != 0;
```
Kha, line 664: Comments! Please document why we don't use `rename` here

Kha:
Thanks!

PR #2547, opened 2023-09-17
fix: uninterpolated error message in `registerRpcProcedure`

thorimur:
In `registerRpcProcedure ` in `Lean.Server.Rpc.RequestHandling`, we have
```
def registerRpcProcedure (method : Name) : CoreM Unit := do
  let env ← getEnv
  let errMsg := "Failed to register RPC call handler for '{method}'"
  ...
```
`errMsg` fails to interpolate `method`. This PR uses `m!` to interpolate it.

(PR'd directly as a simple fix as per [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Small.20bugs.20in.20core.3F).)

### DIFF
--- a/src/Lean/Server/Rpc/RequestHandling.lean
+++ b/src/Lean/Server/Rpc/RequestHandling.lean
@@ -104,11 +104,11 @@ def registerBuiltinRpcProcedure (method : Name) paramType respType
 open Lean Elab Command Term Meta in
 def registerRpcProcedure (method : Name) : CoreM Unit := do
   let env ← getEnv
-  let errMsg := s!"Failed to register RPC call handler for '{method}'"
+  let errMsg := m!"Failed to register RPC call handler for '{method}'"
   if (←builtinRpcProcedures.get).contains method then
-    throwError s!"{errMsg}: already registered (builtin)"
+    throwError m!"{errMsg}: already registered (builtin)"
   if userRpcProcedures.contains env method then
-    throwError s!"{errMsg}: already registered"
+    throwError m!"{errMsg}: already registered"
   let wrappedName := method ++ `_rpc_wrapped
   let procT := mkConst ``RpcProcedure
   let proc ← MetaM.run' <| TermElabM.run' <| withoutErrToSorry do


thorimur:
awaiting-CI

src/Lean/Server/Rpc/RequestHandling.lean
```diff
@@ -104,7 +104,7 @@ def registerBuiltinRpcProcedure (method : Name) paramType respType
 open Lean Elab Command Term Meta in
 def registerRpcProcedure (method : Name) : CoreM Unit := do
   let env ← getEnv
-  let errMsg := "Failed to register RPC call handler for '{method}'"
+  let errMsg := s!"Failed to register RPC call handler for '{method}'"
```
Vtec234, line 107: I am guessing it could even be an `m!`, if you adjust the `throwError` also.

thorimur:
> - 💥 Mathlib branch [lean-pr-testing-2547](https://github.com/leanprover-community/mathlib4/compare/master...lean-pr-testing-2547) build failed against this PR. (2023-09-17 02:15:30) [View Log](https://github.com/leanprover-community/mathlib4/actions/runs/6210979102)

I'm guessing `master` breaks mathlib currently; the errors don't seem to be related to this fix.

Maybe in general we could test by rebasing the PR's commits onto whatever commit mathlib relies on, as long as there are no errors/conflicts during rebasing?

src/Lean/Server/Rpc/RequestHandling.lean
thorimur: I could make it `m!`, but I did want to keep this PR as a very tiny bugfix! :) Would `m!` just enable hover/clickability of `method`? Or would it potentially pretty-print the name differently?

semorrison:
> > * 💥 Mathlib branch [lean-pr-testing-2547](https://github.com/leanprover-community/mathlib4/compare/master...lean-pr-testing-2547) build failed against this PR. (2023-09-17 02:15:30) [View Log](https://github.com/leanprover-community/mathlib4/actions/runs/6210979102)
> 
> I'm guessing `master` breaks mathlib currently; the errors don't seem to be related to this fix.
> 
> Maybe in general we could test by rebasing the PR's commits onto whatever commit mathlib relies on, as long as there are no errors/conflicts during rebasing?

Yes, I'd though about this, but hoped it was unnecessary. I think it risks introducing unnecessary complexity, if the toolchain produced for a PR doesn't actually match what merging the PR would product. :-( My preference is to wait a bit longer, and see if better management of releases and the `nightly-testing` branch, as we get used to the processes, makes this problem mostly go away.

src/Lean/Server/Rpc/RequestHandling.lean
thorimur: Ok, fixed in a41bf0. :)

thorimur:
awaiting-review

PR #2551, opened 2023-09-17
chore: Remove unused PrefixTree and NameTrie

nomeata:


tydeu:
@nomeata Why are you removing `NameTrie`? I was considering using it eventually in Lake for [parametric targets](https://github.com/leanprover/lake/issues/114). For example, defining a parametric `foo.*` target that would capture and build both `foo.bar`  and `foo.baz`. My though process was that this could implemented via a `NameTrie` for prefixes (and concrete targets).

nomeata:
I stumbled across this and noticed that the code isn’t used in lean, std or mathlib, nor has been for quite a while; I asked about it in <https://leanprover.zulipchat.com/#narrow/stream/341532-lean4-dev/topic/Unused.20PrefixTrie.20and.20NameTrie> and opened the draft PR just to see if CI is happy, but yes, I should have at least linked back.

If you want to use `NameTrie`, then we should of course keep it!

(If it is used in a hot path, I can imagine some performance tweaks…)

PR #2555, opened 2023-09-18
chore: disable GMP by default

Kha:
Triggered by https://github.com/leanprover/lean4/pull/2533#issuecomment-1723208243; let's start by looking at whether there is any significant impact on our existing benchmarks (and if there isn't, we probably need new ones!)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -146,7 +146,7 @@ jobs:
             release: true
             cross: true
             shell: bash -euxo pipefail {0}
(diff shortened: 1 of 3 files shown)


Kha:
!bench

Kha:
*what*

Kha:
Ah, very interesting:
```
+/home/runner/work/lean4/lean4/src/runtime/mpz.cpp:430:16: runtime error: negation of -2147483648 cannot be represented in type 'int'; cast to an unsigned type to negate this value to itself
```

Kha:
Obsoleted by #5144

PR #2559, opened 2023-09-19
Structured `Meta.whnf` trace

Kha:
Makes `trace.profiler` much more effective on WHNF bottlenecks. Also make sure we actually know which WHNF is timing out.

### DIFF
--- a/src/Lean/Util/Trace.lean
+++ b/src/Lean/Util/Trace.lean
@@ -98,7 +98,7 @@ where
     else
       false
 
-def isTracingEnabledFor (cls : Name) : m Bool := do
+@[inline] def isTracingEnabledFor (cls : Name) : m Bool := do
   let inherited ← (inheritedTraceOptions.get : IO _)
   pure (checkTraceOption inherited (← getOptions) cls)
 


Kha:
!bench

Kha:
It would be nice to also show the result but that doesn't fit in the current `withTraceNode(Before)` dichotomy

Kha:
> Here are the [benchmark results](http://speed.lean-fro.org/lean4/run-detail/302182d1-9e5f-491a-b978-2f06e59544fb) for commit [3922555](https://github.com/leanprover/lean4/commit/39225557b0e670b976d4075e5605ea6c7def31ad). There were [significant changes](http://speed.lean-fro.org/lean4/compare/903b2e5d-573e-45b3-ad28-11ae014b0aff/to/302182d1-9e5f-491a-b978-2f06e59544fb) against commit [0d5f912](https://github.com/leanprover/lean4/commit/0d5f9122a1fc0cc1751a55eec8368a0d6592722a):
> 
> ```diff
>   Benchmark   Metric             Change
>   ================================================
> - stdlib      instructions         1.4% (2495.6 σ)
> - stdlib      tactic execution     7.6%   (14.3 σ)
> ```

Ahh, I feared so. The fast path without tracing enabled *should* be identical but the generated code is not great, e.g. the message is constructed unconditionally even if the value is not used in some branches. Let's keep around as WIP for specific testing until the code is good enough?

Kha:
!bench

Kha:
Obsoleted by #3774

PR #2562, opened 2023-09-19
perf: inline `isTracingEnabledFor`

Kha:
An easy way to avoid an extremely frequent allocation I noticed when looking at the code of #2559. It was significant there, let's see if it still is in isolation.


Kha:
!bench

PR #2563, opened 2023-09-19
fix: set ref in getCalcFirstStep

PatrickMassot:
---

Suggested by @gebner at https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/Random.20calc.20ranges/near/372704972. Sebastian wrote this is sufficiently tiny fix to allow for a direct PR.


Kha:
Yes, setting the syntax to be destructured as the ref is I think always an uncontroversial move. So much so that it could be automated, except I don't think we want to move back to `match_stx` nor make some `match`es effectful.

PR #2565, opened 2023-09-20
feat: lake: add name to manifest

tydeu:


tydeu:
@semorrison Do you think we can merge this into v4.1.0 as a hotfix? It would be very useful to have this for Reservoir this coming month.

PR #2570, opened 2023-09-21
perf: do not detect `lean`'s toolchain

tydeu:
Addresses a significant cause of startup latency in `v4.1.0-rc1`.

### DIFF
--- a/src/lake/Lake/Config/Env.lean
+++ b/src/lake/Lake/Config/Env.lean
@@ -51,13 +51,10 @@ def compute (lake : LakeInstall) (lean : LeanInstall) (elan? : Option ElanInstal
 
 /--
 The preferred toolchain of the environment. May be empty.
-Order is: `env.lean.toolchain` or `env.initToolchain` or Lake's `Lean.toolchain`
+Tries `env.initToolchain` first and then Lake's `Lean.toolchain`.
 -/
 def toolchain (env : Env) : String :=
-  if env.lean.toolchain.isEmpty then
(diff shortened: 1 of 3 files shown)


tydeu:
Local time measurement with `hyperfine`:

```
$ hyperfine -w 3 "rc1-lake self-check" "new-lake self-check"
Benchmark 1: rc1-lake self-check
  Time (mean ± σ):     403.8 ms ±   3.9 ms    [User: 280.9 ms, System: 120.3 ms]
  Range (min … max):   400.6 ms … 412.4 ms    10 runs

Benchmark 2: new-lake self-check
  Time (mean ± σ):     220.6 ms ±   2.7 ms    [User: 138.7 ms, System: 79.7 ms]
  Range (min … max):   216.9 ms … 227.2 ms    12 runs

Summary
  'new-lake self-check' ran
    1.83 ± 0.03 times faster than 'rc1-lake self-check'

Kha:
Nice work! What about the `lean --githash` invocation just below? It should be a bit faster, but in the standard colocated scenario I assume is it identical to `Lean.githash`.

By the way, have you benchmarked before how costly the `git` invocations on Windows are? On Linux they are super fast but I know that process creation is much more costly on Windows.

tydeu:
@Kha 

> Nice work! What about the `lean --githash` invocation just below? It should be a bit faster

That was my first thought after this as well! On my machine, hyperfine indicates `lean --githash` takes 100ms, so it is thus have the startup time. Thus, one of my next TODOs is to avoid it.

PR #2573, opened 2023-09-22
perf: lake: build lakefile environments incrementally

Kha:
Result of a Monday brainstorming between @tydeu and me. This change mostly uncouples Lake's workspace load time from the number of dependencies. In mathlib, we obtain a ~60% reduction in load time (as measure by `lake env true`), and ~40% in no-op `build` and `cache get`. Reductions for opening a file, i.e. `print-paths`, should be somewhere in between those two numbers.
```
$ nix run nixpkgs#hyperfine -- "lake env true" "lake +result env true"
Benchmark 1: lake env true
  Time (mean ± σ):     462.9 ms ±   2.6 ms    [User: 297.9 ms, System: 112.9 ms]
  Range (min … max):   458.5 ms … 466.4 ms    10 runs

Benchmark 2: lake +result env true
  Time (mean ± σ):     169.9 ms ±   1.1 ms    [User: 89.2 ms, System: 82.9 ms]
  Range (min … max):   168.1 ms … 173.0 ms    17 runs

Summary
  'lake +result env true' ran
    2.72 ± 0.02 times faster than 'lake env true'
    
$ nix run nixpkgs#hyperfine -- "lake exe cache get" "lake +result exe cache get"
Benchmark 1: lake exe cache get
  Time (mean ± σ):     887.7 ms ±   4.3 ms    [User: 695.1 ms, System: 259.1 ms]
  Range (min … max):   881.2 ms … 893.7 ms    10 runs

Benchmark 2: lake +result exe cache get
  Time (mean ± σ):     505.8 ms ±   3.8 ms    [User: 389.4 ms, System: 241.6 ms]
  Range (min … max):   502.0 ms … 514.1 ms    10 runs

Summary
  'lake +result exe cache get' ran
    1.76 ± 0.02 times faster than 'lake exe cache get'
    
$ nix run nixpkgs#hyperfine -- "lake build" "lake +result build"
Benchmark 1: lake build
  Time (mean ± σ):     915.5 ms ±   8.8 ms    [User: 1137.0 ms, System: 691.1 ms]
  Range (min … max):   898.4 ms … 934.5 ms    10 runs

Benchmark 2: lake +result build
  Time (mean ± σ):     529.7 ms ±   3.3 ms    [User: 817.2 ms, System: 685.4 ms]
  Range (min … max):   524.0 ms … 533.1 ms    10 runs

Summary
  'lake +result build' ran
    1.73 ± 0.02 times faster than 'lake build'
```

### DIFF
--- a/src/Lean/Environment.lean
+++ b/src/Lean/Environment.lean
@@ -659,7 +659,7 @@ def writeModule (env : Environment) (fname : System.FilePath) : IO Unit := do
 Construct a mapping from persistent extension name to entension index at the array of persistent extensions.
 We only consider extensions starting with index `>= startingAt`.
 -/
-private def mkExtNameMap (startingAt : Nat) : IO (HashMap Name Nat) := do
+def mkExtNameMap (startingAt : Nat) : IO (HashMap Name Nat) := do
   let descrs ← persistentEnvExtensionsRef.get
   let mut result := {}
   for h : i in [startingAt : descrs.size] do
--- a/src/lake/Lake/Load/Elab.lean
+++ b/src/lake/Lake/Load/Elab.lean
@@ -14,29 +14,23 @@ open Lean System
 
 deriving instance BEq, Hashable for Import
 
-/- Cache for the import state of Lake configuration files. -/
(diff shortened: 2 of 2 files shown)


tydeu (approved):
LGTM!

PR #2574, opened 2023-09-22
test: lake: add env & dep cfg benchmarks + cleanup

tydeu:
Follow-up to  #2457 that adds a `lake env` benchmark and a configuration benchmark for a package with multiple dependencies.


tydeu:
!bench

Kha (approved):

tydeu:
!bench

tydeu:
!bench

tydeu:
!bench

tydeu:
(The significant change is because `build no-op` now builds Inundation rather than Lean).

PR #2576, opened 2023-09-24
chore: Remove unused variables from kernel

nomeata:
when I build lean locally, I get a nice and warning-free build
experience with the exception of these two unused variables. They can
probably go?


nomeata:
awaiting-review

semorrison (approved):

PR #2577, opened 2023-09-24
refactor: Path compression for DiscrTree

nomeata:
Trie data structures tend to be very wasteful if one doesn’t apply path compression: If you have a sequence of nodes of degree 1, better put them in a flat array, instead of nested `.nodes` with singleton arrays.

If applied to the `DiscrTree`, this would reduce the size of the `DiscrTree` used by mathlib’s `library_search` from `217 MB` to `41 MB`, and the maximum depth from `2596` to `27`. The maximum depths was causing stack overflow in mathlib PRs.


nomeata:
Could someone !bench this?

I don't know why the mathlib bot reports failure; that failing file works locally in VSCode here. Maybe some toolchain confusion?

Kha:
!bench

nomeata:
> There were [no significant changes](http://speed.lean-fro.org/lean4/compare/55fc863a-73f0-4713-a374-dea774d8add5/to/4e4d1dce-0d96-442b-9bfc-e3a8446f2431) against commit https://github.com/leanprover/lean4/commit/2ac782c31593b0abc9ab84f3614895ef14b74e4f.

That’s a bit disappointing. I’ll make mathlib compile, maybe there is something there.

nomeata:
From 237MB to:
```
$ ls -sh ./build/lib/MathlibExtras/LibrarySearch.extra
58M ./build/lib/MathlibExtras/LibrarySearch.extra
```
That's at least pretty nice.

What do I have to do to get a speedcenter comparision for mathlib, for <https://github.com/leanprover-community/mathlib4/compare/nightly-testing...lean-pr-testing-2577>? Do I have to open a PR with that change first?

nomeata:
Results are in: http://speed.lean-fro.org/mathlib4/compare/e7b27246-a3e6-496a-b552-ff4b45c7236e/to/ee9e4caf-7a54-4dc1-a6d9-19101bc7d2d9?hash1=a477df946a0648231b66e5a7f0f04839546ebf25

Mild improvements across the board, it seems?

nomeata:
awaiting-review

Guess we have the data to weigh this idea. The code certainly gets more complex, but the size gains at least for `library_search` (didn't check the size of other in instances used) are significant, and a 3.5% mathlib compile time improvement is at least nice.

Jovan Gerbscheid pointed out that due to the well-nested structure of key sequences, values are only ever found at the leaves. One could therefore remove the .empty constructor and the recursive argument to .values, and make insertion panic when the key sequence is invalid, for a bit code reduction.

digama0:
A 3.5% wall clock improvement on mathlib is huge. I don't know how this happened, the instruction count is only down by 0.5% but the branch misses are down 2.5% so maybe the difference is that because discrimination trees are smaller there is less cache usage and shorter pointer chains to follow. Unfortunately cache misses do not seem to be a tracked figure so it is hard to say for sure, but this looks like a strong positive sign.

Kha:
Are you both looking at the "lint wall-clock" line instead of "build wall-clock"?

nomeata:
I looked at “build compilation” but thinking about it, it probably means just the .c->.o part, so obviously the wrong line 🤦‍♂️

build wall-time is up 0.5%, how unfortunate.

nomeata:
I updated the branches, and created a new comparison report:
<http://speed.lean-fro.org/mathlib4/compare/63a60bc6-4997-4dbe-a363-9ceff22def1c/to/1c8d5e96-3fad-4da2-b00c-4b1a63cdc75b>

I’m not sure how to best get a good signal from the report.
This time build wall-clock has improved by 0.9% (probably noise). Instruction counts go down through the bank, but less than 1%. All not very exiciting, not sure if this is worth pursuing, it seems only the size improvement to the `library_search` and `rw?` cache still look good.

semorrison:
Oh, I think you're underselling. A 1% improvement on Mathlib is still great news, and the size improvements for the caches are a significant QOL win, because they are usually the last to finish when running `lake exe cache get`.

semorrison:
@nomeata, I'm marking this back to `awaiting-author` now that there are conflicts.

nomeata:
Absolutely! Although I think I’ll see https://github.com/leanprover/std4/pull/285 (which I also have to update now) through first, so that refactorings like the present only affect one downstream repo, and not three.

nomeata:
I’ll close this. Others are working on DiscrTree more actively these days, have a better sense if this is useful, and if it is, can probably cherry-pick the ideas easily.

JovanGerb:
@nomeata, do you have interest in reviving this PR in some form? As mathlib and Lean continue to grow, the number of simp lemmas keeps growing, and building the simp discrimination tree (at import time) is taking a significant amount of time when importing all of mathlib. The refactor in this PR would be one way to mitigate this inefficiency.

nomeata:
Do we have evidence that this PR is helping with import times?

JovanGerb:
I measured that for `import Mathlib`, the `simp` discrimination tree takes about 0.5s and the instances discrimination tree takes about 0.1s. When building mathlib, the imports are typically smaller, so it will take less time per file, but with over 8000 files this will add up. (see [#mathlib4 > Performance cost of environment extensions @ 💬](https://leanprover.zulipchat.com/#narrow/channel/287929-mathlib4/topic/Performance.20cost.20of.20environment.20extensions/near/571450884))

And I recall that when this PR was benchmarked against mathlib there was a significant speedup. So I figured that this speedup is from the reduced import times, and that the effect would be even more significant now that mathlib has grown more.

nomeata:
Improving startup time for `import Mathlib` would be nice. I’ll ask claude to update this PR and we can get some measurement :-)

JovanGerb:
An alternative approach would be to make use of a `Thunk` data structure so as to never compute the entire discrimination tree. However when I tried this, I realized that the implementation would require some more flexibility from the environment extension framework: at import time we would have to first construct the root node of the discrimination tree, and after that is created, map over it with some `Thunk.mk`. Otherwise we would need to use a `Thunk.map` for each imported simp lemma, which I somehow imagine to not be efficient.

nomeata:
Closing this PR in favor of https://github.com/leanprover/lean4/pull/12838

Some form of lazy loading may possibly also be interesting, of course.

Issue #2580, opened 2023-09-25
RFC: style: number of spaces before `|`


mo271:
### Proposal

I propose to make the amount of spaces used before `\` consistent. Currently it is a mix of zero or two spaces, with a majority being two spaces. (There is very few places where another number such as 3 or 4 is used, I'll send a pr for that soon)

Other projects such as std and mathlib are doing two spaces.

  - **User Experience**: How does this feature improve the user experience?
  
  The user does not have to think how many spaces to type, automation tools can make consistent choices.

  - **Beneficiaries**: Which Lean users and projects benefit most from this feature/change?
  
  I suppose mainly those that write code in core lean, but of course this style, when done here, will probably be imitated by many other projects.

  - **Maintainability**: Will this change streamline code maintenance or simplify its structure?

Probably it will minimally simplify maintenance. 

### Community Feedback

Ideas should be discussed on [the Lean Zulip](https://leanprover.zulipchat.com) prior to submitting a proposal. Summarize all prior discussions and link them here.

_Summary_: Two spaces is the de-factor convention, so let's go with this.

https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/style.20for.20proofs.20with.20.60.7C.60.2E

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others benefit from the changes in this proposal being added, please ask them to add :+1: to it.

nomeata:
This is only about definitions defined by `|` clauses, but not about the `match` statement, where the clauses continue to be not indented relative to the `match`, correct?

mo271:
yes, when there is a `match` the `|` should have the same indentation. There are a few handful of cases currently in the library where this is not the case, including an odd number of spaces perhaps also worth cleaning up.

Kha:
Thanks for the RFC. As we do not have a formal style guide nor a code formatter implementing one nor do we want to create/review pure style changes manually, this RFC is not directly actionable. Thus I am closing it for the time being.

PR #2581, opened 2023-09-25
chore: indentation before `|`

mo271:
This fixes indentation before `|` in places where the amount of spaces was not 0 or 2.

I suppose this is the non-controversial part of #2580. Deciding for either 0 or 2 to make it consistent would be a much larger change.

#2580


tests/lean/run/KyleAlg.lean
```diff
@@ -187,7 +187,7 @@ instance [One α] [One β] : One (α × β) where
     one := (1, 1)
 
 theorem Product.ext : {p q : α × β} → p.1 = q.1 → p.2 = q.2 → p = q
-    | (a, b), (c, d) => by simp_all
```
eric-wieser, line 190: This whole file uses four-space indentation

tests/lean/run/KyleAlgAbbrev.lean
```diff
@@ -163,7 +163,7 @@ instance [One α] [One β] : One (α × β) where
     one := (1, 1)
 
 theorem Product.ext : {p q : α × β} → p.1 = q.1 → p.2 = q.2 → p = q
-    | (a, b), (c, d) => by simp_all
+  | (a, b), (c, d) => by simp_all
```
eric-wieser, line 166: As does this one.

tests/lean/run/KyleAlg.lean
mo271: yes, and curly braces and semi-colons.

Do you propose to not change the indentation before `|` in this file or to make the style of this file and `KyleAlgAbbrev` to match the rest of the repo. 
There are currently only 69 occurance of `by \{\n` , one of them seems deliberate (because of the comment "`Same example for curly braces and semicolons aficionados`")

eric-wieser: I can't speak for the lean4 developers, but I think the most sensible choice is either:
* Leave the file alone (reverting the changes here), adapt the rule to "one indentation level" rather than two spaces
* Make a separate PR to clean up files with four-space indents.

zwarich:
This is 2 years stale for a widespread style cleanup. If we still want to do this, someone can open a new PR.

PR #2586, opened 2023-09-26
fix: hover term/tactic confusion

digama0:
Fix for a bug [reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/by_contradiction.20not.20found.3F/near/393124002). By looking up `stx.getKind` even if `stx` is not a `Syntax.node`, we would mistakenly show doc strings for terms when hovering over tokens, or a doc string for `def ident` when hovering over identifiers.


Kha:
I believe I also had a proposal to put token kinds into a `_tokens` namespace or something like that which should have resolved this issue as well. But as it's just one line...

Issue #2329, opened 2023-07-17
Support for mutual and nested inductive types as `DecidableEq` instance generator


leodemoura:
We currently have to manually write `DecidableEq` instances for nested and mutual inductive datatypes. Here is an example.

```lean
namespace Example

inductive Min where
  | Base
  | Const (a : List Min)

mutual
def decMin (a b : Min) : Decidable (a = b) :=
  match a, b with
  | .Base, .Base         => isTrue rfl
  | .Base, .Const ..     => isFalse (by intro; contradiction)
  | .Const .., .Base     => isFalse (by intro; contradiction)
  | .Const as, .Const bs => match decMinList as bs with
    | isTrue h => isTrue (by rw [h])
    | isFalse _ => isFalse (by intro h; injection h; contradiction)

def decMinList (as bs : List Min) : Decidable (as = bs) :=
  match as, bs with
  | [], [] => isTrue rfl
  | _::_, [] => isFalse (by intro; contradiction)
  | [], _::_ => isFalse (by intro; contradiction)
  | a::as, b::bs =>
    match decMin a b with
    | isTrue h₁ => match decMinList as bs with
      | isTrue h₂ => isTrue (by rw [h₁, h₂])
      | isFalse _ => isFalse (by intro h; injection h; contradiction)
    | isFalse _ => isFalse (by intro h; injection h; contradiction)
end

instance : DecidableEq Min :=
  decMin
```

nomeata:
For mutual, this has been fixed a while ago in #2591.

For nested we are now in a good position to have it, with mutual recursion over nested inductives. @arthur-adjedj said he'd look into this. Thanks!

raoxiaojia:
@nomeata I forgot to give an update -- this is following the discussion at Wasm-DSL/spectec#192.

(For context: I needed `DecidableEq` for a part of Wasm AST involving nested mutual inductives, which the built-in deriver doesn't handle. My first approach was to instead apply the recursor, setting the motive to pointwise `DecEq`. However, the result had to be noncomputable since Lean doesn't compile recursors to C (leanprover/lean4#2049). @nomeata suggested analyzing the recursor to learn the structure, then generating compilable defs that follow the same decomposition.)

[...]

PR #2591, opened 2023-09-26
feat : derive `DecidableEq` for mutual inductives

arthur-adjedj:

This partly implements #2329 by making the derive handler manage mutual inductive types. The general framework for derive handlers is currently not well-suited to manage nested inductives without resorting to making the derived instances partial, and managing them would require big changes to the framework, and thus, to other handlers as well. Such work/design decisions should be discussed in a subsequent RFC.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -11,11 +11,15 @@ v4.3.0 (development in progress)
 ---------
 
 * The derive handler for `DecidableEq` [now handles](https://github.com/leanprover/lean4/pull/2591) mutual inductive types.
+* [Fix linker warnings on macOS](https://github.com/leanprover/lean4/pull/2598).
 
 v4.2.0
 ---------
 
-No breaking changes.
+* Improvements to Lake startup time ([#2572](https://github.com/leanprover/lean4/pull/2572), [#2573](https://github.com/leanprover/lean4/pull/2573))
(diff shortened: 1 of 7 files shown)


arthur-adjedj:
awaiting-review

semorrison:
I was slightly surprised to find that writing `deriving DecidableEq` on *both* inductives in a mutual block didn't error. That's great, but could you explain to me why? :-)

semorrison:
Would you mind adding a sentence about this to `RELEASES.md` in the `v4.3.0` section (you will need to rebase on master for this section to appear)? We're trying to get better about recording the lovely new features arriving in Lean!

semorrison (approved):

arthur-adjedj:
> I was slightly surprised to find that writing `deriving DecidableEq` on _both_ inductives in a mutual block didn't error. That's great, but could you explain to me why? :-)

I was surprised by this behaviour too ! As it turns out, other handlers share the same behaviour, so I figured I shouldn't change its behaviour here. Another thing that surprised me is that deriving on one inductive doesn't lead to the other one(s) also getting the instances generated. Furthermore, you can't use `deriving _` on a whole `mutual` block, despite handlers being handled arrays of names instead of just a single one. Because of this, you need to derive on each and every inductive, which leads to code duplication of the definitions behind the instances (!). I believe such fixes/changes could (should ?) come through reworking the whole framework as well, instead of it being on a case-by-case basis.

Issue #2361, opened 2023-07-27
Missing `withAssignableSyntheticOpaque` for `assumption` tactic?


kmill:
When you use `refine` in place of `constructor`, then the metavariables you get are all synthetic opaque. If you focus on a later goal, then you cannot use `assumption` to assign both the goal and a preceding metavariable since this causes it to fail the `isDefEq` check.

```lean
import Lean

example {α : Type} {P : α → Prop} (a : α) (h : P a) : ∃ x, P x := by
  constructor
  assumption -- succeeds

example {α : Type} {P : α → Prop} (a : α) (h : P a) : ∃ x, P x := by
  refine ⟨?m, ?c⟩
  case c =>
    /-
    α : Type
    P : α → Prop
    a : α
    h : P a
    ⊢ P ?m
    -/
    assumption -- fails

open Lean Elab Tactic Meta in
elab "assumption'" : tactic => do
  liftMetaTactic fun mvarId => withAssignableSyntheticOpaque do mvarId.assumption; pure []

example {α : Type} {P : α → Prop} (a : α) (h : P a) : ∃ x, P x := by
  refine ⟨?m, ?c⟩
  case c => assumption' -- succeeds
```

semorrison:
@kmill, I made #2596 following this suggestion. Mathlib builds fine. Would you be interested in adding a few tests?

PR #2596, opened 2023-09-27
feat: withAssignableSyntheticOpaque in assumption

kim-em:

* Fixes https://github.com/leanprover/lean4/issues/2361

### DIFF
--- a/tests/lean/2361.lean
+++ b/tests/lean/2361.lean
@@ -1,0 +1,16 @@
+import Lean
+
+-- This worked before issue #2361
+example {α : Type} {P : α → Prop} (a : α) (h : P a) : ∃ x, P x := by
+  constructor
+  assumption
+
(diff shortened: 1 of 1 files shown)


kmill:
I don't have permission to push to this branch, but here's a test for `tests/lean/2361.lean`:
```lean
import Lean

-- This worked before issue #2361
example {α : Type} {P : α → Prop} (a : α) (h : P a) : ∃ x, P x := by
  constructor
  assumption

example {α : Type} {P : α → Prop} (a : α) (h : P a) : ∃ x, P x := by
  refine ⟨?m, ?c⟩
  -- `?m` is synthetic opaque, so `assumption` needs to be able to assign such metavariables
  -- for this to succeed.
  case c => assumption

example {α : Type} {P : α → Prop} (a : α) (h : P a) : ∃ x, P x := by
  refine' ⟨_, ?c⟩
  case c => assumption
```

PR #2597, opened 2023-09-27
chore: add release note about lake startup time

kim-em:


tydeu:
@semorrison Nice! There is also #2583.

PR #2598, opened 2023-09-27
Fix linker warnings on macOS

Kha:


semorrison:
Yes, this toolchain successfully suppresses the linking warning for me! 🎉

Kha:
Great, thanks for testing

PR #2599, opened 2023-09-27
feat: Web Assembly Build

abentkamp:
This PR revives the web assembly build originally implemented by @Vtec234. Thanks to @rujialiu for the initial work on this.

The WebAssembly build had been abandoned because it was too slow for CI. I have now implemented a CI workflow suggested by @Kha that is a bit faster: We first build the oleans using native 32bit in stage0 and then we build the web assembly version in stage1 only.

This PR only provides a web assembly build of the batch compiler. The interactive mode (`lean --server`) is not yet supported.

Further discussion: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/wasm.20build.

### DIFF
--- a/stage0/src/CMakeLists.txt
+++ b/stage0/src/CMakeLists.txt
@@ -106,11 +106,10 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Emscripten")
 
     # From https://emscripten.org/docs/compiling/WebAssembly.html#backends:
     # > The simple and safe thing is to pass all -s flags at both compile and link time.
-    set(EMSCRIPTEN_SETTINGS "-s ALLOW_MEMORY_GROWTH=1 -s MAIN_MODULE=1 -fwasm-exceptions -pthread -flto")
+    set(EMSCRIPTEN_SETTINGS "-s ALLOW_MEMORY_GROWTH=1 -fwasm-exceptions -pthread -flto")
     string(APPEND LEANC_EXTRA_FLAGS " -pthread")
     string(APPEND LEAN_EXTRA_CXX_FLAGS " -D LEAN_EMSCRIPTEN ${EMSCRIPTEN_SETTINGS}")
     string(APPEND LEAN_EXTRA_LINKER_FLAGS " ${EMSCRIPTEN_SETTINGS}")
-    unset(CMAKE_EXECUTABLE_SUFFIX)
 endif()
 
 # Added for CTest
@@ -359,7 +358,11 @@ elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 elseif(${CMAKE_SYSTEM_NAME} MATCHES "Emscripten")
   string(APPEND CMAKE_CXX_FLAGS " -fPIC")
   string(APPEND LEANC_EXTRA_FLAGS " -fPIC")
-  string(APPEND CMAKE_EXE_LINKER_FLAGS " -Wl,--whole-archive -lInit -lLean -lleancpp -lleanrt ${EMSCRIPTEN_SETTINGS} -lnodefs.js -s EXIT_RUNTIME=1 -s LINKABLE=1 -s EXPORT_ALL=1")
+  # We do not use dynamic linking via leanshared for Emscripten to keep things
+  # simple. (And we are not interested in `Lake` anyway.) To use dynamic
+  # linking, we would probably have to set MAIN_MODULE=2 on `leanshared`,
+  # SIDE_MODULE=2 on `lean`, and set CMAKE_SHARED_LIBRARY_SUFFIX to ".js".
+  string(APPEND CMAKE_EXE_LINKER_FLAGS " -Wl,--whole-archive -lInit -lLean -lleancpp -lleanrt ${EMSCRIPTEN_SETTINGS} -lnodefs.js -s EXIT_RUNTIME=1 -s MAIN_MODULE=1 -s LINKABLE=1 -s EXPORT_ALL=1")
 elseif(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
   string(APPEND CMAKE_EXE_LINKER_FLAGS " -lleanshared")
 endif()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ endforeach()
 include(ExternalProject)
 project(LEAN CXX C)
 
-if("${CMAKE_SYSTEM_NAME}" MATCHES "Emscripten")
+if(("${CMAKE_SYSTEM_NAME}" MATCHES "Emscripten") AND ("${USE_GMP}" MATCHES "ON"))
```
Kha, line 29: Let's remove any code for Emscripten+GMP

src/runtime/mutex.cpp
```diff
@@ -51,9 +51,11 @@ extern "C" LEAN_EXPORT obj_res lean_io_condvar_new(obj_arg) {
 }
 
 extern "C" LEAN_EXPORT obj_res lean_io_condvar_wait(b_obj_arg condvar, b_obj_arg mtx, obj_arg) {
+#if defined(LEAN_MULTI_THREAD)
     unique_lock<mutex> lock(*basemutex_get(mtx), std::adopt_lock_t());
     condvar_get(condvar)->wait(lock);
     lock.release();
+#endif
```
Kha, line 58: Is it clear why only this function is affected?

src/shell/CMakeLists.txt
```diff
@@ -46,8 +46,14 @@ add_test(lean_ghash2   "${CMAKE_BINARY_DIR}/bin/lean" --githash)
 add_test(lean_unknown_option bash "${LEAN_SOURCE_DIR}/cmake/check_failure.sh" "${CMAKE_BINARY_DIR}/bin/lean" "-z")
 add_test(lean_unknown_file1 bash "${LEAN_SOURCE_DIR}/cmake/check_failure.sh" "${CMAKE_BINARY_DIR}/bin/lean" "boofoo.lean")
 
-# LEANC_OPTS is necessary for macOS c++ to find its headers
-set(TEST_VARS "PATH=${LEAN_BIN}:$PATH ${LEAN_TEST_VARS} CXX='${CMAKE_CXX_COMPILER} ${LEANC_OPTS}'")
+if(${EMSCRIPTEN})
+  set(LEAN_EXE "node --stack-size=8192 ${CMAKE_BINARY_DIR}/bin/lean")
```
Kha, line 50: What exactly is `lean` and how does it relate to `lean.wasm`?

Kha (changes requested):
Looks good! For any non-obvious wasm customization we should document though why it is necessary, e.g. what error is avoided by it

Kha:
Do we need the full stage 0 update or just a few select files?

src/runtime/mutex.cpp
abentkamp: Rujia added this part, but I think it's no longer needed, actually. `LEAN_MULTI_THREAD` is on anyway. I've removed it; let's see what the tests say.

src/shell/CMakeLists.txt
abentkamp: `lean` is a javascript file. It loads `lean.wasm`.

We could rename `lean` to `lean.js`  by removing `unset(CMAKE_EXECUTABLE_SUFFIX)` in `src/CMakeLists.txt` if you prefer. However, this causes an issue with the line `"${PREV_STAGE}/bin/lean${CMAKE_EXECUTABLE_SUFFIX}"` in `stdlib.make.in` because our stage0 does not have the suffix `.js`. We would need something like `${PREV_STAGE_CMAKE_EXECUTABLE_SUFFIX}` or so.

abentkamp:
I've removed the stage0 files that I suspect are not necessary.

abentkamp:
> Looks good! For any non-obvious wasm customization we should document though why it is necessary, e.g. what error is avoided by it

The one change that is puzzling me is why `leanshared` is not working and we need to link `Init`, `Lean` and `leancpp` directly. Maybe Rujia knows.

Anything else that is "non-obvious"?

abentkamp:
I still don't fully understand why `leanshared` does not work, but I have a suspicion now. I have added a comment in the code about this.

src/shell/CMakeLists.txt
Kha: Why is the JS file so big, is it all the exported symbols? `PREV_STAGE_CMAKE_EXECUTABLE_SUFFIX` doesn't sound too weird to me. The upside I assume would be that we could replace `LEAN_EXE` with a `lean` shell script that does the `node` invocation. Not sure. What do you think?

Kha:
> Anything else that is "non-obvious"?

Only other thing I noticed is why we disable the stack overflow module

rujialiu:
> Only other thing I noticed is why we disable the stack overflow module

I made that change. I don't remember the details but I couldn't make it work. It crashed at startup. Since I wanted to move on quickly (I wasn't sure how many obstacles we'll have) , I ended up commenting it without too much effort. Maybe we can try again and see whether we can keep it.

abentkamp:
Wojciech had also essentially disabled the stack overflow module in `stack_overflow.cpp`. I enabled everything and it seems to work fine. Let's see what CI says.

rujialiu:
Probably recent emscripten fixed that. Unfortunately I didn't realize I was using a rather old version of emscripten :(

src/shell/CMakeLists.txt
abentkamp: Yes, but I think we need to export all symbols because the interpreter can't find them otherwise.

I have added `PREV_STAGE_CMAKE_EXECUTABLE_SUFFIX` and replaced `LEAN_EXE` by a script.

Kha (approved):

abentkamp:
Okay, getting `PREV_STAGE_CMAKE_EXECUTABLE_SUFFIX` to work was harder than I thought, but now it all works.

Kha:
Great! Feel free to log follow-up issues on what remains to be solved in this repo related to wasm

PR #2601, opened 2023-09-28
fix: XML parsing bugs

kuruczgy:

* I didn't open issues since these are such simple bugs, please tell me if you want me to do so anyway.

Since this PR is for two different bugs I separated it into two commits. Please see the commit messages and the docstrings in the tests for explanations.

Results of the tests **before** the fixes:
```diff
-ok: <a><b a="v"></b></a>
-ok: <a><!-- comment --></a>
+error: offset 3: expected a name character
+<a
+error: offset 17: condition not satisfied
+ comment -
```

### DIFF
--- a/src/Lean/Data/Xml/Parser.lean
+++ b/src/Lean/Data/Xml/Parser.lean
@@ -119,7 +119,7 @@ def XMLdecl : Parsec Unit := do
 def Comment : Parsec String :=
   let notDash := Char.toString <$> satisfy (λ c => c ≠ '-')
   skipString "<!--" *>
-  Array.foldl String.append "" <$> many (notDash <|> (do
+  Array.foldl String.append "" <$> many (attempt <| notDash <|> (do
     let d ← pchar '-'
     let c ← notDash
     pure $ d.toString ++ c))
--- a/tests/lean/xmlParsing.lean
+++ b/tests/lean/xmlParsing.lean
@@ -5,3 +5,6 @@ open Lean.Xml
 
 /-! Test whether trailing whitespace in opening tags is handled correctly. -/
 #eval parse "<a ><b a=\"v\" ></b></a>"
+
+/-! Test whether comments are parsed correctly. -/
+#eval parse "<a><!-- comment --></a>"
--- a/tests/lean/xmlParsing.lean.expected.out
+++ b/tests/lean/xmlParsing.lean.expected.out
@@ -1,1 +1,2 @@
 ok: <a><b a="v"></b></a>
+ok: <a><!-- comment --></a>


kuruczgy:
Also tagging @DanielFabian since he seems to be the sole author of the XML parser.

Kha:
@digama0 @semorrison Do you think std4 would be a better home for this module?

digama0:
why does lean even have an xml parser?

hargoniX:
> why does lean even have an xml parser? Yes I would be fine with having this in std, along with the JSX style syntax in docgen.

Doesn't this make doc-gen depend on std4 so we end up in a cylic dependency chain when trying to build documentation for std4?

digama0:
> Doesn't this make doc-gen depend on std4 so we end up in a cylic dependency chain when trying to build documentation for std4?

I didn't necessarily mean to imply that docgen would use it; if there are cyclic dependency issues then docgen will just keep on having its own copy. That said, this is something I really want to fix about lake, there should be no need to have docgen as a dependency to run it over your project. There are many projects which are like lean addons that look *at* a lean project and produce some output, and we should not need to deal with observer-effect issues caused by adding the project as an additional dependency of the target project.

Docgen should be a standalone project which depends on whatever it wants, and is not a dependency of anything (other than other documentation tools); then you would just compile it and point it at a project you want documented with no lakefile changes. (Lean4checker is a recent tool that works in this way - you just point it at a module you want verified, you don't add it as a dependency of the module you want verified.)

hargoniX:
There used to be a time where doc-gen did behave like this independently of lake. But there were two main reasons to do the lake thing:
1. .olean breakages. Quite often doc-gen would crash because parts of the .olean format were changed (e.g. by doing changes to env extensions) such that doc-gen would access it in a bad way since it was compiled with a different version than the target project which would lead to a crash
2. Lake allows doc-gen to profit from parallelism for 0 cost and in addition also smarter rebuilds in case you are generating documentation locally.

How do you see these two things being addressed without lake? I guess the first one should eventually be fine as Lean stabilizes. For the second one should I just build my own parallelism?

digama0:
Both issues should be solved by using lake, although in different ways. For (1) the solution is to have `lake` compile `doc-gen` using the version in the lean-toolchain. This is however not a regular build to go into the build directory of the current project, it is more like a global install. My model for this is `cargo install`: you can download a project, compile it to an executable, then run it as a cargo subcommand, so that you can have things like `cargo expand`, `cargo msrv` or `cargo flamegraph`. These can then be directed to run on the current project.

For (2), you would want to use lake as a harness, setting up a custom build graph and getting lake to execute it. Again, it would not be the normal frontend, it would be somewhat analogous to calling `runFrontend` or `elabCommand` directly from a `main` function to drive elaboration in a specialized way. As far as I know lake is not set up to do this particularly well, but it could be a feature request.

semorrison:
I agree we should move XML parsing out of the core repository, but this bugfix PR isn't the place to do it!

PR #2603, opened 2023-09-29
feat: lake: `postUpdate?` + test

tydeu:
Closes leanprover/lake#185. 

This feature is implemented in a very straightforward mannaer. It is simply a monadic package configuration option.  `lake update` runs all the workspace's packages' `postUpdate?` actions after resolving dependencies and finalizing the workspace but before writing the manifest to its file.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -13,6 +13,7 @@ v4.3.0 (development in progress)
 * The derive handler for `DecidableEq` [now handles](https://github.com/leanprover/lean4/pull/2591) mutual inductive types.
 * [Show path of failed import in Lake](https://github.com/leanprover/lean4/pull/2616).
 * [Fix linker warnings on macOS](https://github.com/leanprover/lean4/pull/2598).
+* **Lake:** Add `postUpdate?` package configuration option. Used by a package to specify some code which should be run after a successful `lake update` of the package or one of its downstream dependencies. ([lake#185](https://github.com/leanprover/lake/issues/185))
 
 v4.2.0
 ---------
--- a/src/lake/README.md
+++ b/src/lake/README.md
@@ -164,6 +164,7 @@ Lake provides a large assortment of configuration options for packages.
 
 ### Build & Run
 
+* `postUpdate?`: A post-`lake update` hook. The monadic action is run after a successful `lake update` execution on this package or one of its downstream dependents. Defaults to `none`. See the option's docstring for a complete example.
 * `precompileModules`:  Whether to compile each module into a native shared library that is loaded whenever the module is imported. This speeds up the evaluation of metaprograms and enables the interpreter to run functions marked `@[extern]`. Defaults to `false`.
 * `moreServerArgs`:  Additional arguments to pass to the Lean language server (i.e., `lean --server`) launched by `lake serve`.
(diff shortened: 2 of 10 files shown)


RELEASES.md
```diff
@@ -10,6 +10,8 @@ Please check the [releases](https://github.com/leanprover/lean4/releases) page f
 v4.3.0 (development in progress)
 ---------
 
+* **Lake:** Add `postUpdate?` package configuration option. Used by a package to specify some code which should be run after a successful `lake update` of the package or one of its downstream dependencies. ([lake#185](https://github.com/leanprover/lake/issues/185))
```
eric-wieser, line 13: Do you think it's worth mentioning `lake exe cache get` explicitly here, and noting the mathlib motivation?

tydeu: @eric-wieser I tend to think of a changelog as standalone briefs of *what* was changed and leave the *why* to referenced issues, especially for new features (whereas, e.g., performance improvements, may need to the *why* to understand the significance of the the *what*). However, as the `RELEASES.md` is meant, at least in part, to be a tool for the community. it may be worth further community discussion as to what style of notes would be most helpful.

src/lake/README.md
```diff
@@ -164,6 +164,7 @@ Lake provides a large assortment of configuration options for packages.
 
 ### Build & Run
 
+* `postUpdate?`: A post-`lake update` hook. The monadic action is run after a successful `lake update` execution on this package or one of its downstream dependents. Defaults to `none`.
```
semorrison, line 167: Could we have an example? Looking at this diff, the only clue as to how to actually define such a hook is:
* reading the type `Option (LakeT LogIO PUnit)`
* looking at  `src/lake/tests/postUpdate/dep/lakefile.lean`.

If this is meant to help Mathlib run `lake exe cache get` after an update, then there needs to be an example in a documentation file that shows how to run an `exe` target as an update hook.

tydeu: @semorrison While I agree in spirit, I am not quite sure where to put such an example. Putting it here in the README would be awkward as none of the other configuration options have examples. Thus, giving it one would would draw the reader's attention, focusing them on an option that is likely not useful for most users. Would putting the example in the docstring be sufficient?

semorrison: Maybe start a new file? I don't really mind where it goes. If one option is better documented than others that's a good thing, and the "fix" is to document the others as well...

semorrison: I'm just hoping that we can say "Patrick, do you want to see if you can make `lake exe cache get` run automatically as part of `lake update`", and he replies with "Wow, that was easy" rather than "`Option (LakeT LogIO PUnit)`???".

tydeu: @semorrison Does the new docstring look good?

semorrison: The docstring is great, and I've verified that it works.

My only concern is about the code block in the comment eventually rotting.

* Maybe the API is uses is sufficiently stable to just not worry?
* Alternatively we could have a test that is effectively isomorphic (in terms of API calls) to the example in the doc-string, and a comment saying "If you need to fix this test, make sure you also update the doc-string on `postUpdate?`".

I'm happy either way.

tydeu: @semorrison  I went with option 2. If everything looks good, this feature should be ready to merge. 😄

semorrison (approved):

PR #2604, opened 2023-09-29
feat: make `Environment.mk` private

digama0:
As one follow-up to the discussion [here](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/PSA.20about.20trusting.20Lean.20proofs) and [here](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/More.20on.20trusting.20Lean.20proofs), one easy way to make it harder for tactic-writers to accidentally modify the environment to add unchecked constants is to make the constructor private.

Against my expectations, this broke literally nothing in lean itself, and only one test (which ironically was testing private constructors); this was switched to use the API function and everything else worked. We'll see what the mathlib breakage is.

### DIFF
--- a/src/Lean/Environment.lean
+++ b/src/Lean/Environment.lean
@@ -852,7 +852,7 @@ private def registerNamePrefixes : Environment → Name → Environment
   | env, _        => env
 
 @[export lean_environment_add]
-private def add (env : Environment) (cinfo : ConstantInfo) : Environment :=
+def add (env : Environment) (cinfo : ConstantInfo) : Environment :=
   let env := registerNamePrefixes env cinfo.name
   env.addAux cinfo
 
--- a/src/lake/Lake/Load/Elab.lean
+++ b/src/lake/Lake/Load/Elab.lean
@@ -72,13 +72,6 @@ def elabConfigFile (pkgDir : FilePath) (lakeOpts : NameMap String)
     return s.commandState.env
 
 /--
-`Lean.Environment.add` is now private, but exported as `lean_environment_add`.
-We call it here via `@[extern]` with a mock implementation.
--/
-@[extern "lean_environment_add"]
-private def add (env : Environment) (_ : ConstantInfo) : Environment := env
-
-/--
 Import the OLean for the configuration file if `reconfigure` is not set
 and an up-to-date one exists (i.e., one newer than the configuration and the
 toolchain). Otherwise, elaborate the configuration and save it to the OLean.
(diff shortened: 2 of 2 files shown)


src/Lean/Environment.lean
```diff
@@ -117,6 +117,9 @@ declared by users are stored in an environment extension. Users can declare new
 using meta-programming.
 -/
 structure Environment where
+  /-- The constructor of `Environment` is private to protect against modification
+  that bypasses the kernel. -/
```
semorrison, line 121: I think it would be better to say "discourage" or "inhibit" rather than "protect against", given the existence of `open private`! 

(I don't want the comment to encourage someone feeling clever to announce that `open private` leads to unsoundness.)

SchrodingerZhu:
How can I obtain an empty Environment for mock testing?

semorrison:
> How can I obtain an empty Environment for mock testing?

Would using `Environment.mk` via `open private` from Std suffice for your use case?

SchrodingerZhu:
> > How can I obtain an empty Environment for mock testing?
> 
> 
> 
> Would using `Environment.mk` via `open private` from Std suffice for your use case?

Interesting, just learned that private is not definite Lean.

semorrison:
> How can I obtain an empty Environment for mock testing?

@SchrodingerZhu, sorry, I gave you the wrong answer previously: you can use `Lean.mkEmptyEnvironment`.

semorrison:
Making `Environment.add` private is more complicated: it currently breaks Aesop, but if fixable by using the "replay" functionality of `lean4checker` directly.

In order to keep things moving, I'm now going to revert the commit to this PR in which I made `Environment.add` private --- we'll do that later.

semorrison:
@digama0, if you're happy with my proposed change to the comment, I can merge this.

src/Lean/Environment.lean
digama0: another possible word is "defend against", I don't think this implies indefeasibility. But I didn't really want to go into an extended discussion about the soundness implications of violating privacy here.

digama0:
Hm, the `Environment.add` part was a mistake, it should be private as well. Can we go forward with the change and calling it via `extern` as you did? (It would be good if `extern` was flagged a bit more clearly as unsafe, but it seems visible enough for review purposes. Is there a reason the same technique doesn't work for Aesop?)

Kha:
> Is there a reason the same technique doesn't work for Aesop?

The interpreter only knows how to handle `[extern]` from compiled modules

semorrison:
My suggestion is that we merge as is, and then when #2617 is merged Aesop can use that, and then we can make `Environment.add` private as well.

Issue #2150, opened 2023-03-13
Missing `withMainContext` in `Lean.Elab.Tactic.withLocation`


kmill:
I'm not completely sure if this is an issue or working as intended, but, in consideration of the `Location.wildcard` case, it appears that [this line](https://github.com/leanprover/lean4/blob/0cc9d7a43de73af086597c48e2ff427f31fef4f3/src/Lean/Elab/Tactic/Location.lean#L52) of `Lean.Elab.Tactic.withLocation` should be `withMainContext atTarget` rather than just `atTarget`.

There is an easy workaround for this, but I figured I should create an issue to track it.

PR #2607, opened 2023-10-01
fix: withLocation should use withMainContext for target

alexjbest:
* [x] Put an X in this bracket to confirm you have read the

This is done in `withLocation` for all other calls to `atLocal` and `atTarget` (in the wildcard case).
The fact this is missing causes mathlib tactics to fail as metaprogrammers are not aware of this inconsistency and hence may miss this even when tactics work on some tests. Therefore this PR aims to make downstream tactic writing more consistent.

This issue was discovered and worked around in https://github.com/leanprover-community/mathlib4/pull/7437, though other mathlib tactics may still be affected by it. Downstream it causes the important tactic linarith to not work properly on goals containing local variables, but in a way opaque to the end user.

A minimalistic test is provided with the simplest example I can think of, a tactic that simply prints the user name of some FVar in the goal.

### DIFF
--- a/src/Lean/Elab/Tactic/Location.lean
+++ b/src/Lean/Elab/Tactic/Location.lean
@@ -49,7 +49,7 @@ def withLocation (loc : Location) (atLocal : FVarId → TacticM Unit) (atTarget 
       let fvarId ← getFVarId hyp
       atLocal fvarId
     if type then
-      atTarget
+      withMainContext atTarget
   | Location.wildcard =>
     let worked ← tryTactic <| withMainContext <| atTarget
     withMainContext do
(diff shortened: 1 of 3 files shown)


alexjbest:
awaiting-review

semorrison:
@alexjbest, the `withMainContext` in 
```
worked := worked || (← tryTactic <| withMainContext <| atLocal fvarId)
```
(i.e. a few lines past your change) is redundant, right? Perhaps doesn't need to change now, what do you think?

alexjbest:
> @alexjbest, the `withMainContext` in
> 
> ```
> worked := worked || (← tryTactic <| withMainContext <| atLocal fvarId)
> ```
> 
> (i.e. a few lines past your change) is redundant, right? Perhaps doesn't need to change now, what do you think?

I think so but I am hesitant to change anything in core / that Leo wrote exactly that way without understanding everything fully (given my track record with subtly broken PRs!). So I'd rather not change that here.

kmill:
Fixes #2150

PR #2610, opened 2023-10-01
feat: mark predefinitions that failed to compile as noncomputable

alexjbest:

Currently definitions inside a noncomputable section that fail to compile are not explicitly tagged as noncomputable.
A noncomputable section simply attempts to compile all definitions but returns false instead of erroring when a compilation error is encountered, however this information is discarded in `addNonRecAux`.
It seems more useful for downstream tactics / users for the fact that compilation failed to be recorded by tagging the definition with `noncomputable`.
For example the `to_additive` tactic in mathlib will always try to compile declarations generated from not noncomputable multiplicative versions, which means that currently to_additive breaks inside `noncomputable section`s see https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/noncomputable.20and.20to_additive/near/394068118, this particular issue could be dealt with downstream, but adding the noncomputable tag seems to be the cleanest implementation to me.

### DIFF
--- a/src/Lean/Elab/PreDefinition/Basic.lean
+++ b/src/Lean/Elab/PreDefinition/Basic.lean
@@ -117,10 +117,13 @@ private def addNonRecAux (preDef : PreDefinition) (compile : Bool) (all : List N
     withSaveInfoContext do  -- save new env
       addTermInfo' preDef.ref (← mkConstWithLevelParams preDef.declName) (isBinder := true)
     applyAttributesOf #[preDef] AttributeApplicationTime.afterTypeChecking
-    if preDef.modifiers.isNoncomputable then
+    let didCompile ←
+      if compile && shouldGenCodeFor preDef then
+        compileDecl decl
+      else
+        pure false
+    if preDef.modifiers.isNoncomputable || !didCompile then
       modifyEnv fun env => addNoncomputable env preDef.declName
-    if compile && shouldGenCodeFor preDef then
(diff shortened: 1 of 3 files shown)


alexjbest:
awaiting-review

semorrison (approved):

src/Lean/Elab/PreDefinition/Basic.lean
```diff
@@ -117,10 +117,13 @@ private def addNonRecAux (preDef : PreDefinition) (compile : Bool) (all : List N
...
       addTermInfo' preDef.ref (← mkConstWithLevelParams preDef.declName) (isBinder := true)
     applyAttributesOf #[preDef] AttributeApplicationTime.afterTypeChecking
-    if preDef.modifiers.isNoncomputable then
+    let didCompile ←
+      if compile && shouldGenCodeFor preDef then
+        compileDecl decl
+      else
+        pure false
+    if preDef.modifiers.isNoncomputable || !didCompile then
       modifyEnv fun env => addNoncomputable env preDef.declName
```
semorrison, line 126: This means that all theorems will now be marked as noncomputable, whereas before they were not so marked. While it's unclear how much that matters, I think it would be good to not change things unnecessarily (and moreover, it would be great if we could have sufficiently thorough tests here that such changes are trivial to notice!)

zwarich:
I'm going to close this old PR in favor of #9021.

PR #2616, opened 2023-10-03
    feat: show path of failed import

bollu:
### Old error

```
    error: no such file or directory (error code: 2)
      file: ./././Bug/Foo.lean
```
### New error
```
    error: 'bug.Main' ▸ 'Bug' ▸ 'Bug.Foo' ▸ 'Bug.Foo': no such file or directory (error code: 2)
      file: ./././Bug/Foo.lean
```

@tydeu do you know why the path contains repeated references to the module `Bug.Foo`?

awaiting-review

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -11,6 +11,7 @@ v4.3.0 (development in progress)
 ---------
 
 * The derive handler for `DecidableEq` [now handles](https://github.com/leanprover/lean4/pull/2591) mutual inductive types.
+* [Show path of failed import in Lake](https://github.com/leanprover/lean4/pull/2616).
 * [Fix linker warnings on macOS](https://github.com/leanprover/lean4/pull/2598).
 
 v4.2.0


bollu:
awaiting-review

tydeu:
@bollu 

> @tydeu do you know why the path contains repeated references to the module `Bug.Foo`?

This is because Lake visits multiple nested facets of the same module when building it (and `parseImports` is not the top-level one).

src/lake/Lake/Build/Module.lean
```diff
@@ -13,6 +13,15 @@ Build function definitions for a module's builtin facets.
 
 open System
 
+
+/-- Remove adjacent duplicates. -/
+def List.squeeze [BEq α] : List α → List α
```
tydeu, line 18: @bollu Sorry, minor style quibble: Lake does not put things outside the `Lake` namespace (e.g., see `Lake.Util.Name`). Unfortunately, this does mean one cannot use dot notation with new definitions, but I think that is worth it to avoid potential clashes (e.g., with Std) when using Lake as a library.

bollu:
@tydeu I pulled the auxiliary definition into `Lake.List.squeeze`.

tydeu:
LGTM!

tydeu:
@bollu I forgot about this myself as well, but this should probably come with a brief mention in `RELEASES.md`. Sorry for another nitpick!

bollu:
@tydeu Just added a release note!

PR #2617, opened 2023-10-03
feat: replay constants into an Environment

kim-em:
This is the core functionality of `lean4checker`.

It is separately useful for users who need to copy constants from one `Environment` to another safely, without using `Environment.add`, which we will like to make private as in https://github.com/leanprover/lean4/pull/2604.

### DIFF
--- a/src/Lean/Replay.lean
+++ b/src/Lean/Replay.lean
@@ -17,7 +17,7 @@ but rather checks that they are identical to constructors or recursors generated
 after replaying any inductive definitions occurring in `constantMap`.
 
 `replay` can be used either as:
-* a verifier for an `Enviroment`, by sending everything to the kernel, or
+* a verifier for an `Environment`, by sending everything to the kernel, or
 * a mechanism to safely transfer constants from one `Environment` to another.
 
 -/
@@ -72,7 +72,7 @@ and add it to the environment.
 partial def replayConstant (name : Name) : M Unit := do
   if ← isTodo name then
     let some ci := (← read).newConstants.find? name | unreachable!
-    replayConstants ci.getUsedConstants
+    replayConstants ci.getUsedConstantsAsSet
     -- Check that this name is still pending: a mutual block may have taken care of it.
     if (← get).pending.contains name then
       match ci with
@@ -97,7 +97,7 @@ partial def replayConstant (name : Name) : M Unit := do
         -- Make sure we are really finished with the constructors.
         for (_, ctors) in ctorInfo do
           for ctor in ctors do
-            replayConstants ctor.getUsedConstants
+            replayConstants ctor.getUsedConstantsAsSet
         let types : List InductiveType := ctorInfo.map fun ⟨ci, ctors⟩ =>
           { name := ci.name
             type := ci.type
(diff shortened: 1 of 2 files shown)


src/Lean/Replay.lean
```diff
@@ -0,0 +1,169 @@
...
+
+`replay env constantMap` will "replay" all the constants in `constantMap : HashMap Name ConstantInfo` into `env`,
+sending each declaration to the kernel for checking.
+
+`replay` does not send constructors or recursors in `constantMap` to the kernel,
+but rather checks that they are identical to constructors or recursors generated in the enviroment
+after replaying any inductive definitions occurring in `constantMap`.
+
+`replay` can be used either as:
+* a verifier for an `Enviroment`, by sending everything to the kernel, or
```
david-christiansen, line 20:
```suggestion
* a verifier for an `Environment`, by sending everything to the kernel, or
```

src/Lean/Util/FoldConsts.lean
```diff
@@ -66,8 +66,26 @@ opaque foldConsts {α : Type} (e : Expr) (init : α) (f : Name → α → α) :
 def getUsedConstants (e : Expr) : Array Name :=
   e.foldConsts #[] fun c cs => cs.push c
 
+/-- Like `Expr.getUsedConstants`, but produce a `NameSet`. -/
+def getUsedConstants' (e : Expr) : NameSet :=
```
david-christiansen, line 70: What about calling it `getUsedConstantsAsSet` so that the name says more than "a bit like that other one"?

src/Lean/Util/FoldConsts.lean
```diff
@@ -66,8 +66,26 @@ opaque foldConsts {α : Type} (e : Expr) (init : α) (f : Name → α → α) :
...
+  e.foldConsts {} fun c cs => cs.insert c
+
 end Expr
 
+namespace ConstantInfo
+
+/-- Return all names appearing in the type or value of a `ConstantInfo`. -/
+def getUsedConstants (c : ConstantInfo) : NameSet :=
```
david-christiansen, line 78: Just above, `getUsedConstants'` is the version that returns a `NameSet`, but down here it's the version without the `'` that does it. This seems likely to confuse someone in the future.

src/Lean/Replay.lean
semorrison: Done.

src/Lean/Util/FoldConsts.lean
semorrison: Done.

src/Lean/Util/FoldConsts.lean
semorrison: Done.

enricozb:
Does this not conflict with the `append` in mathlib4?

https://github.com/leanprover-community/mathlib4/blob/c98e6b7f0f028f7958b90fe156a7756f9b7a1939/Mathlib/Lean/Expr/Basic.lean#L118-L127

Having some issues trying to write a nix derivation for the current mathlib.

semorrison:
@enricozb, yes, it conflicts. The `nightly-testing` branch of Mathlib has already removed `NameSet.append`, and when Mathlib next moves to a new release candidate or stable release we will remove it from the `master` branch too.

enricozb:
Ah got it thanks for the info, didn't realize mathlib had a nightly branch. Thanks!

PR #2621, opened 2023-10-04
chore: simp tracing reports ←

### LINKED ISSUE #2131: Incorrect simp tracing for local right-to-left simp theorems (`[<- x]`)
MWE:

```lean
structure A
structure B

theorem stuff : A = B := sorry

example : B := by
  set_option tactic.simp.trace true in
  simp [← stuff] -- Try this: simp only [stuff] // should be: simp only [← stuff]
  exact A.mk
```

Tested with Lean 4 nightly 2023-03-01.

Reported on Zulip: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.60simp.3F.60.20dropping.20.60.3C-.60

kim-em:
Fixes #2131 and https://github.com/leanprover/std4/issues/264

### DIFF
--- a/src/Lean/Meta/Tactic/Simp/SimpTheorems.lean
+++ b/src/Lean/Meta/Tactic/Simp/SimpTheorems.lean
@@ -186,10 +186,10 @@ def SimpTheorems.registerDeclToUnfoldThms (d : SimpTheorems) (declName : Name) (
 
 partial def SimpTheorems.eraseCore (d : SimpTheorems) (thmId : Origin) : SimpTheorems :=
   let d := { d with erased := d.erased.insert thmId, lemmaNames := d.lemmaNames.erase thmId }
-  if let .decl declName false := thmId then
+  if let .decl declName := thmId then
     let d := { d with toUnfold := d.toUnfold.erase declName }
     if let some thms := d.toUnfoldThms.find? declName then
-      thms.foldl (init := d) (eraseCore · <| .decl · false)
+      thms.foldl (init := d) (eraseCore · <| .decl ·)
     else
       d
   else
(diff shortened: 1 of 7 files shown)


semorrison:
@digama0 if it's useful for reviewing I can make proposed downstream changes to Std and Mathlib as well, but won't immediately unless you say so.

semorrison:
@digama0, I'd like to get this in, but am uncertain whether your self-assignment above means:
* that I shouldn't do further work on this because you have in mind a better solution?
* that I should do the downstream fixes to Std and Mathlib required for this approach, and then you'll review?

digama0:
It means I was planning on reviewing it, but I was at a conference last week so I'm just now working down my backlog. I'll take a look at it now or tomorrow.

src/Lean/Meta/Tactic/Simp/SimpTheorems.lean
```diff
@@ -18,7 +18,7 @@ what action the user took which lead to this theorem existing in the simp set.
 -/
 inductive Origin where
   /-- A global declaration in the environment. -/
-  | decl (declName : Name)
+  | decl (declName : Name) (inv : Bool)
```
digama0, line 21:
```suggestion
  | decl (declName : Name) (inv := false)
```

src/Lean/Meta/Tactic/Simp/SimpTheorems.lean
```diff
@@ -42,9 +42,9 @@ inductive Origin where
 
 /-- A unique identifier corresponding to the origin. -/
 def Origin.key : Origin → Name
-  | .decl declName => declName
+  | .decl declName _ => declName
   | .fvar fvarId => fvarId.name
-  | .stx id _ => id
+  | .stx id _  => id
```
digama0, line 47:
```suggestion
  | .stx id _ => id
```

digama0 (approved):

semorrison:
Phew, lots of downstream changes required for this one:
* https://github.com/JLimperg/aesop/pull/72
* https://github.com/leanprover/std4/pull/284

semorrison:
I've checked that downstream projects survive okay. There is some unfortunate copypasta there that ideally we can clean up later.

digama0 (approved):

PR #2625, opened 2023-10-05
refactor: simplify excluded middle proof

james-smith-a8e6:
The current `Classical.em` proof isn't that easy to follow. Since Lean has quotients, we can take a two-element type like `Bool`, and quotient it by `x = y ∨ P`. Then, the surjection `Bool → A` splits and we use the fact that `Bool` has decidable equality to decide `P`.

Based on proof at https://ncatlab.org/nlab/show/Diaconescu-Goodman-Myhill+theorem.


leodemoura:
Marking as low priority. Lean has proof irrelevance. Thus, this commit does not benefit users.

TwoFX:
As [discussed](https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/Excluded-middle.20without.20funext/with/611594405), reducing axioms beyond the three standard axioms is not a goal of the Lean standard library, and neither is the proofs being useful pedagogically. So I'm closing this PR.

PR #2629, opened 2023-10-05
Tag names

david-christiansen:

Right now, Lean's error message when providing a nonexistent case tag is somewhat unhelpful. It says `tag not found`.

With these changes, Lean instead lists the available tags that the user could have written, making the message more helpful. I also think that this phrasing is somewhat more likely to inform a new user that a "tag" is the name of the case that comes out of `cases` or `induction`.

### DIFF
--- a/src/Lean/Elab/Tactic/BuiltinTactic.lean
+++ b/src/Lean/Elab/Tactic/BuiltinTactic.lean
@@ -369,11 +369,39 @@ private def getCaseGoals (tag : TSyntax ``binderIdent) : TacticM (MVarId × List
   let gs ← getUnsolvedGoals
   let g ← if let `(binderIdent| $tag:ident) := tag then
     let tag := tag.getId
-    let some g ← findTag? gs tag | throwError "tag not found"
+    let some g ← findTag? gs tag | notFound gs tag
     pure g
   else
     getMainGoal
   return (g, gs.erase g)
+
+where
+  -- When the case tag is not found, construct a message that tells
+  -- the user what they could have written
+  notFound (available : List MVarId) (tag : Name) := do
+    let firstLine := m!"Case tag {showTagName tag} not found."
+    -- We must filter out the anonymous name because there may be an
+    -- anonymous goal, but users shouldn't be mistakenly encouraged
+    -- to write `case anonymous`
+    match (← available.mapM getUserName).filter (· ≠ Name.anonymous) with
+    | [] =>
+      throwError "{firstLine}\n\nThere are no cases to select."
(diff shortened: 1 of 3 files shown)


src/Lean/Elab/Tactic/BuiltinTactic.lean
```diff
@@ -369,12 +369,36 @@ private def getCaseGoals (tag : TSyntax ``binderIdent) : TacticM (MVarId × List
...
   let g ← if let `(binderIdent| $tag:ident) := tag then
     let tag := tag.getId
-    let some g ← findTag? gs tag | throwError "tag not found"
+    let some g ← findTag? gs tag | notFound tag gs
     pure g
   else
     getMainGoal
   return (g, gs.erase g)
 
+  where
```
Kha, line 378: Style nit: we usually dedent the `where` block

tests/lean/caseSuggestions.lean
```diff
@@ -0,0 +1,45 @@
+/-!
+
+  # Provide case alternatives in "nonexistent tag" message
+
+  Test that the available case tags are suggested when a nonexistent
+  tag is requested. -/
```
Kha, line 6: Great!

src/Lean/Elab/Tactic/BuiltinTactic.lean
```diff
@@ -369,12 +369,36 @@ private def getCaseGoals (tag : TSyntax ``binderIdent) : TacticM (MVarId × List
...
+    -- When the case tag is not found, construct a message that tells
+    -- the user what they could have written
+    notFound (tag : Name) (available : List MVarId) := do
+      let firstLine := m!"Case tag {tag} not found."
```
Kha, line 382: I believe we usually put names like `{tag}` from the input in single quotes in messages. Not sure how consistent we are about it though, so I'll leave making the call on it to you.

Kha (approved):

src/Lean/Elab/Tactic/BuiltinTactic.lean
david-christiansen: I would certainly prefer this - I'm updating the code now.

src/Lean/Elab/Tactic/BuiltinTactic.lean
david-christiansen: Thanks, I'll work on my Haskell accent :)

david-christiansen: Done

PR #2633, opened 2023-10-07
fix: eliminate widestring uses

Kha:
This is always incorrect and in fact led to Windows users with Unicode account names being unable to use Lean

### DIFF
--- a/src/runtime/io.cpp
+++ b/src/runtime/io.cpp
@@ -690,11 +690,10 @@ extern "C" LEAN_EXPORT obj_res lean_io_remove_file(b_obj_arg fname, obj_arg) {
 
 extern "C" LEAN_EXPORT obj_res lean_io_app_path(obj_arg) {
 #if defined(LEAN_WINDOWS)
-    HMODULE hModule = GetModuleHandleW(NULL);
-    WCHAR path[MAX_PATH];
-    GetModuleFileNameW(hModule, path, MAX_PATH);
-    std::wstring pathwstr(path);
-    std::string pathstr(pathwstr.begin(), pathwstr.end());
+    HMODULE hModule = GetModuleHandle(NULL);
+    char path[MAX_PATH];
+    GetModuleFileName(hModule, path, MAX_PATH);
+    std::string pathstr(path);
     // Hack for making sure disk is lower case
     // TODO(Leo): more robust solution
     if (pathstr.size() >= 2 && pathstr[1] == ':') {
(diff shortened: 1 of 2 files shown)


src/util/path.cpp
```diff
@@ -42,11 +42,10 @@ static char g_path_sep     = ';';
...
 std::string get_exe_location() {
-    HMODULE hModule = GetModuleHandleW(NULL);
-    WCHAR path[MAX_PATH];
-    GetModuleFileNameW(hModule, path, MAX_PATH);
-    std::wstring pathstr(path);
-    return std::string(pathstr.begin(), pathstr.end());
+    HMODULE hModule = GetModuleHandle(NULL);
+    char path[MAX_PATH];
+    GetModuleFileName(hModule, path, MAX_PATH);
+    return std::string(path);
```
Kha, line 48: TODO: remove this entire file; I believe the only remaining uses are
```
leancpp> /build/source/util/shell.cpp:317:12: error: use of undeclared identifier 'lrealpath'
leancpp>     path = lrealpath(path);
leancpp>            ^
leancpp> /build/source/util/shell.cpp:318:23: error: use of undeclared identifier 'stem'
leancpp>     std::string pkg = stem(path);
leancpp>                       ^
leancpp> /build/source/util/shell.cpp:698:24: error: use of undeclared identifier 'read_file'
leancpp>             contents = read_file(mod_fn);
leancpp>                        ^
```

src/runtime/io.cpp
```diff
@@ -690,11 +690,10 @@ extern "C" LEAN_EXPORT obj_res lean_io_remove_file(b_obj_arg fname, obj_arg) {
 
 extern "C" LEAN_EXPORT obj_res lean_io_app_path(obj_arg) {
 #if defined(LEAN_WINDOWS)
-    HMODULE hModule = GetModuleHandleW(NULL);
-    WCHAR path[MAX_PATH];
-    GetModuleFileNameW(hModule, path, MAX_PATH);
-    std::wstring pathwstr(path);
-    std::string pathstr(pathwstr.begin(), pathwstr.end());
+    HMODULE hModule = GetModuleHandle(NULL);
```
eric-wieser, line 693:
```suggestion
    HMODULE hModule = GetModuleHandleA(NULL);
```
otherwise this does the wrong thing with `UNICODE` defined

eric-wieser: (and throughout)

Kha: Don't define `UNICODE` then!

PR #2636, opened 2023-10-08
chore: fix typos in comments

int-y1:
I ran `codespell src/Lean` and fixed the typos that were in a comment. (and also, update `tests/lean/docStr.lean.expected.out`)


Kha:
Nice, much more efficient than 70 individual PRs!

PR #2640, opened 2023-10-09
fix: simp: use withoutModifyingState rather than withoutModifyingStateWithInfoAndMessages in tacticToDischarge

### LINKED ISSUE #2634: `simp` fails when custom discharger makes no progress
### Prerequisites

### Steps to Reproduce

```
example
  {p: Prop}
  (h: True → p)
  : p := by 
  simp (discharger:= skip) [h]
```

**Expected behavior:** `simp` fails with message:

`simp made no progress` 

**Actual behavior:** `simp` fails with message: 

```
unsolved goals

case simp.discharger
p: Prop
h: True → p
⊢ True
```

### Versions

`nightly-2023-10-07`

### Additional Information

It's also would be useful to have an option to fallback to the default discharger when the custom provided one will be unable to close the goal. 

### Impact

kim-em:
Fixes #2634. When calling the discharger we should suppress error messages, to avoid having irrelevant discharger messages from failed invocations propagate to the user.


semorrison:
@leodemoura, do you know why this code was using `withoutModifyingStateWithInfoAndMessages`? There's no test that requires it. If it's okay to change to `withoutModifyingState` then that is an easy fix for #2634.

leodemoura:
@semorrison We had many changes in how we save info for tactics in the last 2 years. I am worried we don't have many tests at `tests/lean/interactive`. Could you please confirm that we have tests covering this change?

semorrison:
> @semorrison We had many changes in how we save info for tactics in the last 2 years. I am worried we don't have many tests at `tests/lean/interactive`. Could you please confirm that we have tests covering this change?

@leodemoura, I've added documentation to the test explaining what used to happen, and a comment in the code explaining why we want `withoutModifyingState` rather than `withoutModifyingStateWithInfoAndMessages`.

PR #2642, opened 2023-10-09
chore: make `Environment.add` private

kim-em:
This is currently stacked on top of #2617, because Aesop will need that to be merged first.


vleni:
What can we use in place of `Environment.add`?

digama0:
You should be using `addDecl` in the first place, the use cases for bypassing the kernel using `Environment.add` are very limited. That said, you can call it using an `extern`, as demonstrated here for lake.

Issue #2635, opened 2023-10-07
RFC: Allow trailing comma after the last element of a list


vleni:
### Proposal

Currently this is not legal Lean:
```lean
let li = [
  item1,
  item2,
]
```
but if it were legal, adding new elements would not entail changes to existing lines:
```lean
let li = [
  item1,
  item2,
  item3,
  item4,
]
```
This has two main benefits:
1. Machine based code generation is simpler
2. The diff generated by version control is easier to read when new elements are added. This is especially relevant to unit testing.

Trailing comma is already supported in Rust and Python.

### Community Feedback

Ideas should be discussed on [the Lean Zulip](https://leanprover.zulipchat.com) prior to submitting a proposal. Summarize all prior discussions and link them here.

[Discussion thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Trailing.20comma.20for.20lists)

James Gllicchio proposed that we could also have preceding commas:
```lean
let li = [
  , item1
  , item2
]
```

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others benefit from the changes in this proposal being added, please ask them to add :+1: to it.

leodemoura:
@Kha I think this is a good change, and I recall discussing it with you in the past. However, I don't remember why we didn't move forward. I don't have time to do it myself, but I will support a PR. BTW, let's ensure the trailing comma is accepted in other similar notation.

vleni:
> @Kha I think this is a good change, and I recall discussing it with you in the past. However, I don't remember why we didn't move forward. I don't have time to do it myself, but I will support a PR. BTW, let's ensure the trailing comma is accepted in other similar notation.

What do you mean by accepted in other similar notation?

semorrison:
e.g. if this is allowed, then `rw [x, y, ]` should be allowed as well.

Kha:
> However, I don't remember why we didn't move forward.

I am quite sure it was lack of sufficient motivation. So I agree with

> I don't have time to do it myself, but I will support a PR.

vleni:
> @Kha I think this is a good change, and I recall discussing it with you in the past. However, I don't remember why we didn't move forward. I don't have time to do it myself, but I will support a PR. BTW, let's ensure the trailing comma is accepted in other similar notation.

I have a PR here https://github.com/leanprover/lean4/pull/2643.

[...]

PR #2643, opened 2023-10-09
Allow trailing comma in tuples, lists, and tactics

lenianiva:

https://github.com/leanprover/lean4/issues/2635

### DIFF
--- a/tests/lean/StxQuot.lean.expected.out
+++ b/tests/lean/StxQuot.lean.expected.out
@@ -6,7 +6,7 @@ StxQuot.lean:8:12-8:13: error: unexpected token ')'; expected identifier or term
 "(«term_+_» <missing> \"+\" (num \"1\"))"
 "(«term_+_» <missing> \"+\" (num \"1\"))"
 "(«term_+_» (num \"1\") \"+\" (num \"1\"))"
-"(«term[_]» \"[\" [`x._@.UnhygienicMain._hyg.1 \",\"] \"]\")"
+StxQuot.lean:19:15-19:16: error: unexpected token ']'; expected term
 "(Term.fun \"fun\" (Term.basicFun [`a._@.UnhygienicMain._hyg.1] [] \"=>\" `a._@.UnhygienicMain._hyg.1))"
 "(Command.declaration\n (Command.declModifiers [] [] [] [] [] [])\n (Command.def\n  \"def\"\n  (Command.declId `foo._@.UnhygienicMain._hyg.1 [])\n  (Command.optDeclSig [] [])\n  (Command.declValSimple \":=\" (num \"1\") [])\n  []\n  []\n  []))"
 "[(Command.declaration\n  (Command.declModifiers [] [] [] [] [] [])\n  (Command.def\n   \"def\"\n   (Command.declId `foo._@.UnhygienicMain._hyg.1 [])\n   (Command.optDeclSig [] [])\n   (Command.declValSimple \":=\" (num \"1\") [])\n   []\n   []\n   []))\n (Command.declaration\n  (Command.declModifiers [] [] [] [] [] [])\n  (Command.def\n   \"def\"\n   (Command.declId `bar._@.UnhygienicMain._hyg.1 [])\n   (Command.optDeclSig [] [])\n   (Command.declValSimple \":=\" (num \"2\") [])\n   []\n   []\n   []))]"
--- a/tests/lean/StxQuot.lean
+++ b/tests/lean/StxQuot.lean
@@ -16,7 +16,7 @@ namespace Lean.Syntax
 #eval run $ let id := miss; `($id + 1)
 end Lean.Syntax
 #eval run `(1 + 1)
-#eval run `([x,])
+#eval run `([x+])
 #eval run $ `(fun a => a) >>= pure
 #eval run $ `(def foo := 1)
 #eval run $ `(def foo := 1 def bar := 2)


vleni:
I'm modifying the array syntax and for some reason this doesn't work
```lean
syntax "#[" withoutPosition(sepBy(term, ", ", allowTrailingSep)) "]" : term
```

digama0:
Because there isn't any disambiguator between the two optional fields, it just prefers the first parse in which `allowTrailingSep` is treated as a syntax element referring to a (nonexistent) parser. So you have to give the third argument explicitly if you want to pass the fourth one, that is, `sepBy(term, ",", ", ", allowTrailingSep)`. Or better yet, just use `term,*,?` since this isn't setting anything different from the default (see Init/Notation.lean#L238).

vleni:
> Because there isn't any disambiguator between the two optional fields, it just prefers the first parse in which `allowTrailingSep` is treated as a syntax element referring to a (nonexistent) parser. So you have to give the third argument explicitly if you want to pass the fourth one, that is, `sepBy(term, ",", ", ", allowTrailingSep)`. Or better yet, just use `term,*,?` since this isn't setting anything different from the default (see Init/Notation.lean#L238).

I tried `sepBy(term, ",", ", ", allowTrailingSep)` and it doesn't work either. I'll use `term,*,?` like what we have in lists.

digama0:
It should... one is literally just a macro for the other. What error do you get, and when?

vleni:
> It should... one is literally just a macro for the other. What error do you get, and when?

This is what I get when I write either of
```lean
-- Array/Basic.lean
syntax "#[" withoutPosition(term,*,?) "]" : term
syntax "#[" withoutPosition(sepBy(term, ",", ", ", allowTrailingSep)) "]" : term
```

The same schematic allowed me to implement trailing comma for lists though

```
Lean/Widget/TaggedText.lean:24:43: error: elaboration function for '«term#[_,]»' has not been implemented
  [Error pretty printing syntax: unknown constant 'term#[_,]'. Falling back to raw printer.]
  («term#[_,]»
   "#["
   [(Term.app `toJson._@.Lean.Widget.TaggedText._hyg.4 [`a._@.Lean.Widget.TaggedText._hyg.640])
    ","
    (Term.app `toJsonTaggedText._@.Lean.Widget.TaggedText._hyg.635 [`a._@.Lean.Widget.TaggedText._hyg.641])]
   "]")
make[7]: *** [../build/debug/stage1/lib/lean/Lean/Widget/TaggedText.olean] Error 1

```

digama0:
That's a bootstrapping issue, you will get the same error making any change to a builtin parser. See https://leanprover.zulipchat.com/#narrow/stream/341532-lean4-dev/topic/Bootstrapping.20issue.3F for a recent similar discussion, and see if setting parseQuotWithCurrentStage in stage0/src/stdlib_flags.h works for you.

vleni:
> That's a bootstrapping issue, you will get the same error making any change to a builtin parser. See https://leanprover.zulipchat.com/#narrow/stream/341532-lean4-dev/topic/Bootstrapping.20issue.3F for a recent similar discussion, and see if setting parseQuotWithCurrentStage in stage0/src/stdlib_flags.h works for you.

Would the unit tests pass if I just change the file in `stage0`?

digama0:
You should make the stage0 change (and commit it), then run `make update-stage0` which will clobber the change in addition to copying stage1 to stage0, then compile again and ensure the tests pass in the second run.

vleni:
> You should make the stage0 change (and commit it), then run `make update-stage0` which will clobber the change in addition to copying stage1 to stage0 (and commit this too as `chore: update stage0`), then compile again and ensure the tests pass in the second run. See https://github.com/leanprover/lean4/blob/master/doc/dev/bootstrap.md for more information.

I think I'm doing this correctly but it is not working:
1. Make the necessary stage 0 changes to `Array/Basic.lean`
2. Commit
3. `make update-stage0`

which gives
```lean
Lean/SubExpr.lean:198:21: error: elaboration function for '«term#[_,]»' has not been implemented
  [Error pretty printing syntax: unknown constant 'term#[_,]'. Falling back to raw printer.]
  («term#[_,]»
   "#["
   [(Term.app `toJson._@.Lean.SubExpr._hyg.1526 [`a._@.Lean.SubExpr._hyg.1870])
    ","
    (Term.app `toJson._@.Lean.SubExpr._hyg.1526 [`a._@.Lean.SubExpr._hyg.1871])]
   "]")
Lean/SubExpr.lean:198:21: error: elaboration function for '«term#[_,]»' has not been implemented
  [Error pretty printing syntax: unknown constant 'term#[_,]'. Falling back to raw printer.]
  («term#[_,]»
   "#["
   [(Term.app `toJson._@.Lean.SubExpr._hyg.1526 [`a._@.Lean.SubExpr._hyg.1872])
    ","
    (Term.app `toJson._@.Lean.SubExpr._hyg.1526 [`a._@.Lean.SubExpr._hyg.1873])]
   "]")
[    ] Building ../build/debug/stage1/lib/temp/Lean/Elab/Config.c
```
This fails even if I just write `withoutPosition(term,*)`, which should have identical behaviour as before, instead of `withoutPosition(term,*,?)`

digama0:
> This fails even if I just write `withoutPosition(term,*)`, which should have identical behaviour as before, instead of `withoutPosition(term,*,?)`

I believe the reason for the change is that the default name for the `term,*` variant is `term#[_]` instead of `term#[_,]` (this name is constructed by looking at the syntax used to specify the parser).

vleni:
> > This fails even if I just write `withoutPosition(term,*)`, which should have identical behaviour as before, instead of `withoutPosition(term,*,?)`
> 
> I believe the reason for the change is that the default name for the `term,*` variant is `term#[_]` instead of `term#[_,]` (this name is constructed by looking at the syntax used to specify the parser).

How come the same change for list was fine?

digama0:
Most likely because the derive handler (running stage0 code which is no longer correct) happens to be constructing arrays and not lists.

vleni:
> Most likely because the derive handler in question (running stage0 code which is no longer correct) happens to be constructing arrays and not lists.

so should I just leave the stage 0 code alone and add a separate syntax for arrays with a trailing comma?

vleni:
> Most likely because the derive handler in question (running stage0 code which is no longer correct) happens to be constructing arrays and not lists.

I dug into this a bit more and all of the errors occur in `Lean/Elab/Deriving/FromToJson.lean`, but I could not pinpoint exactly where does compilation fail.

src/Init/Notation.lean
```diff
@@ -471,28 +471,34 @@ which uses let bindings as intermediates as in
...
     let rec expandListLit (i : Nat) (skip : Bool) (result : TSyntax `term) : MacroM Syntax := do
       match i, skip with
       | 0,   _     => pure result
       | i+1, true  => expandListLit i false result
       | i+1, false => expandListLit i true  (← ``(List.cons $(⟨elems.elemsAndSeps.get! i⟩) $result))
-    if elems.elemsAndSeps.size < 64 then
-      expandListLit elems.elemsAndSeps.size false (← ``(List.nil))
+    let size := elems.elemsAndSeps.size
+    if size < 64 then
+      expandListLit size (isEven size) (← ``(List.nil))
```
digama0, line 501:
```suggestion
      expandListLit size (size % 2 == 0) (← ``(List.nil))
```

vleni: Doesn't work:
```
Init/Notation.lean:496:26: error: failed to synthesize instance
  HMod Nat Nat Nat
```

digama0: I got it to compile by moving this declaration to the beginning of `Init.Data.List.Basic` and making that file import `Init.Data.Nat.Div`

tests/compiler/array.lean
```diff
@@ -11,3 +11,7 @@ def main : IO Unit := do
   IO.println (f #[2, 3, 4])
   IO.println (g #[2, 3, 4])
   IO.println (h [2, 3, 4])
+
+  -- Trailing comma
+  --IO.println (g #[5, 6, 7,])
```
semorrison, line 16: Leave a note here?

tests/lean/rewrite.lean
```diff
@@ -34,7 +34,7 @@ subst y;
 exact rfl
 
 theorem ex4 (x y z) (h₁ : 0 + x = y) (h₂ : 0 + y = z) : x = z := by
-rewrite [appendAssoc] at *; -- Error
+rewrite [appendAssoc,] at *; -- Error
```
semorrison, line 37: Perhaps best to assemble all of these tests into a single file, so it's easier to understand? e.g. `test/lean/trailingComma`.

It would also be good to exercise it further. Have the tests demonstrate that whitespace (including new lines) before or after the comma has no effect. 

Possibly also a test that checks for the appropriate error message on a double trailing comma?

tests/compiler/array.lean
vleni: I'll just delete it

tests/lean/rewrite.lean
vleni: I moved all of them into one file

vleni:
@semorrison Could you take a look to see if this is ready to merge?

tests/lean/StxQuot.lean.expected.out
```diff
@@ -6,7 +6,7 @@ StxQuot.lean:8:12-8:13: error: unexpected token ')'; expected identifier or term
 "(«term_+_» <missing> \"+\" (num \"1\"))"
 "(«term_+_» <missing> \"+\" (num \"1\"))"
 "(«term_+_» (num \"1\") \"+\" (num \"1\"))"
-StxQuot.lean:19:15-19:16: error: unexpected token ']'; expected term
+"(«term[_]» \"[\" [`x._@.UnhygienicMain._hyg.1 \",\"] \"]\")"
```
Kha, line 9: Let's change this to a different invalid syntax like `[x+]` in the test file

Kha (commented):
LGTM, but just to double-check: are we happy with `(1,)` meaning the same as `(1)`? It has a special meaning (1-tuple) in Python but I don't think that has to affect the design of a statically typed language like Lean.

tests/lean/StxQuot.lean.expected.out
vleni: done

semorrison:
> LGTM, but just to double-check: are we happy with `(1,)` meaning the same as `(1)`? It has a special meaning (1-tuple) in Python but I don't think that has to affect the design of a statically typed language like Lean.

Hmm this is a good point. I think tuples should not allow trailing commas.

digama0:
I lean toward allowing this, with `(1,)` meaning the same as `(1)`, because it is useful for macro-style code to be able to just have consistent behavior in the base case, and tuples are iterated cases of `Prod` constructor with the base case just being the underlying type.

Kha (approved):

Kha:
Thanks!

PR #2645, opened 2023-10-10
chore: syntax documentation for `universe`, `open`, `export`, `variable`

AdrienChampion:
* [x] Put an X in this bracket to confirm you have read the

Mentored by @digama0, adds syntax-documentation with examples to `universe`, `open`, `export`, and `variable`.

The documentation shows up when hovering over keywords, hopefully improving the experience for beginners.

### DIFF
--- a/src/Lean/Elab/BuiltinCommand.lean
+++ b/src/Lean/Elab/BuiltinCommand.lean
@@ -131,7 +131,7 @@ universes].
 Just as type arguments allow polymorphic definitions to be used at many different types, universe
 parameters, represented by universe variables, allow a definition to be used at any required level.
 While Lean mostly handles universe levels automatically, declaring them explicitly can provide more
-control when writing signatures. The universe keyword allows the declared universe variables to be
+control when writing signatures. The `universe` keyword allows the declared universe variables to be
 used in a collection of definitions, and Lean will ensure that these definitions use them
 consistently.
 
@@ -189,7 +189,7 @@ structure Pair (α : Type u) (β : Type v) : Type (max u v) where
 The command `export Some.Namespace (name₁ name₂)` makes `name₁` and `name₂`:
 
 - visible in the current namespace without prefix `Some.Namespace`, like `open`, and
-- visible from outside current namespace `N` as `N.name₁` and `N.name₂`.
+- visible from outside the current namespace `N` as `N.name₁` and `N.name₂`.
 
 ## Examples
 
@@ -250,7 +250,7 @@ The `open` command can be used in a few different ways:
   This works even if `def1` and `def2` are `protected`.
 
 * `open scoped Some.Namespace.Path1 Some.Namespace.Path2` **only** opens [scoped instances],
-  notations, and attributes from `Namespace1` and `Namespace2`; does **not** make any other name
+  notations, and attributes from `Namespace1` and `Namespace2`; it does **not** make any other name
   available.
 
 * `open <any of the open shapes above> in` makes the names `open`-ed visible only in the next
@@ -264,35 +264,36 @@ The `open` command can be used in a few different ways:
 
 ```lean
 /-- SKI combinators https://en.wikipedia.org/wiki/SKI_combinator_calculus -/
-namespace Ident.Path
+namespace Combinator.Calculus
   def I (a : α) : α := a
   def K (a : α) : β → α := fun _ => a
   def S (x : α → β → γ) (y : α → β) (z : α) : γ := x z (y z)
-end Ident.Path
+end Combinator.Calculus
 
 section
-  -- open everything under `Ident.Path`, *i.e.* `I`, `K` and `S` here
-  open Ident.Path
+  -- open everything under `Combinator.Calculus`, *i.e.* `I`, `K` and `S`,
+  -- until the section ends
+  open Combinator.Calculus
 
   theorem SKx_eq_K : S K x = I := rfl
 end
 
--- open everything under `Ident.Path` only for the next command (the next `theorem`, here)
-open Ident.Path in
+-- open everything under `Combinator.Calculus` only for the next command (the next `theorem`, here)
+open Combinator.Calculus in
 theorem SKx_eq_K' : S K x = I := rfl
 
 section
-  -- open only `S` and `K` under `Ident.Path`
-  open Ident.Path (S K)
+  -- open only `S` and `K` under `Combinator.Calculus`
+  open Combinator.Calculus (S K)
 
   theorem SKxy_eq_y : S K x y = y := rfl
 
   -- `I` is not in scope, we have to use its full path
-  theorem SKxy_eq_Iy : S K x y = Ident.Path.I y := rfl
+  theorem SKxy_eq_Iy : S K x y = Combinator.Calculus.I y := rfl
 end
 
 section
-  open Ident.Path
+  open Combinator.Calculus
     renaming
       I → identity,
       K → konstant
@@ -302,7 +303,7 @@ section
 end
 
 section
-  open Ident.Path
+  open Combinator.Calculus
     hiding S
 
   #check I


src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -138,6 +176,86 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
     aliases := aliases.push (currNamespace ++ id, declName)
   modify fun s => { s with env := aliases.foldl (init := s.env) fun env p => addAlias env p.1 p.2 }
 
+/-- Brings some definitions in scope.
+
+* `open Some.Namespace.Ident` brings all definitions in `Some.Namespace.Ident` in scope.
```
eric-wieser, line 181: Presumably this should also document `open foo bar`

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -138,6 +176,86 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
...
+* `open Some.Namespace.Ident (def1 def2)` only brings `Some.Namespace.Ident.def1` and
+  `Some.Namespace.Ident.def2` in scope.
+* `open Namespace renaming def1 → def1', def2 → def2'` brings `def1` and `def2` in scope under the
```
eric-wieser, line 184: Is there any way to have a specialized docstring show up for hovers on the "renaming" keyword?

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -223,6 +341,42 @@ private def replaceBinderAnnotation (binder : TSyntax ``Parser.Term.bracketedBin
...
 
+/-- Declares a list of typed variables.
+
+Introduces variables that can be used in definitions within the same scope. When a definition
+mentions a variable, Lean will add it as an argument of the definition. This process is also able to
+infer typeclass parameters.
+
+Variable declarations have the same flexibility as regular function paramaters. In particular they
+can be explicit, implicit, or instance implicit (in which case they can be anonymous). This can be
+changed, for instance one can turn explicit variable `x` into an implicit one with `variable {x}`.
```
eric-wieser, line 352: IIRC it's not advisable to use this feature in the same `variable` command as declaring other variables; is it worth mentioning that here?

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -223,6 +341,42 @@ private def replaceBinderAnnotation (binder : TSyntax ``Parser.Term.bracketedBin
   else
     return #[binder]
 
+/-- Declares a list of typed variables.
+
+Introduces variables that can be used in definitions within the same scope. When a definition
+mentions a variable, Lean will add it as an argument of the definition. This process is also able to
```
david-christiansen, line 347:
```suggestion
mentions a variable, Lean will add it as an argument of the definition. The `variable` command is also able to
```

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -223,6 +341,42 @@ private def replaceBinderAnnotation (binder : TSyntax ``Parser.Term.bracketedBin
   else
     return #[binder]
 
+/-- Declares a list of typed variables.
```
david-christiansen, line 344: I think that this docstring is also missing a motivation for why someone would want to use the feature. After all, one might think, Lean already generalized unbound names to implicit arguments! What gives?

I believe the answer is:
 * Lean doesn't always pick the correct order of the arguments, but that can matter in the presence of dependent types
 * Lean can't always infer the type, and it can be nicer to write it once rather than in many related definitions/declarations
 * Lean won't always add enough dependencies when complex dependent types are involved
 * Being explicit about variable types improves error messages
 * This feature allows overriding a globally-defined name to use it as an implicit argument again in the section

Did I miss any others?

Are there other use cases?

david-christiansen (changes requested):
This is a great improvement, thanks!

I think it can be even better with some small modifications. They come in a few categories:
1. Describing not just what the feature does, but also the problems that it solves for users
2. Using less-abstract examples that are more likely to be understood by our audiences
3. Reducing the technical programming-languages jargon
4. Picking syntactic metavariables that have less potential for confusion

Thanks again, and please let me know if there's anything unclear in my comments!

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -118,6 +118,21 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
 @[builtin_command_elab choice] def elabChoice : CommandElab := fun stx =>
   elabChoiceAux stx.getArgs 0
 
+/-- Declares a list of type universes.
```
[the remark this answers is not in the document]
eric-wieser: One answer to that question is "you need it when `autoImplicit` is `false`, which it is in mathlib".

When it's not, the answer is "it let's you choose the *order* of the universe variables.

src/Lean/Elab/BuiltinCommand.lean
AdrienChampion: Not as far as I know, no :/

src/Lean/Elab/BuiltinCommand.lean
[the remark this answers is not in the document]
AdrienChampion: I don't know about the precis rules either. My experience is that I want type universe variables when I want to control precisely what the type are, most likely because there are relatively complex type universe constraints. Another one is when I want to be sure what universe a type is in.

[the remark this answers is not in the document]
AdrienChampion: I think the broader question is whether this is the right place to have extensive discussions about these concepts. When we document `namespace` for instance, do we want a full explanation of the namespace system? I don't have an answer, one could argue it would be better to point to the relevant part of the reference document (type universes, namespaces...) and only provide a relatively short of the syntax here.

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -126,6 +141,29 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
   | Except.ok env   => setEnv env
   | Except.error ex => throwError (ex.toMessageData (← getOptions))
 
+/-- Brings some definitions in scope and exports them.
+
+## Examples
+
+```lean
+namespace Ident.Path
```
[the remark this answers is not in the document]
AdrienChampion: fixed in 612d6b4

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -138,6 +176,86 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
     aliases := aliases.push (currNamespace ++ id, declName)
   modify fun s => { s with env := aliases.foldl (init := s.env) fun env p => addAlias env p.1 p.2 }
 
+/-- Brings some definitions in scope.
```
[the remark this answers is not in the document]
AdrienChampion: As I said somewhere else in this PR, I do think there is an issue of where the detailed discussion happens. To take the `namespace` example again, I would argue that the docstring should discuss the syntax and how it works, but not include a lengthy discussion of what a namespace is. Same for `universe`
One of the main arguments I see is that the manual must, I think, feature lengthy discussions about namespaces, universes and so on anyways. So personally I'd rather point to that in the docstring than maintaining two versions.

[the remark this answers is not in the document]
AdrienChampion: As you said, if we go that route we should probably postpone this PR until we have something to point to in the manual

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -138,6 +176,86 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
...
+
```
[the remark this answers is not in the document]
AdrienChampion: fixed in caa91b43

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -223,6 +341,42 @@ private def replaceBinderAnnotation (binder : TSyntax ``Parser.Term.bracketedBin
   else
     return #[binder]
 
+/-- Declares a list of typed variables.
+
+Introduces variables that can be used in definitions within the same scope. When a definition
```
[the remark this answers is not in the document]
AdrienChampion: > Also, this is when the LHS of the definition mentions the name, right?

You mean RHS right? As in the (type) definition's body?

```lean
variable (n : Nat)

def addOne : Nat := n + 1
#check addOne
-- addOne (n : Nat) : Nat

variable (MyType : Type)

structure Opt where
  field : MyType
#check Opt
-- Opt (MyType : Type) : Type
```

src/Lean/Elab/BuiltinCommand.lean
AdrienChampion: Well I would say the main use case is to make (type) definitions much easier to write when you're writing a bunch of things that have roughly the same signature, especially when you have typeclass parameters all over the place.

src/Lean/Elab/BuiltinCommand.lean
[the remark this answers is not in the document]
david-christiansen: I meant LHS but I was just wrong :-)

src/Lean/Elab/BuiltinCommand.lean
david-christiansen: This would be a great second sentence in your docstring :-)

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -138,6 +173,106 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
     aliases := aliases.push (currNamespace ++ id, declName)
   modify fun s => { s with env := aliases.foldl (init := s.env) fun env p => addAlias env p.1 p.2 }
 
+/-- Makes names from other namespaces visible without writing the namespace prefix.
+
+This command only makes *public* names visible, *i.e.* it does not bring `protected` and `private`
+names in scope.
```
eric-wieser, line 179: This is not true:
```lean
private def foo.bar := 1

open foo (bar)

#check bar
```

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -138,6 +173,106 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
...
+
+Names that are made available with `open` are visible within the current `section` or `namespace`
+block. The `open` command can be used in few different ways:
+
+* `open Some.Namespace.Path1 Some.Namespace.Path2` makes all names in `Some.Namespace.Path1` and
+  `Some.Namespace.Path2` available without the prefix, so that `Some.Namespace.Path1.x` and
+  `Some.Namespace.Path2.y` can be referred to by writing only `x` and `y`.
+* `open Some.Namespace.Path (def1 def2)` only makes `Some.Namespace.Path.def1` and
+  `Some.Namespace.Path.def2` available without the full prefix, so `Some.Namespace.Path.def3` would
+  be unaffected.
```
eric-wieser, line 218:
```suggestion
* `open Some.Namespace.Path (def1 def2)` only makes `Some.Namespace.Path.def1` and
  `Some.Namespace.Path.def2` available without the full prefix, so `Some.Namespace.Path.def3` would
  be unaffected. This works even if `def1` and `def2` are `protected`.
```

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -118,6 +118,50 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
 @[builtin_command_elab choice] def elabChoice : CommandElab := fun stx =>
   elabChoiceAux stx.getArgs 0
 
+/-- Declares one or more universe variable(s).
+
+Type universes are used in `Type u` and `Sort u` types. While Lean mostly handles universes
+automatically, explicitely declaring some eases the process of writing signatures and type
+definitions by giving us more control and factoring type universe declarations.
```
david-christiansen, line 125: We talked about this wording in the Zulip thread - it is still difficult for me to parse it. What about the suggestions we made over there?

src/Lean/Elab/BuiltinCommand.lean
david-christiansen: We should definitely clear this up.

david-christiansen (changes requested):
It looks like you've made great progress on it!

src/Lean/Elab/BuiltinCommand.lean
AdrienChampion: Right, I meant that `open My.Namespace` with no `(Ident1 Ident2 ...)`, consistently with your other remark on this topic. Definitely needs a clarification indeed.

src/Lean/Elab/BuiltinCommand.lean
AdrienChampion: Added, along with an example as I was worried it was not understandable unless you already know what it means

AdrienChampion: I'm a bit worried the example is too long for what it conveys, I'm interested in feedback

src/Lean/Elab/BuiltinCommand.lean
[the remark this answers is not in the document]
AdrienChampion: Well you can use them on either side actually, right?

```lean
variable {α : Type}

def len (l : List α) : Nat :=
  l.length
```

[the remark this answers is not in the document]
AdrienChampion: Did you want me to discuss this topic in the docstring?

I'm not completely sure what to say about it though

src/Lean/Elab/BuiltinCommand.lean
AdrienChampion: My bad, I thought I pushed this already.

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -138,6 +203,117 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
     aliases := aliases.push (currNamespace ++ id, declName)
   modify fun s => { s with env := aliases.foldl (init := s.env) fun env p => addAlias env p.1 p.2 }
 
+/-- Makes names from other namespaces visible without writing the namespace prefix.
+
+Names that are made available with `open` are visible within the current `section` or `namespace`
+block. The `open` command can be used in few different ways:
+
+* `open Some.Namespace.Path1 Some.Namespace.Path2` makes all names in `Some.Namespace.Path1` and
```
eric-wieser, line 211: I think it's clearer to write this up front:
```suggestion
* `open Some.Namespace.Path1 Some.Namespace.Path2` makes all non-`protected` names in `Some.Namespace.Path1` and
```

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -138,6 +203,117 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
...
+  (def1 def2)` but `def1`/`def2`'s names are changed to `def1'`/`def2'`.
+
+  This works even if `def1` and `def2` are `protected`.
+
+* `open Some.Namespace.Path hiding def1 def2` opens everything in `Some.Namespace.Path` except
+  `def1` and `def2`.
+
+  Only makes public names visible.
+
+* `open scoped Some.Namespace.Path1 Some.Namespace.Path2` opens [scoped instances], notations, and
```
eric-wieser, line 234: Do all of the commands above also do this?

AdrienChampion: So I moved the `open ... hiding ...` right after the `open Namespace` as the former is a refinement of the latter. I also mention upfront that it only works on non-protected names, as in one of your remarks on `open Namespace`.

```markdown
* `open Some.Namespace.Path1 Some.Namespace.Path2` makes all non-protected names in
  `Some.Namespace.Path1` and `Some.Namespace.Path2` available without the prefix, so that
  `Some.Namespace.Path1.x` and `Some.Namespace.Path2.y` can be referred to by writing only `x` and
  `y`.

* `open Some.Namespace.Path hiding def1 def2` opens all non-protected names in `Some.Namespace.Path`
  except `def1` and `def2`.
```

src/Lean/Elab/BuiltinCommand.lean
AdrienChampion: Agreed, done.

src/Lean/Elab/BuiltinCommand.lean
eric-wieser: To be clear, this comment is attached to the line about `open scoped`.

AdrienChampion: My bad. You're right, `open Namespace` also brings in scoped instances and notations

AdrienChampion: I pushed a fix for this. Before the variants, we now clarify that

```markdown
Names that are made available with `open` are visible within the current `section` or `namespace`
block. This makes referring to (type) definitions and theorems easier, but note that it can also make
[scoped instances], notations, and attributes from a different namespace available.
```

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -138,6 +176,86 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
...
+  name `def1'` and `def2'` respectively.
+* `open Namespace hiding def1 def2` opens everything in `Namespace` except `def1` and `def2`.
+* `open scoped Namespace1 Namespace2` opens scoped notations, instances, and attributes from
+  `Namespace1` and `Namespace2`, does **not** bring anything else in scope.
+
+
+## Examples
+
+```lean
+namespace Ident.Path
```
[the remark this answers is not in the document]
david-christiansen: Any thoughts on the name of the namespace?

david-christiansen (changes requested):
Almost there! I'm happy to finish this up if you're getting discouraged, but I think there's just a couple minor details and then I'll think we're good to go. Thank you so much for your patience!

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -118,6 +118,51 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
 @[builtin_command_elab choice] def elabChoice : CommandElab := fun stx =>
   elabChoiceAux stx.getArgs 0
 
+/-- Declares one or more universe variable(s).
+
+Type universes are used in `Type u` and `Sort u` types. While Lean mostly handles universes
+automatically, declaring them explicitly can provide more control when writing signatures. The
+`universe` keyword allows using the declared universes in a collection of definitions, and Lean will
+ensure that these definitions use them consistently.
```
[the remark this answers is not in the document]
AdrienChampion: The current version addresses this as discussed in private elsewhere

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -223,6 +399,73 @@ private def replaceBinderAnnotation (binder : TSyntax ``Parser.Term.bracketedBin
...
+
+namespace Logger
+  variable {Src : Type}
+
+  def empty : Logger Src where
+    trace := []
+
+  variable (log : Logger Src)
+
+  def len :=
```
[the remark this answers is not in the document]
AdrienChampion: Let me know if the new version addresses your remark

src/Lean/Elab/BuiltinCommand.lean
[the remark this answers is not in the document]
david-christiansen: I still think this is relevant - do you have an objection to this, or should I just go ahead and make the change?

src/Lean/Elab/BuiltinCommand.lean
david-christiansen: @eric-wieser Do you have a quick description of this available?

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -223,6 +399,73 @@ private def replaceBinderAnnotation (binder : TSyntax ``Parser.Term.bracketedBin
...
+    log.trace.length
+  def mapSrc (f : Src → Src') :=
+    log.trace.map fun (src, str) => (f src, str)
+
+  variable (src : Src) [BEq Src]
+
+  def filterSrc :=
```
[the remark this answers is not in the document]
david-christiansen: Looks good :) And yes. Thanks!

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -118,6 +118,74 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
...
+  universe above both `v` and `v'`, meaning it must be at least at level `(max v v') + 1`.
+
+  So `WitnessPair` is (at least) in `Sort ((max v v') + 1)`, which is the same as `Type (max v v')`.
+-/
+universe v v'
+structure WitnessPair : Type (max v v') where
+  α : Sort v
+  val : α
+  α' : Sort v'
+  val' : α'
```
eric-wieser, line 161: I think this would be better down the bottom in the `Examples` section; or just removed entirely; this is not really about the `universe` keyword at all at this  point, as `structure WitnessPair.{v,v'}` would demonstrate the same thing.

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -126,6 +194,26 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
   | Except.ok env   => setEnv env
   | Except.error ex => throwError (ex.toMessageData (← getOptions))
 
+/-- Adds aliases to the current namespace, pointing at names that are defined in some other
+namespace.
```
eric-wieser, line 198: I think "alias" is misleading, because Std has an `alias` command and it means something different.

src/Lean/Elab/BuiltinCommand.lean
eric-wieser: I don't have a description, but I have a pathological example:
```lean
class Foo (α : Type) where

def Bar (α : Type) [Foo α] := Nat

variable {α : Type} [Foo α]

-- now let's do a "binder update":

section
variable (α) (x : Bar α) -- fails; put your cursor before `x` to see why
end

section
variable (α)
variable (x : Bar α) -- same thing on two separate lines: ok
end
```

eric-wieser: So the short answer is "don't put them on the same line because you'll often get a confusing error message that leads you into even more of a mess".

eric-wieser: [No one knows](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/unexpected.20.60variable.60.20behaviour.3F/near/377995272) whether this behavior is a bug or by design.

eric-wieser: Update: I filed #2789, so I guess we just mention it in the docstring as justification for not mixing variables and binder updates.

eric-wieser: Update: I filed #2789, so I guess we just mention it in the docstring as justification for not mixing variables and binder updates.

src/Lean/Elab/BuiltinCommand.lean
AdrienChampion: Right, it attempts to give a more interesting universe example for newbies, but I agree that ideally this should be in a manual somewhere with a broader discussion of universes.

But the thing is, we already reference TPIL which does provide a discussion of universes, so I think we might as well just drop it.

src/Lean/Elab/BuiltinCommand.lean
AdrienChampion: Right, I think the new version addresses this issue, let me know what you think

src/Lean/Elab/BuiltinCommand.lean
AdrienChampion: Fixed

src/Lean/Elab/BuiltinCommand.lean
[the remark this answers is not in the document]
david-christiansen: For now, we don't have a comprehensive reference to point at, but once we do, I think the details clearly belong there. In the meantime, I don't think there's a need to be comprehensive here.

src/Lean/Elab/BuiltinCommand.lean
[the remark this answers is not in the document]
david-christiansen: I don't think we should postpone the PR for this, we should just merge the docs, and then later update things as our documentation system improves. The plan for the reference manual is to include docstrings for functions and syntax, so that we maintain a single canonical reference for everything, but this process will naturally result in reviewing and updating everything anyway.

src/Lean/Elab/BuiltinCommand.lean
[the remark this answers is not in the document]
david-christiansen: I don't think it needs anything here

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -118,6 +118,62 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
...
+On a more technical note, using a universe variable only in the right-hand side of a definition
+causes an error if the universe has not been declared previously.
+
+```lean
+def L₁.{u} := List (Type u)
+
+-- def L₂ := List (Type u) -- error: `unknown universe level 'u'`
+
+universe u
+universe L₃ := List (Type u)
```
digama0, line 161: typo?

eric-wieser:
```suggestion
def L₃ := List (Type u)
```

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -126,6 +182,30 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
   | Except.ok env   => setEnv env
   | Except.error ex => throwError (ex.toMessageData (← getOptions))
 
+/-- Adds names from other namespaces to the current namespace.
+
+The command `export Some.Namespace (name₁ name₂)` makes `name₁` and `name₂`
```
eric-wieser, line 187:
```suggestion
The command `export Some.Namespace (name₁ name₂)` makes `name₁` and `name₂`:
```

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -126,6 +182,30 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
...
+end Morning.Sky
+
+namespace ReExport
+  export Morning.Sky (star)
+  -- `star` is now in scope
+  #check star
+end ReExport
+
+-- `star` is visible in `ReExport`
+#check ReExport.star
```
digama0, line 206:
```suggestion
namespace Evening.Sky
  export Morning.Sky (star)
  -- `star` is now in scope
  #check star
end Evening.Sky

-- `star` is visible in `Evening.Sky`
#check Evening.Sky.star
```
I feel like it defeats the point of the "morning star" reference if you don't use "evening star" as the re-export name.

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -118,6 +118,62 @@ private partial def elabChoiceAux (cmds : Array Syntax) (i : Nat) : CommandElabM
 @[builtin_command_elab choice] def elabChoice : CommandElab := fun stx =>
   elabChoiceAux stx.getArgs 0
 
+/-- Declares one or more universe variables.
+
+`Prop`, `Type`, `Type u` and `Sort u` are types that classify other types, also known as
```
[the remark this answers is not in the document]
AdrienChampion: fixed

src/Lean/Elab/BuiltinCommand.lean
AdrienChampion: fixed

src/Lean/Elab/BuiltinCommand.lean
AdrienChampion: fixed

src/Lean/Elab/BuiltinCommand.lean
AdrienChampion: fixed

src/Lean/Elab/BuiltinCommand.lean
```diff
@@ -225,6 +416,102 @@ private def replaceBinderAnnotation (binder : TSyntax ``Parser.Term.bracketedBin
...
+variable (Src : Type)
+
+structure Logger where
+  trace : List (Src × String)
+#check Logger
+-- Logger (Src : Type) : Type
+
+namespace Logger
+  -- switch `Src : Type` to be implicit until the `end Logger`
+  variable {Src}
```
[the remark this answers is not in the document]
AdrienChampion: Is it? I agree for the first `namespace` in a file if everything is in that namespace, as you don't want to indent the whole file (and there's no closing delimiter anyways).

But besides that I think indentation is important, especially for beginners. I find forcing an incorrect indentation for a block is very confusing, especially when the block's delimiters are not recognizable characters and are instead just words.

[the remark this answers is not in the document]
digama0: See #2681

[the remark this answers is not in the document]
AdrienChampion: Keeping this style until #2681 is settled

src/Lean/Elab/BuiltinCommand.lean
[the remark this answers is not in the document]
david-christiansen: I did it

PR #2646, opened 2023-10-10
Support LSP encoding negotiation

david-christiansen:
In version 3.17, LSP introduced the ability for clients and servers to [negotiate which encoding they'd like to use](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#serverCapabilities) for source positions. In previous versions of LSP, source positions were always expressed in terms of variable-width 16-bit Unicode, which many clients have not properly implemented.

These changes allow Lean to negotiate which position style should be used. In the short term, this is a workaround for [a long-standing lsp-mode issue](https://github.com/emacs-lsp/lsp-mode/issues/2080) that has bitten a number of Lean Emacs users - given the repeated offers of support in that issue's thread that were ignored by the maintainer, I figured it'd be faster to just fix it here. In the long term, this will also make it easier for others to build innovative tools on top of Lean, as it removes one of the trickiest parts of LSP to get right. Lean code is more likely to be affected by 16-bit Unicode problems than other languages, due to the heavy idiomatic use of high-range Unicode for mathematics.

This is the first time I've had to work around [bootstrapping issues](https://leanprover.zulipchat.com/#narrow/stream/341532-lean4-dev/topic/Bootstrapping.20issue.3F), so this includes an updated `stage0`. In case I did something silly with it, I haven't squashed any of my "wip" commits - that should definitely be done prior to merging! You'll also notice lots of bogus little commits, so please ignore the commit history's structure. I'll squash once I get the go-ahead.

I also updated the LSP tests to take Unicode into account, and added the ability to run them with a variety of position encodings using a configuration line at the top.

Summary of changes:

 * Added module `Lean.Data.Lsp.PositionEncodingKind` - it's its own module because it ends up being used in a few places, and this avoids tricky import cycles.
 * Renamed module `Lean.Data.Lsp.Utf16` to `Lean.Data.Lsp.PositionEncoding` and added functions that work with 8- and 32-bit Unicode.
 * Added versions of `Lean.Data.Lsp.Position` and `Lean.Data.Lsp.Range` that contain all three supported encodings of column positions, allowing them to be generated when the text is present and avoiding a tricky global invariant (the code already did LSP UTF-16 conversions for the same reason).
 * Added Unicode support to the LSP testing infrastructure, including the ability to select an encoding (see tests/lean/interactive/run.lean and tests/lean/interactive/fancyUnicode.lean).

Everything seems to work well in Emacs now, which it didn't before, and in VS Code.

### DIFF
--- a/stage0/stdlib/Lean/Data/Lsp/PositionEncoding.c
+++ b/stage0/stdlib/Lean/Data/Lsp/PositionEncoding.c
@@ -259,20 +259,26 @@ x_5 = lean_unsigned_to_nat(0u);
 x_6 = lean_nat_dec_eq(x_2, x_5);
 if (x_6 == 0)
 {
-lean_object* x_7; lean_object* x_8; lean_object* x_9; uint32_t x_10; lean_object* x_11; lean_object* x_12; lean_object* x_13; 
+lean_object* x_7; lean_object* x_8; lean_object* x_9; uint32_t x_10; lean_object* x_11; lean_object* x_12; 
 x_7 = lean_unsigned_to_nat(1u);
 x_8 = lean_nat_sub(x_2, x_7);
+lean_dec(x_2);
 x_9 = lean_string_utf8_next(x_1, x_3);
 x_10 = lean_string_utf8_get(x_1, x_3);
-x_11 = l___private_Lean_Data_Lsp_PositionEncoding_0__String_csize16(x_10);
+lean_dec(x_3);
+x_11 = l___private_Lean_Data_Lsp_PositionEncoding_0__String_csize8(x_10);
 x_12 = lean_nat_add(x_4, x_11);
 lean_dec(x_11);
-x_13 = l___private_Lean_Data_Lsp_PositionEncoding_0__String_codepointPosToUtf16PosFromAux(x_1, x_8, x_9, x_12);
-return x_13;
+lean_dec(x_4);
+x_2 = x_8;
+x_3 = x_9;
+x_4 = x_12;
+goto _start;
 }
 else
 {
-lean_inc(x_4);
+lean_dec(x_3);
+lean_dec(x_2);
 return x_4;
 }
 }
@@ -282,9 +288,6 @@ _start:
 {
 lean_object* x_5; 
 x_5 = l___private_Lean_Data_Lsp_PositionEncoding_0__String_codepointPosToUtf8PosFromAux(x_1, x_2, x_3, x_4);
-lean_dec(x_4);
-lean_dec(x_3);
-lean_dec(x_2);
 lean_dec(x_1);
 return x_5;
 }
@@ -303,8 +306,6 @@ _start:
 {
 lean_object* x_4; 
 x_4 = l_String_codepointPosToUtf8PosFrom(x_1, x_2, x_3);
-lean_dec(x_3);
-lean_dec(x_2);
 lean_dec(x_1);
 return x_4;
 }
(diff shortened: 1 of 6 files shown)


david-christiansen:
It looks like I'll have to update `Std` to keep compatibility here. What's the way to prepare a PR to it that the Mathlib compatibility check will take into account?

david-christiansen:
!bench

src/Lean/Data/Lsp/PositionEncoding.lean
```diff
@@ -0,0 +1,140 @@
...
+import Lean.Data.Position
+
+
+/-! LSP uses UTF-16 for indexing, so we need to provide some primitives
+to interact with Lean strings using UTF-16 indices. -/
+
+namespace Char
+
+def utf16Size (c : Char) : UInt32 :=
+  if c.val ≤ 0xFFFF then 1 else 2
```
Vtec234, line 20: Bikeshed, but is `Lean.Data.Lsp.PositionEncoding` the right name for this? The module contains a bunch of general UTF utilities that are not about LSP positions. Maybe these could be moved to something like `Lean.Data.Utf`?

david-christiansen: I think that the underlying principle would be "When the compiler needs a thing that seems useful, make it available in a plausible module that people can import." I don't really believe this principle - I think the result will be either a compiler that contains too many things that it doesn't really need, or an incomplete API that's full of holes for other uses.

It could be that we need a relatively complete Unicode encodings API - then we might as well do it, as we've done for e.g. lists and strings. But so far, we seem to just need a little bit, so I'd rather keep them next to the code that uses them. Perhaps it should be marked `private` though :-)

david-christiansen: (also, LSP positions are all about UTF encodings, due to the decision of the protocol designers to mandate UTF-8 for messages, and UTF-16 for source positions - a correct LSP implementation needs lots of this kind of thing)

Vtec234: That's fair enough! I wouldn't `private` things though, historically that mostly leads to annoyed people copy-pasting code into their own repository.

semorrison:
> It looks like I'll have to update `Std` to keep compatibility here. What's the way to prepare a PR to it that the Mathlib compatibility check will take into account?

Apologies, @david-christiansen, that I never answered this. I'm finally getting to the end of the backlog of `breaks-mathlib` labels. :-)

Things you need to know to defeat the dreaded `breaks-mathlib` label:
* You need to rebase your PR onto `origin-nightly`. (Just once, it doesn't matter that `nightly` will move on, although you can keep rebasing if otherwise needed.)
* This will automatically create a `lean-pr-testing-NNNN` branch at Mathlib, on which adaptations can be made (or, realistically, you can ask me to make adaptations on!)
* If it also causes breakages in Std, we need to manually create a `lean-pr-testing-NNNN` branch at Std, based off the `nightly-testing-YYYY-MM-DD` branch of `Std` for whichever date `script/most-recent-nightly-tag.sh` reports in your Lean PR. (This step is rarer, so not as automated yet.) You then need to change the `lakefile.lean` in the Mathlib `lean-pr-testing-NNNN` branch to point to the `lean-pr-testing-NNNN` branch of Std, and run `lake update std`.

There is further documentation at https://leanprover-community.github.io/contribute/tags_and_branches.html, although it is more descriptive of the setup rather than instructional.

As this PR is currently broken in Lean CI, I'm just going to remove the `breaks-mathlib` label. Please feel free to ping me if/when you want Mathlib CI to resume, of the instructions above are insufficient.

PR #2644, opened 2023-10-10
isDefEq cache for terms not containing metavariables.

leodemoura:

PR #2647, opened 2023-10-10
fix: false positives in defeq cache

Kha:
This reproduces at least one error seen in https://github.com/leanprover-community/mathlib4/pull/7606, demonstrating that these are most likely false cache positives in the *old* code, not in the new code.


Kha:
Superseded by [a84dad5274872dd4d0f9f91fc3d72226983c621e](https://github.com/leanprover/lean4/pull/2644/commits/a84dad5274872dd4d0f9f91fc3d72226983c621e)

PR #2648, opened 2023-10-10
Cancel outstanding tasks on document edit in the language server

Kha:
Dramatically reduces system load and increases responsiveness of the language server. In my tests, I was not able to get the server above 100% CPU anymore.

The `check_system` calls inside pure Lean functions are a bit problematic. I did not find any process terminations from cancellation anymore after blacklisting two of them in the last commit but I may have missed some of them.

### DIFF
--- a/src/Lean/CoreM.lean
+++ b/src/Lean/CoreM.lean
@@ -199,7 +199,7 @@ instance [MetaEval α] : MetaEval (CoreM α) where
 protected def withIncRecDepth [Monad m] [MonadControlT CoreM m] (x : m α) : m α :=
   controlAt CoreM fun runInBase => withIncRecDepth (runInBase x)
 
-def checkInterrupted : CoreM Unit := do
+@[inline] def checkInterrupted : CoreM Unit := do
   if (← IO.checkCanceled) then
     -- should never be visible to users!
     throw <| Exception.error .missing "elaboration interrupted"


src/kernel/kernel_exception.h
```diff
@@ -145,21 +146,6 @@ an `kernel_exception` or `exception`. Then, convert result into `Except KernelEx
...
-2  | declTypeMismatch (env : Environment) (decl : Declaration) (givenType : Expr)
-3  | declHasMVars     (env : Environment) (name : Name) (expr : Expr)
-4  | declHasFVars     (env : Environment) (name : Name) (expr : Expr)
-5  | funExpected      (env : Environment) (lctx : LocalContext) (expr : Expr)
-6  | typeExpected     (env : Environment) (lctx : LocalContext) (expr : Expr)
-7  | letTypeMismatch  (env : Environment) (lctx : LocalContext) (name : Name) (givenType : Expr) (expectedType : Expr)
-8  | exprTypeMismatch (env : Environment) (lctx : LocalContext) (expr : Expr) (expectedType : Expr)
-9  | appTypeMismatch  (env : Environment) (lctx : LocalContext) (app : Expr) (funType : Expr) (argType : Expr)
-10 | invalidProj      (env : Environment) (lctx : LocalContext) (proj : Expr)
-11 | other            (msg : String)
```
Kha, line 161: This listing was outdated and is replicated below

Kha:
!bench

collares:
@Kha This was reverted due to crashes, and I have a worker stack trace at https://gist.github.com/collares/ae3a8d5b5f7c6976b8d46b5bfb6e945b. It seems like the kernel calls `check_interrupted()` directly in [`for_each_fn::apply`](https://github.com/leanprover/lean4/blob/master/src/kernel/for_each_fn.cpp#L65).

gebner:
FWIW, https://github.com/leanprover/lean4/pull/2036#issuecomment-1382659486

PR #2651, opened 2023-10-10
feat: lake: `--no-build` option to exit before a build

tydeu:
@Kha Is the `IO.Process.exit` signature change fine?

### DIFF
--- a/src/lake/Lake/Build/Common.lean
+++ b/src/lake/Lake/Build/Common.lean
@@ -31,6 +31,8 @@ of the `traceFile`. If rebuilt, save the new `depTrace` to the `tracefile`.
 @[inline] def buildUnlessUpToDate [CheckExists ι] [GetMTime ι] (info : ι)
 (depTrace : BuildTrace) (traceFile : FilePath) (build : JobM PUnit) : JobM PUnit := do
   unless (← depTrace.checkUpToDate info traceFile) do
+    if (← getNoBuild) then
+      IO.Process.exit noBuildCode.toUInt8
     build
     depTrace.writeToFile traceFile
 
@@ -69,6 +71,8 @@ def buildFileUnlessUpToDate (file : FilePath)
   if (← depTrace.checkUpToDate file traceFile) then
     fetchFileTrace file
   else
+    if (← getNoBuild) then
+      IO.Process.exit noBuildCode.toUInt8
     build
     depTrace.writeToFile traceFile
     return .mk (← cacheFileHash file) (← getMTime file)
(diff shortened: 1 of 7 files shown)


Kha:
> @Kha Is the `IO.Process.exit` signature change fine?

Where is the motivation? 3 doesn't strike me as a particularly large uint.

tydeu:
@Kha My motivation was consistent with the rest of the exit-code-related API (which deals with `UInt32` and thus surprised me when a `UInt32` did not work for `exit`). It is also in line with C [`exit`](https://en.cppreference.com/w/c/program/exit) and the Rust [`exit`](https://doc.rust-lang.org/std/process/fn.exit.html) (which, to my understanding much of the rest of the API is inspired by). Windows also has a number of standard >255 exit codes, so this could easily pose a problem for future Windows applications. All in all, since it came up and this seem like a bug/oversight to me which could be easily fixed, it seemed worth fixing.

Kha:
I look at these functions and see they both take a signed int. Which is not a huge difference but significant enough to make me hesitate on what the correct Lean signature should be, and makes me wonder whether sufficient care has been applied in the making of this commit. Taken together, it makes me think that there is no reason for that commit to be in this PR.

Kha:
The Lake changes look good!

tydeu:
@Kha 

> I look at these functions and see they both take a signed int. Which is not a huge difference but significant enough to make me hesitate on what the correct Lean signature should be

This concern confuses me. Lean does not have signed fixed-width integers? Furthermore, `Child.wait` already represents a signed integer via `UInt32` by explicitly casting the technically signed exit status of  `waitpid`  on Linux to `unsigned`. And `exit` itself was already representing the signed exit status as an unsigned `UInt8`? To me, this seems like a strict improvement (after all, the one other non-constant also would prefer a `UInt32`) and one that also has the advantage of being ABI-compatible with any shift in the future to primitive signed integers.

Is the main thrust here that the change should be in a separate PR? Or is this something we really do not want to change?

Kha:
> this seems like a strict improvement

That's the thing, isn't it -- going from UInt8 to UInt32 or to a future Int32 can be considered a strict improvement, but going from UInt32 to Int32 in the future because we didn't plan ahead sufficiently is slightly less nice. Well, I was about to comment before I saw your revert that after more consideration, being consistent with the currently existing API seems fine compared to waiting for some types in the future to appear. But all this discussion I would have preferred to hold at a time and place with a use case in hand.

david-christiansen:
What is the use case for this feature?

Kha:
@david-christiansen The immediate use is https://github.com/leanprover/lean4/issues/2154#issuecomment-1655471667; a more general `--dry-run` flag that also provides nice output is I think a staple of any build system

PR #2652, opened 2023-10-10
fix: treat pretty-printed names as strings

Vtec234:
I initially expected `Name`s to always faithfully represent internal data, in particular that a name with macro scopes would have a form such as  `` `foo._@.Module._hyg.1 ``, and that tombstones would only appear in types that represent pretty-printed output such as as `String` or `Format`. However, that is not what happens. We have `sanitizeNames` which rewrites the `userName` field of local hypotheses to be `Name.str .anonymous "blah✝"`.

Then in the server code, we put these into `names : Array Name`. This works fine for displaying in the infoview, but if we try to deserialize an `InteractiveHypothesisBundle` inside an RPC method for widget purposes, the `FromJson Name` instance blows up in `String.toName`.

I think my preferred solution is to, rather than 'fix' `String.toName` to accept these names with tombstones, stop pretending that they are actual `Name`s and re-type `InteractiveHypothesisBundle.names : Array String`. This should be a backwards-compatible change w.r.t. infoview code as the JSON representation is a string in either case. It is not backwards compatible w.r.t. meta code that uses this field.

See also [Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/341532-lean4-dev/topic/What's.20in.20a.20sanitized.20.60Name.60.3F).

### DIFF
--- a/src/Lean/Widget/InteractiveGoal.lean
+++ b/src/Lean/Widget/InteractiveGoal.lean
@@ -17,8 +17,10 @@ open Server
 as `h₁ h₂ : α`. We call this a 'hypothesis bundle'. We use `none` instead of `some false` for
 booleans to save space in the json encoding. -/
 structure InteractiveHypothesisBundle where
-  /-- The user-friendly name for each hypothesis. -/
-  names : Array Name
+  /-- The user-friendly name for each hypothesis.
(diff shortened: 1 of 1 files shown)


Vtec234:
awaiting-review

Kha (approved):

PR #2654, opened 2023-10-11
chore: add axiom for tracing use of reduceBool / reduceNat

kim-em:
Adds an axiom so `#print axioms` can track use of `reduceBool` and `reduceNat`.

### DIFF
--- a/src/Init/Core.lean
+++ b/src/Init/Core.lean
@@ -1613,6 +1613,11 @@ namespace Lean
 /-! # Kernel reduction hints -/
 
 /--
+Depends on the correctness of the Lean compiler, interpreter, and all `[implemented_by ...]` and `[extern ...]` annotations.
+-/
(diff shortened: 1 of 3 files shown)


src/Init/Core.lean
```diff
@@ -1612,6 +1612,11 @@ class Antisymm {α : Sort u} (r : α → α → Prop) where
 namespace Lean
 /-! # Kernel reduction hints -/
 
+/--
+A placeholder axiom so `#print axioms` can report use of `reduceBool` and `reduceNat`.
+-/
+axiom trustedCodeBase : False
```
digama0, line 1618: This isn't a great name for an axiom, it's like naming an axiom `axiom`. Maybe `trustTheCompiler` is a better name, or `trustCompilerAndImplementedBy`...

PR #2667, opened 2023-10-11
Add left/right actions to term tree coercion and make `^` a right action

### LINKED ISSUE #2220: elaboration of `^` shouldn't try to put the arguments in the same type
### Description

The macro for the `^` notation for exponentiation is defined as
```lean
macro_rules | `($x ^ $y)   => `(binop% HPow.hPow $x $y)
```
However, the `binop%` elaborator tries to make the type of `x` and `y` the same, which is not desired for exponentiation.

### MWE

```lean
class NatCast (R : Type u) where
  natCast : Nat → R

instance [NatCast R] : CoeTail Nat R where coe := NatCast.natCast
instance [NatCast R] : CoeHTCT Nat R where coe := NatCast.natCast

axiom Real : Type
notation "ℝ" => Real

instance (n : Nat) : OfNat ℝ n := sorry

variable (a : ℝ)

instance natCast : NatCast ℝ where natCast n := sorry
instance : HPow ℝ Nat ℝ := sorry

set_option pp.all true


#check a ^ 4
-- @HPow.hPow.{0, 0, 0} Real Nat Real instHPowRealNat a (@OfNat.ofNat.{0} Nat 4 (instOfNatNat 4)) : Real

instance : HPow ℝ ℝ ℝ := sorry

#check a ^ 4
-- @HPow.hPow.{0, 0, 0} Real Real Real instHPowReal a (@OfNat.ofNat.{0} Real 4 (instOfNatReal 4)) : Real

#check a ^ (4 : Nat)
-- @HPow.hPow.{0, 0, 0} Real Real Real instHPowReal a (@NatCast.natCast.{0} Real natCast (@OfNat.ofNat.{0} Nat 4 (instOfNatNat 4))) : Real
```

The desired result is that all checks use the `instHPowRealNat` instance, i.e. that `4` is interpreted as a `Nat`.

### Additional Notes

Just removing `binop%` and replacing the macro with the following fixes the current issue, but it causes problems in other elaboration problems.
```lean
macro_rules | `($x ^ $y)   => `(HPow.hPow $x $y)
```

With this macro an example like the following fails.
```lean
example (n : Nat) := ∃ k, n = 2 ^ k -- fails to find any instance, needs to know the type of `k`
```

### Zulip Discussion

[link](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/!4.234085)

Kha:
Fixes #2220. This is a slightly refactored core integration of @kmill's https://gist.github.com/kmill/bc58941368cc60738f3c932fe2546c9d.

We/someone might first want to update https://github.com/leanprover-community/mathlib4/pull/6852 before merging this.

### DIFF
--- a/src/Lean/Elab/BuiltinNotation.lean
+++ b/src/Lean/Elab/BuiltinNotation.lean
@@ -252,8 +252,18 @@ def elabCDotFunctionAlias? (stx : Term) : TermElabM (Option Expr) := do
       try Term.resolveId? f catch _ => return none
     else
       return none
-  | `(fun $binders* => binop% $f $a $b) =>
+  | `(fun $binders* => binop% $f $a $b)
+  | `(fun $binders* => binop_lazy% $f $a $b)
+  | `(fun $binders* => leftact% $f $a $b)
+  | `(fun $binders* => rightact% $f $a $b)
+  | `(fun $binders* => binrel% $f $a $b)
+  | `(fun $binders* => binrel_no_prop% $f $a $b) =>
     if binders == #[a, b] then
+      try Term.resolveId? f catch _ => return none
+    else
+      return none
+  | `(fun $binders* => unop% $f $a) =>
+    if binders == #[a] then
       try Term.resolveId? f catch _ => return none
     else
       return none
--- a/tests/lean/cdotAtSimpArg.lean
+++ b/tests/lean/cdotAtSimpArg.lean
@@ -1,5 +1,9 @@
+/-! Test `·` being able to refer to constants in `simp` -/
+
 example : ¬ true = false := by
   simp [(¬ ·)]
+
+/-! Test `binop%` -/
 
 example (h : y = 0) : x + y = x := by
   simp [(.+.)] -- Expands `HAdd.hAdd
@@ -14,3 +18,8 @@ example (h : y = 0) : x + y = x := by
   simp [Add.add]
   simp [h, Nat.add]
   done
+
+/-! Test `binop%` variant as well -/
+
+example (x y : Nat) : x ^ y = y ^ x := by
+  simp only [.^.]
--- a/tests/lean/cdotAtSimpArg.lean.expected.out
+++ b/tests/lean/cdotAtSimpArg.lean.expected.out
@@ -4,3 +4,6 @@ h : y = 0
 y x : Nat
 h : y = 0
 ⊢ Add.add x y = x
+cdotAtSimpArg.lean:24:39-25:17: error: unsolved goals
+x y : Nat
+⊢ Pow.pow x y = Pow.pow y x


Kha:
To my shame I did not get `parseQuotWithCurrentStage` to work

src/Lean/Parser/Term.lean
```diff
@@ -555,11 +555,18 @@ def matchAltsWhereDecls := leading_parser
 /-- Similar to `binrel`, but coerce `Prop` arguments into `Bool`. -/
 @[builtin_term_parser] def binrel_no_prop := leading_parser
   "binrel_no_prop% " >> ident >> ppSpace >> termParser maxPrec >> ppSpace >> termParser maxPrec
-@[builtin_term_parser] def binop  := leading_parser
+@[builtin_term_parser] def binop := leading_parser
   "binop% " >> ident >> ppSpace >> termParser maxPrec >> ppSpace >> termParser maxPrec
-@[builtin_term_parser] def binop_lazy  := leading_parser
+-- like `binop%` but wrap arguments in `fun () => _`
```
eric-wieser, line 560: I think this should be a docstring, so that it appears when you hover over `binop_lazy%` in the source.

src/Init/Data/Nat/Power2.lean
```diff
@@ -24,7 +24,7 @@ where
 termination_by go p h => n - p
 decreasing_by simp_wf; apply nextPowerOfTwo_dec <;> assumption
 
-def isPowerOfTwo (n : Nat) := ∃ k, n = 2 ^ k
+def isPowerOfTwo (n : Nat) := ∃ k : Nat, n = 2 ^ k
```
leodemoura, line 27: :(

src/Lean/Parser/Term.lean
Kha: It's not a good docstring though :) . We should add self-contained docstrings to all of these but that's a task for another PR

src/Init/Data/Nat/Power2.lean
Kha: I hope this regression is still acceptable because preventing this information flow is precisely the issue at hand, it's not a side effect we could fix

kmill: I believe there's a default instance in mathlib for nat powers that makes this work without modification there. I'll have to check. I'm not sure this instance can be brought to core because there are lots of defeq issues here that have to be solved in the algebra hierarchy.

src/Lean/Elab/Extra.lean
```diff
@@ -79,13 +79,13 @@ private def throwForInFailure (forInInstance : Expr) : TermElabM Expr :=
 namespace Op
 /-!
 
-The elaborator for `binop%`, `binop_lazy%`, and `unop%` terms.
+The elaborator for `binop%`, `binop_lazy%`, `leftact%`, `rightact%`, and `unop%` terms.
```
leodemoura, line 83: Could you please add a summary describing each one of the operators?
We need to raise the bar for files being touched by multiple people.

leodemoura: Recall we originally only had `binop%`, and the lists keeps getting bigger :)

src/Init/Data/Nat/Power2.lean
leodemoura: @kmill Good to know. I am hopping you will fix the defeq issues in core when you start at the end of the month :)

src/Lean/Elab/Extra.lean
kmill: I'm just on my phone and I haven't checked what got included, but I had previously worked out what would be needed to PR this to core and included additional docstings in a few places: https://github.com/leanprover/lean4/compare/master...kmill:kmill_binop2?expand=1

leodemoura: @kmill Let's get them in as soon as you start next week.

semorrison:
@kmill is it okay if I leave fixing the Mathlib build to you here?

src/Lean/Parser/Term.lean
kmill: I've added docstrings in #2778

src/Lean/Elab/Extra.lean
kmill: I expanded the docstring and added summaries (Sebastian and I decided that I'd create a new PR since I can't push to this branch, so it's over at #2778)

kmill:
I have a new version of this PR at #2778. I'll fix the mathlib build over there.

PR #2675, opened 2023-10-12
fix: level is_geq is missing a case

digama0:
For the standard model of levels as natural numbers, it is true that `u <= imax u v`:
```lean
import Std.Data.Nat.Init.Lemmas

def imax : Nat → Nat → Nat
  | _a, 0 => 0  
  | a, b => max a b

example {u v w} (h : v ≥ w) : imax u v ≥ w := by
  apply Nat.le_trans h
  cases v with
  | zero => apply Nat.le_refl
  | succ v => apply Nat.le_max_right
```

As such, the current condition for `imax u v ≥ w` is overly pessimistic, since it lacks the `u ≥ w` case. Here's an example which is affected by this case:
```lean
inductive Foo.{u, v} : Sort ((imax u v)+1)
  | x : Sort u → Foo
```
```
invalid universe level in constructor 'Lean.Level.Foo.x', parameter has type
  Sort u
at universe level
  u+1
it must be smaller than or equal to the inductive datatype universe level
  (imax u v)+1
```

### DIFF
--- a/src/kernel/level.cpp
+++ b/src/kernel/level.cpp
@@ -505,7 +505,7 @@ bool is_geq_core(level l1, level l2) {
     if (is_imax(l2))
         return is_geq(l1, imax_lhs(l2)) && is_geq(l1, imax_rhs(l2));
     if (is_imax(l1))
-        return is_geq(imax_rhs(l1), l2);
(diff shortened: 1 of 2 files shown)


digama0:
Oops, the proof above shows `v <= imax u v`, not `u <= imax u v`, and the other one is not true. Sorry for the noise.

PR #2676, opened 2023-10-12
fix: bug in do expansion

### LINKED ISSUE #2663: name resolution bug in do elaboration
```lean
example : Id Unit := do
  let mut x ← if true then pure true else pure false
  if let .true := x then
    x := false
 -- ^
-- unknown identifier 'x'
-- unknown identifier 'x'
-- unknown identifier 'x'
```
The `if` doElem on the first line seems to be a factor - enclosing it in parentheses fixes the issue. I'm not entirely sure why the error message is repeated three times; my guess is that the desugared expression contains three occurrences of the variable but the variable declaration is missing.

digama0:
As [reported on zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/unknown.20identifier.20in.20.60let.20var.20.3A.3D.20if.20let.60/near/396272940). The second commit fixes #2663, which looks similar but is a separate issue. I also fixed some typos and formatting issues in the debugging code in this file.

This change now causes an unreachable warning on the following code:
```lean
example := do
  let x <- return
```
which is kind of true? But the error message points at the following doElem and says that is unreachable, which doesn't make much sense here since it points at the whole statement including the `return`. So I added a mechanism to suppress this error. (PS: this should really be a warning, not an error. The code is perfectly well interpretable even if parts of it are not reachable, and this can make editing annoying.)

### DIFF
--- a/tests/lean/run/doNotation6.lean
+++ b/tests/lean/run/doNotation6.lean
@@ -12,7 +12,7 @@ if (← get) == 0 then
 modify (· - x)
 
 def f1 (x : Nat) : M Nat := do
-let v ←
+let v : Bool ←
   try
     dec x
     return x
--- a/src/Lean/Elab/Do.lean
+++ b/src/Lean/Elab/Do.lean
@@ -576,6 +576,8 @@ def mkMatch (ref : Syntax) (genParam : Syntax) (discrs : Syntax) (optMotive : Sy
 def concat (terminal : CodeBlock) (kRef : Syntax) (y? : Option Var) (k : CodeBlock) : TermElabM CodeBlock := do
   unless hasTerminalAction terminal.code do
     throwErrorAt kRef "`do` element is unreachable"
+  -- remove updates to variable `y` in `k` because `y` is being defined (see #2663)
+  let k := if let some y := y? then { k with uvars := eraseVars k.uvars #[y] } else k
   let (terminal, k) ← homogenize terminal k
   let xs := varSetToArray k.uvars
   let y ← match y? with | some y => pure y | none => `(y)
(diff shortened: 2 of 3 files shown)


tests/lean/run/doNotation6.lean
```diff
@@ -12,7 +12,7 @@ if (← get) == 0 then
 modify (· - x)
 
 def f1 (x : Nat) : M Nat := do
-let v ←
+let v : Bool ←
```
leodemoura, line 15: Is it clear to you why this annotation is necessary now?

digama0: Actually I'm not sure anymore. The straightforward answer is that as the type annotation `Bool`, unrelated to anything in the expression, should indicate, the type of this variable is unconstrained, because the doElem before it doesn't return a value to the expression, so erroring with that would seem to make sense. On the other hand, the previous behavior was to not even elaborate the tail part of the expression (meaning for example that `v` would not be semantically highlighted as a variable) because that subterm ends up floating and not used in the final term output, in which case I don't think you would get the unbound type variable error, and testing it locally I am getting this behavior -- the type annotation is necessary, `v` is not highlighted and there are no errors.

I added this annotation in response to an error in CI: https://github.com/leanprover/lean4/actions/runs/6498883602/job/17651115144#step:11:2876 , but I'm personally fine with either behavior, it's a somewhat unimportant edge-case. LMK what behavior you want to see and I can try to implement it. For now I will remove the annotation and see if the error in CI persists.

sgraf812:
The new do elaborator warns on `x` for the program in the OP and fixes #2663 in a different way. Closing.

PR #2679, opened 2023-10-12
feat: mark `rfl` proofs of `Iff` for use by `dsimp`

### LINKED ISSUE #2678: `dsimp` regression: `Iff.rfl` proofs are not used by `dsimp`
### Prerequisites

### Description

Users of Lean 3 and mathlib3 are used to seeing `dsimp` make use of proofs of `Iff` consisting of `Iff.rfl`. For example, in Lean 3,
```
def w : Prop → Prop := fun p, p

@[simp] theorem iff_thm (a : Prop) : w a ↔ a := iff.rfl

example (a : Prop) : w a := by dsimp -- rewrites `w a` to `a`
```
Whereas in Lean 4,
```
def w : Prop → Prop | p => p

@[simp] theorem iff_thm : w a ↔ a := Iff.rfl

example : w a := by dsimp -- dsimp made no progress
```
Ideally, `dsimp` should be able to rewrite the goal using `iff_thm` in Lean 4 as well.

### Context

This issue first arose in March 2023 [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/dsimp.20and.20Iff.2Erfl), where the notion of being able to use `Iff.rfl` proofs with `dsimp` in Lean 4 was met with support. No issue was filed at the time.

Presently (October 2023), a different case of `dsimp` not working as expected for the same reason arose [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Tactic.20state.20explosion). (Note: the linked thread also addresses other issues.)

### Steps to Reproduce

1. Create a `@[simp]` theorem `iff_thm` whose proof consists of `Iff.rfl`
2. Use `dsimp` on a goal which includes a match of the lhs or rhs of that lemma

**Expected behavior:** `dsimp` rewrites the goal via `iff_thm`

**Actual behavior:** `dsimp` does not use `iff_thm`

### Versions

Lean (version 4.0.0-nightly-2023-08-18, commit 9d05b5f081c2, Release)
macOS Ventura 13.4

### Impact

thorimur:
This PR modifies the behavior of `dsimp` to allow it to use `rfl` proofs of `Iff` statements. It does so by marking `Iff` theorems that have proofs which consist of `Iff.rfl`, `Iff.refl`, `Iff.symm`, `propext`, or any combination thereof as `rfl` theorems via `isRflProof`.

Fixes #2678.

Note: this PR is currently experimental, and CI is being used for debugging; this branch will likely be rebased for a cleaner commit history.

⚠️ Various issues arise from adding lemmas to the `dsimp` set. See [zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/using.20.60Iff.60.20rfl.20proofs.20in.20.60dsimp.60) for more details.

### DIFF
--- a/src/Lean/Elab/Tactic/Simp.lean
+++ b/src/Lean/Elab/Tactic/Simp.lean
@@ -116,19 +116,25 @@ private def addDeclToUnfoldOrTheorem (thms : Meta.SimpTheorems) (id : Origin) (e
     thms.add id #[] e (post := post) (inv := inv)
 
 private def addSimpTheorem (thms : Meta.SimpTheorems) (id : Origin) (stx : Syntax) (post : Bool) (inv : Bool) : TermElabM Meta.SimpTheorems := do
-  let (levelParams, proof, type) ← Term.withoutModifyingElabMetaStateWithInfo <| withRef stx <| Term.withoutErrToSorry do
+  let (levelParams, proof) ← Term.withoutModifyingElabMetaStateWithInfo <| withRef stx <| Term.withoutErrToSorry do
     let e ← Term.elabTerm stx none
     Term.synthesizeSyntheticMVars (mayPostpone := false) (ignoreStuckTC := true)
     let type ← inferType e
     let e ← instantiateMVars e
     let e := e.eta
-    trace[debug] "addSimpTheorem: type: {type}\nexpr: {e}\ne.hasMVar == {e.hasMVar}\ntype.hasMVar == {type.hasMVar}"
+    trace[debug] "addSimpTheorem: type: {type}\nexpr: {e}\ne.hasMVar == {e.hasMVar}\ntype.hasMVar == {(← instantiateMVars type).hasMVar}"
     if e.hasMVar then
       let r ← abstractMVars e
-      return (r.paramNames, r.expr, none)
+      return (r.paramNames, r.expr)
     else
-      return (#[], e, some (← instantiateMVars type))
-  thms.add id levelParams proof (post := post) (inv := inv) (type := type)
+      let type ← instantiateMVars type
+      --!! might add a type hint too often?
+      trace[debug] "addSimpTheorem₂: type.hasMVar == {type.hasMVar}\ntype == ← inferType e: {type == (← inferType e)}"
+      if ← (pure !type.hasMVar) <&&> return type != (← inferType e) then
+        return (#[], ← mkExpectedTypeHint e type)
+      else
+        return (#[], e)
+  thms.add id levelParams proof (post := post) (inv := inv)
 
 structure ElabSimpArgsResult where
   ctx     : Simp.Context


thorimur:
WIP

semorrison:
@thorimur, could you run `!bench` on the Mathlib branch? You will need to make a PR out of the branch lean-pr-testing-2679.

semorrison:
!bench

thorimur:
> @thorimur, could you run `!bench` on the Mathlib branch? You will need to make a PR out of the branch lean-pr-testing-2679.

Will do—as soon as I make sure this is working! :) Turns out I needed to account for `simp`'s preprocessing adding `propext` to these lemmas. (I also plan to add some tests.) Depending on how long it takes to compile and produce a toolchain, I might not have a chance to test and bench until tomorrow.

thorimur:
!bench

eric-wieser:
> Will do—as soon as I make sure this is working! :) Turns out I needed to account for `simp`'s preprocessing adding `propext` to these lemmas.

What do you mean by this? IIRC, the strategy in lean 3 was just to recreate a copy of the lemma with `=` instead of `Iff`, to avoid needing `propext`.

thorimur:
> > Will do—as soon as I make sure this is working! :) Turns out I needed to account for `simp`'s preprocessing adding `propext` to these lemmas.
> 
> What do you mean by this? IIRC, the strategy I implemented in lean 3 was just to recreate a copy of the lemma with `=` instead of `Iff`, to avoid needing `propext`. The motivation is not to avoid the axiom per se, but to ensure everything goes through definitionally, which avoids problems with dependent rewrites.

@eric-wieser The reasons lie in the code `mkSimpTheoremsFromConst` in `Lean.Meta.Tactic.Simp.SimpTheorems`. This calls `preprocess` on the type and value, which in the case of `Iff`, attaches `propext`. In `preprocess` in that same file:
```lean
  ...
  else if let some (lhs, rhs) := type.iff? then
      ...
      let type ← mkEq lhs rhs
      let e    ← mkPropExt e
      return [(e, type)]
```
It is this `e` and `type` which is used to create an `auxLemma`, which is in turn fed to `mkSimpTheoremsCore`—which in turn calls `isRflProof` to set the `dsimp` flag. At that stage, we therefore need to unwrap the `propext` to check if it's `dsimp`-worthy. (I was reluctant to change the flow and wanted to keep my changes minimal, but one could imagine e.g. including the output of `isRflProof` in the `preprocess` result.)

I want to be clear, though, that `isRflProof` *already* unwraps this `auxLemma` anyway, and unwraps constants it encounters recursively (as long as they're applied to something, for some reason). This isn't something I introduced; I just added an `if` branch to handle the `propext` case.

What sorts of problems with dependent rewrites might we worry about here?

semorrison:
@thorimur, you might want to rebase this onto `master`; I think we won't be able to get the Mathlib testing to work otherwise, as other things have moved on in the meantime! :-)

nomeata:
What is the status of this PR, @thorimur?

PR #2681, opened 2023-10-13
doc: explain namespace indentation

Kha:


markusschmaus:
I think it would be better to just remove the indentation. If indentation is not desired in actual code, why would should it be included in these examples. It only adds mental workload for anybody trying to learn Lean.

Kha:
> I think it would be better to just remove the indentation. If indentation is not desired in actual code, why would should it be included in these examples. It only adds mental workload for anybody trying to learn Lean.

@david-christiansen Happy to leave that decision to you :)

doc/namespaces.md
```diff
@@ -106,3 +106,6 @@ Sections are also useful for delimiting the scope of commands such as ``set_opti
 In many respects, however, a ``namespace ... end`` block behaves the same as a ``section ... end`` block.
 In particular, if you use the ``variable`` command within a namespace, its scope is limited to the namespace.
 Similarly, if you use an ``open`` command within a namespace, its effects disappear when the namespace is closed.
+
+In all the examples above, we have indented the namespace contents for demonstration purposes.
+In practice, we elide indentation of sections and namespaces in favor of the document structure hints the editor extensions provide.
```
david-christiansen, line 111: I think that on the one hand, it's good to explain the common practice (after all, I would initially want to indent them), but on the other, example code should be idiomatic.

What about removing the indentation from the examples, but then using a text like this?

```suggestion
In idiomatic Lean code, namespaces and sections are typically not indented. This is both because they are very common and because we can rely on the hints about document structure provided in our text editors.
```

Kha: @david-christiansen Just to make sure, are you in favor of removing the indentation? Your comment reads like there may be a "not" missing but I'm not sure.

Kha:
Note, `section.md` should be adjusted correspondingly. Perhaps we should merge them even?

doc/namespaces.md
markusschmaus: Sorry for the late reply. Yes, I am in favor for removing indentation. Don't indent namespaces here, because they aren't indented anywhere else.

PR #2686, opened 2023-10-15
fix: allow constants to be marked for `dsimp`

### LINKED ISSUE #2685: `dsimp` doesn't use `rfl` theorems which consist of an unapplied constant
### Prerequisites

### Description

if a `@[simp]` theorem `foo` has a proof consisting of a constant which is a `rfl` theorem, then whether or not it gets used by `dsimp` is determined by whether or not it has any arguments. Constants without arguments fail to be used.
```
/-! Not used: -/
theorem const_thm_internal_no_args : (or true false) = true := rfl
@[simp] theorem foo_doesnt_work : (or true false) = true := const_thm_internal_no_args

example : (or true false) = true := by dsimp -- dsimp made no progress

/-! Used: -/
theorem const_thm_internal_with_arg (_ : Nat) : (or true false) = true := rfl
@[simp] theorem foo_works : (or true false) = true := const_thm_internal_with_arg 0

example : (or true false) = true := by dsimp -- success
```

### Context

The use of unapplied constants is artificially prevented by the check `proof.isApp` in `isRflProofCore` in `Lean.Meta.Tactic.Simp.SimpTheroems`.

Note that whether a theorem is used by `dsimp` is determined recursively; this example only unwraps once.

This is an edge case, so fixing it is more preventative than reactive.

### Steps to Reproduce

1. Create a constant with no arguments which proves an equality via `rfl`.
2. Use that constant as the proof of a theorem `foo` marked `@[simp]`.

**Expected behavior:** `foo` can be used by `dsimp`.

**Actual behavior:** `foo` cannot be used by `dsimp`

### Impact

thorimur:

Fixes #2685.

-----

(Note that this only broke mathlib briefly because 66ab016 (which landed not more than an hour before this PR) had not been incorporated into std yet (leanprover/std4#284); the subsequent two breakages were for the same reason. I'm not actually sure how it managed to build once before the present build; I merely added a test after the successful first build.)

### DIFF
--- a/tests/lean/dsimpViaConst.lean
+++ b/tests/lean/dsimpViaConst.lean
@@ -1,0 +1,52 @@
+/-! # `dsimp` uses theorems whose proofs are unapplied constants
+
+This test ensures that `dsimp` can take advantage of `@[simp]` theorems whose proofs are unapplied
+constants which are marked as `rfl` theorems.
+
+## Context
+
+To determine if a `@[simp]` theorem is a `rfl` theorem (and therefore usable by `dsimp`), we check
+if its proof is one of the following:
+
+1. directly a `rfl` proof, e.g. `Eq.rfl`
+2. a `symm` proof of `rfl` proofs
+3. an application of a constant which is considered a `rfl` theorem
+
+This means that a theorem `foo` whose proofs is of the form `foo_internal_with_arg 0` where
+`foo_internal_with_arg` is a `rfl` theorem. The first section
+demonstrates that behavior as a control case.
+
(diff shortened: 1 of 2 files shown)


thorimur:
WIP

semorrison:
> (Note: currently seems to only break mathlib because [66ab016](https://github.com/leanprover/lean4/commit/66ab016723adf233f0c511e2e82cd5884d740614) (which landed not more than an hour before this PR) has not been incorporated into std yet.)

If you rebase your branch on `master` hopefully CI will sort this out now!

thorimur:
awaiting-review

PR #2688, opened 2023-10-15
chore: remove unnecessary `%` operations at `Fin.mod` and `Fin.div`

leodemoura:
We now have the missing proofs `Nat.mod_le` and `Nat.div_le_self` in core.
See: https://github.com/leanprover/std4/pull/286#discussion_r1359807875

### DIFF
--- a/src/Init/Data/Fin/Basic.lean
+++ b/src/Init/Data/Fin/Basic.lean
@@ -45,7 +45,7 @@ protected def sub : Fin n → Fin n → Fin n
   | ⟨a, h⟩, ⟨b, _⟩ => ⟨(a + (n - b)) % n, mlt h⟩
 
 /-!
-Remark: mod/div/modn/land/lor can be defined without using (% n), but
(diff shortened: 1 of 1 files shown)


digama0:
This will have a fair amount of fallout in mathlib.

src/Init/Data/Fin/Basic.lean
```diff
@@ -51,10 +51,10 @@ needed to bootstrap Lean.
 -/
```
eric-wieser, line 51: This comment block is outdated now, and can be deleted

leodemoura:
> This will have a fair amount of fallout in mathlib.

There is no rush to merge this PR. It is here just to address a suggestion at https://github.com/leanprover/std4/pull/286

digama0:
(I am in favor of the change, but it seems low priority and will need some mathlib cleanup. Maybe when Scott has some free time... :wink: )

semorrison:
Strangely, it required only trivial cleanup in Std, and then nothing in Mathlib!

PR #2689, opened 2023-10-15
fix: add a missing case to `Level.geq`

digama0:
This PR adds a case to `Level.geq` that is present in the kernel's level `is_geq` procedure, making them consistent with one another.

This came up during testing of `lean4lean`. Currently `Level.geq` differs from `level::is_geq` in the case of `max u v >= imax u v`. The elaborator function is overly pessimistic and yields `false` on this while the kernel function yields true. This comes up concretely in the `Trans` class:
```lean
class Trans (r : α → β → Sort u) (s : β → γ → Sort v) (t : outParam (α → γ → Sort w)) where
  trans : r a b → s b c → t a c
```
The type of this class is `Sort (max (max (max (max (max (max 1 u) u_1) u_2) u_3) v) w)` (where `u_1 u_2 u_3` are the levels of `α β γ`), but if you try writing that type explicitly then the `class` command fails. Omitting the type leaves the `class` to infer the universe level (the command assumes the level is correct, and the kernel agrees it is), but including the type then the elaborator checks the level inequality with `Level.geq` and fails.

### DIFF
--- a/tests/lean/run/2689.lean
+++ b/tests/lean/run/2689.lean
@@ -1,13 +1,36 @@
+/-!
+# Tests of universe constraints for inductive types
+
+https://github.com/leanprover/lean4/pull/2689 corrected the elaborator's `Level.geq`,
+which was missing cases that the kernel `is_geq` could handle.
+-/
+
+/-!
+This always worked. The universe level of `Trans₁` is inferred from the fields.
+-/
 structure Trans₁ {α : Sort a} {β : Sort b} {γ : Sort c}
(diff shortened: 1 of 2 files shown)


semorrison:
Just noting there's a conflict here now.

Would it be possible to include the `Trans` example as a test?

digama0:
Ping @leodemoura on this since apparently #2750 was missed so probably this long standing bug was too.

leodemoura (approved):

PR #2694, opened 2023-10-16
chore: use more macOS versions in CI

david-christiansen:
This will help catch failures due to linker and library location changes on OS updates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,16 +132,52 @@ jobs:
             CMAKE_OPTIONS: -DLEAN_EXTRA_CXX_FLAGS=-fsanitize=address,undefined -DLEANC_EXTRA_FLAGS='-fsanitize=address,undefined -fsanitize-link-c++-runtime' -DSMALL_ALLOCATOR=OFF -DBSYMBOLIC=OFF
             # exclude seriously slow/problematic tests (laketests crash)
             CTEST_OPTIONS: -E 'interactivetest|leanpkgtest|laketest|benchtest'
-          - name: macOS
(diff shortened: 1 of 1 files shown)


david-christiansen:
This is interesting - according to our prior builds, `macos-latest` was identifying itself as:
```
Current runner version: '2.309.0'
Operating System
  macOS
  12.7
  21G816
Runner Image
  Image: macos-12
  Version: 20230921.4
  Included Software: https://github.com/actions/runner-images/blob/macOS-12/20230921.4/images/macos/macos-12-Readme.md
  Image Release: https://github.com/actions/runner-images/releases/tag/macOS-12%2F20230921.4
```
and
```
Current runner version: '2.309.0'
Operating System
  macOS
  12.7
  21G816
Runner Image
  Image: macos-12
  Version: 20230921.4
  Included Software: https://github.com/actions/runner-images/blob/macOS-12/20230921.4/images/macos/macos-12-Readme.md
  Image Release: https://github.com/actions/runner-images/releases/tag/macOS-12%2F20230921.4
```

I would have thought that copy-pasting their setups and explicitly referencing Mac OS versions would have at least resulted in a successful build on `macos-12`.

I'm out of my depth - @Kha, your feedback is welcome when you return!

Kha:
> gcp: command not found

The `brew` step is missing.

However, if this is just about validating our *development* setup, I'm hesitant on spending CI resources on it.

david-christiansen:
The urgency of it was based on faulty assumptions about Mac deployment targets. I'll close, and then we can follow up about e.g. releases. Thanks!

PR #2695, opened 2023-10-16
fix: universe inference for `Prop`-valued structures/inductives

### LINKED ISSUE #2690: `structure` sometimes makes a `Type` not a `Prop`.
### Prerequisites

### Description

Structures which should be `Prop`s (because all their fields are Props) sometimes come out as `Type`s and Lean needs to be explicitly told that they are `Prop`s. 

### Steps to Reproduce

```
structure foo (α : Type) where
  h : ∀ a : α, a = a

#check foo Nat -- Type?!

structure foo' (α : Type) : Prop where
  h : ∀ a : α, a = a

#check foo' Nat -- Prop, as expected
```

**Expected behavior:** 

`foo α : Prop`

**Actual behavior:** 

`foo α : Type`

### Versions

This bug has existed since Lean 3 and the community has been using the obvious workaround for years ("when you randomly notice an instance of this, typically because [a new user is confused on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Object.20not.20being.20a.20structure/near/396760553), just tell the structure that it's a Prop explicitly")

[Output of `lean --version` in the folder that the issue occured in]

Lean (version 4.2.0-rc1, commit a62d2fd49796, Release)

[OS version]

Ubuntu 22.04

### Impact

Very minor, occasionally confusing, little more than annoying once you know about it.

arthur-adjedj:
Closes #2690

### DIFF
--- a/src/Init/Core.lean
+++ b/src/Init/Core.lean
@@ -122,7 +122,7 @@ If and only if, or logical bi-implication. `a ↔ b` means that `a` implies `b` 
 By `propext`, this implies that `a` and `b` are equal and hence any expression involving `a`
 is equivalent to the corresponding expression with `b` instead.
 -/
-structure Iff (a b : Prop) : Prop where
+structure Iff (a b : Prop) where
   /-- If `a → b` and `b → a` then `a` and `b` are equivalent. -/
   intro ::
   /-- Modus ponens for if and only if. If `a ↔ b` and `a`, then `b`. -/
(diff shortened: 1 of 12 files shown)


src/Init/Core.lean
```diff
@@ -1684,15 +1684,15 @@ axiom ofReduceNat (a b : Nat) (h : reduceNat a = b) : a = b
 `IsAssociative op` says that `op` is an associative operation,
 i.e. `(a ∘ b) ∘ c = a ∘ (b ∘ c)`. It is used by the `ac_rfl` tactic.
 -/
-class IsAssociative {α : Sort u} (op : α → α → α) where
+class IsAssociative {α : Sort u} (op : α → α → α) : Type where
```
digama0, line 1687: IMO these should be Props, this is exactly the kind of issue we are hoping to avoid by changing the default

src/Init/Prelude.lean
```diff
@@ -807,7 +807,7 @@ decidability instance instead of the proposition, which has no code).
 If a proposition `p` is `Decidable`, then `(by decide : p)` will prove it by
 evaluating the decidability instance to `isTrue h` and returning `h`.
 -/
-class inductive Decidable (p : Prop) where
+class inductive Decidable (p : Prop) : Type where
```
digama0, line 810: This is a bug, it should not infer `Prop` here because the inductive has two constructors.

arthur-adjedj: Yes, the current draft makes a hacky tweak to `accLevel`, which I believe isn't the solution (as explained in the issue). This is not meant for review ! When tweaking `mkResultUniverse`, I end up getting some kernel errors regarding loose universe variables in sort-polymorphic inductives like `PProd`, so there's still work to do.

PR #2699, opened 2023-10-16
feat: remove configuration olean on `lake clean`

### LINKED ISSUE #2592: `lake clean` should remove `lakefile.olean` in the root directory and under `lake-packages`
(And possibly its location outside of `build` should be reconsidered.)

tydeu:
Closes #2592.


tydeu:
After some debugging (including with `gdb`), it appears that the root olean (and only the root olean) handle stays open for some reason on Windows. This causes the remove to fail. I am not sure why it is only the root olean that exhibits. I would expect either all or none, but that does not happen.

PR #2700, opened 2023-10-16
fix: bug in elab TC for projections

### LINKED ISSUE #2697: (kernel) invalid projection when extracting the proposition from an `Exists` statement.
### Prerequisites

### Description

The following code explains the issue:

```lean
set_option pp.rawOnError true
/- Error:
(kernel) invalid projection
  h.2
-/
example (h : ∃ _ : Nat, P) : P := h.2

/- Works -/
example (h : ∃ _ : Nat, P) : P := by
  let ⟨_, P⟩ := h
  exact P
```

### Context

The issue is discussed [here](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Kernel.20Invalid.20Projection). It seems that the error *makes sense*, but Alex suggested me to open an issue since it's a `kernel` error.

### Versions

The error appears since `Lean (version 4.0.0, commit ec941735c80d, Release)` up til `v4.2.0-rc1`, on a Mac Intel.

### Others

As mentioned above, the error makes sense from a human perspective, so if this `kernel` error is intended, feel free to close this issue :)

digama0:
This adapts the code from the `inferProjType` function and the kernel projection typechecker so that the elaborator correctly reports errors when the projection would be rejected by the kernel. Fixes #2697.

### DIFF
--- a/src/Lean/Elab/App.lean
+++ b/src/Lean/Elab/App.lean
@@ -50,9 +50,26 @@ private def mkProjAndCheck (structName : Name) (idx : Nat) (e : Expr) : MetaM Ex
   let r := mkProj structName idx e
   let eType ← inferType e
   if (← isProp eType) then
-    let rType ← inferType r
-    if !(← isProp rType) then
-      throwError "invalid projection, the expression{indentExpr e}\nis a proposition and has type{indentExpr eType}\nbut the projected value is not, it has type{indentExpr rType}"
+    let eType ← whnf eType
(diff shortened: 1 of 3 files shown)


digama0:
This PR fixes part of the issue (it prevents
```lean
inductive Foo : Prop where
  | mk (x : Nat) (h : x = x) : Foo

example (h : Foo) := h.2
```
from working), but it does not fix the example from #2697, and I think a proper fix would require more than just this, so this PR is on hold until I get approval for making some changes that will affect the kernel.

Issue #2632, opened 2023-10-06
Lake segmentation fault


yangky11:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to mathlib4 or std4.

### Description

I get segmentation faults when running any lake commands, including `lake build` and `lake clean`. The problem only happens when all of the following three conditions are met:

1. The package depends on [LeanInfer](https://github.com/lean-dojo/LeanInfer) or is LeanInfer itself.
2. The package uses `leanprover/lean4:v4.2.0-rc1` (No problem with `leanprover/lean4:v4.0.0`).
3. The package has been built or partially built (No problem with the first run of `lake build`).

### Steps to Reproduce

1. Check out the `LeanInfer-demo` branch of [lean4-example](https://github.com/yangky11/lean4-example/tree/LeanInfer-demo).
2. Edit lean4-example's lean-toolchain file to use `leanprover/lean4:v4.2.0-rc1`.
3. `lake build`
4. Either wait `lake build` to finish or just kill it in the middle. Then run `lake build` for a second time and you'll see `Segmentation fault (core dumped)`. In this state, running any lake command will run into segmentation faults, unless you remove the lakefile.olean of LeanInfer.

### Versions

`Lean (version 4.2.0-rc1, commit a62d2fd49796, Release)`

```
NAME="Ubuntu"
VERSION="20.04.6 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.6 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal
```

tydeu:
[...]

Issue #2631, opened 2023-10-06
compiler slowdown optimizing Option/List uses


JamesGallicchio:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to mathlib4 or std4.

### Description

The (old) compiler spends exponentially long optimizing the MWE below.

### Context

I am auto-generating definitions that encounter this compilation slowdown; I can probably restructure the generated definitions to sidestep optimization and prevent the slowdown, so it is not urgent.

[Zulip thread here](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/slow.20command.20elab) (no substantial discussion/investigation yet).

### Steps to Reproduce

MWE:
```
def oops
  (x0 : Option String)
  (x1 : Option String)
  (x2 : Option String)
  (x3 : Option String)
  (x4 : Option String)
  (x5 : Option String)
  (x6 : Option String)
  (x7 : Option String)
  (x8 : Option String)
  (x9 : Option String)
  (y0 : Option String)
  (y1 : Option String)
  (y2 : Option String)
  :  List String
  :=
  let x0 := Option.map toString x0
  let x1 := Option.map toString x1
  let x2 := Option.map toString x2
  let x3 := Option.map toString x3
  let x4 := Option.map toString x4
  let x5 := Option.map toString x5
  let x6 := Option.map toString x6
  let x7 := Option.map toString x7
  let x8 := Option.map toString x8
  let x9 := Option.map toString x9
  let y0 := Option.map toString y0
  let y1 := Option.map toString y1
  let y2 := Option.map toString y2
  List.filterMap  (@id (Option String))
    [ x0.map (fun s => "x0" ++ s)
    , x1.map (fun s => "x1" ++ s)
    , x2.map (fun s => "x2" ++ s)
    , x3.map (fun s => "x3" ++ s)
    , x4.map (fun s => "x4" ++ s)
    , x5.map (fun s => "x5" ++ s)
    , x6.map (fun s => "x6" ++ s)
    , x7.map (fun s => "x7" ++ s)
    , x8.map (fun s => "x8" ++ s)
    , x9.map (fun s => "x9" ++ s)
  -- uncomment these lines to get progressively much slower
  --, y0.map (fun s => "y0" ++ s)
[...]

PR #2707, opened 2023-10-17
feat: lake: `run_io` to execute IO at term elab time

tydeu:
Workaround for #2632 to avoid defining elaborators in a configuration file for the common case of running IO.


tydeu:
@semorrison As our primary community ambassador, do you think `run_io` is good name for this elaborator?

src/lake/Lake/DSL/Meta.lean
```diff
@@ -53,9 +56,42 @@ scoped syntax (name := metaIf)
...
+    if (← logUnassignedUsingErrorInfos (← getMVars v)) then throwAbortTerm
+    let v ← mkAppM ``toExprIO #[v]
+    liftM <| ← evalExpr (IO Expr) (mkApp (mkConst ``IO) (mkConst ``Expr)) v
+  | _ => Elab.throwUnsupportedSyntax
+
+/-! ## ToExpr Instances -/
+
+instance : ToExpr System.FilePath where
+  toExpr p := mkApp (mkConst ``System.FilePath.mk) (toExpr p.toString)
+  toTypeExpr := mkConst ``System.FilePath
```
semorrison, line 96: Given that both `Lake` and `Std` need this now, perhaps it would be better to place in lean itself, to avoid duplication?

semorrison:
I think #2631 is not the issue you intended to link to, or otherwise I don't understand the connection.

tydeu:
@semorrison 

> I think #2631 is not the issue you intended to link to, or otherwise I don't understand the connection.

Oops, yeah, typo: it was suppose to be #2632 (I updated the description).

src/lake/Lake/DSL/Meta.lean
tydeu: I agree. I have just become somewhat wary of updating core unless necessary because it generally Sebastian or Leo to review and the are rather busy at the moment. This is probably small enough it does not matter. Regardless, it should probably be a separate PR because I did was move it around Lake in this one (it was previously in the deleted `Lake.Util.EvalTerm`).

src/lake/Lake/DSL/Meta.lean
```diff
@@ -53,9 +56,42 @@ scoped syntax (name := metaIf)
...
+opaque evalExpr (α) (expectedType : Expr) (value : Expr) (safety := DefinitionSafety.safe) : MetaM α
+
+def toExprIO [ToExpr α] (x : IO α) : IO Expr :=
+  toExpr <$> x
+
+/--
+Executes a term of type `IO α` at elaboration-time
+and produces an expression corresponding to the result via `ToExpr α`.
+-/
+scoped syntax:lead (name := runIO) "run_io " term : term
```
digama0, line 76:
```suggestion
scoped syntax:lead (name := runIO) "run_io " doSeq : term
```
It is generally more convenient to have monad-taking elabs directly receive a `doSeq` so you don't need to do `run_io do` and `<-` works as intended.

PR #2709, opened 2023-10-18
feat: `LAKE_PKG_URL_MAP`

tydeu:
Initial prototype. Overrides are of the form: `LAKE_PKG_URL_MAP={"pkg-name": "git-url", ...}`.

### DIFF
--- a/src/lake/Lake/Config/Env.lean
+++ b/src/lake/Lake/Config/Env.lean
@@ -3,10 +3,11 @@ Copyright (c) 2022 Mac Malone. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mac Malone
 -/
+import Lake.Util.Name
 import Lake.Util.NativeLib
 import Lake.Config.InstallPath
 
-open System
+open System Lean
 
 /-! # Lake's Environment
 Definitions related to a Lake environment.
@@ -24,6 +25,8 @@ structure Env where
   lean : LeanInstall
   /-- The Elan installation (if any) of the environment. -/
   elan? : Option ElanInstall
+  /-- A name-to-URL mapping of URL overwrites for the named packages. -/
+  pkgUrlMap : NameMap String
   /-- The initial Elan toolchain of the environment (i.e., `ELAN_TOOLCHAIN`). -/
   initToolchain : String
   /-- The initial Lean library search path of the environment (i.e., `LEAN_PATH`). -/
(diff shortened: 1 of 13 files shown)


src/lake/Lake/Config/Env.lean
```diff
@@ -34,20 +37,26 @@ structure Env where
...
     initToolchain := (← IO.getEnv "ELAN_TOOLCHAIN").getD ""
     initLeanPath := ← getSearchPath "LEAN_PATH",
     initLeanSrcPath := ← getSearchPath "LEAN_SRC_PATH",
     initSharedLibPath := ← getSearchPath sharedLibPathEnvVar,
     initPath := ← getSearchPath "PATH"
   }
+where
+  computePkgUrlMap := do
+    let some urlMapStr ← IO.getEnv "LAKE_PKG_URL_MAP" | return {}
+    let .ok urlMap := Json.parse urlMapStr |>.bind fromJson? | return {}
```
semorrison, line 58: This should error, rather than silently ignoring the bad JSON?

src/lake/Lake/Load/Main.lean
```diff
@@ -131,13 +131,13 @@ def buildUpdatedManifest (ws : Workspace)
   match res with
   | (.ok root, deps) =>
     let ws : Workspace ← {ws with root}.finalize
+    let manifest : Manifest := {name? := ws.root.name, packagesDir? := ws.relPkgsDir}
+    let manifest := deps.foldl (·.addPackage ·.manifestEntry) manifest
+    manifest.saveToFile ws.manifestFile
     LakeT.run ⟨ws⟩ <| ws.packages.forM fun pkg => do
       if let some postUpdate := pkg.postUpdate? then
         logInfo s!"{pkg.name}: running post-update hook"
         postUpdate
```
semorrison, line 140: Is the intention here that the post-update hooks might want to rewrite the manifest?

tydeu: Precisely!

src/lake/Lake/Config/Env.lean
tydeu: This does mean that an improper environment will kill virtually any `lake` command which I was not sure was desirable. But it sounds like that is fine, so will do.

semorrison: I think given that JSON-via-environment-variable is pretty dodgy to begin with, failing fast is better than the user not noticing a stray comma and not understanding why they can't get their packages.

tydeu:
@semorrison Is the mathlib CI broken on Lean master?  All I did was rebase the PR onto master and it switched from `builds-mathlib` to `breaks-mathlib`.

semorrison:
Sorry, combined Mathlib/Lean CI is really struggling at the moment, due to the number of simultaneous breaking changes we have going.

Since we previously got a green tick, let's decide to ignore subsequent failures.

PR #2717, opened 2023-10-20
feat: Use private names in where and local let

nomeata:
(no need to look yet, running this to see what breaks)

Motivation:

Hide local definitions from generated documentation, and prevent docstring
linters from complaining about missing docstrings in where clauses.

### DIFF
--- a/src/Lean/Elab/LetRec.lean
+++ b/src/Lean/Elab/LetRec.lean
@@ -41,7 +41,7 @@ private def mkLetRecDeclView (letRec : Syntax) : TermElabM LetRecView := do
         throwErrorAt declId "'let rec' expressions must be named"
       let shortDeclName := declId.getId
       let currDeclName? ← getDeclName?
-      let declName := currDeclName?.getD Name.anonymous ++ shortDeclName
+      let declName := mkPrivateName (← getEnv) <| currDeclName?.getD Name.anonymous ++ shortDeclName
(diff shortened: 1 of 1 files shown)


nomeata:
wip

nomeata:
WIP

nomeata:
Some tests fail, for example `./tests/lean/run/inlineWithNestedRecIssue.lean`:
```
import Lean

@[inline] private unsafe def mapMonoMImp [Monad m] (as : Array α) (f : α → m α) : m (Array α) :=
  go 0 as
where
  @[specialize] go (i : Nat) (as : Array α) : m (Array α) := do
    if h : i < as.size then
      let a := as[i]
      let b ← f a
      if ptrEq a b then
        go (i+1) as
      else
        go (i+1) (as.set ⟨i, h⟩ b)
    else
      return as

set_option trace.Compiler.result true
#eval Lean.Compiler.compile #[``mapMonoMImp, ``mapMonoMImp.go]
```

So there is a need to access the local definitions from the outside. And test `./tests/lean/docStr.lean` also shows that by design local definitions are available from the outside…

So maybe another way must be found to let documentation and linters recognize local definitions (and, for example, hide them from the API _unless_ they have a docstring).

Or maybe local definitoins should support `private`, `public`, `protected` qualifiers, maybe even defaulting to `private`.

PR #2720, opened 2023-10-20
feat: nix: ability to build packages which require presence of external files

infinitisno:
Some packages rely on presence of external files, for example, [proof-widget4](https://github.com/leanprover-community/ProofWidgets4) relies on presence of javascript files.

When the `modHasDependencies` option is true, source directory will be used as is instead of using directory containing only corresponding .lean file for certain component.

With this option, we can build `proof-widget4` and `mathlib4` via `nix`.

  There is no RFC/issue related to this.
  
  Decided to create PR without creating issue first, because there is not much changes.


0x450x6c:
Found that I can change `buildMod` attributes via `overrideBuildModAttrs`, so will close this PR.

PR #2722, opened 2023-10-20
chore: change simp default to decide := false

collares:

This is a follow-up to #1888, which changed `dsimp` defaults but intended to change `simp` as well, and has been tried before in #2068, #2166 and #2192. I understand this was probably discussed in a dev meeting a long time ago (judging from the labels in #2068) and Scott Morrison [states on Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Regression.20in.20simp/near/396366051) that "in principle there is agreement to merge such a change". Quoting Heather Macbeth in #2068,

> it seems like users will write inefficient decidability instances often enough that the same point of confusion [with `decide := true` defaults] will occur again.

One example which [happened recently to Bhavik Mehta](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Regression.20in.20simp/near/396322203) is quoted below. `simp` takes a long time to close the following goal because it tries and fails to use various decidability instances in Mathlib.

```
import Mathlib
example : (fun i : Nat => 3) ≠ fun i : Nat => 4 := by simp [Function.funext_iff]
```
---

The simp_arith changes are only needed because I couldn't figure out how to pass `config := {decide := true}` to https://github.com/leanprover/lean4/blob/583e023314a61efa546e8c32e6395ed83744ef8b/src/Lean/Compiler/LCNF/PassManager.lean#L49 but they might be desirable anyway.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -11,6 +11,8 @@ of each version.
 v4.4.0 (development in progress)
 ---------
 
+* By default, `simp` will no longer try to use Decidable instances to rewrite terms. In particular, not all decidable goals will be closed by `simp`, and the `decide` tactic may be useful in such cases. The `decide` simp configuration option can be used to locally restore the old `simp` behavior, as in `simp (config := {decide := true})`; this includes using Decidable instances to verify side goals such as numeric inequalities.
+
 v4.3.0
 ---------
 
--- a/tests/lean/run/simpDecide.lean
+++ b/tests/lean/run/simpDecide.lean
@@ -1,0 +1,18 @@
+example : 0 < 4 := by decide
+
+example : 0 < 4 := by
+  fail_if_success { simp only }
+  simp (config := { decide := true }) only
+
+def fib : Nat → Nat
+| 0 => 0
+| 1 => 1
+| n+2 => fib (n+1) + fib n
+
+theorem fib_two : fib 2 = 1 := rfl
+
+example : 1 = fib (1 + 1) := by
+  fail_if_success { simp only [fib_two] }
+  simp (config := { decide := true }) only [fib_two]
+
+example : 1 = fib (1 + 1) := by simp only [fib_two, (by decide : 1 + 1 = 2)]


collares:
WIP

collares:
awaiting-review

semorrison:
So close! Mathlib builds, just some tests failing. :-) Please ping me when Mathlib builds!

src/Init/Data/Nat/Basic.lean
```diff
@@ -629,7 +629,7 @@ protected theorem sub_lt_sub_left : ∀ {k m n : Nat}, k < m → k < n → m - n
 @[simp] protected theorem zero_sub (n : Nat) : 0 - n = 0 := by
   induction n with
   | zero => rfl
-  | succ n ih => simp [ih, Nat.sub_succ]
+  | succ n ih => simp [ih, Nat.sub_succ]; decide
```
semorrison, line 632: Could you please use `(config := ...)` here.

src/Init/SimpLemmas.lean
```diff
@@ -152,7 +153,7 @@ theorem Bool.or_assoc (a b c : Bool) : (a || b || c) = (a || (b || c)) := by
 @[simp] theorem bne_self_eq_false' [DecidableEq α] (a : α) : (a != a) = false := by simp [bne]
 
 @[simp] theorem Nat.le_zero_eq (a : Nat) : (a ≤ 0) = (a = 0) :=
-  propext ⟨fun h => Nat.le_antisymm h (Nat.zero_le ..), fun h => by simp [h]⟩
+  propext ⟨fun h => Nat.le_antisymm h (Nat.zero_le ..), fun h => by simp [h]; decide⟩
```
semorrison, line 156: and here, and etc.

stage0/src/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ endif()
 include(ExternalProject)
 project(LEAN CXX C)
 set(LEAN_VERSION_MAJOR 4)
-set(LEAN_VERSION_MINOR 2)
+set(LEAN_VERSION_MINOR 3)
```
semorrison, line 12: Please rebase your changes onto master. We shouldn't be seeing diffs like this.

src/Init/SimpLemmas.lean
collares: This one failed with:

```
Init/SimpLemmas.lean:156:84: error: unknown constant 'Lean.Meta.Simp.Config'
```

Any suggestions/workarounds?

src/Init/Data/Nat/Basic.lean
collares: This one failed with:

```
Init/SimpLemmas.lean:156:84: error: unknown constant 'Lean.Meta.Simp.Config'
```

Any suggestions/workarounds?

collares: I replaced `simp` by `simp only`

src/Init/SimpLemmas.lean
collares: I replaced `simp` by `rw`

stage0/src/CMakeLists.txt
collares: I rebased the PR onto master. Perhaps there has been no stage0 update on master in a while?

src/Init/SimpLemmas.lean
semorrison: No, this is some staging issue. Maybe we can ask @Kha or @digama0 for assistance on this, I still don't understand how to do these updates.

Please revert changes you've made to avoid the `unknown constant 'Lean.Meta.Simp.Config'` changes.

digama0: This is an interesting edge case. We indeed will not be able to use `simp` config options this early in `Init`, because we have to elaborate the simp config and to do that we need `Lean.Meta.Simp.Config` to exist in the environment. We might be able to move it earlier but dependency order in Init is challenging. IMO `simp; decide` is a reasonable alternative solution since this issue only exists for a few proofs in Init.

stage0/src/CMakeLists.txt
collares: @semorrison I've now rebased this PR onto the recent stage0 update in master. The stage0 diff is a lot smaller.

semorrison:
We probably also need a test that is explicitly about this configuration switch. i.e. a new test file that:

* demonstrates that with the default settings `simp` fails on something that can be done with `simp (config := {decide := true})`

semorrison:
Oh dear, there's a merge conflict in `stage0` now. @collares, do you think you can sort this out? Ping if you need help.

collares:
awaiting-review

collares:
Moved the relnote to v4.4.0, given that v4.3.0-rc1 was just released. I can revert it if desired.

PR #2724, opened 2023-10-21
chore: remove "paper cut" when using `Fin USize.size`

leodemoura:


leodemoura:
!bench

leodemoura:
@semorrison Could you please help me with this one?

semorrison:
Okay, hopefully Mathlib is happy with this one now.

semorrison:
Mathlib CI integration is just a mess at the moment due to the number of breaking changes we have in the air at the same time.

Since this one has previously succeeded with Mathlib, but doesn't now, I think we should declare success, and I'll sort out the details once it is actually in the nightly.

PR #2725, opened 2023-10-21
feat: reorder Lake help

kim-em:
Proposed reordering of lake help, so the most important things come first.

### DIFF
--- a/src/lake/Lake/CLI/Help.lean
+++ b/src/lake/Lake/CLI/Help.lean
@@ -13,6 +13,20 @@ uiVersionString ++ "
 USAGE:
   lake [OPTIONS] <COMMAND>
 
+COMMANDS:
+  new <name> <temp>     create a Lean package in a new directory
+  init <name> <temp>    create a Lean package in the current directory
+  build <targets>...    build targets
+  exe <exe> <args>...   build an exe and run it in Lake's environment
+  clean                 remove build outputs
+  env <cmd> <args>...   execute a command in Lake's environment
+  update                update dependencies and save them to the manifest
+  upload <tag>          upload build artifacts to a GitHub release
+  script                manage and run workspace scripts
+  scripts               shorthand for `lake script list`
+  run <script>          shorthand for `lake script run`
+  serve                 start the Lean language server
+
 OPTIONS:
   --version             print version and exit
   --help, -h            print help of the program or a command and exit
@@ -26,20 +40,6 @@ OPTIONS:
   --rehash, -H          hash all files for traces (do not trust `.hash` files)
   --update, -U          update manifest before building
   --reconfigure, -R     elaborate configuration files instead of using OLeans
-
-COMMANDS:
-  new <name> <temp>     create a Lean package in a new directory
-  init <name> <temp>    create a Lean package in the current directory
-  build <targets>...    build targets
-  update                update dependencies and save them to the manifest
-  upload <tag>          upload build artifacts to a GitHub release
-  clean                 remove build outputs
-  script                manage and run workspace scripts
-  scripts               shorthand for `lake script list`
-  run <script>          shorthand for `lake script run`
-  serve                 start the Lean language server
-  env <cmd> <args>...   execute a command in Lake's environment
-  exe <exe> <args>...   build an exe and run it in Lake's environment
 
 See `lake help <command>` for more information on a specific command."
 


tydeu:
@semorrison Unfortunately, this is non-standard. Both other lean programs (e.g.,  `elan`) and other similar systems (e.g., `cargo`) use the current ordering. My reasoning as to why would be that, for CLIs, the-easiest-to-see information is the stuff at the bottom (because of terminal auto-scrolling). That is, other CLIs agree that commands are the most important; they just believe the the most important information should come *last* in in the output (due to scrolling).

digama0:
> That is, other CLIs agree that commands are the most important; they just believe the the most important information should come last in in the output (due to scrolling).

I'm not sure I believe this, since the most information-dense part is the BNF-ish usage line and this invariably comes at the beginning.

tydeu:
@digama0 

> I'm not sure I believe this, since the most information-dense part is the BNF-ish usage line and this invariably comes at the beginning.

Good point! Maybe its more of a book-end thing? If you pipe the help into e.g. `less` then what comes first is more prominent, but if you are on the terminal what comes less is more prominent. Usage is also generally more important for commands with little text (and no subcommands) than ones with many subcommands (anecdotally, rarely do I look at such commands usage text).  Regardless, there is a clear standard here and it is curious why.

digama0:
> Regardless, there is a clear existing standard here and it is curious why.

I'm also not really convinced of this.
* `npm` has commands on the front page and no options at all, and in subcommands the commands come first (listed under "usage") and the options come after.
* `gcc`/`clang` has only options, no subcommands (the main argument is apparently self explanatory)
* `7z` has commands first, then "switches" (options)
* `tar` has "examples", then "main operation mode" then "operation modifiers" and other options grouped by level of obscurity
* `grep` has "patterns" (the main argument) then options
* `hg` has only subcommands, with `hg --help -v` it shows subcommands then global options
* `git` has only subcommands, I don't think it has global options
* `less` has key commands, then options (oddly it doesn't have any usage line or indication of the CLI structure)
* `dune` has a manpage style with "synopsis", "description", "commands", "common options"

Most manpage-styled things like `git reset --help` will start with an explanation of the main part of the command (in prose) in a "description" section and do options after.

tydeu:
@digama0 

Your examples are bit weird to me. The majority of them don't even have both options and subcommands and thus are not applicable to topic at hand. In a few of them (such as NPM), you also conflate the command used in the usage section with a subcommand section for reasons I do not understand. Man-page style (e.g., Git) are naturally read top-to-bottom because `man` incorporates a version of `less` and e.g. Git displays such help pages via opening an HTML page which would also be read top-to-bottom without the problem of terminal scroll. 

`7z` is a good example of a utility that lists commands first, but it also not following a standard CLI template.  On the other hand, your example of `rg` does use a standard CLI template in the form `clap`. But it so happens that `clap` does put subcommands after options (and then the discussion after the options) -- `rg` doesn't demonstrate this only because it has no subcommands. For instance, my previous example `elan` also uses `clap` and organizes them this way. Similarly, the `argparse` library for Python also organizes options and [subcommands](https://docs.python.org/3/library/argparse.html#sub-commands) in this order. This is why I would assert  the current order is a rather standard one because two of the biggest CLI libraries use it.

digama0:
My examples were a random sampling of popular CLI utilities, not selecting for those that have both subcommands and options because I think there is a broader point to be made that CLI help is meant to be read top to bottom. Some of them, like `clang`, have laughably useless output if you only look at the end (and the `clang` help is so long that the top is out of the scroll-back by the time it finishes).

In any case, I don't think it's a very interesting question. Most of these examples have the important information at the top, usually "examples" or "usage" rather than subcommands specifically, but lake doesn't do that, so subcommands are the next best thing. The options are not relevant to most users.

tydeu:
@digama0 

> Most of these examples have the important information at the top, usually "examples" or "usage" rather than subcommands specifically, but lake doesn't do that

Lake does do that. It has a usage line, just like Cargo's and its help docs share Cargo's order (as of the latest Rust v1.73). Also, anecdotally, from my personal experience, the stuff at the bottom of a CLI is most useful (as that is what I see first), and I think I have made a reasonable case that commands often go there.

> By the way, `clap` does not seem to use this order anymore, see [this docs example](https://docs.rs/clap/latest/clap/_derive/_tutorial/chapter_0/index.html#quick-start) which has "usage", "commands", "arguments", "options" in that order.

Interesting. This was a v4.0 change (clap#4125) that appears to have been made by the preference of the developer (and has not been adopted by Cargo). The developer's example of choice was `git-stash`, which as I said above I believe to be a bad example because Git docs are viewed via HTML which is naturally top-down while CLIs are bottom-up.

tydeu:
@semorrison As this is primarily an issue of user preference and there appear to be conflicting signals in other CLIs as to what is best, I have discussed this with @digama0 and it seems like it might be wise to hold a poll on which ordering is best and use that to determine whether to merge. What do you think?

semorrison:
Okay. I will do so, but I have to admit I think this is a silly discussion. The useful and important information must always come first in any explanation.

I've always used the rule that you lose 90% of your audience after every "section" of any written piece. (This is true of CLI help, documentation, academic papers, etc...) You can hope to reduce the 90% loss to 80% by writing clearly and informatively (perhaps even entertainingly), but an author's job is nevertheless to maximize the information conveyed to an average reader!

semorrison:
Poll on [zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/reorder.20Lake.20help.20text/near/398590616), per request.

tydeu:
@semorrison 

>  The useful and important information must always come first in any explanation.

I agree. I think the core of the debate is whether people read CLI help bottom-up or top-down, which determines whether the top or the bottom is "first". Personally, I read them bottom-up because that is the part I see first (due to how CLI scroll works)

semorrison:
https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/reorder.20Lake.20help.20text/near/398590657 is pretty unanimous.

PR #2732, opened 2023-10-21
fix: fixes #1926

### LINKED ISSUE #1926: `simp` not rewriting in binder
### Prerequisites

### Description

In Lean 3, `simp` works in the following snippet while in Lean 4 it doesn't. This is maybe due to `simp` using `rw` instead of `subst` internally?

```lean
example (q : p → Prop) (h : p = True)
  (h' : ∀(q : True → Prop), (∀ x, q x) ↔ q True.intro) :
    (∀ h', q h') ↔ q (h.symm ▸ trivial) := by
  simp only [h, h'] -- no change in goal in Lean 4
  -- rw [h] -- motive not type correct :-/
  subst h
  simp [h'] -- works fine now

```

leodemoura:


leodemoura:
!bench

src/Init/SimpLemmas.lean
```diff
@@ -41,6 +41,12 @@ theorem implies_congr_ctx {p₁ p₂ q₁ q₂ : Prop} (h₁ : p₁ = p₂) (h
 theorem forall_congr {α : Sort u} {p q : α → Prop} (h : ∀ a, p a = q a) : (∀ a, p a) = (∀ a, q a) :=
   (funext h : p = q) ▸ rfl
 
+theorem forall_prop_domain_congr {p₁ p₂ : Prop} {q₁ : p₁ → Prop} {q₂ : p₂ → Prop}
+    (h₁ : p₁ = p₂)
+    (h₂ : ∀ a : p₂, q₁ (h₁.substr a) = q₂ a)
+    : (∀ a : p₁, q₁ a) = (∀ a : p₂, q₂ a) := by
+  subst h₁; simp [← h₂]
```
eric-wieser, line 48: This is true more generally for `Sort u`:
```suggestion
theorem forall_domain_congr {p₁ p₂ : Sort _} {q₁ : p₁ → Sort _} {q₂ : p₂ → Sort _}
    (h₁ : p₁ = p₂)
    (h₂ : ∀ a : p₂, q₁ (h₁.mpr a) = q₂ a)
    : (∀ a : p₁, q₁ a) = (∀ a : p₂, q₂ a) := by
  subst h₁
  have : q₁ = q₂ := funext h₂
  subst this
  rfl
```

src/Lean/Meta/Tactic/Simp/Main.lean
```diff
@@ -723,7 +723,41 @@ where
...
+          /- Using
+          ```
+          theorem forall_prop_domain_congr {p₁ p₂ : Prop} {q₁ : p₁ → Prop} {q₂ : p₂ → Prop}
+              (h₁ : p₁ = p₂)
+              (h₂ : ∀ a : p₂, q₁ (h₁.substr a) = q₂ a)
+              : (∀ a : p₁, q₁ a) = (∀ a : p₂, q₂ a)
+          ```
+          Remark: we should consider whether we want to add congruence lemma support for arbitrary `forall`-expressions.
+          Then, the theroem above can be marked as `@[congr]` and the following code deleted.
+          -/
```
eric-wieser, line 744: Probably you already saw https://github.com/leanprover/lean4/issues/1926#issuecomment-1759397589, but in case you didn't; I stumbled across what might be a neat trick to enable this behavior (and probably also arrow support in `calc`); define
```lean
-- an alias for `∀` to thwart checks that complain that `∀` is not a function
def Forall {α} (β : α → Sort _) := ∀ a, β a
```
and temporarily replace `∀`  with `Forall` when looking for congruence lemmas / `Trans` instances, such that the existing code-path for `Expr.app` can be reused.

PR #2733, opened 2023-10-22
fix: avoid "Argument list is too large" on building via nix.

infinitisno:
The `buildLeanPackage` uses environment variable to store a list of libraries,
then passes that variable to the lean as CLI arguments.
According to this (https://stackoverflow.com/a/28865503), CLI arguments and environment variables share same space, therefore there is an error "Argument list is too large" when building a package with too many components.
(for example - mathlib4)

This change makes `buildLeanPackage` to use a file instead of environment variable to store a list of libraries.

  Didn't created an issue, because patch is trivial.


Kha:
The Nix integration has been deprecated

Issue #2669, opened 2023-10-11
simp unfolding let even with zeta := false option


PatrickMassot:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to mathlib4 or std4.

Note this issue may be related to  #2281, I am not sure.

### Description

`simp (config :={zeta := false})` sometimes unfold let values.

### Context

This is a regression compared to Lean 3 that complicates porting the sphere eversion project for instance. It was discussed [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/simp.20unfolding.20let.20bindings/near/391576668).

### Steps to Reproduce

```lean
def f : Nat → Nat := fun x ↦ x - x

@[simp]
theorem f_zero (n : Nat) : f n = 0 := Nat.sub_self n

example (n : Nat) : False := by
  let g := f n
  have : g + n = n := by 
    simp (config := {zeta := false}) [Nat.zero_add]
  sorry
```

**Expected behavior:** 

The simp should fail. In Lean 3 the `dsimp` line below is needed.

```lean
-- Lean 3 code
def f : nat → nat := fun x, x - x

@[simp]
lemma f_zero (n : ℕ) : f n = 0 := nat.sub_self n

example (n : nat) : false :=
begin
  let g := f n,
  have : g + n = n,
  { -- dsimp only [g],
    simp [nat.zero_add] },
  sorry
end
```

**Actual behavior:**

Lean 4 happily closes the goal. Of course in my example this looks like a feature, but in actual examples it prevents `simp` from closing the goal, by sending it to a dead end.

### Versions

Lean (version 4.2.0-rc1, commit https://github.com/leanprover/lean4/commit/a62d2fd4979671b76b8ab13ccbe4fdf410ec0d9d, Release)
Ubuntu 22.04

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

Issue #2281, opened 2023-06-22
simp (with beta/zeta disabled) and discrimination trees


leodemoura:
We disabled beta, zeta, iota reduction in `simp`. However, there is a nasty interaction with the discrimination tree module.
Consider the following example:
```lean
example (h : a = b) : (fun x => a + x) 0 = b := by
  simp (config := { beta := false })
  -- h : a = b |- a = b    <<<< Goal here
  done
```
We requested beta reduction to be disabled, but `simp` performed indirectly using the discrimination tree module.
Note that the following example works as expected
```lean
example (h1 : a = b) (h2 : c = 0) : (fun x => a + x) c = b := by
  simp (config := { beta := false })
  -- h : a = b |- (fun x => a + x) c = b
  done
```
The issue is that the discrimination tree module retrieves rewriting rules modulo `whnf` with `[reducible]` setting. Then, when `simp` tries to retrieve rewriting rules for the term `(fun x => a + x) 0`, it finds `Nat.add_zero : a + 0 = a`. We need a flag in the discrimination tree modulo to control reduction when retrieving candidates. As an extra complication, we may still want to reduce `[reducible]` definitions, and beta-reduction here should be allowed.

The issue above may also cause performance issues.

PR #2734, opened 2023-10-22
fix: issues #2669 #2281

leodemoura:
@semorrison This PR fixes a high priority issue, but will probably break user-defined functions that manipulate discrimination trees.

### DIFF
--- a/tests/lean/run/letDeclSimp.lean
+++ b/tests/lean/run/letDeclSimp.lean
@@ -1,0 +1,12 @@
+example (a : Nat) : let n := 0; n + a = a := by
+  intro n
+  fail_if_success simp (config := { zeta := false })
+  simp (config := { zeta := false }) [n]
+
+example (a b : Nat) (h : a = b) : let n := 0; n + a = b := by
+  intro n
+  fail_if_success simp (config := { zeta := false })
+  trace_state
+  simp (config := { zeta := false }) [n]
+  trace_state
+  simp [h]


leodemoura:
!bench

src/Lean/Meta/Tactic/Simp/Rewrite.lean
```diff
@@ -142,11 +142,24 @@ def tryTheorem? (e : Expr) (thm : SimpTheorem) (discharge? : Expr → SimpM (Opt
         tryTheoremCore lhs xs bis val type e thm (eNumArgs - lhsNumArgs) discharge?
       else
         return none
+
+/--
+Reture a WHNF configuration for retrieving `[simp]` from the discrimination tree.
```
semorrison, line 147:
```suggestion
Reture=n a WHNF configuration for retrieving `[simp]` from the discrimination tree.
```

src/Lean/Meta/Tactic/Simp/Rewrite.lean
```diff
@@ -142,11 +142,24 @@ def tryTheorem? (e : Expr) (thm : SimpTheorem) (discharge? : Expr → SimpM (Opt
...
+If user has disabled `zeta` and/or `beta` reduction in the simplifier, we must also
+disabled them when retrieving lemmas from discrimination tree. See issues: #2669 and #2281
```
semorrison, line 149:
```suggestion
disable them when retrieving lemmas from discrimination tree. See issues: #2669 and #2281
```

src/Lean/Meta/UnificationHint.lean
```diff
@@ -26,8 +26,10 @@ structure UnificationHints where
 instance : ToFormat UnificationHints where
   format h := format h.discrTree
 
+def UnificationHints.config : WhnfCoreConfig := {}
```
semorrison, line 29: Is this is right value? Unification hints were previously using `simpleReduce := true`, so I thought we should have `iota := ff` here.

src/Lean/Meta/DiscrTree.lean
```diff
@@ -249,16 +249,16 @@ def hasNoindexAnnotation (e : Expr) : Bool :=
 
 /--
 Reduction procedure for the discrimination tree indexing.
-The parameter `simpleReduce` controls how aggressive the term is reduced.
+The parameter `config` controls how aggressive the term is reduced.
```
semorrison, line 252:
```suggestion
The parameter `config` controls how aggressively the term is reduced.
```

src/Lean/Meta/UnificationHint.lean
leodemoura: Good catch. I fill fix it.

PR #2738, opened 2023-10-23
fix: make `rw [foo]` look in the local context for `foo` before it looks in the environment

### LINKED ISSUE #2729: `rw [h]` uses `h` from the environment in preference to `h` from the local context
### Prerequisites

### Description

`rw` seems to priorities other statements over variables in context. I would expect the reverse.

### Context

[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/rw.20bug.20priority.20of.20variables/near/397755221) and (non-minimised) [question in New Members stream](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/.E2.9C.94.20Lean.204.20web.20and.20VS/near/397755570)

### Steps to Reproduce

See the following MWE:

```lean
import Lean

open List

example (A B : Prop) (all : A ↔ B) (b : B) : A := by
  rw[all] -- error: failed to rewrite using equation theorems for 'List.all'
  assumption
```

**Expected behavior:** The above MWE should succeed.

**Actual behavior:** `rw` fails.

Note that all other tactics - `simp`, `apply`, `exact`, ... do correctly use the local `all` instead of `List.all`, so it seems to be a `rw`-specific issue.

Moreover, the same behaviour shows it `open List` is replaced with `def all := 3`.

### Versions

(lean4web)
4.2.0-rc3

### Impact

thorimur:

Fixes #2729.

This requires some changes to mathlib4 where the opposite behavior was being exploited; we now must namespace `foo` to make it refer to a constant in the environment (e.g. `rw [Bar.foo]`), which has the nice side effect of making explicit which `foo` we're referring to. See leanprover-community/mathlib4#7872.

### DIFF
--- a/tests/lean/rwPrioritizesLCtxOverEnv.lean
+++ b/tests/lean/rwPrioritizesLCtxOverEnv.lean
@@ -1,0 +1,21 @@
+/-! # Ensure `rw` prioritizes the local context over the environment
+
+This test ensures that `rw [foo]` looks for `foo` in the local context before it looks for a
+constant named `foo` in the environment. See issue #2729.
+-/
+
+-- Introduce `foo` to the environment.
+/-- A constant whose name should conflict with that of a local declaration. -/
+def foo : List Nat := [0]
+
+/-! ## Ensure that `foo` from the LCtx is used instead of the constant `foo` -/
+
+example (A B : Prop) (foo : A ↔ B) (b : B) : A := by
+  rw [foo] -- should be interpreted as `foo : A ↔ B`, not `foo : List Nat`, and succeed
+  assumption
+
+/-! ## Ensure that we can access the constant `foo` by writing `rw [_root_.foo]` -/
+
+set_option linter.unusedVariables false in
+example (A B : Prop) (foo : A ↔ B) : _root_.foo = [0] := by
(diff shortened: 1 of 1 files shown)


thorimur:
WIP

thorimur:
awaiting-review

tests/lean/rwPrioritizesLCtxOverEnv.lean
```diff
@@ -0,0 +1,13 @@
...
+constant named `foo` in the environment. See issue #2729.
+-/
+
+-- Introduce `foo` to the environment.
+/-- A constant we will *not* be rewriting with. -/
+def foo : List Nat := [0]
+
+example (A B : Prop) (foo : A ↔ B) (b : B) : A := by
+  rw [foo] -- should be interpreted as `foo : A ↔ B`, not `foo : List Nat`, and succeed
+  assumption
```
semorrison, line 13: I think we need a test that shows that we can still rewrite by the global constant, by fully specifying it (e.g. with `_root_`).

thorimur: Good point; added in c0e3492 (rebased for nicer commit history). :)

PR #2743, opened 2023-10-24
fix: don't pack `._` files on MacOS

### LINKED ISSUE #2742: `lake upload` run on macOS produced a tar with fake `.olean`s beginning with `._`
### Description

See for example: https://github.com/leanprover-community/ProofWidgets4/releases/download/v0.0.19/macOS-64.tar.gz which was created via `lake upload` on a macos computer.

In order to see the spurious files, you'll need to unpack this file on a non-macos computer.

See https://apple.stackexchange.com/questions/14980/why-are-dot-underscore-files-created-and-how-can-i-avoid-them for a description of these files.

Note that these files then break all sorts of things. e.g. they break `lean4checker` which runs

```
      for path in (← SearchPath.findAllWithExt sp "olean") do
        if let some m := (← searchModuleNameOfFileName path sp) then
```

but then it can't cope with these files (because despite looking like `.olean` files they are not). 

### Impact

digama0:
Fixes #2742. This needs to be tested on MacOS; this fix is based on https://stackoverflow.com/questions/8766730/tar-command-in-mac-os-x-adding-hidden-files-why .

### DIFF
--- a/src/lake/Lake/Build/Actions.lean
+++ b/src/lake/Lake/Build/Actions.lean
@@ -117,4 +117,6 @@ def tar (name : String) (dir : FilePath) (file : FilePath)
   proc {
     cmd := "tar"
     args := args ++ #["-f", file.toString, "-C", dir.toString, "."]
+    -- don't pack `._` files on MacOS
+    env := if Platform.isOSX then #[("COPYFILE_DISABLE", "true")] else #[]
   }


semorrison:
Checked locally by upload to ProofWidgets from macos, and downloading the zip to linux. Looks good!

PR #2749, opened 2023-10-25
refactor: `.lake` directory for Lake outputs

### LINKED ISSUE #2713: `lake upload` fails with `tar: file changed as we read it`
### Prerequisites

### Description

`lake upload <tag>` fails with

```
error: > tar -c -z --exclude=*.tar.gz --exclude=*.tar.gz.trace -f ./build/linux-64.tar.gz -C ./build .
error: stderr:
tar: .: file changed as we read it
```

for a large `build` directory (~100M) on GitHub actions with Ubuntu, the same command succeeds for Mac.

### Context

It's discussed on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/lake.20upload.20for.20Cloud.20Release), the probable cause is identified.

### Steps to Reproduce

On Ubuntu, e.g.`mcr.microsoft.com/devcontainers/base:ubuntu`

1. For an existing Lean repo, create a release with tag <tag> 
2. Put a large trunk of files in `build`, e.g. the source and built binaries of a C++ project
3. Run `lake upload <tag>` locally from the root of the repository.

More minimized way to reproduce it, is:

1. For an existing local Lean repo
2. Put a large trunk of files in `build`
3. Run `tar -c -z --exclude=*.tar.gz --exclude=*.tar.gz.trace -f ./build/linux-64.tar.gz -C ./build .` which is how lake calls tar

**Expected behavior:**

The tar is successfully generated and uploaded to the release.

**Actual behavior:**

As shown above, tar fails with `tar: .: file changed as we read it`, and the archive is not created in full (the file is created but incomplete)

### Versions

`Lean (version 4.2.0-rc3, commit 5e0f6049c002, Release)`

`Linux d4df9d079930 6.3.13-linuxkit #1 SMP PREEMPT Thu Sep  7 07:48:47 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux`

```
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.3 LTS
Release:        22.04
Codename:       jammy
```

### Additional Information

After debugging, the cause is identified to be

> The issue lies in the way lake calls tar, the archive it generates is inside `build` which changes the modified time of directory `build`, according to [the source code of GNU tar](https://git.savannah.gnu.org/cgit/tar.git/tree/src/create.c), it fails at checking the mtime of the dir, without taking the exclude patterns into account.

The issue doesn't happen if tar takes too short to finish, or if the `tar.gz` files already exists.

The issue can be bypassed 100% by using raw `tar` to generate the archive elsewhere.

The issue may be solved following [this SO answer](https://stackoverflow.com/a/37993307), and they seem to work after repeated local tests.

### Impact

When using `lake upload` for its Cloud Release, FFI projects will fail to upload the prebuilt binaries for Linux.

tydeu:
Store the build directory, build archives, package directory, and compiled configuration inside a dedicated directory for Lake outputs (i.e., `.lake`). ~~There is also now a more rigorous check on `lakefile.olean` compatibility.~~ (**edit:** split out into #2842)

Closes #2713.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -12,6 +12,7 @@ v4.4.0 (development in progress)
 ---------
 
 * By default, `simp` will no longer try to use Decidable instances to rewrite terms. In particular, not all decidable goals will be closed by `simp`, and the `decide` tactic may be useful in such cases. The `decide` simp configuration option can be used to locally restore the old `simp` behavior, as in `simp (config := {decide := true})`; this includes using Decidable instances to verify side goals such as numeric inequalities.
+* **Lake:** Moved the default build directory (e.g., `build`), default packages directory (e.g., `lake-packages`), and the compiled configuration (e.g., `lakefile.olean`) into a new dedicated directory for Lake outputs, `.lake`. The cloud release build archives are also stored here, fixing [#2713](https://github.com/leanprover/lean4/issues/2713).
 
 v4.3.0
 ---------


tydeu:
@Kha I was going to ask you if the `IO.FS` additions looked good since they were outside my module ownership. Does your 🎉 imply you are satisfied with them?

tydeu:
@semorrison A note for you when you can take a look: The mathlib CI is currently breaking for reasons other than this PR and I think this PR will likely need a proper CI run on mathlib since it may break some assumptions there about the location of the build directory.

semorrison:
Options for a successful PR run: 
* rebase this PR onto d126c099f4d15282ed5051c8c440107cf514da2b (the commit underlying [`nightly-2023-10-25`](https://github.com/leanprover/lean4-nightly/releases/tag/nightly-2023-10-25)), which should give an immediate run.
* wait until `nightly-2023-10-26` lands and I have fixed Std+Mathlib for it (approximately +10 hours from now), and then push an empty commit (or anything) to this branch.

Kha:
> @Kha I was going to ask you if the `IO.FS` additions looked good since they were outside my module ownership. Does your 🎉 imply you are satisfied with them?

I see they are being used in the lockfile code but I can't tell what's going on for lack of comments. What is the difference to `readFile/writeFile`?

src/lake/Lake/Load/Elab.lean
```diff
@@ -76,48 +76,74 @@ def elabConfigFile (pkgDir : FilePath) (lakeOpts : NameMap String)
...
+    if (← lockFile.pathExists) then
+      let h ← IO.FS.Handle.mk lockFile .readWrite
+      let githash := (← h.readToEnd).trim
+      h.rewind
+      if reconfigure then return (h, false)
+      let .ok oleanMTime ← getMTime olean |>.toBaseIO | return (h, false)
+      unless oleanMTime > (← getMTime configFile) do
+        -- reconfigure if configuration file is newer than the olean
+        return (h, false)
+      return (h, githash = Lean.githash)
```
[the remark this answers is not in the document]
tydeu: Yep, we could hash the `lakefile.lean`. Honestly, I did not consider that. A possible counterpoint, though, is that !00% correctness is not vital here and hashing is probably more time consuming than a simple mtime check.

tydeu:
@Kha

> What is the difference to `readFile/writeFile`?

`readFile`/`writeFile` would allow two racing `lake` processes to load a bad olean (e.g., one writes an updating lock and then another loads the old olean before the other writes the new one, or one writes a new olean and another loads it before the other writes the new lock). The new primitives allow us to open a handle, read the old hash, and not release it until we write the new one. I will add a comment to this effect in the code.

Kha:
> The new primitives allow us to open a handle, read the old hash, and not release it until we write the new one.

Mmh, if I understand you correctly and you believe that `open`/`fdopen` locks the file against concurrent access from other processes, that is simply untrue I'm afraid.

src/lake/Lake/Load/Elab.lean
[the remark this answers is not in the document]
tydeu: @Kha Yeah, sorry, I meant I had not considered it while writing the code (i.e., I forgot).

tydeu:
@Kha 

> that is simply untrue I'm afraid. That is, it is still racy.

Oops, yeah, it appears I confused the exclusion offered by e.g. `O_EXCL` in `writeNew` with general file handles. I guess I actually needed `lock`/`unlock` file handle bindings for [`flock`](https://www.man7.org/linux/man-pages/man2/flock.2.html), [`LockFileEx`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-lockfileex), and [`UnlockFileEx`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-lockfileex).

Kha:
> I guess I actually also needed `lock`/`unlock` file handle bindings for [`flock`](https://www.man7.org/linux/man-pages/man2/flock.2.html), [`LockFileEx`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-lockfileex), and [`UnlockFileEx`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-lockfileex).

Phew, I'm impressed you made that work in quick time but this looks like it will take some time to review. Why not leave worrying about races to #2766 as mentioned there? On a positive note, this looks like the right functions to implement a general build lock that avoids the issues of the first try.

Kha:
@joehendrix Speaking in general, I think it would be great if you could review additions to `Init.System.IO` and similar modules from a standard library perspective - they may as well be in std4 if it wasn't for bootstrapping. But for this one, let's first wait which functions we actually need in the end.

src/lake/Lake/Load/Elab.lean
```diff
@@ -76,48 +76,98 @@ def elabConfigFile (pkgDir : FilePath) (lakeOpts : NameMap String)
...
+    if (← lockFile.pathExists) then
+      let h ← IO.FS.Handle.mk lockFile .readWrite
+      unless (← h.tryLock) do
+        error "another process configured the package first"
+      if reconfigure then
+        return (h, false, lakeOpts)
+      let contents ← h.readToEnd; h.rewind
+      let .ok (lock : ConfigLock) := Json.parse contents >>= fromJson?
+        | error "compiled configuration is invalid; you will need to reconfigure"
+      return (h, lock.leanHash = Lean.githash ∧ lock.configHash = configHash, lock.options)
```
Kha, line 152: Apologies for nitpicking changes I requested, but I find this code curious (i.e. not a merge blocker), why does it stray so much from all other traces in Lake work? Why not just `mix` the hashes together instead of putting them in a JSON? It *is* a file trace in the end, isn't it?

tydeu: @Kha There at two different things at play here:
1. Lake stores the options verbatim so the olean can be created with the same options when the cache is invalidate without requiring the user to provide them again.
2. I avoided build-style trace mixing because our hashing algorithm is not cryptographically secure and could have clashes and I want to avoid another `lake clean` bug as much as possible.

Kha: Ok, I wasn't sure whether the option persistence feature was more accidental or requested

tydeu:
Anyway, this should be done. The last `platform` addition was some thing I was supposed to add after a discussion with Mario, but it slipped my mind until now. The `lakefile.olean` should now be properly lock-guarded to avoid race conditions and store enough information to be invalidated by Lake on any relevant change.

tydeu:
@Kha Am I correct in understanding that this is approved for merge (from your previous statement that your comment was not a merge blocker)?

semorrison:
@tydeu, are you sure the Mathlib build error is not due to this change? It is ProofWidgets not being able to find a build artifact.

tydeu:
@semorrison Yeah, the error now is solely due to this change. Previously there were others mixed in which resulted in me not noticing it at first.

tydeu:
@semorrison I fixed half of the mathlib breakage that is caused by this PR.  That last breakage is due to a hard-coded reference to the `build` directory in ProofWidgets. That one will require an update to ProofWidgets itself. 

On that note, I think everything that can be done in this PR is now done. However, I will wait until @Kha comes back from vacation and approves the core changes to merge it. From his previous comment, @joehendrix may also wish to take a look.

Kha:
> Why not leave worrying about races to #2766 as mentioned there?

I don't see any reply to this question

tydeu:
@Kha 

> > Why not leave worrying about races to #2766 as mentioned there?
> 
> I don't see any reply to this question

Oh, sorry, I somehow missed that message. For racing, the problem is that we want to ensure the other inputs are correct as well (the lakefile and platform) and do so even if the Lean version hash in the `.olean` is correct. For example, we don't want Lake to race and load the `.olean` for the wrong version of the configuration file if a change to the `lakefile.lean` it is made.

tydeu:
@Kha 

>  Thus my mention of whether these APIs should be used to bring back a general build lock instead. That I assume would eliminate the need for `rewrite/truncate` which I was a bit concerned about with that scary docstring but happy to leave that part to @joehendrix if you need them.

I agree that the new locking primitives could definitely be used to implement a lock file that is properly managed (i.e., is kill-proof). But, I assumed that would be in a separate PR. For one, there are also many lock file design decisions that may inhibit it use as guarantee of a single-`lake`-process-only global invariant.

First, the original lock file only locked builds, it did not prohibit, for example, concurrent `lake run` or `lake env` processes. To provide a lock for configuration loading, the lock file would have to be Lake-global. Second, it is not clear that a general lock file is even desirable. One of the complaints about the lock file was that different VS Code tabs wait for each other was often not desirable, which would still be the case even with a properly managed one. This concern could be addressed by more fine-grained locking (e.g., locking individual `.olean` files), but does the exact opposite of what be required for single-`lake`-process-only global invariant.

Alternatively, I guess Lake could just use a separate lock file for the configuration import lock and then only read/write the trace to a separate trace file after acquiring the lock to avoid the `rewind`/ `truncate` calls. That approach had not previously occurred to me. Personally, though, that seems like unnecessary file I/O overhead.

As for the calls themselves, I modeled them off Rust's [`rewind`](https://doc.rust-lang.org/std/io/trait.Seek.html#method.rewind) and Python's [`truncate`](https://docs.python.org/3/library/io.html#io.IOBase.truncate). The scary remark for `truncate` comes from a wrinkle ([discussed here on StackOverflow](https://stackoverflow.com/a/13813461)) introduced by the fact that Lean uses C++'s buffered streams for reading/writing rather than direct file descriptors ([which Python does instead](https://github.com/python/cpython/blob/f4b5588bde656d8ad048b66a0be4cb5131f0d83f/Modules/_io/fileio.c#L1018)).

Maybe the way forward here is for @joehendrix to say whether the `rewind`/`truncate` implementations seems reasonable. Then, if so, keep them; otherwise, try the separate lock file workaround?

joehendrix:
For `truncate`, could it just flush automatically so we don't need a comment?

tydeu:
@joehendrix 

> I've usually seen people take the act of creating a file (with a failure if the file exists) as acquiring a lock and deleting it as releasing it. 

This is how the old Lake lock file was implemented. This caused problems because the file was not deleted if Lake did not complete cleanly (e.g., it was killed). This created a common need to `rm <lock>` for users that frustrated them, hence the the choice to use OS locks here.

> I've also implemented atomic file modifications by writing to a temp file and renaming rather than acquiring a file exclusively.

Unfortunately, the olean needs to be synchronized with trace data that is stored separately outside it (currently in the lock file). Since it is not possible to atomically update both files, a race condition would still persist with such an approach here.

> If we do go this route, I think the Lean code should be very clear on the semantics of any locking operations. (e.g., if `IO.lock h true` succeeds, then no other process (or handle within this process) may be writing or reading the file until it is closed).

Being clear on the semantics does have an unfortunate effect of tying us to an OS's particular implementation and may require adjustments to make them all agree. For example, file locking is not supported by Web Assembly (its `flock` just invisible succeeds no matter what), that would require special casing to produce an error (not that it would be significant trouble to do so). Thus, if we wish to be specific, we should probably hammer out the details of what we want to be specific about in a discussion.

> For` truncate`, could it just flush automatically so we don't need a comment?

 It could flush automatically, but then it would need a comment to that effect. (It would also, of course, add overhead in the cases where it is unnecessary). Either is fine by me.

tydeu:
@joehendrix 

> I haven't used `flock` before, but the fact that the semantics varies between OSes and file systems is concerning. I noticed this is a common complaint with file locking in some brief google searches.

I don't think is `flock`'s fault (at least in the Web Assembly case). The underlying notion of file locking is well-specified in POSIX (via [`fcntl`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fcntl.html)).  `flock` is just a more convenient -- and, on Linux, more optimized -- interface for file-descriptor-level locks (see [here](https://stackoverflow.com/questions/29611352/what-is-the-difference-between-locking-with-fcntl-and-flock) for a discussion). (If we care a lot about OS generality, we could switch to `fnctl` for total POSIX compatibility -- but I think other parts of the C runtime already presume Linux/BSD-style). The problem here is that the Emscripten transpiler POSIX shim (which is not truly an OS) just doesn't implement file locking in a compliant manner (as reported in [an old issue](https://github.com/emscripten-core/emscripten/issues/11797) over there), `fnctl` is similarly stubbed out (as file system locking doesn't make much sense in a web browser context).

> Could we make the lock file a pid file? The idea would be to create the pid file atomically via a rename and if that fails, then lake would open it to figure out what process created it. If the process that created it is still around, then lake could print an error if the process exists or delete the file and try again if the process doesn't exist.

Yep, this was an idea for the original lock file. The problem is that deleting and recreating it create a race condition (i.e., another process could have done this first and then Lake would end up stealing the lock).

>  [T]he problem then reduces to just making sure the versions were both written by the same step. Perhaps store a hash from/of the olean in the trace?

Yes, but this the problem. Even with a hash the file could be overwritten between the olean check and the olean load by a racing process.

joehendrix:
Ok, I see.  I think I need to switch pseudo code to make sure we are agreeing:

```
def acquirePidFile (pidPath : String) : IO Unit := do
  let tempPath <- createTempFileWithContents currentPid
  if <- doRename tempPath pidPath then
    return () -- pidPath created successfully
  -- pid path exists so rename fails
  let oldPid <- getContents pidPath
  if oldPid.running then
    fail "Process {oldPid} has locked file"
  deleteFile pidPath 
  -- ^ Above is Race: Another lake started at same time and called doRename
  -- between doRename in this block and deleteFile
  acquirePidFile  
```

It's a shame there isn't a way to delete a file using the file descriptor, then we could open the file, read the pid and delete by file descriptor (which would be a no-op if the rename moved the file out from under us.

However, I think since `acquirePidFile` should be fast, one approach is to guard it with it's own lockfile.

```
def withLockFile (lockFile : String) (action : IO Unit) : IO Unit := do
  match <- createNewFile lockFile with
  | None -> sleep 1; withLockFile lockFile action
  | Some(h) ->
    try
      action
    finally
      closeAndDelete h
```

It's clunky, but I think then `withLockFile "lake.lock" (acquirePidFile "lake.pid")` would avoid race conditions at the cost of a lingering "lake.lock" if one managed to kill lake while running acquirePidFile.  There is still a risk of process id recycling, but I think the worst that happens is one gets a spurious messaged that another process owns lake when it doesn't.  Perhaps one could also validate that the oldPid is another lake instance and not the current pid?

```
Even with a hash the file could be overwritten between the olean check and the olean load by a racing process.
```

Are the olean check and olean load done within the same process?  If you open by file descriptor and only overwrite the olean using a rename, then there isn't a way for the contents to be changed between the check and load.  The loaded value would be obsolete, but at least consistent.

tydeu:
@joehendrix 


> If you open by file descriptor and only overwrite the olean using a rename, then there isn't a way for the contents to be changed between the check and load. The loaded value would be obsolete, but at least consistent.

Unfortunately, loading an olean file is a black box operation so there isn't a way open the file, perform the check, and then load from the opened file (and this would be difficult to change because the olean loading does a lot of platform magic to support memory mapping of the files when opening them).

A more general concern I have with the atomic rename is that I am not sure that works as desired on Windows (from the discussion in [this StackOverflow question](https://stackoverflow.com/questions/167414/is-an-atomic-file-rename-with-overwrite-possible-on-windows) it appears it does not). There do seem to be workarounds, but they seem like they will require Windows-specific primitives for managing file information (which seems to working against the goal of a clean OS interface).

Finally, all of this seems like heavy overhead for what is otherwise accomplishable by using the file locking primitives that are generally supported by the platforms Lean is compatibile with (minus Emscripten, but I doubt file system race conditions in Lake are a concern for web users of Lean, since the notion of a file system doesn't really properly exist in the browser and Lake is unlikely to be used there). I understand the desire for a clean Lean interface, but I think  when it comes to bulletproof lock files, this is about as clean as it will get (for reference, the interface naming was modeled after Rust's [fslock](https://docs.rs/fslock/latest/fslock/struct.LockFile.html), which has some wrinkles of its own in the docstrings).

joehendrix:
I may be misreading that but that stack overflow concerns are about the atomicity of metadata rather than the contents.  It's a moot point though if you otherwise can't open oleans by file descriptors.

I don't quite see the connection between the Rust interface and the interface use used.  It does look like Rust Windows and Unix packages implement these lock and try_lock.

I guess in the end this does seem simpler.  If I am reading the [man](https://linux.die.net/man/2/flock) correctly, one potential error could arise from the following:

```
  h <- acquireLock
  fork_long_running_child -- Inherits file descriptor
  random_crash
  unlock h
```

In this case, I think the child process would inherit `h` and the random crash would cause unlock not to be called.  The lock wouldn't be released until the child closes it's copy of `h`.  You just need to make sure the child process closes all file descriptors other than ones it is intentionally supposed to inherit.

I think I'm comfortable with `lock` and `tryLock`.  I'd recommend dropping the exclusive flag for now though and just making it always exclusive.  The non-exclusive locking isn't exercised by your code and thus isn't getting any testing.

tydeu:
@joehendrix 

> In this case, I think the child process would inherit `h` and the random crash would cause unlock not to be called. The lock wouldn't be released until the child closes it's copy of `h`. You just need to make sure the child process closes all file descriptors other than ones it is intentionally supposed to inherit.

I believe you are correct. The good news is that you cannot currently fork processes without defining your own primitive in Lean, so this is not an issue as is (there also may already be other problems with the current IO handle interface outside the new primitives that might be caused by forking and duplicating the file descriptor -- I am not sure if that was a case tested by the existing API).

> I think I'm comfortable with `lock` and `tryLock`. I'd recommend dropping the exclusive flag for now though and just making it always exclusive. The non-exclusive locking isn't exercised by your code and thus isn't getting any testing.

Sounds good. Is `unlock`  fine as well?

semorrison:
> @semorrison I fixed half of the mathlib breakage that is caused by this PR. That last breakage is due to a hard-coded reference to the `build` directory in ProofWidgets. That one will require an update to ProofWidgets itself. So, my understanding of procedure is for that fix to come after this is merged.

Ideally actually updates to packages are all ready to go before we merge the Lean PR, although this is not a rule, just an attempt to minimise work wasted on unanticipated breakages. 

I made https://github.com/leanprover-community/ProofWidgets4/pull/32, which puts `ProofWidgets4` on the toolchain generated for this PR, but breaks. If @tydeu and/or @Vtec234 were able to get that working, please let me know and I can do the rest of configuring the Mathlib test build.

semorrison:
If you're able to rebase this on to `origin/nightly`, future rounds of Mathlib CI should hopefully be much smoother. I've replaced the label `breaks-mathlib` with `awaiting-mathlib`, as we won't get a Mathlib build under the new system until it is rebased.

src/runtime/io.cpp
```diff
@@ -294,6 +295,93 @@ extern "C" LEAN_EXPORT obj_res lean_io_prim_handle_mk(b_obj_arg filename, uint8
...
+        return io_result_mk_error((sstream() << GetLastError()).str());
+    }
+#else
+    if (!flock(fileno(fp), x ? LOCK_EX : LOCK_SH)) {
+        return io_result_mk_ok(box(0));
+    } else {
+        return io_result_mk_error(decode_io_error(errno, nullptr));
+    }
+#endif
+}
```
joehendrix, line 320: I'd suggest lifting the bloc inside the `#ifdef q4_WCE block` into it's own function.

I also have a preference for lifting the `#ifdef LEAN_WINDOWS` outside of `_lock`, `_try_lock` and `_unlock`.  I think that would make it easier to match up the calls to the underlying OS functions across versions.

tydeu:
I split the `lakefile.olean` improvement out into a separate PR that builds on this one (#2842). Hopefully, this will be cleaner.

src/runtime/io.cpp
tydeu: @joehendrix I have implemented these changes (and removed the shared option) over at #2842.

Kha:
I'll remove my review request, unless there's something specific you want me to look at

tydeu:
Following the [Zulip poll](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/RFC.3A.20.60_lake.60.20vs.20.60.2Elake.60/near/400059584), I have decided to follow the majority and make the default Lake directory `.lake`. Furthermore, as this is now a pure-Lake PR, I will likely merge soon after the mathlib incompatibility is resolved, unless there are any objections.  

/cc @semorrison, @Kha, @joehendrix: Since you all commented, it seems proper to notify all of you before I merge in case there is anything else you wish to add.

PR #2750, opened 2023-10-25
fix: ReducibilityHints.lt

digama0:
The implementation of `ReducibilityHints.lt` was previously giving the ordering
```
.abbrev < .regular 1 < .regular 10 < .opaque
``` 
while the kernel is using a three-way comparison function `compare(reducibility_hints, reducibility_hints)` which returns `-1` when the first argument should be unfolded and `1` when the second argument should be unfolded, where the one that should be unfolded is larger in the following ranking:
```
.opaque < .regular 1 < .regular 10 < .abbrev
``` 
Both of these are a bit weird (the `compare` function has reversed interpretation compared to `strcmp`, and the lean function gets the roles of `.abbrev` and `.opaque` mixed up). What is more, the interpretation of the lean ordering is also reversed in `ExprDefEq`, meaning that it actually implements the ordering
```
.opaque < .regular 10 < .regular 1 < .abbrev
``` 
which is almost but not quite the same as the kernel ordering.

My fix is to make the `.lt` function clearly represent "more reducible" with `.abbrev` at the top and `.opaque` at the bottom, and covariant in the definitional height (higher = more reducible). I have chosen not to negate the output values of `compare`, although this would also be a reasonable refactor. Should lean just have a three-way comparison as well?

### DIFF
--- a/src/Lean/Declaration.lean
+++ b/src/Lean/Declaration.lean
@@ -48,12 +48,14 @@ def ReducibilityHints.getHeightEx (h : ReducibilityHints) : UInt32 :=
 
 namespace ReducibilityHints
 
+/-- `a.lt b` means that `b` is "more reducible" than `a`,
+so we will prefer unfolding `b` instead of `a`. -/
 def lt : ReducibilityHints → ReducibilityHints → Bool
-  | .abbrev,     .abbrev     => false
-  | .abbrev,     _           => true
+  | _,           .opaque     => false
+  | .abbrev,     _           => false
+  | .opaque,     _           => true
+  | _,           .abbrev     => true
   | .regular d₁, .regular d₂ => d₁ < d₂
-  | .regular _,  .opaque     => true
-  | _,           _           => false
 
 def isAbbrev : ReducibilityHints → Bool
   | .abbrev => true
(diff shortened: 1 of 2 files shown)


semorrison:
Apologies for the `breaks-mathlib`. Until #2734 hits a nightly release anything on top of that is going to break mathlib. After the next nightly release, touching this PR should be sufficient to make the Mathlib CI come good.

semorrison:
Apparently there is a small breakage in Mathlib (a `dsimp` fixes a universe unification problem)? This seems slightly surprising to me; @digama0, could you please check it out? https://github.com/leanprover-community/mathlib4/pull/8289

semorrison:
@digama0, Leo has merged #3964 which also addresses this. Do we need to keep this PR?

digama0:
Yes, this PR and #3964 are identical up to a global flip of the ordering relation. Although I would like to ask @leodemoura to comment on this choice explicitly: If `a.lt b` are reducibilities then intuitively b should be "more reducible" and a should be "less reducible", such that when deciding between them we would prefer to unfold b. #3964 makes the opposite choice without comment. However it is consistent with the `reducibility_hints::compare` and clearly a smaller diff, so in any case this PR is no longer needed.

PR #2766, opened 2023-10-25
feat: embed and check githash in .olean

Kha:
This is an additional safety net on top of #2749: it protects users that circumvent the build system (e.g. with `lake env`) as well as obviates the need for TOCTOU-like race condition checks in the build system.

The check is activated by `CHECK_OLEAN_VERSION=ON`, which now defaults to `OFF` as the sensible default for local development. When activated, `USE_GITHASH=ON` is also force-enabled for stage 0 in order to make sure that stage 1 can load its own core library.

### DIFF
--- a/src/library/module.cpp
+++ b/src/library/module.cpp
@@ -3,6 +3,8 @@ Copyright (c) 2014-2015 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 
 Authors: Leonardo de Moura, Gabriel Ebner, Sebastian Ullrich
+
+.olean serialization and deserialization.
 */
 #include <unordered_map>
 #include <vector>
@@ -25,6 +27,7 @@ Authors: Leonardo de Moura, Gabriel Ebner, Sebastian Ullrich
 #include "library/constants.h"
 #include "library/time_task.h"
 #include "library/util.h"
+#include "githash.h"
 
 #ifdef LEAN_WINDOWS
 #include <windows.h>
@@ -41,8 +44,22 @@ Authors: Leonardo de Moura, Gabriel Ebner, Sebastian Ullrich
 #endif
 
 namespace lean {
-// manually padded to multiple of word size, see `initialize_module`
-static char const * g_olean_header   = "oleanfile!!!!!!!";
+
+/** On-disk format of a .olean file. */
+struct olean_header {
+    // 5 bytes: magic number
+    char marker[5] = {'o', 'l', 'e', 'a', 'n'};
+    // 1 byte: version, currently always `1`
+    uint8_t version = 1;
+    // 42 bytes: build githash, padded with `\0` to the right
+    char githash[42];
+    // address at which the beginning of the file (including header) is attempted to be mmapped
+    size_t base_addr;
+    // payload, a serialize Lean object graph; `size_t` has same alignment requirements as Lean objects
+    size_t data[];
+};
+// make sure we don't have any padding bytes, which also ensures `data` is properly aligned
+static_assert(sizeof(olean_header) == 5 + 1 + 42 + sizeof(size_t), "olean_header must be packed");
 
 extern "C" LEAN_EXPORT object * lean_save_module_data(b_obj_arg fname, b_obj_arg mod, b_obj_arg mdata, object *) {
     std::string olean_fn(string_cstr(fname));
@@ -73,10 +90,14 @@ extern "C" LEAN_EXPORT object * lean_save_module_data(b_obj_arg fname, b_obj_arg
         // `MapViewOfFileEx` addresses must be aligned to the "memory allocation granularity", which is 64KB.
         base_addr = base_addr & ~((1LL<<16) - 1);
 
-        object_compactor compactor(reinterpret_cast<void *>(base_addr + strlen(g_olean_header) + sizeof(base_addr)));
+        object_compactor compactor(reinterpret_cast<void *>(base_addr + offsetof(olean_header, data)));
         compactor(mdata);
-        out.write(g_olean_header, strlen(g_olean_header));
-        out.write(reinterpret_cast<char *>(&base_addr), sizeof(base_addr));
+
+        // see/sync with file format description above
+        olean_header header = {};
+        header.base_addr = base_addr;
+        strncpy(header.githash, LEAN_GITHASH, sizeof(header.githash));
+        out.write(reinterpret_cast<char *>(&header), sizeof(header));
         out.write(static_cast<char const *>(compactor.data()), compactor.size());
         out.close();
         while (std::rename(olean_tmp_fn.c_str(), olean_fn.c_str()) != 0) {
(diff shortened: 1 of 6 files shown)


src/library/module.cpp
```diff
@@ -73,9 +85,13 @@ extern "C" LEAN_EXPORT object * lean_save_module_data(b_obj_arg fname, b_obj_arg
         // `MapViewOfFileEx` addresses must be aligned to the "memory allocation granularity", which is 64KB.
         base_addr = base_addr & ~((1LL<<16) - 1);
 
-        object_compactor compactor(reinterpret_cast<void *>(base_addr + strlen(g_olean_header) + sizeof(base_addr)));
+        object_compactor compactor(reinterpret_cast<void *>(base_addr + strlen(g_olean_header) + g_githash_header_size + sizeof(base_addr)));
         compactor(mdata);
         out.write(g_olean_header, strlen(g_olean_header));
+        out.write(LEAN_GITHASH, std::min(g_githash_header_size, strlen(LEAN_GITHASH)));
+        if (g_githash_header_size > strlen(LEAN_GITHASH)) {
+            out << std::string(g_githash_header_size - strlen(LEAN_GITHASH), '0');
```
digama0, line 93: I would prefer if this was padded with null bytes or spaces, so that you can tell the difference between a missing githash and a real one, or hashes of different lengths.

semorrison:
Options for a successful PR run: 
* rebase this PR onto d126c099f4d15282ed5051c8c440107cf514da2b (the commit underlying [`nightly-2023-10-25`](https://github.com/leanprover/lean4-nightly/releases/tag/nightly-2023-10-25)), which should give an immediate run.
* wait until `nightly-2023-10-26` lands and I have fixed Std+Mathlib for it (approximately +10 hours from now), and then push an empty commit (or anything) to this branch.

digama0:
Note: this will break `leantar` / `lake exe cache`, keep me appraised when it hits mathlib.

digama0:
While we're changing the olean format, I also have a request: add a version field, pinned to `1` for the time being. We could make it part of the oleanfile header, say using 12 bytes for `oleanfile!!!` and then a 4 byte version field.

I don't want to start the discussion of ABI versioning yet, but it would be nice if we had a little foresight...

src/library/module.cpp
digama0: Also, why 32 bytes? Git hashes are 40 bytes, or 20 if you use a sensible encoding.

digama0:
I have a patch for leantar up at https://github.com/digama0/leangz/tree/olean_v1 . It is backward compatible with earlier versions, so I will make a release for it as soon as the new olean format is finalized and this PR is merged.

digama0:
I'll repeat my request for adding a version field. `leantar` needs this to detect any ABI breaking change to core data structures, in particular `Expr`, for which there have been [two](https://github.com/leanprover/lean4/pull/2747) [PRs](https://github.com/leanprover/lean4/pull/2805) in the recent past (both from me and not yet accepted, but still) to be able to detect changes. `leantar` has to work with multiple versions of lean simultaneously, because users can jump between different versions of the toolchain when switching branches, so it is helpful for any changes to the olean format to be accompanied by a version bump. There are generally very few of these, but this PR and the two mentioned ones would involve an olean version bump. Git hashes are not sufficient, they allow detecting that the version is wrong but not whether it is an old version or a new version, and neither git hashes nor toolchain names are easily placed in a linear order.

Kha:
I added a version field for the .olean header format itself but I don't see how we could realistically version the payload format itself given that any change to a user environment extension can be an ABI change. And even if we restricted ourselves to core, it seems infeasible to me without some automation double-checking the contract.

> leantar has to work with multiple versions of lean simultaneously

It doesn't *have* to, `cache` could refer to and download an exact leantar version instead of a lower bound. It would be a simpler solution than olean versioning by orders of magnitude.

digama0:
> I added a version field for the .olean header format itself but I don't see how we could realistically version the payload format itself given that any change to a user environment extension can be an ABI change. And even if we restricted ourselves to core, it seems infeasible to me without some automation double-checking the contract.

Thanks. A "good enough" solution is fine for the leantar usecase, as if it gets the format wrong the worst that happens is the compression ratio goes way down. The main idea here is that when we anticipate breakage due to use cases in the wild we have a number we can bump to fix them. For lean itself it makes sense that you would want to use the exact githash.

> It doesn't have to, cache could refer to and download an exact leantar version instead of a lower bound. It would be a simpler solution than olean versioning by orders of magnitude.

The problem here is that one would have to know which leantar version to use given an olean, and unless it is being distributed with the toolchain and being released with every commit I don't how to do that. The number of olean version changes that are relevant to leantar that have happened in the last year can be counted on one hand, so it is perfectly feasible to release new leantar versions when needed and have complete backward compatibility to every change made to the format so far. Maybe this won't be the case forever but the changes so far have been quite tame and predictable.

I have some ideas for how to do complete olean versioning using "schema generation" from the lean environment; a lot of this is already implemented in oleandump, which reads lean declarations to determine how to deserialize oleans. That would make it possible to detect and automatically bump the version whenever there is an ABI breaking change. But I'm not proposing that at this time, only a version that is manually bumped to avoid the worst fallout of ABI breaks.

Kha:
> The problem here is that one would have to know which leantar version to use given an olean, and unless it is being distributed with the toolchain and being released with every commit I don't how to do that

Why does my suggestion of hardcoding the leantar version in the `cache` source code not work for that? As you said it would only have to be adjusted in a rare few `lean-toolchain` bumps in mathlib.

digama0:
> Why does my suggestion of hardcoding the leantar version in the cache source code not work for that? As you said it would only have to be adjusted in a rare few lean-toolchain bumps in mathlib.

Oh I see, yes that would work, you are saying that the "olean version" would be tracked in the cache source code instead of the lean core source code. This can be done without needing to pin leantar versions too, cache could pass an olean version as a command line argument.

I would go for all of the above: a lean embedded olean version number, a cache version command line arg, and leantar versioned releases. The reason is because these each have slightly different development processes and this gives us flexibility to detect and correct issues caused by any of these places.

digama0 (approved):

Kha:
(ready for review but I want to do one final test round before merge)

digama0:
Updated https://github.com/digama0/leangz/tree/olean_v1 , I will make a release once this PR is merged. Fortunately or unfortunately, old leantar versions will detect that these oleans have a funny format and will instead use a generic (and much less efficient) compression format, so if we are not on the ball bumping the leantar version at the right time we will simply have some large cache files. The new leantar is backward compatible though, so I will try to get it merged in advance of this change landing in mathlib, and then hopefully it will not be an issue.

semorrison:
The `breaks-mathlib` tag is (finally) telling us something real: the release oleans from ProofWidgets are being rejected, but othewise Mathlib is passing. I think for this PR we should interpret this as success. :-)

I can coordinate making a pre-release of ProofWidgets on this toolchain if we really want an end-to-end test.

semorrison:
I've made a [pre-release](https://github.com/leanprover-community/ProofWidgets4/releases/tag/v0.0.23-pre) of ProofWidgets, and update Mathlib's `lean-pr-testing-2766` branch to use it.

semorrison:
Unfortunately I can't get this to work. Mathlib's `lean-pr-testing-2766` branch is rejecting the `.olean`s from ProofWidgets' `v0.0.23-pre` release, which was created with this toolchain:

```
lake env lean test/CommDiag.lean
test/CommDiag.lean:1:0: error: failed to read file './lake-packages/proofwidgets/build/lib/ProofWidgets/Component/Panel/GoalTypePanel.olean', invalid header
```

Kha:
@semorrison You have to run `lake update proofwidgets` after changing the lakefile

semorrison:
Mathlib used parts of ProofWidgets in tests, that were not being built by `lake build`. Hence this had the wrong hash, and `lake` was not being given the chance to rebuild them. I've added `lake build ProofWidgets` in the CI test step for Mathlib, and cherry-picked that to `lean-pr-testing-2766`. Hopefully this will go green again shortly!

Kha:
@digama0 It's merged!

Issue #2701, opened 2023-10-16
PR releases release (unwanted) merge state


nomeata:
This is a follow up to https://github.com/leanprover/lean4/issues/2441

The PR releases are great! But I think they don’t release what they should.

Today, I pushed https://github.com/leanprover/lean4/pull/2577/commits/cf120acb2c43f3922ac2a2c9115b3353269060ff to https://github.com/leanprover/lean4/pull/2577/. This commit is behind the latest `master` (intentionally; I want the delta to `v4.2.0-rc2` small for sensible benchmarks). 

The release at <https://github.com/leanprover/lean4-pr-releases/releases/tag/pr-release-2577> however isn't built from `cf120acb2c43f3922ac2a2c9115b3353269060ff`, but rather from a synthetic merge commit with lean’s latest `master`, namely <https://github.com/leanprover/lean4-pr-releases/commit/96a003904b50ebd7e08c6e2228b202ec57cd6668> which does not exist on any branch, and has 

> Merge https://github.com/leanprover/lean4-pr-releases/commit/cf120acb2c43f3922ac2a2c9115b3353269060ff into https://github.com/leanprover/lean4-pr-releases/commit/44416624904d389f6fb2cddf018594d34624b2bf

as the commit message.

Looking at <https://github.com/semorrison/lean4/blob/master/.github/workflows/pr-release.yml> it seems the issue isn’t actually with this job; this job simply uses the commit that was tested by CI.  The problem is thus with <https://github.com/semorrison/lean4/blob/master/.github/workflows/ci.yml> using `actions/checkout@v3`:

That action has the very annoying default to not check out the tested branch as it is, but rather as it would look like if it were merged into `master` at that particular point (see https://github.com/actions/checkout#Checkout-pull-request-HEAD-commit-instead-of-merge-commit for docs). This is of course going to be very confusing: The CI results and the toolchain obtained this way describe some rather arbitrary state that depends on both the branch and `master` at the time it runs, and will be something else at some later time, even if the branch hasn’t changed. (Note that Github does _not_ update the CI status when `master` evolves, so this logic is surprisingly broken for a default.)

If others agree that this is a problem then my suggested way to fixing this would be to make all CI jobs test the head commit of the feature branch, and _not_ some merge state, and additionally enable “strict merges” so that only PRs that are up-to-date with regard to master can be merged.

nomeata:
I guess I should @semorrison since it is related to #2441.

semorrison:
[...]

PR #2767, opened 2023-10-26
chore: update 'nightly' branch to track nightly releases

kim-em:
This PR updates our CI to automatically update the `nightly` branch to follow the most recent commit used to generate a nightly release.

This opens the possibility of encouraging contributors to start their new features branches from `nightly` rather than `master`. This should drastically reduce the number of Mathlib CI failures we see because of unrelated changes.

(We would also need to adjust how github CI works for this to pay off: by default it does a "virtual rebase" onto master before starting CI. See https://github.com/leanprover/lean4/issues/2701)


semorrison:
Note that the Mathlib failure here is exactly because of github CI doing the virtual rebase onto `master`.

PR #2769, opened 2023-10-26
chore: run CI against the head of the branch, not a virtual merge with master

### LINKED ISSUE #2701: PR releases release (unwanted) merge state

kim-em:
Fixes #2701

The default behaviour in Github CI is not to run CI against the head of the PR branch, but against a virtual merge with `master`. See discussion at #2701 for details.

This PR changes that behaviour to actually run CI (and hence also to build toolchains and `lean-pr-testing-NNNN` Mathlib branches) against the PR itself.

This should help reduce the number of CI failures against Mathlib due to unrelated changes, particularly if authors start making feature branches by branching off the `nightly` branch (as introduced in #2767), rather than `master`. This is much more likely to have a successful Mathlib build!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -197,6 +197,8 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: true
+          # the default is to use a virtual merge commit between the PR and master: just use the PR
+          ref: ${{ github.event.pull_request.head.sha }}
       - name: Install Nix
         uses: cachix/install-nix-action@v18
         with:
--- a/.github/workflows/nix-ci.yml
+++ b/.github/workflows/nix-ci.yml
@@ -34,6 +34,9 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
+          # the default is to use a virtual merge commit between the PR and master: just use the PR
+          ref: ${{ github.event.pull_request.head.sha }}
       - name: Install Nix
         uses: cachix/install-nix-action@v18
         with:


semorrison:
And look at that, magically succeeding on Mathlib, unlike #2767, made at about the same time!

Kha:
I assume the more proper way to eventually test-build the PR merged/rebased onto master would be a future merge queue? LGTM then.

semorrison:
Yes, the merge queue is starting to look essential. I will try to set that up next week.

nomeata:
Note that this setting is somewhat dangerous without the “require branches to be up to date before merging”, because now one can merge a green PR without noticing that the merged state is bad, and would lead to a broken master.

With that setting merging requires updating first, which is ok for slow throughput repos.

Later, merge automation like mergify or GitHub's merge queue helps a lot.

Kha:
This was already true before because CI is not restarted when master is updated, no? That is, the window of CI/master divergence merely widened from "most recent PR push to now" to "most recent PR rebase to now".

nomeata:
Indeed, its just more likely now. But yes, GitHub's default is broken in surprisingly many ways :-)

PR #2770, opened 2023-10-26
feat: Sensible defaults arguments for math project

PatrickMassot:
Those are copy-pasted from the Mathlib lakefile. See discussion [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Project.20configuration.20interface/near/398712899).


tydeu:
@PatrickMassot Unfortunately, the quotes need escaping because Lean does not have triple quote multiline strings like Python. :(

PR #2771, opened 2023-10-26
doc: Improve docstrings around Array.mk,.data,.toList

nomeata:
following a discussion at
<https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Understanding.20the.20docstring.20for.20docs.23Array.2Edata/near/398705430>

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -2507,7 +2507,7 @@ as they are used "linearly" all updates will be performed destructively on the
 array, so it has comparable performance to mutable arrays in imperative
 programming languages.
 
-From the point of proofs `Array α` is just a wrapper around `List α`.
+From the point of view of proofs `Array α` is just a wrapper around `List α`.
 -/
 structure Array (α : Type u) where
   /--


nomeata:
On Zulip there was some disagreement which function should be recommended to be used, without conclusion. To not have this PR (which makes no such recommendation either way) sitting around for much longer, I’ll merge it if noone complains. Further improvements are still possible of course.

src/Init/Prelude.lean
```diff
@@ -2506,13 +2506,22 @@ is not observable from lean code. Arrays perform best when unshared; as long
 as they are used "linearly" all updates will be performed destructively on the
 array, so it has comparable performance to mutable arrays in imperative
 programming languages.
+
+From the point of proofs `Array α` is just a wrapper around `List α`.
```
digama0, line 2510:
```suggestion
From the point of view of proofs `Array α` is just a wrapper around `List α`.
```

digama0:
Regarding the main question, my opinion is [as mentioned on zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Understanding.20the.20docstring.20for.20docs.23Array.2Edata/near/398772743) - we should rename `data` to `toList` and `toList` to `toListImpl` - but as a pure docs PR the docstrings look fine to me.

PR #2774, opened 2023-10-26
feat: Array.mem: Avoid DecidableEq, set up decreasing_trivial

nomeata:
The notation `a ∈ as` for Arrays was previously only defined with
`DecidableEq` on the elements, for (apparently) no good reason. This
drops this requirements (by using `a ∈ as.data`), and simplifies a bunch
of proofs by simply lifting the corresponding proof from lists.

Also, `sizeOf_lt_of_mem` was defined, but not set up to be picked up by
`decreasing_trivial` in the same way that the corresponding List lemma
was set up, so this adds the tactic setup.

The definition for `a ∈ as` is intentionally not defeq to `a ∈ as.data`
so that the termination tactics for Arrays don’t spuriously apply when
recursing through lists.

### DIFF
--- a/src/Init/Data/Array/Mem.lean
+++ b/src/Init/Data/Array/Mem.lean
@@ -1,7 +1,7 @@
 /-
 Copyright (c) 2022 Microsoft Corporation. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Leonardo de Moura
+Authors: Leonardo de Moura, Joachim Breitner
 -/
 prelude
 import Init.Data.Array.Basic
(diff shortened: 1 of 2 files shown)


nomeata:
awaiting-review

nomeata:
I am a bit confused how this change can affect tests like 
```
inductive Term where
  | app (f : String) (args : List Term)

def printFns : Term → IO Unit
  | Term.app f args => do
     IO.println f
     for h : arg in args do
       have : sizeOf arg < 1 + sizeOf f + sizeOf args := Nat.lt_trans (List.sizeOf_lt_of_mem h) (by simp_arith)
       printFns arg
```
where there is no `Array` around. Maybe because `apply Nat.lt_trans (Array.sizeOf_lt_of_mem ?h)` can fire due to unwanted defeqs even if the assumption is only talking about lists?

digama0:
You should make it a `structure` instead of an `inductive`, assuming `def` doesn't work (it should work though).

nomeata:
Will do. Is there a difference between a one field structure and a one field one constructor inductive? Ive seen both patterns so far.

digama0:
Structures have various additional niceties, I don't see why you would ever not use them unless you don't meet the restrictions. Do you have an example?

nomeata:
I've seen it mostly with rose trees recently: https://github.com/leanprover/lean4/blob/6c5f79c0df01e9cd74ac4ee411ec392a5ac1b5d3/src/Lean/Data/PrefixTree.lean#L11 (and similar data structures)

Maybe the benefit of an inductive is that you don't have to think of a name for the projection :-)

nomeata:
Fixed in 9a7d35063c, PTAL.

leodemoura (approved):

nomeata:
Squashed.

(And because Github is very strange in that squashing is just one click away when merging, but basically impossible without a laptop and a command line for the PR authors, in the process I wrote half a Github App that squashes PRs… :-)

digama0:
> Maybe the benefit of an inductive is that you don't have to think of a name for the projection :-)

My suggestion would be to just use `val` for one-field structures, rather than coming up with bespoke names that have to be remembered.

nomeata:
> My suggestion would be to just use val for one-field structures, rather than coming up with bespoke names that have to be remembered.

done

PR #2776, opened 2023-10-27
fix: dsimp missing consumeMData when closing goals by rfl

### LINKED ISSUE #2514: bug: `dsimp` does not close reflexive equality goals if they are wrapped in metadata
### Prerequisites

### Description

`dsimp` has a feature where it can close goals that can be solved by `Eq.refl` if the goal is obviously an equality and the LHS and RHS are defeq (up to reducible transparency). For example
```lean
example : ∀ (x : α × α), x = id x := by
  intro p
  -- `⊢ (x, y) = id (x, y)`
  dsimp
  -- closed goal
```
However, sometimes tactics can insert metadata around this equality and then `dsimp` no longer tries to use `Eq.refl`.
```lean
example : ∀ (x : α × α), x = id x := by
  intro (x, y)
  -- `⊢ (x, y) = id (x, y)`
  dsimp
  -- not done yet, goal `⊢ (x, y) = (x, y)`
  rfl
```
One can see the metadata (`noImplicitLambda`) by setting `set_option pp.raw true`.

(This was reported by Heather Macbeth [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/intro.20.28x.2C.20y.29/near/389257895))

Here is another test that demonstrates the issue. Removing the `have` line removes the need for the `rfl`. The `have` line can also be replaced by `refine no_implicit_lambda% ?_` to more explicitly introduce metadata.
```lean
example {α : Type _} (n : α) : n = n := by
  have := 0
  dsimp
  rfl
```

In `Lean.Meta.dsimpGoal` one can see that there is `targetNew.eq?` instead of `targetNew.consumeMData.eq?`, which is likely the cause. Other pattern matches in that function refer to `targetNew.consumeMData` rather than `targetNew`.

### Versions

leanprover/lean4:v4.0.0-rc4

kmill:
Fixes #2514

### DIFF
--- a/src/Lean/Meta/Tactic/Simp/Main.lean
+++ b/src/Lean/Meta/Tactic/Simp/Main.lean
@@ -1123,7 +1123,7 @@ def dsimpGoal (mvarId : MVarId) (ctx : Simp.Context) (simplifyTarget : Bool := t
       if targetNew.consumeMData.isConstOf ``True then
         mvarIdNew.assign (mkConst ``True.intro)
         return (none, usedSimps)
-      if let some (_, lhs, rhs) := targetNew.eq? then
+      if let some (_, lhs, rhs) := targetNew.consumeMData.eq? then
         if (← withReducible <| isDefEq lhs rhs) then
           mvarIdNew.assign (← mkEqRefl lhs)
           return (none, usedSimps)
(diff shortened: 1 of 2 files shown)


kmill:
awaiting-review

leodemoura (approved):

PR #2777, opened 2023-10-27
doc: expand docstring for `intros`

kmill:
The docstring for `intros` did not explain the difference between the zero-argument and the one-or-more-argument cases.

### DIFF
--- a/src/Init/Tactics.lean
+++ b/src/Init/Tactics.lean
@@ -39,8 +39,75 @@ be a `let` or function type.
 syntax (name := intro) "intro" notFollowedBy("|") (ppSpace colGt term:max)* : tactic
 
 /--
-`intros x...` behaves like `intro x...`, but then keeps introducing (anonymous)
-hypotheses until goal is not of a function type.
+Introduces zero or more hypotheses, optionally naming them.
+
+- `intros` is equivalent to repeatedly applying `intro`
+  until the goal is not an obvious candidate for `intro`, which is to say
+  that so long as the goal is a `let` or a pi type (e.g. an implication, function, or universal quantifier),
+  the `intros` tactic will introduce an anonymous hypothesis.
+  This tactic does not unfold definitions.
+
+- `intros x y ...` is equivalent to `intro x y ...`,
+  introducing hypotheses for each supplied argument and unfolding definitions as necessary.
+  Each argument can be either an identifier or a `_`.
+  An identifier indicates a name to use for the corresponding introduced hypothesis,
+  and a `_` indicates that the hypotheses should be introduced anonymously.
+
+## Examples
+
+Basic properties:
+```lean
+def AllEven (f : Nat → Nat) := ∀ n, f n % 2 = 0
+
+-- Introduces the two obvious hypotheses automatically
+example : ∀ (f : Nat → Nat), AllEven f → AllEven (fun k => f (k + 1)) := by
+  intros
+  /- Tactic state
+     f✝ : Nat → Nat
+     a✝ : AllEven f✝
+     ⊢ AllEven fun k => f✝ (k + 1) -/
+  sorry
+
+-- Introduces exactly two hypotheses, naming only the first
+example : ∀ (f : Nat → Nat), AllEven f → AllEven (fun k => f (k + 1)) := by
+  intros g _
+  /- Tactic state
+     g : Nat → Nat
+     a✝ : AllEven g
+     ⊢ AllEven fun k => g (k + 1) -/
+  sorry
+
+-- Introduces exactly three hypotheses, which requires unfolding `AllEven`
+example : ∀ (f : Nat → Nat), AllEven f → AllEven (fun k => f (k + 1)) := by
+  intros f h n
+  /- Tactic state
+     f : Nat → Nat
+     h : AllEven f
+     n : Nat
+     ⊢ (fun k => f (k + 1)) n % 2 = 0 -/
+  apply h
+```
+
+Implications:
+```lean
+example (p q : Prop) : p → q → p := by
+  intros
+  /- Tactic state
+     a✝¹ : p
+     a✝ : q
+     ⊢ p      -/
+  assumption
+```
+
+Let bindings:
+```lean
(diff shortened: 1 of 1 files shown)


src/Init/Tactics.lean
```diff
@@ -39,8 +39,10 @@ be a `let` or function type.
 syntax (name := intro) "intro" notFollowedBy("|") (ppSpace colGt term:max)* : tactic
 
 /--
-`intros x...` behaves like `intro x...`, but then keeps introducing (anonymous)
-hypotheses until goal is not of a function type.
+Introduces zero or more hypotheses, optionally naming them.
+- `intros` is equivalent to repeatedly applying `intro` so long as the goal is of a function type.
```
faenuccio, line 43:
```suggestion
- `intros` is equivalent to repeatedly applying `intro`, introducing anonymous variables so long as the goal is of a function type.
```

leodemoura:
@semorrison Should we ask people to submit PRs on top of current mathlib lean-toolchain (currently: leanprover/lean4:v4.2.0-rc4)? This PR breaks mathlib because it is on top of `main`. What do you think?

semorrison:
> @semorrison Should we ask people to submit PRs on top of current mathlib lean-toolchain (currently: leanprover/lean4:v4.2.0-rc4)? This PR breaks mathlib because it is on top of `main`. What do you think?

In fact, this PR is on top of `nightly` (as I'd asked @kmill to do), which was what I'd hoped would be sufficient. Then the Mathlib testing happens on a branch coming off the latest `nightly-testing`.

The problem with this approach is that `nightly-testing` is not reliable enough, particularly if I am away over the weekend!

As I've just fixed up `nightly-testing`, and Kyle has just force pushed to this branch, it is not impossible that Mathlib CI will come good now. But it is clearly not reliable enough!

I've just written up an RFC at https://github.com/leanprover/lean4/issues/2785 explaining the current setup, why it fails, and a proposal for further improvement.

semorrison:
Ah... This PR is based off a quite old commit on the `nightly` branch. I've rebased now.

src/Init/Tactics.lean
```diff
@@ -39,8 +39,11 @@ be a `let` or function type.
 syntax (name := intro) "intro" notFollowedBy("|") (ppSpace colGt term:max)* : tactic
 
 /--
-`intros x...` behaves like `intro x...`, but then keeps introducing (anonymous)
-hypotheses until goal is not of a function type.
+Introduces zero or more hypotheses, optionally naming them.
+- `intros` is equivalent to repeatedly applying `intro`,
+  introducing anonymous variables so long as the goal is of a function type.
```
david-christiansen, line 44: Will all readers be comfortable thinking of implication as a function type? `intros` is one of the first tactics that one learns, after all.

```suggestion
  introducing anonymous variables so long as the goal is an implication or a function.
```

faenuccio: Nice point!

src/Init/Tactics.lean
```diff
@@ -39,8 +39,11 @@ be a `let` or function type.
...
+- `intros x y ...` is equivalent to `intro x y ...`.
+  Each argument can be either an identifier or a `_`.
```
david-christiansen, line 46: Likewise, thinking about people who haven't learned `intro` yet:
```suggestion
  Each argument can be either an identifier or a `_`. Identifiers are used to name the corresponding hypotheses and each use of `_` results in an anonymous hypothesis.
```

src/Init/Tactics.lean
kmill: I think we want to be careful that we don't accidentally suggest that these are different concepts, but it's a good point that we should point this out. I've made an attempt at wording something, and I've also added examples to illustrate the scope of this tactic.

src/Init/Tactics.lean
```diff
@@ -39,8 +39,74 @@ be a `let` or function type.
 syntax (name := intro) "intro" notFollowedBy("|") (ppSpace colGt term:max)* : tactic
 
 /--
-`intros x...` behaves like `intro x...`, but then keeps introducing (anonymous)
-hypotheses until goal is not of a function type.
+Introduces zero or more hypotheses, optionally naming them.
+
+- `intros` is equivalent to repeatedly applying `intro`
+  until the goal is not an obvious candidate for `intro`.
```
faenuccio, line 45:
```suggestion
  until the goal is not a a function type (this includes implications, since these are regarded as functions).
```

src/Init/Tactics.lean
```diff
@@ -39,8 +39,74 @@ be a `let` or function type.
...
+  That is, so long as the goal is a `let` or a pi type (e.g. an implication, function, or universal quantifier),
```
faenuccio, line 46: What do you mean by a `let` here?

kmill: I mean a `let`, like in `let x := 1; x = x`. There's an example in the examples section.

src/Init/Tactics.lean
kmill: Do you want this redundancy with the "That is" line that follows?

faenuccio: Well, I somehow find it clearer in this version that with the "That is" line that follows: I tend to think that someone who is lost in understanding the expression "function type" (or in seeing why an implication is a term of a function type) would not really be helped by reading `let` or `pi type`. Perhaps we can ask some third opinion?

src/Init/Tactics.lean
faenuccio: Well, I must say that the example section is very well written and clear, so I have no further comments/suggestions!

src/Init/Tactics.lean
kmill: Ok, I've merged the sentences together. Was the "e.g." not enough to explain what a pi type was?

faenuccio: I like it this way! Thanks.

semorrison:
(I'm aware this PR can't possibly really break Mathlib, but as I'm trying to test the new CI update I'm going to include this on my list of PRs to get Mathlib combined CI working.)

PR #2778, opened 2023-10-27
feat: add left/right actions to term tree coercion elaborator and make `^` a right action

### LINKED ISSUE #2220: elaboration of `^` shouldn't try to put the arguments in the same type

kmill:
Implements RFC #2854 and fixes #2220

* Adds `leftact%` and `rightact%` syntax to the expression tree elaborator. These are like `binop%`, but instead of being for homogeneous binary operations, these are for *actions*, which are binary operations where one operand "acts" on the other. Left action tend to have the type `α → β → β` and right actions tend to have the type `α → β → α`.

  The difference in elaboration is that `leftact%` notations only propagate types through the right-hand argument and `rightact%` notations only propagate types through the left-hand argument.

* Changes the `^` notation to use `rightact%` because exponentiation is an action of the exponent upon the base. This is supported by the `Pow` class already having the type of a right action -- often there isn't a completely homogeneous exponentiation operation.

* Adds `NatPow` and `HomogeneousPow` classes with default `Pow` instances. These are here to allow. types to subscribe to a particular defaulting behavior. Types that prefer the exponent to be a `Nat` can implement `NatPow`, and types that prefer a completely homogeneous operation can implement `HomogeneousPow`. For example, `Float` implements `HomogeneousPow`. The `NatPow` behavior is widely expected in Mathlib, and it can later be used for its [Pow instance for monoids](https://github.com/leanprover-community/mathlib4/blob/77ccc30d9f983d3c9bfbc6a795623d404227f4e7//Mathlib/Algebra/Group/Defs.lean#L617-L618). Without `HomogeneousPow`, sensible expressions like `(2.2 ^ 2.2 : Float)` would not elaborate.

### DIFF
--- a/src/Lean/Elab/BuiltinNotation.lean
+++ b/src/Lean/Elab/BuiltinNotation.lean
@@ -252,8 +252,18 @@ def elabCDotFunctionAlias? (stx : Term) : TermElabM (Option Expr) := do
       try Term.resolveId? f catch _ => return none
     else
       return none
-  | `(fun $binders* => binop% $f $a $b) =>
+  | `(fun $binders* => binop% $f $a $b)
+  | `(fun $binders* => binop_lazy% $f $a $b)
+  | `(fun $binders* => leftact% $f $a $b)
+  | `(fun $binders* => rightact% $f $a $b)
+  | `(fun $binders* => binrel% $f $a $b)
+  | `(fun $binders* => binrel_no_prop% $f $a $b) =>
     if binders == #[a, b] then
+      try Term.resolveId? f catch _ => return none
+    else
+      return none
+  | `(fun $binders* => unop% $f $a) =>
+    if binders == #[a] then
       try Term.resolveId? f catch _ => return none
     else
       return none
--- a/tests/lean/cdotAtSimpArg.lean
+++ b/tests/lean/cdotAtSimpArg.lean
@@ -1,5 +1,9 @@
+/-! Test `·` being able to refer to constants in `simp` -/
+
 example : ¬ true = false := by
   simp [(¬ ·)]
+
+/-! Test `binop%` -/
 
 example (h : y = 0) : x + y = x := by
   simp [(.+.)] -- Expands `HAdd.hAdd
@@ -14,3 +18,8 @@ example (h : y = 0) : x + y = x := by
   simp [Add.add]
   simp [h, Nat.add]
   done
+
+/-! Test `binop%` variant `rightact%` as well -/
+
+example (x y : Nat) : x ^ y = y ^ x := by
+  simp only [.^.]
--- a/tests/lean/cdotAtSimpArg.lean.expected.out
+++ b/tests/lean/cdotAtSimpArg.lean.expected.out
@@ -4,3 +4,6 @@ h : y = 0
 y x : Nat
 h : y = 0
 ⊢ Add.add x y = x
+cdotAtSimpArg.lean:24:39-25:17: error: unsolved goals
+x y : Nat
+⊢ Pow.pow x y = Pow.pow y x


kmill:
(I've based this on 4.2.0-rc4 for now to make it easier to fix mathlib. Once I've made the fixes, I'll put it back on master, since that seems like what I'm supposed to be doing.)

kmill:
@leodemoura I found a way to prevent a couple of regressions. Now if `x : Nat` then `x ^ 2` elaborates with `2 : Nat` (and in `∃ k, n = 2 ^ k` then `k` elaborates as a `Nat`). We also have `(2 ^ 64 : Float)` elaborating with `64 : Float`.

The trick is to create two typeclasses that provide default instances for `Pow`. The first, `NatPow`, is for the case where a type wants the exponent to be a `Nat`, and the second, `RPow`, is for the case where the type wants `Pow` to be completely homogeneous. We seem to need separate typeclasses like this so that the `default_instance` mechanism doesn't overzealously specialize types (for example, a `Pow Float Float` default instance would cause `#check 2 ^ 2` to yield `2 ^ 2 : Float`).

Without `RPow`, I'd experimented with giving `Float` a `NatPow` instance, but an unfortunate regression was that `x ^ 2.2` would lead to a "failed to synthesize instance OfScientific Nat" error. I felt that needing to write `x ^ (2.2 : Float)` would be unacceptable to anyone using Lean for numerical computations.

I chose the name `RPow` just for sake of having a name -- it means "reflexive pow" -- and I'm happy to change it.

Does this seem like a sensible design?

PatrickMassot:
Note that Mathlib uses `rpow` to mean "real power", so the proposed name "reflexive power" would create some confusion.

ericrbg:
Will this cause similar issues with trying to find `Neg Nat` if we want to write `(z : C) ^ (-1)`?

kmill:
@ericrbg Yes, unfortunately. You have to write `(z : C) ^ (-1 : Int)` to get that to elaborate. It seems like it's not possible with this design to be able to elaborate `(z : C) ^ (-1)` with an `Int` power while simultaneously elaborating `(z : C) ^ 2` with a `Nat` power -- and I spent some time trying to think of how to get it to work before settling on this one. If you use the default instance mechanism to get the first to work, then the second will come out with an `Int` power.

ericrbg:
Is it not possible to have different "priority" default instances, e.g. it can try the `Nat` one first and then if that fails the `Int` one?

kmill:
Yes, but no.
```lean
class F (α : Type) where
  f : α → Bool

@[default_instance high]
instance : F Nat where
  f _ := true

@[default_instance low]
instance : F Int where
  f _ := false

#eval F.f 2
-- true
#eval F.f (-2)
/-
failed to synthesize instance
  Neg Nat
-/
#eval F.f (-2 : Int)
-- false
```

semorrison:
@kmill, when you want to do Mathlib testing, could you:
* rebase this PR onto `origin/nightly-testing`
* make the Mathlib fixes on the `lean-pr-testing-2778` branch of Mathlib?

src/Init/Prelude.lean
```diff
@@ -1297,11 +1297,38 @@ class Mod (α : Type u) where
...
+Types can choose to subscribe to particular defaulting behavior by providing
+an instance to either `NatPow` or `HomogeneousPow`:
+- `NatPow` is for types whose exponents is preferentially a `Nat`.
+- `HomogeneousPow` is for types whose base and exponent are preferentially the same.
 -/
 class Pow (α : Type u) (β : Type v) where
   /-- `a ^ b` computes `a` to the power of `b`. See `HPow`. -/
   pow : α → β → α
 
+/-- The homogenous version of `Pow` where the exponent is a `Nat`.
```
semorrison, line 1310:
```suggestion
/-- The heterogenous version of `Pow` where the exponent is a `Nat`.
```

src/Lean/Elab/Extra.lean
```diff
@@ -396,22 +435,24 @@ end
...
-  Elaboration functionf for `binrel%` and `binrel_no_prop%` notations.
+  Elaboration functions for `binrel%` and `binrel_no_prop%` notations.
   We use the infrastructure for `binop%` to make sure we propagate information between the left and right hand sides
   of a binary relation.
 
-  Recall that the `binrel_no_prop%` notation is used for relations such as `==` which do not support `Prop`, but
-  we still want to be able to write `(5 > 2) == (2 > 1)`.
+  - `binrel% R x y` elaborates `R x y` using the `binop%/...` expression trees in both `x` and `y`.
+    It is similar to how `binop% R x y` elaborates but with a significant difference:
+    it does not use the expected type when computing the types of the operads.
```
semorrison, line 451:
```suggestion
    it does not use the expected type when computing the types of the operands.
```

src/Init/Prelude.lean
semorrison: Kyle has explained the naming scheme; I'm happy as is.

PR #2781, opened 2023-10-29
perf: `infer_app` with `infer_only := false` (WIP)

leodemoura:


leodemoura:
!bench

semorrison:
@leodemoura here is the comparison between Mathlib runs: 
http://speed.lean-fro.org/mathlib4/compare/e7b27246-a3e6-496a-b552-ff4b45c7236e/to/e7b27246-a3e6-496a-b552-ff4b45c7236e?hash1=1ca67b1a4b326c4224bb25a792b22c304c83b597&hash2=bb1b2648e3674b537c07efac0d4e66162fb7dab5

Unfortunately it doesn't look good: a 2.4% wallclock regression.

leodemoura:
> Unfortunately it doesn't look good: a 2.4% wallclock regression.

I took a look at the results. This commit affects "type checking" time, and it seems it went down. 

image

This commit minimizes the number of `instantiate` operations, but it is not caching the type for partial applications. That is, if we have `f a b` and `f a c`. The previous approach caches the result for `f a`, but this commit skips it. That being said, it seems "type checking" time still went down. It is unclear to me how the "wall clock" time went up.

PR #2782, opened 2023-10-29
perf: use pointer equality at kernel `inferType` and `whnf` caches  (WIP)

leodemoura:
Motivation: when type checking big terms, we may get hash collisions for big terms that are structurally different, but very similar. In this kind of situation we may spend we often get hash collisions

See new test.

### DIFF
--- a/tests/bench/dag_issues/dag2.lean
+++ b/tests/bench/dag_issues/dag2.lean
@@ -1,0 +1,71 @@
+import Lean
+
+/-!
+This test produces a timeout if we don't use pointer equality at the
+kernel infer_type and whnf caches.
+Reason: it produces a hash collision for big, very similar, but structurally different terms.
+Then, `m_infer_type` cache gets lost comparing terms for equality.
+-/
+
+open Lean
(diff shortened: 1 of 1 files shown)


leodemoura:
!bench

leodemoura:
@semorrison I am performing some performance experiments to address kernel performance issues in benchmarks coming from software verification. I would love to have a `!bench` that reports the impact on Mathlib. Do you see a way to have to make this happen? :)

leodemoura:
TODO: try to cache small expressions using `==` and big ones using pointer equality.

semorrison:
@leodemoura, certainly I can manually run benchmarks for Mathlib, although it requires a  successful CI run here for the toolchain is created.

See https://github.com/leanprover-community/mathlib4/pull/8021 where I am in the process of benchmarking the performance impact of #2781.

I can do the same here once there is a toolchain.

I don't think it makes sense to automate this quite yet: we need to get the reliability of the Mathlib builds better still. (Working on it! :-)

PR #2783, opened 2023-10-29
fix: issue 2042

### LINKED ISSUE #2042: Equation lemmas are longer generated by a simple match
### Prerequisites

### Description

Compare this Lean 3 code
```lean
@[simp]
def foo : ℕ → ℕ
| a := 2 * a

example : foo = λ a, a + a :=
begin
  success_if_fail { simp },  -- only `foo x` is matched by the equation lemma
  funext,
  simp, -- now matches
  sorry
end
```
vs the same code in Lean 4
```lean
@[simp]
def foo : Nat → Nat
| a => 2 * a

example : foo = fun a => a + a :=
by
  simp -- unfolds foo into a lambda
  sorry
```

The Lean3 behavior was very useful in mathlib; for instance, in the definition of
```lean
/-- The transpose of a matrix. -/
def transpose (M : matrix m n α) : matrix n m α
| x y := M y x
```
it means that `simp [transpose]` only simplifies terms of the form `transpose M i j` and not `transpose M`.

### Steps to Reproduce

1. Run the code above in Lean 3 and Lean 4
2. Observe the behavior difference

**Expected behavior:** Lean 4 behaves like Lean 3 and generates an equation lemma `foo a = 2 * a`

**Actual behavior:** Lean 4 generates an equation lemma ` foo = fun a => 2 * a` 

### Versions

Lean 4.0.0-nightly-2023-01-16

### Additional Information

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.40.5Bsimp.5D.20on.20equation.20compiler.20definitions/near/322022846), and an [older Zulip message](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.E2.9C.94.20simp.20normal.20form/near/318471016)

leodemoura:
Fixes #2042 

To fix proofs affected by this change, use `unfold f` or `simp (config := { unfoldPartialApp := true }) [f]`.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -41,6 +41,9 @@ v4.3.0
 * The `:=` syntax for configuration declarations (i.e., `package`, `lean_lib`, and `lean_exe`) has been deprecated. For example, `package foo := {...}` is deprecated.
 * [support for overriding package URLs via `LAKE_PKG_URL_MAP`](https://github.com/leanprover/lean4/pull/2709)
 
+* `simp [f]` does not unfold partial applications of `f` anymore. See issue [#2042](https://github.com/leanprover/lean4/issues/2042).
+  To fix proofs affected by this change, use `unfold f` or `simp (config := { unfoldPartialApp := true }) [f]`.
+
 v4.2.0
 ---------
 


semorrison:
I've rebased this onto `nightly`, and will watch the Mathlib CI output.

semorrison:
This one requires fixes to both Std and Mathlib. Working on it!

semorrison:
This would require *many* uses of `(config := { unfoldPartialApp := true })` to naively adapt Mathilb. I've made a representative sample of such changes in https://github.com/leanprover-community/mathlib4/pull/8023, but we will wait to see what users think before proceeding!

leodemoura:
@semorrison I considered adding the notation `simp [!f]` to force `simp` to unfold `f` even if it is partially applied.

semorrison:
Sorry about the long delay here. I've finished making the Mathlib adaptations, and hopefully CI will confirm shortly that this builds Mathlib. We can survive without the proposed new `!` notation, I think.

PR #2786, opened 2023-10-30
feat: lake: Sort manifest packages by name

### LINKED ISSUE #2664: RFC: lake update reorders the manifest
### Proposal

The `lake update <package>` command reorders the manifest, because it leaves non-updated packages at the top. See discussion at https://github.com/leanprover-community/mathlib4/pull/7602#discussion_r1351972534.

Reordering makes reviewing these changes more difficult. Given that `lake update` is already a sharp knife that people are prone to cutting themselves with, it would be good to make reviewing changes made by `lake update` as simple as possible, in particularly by never reordering the manifest.

### Impact

nomeata:
This fixes #2664

The manifest’s `packages` field is only assembled in
`Workspace.updateAndMaterialize`, where previously a `OrdNameMap` was
filled. The `OrdNameMap` builds on `RBArray` which explicity preserves
the insertion order. But for the manifest, it seems desirable to keep
the entries independent of insertion order in this code, so this now
uses a name map that keeps names sorted.

Note that Lean’s `Lean.NameMap` is not a great choice, beceause it sorts
names by their Hash first; using this order would be counter-intuitive
for users, and possibly annoying if the hash function changes.

Therfore, we use `RBMap Name α Name.ord` (abbreviated to
`SortedNameMap`) here.

### DIFF
--- a/src/lake/Lake/Load/Main.lean
+++ b/src/lake/Lake/Load/Main.lean
@@ -96,13 +96,13 @@ If `reconfigure`, elaborate configuration files while updating, do not use OLean
 -/
 def Workspace.updateAndMaterialize (ws : Workspace)
 (toUpdate : NameSet := {}) (reconfigure := true) : LogIO Workspace := do
-  let res ← StateT.run (s := mkOrdNameMap MaterializedDep) <| EStateT.run' (mkNameMap Package) do
+  let res ← StateT.run (s := mkSortedNameMap MaterializedDep) <| EStateT.run' (mkNameMap Package) do
     -- Use manifest versions of root packages that should not be updated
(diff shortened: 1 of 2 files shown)


nomeata:
awaiting-review

tydeu:
Oh, I was hoping to pre-empt you spending time on this with #2788, but I guess I was too late.

semorrison:
@nomeata, sorry, this was my fault for not communicating the priority of this issue with @tydeu. Apologies for the duplicated work, I know that is always frustrating.

nomeata:
No worries, for me it was useful to play around with the lake codebase

PR #2787, opened 2023-10-30
doc: Add docstrings to `dbg_trace` and `assert!` in `do` blocks

nomeata:
they had doc strings in their term forms, but the doElem variant did
not, as noted [on zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Infoview.20hangs.20after.20using.20.60IO.2Eprintln.60.20in.20.60Delab.60/near/399317734)


leodemoura (approved):

Kha:
Note: unlike in mathlib, changing the title does not change the eventual commit message :)

nomeata:
I guess I should have triggered https://squasher.nomeata.de/ before.

Can we haz Borg or mergify soon, to reduce this friction? :-)

PR #2788, opened 2023-10-30
fix: lake: consistent order for manifest packages

### LINKED ISSUE #2664: RFC: lake update reorders the manifest

tydeu:
Closes #2664.

Packages in the Lake manifest now consistently follow the order in the workspace (which is based on the order of the `require` statements in the configuration files). Different invocations of `lake update` thus no longer change the order (assuming the order in the configuration files has not changed). 

One side effect of this approach is that `lake update` will now remove packages even if they are not explicitly updated (i.e., removing a `require a` before `lake update b` will remove `a` from the manifest as would the usual `lake update a`). This seems potentially fine to me, so I have not worked around this side effect yet, but whether it is actually fine is probably worth discussing.

### DIFF
--- a/src/lake/tests/depTree/test.sh
+++ b/src/lake/tests/depTree/test.sh
@@ -129,14 +129,13 @@ sed_i 's/third commit/fourth commit/' A.lean
 git commit -am 'fourth commit in a'
 popd
 pushd d
-# d: no require c
-sed_i '/require c/d' lakefile.lean
 # d: b@1 -> b@2 => a@1 -> a@3
 $LAKE update b -v
 # test 119: pickup a@3 and not a@4
 grep 'third commit in a' lake-packages/a/A.lean
 # test the removal of `c` from the manifest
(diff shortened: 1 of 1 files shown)


tydeu:
@semorrison I requested a review because I am curious whether you think the side-effect is reasonable.

semorrison:
I think it's reasonable, if only because removing a requirement is a rather rare event, and so there aren't too many opportunities to be confused by `lake update`'s behaviour there!

PR #2800, opened 2023-11-01
perf: minor codegen optimizations

SchrodingerZhu:
# Read and remove this section before submitting

* Ensure your PR follows the [External Contribution Guidelines](https://github.com/leanprover/lean4/blob/master/CONTRIBUTING.md).
* Please make sure the PR has excellent documentation and tests. If we label it `missing documentation` or `missing tests` then it needs fixing!
* Add the link to your `RFC` or `bug` issue below.
* If the issue does not already have approval from a developer, submit the PR as draft.
* Remove this section before submitting.

You can manage the `awaiting-review`, `awaiting-author`, and `WIP` labels yourself, by writing a comment containing one of these labels on its own line.

# Summary

Link to `RFC` or `bug` issue:

### DIFF
--- a/src/include/lean/lean.h
+++ b/src/include/lean/lean.h
@@ -1178,26 +1178,57 @@ static inline lean_obj_res lean_nat_succ(b_lean_obj_arg a) {
         return lean_nat_big_succ(a);
 }
 
+#ifdef __has_builtin
+#if __has_builtin(__builtin_add_overflow)
+#define LEAN_H_HAS_ADD_OVERFLOW
+#endif
+#endif
+
+#ifndef LEAN_H_HAS_ADD_OVERFLOW
 static inline lean_obj_res lean_nat_add(b_lean_obj_arg a1, b_lean_obj_arg a2) {
     if (LEAN_LIKELY(lean_is_scalar(a1) && lean_is_scalar(a2)))
         return lean_usize_to_nat(lean_unbox(a1) + lean_unbox(a2));
     else
         return lean_nat_big_add(a1, a2);
 }
+#else
+static inline lean_obj_res lean_nat_add(b_lean_obj_arg a1, b_lean_obj_arg a2) {
+    if (LEAN_LIKELY(lean_is_scalar(a1) && lean_is_scalar(a2))) {
+        size_t n1 = (size_t)a1;
+        size_t n2 = ((size_t)a2) & (~1);
+        size_t res;
+        if (LEAN_UNLIKELY(__builtin_add_overflow(n1, n2, &res)))
+            // rotate right by 1 bit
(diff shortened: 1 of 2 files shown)


src/include/lean/lean.h
```diff
@@ -1213,6 +1244,24 @@ static inline lean_obj_res lean_nat_mul(b_lean_obj_arg a1, b_lean_obj_arg a2) {
         return lean_nat_big_mul(a1, a2);
     }
 }
+#else
+static inline lean_obj_res lean_nat_mul(b_lean_obj_arg a1, b_lean_obj_arg a2) {
+    if (LEAN_LIKELY(lean_is_scalar(a1) && lean_is_scalar(a2))) {
+        ptrdiff_t n1 = (ptrdiff_t)lean_unbox(a1);
```
eric-wieser, line 1250: Why the cast to `ptrdiff_t`?

SchrodingerZhu: Sadly, ssize_t does not seem to be in C standard. 

https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html

According to C++, std::ssize returns ptrdiff_t as a signed counterpart of size_t.

Please also notice that intptr_t/uintptr_t is marked as optional while ptrdiff_t is a part of standard since C99 standard (ISO/IEC 9899:1999).

eric-wieser: Why do you need a signed type here? `__builtin_mul_overflow` appears to work for unsigned multplication too, as far as I can tell.

SchrodingerZhu: We are multiplying two unsigned integers. If result spans to last sign bit, then it must be the case where a plain box cannot encode the whole integer result.

SchrodingerZhu: For example, the following function has different behaviors:
```
#include <cstddef>
#include <iostream>
#include <cstdint>
bool f(size_t i, size_t j) {
    ptrdiff_t res;
    return __builtin_mul_overflow((ptrdiff_t)i, (ptrdiff_t)j, &res);
}

bool g(size_t i, size_t j) {
    size_t res;
    return __builtin_mul_overflow((size_t)i, (size_t)j, &res);
}

int main() {
    std::cout << std::boolalpha << f(SIZE_MAX/2, 2) << " " << g(SIZE_MAX/2, 2)  << std::endl;
}
```
Consider multiplying
```
lean_box(SIZE_MAX/2), lean_box(2)
```
the result will not be able to be stored in `lean_box`. Signed `mul` will detect this correctly.

semorrison:
!bench

SchrodingerZhu:
change is not interesting enough. closing this.

PR #2801, opened 2023-11-01
refactor: lake: more flexible manifest

tydeu:
Refactored manifest code to make it easier to mdodify the its format across future versions while mataining backwards compatibility. 

For actual format changes: removed `opts` and added `configFile` and `manifestFile` to the package entry.  Options can be platform-specific are already locked via the configuration olean. Duplication here is unnecessary and possibly incorrect. Adding the paths for the configuraiton and manifest makes it possible to find them in the future should the defaults change (enabling backwards compatibility).

Deprecated the `manifestFile` package configuration option. This creates complications should we ever wish to load the manifest prior to the package configuration.


tydeu:
!bench

PR #2805, opened 2023-11-02
chore: remove `nonDep` flag from `Expr.letE`

digama0:
The `nonDep` flag in letE has the following scary warning:

https://github.com/leanprover/lean4/blob/1e915b12486395e0fac028db734f6e5d5fd8e1a5/src/Lean/Expr.lean#L405-L413

Because of this, most modules are forced to ignore it and basically do random useless things with the field. However, it is actually completely unused (no one actually uses the value for anything), and while the text above implies local usage in the compiler, it seems likely given the current state of the new compiler that it would use a different data structure than `Expr` anyway for storing this kind of information, so if both the old and new compilers are not already using this information (considering that the whole frontend transforming `Expr` into LCNF is done) I think it is safe to say that it will not be used in the future. Plus, we already have `mdata` for storing auxiliary data like this in exprs in a pinch, so I don't think it is necessary to include an extra field in `Expr` for something so rarely used.

It should also be a minor performance win since we are doing strictly less work.


digama0:
The build seems to be failing due to a bootstrapping issue with LeanInk. (?!) Ideally I would be able to supply a patch for the LeanInk repo using the PR release version of lean, but it doesn't build one (AFAIK) unless the CI passes so it seems like a catch-22. cc: @semorrison

leodemoura:
I have future plans for this flag. Please do not merge.

PR #2814, opened 2023-11-03
fix: Move RecApp marker further inside the expression

### LINKED ISSUE #2810: Superfluous parentheses can inhibit structural recursion
### Prerequisites

### Description

When additional parentheses are inserted around a function, this can prevent structural recursion from succeeding.

Simple example:
```lean
def f (n : Nat) : Nat :=
  match n with
  | 0 => 0
  | n + 1 => (f) n
```
With the parentheses, "structural recursion cannot be used." Replacing `(f) n` with `f n` makes it succeed. Looking at the terms `(f) n` and `f n`, the difference is whether `n` is included within the recApp metadata expression.

This behavior indirectly appeared in a Zulip question when trying to make use of structural recursion within a theorem. This is a minified example:
```lean
theorem thm (N : Nat) :
    N * 0 = 0 :=
  match N with
  | Nat.zero => rfl
  | Nat.succ n => by
    rewrite [Nat.succ_mul, Nat.add_zero]
    rw [thm]
```
Replacing `rw [thm]` with `rw [thm n]` makes structural recursion succeed -- the terms are identical except for whether `n` is included within a recApp metadata expression. Replacing `rw [thm]` by `simp only [thm]` also makes it succeed -- in this case `simp` does not create an appRec metadata expression.

### Context

[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/.60Nat.2Efold.60.20with.20push.20gives.20an.20exact.20size/near/399639322)

### Versions

leanprover/lean4:v4.3.0-rc1

### Impact

nomeata:
This didn't work before
```
def f (n : Nat) : Nat :=
  match n with
  | 0 => 0
  | n + 1 => (f) n
```
because the `RecApp` metadata marker gets in the way. More practical,
such code seems to be produced when using `rw` or `simp` in recursive
theorems.

We can fix the issue by storing the `.mdata` not around the full
application expression, but rather around the inner `.const`. This way,
`Expr.withApp` and similar traversals work without without issue.

While I’m at it: So far only `WellFounded` was actually making use of
this. Now (unlike before)
```
def f (n : Nat) : Nat :=
  match n with
  | 0 => 0
  | n + 1 => f (n + 1)
```
will show the error with squiggly lines under `f (n + 1)`.

Test case is included. Fixes #2810.

### DIFF
--- a/tests/lean/1673.lean.expected.out
+++ b/tests/lean/1673.lean.expected.out
@@ -3,4 +3,4 @@
 with errors
 structural recursion cannot be used
 
-failed to prove termination, use `termination_by` to specify a well-founded relation
+well-founded recursion cannot be used, 'foo.a' does not take any (non-fixed) arguments
--- a/tests/lean/terminationFailure.lean.expected.out
+++ b/tests/lean/terminationFailure.lean.expected.out
@@ -9,7 +9,7 @@ f (x : Nat) : Nat
 f.g (a✝ : Nat) : Nat
 1
 2
-terminationFailure.lean:20:4-20:5: error: fail to show termination for
+terminationFailure.lean:24:9-24:12: error: fail to show termination for
   h
 with errors
 argument #1 was not used for structural recursion
--- a/tests/lean/treeMap.lean.expected.out
+++ b/tests/lean/treeMap.lean.expected.out
@@ -1,4 +1,4 @@
-treeMap.lean:8:59-8:69: error: failed to prove termination, possible solutions:
+treeMap.lean:9:15-9:23: error: failed to prove termination, possible solutions:
   - Use `have`-expressions to prove the remaining goals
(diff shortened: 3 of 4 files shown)


nomeata:
Ok, finally there, it seems. 

This approach (keeping `.mdata` around the `.const`) has its downsides as well; there are quite a few places where the `.const` is replaced with an expression, and thus we have to move the `.mdata` along. But probably still better than #2813.

In a few places I wrote code does what it should, but in a non-pretty way that needs cleanup. Hence leaving this as WIP for now.

The test output changes arebecause of improved error locations in structural recursion, and in one case a better termination check error message.

nomeata:
Learned a lot while doing this, in particular, that there is a much more elegant approach: #2818.

PR #2816, opened 2023-11-04
feat: find Decidable instances via unification

kim-em:
I've been finding situations where these `simp` lemmas don't fire, because the decidable instance is not the one found by typeclass inference.

This perhaps deserves a deeper fix, but I want to see if just changing the binder types here suffices for my purposes.

(Previously this behaviour was being hidden by the default `decide := true` in `simp`.)

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -806,6 +806,12 @@ decidability instance instead of the proposition, which has no code).
 
 If a proposition `p` is `Decidable`, then `(by decide : p)` will prove it by
 evaluating the decidability instance to `isTrue h` and returning `h`.
+
+Because `Decidable` carries data,
+when writing `@[simp]` lemmas which include a `Decidable` instance on the LHS,
+it is best to use `{_ : Decidable p}` rather than `[Decidable p]`
+so that non-canonical instances can be found via unification rather than
+typeclass search.
 -/
 class inductive Decidable (p : Prop) where
   /-- Prove that `p` is decidable by supplying a proof of `¬p` -/


leodemoura:
> This perhaps deserves a deeper fix,

Agreed.

> but I want to see if just changing the binder types here suffices for my purposes.

Happy to merge the workaround. We have used it in other places.

leodemoura (approved):

semorrison:
Mathlib adaptation PR at https://github.com/leanprover-community/mathlib4/pull/8285

src/Init/SimpLemmas.lean
```diff
@@ -143,9 +143,9 @@ theorem Bool.or_assoc (a b c : Bool) : (a || b || c) = (a || (b || c)) := by
 @[simp] theorem Bool.not_eq_true (b : Bool) : (¬(b = true)) = (b = false) := by cases b <;> decide
 @[simp] theorem Bool.not_eq_false (b : Bool) : (¬(b = false)) = (b = true) := by cases b <;> decide
 
-@[simp] theorem decide_eq_true_eq [Decidable p] : (decide p = true) = p := propext <| Iff.intro of_decide_eq_true decide_eq_true
-@[simp] theorem decide_not [h : Decidable p] : decide (¬ p) = !decide p := by cases h <;> rfl
-@[simp] theorem not_decide_eq_true [h : Decidable p] : ((!decide p) = true) = ¬ p := by cases h <;> simp [decide, *]
+@[simp] theorem decide_eq_true_eq {_ : Decidable p} : (decide p = true) = p := propext <| Iff.intro of_decide_eq_true decide_eq_true
```
eric-wieser, line 146: Can you add a note to the docstring of `Decidable` that describes this style and when/why its useful?

src/Init/Prelude.lean
```diff
@@ -806,6 +806,12 @@ decidability instance instead of the proposition, which has no code).
 
 If a proposition `p` is `Decidable`, then `(by decide : p)` will prove it by
 evaluating the decidability instance to `isTrue h` and returning `h`.
+
+Because `Decidable` carries data,
+when writing `@[simp]` lemmas which include a `Decidable` instance on the LHS,
```
eric-wieser, line 811: Is the corollary here that you can no longer use the lemma with `←` if no `Decidable` instance is on the *R*HS? And so in those cases, you need to write two versions of the same lemma?

(Re-posted as this got lost as a non-PR comment)

PR #2821, opened 2023-11-05
feat: string gaps for continuing string literals across multiple lines

kmill:
Implements "gaps" in string literals. These are escape sequences of the form `"\" newline whitespace+` that have the interpretation of an empty string. For example,
```
  "this is \
     a string"
```
is equivalent to `"this is a string"`. These are modeled after string continuations in [Rust](https://doc.rust-lang.org/beta/reference/tokens.html#string-literals).

Implements RFC #2838

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -11,6 +11,15 @@ of each version.
 v4.5.0 (development in progress)
 ---------
 
+* Modify the lexical syntax of string literals to have string gaps, which are escape sequences of the form `"\" newline whitespace*`.
+  These have the interpetation of an empty string and allow a string to flow across multiple lines without introducing additional whitespace.
+  The following is equivalent to `"this is a string"`.
+  ```lean
+  "this is \
+     a string"
+  ```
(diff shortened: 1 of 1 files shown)


tests/lean/string_gaps.lean
```diff
@@ -0,0 +1,76 @@
+import Lean
```
Kha, line 1: Please minimize the import and add docstrings https://lean-lang.org/lean4/doc/dev/testing.html#writing-good-tests

Kha (changes requested):

kmill:
@Kha What are you looking for with the changelog entry? Do you mean I should write something in `RELEASES.md` for the development version?

By the way, once the string literal PRs are all sorted out, I can write a rough draft of the strings section of the reference manual.

Kha:
> Do you mean I should write something in `RELEASES.md` for the development version?

Yes, a brief sentence similar to the PR title and a PR link are sufficent.

> By the way, once the string literal PRs are all sorted out, I can write a rough draft of the strings section of the reference manual.

Great! You should probably get in touch with @david-christiansen for that.

PR #2825, opened 2023-11-05
Ensure nested proofs in equation and unfold lemmas have been abstracted

leodemoura:


tests/lean/run/PPTopDownAnalyze.lean
```diff
@@ -310,7 +310,8 @@ def takesStrictMotive ⦃motive : Nat → Type⦄ {n : Nat} (x : motive n) : mot
 def arrayMkInjEqSnippet :=
   fun {α : Type} (xs : List α) => Eq.ndrec (motive := fun _ => (Array.mk xs = Array.mk xs)) (Eq.refl (Array.mk xs)) (rfl : xs = xs)
 
-#testDelabN arrayMkInjEqSnippet
+-- TODO: fix following test
+-- #testDelabN arrayMkInjEqSnippet
```
leodemoura, line 314: @Kha This issue was being masked by `abstractNestedProofs`. It was incorrectly abstracting the `arrayMkInjEqSnippet` since it was marked as `def`. So, it was not testing anything since the body was just an application after the abstraction step.

leodemoura: There is no rush to fix this one.

leodemoura:
@semorrison I think the CI failed here. No feedback.

semorrison:
I haven't seen that before, and I don't think it was related to my recent changes. Pushing an empty commit got it out of its stuck state.

tests/lean/run/PPTopDownAnalyze.lean
Kha: Good, because I don't know anything about `pp.analyze` :)

semorrison (approved):
Mathlib is working with this toolchain: all the changes are positive (many unused arguments that previously weren't being noticed by the linter).

Issue #2829, opened 2023-11-06
C Compiler warning "trigraph ignored" during lake build


PratherConid:
### Description

```
[1/6] Building Cwarning
stderr:
./build/ir/Cwarning/Basic.c:22:35: warning: trigraph ignored [-Wtrigraphs]
x_1 = lean_mk_string_from_bytes("(??)", 4);
```

### Steps to Reproduce

```
git clone https://github.com/PratherConid/lean-cwarning.git; cd lean-cwarning; lake build
```
Alternatively, run ``lake build`` on a repo containing the following definition:
```
def hello := "(??)"
```

### Versions

Lean (version 4.3.0-rc1, commit baa4b68a7192, Release)
Ubuntu 20.04.6 LTS (WSL)

nomeata:
It doesn’t look like it needs lake to reproduce this:

```
/tmp/test $ LANG=C lean -c test.c test.lean  && LANG=C leanc test.c
test.c: In function '_init_l_main___closed__1':
test.c:23:35: warning: trigraph ??) ignored, use -trigraphs to enable [-Wtrigraphs]
   23 | x_1 = lean_mk_string_from_bytes("(??)", 4);
      |
At top level:
cc1: note: unrecognized command-line option '-Wno-unused-command-line-argument' may have been intended to silence earlier diagnostics
```

Kha:
We can also stick it here statically, there's already another one: https://github.com/leanprover/lean4/blob/a2afdb62c5c95a8927ed32680df0096942432d5c/src/Leanc.lean#L56

nomeata:
Do we have to worry about people using `leanc` with other C compilers? Or is that flag standard enough? I have no idea about these things.

nomeata:
I suggest to fix it in `quoteString`. This avoids worrying about where to exactly insert the flag, and also benefits whoever processes the `.c` file outside of `leanc`.

* https://github.com/leanprover/lean4/pull/2833

digama0:
Note that trigraphs are deprecated in C, and not supported in later versions of the standard at all. Given that this is only a warning, it seems like suppressing the warning is the simplest way to get the desired behavior.

nomeata:
The issue is worth fixing; not sure if the fixes differ so much that it’s worth spending discussion bandwidth on them. I prefer simply avoiding digraphs (e.g. I expect more people to read command line flags as they debug stuff compared to people reading string literals in generated C codes, so adding noise to the latter seems less invasive than the former).

[...]

PR #2832, opened 2023-11-06
test: C trigraph

nomeata:
add a test file that checks that C trigraphs in string literals are not
miscompiled.


digama0:
is there a way to set this test up to also ensure you don't get the warning from #2829?

nomeata:
Yeah, that would be good. This use case (check stderr/out of `leanc`) doesn't seem to be supported by the test suite at the moment, or I didn’t see it.

A simple hammer would be to make tests in `lean/compiler` fail whenever they have output, but that would break existing files:
```
1549: Test command: /nix/store/m6xrxfbgvm8791x09f7ims712hwy3aqh-bash-interactive-5.2-p15/bin/bash "-c" "PATH=/home/jojo/build/lean/lean4/build/release/stage1/bin:$PATH LEAN_CC=/nix/store/xkj1aa2f8gmqs1q3vqxcnl89n1iqwakv-clang-wrapper-15.0.7/bin/clang CXX='/nix/store/xkj1aa2f8gmqs1q3vqxcnl89n1iqwakv-clang-wrapper-15.0.7/bin/clang++  -O3 -DNDEBUG' LEANC_OPTS=' -O3 -DNDEBUG' ./test_single.sh append.lean"
1549: Working Directory: /home/jojo/build/lean/lean4/src/../tests/compiler
1549: Test timeout computed to be: 1500
1549: running C program...
1549: append.lean:3:10: warning: unused variable `xs` [linter.unusedVariables]
 4/53 Test #1549: leancomptest_append.lean .....................   Passed    0.33 sec
test 1550
      Start 1550: leancomptest_array.lean
```

The next level of sophistication is to have `.expect.out` files for both `lean` and `leanc` in `tests/compiler` tests. That would require a lot of empty files.

The next next level of sophistication is to treat a missing `.expect.out` as an empty file. That’s what I did when setting up the Motoko compiler test suite, and found it quite convenient and effective.

digama0:
> The next next level of sophistication is to treat a missing .expect.out as an empty file. That’s what I did when setting up the Motoko compiler test suite, and found it quite convenient and effective. Woulds also require changes to diff_produced.

I was going to suggest treating a missing `.expect.out` as a run test, i.e. you just ignore the output unless you are told to compare it to something.

nomeata:
Also a viable option. I like “assert output is empty” is a good implicit default (also because it means that many tests only need the `.lean` file and nothings else) and “ignore output” something that should be explicit. In general, test suite organization is very prone to developer bikeshedding :-)

digama0:
If all existing tests have no output, then that's fine, but my guess is that the status quo is that the output is ignored so checking / asserting all the tests will mean lots of new files.

Kha:
Not checking the warning seems good enough for now

PR #2836, opened 2023-11-06
fix: whitelist environment extensions compatible with lakefile.olean fast-loading

### LINKED ISSUE #2632: Lake segmentation fault

Kha:
Fixes #2632

TODO:
- [ ] do we need a faster check? Probably not, we're talking about `23 * #lakefiles` name comparisons.
- [ ] use previous approach instead of elaboration as fallback
- [ ] check and add more envexts if desired, at least the Lake ones

### DIFF
--- a/src/lake/Lake/Load/Elab.lean
+++ b/src/lake/Lake/Load/Elab.lean
@@ -86,6 +86,10 @@ toolchain). Otherwise, elaborate the configuration and save it to the OLean.
 def importConfigFile (wsDir pkgDir : FilePath) (lakeOpts : NameMap String)
 (leanOpts := Options.empty) (configFile := pkgDir / defaultConfigFile) (reconfigure := true) : LogIO Environment := do
   let olean := configFile.withExtension "olean"
+  let useElab := do
+    let env ← elabConfigFile pkgDir lakeOpts leanOpts configFile
+    Lean.writeModule env olean
+    return env
   let useOLean ← id do
     if reconfigure then return false
     let .ok oleanMTime ← getMTime olean |>.toBaseIO | return false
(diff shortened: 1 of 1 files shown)


tydeu:
@Kha Do we even need to fallback to elaboration rather than just skipping incompatible ones (and maybe even skip Lake-unutilized extensions). All Lake is doing with this environment is loading constant names from the Lake attributes and then evaluating the relevant constants, so it only needs proper extension values for its attributes and the IR, right?

Kha:
@tydeu That sounds correct, yes!

tydeu:
Superseded by #2836.

Issue #2812, opened 2023-11-03
RFC: Extensible pre-definition-elaboration


nomeata:
#### Context

Recursive function definitions are currently elaborated to a `PreDefinition`, which is what’s sent to the compiler/interpreter. On the way to the kernel, the PreDefinitions are grouped by strongly connected component, and then pass through [`Lean.Elab.addPreDefinition`](https://leanprover-community.github.io/mathlib4_docs/Lean/Elab/PreDefinition/Main.html#Lean.Elab.addPreDefinitions) which turns them into non-recursive `Definition` suitable for the kernel.

#### Status quo

Currently, Lean has a few built-in modes to do this step:

* Non-recursive definitions are translated straight-forwardly
* `unsafe` definitions are turned into axioms
* `partial` definitions are made opaque (but don’t require axioms because the result type must be `Inhabited`)
* structurally non-mutual recursive functions are expressed using bounded recursion `brec` recursor of the underlying data type
* well-founded recursive functions are expressed using `WellFounded.fix`

The code makes educated guesses: Certain annotations cause it to go for well-founded recursion, otherwise it tries structurally first and then well-founded.

This also means that these strategies have a rather simple interface:

> (preDefs : Array PreDefinition) : TermElabM Unit

#### Motivation for more strategies

There are possibly a large number of alternative ways to process a PreDefinition. Some particularly interesting ones are

* Partial functions with an explicit inductive predicate for the domain of the function
* Monotone functions in a CPO, where the function exists by construction, even for non-terminating equations. In particular
  * Monadic functions in the `Option` monad (and inhabited result type)
  * Tail-recursive functions with inhabited result type.
* Well-founded recursion using _extrinsic_ “termination” proof using congruence lemmas
* Coinductive definitions

It’s easy to tell that I am approaching this from an Isabelle background, where [all these are possible](https://isabelle.in.tum.de/doc/functions.pdf). Of course it’s not completely trivial to port them to Lean, but at least with enough `Inhabited` side conditions and expecting only propositional equalities I believe they can be used here, and will be useful.

There many be many more interesting takes here that people could experiment with without taking the high hurdle of changing core code.

Finally, it reduces friction even for those working on the “core strategies” if bigger changes can be iterated upon in a separate project.

#### Proposed change (high-level)

Lean is generally a very extensible system. So it beholds of Lean to allow users to write their own `PreDefinition` processor, without having to touch the core repo.

So this requires

* A good name for the concept
* An API interface
* Surface syntax

[...]

PR #2839, opened 2023-11-07
feat: Extensible derecursifier

nomeata:
This is not meant to be merged!

This branch can be used to prototype derecursifiers outside the lean4
tree. Example:

```
import Lean.Elab.PreDefinition.Main

set_option autoImplicit false

open Lean Elab

def foo (n : Nat) : Nat := match n with
  | .zero => .zero
  | .succ n => foo n
derecursify_with Structural.structuralRecursion

def derec (preDefs : Array PreDefinition) : TermElabM Unit := do
  logInfo m!"{preDefs.map (·.declName)}"
  return

def foo2 (n : Nat) : Nat := match n with
  | .zero => .zero
  | .succ n => foo2 n
derecursify_with derec
```

_Maybe_ a proper designed version of this feature is useful at some
point, see #2812 for that.

### DIFF
--- a/src/Lean/Elab/Declaration.lean
+++ b/src/Lean/Elab/Declaration.lean
@@ -334,7 +334,11 @@ def expandMutualPreamble : Macro := fun stx =>
 
 @[builtin_command_elab «mutual»]
 def elabMutual : CommandElab := fun stx => do
-  let hints := { terminationBy? := stx[4].getOptional?, decreasingBy? := stx[5].getOptional? }
+  let hints := {
+      derecursifyWith? := stx[3].getOptional?
+      terminationBy? := stx[4].getOptional?
+      decreasingBy? := stx[5].getOptional?
+  }
   if isMutualInductive stx then
     if let some bad := hints.terminationBy? then
       throwErrorAt bad "invalid 'termination_by' in mutually inductive datatype declaration"


nomeata:
Odd, the failing test cases say
```
  [failed to pretty print term (use 'set_option pp.rawOnError true' for raw representation)
```

Why does extending some syntax, but not even using it, affect the pretty printers?

digama0:
Changing existing syntax generally requires an update-stage0, because stage0 syntax is used inside syntax match expressions in the stage1 compiler, meaning that it will incorrectly parse stage1 syntax which can cause a variety of random errors. You can also try setting `parseQuotWithCurrentStage` to true in `stage0/stdlib_flags.h`, which will make stage1 do its best to use stage1 syntax, although I suspect it won't work in this case because the syntax in question is builtin syntax defined in the `Lean` package rather than regular `syntax` defined in the `Init` package and used after definition.

nomeata:
Ah, I feared it was something about bootstrapping. Thanks for the pointers!

Kha:
> I suspect it won't work in this case because the syntax in question is builtin syntax defined in the `Lean` package rather than regular `syntax` defined in the `Init` package and used after definition.

It should work *if* the module of the quotation imports the parser

PR #2841, opened 2023-11-07
Mark environments not freed before process exit persistent eagerly

Kha:
Another ~-10% on `import Mathlib`

### DIFF
--- a/src/Init/Util.lean
+++ b/src/Init/Util.lean
@@ -96,11 +96,10 @@ macro:max x:term noWs "[" i:term "]" noWs "!" : term => `(getElem! $x $i)
 def Runtime.markMultiThreaded (a : α) : α := a
 
 /--
-  Marks given value and its object graph closure as multi-threaded if currently
-  marked single-threaded. This will remove reference counter updates but prevent
-  the closure from being deallocated until the end of the process! It can still
-  be useful to do eagerly when the value will be marked persistent later anyway
-  and there is available time budget to mark it now or it would be unnecessarily
-  marked multi-threaded in between. -/
+  Marks given value and its object graph closure as persistent. This will remove
+  reference counter updates but prevent the closure from being deallocated until
+  the end of the process! It can still be useful to do eagerly when the value
+  will be marked persistent later anyway and there is available time budget to
+  mark it now or it would be unnecessarily marked multi-threaded in between. -/
 @[extern "lean_runtime_mark_persistent"]
 def Runtime.markPersistent (a : α) : α := a
--- a/src/Lean/Environment.lean
+++ b/src/Lean/Environment.lean
@@ -765,7 +765,17 @@ partial def importModulesCore (imports : Array Import) : ImportStateM Unit := do
       moduleNames := s.moduleNames.push i.module
     }
 
-def finalizeImport (s : ImportState) (imports : Array Import) (opts : Options) (trustLevel : UInt32 := 0) : IO Environment := do
+/--
+  Construct environment from `importModulesCore` results.
+
+  If `leakEnv` is true, we mark the environment as persistent, which means it
+  will not be freed. We set this when the object would survive until the end of
+  the process anyway. In exchange, RC updates are avoided, which is especially
+  important when they would be atomic because the environment is shared across
+  threads (potentially, storing it in an `IO.Ref` is sufficient for marking it
+  as such). -/
+def finalizeImport (s : ImportState) (imports : Array Import) (opts : Options) (trustLevel : UInt32 := 0)
+    (leakEnv := false) : IO Environment := do
   let numConsts := s.moduleData.foldl (init := 0) fun numConsts mod =>
     numConsts + mod.constants.size + mod.extraConstNames.size
   let mut const2ModIdx : HashMap Name ModuleIdx := mkHashMap (capacity := numConsts)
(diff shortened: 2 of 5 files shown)


Kha:
!bench

src/Lean/Environment.lean
```diff
@@ -765,7 +765,8 @@ partial def importModulesCore (imports : Array Import) : ImportStateM Unit := do
       moduleNames := s.moduleNames.push i.module
     }
 
-def finalizeImport (s : ImportState) (imports : Array Import) (opts : Options) (trustLevel : UInt32 := 0) : IO Environment := do
+def finalizeImport (s : ImportState) (imports : Array Import) (opts : Options) (trustLevel : UInt32 := 0)
```
leodemoura, line 768: Please add documentation for `leakEnv`

leodemoura: And motivate why this was added. We will forget the discussion we had this week.

src/Lean/Environment.lean
```diff
@@ -797,18 +798,28 @@ def finalizeImport (s : ImportState) (imports : Array Import) (opts : Options) (
       moduleData   := s.moduleData
     }
   }
-  let env ← setImportedEntries env s.moduleData
-  let env ← finalizePersistentExtensions env s.moduleData opts
+  env ← setImportedEntries env s.moduleData
+  if leakEnv then
+    /- When allowed to leak `env`, we eagerly mark it persistent here, which
+       avoids RC updates as well as costly MT markings when e.g. an interpreter
+       closure (wich contains the environment) is put in an `IO.Ref` like would
```
leodemoura, line 805: wich => which

src/Lean/Environment.lean
```diff
@@ -797,18 +798,28 @@ def finalizeImport (s : ImportState) (imports : Array Import) (opts : Options) (
...
+       happen with user environment extensions loaded in the next step. -/
+    env := Runtime.markPersistent env
+  env ← finalizePersistentExtensions env s.moduleData opts
+  if leakEnv then
+    /- Also mark environment extension states. -/
+    env := Runtime.markPersistent env
```
leodemoura, line 811: You are calling `markPersistent` twice. It is worth explaining. I am guessing that you are concerned some extension may mark the environment as `MT` at `finalizePersistentExtensions`.

src/Lean/Environment.lean
Kha: Yes, I wasn't sure where to document it. I've split the comment below, moved the first part into the docstring, and extended comments in general.

src/Lean/Environment.lean
Kha: That could happen; mostly I just want to give a consistent guarantee about what is returned with `leakEnv`.

Kha:
The `import Lean` regression is interesting, it shows it is not quite like `import Mathlib` after all: it doesn't have any interpreted code of course so before this PR, the environment object stays ST and marking it persistent is additional overhead. But as we want to introduce parallelism in basically any file, and having the environment be MT is definitely worse than persistent, I think this regression is acceptable. @leodemoura Do you agree?

I double-checked the `import Mathlib` speedup on top of https://github.com/leanprover-community/mathlib4/pull/8199:
```
Benchmark 1: lean Mathlib.lean
  Time (mean ± σ):      1.464 s ±  0.032 s    [User: 1.145 s, System: 0.311 s]
  Range (min … max):    1.437 s …  1.544 s    10 runs

Benchmark 2: lean +lean4 Mathlib.lean
  Time (mean ± σ):      1.336 s ±  0.011 s    [User: 1.041 s, System: 0.294 s]
  Range (min … max):    1.319 s …  1.355 s    10 runs

Summary
  'lean +lean4 Mathlib.lean' ran
    1.10 ± 0.03 times faster than 'lean Mathlib.lean'
    Benchmark 1: lean Mathlib.lean
  Time (mean ± σ):      1.464 s ±  0.032 s    [User: 1.145 s, System: 0.311 s]
  Range (min … max):    1.437 s …  1.544 s    10 runs

Benchmark 2: lean +lean4 Mathlib.lean
  Time (mean ± σ):      1.336 s ±  0.011 s    [User: 1.041 s, System: 0.294 s]
  Range (min … max):    1.319 s …  1.355 s    10 runs

Summary
  'lean +lean4 Mathlib.lean' ran
    1.10 ± 0.03 times faster than 'lean Mathlib.lean'
```

Kha:
> having the environment be MT is definitely worse than persistent

If we want to 100% make sure about this, we can also postpone the PR until such time

src/Init/Util.lean
```diff
@@ -85,3 +85,22 @@ private def outOfBounds [Inhabited α] : α :=
...
+  marked single-threaded. This will make reference counter updates atomic and
+  thus more costly. It can still be useful to do eagerly when the value will be
+  shared between threads later anyway and there is available time budget to mark
+  it now. -/
+@[extern "lean_runtime_mark_multi_threaded"]
+def Runtime.markMultiThreaded (a : α) : α := a
+
+/--
+  Marks given value and its object graph closure as multi-threaded if currently
+  marked single-threaded. This will remove reference counter updates but prevent
```
eric-wieser, line 100: Is this first sentence deliberately identical to the one in the other docstring?

Kha: Wow, thanks, I must have overwritten it in a later revision

leodemoura:
> But as we want to introduce parallelism in basically any file, and having the environment be MT is definitely worse than persistent, I think this regression is acceptable as it should not affect real-life files. @leodemoura Do you agree?

Agreed. Let's merge the PR.
Before merging, please include your analysis as a comment in the file. We don't need the actual numbers.

leodemoura (approved):

PR #2842, opened 2023-11-07
fix: stricter `lakefile.olean` compatibility check

tydeu:
Split from/builds on #2749.

Previously, compatibility of the elaborated configuration file (`lakefile.olean`) with the current configuration was determined via comparing the modification of the olean with the modification time of source configuration file (`lakefile.lean`) and the workspace `lean-toolchain`. The check now saves a trace (`lakefile.olean.trace`) containing information on the platform, Lean Git hash, and the hash of the source lakefile. If there is a mismatch in either of these three details, the `lakefile.olean` is rebuilt with the same options (which are also stored in the trace). The trace file is also locked at file descriptor level to prevent racing between the trace verification and the importing / elaboration of the lakefile.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -15,6 +15,7 @@ v4.4.0 (development in progress)
 * **Lake:** Moved the default build directory (e.g., `build`), default packages directory (e.g., `lake-packages`), and the compiled configuration (e.g., `lakefile.olean`) into a new dedicated directory for Lake outputs, `.lake`. The cloud release build archives are also stored here, fixing [#2713](https://github.com/leanprover/lean4/issues/2713).
 * **Lake:** Update manifest format to version 7 (see [lean4#2801](https://github.com/leanprover/lean4/pull/2801) for details on the changes).
 * **Lake:** Deprecate the `manifestFile` field of a package configuration.
+* **Lake:** There is now a more rigorous check on `lakefile.olean` compatibility (see [#2842](https://github.com/leanprover/lean4/pull/2842) for more details).
 
 v4.3.0
 ---------


semorrison:
Just recording details here: the latest Mathlib failure is real, as Lake's use of lockfiles is preventing us from running Mathlib tests in parallel. Mac will sort this out.

Kha (commented):
The lakefile locking code looks sensible to me. I'll leave review of the IO additions to @joehendrix as before if that hasn't already included. But let's please add tests for them.

Kha (approved):

tydeu:
@Kha Sorry, that was just a rebase cleanup push. I am still working on the mathlib breakage @semorrison mentioned.

src/Init/System/IO.lean
```diff
@@ -328,7 +328,35 @@ namespace FS
 namespace Handle
 
 @[extern "lean_io_prim_handle_mk"] opaque mk (fn : @& FilePath) (mode : FS.Mode) : IO Handle
+
+/--
+Acquire an exclusive lock on the handle.
```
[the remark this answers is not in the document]
tydeu: @Kha There seems to be a mix of styles currently used in the docs here. For example, `read`, `getLine`, and `removeDir` use the imperative whereas `rename` uses the indicative. Should I take this to mean we wish to to transition to using the indicative everywhere?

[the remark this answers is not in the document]
Kha: Yes, we are transitioning to indicative. Which we should also document somewhere.

tydeu:
@semorrison I have hopefully fixed the mathlib incompatibility, but it appears that despite being based on the latest nightly the bot does not wish to test it. Is this expected (i.e., is their some behind-the-scenes work that has to been done before we can test on a particularly nightly)?

semorrison:
Yes, unless the bot can find a commit on the `nightly-testing` branch at Mathlib that already compiles using `nightly-2023-11-14`, it won't start. Because of the backlog of divergences, this might be the case now. I'll manually intervene. :-)

semorrison:
Ah, this is not happening because #2749 broke `cache`.

tydeu:
> Ah, this is not happening because #2749 broke `cache`.
> 
> I've fixed `cache`, and a build of `nightly-testing` is hopefully on the way soon, after which we'll be able to test this PR.

@semorrison, I'm confused, I thought those breakages were already fixed by mathlib4#8216 (which is why #2749) got a checkmark from the bot?

semorrison:
The problem is that that leanprover-community/mathlib4#8216 was not yet merged into `nightly-testing`. This step is still entirely manual, but will not be next time around!

(And the reason for that was that my `nightly-testing` failure bot so thoroughly failed that I didn't get a message about it.)

semorrison:
Okay, `lean-pr-testing-2842` now compiles (and tests, in parallel!) locally for me now. Hopefully the bot will report the same shortly.

tydeu:
I addressed @Kha's comments, @semorrison helped me get a successful build of mathlib, and I have now cleaned up the commit history, so this should be ready for review! @joehendrix, as the IO tsar, I leave the final word to you.

src/Init/System/IO.lean
```diff
@@ -328,7 +328,41 @@ namespace FS
...
+Releases any previously acquired lock on the handle.
+Will succeed even if no lock has been acquired.
+-/
+@[extern "lean_io_prim_handle_unlock"] opaque unlock (h : @& Handle) : IO Unit
+
 @[extern "lean_io_prim_handle_flush"] opaque flush (h : @& Handle) : IO Unit
+/-- Rewind the read/write cursor to the beginning of the handle. -/
+@[extern "lean_io_prim_handle_rewind"] opaque rewind (h : @& Handle) : IO Unit
+/--
+Truncate the handle to the read/write cursor.
```
semorrison, line 358:
```suggestion
Truncates the handle to the read/write cursor.
```

src/lake/Lake/Load/Elab.lean
```diff
@@ -76,51 +77,113 @@ def elabConfigFile (pkgDir : FilePath) (lakeOpts : NameMap String)
...
+def importConfigFileCore (olean : FilePath) (leanOpts : Options) : IO Environment := do
+  /-
+  We could import the olean together with its imports using one
+  `importModules` call, but that would mean we pay for a full
+  `finalizeImports` each time, which is linear in the number of imported
+  constants and extension entries (in fact, it is paid two more times: when
+  marking the `Environment` object as multi-threaded, and when releasing
+  it). As most lakefiles use the same set of imports, we instead construct
+  an `Environment` for it only once, and then apply the lakefile contents
+  on top of it like the elaborator would. Thus the non-shared, part of the
```
semorrison, line 95:
```suggestion
  on top of it like the elaborator would. Thus the non-shared part of the
```

src/lake/Lake/Load/Elab.lean
```diff
@@ -76,51 +77,113 @@ def elabConfigFile (pkgDir : FilePath) (lakeOpts : NameMap String)
...
+  `Environment` is very small.
+  -/
+  let (mod, _) ← readModuleData olean
+  let mut env ← importModulesUsingCache mod.imports leanOpts 1024
+  -- Apply constants (does not go through the kernel, so order is irrelevant)
+  env := mod.constants.foldl addToEnv env
+  /-
+  Apply extension entries (`PersistentEnvExtension.addEntryFn` is pure and
+  does not have access to the whole environment, so no dependency worries
+  here either
```
semorrison, line 105:
```suggestion
  here either)
```

src/lake/Lake/Load/Elab.lean
```diff
@@ -76,51 +77,113 @@ def elabConfigFile (pkgDir : FilePath) (lakeOpts : NameMap String)
...
+  deriving ToJson, FromJson
+
+inductive ConfigLock
+| olean (h : IO.FS.Handle)
+| lean (h : IO.FS.Handle) (lakeOpts : NameMap String)
+
+/--
+Import the `.olean` for the configuration file if `reconfigure` is not set and
+an up-to-date one exists (i.e., one with matching configuration and on the same
+toolchain). Otherwise, elaborate the configuration and save it to the `.olean`.
```
semorrison, line 132:
```suggestion
Imports the `.olean` for the configuration file if `reconfigure` is not set and
an up-to-date one exists (i.e., one with matching configuration and on the same
toolchain). Otherwise, elaborates the configuration and saves it to the `.olean`.
```

joehendrix:
I am fine with the IO changes.

joehendrix (approved):

src/lake/Lake/Load/Elab.lean
tydeu: I did not do this one for now, because the rest of the file is currently in the imperative mood. I think this change would be best done in a full sweep refactor.

digama0:
Both the PR title and changelog say that there is a "stricter check", but what precisely has changed? What was the check doing before and what is it doing now and why will it help?

tydeu:
@digama0 I add a brief summary to the PR body and link to the PR in the release note.

PR #2855, opened 2023-11-10
feat: release web assembly

abentkamp:
# Summary

This PR adds the web assembly build to future releases of Lean. It will help to build Lean code for web assembly (e.g. `lean4game`) without having to compile Lean itself manually. From my local testing, it seems that all relevant files will be included. Moreover, I have added the Web Assembly build to the list of build targets in the documentation.

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -17,7 +17,7 @@ Issue reports and fixes are welcome.
 
 * aarch64 Linux with glibc 2.27+
 * aarch64 (Apple Silicon) macOS
-* Linux x86 (32-bit)
+* x86 (32-bit) Linux
 * Emscripten Web Assembly
 
 <!--


abentkamp:
Actually, for a wasm version of `lean4game`, it would be good to have the native 32bit binary as well. Should that be included in the same release or in another one?

Kha:
You mean for building the game .oleans? That would make sense, feel free to add it in this PR

doc/setup.md
```diff
@@ -17,6 +17,7 @@ Issue reports and fixes are welcome.
 
 * aarch64 Linux with glibc 2.27+
 * aarch64 (Apple Silicon) macOS
+* Emscripten 3.1.44 Web Assembly
```
Kha, line 20: I would avoid mentioning any specific version numbers, they will invariably go out of sync

Kha:
Looks good, let's add "Linux x86 (32-bit)" to Tier 2 as well and then this is ready?

abentkamp:
Yes, ready!

Kha:
Oh, I think we also need a `-DLEAN_INSTALL_SUFFIX=` each to set the name of the release tarball. Let's go with `-linux_x86` and `-linux_wasm32`?

abentkamp:
Oh, good catch. I missed that.

doc/setup.md
```diff
@@ -17,6 +17,8 @@ Issue reports and fixes are welcome.
 
 * aarch64 Linux with glibc 2.27+
 * aarch64 (Apple Silicon) macOS
+* Linux x86 (32-bit)
```
Kha, line 20:
```suggestion
* x86 (32-bit) Linux
```
I guess this would be more consistent. I assume we can't/don't want to make any glibc promises as it's from multilib and from a pinned Nix version

PR #2858, opened 2023-11-10
feat: per-package server options

### LINKED ISSUE #2455: RFC: per-package server options

mhuisi:
This PR adds per-package server options to resolve #2455. It is based on the previous work in #2456, but takes a different approach: options are loaded for the specific file in the file worker when `print-paths` is called, instead of loading them in the watchdog with a separate Lake command. This change addresses review comments made in #2456.

In doing so, it introduces two new Lake config fields: 
- `leanOptions`: `-D` flag options that are passed to both the language server and `lean` when building.
- `moreServerOptions`: `-D` flag options that are passed to the language server.

Since `print-paths` must also accept a file path to compute the options for that file, this PR is changing the API for `print-paths`. As there have been numerous complaints about the name `print-paths`, I also decided to change it to `setup-file` in this PR, since it would break compatibility with the old Lake API anyways.

This PR deprecates the Lakefile field `moreServerArgs` in favor of `moreGlobalServerArgs`, as suggested in the review for #2456.

### Updating your Lakefile

A Lakefile with the following package declaration:
```lean
def moreServerArgs := #[
  "-Dpp.unicode.fun=true"
]
def moreLeanArgs := moreServerArgs

package SomePackage where
  moreServerArgs := moreServerArgs
  moreLeanArgs := moreLeanArgs
```

... can be updated to the following package declaration to use per-package options:
```lean
package SomePackage where
  leanOptions := ⟨`pp.unicode.fun, true⟩
```

A package configuration that does not set `moreLeanArgs` can use `moreServerOptions` instead of `leanOptions`:
```lean
package SomePackage where
  moreServerOptions := ⟨`pp.unicode.fun, true⟩
```

Fixes #2455

### DIFF
--- a/src/Lean/Server/FileWorker/SetupFile.lean
+++ b/src/Lean/Server/FileWorker/SetupFile.lean
@@ -1,3 +1,8 @@
+/-
+Copyright (c) 2023 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Sebastian Ullrich, Marc Huisinga
+-/
 import Init.System.IO
 import Lean.Server.Utils
 import Lean.Util.FileSetupInfo


mhuisi:
(cc @Kha to adjust the Nix stuff)

digama0 (approved):

mhuisi:
The tests are red because of a bug in `System.Uri.fileUriToPath?`. I will investigate it later.

src/lake/Lake/CLI/Serve.lean
```diff
@@ -6,11 +6,12 @@ Authors: Mac Malone
 import Lake.Load
 import Lake.Build
 import Lake.Util.MainM
+import Lean.Util.FileSetupInfo
 
 namespace Lake
-open Lean (Json toJson fromJson? LeanPaths)
+open Lean (Json toJson fromJson? LeanPaths ServerOptions searchModuleNameOfFileName FileSetupInfo)
```
tydeu, line 12: Minor suggestion: At this point, I think just `open Lean` is fine.

tydeu (approved):
LGTM!

mhuisi:
> The tests are red because of a bug in `System.Uri.fileUriToPath?`. I will investigate it later.

It's worse: The tests used an incorrect URI format and it never caused problems before!

tydeu:
I missed something! Since you renamed `moreServerArgs`, the `math` package template needs to be updated as well:
https://github.com/mhuisi/lean4/blob/aa0774467e9ce6ab5f911bd06674593967283bbd/src/lake/Lake/CLI/Init.lean#L98-L118

Also, for temporary backwards compatibility, it might be worth it for one version to leave the `PackageConfig.moreServerArgs`  field and have `moreGlobalServerArgs` default to it. To deprecate a property like this one, I usually put a note on its deprecation in the docstring, remove the mention of it in the README, and potentially a warning issue during package config load if it is non-empty (for reference, see how I deprecated [`manifestFile`](https://github.com/leanprover/lean4/pull/2801/commits/11fd5b06923635a3593d1fa1318f0165a2f679ee)).

semorrison:
Note that the Mathlib bot reporting that PR didn't complete normally is actually a failure:

```
Run lake exe cache clean
error: ./lakefile.lean:26:2: error: 'moreServerArgs' is not a field of structure 'Lake.PackageConfig'
error: ./lakefile.lean: package configuration has errors
Error: Process completed with exit code 1.
```

If you're able to make changes to the `lean-pr-testing-2858` branch of Mathlib directory to make it compatible with these chaanges, please go ahead. Ping me if you need assistance on the Mathlib side.

src/lake/Lake/CLI/Init.lean
```diff
@@ -93,13 +93,16 @@ def mathConfigFileContents (pkgName libRoot : String) :=
...
+  ⟨`pp.unicode.fun, true⟩, -- pretty-prints `fun a ↦ b`
+  ⟨`pp.proofs.withType, false⟩
 ]
 
 -- These settings only apply during `lake build`, but not in VSCode editor.
-def moreLeanArgs := moreServerArgs
+def moreLeanArgs := #[
+  \"-Dpp.unicode.fun=true\", -- pretty-prints `fun a ↦ b`
+  \"-Dpp.proofs.withType=false\"
+]
```
semorrison, line 105: @mhuisi, is there some way we can keep these in sync?

mhuisi: I could provide a function to format `ServerOption` as a command line argument.

digama0: And every newbie user is seeing this metaprogramming code in their `lake new my-first-project math` lakefile? :sob:

mhuisi: I honestly don't care if we have the duplication in the template or not. Please just agree on either one!

src/lake/Lake/CLI/Init.lean
```diff
@@ -93,13 +93,13 @@ def mathConfigFileContents (pkgName libRoot : String) :=
...
-  \"-Dpp.unicode.fun=true\", -- pretty-prints `fun a ↦ b`
-  \"-Dpp.proofs.withType=false\"
+def moreServerOptions : Array ServerOption := #[
+  ⟨`pp.unicode.fun, true⟩, -- pretty-prints `fun a ↦ b`
+  ⟨`pp.proofs.withType, false⟩
 ]
 
 -- These settings only apply during `lake build`, but not in VSCode editor.
-def moreLeanArgs := moreServerArgs
+def moreLeanArgs := moreServerOptions.map ServerOption.asCliArg
```
digama0, line 102: Do we actually need this now? Lean should be passed these options automatically. This should be the default behavior, any option that affects the build has to be added in both places.

src/lake/Lake/Load/Package.lean
```diff
@@ -107,6 +107,10 @@ def Package.finalize (self : Package) (deps : Array Package) : LogIO Package :=
         error s!"post-update hook was defined in `{decl.pkg}`, but was registered in `{self.name}`"
     | .error e => error e
 
+  -- Deprecation warnings
+  unless self.config.moreServerArgs.isEmpty do
+    logWarning s!"{self.name}: package configuration option `moreServerArgs` is deprecated in favor of `moreGlobalServerArgs`"
```
digama0, line 112: This should suggest `moreServerOptions`, not `moreGlobalServerArgs`, since most uses of this flag in the wild as global server args are incorrect and following the suggestion will cause the same issue that this PR is trying to solve.

src/lake/Lake/CLI/Init.lean
digama0: I don't want duplication for sure, but I also want the builtin templates to not contain any dynamic features, i.e. they are expressible as JSON with fancy syntax. In this case that could be addressed by having the defaults set up such that you don't have to specify `moreLeanArgs` at all, or there is some common `buildOptions` thing that feeds into both `moreServerOptions` and `moreLeanArgs`. Copy pasting a nontrivial program into everyone's lakefile is going to make it harder to analyze lean projects at large going forward.

src/Lean/Util/ServerOptions.lean
```diff
@@ -0,0 +1,87 @@
+import Lean.Util.Paths
+
+namespace Lean
+
+/-- Restriction of `DataValue` that covers exactly those cases that Lean is able to handle when passed via the `-D` flag. -/
+inductive ServerOptionValue where
```
tydeu, line 6: Inspired by the discussion above, since these `-D` options also apply on the command, line and to builds it might be worth naming this type `CliOptionValue` or `LeanOptionValue` or `DOptionValue` instead.

src/lake/Lake/CLI/Init.lean
digama0: I'm not saying there should not be a split, but people shouldn't have to repeat themselves and a place to put the common data would be useful for the common case.

mhuisi: If both of you are happy with `leanOptions`, then I will add it.

src/lake/Lake/Config/Package.lean
```diff
@@ -310,9 +319,13 @@ namespace Package
...
-/-- The package's `moreServerArgs` configuration. -/
-@[inline] def moreServerArgs (self : Package) : Array String :=
-  self.config.moreServerArgs
+/-- The package's `moreGlobalServerArgs` configuration. -/
+@[inline] def moreGlobalServerArgs (self : Package) : Array String :=
+  self.config.moreGlobalServerArgs
+
+/-- The package's `moreServerOptions` configuration. -/
+@[inline] def moreServerOptions (self : Package) : Array LeanOption :=
+  self.config.moreServerOptions
```
tydeu, line 328:
```suggestion
  self.leanOptions ++ self.config.moreServerOptions
```

src/lake/Lake/Config/Package.lean
```diff
@@ -322,6 +335,10 @@ namespace Package
 @[inline] def backend (self : Package) : Backend :=
   self.config.backend
 
+/-- The package's `leanOptions` configuration. -/
+@[inline] def leanOptions (self : Package) : Array LeanOption :=
+  self.config.leanOptions
+
 /-- The package's `moreLeanArgs` configuration. -/
 @[inline] def moreLeanArgs (self : Package) : Array String :=
   self.config.moreLeanArgs
```
tydeu, line 344:
```suggestion
  self.leanOptions.map (·.asCliArg) ++ self.config.moreLeanArgs
```

src/lake/Lake/Config/Package.lean
```diff
@@ -322,6 +335,10 @@ namespace Package
 @[inline] def backend (self : Package) : Backend :=
   self.config.backend
 
+/-- The package's `leanOptions` configuration. -/
+@[inline] def leanOptions (self : Package) : Array LeanOption :=
+  self.config.leanOptions
```
tydeu, line 340: As noted above, these should be combined with package's other options in the package definitions themselves (which will require reordering these definitions).

src/lake/Lake/Config/LeanLib.lean
```diff
@@ -103,10 +117,17 @@ then the default (which is C for now).
...
-That is, the package's `moreLeanArgs` plus the library's  `moreLeanArgs`.
+`leanArgs` is the accumulation of:
+- the package's `leanOptions`
+- the library's `leanOptions`
+- the package's `moreLeanArgs`
+- the library's `moreLeanArgs`
 -/
 @[inline] def leanArgs (self : LeanLib) : Array String :=
-  self.pkg.moreLeanArgs ++ self.config.moreLeanArgs
+  self.pkg.leanOptions.map (·.asCliArg)
```
tydeu, line 127: The comments above should help make these two combinations a bit shorter.

tydeu (commented):
Some minor suggestions that help the code of the new options fit the style of other package configuration options.

src/lake/Lake/Config/Package.lean
mhuisi: In this case, it should probably be named `serverOptions`, though.

src/lake/Lake/Config/LeanLib.lean
```diff
@@ -87,6 +87,17 @@ Is true if either the package or the library have `precompileModules` set.
...
+/--
+The arguments to pass to `lean --server` when running the Lean language server.
+`serverOptions` is the the accumulation of:
+- the package's `leanOptions`
+- the package's `moreServerOptions`
+- the library's `leanOptions`
+- the library's `moreServerOptions`
+-/
+@[inline] def serverOptions (self : LeanLib) : Array LeanOption :=
+  self.pkg.moreServerOptions ++ self.config.moreServerOptions
```
tydeu, line 99:
```suggestion
  self.pkg.moreServerOptions ++ self.config.leanOptions ++ self self.config.moreServerOptions
```

src/lake/Lake/Config/LeanLib.lean
```diff
@@ -103,7 +114,11 @@ then the default (which is C for now).
...
 The arguments to pass to `lean` when compiling the library's Lean files.
-That is, the package's `moreLeanArgs` plus the library's  `moreLeanArgs`.
+`leanArgs` is the accumulation of:
+- the package's `leanOptions`
+- the package's `moreLeanArgs`
+- the library's `leanOptions`
+- the library's `moreLeanArgs`
 -/
 @[inline] def leanArgs (self : LeanLib) : Array String :=
   self.pkg.moreLeanArgs ++ self.config.moreLeanArgs
```
tydeu, line 124:
```suggestion
  self.pkg.moreLeanArgs ++ self.config.leanOptions.map (·.asCliArg) ++ self.config.moreLeanArgs
```

tydeu (commented):
@mhuisi I think you lost the library's `leanOptions` when addressing my suggestions.

src/lake/Lake/Config/LeanLib.lean
tydeu: @mhuisi Yes. They are combined for all derivative types (e.g., `Package`, `LeanLib`, `LeanExe`, etc.). After all, that is what this definition is doing, combining them for the `LeanLib` type. What makes `LeanLib` different from `Package` is that `Package` just needs too combine its two user-provided `config.*` options whereas `LeanLib` needs to combine its user-povided `config.*` options plus the combined option from `Package`.

mhuisi: Ok. I somewhat disagree about the derivative types being the right place to do this (I'd probably expect it on `LeanConfig` with a name that captures the notion that it is not simply e.g. the `moreLeanArgs` field from the corresponding config), but will happily implement your suggestion :)

tydeu: @mhuisi That is another design that I had not really given much thought to before! It is worth considering for a future refactor.

tydeu (approved):

tydeu:
@semorrison I noticed the mathlib bot has not posted a comment here. Do you know why?

semorrison:
> @semorrison I noticed the mathlib bot has not posted a comment here. Do you know why?

It's here: https://github.com/leanprover/lean4/pull/2858#issuecomment-1806444671.

I'll prod it into rebuilding now.

mhuisi:
@Kha decided that we do not want to adjust the Nix side for this PR. If you are using Lean 4 with Nix and stumbled upon this PR because it broke your environment, feel free to submit a PR to adjust the Nix side as well!

src/Lean/Server/FileWorker/SetupFile.lean
```diff
@@ -0,0 +1,128 @@
...
+  kind          := FileSetupResultKind.importsOutOfDate
+  srcSearchPath := ← initSrcSearchPath
+  fileOptions   := Options.empty
+}
+
+def FileSetupResult.ofError (msg : String) : IO FileSetupResult := do return {
+  kind          := FileSetupResultKind.error msg
+  srcSearchPath := ← initSrcSearchPath
+  fileOptions   := Options.empty
+}
```
Kha, line 88: I would probably have compressed these into one function (`ofUnsuccessful (kind : FileSetupResultKind)`?) but it is definitely fine either way

Kha (approved):

src/Lean/Server/FileWorker/SetupFile.lean
mhuisi: Good point!

mhuisi: Ah, I think I actually prefer it the way it is now. `ofUnsuccessful FileSetupResultKind.successful` would be strange.

Kha: Indeed it would be because for `successful` we should still use a separate function using the correct search path and options :) . Or do you mean the mere possibility of that call would be strange?

mhuisi: The possibility for the call! The idea of this "design pattern" is that as long as you use the constructor functions that the type provides instead of the primitive constructor, you get a type that fulfills its invariants. `ofUnsuccessful` would violate this.

PR #2860, opened 2023-11-10
experiment: change typeclass search order for CoeTC CoeOTC

kim-em:
I'd like to see the Mathlib benchmark for this.

### DIFF
--- a/src/Init/Coe.lean
+++ b/src/Init/Coe.lean
@@ -139,7 +139,7 @@ class CoeTC (α : Sort u) (β : Sort v) where
   coe : α → β
 attribute [coe_decl] CoeTC.coe
 
-instance [Coe β γ] [CoeTC α β] : CoeTC α γ where coe a := Coe.coe (CoeTC.coe a : β)
+instance [CoeTC α β] [Coe β γ] : CoeTC α γ where coe a := Coe.coe (CoeTC.coe a : β)
 instance [Coe α β] : CoeTC α β where coe a := Coe.coe a
 instance : CoeTC α α where coe a := a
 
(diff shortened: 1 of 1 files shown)


src/Init/Coe.lean
```diff
@@ -139,7 +139,7 @@ class CoeTC (α : Sort u) (β : Sort v) where
   coe : α → β
 attribute [coe_decl] CoeTC.coe
 
-instance [Coe β γ] [CoeTC α β] : CoeTC α γ where coe a := Coe.coe (CoeTC.coe a : β)
+instance [CoeTC α β] [Coe β γ] : CoeTC α γ where coe a := Coe.coe (CoeTC.coe a : β)
```
gebner, line 142: FYI, this change shouldn't have any effect on the search order at all.  Since #2174, the order in which arguments to typeclass instances are processed is determined automatically by looking at which arguments are marked as `outParam`/`semiOutParam`.  (And the order is determined at instance declaration time.  You can see it with `set_option Meta.synthOrder true`.)

There are two arguments here that we could process first, `CoeTC α ?β` and `Coe ?β γ`.  Recall that a *non*-outparam argument is assumed not to contain assignable metavariables, so `α` and `γ` won't have mvars.  Out of the two choices, we can't process `CoeTC α ?β` first because the second argument will always be a metavariable (and the spec for `CoeTC` says that both arguments must be mvar-free).  So the first one will always be `Coe ?β γ` (where the first argument may be a metavariable because it is marked as `semiOutParam`).

semorrison: Thanks. I'll write up soon the very unfortunate typeclass search disaster that prompted me to look at this!

PR #2861, opened 2023-11-10
chore: fix commit used for PR release

kim-em:
We were checking out the synthetic merge commit between the PR `HEAD` and `master`, and this was then breaking the logic to determine which nightly-testing branches to use in Mathlib and Std.


nomeata:
GitHub's off defaults strike again… goof catch!

PR #2864, opened 2023-11-11
feat: `pp.beta` to apply beta reduction when pretty printing

### LINKED ISSUE #715: Add option `pp.beta`
When `pp.beta` is set to `true`, the pretty-printer (delaborator) should apply beta reduction.
Lean 3 implements this feature, but we had *not* implemented it yet in Lean 4.

kmill:
This was a Lean 3 pretty printer option. While this pretty printer option tends to lead to confusing situations when set, it has been frequently requested. [It is possible](https://github.com/leanprover-community/mathlib4/pull/7910) to implement this pretty printer option as a user, but it comes with some artifacts -- for instance, expressions in hovers are not beta reduced. Adding this as a core pp option is cleanest.

(We should consider having hooks into the tactic evaluator to allow users to transform the tactic state between tactics. This would enable beta reducing the entire local context for real, which would be useful for teaching.)

Closes #715

### DIFF
--- a/src/Lean/PrettyPrinter/Delaborator/Basic.lean
+++ b/src/Lean/PrettyPrinter/Delaborator/Basic.lean
@@ -277,6 +277,17 @@ end Delaborator
 open SubExpr (Pos PosMap)
 open Delaborator (OptionsPerPos topDownAnalyze)
 
+/-- Custom version of `Lean.Core.betaReduce` to beta reduce expressions for the `pp.beta` option.
+We do not want to beta reduce the application in `let_fun` annotations. -/
+private partial def betaReduce' (e : Expr) : CoreM Expr :=
+  Core.transform e (pre := fun e => do
+    if isLetFun e then
+      return .done <| e.updateMData! (.app (← betaReduce' e.mdataExpr!.appFn!) (← betaReduce' e.mdataExpr!.appArg!))
+    else if e.isHeadBetaTarget then
+      return .visit e.headBeta
+    else
+      return .continue)
+
 def delabCore (e : Expr) (optionsPerPos : OptionsPerPos := {}) (delab := Delaborator.delab) : MetaM (Term × PosMap Elab.Info) := do
   /- Using `erasePatternAnnotations` here is a bit hackish, but we do it
      `Expr.mdata` affects the delaborator. TODO: should we fix that? -/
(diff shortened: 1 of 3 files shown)


leodemoura:
@kmill This is a good PR. BTW, I want to change how we encode `let_fun` and use a more explicit encoding that uses an explicit auxiliary function instead of just `Expr.mdata`. The auxiliary function would be:
```lean
def letFun {A : Type u} {B : Type v} (x : A) (f : A → B) := f x
```

kmill:
@leodemoura I just brought this PR out of "draft" status (the "break mathlib" appears to be a false alarm). Should we wait at all for any changes to `let_fun`? Or is that a project suggestion? 😃

leodemoura:
> Should we wait at all for any changes to let_fun? 

No.

> Or is that a project suggestion? 😃

Yes, I will be really happy if you make it happen. :) 
The new encoding with the auxiliary function `letFun` is more robust.
I am hoping the new encoding will not break too many things.

semorrison (approved):

src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -277,6 +277,17 @@ end Delaborator
 open SubExpr (Pos PosMap)
 open Delaborator (OptionsPerPos topDownAnalyze)
 
+/-- Custom version of `Lean.Core.betaReduce` to beta reduce expressions for the `pp.beta` option.
+We do not want to beta reduce the application in `let_fun` annotations. -/
+private partial def betaReduce' (e : Expr) : CoreM Expr :=
+  Core.transform e (pre := fun e => do
+    if isLetFun e then
+      return .done <| e.updateMData! (.app (← betaReduce' e.mdataExpr!.appFn!) (← betaReduce' e.mdataExpr!.appArg!))
```
Kha, line 285: Mmh, is that different from `.continue`?

Kha (approved):

src/Lean/PrettyPrinter/Delaborator/Basic.lean
kmill: `.continue` would visit the mdata value again, but here we want to skip visiting the `.app` itself and jump into its arguments. (It would be nice if `Core.transform` had a way to do these finer-grained recursions.)

kmill: Yeah, "mdata value" was supposed to mean the `.app` itself, which is the immediate subterm of the mdata.

PR #2866, opened 2023-11-12
fix: have parenthesizer copy source info to parenthesized term

### LINKED ISSUE #1854: Widgets: Mousing over parentheses should highlight the term inside the parentheses
### Description

When mousing a parenthesis in an output window, I would like that only the term inside those parentheses is highlighted, not the term containing the parentheses.

### Steps to Reproduce

In 
```lean
theorem foo (a b c : Nat) : a + (b + c) = a + (b + c) := rfl
```
Navigate to the `rfl` and hover your mouse over any parenthesis in the output view

**Expected behavior:** Only `(b + c)` is highlighted and you receive a pop-up for that term. 

**Actual behavior:** `a + (b + c)` is highlighted.

The expected behavior is better. On large terms, it can be hard to find the infix that corresponds to the expression inside the parentheses.

kmill:
The effect of this change is that hovering your mouse over a parenthesis in the info view in VS Code highlights the entire parenthesized expression, rather than the next expression outwards.

Closes #1854

### DIFF
--- a/src/Lean/PrettyPrinter/Parenthesizer.lean
+++ b/src/Lean/PrettyPrinter/Parenthesizer.lean
@@ -340,8 +340,18 @@ def parserOfStack.parenthesizer (offset : Nat) (_prec : Nat := 0) : Parenthesize
 
 @[builtin_category_parenthesizer term]
 def term.parenthesizer : CategoryParenthesizer | prec => do
-  maybeParenthesize `term true (fun stx => Unhygienic.run `(($(⟨stx⟩)))) prec $
+  maybeParenthesize `term true wrapParens prec $
     parenthesizeCategoryCore `term prec
+where
+  /-- Wraps the term `stx` in parentheses and then copies its `SourceInfo` to the result.
+  The purpose of this is to copy synthetic delaborator positions from the `stx` node to the parentheses node,
+  which causes the info view to view both of these nodes as referring to the same expression.
+  If we did not copy info, the info view would consider the parentheses to belong to the outer term.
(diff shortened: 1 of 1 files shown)


kmill:
@Kha I've only tested this by hand in a few places, and I'm a little uncertain about whether `SourceInfo.fromRef` is the right choice. However, it seems to work (mouseovers in the info view on "(" or ")" highlight the entire "( ... )" expression and show the right type).

Is this something we can write tests for?

Kha:
Nice find. I think using `withRef stx` inside `Unhygienic.run` should have the same effect?

Kha:
> Is this something we can write tests for?

Unfortunately we don't have any widget tests at all afaik

kmill:
I had tried `withRef stx` first, but that puts position information on the parenthesis tokens themselves, which makes it so each parenthesis highlights on its own.

It's sort of silly copying the source info for level and tactic expressions right now, since they don't get synthetic positions as far as I know, but if they ever do then this change will probably make parenthesis highlighting will work for them.

Kha:
Ok, makes sense but we should document that this is copying the synthetic delaborator position such that the info view will consider the two syntax nodes the same.

Kha:
And specifically what is fixed, i.e. that without copying the info, the info view would consider the parentheses to belong to the outer term

kmill:
@Kha Done. I also limited the changes to just the term parenthesizer, since levels and tactics don't use synthetic source positions, and they don't have any subexpression support in the info view.

Kha:
Perfect!

PR #2870, opened 2023-11-13
fix: re-read HTTP header when skipping notification in `Ipc.readResponseAs`

Kha:
No tests as this is purely testing code; starting breaking with the new snapshot reporter

### DIFF
--- a/src/Lean/Data/Lsp/Ipc.lean
+++ b/src/Lean/Data/Lsp/Ipc.lean
@@ -62,8 +62,23 @@ def readMessage : IpcM JsonRpc.Message := do
 def readRequestAs (expectedMethod : String) (α) [FromJson α] : IpcM (Request α) := do
   (←stdout).readLspRequestAs expectedMethod α
 
-def readResponseAs (expectedID : RequestID) (α) [FromJson α] : IpcM (Response α) := do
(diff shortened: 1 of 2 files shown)


src/Lean/Data/Lsp/Ipc.lean
```diff
@@ -62,8 +62,20 @@ def readMessage : IpcM JsonRpc.Message := do
 def readRequestAs (expectedMethod : String) (α) [FromJson α] : IpcM (Request α) := do
   (←stdout).readLspRequestAs expectedMethod α
 
-def readResponseAs (expectedID : RequestID) (α) [FromJson α] : IpcM (Response α) := do
-  (←stdout).readLspResponseAs expectedID α
+/-- Reads response, discarding notifications in between (as this is purely for testing). -/
+partial def readResponseAs (expectedID : RequestID) (α) [FromJson α] :
```
Kha, line 66: Could also be renamed if desired

mhuisi: I'm O.K. with keeping the name if the doc comment describes why discarding notifications is desirable for testing.

mhuisi (approved):

PR #2879, opened 2023-11-15
chore: fix conditional syntax in pre-release.yml

kim-em:
An error in the syntax was result in some tasks in `pr-release.yml` always running, given nonsense errors (and creating unwanted branches at Mathlib).


tydeu:
Ah yes, this a error I have made many times when writing GitHub action workflows! 😭

PR #2874, opened 2023-11-14
feat: Guess lexicographic order for well-founded recursion

nomeata:
This improves Lean’s capabilities to guess the termination measure for well-founded
recursion, by also trying lexicographic orders.  For example:

    def ackermann (n m : Nat) := match n, m with
      | 0, m => m + 1
      | .succ n, 0 => ackermann n 1
      | .succ n, .succ m => ackermann n (ackermann (n + 1) m)

now just works.

The module docstring of `Lean.Elab.PreDefinition.WF.GuessLex` tells the technical story.
Fixes #2837

PR #2882, opened 2023-11-15
feat: Add MatcherApp. and CasesOnApp.refineThrough

nomeata:
these are compagnions to `MatcherApp.addArg` and `CasesOnApp.addArg` when one only has an
expression (which may not be a type) to transform, but not a concret values.

This is a prerequisite for guessing lexicographic order (#2874). Keeping this on a separate PR because it’s sizable, and has a clear independent specification.

### DIFF
--- a/src/Lean/Meta/CasesOn.lean
+++ b/src/Lean/Meta/CasesOn.lean
@@ -171,7 +171,7 @@ def CasesOnApp.refineThrough (c : CasesOnApp) (e : Expr) : MetaM (Array Expr) :=
           -- and abstract over the parameters of the alternatives, so that we can safely pass the Expr out
           Expr.abstractM body fvs
 
-/-- A non-failing version of `CasesOnApp.transform` -/
+/-- A non-failing version of `CasesOnApp.refineThrough` -/
 def CasesOnApp.refineThrough? (c : CasesOnApp) (e : Expr) :
     MetaM (Option (Array Expr)) :=
   try
--- a/src/Lean/Meta/Match/Match.lean
+++ b/src/Lean/Meta/Match/Match.lean
@@ -1023,7 +1023,7 @@ def MatcherApp.refineThrough (matcherApp : MatcherApp) (e : Expr) : MetaM (Array
           -- and abstract over the parameters of the alternatives, so that we can safely pass the Expr out
           Expr.abstractM body fvs
 
-/-- A non-failing version of `MatcherApp.transform` -/
+/-- A non-failing version of `MatcherApp.refineThrough` -/
 def MatcherApp.refineThrough? (matcherApp : MatcherApp) (e : Expr) :
     MetaM (Option (Array Expr)) :=
   try


nomeata:
These two functions are a big chunk of #2874, and also the chunk where I am least confident about the implementation, so I figured it makes sense to PR and review them in isolation.

I’m somewhat confident that their specification and usefulness is ok.

src/Lean/Meta/CasesOn.lean
```diff
@@ -106,11 +111,71 @@ where
...
+  with `ys_i` as loose bound variables, ready to be `.instantiateRev`d; arity according to `CasesOnApp.altNumParams`.
+
+  This is similar to `CasesOnApp.addArg` when you only have an expression to
+  refined, and not a type with a value.
+
+  This is used in in `Lean.Elab.PreDefinition.WF.GuessFix` when constructing the context of recursive
+  calls to refine the functions' paramter, which may mention `major`.
+  See there for how to use this function.
+-/
+def CasesOnApp.transform (c : CasesOnApp) (e : Expr) : MetaM (Array Expr) :=
```
nomeata, line 139: TODO: Explain why this returns an open term, rather than accept an continuation.

src/Lean/Meta/CasesOn.lean
```diff
@@ -106,11 +111,71 @@ where
...
     return some (← c.addArg arg checkIfRefined)
   catch _ =>
     return none
 
+/--
+  Given a `casesOn` application `c` of the form
+  ```
+  casesOn As (fun is x => motive[is, xs]) is major  (fun ys_1 => (alt_1 : motive (C_1[ys_1])) ... (fun ys_n => (alt_n : motive (C_n[ys_n]) remaining
+  ```
+  and an expression `B[is, major]`, for every alternative `i`, construct the type
```
nomeata, line 126:
```suggestion
  and an expression `B[is, major]` (which may not be a type, e.g. `n : Nat`), for every alternative `i`, construct the type
```

src/Lean/Meta/CasesOn.lean
```diff
@@ -106,11 +111,71 @@ where
...
+    trace[Elab.definition.wf] "CasesOnApp.transform: {indentExpr e}"
+    unless motiveArgs.size == c.indices.size + 1 do
+      throwError "failed to transfer argument through `casesOn` application, motive must be lambda expression with #{c.indices.size + 1} binders"
+    let discrs := c.indices ++ #[c.major]
+    let mut eAbst := e
+    for motiveArg in motiveArgs.reverse, discr in discrs.reverse do
+      eAbst ← kabstract eAbst discr
+      eAbst := eAbst.instantiate1 motiveArg
+    -- Let's create something that’s a `Sort` and mentions `e`, by writing `e = e`,
+    -- so that we can use it as a motive
```
nomeata, line 150:
```suggestion
    -- Let's create something that’s a `Sort` (recall that `e` possibly is not a type) and mentions `e`, by writing `e = e`,
    -- so that we can use it as a motive
```

PR #2887, opened 2023-11-15
feat: set_option for conv mode

lecopivo:
Setting options with `set_option ... in ...` was only possible in tactic mode. This PR enables it to do it in conv mode too. 

The only issue with my fix is that the syntax definition is in `src/Lean/Elab/Tactic/Conv/Basic.lean` which is clearly a bad place for it. Ideally it should be in `src/Lean/Parser/Command.lean` just next to the definition of `set_option` tactic but I do not know how to write the parser with `convSeq`

### DIFF
--- a/src/Lean/Elab/Tactic/Conv/Basic.lean
+++ b/src/Lean/Elab/Tactic/Conv/Basic.lean
@@ -187,3 +187,17 @@ private def convLocalDecl (conv : Syntax) (hUserName : Name) : TacticM Unit := w
   Tactic.evalFirst
 
 end Lean.Elab.Tactic.Conv
+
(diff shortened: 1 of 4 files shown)


digama0:
You need
```lean
  register_parser_alias                                                 optionValue
```
around [here](https://github.com/leanprover/lean4/blob/7cc2c9f1c9d454efaff1c3d236fff1b9205548b6/src/Lean/Parser/Command.lean#L316) and then (after a stage0 update)
```lean
syntax (name := «set_option») "set_option " ident ppSpace optionValue " in " convSeq : conv
```
in `Init/Conv.lean` to define the syntax here.

kim-em:

PR #2890, opened 2023-11-16
feat: bare `lake init` & validated pkg names

### LINKED ISSUE #2637: validate project names
### Description

Creating a new project called `lean` (`lake init lean`) kinda works, except the built artifact from `lake build` segfaults. Any other name appears to work fine.

This command should validate that the name is not "lean" and is alphanumeric. I also expected `.` to use the current directory's name, but instead it created a new project literally called `.`

### Steps to Reproduce

1. `lake init lean`
2. `lake build`
3. `./build/.../lean`
4. Segmentation fault

**Expected behavior:** I shouldn't be allowed to make a project called lean, or this shouldn't segfault.

**Actual behavior:** I am allowed to make a project called lean, and it segfaults when built.

### Versions

[Output of `lean --version` in the folder that the issue occured in] - Lean (version 4.1.0, commit a832f398b80a, Release)
[OS version] - Sonoma 14.0

tydeu:
Closes #2637. 

As a new feature, `lake init .` or a bare `lake init` will now create a package with the current directory's name.

Will also validate that a package name:
* Is not one of `Init`, `Lean`, `Lake`, or `Main` (case-insensitive)
* Is not all `.` or all whitespace
* Does not contain `/` or `\`.

### DIFF
--- a/src/lake/Lake/CLI/Init.lean
+++ b/src/lake/Lake/CLI/Init.lean
@@ -233,10 +233,28 @@ def initPkg (dir : FilePath) (name : String) (tmp : InitTemplate) (env : Lake.En
     else
       logWarning "failed to initialize git repository"
 
-def init (pkgName : String) (tmp : InitTemplate) (env : Lake.Env) : LogIO PUnit :=
-  initPkg "." pkgName tmp env
+def validatePkgName (pkgName : String) : LogIO PUnit := do
+  if pkgName.isEmpty || pkgName.all (· == '.') || pkgName.any (· ∈ ['/', '\\']) then
+    error "illegal package name"
+  if pkgName.toLower ∈ ["init", "lean", "lake", "main"] then
+    error "reserved package name"
 
-def new (pkgName : String) (tmp : InitTemplate) (env : Lake.Env) : LogIO PUnit := do
-  let dirName := pkgName.map fun chr => if chr == '.' then '-' else chr
-  IO.FS.createDir dirName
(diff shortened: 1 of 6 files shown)


src/lake/Lake/CLI/Init.lean
```diff
@@ -233,10 +233,28 @@ def initPkg (dir : FilePath) (name : String) (tmp : InitTemplate) (env : Lake.En
...
-  initPkg "." pkgName tmp env
-
-def new (pkgName : String) (tmp : InitTemplate) (env : Lake.Env) : LogIO PUnit := do
-  let dirName := pkgName.map fun chr => if chr == '.' then '-' else chr
-  IO.FS.createDir dirName
+def validatePkgName (pkgName : String) : LogIO PUnit := do
+  if pkgName.isEmpty || pkgName.all (· == '.') || pkgName.any (· ∈ ['/', '\\']) then
+    error "illegal package name"
+  if pkgName.toLower ∈ ["init", "lean", "lake", "main"] then
+    error "reserved package name"
```
semorrison, line 240: I am tempted to say "some validation is better than none, but couldn't we restrict to `[0-9A-Za-z-_]*`?, but won't. :-)

semorrison (approved):

src/lake/Lake/CLI/Init.lean
tydeu: @semorrison My current reason for avoiding this is twofold. We might wish to permit non-English or Mathematical characters in package names for broader appeal. Second, this validation currently only applies to the `init`/`new` not to the package declaration itself, so any rules here can currently be circumvented. In the future, though, have a more restricted subset for package names in general would likely be a good idea.

Issue #2628, opened 2023-10-05
(kernel) unknown constant in mutually recursive definition


nomeata:
### Description

In some (possibly not-well-founded) mutually recursive definitions, we get `(kernel) unknown constant`.

### Context

I stumbled over this while [trying to help someone on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/induction.20for.20mutually.20inductive.20types/near/395089656).

### Steps to Reproduce

MWE:

```
mutual
  inductive Thing : Type
    | Bag : Object → Thing
  inductive Object : Type
    | Bike : Object
    | Box : Thing -> Object
end

mutual
  def sizeOfThing : Thing → Nat
    | .Bag o => 1 + sizeOfObject o
  def sizeOfObject : Object → Nat
    | .Bike => 42
    | .Box t => 1 + sizeOfThing t
end

axiom Nat.add_mod (a b n : Nat) : (a + b) % n = ((a % n) + (b % n)) % n

mutual
theorem aboutThings : ∀ t, sizeOfThing t % 2 = 1
  | .Bag o => by
    unfold sizeOfThing
    rw [Nat.add_mod, aboutObjects /- o-/]
    rfl

theorem aboutObjects : ∀ o, sizeOfObject o % 2 = 0
  | .Bike => rfl
  | .Box t => by
    unfold sizeOfObject
    rw [Nat.add_mod, aboutThings t]
    rfl
end
decreasing_by sorry
```

This prints
```
(kernel) unknown constant 'aboutObjects'
declaration uses 'sorry'
```

The kernel error goes away if I use `aboutObjects o` in the call to `rw`, i.e. with an explicit argument.

**Expected behavior:** No kernel errors.

**Actual behavior:** The kernel prints an error

### Versions

```
~/build/lean/mathlib4 $ lean --version
Lean (version 4.2.0-rc1, commit a62d2fd49796, Release)
```

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

nomeata:
Smaller reproducer:
```
opaque myId : (Nat → Nat) → (Nat → Nat) := id

mutual
def foo : Nat → Nat
  | .zero => 0
  | .succ n => (myId bar) n
def bar : Nat → Nat
  | .zero => 0
[...]

PR #2892, opened 2023-11-16
fix: PackMutual: Deal with extra arguments

### LINKED ISSUE #2883: unknown constant in mutually recursive definition
### Description

In some mutually recursive definitions, we get `unknown constant`

### Context

This [was reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/.22unknown.20constant.22.20in.20recursive.20mutual.20def/near/402277011).

### Steps to Reproduce

```
mutual
  inductive A : Type
  | baseA  : A
  | fromB : B → A

  inductive B : Type
  | fromA  : A → B
end

mutual
  def foo : B → Prop    -- unknown constant 'bar'
  | .fromA a => bar a 0

  def bar : A → Nat → Prop
  | .baseA   => (fun _ => True)
  | .fromB b => (fun (c : Nat) => ∃ (t : Nat), foo b)
end
termination_by
  foo x => sizeOf x
  bar x => sizeOf x
```

### Versions

4.3.0-rc1 on https://live.lean-lang.org/

### Additional Information

This may or may not be the same or related to https://github.com/leanprover/lean4/issues/2628.

### Impact

nomeata:
previously, it would ignore a recursive call that has extra arguments,
which can happen when the recursive functions return something of
function type. Therefore just leave them extra arguments in place.

Fixes #2883.

### DIFF
--- a/src/Lean/Elab/PreDefinition/WF/PackMutual.lean
+++ b/src/Lean/Elab/PreDefinition/WF/PackMutual.lean
@@ -95,7 +95,7 @@ private partial def packValues (x : Expr) (codomain : Expr) (preDefValues : Arra
   See: `packMutual`
 -/
 private partial def post (fixedPrefix : Nat) (preDefs : Array PreDefinition) (domain : Expr) (newFn : Name) (e : Expr) : MetaM TransformStep := do
-  if e.getAppNumArgs != fixedPrefix + 1 then
+  if e.getAppNumArgs < fixedPrefix + 1 then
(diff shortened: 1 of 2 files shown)


nomeata:
NB: This does _not_ fix #2628, which it separate, but related: There, the recursive call seems to be _missing_ an argument. But that should probably be handled by `packDomain` by eta-expanding first, or simply failing gracefully

leodemoura (approved):

PR #2900, opened 2023-11-17
feat: make `mkApp`, `mkApp2`, ..., `mkApp10` have `@[match_pattern]` attribute

kmill:
Give n-ary `Expr.app` constructors such as `mkApp2`, `mkApp3`, ..., `mkApp10` the `@[match_pattern]` attribute so that it is easier to read and write pattern matching for applications.


kmill:
The build failed because some numbering has subtly changed in one test. I suspect it's changing the evaluation order subtly somewhere, and I'll keep this as a draft until I work that out (or delete it if I can't).

digama0:
It appears that adding
```lean
def mkAppN' (f : Expr) (args : Array Expr) : Expr :=
  args.foldl mkApp f

@[inherit_doc mkAppN']
macro_rules
  | `(mkAppN' $f #[$xs,*]) => show MacroM Term from xs.getElems.foldlM (fun x e => `(Expr.app $x $e)) f
```
alone already accounts for the `inst.71` -> `inst.74` name change. So most likely this is due to additional code that runs at initialization time caused by adding this macro to the environment.

I think this is harmless and it is okay to just accept the change in test output.

kmill:
@digama0 Thanks for investigating. I think I'll cook up an evaluation order test just to be safe and move forward with this PR. I completely expect the macro to be equivalent, but I want to validate that it interacts with `do` arrow lifting correctly.

digama0:
> @digama0 Thanks for investigating. I think I'll cook up an evaluation order test just to be safe and move forward with this PR. I completely expect the macro to be equivalent, but I want to validate that it interacts with `do` arrow lifting correctly.

Note that the evaluation order is probably *not* preserved, but this is because [long list literals have a bonkers evaluation order](https://github.com/leanprover/lean4/blob/cddc8089bc736a1532d6092f69476bd2d205a9eb/src/Init/Data/List/Basic.lean#L13-L24) and `.app` chains will not do that.

kmill:
@semorrison I've got this PR ready for review now.

Something I'm not sure about is whether the macro should do anything special for long `.app` chains. For example, it could use `mkApp2` through `mkApp10` to reduce expression depth if that's somehow necessary, but in the meantime I went with the simplest implementation.

digama0:
Using `mkApp10` doesn't help reduce expression depth, because the first argument is still the most deeply nested. It would need to take the `f` argument last if you wanted to reduce depth by ~10x.

I don't think it's a major concern in any case, I haven't seen any examples of code using `mkApp` with more than about 7 arguments anyway. Most of the larger examples have moved to Qq for expr construction anyway.

Kha:
The macro looks a bit concerning to me, as a macro it cannot use correct scoping rules, and what does running it for every application mean for performance?

digama0:
I am also a bit dubious of the use of this macro in pattern position. Examples I have seen of it like [this](https://github.com/leanprover/std4/blob/a132de4bc4f8c7f4c0dc72745018d6b590664b5d/Std/Tactic/Omega/Frontend.lean#L285-L323) will end up generating a ton of `Expr.casesOn` applications; it's usually more efficient to use array patterns with a pattern match on `getAppFnArgs` instead.

kmill:
@semorrison Sebastian and I decided that we won't make `mkAppN` a macro, but we can at least make various `mkApp*` constructors have the `@[match_pattern]` attribute, which can serve the same purpose.

digama0:
> The macro looks a bit concerning to me, as a macro it cannot use correct scoping rules,

Not disagreeing with @kmill 's last message, but wouldn't scoping work out if you used `scoped macro_rules` instead? The macro would only apply when you have `open Lean` which is the same condition under which it would be legal to write `mkAppN` unqualified.

Kha:
That would be close, yes. It still doesn't give you `Lean.mkAppN`. Which really should be `Lean.Expr.mkAppN` anyway.

nomeata (approved):
I’d find this useful. Is there anything wrong with the `mkApp5` `match_pattern` or should we merge this?

PR #2902, opened 2023-11-17
fix: `PackMutual`: Eta-Expand as needed

### LINKED ISSUE #2628: (kernel) unknown constant in mutually recursive definition

nomeata:
The `packMutual` code ought to reliably replace all recursive calls to
the functions in `preDefs`, even when they are under- or over-applied.
Therefore eta-expand if need rsp. keep extra arguments around.

Needs a tweak to `Meta.transform` to avoid mistaking the `f` in
`f x1 x2` as a zero-arity application.

Includes a test case.

This fixes #2628 and #2883.

### DIFF
--- a/src/Lean/Elab/PreDefinition/WF/PackMutual.lean
+++ b/src/Lean/Elab/PreDefinition/WF/PackMutual.lean
@@ -93,6 +93,8 @@ private partial def packValues (x : Expr) (codomain : Expr) (preDefValues : Arra
 /--
   Pass the first `n` arguments of `e` to the continuation, and apply the result to the
   remaining arguments. If `e` does not have enough arguments, it is eta-expanded as needed.
+
+  Unlike `Meta.etaExpand` does not use `withDefault`.
 -/
 def withAppN (n : Nat) (e : Expr) (k : Array Expr → MetaM Expr) : MetaM Expr := do
   let args := e.getAppArgs
@@ -100,13 +102,12 @@ def withAppN (n : Nat) (e : Expr) (k : Array Expr → MetaM Expr) : MetaM Expr :
     let e' ← k args[:n]
     return mkAppN e' args[n:]
   else
-    withDefault do -- TODO: Copied from `etaExpand`. Needed? Harmful?
-      let missing := n - args.size
-      forallBoundedTelescope (← inferType e) missing fun xs _ => do
-        if xs.size < missing then
-          throwError "Failed to eta-expand partial application"
-        let e' ← k (args ++ xs)
-        mkLambdaFVars xs e'
+    let missing := n - args.size
+    forallBoundedTelescope (← inferType e) missing fun xs _ => do
+      if xs.size < missing then
+        throwError "Failed to eta-expand partial application"
+      let e' ← k (args ++ xs)
+      mkLambdaFVars xs e'
 
 /--
   Auxiliary function for replacing nested `preDefs` recursive calls in `e` with the new function `newFn`.
--- a/tests/lean/run/issue2628.lean
+++ b/tests/lean/run/issue2628.lean
@@ -70,3 +70,39 @@ termination_by foo n m => (n,m); bar n m => (n,m)
 decreasing_by sorry
 
 end Ex4
+
+-- Check that eta-expansion works even if the function does not
+-- have a function type
+namespace Ex5
+def FunType := Nat → Nat
+
+mutual
+def foo : FunType
+  | .zero => 0
+  | .succ n => (id bar) n
+def bar : Nat → Nat
+  | .zero => 0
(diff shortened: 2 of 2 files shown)


nomeata:
Builds upon #2892. Might be worth to merge #2892 first for bisectability.

src/Lean/Elab/PreDefinition/WF/PackMutual.lean
```diff
@@ -90,34 +90,51 @@ private partial def packValues (x : Expr) (codomain : Expr) (preDefValues : Arra
...
+    let e' ← k args[:n]
+    return mkAppN e' args[n:]
+  else
+    withDefault do -- TODO: Copied from `etaExpand`. Needed? Harmful?
+      let missing := n - args.size
+      forallBoundedTelescope (← inferType e) missing fun xs _ => do
+        if xs.size < missing then
+          throwError "Failed to eta-expand partial application"
+        let e' ← k (args ++ xs)
+        mkLambdaFVars xs e'
```
nomeata, line 109:
```suggestion
/--
  Pass the first `n` arguments of `e` to the continuation, and apply the result to the
  remaining arguments. If `e` does not have enough arguments, it is eta-expanded as needed.
  
  Unlike `etaExpand` does not turn on `withDefault`; take care when using outside the current module.
-/
def withAppN (n : Nat) (e : Expr) (k : Array Expr → MetaM Expr) : MetaM Expr := do
  let args := e.getAppArgs
  if n ≤ args.size then
    let e' ← k args[:n]
    return mkAppN e' args[n:]
  else
    let missing := n - args.size
    forallBoundedTelescope (← inferType e) missing fun xs _ => do
      if xs.size < missing then
        throwError "Failed to eta-expand partial application"
      let e' ← k (args ++ xs)
      mkLambdaFVars xs e'
```

nomeata: TODO: write test with non-function-type at the function’s type

src/Lean/Meta/Transform.lean
```diff
@@ -73,12 +73,18 @@ namespace Meta
 
 /--
   Similar to `Core.transform`, but terms provided to `pre` and `post` do not contain loose bound variables.
-  So, it is safe to use any `MetaM` method at `pre` and `post`. -/
+  So, it is safe to use any `MetaM` method at `pre` and `post`.
+
+  If `constIsApp := true`, then for an expression `mkAppN (.const f) args`, the subexpression
+  `.const f` is not visited again. Put differently: every `.const f` is visited once, with its
+  arguments.
+ -/
```
nomeata, line 81:
```suggestion
  So, it is safe to use any `MetaM` method at `pre` and `post`.

  If `skipConstInApp := true`, then for an expression `mkAppN (.const f) args`, the subexpression
  `.const f` is not visited again. Put differently: every `.const f` is visited once, with its
  arguments.
 -/
```

nomeata:
I’ll get #2926 in first, and then merge this.

PR #2903, opened 2023-11-17
fix: segfault in old compiler due to noConfusion assumptions 

### LINKED ISSUE #2901: Segfault in old compiler
### Prerequisites

### Description

The lean server crashes when type checking a function. A similar function does not lead to a crash

The issue was discussed on [zulip](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/bug.20.3F) where it was analyzed as 

> Mario Carneiro :
this seems like a rather severe issue, it is a SIGSEGV which normally shouldn't happen even if there was a bug in the elaborator
> 
> it seems that the compiler had difficulty determining the arity of a constant so it threw an error, but it messed up the error payload and it crashes while trying to print the error

### Steps to Reproduce

Typechecking the following code crashes the Lean Server.

```lean
import Mathlib.Data.Vector

inductive HVect : (n : Nat) -> (Vector (Type v) n) -> Type (v+1)  where
   | Nil  : HVect 0 ⟨ [], simp ⟩
   | Cons : (t : Type v) -> (x : t) -> HVect n ⟨ts, p⟩ -> HVect (n+1) ⟨t::ts, by simp [p]⟩


def printHOK (v : HVect (n+1) ⟨String::ts, p'⟩) : String :=
   match v with
   | HVect.Cons _ x _ => (x : String)


def printHKO (v : HVect (n+1) ⟨String::ts, p'⟩) : String :=
   match v with
   | HVect.Cons _ x _ => "Hi"
```

**Expected behavior:** `x` is of type `String` so we should be able to substitute an actual string for it.

**Actual behavior:** at the lean server level we have 

> LEAN ASSERTION VIOLATION
> File: /Users/runner/work/lean4/lean4/src/runtime/io.cpp
> Line: 92
> fclose(static_cast<FILE *>(h)) == 0
> 'unreachable' code was reached

### Versions

4.3.0-rc2 

I think previous versions are concerned. checking now.

### Impact

PS : I would love to Teams/Zoom to see you work on that issue.

digama0:
This fixes #2901, a bug in the old compiler which causes a segfault. The issue is that when visiting `noConfusion` applications, it assumes that each constructor case has `nfields` arguments, e.g. `head1 = head2 -> tail1 = tail2 -> P` has two arguments because `List.cons` has 2 fields, but in fact propositional fields are skipped by the noConfusion type generator, so for example `Subtype.noConfusionType` is:
```lean
@[reducible] protected def Subtype.noConfusionType.{u_1, u} : {α : Sort u} →
  {p : α → Prop} → Sort u_1 → Subtype p → Subtype p → Sort u_1 :=
fun {α} {p} P v1 v2 ↦
  Subtype.casesOn v1 fun val property ↦ Subtype.casesOn v2 fun val_1 property ↦ 
    (val = val_1 → P) → P
```
where `val = val_1 → P` only has the one argument even though `Subtype.mk` has two fields, presumably because it is useless to have an equality of propositions. Unfortunately there isn't any easy cache or getter to use here to get the number of non-propositional fields, so we just calculate it on the spot.

### DIFF
--- a/src/library/compiler/lcnf.cpp
+++ b/src/library/compiler/lcnf.cpp
@@ -266,6 +266,23 @@ public:
             return e;
     }
 
+    unsigned get_constructor_non_prop_nfields(name ctor, unsigned nparams) {
+        local_ctx lctx;
+        expr type = env().get(ctor).get_type();
+        for (unsigned i = 0; i < nparams; i++) {
+            lean_assert(is_pi(type));
+            expr local = lctx.mk_local_decl(ngen(), binding_name(type), binding_domain(type), binding_info(type));
+            type = instantiate(binding_body(type), local);
+        }
+        unsigned nfields = 0;
+        while (is_pi(type)) {
+            if (!type_checker(m_st, lctx).is_prop(binding_domain(type))) nfields++;
+            expr local = lctx.mk_local_decl(ngen(), binding_name(type), binding_domain(type), binding_info(type));
+            type = instantiate(binding_body(type), local);
+        }
+        return nfields;
+    }
+
     expr visit_no_confusion(expr const & fn, buffer<expr> & args, bool root) {
         name const & no_confusion_name  = const_name(fn);
         name const & I_name             = no_confusion_name.get_prefix();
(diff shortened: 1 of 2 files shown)


nomeata:
I took the liberty to make the PR title more explicit, it helps at least me when looking at notifications. I hope that doesn't step on anyone's toes.

nrolland:
will this go to 4.4 or nightly ?

semorrison:
> will this go to 4.4 or nightly ?

When this is merged it will landed in the nightly releases. It will then be rolled out as part of `v4.5.0-rc1` (likely on Dec 21).

digama0:
Ping on this old PR.

semorrison:
Rebased to make sure there are no interactions with recent changes.

PR #2904, opened 2023-11-17
feat: import auto-completion

### LINKED ISSUE #2655: RFC: autocompletion of imports
### Proposal

Autocompletion of imports is currently missing in Lean 4, and is an important usability issue.

It is slightly difficult to implement, as when the imports are incomplete we likely can't even start the server, so it will require custom support in the VSCode extension.

### Community Feedback

* https://leanprover.zulipchat.com/#narrow/stream/113486-announce/topic/Lean.20FRO/near/388986233
* https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/Lean4.20autocomplete

I think this is already on @mhuisi's todo list, but I'm creating this issue to make sure there is a public tracker for it!

### Impact

mhuisi:
This PR adds basic auto-completion support for imports. Since it still lacks Lake support for accurate completion suggestions (cc @tydeu - we already know what needs to be done), it falls back to traversing the `LEAN_SRC_PATH` for available imports.

Three kinds of import completion requests are supported:

- Completion of the full `import` command. Triggered when requesting completions in an empty space within the header.
  - Known issue: It is possible to trigger this completion within a comment in the header. Fixing this would require architecture for parsing some kind of sub-syntax between individual commands.
- Completion of the full module name after an incomplete `import` command.
- Completion of a partial module name with a trailing dot.

Since the set of imports is potentially expensive to compute, they are cached for 10 seconds after the last import auto-completion request.

Closes #2655.

### Changes

This PR also makes the following changes:
- To support completions on the trailing dot, the `import` syntax was adjusted to provide partial syntax when a trailing dot is used.
- `FileWorker.lean` was refactored lightly with some larger definitions being broken apart.
- The `WorkerState` gained two new fields:
  - `currHeaderStx` tracks the current header syntax, as opposed to tracking only the initial header syntax in `initHeaderStx`. When the header syntax changes, a task is launched that restarts the file worker after a certain delay to avoid constant restarts while editing the header. During this time period, we may still want to serve import auto-completion requests, so we need to know the up-to-date header syntax.
  - `importCachingTask?` contains a task that computes the set of available imports.
- `determineLakePath` has moved to a new file `Lean/Util/LakePath.lean` as it is now needed both in `ImportCompletion.lean` and `FileWorker.lean`.
- `forEachModuleIn` from `Lake/Config/Blob.lean` has moved to `Lean/Util/Path.lean` as it is a generally useful utility function that was useful for traversing the `LEAN_SRC_PATH` as well.

### Tests

Unfortunately, this PR lacks tests since the set of imports available in `tests/lean/interactive` will not be stable. In the future, I will add support for testing LSP requests in full project setups, which is when tests for import auto-completion will be added as well.

### DIFF
--- a/src/Lean/Server/ImportCompletion.lean
+++ b/src/Lean/Server/ImportCompletion.lean
@@ -110,7 +110,7 @@ def collectAvailableImportsFromSrcSearchPath : IO AvailableImports :=
       if ! (← p.isDir) then
         continue
       Lean.forEachModuleInDir p fun mod => do
-        set <| (← get).push mod
+        modify (·.push mod)
 
 def collectAvailableImports : IO AvailableImports := do
   match ← ImportCompletion.collectAvailableImportsFromLake with


tydeu (approved):
LGTM! I imagine I may need to tweak a few things when I implement `lake available-imports`, but I presume that is okay.

src/Lean/Server/FileWorker.lean
```diff
@@ -181,65 +193,97 @@ section Initialization
...
-    | 3 => throwServerError s!"Imports are out of date and must be rebuilt; use the \"Restart File\" command in your editor.\n\n{stdout}"
-    | _ => throwServerError s!"`{cmdStr}` failed:\n{stdout}\nstderr:\n{stderr}"
+    | 3 => throwServerError s!"Imports are out of date and must be rebuilt; use the \"Restart File\" command in your editor.\n\n{result.stdout}"
+    | _ => throwServerError s!"`{cmdStr}` failed:\n{result.stdout}\nstderr:\n{result.stderr}"
+
+  def setupSrcSearchPath (m : DocumentMeta) (hOut : FS.Stream) (headerStx : Syntax) : IO SearchPath := do
+    let some path := System.Uri.fileUriToPath? m.uri
+      | return ← initSrcSearchPath
+
+    -- NOTE: we assume for now that `lakefile.lean` does not have any non-stdlib deps
```
semorrison, line 221: Suggest using "Lean" or "core Lean" rather than "stdlib", which should refer to Std.

src/Lean/Server/FileWorker.lean
```diff
@@ -181,65 +193,97 @@ section Initialization
...
+    return ← initSrcSearchPath pkgSearchPath
+
+  def buildHeaderEnv (m : DocumentMeta) (hOut : FS.Stream) (opts : Options) (headerStx : Syntax) : IO (Environment × MessageLog × SearchPath) := do
+    let (headerEnv, msgLog, srcSearchPath) ←
+      try
+        let srcSearchPath ← setupSrcSearchPath m hOut headerStx
+        -- allow `headerEnv` to be leaked, which would live until the end of the process anyway
+        let (headerEnv, msgLog) ← Elab.processHeader (leakEnv := true) headerStx opts MessageLog.empty m.mkInputContext
+        pure (headerEnv, msgLog, srcSearchPath)
+      catch e =>  -- should be from `lake print-paths`
```
semorrison, line 241: You could move the `catch` to before `processHeader` to be sure!

src/Lean/Server/FileWorker.lean
mhuisi: Agreed. By the way, this comment is two years old at this point :)

src/Lean/Util/LakePath.lean
```diff
@@ -0,0 +1,8 @@
+def Lean.determineLakePath : IO System.FilePath := do
```
Kha, line 1: missing header

src/Lean/Util/Path.lean
```diff
@@ -10,6 +10,17 @@ with a directory `A/`. `import A` resolves to `path/A.olean`.
 -/
 import Lean.Data.Name
 
+partial def System.FilePath.forEachModuleIn [Monad m] [MonadLiftT IO m]
```
Kha, line 13: Personally speaking I don't like attaching use topic-specific methods to generic types

Kha (approved):
I didn't review `ImportCompletion` in much detail

src/Lean/Util/Path.lean
mhuisi: I moved it to `Lean`, like the other functions in the file.

src/Lean/Server/ImportCompletion.lean
```diff
@@ -0,0 +1,141 @@
...
+    return none
+
+def collectAvailableImportsFromSrcSearchPath : IO AvailableImports :=
+  (·.2) <$> StateT.run (s := #[]) do
+    let srcSearchPath ← initSrcSearchPath
+    for p in srcSearchPath do
+      if ! (← p.isDir) then
+        continue
+      Lean.forEachModuleInDir p fun mod => do
+        set <| (← get).push mod
```
tydeu, line 113:
```suggestion
       modify (·.push mod)
```

Suggestion for code style that just happen to stick out to me while looking at this once more.

mhuisi: Good point. Note that this is not just a code style issue but also a performance issue because `get` duplicates the array reference, preventing in-place updates of the array, whereas `modify` does not.

PR #2908, opened 2023-11-18
fix: don't write padding bits to olean files

digama0:
With `USE_GMP=OFF`, the implementation of `mpz` used a `bool` next to a `size_t`, resulting in 7 bytes of padding that are faithfully copied into the `.olean` files on affected systems, and resulting in nondeterminism in lake hashes. (This is not a problem for the mathlib cache, because the CI machines are running Linux x86, which enables GMP.)

PS: Rather than this representation, it would be nice if we could use 
```c
struct mpz {
  uint32_t capacity;
  int32_t sign_size;
  mpn_digit * m_digits;
};
```
(where `capacity == abs(sign_size)`) for ABI compatibility between GMP and non-GMP variants of lean built on the same system. Otherwise one has to know whether GMP is enabled or not in order to encode/decode these files (I'm going to implement this in leantar using the architecture, but if we decide to change the GMP setting at some point this will be an ABI break with no indicator).


digama0:
I could also use `uint8_t m_padding[7]; bool m_size;` there if preferred.

eric-wieser:
> An alternative fix is to zero the padding bits in the compactor after writing the struct.

Another solution is to copy across the fields one-by-one, since the memory was zeroed to start with:
```diff
--- a/stage0/src/runtime/compact.cpp
+++ b/stage0/src/runtime/compact.cpp
@@ -278,7 +278,10 @@ void object_compactor::insert_mpz(object
     size_t data_sz = sizeof(mpn_digit) * to_mpz(o)->m_value.m_size;
     size_t sz      = sizeof(mpz_object) + data_sz;
     mpz_object * new_o = (mpz_object *)alloc(sz);
-    memcpy(new_o, to_mpz(o), sizeof(mpz_object));
+    // struct has padding, which we want to be deterministic
+    memcpy(new_o, o, sizeof(lean_object));
+    new_o->m_value.m_sign = to_mpz(o)->m_value.m_sign;
+    new_o->m_value.m_size = to_mpz(o)->m_value.m_size;
```

eric-wieser:
I've made a PR with the above patch.

PR #2909, opened 2023-11-18
doc: clarify doc-string for Lean.Elab.Tactic.withLocation

kim-em:
In the previous doc-string, the sentence

> "If any of the selected tactic applications fail, it will call `failed` with the main goal mvar."

was false both for `Location.wildcard` (where it should have said "If all", not "If any") or for `Location.targets` (where `failed` is never called).

### DIFF
--- a/src/Lean/Elab/Tactic/Location.lean
+++ b/src/Lean/Elab/Tactic/Location.lean
@@ -47,7 +47,7 @@ Runs the given `atLocal` and `atTarget` methods on each of the locations selecte
 * If `loc` is `*`, runs at the target first and then the hypotheses in reverse order.
   If all tactic applications fail, `withLocation` with call `failed` with the main goal mvar.
   If the tactic application closes the main goal, `withLocation` will then fail with `no goals to be solved`
-  when it reaches the first hypothesis.
+  upon reaching the first hypothesis.
 -/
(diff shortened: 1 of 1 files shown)


src/Lean/Elab/Tactic/Location.lean
```diff
@@ -39,9 +39,16 @@ def expandOptLocation (stx : Syntax) : Location :=
...
-If any of the selected tactic applications fail, it will call `failed` with the main goal mvar.
- -/
+/--
+Runs the given `atLocal` and `atTarget` methods on each of the locations selected by the given `loc`.
+
+* If `loc` is a list of locations, runs at each specified hypothesis (and finally the goal if `⊢` is included),
+  and fails if any of the tactic applications fail.
+* If `loc` is `*`, runs at the target first and then the hypotheses in reverse order.
+  If all tactic applications fail, `withLocation` with call `failed` with the main goal mvar.
+  If the tactic application closes the main goal, `withLocation` will then fail with `no goals to be solved`
```
leodemoura, line 49: Now that you documented it, it looks weird, right? Do users expect this behavior?

digama0: Agreed that this is weird behavior, it should just close the goal in this case

semorrison: I will separately PR a change in behaviour.

semorrison: #2917

PR #2911, opened 2023-11-19
fix: `eq_refl` tactic’s name is `eqRefl`

nomeata:
Previously, it has `name := refl`, which looked confusing in [the docs](https://leanprover-community.github.io/mathlib4_docs/Init/Tactics.html#Lean.Parser.Tactic.refl) – there is no `refl` tactic.


nomeata:
Or maybe there is a good reason for this? If so, I’ll change the PR to include a comment to explain.

stage0/src/stdlib_flags.h
```diff
@@ -6,7 +6,7 @@ options get_default_options() {
     // see https://lean-lang.org/lean4/doc/dev/bootstrap.html#further-bootstrapping-complications
 #if LEAN_IS_STAGE0 == 1
     // switch to `true` for ABI-breaking changes affecting meta code
-    opts = opts.update({"interpreter", "prefer_native"}, false);
+    opts = opts.update({"interpreter", "prefer_native"}, true);
```
semorrison, line 9: Presumably this is spurious.

src/Init/Tactics.lean
```diff
@@ -250,7 +250,7 @@ macro:1 x:tactic tk:" <;> " y:tactic:2 : tactic => `(tactic|
     all_goals $y:tactic)
 
 /-- `eq_refl` is equivalent to `exact rfl`, but has a few optimizations. -/
-syntax (name := refl) "eq_refl" : tactic
+syntax (name := eqRefl) "eq_refl" : tactic
```
semorrison, line 253: This is fine, but a more radical suggestion is to simply remove it. It is not used in Lean/Std/Mathlib currently.

stage0/src/stdlib_flags.h
digama0: Probably not. It will be automatically reverted by the next update-stage0, see https://github.com/leanprover/lean4/blob/master/doc/dev/bootstrap.md#further-bootstrapping-complications.

src/Init/Tactics.lean
nomeata: Besides as one of the implementations of `rfl`, it seems:
```
macro "rfl" : tactic => `(tactic| eq_refl)
```

PR #2918, opened 2023-11-20
fix: DecidableEq deriving handler could not handle fields whose types start with an implicit argument

### LINKED ISSUE #2914: Deriving DecidableEq fails in implicit argument
### Prerequisites

### Description

Can't derive DecidableEq when implicit argument is used in structure

### Context

```
structure Foo where
  a : List Nat
  ha : ∀ {i}, i ∈ a → 0 < i
deriving DecidableEq
```
Zulip discussion: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/deriving.20DecidableEq.20fails.20with.20implicit.20argument. The equivalent in Lean 3 succeeds. This was observed in a porting note in mathlib from February 2023 but seemingly never reported.

### Steps to Reproduce

Code as above.

**Expected behavior:** A DecidableEq instance is generated for Foo.

**Actual behavior:** Fails with error:
```
tactic 'subst' failed, invalid equality proof, it is not of the form (x = t) or (t = x)
  _ = _
x✝¹x✝: Foo
a✝¹: List Nat
a✝b✝: ∀ {i : Nat}, i ∈ a✝¹ → 0 < i
h✝: _ = _
⊢ Decidable ({ a := a✝¹, ha := a✝ } = { a := a✝¹, ha := b✝ })
```

### Versions

Lean (version 4.3.0-rc2)
MacOS

### Additional Information

None

### Impact

kmill:
Fixes #2914

### DIFF
--- a/src/Lean/Elab/Deriving/DecEq.lean
+++ b/src/Lean/Elab/Deriving/DecEq.lean
@@ -24,15 +24,15 @@ where
     | [] => ``(isTrue rfl)
     | (a, b, recField, isProof) :: todo => withFreshMacroScope do
       let rhs ← if isProof then
-        `(have h : $a = $b := rfl; by subst h; exact $(← mkSameCtorRhs todo):term)
+        `(have h : @$a = @$b := rfl; by subst h; exact $(← mkSameCtorRhs todo):term)
       else
-        `(if h : $a = $b then
+        `(if h : @$a = @$b then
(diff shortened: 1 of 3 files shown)


tests/lean/2914.lean.expected.out
timotree3, line 1: This test should probably be in `tests/lean/run/` instead if its output doesn't need to be checked against anything

semorrison (approved):

PR #2919, opened 2023-11-20
feat: `Lean.MVarId.cleanup` configuration

kmill:
Modifies `cleanup` so that it takes (1) an array of additional fvarids to preserve and (2) a flag to control whether to include indirect propositions.

(This is wanted in mathlib for the `extract_goal` tactic.)


semorrison (approved):

leodemoura:
> (This is wanted in mathlib for the extract_goal tactic.)

@kmill Please include that in the `cleanup` docstring

kmill:
@leodemoura Done

PR #2923, opened 2023-11-20
chore: fix superfluous lemmas in simp.trace

kim-em:
Fixes an issue reported on Zulip; see the test case.

* Modifies the `MonadBacktrack` instance for `SimpM` to also backtrack the `UsedSimps` field.
* When calling the discharger, `saveState`, and then `restoreState` if something goes wrong.

I'm not certain that it makes sense to restore the `MetaM` state if discharging fails. I can easily change this to more conservatively just backtrack the `UsedSimps` after failed discharging.

### DIFF
--- a/src/Lean/Meta/Tactic/Simp/Rewrite.lean
+++ b/src/Lean/Meta/Tactic/Simp/Rewrite.lean
@@ -37,13 +37,18 @@ def synthesizeArgs (thmId : Origin) (xs : Array Expr) (bis : Array BinderInfo) (
         if (← synthesizeInstance x type) then
           continue
       if (← isProp type) then
+        -- We save the state, so that `UsedTheorems` does not accumulate
+        -- `simp` lemmas used during unsuccessful discharging.
+        let usedTheorems := (← get).usedTheorems
         match (← discharge? type) with
         | some proof =>
           unless (← isDefEq x proof) do
             trace[Meta.Tactic.simp.discharge] "{← ppOrigin thmId}, failed to assign proof{indentExpr type}"
+            modify fun s => { s with usedTheorems }
             return false
         | none =>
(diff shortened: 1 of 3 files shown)


src/Lean/Meta/Tactic/Simp/Rewrite.lean
```diff
@@ -37,13 +37,18 @@ def synthesizeArgs (thmId : Origin) (xs : Array Expr) (bis : Array BinderInfo) (
         if (← synthesizeInstance x type) then
           continue
       if (← isProp type) then
+        -- We save the state, so that `UsedSimps` does not accumulate
+        -- `simp` lemmas used during unsuccessful discharging.
+        let s ← saveState
```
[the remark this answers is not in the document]
semorrison: (My first attempt at this was changing the `MonadBacktrack` instance, then realizing we weren't even backtracking.)

Changed to just saving the `UsedTheorems`.

src/Lean/Meta/Tactic/Simp/Types.lean
```diff
@@ -45,9 +45,15 @@ structure State where
...
 abbrev SimpM := ReaderT Context $ StateRefT State MetaM
 
+structure SavedState where
+  meta : Meta.SavedState
+  usedTheorems : UsedSimps
+
 instance : MonadBacktrack SavedState SimpM where
-  saveState      := Meta.saveState
-  restoreState s := s.restore
+  saveState      := return ⟨← Meta.saveState, (← get).usedTheorems⟩
```
leodemoura, line 53: Do you still need this modification?

semorrison: No, I don't. In fact, this instance is dead code which I'll separately propose deleting.

semorrison:
Lovely -- the mathlib failures are `simp? says` complaining that the `simp?` output has improved!

leodemoura (approved):

Issue #1422, opened 2022-08-05
Raw strings


Kha:
We should have a new string literal that accepts double quotes as is. My suggestion in a prior discussion was to follow Rust 1:1: https://github.com/leanprover/lean4/issues/407#issuecomment-819772315

alok:
@Kha this can be closed now?

semorrison:
Yes, #2929 implemented this.

PR #2929, opened 2023-11-20
feat: Rust-style raw string literals

kmill:
For example, `r"\n"` and `r#"The word "this" is in quotes."#`.

Implements #1422

### DIFF
--- a/src/Lean/Parser/Basic.lean
+++ b/src/Lean/Parser/Basic.lean
@@ -670,11 +670,6 @@ partial def strLitFnAux (startPos : String.Pos) : ParserFn := fun c s =>
     else strLitFnAux startPos c s
 
 /--
-Enables `rawStrLitStart` being defined using `partial`.
--/
-local instance : Inhabited (ParserState → ParserState × Option Nat) := ⟨fun s => (s, none)⟩
-
-/--
 Raw strings have the syntax `r##...#"..."#...##` with zero or more `#`'s.
 If we are looking at a valid start to a raw string (`r##...#"`),
 returns true.


hargoniX:
Is there a particular reason (apart from copying Rust syntax) that this breaks the current Lean convention of having modifiers before strings with exclamation marks (think `s!` and `m!`). I feel like it would be rather confusing for a new user that some modifiers come with `!` and others with `#`.

src/Lean/Parser/Basic.lean
```diff
@@ -627,6 +627,63 @@ partial def strLitFnAux (startPos : String.Pos) : ParserFn := fun c s =>
     else if curr == '\\' then andthenFn quotedCharFn (strLitFnAux startPos) c s
     else strLitFnAux startPos c s
 
+/-- Raw strings have the syntax `r##...#"..."#...##` with zero or more `#`'s.
+If this is a valid start to a raw string (`r##...#"`),
+returns the parser state after the `"` along with the number of `#`'s -/
+partial def rawStrLitStart (c : ParserContext) (s : ParserState) : Option (ParserState × Nat) :=
```
digama0, line 633: You should never pass the parser state in an Option because this requires the caller to use the state nonlinearly, since it has to hold on to the original state for use in the none case. Instead you should just always return a `ParserState` and ensure that it has its position reset to the start in that case.

kmill: Thanks, I pushed a fix to that nonlinearity.

kmill:
@Kha I think this is ready for review

semorrison (approved):

Kha:
!bench

Kha:
There is a ~1% parser run time regression. That sounds acceptable, but it might be worth a try avoiding the allocation for the `rawStrLitStart` result by only calling it when we're looking at an `r`. Sounds sensible?

kmill:
@Kha I modified it to avoid allocation and to use the cheap `curr == 'r'` check. I also simplified the raw string parser state machine by making the states more explicit, which also saves some branching. One thing I decided is that it is OK to do lookahead for the initial `##...#"` for a fast check and then parse it again to actually count the `#`'s, since then this fast check definitely does not need to allocate anything.

I'll run the benchmarks again once the build succeeds.

kmill:
!bench

tydeu:
!bench

PR #2930, opened 2023-11-20
refactor: lake: simplify math template & test it

tydeu:
Removes the `CI` option from the `math` template. Since the template does not currently generate a GitHub workflow, it does not do anything out of the box except add unnecessary complexity.

The `math` template is also now tested in `tests/init` (minus the Mathlib `require`).

### DIFF
--- a/src/lake/Lake/CLI/Init.lean
+++ b/src/lake/Lake/CLI/Init.lean
@@ -93,31 +93,19 @@ def mathConfigFileContents (pkgName libRoot : String) :=
 s!"import Lake
 open Lake DSL
 
-def leanOptions : Array LeanOption := #[
-  ⟨`pp.unicode.fun, true⟩, -- pretty-prints `fun a ↦ b`
-  ⟨`pp.proofs.withType, false⟩
-]
-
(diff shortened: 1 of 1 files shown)


src/lake/tests/init/test.sh
```diff
@@ -34,6 +40,14 @@ $LAKE1 build
 .lake/build/bin/hello
 popd
 
+# Test math template
+
+$LAKE new qed math
+# Remove the require, since we do not wish to download mathlib during tests
+sed_i '/^require.*/{N;d}' qed/lakefile.lean
```
digama0, line 47: Won't this just delete the first line? The `require mathlib` line is broken in the middle in the template.

tydeu: It should delete both the `N;` reads an additional line. See [this StackOverflow question](https://stackoverflow.com/questions/28021351/sed-remove-multiline-patterns-regexp) for an explanation. Also, it worked locally.

tydeu:
@semorrison Any comments? As our math community specialist, it seems appropriate to notify you of this PR before I merge.

PR #2933, opened 2023-11-21
feat: `pp.numericTypes` option for printing number literals with type ascriptions

### LINKED ISSUE #3021: RFC: Pretty printer options to see types of numeric literals
## Summary

* Add a boolean pretty printer option `pp.numericTypes` that causes numeric literals to pretty print with type ascriptions. For example, `(2 : Nat)`, `(-2 : Int)`, and `(-2 / 3 : Rat)`. The scope of a "numeric literal" is defined below, but it follows the convention of `norm_num`.
* Add a boolean pretty printer option `pp.natLit` that causes raw natural number literals to pretty print with `nat_lit`. For example, `nat_lit 2`. If `pp.natLit` is not set, it defaults to the value of `pp.numericType`.

## Motivation

When working with numeric literals in multiple different number systems simultaneously, it can be confusing making sense of a goal state. For example, in `x ^ 5`, is this `x ^ (5 : Nat)` or `x ^ (5 : Real)`? While both interpretations are equal to one another, they are not defeq so one needs to know which it is to know which lemmas immediately apply. Having a pretty printer option to see `5` along with its type ascription can help debug such issues when writing proofs.

Similarly, sometimes there are long algebraic expressions that appear to be defeq, and having a way to see the types of numeric literals at a glance is helpful -- without needing to hover over each term one at a time in the Infoview or to resort to using `pp.all`.

This option was present in the Lean 3 community edition as `pp.numeralTypes`, and it was added specifically to help with these issues.

The `pp.natLit` option is to help with a rarer scenario -- debugging when a natural number literal is a raw natural number literal or not. Knowing types of number literals seems related to `nat_lit` due to both having to do with numeric literals, so we include it in this RFC.

## Manual-level explanation

The natural number literal syntax is elaborated into an expression that uses the `OfNat.ofNat` function to convert a raw natural number literal into a natural number literal for a given type. This allows different types to each have their own interpretation of a natural number literal, and this way one can write `(2 : Nat)`, `(2 : Int)`, `(2 : Fin 10)`, and so on. Each of these examples elaborate into `OfNat.ofNat (nat_lit 2)` using some appropriate `OfNat` instance, where `nat_lit 2` is a raw natural number literal, which is a term that stands for an actual `Nat` constructor expression. (Note: while `nat_lit 2` stands for `Nat.succ (Nat.succ Nat.zero))`, the elaborator and kernel use a more-efficient internal representation.)

When speaking of expressions:
* we refer to these `OfNat.ofNat` expressions themselves as being natural number literals;
* an *integer literal* is an expression that is either a natural number literal or the negative of a natural number literal; and
* a *rational number literal* is an expression that is either an integer literal or the quotient of an integer literal and a natural number literal. The division operator in this quotient must be homogeneous for it to count as a rational number literal.

The collection of natural number literals, integer literals, and rational number literals are known as *numeric literals*.

By default, the pretty printer does not do anything special for numeric literals other than printing natural number literals as natural number literals. That is to say, it pretty prints, for example, `OfNat.ofNat (nat_lit 2)` as `2`.

If you set `set_option pp.numericTypes true`, then numeric literals print with a type ascription. For example, you can see expressions such as `(2 : Nat)`, `(-2 : Int)`, and `(-2 / 3 : Rat)` rather than just `2`, `-2`, and `-2 / 3`.

While raw natural number literals by default pretty print as the natural number itself, the `pp.numericTypes` option also causes the `nat_lit` prefix to print. One can use the `pp.natLit` option to enable just this feature, or even set `set_option pp.natLit false` to override `pp.numericTypes`. Normally, one should never see raw natural number literals, but this option is available to help detect any that might be present.

## Reference-level explanation

The `pp.natLit` and `pp.numericTypes` options control the delaboration of numeric literals. By default, `pp.numericTypes` is false, and by default `pp.natLit` is the current setting of `pp.numericTypes`.

When `pp.natLit` is false, then the `nat_lit n` expression delaborates as `n`, and otherwise it delaborates as `nat_lit n`.

When `pp.numericTypes` is false, then natural number literals (i.e., `OfNat.ofNat (nat_lit n)` expressions) delaborate as `n`. Otherwise, entire numeric literal expressions delaborate wrapped in a type ascription; a *numeric literal* consists of one of the following:
- a natural number literal; or
- an *integer literal*, which is a natural number literal or `Neg.neg n` with `n` a natural number literal; or
- a *rational number literal*, which is an integer literal or a homogeneous `HDiv.hDiv m n` with `m` an integer literal and `n` a natural number literal.

The type ascription obtains its type from either the `HDiv.hDiv`, the `Neg.neg`, or the `OfNat.ofNat`, whichever happens to be the outermost expression.

TermInfo is still associated to each subexpression that appears in integer literals and rational number literals.

## Drawbacks

* This adds complexity to the delaborator, and it adds new pretty printing options that users might need to learn to become competent with the system.
* The option might be "over-active" in that, for example, one would see `(2 : Fin (5 : Nat))` rather than the likely sufficient `(2 : Fin 5)`.
* Non-reduced expressions such as `(2 / 3 : Nat)` might be confusing.

## Rationale and alternatives

Naming-wise, the option is modeled after `pp.funBinderTypes`, `pp.piBinderTypes`, and `pp.instanceTypes`. While `pp.proofs.withType` seems related, this is instead controlling the behavior when `pp.proofs` is false (and perhaps it should be renamed `pp.proofs.withTypes` to follow how "types" is plural in most other options).

We considered `pp.numeralTypes`, since that was what it was called in the Lean 3 community edition. While the word "numeral" is frequently used in the internals to describe natural number literal expressions, and it does appear in the reference manual for this purpose, the option in this RFC encompasses numeric literals in a wider sense.

The Lean 3 option only added type ascriptions for natural number literals and integer literals, but not rational number literals. For this option, matching what `norm_num` considers to be a natural number literal seems to be consistent, and there is no confusion in an expression such as `(2 / 3 : Rat)` that `(2 : Rat) / (3 : Rat)` would reduce.

The `pp.numLit` option could be `pp.num_lit`, but that would be the only pretty printing option with an underscore.

While we could try to come up with an automated method for deciding whether to show types of numeric literals (like a `pp.analyze` system), we still need support for pretty printing numeric literals with type ascriptions, and that system could be implemented by setting `pp.numeralTypes` on subexpressions.

This proposal could be implemented in std or mathlib. However, the implementations would completely override the core delaborators for raw natural number literals and natural number literals. It would be cleaner to add this as a core Lean option, and of course then this option would be available when debugging Lean itself.

## Future possibilities

* There could be a pretty printing option for giving type ascriptions around arithmetic expressions whose components are all of the same type. This is a generalization of the notion of a numeric literal, and it can be used to help debug how the expression tree elaborator elaborates a given expression.
* We could augment `pp.analyze` with support for numeric literals, where it tags a subexpression with `pp.numeralTypes` if the type is ambiguous (like in the `x ^ 2` example).
* Similarly, we could augment it to tag subexpressions with `pp.numeralTypes` if they are arguments to polymorphic functions. Such cases might or might not be ambiguous (for example, `max 1 2 = (2 : Int)` is technically not ambiguous), but it might be useful to see `max 1 2` as `max (1 : Int) (2 : Int)` to make the types abundantly clear.
* We could write analysis delaborators for functions such as `Fin` so that they turn off `pp.numeralTypes` for their numeric argument. That way, we would see `(2 : Fin 5)` when the option is set.
* There could be a per-type option for pretty printing with type ascriptions. One can implement this as a library by adding a delaborator for, for example, `Fin` numeric literal expressions. We can expose numeric literal recognizers as part of the implementation of this RFC.
* The `pp.natLit` option could also take into account `Nat.zero` and `Nat.succ`. We suggest deferring this to a future RFC since in an obvious implementation, `Nat.zero` might pretty print as `0` when `pp.natLit` is false. This seems too controversial.

## Community Feedback

Prior discussion is on the Lean Zulip in [this thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/pre-RFC.3A.20pretty.20printer.20option.20for.20numeric.20literal.20types/near/404932599). (Note: there is already a partial implementation of this RFC at #2933.)

### Impact

kmill:
Implements the pretty printer option `pp.numericTypes` for including a type ascription for numeric literals. For example, `(2 : Nat)`, `(-2 : Int)`, and `(-2 / 3 : Rat)`. This is useful for debugging how arithmetic expressions have elaborated or have been otherwise transformed. For example, with exponentiation is is helpful knowing whether it is `x ^ (2 : Nat)` or `x ^ (2 : Real)`. This is like the Lean 3 option `pp.numeralTypes` but it has a wider notion of a numeric literal.

Also implements the pretty printer option `pp.natLit` for including the `nat_lit` prefix for raw natural number literals.

Closes #3021

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -215,6 +215,11 @@ simproc [my_simp] reduceFoo (foo _) := ...
 
 * Add pretty printer settings to omit deeply nested terms (`pp.deepTerms false` and `pp.deepTerms.threshold`) ([PR #3201](https://github.com/leanprover/lean4/pull/3201))
 
+* Add pretty printer options `pp.numeralTypes` and `pp.natLit`.
+  When `pp.numeralTypes` is true, then natural number literals, integer literals, and rational number literals
+  are pretty printed with type ascriptions, such as `(2 : Rat)`, `(-2 : Rat)`, and `(-2 / 3 : Rat)`.
+  When `pp.natLit` is true, then raw natural number literals are pretty printed as `nat_lit 2`.
+  [PR #2933](https://github.com/leanprover/lean4/pull/2933) and [RFC #3021](https://github.com/leanprover/lean4/issues/3021).
 
 v4.5.0
 ---------


Kha (approved):

tydeu:
I am a bit concerned about the name `numeralTypes` and its discoverability because, as far as I know, as numeric literals are not called "numerals" anywhere else in Lean. Thus, it would be hard to connect the dots between the option in an autocomplete list and its functionality. Maybe `numLitTypes` instead?

kmill:
Sebastian suggested that we go through the RFC process for this, so I'll put this back to draft and we can discuss these issues on Zulip.

kmill:
I'm re-opening this now that it's implementing an RFC.

Something I checked by hand is that the delaborators ensure that each subexpression is still hoverable in the infoview.

semorrison:
This looks good.
* Could you add a sentence to RELEASES.md for visibility?
* You wrote such great text on the RFC (manual level explanation, reference level explanation). Should any of that text be copied into `docs/`? Happy either way.

kmill:
@semorrison I added a RELEASES entry. Regarding the RFC, I'm planning on eventually going through the RFCs I've written and copying the explanations to relevant parts of the docs (in future PRs).

PR #2934, opened 2023-11-21
feat: `pp.parens` option to pretty print with all parentheses

kmill:
This PR adds the option `pp.parens` (default: false) that causes the pretty printer to eagerly insert parentheses, which can be useful for teaching and for understanding the structure of expressions. For example, it causes `p → q → r` to pretty print as `p → (q → r)`.

Any notations with precedence greater than or equal to `maxPrec` do not receive such discretionary parentheses, since this precedence level is considered to be infinity.

This option was a feature in the Lean 3 community edition.

### DIFF
--- a/src/Lean/PrettyPrinter/Delaborator/Options.lean
+++ b/src/Lean/PrettyPrinter/Delaborator/Options.lean
@@ -24,6 +24,11 @@ register_builtin_option pp.notation : Bool := {
   group    := "pp"
   descr    := "(pretty printer) disable/enable notation (infix, mixfix, postfix operators and unicode characters)"
 }
+register_builtin_option pp.parens : Bool := {
+  defValue := false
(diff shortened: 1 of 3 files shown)


kmill:
@Kha What do you think about this implementation? Do you have any ideas how to make it not parenthesize atomic syntax? In particular, is there some condition that implies parentheses are never needed?

Kha:
Perhaps something like `(← read).forceParens && prec < maxPrec`? That is, if the precedence of the thing to be parenthesized is already at least the precedence of parens, there is no need to do it.

kmill:
@Kha I've tried a few variations on that, but they didn't seem to have any effect. I think I've spent my time budget on this feature -- maybe if it the basic version of the feature exists, someone who is keen on having it prettier will figure out a solution?

PR #2935, opened 2023-11-21
fix: ignore errors on `IO.FS.Handle` finalization

Kha:


src/runtime/io.cpp
```diff
@@ -89,7 +89,14 @@ static obj_res mk_file_not_found_error(b_obj_arg fname) {
 static lean_external_class * g_io_handle_external_class = nullptr;
 
 static void io_handle_finalizer(void * h) {
-    lean_always_assert(fclose(static_cast<FILE *>(h)) == 0);
+    int errnum = fclose(static_cast<FILE *>(h));
+    if (errnum != 0) {
```
tydeu, line 93:
```suggestion
    if (!fclose(static_cast<FILE *>(h))) {
```
For consistency with the style used elsewhere in the file.

src/runtime/io.cpp
```diff
@@ -89,7 +89,14 @@ static obj_res mk_file_not_found_error(b_obj_arg fname) {
...
+        object * str = lean_io_error_to_string(decode_io_error(errnum, nullptr));
+        sstream ss;
+        ss << "could not close finalized `IO.FS.Handle`: " << string_cstr(str);
+        lean_internal_panic(ss.str().c_str());
+        dec_ref(str);
+    }
 }
```
joehendrix, line 100: I was surprised to see there is no `IO.FS.Handle.close` operation.  That seems like it'd be sensible to add, and then this code would need to handle that case (`fclose` returning non-zero and then checking if `errno` is `ENOTOPEN`).

I could add that as a separate commit though.

src/runtime/io.cpp
Kha: errnum is used in the next line

src/runtime/io.cpp
Kha: Yes, sounds good to me as a follow-up PR

PR #2954, opened 2023-11-23
feat: WF.GuessLex: If there is only one plausible measure, use it

nomeata:
If here is only one plausible measure, there is no point having the `GuessLex` code see if it
is terminating, running all the tactics, only for the `MkFix` code then run the tactics again.

So if there is only one plausible measure (non-mutual recursion with only one varying
parameter), just use that measure.

Side benefit: If the function isn’t terminating, more detailed error messages are shown 
(failing proof goals), located at the recursive calls.


nomeata:
!bench

tests/lean/guessLexFailures.lean.expected.out
```diff
@@ -7,7 +7,27 @@ argument #1 was not used for structural recursion
...
+  - Use `termination_by` to specify a different well-founded relation
+  - Use `decreasing_by` to specify your own tactic for discharging this kind of goal
+n✝ n : Nat
+⊢ n < n✝
+guessLexFailures.lean:33:14-33:31: error: failed to prove termination, possible solutions:
+  - Use `have`-expressions to prove the remaining goals
+  - Use `termination_by` to specify a different well-founded relation
+  - Use `decreasing_by` to specify your own tactic for discharging this kind of goal
+n✝ n : Nat
+⊢ n < Nat.succ n✝
```
nomeata, line 33: TODO:

These tests need to be updated/extended to have more than one parameter, else the code that they should test would just be skipped.

Same also for `guessLex.lean`.

nomeata:
This seems straightforward and reasonable, and fixes some of the performance regressions of introducing GuessLex (#2960). Will merge right after that (but separately for bisecting and more helpful history).

PR #2958, opened 2023-11-24
refactor: rewrite TerminationHint elaborators

nomeata:
In order to familiarize myself with this code, and so that the next
person has an easier time, I

* added docstrings explaining what I found out these things to
* rewrote the syntax expansion functions using syntax pattern matches,
  to the extend possible


src/Lean/Elab/PreDefinition/WF/TerminationHint.lean
```diff
@@ -20,31 +20,38 @@ inductive TerminationHint where
...
+      | some declName => return m.insert declName { ref := arg, value := arg[2] }
+    for clique in cliques do
+      let mut found? := Option.none
+      for declName in clique do
+        if let some { ref, .. } := m.find? declName then
+          if let some found := found? then
+            Macro.throwErrorAt ref s!"invalid termination hint element, '{declName}' and '{found}' are in the same clique"
+          found? := some declName
+    return TerminationHint.many m
+  | _ => Macro.throwUnsupported
```
nomeata, line 54: Here I wasn’t able to use more syntax matches. Is that because `terminationHint1` and [`terminationHintMany`](https://github.com/leanprover/lean4/blob/8e5cf6466061e7660353d7c3d9b69f744ac29edd/src/Lean/Parser/Command.lean#L38-L40) are parametric parsers, so the syntax match elaboration doesn’t work? Or was I just not trying the right things?

nomeata:
Could someone with some understanding of syntax pattern macros have a look: Is the code I wrote sensible and idiomatic (if not, better tell me before I get used to bad ways), and can something be done about those parametric parsers, or is that the best I can do? Thanks!

src/Lean/Elab/PreDefinition/WF/TerminationHint.lean
Kha: Yes, parametric parsers cannot be used as quotation parsers but you can still match through them:
```
`(terminationByCore| termination_by' $e:term)
```
Now, for the `Many` case something like
```
`(terminationByCore| termination_by' $[$ids:ident => $es $[;]?]*)
```
*should* work but I believe the `lookahead` is preventing that as it's not happy to see the `$` instead of an actual identifier :/

nomeata: I never figured out complex syntax like that, with nested parsers. Thanks for the example!

Wrapping `patternIgnore` around the lookahead won’t help, I assume.

Kha: It won't, no. I suppose the lookahead can be called a workaround for tactics not currently being (contextual) keywords, which we want to fix. But also your "per-declaration `termination_by`" would fix it as well I assume as there would be no more need for `Many`?

nomeata: Yes, was about to say that; no need to dwell on this too much. 

Happy with the rest of this PR?

PR #2959, opened 2023-11-24
fix: `run_task`/`deactivate_task` race condition on `m_imp->m_closure`

### LINKED ISSUE #2853: Segmentation fault in `Task.bind` task
Executing the last portion of a `Task.bind` can run into the closure for it already having been nulled.
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007ff709a56a7c in lean_ptr_tag (o=0x0) at /home/sebastian/lean/lean/build/debug/stage1/include/lean/lean.h:386
386	    return o->m_tag;
[Current thread is 1 (Thread 0x7ff6feffd6c0 (LWP 1784522))]
(gdb) up
...
#5  0x00007ff709a6498c in lean::task_manager::run_task (this=0x55b73faa1380, lock=..., t=0x7ff6fdc4c098) at /home/sebastian/lean/lean/src/runtime/object.cpp:718
718	            add_dep(lean_to_task(closure_arg_cptr(t->m_imp->m_closure)[0]), t);
```

To reproduce:
Not easily. This is a somewhat rare failure in my new snapshot types branch.

Analysis:
This appears to be a plain race condition in the following code: https://github.com/leanprover/lean4/blob/b278172b7c8be73678e40cb7fc38c7b2db2bbe1b/src/runtime/object.cpp#L707-L718
We check that the task is not deleted then, in the last branch, we drop the lock and access the closure, but a concurrent thread can empty the closure field in `deactivate_task_core` in between! I've confirmed that retrieving the closure within the lock like is done in the same function before fixes my test failures: https://github.com/Kha/lean4/commit/994a088e6149c83b5ec252173ed5d65949c61c53. But it is not actually clear to me why/whether this is sufficient in either of these closure accesses: who is actually keeping the closure alive after we've dropped the lock? `deactivate_task_core` will `dec` it as soon as the task is unreferenced but I don't see or remember a counteracting `inc`.

@leodemoura In general I am concerned about the cost/benefit ratio of the two following `task_object` features that are involved in the bug:
* the `deactivated` state that eagerly clears up specific fields (such as `m_closure`) even while the task is queued or running. If it merely set `m_canceled` and left all cleanup to when the `task_manager` is completely done with the task, there would be no interference with other task functions.
* reusing the task object in `task_bind_fn1` for a second logical task instead of allocating a new task object, which requires the special case branch in `task_manager` that broke above.

As far as I can see, both of these provide marginal memory use benefits while they significantly complicate the `task_object` invariants and thus its maintainability. Both parts had to be fixed before.

Kha:
Fixes #2853, unblocking my work before I get to refactoring this part of the task manager.

### DIFF
--- a/src/runtime/object.cpp
+++ b/src/runtime/object.cpp
@@ -714,8 +714,12 @@ class task_manager {
             resolve_core(t, v);
         } else {
             // `bind` task has not finished yet, re-add as dependency of nested task
+            // NOTE: closure MUST be extracted before unlocking the mutex as otherwise
+            // another thread could deactivate the task and empty `m_clousure` in
+            // between.
+            object * c = t->m_imp->m_closure;
             lock.unlock();
-            add_dep(lean_to_task(closure_arg_cptr(t->m_imp->m_closure)[0]), t);
+            add_dep(lean_to_task(closure_arg_cptr(c)[0]), t);
             lock.lock();
         }
     }


leodemoura (approved):

semorrison:
This needed to be rebased so that the new CI steps were triggered, before the merge queue would pick this up.

PR #2960, opened 2023-11-24
feat: GuessLex: if no measure is found, explain why

nomeata:
by showing the matrix of calls and measures, and what we know about that
call (=, <, ≤, ?), e.g.

    guessLexFailures.lean:27:0-33:31: error: Could not find a decreasing measure.
    The arguments relate at each recursive call as follows:
    (<, ≤, =: relation proved, ? all proofs failed, _: no proof attempted)
               x1 x2 x3
    1) 29:6-25  =  =  =
    2) 30:6-23  =  ?  <
    3) 31:6-23  <  _  _
    Please use `termination_by` to specify a decreasing measure

It’s a bit more verbose for mutual functions.

It will use the user-specified argument names for functions written
```
foo (n : Nat) := …
```
but not with pattern matching like
```
foo : Nat → … 
  | n => …
```
This can be refined later and separately (and maybe right away in `expandMatchAltsWhereDecls`).


nomeata:
TODO: 

* [ ] Also for mutual recursion
* [ ] More tests

I am not sure how well an ASCII-table works everywhere; the “Problems” pane of VSCode uses a proportional font.

Issue #2963, opened 2023-11-26
RFC: Bundle of widget improvements


Vtec234:
### Proposal

- This RFC proposes a bundle of improvements to user widgets that I identified over the year from bug reports to [ProofWidgets4](https://github.com/leanprover-community/ProofWidgets4), as well as my own usage.
- Generalize basic widget declarations.
	- Motivation. When designing user-widgets, our original understanding of "widget" was roughly as "something that could be displayed in the infoview next to the tactic state". With ProofWidgets it became clear that this isn't general enough; for instance, it is useful to be able to declare widgets that are merely subcomponents (for instance, a single button) to be used in other components but which shouldn't be displayed at the top-level.
	- Thus we generalize "user widget" which has a mandatory user-visible name (under the assumption that it will be shown at the top-level) to "widget module" which is just an arbitrary JS module. We now call JS module that produce a React component suitable for top-level display "panel widgets".
	- Implementation.
		- Deprecate `Widget.UserWidgetDefinition` in favour of `Widget.Module` which does not require a `name` field.
			- In pictures, before the change the widget writer controls the blue box and the string contents of the title. After the change, they control the orange box. <img width="255" alt="Screenshot 2023-11-26 at 1 03 18 AM" src="https://github.com/leanprover/lean4/assets/13901751/7a2096fa-ec21-4888-9916-a60649f21940">
		- Deprecate the `@[widget]` attribute in favour of a `@[widget_module]` attribute.
		- ~~Rename `ofUserWidget` to `ofPanelWidget` to clarify when panel widgets are being used, and other definitions accordingly.~~ (This change was abandoned.)
- Allow storing RPC references in widget props.
	- Motivation. Panel widget nodes in the infotree currently cannot store RPC references, which limits what can be expressed and necessitates various hacks.
	- Implementation. `RpcEncodable` data can now be stored in `ofUserWidget`: https://github.com/Vtec234/lean4/commit/f7e8c082cbac62d85800b09daea6792069f6efa9#diff-513721adf31ab2bf9c9bd846073e7d71720d0f67cef2bbb81a76887071c703daR110-R115
- Add ability to display a widget everywhere (global), and everywhere within a given file (local).
	- Motivation.
		- Users want to be able to declare a widget as always-on within a given file, or when a given namespace is open: e.g. a geometry widget popping up after `open GeometryLib`. See also https://github.com/leanprover-community/ProofWidgets4/issues/5.
	- Implementation.
[...]

PR #2964, opened 2023-11-26
feat: bundle of widget improvements

Vtec234:
Implements RFC #2963.

Leftover tasks:
- [x] Provide companion PR to vscode-lean4 (leanprover/vscode-lean4#376)
- [x] Companion PR to std4 (leanprover/std4#467)
- [x] Companion PR to ProofWidgets4 (leanprover-community/ProofWidgets4#36)
- [X] Companion commit to mathlib4 (https://github.com/leanprover-community/mathlib4/commit/0f4660f6552d2a6aaae7adb90a6a3a830cd7fa1c)
- [ ] ~~Update the manual chapter~~ (will do in a follow-up)

### DIFF
--- a/src/Lean/Widget/UserWidget.lean
+++ b/src/Lean/Widget/UserWidget.lean
@@ -365,8 +365,6 @@ def widgetInfosAt? (text : FileMap) (t : InfoTree) (hoverLine : Nat) : List User
         failure
     | _, _, _ => none
 
-deriving instance Server.RpcEncodable for WidgetInstance
-
 structure PanelWidgetInstance extends WidgetInstance where
   /-- The syntactic span in the Lean file at which the panel widget is displayed. -/
   range? : Option Lsp.Range := none


src/Lean/Widget/UserWidget.lean
```diff
@@ -2,189 +2,395 @@
...
 -/
 import Lean.Elab.Eval
 import Lean.Server.Rpc.RequestHandling
 
-open Lean
+namespace Lean.Server
+open Elab Command in
+/-- Derive `Lean.Server.RpcEncodable` for a type.
+
+TODO: remove after update-stage0 -/
```
Vtec234, line 14: This needs an `update-stage0`. Should I run it myself?

tests/lean/derivingRpcEncoding.lean
```diff
@@ -16,7 +16,6 @@ structure FooRef where
   a : Array Nat
   deriving Inhabited, TypeName
 
-#check instTypeNameFooRef
```
Vtec234, line 19: I guess the test shouldn't rely on autogenerated names anyway.

src/Lean/Server/InfoUtils.lean
```diff
@@ -108,7 +108,7 @@ def Info.stx : Info → Syntax
   | ofFieldInfo i          => i.stx
   | ofCompletionInfo i     => i.stx
   | ofCustomInfo i         => i.stx
-  | ofUserWidgetInfo i     => i.stx
+  | ofPanelWidgetInfo i    => i.stx
```
Kha, line 111: FWIW, @david-christiansen mentioned that he had great success with using `ofCustomInfo`, which begs the question whether we want to migrate all other constructors to it /cc @mhuisi

Kha (approved):
There isn't really any non-widget code to review, so just a comment from me

src/Lean/Widget/UserWidget.lean
Kha: Yes, feel free to do it when the PR is ready for merge

src/Lean/Server/InfoUtils.lean
Vtec234: Ah, I was [using that](https://github.com/leanprover-community/ProofWidgets4/blob/28046b1c0d70de55bc3ef33abef635ac4dc9a5ab/ProofWidgets/Compat.lean#L159) in ProofWidgets to hack around things that this PR is changing. It has the added benefits of cutting some dependencies, i.e., `InfoTree.Types` would not need to know about widgets anymore. It may also mesh well with your plans to generalize info trees. However, I had undone this renaming now as it's more annoying than helpful, so I'd leave that as future work.

Kha: Great, fine with me. @mhuisi mentioned that it might be a good idea to keep separate constructors for "standard" cases - which arguably widgets would be more so than the other constructors that are specific to the Lean language

semorrison (approved):
The necessary changes in Mathlib look fine with me, so I would be okay with merging this. We are now extremely close to the cut-off for `v4.5.0-rc1`, however.

src/Lean/Widget/UserWidget.lean
```diff
@@ -2,189 +2,421 @@
...
```
semorrison, line 14: Does this still need to be removed?

PR #2968, opened 2023-11-27
doc: update quickstart guide to reference vs code setup guide

mhuisi:
Since the vscode-lean4 setup guide allows us to provide information on setting up Lean 4 tailored to the user's operating system, this PR adjusts the quickstart guide to reference the vscode-lean4 setup guide instead.


Kha (approved):
LGTM! @semorrison @PatrickMassot We should also sync the community setup pages with this.

Issue #2953, opened 2023-11-23
`simp_all?` suggests wrong/misleading `simp_all only [...]`


JLimperg:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to mathlib4 or std4.

### Description

Consider this `simp_all?` call:

```lean
example {m n k : Nat} (h₁: n = m) (h₂ : m = k) : n = 0 := by
  set_option tactic.simp.trace true in
  simp_all -- Try this: simp_all only [h₂]
  -- m n k : Nat
  -- h₁ : n = k
  -- h₂ : m = k
  -- ⊢ k = 0
```

The `Try this` suggestion is (a) wrong, since `h₁` was also used; (b) misleading since `simp_all only [h₂]` doesn't appear to actually restrict the set of local hypotheses used. In the following example, `h₁` is used despite `only [h₂]`:

```lean
example {m n k : Nat} (h₁: n = m) (h₂ : m = k) : n = 0 := by
  simp_all only [h₂]
  -- m n k : Nat
  -- h₁ : n = k
  -- h₂ : m = k
  -- ⊢ k = 0
```

I would therefore suggest that `simp_all?` should omit local hypotheses from the generated `simp_all only` list. Hence, the **expected suggestion** above would be `simp_all only`.

Are there any hypotheses whose presence in the `simp_all only` list does make a difference? I thought `simp_all` might not pick up dependent propositional hypotheses by default, but apparently it does:

```lean
example {m n : Nat} (P : ∀ {n m}, n = m → Prop) (h₁ : n = m) (h₂ : P h₁) : n = 0 := by
  simp_all
  -- m n : Nat
  -- P : {n m : Nat} → n = m → Prop
  -- h₁ : n = m
  -- h₂ : P _
  -- ⊢ m = 0
```

### Context

Issue #1853  is somewhat related. The very old PR #1862 addresses the issue; I'll try to resurrect it.

### Versions

Lean 4.3.0-rc2

### Impact

Aesop [currently works around this](https://github.com/leanprover-community/aesop/commit/bbba41db5f1ffd3b21e0c53d6e6e4f00660e2f3c).

[...]

PR #2969, opened 2023-11-27
fix: omit fvars from simp_all? theorem list

JLimperg:
Removes local hypotheses from the simp theorem list generated by `simp_all?`.

Fixes: #2953

---

Supersedes PR #1862

### DIFF
--- a/src/Lean/Elab/Tactic/Simp.lean
+++ b/src/Lean/Elab/Tactic/Simp.lean
@@ -264,7 +264,14 @@ register_builtin_option tactic.simp.trace : Bool := {
   descr    := "When tracing is enabled, calls to `simp` or `dsimp` will print an equivalent `simp only` call."
 }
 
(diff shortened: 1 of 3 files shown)


JLimperg:
awaiting-review

semorrison:
Perhaps for the future, if `usedSimps` has no local hypotheses, we could suggest replacing `simp_all` with `simp`. But perhaps there aare strange corner cases where this is not correct.

semorrison:
Could you please rebase onto `origin/nightly` (alternatively `origin/nightly-with-mathlib`), so we can get a Mathlib CI run, just to be sure?

JLimperg:
Looks green on nightly as well. :)

awaiting-review

leodemoura (approved):

semorrison:
Rebasing this onto `nightly`, so our updated CI is happy.

PR #2970, opened 2023-11-28
feat: test_extern command

kim-em:
This adds a `test_extern` command.

Usage:
```
import Lean.Util.TestExtern

test_extern Nat.add 17 37
```

This:
* Checks that the head symbol has an `@[extern]` attribute.
* Writes down `t == t'`, where `t` is the term provided, and `t'` is the reference implementation (specifically, `t` with the head symbol unfolded).
* Tries to reduce this to `true`, and complains if this fails.

Note that the type of the term must have a `BEq` instance for this to work: there's a self-explanatory error message if it isn't available.

### DIFF
--- a/tests/pkg/test_extern/test.sh
+++ b/tests/pkg/test_extern/test.sh
@@ -21,13 +21,13 @@ verify_output() {
 lake build 2>&1 | verify_output > produced.txt
 
 # Compare the actual output with the expected output
-if diff -q produced.txt expected.txt > /dev/null; then
+if diff --strip-trailing-cr -q produced.txt expected.txt > /dev/null; then
     echo "Output matches expected output."
     rm produced.txt
     exit 0
 else
     echo "Output differs from expected output:"
-    diff produced.txt expected.txt
+    diff --strip-trailing-cr produced.txt expected.txt
     rm produced.txt
     exit 1
 fi


tests/lean/test_extern.lean.expected.out
```diff
@@ -0,0 +1,11 @@
...
+test_extern.lean:9:0-9:86: error: native implementation did not agree with reference implementation!
+Compare the outputs of:
+#eval ByteArray.copySlice { data := #[1, 2, 3] } 1 { data := #[4, 5, 6, 7, 8, 9, 10, 11, 12, 13] } 0 6
+ and
+#eval {
+  data :=
+    Array.extract { data := #[4, 5, 6, 7, 8, 9, 10, 11, 12, 13] }.data 0 0 ++
+        Array.extract { data := #[1, 2, 3] }.data 1 (1 + 6) ++
+      Array.extract { data := #[4, 5, 6, 7, 8, 9, 10, 11, 12, 13] }.data (0 + 6)
+        (Array.size { data := #[4, 5, 6, 7, 8, 9, 10, 11, 12, 13] }.data) }`
```
semorrison, line 11:
```suggestion
        (Array.size { data := #[4, 5, 6, 7, 8, 9, 10, 11, 12, 13] }.data) }
```

nomeata (approved):
Very neat

tests/pkg/test_extern/test.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+#!/bin/bash
+
+rm -rf .lake/build
+
+# Function to process the output
+verify_output() {
+    awk '/error: stdout:/, /error: external command/' | sed '/error: external command/d'
+}
```
nomeata, line 10: On windows we still get
```
2023-11-28T02:45:09.3711308Z < error: stdout:
2023-11-28T02:45:09.3737497Z < .\.\.\TestExtern.lean:5:0: error: native implementation did not agree with reference implementation!
2023-11-28T02:45:09.3739223Z < Compare the outputs of:
2023-11-28T02:45:09.3739977Z < #eval Nat.not_add 4 5
2023-11-28T02:45:09.3741896Z <  and
2023-11-28T02:45:09.3754905Z < #eval match 4, 5 with
2023-11-28T02:45:09.3760717Z < | x, x_1 => 0
2023-11-28T02:45:09.3763062Z ---
2023-11-28T02:45:09.3780816Z > error: stdout:
2023-11-28T02:45:09.3784341Z > ./././TestExtern.lean:5:0: error: native implementation did not agree with reference implementation!
2023-11-28T02:45:09.3785391Z > Compare the outputs of:
2023-11-28T02:45:09.3785950Z > #eval Nat.not_add 4 5
2023-11-28T02:45:09.3786409Z >  and
2023-11-28T02:45:09.3786835Z > #eval match 4, 5 with
2023-11-28T02:45:09.3787300Z > | x, x_1 => 0
```
is that something that the sed should fix?
(Maybe add a comment why the output has to be mangled)

semorrison: Normalized, and added apology for the ad-hoc mangling.

leodemoura (approved):

digama0:
> Note that the type of the term must have a `BEq` instance for this to work: there's a self-explanatory error message if it isn't available.

I think this should be using a `DecidableEq` instance, not a `BEq` instance, because this is guarding against values that are different in any way. Even if the difference is ignored by `BEq` this is still bad.

PR #2973, opened 2023-11-28
feat: encode `let_fun` using a `letFun` function

kmill:
Switches from encoding `let_fun` using an annotated `(fun x : t => b) v` expression to a function application `letFun v (fun x : t => b)`.

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -72,7 +72,7 @@ This is equal to `(fun x => b) v`, so the value of `x` is not accessible to `b`.
 This is in contrast to `let x := v; b`, where the value of `x` is accessible to `b`.
 
 There is special support for `letFun`.
-Both WHNF and `simp` are aware of `letFun` and can reduce it when zeta reduction is enable,
+Both WHNF and `simp` are aware of `letFun` and can reduce it when zeta reduction is enabled,
 despite the fact it is marked `irreducible`.
 For metaprogramming, the function `Lean.Expr.letFun?` can be used to recognize a `let_fun` expression
 to extract its parts as if it were a `let` expression.


kmill:
@leodemoura I carried out your suggestion to re-encode `let_fun`, and hopefully it came out in the way you anticipated!

leodemoura:
> I'm not sure if you were expecting it to be for dependent or non-dependent lets.

It is for non-dependent lets.

> I investigated making it be only for non-dependent lets, but I found that the first example in tests/lean/letFun.lean

I can encode all `let_fun`s in this test file using the auxiliary `letFun` function.  Could you please clarify the issue you are having?

```lean
def letFun {α : Sort u} {β : α → Sort v} (v : α) (f : (x : α) → β x) : β v := f v

#check
  letFun (fun x => x * 2) fun f =>
  letFun 1 fun x =>
  letFun (x + 1) fun y =>
  f (y + x)

example (a b : Nat) (h1 : a = 0) (h2 : b = 0) : (letFun (a + 1) fun x => x + x) > b := by
  simp (config := { beta := false }) [h1]
  trace_state
  simp (config := { decide := true }) [h2]
```

kmill:
@leodemoura I've used the wrong terminology -- using `SimpLetCase` for clarification, I mean to say I wasn't sure if you were expecting the `nondepDepVar` or `nondep` encoding.  I got the `nondepDepVar` encoding of `letFun` working in this PR, and it does exactly what you wrote in your code block.

kmill:
@leodemoura I don't have any obstruction here, and I can handle non-constant type families. I just wanted to be sure that you weren't expecting this simplified `letFun` encoding [that you had proposed in your original message](https://github.com/leanprover/lean4/pull/2864#issuecomment-1807176233) since I instead used [this "`nonDepDepVar`" encoding](https://github.com/leanprover/lean4/pull/2973/files#diff-52ef0c67eea613acf6c0b6284063fd7112cdd40750de38365c214abaef246db4R71), which I switched to at the first sign of trouble. The latter was more obvious how to implement. In particular [`mkLetFun`](https://github.com/leanprover/lean4/pull/2973/files#diff-042e743a1251bf5b246b4ff2fcf7c2258550eaf715fbe888baaea034ecedc167R29-R36), to produce the simple encoding, would need to force `β` to not depend on the fvar `x`, even in the presence of metavariables.

Examples like this one definitely work with this PR's implementation:
```lean
#check let_fun n := 5
  (⟨[], Nat.zero_le n⟩ : { as : List Bool // as.length ≤ n })
```
I haven't evaluated whether these sorts of examples appear in the wild, but I would not be surprised if they do.

You are correct that it does not generalize the value in the expected type, so your examples with `{ as : List Bool // as.length ≤ 5 }` do not work. I personally believe generalization would be a surprising feature for `let_fun`, and if that is at some point a requested feature, that it should be either an option (like `let_fun (generalizing := true) n := 5; ...`) or be a new binding syntax.

leodemoura:
@kmill Let's move forward and merge.

kmill:
@leodemoura Sounds good. Let me know if you need me to do anything. (Maybe I should rebase this onto the newest nightly and regenerate the stage0?)

semorrison:
> (Maybe I should rebase this onto the newest nightly and regenerate the stage0?)

Yes please.

kmill:
@semorrison Done.

By the way, this will take a small change to std. I don't have push access to that project, so didn't update the testing branch. Instead, I made my own branch and modified the mathlib testing branch to use that. It was an easy fix: https://github.com/kmill/std4/commit/7ca1c2c8232a72a10ba14e69869d1efc6666b043

src/Init/Prelude.lean
```diff
@@ -66,6 +66,13 @@ example (b : Bool) : Function.const Bool 10 b = 10 :=
 @[inline] def Function.const {α : Sort u} (β : Sort v) (a : α) : β → α :=
   fun _ => a
 
+/--
+The encoding of `let_fun x := v; b` is `letFun v (fun x => b)`.
+This is an abbreviation for `(fun x => b) v`, so the value of `x` is not accessible to `b`,
+unlike with `let x := v; b`.
+-/
+def letFun {α : Sort u} {β : α → Sort v} (v : α) (f : (x : α) → β x) : β v := f v
```
Kha, line 74: This encoding means that even with `config.zeta` false, we reduce `let_fun`s when at default reducibility, correct? Is that intentional or should it be `[irreducible]`?

kmill: Good point, that is something I did not think enough about. I made it be irreducible and added more tests. (I also extended the optimization for ignoring unnecessary lets in `isDefEqQuick` to also consider `let_fun`.)

src/Init/Prelude.lean
```diff
@@ -66,6 +66,19 @@ example (b : Bool) : Function.const Bool 10 b = 10 :=
 @[inline] def Function.const {α : Sort u} (β : Sort v) (a : α) : β → α :=
   fun _ => a
 
+/--
+The encoding of `let_fun x := v; b` is `letFun v (fun x => b)`.
+This is equal to `(fun x => b) v`, so the value of `x` is not accessible to `b`.
+This is in contrast to `let x := v; b`, where the value of `x` is accessible to `b`.
+
+There is special support for `letFun`.
+Both WHNF and `simp` are aware of `letFun` and can reduce it when zeta reduction is enable,
```
Kha, line 75:
```suggestion
Both WHNF and `simp` are aware of `letFun` and can reduce it when zeta reduction is enabled,
```

Kha (approved):

src/Init/Prelude.lean
Kha: @leodemoura I'll wait for your final confirmation on this point before merging

leodemoura: Agreed. We don't want to reduce `let_fun`s when `config.zeta = false` and using the default reducibility.

leodemoura: Agreed. We don't want to reduce `let_fun`s when `config.zeta = false` and using the default reducibility.

semorrison:
@Kha, @kmill, I don't think we should have merged this while it had a `breaks-mathlib` label!

It has indeed broken both Std and Mathlib's `nightly-testing` branches, and without understanding this PR in detail I don't immediately know how to fix them.

(The [discussion on zulip](https://leanprover.zulipchat.com/#narrow/stream/345428-mathlib-reviewers/topic/CI.20failure.20on.20the.20nightly-testing.20branch/near/409042357) might be a better place for follow-up than here.)

kmill:
@semorrison Did you see my message earlier, https://github.com/leanprover/lean4/pull/2973#issuecomment-1853065287 ?

I'll follow up on zulip too.

PR #2978, opened 2023-11-28
chore: Check PR title, not commit, for commit convention

nomeata:
Also turn this into a proper check, run when a PR is opened or edited.

I took the liberty to rename the workflow file and name, so that one
doesn't have to look inside to guess what the workflow is doing.


nomeata:
Seems to work, as tested with #2979 and #2980

PR #2985, opened 2023-11-29
chore: Just an empty PR with a bad title

nomeata:
from another repo


nomeata:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request says

> The pull_request webhook event payload is empty for merged pull requests and pull requests that come from forked repositories.

But it does not seem to be true.

PR #2986, opened 2023-11-29
chore: trim CI set by default

nomeata:
The goal of this change is to run a trimmed-down CI on PRs by default, but allows opt-in the full CI as necessary.

### Specification

The CI workflow runs in “quick” mode if it was triggered from a pull request, and that pull request does not have the `full-ci` label set.

In “quick” mode the build matrix contains fewer jobs. At the moment only:

* Linux-release, to get the PR releases.

In non-quick mode everything should be as before.

### Implementation notes

I created a `configure` job that combines all the previous `set-` jobs, I guess this is faster than firing up separate jobs.

The matrix is calculated in this job; this seems to be the cleanest way to get a dynamic matrix going (experiments using `exclude` failed). The downside is that the matrix is now in JSON rather than Yaml syntax. The upside is that we can (later) make it’s calculation simpler, e.g. set default `shell` values etc.

I was not able to make it so that CI runs when the `full-ci` label is added, but don’t do anything otherwise. I think it can be done with another workflow listening to `labeled` and then triggering this one, but let’s do that separately. For now, add the label and then push (or close and reopen).

The checks
```
  if: matrix.build-stage2 || matrix.check-stage3
  if: matrix.check-stage3
```
were dead code, we did not have these fields in the matrix anymore, so I replaced them with
```
  if: matrix.test-speedcenter
```


nomeata:
> Note that the CI step only tests that the benchmarks still work, it is not a part of the !bench pipeline. So I think we can leave that check to the merge queue.

Ah, great! Removed from `quick`

nomeata:
> > The downside is that the matrix is now in JSON rather than Yaml syntax.
> 
> One upside is that we could now actually factor out common configuration :) ? 

That’s what I meant to say :-)

> But all the comments are gone :( ?

Ups. While converting I thought “I need to put comments back in”, and then thought got lost… will fix.

Kha (approved):

nomeata:
As expected the PR release only has linux:
https://github.com/leanprover/lean4-pr-releases/releases/tag/pr-release-2986

I’ll rebase this on nightly-with-mathlib to see if the mathlib test job works as expected.

PR #2994, opened 2023-11-29
fix: do not strip dotted components from lean module names

### LINKED ISSUE #2999: Imports not working for module names with dots
### Prerequisites

### Description

To import a file `A/B.C.lean`, `import «A».«B.C»` instead tries to find a file called `A/B.lean`.

The workaround was to use `import «A».«B.C.lean»` instead.

### Context

Bug described [here](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Import.20file.20with.20multiple.20dots.20in.20file.20name)

### Steps to Reproduce

1. Create a file `A/B.C.lean`
2. Try to import it from a different file with `import «A».«B.C»`

**Expected behavior:** Import works

**Actual behavior:**

```lean4
`/home/vscode/.elan/toolchains/leanprover--lean4---v4.2.0/bin/lake print-paths Init «A».«B.C» --no-build` failed:

stderr:
error: '«A».«B.C»': no such file or directory (error code: 2)
  file: ./././A/B.lean
```

### Versions

Lean v4.2.0

### Additional Information

### Impact

eric-wieser:
This introduces `FilePath.addExtension` to take a path that we know has no prior extension, and append a new extension to it.
As this function is simpler than `FilePath.withExtension`, this change eagerly replaces uses of the latter with the former, except in a few cases where stripping the extension really is the right thing to do.

This should fix the bug described at https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Import.20file.20with.20multiple.20dots.20in.20file.20name/near/404508048, where `import «A.B».«C.D.lean»` is needed to import `A.B/C.D.lean`.

Closes #2999

### DIFF
--- a/src/Init/System/FilePath.lean
+++ b/src/Init/System/FilePath.lean
@@ -101,9 +101,9 @@ def withFileName (p : FilePath) (fname : String) : FilePath :=
   | none => ⟨fname⟩
   | some p => p / fname
 
-/-- Append the extension `ext` to a path `p`.
+/-- Appends the extension `ext` to a path `p`.
 
-`ext` should not contain a `.`, as this function adds one.
+`ext` should not contain a leading `.`, as this function adds one.
 If `ext` is the empty string, no `.` is added.
 
 Unlike `System.FilePath.withExtension`, this does not remove any existing extension. -/


src/lake/tests/init/test.sh
```diff
@@ -77,6 +77,12 @@ $LAKE -d lean-data exe lean-data
 $LAKE new 123-hello
 $LAKE -d 123-hello exe 123-hello
 
+# Test creating packages both starting with a digit and containing a `.`
+# https://github.com/leanprover/lean4/issues/2999
+
+$LAKE new 123-hello.world
+$LAKE -d 123-hello-world exe 123-hello.world
+
```
tydeu, line 85: I am pretty sure this does not test what you think it does. This creates a package with a module named `123Hello.World` located `123Hello/World.lean`. If you want to include the do in the module name, you need to use guillemets (e.g., `lake new «A.B».«C.D»`).

eric-wieser: That creates a folder name that contains the guillemets!

tydeu: @eric-wieser Oh, yeah. The guillemets in top-level directory name should be stripped (in another PR). Nonetheless, that should create the proper directory structure underneath it (i.e., ` «A-B»-«C-D»/A.B/C.D.lean`) with the fix in this PR.

src/lake/tests/init/clean.sh
```diff
@@ -5,5 +5,6 @@ rm -rf hello_world
 rm -rf hello-exe
 rm -rf lean-data
 rm -rf 123-hello
+rm -rf «A.B».«C.D»
```
tydeu, line 8:
```suggestion
rm -rf «A-B»-«C-D»
```

tydeu (approved):
LGTM!

tydeu:
@Kha I think the test failure on Nix may be due to the use of guillemets in a file name?

mhuisi (approved):
The changes to `FilePath.lean`, `Watchdog.lean` and `Path.lean` look good to me.

eric-wieser:
It looks to me like the full tests are still not actually running

Kha:
@eric-wieser Could you try merging master?

src/lake/Lake/CLI/Init.lean
```diff
@@ -218,16 +218,17 @@ def initPkg (dir : FilePath) (name : String) (tmp : InitTemplate) (env : Lake.En
 
 def validatePkgName (pkgName : String) : LogIO PUnit := do
   if pkgName.isEmpty || pkgName.all (· == '.') || pkgName.any (· ∈ ['/', '\\']) then
-    error "illegal package name"
+    error ("illegal package name: " ++ pkgName)
```
tydeu, line 221:
```suggestion
    error s!"illegal package name '{pkgName}'"
```

src/lake/Lake/CLI/Init.lean
```diff
@@ -218,16 +218,17 @@ def initPkg (dir : FilePath) (name : String) (tmp : InitTemplate) (env : Lake.En
...
 
 def init (pkgName : String) (tmp : InitTemplate) (env : Lake.Env) (cwd : FilePath := ".") : LogIO PUnit := do
   let pkgName ← do
     if pkgName == "." then
-      match (← IO.FS.realPath cwd).fileName with
+      let path ← IO.FS.realPath cwd
+      match path.fileName with
       | some dirName => pure dirName
-      | none => error "illegal package name"
+      | none => error ("illegal package name: cwd is " ++ path.toString)
```
tydeu, line 231:
```suggestion
      | none => error s!"illegal package name: could not derive a package name from '{path}'"
```

tydeu (commented):
If you intend to keep the error message changes, I suggested some edits I would like to see.

tydeu:
@eric-wieser As to the error itself, I believe it has to do with MSYS2 path conversion. Adding the following to the top of the test file should hopefully fix it:
```shell
# make paths behave on MSYS2
[ "$OSTYPE" == "msys" ] && export MSYS2_ARG_CONV_EXCL=*
```

src/lake/tests/init/test.sh
```diff
@@ -77,6 +77,15 @@ $LAKE -d lean-data exe lean-data
 $LAKE new 123-hello
 $LAKE -d 123-hello exe 123-hello
 
+# Test creating packages with components that contain `.`s
+# https://github.com/leanprover/lean4/issues/2999
+  
+# this fails on windows for unrelated reasons
+if [ "$OSTYPE" != "msys" ]; then
```
eric-wieser, line 84: @tydeu: I couldn't work this one out: but I figure that whatever this bug is, it's not in scope for this PR.

tydeu: Fair enough. Yeah, I think it can be pushed to a later date.

eric-wieser: @tydeu: is this PR good to go then?

tydeu:
@eric-wieser I am happy with the Lake changes, but I think we need @Kha to sign off on the core changes before merging.

src/Init/System/FilePath.lean
```diff
@@ -101,6 +101,11 @@ def withFileName (p : FilePath) (fname : String) : FilePath :=
   | none => ⟨fname⟩
   | some p => p / fname
 
+def addExtension (p : FilePath) (ext : String) : FilePath :=
```
Kha, line 104: Needs docstring

Kha (changes requested):

eric-wieser:
@Kha, is the new docstring ok?

src/Init/System/FilePath.lean
```diff
@@ -101,11 +101,21 @@ def withFileName (p : FilePath) (fname : String) : FilePath :=
   | none => ⟨fname⟩
   | some p => p / fname
 
+/-- Append the extension `ext` to a path `p`.
+
+`ext` should not contain a `.`, as this function adds one.
```
Kha, line 106:
```suggestion
/-- Appends the extension `ext` to a path `p`.

`ext` should not contain a leading `.`, as this function adds one.
```

PR #2998, opened 2023-11-30
chore: CI: in quick mode, only Nix build runs the tests

nomeata:
Following up on #2986, stop running the test suite in ci.yml in quick
mode; the test suite is run in the Nix job, and we do not need to run it
twice.

With a cold nix cache, when `lean` is rebuilt, not much changes, as both
jobs take ~20mins. But when `lean` is unchanged, the nix build should
be faster, and shaving off the (currently) 4mins in the CI.yaml run
should get us to a green PR sooner.

Another benefit is that we get the PR release sooner and even get it
when the test suite fails, which can be useful if you want to test
mathlib or other things before fixing the lean test suite.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -374,7 +374,7 @@ jobs:
         if: (matrix.wasm || !matrix.cross) && !needs.configure.outputs.quick
       - name: Check Test Binary
         run: ${{ matrix.binary-check }} tests/compiler/534.lean.out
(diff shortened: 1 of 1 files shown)


Kha (approved):

nomeata:
I promise I’ll try to write lowercase titles; but it goes very hard against my muscle memory. Maybe I’ll have to extend the commit convention check.

Kha:
Hah, I didn't even notice that this time

Issue #2921, opened 2023-11-20
RFC: Run decreasing_by tactic on all goals, not each goal


nomeata:
### Proposal

Consider the (of course non-terminating) definition:
```
def foo (n : Nat) : Nat :=  foo (n - 1) + foo n + foo (n + 1)
decreasing_by …
```

The tactic passed to `decreasing_by` is currently executed on each recursive call separately, with one goal present. This makes it harder to write these termination tactics individually, for example when different recursive calls need widely different approaches, because whatever you write has to work with all goals at the same time.

As a user I would have expected to be able to focus individual calls with `\.`, like so
```
def foo (n : Nat) : Nat := foo (n - 1) + foo n + foo (n + 1)
decreasing_by
  · trace_state; sorry
  · trace_state; sorry
  · trace_state; sorry
```
but here the second `\.` says “no goals to be solved”.

If instead the tactic state after `decreasing_by` has _all_ goals present and `\.` (or `case <patterm>` or other focusing commands) wold work, interactively building the termination proof would be simpler. I can still use `all_goals` to apply something to all of them, e.g. `all_goals simp_wf`

Open question: If no explicit proof is given, and the default `termination_tactic` is used, should it be applied as before (on each goal separately), to put the squiggly line under the recursive call that fails?? Or to all goals at once, for consistency?

Maybe down the line we want to be able to distinguish, either by different syntax or by educated guesses, whether the user is best served by squiggly lines under failing recursive calls, or by reporting errors at the tactic that fails. `decreasing_by a_single_tactic` is a corner cases where it’s unclear what’s best.

### Community Feedback

[Initial discussion on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Tricky.20lexicographic.20termination.20issue/near/401607174) (but not much activity).

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others benefit from the changes in this proposal being added, please ask them to add :+1: to it.

leodemoura:
Like the idea, but what about mutual recursion? The actual order may not be clear to users. Recall that we compute SCCs.

nomeata:
Hmm, yes, the order may be surprising.

We could sort the MVars by the syntax `ref` maybe to put them back into source code order? (If MVars remember the ref they were created at or if we pass it separately.)

I'll experiment with multiple SCCs and see.

[...]

PR #3004, opened 2023-12-01
fix: GuessLex: deduplicate recursive calls

nomeata:
The elaborator is prone to duplicate terms, including recursive calls,
even if the user only wrote a single one. This duplication is wasteful
if we run the tactics on duplicated calls, and confusing in the output
of GuessLex. So prune the list of recursive calls, and remove those
where another call exists that has the same goal and context that is no
more specific.


nomeata:
!bench

nomeata:
The effect of this will be more pronounced with #2960. Maybe I’ll finish that first, to have a more pleasant diff here :-)


I also plan to make an analogous changes to `WF.Fix`, but that will be part of #2921, or at least a step towards that.

PR #3008, opened 2023-12-01
chore: add vscode cmake configuration

eric-wieser:
This sets the build directory to `build/release` for the "CMake Tools for Visual Studio Code" extension documented at https://vector-of-bool.github.io/docs/vscode-cmake-tools/settings.html#cmake-builddirectory.

It also sets the generator to `make`, since otherwise it tries `Ninja` which doesn't work.

Without these settings, the extension runs configure in a bad place at startup.

This does *not* add the cmake tools extension to the default workspace configuration; the goal is simply to prevent bad behavior for users who already have the extension enabled.

# Summary

Screenshot of this in action:

image

Link to `RFC` or `bug` issue: N/A, this is not a bug nor a user-visible feature.

### DIFF
--- a/lean.code-workspace
+++ b/lean.code-workspace
@@ -13,6 +13,8 @@
 	"settings": {
 		"files.insertFinalNewline": true,
 		"files.trimTrailingWhitespace": true,
+		"cmake.buildDirectory": "${workspaceFolder}/build/release",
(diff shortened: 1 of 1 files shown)


eric-wieser:
awaiting-review

PR #3010, opened 2023-12-01
feat: allow `EStateM` to be used universe-polymorphically

eric-wieser:
This PR makes `EStateM` universe polymorphic.

This doesn't change `IO` or `ST` (such a change would be more invasive and require approval of #3011), but does enable downstream projects like `Mathlib` to use `EStateM` for their own purposes with full universe polymorphism.
The alternative is defining an entire copy of `EStateM`, which is a lot of boilerplate (though in fact Lake has already done this!).

The generalized `EStateM.map` after this change can be invoked (in the non-monadic spelling) with `ULift.up` to move between universes. Previously it was only universe-monomorphic.

There seems to be no cost associated behind this generalization.

Zulip thread: [#lean4 > StateM and Result not universe polymorphic @ 💬](https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/StateM.20and.20Result.20not.20universe.20polymorphic/near/498751195)


tests/lean/CompilerElimDeadBranches.lean.expected.out
eric-wieser, line 1: As far as I can tell this is just a permutation

eric-wieser:
Marking this as "ready-for-review", since it seems that some Lean change in the last 9 month remove any fallout this once had (perhaps it was just a case of stage0 being outdated at the time)

eric-wieser:
An argument in favor of this change; the analogous `EStateT` is already universe polymorphic:

https://github.com/leanprover/lean4/blob/682173d7c084de77bd948b458ef9a7bf48bcb34e/src/lake/Lake/Util/EStateT.lean#L70-L75

eric-wieser:
(cc @tydeu who I assume wrote the above)

kim-em:
!bench

kim-em:
!bench

kim-em:
Oh the second run isn't actually independent, of course. We'd need to push an empty commit or something.

eric-wieser:
!bench

ghost:
Mathlib CI status ([docs](https://leanprover-community.github.io/contribute/tags_and_branches.html)):
- 💥 Mathlib branch [lean-pr-testing-3010](https://github.com/leanprover-community/mathlib4/compare/nightly-testing...lean-pr-testing-3010) build failed against this PR. (2025-02-10 16:40:21) [View Log](https://github.com/leanprover-community/mathlib4/actions/runs/13244621087)
- ✅ Mathlib branch [lean-pr-testing-3010](https://github.com/leanprover-community/mathlib4/compare/nightly-testing...lean-pr-testing-3010) has successfully built against this PR. (2025-02-16 03:10:53) [View Log](https://github.com/leanprover-community/mathlib4/actions/runs/13350852499)
- ❗ Batteries/Mathlib CI will not be attempted unless your PR branches off the `nightly-with-mathlib` branch. Try `git rebase 8e9da7a1bcfbfc313301df86274f9c93f1cc66db --onto deef1c2739a331899b0e536312e65e9815059014`. You can force Mathlib CI using the `force-mathlib-ci` label. (2025-04-14 14:21:32)
- ❗ Mathlib CI can not be attempted yet, as the `nightly-testing-2025-07-27` tag does not exist there yet. We will retry when you push more commits. If you rebase your branch onto `nightly-with-mathlib`, Mathlib CI should run now. You can force Mathlib CI using the `force-mathlib-ci` label. (2025-07-27 12:35:23)

eric-wieser:
@kim-em, can you trigger benchmarking? It looks like the bot only listens to FRO members.

tydeu:
!bench

eric-wieser:
@tydeu, mind doing so one more time now that the downstream issues are gone?

tydeu:
!bench

eric-wieser:
Given every benchmark seems to be totally different and each time the changed code is identical, I assume they can all be written off as noise?

PR #3013, opened 2023-12-02
feat: synchronous execution of task continuations

Kha:
In the new snapshot design, we have a tree of `Task`s that represents the asynchronously processed document structure. When transforming this tree in response to a user edit, we want to quickly run through reusable, already computed nodes of the tree synchronously and then spawn new tasks for the new parts. The new flag allows us to do such mixed sync/async tree transformations uniformly. This flag exists as e.g. [`ExecuteSynchronously`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcontinuationoptions?view=net-8.0) in other runtimes.

### DIFF
--- a/src/Init/Core.lean
+++ b/src/Init/Core.lean
@@ -411,9 +411,10 @@ set_option linter.unusedVariables.funArgs false in
 be available and then calls `f` on the result.
 
 `prio`, if provided, is the priority of the task.
+If `sync` is set to true, `f` is executed on the current thread if `x` has already finished.
 -/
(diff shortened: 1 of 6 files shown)


joehendrix (approved):

Kha:
Just to mention design alternatives I've considered, one could ask whether we should use a forward-compatible flag enum instead like .NET does above. But that's not something we really have as an idiom so far. In any case, I think this feature is advanced enough that I would not expect too much downstream breakage if we do consider a refactoring the next time we add a parameter.

PR #3015, opened 2023-12-03
chore: Nix CI: update setup

Kha:
Now that we're, at least temporarily, relying more on the Nix CI, replace some old hacks of mine with better solutions people have figured out in the meantime.

Cachix support could probably be dropped at this point but it doesn't really hurt.


nomeata (approved):
LGTM

Kha:
Oh well, only superficial changes left

PR #3016, opened 2023-12-04
doc: mention termination_by and decreasing_by

nomeata:
so far, our reference manual did not mention these at all, this takes
the discussion of recursive definition out of the “equation compiler”
section, put it into its own section, and expands it a bit.

This is more a MVP doc change to at least mention the features briefly,
and not the most polished and thought through didactic exposition. But
it provides a start for more improvements.

### DIFF
--- a/doc/declarations.md
+++ b/doc/declarations.md
@@ -512,12 +512,12 @@ end Vector
 Recursive functions
 ===================
 
-Lean must ensure that a recursive functions terminate, for which there are two strategies: _structural recursion_, in which all recursive calls are made on smaller parts of the input data, and _well-founded recursion_, in which recursive calls are justified by showing that arguments to recursive calls are smaller according to some other measure.
+Lean must ensure that a recursive function terminates, for which there are two strategies: _structural recursion_, in which all recursive calls are made on smaller parts of the input data, and _well-founded recursion_, in which recursive calls are justified by showing that arguments to recursive calls are smaller according to some other measure.
 
 Structural recursion
 --------------------
 
-If the definition of a function contains recursive calls, Lean tries to interpret the definition as a structural recursion. In order for that to succeed, the recursive arguments must be subterms of the corresponding arguments on the left-hand side.
+If the definition of a function contains recursive calls, Lean first tries to interpret the definition as a structural recursion. In order for that to succeed, the recursive arguments must be subterms of the corresponding arguments on the left-hand side.
 
 The function is then defined using a *course of values* recursion, using automatically generated functions ``below`` and ``brec`` in the namespace corresponding to the inductive type of the recursive argument. In this case the defining equations hold definitionally, possibly with additional case splits.
 
@@ -584,15 +584,15 @@ Please use `termination_by` to specify a decreasing measure.
 
 This table should be read as follows:
 
- * The first recursive call, in line 39, arguments 1, 2 and 4 are equal to the function's paramters.
- * The second recursive call, in line 40, has an equal first, a smaller fourth argument, and nothing could be inferred for the second argument.
+ * In the first recursive call, in line 39, arguments 1, 2 and 4 are equal to the function's parameters.
+ * The second recursive call, in line 40, has an equal first argument, a smaller fourth argument, and nothing could be inferred for the second argument.
  * The third recursive call, in line 41, has a decreasing first argument.
  * No other proofs were attempted, either because the parameter has a type without a non-trivial ``WellFounded`` instance (parameter 3), or because it is already clear that no decreasing measure can be found.
 
 
 Lean will print the termination argument it found if ``set_option showInferredTerminationBy true`` is set.
 
-If Lean does not find the termination argument, or if you want to be explict, you can append a `termination_by` clause to the function definition, after the function's body, but before the `where` clause if present. It is of the form
+If Lean does not find the termination argument, or if you want to be explicit, you can append a `termination_by` clause to the function definition, after the function's body, but before the `where` clause if present. It is of the form
 ```
 termination_by e
 ```
@@ -687,7 +687,7 @@ termination_by n h => h
 end
 ```
 
-Another way to express mutual recursion using local function definitions in ``where`` or ``let rec`` clauses: these can be mutually recursive with each other and their containing function:
+Another way to express mutual recursion is using local function definitions in ``where`` or ``let rec`` clauses: these can be mutually recursive with each other and their containing function:
 
 ```
 theorem even_of_odd_succ : ∀ n, Odd (n + 1) → Even n


nomeata:
TODO: Also update https://github.com/leanprover/theorem_proving_in_lean4

doc/declarations.md
```diff
@@ -504,7 +539,10 @@ example : append [(1 : Nat), 2, 3] [4, 5] = [1, 2, 3, 4, 5] => rfl
 end Hide
 ```
 
-If structural recursion fails, the equation compiler falls back on well-founded recursion. It tries to infer an instance of ``SizeOf`` for the type of each argument, and then show that each recursive call is decreasing under the lexicographic order of the arguments with respect to ``sizeOf`` measure. If it fails, the error message provides information as to the goal that Lean tried to prove. Lean uses information in the local context, so you can often provide the relevant proof manually using ``have`` in the body of the definition. In this case of well-founded recursion, the defining equations hold only propositionally, and can be accessed using ``simp`` and ``rewrite`` with the name ``foo``.
+Well-founed recursion
```
semorrison, line 542:
```suggestion
Well-founded recursion
```

doc/declarations.md
```diff
@@ -528,9 +566,29 @@ by rw [div]; rfl
 end Hide
 ```
 
+If Lean cannot find a permutation of the arguments for which all recursive calls are decreasing, it will print a table that contains, for every recursive call, which arguments Lean could prove to be decreasing.
+
+Lean will print the termination argument it found if ``set_option showInferredTerminationBy true`` is set.
+
+If Lean does not find the termination argument, or if you want to be explict, you can append a `termination_by` clause to the function definition. It is fo the form
```
semorrison, line 573:
```suggestion
If Lean does not find the termination argument, or if you want to be explict, you can append a `termination_by` clause to the function definition. It is of the form
```

doc/declarations.md
```diff
@@ -528,9 +566,29 @@ by rw [div]; rfl
 end Hide
 ```
 
+If Lean cannot find a permutation of the arguments for which all recursive calls are decreasing, it will print a table that contains, for every recursive call, which arguments Lean could prove to be decreasing.
```
semorrison, line 569: Would it make sense to include an example table here? Or too difficult to keep in sync?

doc/declarations.md
```diff
@@ -528,9 +566,29 @@ by rw [div]; rfl
...
+
+Lean will print the termination argument it found if ``set_option showInferredTerminationBy true`` is set.
+
+If Lean does not find the termination argument, or if you want to be explict, you can append a `termination_by` clause to the function definition. It is fo the form
+```
+termination_by f a₁ … aₙ => e
+```
+where ``f`` is the function name (or ``_``), ``a₁ … aₙ``  are the parameters (or ``_``), and ``e`` is the expression that should be decreasing at each recursive call. The the type of `e` should be an instance of the class ``WellFoundedRelation``, which determins how to compare two values of that type.
```
semorrison, line 577:
```suggestion
where ``f`` is the function name (or ``_``), ``a₁ … aₙ``  are the parameters (or ``_``), and ``e`` is the expression that should be decreasing at each recursive call. The the type of `e` should be an instance of the class ``WellFoundedRelation``, which determines how to compare two values of that type.
```

semorrison, line 577:
```suggestion
where ``f`` is the function name (or ``_``), ``a₁ … aₙ``  are the parameters (or ``_``), and ``e`` is the expression that should be decreasing at each recursive call. The type of `e` should be an instance of the class ``WellFoundedRelation``, which determins how to compare two values of that type.
```

doc/declarations.md
```diff
@@ -528,9 +566,29 @@ by rw [div]; rfl
...
+
+By default, Lean uses the tactic ``decreasing_tactic`` when proving that an argument is decreasing; see its documentation for how to globally extend it. You can also choose to use a different tactic for a given function definition with the clause
+```
+decreasing_by <tac>
+```
+which should come ``termination_by`, if present.
```
semorrison, line 583:
```suggestion
which should come after ``termination_by`, if present.
```

doc/declarations.md
nomeata: Will do

doc/declarations.md
nomeata:
```suggestion
where ``f`` is the function name (or ``_``), ``a₁ … aₙ``  are the parameters (or ``_``), and ``e`` is the expression that should be decreasing at each recursive call. The type of `e` should be an instance of the class ``WellFoundedRelation``, which determins how to compare two values of that type.
```

david-christiansen (changes requested):

nomeata:
We just criss-crossed, I updated it for #3040 and pushed. Darn :-)

doc/declarations.md
```diff
@@ -483,7 +483,42 @@ def baz : Char → Nat
...
+
+def tail {α : Type} : Vector α (n+1) → Vector α n
+| cons h t => t
+
+def map {α β γ : Type} (f : α → β → γ) :
+  ∀ {n}, Vector α n → Vector β n → Vector γ n
+| 0,   nil,       nil       => nil
+| n+1, cons a va, cons b vb => cons (f a b) (map f va vb)
+
+end Vector
```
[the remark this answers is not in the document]
nomeata: Yuck, `autoImplicits` :-)

Is that something we want to encourage, especially in places where people do _not_ have the benefit of IDE support?

doc/declarations.md
```diff
@@ -483,7 +483,42 @@ def baz : Char → Nat
...
+
+Recursive functions
+===================
+
+Lean has to ensure that a recursive function definition is terminating, and has two strategies for that.
+
+Structural recursion
+--------------------
+
+If the definition of a function contains recursive calls, the equation compiler tries to interpret the definition as a structural recursion. In order for that to succeed, the recursive arguments must be subterms of the corresponding arguments on the left-hand side. The function is then defined using a *course of values* recursion, using automatically generated functions ``below`` and ``brec`` in the namespace corresponding to the inductive type of the recursive argument. In this case the defining equations hold definitionally, possibly with additional case splits.
```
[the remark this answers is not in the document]
nomeata: Previouisly, all of this was documented under “equation compiler”, and so far I don’t have good name for the component that processes recursive definitions (besides “derecursifier”, but I doubt that we should really use this.). It may make sense to avoid giving this component a name alltogether.

Feel free to suggest better wordings! But remember that this is a MVP-doc-PR to have something, and maybe more serious copyediting should happen as part of a more complete overhaul of the documentation, and maybe by a professional technical writer.

doc/declarations.md
```diff
@@ -528,9 +566,29 @@ by rw [div]; rfl
 end Hide
 ```
 
+If Lean cannot find a permutation of the arguments for which all recursive calls are decreasing, it will print a table that contains, for every recursive call, which arguments Lean could prove to be decreasing.
+
+Lean will print the termination argument it found if ``set_option showInferredTerminationBy true`` is set.
```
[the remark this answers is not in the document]
nomeata: For the table I wrote it while you posted.

Adding an example for the output from ``showInferredTerminationBy`` here is almost not helpful, because the syntax it is presented in is discussed later, and clearly recognizable as that.

doc/declarations.md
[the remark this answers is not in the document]
digama0: I don't think there is any indication that lean discourages autoimplicits. That's a mathlib thing. (Besides this, I think for type arguments they are quite unobjectionable, being the default behavior in haskell and ocaml.)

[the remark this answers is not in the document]
david-christiansen: I agree - and especially in a context like this, decreasing syntactic noise seems useful.

doc/declarations.md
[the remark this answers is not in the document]
david-christiansen: I like not mentioning a component name. What about "Lean" instead of "the equation compiler" here? I think that's just as clear, and less likely to throw people off.

doc/declarations.md
```diff
@@ -528,9 +566,53 @@ by rw [div]; rfl
 end Hide
 ```
 
+If Lean cannot find a permutation of the arguments for which all recursive calls are decreasing, it will print a table that contains, for every recursive call, which arguments Lean could prove to be decreasing. For example, a function with three recursive calls and four paramters might cause the following message to be printed
```
david-christiansen, line 569:
```suggestion
If Lean cannot find a permutation of the arguments for which all recursive calls are decreasing, it will print a table that contains, for every recursive call, which arguments Lean could prove to be decreasing. For example, a function with three recursive calls and four parameters might cause the following message to be printed
```

doc/declarations.md
```diff
@@ -528,9 +566,53 @@ by rw [div]; rfl
...
+1) 39:6-27  =  =  _  =
+2) 40:6-25  =  ?  _  <
+3) 41:6-25  <  _  _  _
+Please use `termination_by` to specify a decreasing measure.
+```
+
+This table should be read as follows:
+
+ * The first recursive call, in line 39, arguments 1, 2 and 4 are equal to the function's paramters.
+ * The second recursive call, in line 40, has an equal first, a smaller forth argument, and nothing could be inferred for the second argument.
```
david-christiansen, line 585:
```suggestion
 * The second recursive call, in line 40, has an equal first, a smaller fourth argument, and nothing could be inferred for the second argument.
```

doc/declarations.md
```diff
@@ -528,9 +566,53 @@ by rw [div]; rfl
...
+ * The third recursive call, in line 41, has a decreasing first argument.
+ * No other proofs were attempted, either because the paramter has a type without a non-trivial ``WellFounded`` instance (parameter 3), or because it is already clear that no decreasing measure can be found.
```
david-christiansen, line 587:
```suggestion
 * No other proofs were attempted, either because the parameter has a type without a non-trivial ``WellFounded`` instance (parameter 3), or because it is already clear that no decreasing measure can be found.
```

doc/declarations.md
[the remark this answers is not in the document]
nomeata:
```suggestion
If the definition of a function contains recursive calls, Lean tries to interpret the definition as a structural recursion. In order for that to succeed, the recursive arguments must be subterms of the corresponding arguments on the left-hand side.

The function is then defined using a *course of values* recursion, using automatically generated functions ``below`` and ``brec`` in the namespace corresponding to the inductive type of the recursive argument. In this case the defining equations hold definitionally, possibly with additional case splits.
```

doc/declarations.md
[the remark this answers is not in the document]
david-christiansen: Sounds good

doc/declarations.md
```diff
@@ -587,29 +670,31 @@ def num_consts_lst : List Term → Nat
...
-
-```lean
-universe u
-
-inductive Vector (α : Type u) : Nat → Type u
-| nil  : Vector α 0
-| cons : α → Vector α n → Vector α (n+1)
-
-namespace Vector
+In a set of mutually recursive function, either all or no functions must have an explicit termination argument (``termination_by``). A change of the default termination tactic (``decreasing_by``) only affects the proofs about the recursive calls of that function, not the other functiosn in the group.
```
david-christiansen, line 673:
```suggestion
In a set of mutually recursive function, either all or no functions must have an explicit termination argument (``termination_by``). A change of the default termination tactic (``decreasing_by``) only affects the proofs about the recursive calls of that function, not the other functions in the group.
```

doc/declarations.md
```diff
@@ -587,29 +670,31 @@ def num_consts_lst : List Term → Nat
...
 
-end Vector
+```
+theorem even_of_odd_succ : ∀ n, Odd (n + 1) → Even n
+| _, odd_succ n h => h
+termination_by n h => h
+  where
+    theorem odd_of_even_succ : ∀ n, Even (n + 1) → Odd n
+    | _, even_succ n h => h
+    termination_by n h => h
```
david-christiansen, line 697: We typically indent the cases:
```suggestion
theorem even_of_odd_succ : ∀ n, Odd (n + 1) → Even n
  | _, odd_succ n h => h
termination_by n h => h
  where
    theorem odd_of_even_succ : ∀ n, Even (n + 1) → Odd n
      | _, even_succ n h => h
    termination_by n h => h
```

doc/functions.md
```diff
@@ -32,7 +32,10 @@ def fact x :=
 
 #eval fact 100
 ```
-By default, Lean only accepts total functions. The `partial` keyword should be used when Lean cannot
+By default, Lean only accepts total functions (see [The Equation
+Compiler](declarations.md#_the_equation_compiler) for how Lean determines
+whether functions are total).
+The `partial` keyword should be used when Lean cannot
 establish that a function always terminates.
```
david-christiansen, line 39: What about:
```suggestion
The `partial` keyword may be used to define a recursive function without a termination proof; partial functions compute in compiled programs, but are opaque in proofs and during type checking.
```

david-christiansen (approved):
I think it's basically there now, modulo typos and nitpicks.

doc/declarations.md
```diff
@@ -504,7 +539,10 @@ example : append [(1 : Nat), 2, 3] [4, 5] = [1, 2, 3, 4, 5] => rfl
...
+Well-founded recursion
+---------------------
+
+If structural recursion fails, the equation compiler falls back on well-founded recursion. It tries to infer an instance of ``SizeOf`` for the type of each argument, and then tries to find a permutation of the arguments such that each recursive call is decreasing under the lexicographic order with respect to ``sizeOf`` measures.  Lean uses information in the local context, so you can often provide the relevant proof manually using ``have`` in the body of the definition. In this case of well-founded recursion, the defining equations hold only propositionally, and can be accessed using ``simp`` and ``rewrite`` with the name ``foo``.
```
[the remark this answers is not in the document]
nomeata:
```suggestion
If structural recursion fails, the equation compiler falls back on well-founded recursion. It tries to infer an instance of ``SizeOf`` for the type of each argument, and then tries to find a permutation of the arguments such that each recursive call is decreasing under the lexicographic order with respect to ``sizeOf`` measures.  Lean uses information in the local context, so you can often provide the relevant proof manually using ``have`` in the body of the definition.

In the case of well-founded recursion, the equation used to declare the function holds only propositionally, but not definitionally, and can be accessed using ``unfold``, ``simp`` and ``rewrite`` with the function name (for example ``unfold foo`` or ``simp [foo]``, where ``foo`` is the function defined with well-founded recursion).
```

(Not perfect, just aiming for good enough at this point…)

doc/declarations.md
```diff
@@ -528,9 +569,53 @@ by rw [div]; rfl
...
+If Lean does not find the termination argument, or if you want to be explict, you can append a `termination_by` clause to the function definition, after the function's body, but before the `where` clause if present. It is of the form
+```
+termination_by e
+```
+where ``e`` is an expression that depends on the parameters of the function and should be decreasing at each recursive call. The type of `e` should be an instance of the class ``WellFoundedRelation``, which determines how to compare two values of that type.
+
+If ``f`` has parameters “after the ``:``” (a `fun` after the `:=`, or function equations), these must be given names using the syntax
+```
+termination_by a₁ … aₙ => e
+```
```
nomeata, line 604:
````suggestion
If ``f`` has parameters “after the ``:``” (for example when defining functions via patterns using `|`), then these can be brought into scope using the syntax.
```
termination_by a₁ … aₙ => e
```
````

No longer a ”must”.

File: doc/declarations.md [hunk 17]
```diff
@@ -483,7 +483,43 @@ def baz : Char → Nat
...
+
+end Vector
+```
+
+.. _recursive_functions:
+
+Recursive functions
+===================
+
+Lean must ensure that a recursive functions terminate, for which there are two strategies: _structural recursion_, in which all recursive calls are made on smaller parts of the input data, and _well-founded recursion_, in which recursive calls are justified by showing that arguments to recursive calls are smaller according to some other measure.
```
semorrison, line 515:
```suggestion
Lean must ensure that a recursive function terminates, for which there are two strategies: _structural recursion_, in which all recursive calls are made on smaller parts of the input data, and _well-founded recursion_, in which recursive calls are justified by showing that arguments to recursive calls are smaller according to some other measure.
```

doc/declarations.md
```diff
@@ -483,7 +483,43 @@ def baz : Char → Nat
...
+
+Structural recursion
+--------------------
+
+If the definition of a function contains recursive calls, Lean tries to interpret the definition as a structural recursion. In order for that to succeed, the recursive arguments must be subterms of the corresponding arguments on the left-hand side.
```
semorrison, line 520:
```suggestion
If the definition of a function contains recursive calls, Lean first tries to interpret the definition as a structural recursion. In order for that to succeed, the recursive arguments must be subterms of the corresponding arguments on the left-hand side.
```
Helpfully foreshadows that there is a fall-back?

doc/declarations.md
```diff
@@ -528,9 +569,53 @@ by rw [div]; rfl
 end Hide
 ```
 
+If Lean cannot find a permutation of the arguments for which all recursive calls are decreasing, it will print a table that contains, for every recursive call, which arguments Lean could prove to be decreasing. For example, a function with three recursive calls and four parameters might cause the following message to be printed
+
+```
+example.lean:37:0-43:31: error: Could not find a decreasing measure.
+The arguments relate at each recursive call as follows:
+(<, ≤, =: relation proved, ? all proofs failed, _: no proof attempted)
```
semorrison, line 577: Colons are inconsistent here: should there be one after the `?`?

The colons are't that great, in any case, as I'm tempted to misinterpret `=:` as an atom.

doc/declarations.md
```diff
@@ -528,9 +569,53 @@ by rw [div]; rfl
...
+           x1 x2 x3 x4
+1) 39:6-27  =  =  _  =
+2) 40:6-25  =  ?  _  <
+3) 41:6-25  <  _  _  _
+Please use `termination_by` to specify a decreasing measure.
+```
+
+This table should be read as follows:
+
+ * The first recursive call, in line 39, arguments 1, 2 and 4 are equal to the function's paramters.
```
semorrison, line 587:
```suggestion
 * The first recursive call, in line 39, arguments 1, 2 and 4 are equal to the function's parameters.
```

doc/declarations.md
```diff
@@ -528,9 +569,53 @@ by rw [div]; rfl
...
+ * The second recursive call, in line 40, has an equal first, a smaller fourth argument, and nothing could be inferred for the second argument.
```
semorrison, line 588:
```suggestion
 * The second recursive call, in line 40, has an equal first argument, a smaller fourth argument, and nothing could be inferred for the second argument.
```

doc/declarations.md
```diff
@@ -528,9 +569,53 @@ by rw [div]; rfl
...
+ * The third recursive call, in line 41, has a decreasing first argument.
+ * No other proofs were attempted, either because the parameter has a type without a non-trivial ``WellFounded`` instance (parameter 3), or because it is already clear that no decreasing measure can be found.
+
+
+Lean will print the termination argument it found if ``set_option showInferredTerminationBy true`` is set.
+
+If Lean does not find the termination argument, or if you want to be explict, you can append a `termination_by` clause to the function definition, after the function's body, but before the `where` clause if present. It is of the form
```
semorrison, line 595:
```suggestion
If Lean does not find the termination argument, or if you want to be explicit, you can append a `termination_by` clause to the function definition, after the function's body, but before the `where` clause if present. It is of the form
```

doc/declarations.md
```diff
@@ -587,29 +672,31 @@ def num_consts_lst : List Term → Nat
...
+| _, even_succ n h => h
+termination_by n h => h
+end
+```
 
-def map {α β γ : Type} (f : α → β → γ) :
-  ∀ {n}, Vector α n → Vector β n → Vector γ n
-| 0,   nil,       nil       => nil
-| n+1, cons a va, cons b vb => cons (f a b) (map f va vb)
+Another way to express mutual recursion using local function definitions in ``where`` or ``let rec`` clauses: these can be mutually recursive with each other and their containing function:
```
semorrison, line 690:
```suggestion
Another way to express mutual recursion is using local function definitions in ``where`` or ``let rec`` clauses: these can be mutually recursive with each other and their containing function:
```

doc/declarations.md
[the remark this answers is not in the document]
nomeata:
```suggestion
 * In the first recursive call, in line 39, arguments 1, 2 and 4 are equal to the function's parameters.
```

doc/declarations.md
nomeata: Good point. How about putting the prose first and using parentheses?
```suggestion
Relation proved (<, ≤, =), all proofs failed (?), no proof attempted (_)
```

PR #3024, opened 2023-12-05
fix: WF.Fix: deduplicate subsumed goals before running tactic

nomeata:
before code like

    def dup (a : Nat) (b : Nat := a) := a + b

    def rec : Nat → Nat
     | 0 => 1
     | n+1 => dup (dup (dup (rec n)))
    decreasing_by decreasing_tactic

would run the `decreasing_tactic` 8 tims, because the recursive call
`rec n` gets duplicate due to the default paramter. Similar effects can
be observed due to dependent types or tactics like `cases`.

This is wasteful, and is confusing to the user when they use
`decreasing_by` interactively. Therfore, we now go through the proof
obligations (MVars) and if solving one would imply solving another one,
we assign the mvars to each other accordingly.

This PR is a sibling of #3004.

### DIFF
--- a/src/Lean/Data/Array.lean
+++ b/src/Lean/Data/Array.lean
@@ -1,0 +1,46 @@
+/-
+Copyright (c) 2023 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Joachim Breitner
+-/
+
+import Init.Data.Array
+
+namespace Array
+
+/-!
+This module contains utility functions involving Arrays that are useful in a few places
+of the lean code base, but too specialized to live in `Init.Data.Array`, which arguably
+is part of the public, user-facing standard library.
+-/
+
+/--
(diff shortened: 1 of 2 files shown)


nomeata:
I’ll run this against mathlib once `9290b491bb16ec897a5cb9ffd98c8416af30bfb2` is tagged as `nightly-with-mathlib`.

nomeata:
Got mathlib perf results, nothing to see. <http://speed.lean-fro.org/mathlib4/compare/6d60f603-6faa-4727-9a04-42abb15e1477/to/535315ad-b4ab-4d6a-b3fd-67e8b2ff12c7>

There is an “open Mathlib — wall-clock 7.8%” but I think it’s a fluke: It doesn't show up when comparing this branch with the latest `master`, and it it seems unlikely that there is a change in `testing-nightly` that improved `open Mathlib` by exactly the amount that this regressed then. Also “open Mathlib – instructions” goes down.

PR #3025, opened 2023-12-05
doc: add migration guide for per-package server options

mhuisi:
This PR adjusts `RELEASES.md` to match the recently adjusted release notes.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -33,12 +33,6 @@ v4.4.0
   package SomePackage where
     leanOptions := #[⟨`pp.unicode.fun, true⟩]
   ```
-  
-  A package configuration that does not intend to set `moreLeanArgs` can use `moreServerOptions` instead of `leanOptions`:
-  ```lean
-  package SomePackage where
-    moreServerOptions := #[⟨`pp.unicode.fun, true⟩]
-  ```
 * [Rename request handler](https://github.com/leanprover/lean4/pull/2462).
 * [Import auto-completion](https://github.com/leanprover/lean4/pull/2904).
 * [`pp.beta`` to apply beta reduction when pretty printing](https://github.com/leanprover/lean4/pull/2864).


Kha (approved):

RELEASES.md
```diff
@@ -14,10 +14,34 @@ v4.5.0 (development in progress)
...
+  
+  package SomePackage where
+    moreServerArgs := moreServerArgs
+    moreLeanArgs := moreLeanArgs
+  ```
+  
+  ... can be updated to the following package declaration to use per-package options:
+  ```lean
+  package SomePackage where
+    leanOptions := ⟨`pp.unicode.fun, true⟩
```
digama0, line 34:
```suggestion
    leanOptions := #[⟨`pp.unicode.fun, true⟩]
```

RELEASES.md
```diff
@@ -14,10 +14,34 @@ v4.5.0 (development in progress)
...
+  ```
+  
+  A package configuration that does not intend to set `moreLeanArgs` can use `moreServerOptions` instead of `leanOptions`:
+  ```lean
+  package SomePackage where
+    moreServerOptions := ⟨`pp.unicode.fun, true⟩
```
digama0, line 40:
```suggestion
    moreServerOptions := #[⟨`pp.unicode.fun, true⟩]
```

digama0: This should also use an example that makes sense to use only in the server. (Do we have any?) Pretty printing settings should not be server-only, they can break tests.

mhuisi: I think I'll just remove this part of the migration guide entirely.

RELEASES.md
```diff
@@ -14,10 +14,34 @@ v4.5.0 (development in progress)
...
+  ```lean
+  package SomePackage where
+    leanOptions := #[⟨`pp.unicode.fun, true⟩]
+  ```
+  
+  A package configuration that does not intend to set `moreLeanArgs` can use `moreServerOptions` instead of `leanOptions`:
+  ```lean
+  package SomePackage where
+    moreServerOptions := #[⟨`pp.unicode.fun, true⟩]
+  ```
```
mhuisi, line 41:
```suggestion
```

PR #3030, opened 2023-12-05
chore: set `warningAsError` in CI only

Kha:
Don't fail local builds because of this


nomeata (commented):
Also set it in nix, so that fails quickly?

Kha:
> Also set it in nix, so that fails quickly?

Already there. One more good reason for not using Nix for interactive development in fact :) .

PR #3033, opened 2023-12-06
feat: drop support for termination_by'

nomeata:
until around 7fe6881 the way to define well-founded recursions was to
specify a `WellFoundedRelation` on the argument explicitly. This was
rather low-level, for example one had to predict the packing of multiple
arguments into `PProd`s, the packing of mutual functions into `PSum`s,
and the cliques that were calculated.

Then the current `termination_by` syntax was introduced, where you
specify the termination argument at a higher level (one clause per
functions, unpacked arguments), and the `WellFoundedRelation` is found
using type class resolution.

The old syntax was kept around as `termination_by'`. This is not used
anywhere in the lean, std, mathlib or the theorem-proving-in-lean repositories,
and three occurrences I found in the wild can do without

In particular, it should be possible to express anything that the old syntax
supported also with the new one, possibly requiring a helper type with a
suitable instance, or the following generic wrapper that now lives in std
```
def wrap {α : Sort u} {r : α → α → Prop} (h : WellFounded r) (x : α) : {x : α // Acc r x}
```

Since the old syntax is unused, has an unhelpful name and relies on
internals, this removes the support. Now is a good time before the
refactoring that's planned in #2921.

The test suite was updated without particular surprises.

The parametric `terminationHint` parser is gone, which means we can
match on syntax more easily now, in `expandDecreasingBy?`.


nomeata:
Silly commit convention, removal of a feature is strange `feat` :-)

nomeata:
> I think this should not be done without a clear deprecation and migration strategy, including the WithWFRel type.

Yeah, that is something I wonder as well: How fast can we still move? Do you have evidence that this is used anywhere? And moreover that any of these theses are not better served by `termination_by` (rather than `WithWFRel`)?

You are probably not on board with “Renaming the old syntax as `termination_by'`, so that users start using `termination_by` is the deprecation strategy.”?

Also, would you argue we need a migration strategy for #2921 as well?

digama0:
> You are probably not on board with “Renaming the old syntax as `termination_by'`, so that users start using `termination_by` is the deprecation strategy, and two years was enough” :-)?

No, there has never been any indication that this was on its way out. Primes in lean generally don't indicate deprecation, they indicate a less common variation on the operation for which we couldn't think of a better name.

nomeata:
That’s what I thought :-)

I’ll escalate this to @leodemoura.

Personally I hope we are still in a phase of lean development where we can still clean up things, even if it breaks hypothetical code, and where we can use std and mathlib (and soon more stuff on reservior) to distinguish actual breakage from hypothetical breakage.

digama0:
> Personally I hope we are still in a phase of lean development where we can still clean up things,

I think you misunderstand. Having a deprecation period doesn't mean you can't clean up, and even if you don't do a deprecation period a migration guide for breakage is essential to keep users happy. Lean is moving fast and breaking my code all the time, and this is off-putting to many users we want to attract.

nomeata:
>  if you don't do a deprecation period a migration guide for breakage is essential to keep users happy.

Oh, that I am absolutely on board with; this change should be well documented in the release notes.

nomeata:
> this is not used anywhere in the lean, std

I stand corrected, my `std` checkout was a tad old, and just there weeks ago [a use of `termination_by'` was introduced](https://github.com/leanprover/std4/pull/371) there. That's a bit embarrasing.

nomeata:
I agree!

digama0:
There are a few separate concerns here:

1. Usage in mathport: here we want something that can be applied fully automatically as a syntactic transformation
1. Grandfathered usage in user code: here we want something easy to explain, but not necessarily a pure syntactic transformation
1. Deliberate usage in user code: This is for cases where the old syntax is genuinely easier to use than the new syntax.

For (1) I'll do my best with whatever we end up with, but my quick take is that there is no simple replacement currently, because of the aforementioned packing process. For (2) I think your suggestion is sufficient. For (3) I think it might be worth considering a syntax similar to `termination_by` which lets you pass the instance instead of the value, or put an at-sign or optional argument somewhere. It's a bit awkward that the *only* way to pass that argument is via typeclass inference.

digama0:
One thing which I don't like about the current system which seems vaguely (3)-like is that you have to declare an instance in order to make the definition, even though it may just be something private to the definition. This pollutes the namespace with additional declarations (sometimes with [long names](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/ridiculous.20instance.20names/near/406224095)), and unlike most cases you can't use a local `have` to introduce the instance, because `termination_by` lives "outside" the definition and there is nothing "outside" the `termination_by` where you would otherwise be able to put the `have` declaration. That's how you end up with `private local instance` like in the std example.

nomeata:
Right, but you can work around that using `WellFounded.wrap`, where you do not have to define new instances.

So a work-around exists, the question is whether the work-around is ok enough, or bad enough to warrant supporting (and documenting, explaining, including the internals of bundling paramters etc.) a separate `termination_relation`, isn't it?

nomeata:
>  (I would not include the internals of bundling parameters, I would prefer a way to specify the bundling pattern. But I can't think of a nice syntax for this which isn't suspiciously similar to termination_by.)

Right. I guess you’d end up with something like “`termination_by` + explicitly specified relation”, which may be a bit more convenient than `wf.wrap …` inside `termination_by`, but target audience that benefits from this small DX delta might not be worth the effort on the lean side.

PR #3034, opened 2023-12-06
chore: default `compiler.enableNew` to false until development restarts

Kha:


leodemoura (approved):

Kha:
@leodemoura There is a number of tests testing new compiler output that are now broken. We could disable the option for external users only but I think the time savings for building Lean itself are significant enough to justify moving these tests into a new folder that is not part of the suite for now.

Kha:
I've moved the new compiler tests to a separate directory that is currently not included in the test suite

Kha:
!bench

Issue #3031, opened 2023-12-05
Using `open` or `export` does not make names accessible for extended field notation (dot notation)


kmill:
### Description

If you try to use `open` or `export` to import names to use for dot notation, it does not work. Confusingly, you can have `String.a "x"` succeed but `"x".a` fail.

### Context

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/Structuring.20Code.20.2F.20importing.20Modules/near/406170432) where this was brought up, during an Advent of Code problem.

### Steps to Reproduce

Observe that `"x".a` produces an error in the following
```lean
namespace Common
namespace String

def a (s : String) : Nat := s.length

end String
end Common

open Common (String.a)

#eval String.a "x"

/-
invalid field 'a', the environment does not contain 'String.a'
  "x"
has type
  String
-/
#eval "x".a

export Common (String.a)

/-
invalid field notation, function 'Common.String.a' does not have argument
with type (Common.String ...) that can be used, it must be explicit or
implicit with a unique name
-/
#eval "x".a
```

**Expected behavior:** The second and third `#eval`s have the same interpretation as the first.

**Actual behavior:** Only the first succeeds. The second `#eval` (when the name is just `open`) doesn't see `String.a` at all, and the third `#eval` (when the name is `export`ed) it sees it but then fails to process it correctly.

### Versions

MacOS arm64
```lean
/- "4.3.0-rc2" -/
#eval Lean.versionString
```

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

digama0:
I would love this feature, it would be a way to avoid the name-clash-inducing practice of defining functions outside the project namespace so that you get dot notation on types you don't own.

Kha:
I think this is an RFC :) . It is a sensible enhancement suggestion but it is not an accidental oversight in the implementation.

Kha:
I should also mention a much older proposal of allowing *any* top-level name in scope as a target for extended field notation. This is especially interesting for typeclasses where it would allow us to use `x.toString` uniformly instead of the status quo where it only works sometimes depending on whether that exact declaration name was used to construct the instance. It is similar to how in Rust you need to `use` a trait to get dot access to its members.
[...]

PR #3035, opened 2023-12-06
fix: move Lean.List.toSMap to List.toSMap

digama0:
This definition was clearly meant to be in the `List` namespace, but it is also in a `namespace Lean` so it ended up as `Lean.List.toSMap` instead of `List.toSMap`. It would be nice if #3031 made this unnecessary, but for now this seems to be the convention.

I noticed this because of another side effect: it defines `Lean.List` as a namespace, which means that
```lean
import Std

namespace Lean
open List

#check [1] <+ [2]
```
does not work as expected, it opens the `Lean.List` namespace instead of the `List` namespace. Should there be a regression test to ensure that the `Lean.List` namespace (and maybe others) are not accidentally created? (Unfortunately this puts a bit of a damper on #3031.)

### DIFF
--- a/src/Lean/Data/SMap.lean
+++ b/src/Lean/Data/SMap.lean
@@ -97,7 +97,7 @@ def toList (m : SMap α β) : List (α × β) :=
 
 end SMap
 
-def List.toSMap [BEq α] [Hashable α] (es : List (α × β)) : SMap α β :=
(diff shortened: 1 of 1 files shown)


leodemoura (approved):

semorrison:
This needs to be rebased onto something more recent (e.g. `nightly`), so that the new CI requirements for the merge queue run.

PR #3037, opened 2023-12-07
doc: document constructors of `TransparencyMode`

eric-wieser:
Taken from https://github.com/leanprover-community/lean4-metaprogramming-book/blob/master/md/main/04_metam.md#transparency

I can never remember which way around `reducible` and `default` go, and this avoids me needing to leave the editor to find out.


nomeata:
Thanks!

PR #3040, opened 2023-12-08
feat: per-function termination hints

### LINKED ISSUE #2921: RFC: Run decreasing_by tactic on all goals, not each goal

nomeata:
This change

 * moves `termination_by` and `decreasing_by` next to the function they apply to
 * simplify the syntax of `termination_by`
 * apply the `decreasing_by` goal to all goals at once, for better interactive use.

See the section in `RELEASES.md` for more details and migration advise.

This is a hard breaking change, requiring developers to touch every `termination_by` in their code base. We decided to still do it as a hard-breaking change, because supporting both old and new syntax at the same time would be non-trivial, and not save that much. Moreover, this requires changes to some metaprograms that developers might have written, and supporting both syntaxes at the same time would make _their_ migration harder.

Std and mathlib updates are prepared at

* https://github.com/leanprover/std4/pull/446
* https://github.com/leanprover-community/mathlib4/pull/9019

Fixes #2921 and #3081.

### DIFF
--- a/src/Init/Data/Array/Basic.lean
+++ b/src/Init/Data/Array/Basic.lean
@@ -276,8 +276,8 @@ def mapM {α : Type u} {β : Type v} {m : Type v → Type w} [Monad m] (f : α �
       map (i+1) (r.push (← f as[i]))
     else
       pure r
+  termination_by as.size - i
   map 0 (mkEmpty as.size)
-termination_by map => as.size - i
 
 @[inline]
 def mapIdxM {α : Type u} {β : Type v} {m : Type v → Type w} [Monad m] (as : Array α) (f : Fin as.size → α → m β) : m (Array β) :=
@@ -348,12 +348,12 @@ def anyM {α : Type u} {m : Type → Type w} [Monad m] (p : α → m Bool) (as :
           loop (j+1)
       else
         pure false
+      termination_by stop - j
     loop start
   if h : stop ≤ as.size then
     any stop h
   else
     any as.size (Nat.le_refl _)
-termination_by loop i j => stop - j
 
 @[inline]
 def allM {α : Type u} {m : Type → Type w} [Monad m] (p : α → m Bool) (as : Array α) (start := 0) (stop := as.size) : m Bool :=
@@ -523,7 +523,7 @@ def isEqvAux (a b : Array α) (hsz : a.size = b.size) (p : α → α → Bool) (
      p a[i] b[i] && isEqvAux a b hsz p (i+1)
   else
     true
-termination_by _ => a.size - i
+termination_by a.size - i
 
 @[inline] def isEqv (a b : Array α) (p : α → α → Bool) : Bool :=
   if h : a.size = b.size then
@@ -627,7 +627,7 @@ def indexOfAux [BEq α] (a : Array α) (v : α) (i : Nat) : Option (Fin a.size) 
     if a.get idx == v then some idx
     else indexOfAux a v (i+1)
   else none
-termination_by _ => a.size - i
+termination_by a.size - i
 
 def indexOf? [BEq α] (a : Array α) (v : α) : Option (Fin a.size) :=
   indexOfAux a v 0
@@ -659,7 +659,7 @@ where
       loop as (i+1) ⟨j-1, this⟩
     else
       as
-termination_by _ => j - i
+termination_by j - i
 
 def popWhile (p : α → Bool) (as : Array α) : Array α :=
   if h : as.size > 0 then
@@ -669,7 +669,7 @@ def popWhile (p : α → Bool) (as : Array α) : Array α :=
       as
   else
     as
-termination_by popWhile as => as.size
+termination_by as.size
 
 def takeWhile (p : α → Bool) (as : Array α) : Array α :=
   let rec go (i : Nat) (r : Array α) : Array α :=
@@ -681,8 +681,8 @@ def takeWhile (p : α → Bool) (as : Array α) : Array α :=
         r
     else
       r
+    termination_by as.size - i
   go 0 #[]
-termination_by go i r => as.size - i
 
 def eraseIdxAux (i : Nat) (a : Array α) : Array α :=
   if h : i < a.size then
@@ -692,7 +692,7 @@ def eraseIdxAux (i : Nat) (a : Array α) : Array α :=
     eraseIdxAux (i+1) a'
   else
     a.pop
-termination_by _ => a.size - i
+termination_by a.size - i
 
 def feraseIdx (a : Array α) (i : Fin a.size) : Array α :=
   eraseIdxAux (i.val + 1) a
@@ -707,7 +707,7 @@ def eraseIdxSzAux (a : Array α) (i : Nat) (r : Array α) (heq : r.size = a.size
     eraseIdxSzAux a (i+1) (r.swap idx idx1) ((size_swap r idx idx1).trans heq)
   else
     ⟨r.pop, (size_pop r).trans (heq ▸ rfl)⟩
-termination_by _ => r.size - i
+termination_by r.size - i
 
 def eraseIdx' (a : Array α) (i : Fin a.size) : { r : Array α // r.size = a.size - 1 } :=
   eraseIdxSzAux a (i.val + 1) a rfl
@@ -726,10 +726,10 @@ def erase [BEq α] (as : Array α) (a : α) : Array α :=
       loop as ⟨j', by rw [size_swap]; exact j'.2⟩
     else
       as
+    termination_by j.1
   let j := as.size
   let as := as.push a
   loop as ⟨j, size_push .. ▸ j.lt_succ_self⟩
-termination_by loop j => j.1
 
 /-- Insert element `a` at position `i`. Panics if `i` is not `i ≤ as.size`. -/
 def insertAt! (as : Array α) (i : Nat) (a : α) : Array α :=
@@ -779,7 +779,7 @@ def isPrefixOfAux [BEq α] (as bs : Array α) (hle : as.size ≤ bs.size) (i : N
       false
   else
     true
-termination_by _ => as.size - i
+termination_by as.size - i
 
 /-- Return true iff `as` is a prefix of `bs`.
 That is, `bs = as ++ t` for some `t : List α`.-/
@@ -800,7 +800,7 @@ private def allDiffAux [BEq α] (as : Array α) (i : Nat) : Bool :=
     allDiffAuxAux as as[i] i h && allDiffAux as (i+1)
   else
     true
-termination_by _ => as.size - i
+termination_by as.size - i
 
 def allDiff [BEq α] (as : Array α) : Bool :=
   allDiffAux as 0
@@ -815,7 +815,7 @@ def allDiff [BEq α] (as : Array α) : Bool :=
       cs
   else
     cs
-termination_by _ => as.size - i
+termination_by as.size - i
 
 @[inline] def zipWith (as : Array α) (bs : Array β) (f : α → β → γ) : Array γ :=
   zipWithAux f as bs 0 #[]
--- a/src/Init/Data/Array/BasicAux.lean
+++ b/src/Init/Data/Array/BasicAux.lean
@@ -47,7 +47,7 @@ where
       have hlt : i < as.size := Nat.lt_of_le_of_ne hle h
       let b ← f as[i]
       go (i+1) ⟨acc.val.push b, by simp [acc.property]⟩ hlt
-termination_by go i _ _ => as.size - i
+  termination_by as.size - i
 
 @[inline] private unsafe def mapMonoMImp [Monad m] (as : Array α) (f : α → m α) : m (Array α) :=
   go 0 as
--- a/src/Init/Data/Array/DecidableEq.lean
+++ b/src/Init/Data/Array/DecidableEq.lean
@@ -20,7 +20,7 @@ theorem eq_of_isEqvAux [DecidableEq α] (a b : Array α) (hsz : a.size = b.size)
   · have heq : i = a.size := Nat.le_antisymm hi (Nat.ge_of_not_lt h)
     subst heq
     exact absurd (Nat.lt_of_lt_of_le high low) (Nat.lt_irrefl j)
-termination_by _ => a.size - i
+termination_by a.size - i
 
 theorem eq_of_isEqv [DecidableEq α] (a b : Array α) : Array.isEqv a b (fun x y => x = y) → a = b := by
   simp [Array.isEqv]
@@ -36,7 +36,7 @@ theorem isEqvAux_self [DecidableEq α] (a : Array α) (i : Nat) : Array.isEqvAux
   split
   case inl h => simp [h, isEqvAux_self a (i+1)]
   case inr h => simp [h]
-termination_by _ => a.size - i
+termination_by a.size - i
 
 theorem isEqv_self [DecidableEq α] (a : Array α) : Array.isEqv a a (fun x y => x = y) = true := by
   simp [isEqv, isEqvAux_self]
--- a/src/Init/Data/Array/QSort.lean
+++ b/src/Init/Data/Array/QSort.lean
@@ -26,8 +26,8 @@ def qpartition (as : Array α) (lt : α → α → Bool) (lo hi : Nat) : Nat × 
     else
       let as := as.swap! i hi
       (i, as)
+    termination_by hi - j
   loop as lo lo
-termination_by _ => hi - j
 
 @[inline] partial def qsort (as : Array α) (lt : α → α → Bool) (low := 0) (high := as.size - 1) : Array α :=
   let rec @[specialize] sort (as : Array α) (low high : Nat) :=
--- a/src/Init/Data/Nat/Power2.lean
+++ b/src/Init/Data/Nat/Power2.lean
@@ -21,8 +21,8 @@ where
       go (power * 2) (Nat.mul_pos h (by decide))
     else
       power
-termination_by go p h => n - p
-decreasing_by simp_wf; apply nextPowerOfTwo_dec <;> assumption
+  termination_by n - power
+  decreasing_by simp_wf; apply nextPowerOfTwo_dec <;> assumption
 
 def isPowerOfTwo (n : Nat) := ∃ k, n = 2 ^ k
 
@@ -48,7 +48,7 @@ where
     split
     . exact isPowerOfTwo_go (power*2) (Nat.mul_pos h₁ (by decide)) (Nat.mul2_isPowerOfTwo_of_isPowerOfTwo h₂)
     . assumption
-termination_by isPowerOfTwo_go p _ _ => n - p
-decreasing_by simp_wf; apply nextPowerOfTwo_dec <;> assumption
+  termination_by n - power
+  decreasing_by simp_wf; apply nextPowerOfTwo_dec <;> assumption
 
 end Nat
--- a/src/Init/Data/String/Basic.lean
+++ b/src/Init/Data/String/Basic.lean
@@ -159,7 +159,7 @@ def posOfAux (s : String) (c : Char) (stopPos : Pos) (pos : Pos) : Pos :=
       have := Nat.sub_lt_sub_left h (lt_next s pos)
       posOfAux s c stopPos (s.next pos)
   else pos
-termination_by _ => stopPos.1 - pos.1
+termination_by stopPos.1 - pos.1
 
 @[inline] def posOf (s : String) (c : Char) : Pos :=
   posOfAux s c s.endPos 0
@@ -171,7 +171,7 @@ def revPosOfAux (s : String) (c : Char) (pos : Pos) : Option Pos :=
     let pos := s.prev pos
     if s.get pos == c then some pos
     else revPosOfAux s c pos
-termination_by _ => pos.1
+termination_by pos.1
 
 def revPosOf (s : String) (c : Char) : Option Pos :=
   revPosOfAux s c s.endPos
@@ -183,7 +183,7 @@ def findAux (s : String) (p : Char → Bool) (stopPos : Pos) (pos : Pos) : Pos :
       have := Nat.sub_lt_sub_left h (lt_next s pos)
       findAux s p stopPos (s.next pos)
   else pos
-termination_by _ => stopPos.1 - pos.1
+termination_by stopPos.1 - pos.1
 
 @[inline] def find (s : String) (p : Char → Bool) : Pos :=
   findAux s p s.endPos 0
@@ -195,7 +195,7 @@ def revFindAux (s : String) (p : Char → Bool) (pos : Pos) : Option Pos :=
     let pos := s.prev pos
     if p (s.get pos) then some pos
     else revFindAux s p pos
-termination_by _ => pos.1
+termination_by pos.1
 
 def revFind (s : String) (p : Char → Bool) : Option Pos :=
   revFindAux s p s.endPos
@@ -213,8 +213,8 @@ def firstDiffPos (a b : String) : Pos :=
         have := Nat.sub_lt_sub_left h (lt_next a i)
         loop (a.next i)
     else i
+    termination_by stopPos.1 - i.1
   loop 0
-termination_by loop => stopPos.1 - i.1
 
 @[extern "lean_string_utf8_extract"]
 def extract : (@& String) → (@& Pos) → (@& Pos) → String
@@ -240,7 +240,7 @@ where
       splitAux s p i' i' (s.extract b i :: r)
     else
       splitAux s p b (s.next i) r
-termination_by _ => s.endPos.1 - i.1
+termination_by s.endPos.1 - i.1
 
 @[specialize] def split (s : String) (p : Char → Bool) : List String :=
   splitAux s p 0 0 []
@@ -260,7 +260,7 @@ def splitOnAux (s sep : String) (b : Pos) (i : Pos) (j : Pos) (r : List String) 
         splitOnAux s sep b i j r
     else
       splitOnAux s sep b (s.next i) 0 r
-termination_by _ => s.endPos.1 - i.1
+termination_by s.endPos.1 - i.1
 
 def splitOn (s : String) (sep : String := " ") : List String :=
   if sep == "" then [s] else splitOnAux s sep 0 0 0 []
@@ -369,7 +369,7 @@ def offsetOfPosAux (s : String) (pos : Pos) (i : Pos) (offset : Nat) : Nat :=
   else
     have := Nat.sub_lt_sub_left (Nat.gt_of_not_le (mt decide_eq_true h)) (lt_next s _)
     offsetOfPosAux s pos (s.next i) (offset+1)
-termination_by _ => s.endPos.1 - i.1
+termination_by s.endPos.1 - i.1
 
 def offsetOfPos (s : String) (pos : Pos) : Nat :=
   offsetOfPosAux s pos 0 0
@@ -379,7 +379,7 @@ def offsetOfPos (s : String) (pos : Pos) : Nat :=
     have := Nat.sub_lt_sub_left h (lt_next s i)
     foldlAux f s stopPos (s.next i) (f a (s.get i))
   else a
-termination_by _ => stopPos.1 - i.1
+termination_by stopPos.1 - i.1
 
 @[inline] def foldl {α : Type u} (f : α → Char → α) (init : α) (s : String) : α :=
   foldlAux f s s.endPos 0 init
@@ -392,7 +392,7 @@ termination_by _ => stopPos.1 - i.1
     let a := f (s.get i) a
     foldrAux f a s i begPos
   else a
-termination_by _ => i.1
+termination_by i.1
 
 @[inline] def foldr {α : Type u} (f : Char → α → α) (init : α) (s : String) : α :=
   foldrAux f init s s.endPos 0
@@ -404,7 +404,7 @@ termination_by _ => i.1
       have := Nat.sub_lt_sub_left h (lt_next s i)
       anyAux s stopPos p (s.next i)
   else false
-termination_by _ => stopPos.1 - i.1
+termination_by stopPos.1 - i.1
 
 @[inline] def any (s : String) (p : Char → Bool) : Bool :=
   anyAux s s.endPos p 0
@@ -463,7 +463,7 @@ theorem mapAux_lemma (s : String) (i : Pos) (c : Char) (h : ¬s.atEnd i) :
     have := mapAux_lemma s i c h
     let s := s.set i c
     mapAux f (s.next i) s
-termination_by _ => s.endPos.1 - i.1
+termination_by s.endPos.1 - i.1
 
 @[inline] def map (f : Char → Char) (s : String) : String :=
   mapAux f 0 s
@@ -490,7 +490,7 @@ where
       have := Nat.sub_lt_sub_left h (Nat.add_lt_add_left (one_le_csize c₁) off1.1)
       c₁ == c₂ && loop (off1 + c₁) (off2 + c₂) stop1
     else true
-termination_by loop => stop1.1 - off1.1
+  termination_by stop1.1 - off1.1
 
 /-- Return true iff `p` is a prefix of `s` -/
 def isPrefixOf (p : String) (s : String) : Bool :=
@@ -512,8 +512,8 @@ def replace (s pattern replacement : String) : String :=
         else
           have := Nat.sub_lt_sub_left this (lt_next s pos)
           loop acc accStop (s.next pos)
+      termination_by s.endPos.1 - pos.1
     loop "" 0 0
-termination_by loop => s.endPos.1 - pos.1
 
 end String
 
@@ -612,8 +612,8 @@ def splitOn (s : Substring) (sep : String := " ") : List Substring :=
         else
           s.extract b i :: r
         r.reverse
+      termination_by s.bsize - i.1
     loop 0 0 0 []
-termination_by loop => s.bsize - i.1
 
 @[inline] def foldl {α : Type u} (f : α → Char → α) (init : α) (s : Substring) : α :=
   match s with
@@ -640,7 +640,7 @@ def contains (s : Substring) (c : Char) : Bool :=
       takeWhileAux s stopPos p (s.next i)
     else i
   else i
-termination_by _ => stopPos.1 - i.1
+termination_by stopPos.1 - i.1
 
 @[inline] def takeWhile : Substring → (Char → Bool) → Substring
   | ⟨s, b, e⟩, p =>
@@ -661,7 +661,7 @@ termination_by _ => stopPos.1 - i.1
     if !p c then i
     else takeRightWhileAux s begPos p i'
   else i
-termination_by _ => i.1
+termination_by i.1
 
 @[inline] def takeRightWhile : Substring → (Char → Bool) → Substring
   | ⟨s, b, e⟩, p =>
--- a/src/Init/Meta.lean
+++ b/src/Init/Meta.lean
@@ -1039,7 +1039,7 @@ where
       go (i+1) (args.push (quote xs[i]))
     else
       Syntax.mkCApp (Name.mkStr2 "Array" ("mkArray" ++ toString xs.size)) args
-termination_by go i _ => xs.size - i
+  termination_by xs.size - i
 
 instance [Quote α `term] : Quote (Array α) `term where
   quote := quoteArray
--- a/src/Lean/Compiler/LCNF/AlphaEqv.lean
+++ b/src/Lean/Compiler/LCNF/AlphaEqv.lean
@@ -75,10 +75,10 @@ def eqvLetValue (e₁ e₂ : LetValue) : EqvM Bool := do
           go (i+1)
       else
         x
+      termination_by params₁.size - i
     go 0
   else
     return false
-termination_by go i => params₁.size - i
 
 def sortAlts (alts : Array Alt) : Array Alt :=
   alts.qsort fun
--- a/src/Lean/Compiler/LCNF/Specialize.lean
+++ b/src/Lean/Compiler/LCNF/Specialize.lean
@@ -181,8 +181,8 @@ def expandCodeDecls (decls : Array CodeDecl) (body : LetValue) : CompilerM Expr 
       go (i+1) (subst.push value)
     else
       (body.toExpr.abstract xs).instantiateRev subst
+    termination_by values.size - i
   return go 0 #[]
-termination_by go => values.size - i
 
 /--
 Create the "key" that uniquely identifies a code specialization.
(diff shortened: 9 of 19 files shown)


tests/lean/termination_by_where.lean
```diff
@@ -1,86 +1,286 @@
 /-!
-This module tests various indentation variants related to termination_by.
+This module systematically tests various indentation variants related to `where` and
+`termination_by`/`decreasing_by`.
+
+The naming convention is `Ex<a><b><c><d><e>.foo` where
+
+* `a` is the column of `def`
+* `b` is the column of `where
```
semorrison:
```suggestion
* `b` is the column of `where`
```

src/Lean/Elab/PreDefinition/WF/PackMutual.lean
```diff
@@ -36,6 +36,47 @@ private def getCodomainsLevel (preDefsOriginal : Array PreDefinition) (preDefTyp
           throwError "invalid mutual definition, result types must be in the same universe level, resulting type for `{preDefsOriginal[0]!.declName}` is{indentExpr type₀} : {← inferType type₀}\nand for `{preDefsOriginal[i]!.declName}` is{indentExpr typeᵢ} : {← inferType typeᵢ}"
   return r
 
+def unpackMutualArg {m} [Monad m] [MonadError m] (numFuncs : Nat) (e : Expr) :
```
semorrison, line 39: Could new functions please have doc-strings? :-)

nomeata: Yes yes, this is still a draft PR :-)

src/Lean/Elab/DefView.lean
```diff
@@ -127,7 +127,7 @@ def mkDefViewOfOpaque (modifiers : Modifiers) (stx : Syntax) : CommandElabM DefV
       pure <| mkNode ``Parser.Command.declValSimple #[ mkAtomFrom stx ":=", val ]
   return {
     ref := stx, kind := DefKind.opaque, modifiers := modifiers,
-    declId := stx[1], binders := binders, type? := some type, value := val
+    declId := stx[1], binders := binders, type? := some type, value := val,
```
nomeata, line 130:
```suggestion
    declId := stx[1], binders := binders, type? := some type, value := val
```

nomeata:
This is getting ready, and it’s a good time for extra eyeballs.

Two concrete questions for review:

### Are the changes to the syntax sensible?

Here are the definitions in concise form:
```
namespace Termination
def terminationBy := leading_parser
  ppDedent ppLine >>
  "termination_by " >>
  optional (atomic (many (ppSpace >> (ident <|> "_")) >> " => ")) >>
  termParser
def decreasingBy := leading_parser
  ppDedent ppLine >> "decreasing_by " >> Tactic.tacticSeqIndentGt
def suffix := leading_parser
  optional terminationBy >> optional decreasingBy
end Termination
```
used in three places:
```diff
 def letRecDecl       := leading_parser
- optional Command.docComment >> optional «attributes» >> letDecl
+ optional Command.docComment >> optional «attributes» >> letDecl >> Termination.suffix

 @[run_builtin_parser_attribute_hooks]
 def matchAltsWhereDecls := leading_parser
- matchAlts >> optional whereDecls
+ matchAlts >> Termination.suffix >> optional whereDecls

 def declValSimple    := leading_parser
-  " :=" >> ppHardLineUnlessUngrouped >> termParser >> optional Term.whereDecls
+  " :=" >> ppHardLineUnlessUngrouped >> termParser >> Termination.suffix >> optional Term.whereDecls
```

### Expected to hit .missing and .null?

When parsing this syntax I am using this code:

```
def elabTerminationHints {m} [Monad m] [MonadError m] (stx : TSyntax ``suffix) : m TerminationHints := do
  if let .missing := stx.raw then
    return { TerminationHints.none with ref := stx }
  if stx.raw.matchesNull 0 then
    return { TerminationHints.none with ref := stx }
  match stx with
  | `(suffix| $[$t?:terminationBy]? $[$d?:decreasingBy]? ) => do
    let termination_by? ← t?.mapM fun t => match t with
      | `(terminationBy|termination_by $vars* => $body) =>
        if vars.isEmpty then
          throwErrorAt t "no extra parameters bounds, please omit the `=>`"
        else
          pure {ref := t, vars, body}
      | `(terminationBy|termination_by $body:term) => pure {ref := t, vars := #[], body}
      | _ => throwErrorAt t "unexpected `termination_by` syntax"
    let decreasing_by? ← d?.mapM fun d => match d with
      | `(decreasingBy|decreasing_by $tactic) => pure {ref := d, tactic}
      | _ => throwErrorAt d "unexpected `decreasing_by` syntax"
    return { ref := stx, termination_by?, decreasing_by?, extraParams := 0 }
  | _ => throwErrorAt stx s!"Unexpected Termination.suffix syntax: {stx} of kind {stx.raw.getKind}"
```

If I remove the check for `.missing` or null nodes it fails in some examples. Is that worrying and a sign that somewhere I am doing something wrong, or is that expected?

src/lake/Lake/DSL/DeclUtil.lean
```diff
@@ -86,15 +86,15 @@ def mkConfigDecl (name? : Option Name)
...
+  `($[$doc?]? @[$attrs,*] abbrev $(fixName id name?) : $ty := $defn $[$wds?:whereDecls]?)
 | `(structDeclSig| $id:ident $[: $ty?]? :=%$defTk $defn $[$wds?]?) => do
   let notice ← withRef defTk `(#eval IO.eprintln s!" warning: {__dir__}: `:=` syntax for configurations has been deprecated")
   `($notice $[$doc?]? @[$attrs,*] abbrev $(fixName id name?) : $ty := $defn $[$wds?]?)
-| `(structDeclSig| $id:ident { $[$fs $[,]?]* } $[$wds?]?) => do
+| `(structDeclSig| $id:ident { $[$fs $[,]?]* } $[$wds?:whereDecls]?) => do
   let fields ← fs.mapM expandDeclField
   let defn ← `({ name := $(quote id.getId), $fields,* })
-  `($[$doc?]? @[$attrs,*] abbrev $(fixName id name?) : $ty := $defn $[$wds?]?)
+  `(command|$[$doc?]? @[$attrs,*] abbrev $(fixName id name?) : $ty := $defn $[$wds?:whereDecls]?)
```
tydeu, line 99: Minor nitpick: You added `command` to the quotation on line 99 but not on 92. I imagine this was adding during some debugging. For consistency's sake, it should problem be removed (as the other quotations do not make `command` explicit).

nomeata: Good catch, and correct analysis, thanks!

nomeata, line 99:
```suggestion
  `($[$doc?]? @[$attrs,*] abbrev $(fixName id name?) : $ty := $defn $[$wds?:whereDecls]?)
```

tydeu (approved):

Kha:
The grammar changes mostly look sensible. Do we need indentation awareness to separate e1 and e2 in the following case?
```
let rec f := ...
termination_by e1
e2
```

> If I remove the check for .missing or null nodes it fails in some examples. Is that worrying

`missing` is produced by the parser iff there was a parse error, in which case we skip reporting most elaboration errors anyway, so one would think that the default branch in the `match` would be sufficient. And on well-formed syntax, the first pattern should subsume the `matchesNull`, yes. So might be worth looking into, e.g. printing the full syntax in the `matchesNull` case. Do you think this syntax tree comes straight from the parser or some transformation in between?

nomeata:
Re indentation awareness: Good question, I’ll write up some test cases later.

Re .missing and null: I do get test failures without both of these clauses ([run 1](https://github.com/leanprover/lean4/actions/runs/7286114730/job/19854239027), [run 2](https://github.com/leanprover/lean4/actions/runs/7286834964/job/19856313907)):
```
343.lean:18:0-19:54: error: Unexpected Termination.suffix syntax: [] of kind null
1235.lean:1:0-1:26: error: Unexpected Termination.suffix syntax: <missing> of kind missing
```

I’ll dig deeper and try to understand what’s going on.

nomeata:
> missing is produced by the parser iff there was a parse error, in which case we skip reporting most elaboration errors anyway, so one would think that the default branch in the match would be sufficient. 

Found (one of it). The code `opaque f (a b : Nat) : Nat` is transformed by this code

https://github.com/leanprover/lean4/blob/d1a15dea03f935457c6b3cbe99af029652a2584f/src/Lean/Elab/DefView.lean#L120-L131

This manually constructed syntax of course has `.missing` when you try to get the termination suffix (`stx[2]` later) or the where-clauses. Manually constructing the expected syntax, i.e.
```
(Command.declValSimple
 ":="
 …
 (Termination.suffix [] [])
 [])
```
is probably tedious.

How important is the `mkAtomFrom` around `:=`? If not, this would probably work more reliable:
```diff
--- a/src/Lean/Elab/DefView.lean
+++ b/src/Lean/Elab/DefView.lean
@@ -124,7 +124,7 @@ def mkDefViewOfOpaque (modifiers : Modifiers) (stx : Syntax) : CommandElabM DefV
     | some val => pure val
     | none     =>
       let val ← if modifiers.isUnsafe then `(default_or_ofNonempty% unsafe) else `(default_or_ofNonempty%)
-      pure <| mkNode ``Parser.Command.declValSimple #[ mkAtomFrom stx ":=", val ]
+      `(Parser.Command.declValSimple| := $val)
   return {
     ref := stx, kind := DefKind.opaque, modifiers := modifiers,
     declId := stx[1], binders := binders, type? := some type, value := val
```

Or is there a strong reason to avoid syntax quotations in core code (bootstrapping issues maybe)?

nomeata:
With 37ab645, most tests work again, but four still fail:

<details>

```
Test project /home/jojo/build/lean/lean4/build/release/stage1
    Start  270: leantest_doLetLoop.lean
    Start  319: leantest_have.lean
    Start 1718: leaninteractivetest_completion5.lean
    Start 1719: leaninteractivetest_completion6.lean
1/4 Test  #270: leantest_doLetLoop.lean ................***Failed    0.06 sec
--- doLetLoop.lean.expected.out	2023-12-02 12:17:16.269185699 +0100
+++ doLetLoop.lean.produced.out	2023-12-22 14:41:17.452879698 +0100
@@ -1,2 +1,2 @@
 doLetLoop.lean:4:0: error: unexpected end of input
-doLetLoop.lean:2:4-2:5: warning: declaration uses 'sorry'
+doLetLoop.lean:2:0-3:8: error: Unexpected Termination.suffix syntax: <missing> of kind missing
ERROR: file doLetLoop.lean.produced.out does not match doLetLoop.lean.expected.out

2/4 Test  #319: leantest_have.lean .....................***Failed    0.06 sec
--- have.lean.expected.out	2023-12-02 12:17:15.938183664 +0100
+++ have.lean.produced.out	2023-12-22 14:41:17.455879690 +0100
@@ -1,7 +1,4 @@
 have.lean:2:19-4:7: error: unexpected token 'example'; expected term
-have.lean:2:18-2:19: error: don't know how to synthesize placeholder
-context:
-⊢ False
 have.lean:7:2-7:3: error: type mismatch
   f
 has type
ERROR: file have.lean.produced.out does not match have.lean.expected.out

3/4 Test #1718: leaninteractivetest_completion5.lean ...***Failed    1.91 sec
--- completion5.lean.expected.out	2023-12-12 11:26:11.359674934 +0100
+++ completion5.lean.produced.out	2023-12-22 14:41:19.301875008 +0100
@@ -1,7 +1,3 @@
 {"textDocument": {"uri": "file:///completion5.lean"},
  "position": {"line": 9, "character": 15}}
-{"items":
- [{"label": "b1", "kind": 5, "detail": "C → String"},
-  {"label": "f1", "kind": 5, "detail": "C → Nat"},
-  {"label": "f2", "kind": 5, "detail": "C → Bool"}],
- "isIncomplete": true}
+{"items": [], "isIncomplete": true}
ERROR: file completion5.lean.produced.out does not match completion5.lean.expected.out

4/4 Test #1719: leaninteractivetest_completion6.lean ...***Failed    1.94 sec
--- completion6.lean.expected.out	2023-12-12 11:26:11.359674934 +0100
+++ completion6.lean.produced.out	2023-12-22 14:41:19.331874935 +0100
@@ -1,12 +1,6 @@
 {"textDocument": {"uri": "file:///completion6.lean"},
  "position": {"line": 12, "character": 15}}
-{"items":
- [{"label": "b1", "kind": 5, "detail": "C → String"},
-  {"label": "f1", "kind": 5, "detail": "C → Nat"},
-  {"label": "f2", "kind": 5, "detail": "C → Bool"},
-  {"label": "f3", "kind": 5, "detail": "D → Bool"},
-  {"label": "toC", "kind": 5, "detail": "D → C"}],
- "isIncomplete": true}
+{"items": [], "isIncomplete": true}
 {"textDocument": {"uri": "file:///completion6.lean"},
  "position": {"line": 21, "character": 4}}
 {"items":
ERROR: file completion6.lean.produced.out does not match completion6.lean.expected.out
```
</details>

These all relate to partially parsed terms, for example `doLetLoop.lean` is
```
set_option showPartialSyntaxErrors true
def f : IO Unit := do
  if let
```
and now `elabTerminationHints` complains. Probably too loudly given that this is a partial syntax tree anyways?

How should I proceed here? Make `elabTerminationHints` more liberal again and accept `.missing`? Or is there a more principled way?

nomeata:
For the cases where I hit
```
error: Unexpected Termination.suffix syntax: [] of kind null
```
I have the following theory:

Affected are uses of the `unif_hint` command, which is defined in `Init.NotationExtras`:

https://github.com/leanprover/lean4/blob/d1a15dea03f935457c6b3cbe99af029652a2584f/src/Init/NotationExtra.lean#L70-L77

When compiling this code, we are still using stage0 with the old syntax (despite `parseQuotWithCurrentStage := true`), and the old syntax has the `optional whereDelcs` in the splot, and that’s the `[]` we see.

The other failures are related to lakefiles parsing, so probably similar.

So for this one the way forward would probably to keep `elabTerminationHints` liberal at first, merge the PR, do a stage0 update, and then make it stricter?

Kha:
> When compiling this code, we are still using stage0 with the old syntax (despite `parseQuotWithCurrentStage := true`)

Right, the flag only affects code where the changed parser was imported. Your suggested approach makes sense.

> How should I proceed here? Make elabTerminationHints more liberal again and accept .missing? Or is there a more principled way?

The test output suggests that the exception is preventing us from ever elaborating the body, in that case we need to be liberal and just accept .missing here, yes. The parse error makes sure that in the end we will consider this command as failed in any case.

nomeata:
> The grammar changes mostly look sensible. Do we need indentation awareness to separate e1 and e2 in the following case?
> 
> ```
> let rec f := ...
> termination_by e1
> e2
> ```

It seems to work: The `termination_by` only picks up more expressions if they are indented with regard to the `let rec`; same as the body of `f` it seems.

david-christiansen:
Here's some questions that arise for me when reading RELEASES.md.

> The `termination_by` clause must bind exactly those parameters that are not
    already bound by the function header. If there are none, the `=>` can be
    omitted

Just to double-check - this means that
```
def foo (x : Nat) : Nat -> Nat := ...
termination_by x - y
```
is not allowed, right? Even if the RHS of the definition is constructing a function by well-founded recursion on `x`?

All the examples omit `=>` when they may do so. Why is it not mandatory to do so when no names are bound? Lean also doesn't accept `fun => 5` as a natural number, after all.

The examples in the migration guide don't seem to include any `let rec` cases - this will be useful to have. Examples will also help seed the community with a default code style WRT things like indentation, as we haven't had `termination_by` in the middle of definitions before.

Finally, what are the indentation rules for the new `termination_by` and `decreasing_by`? There's some hints in this conversation here, but users shouldn't have to read our PRs to discover this kind of thing. A written specification for the resulting syntax is important - even if it doesn't all belong in the release notes, it should be available somewhere in the `doc` directory.

nomeata:
> If there are none, the `=>` can be
> omitted

That's just wrong, should say “must be”. Thanks for catching!

Will add more examples as well.

I have a PR that mentions these termination hints in the docs in https://github.com/leanprover/lean4/pull/3016, not yet updated to the new rules. I'll update that and merge after I merged this.

nomeata:
I improved the release notes, thanks for the suggestions. It didn't feel quite right to talk much about indentation there: In the final version, there isn’t much to worry about indentation, because we now place `termination_by` before the `where` and the ambiguities (and indentation-based resolutions) that were discussed in this PR are obsolete.

david-christiansen:
My indentation worries are primarily about when the `termination_by` or `decreasing_by` occur in an indented context. My understanding is that there's no rules governing their indentation, but we hope that they'll typically be indented at the same level as either the name in the `where` block that they apply to, or the level of the `let` keyword that defines the function in question. Is that right? If so, I think it's worth mentioning a sentence like this, because users will be faced with the task of migrating their code and needing to put the keywords somewhere.

RELEASES.md
```diff
@@ -22,9 +22,12 @@ v4.5.0 (development in progress)
...
 * The syntax of the `termination_by` and `decreasing_by` termination hints is overhauled:
 
-  * They are now placed directly after the function they apply to, instead of after the whole `mutual` block.
-  * Therefore, the function name no longer has to be repeated.
-  * If the function has a `where`, the `termination_by` and `decreasing_by` clauses come before.
+  * They are now placed directly after the function they apply to, instead of
+    after the whole `mutual` block.
+  * Therefore, the function name no longer has to be mentioned in the hint.
+  * If the function has a `where` clause, the `termination_by` and
+    `decreasing_by` for that function clauses come before the `where`. The
```
david-christiansen, line 29:
```suggestion
    `decreasing_by` for that function come before the `where`. The
```

RELEASES.md
```diff
@@ -22,9 +22,12 @@ v4.5.0 (development in progress)
...
+    functions in the `where` clause can have their own termination hints.
```
david-christiansen, line 30:
```suggestion
    functions in the `where` clause can have their own termination hints, each following the corresponding definition.
```

RELEASES.md
```diff
@@ -68,18 +71,31 @@ v4.5.0 (development in progress)
...
+  -  let rec bar (x : Nat) :  Nat := …
+  -  …
+  -termination_by
+  -  foo a b => a - b
+  -  bar x => x
+  +def foo (a b : Nat) : Nat :=
+  +  let rec bar (x : Nat) : Nat := …
+  +    termination_by x
+  +  …
+  +  termination_by a - b
```
david-christiansen, line 98: We typically indent these at the level of the `def`:
```suggestion
  +termination_by a - b
```

RELEASES.md
```diff
@@ -68,18 +71,31 @@ v4.5.0 (development in progress)
...
+  -def foo (a b : Nat) : Nat := …
+  -  where bar (x : Nat) : Nat := …
   -termination_by
   -  foo a b => a - b
-  -  bar a => a
-  +def foo : Nat → Nat → Nat := …
-  +  termination_by a b => a - b
+  -  bar x => x
+  +def foo (a b : Nat) : Nat := …
+  +  termination_by a - b
```
david-christiansen, line 83: We typically indent these at the level of the `def`:
```suggestion
  +termination_by a - b
```

david-christiansen (changes requested):
A few nits to pick :-)

david-christiansen:
Thanks for the example of the `let rec`, that really helps!

RELEASES.md
```diff
@@ -20,6 +20,121 @@ v4.5.0 (development in progress)
...
+
+  * They are now placed directly after the function they apply to, instead of
+    after the whole `mutual` block.
+  * Therefore, the function name no longer has to be mentioned in the hint.
+  * If the function has a `where` clause, the `termination_by` and
+    `decreasing_by` for that function clauses come before the `where`. The
+    functions in the `where` clause can have their own termination hints.
+  * The `termination_by` clause must bind exactly those parameters that are not
+    already bound by the function header. If there are none, the `=>` can be
+    omitted.
```
david-christiansen, line 33: Does this mean that
```
def f (x : Nat) : Nat -> Nat :=
  if h : x < 4 then id else fun n => 1 + f (x - 10) n
termination_by x
```
is forbidden?

I asked this in the other comment but I think it got lost in the shuffle. If it's allowed, then perhaps the text could be:
```suggestion
  * The `termination_by` clause may bind parameters that are not
    already bound by the function header and are used in the termination
    argument. If there are none, the `=>` must be omitted.
```

nomeata:
> My indentation worries are primarily about when the `termination_by` or `decreasing_by` occur in an indented context. My understanding is that there's no rules governing their indentation, but we hope that they'll typically be indented at the same level as either the name in the `where` block that they apply to, or the level of the `let` keyword that defines the function in question. Is that right? If so, I think it's worth mentioning a sentence like this, because users will be faced with the task of migrating their code and needing to put the keywords somewhere.

They can indent it as they like; the migration guide isn’t really the place for giving style advice (besides implicitly in example, if at all), isn’t it?

Maybe that stance is heavily influence by how I approach indentation, namely intuitively. I can’t explain the rules around indentation in Haskell (let alone Lean), but I hardly ever have trouble with writing correctly indented code.

And personally I also don’t care if people write
```
foo
  where bar (x : Nat) := …
    termination_by x
```
or
```
foo
  where bar (x : Nat) := …
  termination_by x
```
or
```
foo
  where
    bar (x : Nat) := …
    termination_by x
```
or whatever.

RELEASES.md
nomeata: It 's forbidden, correct.


```suggestion
  * The `termination_by` clause must bind exactly those parameters that are not
    already bound by the function header. If there are none, the `=>` must be
    omitted.
```

digama0: I think that example should not be forbidden, since it doesn't introduce the variable. Banning it would mean that specializing an opaque type can cause it to fail:
```lean
opaque A : Type := Nat → Nat
def f (x : Nat) (a : A) : A :=
  if h : x = 0 then a else f (x - 1) a
termination_by x -- works

def B : Type := Nat → Nat
def f (x : Nat) (a : B) : B :=
  if h : x = 0 then a else f (x - 1) a
termination_by x -- fails
```

nomeata: It’s not the type that matters, but rather the function definition (number of manifest lambdas after the `:=`, or scrutinees in the cases syntax. I should clarify that.

The rationale is that that the implementation has to figure out _some_ arity of the function, and arity affects, for example, how arguments are tupled. It seems better to require the user to explicitly acknowledge the arity as inferred and used by the system. It’s a little bit of extra red tape for the user, but it takes away a pointless choice (so less for the user to decide, which is good) and gives more insight in the internals.

I would support allowing `termination_by` with less paramters if that was _prescriptive_, i.e. really affected the argument packing as well, if there was a compelling use case for that, but that’d be a separate change.

digama0: > gives more insight in the internals

I would call that "coupling to internal details" and I don't think it's a good thing. I don't see any reason for this either. If the user only introduces 2 arguments in the `termination_by` clause, then you should only consider those two arguments while reasoning about termination, you shouldn't go ahead and consider 3 arguments and then complain at the user for having done so.

nomeata: I agree that the implementation (existing already) leaks heuristic implementation detail, but fixing that isn't the point of this PR, and I don't want to inflate it more.

And given that the implementation has leaky heuristics, let's be honest about it and reserve “binding fewer parameters” as the syntax for a possible future where this _does_ guide/replace the heuristic.

You are right about ”manifest” being the wrong word.

digama0: By the way, possibly related, I think it is weird that `y` is in scope in the second `termination_by` hint. It is outside the scope of the lambda, so I should not be allowed to write `termination_by _ => y` there, but the current implementation allows me to (only in the second case). So maybe that gives an alternative answer for your arity problem: only stuff left of the colon is in scope, stuff in the body must be introduced explicitly using the `=>` syntax, no matter whether there are manifest lambdas or match statements or anything like that.

nomeata: > The existing implementation does not require any particular number of arguments to be used. 

Well, it syntactically allows you to omit some. But this is very superficial, and does not affect heuristic about how many parameters are actually put in the packed argument in the wf transformation.

The two functions above are translated differently; the first into a recursion on `Nat` and the second into a recursion on `Nat × Nat`.

I agree that this heuristic is maybe suboptimal and this will become more relevant once we infer induction principles from function definitions, which will look quite different depending on the number of paramteres in the recursion. So giving the user some control over this is desirable!

A good syntax for “how many parameters should be packed” is, I’d say, “the parameters given in `termination_by`”. The implementation in this PR is nicely forward-compatible with that interpretation, by only allowing the syntax that matches the (currently implementation defined) heuristic.

The old behavior accepts `termination_by f' => x` and `termination_by f' x n => x`, but there is no actual difference in meaning between these two syntaxes, so to me it’s preferrable to allow only one. Less for the style guide writers to discuss.

david-christiansen: I have strong reservations about a system for which the interface is "guess what the compiler's internals do, and if you don't guess right, then too bad on you". I guess it's fine as a stopgap, but then we should acknowledge this.

In my opinion, if there's an internal notion of function arity that user's can't directly specify, then the interface shouldn't require users to figure out what it is and stick to it. This is fragile, in that changes to elaboration that affect function arity become breaking changes, and it's obscure, in that the core language elaboration leaks into the UI.

Why can't the interface be that users can specify as many of the further arguments as they need in order to specify the termination measure? I don't get why "the internals might be different" is an argument for the current state of things.

david-christiansen:
I think that many users may assume that e.g. the RHS of `termination_by` should be at least as indented as the token, and if this isn't required, it's a good idea to say something like "there are no additional indentation rules" and then show a reasonable default style in the examples. But I don't think this very strongly, the lack of rules may be reason enough to omit the statement :)

RELEASES.md
```diff
@@ -29,7 +29,8 @@ v4.5.0 (development in progress)
     `decreasing_by` for that function come before the `where`. The
     functions in the `where` clause can have their own termination hints, each following the corresponding definition.
   * The `termination_by` clause must bind exactly those parameters that are not
-    already bound by the function header. If there are none, the `=>` must be
+    already bound by the function header, but by a lambda (e.g. `:= fun x y z =>`)
+    or by guards (e.g. `| x, n + 1 => …`). If there are none, the `=>` must be
```
david-christiansen, line 33: Do we call these "guards" in Lean? I thought they were cases. In my mind, a guard is a separate Boolean condition that governs a pattern match case, as e.g. Haskell has.

RELEASES.md
```diff
@@ -29,7 +29,8 @@ v4.5.0 (development in progress)
     `decreasing_by` for that function come before the `where`. The
     functions in the `where` clause can have their own termination hints, each following the corresponding definition.
   * The `termination_by` clause must bind exactly those parameters that are not
-    already bound by the function header. If there are none, the `=>` must be
+    already bound by the function header, but by a lambda (e.g. `:= fun x y z =>`)
```
david-christiansen, line 32: What if the lambda is under an `if`?

RELEASES.md
david-christiansen: (basically, I agree with @digama0's suggestion of "only stuff left of the colon is in scope, stuff in the body must be introduced explicitly using the => syntax, no matter whether there are manifest lambdas or match statements or anything like that")

RELEASES.md
nomeata: Then it doesn’t count. The behavior is unfortunately rather implementation-defined at the moment, and the real specification is “whatever `lambdaTelescope` does after elaboration”, so I don’t have high hopes for a more helpful specification here than a rough intuition.

RELEASES.md
nomeata: Ah, yes, good point. But what are these things called? The parser calls them `matchAlts`, but there is no `alts` around. In https://github.com/leanprover/lean4/blob/master/doc/declarations.md#the-equation-compiler they are documented, but not actually named. I guess “definition by patterns” at least isn’t wrong.

```suggestion
    or matched by patterns (e.g. `| x, n + 1 => …`). If there are none, the `=>` must be
```

digama0: `matchAlts` stands for match alternatives, and they are also called match arms. "Definition by cases" also seems like a reasonable description of the syntax.

RELEASES.md
digama0: > The two functions above are translated differently; the first into a recursion on `Nat` and the second into a recursion on `Nat × Nat`.

By the way, I think this is suboptimal from a proof perspective. The second argument should not participate in the recursion at all, this is a definition by recursion on the first argument (a recursion on `Nat`) which has a motive of type `Nat -> Nat`. Some day I'll go figure out why it's generating a bad proof, if you don't beat me to it.

nomeata: > Why can't the interface be that users can specify as many of the further arguments as they need in order to specify the termination measure? I don't get why "the internals might be different" is an argument for the current state of things.

The current syntax on `master` is rather full of surprises. Consier:
```
def foo (n : Nat) : Nat → Nat → Nat
 | a, b => …
```
which parameter do you think is by each of these:
```
termination_by foo => n        -- 1
termination_by foo n => n      -- 2
termination_by foo _ n => n    -- 3
termination_by foo n _ => n    -- 4
termination_by foo _ _ n => n  -- 5
termination_by foo m => n + m  -- 6
```
All of them are valid! But this is very confusing, especially for the reader of the code.

The confusion is reduced if you _know_ that before the `=>` are exactly and only those obscure “extra parameters” inferred by the system. With the current PR, these 5 arguments are written as
```
termination_by _ _ => n
termination_by _ n => n
termination_by _ n => n
termination_by n _ => n
termination_by _ n => n
termination_by _ m => n + m
```

> “only stuff left of the colon is in scope”

That's the case. Old behavior was that you can rename stuff on the left of the colon (4), new behavior is that you cannot rename stuff on the left (because why should you besides to confuse the reader).

> “stuff in the body must be introduced explicitly using the => syntax”

agreed, and that’s the case.

> “no matter whether there are manifest lambdas or match statements or anything like that")”

That’s not possible at the moment: The system has its current rules for how many parameters there are, and this is not just an implementation detail. It certainly gives an _upper_ limit on the number of parameters here; it may show up in proofs; and will certainly show up once we derive induction principles from function definitions. So I don’t think it’s necessary to hide this detail.

As I said above: We can discuss changing this implementation, and I am sympathetic (up to prioritization of features), but maybe in a separate RFC. I’d say changing that is out of scope for this PR, but at least this PR’s syntax and semantics is a subset of that, so adding these features later will be a strict extension. That’s a good thing, isn’t it?

digama0: It seems unnecessarily obfuscatory to be using the same variable `n` there when there is already an `n` in scope. In the "left of the colon" proposal, those examples would be expressible as
```lean
-- 1
termination_by n
termination_by _ => n
termination_by _ _ => n
-- 2, 3, 5
termination_by _ b => b
-- 4
termination_by a => a
termination_by _ a => a
-- 6
termination_by _ m => n + m
```

> > “no matter whether there are manifest lambdas or match statements or anything like that")”
> 
> That’s not possible at the moment: The system has its current rules for how many parameters there are, and this is not just an implementation detail. It certainly gives an _upper_ limit on the number of parameters here; it may show up in proofs; and will certainly show up once we derive induction principles from function definitions. So I don’t think it’s necessary to hide this detail.

It should ideally not show up in proofs. In fact, if I recall correctly that was a major motivation for removing `termination_by'`, it was said to be too coupled to the implementation because you could see the tupled type explicitly, and this is true. We should hide this detail, because I fully expect it to change when we get better about not adding unnecessary things to the tuple. Maybe we can't do this perfectly, but at least we should avoid the common case having to couple to the implementation detail, and requiring that all arguments in the lambdaTelescope be introduced causes exactly that coupling.

nomeata: Yes, I was sloppy in the example and omitted the `_`.


Ok, implemented in 4a5b39a442997d836e6854f9fc5fc3e577216ecb, and updated the RELEASE notes.

david-christiansen (approved):
I think the migration guide is much improved now, and that allowing fewer argument names to be specified while treating them as a prefix is going to be easier to figure out than both the current behavior and the prior behavior of this PR. As a whole, this new system is easier to explain and to read - great stuff!

nomeata:
Rebased this PR (against all my strongly held convictions) because it needs a stage0 update in the middle. Needs to be rebase-merged. This seems to be good to go, but I’ll wait for mathlib CI to work again, to make sure that the std and mathlib PRs are up-to-date, and check back with Scott about how to merge this with the least release complications.

src/Lean/Elab/PreDefinition/WF/Fix.lean
```diff
@@ -186,22 +187,47 @@ def assignSubsumed (mvars : Array MVarId) : MetaM (Array MVarId) :=
...
+/--
+The subgoals are of the form `rel arg param`, where `param` is the `PackMutual`'ed parameter
+of the current function, and thus we can peek at that to know which function is making the call.
+The close coupling with how arguments are packed and termination goals look like is not great,
+but it works for now.
+-/
+def groupGoalsByFunction (numFuncs : Nat) (goals : Array MVarId) : MetaM (Array (Array MVarId)) := do
+  let mut r := mkArray numFuncs #[]
+  for goal in goals do
+    let (.mdata _ (.app _ param)) ← goal.getType
```
leodemoura, line 199: Could you please add a comment here?

src/Lean/Elab/PreDefinition/WF/GuessLex.lean
```diff
@@ -339,20 +354,18 @@ def evalRecCall (decrTactic? : Option Syntax) (rcc : RecCallWithContext) (paramI
...
+            let remainingGoals ← Tactic.run mvarId do Tactic.withoutRecover do
+              let tacticStx : Syntax ←
+                match decrTactic? with
+                | none => pure (← `(tactic| decreasing_tactic)).raw
+                | some decrTactic =>
+                  trace[Elab.definition.wf] "Using tactic {decrTactic.tactic.raw}"
+                  pure decrTactic.tactic.raw
+              Tactic.evalTactic tacticStx
+            remainingGoals.forM fun _ => throwError "goal not solved"
+            -- trace[Elab.definition.wf] "Found {rel} proof: {← instantiateMVars mvar}"
```
leodemoura, line 367: Leftover?

src/Lean/Elab/PreDefinition/WF/GuessLex.lean
```diff
@@ -538,25 +553,40 @@ def mkTupleSyntax : Array Term → MetaM Term
...
-          let sizeOfIdent := mkIdent (← unresolveNameGlobal ``sizeOf)
+          let varIdx := varIdxs[funIdx]!
+          let v := vars[varIdx]!
+          -- Print `sizeOf` as such, unless it is shadowed.
+          -- Shadowing by a `def` in the current namespace is handled
+          -- by `unresolveNameGlobal`.
+          -- But it could also be shadowed by an earlier parameter (fixed or not),
+          -- which we check in `originalVarNames`
+          let sizeOfIdent :=
+            if originalVarNamess[funIdx]!.any (· = `sizeOf) then
```
leodemoura, line 572: Missing ` at sizeOf?

src/Lean/Elab/PreDefinition/WF/GuessLex.lean
```diff
@@ -538,25 +553,40 @@ def mkTupleSyntax : Array Term → MetaM Term
...
+    (varNamess : Array (Array Name)) (measures : Array MutualMeasure) (trimVars : Bool)
+    : MetaM TerminationWF := do
   let mut termByElements := #[]
   for h : funIdx in [:varNamess.size] do
     let vars := (varNamess[funIdx]'h.2).map mkIdent
     let body ← mkTupleSyntax (← measures.mapM fun
       | .args varIdxs => do
-          let v := vars.get! (varIdxs[funIdx]!)
-          let sizeOfIdent := mkIdent (← unresolveNameGlobal ``sizeOf)
+          let varIdx := varIdxs[funIdx]!
```
leodemoura, line 564: Could you please document the relationship between the different parameters in the `buildTermWF` doc-string?

leodemoura (approved):

src/Lean/Elab/PreDefinition/WF/GuessLex.lean
nomeata: You rightfully called that function out, didn’t like it either. It’s now two functions, and better commented.

src/Lean/Elab/PreDefinition/WF/GuessLex.lean
```diff
@@ -72,27 +72,42 @@ register_builtin_option showInferredTerminationBy : Bool := {
...
-Given a predefinition, find good variable names for its parameters.
-Use user-given parameter names if present; use x1...xn otherwise.
+Given a predefinition, return the variabe names in the outermost lambdas.
+Includes the “fixed prefix”.
 
 The length of the returned array is also used to determine the arity
 of the function, so it should match what `packDomain` does.
+-/
+def originalVarNames (preDef : PreDefinition) : MetaM (Array Name) := do
+  lambdaTelescope preDef.value fun xs _ => xs.mapM (·.fvarId!.getUserName)
```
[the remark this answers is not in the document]
nomeata: Yes! It’s not high priority (only shows up in error/info messages), so I’ll open an issue and do it some other time. Maybe it’s even possible to do it earlier and have more helpful names in the actual predefinition. Ideally in `expandMatchAltsWhereDecls`, but that doesn’t have type information available.

PR #3042, opened 2023-12-08
chore: stage0 autoupdater action

nomeata:
This Github action automatically updates `stage0` on `master` if
`src/stdlib_flags.h` and `stage0/src/stdlib_flags.h`
are out of sync there.

It bypasses the merge queue to be quick, this way, an out-of-date stage0 on on
master should only exist for a few minutes.

Needs access to a _deploy SSH key_ with write permission.

### DIFF
--- a/garnix.yaml
+++ b/garnix.yaml
@@ -1,6 +1,0 @@
-builds:
-  include:
-  - packages.x86_64-linux.cacheRoots
-  - packages.x86_64-linux.update-stage0-commit
-  - checks.x86_64-linux.lean
-  - devShell.x86_64-linux


Kha:
@nomeata Deploy key deployed. Can we get a `workflow_dispatch` as well in order to trigger updates manually?

Kha (approved):

nomeata:
> @nomeata Deploy key deployed. Can we get a `workflow_dispatch` as well in order to trigger updates manually?

Yes, should be relatively straight-forward. Separate PR, I guess.

Would you want it to update even if `stdflags.h` are the same, or still do the same check?

Kha:
> Would you want it to update even if `stdflags.h` are the same, or still do the same check?

No check, we want to do this to e.g. make new LS features available in stage0.

nomeata:
> > @nomeata Deploy key deployed. Can we get a `workflow_dispatch` as well in order to trigger updates manually?
> 
> Yes, should be relatively straight-forward. Separate PR, I guess.

https://github.com/leanprover/lean4/pull/3052

PR #3043, opened 2023-12-09
fix: Option.getD eagerly evaluates dflt

digama0:
Reported [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/Panics.20in.20Std.2EHashMap.2Efind!/near/406872395). The `dflt` argument of `Option.getD` is not evaluated lazily, as the documentation says, because even after `macro_inline` the expression
```lean
match opt, dflt with
| some x, _ => x
| none, e => e
```
still has the semantics of evaluating `dflt` when `opt` is `some x`.


Kha:
!bench

tests/lean/run/optionGetD.lean
```diff
@@ -0,0 +1,12 @@
+import Lean.Data.HashMap
+
+def test (m : Lean.HashMap Nat Nat) : IO (Nat × Nat) := do
+  let start := 1
+  let mut i := start
+  let mut count := 0
+  while i != 0 do
+    i := (m.find? i).getD (panic! "key is not in the map")
```
Kha, line 8: Would this test actually fail on panic? It's not an output test and we're not running them under `LEAN_ABORT_ON_PANIC`

digama0: Really? I would have expected panics to be treated as failures in `run/` tests

digama0: I moved it to `tests/lean` and confirmed that the output contains `PANIC` on master.

Kha: > I would have expected panics to be treated as failures in `run/` tests

Perhaps they should be but right now we only set the flag for the build itself

PR #3046, opened 2023-12-09
chore: fix (PersistentHashMap.insert k v).size

### LINKED ISSUE #3029: PersistentHashMap.insert always increments size.
### Prerequisites

### Description

`PersistentHashMap.insert` increases the recorded size by one even if replacing an existing entry.  The following code returns a size of `2` even though a single entry exists.

```lean
import Lean.Data.PersistentHashMap
open Lean

#eval (PersistentHashMap.empty : PersistentHashMap Nat Nat)
    |> (·.insert 1 1)
    |> (·.insert 1 1)
    |> (·.size)
```
### Context

None

### Steps to Reproduce

See above

### Versions

`4.3.0-rc2`

### Impact

kim-em:
Fixes #3029, in which `(PersistentHashMap.insert k v).size` is wrong if the key already exists.

I think this fix ensures linearity correctly, but I would welcome learning if I've made a mistake!

### DIFF
--- a/src/Lean/Data/PersistentHashMap.lean
+++ b/src/Lean/Data/PersistentHashMap.lean
@@ -76,26 +76,21 @@ private theorem size_set {ks : Array α} {vs : Array β} (h : ks.size = vs.size)
 private theorem size_push {ks : Array α} {vs : Array β} (h : ks.size = vs.size) (k : α) (v : β) : (ks.push k).size = (vs.push v).size := by
   simp [h]
 
-structure WithReplaced (α) where
-  node : α
-  replaced : Bool
-  deriving Inhabited
-
-partial def insertAtCollisionNodeAux [BEq α] : CollisionNode α β → Nat → α → β → WithReplaced (CollisionNode α β)
+partial def insertAtCollisionNodeAux [BEq α] : CollisionNode α β → Nat → α → β → CollisionNode α β
   | n@⟨Node.collision keys vals heq, _⟩, i, k, v =>
     if h : i < keys.size then
       let idx : Fin keys.size := ⟨i, h⟩;
       let k' := keys.get idx;
       if k == k' then
          let j : Fin vals.size := ⟨i, by rw [←heq]; assumption⟩
-         ⟨⟨Node.collision (keys.set idx k) (vals.set j v) (size_set heq idx j k v), IsCollisionNode.mk _ _ _⟩, true⟩
+         ⟨Node.collision (keys.set idx k) (vals.set j v) (size_set heq idx j k v), IsCollisionNode.mk _ _ _⟩
       else insertAtCollisionNodeAux n (i+1) k v
     else
-      ⟨⟨Node.collision (keys.push k) (vals.push v) (size_push heq k v), IsCollisionNode.mk _ _ _⟩, false⟩
+      ⟨Node.collision (keys.push k) (vals.push v) (size_push heq k v), IsCollisionNode.mk _ _ _⟩
   | ⟨Node.entries _, h⟩, _, _, _ => nomatch h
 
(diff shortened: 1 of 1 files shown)


leodemoura:
!bench

Kha:
There were some configuration changes on the speedcenter runner machine (new Spectre mitigations...?), please try benchmarking again after rebasing

semorrison:
!bench

semorrison:
@Kha, I've rebased, but it looks like the benchmark report is pretty much the same.

I don't know how significant the regressions here are.

If someone could look over these changes to try to identify a cause (nonlinearity?) that would be great.

digama0:
I think the cause is most likely not linearity but the additional allocation involved in creating and using tuples. You may be able to fix this by using continuation passing style but I'm not sure.

src/Lean/Data/PersistentHashMap.lean
```diff
@@ -70,20 +70,24 @@ private theorem size_set {ks : Array α} {vs : Array β} (h : ks.size = vs.size)
...
       else insertAtCollisionNodeAux n (i+1) k v
     else
-      ⟨Node.collision (keys.push k) (vals.push v) (size_push heq k v), IsCollisionNode.mk _ _ _⟩
+      (⟨Node.collision (keys.push k) (vals.push v) (size_push heq k v), IsCollisionNode.mk _ _ _⟩, false)
   | ⟨Node.entries _, h⟩, _, _, _ => False.elim (nomatch h)
 
-def insertAtCollisionNode [BEq α] : CollisionNode α β → α → β → CollisionNode α β :=
+/-- Inserts a key-value pair into a CollisionNode, also returning whether an existing value was replaced. -/
+def insertAtCollisionNode [BEq α] : CollisionNode α β → α → β → CollisionNode α β × Bool  :=
   fun n k v => insertAtCollisionNodeAux n 0 k v
```
joehendrix, line 91: If the overhead from the pair is contributing to performance loss, then I think it would be to make `insertAtCollisionNode` not recursive and then inline it into `insertAux` for further optimization.  A non-recursive definition may be something like the following:
```lean
def insertAtCollisionNode [BEq α] : CollisionNode α β → α → β → CollisionNode α β × Bool  :=
| ⟨Node.collision keys vals heq, _⟩, k, v =>
    let i := (keys.findIdx? k).getD key.size 
    if h : i < keys.size then
      let idx : Fin keys.size := ⟨i, h⟩;
      let j : Fin vals.size := ⟨i, by rw [←heq]; assumption⟩
      (⟨Node.collision (keys.set idx k) (vals.set j v) (size_set heq idx j k v), IsCollisionNode.mk _ _ _⟩, true)
    else
      (⟨Node.collision (keys.push k) (vals.push v) (size_push heq k v), IsCollisionNode.mk _ _ _⟩, false)
| ⟨Node.entries _, h⟩, _, _ => False.elim (nomatch h)
```

semorrison:
Actually rebasing this time.

!bench

semorrison:
!bench

semorrison:
Hooray, that's much nicer news!!

src/Lean/Data/PersistentHashMap.lean
semorrison: Looks like the performance loss was not real, so I'm not going to pursue this.

joehendrix (approved):

semorrison:
!bench

semorrison:
If someone is interested in picking this up and seeing if the performance cost coming from the tuples can be avoided, please feel free to take over here.

digama0:
!bench

digama0:
!bench

digama0:
!bench

digama0:
Subsumed by #4503

PR #3050, opened 2023-12-11
chore: nix flake update

### LINKED ISSUE #3048: `nix-shell` fails to build on `aarch64-darwin`/macOS
### Prerequisites

### Description

`nix-shell` fails to build on `aarch64-darwin`/macOS.

### Context

Following the [build setup instruction](https://github.com/leanprover/lean4/blob/e6c0484074806d8a7d54c4a965d0651d0308e8d6/doc/make/index.md), when trying to use `nix-shell` to setup the build environment, the command fails.

I reported the issue in [this Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Cannot.20determine.20lean.20version.20in.20lean4.20repo/near/406873231). The thread is about the need to compile Lean 4 in the Lean 4 repository because `elan` now (correctly) no longer falls back to a default Lean version when it cannot find the artifact.

### Steps to Reproduce

On `aarch64-darwin`/macOS, run `nix-shell` in the root directory.

(Alternatively, build the problematic input with`nix build .#packages.aarch64-darwin.lean-all`.)

**Expected behavior:** build succeeds

**Actual behavior:** build fails with the following log:

```
[...]
 7/42 Test #11: test.fileclone ...................***Failed    1.14 sec
/tmp/nix-build-ccache-4.8.1.drv-2/source/test/suites/base.bash: line 556: syntax error near unexpected token `&&'
/tmp/nix-build-ccache-4.8.1.drv-2/source/test/suites/base.bash: line 556: `       && $COMPILER -E test1.c -gz >preprocessed.i 2>/dev/null \'
Compiler:         clang (/nix/store/l5dr4rvjmiybf3la6mk6gzhszrb9py3v-Toolchains/XcodeDefault.xctoolchain/bin/clang)
Compiler version: clang version 11.1.0
CUDA compiler:    not available

Running test suite fileclone.
FAILED

Test suite:     fileclone (line 32)
Test case:      Base case
Failure reason: Failed to clone
[...]
The following tests FAILED:
         11 - test.fileclone (Failed)
Errors while running CTest
```

### Versions

Lean version: dd42a0919d92e5eb5440317d77f163e9296a290a
OS version:
```
$ uname -a
Darwin rubeno.local 23.1.0 Darwin Kernel Version 23.1.0: Mon Oct  9 21:27:24 PDT 2023; root:xnu-10002.41.9~6/RELEASE_ARM64_T6000 arm64
```

### Additional Information

The issue is a bug in the CCache v4.8.1 test script. It has been [fixed](https://github.com/ccache/ccache/commit/6372b867387648ce79b58c54ce345c2442c9493d) and the patch is available in CCache v4.8.2. I've tested with a local `nix flake update`. The updated flake lock now fetches CCache v4.8.3 which fixes this issue and can successfully build Lean 4. I'm planning to submit a PR for that.

Please review if it breaks other usages that depend on the locked flake versions in that PR.

crvdgc:
In order to fix nix-shell failure due to CCache 4.8.1 test script. Closes #3048.

### DIFF
--- a/doc/flake.lock
+++ b/doc/flake.lock
@@ -17,13 +17,32 @@
         "type": "github"
       }
     },
-    "flake-utils": {
+    "flake-compat": {
+      "flake": false,
       "locked": {
-        "lastModified": 1656928814,
-        "narHash": "sha256-RIFfgBuKz6Hp89yRr7+NR5tzIAbn52h8vT6vXkYjZoM=",
+        "lastModified": 1673956053,
+        "narHash": "sha256-4gtG9iQuiKITOjNQQeQIpoIB6b16fm+504Ch3sNKLd8=",
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "rev": "35bb57c0c8d8b62bbfd284272c928ceb64ddbde9",
+        "type": "github"
+      },
+      "original": {
+        "owner": "edolstra",
+        "repo": "flake-compat",
+        "type": "github"
+      }
+    },
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1701680307,
+        "narHash": "sha256-kAuep2h5ajznlPMD9rnQyffWG8EM/C73lejGofXvdM8=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "7e2a3b3dfd9af950a856d66b0a7d01e3c18aa249",
+        "rev": "4022d587cbbfd70fe950c1e2083a02621806a725",
         "type": "github"
       },
       "original": {
@@ -41,7 +60,7 @@
       },
       "locked": {
         "lastModified": 0,
-        "narHash": "sha256-YnYbmG0oou1Q/GE4JbMNb8/yqUVXBPIvcdQQJHBqtPk=",
+        "narHash": "sha256-7SW745qUDm25qUIQ7pm65rKv6d9LNpVrm2T3mSW8O/g=",
         "path": "../.",
         "type": "path"
       },
@@ -53,11 +72,11 @@
     "lean4-mode": {
       "flake": false,
       "locked": {
-        "lastModified": 1659020985,
-        "narHash": "sha256-+dRaXB7uvN/weSZiKcfSKWhcdJVNg9Vg8k0pJkDNjpc=",
+        "lastModified": 1689286482,
+        "narHash": "sha256-tD5Ysa24fMIS6ipFc50OjabZEUge4riSb7p4BR05ReQ=",
         "owner": "leanprover",
         "repo": "lean4-mode",
-        "rev": "37d5c99b7b29c80ab78321edd6773200deb0bca6",
+        "rev": "d1c936409ade7d93e67107243cbc0aa55cda7fd5",
         "type": "github"
       },
       "original": {
@@ -80,6 +99,22 @@
         "owner": "leanprover",
         "ref": "refs/pull/57/merge",
         "repo": "LeanInk",
+        "type": "github"
+      }
+    },
+    "libgit2": {
+      "flake": false,
+      "locked": {
+        "lastModified": 1697646580,
+        "narHash": "sha256-oX4Z3S9WtJlwvj0uH9HlYcWv+x1hqp8mhXl7HsLu2f0=",
+        "owner": "libgit2",
+        "repo": "libgit2",
+        "rev": "45fd9ed7ae1a9b74b957ef4f337bc3c8b3df01b5",
+        "type": "github"
+      },
+      "original": {
+        "owner": "libgit2",
+        "repo": "libgit2",
         "type": "github"
       }
     },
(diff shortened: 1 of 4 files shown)


nomeata:
Thanks for this PR!

The nix CI is failing. It seems somewhere a rust package vendor hash needs updating. @crvdgc, let us know if you can find and fix that or need advise.

crvdgc:
Hi, @nomeata. The problem is the `lean-mdbook`'s [`cargoDeps.outputHash`](https://github.com/leanprover/lean4/blob/178ab8ef2e7aba9e18c9fcfeda12e763d62983af/doc/flake.nix#L30). I'm not sure what's the dependency requirement for that directory, so I just bumped the hash as well as the `flake.lock` file together. Let me know if this creates any futher problem and I'll dive deeper.

nomeata:
Thanks, CI passes, so that should be sufficient.

@Kha, are there things in the nix setup that aren't covered by CI that we should check manually before merging?

Kha:
We should check the docs after merging, other than that CI is what it's for :)

nomeata:
With full-ci, the sanitize job failed, probably because of newer LLVM or something?
```
[    ] Linking /home/runner/work/lean4/lean4/build/stage1/bin/lake
/nix/store/1fn92b0783crypjcxvdv6ycmvi27by0j-binutils-2.40/bin/ld: /nix/store/07haalw3zm3yl0rbphvrg59fl8ain36k-clang-wrapper-15.0.7/resource-root/lib/linux/libclang_rt.asan_cxx-x86_64.a(ubsan_type_hash_itanium.cpp.o): undefined reference to symbol '_ZTIN10__cxxabiv121__vmi_class_type_infoE@@CXXABI_1.3'
/nix/store/1fn92b0783crypjcxvdv6ycmvi27by0j-binutils-2.40/bin/ld: /nix/store/myw67gkgayf3s2mniij7zwd79lxy8v0k-gcc-12.3.0-lib/lib/libstdc++.so.6: error adding symbols: DSO missing from command line
clang-15: error: linker command failed with exit code 1 (use -v to see invocation)
```

crvdgc:
Hi, looking at the full-ci failure, I added one fix for [a warning](https://github.com/leanprover/lean4/actions/runs/7170442622/job/19523206399#step:11:82) for [CMP0148](https://cmake.org/cmake/help/latest/policy/CMP0148.html) because CMake is bumped to 3.27.

As for the actual error, my hypothesis is that to use `-fsanitize=address`, `compiler-rt` needs to be enabled. Although previously `g++` picked that up by itself. See [the `compiler-rt` page](https://compiler-rt.llvm.org/) and [the `AddressSanitizer` page](https://clang.llvm.org/docs/AddressSanitizer.html).

I should note that I'm not familiar with LLVM/clang. Feel free to edit or invite more experts for reviewing.

nomeata:
> I should note that I'm not familiar with LLVM/clang. Feel free to edit or invite more experts for reviewing.

Me neither, so I'm secretly hoping you'll be able to figure it out so that I don't have to, and will be grateful for that :-)

Kha:
We might want to consider not using shell.nix for these configurations if it makes our lives harder rather than easier :)

nomeata:
If it works it probably makes it easier, e.g. when we have to debug it outside CI, doesn't it?

crvdgc:
What about pinning `nixpkgs` to [the revision where CCache is bumped to 4.8.2](https://github.com/NixOS/nixpkgs/commit/94a9bd0f55cca537d6269163c6bb91590b023ecb)? That should be generate the minimum breakage but also solves #3048. If that doesn't work as well, then maybe we should wait for upstream change.

I can put the above revision in place of `nixpkgs-unstable` in the root `flake.nix`, or just updating the lock files. Pinning in the flake may be ugly, but it makes the intent clear. What do you think, @nomeata?

nomeata:
If you just want to scratch that itch, and aren't motivated to track down the other issue (which is fair), then that's a plausible approach. It's still from nixpkg master, so the cache should work.

I'd pin it in the flake with a comment that explains why that commit.

You could also see if nixpkg-23.11 happens to work; using stable if we can would be good.

crvdgc:
Thank you for the suggestions @nomeata. I tried `nixpkgs` 23.11 and it can compile Lean 4 but fails to build the `doc`. `nixpkgs` 23.05 worked for me locally (on `aarch64-darwin`), so I'll try one more time :)

crvdgc:
Hi, the CI passed, I wonder if I should rebase again and try to merge this PR?

src/CMakeLists.txt
```diff
@@ -245,7 +245,12 @@ if(CCACHE AND NOT CMAKE_CXX_COMPILER_LAUNCHER AND NOT CMAKE_C_COMPILER_LAUNCHER)
 endif()
 
 # Python
-find_package(PythonInterp)
+# https://cmake.org/cmake/help/latest/policy/CMP0148.html
+if(${CMAKE_VERSION} VERSION_LESS_EQUAL "3.12")
+  find_package(PythonInterp)
+else ()
+  find_package(Python COMPONENTS Interpreter)
+endif ()
```
Kha, line 253: This whole block shouldn't be necessary anymore, could you try deleting it?

crvdgc: Hi, had another go at this. The CMake commit is dropped. Let's see if the CI is happy.

PR #3051, opened 2023-12-12
chore: robustify PR release workflow

nomeata:
the workflow is triggered not only by pull-request-CI-runs but also by
others. These should be skipped.

Also, no need to query the Github API to get the pull request number and
head sha, they are part of the payload, it seems.

### DIFF
--- a/.github/workflows/pr-release.yml
+++ b/.github/workflows/pr-release.yml
@@ -16,20 +16,16 @@ on:
 jobs:
   on-success:
     runs-on: ubuntu-latest
-    if: github.event.workflow_run.conclusion == 'success' && github.repository == 'leanprover/lean4'
+    if: github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'pull_request' && github.repository == 'leanprover/lean4'
     steps:
-      - name: Retrieve PR number and head commit
+      - name: Set PR number and head commit
         uses: actions/github-script@v7
         id: workflow-info
         with:
           script: |
-            const reply = await github.rest.actions.getWorkflowRun({
-                owner: context.repo.owner,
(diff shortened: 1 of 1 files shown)


Kha:
If they are part of the payload, can't we directly access them in `${{ }}`?

nomeata:
Yes, but it makes the `if` condition and the value even longer, so I left it like this. I can change it if you dislike it.

Kha:
Ok, fine with me

.github/workflows/pr-release.yml
```diff
@@ -16,20 +16,16 @@ on:
...
         with:
           script: |
-            const reply = await github.rest.actions.getWorkflowRun({
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                run_id: context.payload.workflow_run.id,
-            })
-            core.setOutput('pullRequestNumber', reply.data.pull_requests[0].number)
-            core.setOutput('sourceHeadSha',     reply.data.pull_requests[0].head.sha)
+            // console.log(`context.payload: ${JSON.stringify(context.payload, null, 2)}`)
```
Kha, line 26: Will the comment stay in there? :)

nomeata: Yes, it's useful for debugging to not have to look it up everytime :-)

(unless you mind)

nomeata:
Not robust enough, it seems :-(

https://github.com/leanprover/lean4/actions/runs/7187252254/job/19574429098

Kha:
classic undefined behavior

nomeata:
Should have left the debugging output on :-)

But that was this odd very old PR that got activated, maybe that played a role. I'll keep an eye open.

PR #3054, opened 2023-12-12
feat: snippet extension

Vtec234:
# Summary

This makes a small addition to our take on the LSP protocol
in the form of supporting snippet text edits.
It has been discussed [here](https://github.com/microsoft/language-server-protocol/issues/592) on the LSP issue tracker for a while,
but seems unlikely to be added anytime soon.
This feature was requested by @PatrickMassot for the purposes
of supporting Lean code templates in code actions and widgets.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -36,6 +36,8 @@ v4.5.0 (development in progress)
   +termination_by _ x => hwf.wrap x
   ```
 
+* Support snippet edits in LSP `TextEdit`s. See `Lean.Lsp.SnippetString` for more details.
+
 v4.4.0
 ---------
 


PatrickMassot:
The LSP issue isn't easy to read, so let me quote it. The main point is [here](https://github.com/microsoft/language-server-protocol/issues/592#issuecomment-767367061):
```quote
we decided that we only add something to the spec if we have an implementation in a client. Otherwise things are hard to spec and to get feedback on.
```
You would think the process is: 1) add some feature to the LSP spec 2) get this feature implemented in editors. But the comment above proves it goes the other way around. So indeed "insert snippets as text edits" is not yet part of the LSP specification but it is already [supported by VSCode](https://github.com/microsoft/vscode/issues/145374).

david-christiansen:
@PatrickMassot Requiring implementations before adding something to a spec is very good practice - it's difficult to know what will be good in practice without having first used it. IMHO they should require at least two completely independent implementations - this would also help discover places where the documentation or specification is lacking and help ensure that the protocol actually does meet its goal of working in multiple editors.

src/Lean/Data/Lsp/Basic.lean
```diff
@@ -78,6 +78,18 @@ structure Command where
   arguments? : Option (Array Json) := none
   deriving ToJson, FromJson
 
+/-- A snippet string is a template which allows to insert text
+and to control the editor cursor when insertion happens.
```
david-christiansen, line 82: Style nitpick: "allows" should always take a direct object.


```suggestion
/-- A snippet string is a template that simultaneously provides a string to be inserted
and a set of positions for the editor cursor within the inserted string.
```

src/Lean/Data/Lsp/Basic.lean
```diff
@@ -78,6 +78,18 @@ structure Command where
...
+
+A snippet can define tab stops and placeholders with `$1`, `$2`
```
david-christiansen, line 84: I wouldn't call these "tab stops", as that's already used for the multiples-of-spaces that the editor indents to (for Lean code, typically every other column is a tab stop; for Python, it's typically every fourth).

What about just calling them "placeholders"?

src/Lean/Data/Lsp/Basic.lean
```diff
@@ -78,6 +78,18 @@ structure Command where
...
+and `${3:foo}`. `$0` defines the final tab stop, it defaults to
+the end of the snippet. Variables are defined with `$name` and
```
david-christiansen, line 86: What about "`$0` denotes the final position of the cursor after the placeholders have been filled out, which defaults to the end of the snippet if not provided"?

src/Lean/Data/Lsp/Basic.lean
```diff
@@ -87,6 +99,22 @@ structure TextEdit where
...
+  as fallback in case the editor does not support this feature.
+
+  *Note* that a snippet edit can always be performed as a normal text edit.
+  In VSCode, this will happen when no matching editor is open or when a workspace edit
+  contains snippet edits for multiple files. In that case only those that match the active editor
+  will be performed as snippet edits and the others as normal text edits. -/
+  /- NOTE: Similar functionality may be added to LSP in the future:
+  see [issue #592](https://github.com/microsoft/language-server-protocol/issues/592).
+  If such an addition occurs, this field should be deprecated. -/
+  leanExtSnippet? : Option SnippetString := none
```
david-christiansen, line 117: Multiple independent `*NOTE*`s, in different formats, makes me nervous here - it makes it sound like my files are getting deleted!

What about something like this?

```suggestion
  This is a Lean-specific extension to the LSP standard,
  so `newText` should still be set to a correct value
  as fallback in case the editor does not support this feature. Even editors
  that support snippets may not always use them; for instance, VS Code will
  only use snippets in the currently-active file when the language server
  provides edits to multiple files at once.
  -/
  /- NOTE: Similar functionality may be added to LSP in the future:
  see [issue #592](https://github.com/microsoft/language-server-protocol/issues/592).
  If such an addition occurs, this field should be deprecated. -/
  leanExtSnippet? : Option SnippetString := none
```

src/Lean/Data/Lsp/Basic.lean
Vtec234: Fwiw this is copy-pasta from [VSC docs](https://code.visualstudio.com/api/references/vscode-api#SnippetString), but why not improve on their wording :)

src/Lean/Data/Lsp/Basic.lean
Vtec234: Sure! Except I am now fully convinced of the utility of [semantic linebreaking](https://sembr.org/), so will format slightly differently.

src/Lean/Data/Lsp/Basic.lean
Vtec234: Actually, how about just replacing the entire docstring with a link to that? It will not fix the wording, but it will update automatically whenever they happen to change the docs; additionally, I don't entirely understand the details of their syntax, so am not comfortable trying to explain its minutiae.

david-christiansen: I did a bit more reading, and it seems that the syntax (and the "tab stop" terminology) comes [from TextMate](https://macromates.com/manual/en/snippets) - given that it's established terminology, I think we should just use it, and attempt to resolve reader confusion directly. I find the TextMate docs to be easier to understand than the ones you linked to, and the [non-API VSC docs](https://code.visualstudio.com/docs/editor/userdefinedsnippets) are even better. But neither of these describe [LSP snippets](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax), which are (as far as I can tell) as subset of both TextMate and VSC snippets.

Anyway, with all that background, I'd actually propose a different modification here that acknowledges the overloaded terminology but doesn't try to avoid it, and that calls users' attention to the specific syntax that LSP supports. In particular, the variables supported in LSP are fewer than those in VS Code, LSP regexes have fewer features than Javascript regexes,

What about this alternative approach?
```
A snippet is a template that includes instructions for subsequent edits once the string is inserted, expressed in a domain-specific language based on one from TextMate. Snippets may include:

 * Designated positions for subsequent user input, called "tab stops" after their most frequently-used keybinding, are denoted with `$1`, `$2`, and so forth. `$0` denotes the final cursor position after all edits are completed, defaulting to the end of the string. Snippet tab stops are unrelated to tab stops used for indentation.
 * Tab stops with default values, called _placeholders_, written `${1:default}`. The default may itself contain a tab stop or a further placeholder, and multiple options to select from may be provided by surrounding them with `|`s and separating them with `,`, as in `${1|if $2 then $3 else $4,if let $2 := $3 then $4 else $5|}`. 
 * One of a set of predefined variables that are replaced with values such as the current line number (`$TM_LINE_NUMBER`) or the text that was selected when the snippet was invoked (`$TM_SELECTED_TEXT`).
 * Formatting instructions to modify variables using regular expressions or a set of predefined filters.

The full syntax and semantics of snippets, including the available variables and the rules for escaping control characters, are described in the [LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax).
```

src/Lean/Data/Lsp/Basic.lean
```diff
@@ -87,6 +99,20 @@ structure TextEdit where
   range : Range
   /-- The string to be inserted. For delete operations use an empty string. -/
   newText : String
+  /-- If this field is present and the editor supports it,
+  `newText` is ignored
+  and an interactive snippet edit is performed instead.
+
+  This is a Lean-specific extension to the LSP standard,
```
david-christiansen, line 106: After reading up more on this, I think it's most accurate to say:
```suggestion
  The use of snippets in `TextEdit`s is a Lean-specific extension to the LSP standard,
  which only provides for snippets in completion items,
```

PatrickMassot: Is this sentence missing a word? It doesn't seem to make sense as it is.

Vtec234: Oh huh, I didn't realize that snippets were already supported in completions but not code actions. That's weird! But what you wrote is what I meant by "This" anyway, so may as well disambiguate.

src/Lean/Data/Lsp/Basic.lean
Vtec234: Done (with some rewording)!

PR #3058, opened 2023-12-13
fix: manage all declarations in a given derive

### LINKED ISSUE #3057: `deriving` only derives the first declaration for a few handlers
### Prerequisites

### Description

Derive handlers expect to be given an array of names from which to derive things from. However, many core handlers such as [`mkDecEqInstanceHandler`](https://github.com/leanprover/lean4/blob/2f216b52550e7b12b342d6b1717897ad3630ae63/src/Lean/Elab/Deriving/DecEq.lean#L189) and [`mkReprInstanceCmds` ](https://github.com/leanprover/lean4/blob/2f216b52550e7b12b342d6b1717897ad3630ae63/src/Lean/Elab/Deriving/Repr.lean#L107) only generate the derivations for the first declaration.

### Context

This issue comes up when trying to derive things for multiple inductive types, i.e:

```lean
mutual
  inductive Tree : Type :=
    | node : ListTree → Tree
    deriving Repr, DecidableEq

  inductive ListTree : Type :=
    | nil : ListTree
    | cons : Tree → ListTree → ListTree
    deriving Repr, DecidableEq
end

inductive Wrap : Type :=
  | wrap : Bool → Tree → Wrap
  deriving Repr, DecidableEq -- works :-)

inductive ListWrap : Type :=
  | wrap : Bool → ListTree → ListWrap
  deriving Repr, DecidableEq -- failed to synthesize instance
 ```

In the same manner, writing 
```lean
deriving instance DecidableEq for Tree, ListTree
```
produces the same problem.
### Versions

Lean 4.4.0-rc1

### Impact

arthur-adjedj:
Closes #3057


semorrison:
Could you please rebase onto `nightly-with-mathlib` so we get a CI run against Mathlib?

src/Lean/Elab/Deriving/DecEq.lean
```diff
@@ -169,7 +169,7 @@ def mkEnumOfNatThm (declName : Name) : MetaM Unit := do
       value, type
     }
 
-def mkDecEqEnum (declName : Name) : CommandElabM Unit := do
+def mkDecEqEnum (declName : Name) : CommandElabM Bool := do
```
semorrison, line 172: Why have you changed all the return types to `Bool`, give  that these functions always return `true`? Seems clearer to just add a `return true` at the end of `mkDecEqInstanceHandler`.

arthur-adjedj: `mkDecEq` currently returns false if the inductive type is nested, hence why I've thought it would be better to propagate the bool here, but you're right that it's not useful for `mkDecEqEnum`, I'll change that.
In practice, that boolean is only used to throw an error saying that the type isn't handled when it isn't ([here](https://github.com/leanprover/lean4/blob/4497aba1a985619023adf0e1c655e769b12735e7/src/Lean/Elab/Deriving/Basic.lean#L91)), but if that's the case, why not throw the error locally instead of propagating it that way ?

semorrison: Well, it's actually using that flag to decide to call the default handler, which currently happens to error, but might not in future. (E.g. in Lean3 the default handler unfolded the head constant and then looked for an instance.)

PR #3060, opened 2023-12-13
fix: allow generalization in let

### LINKED ISSUE #3065: `let`/ else pattern matching does not substitute in expected type
### Prerequisites

### Description

`let 3 := n | failure` does not substitute `n` with `3` in the target type, even though equivalent spellings do.

### Context

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Pattern.20matching.20in.20let.20not.20substituting/near/407080090)

This impacts leanprover-community/quote4, which transpiles to this type of `let` statement.

### Steps to Reproduce

The following does not work:
```lean
example (n : Nat) (i : Fin n) : Option (Fin 3) := do
  let 3 := n | failure
  return i -- fails, pattern match did not substitute in hypothesis

example (n : Nat) (i : Fin 3) : Option (Fin n) := do
  let 3 := n | failure
  return i -- fails, pattern match did not substitute in goal
```
but the corresponding `match` does:
```lean
example (n : Nat) (i : Fin n) : Option (Fin 3) := do
  match n with
  | 3 => return i -- ok
  | _ => failure

example (n : Nat) (i : Fin 3) : Option (Fin n) := do
  match n with
  | 3 => return i -- ok
  | _ => failure
```
as does `if let`
```lean
example (n : Nat) (i : Fin n) : Option (Fin 3) := do
  if let 3 := n then
    return i
  else
    failure

example (n : Nat) (i : Fin 3) : Option (Fin n) := do
  if let 3 := n then
    return i
  else
    failure
```
**Expected behavior:** All three examples should behave identically

**Actual behavior:** The first one does not type-check

### Versions

leanprover/lean4:v4.4.0-rc1

### Impact

eric-wieser:
As suggested by @kmill, removing an unnecessary `let` (possibly only there in the first place for copy/paste reasons) seems to fix the included test.

This makes `~q()` matching in quote4 noticeably more useful in things like `norm_num` (as it fixes https://github.com/leanprover-community/quote4/issues/29)

It also makes a quote4 bug slightly more visible (https://github.com/leanprover-community/quote4/issues/30), but the bug there already existed anyway, and isn't caused by this patch.

Fixes #3065


semorrison (approved):

eric-wieser:
@semorrison, is there any hope of this landing in the 4.5.0 release?

semorrison:
> @semorrison, is there any hope of this landing in the 4.5.0 release?

I'm sorry, this didn't make it.

eric-wieser:
And just to check I understand the rc process, the fact this missed 4.5.0rc1 means it certainly will miss any future rcs?

semorrison:
> And just to check I understand the rc process, the fact this missed 4.5.0rc1 means it certainly will miss any future rcs?

Well, hopefully it can make `v4.6.0-rc1`! 

The status of `rc2` is still a bit up in the air. If there is really substantial demand (like 10 maintainers have +1'd a post saying that Mathlib wants something asap), I think we're open to cutting a `rc2` solely for the sake a new features. But the preference is that we only make `rc2` when there is a bad problem that needs fixing.

semorrison:
@Kha, I know @eric-wieser is keen to get this in, as it will enable new uses of Qq in writing tactic (`positivity`) extensions in Mathlib. As it touches `do` notation I think it needs review from you.

digama0 (approved):

Kha:
It looks like a strict improvement to me but I'll give @leodemoura a chance to look at it as well since he wrote the code. In general we haven't gone back to the `do` code in a long time so it's not a good time for any larger changes (which this one isn't).

Kha:
Ok, I still don't see any possible downsides

PR #3067, opened 2023-12-13
chore: remove invalid link

TristanCacqueray:
This change removes the `metaprogramming.md` link as it was renamed to syntax_example in 80c04c56 which no longer seems accurate.


Kha:
Fixed in #3216

PR #3066, opened 2023-12-13
chore: pr-release.yml: use API to get pull request number

nomeata:
partially reverting 6a629f7d7ff20684102a6d6a106b265aea74f4db. What a
mess.

PR #3070, opened 2023-12-14
chore: refactor pr-release.yml to avoid 'await'

kim-em:
#3066 is causing CI failures, e.g. [here](https://github.com/leanprover/lean4/actions/runs/7202184616/job/19619827364).

Although there are plenty of examples of using `await` in a Github workflow script block, the error *seems* to be about this. This refactor hopefully works around that, but I'm still uncertain of a root cause.


semorrison:
I really need PR releases up and running again, so I am going to merge this now and test in production.

PR #3071, opened 2023-12-14
fix: don't panic in leanPosToLspPos

kim-em:
Testing a problem in the REPL.

### DIFF
--- a/src/Lean/Data/Lsp/Utf16.lean
+++ b/src/Lean/Data/Lsp/Utf16.lean
@@ -62,7 +62,7 @@ end String
 namespace Lean
 namespace FileMap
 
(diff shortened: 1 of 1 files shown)


Kha:
Please use the same defaulting as in the function above it

src/Lean/Data/Lsp/Utf16.lean
```diff
@@ -62,21 +62,24 @@ end String
 namespace Lean
 namespace FileMap
 
+private def colPos (text : FileMap) (line : Nat) : String.Pos :=
```
mhuisi, line 65: s/`colPos`/`lineStartPos`?

mhuisi (approved):

PR #3075, opened 2023-12-15
feat: improvements to test_extern command

kim-em:
Two improvements [suggested](https://github.com/leanprover/lean4/pull/2970#issuecomment-1853436906) by @digama0 after the initial PR was merged.

* Allow testing `implemented_by` attributes as well.
* Use `DecidableEq` rather than `BEq` for stricter testing.

### DIFF
--- a/tests/lean/run/utf8英語.lean
+++ b/tests/lean/run/utf8英語.lean
@@ -1,7 +1,6 @@
 import Lean.Util.TestExtern
 
-instance : BEq ByteArray where
-  beq x y := x.data == y.data
+deriving instance DecidableEq for ByteArray
 
 test_extern String.toUTF8 ""
 test_extern String.toUTF8 "\x00"


tests/lean/test_extern.lean.expected.out
```diff
@@ -1,5 +1,5 @@
-test_extern.lean:7:0-7:17: error: test_extern: HAdd.hAdd does not have an @[extern] attribute
-test_extern.lean:9:0-9:86: error: native implementation did not agree with reference implementation!
+test_extern.lean:6:0-6:17: error: test_extern: HAdd.hAdd does not have an @[extern] attribute or @[implemented_by] attribute
+test_extern.lean:8:0-8:86: error: native implementation did not agree with reference implementation!
 Compare the outputs of:
```
digama0, line 3: Why does the error message say to compare the outputs instead of showing what the outputs are? I think it should actually print the outputs (when possible), because these issues can be OS dependent so it's important to know what value was actually produced in CI since they may not reproduce locally.

semorrison: That's a good suggestion, but I'd prefer not to take it on now. Happy to either open an issue, let you do that. Or, for that matter, to accept commits on this PR, or promise to review a new PR! :-)

PR #3078, opened 2023-12-15
feat: add `quot_precheck` for expression tree elaborators (`binop%`, etc.)

kmill:
There were no `quot_precheck` instances registered for the expression tree elaborators, which prevented them from being usable in a `notation` expansion without turning off the quotation prechecker.

Users can evaluate whether `set_option quotPrecheck false` is still necessary for their `notation` definitions.

### DIFF
--- a/src/Lean/Elab/Quotation/Precheck.lean
+++ b/src/Lean/Elab/Quotation/Precheck.lean
@@ -144,31 +144,31 @@ private def isSectionVariable (e : Expr) : TermElabM Bool := do
 section ExpressionTree
 
 @[builtin_quot_precheck Lean.Parser.Term.binrel] def precheckBinrel : Precheck
-  | `(binrel% $_ $a $b) => do precheck a; precheck b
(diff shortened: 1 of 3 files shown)


src/Lean/Elab/Quotation/Precheck.lean
```diff
@@ -141,4 +141,36 @@ private def isSectionVariable (e : Expr) : TermElabM Bool := do
   runPrecheck singleQuot.getQuotContent
   adaptExpander (fun _ => pure singleQuot) stx expectedType?
 
+section ExpressionTree
+
+@[builtin_quot_precheck Lean.Parser.Term.binrel] def precheckBinrel : Precheck
+  | `(binrel% $_ $a $b) => do precheck a; precheck b
```
Kha, line 147: What about the first argument, shouldn't we precheck it as well for the sake of completeness?

kmill: Yes, that makes sense. Done.

PR #3080, opened 2023-12-16
fix: pass CMAKE_C/CXX_COMPILER to all stages

### LINKED ISSUE #2863: [cmake] Some command line arguments not persisting to stage0 step
### Prerequisites

### Description

Some variables passed to `cmake` (such as `CMAKE_C_COMPILER`) at configuration step do not persist when the project is built. 

### Context

The first thing the `CMakeLists.txt` file does is iterate through all cached cmake variables and set some special variables that are passed on to step0. But when the project is created with the `project` command, some of these variables are set to default values. 
Additionally, custom toolchain files are invalidated because they are not read until the project is created. 

### Steps to Reproduce

In a system with both `clang` and `gcc` installed, with `gcc` as the default compiler:

1. `cmake -B build -DCMAKE_C_COMPILER=clang`  ==> The C compiler identification is Clang 17.0.4
2. `make -C build` ==> -- The C compiler identification is GNU 13.2.0

--

I noticed other cmake-related issues could be directly related to this. 
Would a PR to attempt to fix this be acceptable?

Vtec234:
# Summary

This PR fixes #2863, making a guess as to how that should be done. We replace the `if/else/else` with `if/else;if` so that `CMAKE_C/CXX_COMPILER` is stored *both* in `PLATFORM_ARGS` and `CL_ARGS` and consequently passed to all stages. I am not sure what the original intent of the logic was, but the comment below makes it sound like this was indeed a bug.

```lean
# Use standard release build (discarding LEAN_CXX_EXTRA_FLAGS etc.) for stage0 by default since it is assumed to be "good", but still pass through CMake platform arguments (compiler, toolchain file, ..).
```

Link to `RFC` or `bug` issue: #2863


Vtec234:
Hm, it looks like the CI build actually [sets `CMAKE_C_COMPILER` to a custom LLVM under `stage1/bin`](https://github.com/leanprover/lean4/actions/runs/7232967915/job/19707731194#step:11:53). Before this PR, the stage0 build was still [using the system compiler](https://github.com/leanprover/lean4/actions/runs/7228723799/job/19698609546?pr=3078#step:11:79) instead of that. Perhaps this PR is the wrong approach and people should just use `STAGE0_CMAKE_C_COMPILER` explicitly.

Kha:
We can switch to `STAGE0_CMAKE_C_COMPILER=cc` in CI, though I'm actually not sure why that's necessary - we shouldn't be forwarding `-nostdinc` in this case.

Kha:
> We can switch to STAGE0_CMAKE_C_COMPILER=cc in CI

This would likely be the way to go but the PR should be changed to only forward *_COMPILER not any other options

PR #3082, opened 2023-12-16
feat: add call hierarchy support

mhuisi:
This PR adds support for the "call hierarchy" feature of LSP that allows quickly navigating both inbound and outbound call sites of functions. In this PR, "call" is taken to mean "usage", so inbound and outbound references of all kinds of identifiers (e.g. functions or types) can be navigated. To implement the call hierarchy feature, this PR implements the LSP requests `textDocument/prepareCallHierarchy`, `callHierarchy/incomingCalls` and `callHierarchy/outgoingCalls`.

<details>
  <summary>Showing the call hierarchy (click to show image)</summary>
  
  show_call_hierarchy
  
</details>

<details>
  <summary>Incoming calls (click to show image)</summary>
  
  incoming_calls
  
</details>

<details>
  <summary>Outgoing calls (click to show image)</summary>
  
  outgoing_calls
  
</details>

It is based on #3159, which should be merged before this PR.

To route the parent declaration name through to the language server, the `.ilean` format is adjusted, breaking backwards compatibility with version 1 of the ILean format and yielding version 2.

This PR also makes the following more minor adjustments:
- `Lean.Server.findModuleRefs` now also combines the identifiers of constants and FVars and prefers constant over FVars for the combined identifier. This is necessary because e.g. declarations declared using `where` yield both a constant (for usage outside of the function) and an FVar (for usage inside of the function) with the same range, whereas we would typically like all references to refer to the former. This also fixes a bug introduced in #2462 where renaming a declaration declared using `where` would not rename usages outside of the function, as well as a bug in the unused variable linter where `where` declarations would be reported as unused even if they were being used outside of the function.
- The function converting `Lean.Server.RefInfo` to `Lean.Lsp.RefInfo` now also computes the `Lean.DeclarationRanges` for parent declaration names via `MetaM` and must hence be in `IO` now.
- Add a utility function `Array.groupByKey` to `HashMap.lean`.
- Stylistic refactoring of `Watchdog.lean` and `LanguageFeatures.lean`.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -192,7 +192,9 @@ example : x + foo 2 = 12 + x := by
   ought to be applied to multiple functions, the `decreasing_by` clause has to
   be repeated at each of these functions.
 
-* Modify `InfoTree.context` to facilitate augmenting it with partial contexts while elaborating a command. This breaks backwards compatibility with all downstream projects that traverse the `InfoTree` manually instead of going through the functions in `InfoUtils.lean`, as well as those manually creating and saving `InfoTree`s. See https://github.com/leanprover/lean4/pull/3159 for how to migrate your code.
+* Modify `InfoTree.context` to facilitate augmenting it with partial contexts while elaborating a command. This breaks backwards compatibility with all downstream projects that traverse the `InfoTree` manually instead of going through the functions in `InfoUtils.lean`, as well as those manually creating and saving `InfoTree`s. See [PR #3159](https://github.com/leanprover/lean4/pull/3159) for how to migrate your code.
+
+* Add language server support for [call hierarchy requests](https://devblogs.microsoft.com/cppblog/c-extension-in-vs-code-1-16-release-call-hierarchy-more/) ([PR #3082](https://github.com/leanprover/lean4/pull/3082)). The change to the .ilean format in this PR means that projects must be fully rebuilt once in order to generate .ilean files with the new format before features like "find references" work correctly again.
 
 
 v4.5.0


mhuisi:
The previous version of this PR adjusted `withDeclName` to also save the `InfoContext`. This was incorrect because this saved `InfoContext` was not fully complete (e.g. it lacked assignments for metavariables).

The core issue is that we want to save some kind of context while elaborating a term, but cannot meaningfully do so until we are fully done with elaborating the term, lest parts of the saved context may be missing. We already work around this limitation in several other places, e.g. using `Info.updateContext?`, and the parent declaration name is just one more situation where this comes up.

To solve this issue, the previous commit introduces a notion of `PartialContextInfo` that replaces the `ContextInfo` in `InfoTree.context`. The underlying idea is that while constructing the `InfoTree`, we augment it with `PartialContextInfo`s where it makes sense, and then when traversing over the tree, we merge outer `ContextInfo`s with inner `PartialContextInfo`s (with fields from inner contexts shadowing those of outer context) into `ContextInfo`s to provide all the context that is available up to this point.

The old `ContextInfo` that was added by `liftTermElabM` has been renamed to `CommandContextInfo`.

Another adjustment made in the previous commit is to remove the incorrect unused variable linter tests checking for unused `where` declarations. `where` declarations are accessible from outside of a definition and should hence not be marked as unused, something that is also fixed by this PR.

mhuisi:
(CI is red because LeanInk manually traverses the InfoTree, but the tests are green. PRs for fixing downstream repos tbd)

src/Lean/Server/References.lean
```diff
@@ -117,8 +134,8 @@ open Elab
 
 /-- Content of individual `.ilean` files -/
 structure Ilean where
-  version : Nat := 1
-  module : Name
+  version    : Nat := 2
```
digama0, line 137: Why is there a version bump if the format hasn't changed? Is the semantics different?

mhuisi: The format has changed because `RefInfo` changed to store the parent declaration. See the PR description.

src/Lean/Server/References.lean
```diff
@@ -152,64 +169,98 @@ def findReferences (text : FileMap) (trees : Array InfoTree) : Array Reference :
...
-
-This function changes every such group to use a single `FVarId` (the head of the
-chain/DAG) and gets rid of duplicate definitions.
+There are several different identifiers that should be considered equal for the purpose of finding
+all references of an identifier:
+- `FVarId`s of a function parameter in the function's signature and body
+- Chains of helper definitions like those created for do-reassignment `x := e`
+- Overlapping definitions like those defined by `where` declarations that define both an FVar
+  (for local usage) and a constant (for non-local usage)
+- Identifiers connected by `FVarAliasInfo`
```
Kha, line 178:
```suggestion
- Identifiers connected by `FVarAliasInfo` such a variables before and after `match` generalization
```
if you want to include an example

PR #3083, opened 2023-12-17
feat: extract `delabAppCore`, define `withOverApp`, and make over-applied projections pretty print

kmill:
To handle delaborating notations that are functions that can be applied to arguments, extracts the core function application delaborator as a separate function that accepts the number of arguments to process and a delaborator to apply to the "head" of the expression.

Defines `withOverApp`, which has the same interface as the combinator of the same name from std4, but it uses this core function application delaborator.

Uses `withOverApp` to improve a number of application delaborators, notably projections. This means Mathlib can stop using `pp_dot` for structure fields that have function types.

Incidentally fixes `getParamKinds` to specialize default values to use supplied arguments, which impacts how default arguments are delaborated.

### DIFF
--- a/tests/lean/283.lean.expected.out
+++ b/tests/lean/283.lean.expected.out
@@ -1,7 +1,7 @@
 283.lean:1:24-1:25: error: application type mismatch
-  f (f ?m)
+  f f
 argument
-  f ?m
+  f
 has type
   ?m : Sort ?u
 but is expected to have type
--- a/tests/lean/439.lean.expected.out
+++ b/tests/lean/439.lean.expected.out
@@ -1,20 +1,20 @@
 fn.imp : {p : P} → Bar.fn p
 439.lean:18:7-18:12: error: function expected at
-  Fn.imp fn
+  fn.imp
 term has type
   Bar.fn ?m
 439.lean:29:7-29:11: error: function expected at
-  Fn.imp fn
+  fn.imp
 term has type
   Bar.fn ?m
-Fn.imp fn : Bar.fn p
-Fn.imp fn' Bp : Bar.fn p
+fn.imp : Bar.fn p
+fn'.imp Bp : Bar.fn p
 439.lean:39:11-39:12: error: application type mismatch
-  Fn.imp fn' p
+  fn'.imp p
 argument
   p
 has type
   P : Sort u
 but is expected to have type
   Bar.fn ?m : Sort ?u
-Fn.imp fn' (sorryAx (Bar.fn ?m) true) : Bar.fn ?m
+fn'.imp (sorryAx (Bar.fn ?m) true) : Bar.fn ?m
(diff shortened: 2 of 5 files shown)


src/Lean/Expr.lean
```diff
@@ -919,6 +919,14 @@ private def getAppNumArgsAux : Expr → Nat → Nat
 def getAppNumArgs (e : Expr) : Nat :=
   getAppNumArgsAux e 0
 
+/--
+If the given expression is a sequence of function applications `f a₁ .. aₙ`,
+returns `f a₁ .. aₖ` where `k` is minimal such that `n - k ≤ maxArgs`.
```
Kha, line 924: Should it be `maxRemainingArgs` here for clarity?

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
```diff
@@ -133,43 +142,37 @@ where
...
+-/
+def useAppExplicit (paramKinds : Array ParamKind) : DelabM Bool := do
+  if ← getPPOption getPPExplicit then
+    if paramKinds.any (fun param => !param.isRegularExplicit) then return true
+
+  -- If the expression has an implicit function type, fall back to delabAppExplicit.
+  -- This is e.g. necessary for `@Eq`.
+  let isImplicitApp ← try
+      let ty ← whnf (← inferType (← getExpr))
+      pure <| ty.isForall && (ty.binderInfo == BinderInfo.implicit || ty.binderInfo == BinderInfo.instImplicit)
```
Kha, line 161:
```suggestion
      pure <| ty.isForall && (ty.binderInfo matches .implicit | .instImplicit)
```
minor code compression

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
```diff
@@ -247,12 +267,37 @@ def delabAppImplicit : Delab := do
...
     <|> pure stx
   else pure stx
 
+/--
+Delaborates applications. Removes up to `maxArgs` arguments to form
+the "head" of the application and delaborates the head using `delabHead`.
+The remaining arguments are processed depending on whether heuristics indicate that the application
+should be delaborated using `@`.
+-/
+def delabAppCore (maxArgs : Nat) (delabHead : Delab) : Delab := do
```
Kha, line 282: Are there any delabs in core that should already be using `delabAppCore` with a specific `maxArgs`?

Kha (approved):

src/Lean/Expr.lean
kmill: I find my comment to be confusing. I want `getBoundedAppFn m e` to pop off at most `m` arguments, and I want it to be "dual" to `getBoundedAppArgs` in the sense that `mkAppN (e.getBoundedAppFn m) (e.getBoundedAppArgs m) = e`.

Really, the use case is extracting a function of a particular arity, but I found it easier to think about these functions in their current form, where it is up to the caller to decide how many arguments to pop off.

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
kmill: Yeah, many builtin delaborators could use it. One example is `delabProjectionApp`, which could use it to handle over-application.

My plan was to use `delabAppCore` to upstream a more general version of [`Lean.PrettyPrinter.Delaborator.withOverApp`](https://github.com/leanprover/std4/blob/9dd24a3493cceefa2bede383f21e4ef548990b68//Std/Lean/Delaborator.lean#L24-L38) and then use this combinator to improve builtins.

src/Lean/Expr.lean
kmill: I tried to clarify the docstring.

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
kmill: I went ahead and upstreamed the combinator and improved the builtins that could be improved.

PR #3084, opened 2023-12-17
fix: hover info for `cases h : ...`

digama0:
This makes hover info, go to definition, etc work for the `h` in `cases h : e`. The implementation is similar to that used for the `generalize h : e = x` tactic.

### DIFF
--- a/src/Lean/Elab/Tactic/Induction.lean
+++ b/src/Lean/Elab/Tactic/Induction.lean
@@ -588,6 +588,7 @@ def elabCasesTargets (targets : Array Syntax) : TacticM (Array Expr × Array (Id
       liftMetaTacticAux fun mvarId => do
         let argsToGeneralize ← args.filterM fun arg => shouldGeneralizeTarget arg.expr <||> pure arg.hName?.isSome
         let (fvarIdsNew, mvarId) ← mvarId.generalize argsToGeneralize
+        -- note: fvarIdsNew contains the `x` variables from `args` followed by all the `h` variables
         let mut result := #[]
         let mut j := 0
         for arg in args do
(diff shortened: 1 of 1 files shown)


src/Lean/Elab/Tactic/Induction.lean
```diff
@@ -583,16 +596,16 @@ def elabCasesTargets (targets : Array Syntax) : TacticM (Array Expr) :=
             j := j+1
           else
             result := result.push arg.expr
-        return (result, [mvarId])
+        return ((result, hIdents.zip fvarIdsNew[j:]), [mvarId])
```
Kha, line 599: The approach is good but this indexing looks dubious to me, what exactly is happening here? How can this be correct if `shouldGeneralizeTarget` is true for at least one arg?

digama0: The return value of `generalize` is a list of fvarIds with exactly one entry for each `x` variable provided, followed by the `h` variables in order, skipping the ones that are `none`. The `filterM` line trims some of the variables, but because of the `<||> arg.hName?.isSome` part every `h` variable is preserved.

We call `generalize` with the filtered list, so it now has `fvarIdsNew` containing the selected `x` variables and all of the `h` variables. The variable `j` counts the `x` variables which have not been filtered out, so at the end of the loop it has enumerated all the `x` variables and what remains is the list of all the `h` variables, which exactly match what we first put in the `hIdents` array.

(Yes, this indexing design is horribly convoluted, but I don't feel like refactoring `generalize` right now. IMO a better design would be to return a list of `FVar x Option FVarId` pairs matching the input `GeneralizeArg`s.)

Kha: Ohh, I had no idea about the makeup of the `generalize` result. Then the indexing makes sense, thanks. Maybe a short comment to that effect on this line?

PR #3085, opened 2023-12-17
doc: fix typos/indentation

marcusrossel:


nomeata:
Thanks for your contribution! I'm not sure how useful indentation fixes are when you aren't already touching the code (if it causes a merge conflict for someone else probably not), so don't go out of your way to fix them, but typo fixes are certainly appreciated.

marcusrossel:
> typo fixes are certainly appreciated.
Are these kinds of small chore-PRs generally welcome, or are they more noise for you than anything else?

nomeata:
That is a good question. I’d say anything user-visible and obviously annoying (typos, wrong references in docstrings) is very welcome. Fixes to user-visible markup (bold theorem name) maybe as well, if it improves consistency. Typos in comments when they annoy you, but not need to be proactive about them. Indentation changes I personally don’t care too much about and would leave it to whoever touches the code next.

If you expect to do more, then grouping multiple similar ones in one PR makes reviewing quicker.

marcusrossel:
(Sorry, that last commit was an accident, but I guess it's also enough of a chore that I'll just leave it in.)

marcusrossel:
Ok, now I'm confused how the PR even got merged, but does not include the final commit...

nomeata:
Maybe there was a race condition with the Github merge queue. But separate PR is better anyways, since it isn’t about `docs` here.

PR #3087, opened 2023-12-18
feat: error on non-rfl dsimp args

marcusrossel:
This PR implements a TODO for throwing an error on `dsimp` arguments which aren't proofs by reflexivity. I don't know exactly what `resolveSimpIdTheorem?` does, so there might be some cases which slip through - but I think this implementation should at least not produce false positives.

### DIFF
--- a/src/Lean/Elab/Tactic/Simp.lean
+++ b/src/Lean/Elab/Tactic/Simp.lean
@@ -181,7 +181,6 @@ def elabSimpArgs (stx : Syntax) (ctx : Simp.Context) (eraseLocal : Bool) (kind :
           match (← resolveSimpIdTheorem? term) with
           | .expr e  =>
             if kind == .dsimp && !(← isRflProof e) && (← isProof e) then
-              -- TODO: Make this error message more informative, by stating which argument failed.
               throwErrorAt term "'dsimp' tactic only supports theorems which hold by reflexivity"
             let name ← mkFreshId
             thms ← addDeclToUnfoldOrTheorem thms (.stx name arg) e post inv kind


semorrison:
I've rebased so that CI can test Mathlib against this PR.

marcusrossel:
An example of what breaks in Mathlib:

```lean
theorem Perm.permutations' {s t : List α} (p : s ~ t) : permutations' s ~ permutations' t := by
  induction' p with a s t _ IH a b l s t u _ _ IH₁ IH₂; · simp
  · exact IH.bind_right _
  · dsimp [permutations'] -- ERROR
    rw [bind_assoc, bind_assoc]
    apply Perm.bind_left
    intro l' _
    apply perm_permutations'Aux_comm
  · exact IH₁.trans IH₂
```

Should `dsimp` accept this argument?

marcusrossel:
@semorrison I have a fix for the Mathlib breakage laying around locally. Could I please have permission to push to the corresponding Mathlib branch?

eric-wieser:
Does it make sense to fix #2678 first before removing all such lemmas from mathlib?

marcusrossel:
> Does it make sense to fix #2678 first before removing all such lemmas from mathlib `dsimp` calls?

I think that's not necessary. Fixing Mathlib's `dsimp` calls amounted to simply removing non-`rfl` arguments - no other adjustments required. And I'm guessing an implementation of #2678 would add the check for `Iff.rfl` in `isRflProof`, which is precisely the function used in this PR.

eric-wieser:
My point is that if we force mathlib to remove the non-`rfl` lemmas, then we end up removing all the `Iff.rfl` lemmas that are supposed to work in the first place. If we fix `Iff.rfl` first, then we don't end up removing things that in the long run we actually want to keep.

A compromise would be to have this warning only reject non-`rfl` non-`Iff.rfl` proofs, even though for now the latter don't actually work. That way we don't incentivize removing them prematurely from mathlib.

marcusrossel:
Ah ok, then it seems I don't understand the following point yet: What's the benefit of keeping `dsimp` arguments which aren't actually necessary for the proof to go through?

eric-wieser:
The `Iff.rfl` lemmas in `dsimp` in mathlib are most likely left over from porting lean 3 code, where they actually worked correctly. As a result, the proofs that follow those `dsimp`s will likely end up being longer or messier than necessary as a workaround.

Leaving the Iff.rfl lemmas in place means that when we fix dsimp to accept them, the old proof will start working again and CI will encourage us to remove any workarounds later in the proof (resulting in simpler proofs).

marcusrossel:
> the proofs that follow those `dsimp`s will likely end up being longer or messier than necessary as a workaround.

I just compared the proofs which are affected by this PR to the corresponding proofs in mathlib3. All the proofs are identical, so this shouldn't be an issue.

semorrison:
@marcusrossel, apologies for the long delay, but I just gave your write access at mathlib4!

semorrison:
@marcusrossel let me know if you need help with testing Mathlib here. Your changes need to go on the `lean-pr-testing-3087` branch of Mathlib to be reflected here.

src/Lean/Elab/Tactic/Simp.lean
```diff
@@ -180,6 +180,9 @@ def elabSimpArgs (stx : Syntax) (ctx : Simp.Context) (eraseLocal : Bool) (kind :
 
           match (← resolveSimpIdTheorem? term) with
           | .expr e  =>
+            if kind == .dsimp && !(← isRflProof e) && (← isProof e) then
+              -- TODO: Make this error message more informative, by stating which argument failed.
```
semorrison, line 184: This TODO seems unnecessary: you're already calling `throwErrorAt term` here, so the squiggle should come on the bad lemma.

marcusrossel:
@semorrison thanks! I'm just waiting on #2678 to be resolved before continuing with this PR. Is there some way I can indicate that this PR is waiting for another PR/issue?

semorrison:
@marcusrossel, we should check in with @thorimur re: #2678, I guess; it's unclear if it will progress.

eric-wieser:
To be clear on my stance; I don't think we need to wait for https://github.com/leanprover/lean4/issues/2678; I just think we should be conservative with the error message and only emit it on things that would still not be accepted after #2678 goes in.

PR #3090, opened 2023-12-18
fix: collect level parameters in evalExpr

### LINKED ISSUE #3091: evalExpr does not handle universe params, but #eval does
### Prerequisites

### Description

`evalExpr` fails if any universe parameters are present, but `#eval` handles them correctly.

### Context

This impedes https://github.com/leanprover-community/mathlib4/pull/9135

### Steps to Reproduce

```lean
import Lean

open Lean

def foo.{u} : Nat := (ULift.up.{u} Nat.zero).down

-- works
universe u in
#eval foo.{u}

-- fails: (kernel) invalid reference to undefined universe level parameter 'u'
#eval do
  let u : Lean.Level := .param `u
  Meta.evalExpr Nat (.const ``Nat []) (.const ``foo [u])
```

**Expected behavior:** both examples succeed and output 0

**Actual behavior:** the second example fails as shown

### Versions

4.4.0-rc1

### Impact

eric-wieser:
`elabEvalUnsafe` already does something similar: it also instantiates universe metavariables, but it is not clear to me whether that is sensible here.
To be conservative, I leave it out of this PR.

See https://github.com/leanprover/lean4/pull/3090#discussion_r1432007590 for a comparison between `#eval` and `Meta.evalExpr`. This PR is not trying to fully align them, but just to fix one particular misalignment that I am impacted by.

Closes #3091

### DIFF
--- a/src/Lean/Meta/Eval.lean
+++ b/src/Lean/Meta/Eval.lean
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Sebastian Ullrich, Leonardo de Moura
 -/
 import Lean.Meta.Check
+import Lean.Util.CollectLevelParams
 
 namespace Lean.Meta
 
@@ -11,12 +12,13 @@ unsafe def evalExprCore (α) (value : Expr) (checkType : Expr → MetaM Unit) (s
   withoutModifyingEnv do
     let name ← mkFreshUserName `_tmp
     let value ← instantiateMVars value
+    let us := collectLevelParams {} value |>.params
     if value.hasMVar then
       throwError "failed to evaluate expression, it contains metavariables{indentExpr value}"
     let type ← inferType value
     checkType type
     let decl := Declaration.defnDecl {
-       name, levelParams := [], type
+       name, levelParams := us.toList, type
        value, hints := ReducibilityHints.opaque,
        safety
     }
--- a/tests/lean/run/3091.lean
+++ b/tests/lean/run/3091.lean
@@ -1,0 +1,13 @@
+import Lean
+
+open Lean
+
+def foo.{u} : Nat := (ULift.up.{u} Nat.zero).down
+
+universe u in
+#eval foo.{u}
+
+-- this used to produce an error
+#eval do
+  let u : Lean.Level := .param `u
+  Meta.evalExpr Nat (.const ``Nat []) (.const ``foo [u])


semorrison:
It would be good to have a Mathlib build for this one. Could you rebase onto `nightly-with-mathlib`?

eric-wieser:
I'm afraid I probably have no time today to do so, but might later this week.

My figuring is that this is very unlikely to break anything, since the change only affects code that was previously producing kernel errors; but indeed it's good practice to test it anyway.

src/Lean/Meta/Eval.lean
```diff
@@ -4,19 +4,21 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
 import Lean.Meta.Check
+import Lean.Util.CollectLevelParams
 
 namespace Lean.Meta
 
 unsafe def evalExprCore (α) (value : Expr) (checkType : Expr → MetaM Unit) (safety := DefinitionSafety.safe) : MetaM α :=
   withoutModifyingEnv do
     let name ← mkFreshUserName `_tmp
     let value ← instantiateMVars value
+    let us := collectLevelParams {} value |>.params
```
eric-wieser, line 15: `elabEvalUnsafe` uses

https://github.com/leanprover/lean4/blob/8335d4e27d7a44ee13ad46d9aaa33c78b1984183/src/Lean/Elab/BuiltinCommand.lean#L634-L637

which rather puzzlingly uses `instantiateMVars` twice. I don't really know which if either is right, so I've limited this to the minimal change that fixes my bug.

eric-wieser:
> It would be good to have a Mathlib build for this one. Could you rebase onto `nightly-with-mathlib`?

I did this rebase @semorrison, but I guess CI won't run until tomorrow?

nomeata:
> > It would be good to have a Mathlib build for this one. Could you rebase onto `nightly-with-mathlib`?
> 
> I did this rebase @semorrison, but I guess CI won't run until tomorrow?

The workflow is a bit racy; by the time it ran, a new nightly tag appeared (`2023-12-19`), but `nightly-with-mathlib` did not advance. I hope that #3097  will fix that. Once that hits master, I can re-trigger the workflow and then it hopefully works. Sorry for the delays.

nomeata:
Seems to work (https://github.com/leanprover/lean4/actions/runs/7274620683/job/19820906002), mathlib CI should be running

src/Lean/Meta/Eval.lean
eric-wieser: @kmill, do you have any guesses as to why `inferType` happens to the pre-`instantiateMVars`d value in the other file referenced above?

eric-wieser: @digama0, @semorrison, do you have any thoughts on what's going on here?

kmill: I don't see a reason to `instantiateMVars` that second time, but I do wonder if it should be `let type ← instantiateMVars type` instead, since there's no reason why the inferred type will be metavariable-free.

eric-wieser: How can a value have no metavariables but its inferred type have some?

digama0: If the value is `a` where `a : ?m1` is a local constant then this can happen.

eric-wieser: But `eval` can't work if there are free variables anyway, so presumably this doesn't matter?

eric-wieser:
@semorrison, do you think this is ok as is? While there's [some discussion above](https://github.com/leanprover/lean4/pull/3090#discussion_r1485204250) about instantiating metavariables, it's mainly about code that isn't part of this patch anyway; and the actual change is only two lines!

eric-wieser:
Just to note this now also impacts `%eval` in mathlib:
```lean
import Mathlib.Tactic.Eval

def foo.{u} (x : PUnit.{u}) : Nat := 37

-- fails
def bar.{u} (x : PUnit.{u}) : Nat := eval% foo.{u} PUnit.unit
```

PR #3095, opened 2023-12-20
chore: revert #3020, unintentional change to Mathlib CI behaviour

kim-em:
Apologies, I should have caught this change in @nomeata's #3020 during review.

That PR unintentionally changed the behaviour of Mathlib CI.

It's essential that when we set up a branch on Mathlib for testing a Lean PR, that we
* compute the most recent nightly tag that has been merged into the PR
* start the Mathlib branch based off the corresponding `nightly-testing-YYYY-MM-DD` branch at Mathlib.

#3020 changed the behaviour so that it always just took the most recent nightly tag (without checking whether it had been merged into the PR).

#3020 also did more: it removed the need for the `pr-release.yml` task to actually check out the branch. However as I don't see how to easily correctly compute the relevant nightly tag without a local checkout (the problem is that the nightly tags are only on the `lean4-nightly` repo, but we need to see which ones have been merged into the PR branch!), I'm just reverting #3020 here wholesale, and we can try again later.

### DIFF
--- a/.github/workflows/pr-release.yml
+++ b/.github/workflows/pr-release.yml
@@ -169,9 +169,9 @@ jobs:
             else
               echo "The message already exists in the comment body."
             fi
-            echo "::set-output name=mathlib_ready::false"
+            echo "mathlib_ready=false" >> $GITHUB_OUTPUT
           else
-            echo "::set-output name=mathlib_ready::true"
+            echo "mathlib_ready=true" >> $GITHUB_OUTPUT
           fi
 
       # We next automatically create a Mathlib branch using this toolchain.


nomeata:
> However as I don't see how to easily correctly compute the relevant nightly tag without a local checkout (the problem is that the nightly tags are only on the lean4-nightly repo, but we need to see which ones have been merged into the PR branch!),

Well, we have a local checkout still. It's a bare one for extra security, but it should still be possible to fetch the nightly tags into it and make git operations on them.

Let me give it a try today. I have been meaning to do that, but pushing it off.

nomeata:
Alternatve PR at https://github.com/leanprover/lean4/pull/3097

Issue #3081, opened 2023-12-16
RFC: termination_by syntax: less wiggle-room with variables


nomeata:
### Status quo

Currently (assuming #2921), the `termination_by` syntax expect a list, possibly empty, of identifiers (or `_`), then `=>`, and then the body.

The semantics is that the names in the list override any variable names _from the back_. Examples:

```
def foo (n : Nat) : Nat := foo (n-1)
termination_by n => n
```
works but equivalently
```
def foo (n : Nat) : Nat := foo (n-1)
termination_by => n
```
or
```
def foo (n : Nat) : Nat := foo (n-1)
termination_by m => m
```

This is a bit silly: Why would the user _not_ want to use the variable name given in the header? Is there ever a compelling reason?

Of course, the list of variables do have their justification: When the function arguments are not in the header, but bound by a lambda or by match equations:
```
def foo : Nat → Nat := fun n => foo (n-1)
termination_by n => n
```
or
```
def foo : Nat → Nat
 | 0 => 0
 | n => foo (n-1)
termination_by n => n
```

If there are variables both in the header and in the lambda/matches, the list is applied from the right:
```
def foo (a b : Nat) : Nat → Nat → Nat
 | 0, 0 => 0
 | c, d => foo (a-1) (b - 1) (c - 1) (d-1)
termination_by _ x y z => x -- termination goal: b - 1 < b
```
and you can wildly mix overriding variable names and referring to the original name using `_`:
```
termination_by x _  _ y => x + b + y -- termination goal: (a - 1) + (b - 1) + (d - 1) < a + b + d
```

### Oddities

This works, and knowing the implementation it’s not surprised it behaves like this, but it is a bit odd:

* When no variable needs to be introduced, the arrow in `termination_by => n` looks odd.
* The right-to-left rule is probably not intuitive.
* Usually it is easy to predict how many parameters “after the `:`” lean considers, but maybe not always.
* I find hard to justify why the user even _can_ change `a` to `x` here.

### Proposed change

I wonder if our users are not better served with a more rigid syntax:

* The number of variables mentioned _must_ match the number of “unnamed” variables  (lambda after the `:=` or discriminants in matches equations). No fewer and not more.
* If there are no such variables, the `=>` must be omitted.

So above the valid variants would be:
```
def foo (n : Nat) : Nat := foo (n-1)
termination_by n

def foo : Nat → Nat := fun n => foo (n-1)
termination_by n => n

def foo : Nat → Nat
 | 0 => 0
[...]

PR #3099, opened 2023-12-20
feat: strict termination_by variables

nomeata:
Towards #3081


src/Lean/Elab/PreDefinition/WF/TerminationHint.lean
```diff
@@ -41,9 +44,18 @@ structure TerminationHints where
   ref : Syntax
   termination_by? : Option TerminationBy
   decreasing_by?  : Option DecreasingBy
+  /-- Here we record the number of paramters past the `:`. This is currently
```
digama0, line 47:
```suggestion
  /-- Here we record the number of parameters past the `:`. This is currently
```

PR #3102, opened 2023-12-20
fix: remove `Option.lt` and related components

### LINKED ISSUE #3100: `LT` instance for Option is unused/not useful to have at this kind of base level.
### Prerequisites

* [X ] Put an X between the brackets on this line if you have done all of the following:

### Description

The file Init/Data/Option/Basic.lean contains a function `Option.lt`, an instance of LT for `Option A` inherited from a `LT A`, and a decidability instance for that relation.

The definition that it uses is one which sets `none < some a` and `some a < some b` just when `a < b`. This is a perfectly sane definition, but making this choice at such a low-level seems counter-intuitive to me. Different libraries might want to choose different definitions for LT on Option. Back in Mathlib 4 land, `WitBot A` and `WithTop A` are synonyms for `Option` that are based off this ordering and the obvious dual one. The base `Option` type having a LT instance as a core feature of `lean4` - when as far as I can see this isn't used anywhere - removes the ability of downstream projects like `std4` or `mathlib4` to make different choices.

### Context

I was doing some work in Mathlib4 and I used `<` in some statements about Option, It worked (because the instance is defined), and some stuff is RFL-true and so it wasn't until substantially later that I realised that this instance wasn't particular doing what I had expected and didn't have any nice API. I discussed this in a Zulip thread here: https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/Preorder.20on.20Option

I think downstream in Mathlib4 (or maybe std4 but I am not as sure how to contribute to that) we should add an instance, but in the context of this project I think it should simply be removed - it isn't useful and it creates footguns like what happened to me.

### Steps to Reproduce

1. Write down the following code:
``` 
example {A : Type*} [LT A] {a : A} : (none : Option A) < some a := sorry` 
```
2. The code is trivially true (I think `trivial` works as a proof through I am not sure.) 
3. And this typechecks.

**Expected behavior:**
I would expect the default behaviour of LT on Option to be either:
a) LT is not defined, and so this shouldn't typecheck/synthesise a correct instance.
b) An instance is defined, and in it, this is not a true statement. Why should `none < some a` be true, and `some a < none` not? You don't have to make a choice.

**Actual behavior:** 
A choice is made: this statement is true.

### Versions

4.30
Windows 11 running Ubuntu WSL2

### Additional Information

None.

### Impact

linesthatinterlace:
---
This is a draft PR implementing the minimal changes required to fix #3100.

### DIFF
--- a/src/Init/Data/Option/Basic.lean
+++ b/src/Init/Data/Option/Basic.lean
@@ -62,17 +62,6 @@ theorem map_id : (Option.map id : Option α → Option α) = id :=
 instance : OrElse (Option α) where
   orElse := Option.orElse
 
-@[inline] protected def lt (r : α → α → Prop) : Option α → Option α → Prop
-  | none, some _     => True
-  | some x,   some y => r x y
-  | _, _             => False
-
-instance (r : α → α → Prop) [s : DecidableRel r] : DecidableRel (Option.lt r)
-  | none,   some _ => isTrue  trivial
-  | some x, some y => s x y
-  | some _, none   => isFalse not_false
-  | none,   none   => isFalse not_false
-
 /-- Take a pair of options and if they are both `some`, apply the given fn to produce an output.
 Otherwise act like `orElse`. -/
 def merge (fn : α → α → α) : Option α → Option α → Option α
@@ -85,9 +74,6 @@ end Option
 
 deriving instance DecidableEq for Option
 deriving instance BEq for Option
-
-instance [LT α] : LT (Option α) where
-  lt := Option.lt (· < ·)
 
 @[always_inline]
 instance : Functor Option where


semorrison (approved):
I agree this is good.

Choices like this, where all choices (`none` at the bottom, top, or incomparable) should be made via instances on synonyms. Mathlib already has `WithTop` and `WithBot` as synonyms for `Option`.

linesthatinterlace:
I'm personally in favour of Mathlib having the incomparable order as the default on Option - in some ways it's going to be quite useful to have something on it - but that's a discussion to have elsewhere. Point is it's a downstream choice anyway.

nomeata:
Don't you need at least _some_ order on Option if you need to put data types with option in it into, say, a RBTree or some other ordered data structure?

linesthatinterlace:
Yes, precisely so. And the incomparable order is the most neutral of the "obvious" choices I think.

(I don't know if it should be in std4 or mathlib4. Clearly any order hierarchy stuff should be in the latter but I can see computer science applications needing the basic order on Option defined for the reasons you mention.)

linesthatinterlace:
Well, for the function you've given there, it seems like it "makes sense" if any input that involves none outputs false. It probably then depends how qsort is written.

linesthatinterlace:
(If it is written assuming totality, it should really have that as a hypothesis in some form, as a typeclass or something else.)

nomeata:
You have high expectations for how programming APIs are structured :-D

Most ordering based data structures will break of the passed `lt` isn't a linear order, without telling you.

It may make sense to have different type classes for programming (total, canonical  instances for common data type) and mathematics?

Btw the `none < some _` order on Option is very common in programming, like `false < true` and `[] < _ :: _`, so less arbitrary than it maybe looks like from the math point of view.

linesthatinterlace:
I should say that I'm as much as programmer as a mathematician. But yes, it makes some sense to make that choice - most implementations of sorting algorithms seem to basically assume totality and that's as good as anything and not without precedent. Though to me false < true is more obvious somehow (because in some sense it's 0 < 1).

nomeata:
Right, and from `false < true` it's a small step to “empty before filled” order for container types like option and list, if one associates “false” with “no value” or “value”.

So it's rather defensible to have this order, the question is more how to make sure it doesn't get in the way when you want to do math. (but wouldn't you then use `WithBot` and `WithTop`?)

linesthatinterlace:
I certainly feel that `WithTop < Incomparable` and `WithTop < WithBot` - in a sense I personally don't care about the difference between `Incomparable` and `WithBot` too much. BUT: not all the API for Option exists on With*. So the other day when I tried to switch them out it broke more things than I was hoping.

nomeata:
> not all the API for Option exists on With*.

It kinda does, because [WithTop](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Order/WithBot.html#WithTop) α [=](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#Eq) [Option](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#Option) α, but some of the convenience features (dot notation, type class inference) probably doesn’t work out of the box.

But that’s more a reason to improve the With* APIs?

linesthatinterlace:
I don't disagree with that! I was surprised that it didn't Just Work TM tbh.

PR #3105, opened 2023-12-21
chore: remove workaround in widgets

Vtec234:
This is a follow-up on #2964 that ~~updates stage0,~~ removes a workaround ~~, and updates release notes.~~

### DIFF
--- a/src/Lean/Widget/UserWidget.lean
+++ b/src/Lean/Widget/UserWidget.lean
@@ -7,19 +7,6 @@ Authors: E.W.Ayers, Wojciech Nawrocki
 import Lean.Elab.Eval
 import Lean.Server.Rpc.RequestHandling
 
-namespace Lean.Server
-open Elab Command in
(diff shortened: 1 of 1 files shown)


nomeata:
These days we can let GitHub update stage0, which I find a bit cleaner. Should we trigger a stage0 update on master and only the workaround needs to be PRed then, or is that for some reason not desirable here?

Vtec234:
@nomeata it would be desirable, I just couldn't figure out how to do it! It looked like lean4#3042 only runs if `stdlib_flags.h` has changed which is not the case here.

nomeata:
It can only be done by people with write permissions. Started at https://github.com/leanprover/lean4/actions/runs/7287788936, let’s see if this works (this is rather new)

semorrison:
We will merge this, but I'll also cherry-pick the release notes onto the `release/v4.5.0` branch when I start that tomorrow.

nomeata:
@Vtec234, done, you can merge latest master into this.

PR #3106, opened 2023-12-21
feat: new snapshot architecture on the cmdline

Kha:


abentkamp:
There are a couple of restrictions of emscripten pthreads listed here,
which might cause this:

https://emscripten.org/docs/porting/pthreads.html

Unfortunately, I dont think I will have the time to look into this in the
near future.

Sebastian Ullrich ***@***.***> schrieb am Sa., 23. Dez. 2023,
5:06 PM:

Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
!bench

src/Lean/Elab/Frontend.lean
```diff
@@ -81,23 +81,17 @@ end Frontend
 
 open Frontend
```
Kha, line 82: `processCommand` and the whole `FrontendM` are not used by Lean itself anymore now but I think it's fine to leave them around as they have some external uses. They just will not benefit from parallelism etc. going forward.

PR #3110, opened 2023-12-22
refactor: post-stage0 fixes to #3040

nomeata:
This pull request is only opened to run CI on stuff I may do once #3040 is
merged and stage0 updated.

### DIFF
--- a/src/lake/Lake/DSL/Package.lean
+++ b/src/lake/Lake/DSL/Package.lean
@@ -73,4 +73,4 @@ macro_rules
   let attr ← withRef kw `(Term.attrInstance| «post_update»)
   let attrs := #[attr] ++ expandAttrs attrs?
   `($[$doc?]? @[$attrs,*] def postUpdateHook : PostUpdateHookDecl :=
-    {pkg := $pkgName, fn := fun $pkg => $defn} $[$wds?]?)
+    {pkg := $pkgName, fn := fun $pkg => $defn} $[$wds?:whereDecls]?)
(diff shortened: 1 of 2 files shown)


digama0:
Normally I do this kind of thing using a single PR which has a `update stage0` commit in the middle, for example #2247 is structured like this.

nomeata:
I'm creating this mostly to test what would happen. Maybe we should enable CI for branches to not have to create PRa just to run CI.

(Additionally I think it's preferable to use the automatic stage0 update on master, no need to worry if the stage0 code was created cleanly.)

digama0:
The point is not to use the update stage0 commit when it is actually merged, in fact it usually has to be recreated on each rebase. The point is so that the tests pass and the PR is all in one place; handling dependent PRs is more complicated, maintenance-wise, than just replacing the update-stage0 commit when bumping such a PR.

Kha:
Mario, the point is that in this case the stage 0 update is *not* necessary as part of the original PR. I agree with Joachim that a follow-up PR is easier in this case.

nomeata:
In any case I should be better at marking PRs (besides marking them as Draft) that are just me trying something on the side, which may or may not turn in to real PRs :-)

I was hoping I can proceed by merging #3040, then automatic stage0 update on master, then clean-up-PR (like with #2964 and then #3105), but that may not be possible in this case, where updating stage0 will break CI until broken library code is updated.

digama0:
I think a PR which fails the stage2 build should not be merged, any stage0 update and other post-stage0 fixes should be done on the same PR. If that's not the situation here, then feel free to ignore, but it sounds like #3040 still has issues of this form.

nomeata:
It probably has, which I am just discovering. I agree that in that case we have to go back to  the multiple-rebased-commits-including-stage0-on-one-PR workflow.

PR #3113, opened 2023-12-23
feat: optimize computing `BinderInfo`s for discrimination trees

kmill:
When computing paths for discrimination trees, currently it uses `getFunInfoNArgs` to calculate binder infos. This does many other calculations as well, and it also doesn't handle overapplied functions. Switches to a straightforward calculation that accumulates `BinderInfo`s.

The effect of this is a reduction of 0.05% instructions when building mathlib. **This is not worth reviewing for now.**


kmill:
!bench

PR #3115, opened 2023-12-24
feat: make `intro` be aware of `let_fun`

kmill:
Adds support for `let_fun` to the `intro` and `intros` tactics. Also adds support to `intro` for anonymous binder names, since the default variable name for a `letFun` with an eta reduced body is anonymous.

### DIFF
--- a/src/Lean/Meta/Tactic/Intro.lean
+++ b/src/Lean/Meta/Tactic/Intro.lean
@@ -109,7 +109,8 @@ private def mkAuxNameImp (preserveBinderNames : Bool) (hygienic : Bool) (useName
 where
   mkAuxNameWithoutGivenName (rest : List Name) : MetaM (Name × List Name) := do
     -- Use a nicer binder name than `[anonymous]`, which can appear in for example `letFun x f` when `f` is not a lambda expression.
-    let binderName := if binderName.isAnonymous then `a else binderName
+    -- In this case, we make sure the name is hygienic.
+    let binderName ← if binderName.isAnonymous then mkFreshUserName `a else pure binderName
     if preserveBinderNames then
       return (binderName, rest)
     else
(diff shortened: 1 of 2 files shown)


src/Lean/Expr.lean
```diff
@@ -1668,14 +1668,14 @@ Warning: if the `let_fun` is applied to additional arguments (such as in `(let_f
 `let_fun` expressions are encoded as `letFun v (fun (n : t) => b)`.
 They can be created using `Lean.Meta.mkLetFun`.
 
-If in the encoding of `let_fun` the last argument to `letFun` is eta reduced, this returns `Name.anonymous` for the binder name.
+If in the encoding of `let_fun` the last argument to `letFun` is eta reduced, this returns `a` for the binder name.
```
Kha, line 1671: Would it be cleaner to do this defaulting in the tactic, not the "raw" accessor?

Kha (approved):

src/Lean/Expr.lean
kmill: I switched to that, since it feels cleaner, and I'm not sure what I was worried about was thought out. The general issue is that maybe a user has an eta-reduced `let_fun` and then some tactic somewhere creates a local context without processing the `[anonymous]` name first. But, so long as the tactic author reads the documentation or instead relies on `intro`, then this should be OK.

All the uses of `letFun?` and `letFunAppArgs?` that could result in user-facing terms end up doing `instantiate1` anyway. The exception is ToLCNF, which converts a `let_fun` to a `let` and trusts the binding name.

Kha: Hm, now that I look at the test, should the name even be accessible in this case? It is an edge case, so perhaps it does not matter too much.

kmill: I suppose it would make sense for intro1p to give an inaccessible name here. It doesn't even pretty print in the eta reduced case, so I don't think anyone would be surprised.

intro1p doesn't strip macro scopes off binders, which I think is reason enough

kmill: I just pushed a version that makes the name hygienic, and I changed the test to just make sure `intro` works, vs checking what the specific name is (since that doesn't make sense now that it's hygienic).

PR #3118, opened 2023-12-27
perf: simplifier performance issues on `match` and `if-then-else` terms

leodemoura:


leodemoura:
!bench

leodemoura:
!bench

leodemoura:
Subsumed by https://github.com/leanprover/lean4/pull/3124

PR #3120, opened 2023-12-27
refactor: simplify `SimpM` recursive definitions

leodemoura:
Depends on https://github.com/leanprover/lean4/pull/3118.
It breaks `Std`, but it is an easy fix.

### DIFF
--- a/src/Lean/Expr.lean
+++ b/src/Lean/Expr.lean
@@ -959,16 +959,16 @@ where
   loop : Nat → Expr → Array Expr → Array Expr
     | 0,   _,        as => as
     | i+1, .app f a, as => loop i f (as.set! i a)
-    | _,   _,        as => as
+    | _,   _,        _  => panic! "too few arguments at"
 
 /--
 Given `e` of the form `f a_1 ... a_n`, return `f`.
 If `n` is greater than the number of arguments, then return `e.getAppFn`.
 -/
-def extractNumArgs (e : Expr) (n : Nat) : Expr :=
+def stripArgsN (e : Expr) (n : Nat) : Expr :=
   match n, e with
   | 0,   _        => e
-  | n+1, .app f _ => extractNumArgs f n
+  | n+1, .app f _ => stripArgsN f n
   | _,   _        => e
 
 /--
@@ -976,7 +976,7 @@ Given `e` of the form `f a_1 ... a_n ... a_m`, return `f a_1 ... a_n`.
 If `n` is greater than the arity, then return `e`.
 -/
 def getAppPrefix (e : Expr) (n : Nat) : Expr :=
-  e.extractNumArgs (e.getAppNumArgs - n)
+  e.stripArgsN (e.getAppNumArgs - n)
 
 /-- Given `e = fn a₁ ... aₙ`, runs `f` on `fn` and each of the arguments `aᵢ` and
 makes a new function application with the results. -/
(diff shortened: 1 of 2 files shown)


src/Lean/Expr.lean
```diff
@@ -947,6 +947,37 @@ private def getAppRevArgsAux : Expr → Array Expr → Array Expr
...
+The resulting array has size `n` even if `f.getAppNumArgs < n`.
+-/
+@[inline] def getAppArgsN (e : Expr) (n : Nat) : Array Expr :=
+  let dummy := mkSort levelZero
+  loop n e (mkArray n dummy)
+where
+  loop : Nat → Expr → Array Expr → Array Expr
+    | 0,   _,        as => as
+    | i+1, .app f a, as => loop i f (as.set! i a)
+    | _,   _,        as => as
```
nomeata, line 962: This will return an array with dummy entries if there aren't enough arguments, will it? Maybe worth pointing out in the docs, also on which side the dummies are? Or maybe just panic instead, its probably misuse if `n` is too small?

src/Lean/Expr.lean
```diff
@@ -947,6 +947,37 @@ private def getAppRevArgsAux : Expr → Array Expr → Array Expr
...
+
+/--
+Given `e` of the form `f a_1 ... a_n`, return `f`.
+If `n` is greater than the number of arguments, then return `e.getAppFn`.
+-/
+def extractNumArgs (e : Expr) (n : Nat) : Expr :=
+  match n, e with
+  | 0,   _        => e
+  | n+1, .app f _ => extractNumArgs f n
+  | _,   _        => e
```
nomeata, line 972: I'm a bit confused by the naming; this doesn't extract the args, but extracts the function, or strips the args, doesn't it?

src/Lean/Expr.lean
leodemoura: Fixed.

src/Lean/Expr.lean
leodemoura: Fixed.

src/Init/SimpLemmas.lean
```diff
@@ -84,6 +84,10 @@ theorem dite_congr {_ : Decidable b} [Decidable c]
 @[simp] theorem ite_false (a b : α) : (if False then a else b) = b := rfl
 @[simp] theorem dite_true {α : Sort u} {t : True → α} {e : ¬ True → α} : (dite True t e) = t True.intro := rfl
 @[simp] theorem dite_false {α : Sort u} {t : False → α} {e : ¬ False → α} : (dite False t e) = e not_false := rfl
+@[simp ↓] theorem ite_cond_true {_ : Decidable c} (a b : α) (h : c) : (if c then a else b) = a := by simp [h]
+@[simp ↓] theorem ite_cond_false {_ : Decidable c} (a b : α) (h : ¬ c) : (if c then a else b) = b := by simp [h]
+@[simp ↓] theorem dite_cond_true {α : Sort u} {_ : Decidable c} {t : c → α} {e : ¬ c → α} (h : c) : (dite c t e) = t h := by simp [h]
+@[simp ↓] theorem dite_cond_false {α : Sort u} {_ : Decidable c} {t : c → α} {e : ¬ c → α} (h : ¬ c) : (dite c t e) = e h := by simp [h]
```
eric-wieser, line 90: What's the intended difference between these lemmas and `if_pos`/`if_neg`/`dif_pos`/`dif_neg`?

leodemoura: See comment at https://github.com/leanprover/lean4/pull/3120/commits/5b46fde02ec6b1bb560bde7b18fd251c4076d42b

leodemoura:
Subsumed by https://github.com/leanprover/lean4/pull/3124

PR #3121, opened 2023-12-28
refactor: rewrite: produce simpler proof terms

nomeata:
Consider
```
import Std.Tactic.ShowTerm

opaque a : Nat
opaque b : Nat
axiom a_eq_b : a = b
opaque P : Nat → Prop

set_option pp.explicit true

-- Using rw
example (h : P b) : P a := by show_term rw [a_eq_b]; assumption
```

Before, a typical proof term for `rewrite` looked like this:
```
-- Using the proof term that rw produces
example (h : P b) : P a :=
  @Eq.mpr (P a) (P b)
  (@id (@Eq Prop (P a) (P b))
    (@Eq.ndrec Nat a (fun _a => @Eq Prop (P a) (P _a))
      (@Eq.refl Prop (P a)) b a_eq_b))
  h
```
which is rather round-about, applying `ndrec` to `refl`. It would be more direct to write
```
example (h : P b) : P a :=
  @Eq.mpr (P a) (P b)
  (@id (@Eq Prop (P a) (P b))
    (@congrArg Nat Prop a b (fun _a => (P _a)) a_eq_b))
  h
```
which this change does.

This makes proof terms smaller, causing mild general speed up throughout the code; if the brenchmarks don’t lie the highlights are

* olean size -2.034 %
* lint wall-clock -3.401 %
* buildtactic execution s -10.462 %

H'T to @digama0 for advice and help.

NB: One might even expect the even simpler
```
-- Using the proof term that I would have expected
example (h : P b) : P a :=
  @Eq.ndrec Nat b (fun _a => P _a) h a a_eq_b.symm
```
but that would require non-local changes to the source code, so one step at a time.

### DIFF
--- a/tests/lean/motiveNotTypeCorect.lean
+++ b/tests/lean/motiveNotTypeCorect.lean
@@ -9,9 +9,11 @@ def mwe (t : Nat) (f : Nat -> Nat) (h : f t = t)
 
 def K (x : α) (y : β) : β := y
 example (h : true = false) (A : (b : Bool) → K b Type) (h2 : A false) : A true := by
-  rw [h] -- tactic 'rewrite' failed, motive is dependent
+  -- the motive is dependent on `true`, but in a non-essential way, so this works fine
+  rw [h]
   exact h2
 
-example (h : true = false) (A : (b : Bool) → if b then Prop else Nat) : A true := by
-  rw [h] -- tactic 'rewrite' failed, motive is dependent
-  exact 0
+example (h : true = false) (A : (b : Bool) → if b then Prop else Nat) : A true :=
+  by
+    rw [h] -- tactic 'rewrite' failed, motive is dependent
+    exact 0
--- a/tests/lean/motiveNotTypeCorect.lean.expected.out
+++ b/tests/lean/motiveNotTypeCorect.lean.expected.out
@@ -5,12 +5,7 @@ h : f t = t
 d : D (f t)
 P : (t : Nat) → D t → Prop
 ⊢ P (f t) d
-motiveNotTypeCorect.lean:12:6-12:7: error: tactic 'rewrite' failed, motive is dependent
-h : true = false
-A : (b : Bool) → K b Type
-h2 : A false
-⊢ A true
-motiveNotTypeCorect.lean:16:6-16:7: error: tactic 'rewrite' failed, motive is dependent
+motiveNotTypeCorect.lean:18:8-18:9: error: tactic 'rewrite' failed, motive is dependent
 h : true = false
 A : (b : Bool) → if b = true then Prop else Nat
 ⊢ A true


nomeata:
!bench

nomeata:
The test suit succeeds but mathlib fails. I hope I can distill a test case from this.

nomeata:
Hmm, the three breaking cases were easy to fix (https://github.com/leanprover-community/mathlib4/commit/11e0f80d35c7fea610637258c856d301fe362904):

* two have subgoals in a different order, but it’s not clear to me that the original order was better. Also unclear why this happens. Maybe an `instantiateMVars` that wasn’t there before?
* In one case there was a porting note
  ```
      /- porting note: the following is indented two spaces more than it should be due to
        strange behavior of `erw` -/
      all_goals …
  ```
  and indeed the fix was to un-indent the `all_goals`, as one would expect. Did I inadvertently fix a minor bug here? But still a bit mysterious to me why this would happen with this change.

nomeata:
Note to myself:

Using `mkCongrArg` reveals that the way I construct the motive here is not always non-dependent:
```
./././Mathlib/CategoryTheory/Limits/IsLimit.lean:538:20: error: AppBuilder for 'congrArg', non-dependent function expected
  fun _a ↦ { pt := pt✝, π := _a }.π.app j
has type
  (_a : (Functor.cones F).obj (op pt✝)) → ((const J).obj { pt := pt✝, π := _a }.pt).obj j ⟶ F.obj j
```
(but it is after a bit more evaluation; maybe `mkCongrArg` is a bit too picky here.)

semorrison:
> Then mathlib builds, but it’s test suite fails, e.g. at
> 
> ```
> -- Discharge side conditions from local hypotheses.
> /--
> info: Try this: rw [h p]
> -- "no goals"
> -/
> #guard_msgs in
> example {P : Prop} (p : P) (h : P → 1 = 2) : 2 = 1 := by
>   rw?
> ```
> 
> so there is something fishy going on with side-conditions.

@nomeata, the "something fishy" is the `rw?` tactic inspecting the proof term in order to reconstruct the arguments of the successful rewrite lemma: https://github.com/leanprover-community/mathlib4/blob/master/Mathlib/Tactic/Rewrites.lean#L36.

This is obviously evil, but I'm not sure of a better way to do it. In any case it should be easy to update that function to match your new scheme. Let me know if you'd like me to do something in that file.

nomeata:
Ha, should have thought of that!

I can do it myself when I come back to this (non-pressing) experiment, now that I know where to look.

Should `rw?` at least warn if it `by?` doesn’t work (“Could rewrite with `lem`, but failed to extract arguments, please report a bug”) , or is that an expected error mode?

src/Lean/Meta/Tactic/Rewrite.lean
```diff
@@ -40,13 +40,13 @@ def _root_.Lean.MVarId.rewrite (mvarId : MVarId) (e : Expr) (heq : Expr)
...
-          let motive := Lean.mkLambda `_a BinderInfo.default α eEqEAbst
+          let eType ← inferType e
+          let motive := Lean.mkLambda `_a BinderInfo.default α eAbst
           unless (← isTypeCorrect motive) do
             throwTacticEx `rewrite mvarId "motive is not type correct"
-          let eqRefl ← mkEqRefl e
-          let eqPrf ← mkEqNDRec motive eqRefl heq
+          let u1 ← getLevel α
+          let u2 ← getLevel eType
+          let eqPrf ← instantiateMVars (mkApp6 (.const ``congrArg [u1, u2]) α eType lhs rhs motive heq)
```
nomeata, line 49:
```suggestion
          let eqPrf := mkApp6 (.const ``congrArg [u1, u2]) α eType lhs rhs motive heq
```
Let’s see if this is needed, and maybe the cause for the reordering of side-condition goals.

nomeata:
Mathlib is green now. I am still surprised that this refactor causes side goals to appear in a different order: <https://github.com/leanprover-community/mathlib4/commit/af6cccb9252c5a850070d6a9ec68d39e31870f02>

semorrison:
Could you benchmark the Mathlib build? Perhaps it's too difficult to get a baseline, at the moment.

nomeata:
I see lots of green! I just hope I did it right…
The report at
http://speed.lean-fro.org/mathlib4/compare/7e8c981f-5cdb-4101-b993-d94a28a1072b/to/e75bdb5f-fa8a-46ac-a48b-a0cbf8da9cc6
should correspond to
> git diff 8e0fec81ddb8a55c95e7aea124d16684684ff54d..fee326ae5794fb3c55b3fc99fb582513f92e1848

nomeata:
Here another benchmark run:
http://speed.lean-fro.org/mathlib4/compare/473da4ea-9efd-484f-a676-b52133e4ad35/to/77c1316a-877a-44d1-abf7-12b72a0e9306

* olean size -2.18 %
* lint wall-clock -1.144 % %
* build tactic execution s -11.514 % %
* build wall-clock -0.745 % 
* build instructions -1.714 % 

So the gains look relatively stable.

semorrison (approved):

PR #3124, opened 2023-12-28
feat: add `Simproc`s

leodemoura:
Depends on #3123 and #3120

---

Downstream changes required:
* https://github.com/leanprover/std4/pull/496
* https://github.com/leanprover-community/aesop/pull/93
* https://github.com/leanprover-community/mathlib4/pull/9500

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -11,6 +11,61 @@ of each version.
 v4.6.0 (development in progress)
 ---------
 
+* Add custom simplification procedures (aka `simproc`s) to `simp`. Simprocs can be triggered by the simplifier on a specified term-pattern. Here is an small example:
+```lean
+import Lean.Meta.Tactic.Simp.BuiltinSimprocs.Nat
+
+def foo (x : Nat) : Nat :=
+  x + 10
+
+/--
+The `simproc` `reduceFoo` is invoked on terms that match the pattern `foo _`.
+-/
+simproc reduceFoo (foo _) :=
+  /- A term of type `Expr → SimpM (Option Step) -/
+  fun e => OptionT.run do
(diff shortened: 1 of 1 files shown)


tests/lean/run/simproc1.lean
nomeata, line 1: The test suggests that users can rely on the pattern to be matched, and the simpproc doesn't have to check again that the expression really is an application of `foo`. Is that intentional? The docstring for simpproc is not very clear on that point. 

I wonder if it could be that the DiscrTree implementation may take shortcuts and match more expressions than the user expect.

In any case, the docstring for simpproc should maybe be precise here.

leodemoura: Added a new docstring.

src/Init/Data/Fin/Basic.lean
```diff
@@ -100,7 +100,7 @@ instance : ShiftLeft (Fin n) where
 instance : ShiftRight (Fin n) where
   shiftRight := Fin.shiftRight
 
-instance : OfNat (Fin (no_index (n+1))) i where
+instance ofNatInst : OfNat (Fin (no_index (n+1))) i where
```
eric-wieser, line 103: In mathlib, we have been using the convention of naming instances like this `instOfNat` as opposed to `ofNatInst`, which I believe is also consistent with the automatic naming rules. Is this departure from the convention an indication that the automatic naming rules are going to change soon?

leodemoura: Changed to `instOfNat`.

leodemoura:
!bench

semorrison:
Added `full-ci` label so we get a toolchain for all OSes.

PR #3127, opened 2023-12-31
feat: make constructor tactic fail in ambiguous cases

hrmacbeth:
---

~~I am opening a PR to get CI to run.  I will write an RFC later, informed by the results of CI.   I hope this is ok.~~

This is a quick implementation of the RFC #3129 so that its impact on Std/Mathlib can be assessed.

### DIFF
--- a/src/Lean/Meta/Tactic/Constructor.lean
+++ b/src/Lean/Meta/Tactic/Constructor.lean
@@ -11,7 +11,8 @@ namespace Lean.Meta
 
 /--
 When the goal `mvarId` type is an inductive datatype,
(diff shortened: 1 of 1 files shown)


hrmacbeth:
WIP

nomeata:
It's ok, especially if clearly explained like this. Thanks for investigating this!

nomeata:
FYI: If you want to test mathlib, the best bet is to branch off `nightly-with-mathlib`.

semorrison:
@hrmacbeth, if you (or anyone else) have the time to add some test cases exercising the new behaviour, I would be happy to proceed with this PR.

PR #3133, opened 2024-01-03
refactor: WF.Eqns: remove unreachable fix-folding

nomeata:
I was about to to address the TODO

     /- TODO: check arity of the given function. If it takes a PSigma as the last argument,
        this function will produce incorrect results. -/

because we now have an arity-observing variant of `decodePackedArg?` in
`unpackArg` in `PackMutual`, and it would be prudent to use it here.

But I first wanted to create a test case that would actually exhibit
this corner case, and failed.

This code was added in 096e4eb6d0f and it had a test case, but not even
that test case seems to be actually using the `decodePackedArg?`
function, neither back then nor now.

Also, mathlib works without this code.

So this seems to be dead code, possibly due to other changes to the
system, and thus can be removed. A strategically place comments points
back to this PR in case we need to resurrect that code.

### DIFF
--- a/src/Lean/Elab/PreDefinition/WF/Eqns.lean
+++ b/src/Lean/Elab/PreDefinition/WF/Eqns.lean
@@ -80,10 +80,8 @@ private partial def mkProof (declName : Name) (type : Expr) : MetaM Expr := do
     let rec go (mvarId : MVarId) : MetaM Unit := do
       trace[Elab.definition.wf.eqns] "step\n{MessageData.ofGoal mvarId}"
       if (← tryURefl mvarId) then
-        trace[Elab.definition.wf.eqns] "refl!"
         return ()
       else if (← tryContradiction mvarId) then
-        trace[Elab.definition.wf.eqns] "contra"
         return ()
       else if let some mvarId ← simpMatchWF? mvarId then
         go mvarId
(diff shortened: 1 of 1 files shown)


nomeata:
(I am not yet saying that we can and should remove this code, still investigating.)

nomeata:
> Maybe: Bisect lean to find out why this is no longer needed for
the original test case.

Heh, even on the original commit 
```diff
diff --git a/src/Lean/Elab/PreDefinition/WF/Eqns.lean b/src/Lean/Elab/PreDefinition/WF/Eqns.lean
index 38c00903a1..aa3f458a35 100644
--- a/src/Lean/Elab/PreDefinition/WF/Eqns.lean
+++ b/src/Lean/Elab/PreDefinition/WF/Eqns.lean
@@ -44,6 +44,7 @@ private def hasWellFoundedFix (e : Expr) : Bool :=
   Recall that we use `PSum` and `PSigma` for packing the arguments of mutually recursive nary functions.
 -/
 private partial def decodePackedArg? (info : EqnInfo) (e : Expr) : Option (Name × Array Expr) := OptionM.run do
+  panic! "decodePackedArg?!"
   if info.declNames.size == 1 then
     let args := decodePSigma e #[]
     return (info.declNames[0], args)
```
does not fail the `nestedWF` test case added there.

Issue #188, opened 2020-09-21
Enforcing kernel assumptions


leodemoura:
We have added a few kernel extensions that make assumptions about the `Nat` and `String` types. For example, the kernel assumes the term `Expr.lit (literal.natVal 0)` has type `Nat`. So, in principle, adversarial users may try to create their own `Core.lean` (using `prelude`) where `Nat` is, for example, the empty type. Right now, we do not have protections against this kind of abuse. This is a very low priority issue, but it is important to document it. 
There are many possible ways to fix this. For example, we can check in the Kernel whether `Nat` and `String` are the ones
we expect at `addDecl`.

PR #3134, opened 2024-01-03
feat: add bitwise operations to reduceNat? and kernel

joehendrix:
This adds bitwise operations to reduceNat? and the kernel.  It incorporates some basic test cases to validate the correct operations are associated.

### DIFF
--- a/tests/lean/run/lean_nat_bitwise.lean
+++ b/tests/lean/run/lean_nat_bitwise.lean
@@ -24,14 +24,14 @@ example : 0b1111 ^^^ 0      = 0b1111 := rfl
 example : 0b1111 ^^^ 0b1111 = 0      := rfl
 example : 0b1100 ^^^ 0b1010 = 0b0110 := rfl
 example : 0b0110 ^^^ 0b0101 = 0b0011 := rfl
-example : let x := 2^20; x ||| x = x := rfl
+example : let x := 2^20; x ^^^ x = 0 := rfl
 
 example : 0      >>> 0      = 0 := rfl
 example : 0b1011 >>> 1      = 0b0101 := rfl
 example : 0b1011 >>> 2      = 0b0010 := rfl
-example : 0      >>> 2^20 = 0 := rfl
+example : 0      >>> 2^20   = 0 := rfl
 
 example : 0      <<< 0      = 0 := rfl
 example : 0b1011 <<< 1      = 0b010110 := rfl
 example : 0b1011 <<< 2      = 0b101100 := rfl
-example : 0 <<< 2^20 = 0 := rfl
+example : 0      <<< 2^20   = 0 := rfl


joehendrix:
!bench

digama0:
I want to register my concern with adding more primitives to the kernel while https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.5BRFC.5D.20Checking.20primitive.20types.20in.20the.20kernel/near/408455066 is still unresolved. (Short version: Adding a primitive to the kernel is an assertion that it *definitionally* computes to concrete values on concrete values, which means that for it to be a conservative extension we need the definitions to have enough definitional equalities to justify this. Definitions by well founded recursion are particularly bad here because they do not compute definitionally and I would like to replace them with fuel based definitions or similar.)

joehendrix:
!bench

joehendrix:
I'm going to wait for #3139 to be resolved, and then rebase this on top of that.

I'll extend the test case there so we can validate the bitwise operations in `reduceNat?` let Std revert leanprover/std4#501.

semorrison:
> I want to register my concern with adding more primitives to the kernel while https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.5BRFC.5D.20Checking.20primitive.20types.20in.20the.20kernel/near/408455066 is still unresolved. (Short version: Adding a primitive to the kernel is an assertion that it _definitionally_ computes to concrete values on concrete values, which means that for it to be a conservative extension we need the definitions to have enough definitional equalities to justify this. Definitions by well founded recursion are particularly bad here because they do not compute definitionally and I would like to replace them with fuel based definitions or similar.)
> 
> I am willing to put in the leg work for this but I would like affirmation that this is a desirable direction first.

I'm not the one to decide anything here, but this sounds reasonable and desirable to me! Your short version here was much more helpful to me than the original discussion, in fact. Perhaps converting the short version here to an RFC is the right way to go?

joehendrix:
#3139 was just merged, so I will rebase this onto that, and augment with tests to verify `whnf` and kernel performance improvements.

I think the soundness goals in #188 are good ones, but low priority as it isn't doesn't affect users currently.   These additions don't substantially impact the difficulty of that feature, but the performance of these operations is more likely to affect users.

tests/lean/run/lean_nat_bitwise.lean
```diff
@@ -0,0 +1,37 @@
...
+example : 0b1100 ||| 0b1010 = 0b1110 := rfl
+example : let x := 2^20; x ||| x = x := rfl
+
+example : 0      ^^^ 0      = 0 := rfl
+example : 0      ^^^ 0b1111 = 0b1111 := rfl
+example : 0b1111 ^^^ 0      = 0b1111 := rfl
+example : 0b1111 ^^^ 0b1111 = 0      := rfl
+example : 0b1100 ^^^ 0b1010 = 0b0110 := rfl
+example : 0b0110 ^^^ 0b0101 = 0b0011 := rfl
+example : let x := 2^20; x ||| x = x := rfl
```
semorrison, line 27:
```suggestion
example : let x := 2^20; x ^^^ x = x := rfl
```

tests/lean/run/lean_nat_bitwise.lean
```diff
@@ -0,0 +1,37 @@
...
+
+example : 0      >>> 0      = 0 := rfl
+example : 0b1011 >>> 1      = 0b0101 := rfl
+example : 0b1011 >>> 2      = 0b0010 := rfl
+example : 0      >>> 2^20 = 0 := rfl
```
semorrison, line 32:
```suggestion
example : 0      >>> 2^20   = 0 := rfl
```

tests/lean/run/lean_nat_bitwise.lean
```diff
@@ -0,0 +1,37 @@
...
+
+example : 0      <<< 0      = 0 := rfl
+example : 0b1011 <<< 1      = 0b010110 := rfl
+example : 0b1011 <<< 2      = 0b101100 := rfl
+example : 0 <<< 2^20 = 0 := rfl
```
semorrison, line 37:
```suggestion
example : 0      <<< 2^20   = 0 := rfl
```

tests/lean/run/lean_nat_bitwise.lean
digama0: ???

joehendrix:
!bench

PR #3137, opened 2024-01-04
chore: pr-release.yaml: indicate information using github status

nomeata:
When looking at a PR I sometimes wonder which `nightly` release is this based on, and is used for the mathlib testing.

Right now, the action uses a label (`toolchain-available`) for this, but a label cannot easily carry more information.

It seems a rather simple way to communicate extra information is by setting [commit statuses](https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28#create-a-commit-status); with this change the following statuses will appear in the PR:

statusses

One could also use [checks](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#create-a-check-run) to add more information, even with a nicely formatted markdown description as in [this example](https://github.com/nomeata/lean4/pull/1/checks?check_run_id=20165137082), but it seems there you can’t set a summary that’s visible without an extra click, and Github seems to associate these checks to “the first workflow”, which is odd. So using statuses seems fine here.

Often one uses bots writing PR comments for this purpose, but that's a bit noisy (extra notifications etc.), especially for stuff that happens on every PR, but isn’t always interesting/actionable

If this works well, we can use this for more pieces of information, and a link can be added as well.

### DIFF
--- a/.github/workflows/pr-release.yml
+++ b/.github/workflows/pr-release.yml
@@ -6,6 +6,10 @@
 # Instead we use `workflow_run`, which essentially allows us to escalate privileges
 # (but only runs the CI as described in the `master` branch, not in the PR branch).
 
+# The main specification/documentation for this workflow is at
+# https://leanprover-community.github.io/contribute/tags_and_branches.html
+# Keep that in sync!
+
 name: PR release
 
 on:


semorrison:
bors merge

PR #3139, opened 2024-01-04
fix: reduceNat? match terms with free or meta variables

joehendrix:
This removes checks in `Lean.Meta.reduceNat?` that caused it to fail on terms it could handle because they contain meta variables in arguments.  This lead to those operations being reduced using their equational definitions and slow performance on large patterns:

```
set_option profiler true
set_option profiler.threshold 1

def testMod (x:Nat) :=
  match x with
  | 128 % 1024 => true
  | _ => false
-- elaboration took 3.02ms

def testMul (x:Nat) :=
  match x with
  | 128 * 1 => true
  | _ => false
-- type checking took 11.1ms
-- compilation of testMul.match_1 took 313ms
-- compilation of testMul took 65.7ms
-- elaboration took 58.9ms
```

Performance is slower on `testMul` than `testMod` because `whnf` ends up evaluateing `128 * 1` using Peano arithmetic while `128 % 1024` is able to avoid that treatment since `128 < 1024`.

### DIFF
--- a/src/Lean/Meta/WHNF.lean
+++ b/src/Lean/Meta/WHNF.lean
@@ -850,27 +850,26 @@ def reduceBinNatPred (f : Nat → Nat → Bool) (a b : Expr) : MetaM (Option Exp
   return toExpr <| f a b
 
 def reduceNat? (e : Expr) : MetaM (Option Expr) :=
-  if e.hasFVar || e.hasMVar then
-    return none
-  else match e with
-    | .app (.const fn _) a                =>
-      if fn == ``Nat.succ then
-        reduceUnaryNatOp Nat.succ a
-      else
-        return none
-    | .app (.app (.const fn _) a1) a2 =>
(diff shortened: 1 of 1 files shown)


leodemoura:
@joehendrix Thanks for catching this issue. Could you please add new tests using `set_option maxHeartbeats <small-value>` that fail without the fix?

joehendrix:
Yes, I validated on my machine, and reordered the commits too so hopefully the CI will [report the first commit fails](https://github.com/leanprover/lean4/actions/runs/7414007763).

joehendrix:
Oh, it didn't actually run the tests.

joehendrix:
!bench

semorrison:
Uh oh, this has caused a bunch of timeouts in Mathlib. I wish that we had tested Mathlib first before merging! (Per the bot's instructions above, this requires basing the Lean PR off `nightly`, or even better `nightly-with-mathlib`.)

PR #3141, opened 2024-01-05
fix: Initialization and finalization of g_bool_true in kernel

joehendrix:
It looks like when `g_bool_true` was introduced, it didn't follow the pattern of other globals in being made persistent and freed in `finalize_type_checker`.


leodemoura (approved):

joehendrix:
This fix was merged in #3134.

PR #3143, opened 2024-01-07
chore: Partially revert #3139

joehendrix:
This partially reverts #3139 to fix places what that lead to WHNF failures in Mathlib.

It's not quite clear that this is the best way to resolve those failures, but this aims to keep the Lean test case for #3139 working while minimizing the disruption to Mathlib.


semorrison:
Just recording my understanding: for now we are not planning on merging this, instead working on #3144, and leaving Mathlib's `nightly-testing` branch broken in the interim.

PR #3144, opened 2024-01-07
perf: (try to) fix regression introduced by #3139

leodemoura:
#3139 created performance issues in Mathlib.


leodemoura:
Merged with #3124

PR #3147, opened 2024-01-08
doc: Document that `Float` corresponds to 64-bit `double` in C

### LINKED ISSUE #3142: The documentation of `Float` doesn't say if it's 32 or 64 bit
### Prerequisites

### Description

[The documentation for `Float`](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Float.html) doesn't say whether it corresponds to a 32-bit or 64-bit `float` or `double` in C.

### Context

I'm doing some numerics, and for some pieces that aren't verified I'm using `Float`.  It should be easy to know which float type I'm using.

### Versions

N/A, since the bug is in the current website https://leanprover-community.github.io/mathlib4_docs/Init/Data/Float.html.

### Impact

girving:
Closes #3142.

### DIFF
--- a/src/Init/Data/Float.lean
+++ b/src/Init/Data/Float.lean
@@ -26,7 +26,7 @@ opaque floatSpec : FloatSpec := {
   decLe := fun _ _ => inferInstanceAs (Decidable True)
 }
 
-/-- Native floating point type, corresponding to 64-bit `double` in C -/
+/-- Native floating point type, corresponding to 64-bit `double` in C. -/
 structure Float where
   val : floatSpec.float
 


src/Init/Data/Float.lean
```diff
@@ -26,6 +26,7 @@ opaque floatSpec : FloatSpec := {
   decLe := fun _ _ => inferInstanceAs (Decidable True)
 }
 
+/-- Native floating point type, corresponding to 64-bit `double` in C -/
```
semorrison, line 29:
```suggestion
/-- Native floating point type, corresponding to 64-bit `double` in C. -/
```

src/Init/Data/Float.lean
```diff
@@ -26,6 +26,7 @@ opaque floatSpec : FloatSpec := {
   decLe := fun _ _ => inferInstanceAs (Decidable True)
 }
 
+/-- Native floating point type, corresponding to 64-bit `double` in C. -/
```
digama0, line 29:
```suggestion
/-- Native floating point type, corresponding to the IEEE 754 *binary64* format
(also known as `double` in C or `f64` in Rust). -/
```
Avoid defining things in terms of C directly, since those definitions are target-dependent.

girving: Apologies, I didn’t see this until after it had been merged.

However, I believe Lean compiles to C, which does not technically guarantee IEEE compliance.  Assuming no extra step by Lean to ensure compliance, I went with what I thought was more universally correct.

digama0: I believe we want to guarantee IEEE compliance, independent of what C does. We do not generally use C to define the semantics of our operations, and if C has target dependent behavior then we would `#ifdef` our way around it in order to get the semantics we want. The C backend may not even be the main backend in the future, with work on LLVM progressing.

girving: Sounds good, I’ll send another PR in a couple hours.

girving: https://github.com/leanprover/lean4/pull/3157

PR #3149, opened 2024-01-08
feat: `lake update` from unsupported manifest versions

tydeu:
If the current manifest is from unsupported (or has errors), a bare `lake update` will now discard it and create a new one from scratch rather than erroring and requiring you to manually delete the manifest. Lake will produce warnings noting it is ignoring such invalid manifests.

### DIFF
--- a/src/lake/Lake/Load/Main.lean
+++ b/src/lake/Lake/Load/Main.lean
@@ -127,6 +127,8 @@ def Workspace.updateAndMaterialize (ws : Workspace)
     | .error (.noFileOrDirectory ..) =>
       logInfo s!"{ws.root.name}: no previous manifest, creating one from scratch"
     | .error e =>
+      unless toUpdate.isEmpty do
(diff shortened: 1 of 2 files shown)


tydeu:
One design consideration here is whether this should be done on all `lake update` invocations, including ones asking to only update a single package (e.g., `lake update std`) or if this behavior should only occur on a bare `lake update`.

semorrison:
I think I'd personally prefer the more conservative behaviour (only start over on a bare `lake update`), as usually when I'm running `lake update X` I really want to be careful to not update the other dependencies, so it isn't helpful if this happens anyway because there was a merge conflict I missed that has borked the manifest.

But I don't think it's critical, this is a rare enough corner case.

tydeu:
@semorrison Thanks for the input! I was really unsure myself, so I really appreciate it. I think you are right. The current behavior defaults an invalid manifest `lake update foo` to essentially a  bare `lake update`, which is generally never the goal when running `lake update foo`.

PR #3152, opened 2024-01-09
fix: reduce let-bodies correctly in `StructInst`

### LINKED ISSUE #3146: Unification bug
### Prerequisites

### Description

The following snippet produces an error
```lean
def Product (m₁ : Type → Type) (m₂ : Type → Type) (α : Type) := m₁ α × m₂ α

instance [Monad m₁] [Monad m₂] : Monad (Product m₁ m₂) where
  pure x := (pure x, pure x)
  bind o f :=
    let y₁ := do f (← o.1) |>.1
    let y₂ := do f (← o.2) |>.2
    (y₁, y₂)
```
the error is: `missing fields: 'map', 'seq', 'seqLeft'`.

### Context

See [the discussion on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/Monad.20instance.20complains.20about.20missing.20fields).

### Versions

Lean: 4.0.0
OS: NixOS unstable

### Impact

This is not very impactful in my opinion, writing this example in a slightly different way removes the error, eg.
```lean
instance [Monad m₁] [Monad m₂] : Monad (Product m₁ m₂) where
  pure x := (pure x, pure x)
  bind o f :=
    (do f (← o.1) |>.1, do f (← o.2) |>.2)
```
or extract `pure` and `bind` in the instances of `Pure` and `Bind`.

arthur-adjedj:
Closes #3146

Reduction doesn't trigger correctly on the bodies of `let`-expressions in `StructInst`, leading some meta-variables to linger in the terms of some fields. Because of this, default fields may try multiple times (and fail) to be generated, leading to an unexpected error.

The solution implemented here is to modify the values of the introduced variables in the local context so as to reduce them correctly.

### DIFF
--- a/src/Lean/Elab/StructInst.lean
+++ b/src/Lean/Elab/StructInst.lean
@@ -826,9 +826,22 @@ def mkDefaultValue? (struct : Struct) (cinfo : ConstantInfo) : TermElabM (Option
 /-- Reduce default value. It performs beta reduction and projections of the given structures. -/
 partial def reduce (structNames : Array Name) (e : Expr) : MetaM Expr := do
   match e with
-  | .lam ..       => lambdaLetTelescope e fun xs b => do mkLambdaFVars xs (← reduce structNames b)
   | .forallE ..   => forallTelescope e fun xs b => do mkForallFVars xs (← reduce structNames b)
-  | .letE ..      => lambdaLetTelescope e fun xs b => do mkLetFVars xs (← reduce structNames b)
+  | .lam ..
+  | .letE ..      => lambdaLetTelescope e fun xs b => do
+    /- The bodies of let-declarations also need to be reduced.
+       Otherwise, some metavariables may be kept in the terms, leading to errors
+       when trying to generate default values.
+       Fixes `#3146`
+    -/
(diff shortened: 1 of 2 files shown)


arthur-adjedj:
awaiting-review

kmill:
Thanks! Looks correct to me.

It seems there are still bugs in a few cases of this function, like the binder types to foralls, lambdas, and lets. I'll take a look at that soon.

arthur-adjedj:
Do you have any MWE of bugs still happening there ?
AFAIK I tried triggering the bug on lambda/forall binder types, but was not able to reproduce the issue on those back then, hence why I conservatively only reduced the terms in let-bound variables. I'm happy to extend the reduction to binder types as well.

kmill:
No, no MWEs, I'm just commenting based on my reading of the source. I'll let you know if I come across one.

kmill:
@arthur-adjedj While I was fixing another issue, it turned out it was exactly a case where a fun binder needed to be reduced. It was easy enough to rearrange the code, so I went ahead and did it myself, but thank you for your offer. #5844

PR #3155, opened 2024-01-10
chore: rename congrArg to congr_arg, also congrFun to congr_fun

kim-em:
Back in 2022 Mario suggested a long list of naming changes (#1897) in Lean 4, after the naming conventions had been battle tested during the Mathlib port. Unfortunately that PR never went anywhere.

I do think it's important that we resolve some of these naming discrepancies: naming is hard enough for newcomers to Lean, and it doesn't help if some very frequently used names violate the conventions that a new user is learning everywhere else! Changing these names only gets harder over time, so despite it being annoying to do now, it's still the best time to do it.

One that bothers me occasionally still is `congrArg` vs `congr_arg` (Mathlib defines and uses `congr_arg`, but Std doesn't, on the principle that it wants to stick with the core naming, even if it is wrong!)

This PR renames `congrArg` to `congr_arg` and `congrFun` to `congr_fun`, leaving the original definitions in place.

Unfortunately it is not (easily) possible to use the `@[deprecated]` attribute here: `congrArg` is define in `Init.Prelude`, but we can't use `ParametricAttribute` on declarations from imported modules!

We could consider solutions to that, or just go with the "soft" deprecation used here (i.e. just a doc-string), and allow a longer time period before removing the old name.

### DIFF
--- a/src/Lean/Linter/Deprecated.lean
+++ b/src/Lean/Linter/Deprecated.lean
@@ -42,18 +42,3 @@ def checkDeprecated [Monad m] [MonadEnv m] [MonadLog m] [AddMessageContext m] [M
     | none => pure ()
     | some none => logWarning <| .tagged ``deprecatedAttr m!"`{declName}` has been deprecated"
     | some (some newName) => logWarning <| .tagged ``deprecatedAttr m!"`{declName}` has been deprecated, use `{newName}` instead"
-
--- The following deprecations are have their replacements defined in earlier files,
--- and as `deprecated` can not be applied to imported declarations,
--- we have moved the old versions here.
-
-/-- Deprecated synyonym for `congr_arg`. -/
-@[deprecated congr_arg]
-theorem congrArg {α : Sort u} {β : Sort v} {a₁ a₂ : α} (f : α → β) (h : Eq a₁ a₂) : Eq (f a₁) (f a₂) :=
-  h ▸ rfl
-
-
-/-- Deprecated synyonym for `congr_fun`. -/
-@[deprecated congr_fun]
-theorem congrFun {α : Sort u} {β : α → Sort v} {f g : (x : α) → β x} (h : Eq f g) (a : α) : Eq (f a) (g a) :=
-  h ▸ rfl
--- a/src/Init/Notation.lean
+++ b/src/Init/Notation.lean
@@ -490,3 +490,18 @@ a string literal with the contents of the file at `"parent_dir" / "path" / "to" 
 file cannot be read, elaboration fails.
 -/
 syntax (name := includeStr) "include_str " term : term
+
+-- The following deprecations are have their replacements defined in earlier files,
+-- and as `deprecated` can not be applied to imported declarations,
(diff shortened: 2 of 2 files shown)


digama0:
Why not rename `congrArg` to `congr_arg` in Init.Prelude and then introduce `congrArg` at the earliest opportunity as a deprecated alias for `congr_arg`? We could also do this in std with the bump, although this would still cause breakage for people using it with no imports other than `Init`.

semorrison:
> Why not rename `congrArg` to `congr_arg` in Init.Prelude and then introduce `congrArg` at the earliest opportunity as a deprecated alias for `congr_arg`?

Done. This requires at `update-stage0`, so unfortunately it's not possible to get a Mathlib build going quite yet!

src/Lean/Linter/Deprecated.lean
```diff
@@ -42,3 +42,18 @@ def checkDeprecated [Monad m] [MonadEnv m] [MonadLog m] [AddMessageContext m] [M
...
     | some none => logWarning <| .tagged ``deprecatedAttr m!"`{declName}` has been deprecated"
     | some (some newName) => logWarning <| .tagged ``deprecatedAttr m!"`{declName}` has been deprecated, use `{newName}` instead"
+
+-- The following deprecations are have their replacements defined in earlier files,
+-- and as `deprecated` can not be applied to imported declarations,
+-- we have moved the old versions here.
+
+/-- Deprecated synyonym for `congr_arg`. -/
+@[deprecated congr_arg]
+theorem congrArg {α : Sort u} {β : Sort v} {a₁ a₂ : α} (f : α → β) (h : Eq a₁ a₂) : Eq (f a₁) (f a₂) :=
```
digama0, line 52: Why can't this be in Init again? The `deprecated` attribute should be built-in, in which case it should be possible to write this even in the prelude.

semorrison: When I try this in `Init` I get the error message `invalid `[deprecated]` attribute`. (Same if I leave out the identifier.)

The relevant code is:
```
     match stx with
     | `(attr| deprecated $[$id?]?) =>
       let some id := id? | return none
       let declNameNew ← Elab.resolveGlobalConstNoOverloadWithInfo id
       return some declNameNew
     | _  => throwError "invalid `[deprecated]` attribute"
```
Note sure why that `match` is misbehaving in `Init.Prelude`.

digama0: It's because the `deprecated` attribute itself is defined in `Init/Notation`, it's not a builtin. (Maybe it should be.) But it means that you should be able to place this declaration anywhere after the `syntax` declaration there.

semorrison: Done.

digama0:
closed?

eric-wieser:
Kevin also asked about the status of this PR [here](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Naming.20convention/near/420978451)

PR #3156, opened 2024-01-10
chore: CI: add actionlint action, fix actions

nomeata:
I keep messing things up, so time for some guard rails, so check them using
[actionlint](https://github.com/raven-actions/actionlint).

This also runs [shellcheck](https://www.shellcheck.net/) on the files. Shellcheck
is a bit picky about putting double quotes around variables, and will flag many
cases where we know it’s safe, but why not simply always write the safer variant.

Unfortunately, actionlint does not (yet) check `actions/github-script` scripts, which is
unfortunate. Maybe they will in the future (https://github.com/rhysd/actionlint/issues/389)


semorrison:
@nomeata, this came back from the merge queue due to a conflict. I'm happy with this if you want to send it back once the conflict is resolved.

PR #3160, opened 2024-01-10
feat: manage nested inductive types in `deriving`

### LINKED ISSUE #2329: Support for mutual and nested inductive types as `DecidableEq` instance generator

arthur-adjedj:
Adds support for handling nested inductive types in derive handlers

This changes the `Deriving` codebase a fair bit, I'll probably need to split this into multiple PRs.
In particular, the file `FromToJson.lean` has been refactored a fair bit.
(Will add a more thorough description once the PR is stable)

Closes #2329

### DIFF
--- a/src/Lean/Elab/Deriving/Util.lean
+++ b/src/Lean/Elab/Deriving/Util.lean
@@ -46,7 +46,7 @@ def mkImplicitBinders (argNames : Array Name) : TermElabM (Array (TSyntax ``Pars
     `(implicitBinderF| { $(mkIdent argName) })
 
 /--
-    Represents `ind params1 ... paramsn`, where `paramsi` is either a nested occurence, a constant name, or a free variable
+    Represents `ind params1 ... paramsn`, where `paramsi` is either a nested type formers, a constant name, or a free variable
 
     Example :
     ```
@@ -56,7 +56,7 @@ def mkImplicitBinders (argNames : Array Name) : TermElabM (Array (TSyntax ``Pars
         | foo : A -> B -> Foo A (Option Bar) B Nat → Bar
 
     ```
-    this nested occurence `Foo A (Option Bar) B Nat` is encoded as
+    this nested type formers `Foo A (Option Bar) B Nat` is encoded as
     ```lean
       node Foo #[
         .inr (.bvar 2),
@@ -68,19 +68,19 @@ def mkImplicitBinders (argNames : Array Name) : TermElabM (Array (TSyntax ``Pars
     Remark 1 : Since an inductive can only be nested as a parameter, not an index, of another inductive type,
     we assume `params.size == ind.numParams`
 
-    Remark 2 : Free variables are abstracted away as bound variables in nested occurences.
+    Remark 2 : Free variables are abstracted away as bound variables in nested type formerss.
     This is useful when trying to delete duplicate occurences, since the check is now purely syntactical.
   -/
-inductive NestedOccurence : Type :=
-  | node (ind : InductiveVal) (params : Array (NestedOccurence ⊕ Expr))
+inductive IndTypeFormer : Type :=
+  | node (ind : InductiveVal) (params : Array (IndTypeFormer ⊕ Expr))
   | leaf (ind : InductiveVal) (fvars : Array Expr)
 
-namespace NestedOccurence
+namespace IndTypeFormer
 
-instance : Inhabited NestedOccurence := ⟨leaf default #[]⟩
-instance : Inhabited (NestedOccurence ⊕ α) := ⟨.inl default⟩
+instance : Inhabited IndTypeFormer := ⟨leaf default #[]⟩
+instance : Inhabited (IndTypeFormer ⊕ α) := ⟨.inl default⟩
 
-partial instance : BEq NestedOccurence := ⟨go⟩
+partial instance : BEq IndTypeFormer := ⟨go⟩
 where go
   | leaf ind₁ _,.leaf ind₂ _ => ind₁.name == ind₂.name
   | node ind₁ arr₁,.node ind₂ arr₂ => Id.run do
@@ -92,7 +92,7 @@ where go
     return true
   | _,_ => false
 
-partial instance : ToString NestedOccurence := ⟨go⟩
+partial instance : ToString IndTypeFormer := ⟨go⟩
 where go
   | leaf ind e   => s!"leaf {ind.name} {e}"
   | node ind arr =>
@@ -100,42 +100,42 @@ where go
     s!"node {ind.name} {s}"
 
 @[inline]
-def getIndVal :  NestedOccurence → InductiveVal
+def getIndVal :  IndTypeFormer → InductiveVal
   | leaf indVal _ | node indVal _ => indVal
 
 @[inline]
-def getArr :  NestedOccurence → Array (NestedOccurence ⊕ Expr)
+def getArr :  IndTypeFormer → Array (IndTypeFormer ⊕ Expr)
   | leaf ..    => #[]
   | node _ arr => arr
 
 @[inline]
-def isLeaf : NestedOccurence → Bool
+def isLeaf : IndTypeFormer → Bool
   | leaf .. => true
   | node .. => false
 
 @[inline]
-def isNode : NestedOccurence → Bool := not ∘ isLeaf
+def isNode : IndTypeFormer → Bool := not ∘ isLeaf
 
-partial def containsFVar (fvarId : FVarId) : NestedOccurence → Bool
+partial def containsFVar (fvarId : FVarId) : IndTypeFormer → Bool
   | leaf _ e   => e.any (Expr.containsFVar · fvarId)
   | node _ arr => arr.any (Sum.elim (containsFVar fvarId) (Expr.containsFVar · fvarId))
 
-partial def toListofNests (e : NestedOccurence) : List NestedOccurence :=
+partial def toListofTypeFormers (e : IndTypeFormer) : List IndTypeFormer :=
   match e with
   | .leaf _ _   => []
   | .node _ arr =>
-    let l := flip arr.foldr [] fun occ l =>
-      if let .inl occ := occ then
-          occ.toListofNests ++ l
+    let l := flip arr.foldr [] fun indFormer l =>
+      if let .inl indFormer := indFormer then
+          indFormer.toListofTypeFormers ++ l
       else l
     e::l
 
 /-- Return the inductive declaration's type applied to the arguments in `argNames`. -/
-partial def mkAppTerm (nestedOcc : NestedOccurence) (argNames : Array Name) : TermElabM Term := do
-  go nestedOcc argNames
+partial def mkAppTerm (indTypeFormer : IndTypeFormer) (argNames : Array Name) : TermElabM Term := do
+  go indTypeFormer argNames
 where
-  go (nestedOcc : NestedOccurence) (argNames : Array Name) : TermElabM Term := do
-  match nestedOcc with
+  go (indTypeFormer : IndTypeFormer) (argNames : Array Name) : TermElabM Term := do
+  match indTypeFormer with
     | leaf indVal _ =>
       let f := mkCIdent indVal.name
       let numArgs := indVal.numParams + indVal.numIndices
@@ -149,14 +149,14 @@ where
     | node indVal arr =>
       let f := mkCIdent indVal.name
       let mut args := #[]
-      for nestedOcc? in arr do
-        match nestedOcc? with
-          | .inl occ =>
-            let arg ← go occ argNames
+      for indTypeFormer? in arr do
+        match indTypeFormer? with
+          | .inl indFormer =>
+            let arg ← go indFormer argNames
             args := args.push arg
           | .inr (.bvar i) =>
             let some argName := argNames[argNames.size-i-1]?
-              | throwError s!"Cannot instantiate {nestedOcc} : not enough arguments given"
+              | throwError s!"Cannot instantiate {indTypeFormer} : not enough arguments given"
             let id := mkIdent argName
             args   := args.push <| ←`($id)
           | .inr e =>
@@ -165,12 +165,12 @@ where
       `(@$f $args*)
 
 /-- Return the inductive declaration's type applied to the arguments in `argNames`. -/
-partial def mkAppExpr (nestedOcc : NestedOccurence) (argNames : Array Expr) : TermElabM Expr := do
-  let res ← go nestedOcc argNames
+partial def mkAppExpr (indTypeFormer : IndTypeFormer) (argNames : Array Expr) : TermElabM Expr := do
+  let res ← go indTypeFormer argNames
   return res
 where
-  go (nestedOcc : NestedOccurence) (argNames : Array Expr): TermElabM Expr := do
-  match nestedOcc with
+  go (indTypeFormer : IndTypeFormer) (argNames : Array Expr): TermElabM Expr := do
+  match indTypeFormer with
     | leaf indVal _ =>
       let numArgs := indVal.numParams + indVal.numIndices
       unless argNames.size >= numArgs do
@@ -184,14 +184,14 @@ where
       return res
     | node indVal arr =>
       let mut args := #[]
-      for nestedOcc? in arr do
-        match nestedOcc? with
-          | .inl occ =>
-            let arg ← go occ argNames
+      for indTypeFormer? in arr do
+        match indTypeFormer? with
+          | .inl indFormer =>
+            let arg ← go indFormer argNames
             args := args.push arg
           | .inr (.bvar i) =>
             let some argName := argNames[argNames.size-i-1]?
-              | throwError s!"Cannot instantiate {nestedOcc} : not enough arguments given"
+              | throwError s!"Cannot instantiate {indTypeFormer} : not enough arguments given"
             args := args.push argName
           | .inr e =>
             args := args.push e
@@ -199,40 +199,40 @@ where
       return res
 
 structure Result where
-  occ : NestedOccurence
+  indFormer : IndTypeFormer
   args : Subarray Expr
   argNames : Array Name
 
 instance : ToString Result where
-  toString res := s!"⟨{res.occ},{res.args},{res.argNames}⟩"
+  toString res := s!"⟨{res.indFormer},{res.args},{res.argNames}⟩"
 
-instance : BEq Result := ⟨(·.occ == ·.occ)⟩
+instance : BEq Result := ⟨(·.indFormer == ·.indFormer)⟩
 
 structure Context where
   indNames : List Name
   results : List Result
 
-end NestedOccurence
+end IndTypeFormer
 
-abbrev NestedOccM := StateT NestedOccurence.Context TermElabM
+abbrev IndTypeFormerM := StateT IndTypeFormer.Context TermElabM
 
-def withIndNames (indNames : List Name) (f : NestedOccM Unit) : TermElabM NestedOccurence.Context := do
+def withIndNames (indNames : List Name) (f : IndTypeFormerM Unit) : TermElabM IndTypeFormer.Context := do
   let ⟨(),ctx⟩ ← StateT.run f ⟨indNames,[]⟩
   return ctx
 
-def addResult (x : NestedOccurence.Result) : NestedOccM Unit := do
+def addResult (x : IndTypeFormer.Result) : IndTypeFormerM Unit := do
   let ⟨names,res⟩ ← get
-  set (⟨names,x::res⟩ : NestedOccurence.Context)
+  set (⟨names,x::res⟩ : IndTypeFormer.Context)
 
-def addName (n : Name) : NestedOccM Unit := do
+def addName (n : Name) : IndTypeFormerM Unit := do
   let ⟨names,res⟩ ← get
-  set (⟨n::names,res⟩ : NestedOccurence.Context)
+  set (⟨n::names,res⟩ : IndTypeFormer.Context)
 
-partial def getNestedOccurencesOf (inds : List Name) (e: Expr) (fvars : Array Expr): MetaM (Option NestedOccurence) := do
+partial def getIndTypeFormersOf (inds : List Name) (e: Expr) (fvars : Array Expr): MetaM (Option IndTypeFormer) := do
   let .inl occs ← go e | return none
   return occs
 where
-  go (e : Expr) : MetaM (NestedOccurence ⊕ Expr) := do
+  go (e : Expr) : MetaM (IndTypeFormer ⊕ Expr) := do
     let hd := e.getAppFn
     let args := e.getAppArgs
     let fallback _ := return .inr <| e.abstract fvars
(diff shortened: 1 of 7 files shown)


arthur-adjedj:
WIP

nomeata:
Oh, did I miss that this is ready for review? The label still says WIP.

arthur-adjedj:
Whoops, forgot to change that. This is indeed ready for review, I sent Kim some details in DM regarding that PR, but I might as well put it all in here for documentation purposes:
The PR changes a fair amount of things in the `Deriving` codebase, namely:
- The derive handlers for `FromJson` and `ToJson` have been refactored to be similar to the other implementations. Their logic has stayed the same, but the content was cut into more functions to make it more readable (i.e the implementation to derive for structs and inductives used to be in one big function with huge if-then-elses, so these were split).
- The logic in `Util.lean` has been modified. Instead of working on inductive types, it now relies on a type `NestedOccurence`, which represents either base inductives (with the `leaf` constructor), or a nested occurence in the constructor of another inductive type (as `node`s). That way, via an initial traversal the inductive types getting their derivations, we can gather all the different nested occurences appearing in them, to know which functions we need to define in the mutual block to derive the typeclass needed. For example:
```lean4
inductive Foo (α : Type u) : Type u :=
  | foo : List (Option (Foo α)) → Foo α
deriving DecidableEq
```
Will generate a mutual block with the definitions:
```lean4 
mutual
  def decEqOptionFoo {α} [DecidableEq α] (x : Option (Foo α)) (x1 : Option (Foo α)) : Decidable (x = x1) := ...
  def decEqListOptionFoo {α} [DecidableEq α] (x : List (Option (Foo α))) (x1 : List (Option (Foo α))) : Decidable (x = x1) := ...
  def decEqFoo {α} [DecidableEq α] (x : Foo α) (x1 : Foo α) : Decidable (x = x1) := ...
end
```
- Because `Util.lean` has changed, all the derive handlers needed to be adapted to those changes.

When first working on the PR, I was hoping to be able to use these modifications to `Util` to get rid of `usePartial` altogether in the derive handlers. However, the modifications made here are insufficient for this. 
One further change I would like to make in a subsequent PR would be to change how the functions for nested occurences are derived. Currently, if a nested occurence (i.e `Option (Foo A)`) needs to be derived, the derive handler will generate the necessary function. However, if such an instance already exists for i.e `Option A`, using that instance as the function in the mutual block (i.e with `A` replaced with `Foo A`) might prove more sensible. In particular, this would allow us to get rid of `usePartial` in `FromJson`, `ToJson` and `Repr` without changing their current outputs on nested/mutual types. I'm planning on looking into this possibility in the future.

The PR currently passes CI/CL on Lean 4, but fails to compile batteries and mathlib.
- For batteries, because I've needed to upstream  `Sum.elim` from the repo, leading to a duplicate definition
- For Mathlib, because their derive handler for `ToExpr` relies on the lean core deriving mechanism, and would thus need to be adapted to the new API (I am willing to do such a PR)

arthur-adjedj:
awaiting-review

nomeata:
Thanks! Maybe worth doing the mathlib adaption on a branch to be sure that it works before we merge?

Would it be too much work to put the JSON refactoring into a PR of it's own, or would that require work that is then thrown away again?

arthur-adjedj:
WIP

nomeata:
Thanks! Let me know if you need anything from me

arthur-adjedj:
awaiting-review

nomeata:
Thanks! You can push that commit to the branch `lean-pr-testing-3160`, then this should get the nice `builds-mathlib` label and when the maintainers update mathlib after a new lean release they'll find your fixes.

src/Lean/Elab/Deriving/Util.lean
```diff
@@ -36,59 +41,340 @@ open TSyntax.Compat in
...
+
+    Remark 1 : Since an inductive can only be nested as a parameter, not an index, of another inductive type,
+    we assume `params.size == ind.numParams`
+
+    Remark 2 : Free variables are abstracted away as bound variables in nested occurences.
+    This is useful when trying to delete duplicate occurences, since the check is now purely syntactical.
+  -/
+inductive NestedOccurence : Type :=
+  | node (ind : InductiveVal) (params : Array (NestedOccurence ⊕ Expr))
+  | leaf (ind : InductiveVal) (fvars : Array Expr)
```
nomeata, line 76: This seems to the be core concept in this PR. I wonder if we can improve the naming a bit – if `NestedOccurence` shows up in the type signature, it doesn’t really say much.

Also I am not sure if I would call `Foo A (Option Bar) B Nat` a nested occurrence (of `Bar`) – the nested occurrence might be the the `Bar` inside it.

But I see that it’s not easy to come up with a good name. The concept behind it is “in the delaboration of nested inductiction, this corresponds to one of the auxillary inductive definition”, right?

In [Lean.Elab.Structural.IndGroupInst.nestedTypeFormers](https://leanprover-community.github.io/mathlib4_docs/Lean/Elab/PreDefinition/Structural/IndGroupInfo.html#Lean.Elab.Structural.IndGroupInst.nestedTypeFormers) I call them “nested type formers”. Not if that helps.

Maybe `IndTypeFormer`?  At least it has `Ind` in the name.

I’m also a bit surprised that you have to introduce a tree-shaped data type for it. Why not keep the `Expr` around, and traverse and take apart as needed?

(also, `occurrence` with two `r`, right?)

tests/lean/decEqMutualInductives.lean
```diff
@@ -1,18 +1,19 @@
 /-! Verify that the derive handler for `DecidableEq` handles mutual inductive types-/
 
--- Print the generated derivations
-set_option trace.Elab.Deriving.decEq true
```
nomeata, line 4: Why is this gone?

src/Lean/Elab/Deriving/Util.lean
arthur-adjedj: >  if NestedOccurence shows up in the type signature, it doesn’t really say much. 

Right, I agree that this needs renaming

> I am not sure if I would call `Foo A (Option Bar) B Nat` a nested occurrence

I followed the naming done in [this comment](https://github.com/leanprover/lean4/blob/f989520d2b17ec2fe987a69fee44a5272be1ff10/src/kernel/inductive.cpp#L873C71-L873C88) in the kernel, but I'm not sure I agree with that phrasing either. 

> The concept behind it is “in the delaboration of nested inductiction, this corresponds to one of the auxillary inductive definition”, right?

Yes, the idea behind this PR is to take inspiration from the kernel code which converts nested inductives into mutual inductives, so that we may mutually define the definitions necessary for deriving the instances.

> Maybe `IndTypeFormer` ?

Works for me ! I'll rename it.

> I’m also a bit surprised that you have to introduce a tree-shaped data type for it. Why not keep the Expr around, and traverse and take apart as needed ?

I first started my work on this PR about a year ago, so my memory is a bit fuzzy regarding my thought process back then. I definitely did think that this would allow me to work more easily on this, but it is certainly possible to get rid of this tree structure entirely and work on exprs instead, although that would require a fair bit of rewriting.

tests/lean/decEqMutualInductives.lean
arthur-adjedj: That's a mistake, I'll fix this.

nomeata (commented):
Are the changes to stage0 needed?

arthur-adjedj:
`Sum.elim` isn't an absolute necessity for the PR, it's only used [once](https://github.com/leanprover/lean4/blob/29d2785cb56a36181ed45659854eeedb7a143ab2/src/Lean/Elab/Deriving/Util.lean#L121), I just thought this was the best way to code this function and upstreamed what I needed. I could just make the match by hand.

nomeata:
I think I was wondering whether and why it needs a stage0 update.

arthur-adjedj:
Hmm, I remember having some build failures on `stage2` without it at some point, but it seems to be working without it now. All the better ! I'll remove it.

nomeata:
Playing around with it a bit I notice that for
```
inductive Tree where | node : Array Tree → Tree
deriving DecidableEq, BEq
```
we get
```
mutual
       set_option match.ignoreUnusedAlts✝ true
       private partial def beqTree✝ (x✝ : @Tree✝) (x✝¹ : @Tree✝) : Bool✝ :=
         let localinst✝ : BEq✝ (@Tree✝) := ⟨beqTree✝⟩;
         match x✝, x✝¹ with
         | @Tree.node a✝, @Tree.node b✝ => a✝ == b✝
         | _, _ => false✝
     end,
     instance : BEq✝ (@Tree✝) :=
       ⟨beqTree✝⟩
```    
but
```
mutual
       private def decEqListTree✝ (x✝ : @List✝ (@Tree✝)) (x✝¹ : @List✝ (@Tree✝)) : Decidable✝ (x✝ = x✝¹) :=
         match x✝, x✝¹ with
         | @List.nil _, @List.nil _ => isTrue✝ rfl✝
         | List.nil .., List.cons .. => isFalse✝ (by intro h✝; injection h✝)
         | List.cons .., List.nil .. => isFalse✝ (by intro h✝; injection h✝)
         | @List.cons _ a✝ a✝¹, @List.cons _ b✝ b✝¹ =>
           let inst✝ := decEqTree✝ @a✝ @b✝;
           if h✝¹ : @a✝ = @b✝ then by subst h✝¹;
             exact
               let inst✝¹ := decEqListTree✝ @a✝¹ @b✝¹;
               if h✝² : @a✝¹ = @b✝¹ then by subst h✝²; exact isTrue✝¹ rfl✝¹
               else isFalse✝¹ (by intro n✝; injection n✝; contradiction)
           else isFalse✝² (by intro n✝¹; injection n✝¹; contradiction)
       termination_by structural x✝
       private def decEqArrayTree✝ (x✝² : @Array✝ (@Tree✝)) (x✝³ : @Array✝ (@Tree✝)) : Decidable✝ (x✝² = x✝³) :=
         match x✝², x✝³ with
         | @Array.mk _ a✝², @Array.mk _ b✝² =>
           let inst✝² := decEqListTree✝ @a✝² @b✝²;
           if h✝³ : @a✝² = @b✝² then by subst h✝³; exact isTrue✝² rfl✝²
           else isFalse✝³ (by intro n✝²; injection n✝²; contradiction)
       termination_by structural x✝²
       private def decEqTree✝ (x✝⁴ : @Tree✝) (x✝⁵ : @Tree✝) : Decidable✝ (x✝⁴ = x✝⁵) :=
         match x✝⁴, x✝⁵ with
         | @Tree.node a✝³, @Tree.node b✝³ =>
           let inst✝³ := decEqArrayTree✝ @a✝³ @b✝³;
           if h✝⁴ : @a✝³ = @b✝³ then by subst h✝⁴; exact isTrue✝³ rfl✝³
           else isFalse✝⁴ (by intro n✝³; injection n✝³; contradiction)
       termination_by structural x✝⁴
     end,
     instance : DecidableEq✝ (@Tree✝) :=
       decEqTree✝
```

Where in the code is the decision made to translate them differently?

I’m a bit worried about the latter. While all `DecidableEq` are logically equivalent, they do differ in their run-time behavior, and this matters whenever someone writes `if t1 = t2`, right? Otherwise we wouldn’t have a custom implementation for array in https://github.com/leanprover/lean4/blob/e9e858a4484905a0bfe97c4f05c3924ead02eed8/src/Init/Data/Array/DecidableEq.lean#L47-L51.

The author of the `Tree` type above might reasonable expect that decidable equality on their type uses that Array implementation, and not a naive one via list as generated by the deriving code.

But I also don’t see an easy alternative. `partial` doesn’t make sense for `DecidableEq`, I assume. Somehow copying the instance code seems hairy.

Maybe after creating `decEqArrayTree` we could automatically prove a `csimp` lemma that connects it to the inferred implementation? For `DecidableEq` at least that proof is simple (`DecidableEq` is a subsingleton). Would that work?

Which other instances are not using `partial` here? It seems that most use `partial` and hence the “right” instance.

src/Lean/Elab/Deriving/Util.lean
nomeata: I wonder if, now that it’s there, you could use the [Lean.Elab.Structural.IndGroupInst](https://leanprover-community.github.io/mathlib4_docs/Lean/Elab/PreDefinition/Structural/IndGroupInfo.html#Lean.Elab.Structural.IndGroupInst) infrastructure instead. This is what’s used by the structural recursion compiler, which is what you are targetting, so maybe that's natural and easier.

Once you instantiate the `IndGroupInst` with your universe parameters and free variables for the inductive parameters, you can get all typeformers in the group. And when you look at the argument of a constructor, you can find out which type former it’s an instance of, and with which index instantiations, like so:

https://github.com/leanprover/lean4/blob/e9e858a4484905a0bfe97c4f05c3924ead02eed8/src/Lean/Elab/PreDefinition/Structural/FindRecArg.lean#L198-L200

digama0:
I'm not sure I would call the `partial` instance the "right" one. Especially if one hopes to prove properties about the `BEq` instance (I have wanted to implement a `LawfulBEq` deriving instance for some time), `partial` makes that impossible so I would prefer to not see it used any more than it already is. Instead, we should have better combinators to make it possible to actually implement a terminating function over nested recursive types which doesn't have to break the `Array` abstraction.

arthur-adjedj:
>Where in the code is the decision made to translate them differently?

This is dictated by the `withNested` argument of [`mkContext`](https://github.com/leanprover/lean4/blob/5e01b496852891d8d0f80d09c404c640d5b291c0/src/Lean/Elab/Deriving/Util.lean#L353). This differentiation was made to ensure the behaviour of the usual derive handlers wouldn't change. For example, the `FromJson` and `ToJson` instances for `Array` are quite specific, and having them change in nested inductives would break a lot of code, and lead to unreliable derive handlers. 

> I’m a bit worried about the latter. While all DecidableEq are logically equivalent, they do differ in their run-time behavior, and this matters whenever someone writes if t1 = t2, right?

Right. This is the biggest issue with the current PR IMO, and I plann on tackling this in a later PR if that's acceptable (I have not yet started working on any proof-of-concept for such a PR). 

> Somehow copying the instance code seems hairy.

This is my initial solution to the problem, but it would impose some challenges (for example, if some instance is defined using some `extern` binding, how feasable would it be to add that binding to the "new" function as well ?).
I think doing this would be the most reliable solution though, as it would allow most handlers to get rid of `partial` altogether. Adding some `csimp` lemmas to the mix as well might solve all the abovementioned issues, but on non-singleton types, proving these csimp lemmas might prove non-trivial..

> Which other instances are not using partial here ?

For now, only `DecidableEq` gets rid of `partial`.

> I'm not sure I would call the partial instance the "right" one.

I agree that it would be ideal to get rid of `partial` in the derive handlers altogether, any other ideas on how this should be solved are welcome.

src/Lean/Elab/Deriving/Util.lean
arthur-adjedj: This might indeed work. One difference is that I can't exactly rely on `IndGroupInst.nestedTypeFormers` to make this work. In particular, I also need to look for nested instances inside the "auxiliary inductive definitions" generated, until I reach a fixpoint. For example when working with `inductive Tree where | node : Array Tree → Tree`, I'll not only need a nested type former for `Array`, but also one for `List`.

nomeata:
> but on non-singleton types, proving these csimp lemmas might prove non-trivial..

What I mean is that for every Type t, the Type `DecidableEq t` is a subsingleton (isn't it?) so proving a csimp between the derived and the user defined one should always be trivial?

nomeata:
Would be nice indeed, but I don't really have a good idea. Maybe with https://github.com/leanprover/lean4/issues/4749 and with all functions involved refined using primitive recursion? But that won't help when a user wrote an off standard method.

For the present PR I am hesitant to merge it while we get the unexpected behavior. Maybe the `csimp` approach would be good enough.

Or, even less ambitious: somehow keep track of which instances were derived and only derive nested instances where all involved type formers have instances that were also derived and where we thus know that they behave like the one we are writing.

Sorry for not thinking of this issue before!

arthur-adjedj:
I think the `csimp` approach is doable. Another, albeit hacky-ish solution, would be to 
- keep the current derivation
- we could make a `partial` `DecidableEq` instance using `Classical` theorems, and link that to the derived implementation using `implemented_by` to get the correct runtime behaviour

For example, the code generated for the array tree would look like this:
```lean
inductive Tree where | node : Array Tree → Tree

noncomputable instance : Inhabited (DecidableEq α) := ⟨Classical.typeDecidableEq _⟩  in
partial def decEqTreePartial (x4 : @Tree) (x5 : @Tree) : Decidable (x4 = x5) :=
  let inst : DecidableEq Tree := decEqTreePartial
  match x4, x5 with
  | @Tree.node a3, @Tree.node b3 =>
    if h4 : @a3 = @b3 then by subst h4; exact isTrue rfl
    else isFalse (by intro n3; injection n3; contradiction)

mutual
  def decEqListTree (x : List Tree) (x1 : List Tree) : Decidable (x = x1) :=
    match x, x1 with
    | @List.nil _, @List.nil _ => isTrue rfl
    | List.nil .., List.cons .. => isFalse (by intro h; injection h)
    | List.cons .., List.nil .. => isFalse (by intro h; injection h)
    | @List.cons _ a a1, @List.cons _ b b1 =>
      let inst := decEqTree a b;
      if h1 : a = b then by subst h1; exact (
          let inst1 := decEqListTree @a1 @b1
          if h2 : a1 = b1 then 
            by subst h2; exact isTrue rfl
          else 
            isFalse (by intro n; injection n; contradiction)
      )
      else isFalse (by intro n1; injection n1; contradiction)
  termination_by structural x
  
  def decEqArrayTree (x2 : @Array (@Tree)) (x3 : @Array (@Tree)) : Decidable (x2 = x3) :=
    match x2, x3 with
    | @Array.mk _ a2, @Array.mk _ b2 =>
      let inst2 := decEqListTree @a2 @b2;
      if h3 : @a2 = @b2 then by subst h3; exact isTrue rfl
      else isFalse (by intro n2; injection n2; contradiction)
  termination_by structural x2

  @[implemented_by decEqTreePartial]
  def decEqTree (x4 : Tree) (x5 : Tree) : Decidable (x4 = x5) :=
    match x4, x5 with
    | Tree.node a3, Tree.node b3 =>
      let inst3 := decEqArrayTree a3 b3;
      if h4 : a3 = b3 then by subst h4; exact isTrue rfl
      else isFalse (by intro n3; injection n3; contradiction)
  termination_by structural x4
end

instance : DecidableEq Tree :=
  decEqTree
```

nomeata:
Sorry, but I'm not getting it. What is the purpose of `decEqTreePartial`, and how does it help?

arthur-adjedj:
`decEqTreePartial` is "the right" `DecidableEq` instance in that it provides the expected runtime behaviour for `decEqTree`, and will use the already defined `DecidableEq` instances for `List` and `Array` in that example

nomeata:
Ah, now I see it, on the laptop the code is easier to understand :-)


And the purpose of
```
noncomputable local instance : Inhabited (DecidableEq α) := ⟨Classical.typeDecidableEq _⟩  in
```
is merely to to avoid
```
failed to compile partial definition 'decEqTreePartial', failed to show that type is inhabited and non empty
```
right?

Would it be better or worse to omit that and use `unsafe` instead of `partial` then? That definition is only used in `implementedBy`, so we might as well hide it from the kernel completely.


Hmm, not bad overall. Better or worse than this variant?

```lean
inductive Tree where | node : Array Tree → Tree

mutual
  def decEqListTree (x : List Tree) (x1 : List Tree) : Decidable (x = x1) :=
    match x, x1 with
    | @List.nil _, @List.nil _ => isTrue rfl
    | List.nil .., List.cons .. => isFalse (by intro h; injection h)
    | List.cons .., List.nil .. => isFalse (by intro h; injection h)
    | @List.cons _ a a1, @List.cons _ b b1 =>
      let inst := decEqTree a b;
      if h1 : a = b then by subst h1; exact (
          let inst1 := decEqListTree @a1 @b1
          if h2 : a1 = b1 then 
            by subst h2; exact isTrue rfl
          else 
            isFalse (by intro n; injection n; contradiction)
      )
      else isFalse (by intro n1; injection n1; contradiction)
  termination_by structural x
  
  def decEqArrayTree (x2 : @Array (@Tree)) (x3 : @Array (@Tree)) : Decidable (x2 = x3) :=
    match x2, x3 with
    | @Array.mk _ a2, @Array.mk _ b2 =>
      let inst2 := decEqListTree @a2 @b2;
      if h3 : @a2 = @b2 then by subst h3; exact isTrue rfl
      else isFalse (by intro n2; injection n2; contradiction)
  termination_by structural x2

  def decEqTree (x4 : Tree) (x5 : Tree) : Decidable (x4 = x5) :=
    match x4, x5 with
    | Tree.node a3, Tree.node b3 =>
      let inst3 := decEqArrayTree a3 b3;
      if h4 : a3 = b3 then by subst h4; exact isTrue rfl
      else isFalse (by intro n3; injection n3; contradiction)
  termination_by structural x4
end

instance : DecidableEq Tree :=
  decEqTree

-- now patch up the implementations of any nested derived instance

def realDecEqArrayTree (x2 : @Array (@Tree)) (x3 : @Array (@Tree)) : Decidable (x2 = x3) := inferInstance

@[csimp]
theorem decEqArrayTree_eq : @decEqArrayTree = @realDecEqArrayTree := by
  ext t1 t2; apply Subsingleton.allEq 
```

It's still not perfect in the cases where the nested type has a manually written (maybe more efficient) `DecidableEq` instance that the user plans to use in kernel reduction (e.g. using `by decide`).

arthur-adjedj:
The local instance is just used to avoid the error message, and using `unsafe` might just be better indeed.  

In both cases, having a non-ideal kernel reduction seems unavoidable sadly. The ultimate solution would still be to use the already existing instance in the mutual block, but that would open another can of worms I don't have the time to get into currently, hence why I'd like to work on that in a follow-up PR.

If we assume that this is a temporary hack, I would tend towards using the `unsafe` approach instead of the `csimp` one, simply for ease of implementation, but I'm not against choosing `csimp`. Any opinion is welcomed.

nomeata:
Either hack is fine with me, if we want a hack at all. I'm not sure if the feature “derivable decidable equality for nested induction” itself is worth introducing hacks and complications for - how pressing is it for our users? I'll also check back with Leo, as the reporter of #2329.

nomeata:
Ah, forgot to report back.

I don’t think we had a good solution either. 

The safest bet would be to fail instead of deriving possibly bad nested inductives, with a good error message that points to a github issue where users can complain.

The next safest bet might be an internal attribute that we use to track “this type’s `DecidableEq` instance is derived”, and do the construction for those types whose instance is itself derived, because for them we know we are generating equivalent code. Not very satisfying, though.

arthur-adjedj:
> The next safest bet might be an internal attribute that we use to track “this type’s DecidableEq instance is derived”
This would mean that users could derive `DecidableEq` with no issue for an inductive nested by a `List`, but would have an error when trying with an `Array` instead. I don't think this is acceptable

I still believe "specializing" an already existing instance inside our mutual block is the right way to go in the long run, despite the technical challenges it would impose. 

In any case, I do not have time to allocate to this anytime soon. Adding to that that the PR needs a rework to get rid of `IndTypeFormer`s and to rely on the already existing `IndGroupInst` architecture instead, I'm guessing the best choice would be to close the PR ?

nomeata:
It pains me if that’s the conclusion, given the work that went into this PR. Sometimes that’s just how it goes, but still unfortunate.

The refactoring to use `IndGroupInst` could be done by me – that’s a privilege you have as a volunteer contributor over paid employees :-).

But the other, semantic problem remains, and we might have to park this until we have a better plan.

(Just to double-check: For the instances that use `withNested := false` and thus the `partial` translation this PR does not change how it works functionality, does it?)

arthur-adjedj:
>It pains me if that’s the conclusion, given the work that went into this PR.

I don't hold a negative sentiment WRT closing the PR, better have no code merged than some bad one. It's a learning experience either way, I should not have dived into my design choices before first discussing it with someone, doing so would have helped with seeing these issues earlier. 

>For the instances that use `withNested := false` and thus the `partial` translation this PR does not change how it works functionality, does it?

There shouldn't be any change in the behaviour of handlers that use `withNested := false`, no.

PR #3164, opened 2024-01-11
chore: use termination_by in Nat.gcd

joehendrix:
This uses the improved termination_by syntax to give Nat.gcd a cleaner definition.  It removes the last explicit use of WellFounded.fix in Init.

This was also partly motivated by leanprover/std4#520 so that unfold Nat.gcd gives a sensible definition.

### DIFF
--- a/src/Init/Data/Nat/Gcd.lean
+++ b/src/Init/Data/Nat/Gcd.lean
@@ -14,7 +14,7 @@ def gcd (m n : @& Nat) : Nat :=
     n
   else
     gcd (n % m) m
-termination_by gcd m _ => m
+termination_by m
 decreasing_by simp_wf; apply mod_lt _ (zero_lt_of_ne_zero _); assumption
 
 @[simp] theorem gcd_zero_left (y : Nat) : gcd 0 y = y :=


src/Init/Data/Nat/Gcd.lean
```diff
@@ -8,14 +8,14 @@ import Init.Data.Nat.Div
...
 @[extern "lean_nat_gcd"]
-def gcd (a b : @& Nat) : Nat :=
-  WellFounded.fix (measure id).wf gcdF a b
+def gcd (m n : @& Nat) : Nat :=
+  if m = 0 then
+    n
+  else
+    gcd (n % m) m
+termination_by gcd m _ => m
+decreasing_by simp_wf; apply mod_lt _ (zero_lt_of_ne_zero _); assumption
```
nomeata, line 18: Note that your branch point is before  #3040 landed. Once you update, you probably want
```suggestion
termination_by m
decreasing_by simp_wf; apply mod_lt _ (zero_lt_of_ne_zero _); assumption
```

joehendrix: Thanks.  I branched off nightly-with-mathlib.  If mathlib builds ok, I'll rebase and merge this.

PR #3166, opened 2024-01-11
refactor: lake: API cleanup / mark some defs `private`

tydeu:
It has long been requested to better clarify what parts of Lake API are public and which parts are internal implementation details. This change is a small step towards that by marking some definitions that should virtually never be called directly by users as `private`.


digama0:
It would be good to get some data on how much breakage this causes in projects using the Lake API. Is this something we can use the testbed for? Barring that, checking the big projects like doc-gen4 seems prudent.

tydeu:
Superseded by #9749.

PR #3171, opened 2024-01-12
fix: `checkTargets` check for duplicate target

nomeata:
The `checkTargets` function introduced in 4a0f8bf2 as
```
  checkTargets (targets : Array Expr) : MetaM Unit := do
    let mut foundFVars : FVarIdSet := {}
    for target in targets do
      unless target.isFVar do
        throwError "index in target's type is not a variable (consider using the `cases` tactic instead){indentExpr target}"
      if foundFVars.contains target.fvarId! then
        throwError "target (or one of its indices) occurs more than once{indentExpr target}"
```
looks like it tries to check for duplicate indices, but it doesn’t
actually, as `foundFVars` is never written to.

This adds
```
      foundFVars := foundFVars.insert target.fvarId!
```
and a test case.

Maybe a linter that warns about `let mut` that are never writen to would
be useful?

### DIFF
--- a/src/Lean/Elab/Tactic/Induction.lean
+++ b/src/Lean/Elab/Tactic/Induction.lean
@@ -571,6 +571,7 @@ where
         throwError "index in target's type is not a variable (consider using the `cases` tactic instead){indentExpr target}"
       if foundFVars.contains target.fvarId! then
         throwError "target (or one of its indices) occurs more than once{indentExpr target}"
+      foundFVars := foundFVars.insert target.fvarId!
 
 def elabCasesTargets (targets : Array Syntax) : TacticM (Array Expr × Array (Ident × FVarId)) :=
   withMainContext do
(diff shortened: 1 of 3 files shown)


nomeata:
If I read the [lean code in `RecursorInfo`](https://github.com/leanprover/lean4/blob/8e1b51701b929855511762514e528230ea637611/src/Lean/Meta/RecursorInfo.lean) correctly, then this attribute stores the position of the major premise, and runs `mkRecursorInfoCore`, but discards the result of that. 

Maybe it is the case that running `mkRecursorInfoCore` when the attribute is useful in case it throws errors, so that the user knows that the function is unsuitable as a recursor early on, and not just when trying to use it?

digama0:
> Maybe a linter that warns about `let mut` that are never writen to would be useful?

I would implement this, but I'm pretty sure that would involve more changes to Elab/Do.lean than is currently considered acceptable.

PR #3173, opened 2024-01-12
chore: pr-release: more robust comment id recognition

nomeata:
this didn’t recognize the new comments with an intro, and thus the bot
would post multiple comments.

The code was also out of sync with mathlib, fixing.

The `first(…)` in the `jq` program makes it more robust in case this
went wrong once (as on #3171) and there are now multiple PRs matching.

### DIFF
--- a/.github/workflows/pr-release.yml
+++ b/.github/workflows/pr-release.yml
@@ -148,11 +148,15 @@ jobs:
 
             echo "Checking existing messages"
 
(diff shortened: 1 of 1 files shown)


nomeata:
Ok, seems to be working on #3171.

nomeata:
Corresponding mathlib PR: https://github.com/leanprover-community/mathlib4/pull/9698

PR #3182, opened 2024-01-14
fix: make lean4 build on FreeBSD

yurivict:
Upstreaming patches from the FreeBSD port.
The port compiles successfully on FreeBSD and all but 6 tests pass.


src/runtime/io.cpp
```diff
@@ -855,7 +855,13 @@ extern "C" LEAN_EXPORT obj_res lean_io_app_path(obj_arg) {
     char dest[PATH_MAX];
     memset(dest, 0, PATH_MAX);
     pid_t pid = getpid();
-    snprintf(path, PATH_MAX, "/proc/%d/exe", pid);
+#if defined(__linux__)
+     snprintf(path, PATH_MAX, "/proc/%d/exe", pid);
```
semorrison, line 859:
```suggestion
    snprintf(path, PATH_MAX, "/proc/%d/exe", pid);
```

semorrison:
Please don't include the changes to `stage0` in the PR (unless they are needed for subsequent changes, in which case they should be in their own commit. (See https://lean-lang.org/lean4/doc/dev/bootstrap.html#updating-stage0)

yurivict:
Will resubmit.

PR #3183, opened 2024-01-14
doc: clarify and expand docstrings for the `instantiate` functions

kmill:


src/Lean/Expr.lean
```diff
@@ -1228,32 +1228,84 @@ def inferImplicit : Expr → Nat → Bool → Expr
...
+for the binding depth of the replaced loose bound variable.
+-/
 @[extern "lean_expr_instantiate1"]
 opaque instantiate1 (e : @& Expr) (subst : @& Expr) : Expr
 
-/-- Similar to instantiate, but `Expr.bvar i` is replaced with `subst[subst.size - i - 1]` -/
+/--
+Instantiates the loose bound variables in `e` using the `subst` array.
+This is equivalent to `Lean.Expr.instantiate e subst.reverse`, but it avoids reversing the array.
+In particular, rather than instantiating `Expr.bvar i` with `subst[i - d]` it instantiates with `subst[subst.size - i - 1 - d]`,
```
semorrison, line 1275: Surely `subst.size - i - 1 - d` isn't right here?

My linter says that without the reverse `i` and `d` appear with different signs, but now they appear with the same sign, and that can't be right!

kmill: Thanks, that was supposed to be `+ d`! I switched it over to `subst[subst.size - 1 - (i - d)]` since I think that's easier to comprehend.

src/Lean/Expr.lean
```diff
@@ -1228,32 +1228,84 @@ def inferImplicit : Expr → Nat → Bool → Expr
...
 /--
-Instantiate the loose bound variables in `e` using `subst`.
-That is, a loose `Expr.bvar i` is replaced with `subst[i]`.
+Instantiates the loose bound variables in `e` using the `subst` array,
+where a loose `Expr.bvar i` at "binding depth" `d` is instantiated with `subst[i - d]` if `0 <= i - d < subst.size`,
+and otherwise it is replaced with `Expr.bvar (i - subst.size)`; non-loose bound variables are not touched.
+
+If we imagine all expressions as being able to refer to the infinite list of loose bound variables ..., 3, 2, 1, 0 in that order,
+then conceptually `instantiate` is instantiating the last `n` of these and reindexing the remaining ones.
+Warning: `instantiate` uses the de Bruin indexing to index the `subst` array, which might be the reverse order from what you might expect.
```
digama0, line 1237:
```suggestion
Warning: `instantiate` uses the de Bruijn indexing to index the `subst` array, which might be the reverse order from what you might expect.
```

src/Lean/Expr.lean
```diff
@@ -1228,32 +1228,84 @@ def inferImplicit : Expr → Nat → Bool → Expr
...
+See also `Lean.Expr.instantiateRev`.
+
+**Terminology.** The "binding depth" of a subexpression is the number of bound variables available to that subexpression
+by virtue of being in the bodies of `Expr.forallE`, `Expr.lam`, and `Expr.letE` expressions.
+A bound variable `Expr.bvar i` is "loose" if its de Bruijn index `i` is not less than its binding depth.)
+
+**About instantiation.** Instantiation isn't mere substitution.
+When an expression from `subst` is being instantiated, its internal loose bound variables have their de Bruijn indices incremented
+by the binding depth of the replaced loose bound variable.
```
digama0, line 1246: Is this true? My impression was that lean's `instantiate` only works on closed terms for `subst`, which means it doesn't need to lift them when inserting them into `e`.

kmill: Yeah, it's certainly true. I did some tests and checked the source code. Didn't make it up :wink:

I had the same impression as you, and this revelation is what led me to feel the need to update this documentation. I saw Tomas Skriven use `Core.transform` with `instantiate1` and thought that that couldn't be correct, but it was.

(Re your edit: yeah, there's an optimization that if the bvar range of the expression is 0 it can just substitute.)

Here's a test:
```
import Lean

open Lean

#eval
  let e := Expr.lam `a (.const `Nat []) (.bvar 1) .default
  Expr.instantiate1 e (.bvar 0)
/-
Lean.Expr.lam `a (Lean.Expr.const `Nat []) (Lean.Expr.bvar 1) (Lean.BinderInfo.default)
-/
```
If the de Bruijn index lifting weren't present, we'd expect `Lean.Expr.bvar 0` in the output.

src/Lean/Expr.lean
kmill: Thanks, didn't mean one of the members of the Boston hockey team.

src/Lean/Expr.lean
```diff
@@ -1228,32 +1228,84 @@ def inferImplicit : Expr → Nat → Bool → Expr
...
 @[extern "lean_expr_instantiate"]
 opaque instantiate (e : @& Expr) (subst : @& Array Expr) : Expr
 
+/--
+Instantiates loose bound variable `0` in `e` using the expression `subst`,
+where in particular a loose `Expr.bvar i` at binding depth `d` is instantiated with `subst` if `i = d`,
+and otherwise it is replaced with `Expr.bvar (i - 1)`; non-loose bound variables are not touched.
+
+If we imagine all expressions as being able to refer to the infinite list of loose bound variables ..., 3, 2, 1, 0 in that order,
+then conceptually `instantiate` is instantiating the last one of these and reindexing the remaining ones.
```
Kha, line 1261:
```suggestion
then conceptually `instantiate1` is instantiating the last one of these and reindexing the remaining ones.
```

Kha (approved):

PR #3186, opened 2024-01-15
perf: rewrite UnusedVariables lint

digama0:
This is a rewrite of the `UnusedVariables` lint to inline and simplify many of the dependent functions to try to improve the performance of this lint, which quite often shows up in perf reports.

* The mvar assignment scanning is one of the most expensive parts of the process, so we do two things to improve this:
  * Lazily perform the scan only if we need it
  * Use an object-pointer hashmap to ensure that we don't have quadratic behavior when there are many mvar assignments with slight differences.
* The dependency on `Lean.Server` is removed, meaning we don't need to do the LSP conversion stuff anymore. The main logic of reference finding is inlined.
* We take `fvarAliases` into account, and union together fvars which are aliases of a base fvar. (It would be great if we had `UnionFind` here.)

More docs will be added once we confirm an actual perf improvement.

### DIFF
--- a/src/Init/Data/Ord.lean
+++ b/src/Init/Data/Ord.lean
@@ -119,10 +119,11 @@ class Ord (α : Type u) where
 
 export Ord (compare)
 
+set_option linter.unusedVariables false in  -- allow specifying `ord` explicitly
 /--
 Compare `x` and `y` by comparing `f x` and `f y`.
 -/
-@[inline] def compareOn [Ord β] (f : α → β) (x y : α) : Ordering :=
+@[inline] def compareOn [ord : Ord β] (f : α → β) (x y : α) : Ordering :=
   compare (f x) (f y)
 
 instance : Ord Nat where


digama0:
!bench

digama0:
I'm going to try to extract some of those as additional test cases, since the lean4 test suite passed.

digama0:
> The new warnings produced in Std and Mathlib are false positives, e.g.
> 
> ```
> ./.lake/packages/std/././Std/Data/Ord.lean:87:25: warning: unused variable `ord` [linter.unusedVariables]
> ```
> 
> is not unused.

```lean
@[inline] def compareOn [ord : Ord β] (f : α → β) (x y : α) : Ordering :=
  compare (f x) (f y)
```
Are you sure? It looks pretty unused to me. Is there an `IgnoreFunction` which makes this an exception?

digama0:
I reviewed all of the new linter warnings, and all of them appear to be correct (to spec; I know some people have doubts about this algorithm but it's doing what it is designed to do). Do I need to make a fix PR before it is possible to !bench this?

semorrison:
Mathlib's `lean-pr-testing-3186` branch needs to be made into a PR if you want to run `!bench` on Mathlib.

I'm not sure why the bot didn't respond to your `!bench` request above. Let me try it again.

semorrison:
!bench

digama0:
> I'm not sure why the bot didn't respond to your `!bench` request above. Let me try it again.

I'm pretty sure it's because I don't have permission to run the bot (the list of approved users is separate from org members, and I think only you and Sebastian and Leo can run it, and possibly other FRO members as well).

mhuisi:
Note that two tests in the Lean 4 test suite accidentally test false positives. See the `linterUnusedVariables.lean` diff at https://github.com/leanprover/lean4/pull/3082/files#diff-a33d4a408eca48eae24f881090327a42577735140184b3485ec87f08af36a8cc

digama0:
> Mathlib's `lean-pr-testing-3186` branch needs to be made into a PR if you want to run `!bench` on Mathlib.

https://github.com/leanprover-community/mathlib4/pull/9803, are you saying that I should run `!bench` in the comments of that PR or that running it here will work now that the PR exists?

digama0:
> Note that two tests in the Lean 4 test suite accidentally test false positives. See the `linterUnusedVariables.lean` diff at https://github.com/leanprover/lean4/pull/3082/files#diff-a33d4a408eca48eae24f881090327a42577735140184b3485ec87f08af36a8cc

One part of the algorithm is supposed to address this: If there is a global constant info over the same span as a variable declaration, then it is not linted. Apparently this only works for (mutual) `def` though, not `let rec` and `where` declarations, which seems like a bug in the info tree generation. (Go to definition on `whereVariable.x` does seem to work though.)

nomeata:
> > Mathlib's `lean-pr-testing-3186` branch needs to be made into a PR if you want to run `!bench` on Mathlib.
> 
> [leanprover-community/mathlib4#9803](https://github.com/leanprover-community/mathlib4/pull/9803), are you saying that I should run `!bench` in the comments of that PR or that running it here will work now that the PR exists?

If you want to run the mathlib benchmark, you have to say `!bench` there. And very likely you’ll also want to manually create a throw-away PR with the head set to whatever commit you want to compare it to (`nightly-testing-2024-xx-yy` or so). Yes, it’s very tedious.

mhuisi:
> > Note that two tests in the Lean 4 test suite accidentally test false positives. See the `linterUnusedVariables.lean` diff at https://github.com/leanprover/lean4/pull/3082/files#diff-a33d4a408eca48eae24f881090327a42577735140184b3485ec87f08af36a8cc
> 
> One part of the algorithm is supposed to address this: If there is a global constant info over the same span as a variable declaration, then it is not linted. Apparently this only works for (mutual) `def` though, not `let rec` and `where` declarations, which seems like a bug in the info tree generation. (Go to definition on `whereVariable.x` does seem to work though.)

The adjusted implementation of `References.lean` in the linked PR fixes the issue, though.

digama0:
Turns out the reason for the issue is because it is being [explcitly excluded](https://github.com/leanprover/lean4/blob/f8edf452dec50d2d310737c266a1db112386ecda/src/Lean/Linter/UnusedVariables.lean#L259).

digama0:
I don't, but there are asymptotic improvements in this PR, the metavariable scanning was reduced from O(n^2) to O(n) and I believe that's the main bottleneck so I would expect it to have outsized impact on the big proofs that reportedly had issues with this linter.

Kha (commented):
> More docs will be added once we confirm an actual perf improvement.

I found and confirmed a 16x improvement in a particularly gnarly proof term, so this would be greatly appreciated to get it merged :+1:

digama0:
I added a bunch of docs, I think this is ready for review. I still need to double check that all the builtin ignore fns are documented correctly and all do something - I think some of them are currently useless.

Kha:
> I still need to double check that all the builtin ignore fns are documented correctly and all do something - I think some of them are currently useless.

Is that something that is affected by this PR, or should that wait for another PR?

digama0:
Well only insofar as the docs were just added and it would be good to make sure they are correct

src/Lean/Linter/UnusedVariables.lean
```diff
@@ -1,34 +1,109 @@
...
+def getLinterUnusedVariablesFunArgs (o : Options) : Bool :=
+  o.get linter.unusedVariables.funArgs.name (getLinterUnusedVariables o)
 
+/-- Gets the status of `linter.unusedVariables.patternVars` -/
+def getLinterUnusedVariablesPatternVars (o : Options) : Bool :=
+  o.get linter.unusedVariables.patternVars.name (getLinterUnusedVariables o)
+
+/-- An `IgnoreFunction` receives:
+
+* a `Syntax.ident` for the unused variable
```
Kha, line 95: Time to make this an `Ident`? Potential breaking change but downstream users most likely just do a lambda without type annotations; in fact, all uses on GitHub don't use the ident at all.

Kha (approved):
Looks great! Tests need a copy-produced

digama0:
> Tests need a copy-produced

Is this auto-corrupt kicking in? I can't make sense of it.

src/Lean/Linter/UnusedVariables.lean
digama0: The only one that used the ident to begin with was the check for `_x` variables, but this has been incorporated into the main algorithm.

digama0: Based on the way in which this syntax was produced, I'm not sure we can actually say it is definitely an `Ident`. It is the declStx with declId stripped, but if it happened to be some other syntax like an atom that an elaborator decided to use as the span of a variable I think we could end up here with a non- `.ident`. We pass a syntax stack corresponding to an `.ident` but I think the `declStx` doesn't necessarily have to be the same thing as the head of the stack, they are only matched up using syntax ranges.

Kha:
copy-produced is described here: https://lean-lang.org/lean4/doc/dev/testing.html#fixing-tests

src/Lean/Linter/UnusedVariables.lean
Kha: Ok, let's leave it as is then

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -398,7 +398,7 @@ def buildInductionCase (fn : Expr) (oldIH newIH : FVarId) (toClear toPreserve :
   mvarId ← assertIHs IHs mvarId
   for fvarId in toClear do
     mvarId ← mvarId.clear fvarId
-  mvarId ← mvarId.cleanup (toPreserve := toPreserve)
+  _ ← mvarId.cleanup (toPreserve := toPreserve)
```
Kha, line 401: @nomeata This looks like an actual bug, you're never making use of this mvar?

PR #3187, opened 2024-01-16
feat: new implementation for `simp (config := { ground := true })`

leodemoura:


semorrison:
Initial Mathlib errors just look like changed `simp?` output. I've fixed them.

PR #3188, opened 2024-01-16
feat: induction using <term>

nomeata:
right now, the `induction` tactic accepts a custom eliminator using the
`using <ident>` syntax, but is restricted to identifiers. This
limitation becomes annoying when the elminator has explicit parameters
that are not targets, and the user (naturally) wants to be able to write
```
induction a, b, c using foo (x := …)
```

This generalizes the syntax to expressions and changes the code
accordingly.

This can be used to instantiate a multi-motive induction:
```
example (a : A) : True := by
  induction a using A.rec (motive_2 := fun b => True)
  case mkA b IH => exact trivial
  case A => exact trivial
  case mkB b IH => exact trivial
```

For this to work the term elaborator learned the `heedElabAsElim` flag,
`true` by default. But in the default setting, `A.rec (motive_2 := fun b => True)`
would fail to elaborate, because there is no expected type. So the induction
tactic will elaborate in a mode where that attribute is simply ignored.

As a side effect, the “failed to infer implicit target” error message 
is improved and prints the name of the implicit target that could not be
instantiated.

### DIFF
--- a/src/Lean/Meta/AbstractMVars.lean
+++ b/src/Lean/Meta/AbstractMVars.lean
@@ -16,15 +16,15 @@ structure AbstractMVarsResult where
 namespace AbstractMVars
 
 structure State where
-  ngen         : NameGenerator
-  lctx         : LocalContext
-  mctx         : MetavarContext
-  nextParamIdx : Nat := 0
-  paramNames   : Array Name := #[]
-  fvars        : Array Expr  := #[]
-  lmap         : HashMap LMVarId Level := {}
-  emap         : HashMap MVarId Expr  := {}
-  levels       : Bool -- whether to abstract level mvars
+  ngen           : NameGenerator
+  lctx           : LocalContext
+  mctx           : MetavarContext
+  nextParamIdx   : Nat := 0
+  paramNames     : Array Name := #[]
+  fvars          : Array Expr  := #[]
+  lmap           : HashMap LMVarId Level := {}
+  emap           : HashMap MVarId Expr  := {}
+  abstractLevels : Bool -- whether to abstract level mvars
 
 abbrev M := StateM State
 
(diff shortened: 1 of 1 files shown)


nomeata:
This seems to be working in the cases I could think of:

nomeata:
This seems to be working in the trivial cases I could think of, see
https://github.com/leanprover/lean4/blob/joachim/induct-using-term/tests/lean/indUsingTerm.lean

I am reasonably happy with the user-facing behavior of the feature. But I am not not sure if the epicycles I go through in `elabTermForElim` are the best way to go about it. I believe I want to elaborate `term` in  a mode  where, given ``ident : {motive : … } → { motive2 : … } … `,
* `ident` is elaborated like `@ident` and not `@ident ?mvar`
* `ident (motive2 := m2)` is elaborated as `fun m1 => @ident m1 m2`, with type `{motive : … } → …` and not `@ident ?mvar m2`
* `ident a b` is elaborated normally, even if `ident` is marked as “`elab_as_elim`”

I might need some advise here.

src/Lean/Elab/Tactic/Induction.lean
```diff
@@ -499,11 +503,36 @@ def getInductiveValFromMajor (major : Expr) : TacticM InductiveVal :=
...
+    if let some e ← Term.resolveId? stx (withInfo := true) then
+      return e
+  Term.withoutErrToSorry <| Term.withoutHeedElabAsElim do
+    let e ← Term.elabTerm stx none (implicitLambda := false)
+    Term.synthesizeSyntheticMVars (mayPostpone := false) (ignoreStuckTC := true)
+    let e ← instantiateMVars e
+    let e := e.eta
+    if e.hasMVar then
+      let r ← abstractMVars e
+      openAbstractMVarsResultLevels r
```
nomeata, line 527: Better: flag to `abstractMVars` to not abstract level mvars

nomeata:
I showed the PR to Leo today, and he said the approach is reasonable. Will merge soon.

PR #3191, opened 2024-01-17
fix: `derive BEq` on structure with `Prop`-fields

### LINKED ISSUE #3140: `deriving BEq` fails for structures with `Prop` fields
`deriving BEq` currently fails if a `structure` has `Prop` fields, a simple example being

```
structure S where
  p : True
  deriving BEq
```

which produces the error message

```
typeclass instance problem is stuck, it is often due to metavariables
  BEq (?m.302 x✝¹ x✝ a✝ b✝)
```

From discussion [in Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.60deriving.20BEq.60.20breaks.20with.20.60Prop.60.20fields/near/411264738) and here, this is due to `BEq` being defined only over `Type`, not `Sort`.  However, the desire is to keep it  this way, and instead fix `deriving BEq` directly to ignore `Prop` fields (which is fine by proof irrelevance).

### Impact

arthur-adjedj:
Closes #3140

### DIFF
--- a/src/Lean/Elab/Deriving/BEq.lean
+++ b/src/Lean/Elab/Deriving/BEq.lean
@@ -58,8 +58,7 @@ where
             ctorArgs1 := ctorArgs1.push a
             ctorArgs2 := ctorArgs2.push b
             let xType ← inferType x
-            let isProof ← isProp xType
-            if  isProof then
+            if (← isProp xType) then
               continue
             if xType.isAppOf indVal.name then
               rhs ← `($rhs && $(mkIdent auxFunName):ident $a:ident $b:ident)


src/Lean/Elab/Deriving/BEq.lean
```diff
@@ -57,6 +57,9 @@ where
             let b := mkIdent (← mkFreshUserName `b)
             ctorArgs1 := ctorArgs1.push a
             ctorArgs2 := ctorArgs2.push b
+            let isProof := (← inferType (← inferType x)).isProp
+            if  isProof then
+              continue
             if (← inferType x).isAppOf indVal.name then
               rhs ← `($rhs && $(mkIdent auxFunName):ident $a:ident $b:ident)
             else
```
nomeata, line 65: This is running `inferType` twice; maybe store as `xType` and reuse?

Also, maybe use `xType.isProof`:
https://github.com/leanprover/lean4/blob/b614ff1d12bc38f39077f9ce9f2d48b42c003ad0/src/Lean/Meta/InferType.lean#L311-L315

arthur-adjedj: Thanks for pointing it out. Looking that the code for `isProof`, I believe it should be `x.isProof` and not `xType.isProof` Since similar code was present in `DecEq.lean`, I've taken the liberty of using `isProof` there too.

src/Lean/Elab/Deriving/BEq.lean
```diff
@@ -57,7 +57,11 @@ where
             let b := mkIdent (← mkFreshUserName `b)
             ctorArgs1 := ctorArgs1.push a
             ctorArgs2 := ctorArgs2.push b
-            if (← inferType x).isAppOf indVal.name then
+            let xType ← inferType x
+            let isProof ← isProp xType
+            if  isProof then
+              continue
```
nomeata, line 63: Thanks! Now just a minor code style suggestion:
```suggestion
            if (← isProp xType) then
              continue
```

src/Lean/Elab/Deriving/BEq.lean
nomeata: Ah, I guess I meant `isProof x`? Not sure if it’s better if you have to run `inferType x` later anyways.

nomeata: An good point about `Meta.isProp` and `isProp`. This is very confusing indeed, but I think you got it right :-)

arthur-adjedj: My thought process is that `isProofQuick` and `isPropQuick` are very similar, and should be comparable in terms of performance. However, if `isProofQuick` fails, it will have to infer the type, and call `isPropQuick` afterwards. since `inferType x` is already called once, I believe using `isProp` to be more efficient here.

arthur-adjedj:
Perfect. Thanks for taking the time to review this :-)

PR #3195, opened 2024-01-18
feat: cleanups to ACI and Identity classes

joehendrix:
This makes changes to the definitions of Associativity, Commutativity, Idempotence and Identity classes to be more aligned with Mathlib's versions.

The changes are:
*  Move classes are moved from `Lean` to root namespace.
* Drop `Is` prefix from names.
* Rename `IsNeutral` to `LawfulIdentity` and add Left and Right subclasses.
* Change neutral/identity element to outParam.
* Introduce `HasIdentity` for operations not intended for proofs to implement

The identity changes are to make this compatible with [Mathlib](https://github.com/leanprover-community/mathlib4/blob/718042db9d6a9dfa5ba724b8b340000e798263f4/Mathlib/Init/Algebra/Classes.lean) and to enable nicer fold operations in Std that can use type classes to infer the identity/initial element on binary operations.

### DIFF
--- a/src/Init/Core.lean
+++ b/src/Init/Core.lean
@@ -1715,7 +1715,7 @@ class IdempotentOp (op : α → α → α) : Prop where
 This class does not require a proof that `o` is an identity, and
 is used primarily for infering the identity using class resoluton.
 -/
-class LeftIdentity (op : α → β → β) (o : outParam α) : Prop where
+class LeftIdentity (op : α → β → β) (o : outParam α) : Prop
 
 /--
 `LawfulLeftIdentify op o` indicates `o` is a verified left identity of
@@ -1731,7 +1731,7 @@ class LawfulLeftIdentity (op : α → β → β) (o : outParam α) extends LeftI
 This class does not require a proof that `o` is an identity, and is used
 primarily for infering the identity using class resoluton.
 -/
-class RightIdentity (op : α → β → α) (o : outParam β) : Prop where
+class RightIdentity (op : α → β → α) (o : outParam β) : Prop
 
 /--
 `LawfulRightIdentify op o` indicates `o` is a verified right identity of
@@ -1747,7 +1747,7 @@ class LawfulRightIdentity (op : α → β → α) (o : outParam β) extends Righ
 This class does not require a proof that `o` is an identity, and is used
 primarily for infering the identity using class resoluton.
 -/
-class Identity (op : α → α → α) (o : outParam α) extends LeftIdentity op o, RightIdentity op o : Prop where
+class Identity (op : α → α → α) (o : outParam α) extends LeftIdentity op o, RightIdentity op o : Prop
 
 /--
 `LawfulIdentity op o` indicates `o` is a verified left and right


joehendrix:
I've pushed fixes to the Mathlib PR testing branch.  It's been locally tested, but once that verifies, I think this is good for review from my perspective.

urkud:
About dropping `Is`: there is a discussion (in a private stream) about using `Is` prefix for `Prop`-valued mixin classes more consistently in Mathlib.

joehendrix:
Lean 3 used "is_" and "has_" prefixes fairly consistently (e.g., `has_add`, `is_lawful_functor`) , but Lean 4 migrated away from that.  These classes are some of the last ones starting with verbs. 

Mathlib seems to have every style (including these laws as both instances and non-instance defs).  I personally think it'd be good to just adopt a consistent style and I don't think converting every Lean core class back to starting with verbs is a good use of time.

kmill:
For reference, [here's](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/typeclass.20names.20starting.20with.20.60Is.60.3F/near/416458489) a recent Zulip thread about revisiting how `Prop`-valued "mixin" classes are named.

For data-carrying classes like `Add`, it makes sense dropping the verb, but `Prop`-valued classes I think need some more thought about naming. With `is_lawful_functor` -> `LawfulFunctor` for example, this is an adjective-noun naming that adds an extra `Prop` to a pre-existing `Functor`. Adjectives are properties, so there's no need for a helper verb.

There's an example in the thread of `IsTensorProduct`, where `TensorProduct` is a noun, and it's hard to read and understand it as being a `Prop` without the `Is` prefix.

With these `Associative`/`Commutative`/etc classes, I wonder if it would make sense to call them `AssociativeOp`/`CommutativeOp`/etc., since that would be a parallel construction with adjective-noun names such as `LawfulFunctor`, assuming it makes sense to call a `α → α → α` function an "`Op`". Though, "`Binop`" might be better since then that allows having both `IdempotentUnop` and `IdempotentBinop`, since there's a currently a gap in these classes that there's nothing about unary operators.

I would rename `HasIdentity`, since with `Has` I am expecting a data-carrying class rather than a class-level predicate. `IsIdentity` might be ok (it's the is-noun naming pattern), so long as you read it as "is the chosen identity" rather than "is an identity". But, thinking of the outparam as being part of the data of the class, maybe better is just `Identity`. That way `LeftIdentity`, `RightIdentity`, and `LawfulIdentity` are following the adjective-noun naming.

kmill:
Though there's that wrinkle that this isn't exactly the same as `Functor` vs `LawfulFunctor` since these `*Identity` classes are extending `Identity` directly... Is there any need for `HasIdentity`? Can we just delete it, then rename `LawfulIdentity` to `Identity`? "Identity" alone is a common name for a two-sided identity.

src/Init/Core.lean
```diff
@@ -1680,40 +1680,64 @@ So, you are mainly losing the capability of type checking your development using
...
+functions by requiring only left or right identity.
 -/
-class IsNeutral {α : Sort u} (op : α → α → α) (neutral : α) where
-  /-- A neutral element can be cancelled on the left: `e ∘ a = a`. -/
-  left_neutral : (a : α) → op neutral a = a
-  /-- A neutral element can be cancelled on the right: `a ∘ e = a`. -/
-  right_neutral : (a : α) → op a neutral = a
+class LawfulCommIdentity (op : α → α → α) (o : outParam α) [hc : Commutative op] extends LawfulIdentity op o : Prop where
+  left_id a := Eq.trans (hc.comm o a) (right_id a)
+  right_id a := Eq.trans (hc.comm a o) (left_id a)
```
kmill, line 1741: I have a feeling that this class will cause some "impedance mismatch" problems. The issue is that you have to implement it specially if you want to use theorems that refer to `[LawfulCommIdentity op o]`. Maybe instead, we can have instances that derive the lawful identity class from commutativity and either LeftIdentity or RightIdentity?

```lean
/-- A binary operation with a left identity. -/
class LeftIdentity (op : α → α → α) (o : outParam α) : Prop where
  /-- Left identify -/
  left_id : ∀ a, op o a = a

/-- A binary operation with a right identity. -/
class RightIdentity (op : α → α → α) (o : outParam α) : Prop where
  /-- Right identify -/
  right_id : ∀ a, op a o = a

/-- A binary operation with a left and right identity. -/
class Identity (op : α → α → α) (o : outParam α) extends LeftIdentity op o, RightIdentity op o : Prop

instance (op : α → α → α) [LeftIdentity op o] [Commutative op] : Identity op o where
  right_id a := (Commutative.comm a o).trans (LeftIdentity.left_id a)

instance (op : α → α → α) [RightIdentity op o] [Commutative op] : Identity op o where
  left_id a := (Commutative.comm o a).trans (RightIdentity.right_id a)
```
That way theorems that need two-sided identities and a commutative operation can require `[Commutative op] [Identity op o]` and apply more generally than `[Commutative op] [LawfulCommIdentity op o]`.

We could consider making these be lower-priority instances, to keep them from being explored unnecessarily.

fgdorais: I would also be tempted to have 
```lean
class abbrev Identity (op : α → α → α) (o : outParam α) := LeftIdentity op o, RightIdentity op o
```
and perhaps even
```lean
class abbrev Monoid (op : α → α → α) (o : outParam α) := Associative op, Identity op o
```
I think these classes are intended to have very few basic instances, so I don't think these are problematic.

joehendrix:
I'm open to a different name, but `HasIdentity` is important to convey what the unit is for operations like fold:
```lean
def fold {α : Type _} (op : α → α → α) {o : α} [HasIdentity op o] (l:List α) : α :=
  l.foldl (init := o) op
```

Lemmas about fold then use stronger constraints gracefully:
```lean
def fold {α : Type _} (op : α → α → α) {o : α} [Identity op o] (l:List α) : α :=
  l.foldl (init := o) op
```

You make a good point on `Idempotent`.  I'm tempted to rename `Idempotent` to `IdempotentOp` and then Idempotent relations could be `IdempotentRel` and Idempotent functions `IdempotentFn`.  The latter two would be out of scope.  

I'm less persuaded about the proposal to rename `Associative` to `AssociativeOp`.  It's normally going to be written in a class applied to a function and `[Associative f]` seems pretty clear.

For the Identity naming issue, I am wondering to what extent I should provide the full spectrum of left and right identity non-lawful classes.  This would enable us to use `fold` where the left identity and right-identity are distinct (e.g., matrix multiply with type-level dimensions).

 ```lean
class LeftIdentity (op : α → β → β) (o : outParam α) : Prop where

class LawfulLeftIdentity (op : α → β → β) (o : outParam α) extends LeftIdentity op o : Prop where
  left_id : ∀ a, op o a = a

class RightIdentity (op : α → β → α) (o : outParam β) : Prop where

class LawfulRightIdentity (op : α → β → α) (o : outParam β) extends RightIdentity op o : Prop where
  right_id : ∀ a, op a o = a

class Identity (op : α → α → α) (o : outParam α) extends LeftIdentity op o, RightIdentity op o : Prop where

class LawfulIdentity (op : α → α → α) (o : outParam α) extends Identity op o, LeftIdentity op o, RightIdentity op o : Prop
 ```

Feedback would be appreciated on this point.  Is this too much?  Or useful to others?  I consider full heterogeneity out of scope for this change (e.g., an HAssociativity that would work `BitVec` concatenation.)

kmill:
I think your idea of `LeftIdentity`/`LawfulLeftIdentity` etc. is a good one. `LeftIdentity` is saying you have an identity for a left action, which is mathematically meaningful, and it seems to be the right level of generality for folds.

fgdorais:
The distinction between left and right identity is actually fairly common in programming. For example, in the exception monad, `fail` is a left identity for `<|>` but it is not a right identity.

src/Init/Core.lean
```diff
@@ -1680,40 +1680,64 @@ So, you are mainly losing the capability of type checking your development using
 -/
 axiom ofReduceNat (a b : Nat) (h : reduceNat a = b) : a = b
 
+end Lean
+
```
fgdorais, line 1684: I just noticed that the classes were moved out of the `Lean` namespace. This is good, it's a confusing namespace for these algebraic classes. However, the root namespace is perhaps not the right place either. It causes annoyances to anyone who wants to use one of these names in their namespace and these are pretty common names. Root is better than `Lean` but any meaningful namespace is better than either.

joehendrix: Good point.  I'll put them in `Std`.

src/Init/Core.lean
joehendrix: I just pushed some changes to Identity from the other thread.  I'm currently keeping `LawfulCommIdentity` but added a comment that it's not intended for constraints, but only instances.

I don't have a major objection to @kmill's alternative though.  Presumedly there's cycle detection to rule out search loops like searching for `LeftIdentity` to find `Identity` using commutative instance, then searching for `Identity` to find `LeftIdentity` using the extends clause.

joehendrix:
This now builds Mathlib, incorporates the feedback above, and is ready to merge from my perspective.

I have a slight preference for the `LawfulCommIdentity` class over implicitly inferring `Lawfuldentity` via search when `LawfulCommutative` can also be found.  Mainly because I think both are equally easy to implement, but `LawfulCommIdentity` is a bit more explicit.

src/Init/Core.lean
```diff
@@ -1680,40 +1680,92 @@ So, you are mainly losing the capability of type checking your development using
...
+class LawfulIdentity (op : α → α → α) (o : outParam α) extends Identity op o, LawfulLeftIdentity op o, LawfulRightIdentity op o : Prop
+
+/--
+`LawfulCommIdentity` can simplify defining instances of LawfulIdentity
+on commutative functions by requiring only a left or right identity
+proof.
+
+This class is intended for simplifying defining instances of
+`LawfulIdentity` and functions needed commutative operations with
+identity should just add a `LawfulIdentity constraint.
```
kmill, line 1765:
```suggestion
identity should just add a `LawfulIdentity` constraint.
```

src/Init/Core.lean
```diff
@@ -1680,40 +1680,92 @@ So, you are mainly losing the capability of type checking your development using
...
+class Identity (op : α → α → α) (o : outParam α) extends LeftIdentity op o, RightIdentity op o : Prop where
+
+/--
+`LawfulIdentity op o` indicates `o` is a verified left and right
+identity of `op`.
+-/
+class LawfulIdentity (op : α → α → α) (o : outParam α) extends Identity op o, LawfulLeftIdentity op o, LawfulRightIdentity op o : Prop
+
+/--
+`LawfulCommIdentity` can simplify defining instances of LawfulIdentity
```
kmill, line 1759:
```suggestion
`LawfulCommIdentity` can simplify defining instances of `LawfulIdentity`
```

src/Init/Core.lean
```diff
@@ -1680,40 +1680,92 @@ So, you are mainly losing the capability of type checking your development using
...
+on commutative functions by requiring only a left or right identity
+proof.
+
+This class is intended for simplifying defining instances of
+`LawfulIdentity` and functions needed commutative operations with
+identity should just add a `LawfulIdentity constraint.
+-/
+class LawfulCommIdentity (op : α → α → α) (o : outParam α) [hc : Commutative op] extends LawfulIdentity op o : Prop where
```
kmill, line 1767: If this is for defining instances, perhaps making it also extend `Commutative op` would be useful. Is there a reason it's an argument instead?

joehendrix: The one downside of doing that is that it also makes the class convenient for adding commutativity and identity constraints as well: `def foo [LawfulCommIdentity op o] -> ...` which is not the preferred form.

I'm inclined to just leave this as is for now.

kmill: One more thought here: if you made this be a `structure`, you could do `attribute [instance] LawfulCommIdentity.toCommutative LawfulCommIdentity.LawfulIdentity`, and that would protect against it being used for an instance argument and it would give the benefits of it being a `class` (though it'd be worth checking that `LawfulCommIdentity.mk` has its `to*` fields be instance arguments once it's switched to `structure` -- I don't know that off the top of my head).

joehendrix: If I try this:

```lean
structure LawfulCommIdentity (op : α → α → α) (o : α) extends Commutative op, LawfulIdentity op o : Prop where
  left_id a := Eq.trans (comm o a) (right_id a)
  right_id a := Eq.trans (comm a o) (left_id a)

attribute [instance] LawfulCommIdentity.toLawfulIdentity
```

I get the error:
```
instance does not provide concrete values for (semi-)out-params
  @LawfulIdentity α op ?o
```

I'm in favor of leaving this alone and merging as is, but alternatively I'm open to dropping `LawfulCommIdentity` entirely.  I think the benefits are pretty minor and I need to prioritize other topics.

kmill: Thanks for checking (and in retrospect that was a silly suggestion, sorry).

While `LawfulCommIdentity` is an unfamiliar design (I don't think mathlib would have it), so it was worth thinking about some more (as it will probably be here for years if it makes it into core), I think I'm happy enough with this PR as it is. I'll look through once more.

kmill (approved):
Looks good. I have mild concerns about `LawfulCommIdentity`, but it seems useful for the stated use case.

src/Init/Core.lean
joehendrix: Thanks for the feedback though.  I appreciate it.

PR #3199, opened 2024-01-19
chore: CI looks for nightly-testing-YYYY-MM-DD at Mathlib as either a branch or tag

kim-em:
As discussed during the FRO meeting 2024-01-18, we are changing the `nightly-testing-YYYY-MM-DD` branches at Std and Mathlib from branches to tags, in:

* https://github.com/leanprover/std4/pull/545
* https://github.com/leanprover-community/mathlib4/pull/9842

This PR updates the script that creates the `lean-pr-testing-NNNN` branches at Mathlib so it is agnostic about whether `nightly-testing-YYYY-MM-DD` will be a branch or a tag.

### DIFF
--- a/.github/workflows/pr-release.yml
+++ b/.github/workflows/pr-release.yml
@@ -126,14 +126,14 @@ jobs:
           if [ "$NIGHTLY_SHA" = "$MERGE_BASE_SHA" ]; then
             echo "The merge base of this PR coincides with the nightly release"
 
-            REMOTE_BRANCHES="$(git ls-remote https://github.com/leanprover-community/mathlib4.git nightly-testing-"$MOST_RECENT_NIGHTLY")"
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-release.yml
```diff
@@ -239,7 +239,7 @@ jobs:
           git config user.name "leanprover-community-mathlib4-bot"
           git config user.email "leanprover-community-mathlib4-bot@users.noreply.github.com"
 
-          if git branch -r | grep -q "nightly-testing-${MOST_RECENT_NIGHTLY}"; then
+          if git ls-remote --heads --tags origin | grep -q "refs/\(heads\|tags\)/nightly-testing-${MOST_RECENT_NIGHTLY}"; then
```
nomeata, line 242: Not sure it matter, but with the ever-growing list of tags and heads, maybe
```suggestion
          if git ls-remote --heads --tags --exit-code origin "nightly-testing-${MOST_RECENT_NIGHTLY}" >/dev/null; then
```
instead?

Also file says “branch“ when it should be “tag“ in a few more places.

nomeata (approved):

PR #3200, opened 2024-01-19
chore: CI creates lean-pr-testing-NNNN branches at Std too

kim-em:
Currently we create `lean-pr-testing-NNNN` branches at Mathlib automatically for each Lean PR.

We don't automatically create one at Std; mostly simply because Std fails less often, so it has been okay to do this manually as needed. It is conceptually simpler, however, if this is done uniformly.

This PR:
* does not proceed with Std/Mathlib CI unless the appropriate `nightly-testing-YYYY-MM-DD` tag exists at Std (like it already doesn't proceed if that tag is missing at Mathlib)
* creates `lean-pr-testing-NNNN` branches at Std
* when it creates `lean-pr-testing-NNNN` branches at Mathlib, updates the Std dependency to use the `lean-pr-testing-NNNN` branch at Std

- [x] depends on #3199

Note that because most users do not have write access at Std, in order to make updates to `lean-pr-testing-NNNN` branches there they will need to make PRs. These will be merged with a very low bar, and feel free to ping me for assistance on this. If this is annoying we will automate. Also, frequent contributors to Lean may ask @digama0 or @joehendrix for write access in order to easily work on these branches.

This PR requires that we have a secret here with write access at Std. I'm arranging that [on zulip](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/bot.20access/near/416686090).

I will update the documentation at https://leanprover-community.github.io/contribute/tags_and_branches.html to reflect these changes when they are merged.

### DIFF
--- a/.github/workflows/pr-release.yml
+++ b/.github/workflows/pr-release.yml
@@ -241,17 +241,17 @@ jobs:
           ref: nightly-testing
           fetch-depth: 0 # This ensures we check out all tags and branches.
 
-      - name: Check if branch exists
+      - name: Check if tag exists
         if: steps.workflow-info.outputs.pullRequestNumber != '' && steps.ready.outputs.mathlib_ready == 'true'
-        id: check_std_branch
+        id: check_std_tag
         run: |
           git config user.name "leanprover-community-mathlib4-bot"
           git config user.email "leanprover-community-mathlib4-bot@users.noreply.github.com"
 
-          if git ls-remote --heads --tags origin | grep -q "refs/\(heads\|tags\)/nightly-testing-${MOST_RECENT_NIGHTLY}"; then
+          if git ls-remote --heads --tags --exit-code origin "nightly-testing-${MOST_RECENT_NIGHTLY}" >/dev/null; then
             BASE="nightly-testing-${MOST_RECENT_NIGHTLY}"
           else
-            echo "This shouldn't be possible: couldn't find a 'nightly-testing-${MOST_RECENT_NIGHTLY}' branch at Std. Falling back to 'nightly-testing'."
+            echo "This shouldn't be possible: couldn't find a 'nightly-testing-${MOST_RECENT_NIGHTLY}' tag at Std. Falling back to 'nightly-testing'."
             BASE=nightly-testing
           fi
 


nomeata (commented):
LGTM. I wonder if the “`"This shouldn't be possible”` code path should simply error: We expect the tag to exist, so it can just fail and crash if it doesn’t

nomeata (approved):

.github/workflows/pr-release.yml
```diff
@@ -248,20 +314,20 @@ jobs:
...
           echo "Using base tag: $BASE"
 
-          git checkout "$BASE"
-
           EXISTS="$(git ls-remote --heads origin lean-pr-testing-${{ steps.workflow-info.outputs.pullRequestNumber }} | wc -l)"
           echo "Branch exists: $EXISTS"
           if [ "$EXISTS" = "0" ]; then
             echo "Branch does not exist, creating it."
-            git checkout -b lean-pr-testing-${{ steps.workflow-info.outputs.pullRequestNumber }}
+            git switch -c lean-pr-testing-${{ steps.workflow-info.outputs.pullRequestNumber }} "$BASE"
```
nomeata, line 321: Hmm turns out
```
git switch -c foo base
```
works if `base` is a local tag, but not if it is a (remote) branch.

The old code (`git checkout base`) would work for a (remote) branch, but not for a tag. Kinda annoying.

nomeata: Maybe we need to parse this output to go onto a common path in both cases

```
$ git ls-remote --heads --tags --exit-code https://github.com/leanprover-community/mathlib4 nightly-testing-2024-01-22
7d803278d1b5978fccc14f992d9ffb692077ed93	refs/heads/nightly-testing-2024-01-22
$ git ls-remote --heads --tags --exit-code https://github.com/leanprover/std4 nightly-testing-2024-01-22
c15b5a8f04e84bc7aab8b1062a4a80de3046865f	refs/tags/nightly-testing-2024-01-22
```

Alternatively, we can once manually convert all `nightly-testing-YYYY-MM-DD` branches to tags on mathlib4. Maybe easier.

nomeata: Possible fix in https://github.com/leanprover/lean4/pull/3208

PR #3201, opened 2024-01-19
feat: elidible subterms

mhuisi:
This PR adds two new delaboration settings: `pp.deepTerms : Bool` (default: `true`) and `pp.deepTerms.threshold : Nat` (default: `20`). 

Setting `pp.deepTerms` to `false` will make the delaborator terminate early after `pp.deepTerms.threshold` layers of recursion and replace the omitted subterm with the symbol `⋯` if the subterm is deeper than `pp.deepTerms.threshold / 4` (i.e. it is not shallow). To display the omitted subterm in the InfoView, `⋯` can be clicked to open a popup with the delaborated subterm.

<details>
  <summary>InfoView with pp.deepTerms set to false (click to show image)</summary>

  image
</details>

### Implementation

- The delaborator is adjusted to use the new configuration settings and terminate early if the threshold is exceeded and the corresponding term to omit is shallow.
- To be able to distinguish `⋯` from regular terms, a new constructor `Lean.Elab.Info.ofOmissionInfo`  is added to `Lean.Elab.Info` that takes a value of a new type `Lean.Elab.OmissionInfo`.
  - `ofOmissionInfo` is needed in `Lean.Widget.makePopup` for the `Lean.Widget.InteractiveDiagnostics.infoToInteractive` RPC procedure that is used to display popups when clicking on terms in the InfoView. It ensures that the expansion of an omitted subterm is delaborated using `explicit := false`, which is typically set to `true` in popups for regular terms.
  - Several `Info` widget utility functions are adjusted to support `ofOmissionInfo`.
- The list delaborator is adjusted with special support for `⋯` so that long lists `[x₁, ..., xₖ, ..., xₙ]` are shortened to `[x₁, ..., xₖ, ⋯]`.


nomeata:
Neat! 

Just based on the PR description, for lists and arrays, did you consider putting the ellipsis in the middle, in case the user cares more about what's at the end?

mhuisi:
One of the reasons that we are implementing this setting is because complete delaboration takes a significant amount of time for some users with especially large terms - it's not just for readability. Putting it somewhere in the middle would require us to traverse the full term, erasing this benefit of the option. 

I'm also not sure if it's easy to implement just for lists with the existing delaboration architecture (whether a term should be omitted is controlled by the function steering the delaboration, not individual delaborators).

nomeata:
That makes sense, thanks for the explanation!

Vtec234 (approved):
The implementation looks sensible. It's a bit sad that we have to add `OmissionInfo` to `Elab.Info`. I think this is because delab-info is starting to diverge further from elab-info (you'll never get `OmissionInfo` from the elaborator, right?). At one point I believe we discussed splitting out a separate `Delab.Info`. It may be worth considering again.

src/Lean/Server/FileWorker/WidgetRequests.lean
```diff
@@ -53,18 +53,25 @@ def makePopup : WithRpcRef InfoWithCtx → RequestM (RequestTask InfoPopup)
...
         | Elab.Info.ofTermInfo ti =>
-          let ti ← ppExprTagged ti.expr (explicit := true)
-          -- remove top-level expression highlight
-          pure <| some <| match ti with
-            | .tag _ tt => tt
-            | tt => tt
+          pure <| some <| ← ppExprTaggedWithoutTopLevelHighlight ti (explicit := true)
+        | Elab.Info.ofOmissionInfo { toTermInfo := ti } =>
+          -- Omitted terms are simply to be expanded, not printed explicitly.
+          pure <| some <| ← ppExprTaggedWithoutTopLevelHighlight ti (explicit := false)
```
Vtec234, line 59:
```suggestion
          -- Keep the top-level tag so that users can also see the explicit version
          -- of the omitted term.
          pure <| some <| ← ppExprTagged ti.expr (explicit := false)
```
I think you want to keep the top-level tag here. Consider
```lean
set_option pp.deepTerms false in
set_option pp.maxTermDepth 2 in
#check (@id Nat 10).succ.succ.succ
```
In this term `@id Nat 10` is elided, and hovering over `⋯` shows `id 10` with the tag stripped. Then there is no way to see the explicit version `@id Nat 10`.

semorrison:
Added `full-ci` so I can test the impact on the REPL and lean-training-data. No need to wait for me to report back, I'm sure this will be an easy adaptation.

mhuisi:
> The implementation looks sensible. It's a bit sad that we have to add `OmissionInfo` to `Elab.Info`. I think this is because delab-info is starting to diverge further from elab-info (you'll never get `OmissionInfo` from the elaborator, right?). At one point I believe we discussed splitting out a separate `Delab.Info`. It may be worth considering again.

I had the same thought when implementing this. However, it seems that all other `Info` constructors are used in the elaborator, and so I decided that splitting out a separate type for a single field may not be worth it. Regardless, this is worth revisiting if it comes up again.

> At one point I believe we discussed splitting out a separate `Delab.Info`.

Do you remember where or why this was discussed?

src/Lean/Server/FileWorker/WidgetRequests.lean
mhuisi: Good catch!

src/Init/NotationExtra.lean
```diff
@@ -170,16 +170,23 @@ See [Theorem Proving in Lean 4][tpil4] for more information.
...
-  | `($(_) $x [])      => `([$x])
-  | `($(_) $x [$xs,*]) => `([$x, $xs,*])
-  | _                  => throw ()
+  | `($(_) $x $tail) =>
+    match tail with
+    | `([])      => `([$x])
+    | `([$xs,*]) => `([$x, $xs,*])
+    | `(⋯)       => `([$x, $tail]) -- Unexpands to `[x, y, z, ⋯]` for `⋯ : List α`
+    | _          => throw ()
+  | _ => throw ()
```
kmill, line 189: For long lists, I guess this unexpander has been yielding quadratic complexity. Maybe eventually we should have a delaborator instead, so it can allocate the array all at once. That could also implement Joachim's suggestion to show the first and last elements of a long list. That's not for this PR though.

src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -40,6 +40,7 @@ structure Context where
   openDecls      : List OpenDecl
   inPattern      : Bool := false -- true when delaborating `match` patterns
   subExpr        : SubExpr
+  depth          : Nat := 0
```
kmill, line 43: Let's add a docstring. Something like this?
```suggestion
  /-- Current recursion depth for delaborator. Used by the `pp.deepTerms` option to
  omit deep terms. -/
  depth          : Nat := 0
```

src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -216,22 +219,46 @@ where
...
+
+def withIncDepth (act : DelabM α) : DelabM α := fun ctx =>
+  act { ctx with depth := ctx.depth + 1 }
+
+def isMaxDepthReached : DelabM Bool := do
+  let maxDepthCheckEnabled := ! (← getPPOption getPPDeepTerms)
+  let maxDepth ← getPPOption getPPMaxTermDepth
+  let depth := (← read).depth
+
+  return maxDepthCheckEnabled && depth > maxDepth
```
kmill, line 249: This function is reading both pretty printer options every time. I would suggest modifying it in the following way (assuming we switch to `pp.omitDeepTerms`), and also adding a docstring.
```lean
/--
Returns `true` if the delaborator should omit a term and replace it with `⋯`.
-/
def isMaxDepthReached : DelabM Bool := do
  unless (← getPPOption getPPOmitDeepTerms) do return false
  let maxDepth ← getPPOption getPPMaxTermDepth
  let depth := (← read).depth
  return depth > maxDepth
```
(I also wouldn't find it any less readable if the last two lines were coalesced into `return (← read).depth > maxDepth`, but that's a very minor suggestion.)

src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -240,6 +267,9 @@ partial def delabFor : Name → Delab
     <|> if k.isAtomic then failure else delabFor k.getRoot
 
 partial def delab : Delab := do
+  if ← isMaxDepthReached then
```
kmill, line 270: For atomic expressions, it seems to me that there's no need to replace them with `⋯`. There's a chance an atomic expression has a pretty printer that does something complicated (like print a type ascription), but then that type ascription will likely get a `⋯`.

```suggestion
  if ← pure !e.isAtomic <&&> isMaxDepthReached then
```

In any case, I think for consistency we should move this to after `checkSystem`.

kmill (commented):
This is a nice feature to have, and overall it looks like a good implementation.

There are two things I'd like to see before merging:
1. Adjusting the name. I think something like `pp.omitDeepTerms` with the opposite truth value would be clearer. Pretty printing deep terms is the natural default, and omitting deep terms seems like a special setting.
2. Adding a test file. It'd be nice to see that the feature can be turned on and off, and that it can handle long lists. (Probably best to set the maxDepth low.) For test files, Sebastian suggested to me that we are using module comments to say what things are meant to be testing. (`lean/run/rawStrings.lean` is one example)

src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -251,7 +281,7 @@ partial def delab : Delab := do
     else
       return ← annotateTermInfo (← ``(_))
```
kmill, line 282: If we wanted to adjust `pp.proofs` to use this new syntax, I suppose this `if`/`else` could be replaced with something like
```lean
    let pf ← omission
    if ← getPPOption getPPProofsWithType then
      let stx ← withType delab
      return ← `(($pf : $stx)) >>= annotateCurPos
    else
      return pf
```
(That's untested -- I'm just wanting to record how big of a change this would be.)

src/Lean/PrettyPrinter/Delaborator/Options.lean
```diff
@@ -125,6 +125,16 @@ register_builtin_option pp.instanceTypes : Bool := {
   group    := "pp"
   descr    := "(pretty printer) when printing explicit applications, show the types of inst-implicit arguments"
 }
+register_builtin_option pp.deepTerms : Bool := {
+  defValue := true
+  group    := "pp"
+  descr    := "(pretty printer) display deeply nested terms"
+}
```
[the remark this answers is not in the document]
nomeata: Hmm, double negation has its own downsides, and `omit` sounds like a negation, though. `pp.truncateTerms` maybe?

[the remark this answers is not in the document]
kmill: I'm not seeing a double negation here -- `pp.omitDeepTerms` being true means that deep terms are omitted. "Deep" doesn't seem like a negative quality to me.

I think that the concept of depth is important here; with `pp.truncateTerms` I wouldn't be able to guess what it's doing, or what it means to truncate a term (I might guess it would mean print it as a string and then use "⋯" if it exceeds some number of characters, which is very different).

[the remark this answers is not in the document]
digama0: I think it is more consistent to keep the polarity as is; other pp options like `pp.implicit` -> `pp.explicit` have had their polarities changed so that true = more printing. In lean 3 you would have to set some options true and others false to get `pp.all` behavior a la carte, which was confusing.

I have no opinion regarding the name of the option.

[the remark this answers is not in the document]
kmill: One difference between this new option and ones like `pp.explicit` or `pp.proofs` is that turning this one on results in terms that cannot be elaborated. These other options make the pretty printer create terms that are more likely to be successfully elaborated.

src/Lean/PrettyPrinter/Delaborator/Basic.lean
mhuisi: I generalized this idea to "shallow expressions": The delaborator now never omits subterms if they have an `approxDepth <= maxTermDepth / 4`.
Without this kind of generalization, `[1, 2, 3, 4, 5]` would print as `[1, 2, 3, 4, ⋯, ⋯]`, with `⋯` referring to `5` and `[]` respectively because both terms are not atomic, but also exceed the `maxTermDepth` at that point.

Please let me know if you want to keep this generalization.

src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -216,22 +220,54 @@ where
...
+  modify fun s => { s with infos := s.infos.insert pos info }
+where
+  mkOmissionInfo stx e := return {
+    toTermInfo := ← addTermInfo.mkTermInfo stx e (isBinder := false)
+  }
+
+def withIncDepth (act : DelabM α) : DelabM α := fun ctx =>
+  act { ctx with depth := ctx.depth + 1 }
+
+def isMaxDepthReached (e : Expr) : DelabM Bool := do
```
kmill, line 245: Should we rename this to something more general, such as `shouldOmitExpr`? In any case, let's add a docstring.
```suggestion
/--
Returns `true` if, at the current depth, we should omit the term and use `⋯` rather than delaborating it.
This function only returns `true` if `pp.omitDeepTerms` is set to `true`.
It also contains a heuristic to allow "shallow terms" to be delaborated, even if they appear deep in an expression,
which prevents terms such as atomic expressions or `OfNat.ofNat` literals from being delaborated as `⋯`.
-/
def shouldOmitExpr (e : Expr) : DelabM Bool := do
```

src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -216,22 +220,54 @@ where
...
+    !isMaxedOutApproxDepth && approxDepth <= maxDepth/4 - depthExcess
+
+  return depthExcess > 0 && !isShallowExpression
+
 def annotateTermInfo (stx : Term) : Delab := do
   let stx ← annotateCurPos stx
   addTermInfo (← getPos) stx (← getExpr)
   pure stx
 
+def omission : Delab := do
```
kmill, line 265: Docstring suggestion:
```suggestion
/--
Delaborates the current expression as `⋯` and attaches omission info.
This kind of info influences how the term is delaborated when hovered over.
-/
def omission : Delab := do
```

src/Lean/PrettyPrinter/Delaborator/Options.lean
```diff
@@ -125,6 +125,16 @@ register_builtin_option pp.instanceTypes : Bool := {
...
 }
+register_builtin_option pp.omitDeepTerms : Bool := {
+  defValue := false
+  group    := "pp"
+  descr    := "(pretty printer) omit deeply nested terms, replacing them with `⋯`"
+}
+register_builtin_option pp.maxTermDepth : Nat := {
+  defValue := 20
+  group    := "pp"
+  descr    := "(pretty printer) the maximum expression depth used by `pp.omitDeepTerms` when replacing terms with `⋯`"
```
kmill, line 136: Given that there's a heuristic, maybe we reword this?
```suggestion
  descr    := "(pretty printer) when `pp.omitDeepTerms` is true, the depth at which terms begin being replaced with `⋯`"
```

src/Lean/PrettyPrinter/Delaborator/Options.lean
```diff
@@ -125,6 +125,16 @@ register_builtin_option pp.instanceTypes : Bool := {
   group    := "pp"
   descr    := "(pretty printer) when printing explicit applications, show the types of inst-implicit arguments"
 }
+register_builtin_option pp.omitDeepTerms : Bool := {
+  defValue := false
+  group    := "pp"
+  descr    := "(pretty printer) omit deeply nested terms, replacing them with `⋯`"
+}
+register_builtin_option pp.maxTermDepth : Nat := {
```
kmill, line 133: Marc and I were discussing renaming this for discoverability. Some options:
* `pp.omitDeepTerms.threshold`, which I think is a nice name for the concept.
* `pp.omitDeepTerms.maxDepth`, which is parallel to `pp.raw.maxDepth`, but it's not strictly a *max* depth in this case due to the shallow terms heuristic.

kmill (commented):
The heuristic is a good idea for handling some potential annoyances when omitting terms. Maybe in the future we might have some additional pretty printing options to control it, but I believe it's better to add them later if they're requested.

Once we have docstrings on the API functions and we decide whether or not to rename `pp.maxTermDepth`, then I think it'll be ready.

src/Lean/PrettyPrinter/Delaborator/Options.lean
nomeata: Either is probably fine, if you need comments from the peanut gallery: I buy your arguments in favor of `threshold`

src/Lean/PrettyPrinter/Delaborator/Options.lean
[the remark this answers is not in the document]
digama0: @kmill I am confused by your statement. Assuming that the polarity is that of `pp.deepTerms`, i.e. false means use `...` and true means to show everything, it would not be any different from `pp.explicit` or `pp.proofs`, more true equals more printing equals more likely to be successfully elaborated. That's exactly my point.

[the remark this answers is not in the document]
kmill: @digama0 I'm saying that this feature is different because it has to do with a feature that makes terms impossible to elaborate, rather than just less likely.

It's true though that `pp.proofs` being off makes elaboration vanishingly unlikely.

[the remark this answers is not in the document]
digama0: You could use `_` instead of `...` to make it at least syntactically correct, although using a special syntax is also useful for making it clear what has happened.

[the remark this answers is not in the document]
Kha: Indeed this discussion makes me wonder whether `pp.proofs` should be reimplemented to use the same UX as elision by depth.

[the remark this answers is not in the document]
Kha: > I think it is more consistent to keep the polarity as is; other pp options like `pp.implicit` -> `pp.explicit` have had their polarities changed so that true = more printing

I would agree with that, indeed that was my reasoning behind introducing `pp.explicit`

kmill (approved):
Looks good to me!

(The rationale for `pp.deepTerms` is that, even though we think `pp.omitDeepTerms` is a clearer name, it's important to follow Sebastian's design for the pp options and be consistent with the analogous `pp.proofs`. Furthermore, this option will likely soon become a closer analogue since it too should use similar logic and use the omission syntax.)

PR #3202, opened 2024-01-20
refactor: skip `mkExpectedTypeHint` in `replaceTargetEq`

nomeata:
Looking at the proof terms produced by `rewrite` I noticed an extra `id`
call without further explanation. I’m trying to find out whether

 * It’s strictly needed to help the type checker (in which case I’d like
   to add a test case).
 * It’s not strictly needed, but helps with performance (in which case
   I’ll add a comment)
 * Turns out to be not needed and maybe even has a measurable
   performance hit (in which case maybe it should be removed).

### DIFF
--- a/src/Lean/Meta/Tactic/Replace.lean
+++ b/src/Lean/Meta/Tactic/Replace.lean
@@ -24,9 +24,7 @@ def _root_.Lean.MVarId.replaceTargetEq (mvarId : MVarId) (targetNew : Expr) (eqP
     let mvarNew  ← mkFreshExprSyntheticOpaqueMVar targetNew tag
     let target   ← mvarId.getType
     let u        ← getLevel target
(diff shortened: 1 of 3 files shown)


nomeata:
Mathlib builds, and the bench results at http://speed.lean-fro.org/mathlib4/compare/ed8d927b-e006-4c9b-a660-00049b15865c/to/41aeb673-3fe0-4471-afe0-69e16f6f4225 are mostly green (-0.754 % smaller oleans, -0.3% build wall-time/instructions).

So maybe this `mkExpectedTypeHint` is not that useful?

nomeata:
Discussed this with Leo and learned that while it’s not always possible to give hard evidence for why these type hints are absolutely needed, it’s probably better to be conservative here and leave it in, at least unless we have evidence that they hurt.

PR #3208, opened 2024-01-22
chore: pr-release.yaml: create testing branch from tag or branch

nomeata:


nomeata:
At this point the better fix is probably to tell people to upgrade their PR to a newer nightly tag.

PR #3210, opened 2024-01-23
refactor: cleanup `simp` types and `pre` and `post` method semantics

leodemoura:
Before this commit, `Simproc`s were defined as `Expr -> SimpM (Option Step)`, where `Step` is inductively defined as follows:
```
inductive Step where
  | visit : Result → Step
  | done  : Result → Step
```
Here, `Result` is a structure containing the resulting expression and a proof demonstrating its equality to the input. Notably, the proof is optional; in its absence, `simp` assumes reflexivity.

A simproc can:
- Fail by returning `none`, indicating its inapplicability. In this case, the next suitable simproc is attempted, along with other simp extensions.
- Succeed and invoke further simplifications using the `.visit`
constructor. This action returns control to the beginning of the
simplification loop.
- Succeed and indicate that the result should not undergo further
simplifications. However, I find the current approach unsatisfactory, as it does not align with the methodology employed in `Transform.lean`, where we have the type:

```
inductive TransformStep where
  /-- Return expression without visiting any subexpressions. -/
  | done (e : Expr)
  /--
  Visit expression (which should be different from current expression) instead.
  The new expression `e` is passed to `pre` again.
  -/
  | visit (e : Expr)
  /--
  Continue transformation with the given expression (defaults to current expression).
  For `pre`, this means visiting the children of the expression.
  For `post`, this is equivalent to returning `done`. -/
  | continue (e? : Option Expr := none)
```
This type makes it clearer what is going on. The new `Simp.Step` type is similar but use `Result` instead of `Expr` because we need a proof.

Other modifications:
- The `pre` and `post` methods have now type `Simproc`.
- `simprocs` field at `Methods` was deleted.
- Added `andThen` combinator for `Simproc`.
- Cleanup `simpLoop` semantics.

Todo: cleanup `reduceStep`.

Recording adapation PRs here:
- https://github.com/leanprover/std4/pull/571 (working)
- https://github.com/leanprover-community/aesop/pull/98 (working but with one test regression)
- https://github.com/leanprover-community/mathlib4/pull/10133 (still WIP)

### DIFF
--- a/src/Lean/Elab/Tactic/Simp.lean
+++ b/src/Lean/Elab/Tactic/Simp.lean
@@ -137,7 +137,14 @@ inductive ResolveSimpIdResult where
   | none
   | expr (e : Expr)
   | simproc (declName : Name)
-  | ext  (ext₁ : SimpExtension) (ext₂ : Simp.SimprocExtension)
+  /--
+  Recall that when we declare a `simp` attribute using `register_simp_attr`, we automatically
+  create a `simproc` attribute. However, if the user creates `simp` and `simproc` attributes
+  programmatically, then one of them may be missing. Moreover, when we write `simp [seval]`,
+  we want to retrieve both the simp and simproc sets. We want to hide from users that
+  `simp` and `simproc` sets are stored in different data-structures.
+  -/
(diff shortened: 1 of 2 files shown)


leodemoura:
The `register_simp_attr <id>` command now also creates a simproc set.

PR #3215, opened 2024-01-24
feat: enforce indentation of command contents

Kha:
Only affects parts using `colGt/Ge`, e.g. arguments to functions

See test case for motivation

### DIFF
--- a/tests/lean/decreasing_by.lean.expected.out
+++ b/tests/lean/decreasing_by.lean.expected.out
@@ -12,7 +12,7 @@ The arguments relate at each recursive call as follows:
 1) 66:29-43 = ?
 2) 66:46-62 ? _
 Please use `termination_by` to specify a decreasing measure.
-decreasing_by.lean:81:13-83:3: error: unexpected token 'end'; expected '{' or tactic
+decreasing_by.lean:83:0: error: expected '{' or indented tactic sequence
 decreasing_by.lean:81:0-81:13: error: unsolved goals
 n m : Nat
 ⊢ (invImage (fun a => PSigma.casesOn a fun n snd => (n, snd)) Prod.instWellFoundedRelationProd).1


Kha:
Sadly this breaks completely unindented tactic blocks as evidenced by the tests, which we might not be ready to do
```
theorem tst0' {p q : Prop } (h : p ∨ q) : q ∨ p := by
induction h
```
The two non-tactic cases that I would have less qualms about breaking:
```
def run1 (i : Nat) (n : Nat) (xs : List Nat) : Nat :=
n.repeat (fun r =>
  dbg_trace ">> [{i}] {r}";
  xs.foldl (fun a b => a + b) r)
0
```
```
#synth AppendList
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16]
[200, 201, 202, 203, 204, 205, 206, 207, 208, 209, 210, 211]
[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16,
 200, 201, 202, 203, 204, 205, 206, 207, 208, 209, 210, 211]
```

semorrison:
I would be happy to make those examples break, and the error message "expected indented tactic sequence" is clear and actionable.

Kha:
@kmill @digama0 Could I get your input on potential interactions of this PR with the [empty tactic sequences proposal](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/parser.20allowing.20no.20tactics.20after.20by)? The issue I see is that unindented input like above yields an actionable error message with this PR as mentioned by @semorrison but would presumably result in `unexpected token "induction", expected command` with the combination of the two as we would successfully parse an empty `by` instead. I believe this would be terrible UX for beginners who often don't consider proper indentation yet and nothing in the error message would point them to it anymore.

digama0:
My view is that if we go forward with this then we should recognize that the style of no-indent after `:=` is basically no longer legal. This isn't entirely true, but the examples you have shown above mean that we can't really recommend it at all.

The issue of terrible parser error messages is a hard one and probably needs significant additional work to deal with. This wouldn't be the first time a slight mis-indentation results in a completely crazy error message.

nomeata:
Looking at this PR again, I think the possible benefits – in particular better error messages – are high. Any syntax that uses `colGt` probably intents some form of indentation, so it seems reasonable to always have a something.

I’ve revived this in https://github.com/leanprover/lean4/pull/13229.

PR #3216, opened 2024-01-24
fix: broken internal links in the docs

david-christiansen:
I deleted internal links that seemed to have the character of "TODO". I think that the residual TODO is of little value, given that we plan a big revamp and revision soon anyway, but I could do it some other way as well.


david-christiansen:
I'll need to debug the Netlify setup for PRs tomorrow, or perhaps just switch to a local on-disk link checker script.

david-christiansen:
OK, I think this is ready to go. It'll fix the broken links on the generated site and help prevent future issues.

Kha (approved):

PR #3221, opened 2024-01-26
doc: add doc for FileMap

joneugster:


joneugster:
I'm not sure if Pr's only adding documentation, like this one, are generally well received. If you don't have the time/interest to review such Pr's please just close it!

src/Lean/Data/Position.lean
```diff
@@ -29,9 +29,16 @@ instance : ToExpr Position where
...
+/-- Content of a file together with precalculated positions of newlines. -/
 structure FileMap where
+  /-- The content of the file. -/
   source    : String
+  /-- The positions of newline characters.
+  The first entry is always `0` and the last always the index of the last character.
+  In particular, if the last character is a newline, that index will appear twice. -/
   positions : Array String.Pos
+  /-- The line indices. Has the same length as `positions` and is always of the form
+  `[1, 2, …, n-1, n-1]`. -/
```
david-christiansen, line 41: I'm not sure I understand this specification here. There's a few things about it that I find confusing:

1. (minor) It uses list syntax for an array
2. I'm not sure what a line index is. Can you define the term? Or what about calling it "line number associated with each mapped newline"?

david-christiansen:
Thanks for the PR!

src/Lean/Data/Position.lean
joneugster: updated!

One point is that `lines` seems to be pretty redundant, as far as I can tell.

Kha: Yes, if you want to do a follow-up PR removing it, that would be a strict improvement I'd say!

david-christiansen (approved):

PR #3223, opened 2024-01-26
chore: move `ac_rfl` to Std

### LINKED ISSUE #3198: RFC: Move `ac_rfl` tactic to Std
### Proposal

The `ac_rfl` tactic is not used at all within Lean itself. I propose to move it to Std so it could be improved to a more useful tactic.

Users would not be affected much since the tactic would still be available in Std. There are no maintainability issues for Lean but there is an additional burden for Std to repurpose the tactic. The main beneficiaries would be Std developers who are working on the use of AC lemmas for non-arithmetic operations and folds, for example.

### Community Feedback

This came up in a discussion on [std4#535](https://github.com/leanprover/std4/pull/535). It was proposed there to make small adjustments to some classes in Lean. Since these classes are only used by `ac_rfl` I floated the idea of moving the tactic to Std since these classes and `ac_rfl` aren't used anywhere in Lean itself. This was confirmed by an experiment removing `ac_rfl` on a fork of Lean: https://github.com/fgdorais/lean4/tree/ac_rfl

### Impact

fgdorais:
Move `ac_rfl` tactic to Std.

Companion PRs [std4#573](https://github.com/leanprover/std4/pull/573) and [mathlib4#10152](https://github.com/leanprover-community/mathlib4/pull/10152).

Closes #3198


fgdorais:
WIP

fgdorais:
awaiting-review

leodemoura:
Thanks for the PR, but we have decided to keep tactics in core.

Issue #3192, opened 2024-01-17
LLVM vs C backend performance


hargoniX:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to mathlib4 or std4.

### Description

Currently the LLVM backend generates IR that has worse performance than the C backend after optimization. It appears that this issue can be deduced to a failure of inlining Lean runtime
functions for manipulating `lean_object`s.

### Steps to Reproduce

1. Compile Lean with the C and the LLVM backend enabled
2. Inspect the IR of the const_fold.lean bench test generated by LLVM vs generated by C before and after `-O3`
4. Some lean.h functions will be inlined in the C variant but not in the LLVM variant

**Expected behavior:** The same functions should be inlined

**Actual behavior:** They are not

### Impact
As long as this is not fixed the LLVM backend will most likely not be able to be competetive with the C one in terms of performance.

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

Kha:
@hargoniX In #3225 it sounds like this specific issue is likely resolved, yes?

hargoniX:
Yes, in general the performance diff between the LLVM and C backend is very miniscule but still measurable. We investigated the IR for a while and didn't really see any obvious differences in the logic it's implementing so the remaining diff is most likely due to different basic blocks, binary layouts etc.

PR #3225, opened 2024-01-27
fix: type mismatches in the LLVM backend

hargoniX:
Debugged and authored in collaboration with @bollu.

This PR fixes several performance regressions of the LLVM backend compared to the C backend
as described in #3192. We are now at the point where some benchmarks from `tests/bench` achieve consistently equal and sometimes ever so slightly better performance when using LLVM instead of C. However there are still a few testcases where we are lacking behind ever so slightly.

The PR contains two changes:
1. Using the same types for `lean.h` runtime functions in the LLVM backend as in `lean.h` it turns out that:
  a) LLVM does not throw an error if we declare a function with a different type than it actually has. This happened on multiple occasions here, in particular when the function used `unsigned`, as it was wrongfully assumed to be `size_t` sized.
  b) Refuses to inline a function to the call site if such a type mismatch occurs. This means that we did not inline important functionality such as `lean_ctor_set` and were thus slowed down compared to the C backend which did this correctly.
2. While developing this change we noticed that LLVM does treat the following as invalid: Having a function declared with a certain type but called with integers of a different type. However this will manifest in completely nonsensical errors upon optimizing the bitcode file through `leanc` such as:
```
error: Invalid record (Producer: 'LLVM15.0.7' Reader: 'LLVM 15.0.7')
```
Presumably because the generate .bc file is invalid in the first place. Thus we added a call to `LLVMVerifyModule` before serializing the module into a bitcode file. This ended producing the expected type errors from LLVM an aborting the bitcode file generation as expected.

We manually checked each function in `lean.h` that is mentioned in `EmitLLVM.lean` to make sure that all of their types align correctly now.

Quick overview of the fast benchmarks as measured on my machine, 2 runs of LLVM and 2 runs of C to get a feeling for how far the averages move:
- binarytrees: basically equal performance
- binarytrees.st: basically equal performance
- const_fold: equal if not slightly better for LLVM
- deriv: LLVM has 8% more instructions than C but same wall clock time
- liasolver: basically equal performance
- qsort: LLVM is slower by 7% instructions, 4% time. We have identified why the generated code is slower (there is a store/load in a hot loop in LLVM that is not in C) but not figured out why that happens/how to address it.
- rbmap: LLVM has 3% less instructions and 13% less wall-clock time than C (woop woop)
- rbmap_1 and rbmap_10 show similar behavior
- rbmap_fbip: LLVM has 2% more instructions but 2% better wall time
- rbmap_library: equal if not slightly better for LLVM
- unionfind: LLVM has 5% more instructions but 4% better wall time

Leaving out benchmarks related to the compiler itself as I was too lazy to keep recompiling it from scratch until we are on a level with C.

Summing things up, it appears that LLVM has now caught up or surpassed the C backend in the microbenchmarks for the most part. Next steps from our side are:
- trying to win the qsort benchmark
- figuring out why/how LLVM runs more instructions for less wall-clock time. My current guesses would be measurement noise and/or better use of micro architecture?
- measuring the larger benchmarks as well

### DIFF
--- a/src/Lean/Compiler/IR/EmitLLVM.lean
+++ b/src/Lean/Compiler/IR/EmitLLVM.lean
@@ -103,6 +103,9 @@ def getDecl (n : Name) : M llvmctx Decl := do
 def constInt8 (n : Nat) : M llvmctx (LLVM.Value llvmctx) :=  do
     LLVM.constInt8 llvmctx (UInt64.ofNat n)
 
+def constInt64 (n : Nat) : M llvmctx (LLVM.Value llvmctx) :=  do
+    LLVM.constInt64 llvmctx (UInt64.ofNat n)
+
 def constIntSizeT (n : Nat) : M llvmctx (LLVM.Value llvmctx) :=  do
     LLVM.constIntSizeT llvmctx (UInt64.ofNat n)
 
@@ -172,7 +175,7 @@ def callLeanUnsignedToNatFn (builder : LLVM.Builder llvmctx)
   let retty ← LLVM.voidPtrType llvmctx
   let f ←   getOrCreateFunctionPrototype mod retty "lean_unsigned_to_nat"  argtys
   let fnty ← LLVM.functionType retty argtys
-  let nv ← LLVM.constInt32 llvmctx (UInt64.ofNat n)
+  let nv ← constIntUnsigned n
   LLVM.buildCall2 builder fnty f #[nv] name
 
 def callLeanMkStringFromBytesFn (builder : LLVM.Builder llvmctx)
@@ -228,9 +231,9 @@ def callLeanAllocCtor (builder : LLVM.Builder llvmctx)
   let fn ← getOrCreateFunctionPrototype (← getLLVMModule) retty fnName argtys
   let fnty ← LLVM.functionType retty argtys
 
-  let tag ← LLVM.constInt32 llvmctx (UInt64.ofNat tag)
-  let num_objs ← LLVM.constInt32 llvmctx (UInt64.ofNat num_objs)
-  let scalar_sz ← LLVM.constInt32 llvmctx (UInt64.ofNat scalar_sz)
+  let tag ← constIntUnsigned tag
+  let num_objs ← constIntUnsigned num_objs
+  let scalar_sz ← constIntUnsigned scalar_sz
   LLVM.buildCall2 builder fnty fn #[tag, num_objs, scalar_sz] name
 
 def callLeanCtorSet (builder : LLVM.Builder llvmctx)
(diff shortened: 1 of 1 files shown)


digama0:
Does LLVM have a typechecker, and could we use it to validate the output in the tests?

hargoniX:
The strongest thing that I know of is the mentioned validator that we now run per default on the .bc file to avoid producing a corrupted file. Maybe @bollu or @tobiasgrosser know of some additional tool that we could run?

tobiasgrosser:
@hargoniX, how did your flow look like before you started to run the validator? LLVM always calls the validator when loading a `.bc` or `.ll` file into any tool. I am surprised you did not run the validator yourself as part of your flow, as invalid IR is very likely to cause issues.

As allowing for invalid IR at an intermediate stage makes writing transformations easier and running the validator after each pass has a performance cost, LLVM does not run the validator after each pass. However, when debugging its really useful to run the validator.

hargoniX:
> LLVM always calls the validator when loading a .bc or .ll file into any tool.

This statement confuses me a little bit. Once we noticed the type mismatches we changed the types at the declaration of the function types but we forgot to also change the types of the numeric literals that are passed to the functions. This ended up producing things like the above:
```
error: Invalid record (Producer: 'LLVM15.0.7' Reader: 'LLVM 15.0.7')
```
When trying to run the optimization step on the produced `.bc` file with `leanc`. But upon adding the call to `LLVMVerifyModule` it produced proper type errors detailing which functions were expecting what errors etc.

tobiasgrosser:
Did you try to run `opt` on this file?

tobiasgrosser:
Who created this file? Did you export it into a .ll file and see if it validates?

src/Lean/Compiler/IR/EmitLLVM.lean
```diff
@@ -25,9 +25,13 @@ def leanMainFn := "_lean_main"
 
 namespace LLVM
 -- TODO(bollu): instantiate target triple and find out what size_t is.
-def size_tType (llvmctx : LLVM.Context) : IO (LLVM.LLVMType llvmctx) :=
+def size_tType (llvmctx : LLVM.Context) : BaseIO (LLVM.LLVMType llvmctx) :=
   LLVM.i64Type llvmctx
 
+-- TODO(bollu): instantiate target triple and find out what unsigned is.
```
Kha, line 31: We can probably replace this with a static assert that `unsigned` is two bytes because we're definitely not expecting it to ever be something else :)

Kha (approved):

tobiasgrosser:
I am happy to go ahead with this. I suggested further improvements, but those can be submitted in subsequent PRs as they are not directly related to this PR.

src/Lean/Compiler/IR/EmitLLVM.lean
digama0: *two* bytes?

Kha: Uh, that's the minimum size, but we'd rather expect four, yes

digama0: Relatedly: Can we please remove the usage of wobbly types from lean.h? I see no advantage to using `unsigned` over `uint32_t` and a risk of miscompilations if they ever differ.

hargoniX: Strong :+1: on that one from my side.

bollu: Strong +1 from me as well, this will eliminate all of the wobbliness from C in the backend code.

bollu: I suggest performing the refactor (either changing it to 4 bytes, or the more aggressive refactor suggested by @digama0) as a subsequent PR.

tobiasgrosser:
Can we merge this PR? We are currently polishing the LLVM backend further, and it would help to not have to stack PRs.

Kha:
Sorry, didn't have it on my radar anymore!

PR #3226, opened 2024-01-29
feat: lake: improved platform information & control

### LINKED ISSUE #2754: Build archives not differentiated by architecture
I couldn't help but notice these curious-looking file paths in the mathlib4 build:
```
info: Downloading proofwidgets/v0.0.10/linux-64.tar.gz
```
The suffix is unexpected - while we do not officially support any non-64 bit platform, we have supported ARM64 since before the cloud builds existed! If we need a new primitive for `LEAN_INSTALL_SUFFIX` to properly encode that, we should do it.

tydeu:
This combines a few platform-related changes:

* Add a ternary `platformIndependent` Lean configuration option to assert whether Lake should assume Lean code is platform-independent. If `true`, Lake will exclude platform-independent objects like external libraries or dynlibs created through `precompileModules` from module traces. If `false`, Lake will add the platform to module traces. If `none` (the default), Lake will retain the current behavior (modules are platform-dependent if and only if it depends on native objects).
* Use `System.Platform.target` from #3207 as the platform descriptor in Lake for the configuration file trace, the cloud release archive, and as the platform trace in Lean modules and native artifacts (e.g., object files, and static and shared libraries).
* Do not add the platform descriptor into custom build archive names (i.e., a user-set `buildArchive` configuration). This allows users to create cross-platform / platform-independent archives via a name override should they so desire.

Closes #2754.

### DIFF
--- a/src/lake/Lake/Config/LeanConfig.lean
+++ b/src/lake/Lake/Config/LeanConfig.lean
@@ -163,10 +163,25 @@ structure LeanConfig where
   -/
   backend : Backend := .default
   /--
-  Asserts Lake should assume the Lean code is platform-independent.
-  This means Lake will exclude platform-dependent elements (such as shared libraries)
-  from a module's trace and will not re-elaborate Lean code for different platforms.
-  Defaults to `false`.
+  Asserts whether Lake should assume Lean modules are platform-independent.
+
+  * If `false`, Lake will add `System.Platform.target` to the module traces
+  within the code unit (e.g., package or library). This will force Lean code
+  to be re-elaborated on different platforms.
+
+  * If `true`, Lake will exclude platform-dependent elements
+  (e.g., precompiled modules, external libraries) from a module's trace,
+  preventing re-elaboration on different platforms. Note that this will not
+  effect  modules outside the code unit in question. For example, a
+  platform-independent package which depends on a platform-dependent library
+  will still be platform-dependent.
+
+  * If `none`, Lake will construct traces as natural. That is, it will include
+  platform-dependent artifacts in the trace if they module depends on them,
+  but otherwise not force modules to be platform-dependent.
+
+  There is no check  for correctness here, so a configuration can lie
+  and Lake will not catch it. Defaults to `none`.
   -/
-  platformIndependent : Bool := false
+  platformIndependent : Option Bool := none
 deriving Inhabited, Repr
--- a/src/lake/Lake/Build/Common.lean
+++ b/src/lake/Lake/Build/Common.lean
@@ -16,6 +16,13 @@ namespace Lake
 
 /-! ## General Utilities -/
 
+/--
+Build trace for the host platform.
+If an artifact includes this in its trace, it is platform-dependent
+and will be rebuilt on different host platforms.
+-/
+def platformTrace := pureHash System.Platform.target
+
 /-- Check if the `info` is up-to-date by comparing `depTrace` with `traceFile`. -/
 @[specialize] def BuildTrace.checkUpToDate [CheckExists ι] [GetMTime ι]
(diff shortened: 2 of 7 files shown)


src/lake/Lake/Config/LeanConfig.lean
```diff
@@ -157,10 +157,17 @@ structure LeanConfig where
...
   -/
   backend : Backend := .default
+  /--
+  Asserts Lake should assume the Lean code is platform-independent.
+  This means Lake will not include platform-dependent elements (such as
+  shared libraries) in a module's trace and will not re-elaborate Lean code
+  for different platforms.
+  Defaults to `false`.
+  -/
+  platformIndependent : Bool := false
```
Kha, line 172: From the name and description I would expect .oleans even without shared libraries to get platform-specific traces by default but I don't think that's the case, is it? Should it be?

tydeu: No, it is not the case and I do not think it should be (at least as the default). It was not my intention to convey that with the docstring. What gave you that impression? Would "This means Lake will exclude platform-independent elements (such as shared libraries) from the module's trace (if they exist)" be better?

Kha: I don't know how to interpret "the Lean code is platform-independent" then. If the *compile-time* code is platform-dependent, it seems like the .olean hash should depend on the platform under any configuration. Asked differently, why exactly should setting `precompileModules` change the default assumption to "platform-dependent" when no external libraries are involved?

tydeu: I agree with your assessment that "If the output of compile-time code is platform-dependent, it seems like the .olean hash should depend on the platform under". The difficulty is that it is hard for Lake to know whether this is the case. Currently, Lake assumes not unless one of the inputs is naturally platform-dependent (e.g. a shared library). 

What I think you're saying here is that, with this option, we can avoid assumptions and just operate based on this flag? That is, if it is `false`, we add the platform to the module trace. Otherwise, we exclude platform-dependent hashes from the trace. Is that right?

tydeu: I have updated the `platformIndepedent` configuration to a ternary value. `none` will retain the pre-PR behavior, `false` will add the platform to the trace, and `true` will exclude native objects from the trace. I have also ensured that the Lake-provided native objects always include the platform in their trace. Hopefully, this will address this concern.

Kha: Sounds exhaustive, great!

semorrison:
Can we verify that
```
git checkout lean-pr-testing-3226
elan toolchain uninstall leanprover/lean4-pr-releases:pr-release-3226 # make sure we have the latest PR toolchain
lake exe cache get
```
works in Mathlib before merging this? It seems to not retrieve any oleans at present.

tydeu:
@semorrison I am confused, how would this prevent cache from downloading oleans? It is my understanding the cache uses it own custom traces for oleans. If this broke anything, I would think it be `lake build` rebuilding mathlib.

semorrison:
> @semorrison I am confused, how would this prevent cache from downloading oleans? It is my understanding the cache uses it own custom traces for oleans. If this broke anything, I would think it be `lake build` rebuilding mathlib.

I suspect it is just that you force pushed just recently, so no update to date oleans are available right now anyway. So not an indication of a problem, just a request that we run this check successfully at least once before merging!

(I don't want to have something hit `nightly` on release day that could conceivably break the cache, without checking!)

tydeu:
@semorrison I just tested (now the cache is uploaded) and `lake exe cache get && lake build` works for me. Does it work for you as well?

tydeu:
!bench

PR #3231, opened 2024-01-30
feat: set List.toArray equal to Array.mk

joehendrix:
`List.toArray l` is a natural simp normal form for arrays since it unfolds during compilation so that `Array` literals (which are defined as `List.toArray [...]` of an list literal) can unfold.  The current recursive `List.toArray` function makes reasoning about array literals more difficult than necessary however since it is a recursive function distinct from the constructor (though the recursive definition is used by interpreter and compiler).

This makes the two logically the same, but uses `csimp` lemma show equivalence so that compilation behavior is preserved.

As part of this, I removed a lemma that is equivalent to `Array.mk.injEq` and added a test case to check compiler output.

### DIFF
--- a/src/Init/Data/List/Basic.lean
+++ b/src/Init/Data/List/Basic.lean
@@ -77,55 +77,13 @@ theorem length_add_eq_lengthTRAux (as : List α) (n : Nat) : as.length + n = as.
 @[simp] theorem length_nil : length ([] : List α) = 0 :=
   rfl
 
-/-- Auxiliary for `List.reverse`. `List.reverseAux l r = l.reverse ++ r`, but it is defined directly. -/
-def reverseAux : List α → List α → List α
-  | [],   r => r
-  | a::l, r => reverseAux l (a::r)
-
-/--
-`O(|as|)`. Reverse of a list:
-* `[1, 2, 3, 4].reverse = [4, 3, 2, 1]`
-
-Note that because of the "functional but in place" optimization implemented by Lean's compiler,
-this function works without any allocations provided that the input list is unshared:
-it simply walks the linked list and reverses all the node pointers.
--/
-def reverse (as : List α) : List α :=
-  reverseAux as []
-
 theorem reverseAux_reverseAux_nil (as bs : List α) : reverseAux (reverseAux as bs) [] = reverseAux bs as := by
   induction as generalizing bs with
   | nil => rfl
   | cons a as ih => simp [reverseAux, ih]
 
-theorem reverseAux_reverseAux (as bs cs : List α) : reverseAux (reverseAux as bs) cs = reverseAux bs (reverseAux (reverseAux as []) cs) := by
-  induction as generalizing bs cs with
-  | nil => rfl
-  | cons a as ih => simp [reverseAux, ih (a::bs), ih [a]]
-
 @[simp] theorem reverse_reverse (as : List α) : as.reverse.reverse = as := by
   simp [reverse]; rw [reverseAux_reverseAux_nil]; rfl
-
-/--
-`O(|xs|)`: append two lists. `[1, 2, 3] ++ [4, 5] = [1, 2, 3, 4, 5]`.
-It takes time proportional to the first list.
--/
-protected def append : (xs ys : List α) → List α
-  | [],    bs => bs
-  | a::as, bs => a :: List.append as bs
-
-/-- Tail-recursive version of `List.append`. -/
-def appendTR (as bs : List α) : List α :=
-  reverseAux as.reverse bs
-
-@[csimp] theorem append_eq_appendTR : @List.append = @appendTR := by
-  apply funext; intro α; apply funext; intro as; apply funext; intro bs
-  simp [appendTR, reverse]
-  induction as with
-  | nil  => rfl
-  | cons a as ih =>
-    rw [reverseAux, reverseAux_reverseAux]
-    simp [List.append, ih, reverseAux]
 
 instance : Append (List α) := ⟨List.append⟩
 
--- a/src/Init/Core.lean
+++ b/src/Init/Core.lean
@@ -1603,6 +1603,98 @@ instance : Subsingleton (Squash α) where
     apply Quot.sound
     trivial
 
+/-! # List primitives -/
+
+/-- Auxiliary for `List.reverse`. `List.reverseAux l r = l.reverse ++ r`, but it is defined directly. -/
+def List.reverseAux : List α → List α → List α
+  | nil,    r => r
+  | a :: l, r => reverseAux l (a :: r)
+
+/--
+`O(|as|)`. Reverse of a list:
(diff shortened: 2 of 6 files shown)


joehendrix:
!bench

src/Init/Data/List/Basic.lean
```diff
@@ -902,4 +902,35 @@ theorem drop_eq_nil_of_le {as : List α} {i : Nat} (h : as.length ≤ i) : as.dr
   | _::_,  0   => simp at h
   | _::as, i+1 => simp at h; exact @drop_eq_nil_of_le as i (Nat.le_of_succ_le_succ h)
 
+/-- Auxiliary definition for `toArray`. -/
+@[inline_if_reduce]
+def toArrayAux : List α → Array α → Array α
+  | nil,       r => r
+  | cons a as, r => toArrayAux as (r.push a)
+
+/-- A non-tail-recursive version of `List.length`, used for `List.toArray`. -/
```
digama0, line 911: Do we still need this? `List.length` is already non-tail-recursive. I think this dates back to when `List.lengthTR` was called `List.length`.

nomeata:
Here is a tangentially related zulip thread about `Array.data` vs. `Array.toList`. Maybe they can be similarly made “the same” to avoid that confusion:
https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Understanding.20the.20docstring.20for.20docs.23Array.2Edata/near/398705430


On the weekend I was playing around with verifying some code that uses ByteArray and Array, which are isomorphic to Lists, and it wasn’t smooth sailing yet. Pondering this issue a bit more, I think there is a design decision that std4 has to take (if it hasn't sailed yet), between:

1. Each of these types has a full-fledged API, with it’s own simp-canonical form, suitable extensionality lemma and all the simp lemmas you expect (often one per combination of operations).
 
2. Only *one* of these types is meant to be used for verification (likely List). Operations on Array and ByteArray are each specified in terms of the List one and the suitable conversion (`(Array.filter xs p).toList = List.filter xs.toList p`). 
 
Currently, I think we are mostly doing 1, but there are operations (like `Array.findIdx?`) that don’t have the set of lemmas you’d expect. I found I sometimes had to convert from Arrays to Lists and back to get to apply the lemmas that I care about.

I have a gut feeling that 2 will scale better and lead to better, more consistent proofs, but probably only if applied consistently, i.e. there should then be *no* proofs about operations on Arrays besides the lemma that transports to arrays.

But it’s not what we have right now, and be too disruptive. We we’d have to do 1, but then an entitled user like me would probably expect to consistently find that every operation has the same set of lemmas on all three isomorphic data types, plus all the transport lemmas. Quite some explosion.

You probably thought about this too already; just jotting down some late evening thoughts here.

digama0:
I think we can do both, to some extent. In particular, following (2) is not incompatible with having other theorems anyway, as long as all the transfer theorems are there. My inclination is to start with lemmas facilitating (2) and add API lemmas for (1) on an as-needed basis.

nomeata:
That’s true. I’m wondering if there is merit (hypothetically) in an an exclusive (2) approach. With a mixed approach, you start proving things at the abstraction level of ByteStrings, and half-way through you find that some lemmas you need are only there for Lists. Now you maybe start to convert back and forth in an unprincipled way, and things get hairy.

But Joe mentioned plans to have some code in place to check for API completeness. Presumably it can help to achieve (1) completely, and warn if a lemma about ByteStrings is missing based on the existence of a lemma about List.

digama0:
> That’s true. I’m wondering if there is merit (hypothetically) in an an exclusive (2) approach. With a mixed approach, you start proving things at the abstraction level of ByteStrings, and half-way through you find that some lemmas you need are only there for Lists. Now you maybe start to convert back and forth in an unprincipled way, and things get hairy.

All of the (1) lemmas would have very easy proofs as long as a robust (2) bridge exists, and if a (1) lemma is missing you can prove it on the spot by transporting across the bridge (and there is no "and back", it's just lifting everything to List and proving things there). And if you needed to do that, maybe PR the lemma once you're done. It should not get hairy; if it does then there are missing lemmas.

joehendrix:
I've made some experiments with the compiler and this code, and it affects the generated code.

I'm going to see if this change in particular is easy to fix, but more changes may need to wait until there is time for the new compiler.

src/Init/Data/List/Basic.lean
joehendrix: The main current difference is that this has `inline_if_reduce` while `length` does not.

Just going by the name, that could likely be added to `length`, but I'm trying to be minimally disruptive until the new compiler development is resumed.

joehendrix:
!bench

joehendrix:
I've manually inspected compilation results on a small example program as well as the results of `update-stage0`.  This change appears to consistently preserve compiler behavior.

joehendrix:
I think I have resolved the issue.  The `csimp` application seems to stop working with some changes to `Init.Core`, my testing would show that even an empty `List` namespace would cause tests to fail:

e.g., I get different behavior if I add this to `Init.Core`:
```lean
namespace List
end List
``` 

Moving the `Append String` instance from `Init.Data.List.Basic` to `Init.Core` also changed behavior.  The tests are passing now after being very careful to only move essential definitions.

digama0:
What tests failed, and how? Adding a `namespace List` can indeed change things since it will implicitly declare the namespace if it hasn't already been declared, which in turn affects the behavior of the `open` command, but I don't see any particular reason this should cause issues if code was already importing `Init` which of course defines the `List` namespace.

joehendrix:
> What tests failed, and how?

Hi @digama0, It turns out I need to prioritize other issues and so I've closed this for now.

However, if you create a lean namespace in `Init.Core` like below, then I would see test cases that compiled to `appendTR` calls instead compile to `append` calls.  It's pretty strange.

```lean
namespace Lean
end Lean
```

With regards to the broader issue of `Array` and `List` compatibility, I think for some operations close equivalence is good, but others there may be efficient and elegant definitions for arrays since you have fast indexing and a Decidable instance for Fin and quantifiers (will be separate PR).

Here are a couple definitions I created before switching gears:

```lean
namespace Array

-- | This is equivalent to Array.allDiff
def allDiffIdx [BEq α] (as : Array α) : Bool :=
  ∀(i : Fin as.size) (j : Fin i),
    as[i] != as[j]'(Nat.lt_trans j.isLt i.isLt)

-- | This is equivalent to Array.all
-- It has some definite issues due to current omega limitations, but I
-- think this approach has a lot of potential to replace inductive
-- reasoning.
def allIdx (as : Array α) (p : α → Bool) (start : optParam Nat 0) (stop : optParam Nat as.size) : Bool :=
  let stop : Nat := stop
  let n := min stop as.size
  ∀(i : Fin (n - start)),
    let ⟨i, lt⟩ := i
    have _ : n ≤ as.size :=  Nat.min_le_right stop as.size
    p (as[start + i]'(by omega))

end Array
```

PR #3234, opened 2024-01-30
feat: apply’s error message should show implicit arguments as needed

### LINKED ISSUE #3232: “failed to unify” message should not print identical types
Sometimes Lean says

```
  some type
and
  some type
```

and `some type` is visually identical in both cases.

Of course advanced users will be forgiving such silliness, as they know about implicit arguments and things, and can use `set_option pp.explicit true`. But it is still quite silly for Lean to even say something like this.

Is there a reason it couldn’t compare the delaborated `Syntax` for equality, and if it is equal, elaborate again with `pp.explicit := true` (and if that’s still equal maybe with `pp.raw := true`), printing

```
  @some 1 type
and
  @some 2 type
(shown here with pp.explicit := true to reveal differences)
```

### Context

* https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/rage/near/418885474

### Steps to Reproduce

(TODO)

### Versions

Certainly around 0.4.5

### Impact

nomeata:
luckily the necessary functionality already exists in the form of
`addPPExplicitToExposeDiff`. But it is not cheap, and we should not run this code
when the error message isn’t shown, so this also introduces “thunks” to the `MessageData`
data structure.

Unfortunately, positivity restrictions in Lean prevent a simple
```
  | thunk : IO MessageData → MessageData
```
but (maybe for the wrong reasons)
```
  | thunk : Thunk MessageData → MessageData
```
works.

Together with a newly introduced, not-really-fully safe
```
def asThunk (f : BaseIO α) : BaseIO (Thunk α) := do
  return Thunk.mk fun () => unsafeBaseIO f
```
we can implement
```
MessageData.ofMetaMThunk : MetaM MessageData → MetaM MessageData :
```
which stores the MetaM state away (in a `PPContext`) and defers running the
`MetaM MessageData` later.

This allows us to invest into creating good explanatory error messages without paying that cost when the error message is never shown e.g. due to backtracking.

Fixes #3232.

### DIFF
--- a/src/Lean/Meta/Basic.lean
+++ b/src/Lean/Meta/Basic.lean
@@ -350,6 +350,34 @@ instance : MonadBacktrack SavedState MetaM where
 instance [MetaEval α] : MetaEval (MetaM α) :=
   ⟨fun env opts x _ => MetaEval.eval env opts x.run' true⟩
 
+/--
+Saves the current state and context of the `MetaM` (and `CoreM`) monad in a `PPContext`.
+-/
+def MetaM.savePPContext : MetaM PPContext := do
+  let env ← getEnv
+  let mctx ← getMCtx
+  let lctx ← getLCtx
+  let opts ← getOptions
+  let { currNamespace, openDecls, .. } ← readThe Core.Context
+  return { env, mctx, lctx, opts, currNamespace, openDecls }
+
+/--
+Runs a `CoreM` action in the context stored in the `ppCtxt`, typically for pretty-printing.
+Changes to the state are discarded.
+-/
+def _root_.Lean.PPContext.runCoreM {α : Type} (ppCtx : PPContext) (x : CoreM α) : IO α :=
+  Prod.fst <$> x.toIO { options := ppCtx.opts, currNamespace := ppCtx.currNamespace,
+                        openDecls := ppCtx.openDecls, fileName := "<PrettyPrinter>",
+                        fileMap := default }
+                      { env := ppCtx.env, ngen := { namePrefix := `_pp_uniq } }
+
+/--
+Runs a `MetaM` action in the context stored in the `ppCtxt`, typically for pretty-printing.
+Changes to the state are discarded.
+-/
+def _root_.Lean.PPContext.runMetaM {α : Type} (ppCtx : PPContext) (x : MetaM α) : IO α :=
+  ppCtx.runCoreM <| x.run' { lctx := ppCtx.lctx } { mctx := ppCtx.mctx }
+
 protected def throwIsDefEqStuck : MetaM α :=
   throw <| Exception.internal isDefEqStuckExceptionId
 
@@ -1551,12 +1579,9 @@ Runs the `MetaM` action lazily, when the resulting `Thunk` is forced.
 This forks the state of the `MetaM` and `CoreM` monad, and discards changes to that state.
 Caveats of `BaseIO.asThunk` apply.
 -/
-def MetaM.asThunk (m : MetaM α) : MetaM (Thunk (Except Exception α)) := do
-  let ctx ← readThe Meta.Context
-  let state ← saveState
-  let coreState ← getThe Core.State
-  let coreCtxt ← readThe Core.Context
-  (do restoreState state; m).run' (ctx := ctx) |>.run' coreCtxt coreState |>.toBaseIO.asThunk
+def MetaM.asThunk (m : MetaM α) : MetaM (Thunk (Except IO.Error α)) := do
+  let ppCtx ← savePPContext
+  BaseIO.asThunk (ppCtx.runMetaM m |>.toBaseIO)
 
 /--
 Delays computation of `MessageData` until the message is used. This can be used to avoid constructing
@@ -1565,7 +1590,7 @@ a complex error message when it woudn’t be shown anyways.
 def _root_.MessageData.ofMetaMThunk (m : MetaM MessageData) : MetaM MessageData :=
   return .thunk <| (← m.asThunk).map fun
     | .ok m => m
-    | .error e => m!"Exception while constructing message:\n{e.toMessageData}"
+    | .error e => m!"Exception while constructing message:\n{e}"
 
 /--
   Sort free variables using an order `x < y` iff `x` was defined before `y`.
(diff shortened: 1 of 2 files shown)


tests/lean/issue3232.lean.expected.out
```diff
@@ -0,0 +1,6 @@
+issue3232.lean:5:2-5:9: error: tactic 'apply' failed, failed to unify
+  @foo 42
+with
+  @foo 23
```
nomeata, line 4: Was
```
  foo
with
  foo
```
before

nomeata:
This has as a side effect to instantiate MVars, as one test shows. Probably a good thing, or at least worth the UX improvement?

tests/lean/issue3232.lean.expected.out
kmill: Could you add tests to show what happens when the differences are hidden by notation?

nomeata: I don't know much about notation, but happy to add any tests that you paste here :-)

(Even if it doesn’t work then yet it’s probably worth merging this as an incremental improvement.)

kmill: Maybe this?
```
example : (1 : Nat) = 1 := by
  apply (rfl : (1 : Int) = 1)
```

nomeata:
Hmm, I'm just not used to strict programming languages, it seems
```

Build: Mathlib/RingTheory/Finiteness.lean#L345
(deterministic) timeout at 'isDefEq', maximum number of heartbeats (200000) has been reached (use 'set_option maxHeartbeats <num>' to set the limit)
```

Of course apply is used a lot in tactics where we expect it to fail, and where the error message isn't shown, but my change would still prepare the error message and that's expensive. (Is this analysis correct?)

Is there a way to defer some work until the message is actually printed?

Kha:
I believe this should be possible using `MessageData.ofPPFormat`

nomeata:
> I believe this should be possible using `MessageData.ofPPFormat`, see also the implementation of `ofExpr`

Thanks for the pointer. I played around a bit with this, but it seems non-trivial to make this work (may need a bit of refactoring in various places); putting this back on the backlog for now.

nomeata:
Speedcenter report shows no significant changes: http://speed.lean-fro.org/mathlib4/compare/13cdd4f5-e329-4436-8167-85d13c940c0b/to/4aff2e40-f718-4e3e-be64-b37a963758fc

nomeata:
Ready for review, in particular the dangerous looking `BaseIO.asThunk` and `MetaM.asThunk` functions. Is this sufficiently kosher?

tests/lean/issue3232.lean.expected.out
nomeata: Before:
```
  1 = 1
with
  1 = 1
```
now
```
  @OfNat.ofNat Int 1 instOfNat = 1
with
  @OfNat.ofNat Nat 1 (instOfNatNat 1) = 1
```

src/Lean/Message.lean
```diff
@@ -64,7 +64,10 @@ inductive MessageData where
   |  compose           : MessageData → MessageData → MessageData
   /-- Tagged sections. `Name` should be viewed as a "kind", and is used by `MessageData` inspector functions.
     Example: an inspector that tries to find "definitional equality failures" may look for the tag "DefEqFailure". -/
-  | tagged            : Name → MessageData → MessageData
+  | tagged             : Name → MessageData → MessageData
+  /-- A lazy message data. This can be used to delay possibly costly computations of `MessageData`
+    that may not be used in the end, such as a `throwError` in a tactic inside `try`.  -/
+  | thunk              : Thunk MessageData → MessageData
```
Vtec234, line 70: Hey @nomeata! I'm still confused as to why this needs a new constructor given that `ofPPFormat` exists. The issue was that using `ofPPFormat` does not properly generate highlightable spans. Could that not be fixed without adding a new constructor?

nomeata: Worth exploring. Can you define `++` on `FormatWithInfos`?

kmill: What's the difference between implementing this as a new constructor and defining `MessageData.thunk x` as `.ofPPFormat fun _ => x.get`?

Vtec234: I am assuming that if the new constructor isn't needed then neither is `BaseIO.ofThunk`, so that all concerns about purity/evaluation order go away.

nomeata: Because `x` can return `MessageData`, not just `FormatWithInfo`, so you can use all the message data related functionality – embedding goals with `ofGoal`, different contexts for different subcomponents, etc.

In a way, `.ofPPFormat` is a constructor that bundles three orthogonal features that could all be their own constructor (ignoring positivity requirements):

 * It allows lazy creation of a message (with a `hasSyntheticSorry` peeking)
 * It gives access to an implicit context, so that you compose MessageData in pure code without having to say `(← ppTerm e)` everywhere
 * It extends `.ofFormat` with the ability to add Info annotations
 
One might argue that `.ofPPFormat` could be split into three constructors, one for each of these features, then it would be easier to put them together in more flexible ways.

I may well be missing some obvious solution here, though, I don’t think I have completely penetrated this part of the code base.

nomeata: Hmm, reading the code some more it seems that renumbering the tags in `FormatWithInfos` to be able to concatenate two of them, and thus life the `Format` API to `FormatWithInfos`, may be possible as well. Doesn’t feel particularly great to have yet another Format-API-wrapper (besides `Format` itself and `MessageData`).

nomeata: > I am assuming that if the new constructor isn't needed then neither is BaseIO.ofThunk, so that all concerns about purity/evaluation order go away.

Yes, that would be good indeed, it’s a bit of a cannon and sparrows situation. We might for example define a `IOThunk` variant where `.get` is monadic, to fix the purity worries.

digama0: Isn't `IOThunk` just spelled `IO`?

nomeata: I'd expect an `IOThunk` to internally also store the result and run the action at most once, something we want for lazy MessageData.

nomeata:
Alternative without thunks in #3926

PR #3241, opened 2024-02-01
feat: have `pp.proofs` use `⋯` for omission

kmill:
By having the `pp.proofs` feature use `⋯` when omitting proofs, when users copy/paste terms from the InfoView the elaborator can give an error message explaining why the term cannot be elaborated.

Also adds `pp.proofs.threshold` option to allow users to pretty print shallow proof terms. By default, only atomic proof terms are pretty printed.

This adjustment was suggested in PR #3201, which added `⋯` and the related `pp.deepTerms` option.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -11,6 +11,11 @@ of each version.
 v4.7.0 (development in progress)
 ---------
 
+* When the `pp.proofs` is false, now omitted proofs use `⋯` rather than `_`,
+  which gives a more helpful error message when copied from the Infoview.
+  The `pp.proofs.threshold` option lets small proofs always be pretty printed.
+  [#3241](https://github.com/leanprover/lean4/pull/3241).
+
 v4.6.0
 ---------
 


src/Lean/Widget/InteractiveCode.lean
```diff
@@ -82,7 +82,8 @@ def ppExprTagged (e : Expr) (explicit : Bool := false) : MetaM CodeWithInfos :=
       withOptionAtCurrPos pp.explicit.name true do
         delabApp
     else
-      delab
+      withOptionAtCurrPos pp.proofs.name true do
```
kmill, line 85: @Vtec234 Does this change seem fine? It causes just the top-level of delaboration to show the proof term. I made this change since otherwise if you hover over the `⋯` in `(⋯ : 1 ≤ Nat.succ 1)`, you just see something like `(⋯ : 1 ≤ Nat.succ 1) : 1 ≤ Nat.succ 1`.

There might be another design where OmissionInfo is processed differently for top-level expressions.

Vtec234: Hm, `ppExprTagged` is used to print a bunch of things. This includes the target type, and hypothesis types/values, in tactic states, as well as expressions in widgets. So changing it has global consequences. But I think it's a defensible choice to set `pp.proofs := true` at the root position, since if the only thing being printed in a given situation is a proof, you probably want to show that proof and not ⋯. In that case I guess you want to set it regardless of `explicit` since `explicit` doesn't seem to imply `proofs`.

kmill: Thanks for taking a look.

The function that's the main frontend to the delaborator, `delabCore`, does something stronger, which is if the thing being pretty printed is a proof, then [it overrides the option `pp.proofs` and sets it to true](https://github.com/leanprover/lean4/blob/43bbedca46f890e2d2b29d92f71b1e7b76aa0e93/src/Lean/PrettyPrinter/Delaborator/Basic.lean#L343-L347). I had thought about doing that here, but I wasn't sure and left it as a positional option.

src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -247,6 +247,20 @@ See also `Lean.PrettyPrinter.Delaborator.Context.depth`.
 def withIncDepth (act : DelabM α) : DelabM α := fun ctx =>
   act { ctx with depth := ctx.depth + 1 }
 
+/--
+Returns true if `e` is a "shallow" expression.
+Local variables, constants, and other atomic expressions are always shallow.
+The `threshold` variable adjusts what else is considered to be shallow,
+and it's compared against the depth of the expression.
```
mhuisi, line 254: "by comparing it against the depth of the expression."? Otherwise, the doc comment suggests that multiple things are happening here.

src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -255,19 +269,39 @@ an expression, which prevents terms such as atomic expressions or `OfNat.ofNat`
...
-  let isMaxedOutApproxDepth := approxDepth >= 255
-  let isShallowExpression :=
-    !isMaxedOutApproxDepth && approxDepth <= depthThreshold/4 - depthExcess
+  return depthExcess > 0 && !isShallowExpression threshold e
 
-  return depthExcess > 0 && !isShallowExpression
+/--
+Returns `true` if the given expression is a proof and it should be omitted.
+This function only returns `true` if `pp.proofs` is set to `false`.
+It uses `pp.proofs.threshold` to determine what is a "shallow" proof -- shallow proofs are not omitted.
```
mhuisi, line 291: "It uses `pp.proofs.threshold` to determine "shallow" proofs that are not omitted"?

mhuisi (approved):
These changes look good to me!

src/Lean/PrettyPrinter/Delaborator/Basic.lean
kmill: Thanks, these were confusingly worded. I used your ideas and tried refining them even more. How does it look now?

mhuisi: I'm happy with your refinement.

PR #3247, opened 2024-02-02
feat: diagnostics for stale dependencies

mhuisi:
Sends a diagnostic informing the user to run Restart File when a file dependency is saved.

Based on #3014 because this feature was easier to implement with the new architecture.

ToDo:
- [x] Adjust vscode-lean4 to display a notification when this diagnostic appears in a non-annoying way (https://github.com/leanprover/vscode-lean4/pull/393)
- [x] Use a file watcher to identify changes to files not tracked by VS Code
- [x] Rebase onto master when #3014 is merged

### DIFF
--- a/src/Lean/Data/Lsp/Diagnostics.lean
+++ b/src/Lean/Data/Lsp/Diagnostics.lean
@@ -129,15 +129,8 @@ private structure DiagnosticWith.UserVisible (α : Type) where
 
 /-- Extracts user-visible properties from the given `DiagnosticWith`. -/
 private def DiagnosticWith.UserVisible.ofDiagnostic (d : DiagnosticWith α)
-    : DiagnosticWith.UserVisible α where
-  range               := d.range
-  fullRange?          := d.fullRange?
-  severity?           := d.severity?
-  code?               := d.code?
-  source?             := d.source?
-  message             := d.message
-  tags?               := d.tags?
-  relatedInformation? := d.relatedInformation?
+    : DiagnosticWith.UserVisible α :=
+  { d with }
 
 /-- Compares `DiagnosticWith` instances modulo non-user-facing properties. -/
 def compareByUserVisible [Ord α] (a b : DiagnosticWith α) : Ordering :=


src/Lean/Server/FileWorker.lean
```diff
@@ -59,24 +59,31 @@ open IO
...
-  maxDocVersionRef : IO.Ref Int
-  hLog             : FS.Stream
-  initParams       : InitializeParams
-  processor        : Parser.InputContext → BaseIO Lean.Language.Lean.InitialSnapshot
-  clientHasWidgets : Bool
+  maxDocVersionRef     : IO.Ref Int
+  /--
+  Diagnostics that are included in every single `textDocument/publishDiagnostics` notification.
+  -/
+  stickyDiagnosticsRef : IO.Ref StickyDiagnostics
```
Vtec234, line 77: Might it make sense to make this a map `StaleDepId -> InteractiveDiagnostic` instead? It would make for an easier comparison in `handleStaleDependency` (so that you wouldn't need `compareAsDiagnostics`), and you still get to grab all the cached sticky diagnostics quickly by iterating over the values of the map. If other kinds of sticky diagnostics are added later, it could be keyed by `inductive StickyDiagId | staleDep id | ...`.

Vtec234 (approved):
LGTM on a very cursory look.

src/Lean/Server/FileWorker.lean
mhuisi: The goal here is specifically to filter sticky diagnostics that look identical (otherwise an `Array` would suffice as well, see also the `fix: deduplicate sticky diagnostics` commit). Without this map, if you save a dependency multiple times, there will be multiple diagnostics in the same location with the same message informing the user that this specific dependency is stale.
Since displaying the same sticky diagnostic multiple times seems like noise in general, I figured it would be best to just implement this at the level of the sticky diagnostics themselves.

src/Lean/Data/Lsp/Diagnostics.lean
```diff
@@ -113,6 +113,36 @@ structure DiagnosticWith (α : Type) where
...
+private def DiagnosticWith.UserVisible.ofDiagnostic (d : DiagnosticWith α)
+    : DiagnosticWith.UserVisible α where
+  range               := d.range
+  fullRange?          := d.fullRange?
+  severity?           := d.severity?
+  code?               := d.code?
+  source?             := d.source?
+  message             := d.message
+  tags?               := d.tags?
+  relatedInformation? := d.relatedInformation?
```
Kha, line 140:
```suggestion
    : DiagnosticWith.UserVisible α :=
  { d with }
```
should work I think

Kha (approved):

src/Lean/Server/FileWorker.lean
Vtec234: No I know, but I think the map I proposed would also solve that? If it is keyed by the ID (module name, for instance) of the stale dependency, calling `.insert staleDepName staleDepDiag` would replace the previous entry, not add a new one.

mhuisi: I think I'd like to keep it as it is. The sticky diagnostics shouldn't need to know a lot about the diagnostics that are contained in them. I know that we can generalize this further to a sigma type, but I don't really see the point, since I doubt that the sticky diagnostics lookup will ever be a bottleneck.

PR #3249, opened 2024-02-03
doc: add links to folder references

jeswr:
This PR adds links to some folder references in the docs, making them easier to navigate.

Please advise if these need to be made to be full URIs rather than relative paths in order to work correctly with the doc generation tooling that is in place.


david-christiansen:
These should indeed be made into explicit GitHub links - the contents of the `doc` directory are rendered into HTML, and this would show up as an absolute link in the context of the generated site, rather than the original context of the repository.

david-christiansen:
(and thanks!)

david-christiansen:
This is great, thanks!

Issue #3245, opened 2024-02-02
Using split on if-then-else simplifies unrelated nested if-then-else


nomeata:
This is rather minor, but I stumbled over this while debugging #3229, and I’d like to record it as an instance of a non-finishing tactic doing possibly unexpected things on the side.

Consider this example:
```
example (b1 b2 : Bool) :
  (match b1 with
  | true => 0
  | false => (if b2 then 0 else if b2 then 42 else 0)) = 0 := by
  split
  rfl
```
Now the goal is, as expected
```
b2 b1✝: Bool
⊢ (if b2 = true then 0 else if b2 = true then 42 else 0) = 0
```
The `split` tactic split the `match`, and leaves the branch in place.

Using `if-then-else` however:
```
example (b1 b2 : Bool) :
  (if b1
  then 0
  else (if b2 then 0 else if b2 then 42 else 0)) = 0 := by
  split
  rfl
```
the goal becomes
```
b1b2: Bool
h✝: ¬b1 = true
⊢ (if b2 = true then 0 else 0) = 0
```

Note that suddenly, the totally unrelated nested `if b2` was reduced.

The problem lies is with `Lean.Meta.simpIfTarget`, which 

 * uses the if-then-else congruence theorems and
 * a discharger that looks into the hypotheses.

I wonder if it really needs to enable `ite_congr` – after all, for the `if-then-else` that was `split` the condition is already in the context, so no need for the “better” congruence rule and the standard rule should suffice?

### Versions

v4.5.0

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

semorrison:
I realise that I've encountered this several time.

nomeata:
> I wonder if it really needs to enable ite_congr – after all, for the if-then-else that was split the condition is already in the context, so no need for the “better” congruence rule and the standard rule should suffice?

This is barking up the wrong tree, and seems to have incidentially be fixed by #3258.

But there is still a variant of the issue: After

```
example (b1 b2 : Bool) (h : b2 = true) :
  (if b1 then 0 else (if b2 then 0 else if b2 then 42 else 0)) = 0 := by
  split
```
we get 
```
2 goals
case inl
b1 b2 : Bool
h : b2 = true
h✝ : b1 = true
⊢ 0 = 0

case inr
[...]

PR #3253, opened 2024-02-04
fix: try out proposed fix for #3245

kim-em:


semorrison:
Ah, we need at least a congruence theorem to descend into the condition, but we don't want to descend into the branches. Trying again.

PR #3255, opened 2024-02-05
feat: improve termination_by error messages

nomeata:
as suggested in
<https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/termination_by.20regression/near/419786430>

Also refactored the code a bit and removed the code smell around
`GuessLex`-produced termination arguments (which may not be
surface-syntactically expressible) a bit by introducing an explicit flag
for those.

### DIFF
--- a/src/Lean/Elab/PreDefinition/WF/TerminationHint.lean
+++ b/src/Lean/Elab/PreDefinition/WF/TerminationHint.lean
@@ -51,15 +51,13 @@ structure TerminationHints where
   ref : Syntax
   termination_by? : Option TerminationBy
   decreasing_by?  : Option DecreasingBy
-  /-- Here we record the number of parameters past the `:`. This is used in two places:
+  /-- Here we record the number of parameters past the `:`. It is set by
+  `TerminationHints.rememberExtraParams` and used as folows:
 
   * When we guess the termination argument in `GuessLex` and want to print it in surface-syntax
     compatible form.
   * If there are fewer variables in the `termination_by` annotation than there are extra
-    parameters, we know which parameters they should apply to.
-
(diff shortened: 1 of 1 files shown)


src/Lean/Elab/PreDefinition/WF/TerminationHint.lean
```diff
@@ -70,19 +78,31 @@ def TerminationHints.ensureNone (hints : TerminationHints) (reason : String): Co
...
     if tb.vars.size > extraParams then
-      logErrorAt tb.ref <| m!"Too many extra parameters bound; the function definition only " ++
-        m!"has {extraParams} extra parameters."
-  return { hints with extraParams := extraParams }
+      let mut msg := m!"{parameters tb.vars.size} bound in `termination_by`, but the body of " ++
+        m!"{funName} only binds {parameters extraParams}."
+      if let `($ident:ident) := tb.vars[0]! then
+        if ident.getId.isSuffixOf funName then
+            msg := msg ++ m!"\nNB: Since Lean v4.6, the `termination_by` clause no longer " ++
+              "expects the function name here."
```
semorrison, line 100:
```suggestion
            msg := msg ++ m!"\nNB: Since Lean v4.6.0, the `termination_by` clause no longer " ++
              "expects the function name here."
```

(Maybe parenthetical, rather than "NB"? Easier to parse, I think.)

semorrison: Also, you can now use multiline strings just by ending with a line with a `\`!

nomeata: Will do.

Too bad we don’t have a “update all expected test output” github action yet…

PR #3256, opened 2024-02-05
fix: let induction handle parameters

### LINKED ISSUE #3212: bug: induction intros reverted assumptions even in goals that don’t result in the motive
### Prerequisites

### Description

The induction tactic does not behave as intended when there are restrictions on motive.

### Context

[Zulip chat](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/restricted.20induction)

Thanks to @nomeata for pointing out that this might be a bug.

### Steps to Reproduce

Execute the following code in web editor.
```lean
axiom N : Type

axiom zero : N

axiom succ : N → N

axiom add : N → N → N

axiom Restriction : (N → Prop) → Prop

axiom restricted_induction {motive : N → Prop} (h : Restriction motive)
  (zero : motive zero) (succ : ∀ n, motive n → motive (succ n)) : ∀ n, motive n

example (n m : N) (h : n ≠ zero) : add m n ≠ zero := by
  induction n using restricted_induction
```

**Expected behavior:**
 Three goals are generated: `Restriction fun n => n ≠ zero → add m n ≠ zero`, `add m 0 ≠ zero` and `add m (succ n) ≠ zero`. In later two, the hypothesis `h` are rewriten to `zero ≠ zero` and `succ n ≠ zero`, respectively.

**Actual behavior:** 
Returns following error.
```
tactic 'introN' failed, insufficient number of binders

nm: N
⊢ Restriction fun n => n ≠ zero → add m n ≠ zero
```

### Versions

4.5.0-rc1
Web editor

### Impact

nomeata:
The induction principle used by `induction` may have explicit parameters that are
not motive, target or “real” alternatives (that have the `motive` as
conclusion), e.g. restrictions on the `motive` or other parameters.

Previously, `induction` would treat them as normal alternatives, and try
to re-introduce the automatically reverted hypotheses. But this only
works when the `motive` is actually the conclusion in the type of that
alternative.

We now pay attention to that, thread that information through, and only
revert when needed.

Fixes #3212.

### DIFF
--- a/tests/lean/run/issue3212.lean
+++ b/tests/lean/run/issue3212.lean
@@ -50,4 +50,11 @@ example (n m : Nat) (h : m * m < 100) (h2 : n ≤ m) : n * n < 100 := by
     have IH := IH hlt
     sorry
 
+-- Unfortunately, this doesn’t work either:
+--  example (n m : Nat) (h : m * m < 100) (h2 : n ≤ m) : n * n < 100 := by
+--    induction n using down_induction (le_u := h2)
+-- because after this, the target `n` is no longer a universally qualified variable
+--
+-- This probably could be made to work with a bit more refactoring.
+
 end DownInduction


nomeata:
I’m fairly confident, but will wait merging for a day or two if someone wants to have a second look.

PR #3258, opened 2024-02-05
fix: cache issue at `split` tatic

### LINKED ISSUE #3229: loose fvar in `split`
```lean
example (b : Bool) :
    (if b then 1 else if b then 1 else 0) = (if b then 1 else 0) := by
  split <;> sorry
```
```
(kernel) declaration has free variables '_example'
```

Observed on `leanprover/lean4:v4.5.0-rc1`

leodemoura:
closes #3229

### DIFF
--- a/tests/lean/run/3229.lean
+++ b/tests/lean/run/3229.lean
@@ -1,4 +1,4 @@
 -- This example exposed a caching issue with the `discharge?` method used by the `split` tactic.
 example (b : Bool) :
     (if b then 1 else if b then 1 else 0) = (if b then 1 else 0) := by
-  split <;> sorry
+  split <;> rfl


tests/lean/run/3229.lean
```diff
@@ -0,0 +1,4 @@
+-- This example exposed a caching issue with the `discharge?` method used by the `split` tactic.
+example (b : Bool) :
+    (if b then 1 else if b then 1 else 0) = (if b then 1 else 0) := by
+  split <;> sorry
```
nomeata, line 4: A bit more robust
```suggestion
  split <;> rfl
```

nomeata (commented):
I expect that this will, as a side effect, fix most (but not all) instances of #3245.

One could make `simpIfTarget` even more robust if it receives as a parameter the name of the hypotheses actually added by `split`, and discharge could simply use precisely that, instead of going through the local context. Not sure if it’s worth it, or if maybe using any available hypothesis is actually desirable here.

semorrison:
Mathlib fix is trivial, CI should go green shortly.

PR #3260, opened 2024-02-06
feat: upstream some Syntax/Position helper functions used in code actions in Std

kim-em:


src/Init/Data/String/Basic.lean
```diff
@@ -515,6 +515,21 @@ def replace (s pattern replacement : String) : String :=
       termination_by s.endPos.1 - pos.1
     loop "" 0 0
 
+/-- Return the beginning of the line contatining character `pos`. -/
```
david-christiansen, line 518:
```suggestion
/-- Return the beginning of the line that contains character `pos`. -/
```

david-christiansen: Typo fix seems important, other thing less so

src/Lean/Syntax.lean
```diff
@@ -305,6 +305,10 @@ def getRange? (stx : Syntax) (canonicalOnly := false) : Option String.Range :=
   | some start, some stop => some { start, stop }
   | _,          _         => none
 
+/-- Returns a synthetic Syntax which has the specified `String.Range`. -/
+def ofRange (range : String.Range) (canonical := true) : Lean.Syntax :=
```
Kha, line 309: This just screams "workaround" to me, I'd really prefer it to be in a more specific place and namespace

Kha (changes requested):

src/Init/Data/String/Basic.lean
```diff
@@ -515,6 +515,21 @@ def replace (s pattern replacement : String) : String :=
       termination_by s.endPos.1 - pos.1
     loop "" 0 0
 
+/-- Return the beginning of the line that contains character `pos`. -/
+def findLineStart (s : String) (pos : String.Pos) : String.Pos :=
+  match s.revFindAux (· = '\n') pos with
+  | none => 0
+  | some n => ⟨n.byteIdx + 1⟩
```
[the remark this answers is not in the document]
digama0: `FileMap.lineStart` takes a `line` number, `findLineStart` takes a `String.Pos`.

src/Lean/Syntax.lean
digama0: It is, kind of, but I do think it deserves to be centrally located. What it is working around is the fact that lots of server / info tree stuff uses `Syntax` where it really means `String.Range`, which causes many unnecessary optional unwraps and gives no good option for cases where the range being highlighted does not correspond to a single syntax. There are already a few places in lean core where the workaround of "make a null node with two `Syntax`es for the start and end bit" is used to graft a range together from pieces. I think it would be better to standardize on a single way of representing such dummy spans.

src/Lean/Data/Position.lean
```diff
@@ -84,6 +84,23 @@ partial def toPosition (fmap : FileMap) (pos : String.Pos) : Position :=
...
+    if h : pos.line - 1 < text.positions.size then
+      text.positions.get ⟨pos.line - 1, h⟩
+    else if text.positions.isEmpty then
+      0
+    else
+      text.positions.back
+  String.Iterator.nextn ⟨text.source, colPos⟩ pos.column |>.pos
+
+/-- Returns the position of the start of (1-based) line `line`. -/
+def lineStart (map : FileMap) (line : Nat) : String.Pos :=
```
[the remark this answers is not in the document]
digama0: Yes, but faster. (And also clearer to some extent; I always have trouble working with the sometimes 1-based and sometimes 0-based indexing of positions in lean and LSP.)

src/Init/Data/String/Basic.lean
```diff
@@ -515,6 +515,21 @@ def replace (s pattern replacement : String) : String :=
...
+
+/--
+Return the indentation (number of leading spaces) of the line containing `pos`,
+and whether `pos` is the first non-whitespace character in the line.
+-/
+def findIndentAndIsStart (s : String) (pos : String.Pos) : Nat × Bool :=
```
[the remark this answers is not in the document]
semorrison: I will remove from this PR, and then relocate the code close to the point of use in #3266.

src/Lean/Data/Position.lean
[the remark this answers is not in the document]
semorrison: I propose keeping, with the additional doc-string "This gives the stame result as `map.ofPosition ⟨line, 0⟩`, but is more efficient." which I've now added.

@Kha?

src/Init/Data/String/Basic.lean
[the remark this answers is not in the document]
Kha: Ah, makes sense now. Thanks!

src/Lean/Data/Position.lean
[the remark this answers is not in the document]
Kha: I would contest this being measurably faster in any real-life context :) but I'm fine with the current version

src/Lean/Syntax.lean
Kha: Is this the only std use? https://github.com/leanprover/std4/blob/453fb67573767ca73eb4f08710c6cb352a40aaa5/Std/Tactic/TryThis.lean#L350
Isn't that just
```
stx := origSpan?.getD ref
```
?

semorrison: @Kha, I don't know the exact intended behaviour in the place you linked, but the code is not equivalent to `orginSpan?.getD ref`.

The code is using the range of `ref` if it has a range, and otherwise insisting that `origSpan?.isSome` and that it has a range, and using that (and then creating a dummy Syntax from that).

Kha: Yes, I noticed that but found it unlikely to be intentional. Anyway, this function is used in the wild so I guess let's merge.

Kha (approved):

src/Lean/Syntax.lean
digama0: This is paged out at the moment, but I thought quite carefully about that line and I'm quite sure the semantics there are intentional.

PR #3261, opened 2024-02-06
feat: upstream TSyntax helper functions

kim-em:
From Std.Lean.Syntax.

### DIFF
--- a/src/Lean/Syntax.lean
+++ b/src/Lean/Syntax.lean
@@ -531,26 +531,4 @@ def Stack.matches (stack : Syntax.Stack) (pattern : List $ Option SyntaxNodeKind
 
 end Syntax
 
-namespace TSyntax
-
-/--
-Applies the given function to every subsyntax.
-
-Like `Syntax.replaceM` but for typed syntax.
-(Note there are no guarantees of type correctness here.)
--/
-def replaceM [Monad m] (f : Syntax → m (Option Syntax)) (stx : TSyntax k) : m (TSyntax k) :=
-  .mk <$> stx.1.replaceM f
-
-end TSyntax
-
-/--
-Constructs a typed separated array from elements.
-The given array does not include the separators.
-
-Like `Syntax.SepArray.ofElems` but for typed syntax.
--/
-def Syntax.TSepArray.ofElems {sep} (elems : Array (TSyntax k)) : TSepArray k sep :=
-  .mk (SepArray.ofElems (sep := sep) elems).1
-
 end Lean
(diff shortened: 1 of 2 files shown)


src/Lean/Syntax.lean
```diff
@@ -530,4 +530,27 @@ def Stack.matches (stack : Syntax.Stack) (pattern : List $ Option SyntaxNodeKind
...
+
+namespace TSyntax
+
+/--
+Applies the given function to every subsyntax.
+
+Like `Syntax.replaceM` but for typed syntax.
+(Note there are no guarantees of type correctness here.)
+-/
+def replaceM [Monad M] (f : Syntax → M (Option Syntax)) (stx : TSyntax k) : M (TSyntax k) :=
```
david-christiansen, line 542:
```suggestion
def replaceM [Monad m] (f : Syntax → M (Option Syntax)) (stx : TSyntax k) : m (TSyntax k) :=
```

semorrison:
```suggestion
def replaceM [Monad m] (f : Syntax → m (Option Syntax)) (stx : TSyntax k) : m (TSyntax k) :=
```
Missed one!

src/Lean/Syntax.lean
```diff
@@ -530,4 +530,27 @@ def Stack.matches (stack : Syntax.Stack) (pattern : List $ Option SyntaxNodeKind
...
 
 end Syntax
+
+namespace TSyntax
+
+/--
+Applies the given function to every subsyntax.
+
+Like `Syntax.replaceM` but for typed syntax.
+(Note there are no guarantees of type correctness here.)
```
Kha, line 540: I think I'd prefer the abstraction to just not exist then, unsafely creating `TSyntax` should be explicit

src/Lean/Syntax.lean
```diff
@@ -530,4 +530,27 @@ def Stack.matches (stack : Syntax.Stack) (pattern : List $ Option SyntaxNodeKind
...
+
+end TSyntax
+
+/--
+Constructs a typed separated array from elements.
+The given array does not include the separators.
+
+Like `Syntax.SepArray.ofElems` but for typed syntax.
+-/
+def Syntax.TSepArray.ofElems {sep} (elems : Array (TSyntax k)) : TSepArray k sep :=
```
Kha, line 553: We should move this in front of https://github.com/leanprover/lean4/blob/5cc0c3f145ffa623a617ca31387d7f589802c1d3/src/Init/Meta.lean#L566-L567 and make use of it there

Kha (changes requested):

src/Lean/Syntax.lean
semorrison: Done.

src/Lean/Syntax.lean
semorrison: Turns out it is not even used. I will remove from this PR and maybe clean it up in Std later.

PR #3262, opened 2024-02-06
feat: InfoTree helper function used in code actions

kim-em:


src/Lean/Server/InfoUtils.lean
```diff
@@ -86,6 +86,20 @@ where go ctx? a
     ts.foldl (init := a) (go <| i.updateContext? ctx?)
   | _ => a
 
+/-- Like `InfoTree.foldInfo`, but also passes the whole node to `f` instead of just the head. -/
```
david-christiansen, line 89:
```suggestion
/--
Fold an info tree as follows, while ensuring that the correct `ContextInfo` is supplied at each stage:

 * Nodes are combined with the initial value `init` using `f`, and the result is then combined with the children using a left fold
 * Everything else returns the initial value

This is like `InfoTree.foldInfo`, but it also passes the whole node to `f` instead of just the head.
-/
```

src/Lean/Server/InfoUtils.lean
```diff
@@ -86,6 +86,27 @@ where go ctx? a
...
+
+ * Nodes are combined with the initial value `init` using `f`, and the result is then combined with the children using a left fold
+ * Everything else returns the initial value
+
+This is like `InfoTree.foldInfo`, but it also passes the whole node to `f` instead of just the head.
+-/
+partial def InfoTree.foldInfoTree (init : α) (f : ContextInfo → InfoTree → α → α) : InfoTree → α :=
+  go none init
+where
+  /-- `foldInfoTree.go` is like `foldInfo'` but with an additional outer context parameter `ctx?`. -/
```
mhuisi, line 100: `foldInfo'`?

src/Lean/Server/InfoUtils.lean
```diff
@@ -86,6 +86,27 @@ where go ctx? a
     ts.foldl (init := a) (go <| i.updateContext? ctx?)
   | _ => a
 
+/--
+Fold an info tree as follows, while ensuring that the correct `ContextInfo` is supplied at each stage:
+
+ * Nodes are combined with the initial value `init` using `f`, and the result is then combined with the children using a left fold
+ * Everything else returns the initial value
```
mhuisi, line 93: It may be a good idea to be explicit about "everything else" (perhaps in the code as well). The remaining case are holes.

mhuisi (approved):

src/Lean/Server/InfoUtils.lean
semorrison:
```suggestion
  /-- `foldInfoTree.go` is like `foldInfoTree` but with an additional outer context parameter `ctx?`. -/
```
Lesson: never renaming while moving. Thanks.

src/Lean/Server/InfoUtils.lean
semorrison: Done.

PR #3263, opened 2024-02-06
feat: upstream helper functions for Name

kim-em:
This does not completely empty `Std.Lean.Name`, as working out how to document the difference between `Name.isInternalDetail` and `Name.isImplementationDetail` requires further thought.


david-christiansen:
So far as I can tell, these functions are both essentially checking whether the name should show up in various bits of the UI. I think the distinction is that they're used from UIs (one for local context entries in a tactic state, the other for definitions in the environment), but I don't see another difference.

I _think_ they could be united into one function that returns true if either would have, but that'd require a quick experiment.

PR #3265, opened 2024-02-06
feat: upstreaming the json% term elaborator

kim-em:
This is used in the "Try this:" widget machinery powering `simp?`.

There is a test file in Std, which I am not upstreaming at the same time, as that relies on more code actions / #guard_msgs material. That test file will still of course test things from Std, and later it can be reunited with the code it is testing.

### DIFF
--- a/tests/lean/run/toFromJson.lean
+++ b/tests/lean/run/toFromJson.lean
@@ -2,13 +2,13 @@ import Lean
 
 open Lean
 
-declare_syntax_cat json
-syntax str : json
-syntax num : json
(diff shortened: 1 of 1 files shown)


leodemoura:
It seems it failed building stage2

semorrison:
Indeed:
```
make[8]: *** [/home/runner/work/lean4/lean4/build/stage2/bin/../share/lean/lean.mk:70: ../build/stage2/lib/lean/Lean/Data/Json/Elab.olean] Segmentation fault (core dumped)
```

PR #3266, opened 2024-02-06
feat: upstream 'Try this:' widgets

kim-em:
There is a test file in Std that should later be reunited with this code.

### DIFF
--- a/src/Init/Tactics.lean
+++ b/src/Init/Tactics.lean
@@ -398,7 +398,7 @@ syntax locationWildcard := " *"
 A hypothesis location specification consists of 1 or more hypothesis references
 and optionally `⊢` denoting the goal.
 -/
-syntax locationHyp := (ppSpace colGt term:max)+ ppSpace patternIgnore( atomic("|" noWs "-") <|> "⊢")?
+syntax locationHyp := (ppSpace colGt term:max)+ patternIgnore(ppSpace (atomic("|" noWs "-") <|> "⊢"))?
 
 /--
 Location specifications are used by many tactics that can operate on either the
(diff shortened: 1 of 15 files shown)


src/Lean/Meta/Tactic/TryThis.lean
```diff
@@ -0,0 +1,585 @@
...
+namespace SuggestionText
+
+/-- Pretty-prints a `SuggestionText` as a `Format`. If the `SuggestionText` is some `TSyntax kind`,
+we use the appropriate pretty-printer; strings are coerced to `Format`s as-is. -/
+def pretty : SuggestionText → CoreM Format
+  | .tsyntax (kind := kind) stx => ppCategory kind stx
+  | .string text => return text
+
+/- Note that this is essentially `return (← s.pretty).prettyExtra w indent column`, but we
+special-case strings to avoid converting them to `Format`s and back. -/
```
Kha, line 202: Does this really exist to save something like a few allocations per suggestion item?

src/Lean/Meta/Tactic/TryThis.lean
```diff
@@ -0,0 +1,585 @@
...
+
+/--
+Style hooks for `Suggestion`s. See `SuggestionStyle.error`, `.warning`, `.success`, `.value`,
+and other definitions here for style presets. This is an arbitrary `Json` object, with the following
+interesting fields:
+* `title`: the hover text in the suggestion link
+* `className`: the CSS classes applied to the link
+* `style`: A `Json` object with additional inline CSS styles such as `color` or `textDecoration`.
+-/
+def SuggestionStyle := Json deriving Inhabited, ToJson
```
Kha, line 224: I'm actually confused again now about the need for unstructured JSON on the server side. Wouldn't it be nicer to do the style computations below in the client (which would allow for alternative clients, though that is probably not a priority)? Or do we want the style be open to extensibility?

Kha (approved):
a few nits but nothing that should block merging

src/Lean/Meta/Tactic/TryThis.lean
digama0: I believe Strings are converted to Format in a complicated way which involves splitting on newlines. I forget the details but I don't think this is just an optimization.

src/Lean/Meta/Tactic/TryThis.lean
digama0: This is for extensibility. The previous version had a variant for all of the `def`s below it, including the rainbow colored one, and it seemed better to just let clients set their own colors if they want to customize to that extent.

src/Lean/Meta/Tactic/TryThis.lean
```diff
@@ -0,0 +1,585 @@
...
+  s.replaceM fun s => do
+    let `(?$id:ident) := s | pure none
+    if id.getId.hasNum || id.getId.isInternal then `(?_) else pure none
+
+/-- Delaborate `e` into syntax suitable for use by `refine`. -/
+def delabToRefinableSyntax (e : Expr) : MetaM Term :=
+  return ⟨← replaceMVarsByUnderscores (← delab e)⟩
+
+/-- The default maximum width of an ideal line in source code, 100 is the current convention. -/
+def inputWidth : Nat := 100
```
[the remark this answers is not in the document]
semorrison: Done.

src/Lean/Meta/Tactic/TryThis.lean
Kha: Oh my, I think you are right when it comes to indentation

Kha, line 202:
```suggestion
special-case strings to avoid converting them to `Format`s and back, which adds indentation after each newline. -/
```

PR #3268, opened 2024-02-06
feat: introduce auto-imported `Base` package

Kha:
We are planning to introduce a new package to core for functionality that a) is not strictly necessary for building Lean (so it should not be in `Init` or `Lean`) but b) is too fundamental to keep in a separate repo (like `std4`) for technical and organizational reasons. In particular, this includes theorems about `Init` types.

This PR introduces `Base`, which specifically
* like `Init`, is automatically imported unless the module is marked `prelude`
* like `Lean`, depends on nothing but `Init` (and so can be built in parallel with `Lean`)
* like both, is included in `libleanshared` and so has built-in native code support

For end users, it really shouldn't matter whether something is in `Init` or `Base`, they will get it implicitly in either case

TODO
- [ ] sanity-check building native code against it
- [ ] fix stage0, for which we will want to prefix all of `Lean` with `prelude`

### DIFF
--- a/src/Lean/Elab/Import.lean
+++ b/src/Lean/Elab/Import.lean
@@ -9,7 +9,7 @@ import Lean.Data.Json
 namespace Lean.Elab
 
 def headerToImports (header : Syntax) : Array Import :=
-  let imports := if header[0].isNone then #[{ module := `Init : Import }] else #[]
+  let imports := if header[0].isNone then Import.prelude else #[]
   imports ++ header[1].getArgs.map fun stx =>
     -- `stx` is of the form `(Module.import "import" "runtime"? id)
     let runtime := !stx[1].isNone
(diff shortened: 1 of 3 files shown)


digama0:
> This PR introduces `Base`, which specifically
> 
>   * like `Init`, is automatically imported unless the module is marked `prelude`
> 
>   * like `Lean`, depends on nothing but `Init` (and so can be built in parallel with `Lean`)
> 
>   * like both, is included in `libleanshared` and so has built-in native code support

I do like the clarity of having `Init` be the single thing that is imported if you don't use `prelude`, and the description here makes `Base` much more similar to `Init` than anything else, as far as core language behavior is concerned. It seems like the requirements here would also be satisfied, with less impact downstream, if individual files in `Lean` were to use more precise imports with `prelude`? We could even make use of `shake` here to automatically generate the list of imports.

Kha:
You're quite right, we're having an internal discussion on whether this wins us anything compared to just using `Init`

PR #3276, opened 2024-02-08
chore: upstream Std.Util.TermUnsafe

kim-em:


digama0:
This should probably be followed up with a PR using it in core: there are a *ton* of potential use cases for this!

src/Lean/Util/TermUnsafe.lean
```diff
@@ -0,0 +1,62 @@
...
+#eval cool
+```
+-/
+
+namespace Lean.TermUnsafe
+
+open Lean Meta Elab Term
+
+/-- Construct an auxiliary name based on the current declaration name and the given `hint` base. -/
+def mkAuxName (hint : Name) : TermElabM Name :=
```
Kha, line 26: This feels like a function that wants to be relocated and used more broadly but it can start here

src/Lean/Util/TermUnsafe.lean
```diff
@@ -0,0 +1,62 @@
...
+body of `t : α`, by creating an auxiliary definition containing `t` and using `implementedBy` to
+wrap it in a safe interface. It is required that `α` is nonempty for this to be sound,
+but even beyond that, an `unsafe` block should be carefully inspected for memory safety because
+the compiler is unable to guarantee the safety of the operation.
+
+For example, the `evalExpr` function is unsafe, because the compiler cannot guarantee that when
+you call ```evalExpr Foo ``Foo e``` that the type `Foo` corresponds to the name `Foo`, but in a
+particular use case, we can ensure this, so `unsafe (evalExpr Foo ``Foo e)` is a correct usage.
+-/
+elab "unsafe " t:term : term <= expectedType => do
```
Kha, line 42: Syntax decl and docstring should probably be moved into Init?

Kha (changes requested):

src/Lean/Util/TermUnsafe.lean
semorrison: I moved the syntax and docstring to Init/Notation.lean (is that correct?)

I changed the elab statement to use `builtin_term_elab`.

src/Lean/Util/TermUnsafe.lean
semorrison: I found in Lean.Elab.Term:
```
def Lean.Elab.Term.mkAuxName (suffix : Name) : TermElabM Name := do
  match (← read).declName? with
  | none          => throwError "auxiliary declaration cannot be created when declaration name is not available"
  | some declName => Lean.mkAuxName (declName ++ suffix) 1
```

which is calling another `mkAuxName` in Lean.MonadEnv
```
private partial def mkAuxNameAux (env : Environment) (base : Name) (i : Nat) : Name :=
  let candidate := base.appendIndexAfter i
  if env.contains candidate then
    mkAuxNameAux env base (i+1)
  else
    candidate

def mkAuxName [Monad m] [MonadEnv m] (baseName : Name) (idx : Nat) : m Name := do
  return mkAuxNameAux (← getEnv) baseName idx
```

I am inclined not to spend cycles trying to unify this, just mark this as private (I can mention the existence of the others), and keep moving.

PR #3282, opened 2024-02-08
chore: upstream Std.Data.Array.Init.Basic

joehendrix:
This migrates the handful of array operations in [Std.Data.Array.Init.Basic](https://github.com/leanprover/std4/blob/main/Std/Data/Array/Init/Basic.lean).

### DIFF
--- a/src/Init/Data/Array/Basic.lean
+++ b/src/Init/Data/Array/Basic.lean
@@ -21,6 +21,21 @@ def mkArray {α : Type u} (n : Nat) (v : α) : Array α := {
   data := List.replicate n v
 }
 
+/--
+`ofFn f` with `f : Fin n → α` returns the list whose ith element is `f i`.
+```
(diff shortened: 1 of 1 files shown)


src/Init/Data/Array/Basic.lean
```diff
@@ -487,6 +506,11 @@ def elem [BEq α] (a : α) (as : Array α) : Bool :=
 def toList (as : Array α) : List α :=
   as.foldr List.cons []
 
+/-- Prepends an `Array α` onto the front of a list.  Equivalent to `as.toList ++ l`. -/
+@[inline]
+def toListAppend (as : Array α) (l : List α) : List α :=
+  as.foldr List.cons l
+
```
semorrison, line 513: Just noting that while this one looks a little silly, it is quite useful in setting up tail-recursive versions of List operations in Std.

semorrison (approved):

PR #3287, opened 2024-02-09
doc: mention maxSteps setting in its error message

BoltonBailey:
This small PR changes the error message that triggers when simp reaches the `maxSteps` parameter to mention that this is a configurable setting and say what the name of the setting is.


Kha:
I don't think people will know what to do from this description alone

PR #3290, opened 2024-02-09
chore: upstream Std.CodeAction.* and Std.Tactic.GuardMsgs

kim-em:


semorrison:
Not clear to me which directory / namespace this should be in, so it's top-level for now.

digama0:
> Not clear to me which directory / namespace this should be in, so it's top-level for now.

I think this could go in `Lean.Server`

Kha:
Marking as draft for now; no immediate need for upstreaming and I'd like to consider the action kinds in more detail before we do

src/Lean/Elab/Tactic/GuardMsgs.lean
```diff
@@ -0,0 +1,133 @@
...
+
+/-! `#guard_msgs` command for testing commands
+
+This module defines a command to test that another command produces the expected messages.
+See the docstring on the `#guard_msgs` command.
+-/
+
+open Lean Parser.Tactic Elab Command
+
+namespace Std.Tactic.GuardMsgs
```
david-christiansen, line 17: Perhaps we should take this opportunity to reconsider this namespace? It does leak out from time to time, and `#guard_msgs` is not a tactic, nor is its usefulness restricted to tactics.

src/Lean/Elab/Tactic/GuardMsgs.lean
```diff
@@ -0,0 +1,133 @@
...
+/-- Gives a string representation of a message without source position information.
+Ensures the message ends with a '\n'. -/
+private def messageToStringWithoutPos (msg : Message) : IO String := do
+  let mut str ← msg.data.toString
+  unless msg.caption == "" do
+    str := msg.caption ++ ":\n" ++ str
+  match msg.severity with
+  | MessageSeverity.information => str := "info: " ++ str
+  | MessageSeverity.warning     => str := "warning: " ++ str
+  | MessageSeverity.error       => str := "error: " ++ str
```
david-christiansen, line 28: If `str` begins with a newline, then this ends up with trailing whitespace that has a tendency to get auto-deleted by editors. Should the space be suppressed if `str` begins with `\n`?

src/Lean/Elab/Tactic/GuardMsgs.lean
```diff
@@ -0,0 +1,133 @@
...
+    | _ => throwErrorAt spec "Invalid #guard_msgs specification"
+  else
+    return fun _ => .check
+
+/-- An info tree node corresponding to a failed `#guard_msgs` invocation,
+used for code action support. -/
+structure GuardMsgFailure where
+  /-- The result of the nested command -/
+  res : String
+  deriving TypeName
```
david-christiansen, line 73:
```suggestion
deriving TypeName
```

src/Lean/Elab/Tactic/GuardMsgs.lean
semorrison: Done.

semorrison:
Where should the implementation of `#guard_msgs` live? Currently I have it in `Lean/Elab/Tactic/GuardMsgs.lean`, but it is a command, not a tactic.

david-christiansen:
> Where should the implementation of `#guard_msgs` live? Currently I have it in `Lean/Elab/Tactic/GuardMsgs.lean`, but it is a command, not a tactic.

What about something like `Lean/Testing/GuardMsgs.lean`? And a corresponding namespace?

semorrison:
This is not ready to go: the code action is not working because it needs to be made builtin.

src/Lean/Server/CodeActions/Provider.lean
```diff
@@ -0,0 +1,185 @@
...
+  cursor was indented to align with `baz` then we would select the `bar; baz` sequence instead.
+
+* `range`: the cursor selection. We do not do much with range selections; if a range selection
+  covers more than one tactic then we abort.
+
+* `root`: the root syntax to process
+
+The return value is either a selected tactic, or a selected point in a tactic sequence.
+-/
+partial def findTactic? (preferred : String.Pos → Bool) (range : String.Range)
```
Kha, line 76: @semorrison There are some remnants of the postponed `tactic_code_action`

PR #3291, opened 2024-02-09
chore: upstream Std's linter framework

kim-em:
It is the second linter framework from Std. It runs *after* declarations have been made, and e.g. can verify global properties of simp sets.

WIP, there are a number of problems here:

* Where does this go? 
* What namespace does it live in?
* The `simp` linter, which I've included here, uses `open private` to get access to simp internals.
* Do we actually want this? We could proceed without it for now, at the cost of separating some `@[nolint]` attributes from their declarations.

### DIFF
--- a/src/Lean/Linter/OtherLinter/Basic.lean
+++ b/src/Lean/Linter/OtherLinter/Basic.lean
@@ -1,0 +1,132 @@
+/-
+Copyright (c) 2020 Floris van Doorn. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Floris van Doorn, Robert Y. Lewis, Gabriel Ebner
+-/
+import Lean.Structure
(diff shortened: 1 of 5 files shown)


semorrison:
To be discussed at the 12 Feb all-hands.

digama0:
(For clarity, I will be calling the two linter frameworks "syntax linters" and "environment linters" from now on.)

Remark: In the last meeting we discussed the possibility of "finalizers" which can be registered similarly to syntax linters and trigger at the end of the file. It is possible that we can use finalizers as a substitute for environment linters. The main downside is that if we were to do this with all of mathlib's linters it would have the effect of running linting at the same time as the build, which would mean lost parallelism unless we can generate the .olean file before running finalizers and signal to lake that it can continue with dependent files even though the lean process has not exited yet.

PR #3292, opened 2024-02-09
chore: upstream rcases

kim-em:
This moves the `rcases` and `obtain` tactics from Std, and makes them built-in tactics.

We will separately move the test cases from Std after #3297 (`guard_expr`).

### DIFF
--- a/src/Init.lean
+++ b/src/Init.lean
@@ -8,6 +8,7 @@ import Init.Prelude
 import Init.Notation
 import Init.Tactics
 import Init.TacticsExtra
+import Init.RCases
 import Init.Core
 import Init.Control
 import Init.Data.Basic
--- a/src/Init/RCases.lean
+++ b/src/Init/RCases.lean
@@ -1,0 +1,192 @@
+/-
+Copyright (c) 2017 Mario Carneiro. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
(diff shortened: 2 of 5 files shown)


digama0:
I just want to make a remark about maintenance here for this and other PRs in this series: Please continue to route bug reports on these files with my name on them to me. I'm fine with them leaving the std repo but I hope this doesn't interfere with my continued maintenance of the files.

leodemoura:
@digama0 This is a valid concern. Thanks for sharing it. Happy to hear you are willing to continue maintaining the file. 
`rcases` has a nontrivial syntax, if you are planning to keep tweaking it, I suggest we avoid using `rcases` in this repo. Otherwise, you may waste a lot of time dealing with bootstrapping issues.

leodemoura:
@semorrison I added only a few sanity checks. I agree with you that we should move the test file from `Std` as soon as we move `GuardExpr.lean`.

digama0:
The syntax of rcases is not likely to change, unless it is changed as a part of upstreaming now (as was done for `fun.` syntax). But there are routinely issues that appear in how metavariables are handled, or how things are unfolded, whether it works on other inductive-like types, appearance of hover text, etc., much the same as with the `cases` tactic. Furthermore there is a big TODO still remaining in the code for implementing the `rcases?` tactic.

PR #3298, opened 2024-02-10
feat: lake: alternative TOML config

tydeu:
Adds an alternative TOML configuration format to Lake. 

* Uses TOML v1.0.0 and is fully specification compliant (tested via [toml-test v1.4.0](https://github.com/toml-lang/toml-test/tree/v1.4.0)).
* Supports package configuration options, Lean libraries, Lean executables, and dependencies.
* TOML configurations can be generated for new projects via `lake new|init <pkg> <template>.toml`.
* Supported configurations can be converted to/from TOML via `lake translate-config <lang>`.

### DIFF
--- a/src/lake/Lake/Build/Trace.lean
+++ b/src/lake/Lake/Build/Trace.lean
@@ -3,6 +3,7 @@ Copyright (c) 2021 Mac Malone. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mac Malone
 -/
+import Lake.Util.Newline
 
 open System
 namespace Lake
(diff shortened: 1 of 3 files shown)


src/lake/Lake/CLI/Init.lean
```diff
@@ -108,17 +134,33 @@ lean_lib {libRoot} where
   -- add any library configuration options here
 "
 
+def mathTomlConfigFileContents (pkgName libRoot : String) :=
+s!"name = {repr pkgName}
+defaultTargets = [{repr libRoot}]
+
+[leanOptions]
+pp.unicode.fun = true # pretty-prints `fun a ↦ b`
+pp.proofs.withType = false
```
Kha, line 143: Now the default! :)

semorrison:
Sounds great. Looking forward to seeing this in a nightly so we can start translating ahead of v4.8.

PR #3299, opened 2024-02-10
chore: `rwa` tactic macro

leodemoura:


src/Init/Tactics.lean
```diff
@@ -323,9 +323,14 @@ syntax (name := eqRefl) "eq_refl" : tactic
 `rfl` tries to close the current goal using reflexivity.
 This is supposed to be an extensible tactic and users can add their own support
 for new reflexive relations.
+
+Remark: `rfl` is an extensible tactic. We later add `macro_rules` to try different
+reflexivity theorems (e.g., `Iff.rfl`).
```
nomeata, line 328: “We later add” doesn't quite work well for a user facing docstring, does it?

src/Init/Tactics.lean
```diff
@@ -432,13 +437,17 @@ syntax (name := rewriteSeq) "rewrite" (config)? rwRuleSeq (location)? : tactic
...
   match s with
   | `(rwRuleSeq| [$rs,*]%$rbrak) =>
     -- We show the `rfl` state on `]`
     `(tactic| (rewrite $(c)? [$rs,*] $(l)?; with_annotate_state $rbrak (try (with_reducible rfl))))
   | _ => Macro.throwUnsupported
 
+/-- `rwa` calls `rw`, then closes any remaining goals using `assumption`. -/
+macro "rwa " rws:rwRuleSeq loc:(location)? : tactic =>
+  `(tactic| (rw $rws:rwRuleSeq $[$loc:location]?; assumption))
+
```
nomeata, line 450: The docstring says “any goals” but the code looks like it will close exactly one goal. Should the docstring say “the remaining goal”?

tests/lean/runTacticMustCatchExceptions.lean.expected.out
```diff
@@ -1,27 +1,36 @@
...
+has type
+  ?m ↔ ?m : Prop
+but is expected to have type
+  a + b ≤ b : Prop
+runTacticMustCatchExceptions.lean:11:14-11:17: error: type mismatch
+  Iff.rfl
+has type
+  ?m ↔ ?m : Prop
+but is expected to have type
+  b ≤ 2 : Prop
```
nomeata, line 36: This seems to be a slight regression of error reporting, but for something so elementary as `rfl` not irrelevant. 

If macro rules are used, which tactic decides about the error message? Last or first? Can we somehow register a `fail "trivial equality (or equivalence) expected"` or simply `fail "rfl failed"` or something like this as the last tactic to try with `rfl`?

PR #3300, opened 2024-02-10
chore: add documentation for the `String.iterator` API

AdrienChampion:
Adds documentation to the `String.Iterator` API, mentored by @eric-wieser and @david-christiansen

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -2375,10 +2375,7 @@ Indexing a `String` by a byte position is constant-time, while codepoint
 positions need to be translated internally to byte positions in linear-time.
 
 A byte position `p` is *valid* for a string `s` if `0 ≤ p ≤ s.endPos` and `p`
-lies on a UTF8 byte boundary. This notion is properly introduced in [std's
-`String.Pos.Valid`][valid].
-
-[valid]: https://leanprover-community.github.io/mathlib4_docs/Std/Data/String/Lemmas.html#String.Pos.Valid
+lies on a UTF8 byte boundary.
 -/
 structure String.Pos where
   /-- Get the underlying byte index of a `String.Pos` -/


src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,111 @@ where go (acc : String) (s : String) : List String → String
...
 instance : SizeOf String.Iterator where
   sizeOf i := i.1.utf8ByteSize - i.2.byteIdx
 
 theorem Iterator.sizeOf_eq (i : String.Iterator) : sizeOf i = i.1.utf8ByteSize - i.2.byteIdx :=
   rfl
 
 namespace Iterator
+/-- The *full* string the iterator is for. -/
 def toString : Iterator → String
   | ⟨s, _⟩ => s
```
eric-wieser, line 324: I'm puzzled by how we have both this and `Iterator.s`; that is, I think this is the same as
```suggestion
/-- The *full* string the iterator is for. -/
abbrev toString := Iterator.s
```
Should the docstring explain which one is preferred?

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,111 @@ where go (acc : String) (s : String) : List String → String
...
 
+/-- Number of bytes remaining in the iterator. -/
 def remainingBytes : Iterator → Nat
   | ⟨s, i⟩ => s.endPos.byteIdx - i.byteIdx
 
+/-- Current character position. -/
```
eric-wieser, line 330:
```suggestion
@[inherit_doc Iterator.i]
```

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,111 @@ where go (acc : String) (s : String) : List String → String
...
+/-- The notion of size for a string iterator is the number of bytes remaining. -/
 instance : SizeOf String.Iterator where
   sizeOf i := i.1.utf8ByteSize - i.2.byteIdx
 
 theorem Iterator.sizeOf_eq (i : String.Iterator) : sizeOf i = i.1.utf8ByteSize - i.2.byteIdx :=
   rfl
 
 namespace Iterator
+/-- The *full* string the iterator is for. -/
 def toString : Iterator → String
```
eric-wieser, line 323:
```suggestion
@[inherit_doc Iterator.s]
def toString : Iterator → String
```

src/Init/Data/String/Basic.lean
AdrienChampion: Same for `forward` and `nextn`

eric-wieser (commented):
A few more places where there are duplicate definitions for `inherit_doc`

src/Init/Data/String/Basic.lean
AdrienChampion: Could be an `abbrev` maybe?

src/Init/Data/String/Basic.lean
AdrienChampion: Could also be an `abbrev`, as in your first comment?

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,111 @@ where go (acc : String) (s : String) : List String → String
...
+
+Typically created by `s.iter`, where `s` is a `String`. -/
 structure Iterator where
+  /-- The string the iterator is for. -/
   s : String
+  /-- The current position.
+
+  This position is not necessarily legal for the string, for instance if one keeps calling
+  `Iterator.next` when `Iterator.atEnd` is true. If the position is not a legal one, then the
+  current character is `(default : Char)`, similar to `String.get` on an illegal position. -/
```
david-christiansen, line 303:
```suggestion
  This position is not necessarily legal for the string, for instance if one keeps calling
  `Iterator.next` when `Iterator.atEnd` is true. If the position is not valid, then the
  current character is `(default : Char)`, similar to `String.get` on an invalid position. -/
```

The other docs for strings refer to positions as "valid" or "invalid" rather than "legal" or "illegal" - I think it's good to be consistent

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,111 @@ where go (acc : String) (s : String) : List String → String
...
   i : Pos
   deriving DecidableEq
 
+/-- The iterator positioned at the first character of a string (if any). -/
```
david-christiansen, line 307:
```suggestion
/-- Create an iterator that points at the first character of the string `s` (if any). -/
```

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,111 @@ where go (acc : String) (s : String) : List String → String
...
 def mkIterator (s : String) : Iterator :=
   ⟨s, 0⟩
 
+@[inherit_doc mkIterator]
 abbrev iter := mkIterator
 
+/-- The notion of size for a string iterator is the number of bytes remaining. -/
```
david-christiansen, line 314:
```suggestion
/-- The size of a string iterator is the number of bytes remaining. -/
```
I think this reads better, but the original is not incorrect :)

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,111 @@ where go (acc : String) (s : String) : List String → String
...
 def remainingBytes : Iterator → Nat
   | ⟨s, i⟩ => s.endPos.byteIdx - i.byteIdx
 
+/-- Current character position. -/
 def pos : Iterator → Pos
   | ⟨_, i⟩ => i
 
+/-- The character at the current position.
+
+On an illegal position, returns `(default : Char)`. -/
```
david-christiansen, line 336:
```suggestion
On an invaid position, returns `(default : Char)`. -/
```

Or even
```suggestion
If the iterator points outside the bounds of the string, this returns `(default : Char)`. -/
```

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,111 @@ where go (acc : String) (s : String) : List String → String
...
 def curr : Iterator → Char
   | ⟨s, i⟩ => get s i
 
+/-- Increases the iterator's position by one, unconditionally. -/
```
david-christiansen, line 340:
```suggestion
/-- Increases the iterator's position by one character, unconditionally. -/
```

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,111 @@ where go (acc : String) (s : String) : List String → String
...
 
+/-- True if the iterator is not past the string's last character. -/
 def hasNext : Iterator → Bool
   | ⟨s, i⟩ => i.byteIdx < s.endPos.byteIdx
 
+/-- True if the position is not zero. -/
 def hasPrev : Iterator → Bool
   | ⟨_, i⟩ => i.byteIdx > 0
 
+/-- Replaces the current character in the string.
```
david-christiansen, line 362:
```suggestion
/-- Construct an iterator into a string in which the current character is replaced with some other character.

If the iterator contains the only reference to the string, then this will mutate the string rather than allocating a new one.
```

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,111 @@ where go (acc : String) (s : String) : List String → String
...
+/-- The remaining characters in an iterator, as a string. -/
 def remainingToString : Iterator → String
   | ⟨s, i⟩ => s.extract i s.endPos
 
+@[inherit_doc forward]
 def nextn : Iterator → Nat → Iterator
   | it, 0   => it
   | it, i+1 => nextn it.next i
 
+/-- Moves the iterator's position several characters back. -/
```
david-christiansen, line 397:
```suggestion
/-- Moves the iterator's position several characters back.

If there aren't enough characters in the string, then the iterator is moved to the beginning. -/
```

src/Init/Data/String/Basic.lean
AdrienChampion: Being outside of the bounds of the string is not the only way to have an invalid position if there are UTF8 characters.

```lean
def main : IO Unit := do
  let i : String.Iterator := ⟨"🙀", ⟨1⟩⟩
  println! "curr: `{i.curr}`"
  -- curr: `A`
```

This should not happen on an iterator that's used the way it's supposed to be, but there's no proof that the byte index actually corresponds to a character so...

src/Init/Data/String/Basic.lean
digama0: I think the wording here should discourage invalid positions a bit more strongly. There is no proof witness in the type asserting that the position is valid, but it should be unequivocally a logic bug to have such a misaligned iterator and all string functions should be allowed to have arbitrary behavior in such cases.

digama0, line 303: I think the wording here should discourage invalid positions a bit more strongly. There is no proof witness in the type asserting that the position is valid, but it should be unequivocally a logic bug to have such a misaligned iterator and all string functions should be allowed to have arbitrary behavior in such cases.

src/Init/Data/String/Basic.lean
digama0: I think we could just rename the field names

src/Init/Data/String/Basic.lean
digama0:
```suggestion
/-- Increases the iterator's position by one character, unconditionally.

It is only valid to call this function if `Iterator.atEnd` is false, otherwise the resulting iterator will be invalid. -/
```

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,111 @@ where go (acc : String) (s : String) : List String → String
...
+/-- Extracts the substring between the positions of two iterators.
+
+Returns the empty string if the iterators are for different strings, or if the position of the first
+iterator is past the position of the second iterator. -/
 def extract : Iterator → Iterator → String
   | ⟨s₁, b⟩, ⟨s₂, e⟩ =>
     if s₁ ≠ s₂ || b > e then ""
     else s₁.extract b e
 
+/-- Moves the iterator's position several characters forward. -/
```
digama0, line 383:
```suggestion
/-- Moves the iterator's position several characters forward.

If there aren't enough characters in the string, the resulting iterator will be invalid. -/
```

src/Init/Data/String/Basic.lean
eric-wieser: That sounds like it probably belongs in a separate PR

src/Init/Data/String/Basic.lean
AdrienChampion: I understand what you're saying but I find this a bit tricky. I think one expects "position `i` is valid for string `s`" to mean that `i` is in-bounds, but then `"".iter` is invalid, and so is `"1".iter.next` although `"1".iter.atEnd` is false.

So I assume you mean that `i` is valid for `s` if `i` is in-bounds or `i` is the successor of the index of the last character in `s`. Is that what you mean?

digama0: A `p : String.Pos` is valid for `s : String` when `0 <= p <= s.endPos` and `p` lies on a UTF8 byte boundary. This is defined in std.

eric-wieser: Yes, there are two options of validity here: valid for character access, and valid for iterator comparison (which also allows `atEnd`).

Maybe the C++ documentation has some terminology for this distinction that we can copy.

src/Init/Data/String/Basic.lean
AdrienChampion: I pushed a slightly reworded version.

src/Init/Data/String/Basic.lean
digama0: In C++ a pointer is considered valid (not immediate UB) if it is within the allocation or one-past-the-end. I'd rather not borrow the one-past-the-end terminology though, it's very awkward and reinforces the idea that a pointer points to an element rather than a separation point "between characters". Std currently does not have any terminology for "valid and not at the end"; if necessary it would just say that as two things. [`String.get_of_valid`](https://leanprover-community.github.io/mathlib4_docs/find/?pattern=String.get_of_valid#src) is not stated in a way that directly uses `String.Pos.Valid` because it needs to know more specifically what is the list of characters on the right of the partition.

AdrienChampion: How about this, with the exceptions mentioned properly documented?

```lean
/--

An iterator is *valid* if the position `i` is valid for the string `s`, meaning `0 ≤ i ≤ s.endPos`
and `i` lies on a UTF8 by boundary.

Most operations on iterators return arbitrary values if the iterator is not valid. The functions in
the `String.Iterator` API should prevent you from dealing with invalid iterators, with two
exceptions:

- `Iterator.next iter` is invalid if `iter` is already at the end of the string (`iter.atEnd` is
  true), and
- `Iterator.forward iter n`/`Iterator.nextn iter n` if `n` is strictly greater than the number of
  remaining characters. -/
```

digama0: The definition of string position validity should also be moved or copied to the `String.Pos` docs.

AdrienChampion: Done

AdrienChampion: Done

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,135 @@ where go (acc : String) (s : String) : List String → String
...
+
+/-- The character at the current position.
 
+On an invalid position, returns `(default : Char)`. -/
 def curr : Iterator → Char
   | ⟨s, i⟩ => get s i
 
+/-- Moves the iterator's position forward by one character, unconditionally.
+
+It is only valid to call this function if the iterator is not a the end of the string, *i.e.*
```
digama0, line 356:
```suggestion
It is only valid to call this function if the iterator is not at the end of the string, *i.e.*
```

src/Init/Data/String/Basic.lean
AdrienChampion: Are you sure about the "Constructs an iterator..." description?

I find it verbose and daunting as a first description, IMHO it hides the human-level abstraction of the function which is the current description.

I'm also worried it can be confusing because it is inconsistent with other iterator-transforming functions such as `Iterator.next`, which does not really "move the position", but rather (in general) "constructs an iterator ...".

AdrienChampion: A reasonable compromise could be to have your "Constructs an iterator ..." below the current description, as a more detailed explanation.

I still think that's a bit inconsistent if, for instance, `Iterator.next` does not have a similar clarification, though I think most readers at this point should have at least an intuitive understanding of what "purely functional" means and don't need the clarification anyways.

src/Init/Data/String/Basic.lean
AdrienChampion: Pushed a reworded version, not sure what else to do about the footgun aspect, did you have something in mind?

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,135 @@ where go (acc : String) (s : String) : List String → String
...
+Most operations on iterators return arbitrary values if the iterator is not valid. The functions in
+the `String.Iterator` API should prevent you from dealing with invalid iterators, with two
+exceptions:
+
+- `Iterator.next iter` is invalid if `iter` is already at the end of the string (`iter.atEnd` is
+  true), and
+- `Iterator.forward iter n`/`Iterator.nextn iter n` is invalid if `n` is strictly greater than the
+  number of remaining characters.
+
+[valid]: https://leanprover-community.github.io/mathlib4_docs/Std/Data/String/Lemmas.html#String.Pos.Valid
```
david-christiansen, line 310: We shouldn't link to external generated doc HTML from docstrings - this will be too fragile. We need to have notation for stable doc links, but we don't have it yet, so for now we should just have the names here.

david-christiansen:
```suggestion
```

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,135 @@ where go (acc : String) (s : String) : List String → String
...
   | []      => acc
 
-/-- Iterator for `String`. That is, a `String` and a position in that string. -/
+/-- Iterator over the characters (`Char`) of a `String`.
+
+Typically created by `s.iter`, where `s` is a `String`.
+
+An iterator is *valid* if the position `i` is *valid* for the string `s`, meaning `0 ≤ i ≤ s.endPos`
+and `i` lies on a UTF8 byte boundary consistently with [std's `String.Pos.Valid`][valid]. If `i =
+s.endPos`, the iterator is *at the end (of `s`)*.
```
david-christiansen, line 299: I like the use of italic to introduce technical terms here, but I think it should do only that, and not emphasize being at the end.
```suggestion
An iterator is *valid* if the position `i` is *valid* for the string `s`, meaning `0 ≤ i ≤ s.endPos`
and `i` lies on a UTF8 byte boundary. If `i = s.endPos`, the iterator is at the end (of `s`).
```

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,135 @@ where go (acc : String) (s : String) : List String → String
...
+
+Most operations on iterators return arbitrary values if the iterator is not valid. The functions in
+the `String.Iterator` API should prevent you from dealing with invalid iterators, with two
```
david-christiansen, line 302:
```suggestion
the `String.Iterator` API should rule out the creation of invalid iterators, with two
```

david-christiansen: Working on consistent (and less) use of personal pronouns in the reference docs

src/Init/Data/String/Basic.lean
david-christiansen: I think there's no need to emphasize `*at the end of the string*` here - this is an ordinary and expected state for an iterator to have.

How about this minor edit:
```
An iterator is *valid* if its current position `i` is valid for the string `s`, meaning `0 ≤ i ≤ s.endPos`
and `i` lies on a UTF8 byte boundary. If `i = s.endPos`, the iterator is at the end of the string.

Most operations on iterators return arbitrary values if the iterator is not valid. The functions in
the `String.Iterator` API should rule out the creation of invalid iterators, with two
exceptions:

- `Iterator.next iter` is invalid if `iter` is already at the end of the string (`iter.atEnd` is
  true), and
- `Iterator.forward iter n`/`Iterator.nextn iter n` if `n` is strictly greater than the number of
  remaining characters.
```

src/Init/Data/String/Basic.lean
david-christiansen: I think it's fine to not say "contructs" here.

I think the bit about mutation is important to have, but the original first paragraph is fine.

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,135 @@ where go (acc : String) (s : String) : List String → String
...
+
+Does nothing if the iterator is at the end of the string. If the iterator contains the only
+reference to its string, this function will mutate the string in-place instead of allocating a new
+one. -/
 def setCurr : Iterator → Char → Iterator
   | ⟨s, i⟩, c => ⟨s.set i c, i⟩
 
+/-- Moves the iterator's position to the end of the string.
+
+Note that `i.toEnd.atEnd` is always true. -/
```
david-christiansen, line 389:
```suggestion
Note that `i.toEnd.atEnd` is always `true`. -/
```

src/Init/Data/String/Basic.lean
```diff
@@ -290,73 +290,135 @@ where go (acc : String) (s : String) : List String → String
...
+Returns the empty string if the iterators are for different strings, or if the position of the first
+iterator is past the position of the second iterator. -/
 def extract : Iterator → Iterator → String
   | ⟨s₁, b⟩, ⟨s₂, e⟩ =>
     if s₁ ≠ s₂ || b > e then ""
     else s₁.extract b e
 
+/-- Moves the iterator's position several characters forward.
+
+Calling this function is only legal if the number of character to skip is less than or equal to the
```
david-christiansen, line 404:
```suggestion
The resulting iterator is only valid if the number of characters to skip is less than or equal to the
```

src/Init/Prelude.lean
```diff
@@ -2373,6 +2373,12 @@ Codepoint positions (counting the Unicode codepoints rather than bytes)
 are represented by plain `Nat`s instead.
 Indexing a `String` by a byte position is constant-time, while codepoint
 positions need to be translated internally to byte positions in linear-time.
+
+A byte position `p` is *valid* for a string `s` if `0 ≤ p ≤ s.endPos` and `p`
+lies on a UTF8 byte boundary. This notion is properly introduced in [std's
+`String.Pos.Valid`][valid].
+
+[valid]: https://leanprover-community.github.io/mathlib4_docs/Std/Data/String/Lemmas.html#String.Pos.Valid
```
david-christiansen, line 2381:
```suggestion
lies on a UTF8 byte boundary.
```
We shouldn't be linking to external URLs like this before we get a way to keep them working over time, even if it would be nice.

src/Init/Data/String/Basic.lean
digama0: I think if we use the same URL as `docs#String.Pos.Valid`, that is https://leanprover-community.github.io/mathlib4_docs/find/?pattern=String.Pos.Valid#src , that is as good as we are going to get for the near term, and removing the links does users a disservice. Moreover, if and when we get a stable doc link syntax, it won't work here anyway because this is a *forward reference* to a definition in another repository.

src/Init/Data/String/Basic.lean
digama0: I believe the use of italics for "at the end" was also intended to be introducing a technical term, although it sounds a bit too colloquial to pass as a technical term.

src/Init/Prelude.lean
digama0: Idem ditto. We've been using permalinks on zulip for years.

src/Init/Data/String/Basic.lean
AdrienChampion: Okay, the discussion on mutation (assuming that's the "immutable beans" bit) is already in there, slightly reworded.

src/Init/Data/String/Basic.lean
AdrienChampion: > I believe the use of italics for "at the end" was also intended to be introducing a technical term, although it sounds a bit too colloquial to pass as a technical term.

Correct.

src/Init/Data/String/Basic.lean
david-christiansen: We don't want to have links to external domains and downstream projects in Lean - it makes everything too fragile. I totally get the concern here! But we should provide enough information for people to understand the validity of the iterator in these docstrings, not put it elsewhere, and I think we do that now. And this is largely your doing - thanks!

src/Init/Data/String/Basic.lean
david-christiansen: I think this API is just a footgun. We could look into changing it, but that's outside the scope of this docs PR!

src/Init/Data/String/Basic.lean
david-christiansen: Do we need that as a special technical term? It seems to me that we use "at the end" in a totally ordinary sense here, while "valid" is being restricted to a very specific subset of the meanings that "valid" can take on.

src/Init/Data/String/Basic.lean
AdrienChampion: Do we agree to resolve this conversation then?

src/Init/Data/String/Basic.lean
AdrienChampion: Done, with the link to std4 removed

src/Init/Data/String/Basic.lean
AdrienChampion: Not italized any more

src/Init/Data/String/Basic.lean
AdrienChampion: I pushed that change, but I have to say I'm not a fan of it as I don't feel like it really clarifies anything for the reader, and is cumbersome to make consistent. For instance, there are a bunch of functions with a documentation looking like

```lean
/-- True if the iterator is past the string's last character. -/
def atEnd : Iterator → Bool
  | ⟨s, i⟩ => i.byteIdx ≥ s.endPos.byteIdx
```

We can't just write "`True` when ..." since that would be a `Prop`, so we need to say something like "Yields/Produces/Returns/Is `true` when ...".

Do you want me to make these changes too?

src/Init/Data/String/Basic.lean
digama0: For now, all doc-gen generated lean documentation is on the leanprover-community web site, for better or worse. If and when we start generating doc-gen for leanprover projects somewhere on the FRO website we can retarget these URLs if desired. This is a link from core to std, hardly "external". In fact, ideally users would not even need to know that half of the String API is in `Init` (in core) and the other half is in `Std`. Validity is an essential characterizing property of these types, and I think it is important to point to the definition itself by name. (If I could, I would put validity directly into the types themselves and save users the trouble of having to ever deal with invalid iterators. But this has its own issues.)

src/Init/Data/String/Basic.lean
david-christiansen: Yes, I think so. It'd be good to improve this, but at least a documented footgun is better than an undocumented one.

src/Init/Data/String/Basic.lean
david-christiansen: I don't think there's a need to go overboard here, and there's good arguments for either style. I think it makes things slightly more clear here, but only very slightly :)

You're doing the work here, let's defer to you on this one.

AdrienChampion: > I don't think there's a need to go overboard here, and there's good arguments for either style.

Agreed. Your version is the current one, might as well keep it :)

src/Init/Data/String/Basic.lean
david-christiansen: From what I can see, it's still using the "legal" terminology...

david-christiansen:
I went ahead and resolved the two outstanding issues. Thanks again!

PR #3305, opened 2024-02-12
chore: upstream repeat/split_ands/subst_eqs

kim-em:
Small tactics used in the implementation of `ext`.

### DIFF
--- a/src/Init/Tactics.lean
+++ b/src/Init/Tactics.lean
@@ -208,6 +208,28 @@ the first matching constructor, or else fails.
 syntax (name := constructor) "constructor" : tactic
 
 /--
+Applies the second constructor when
+the goal is an inductive type with exactly two constructors, or fails otherwise.
+```
+example : True ∨ False := by
+  left
+  trivial
+```
+-/
+syntax (name := left) "left" : tactic
+
+/--
+Applies the second constructor when
+the goal is an inductive type with exactly two constructors, or fails otherwise.
+```
+example {p q : Prop} (h : q) : p ∨ q := by
+  right
+  exact h
+```
+-/
+syntax (name := right) "right" : tactic
+
+/--
 * `case tag => tac` focuses on the goal with case name `tag` and solves it using `tac`,
   or else fails.
 * `case tag x₁ ... xₙ => tac` additionally renames the `n` most recent hypotheses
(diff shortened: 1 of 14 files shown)


david-christiansen:
Oh no, I didn't realize that #3306 was based on this one! There's some comments there on docstrings and names.

leodemoura (approved):

src/Init/Tactics.lean
```diff
@@ -872,6 +872,22 @@ The tactic `nomatch h` is shorthand for `exact nomatch h`.
...
+
+/--
+`repeat1' tac` applies `tac` to main goal at least once. If the application succeeds,
+the tactic is applied recursively to the generated subgoals until it eventually fails.
+-/
+syntax (name := repeat1') "repeat1' " tacticSeq : tactic
+
+/-- `split_ands` applies `And.intro` until it does not make progress. -/
+syntax "split_ands" : tactic
+macro_rules | `(tactic| split_ands) => `(tactic| repeat' refine And.intro ?_ ?_)
```
semorrison, line 889:
```suggestion
/-- `and_intros` applies `And.intro` until it does not make progress. -/
syntax "and_intros" : tactic
macro_rules | `(tactic| and_intros) => `(tactic| repeat' refine And.intro ?_ ?_)
```

tests/lean/run/split_and.lean
```diff
@@ -0,0 +1,7 @@
+example (hp : p) (hq : q) (hr : r) : (p ∧ q) ∧ (q ∧ (r ∧ p)) := by
+  split_ands
```
semorrison, line 2:
```suggestion
  and_intros
```

src/Lean/Elab/Tactic/BuiltinTactic.lean
```diff
@@ -324,6 +324,12 @@ def forEachVar (hs : Array Syntax) (tac : MVarId → FVarId → MetaM MVarId) :
 @[builtin_tactic Lean.Parser.Tactic.substVars] def evalSubstVars : Tactic := fun _ =>
   liftMetaTactic fun mvarId => return [← substVars mvarId]
 
+/--
+`subst_eq` repeatedly substitutes according to the equality proof hypotheses in the context,
+replacing the left side of the equality with the right, until no more progress can be made.
+-/
```
kmill, line 330: This docstring isn't accurate. The `Lean.MVarId.substEqs` function will try doing `cases` on every `Eq` and `HEq` in the local context. That means it even will succeed in cases like `Nat.succ a = Nat.succ b`, substituting using `a = b`.

It doesn't have any preference for the LHS or the RHS -- if either side is a variable, it will be substituted.

PR #3306, opened 2024-02-12
chore: upstream `ext` tactic

kim-em:


src/Lean/Elab/Command.lean
```diff
@@ -503,6 +504,45 @@ def expandDeclId (declId : Syntax) (modifiers : Modifiers) : CommandElabM Expand
 
 end Elab.Command
 
+open Elab Command MonadRecDepth
+
+/--
+Lift an action in `CommandElabM` into `CoreM`, updating the traces and the environment.
+This does not preserve things like `open` and `namespace` declarations.
```
david-christiansen, line 511: What does "preserve" mean in this context?

Perhaps this means the same thing?
```suggestion
Commands that modify the processing of subsequent commands, such as `open` and `namespace`
declarations, only have an effect for the remainder of the `CommandElabM` computation passed here,
and do not affect subsequent commands.
```

src/Lean/Elab/Command.lean
```diff
@@ -503,6 +504,45 @@ def expandDeclId (declId : Syntax) (modifiers : Modifiers) : CommandElabM Expand
 
 end Elab.Command
 
+open Elab Command MonadRecDepth
+
+/--
+Lift an action in `CommandElabM` into `CoreM`, updating the traces and the environment.
```
david-christiansen, line 510:
```suggestion
Lifts an action in `CommandElabM` into `CoreM`, updating the traces and the environment.
```

src/Lean/Elab/Command.lean
```diff
@@ -503,6 +504,45 @@ def expandDeclId (declId : Syntax) (modifiers : Modifiers) : CommandElabM Expand
...
+This does not preserve things like `open` and `namespace` declarations.
+-/
+def liftCommandElabM (k : CommandElabM α) : CoreM α := do
```
david-christiansen, line 513: The parameter name `k` suggests a continuation, but this is not that. What about calling it `commands` instead?

src/Lean/Elab/Command.lean
```diff
@@ -503,6 +504,45 @@ def expandDeclId (declId : Syntax) (modifiers : Modifiers) : CommandElabM Expand
...
+    traceState.traces := coreState.traceState.traces ++ commandState.traceState.traces
+    env := commandState.env
+  }
+  if let some err := commandState.messages.msgs.toArray.find? (·.severity matches .error) then
+    throwError err.data
+  pure a
+
+/--
+Evaluate any `set_option in` commands before the given `stx`, and pass the inner `stx` with the
+updated environment to the continuation `k`.
```
david-christiansen, line 535: This fixes indicative vs imperative:
```suggestion
Evaluates any `set_option in` commands before the given `stx`, and pass the inner `stx` with the
updated environment to the continuation `k`.
```
But I'd write it a bit differently, because `stx` isn't a named argument, and `withSetOptionIn` presents itself as a transformer of command elaborators, rather than as a monadic action that consumes syntax:
```suggestion
Given a command elaborator `elab`, returns a new command elaborator that first evaluates any local `set_option ... in ...` clauses and then invokes `elab` on what remains.
```
(while also renaming the argument to `elab`)

src/Lean/Elab/Command.lean
```diff
@@ -503,6 +504,45 @@ def expandDeclId (declId : Syntax) (modifiers : Modifiers) : CommandElabM Expand
...
+-/
+partial def withSetOptionIn (k : CommandElab) : CommandElab := fun stx => do
+  if stx.getKind == ``Lean.Parser.Command.in &&
+     stx[0].getKind == ``Lean.Parser.Command.set_option then
+      let opts ← Elab.elabSetOption stx[0][1] stx[0][2]
+      Command.withScope (fun scope => { scope with opts }) do
+        withSetOptionIn k stx[1]
+  else
+    k stx
```
david-christiansen, line 544:
```suggestion
partial def withSetOptionIn (elab : CommandElab) : CommandElab := fun stx => do
  if stx.getKind == ``Lean.Parser.Command.in &&
     stx[0].getKind == ``Lean.Parser.Command.set_option then
      let opts ← Elab.elabSetOption stx[0][1] stx[0][2]
      Command.withScope (fun scope => { scope with opts }) do
        withSetOptionIn elab stx[1]
  else
    elab stx
```

src/Lean/Elab/Tactic/BuiltinTactic.lean
```diff
@@ -323,6 +323,9 @@ def forEachVar (hs : Array Syntax) (tac : MVarId → FVarId → MetaM MVarId) :
 @[builtin_tactic Lean.Parser.Tactic.substVars] def evalSubstVars : Tactic := fun _ =>
   liftMetaTactic fun mvarId => return [← substVars mvarId]
 
+/-- `subst_eqs` applies `subst` to all equalities in the context as long as it makes progress. -/
```
david-christiansen, line 326: I think this formulation is easier to understand:
```suggestion
/-- Repeatedly substitutes according to the equality proof hypotheses in the context, replacing the left side of the equality with the right, until no more progress can be made. -/
```

But what happens if there are no such hypotheses? Does it fail or is it a no-op?

And from what I can see, the implementation doesn't use `subst`, but rather `casesOn`. But that's a bit beside the point - tactic users shouldn't have to care which term they get, at least not for things in `Prop`.

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+  let flat ← match flat with
+  | `(true) => pure true
+  | `(false) => pure false
+  | _ => throwErrorAt flat "expected 'true' or 'false'"
+  unless isStructure (← getEnv) struct do throwError "not a structure: {struct}"
+  let structC ← mkConstWithLevelParams struct
+  forallTelescope (← inferType structC) fun params _ => do
+  withNewBinderInfos (params.map (·.fvarId!, BinderInfo.implicit)) do
+  withLocalDeclD `x (mkAppN structC params) fun x => do
+  withLocalDeclD `y (mkAppN structC params) fun y => do
```
david-christiansen, line 31: This could just be a "me" thing, but I find this style without indentation to be quite difficult to read - the indentation sends a message that we're in the telescope, or in the local decl's scope, and missing that message throws me off.

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+import Lean.Linter.Util
+
+namespace Lean.Elab.Tactic.Ext
+open Lean Meta Elab Tactic
+
+/--
+Constructs the hypotheses for the extensionality lemma.
+Calls the continuation `k` with the list of parameters to the structure,
+two structure variables `x` and `y`, and a list of pairs `(field, ty)`
+where `ty` is `x.field = y.field` or `HEq x.field y.field`.
```
david-christiansen, line 18:
```suggestion
Constructs the hypotheses for the structure extensionality lemma that
states that two structures are equal if their fields are equal.
 
Calls the continuation `k` with the list of parameters to the structure,
two structure variables `x` and `y`, and a list of pairs `(field, ty)`
where `ty` is `x.field = y.field` or `HEq x.field y.field`.
```

david-christiansen: I'd also add something like:
```
`flat` is expected to be the syntax of a constructor of `Bool`. If it is `true`, any fields inherited from parent structures are treated fields of the given structure type. If it is `false`, then the behind-the-scenes encoding of inherited fields is visible in the extensionality lemma.
```

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+      else if ← isDefEq (← inferType x_f) (← inferType y_f) then
+        hyps := hyps.push (field, ← mkEq x_f y_f)
+      else
+        hyps := hyps.push (field, ← mkHEq x_f y_f)
+    k params x y hyps
+
+/--
+Creates the type of the extensionality lemma for the given structure,
+elaborating to `x.1 = y.1 → x.2 = y.2 → x = y`, for example.
+-/
```
david-christiansen, line 51: See suggestion for `flat` in the docstring above

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+/-- Make an n-ary `And` application. `mkAndN []` returns `True`. -/
+def mkAndN : List Expr → Expr
+  | [] => mkConst ``True
+  | [p] => p
+  | p :: ps => mkAnd p (mkAndN ps)
+
+/--
+Creates the type of the iff-variant of the extensionality lemma for the given structure,
+elaborating to `x = y ↔ x.1 = y.1 ∧ x.2 = y.2`, for example.
+-/
```
david-christiansen, line 70: See comment above about `flat` - docstrings mostly appear in hovers, so a bit of repetition is useful.

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+    protected theorem $extIffName:ident : ext_iff_type% $flat $struct:ident :=
+      fun {..} {..} =>
+        ⟨fun h => by cases h; split_ands <;> rfl,
+         fun _ => by (repeat cases ‹_ ∧ _›); subst_eqs; rfl⟩)
+
+/-- Apply a single extensionality lemma to `goal`. -/
+def applyExtLemma (goal : MVarId) : MetaM (List MVarId) := goal.withContext do
+  let tgt ← goal.getType'
+  unless tgt.isAppOfArity ``Eq 3 do
+    throwError "applyExtLemma only applies to equations, not{indentExpr tgt}"
```
david-christiansen, line 96: Would it be more user-friendly to also check here that the thing being applied is `Eq` or `HEq`?

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+  throwError "no applicable extensionality lemma found for{indentExpr ty}"
+
+/-- Apply a single extensionality lemma to the current goal. -/
+elab "apply_ext_lemma" : tactic => liftMetaTactic applyExtLemma
+
+/--
+Postprocessor for `withExt` which runs `rintro` with the given patterns when the target is a
+pi type.
+-/
+def tryIntros [Monad m] [MonadLiftT TermElabM m] (g : MVarId) (pats : List (TSyntax `rcasesPat))
```
david-christiansen, line 123: The parameter name `g` here and elsewhere in the file isn't so explanatory. Is it "goal"?

I think single-letter parameter names are great when the argument is something truly arbitrary, like the elements of a list when defining `map`, but here it has a specific function. An informative name makes the API easier to figure out.

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+-/
+def withExt1 [Monad m] [MonadLiftT TermElabM m] (g : MVarId) (pats : List (TSyntax `rcasesPat))
+    (k : MVarId → List (TSyntax `rcasesPat) → m Nat) : m Nat := do
+  let mut n := 0
+  for g in ← (applyExtLemma g : TermElabM _) do
+    n := n.max (← tryIntros g pats k)
+  pure n
+
+/--
+Applies a extensionality lemmas recursively, using `pats` to introduce variables in the result.
```
david-christiansen, line 147: Does this apply a single lemma recursively, or does it apply arbitrary known extensionality lemmas recursively? I think the latter?

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+    if failIfUnchanged then
+      withExt1 g pats fun g pats => withExtN g pats k depth (failIfUnchanged := false)
+    else try
+      withExt1 g pats fun g pats => withExtN g pats k depth (failIfUnchanged := false)
+    catch _ => k g pats
+
+/--
+Apply extensionality lemmas as much as possible, using `pats` to introduce the variables
+in extensionality lemmas like `funext`. Returns a list of subgoals.
+-/
```
david-christiansen, line 165: How is this different from `withExtN`?

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+def extCore (g : MVarId) (pats : List (TSyntax `rcasesPat))
+    (depth := 1000000) (failIfUnchanged := true) :
+    TermElabM (Nat × Array (MVarId × List (TSyntax `rcasesPat))) := do
+  StateT.run (m := TermElabM) (s := #[])
+    (withExtN g pats (fun g qs => modify (·.push (g, qs)) *> pure 0) depth failIfUnchanged)
+
+/--
+* `ext pat*` applies extensionality lemmas as much as possible,
+  using `pat*` to introduce the variables in extensionality lemmas using `rintro`.
```
david-christiansen, line 174: What if the `pat`s don't match? Does that result in anonymous hypotheses or in the tactic halting?

It's also nice to expand abbreviations:
```suggestion
  using the patterns `pat*` to introduce the variables in extensionality lemmas using `rintro`.
```

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+  For example, this names the variables introduced by lemmas such as `funext`.
```
david-christiansen, line 175:
```suggestion
  For example, the patterns are used to name the variables introduced by lemmas such as `funext`.
```

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+* `ext` applies extensionality lemmas as much as possible
+  but introduces anonymous variables whenever needed.
```
david-christiansen, line 177:
```suggestion
* Without patterns,`ext` applies extensionality lemmas as much
  as possible but introduces anonymous hypotheses whenever needed.
```

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+Apply extensionality lemmas as much as possible, using `pats` to introduce the variables
+in extensionality lemmas like `funext`. Returns a list of subgoals.
+-/
+def extCore (g : MVarId) (pats : List (TSyntax `rcasesPat))
+    (depth := 1000000) (failIfUnchanged := true) :
+    TermElabM (Nat × Array (MVarId × List (TSyntax `rcasesPat))) := do
+  StateT.run (m := TermElabM) (s := #[])
+    (withExtN g pats (fun g qs => modify (·.push (g, qs)) *> pure 0) depth failIfUnchanged)
+
+/--
```
david-christiansen, line 172:
```suggestion
/--
Applies extensionality lemmas that are registered with the `@[ext]` attribute.
```

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+  using `pat*` to introduce the variables in extensionality lemmas using `rintro`.
+  For example, this names the variables introduced by lemmas such as `funext`.
+* `ext` applies extensionality lemmas as much as possible
+  but introduces anonymous variables whenever needed.
+* `ext pat* : n` applies ext lemmas only up to depth `n`.
+
+The `ext1 pat*` tactic is like `ext pat*` except that it only applies a single extensionality lemma.
+
+The `ext?` tactic (note: unimplemented) has the same syntax as the `ext` tactic,
+and it gives a suggestion of an equivalent tactic to use in place of `ext`.
```
david-christiansen, line 183: Perhaps this should be deleted until the tactic is implemented? I don't think having unimplemented API in the docs gives a good impression.

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+
+/--
+`ext1 pat*` is like `ext pat*` except that it only applies a single extensionality lemma rather
+than recursively applying as many extensionality lemmas as possible.
+
+The `pat*` patterns are processed using the `rintro` tactic.
+If no patterns are supplied, then variables are introduced anonymously using the `intros` tactic.
+
+The `ext1?` tactic (note: unimplemented) has the same syntax as the `ext1?` tactic,
+and it gives a suggestion of an equivalent tactic to use in place of `ext1`.
```
david-christiansen, line 205:
```suggestion
```

src/Lean/Elab/Tactic/Ext/Attr.lean
```diff
@@ -0,0 +1,111 @@
...
+Authors: Gabriel Ebner, Mario Carneiro
+-/
+import Lean.Elab.Tactic.RCases
+import Lean.Meta.DiscrTree
+import Lean.Elab.Command
+
+namespace Lean.Elab.Tactic.Ext
+open Lean Meta
+
+/-- `declare_ext_theorems_for A` declares the extensionality theorems for the structure `A`. -/
```
david-christiansen, line 13:
```suggestion
/-- `declare_ext_theorems_for A` declares the extensionality lemmas for the structure `A`. -/
```
For consistency.

I'd also say a bit more:
```suggestion
/-- `declare_ext_theorems_for A` declares the extensionality lemmas for the structure `A`.

These lemmas state that two expressions with the structure type are equal if their fields are equal.
-/
```

src/Lean/Elab/Tactic/Ext/Attr.lean
```diff
@@ -0,0 +1,111 @@
...
+syntax "declare_ext_theorems_for " ("(" &"flat" " := " term ") ")? ident (ppSpace prio)? : command
+
+/-- Information about an extensionality theorem, stored in the environment extension. -/
```
david-christiansen, line 16: We should be consistent about calling them "theorems" or "lemmas".

src/Lean/Elab/Tactic/Ext/Attr.lean
```diff
@@ -0,0 +1,111 @@
...
+structure ExtTheorem where
+  /-- Declaration name of the extensionality theorem. -/
+  declName : Name
+  /-- Priority of the extensionality theorem. -/
+  priority : Nat
+  /-- Key in the discrimination tree. -/
```
david-christiansen, line 22: Perhaps a bit should be said about what this key actually contains? It's the type for which the lemma applies, right?

src/Lean/Elab/Tactic/Ext/Attr.lean
```diff
@@ -0,0 +1,111 @@
...
+/-- The environment extension to track `@[ext]` lemmas. -/
+initialize extExtension :
+    SimpleScopedEnvExtension ExtTheorem ExtTheorems ←
+  registerSimpleScopedEnvExtension {
+    addEntry := fun { tree, erased } thm =>
+      { tree := tree.insertCore thm.keys thm, erased := erased.erase thm.declName }
+    initial := {}
+  }
+
+/-- Get the list of `@[ext]` lemmas corresponding to the key `ty`,
```
david-christiansen, line 46:
```suggestion
/-- Gets the list of `@[ext]` lemmas corresponding to the key `ty`,
```

src/Lean/Elab/Tactic/Ext/Attr.lean
```diff
@@ -0,0 +1,111 @@
...
+  priority : Nat
+  /-- Key in the discrimination tree. -/
+  keys : Array DiscrTree.Key
+  deriving Inhabited, Repr, BEq, Hashable
+
+/-- The state of the `ext` extension environment -/
+structure ExtTheorems where
+  /-- The tree of `ext` extensions. -/
+  tree   : DiscrTree ExtTheorem := {}
+  /-- Erased `ext`s. -/
```
david-christiansen, line 30: What does it mean to be "erased"? That's not immediately apparent when looking through this API.

src/Lean/Elab/Tactic/Ext/Attr.lean
```diff
@@ -0,0 +1,111 @@
...
+  -- Most ext lemmas have default priority.
+  return erasedArr.insertionSort (·.priority < ·.priority) |>.reverse
+
+/-- Erases a name marked `ext` by adding it to the state's `erased` field and
+  removing it from the state's list of `Entry`s. -/
+def ExtTheorems.eraseCore (d : ExtTheorems) (declName : Name) : ExtTheorems :=
+ { d with erased := d.erased.insert declName }
+
+/--
+  Erase a name marked as a `ext` attribute.
```
david-christiansen, line 62:
```suggestion
  Erases a name marked as a `ext` attribute.
```

It'd also be nice to include what erasure means in this context.

src/Lean/Elab/Tactic/Ext/Attr.lean
```diff
@@ -0,0 +1,111 @@
...
+  return d.eraseCore declName
+
+/-- Registers an extensionality lemma.
+
+* When `@[ext]` is applied to a structure, it generates `.ext` and `.ext_iff` theorems and registers
+  them for the `ext` tactic.
+
+* When `@[ext]` is applied to a theorem, the theorem is registered for the `ext` tactic.
+
+* You can use `@[ext 9000]` to specify a priority for the attribute.
```
david-christiansen, line 79: What's the default priority?

david-christiansen: Also, the "you" here doesn't match our other docstrings in style. What about:
```suggestion
* An optional natural number argument, e.g. `@[ext 9000]`, specifies a priority for the lemma. Higher-priority lemmas are chosen first, and the default is `0`.
```

src/Lean/Elab/Tactic/Ext/Attr.lean
```diff
@@ -0,0 +1,111 @@
...
+
+* You can use the flag `@[ext (flat := false)]` to prevent flattening all fields of parent
```
david-christiansen, line 81:
```suggestion
* The flag `@[ext (flat := false)]` causes generated structure extensionality lemmas to show inherited fields based on their representation, rather than flattening the parents' fields into the lemma's equality hypotheses.
```

src/Lean/Elab/Tactic/Repeat.lean
```diff
@@ -0,0 +1,107 @@
...
+-/
+
+import Lean.Elab.Tactic.Basic
+import Lean.Meta.Tactic.Cases
+
+open Lean Elab Tactic
+
+/--
+`iterate n tac` runs `tac` exactly `n` times.
+`iterate tac` runs `tac` repeatedly until failure.
```
david-christiansen, line 14: Perhaps this should be removed, as `repeat` already does this? Or alternatively, can this be folded into `repeat` so we have one tactic rather than two?

src/Lean/Elab/Tactic/Repeat.lean
```diff
@@ -0,0 +1,107 @@
...
+
+To run multiple tactics, one can do `iterate (tac₁; tac₂; ⋯)` or
+```lean
+iterate
+  tac₁
+  tac₂
+  ⋯
+```
```
david-christiansen, line 22:
````suggestion
`iterate`'s argument is a tactic sequence, so multiple tactics can be run using `iterate n (tac₁; tac₂; ⋯)` or
```lean
iterate n
  tac₁
  tac₂
  ⋯
```
````

File: src/Lean/Elab/Tactic/Repeat.lean [hunk 31]
```diff
@@ -0,0 +1,107 @@
...
+/--
+Implementation of `repeat'` and `repeat1'`.
+
+`repeat'Core f` runs `f` on all of the goals to produce a new list of goals,
+then runs `f` again on all of those goals, and repeats until `f` fails on all remaining goals,
+or until `maxIters` total calls to `f` have occurred.
+
+Returns a boolean indicating whether `f` succeeded at least once, and
+all the remaining goals (i.e. those on which `f` failed).
+-/
```
david-christiansen, line 42: This is a great docstring!

src/Lean/Elab/Tactic/Repeat.lean
```diff
@@ -0,0 +1,107 @@
...
+def repeat'Core [Monad m] [MonadExcept ε m] [MonadBacktrack s m] [MonadMCtx m]
+    (f : MVarId → m (List MVarId)) (gs : List MVarId) (maxIters := 100000) :
```
david-christiansen, line 44:
```suggestion
    (f : MVarId → m (List MVarId)) (goals : List MVarId) (maxIters := 100000) :
```

src/Lean/Elab/Tactic/Repeat.lean
```diff
@@ -0,0 +1,107 @@
...
+    m (Bool × List MVarId) := do
+  let (progress, acc) ← go maxIters false gs [] #[]
```
david-christiansen, line 46:
```suggestion
  let (progress, acc) ← go maxIters false goals [] #[]
```

src/Lean/Elab/Tactic/Repeat.lean
```diff
@@ -0,0 +1,107 @@
...
+
+/--
+`repeat' f` runs `f` on all of the goals to produce a new list of goals,
+then runs `f` again on all of those goals, and repeats until `f` fails on all remaining goals,
+or until `maxIters` total calls to `f` have occurred.
+Always succeeds (returning the original goals if `f` fails on all of them).
+-/
+def repeat' [Monad m] [MonadExcept ε m] [MonadBacktrack s m] [MonadMCtx m]
+    (f : MVarId → m (List MVarId)) (gs : List MVarId) (maxIters := 100000) : m (List MVarId) :=
+  repeat'Core f gs maxIters <&> (·.2)
```
david-christiansen, line 75:
```suggestion
def repeat' [Monad m] [MonadExcept ε m] [MonadBacktrack s m] [MonadMCtx m]
    (f : MVarId → m (List MVarId)) (goals : List MVarId) (maxIters := 100000) : m (List MVarId) :=
  repeat'Core f goals maxIters <&> (·.2)
```

src/Lean/Elab/Tactic/Repeat.lean
```diff
@@ -0,0 +1,107 @@
...
+/--
+`repeat1' f` runs `f` on all of the goals to produce a new list of goals,
+then runs `f` again on all of those goals, and repeats until `f` fails on all remaining goals,
+or until `maxIters` total calls to `f` have occurred.
+Fails if `f` does not succeed at least once.
+-/
+def repeat1' [Monad m] [MonadError m] [MonadExcept ε m] [MonadBacktrack s m] [MonadMCtx m]
+    (f : MVarId → m (List MVarId)) (gs : List MVarId) (maxIters := 100000) : m (List MVarId) := do
+  let (.true, gs) ← repeat'Core f gs maxIters | throwError "repeat1' made no progress"
+  pure gs
```
david-christiansen, line 86:
```suggestion
def repeat1' [Monad m] [MonadError m] [MonadExcept ε m] [MonadBacktrack s m] [MonadMCtx m]
    (f : MVarId → m (List MVarId)) (goals : List MVarId) (maxIters := 100000) : m (List MVarId) := do
  let (.true, gs) ← repeat'Core f goals maxIters | throwError "repeat1' made no progress"
  pure goals
```

src/Lean/Elab/Tactic/Repeat.lean
```diff
@@ -0,0 +1,107 @@
...
+  setGoals (← repeat' (evalTacticAtRaw tac) (← getGoals))
+
+/--
+`repeat1' tac` applies `tac` to main goal at least once. If the application succeeds,
+the tactic is applied recursively to the generated subgoals until it eventually fails.
+-/
+elab "repeat1' " tac:tacticSeq : tactic => do
+  setGoals (← repeat1' (evalTacticAtRaw tac) (← getGoals))
+
+/-- `subst_eqs` applies `subst` to all equalities in the context as long as it makes progress. -/
```
david-christiansen, line 102: This is also defined in BuiltInTactic.lean - is this a mistake?

src/Lean/Elab/Tactic/Repeat.lean
```diff
@@ -0,0 +1,107 @@
...
+elab "subst_eqs" : tactic => Elab.Tactic.liftMetaTactic1 (·.substEqs)
+
+/-- `split_ands` applies `And.intro` until it does not make progress. -/
```
david-christiansen, line 105: Is this the right tactic name?

In Coq, `split` can be used to do one level of this on a goal that is a conjunction. But Lean uses `split` to talk about case splitting - it has more the feel of a proof by cases about a program that itself case-splits, rather than a conjunction. This seems to have little to do with destructuring `And` goals.

david-christiansen:
Lots of docstring and parameter name nitpicks :-)

I worry about the API overlap between `iterate` and `repeat`. Can they be made into a single tactic (e.g. by deleting `repeat` and putting `iterate`'s implementation where it was)?

src/Lean/Elab/Tactic/Repeat.lean
semorrison: Yes, thanks.

src/Lean/Elab/Tactic/Ext/Attr.lean
```diff
@@ -0,0 +1,111 @@
...
+-/
+import Lean.Elab.Tactic.RCases
+import Lean.Meta.DiscrTree
+import Lean.Elab.Command
+
+namespace Lean.Elab.Tactic.Ext
+open Lean Meta
+
+/-- `declare_ext_theorems_for A` declares the extensionality theorems for the structure `A`. -/
+syntax "declare_ext_theorems_for " ("(" &"flat" " := " term ") ")? ident (ppSpace prio)? : command
```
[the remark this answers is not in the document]
semorrison: Hmm, `lemma` isn't a keyword in Lean (although Mathlib adds it). I'd prefer not to make this change.

src/Lean/Elab/Tactic/Ext/Attr.lean
semorrison: Added "via `attribute [-ext]`"

src/Lean/Elab/Tactic/Ext/Attr.lean
semorrison: Done.

src/Lean/Elab/Tactic/Ext.lean
semorrison: Acknowledged, but not acted on. :-)

src/Lean/Elab/Tactic/Ext.lean
semorrison: These are not intended for use outside this module, so I'm not too concerned about the hover explaining things in isolation.

src/Lean/Elab/Tactic/Ext.lean
semorrison: As above.

src/Lean/Elab/Tactic/Ext.lean
semorrison: I did a quick count. We have 383 `mvarId : MVarId`, 15 `g : MVarId`, and 5 `goal : MVarId` (and 4 of those were upstreamed yesterday).

`mvarId` is even less informative than `g`! I think I might punt on this one.

src/Lean/Elab/Tactic/Ext.lean
semorrison: The latter.

src/Lean/Elab/Tactic/Ext.lean
```diff
@@ -0,0 +1,234 @@
...
+-/
+macro "ext1" xs:(colGt ppSpace rintroPat)* : tactic =>
+  if xs.isEmpty then `(tactic| apply_ext_lemma <;> intros)
+  else `(tactic| apply_ext_lemma <;> rintro $xs*)
+
+-- TODO
+/-- `ext1? pat*` is like `ext1 pat*` but gives a suggestion on what pattern to use -/
+syntax "ext1?" (colGt ppSpace rintroPat)* : tactic
+/-- `ext? pat*` is like `ext pat*` but gives a suggestion on what pattern to use -/
+syntax "ext?" (colGt ppSpace rintroPat)* (" : " num)? : tactic
```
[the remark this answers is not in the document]
semorrison: Removed.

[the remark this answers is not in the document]
semorrison: Tracking issue at #3325

src/Lean/Elab/Tactic/Repeat.lean
semorrison: I've made an internal TODO.

src/Lean/Elab/Tactic/Repeat.lean
semorrison: I've renamed it `and_intros`.

src/Lean/Elab/Tactic/Ext.lean
semorrison: I've added a note.

src/Lean/Elab/Tactic/Ext.lean
semorrison: I've added two sentences.

src/Lean/Elab/Tactic/Ext.lean
semorrison: removed `a`.

src/Lean/Elab/Tactic/Ext.lean
semorrison: I'm not sure what you mean here.

david-christiansen: I'm also not sure what I mean - I think I read it incorrectly the first time around, and missed the `Eq` on line 95. Sorry for the false alarm :)

PR #3307, opened 2024-02-12
chore: upstream left/right tactics

kim-em:


src/Lean/Meta/Tactic/Apply.lean
```diff
@@ -230,4 +230,22 @@ def _root_.Lean.MVarId.exfalso (mvarId : MVarId) : MetaM MVarId :=
...
+    MetaM (List MVarId) := do
+  goal.withContext do
+    goal.checkNotAssigned name
+    matchConstInduct (← goal.getType').getAppFn
+      (fun _ => throwTacticEx name goal "target is not an inductive datatype")
+      fun ival us => do
+        if let some e := expected? then unless ival.ctors.length == e do
+          throwTacticEx name goal
+            s!"{name} tactic works for inductive types with exactly {e} constructors"
+        goal.apply <| mkConst ival.ctors[idx]! us
```
nomeata, line 249: If this is called without `expected?`, it will panic. The only uses here are `left` and `right`, which do set `expected`.  Is it worth making this function safer by requiring `expected?` and maybe even setting `idx : Fin expected`? Or else at least check `idx < ival.ctrs.size` here?

src/Init/Tactics.lean
```diff
@@ -207,6 +207,28 @@ the first matching constructor, or else fails.
...
 syntax (name := constructor) "constructor" : tactic
 
+/--
+Apply the first constructor,
+in the case that the goal is an inductive type with exactly two constructors.
+```
+example : True ∨ False := by
+  left
+  trivial
+```
```
david-christiansen, line 217: We decided a while back to standardize on the indicative mood in docstrings. I'm not great at remembering it, but I'm working on it!

```suggestion
Applies the first constructor when the goal is an inductive type with exactly two constructors, and fails otherwise.
```
example : True ∨ False := by
  left
  trivial
```
```

src/Lean/Meta/Tactic/Apply.lean
david-christiansen: Having `expected?` be optional allows this to be used for tactics that work for more general types. But I think we should not generalize to cases that we may someday want, as it makes the code harder to use if we don't actually want to support them. In fact, I think this could even be simplified to a version that supports just 2-constructor datatypes, unless we have further plans to upstream more general tactics in this family.

semorrison: I think just an informative error unless `idx < ival.ctrs.size`. Want to push the change? :-)

nomeata: Can do after breakfast

nomeata:
```suggestion
        unless (idx < ival.ctors.length) do
          throwError m!"nthConstructor: idx {idx} out of bounds, only {ival.ctors.length} constructors"
        goal.apply <| mkConst ival.ctors[idx]! us
```

src/Lean/Elab/Tactic/BuiltinTactic.lean
```diff
@@ -468,4 +469,10 @@ where
   | none    => throwIllFormedSyntax
   | some ms => IO.sleep ms.toUInt32
 
+@[builtin_tactic left] def evalLeft : Tactic := fun stx => do
+  liftMetaTactic (fun g => g.nthConstructor `left 0 (some 2))
+
+@[builtin_tactic right] def evalRight : Tactic := fun stx => do
+  liftMetaTactic (fun g => g.nthConstructor `right 1 (some 2))
```
nomeata, line 476:
```suggestion
@[builtin_tactic left] def evalLeft : Tactic := fun _stx => do
  liftMetaTactic (fun g => g.nthConstructor `left 0 (some 2))

@[builtin_tactic right] def evalRight : Tactic := fun _stx => do
  liftMetaTactic (fun g => g.nthConstructor `right 1 (some 2))
```

src/Init/Tactics.lean
david-christiansen: I actually like the imperative one better, but I like consistency most of all!

I'll try to be aware of incoming changes, and we can chip away at it as we go.

src/Init/Tactics.lean
```diff
@@ -207,6 +207,28 @@ the first matching constructor, or else fails.
 -/
 syntax (name := constructor) "constructor" : tactic
 
+/--
+Applies the second constructor when
+the goal is an inductive type with exactly two constructors, or fails otherwise.
```
david-christiansen, line 212: This is also some odd word-wrapping.

[Rewrap](https://open-vsx.org/extension/stkb/rewrap) is a good way to make it easy to fix these.

src/Lean/Meta/Tactic/Apply.lean
```diff
@@ -230,4 +230,25 @@ def _root_.Lean.MVarId.exfalso (mvarId : MVarId) : MetaM MVarId :=
...
+    MetaM (List MVarId) := do
+  goal.withContext do
+    goal.checkNotAssigned name
+    matchConstInduct (← goal.getType').getAppFn
+      (fun _ => throwTacticEx name goal "target is not an inductive datatype")
+      fun ival us => do
+        if let some e := expected? then unless ival.ctors.length == e do
+          throwTacticEx name goal
+            s!"{name} tactic works for inductive types with exactly {e} constructors"
+        if h: idx < ival.ctors.length
```
david-christiansen: Style nitpick:
```suggestion
        if h : idx < ival.ctors.length
```

nomeata:
Ok, shouldn’t do coding using github suggestions. Mea culpa. I was too optimistic.

I actually used `throwErrow` intentionally, because I thought this error conditions is more an internal error than a tactic failure, but maybe I was overthinking this. Hard to tell when coding for an unknown use case :-)

src/Init/Tactics.lean
semorrison: Oh. It was my attempt at semantic line-wrapping.

PR #3308, opened 2024-02-12
chore: upstream change tactic

kim-em:
We previously had the syntax for `change` and `change at`, but no implementation.

This moves Kyle's implementation from Std.

This also changes the `changeLocalDecl` function to push nodes to the infotree about FVar aliases.


src/Lean/Elab/Tactic/Change.lean
```diff
@@ -0,0 +1,47 @@
...
+```lean
+change _ + 1 = _ at h h'
+```
+changes their types to be `h : n + 1 + 1 = 2` and `h' : n + 2 + 1 = 4`.
+
+Change is like `refine` in that every placeholder needs to be solved for by unification,
+but you can use named placeholders and `?_` where you want `change` to create new goals.
+
+The tactic `show e` is interchangeable with `change e`, where the pattern `e` is applied to
+the main goal. -/
```
david-christiansen, line 33: This seems to needlessly increase our API surface. Why can't we have just one of `change` and/or `show`?

I don't think it's a good user experience to ask why a given operator seems to just be a special case of another, and then find out "oh, yeah, this is some historical thing". We'll eventually be there for reasons of compatibility, but let's keep things consistent as long as we have users who will put up with it!

david-christiansen:
```suggestion
/-- `change` can be used to replace the main goal or its hypotheses with
different, yet definitionally equal, goal or hypotheses.

For example, if `n : Nat` and the current goal is `⊢ n + 2 = 2`, then
```lean
change _ + 1 = _
```
changes the goal to `⊢ n + 1 + 1 = 2`.

The tactic also applies to hypotheses. If `h : n + 2 = 2` and `h' : n + 3 = 4`
are hypotheses, then
```lean
change _ + 1 = _ at h h'
```
changes their types to be `h : n + 1 + 1 = 2` and `h' : n + 2 + 1 = 4`.

Change is like `refine` in that every placeholder needs to be solved for by unification,
but using named placeholders or `?_` results in `change` to creating new goals.

The tactic `show e` is interchangeable with `change e`, where the pattern `e` is applied to
the main goal. -/
```

Where are named placeholders described? A quick grep of the Lean source doesn't turn much up. Can we add an example here?

Comment (david-christiansen) [reply src/Lean/Elab/Tactic/Change.lean:33]: (that last mega-suggestion is mostly just written style harmonization/consistency with other docstrings)

Comment (digama0) [reply src/Lean/Elab/Tactic/Change.lean:33]: `show T` and `change T` are the same, but they each generalize in different directions.

* `show T` is also a term constructor `show T from e` / `show T by tac`, and we like to have tactics which mimic the syntax of their term variants. (I'm not a big fan of the use of `from` in this syntax and would prefer `:=`, but that's another matter.)
* `change T` also supports `change T at h` and (not upstreamed) `change e1 with e2 (at h)?`, which allow defeq rewriting in the goal or a hypothesis. While one could use `show` here instead, the resulting tactic `show e1 with e2` doesn't make a whole lot of sense grammatically (given that it means to replace occurrences of `e1` in the goal with `e2`).

Comment (digama0) [reply src/Lean/Elab/Tactic/Change.lean:33]: An alternative tactic design for `change e1 with e2 at h` is `drw [e1 = e2] at h`. This is using the same syntax as `rw` with a `d` prefix for "definitional" similar to `dsimp` and others, but it takes equalities directly instead of proofs (the equivalent of `show e1 = e2 from rfl` in a `rw` proof). I don't see any obvious analogue for the `change T at h` though, except `drw [_ = T] at h` which has some unnecessary syntax.

File: src/Lean/Meta/Tactic/Replace.lean [hunk 2]
```diff
@@ -139,29 +140,64 @@ def _root_.Lean.MVarId.change (mvarId : MVarId) (targetNew : Expr) (checkDefEq :
...
+  and a mutated `MVarId`.
+
+The `a : Array (Option FVarId)` array returned by `k` is interpreted in the following way.
+The function will intro `a.size` variables, and then for each non-`none` entry we
+create an FVar alias between it and the corresponding `intro`ed variable.
+For example, having `k` return `fvars.map .some` causes all reverted variables to be
+`intro`ed and linked.
+
+Returns the value returned by `k` along with the resulting goal.
+ -/
```
david-christiansen, line 162: I think this docstring mixes too many separate concerns at one time. How about this?
```suggestion
/-- Runs the continuation `k` after temporarily reverting some variables from the local context of a metavariable (identified by `mvarId`), then reintroduces local variables as specified by `k`.

The argument `fvarIds` is an array of `fvarIds` to revert in the order specified. An error is thrown if they cannot be reverted in order.

Once the local variables have been reverted, `k` is passed `mvarId` along with an array of local variables that were reverted. This array always has `fvarIds` as a prefix, but it may contain additional variables that were reverted due to dependencies. `k` returns a value, a goal, an array of _link variables_.

Once `k` has completed, one variable is introduced for each link variable returned by `k`. If the returned variable is `none`, the variable is just introduced. If it is `some fv`, the variable is introduced and then linked as an alias of `fv` in the info tree. For example, having `k` return `fvars.map .some` as the link variables causes all reverted variables to be introduced and linked.

Returns the value returned by `k` along with the resulting goal.
 -/
```
I think this formulation separates the context manipulation from the UI concern of linking variables a bit more.

src/Lean/Meta/Tactic/Replace.lean
```diff
@@ -139,29 +140,64 @@ def _root_.Lean.MVarId.change (mvarId : MVarId) (targetNew : Expr) (checkDefEq :
...
+  let (r, xs', mvarId) ← k mvarId xs
+  let (ys, mvarId) ← mvarId.introNP xs'.size
+  mvarId.withContext do
+    for x? in xs', y in ys do
+      if let some x := x? then
+        Elab.pushInfoLeaf (.ofFVarAliasInfo { id := y, baseId := x, userName := ← y.getUserName })
+  return (r, mvarId)
+
 /--
 Replace the type of the free variable `fvarId` with `typeNew`.
```
david-christiansen, line 176:
```suggestion
Replaces the type of the free variable `fvarId` with `typeNew`.
```

src/Lean/Meta/Tactic/Replace.lean
```diff
@@ -139,29 +140,64 @@ def _root_.Lean.MVarId.change (mvarId : MVarId) (targetNew : Expr) (checkDefEq :
...
-If `checkDefEq = false`, this method assumes that `typeNew` is definitionally equal to `fvarId` type.
-If `checkDefEq = true`, throw an error if `typeNew` is not definitionally equal to `fvarId` type.
+
+If `checkDefEq = true` then throws an error if `typeNew` is not definitionally
```
david-christiansen, line 178:
```suggestion
If `checkDefEq` is `true` then an error is thrown if `typeNew` is not definitionally
```
This is not a propositional equality!

src/Lean/Meta/Tactic/Replace.lean
```diff
@@ -139,29 +140,64 @@ def _root_.Lean.MVarId.change (mvarId : MVarId) (targetNew : Expr) (checkDefEq :
...
+equal to the type of `fvarId`. Otherwise this function assumes `typeNew` and the type
+of `fvarId` are definitionally equal.
+
+This function is the same as `Lean.MVarId.changeLocalDecl` but makes sure to push substitution
+information into the infotree.
```
david-christiansen, line 183:
```suggestion
information into the info tree.
```

src/Lean/Elab/Tactic/Change.lean
david-christiansen: OK, this makes sense - one is not more general than the other.

I think we should update the docstrings of both to make this more clear, then, using text similar to @digama0's comment.

I can have a go, but I want to check my understanding first: there's no context in a tactic proof where `show` would be applicable but `change` would not be, right? In other words, we can consider `show` to be something you can use if you want to, but `change` is the appropriate default for people who want to accomplish this task?

digama0: Yes, you can always use `change` in place of `show`. But as a matter of practice `show` is considerably more popular than `change` for the case where it is applicable. (I think; it's a bit tricky to directly get evidence using grep, because `show` has lots of other uses.)

PR #3309, opened 2024-02-12
doc: fix typos around inductiveCheckResultingUniverse

eric-wieser:
The unpaired backtick was causing weird formatting in vscode doc hovers.

Also closes an unpaired `(` in an error message.


nomeata:
Looks like you need to update some test output

eric-wieser:
Good catch, thanks! Note you'll need to re-enable automerge.

PR #3277, opened 2024-02-08
chore: upstream Std.Tactic.ByCases

kim-em:

PR #3312, opened 2024-02-12
chore: upstream Std.Logic

joehendrix:
This will collect definitions from Std.Logic

### DIFF
--- a/src/Init/Core.lean
+++ b/src/Init/Core.lean
@@ -1221,7 +1221,6 @@ gen_injective_theorems% Lean.Syntax
   ⟨eq_of_beq, by intro h; subst h; exact LawfulBEq.rfl⟩
 
 /-! # Prop lemmas -/
--- These may depend on propext
 
 /-- *Ex falso* for negation: from `¬a` and `a` anything follows. This is the same as `absurd` with
 the arguments flipped, but it is in the `Not` namespace so that projection notation can be used. -/
--- a/src/Init/PropLemmas.lean
+++ b/src/Init/PropLemmas.lean
@@ -231,7 +231,6 @@ theorem not_forall_of_exists_not {p : α → Prop} : (∃ x, ¬p x) → ¬∀ x,
 
 @[simp] theorem exists_eq_left' : (∃ a, a' = a ∧ p a) ↔ p a' := by simp [@eq_comm _ a']
 
--- this theorem is needed to simplify the output of `List.mem_cons_iff`
 @[simp] theorem forall_eq_or_imp : (∀ a, a = a' ∨ q a → p a) ↔ p a' ∧ ∀ a, q a → p a := by
   simp only [or_imp, forall_and, forall_eq]
 
@@ -324,10 +323,10 @@ theorem Decidable.or_iff_not_imp_left [Decidable a] : a ∨ b ↔ (¬a → b) :=
   ⟨Or.resolve_left, fun h => dite _ .inl (.inr ∘ h)⟩
 
 theorem Decidable.or_iff_not_imp_right [Decidable b] : a ∨ b ↔ (¬b → a) :=
-or_comm.trans or_iff_not_imp_left
+  or_comm.trans or_iff_not_imp_left
 
 theorem Decidable.not_imp_not [Decidable a] : (¬a → ¬b) ↔ (b → a) :=
-⟨fun h hb => byContradiction (h · hb), mt⟩
+  ⟨fun h hb => byContradiction (h · hb), mt⟩
 
 theorem Decidable.not_or_of_imp [Decidable a] (h : a → b) : ¬a ∨ b :=
   if ha : a then .inr (h ha) else .inl ha
@@ -410,9 +409,9 @@ theorem Decidable.or_congr_right' [Decidable a] (h : ¬a → (b ↔ c)) : a ∨ 
   rw [or_iff_not_imp_left, or_iff_not_imp_left]; exact imp_congr_right h
 
 /-- Transfer decidability of `a` to decidability of `b`, if the propositions are equivalent.
-**Important**: this function should be used instead of `rw` on `decidable b`, because the
+**Important**: this function should be used instead of `rw` on `Decidable b`, because the
 kernel will get stuck reducing the usage of `propext` otherwise,
-and `dec_trivial` will not work. -/
+and `decide` will not work. -/
 @[inline] def decidable_of_iff (a : Prop) (h : a ↔ b) [Decidable a] : Decidable b :=
   decidable_of_decidable_of_iff h
 
--- a/src/Init/SimpLemmas.lean
+++ b/src/Init/SimpLemmas.lean
@@ -111,7 +111,6 @@ end SimprocHelperLemmas
 @[simp] theorem true_or (p : Prop) : (True ∨ p) = True := eq_true (.inl trivial)
 @[simp] theorem or_false (p : Prop) : (p ∨ False) = p := propext ⟨fun (.inl h) => h, .inl⟩
 @[simp] theorem false_or (p : Prop) : (False ∨ p) = p := propext ⟨fun (.inr h) => h, .inr⟩
-
 @[simp] theorem iff_self (p : Prop) : (p ↔ p) = True := eq_true .rfl
 @[simp] theorem iff_true (p : Prop) : (p ↔ True) = p := propext ⟨(·.2 trivial), fun h => ⟨fun _ => trivial, fun _ => h⟩⟩
 @[simp] theorem true_iff (p : Prop) : (True ↔ p) = p := propext ⟨(·.1 trivial), fun h => ⟨fun _ => h, fun _ => trivial⟩⟩
@@ -124,7 +123,6 @@ end SimprocHelperLemmas
 @[simp] theorem not_true_eq_false : (¬ True) = False := by decide
 
 @[simp] theorem not_iff_self : ¬(¬a ↔ a) | H => iff_not_self H.symm
-
 
 /-! ## and -/
 


src/Init/Core.lean
```diff
@@ -34,6 +36,25 @@ and `flip (·<·)` is the greater-than relation.
 
 attribute [simp] namedPattern
 
+/--
+`Empty.elim : Empty → C` says that a value of any type can be constructed from
+`Empty`.
```
david-christiansen, line 41: I think this is a useful addition for new users:

```suggestion
`Empty`. This can be thought of as a compiler-checked assertion that a code path is unreachable.
```

src/Init/Core.lean
```diff
@@ -34,6 +36,25 @@ and `flip (·<·)` is the greater-than relation.
...
+This is a non-dependent variant of `Empty.rec`.
+-/
+@[macro_inline] def Empty.elim {C : Sort u} : Empty → C := Empty.rec
+
+/-- Decidable equality for Empty -/
+instance : DecidableEq Empty := fun a => a.rec
+
+/--
+`PEmpty.elim : Empty → C` says that a value of any type can be constructed from
+`PEmpty`.
```
david-christiansen, line 52:
```suggestion
`PEmpty`. This can be thought of as a compiler-checked assertion that a code path is unreachable.
```

src/Init/Core.lean
```diff
@@ -78,6 +99,13 @@ instance thunkCoe : CoeTail α (Thunk α) where
 abbrev Eq.ndrecOn.{u1, u2} {α : Sort u2} {a : α} {motive : α → Sort u1} {b : α} (h : a = b) (m : motive a) : motive b :=
   Eq.ndrec m h
 
+/-! # definitions  -/
+
+@[inherit_doc True.intro] def trivial : True := ⟨⟩
+
+-- proof irrelevance is built in
+theorem proofIrrel {a : Prop} (h₁ h₂ : a) : h₁ = h₂ := rfl
```
david-christiansen, line 107: This doesn't follow our usual naming convention for theorems

src/Init/Core.lean
```diff
@@ -1174,12 +1206,122 @@ gen_injective_theorems% Lean.Syntax
...
 
-/-! # Quotients -/
+/-! # Prop lemmas -/
+-- These may depend on propext
+
+theorem Eq.comm {a b : α} : a = b ↔ b = a := Iff.intro Eq.symm Eq.symm
+theorem eq_comm {a b : α} : a = b ↔ b = a := Eq.comm
+
+/-- Ex falso for negation. From `¬a` and `a` anything follows. This is the same as `absurd` with
+the arguments flipped, but it is in the `not` namespace so that projection notation can be used. -/
```
david-christiansen, line 1216:
```suggestion
the arguments flipped, but it is in the `Not` namespace so that projection notation can be used. -/
```

src/Init/Core.lean
```diff
@@ -1174,12 +1206,122 @@ gen_injective_theorems% Lean.Syntax
...
   ⟨eq_of_beq, by intro h; subst h; exact LawfulBEq.rfl⟩
 
-/-! # Quotients -/
+/-! # Prop lemmas -/
+-- These may depend on propext
+
+theorem Eq.comm {a b : α} : a = b ↔ b = a := Iff.intro Eq.symm Eq.symm
+theorem eq_comm {a b : α} : a = b ↔ b = a := Eq.comm
+
+/-- Ex falso for negation. From `¬a` and `a` anything follows. This is the same as `absurd` with
```
david-christiansen, line 1215:
```suggestion
/-- *Ex falso* for negation: from `¬a` and `a` anything follows. This is the same as `absurd` with
```

src/Init/Core.lean
```diff
@@ -1174,12 +1206,122 @@ gen_injective_theorems% Lean.Syntax
...
+theorem not_true : (¬True) ↔ False := iff_false_intro (not_not_intro True.intro)
+
+theorem not_false_iff : (¬False) ↔ True := iff_true_intro not_false
+
+theorem Eq.to_iff : a = b → (a ↔ b) := Iff.of_eq
+theorem iff_of_eq : a = b → (a ↔ b) := Iff.of_eq
+
+theorem iff_iff_eq : (a ↔ b) ↔ a = b := Iff.intro propext Iff.of_eq
+theorem eq_iff_iff : (a = b) ↔ (a ↔ b) := iff_iff_eq.symm
+-- Jhx note. `eq_iff_iff` is labeled @[simp] in Std/Mathlib, but that current breaks Lean core eproofs.
```
david-christiansen, line 1260: Can we date these notes as well? Git blame always works, of course, but it does help give readers context as to what "currently" means.

```suggestion
-- Jhx note, Feb. 2024: `eq_iff_iff` is labeled @[simp] in Std/Mathlib, but that currently breaks Lean core proofs.
```

src/Init/Core.lean
```diff
@@ -1174,12 +1206,122 @@ gen_injective_theorems% Lean.Syntax
...
+@[simp] theorem imp_not_self : (a → ¬a) ↔ ¬a := Iff.intro (fun h ha => h ha ha) (fun h _ => h)
+
+
+theorem imp_intro {α β : Prop} (h : α) : β → α := fun _ => h
+
+theorem imp_imp_imp {a b c d : Prop} (h₀ : c → a) (h₁ : b → d) : (a → b) → (c → d) := (h₁ ∘ · ∘ h₀)
+
+theorem imp_iff_right {a : Prop} (ha : a) : (a → b) ↔ b := Iff.intro (· ha) (fun a _ => a)
+
+-- This is not marked `@[simp]` because we have `implies_true : (α → True) = True` in core.
```
david-christiansen, line 1296:
```suggestion
-- This is not marked `@[simp]` because we have `implies_true : (α → True) = True`
```

src/Init/SimpLemmas.lean
```diff
@@ -114,6 +117,94 @@ end SimprocHelperLemmas
 @[simp] theorem not_false_eq_true : (¬ False) = True := eq_true False.elim
 @[simp] theorem not_true_eq_false : (¬ True) = False := by decide
 
+-- From Std.
```
david-christiansen, line 120:
```suggestion
```
Do we need to track bits of history in comments like this? I suspect it will be more confusing than anything a year from now.

src/Init/SimpLemmas.lean
```diff
@@ -114,6 +117,94 @@ end SimprocHelperLemmas
...
+@[simp] theorem not_iff_self : ¬(¬a ↔ a) | H => iff_not_self H.symm
+
+/-! ## and -/
+
+/-- Non-dependent eliminator for `And`. -/
+abbrev And.elim (f : a → b → α) (h : a ∧ b) : α := f h.left h.right
+
+-- TODO: rename and_self to and_self_eq
```
david-christiansen, line 128: Is this in the wrong place?

src/Init/Core.lean
digama0: This was on the list from #1897

joehendrix: I made the proofIrrel since it seemed to require minimal changes.  I'll try to re-evaluate on a case-by-case basis before merging.

src/Init/Core.lean
joehendrix: I was informally using "Jhx note" to mean a note to myself for porting.  I've added the simp attribute now and fixed the proofs that broke.

src/Init/SimpLemmas.lean
joehendrix: Good catch!

src/Init/Core.lean
```diff
@@ -575,8 +605,9 @@ theorem Ne.elim (h : a ≠ b) : a = b → False := h
 
 theorem Ne.irrefl (h : a ≠ a) : False := h rfl
 
-theorem Ne.symm (h : a ≠ b) : b ≠ a :=
-  fun h₁ => h (h₁.symm)
+theorem Ne.symm (h : a ≠ b) : b ≠ a := fun h₁ => h (h₁.symm)
+
+theorem ne_comm {α} {a b : α} : a ≠ b ↔ b ≠ a := ⟨Ne.symm, Ne.symm⟩
```
joehendrix, line 610: I know the naming looks inconsistent here.  I could make a pass through to clean it up, but those changes would make the diff even bigger.

semorrison: It is inconsistent for a good reason: in the first case we can use dot notation, and this is done frequently with `Ne`.

semorrison:
Adding `full-ci` so I can look at Std/Mathlib with this PR.

src/Init/PropLemmas.lean
```diff
@@ -0,0 +1,442 @@
...
+theorem not_and_of_not_or_not (h : ¬a ∨ ¬b) : ¬(a ∧ b) := h.elim (mt (·.1)) (mt (·.2))
+
+/-! ## exists and forall -/
+
+section quantifiers
+variable {p q : α → Prop} {b : Prop}
+
+theorem forall_imp (h : ∀ a, p a → q a) : (∀ a, p a) → ∀ a, q a := fun h' a => h a (h' a)
+
+@[simp] theorem forall_exists_index {q : (∃ x, p x) → Prop} :
```
semorrison, line 124:
```suggestion
/--
As `simp` does not index foralls, this `@[simp]` lemma is tried on every `forall` expression.
This is not ideal, and likely a performance issue, but it is difficult to remove this attribute at this time.
-/
@[simp] theorem forall_exists_index {q : (∃ x, p x) → Prop} :
```

src/Init/PropLemmas.lean
```diff
@@ -0,0 +1,442 @@
...
+
+theorem exists_imp : ((∃ x, p x) → b) ↔ ∀ x, p x → b := forall_exists_index
+
+@[simp] theorem exists_const (α) [i : Nonempty α] : (∃ _ : α, b) ↔ b :=
+  ⟨fun ⟨_, h⟩ => h, i.elim Exists.intro⟩
+
+section forall_congr
+
+-- Port note: this is `forall_congr` from Lean 3. In Lean 4, there is already something
+-- with that name and a slightly different type.
```
semorrison, line 143:
```suggestion
```

src/Init/PropLemmas.lean
```diff
@@ -0,0 +1,442 @@
...
+  (exists_congr <| by exact fun a => And.comm).trans exists_eq_left
+
+@[simp] theorem exists_and_left : (∃ x, b ∧ p x) ↔ b ∧ (∃ x, p x) :=
+  ⟨fun ⟨x, h, hp⟩ => ⟨h, x, hp⟩, fun ⟨h, x, hp⟩ => ⟨x, h, hp⟩⟩
+
+@[simp] theorem exists_and_right : (∃ x, p x ∧ b) ↔ (∃ x, p x) ∧ b := by simp [And.comm]
+
+@[simp] theorem exists_eq_left' : (∃ a, a' = a ∧ p a) ↔ p a' := by simp [@eq_comm _ a']
+
+-- this theorem is needed to simplify the output of `List.mem_cons_iff`
```
semorrison, line 232:
```suggestion
-- this theorem is needed to simplify the output of `List.mem_cons_iff`
```
Don't mind if we drop this comment. The theorem statement looks crazy at first, but is actually pretty reasonable, and hard to build out of smaller simps.

src/Init/PropLemmas.lean
```diff
@@ -0,0 +1,442 @@
...
+  ⟨not_imp_symm, not_imp_symm⟩
+
+theorem Decidable.not_imp_self [Decidable a] : (¬a → a) ↔ a := by
+  have := @imp_not_self (¬a); rwa [not_not] at this
+
+theorem Decidable.or_iff_not_imp_left [Decidable a] : a ∨ b ↔ (¬a → b) :=
+  ⟨Or.resolve_left, fun h => dite _ .inl (.inr ∘ h)⟩
+
+theorem Decidable.or_iff_not_imp_right [Decidable b] : a ∨ b ↔ (¬b → a) :=
+or_comm.trans or_iff_not_imp_left
```
semorrison, line 325:
```suggestion
  or_comm.trans or_iff_not_imp_left
```

src/Init/PropLemmas.lean
```diff
@@ -0,0 +1,442 @@
...
+
+theorem Decidable.not_imp_not [Decidable a] : (¬a → ¬b) ↔ (b → a) :=
+⟨fun h hb => byContradiction (h · hb), mt⟩
```
semorrison, line 328:
```suggestion
  ⟨fun h hb => byContradiction (h · hb), mt⟩
```

src/Init/PropLemmas.lean
```diff
@@ -0,0 +1,442 @@
...
+  else by simp only [ha, false_or, false_and, false_imp_iff]
+
+theorem Decidable.or_congr_left' [Decidable c] (h : ¬c → (a ↔ b)) : a ∨ c ↔ b ∨ c := by
+  rw [or_iff_not_imp_right, or_iff_not_imp_right]; exact imp_congr_right h
+
+theorem Decidable.or_congr_right' [Decidable a] (h : ¬a → (b ↔ c)) : a ∨ b ↔ a ∨ c := by
+  rw [or_iff_not_imp_left, or_iff_not_imp_left]; exact imp_congr_right h
+
+/-- Transfer decidability of `a` to decidability of `b`, if the propositions are equivalent.
+**Important**: this function should be used instead of `rw` on `decidable b`, because the
```
semorrison, line 411:
```suggestion
**Important**: this function should be used instead of `rw` on `Decidable b`, because the
```

src/Init/PropLemmas.lean
```diff
@@ -0,0 +1,442 @@
...
+kernel will get stuck reducing the usage of `propext` otherwise,
+and `dec_trivial` will not work. -/
```
semorrison, line 413:
```suggestion
and `decide` will not work. -/
```

src/Init/ByCases.lean
```diff
@@ -0,0 +1,74 @@
...
+`by_cases (h :)? p` splits the main goal into two cases, assuming `h : p` in the first branch, and `h : ¬ p` in the second branch.
+-/
+syntax "by_cases " (atomic(ident " : "))? term : tactic
+
+macro_rules
+  | `(tactic| by_cases $e) => `(tactic| by_cases h : $e)
+macro_rules
+  | `(tactic| by_cases $h : $e) =>
+    `(tactic| open Classical in refine if $h:ident : $e then ?pos else ?neg)
+
```
semorrison, line 21: This can be removed, no?

semorrison: Oh, or is there some boostrapping issue? We already have `by_cases` and should be duplicating the syntax here.

semorrison: c.f. https://github.com/leanprover/lean4/pull/3277/files

src/Init/Core.lean
```diff
@@ -1174,12 +1220,118 @@ gen_injective_theorems% Lean.Syntax
 @[simp] theorem beq_iff_eq [BEq α] [LawfulBEq α] (a b : α) : a == b ↔ a = b :=
   ⟨eq_of_beq, by intro h; subst h; exact LawfulBEq.rfl⟩
 
-/-! # Quotients -/
+/-! # Prop lemmas -/
+-- These may depend on propext
```
joehendrix, line 1224:
```suggestion
```

src/Init/SimpLemmas.lean
```diff
@@ -93,16 +96,22 @@ theorem dite_cond_eq_true {α : Sort u} {c : Prop} {_ : Decidable c} {t : c →
...
+@[simp] theorem and_not_self : ¬(a ∧ ¬a) | ⟨ha, hn⟩ => absurd ha hn
+@[simp] theorem not_and_self : ¬(¬a ∧ a) := and_not_self ∘ And.symm
+@[simp] theorem and_imp : (a ∧ b → c) ↔ (a → b → c) := ⟨fun h ha hb => h ⟨ha, hb⟩, fun h ⟨ha, hb⟩ => h ha hb⟩
+@[simp] theorem not_and : ¬(a ∧ b) ↔ (a → ¬b) := and_imp
 @[simp] theorem or_self (p : Prop) : (p ∨ p) = p := propext ⟨fun | .inl h | .inr h => h, .inl⟩
 @[simp] theorem or_true (p : Prop) : (p ∨ True) = True := eq_true (.inr trivial)
 @[simp] theorem true_or (p : Prop) : (True ∨ p) = True := eq_true (.inl trivial)
 @[simp] theorem or_false (p : Prop) : (p ∨ False) = p := propext ⟨fun (.inl h) => h, .inl⟩
 @[simp] theorem false_or (p : Prop) : (False ∨ p) = p := propext ⟨fun (.inr h) => h, .inr⟩
+
```
joehendrix, line 114:
```suggestion
```

src/Init/SimpLemmas.lean
```diff
@@ -114,6 +123,59 @@ end SimprocHelperLemmas
 @[simp] theorem not_false_eq_true : (¬ False) = True := eq_true False.elim
 @[simp] theorem not_true_eq_false : (¬ True) = False := by decide
 
+@[simp] theorem not_iff_self : ¬(¬a ↔ a) | H => iff_not_self H.symm
+
+
```
joehendrix, line 128:
```suggestion
```

semorrison:
Looks good to me. We have clean builds of Std and Mathlib on `lean-pr-testing-3312` on each repo.

PR #3318, opened 2024-02-13
fix: trailing whitespace in `location` formatter

digama0:
This causes problems when used in conjunction with `#guard_msgs` (which checks whitespace) and trailing whitespace removal. Discovered by @PatrickMassot in verbose-lean4.


PatrickMassot:
Thank you so much @digama0! I would have had a really hard time debugging this.

PR #3320, opened 2024-02-13
chore: upstream Std.Util.ExtendedBinders

kim-em:
This is not a complete upstreaming of that file (it also supports `∀ᵉ (x < 2) (y < 3), p x y` as shorthand for `∀ x < 2, ∀ y < 3, p x y`, but I don't think we need this; it is used in Mathlib).

Syntaxes still need to be made built-in.

### DIFF
--- a/tests/lean/interactive/keywordCompletion.lean.expected.out
+++ b/tests/lean/interactive/keywordCompletion.lean.expected.out
@@ -2,6 +2,7 @@
  "position": {"line": 4, "character": 10}}
 {"items":
  [{"label": "bin", "kind": 21, "detail": "Type 1"},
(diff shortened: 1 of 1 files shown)


digama0:
> This is not a complete upstreaming of that file (it also supports `∀ᵉ (x < 2) (y < 3), p x y` as shorthand for `∀ x < 2, ∀ y < 3, p x y`, but I don't think we need this; it is used in Mathlib).

I believe that if we upstream this, it can lose the `ᵉ` and just become an official part of the `∀` syntax. This was a workaround for not being able to extend the existing grammar construction.

semorrison:
Upstreaming `∀ᵉ` (without the `ᵉ`) is on my todo list, but not for the short term.

PR #3322, opened 2024-02-13
chore: upstream norm_cast tactic

kim-em:
This is a quite substantial tactic.

It also includes the infamour `NatCast` typeclass (which I've equipped with a module-doc). I wasn't at all sure where that should live, so it is currently randomly in `Lean/Elan/Tactic/NatCast.lean`: presumably if we're doing this it will go somewhere in `Init`.

### DIFF
--- a/src/Init/Notation.lean
+++ b/src/Init/Notation.lean
@@ -484,7 +484,7 @@ instance : Coe Syntax (TSyntax `rawStx) where
 /-- `with_annotate_term stx e` annotates the lexical range of `stx : Syntax` with term info for `e`. -/
 scoped syntax (name := withAnnotateTerm) "with_annotate_term " rawStx ppSpace term : term
 
-/-- Term elaborator which uses the expected type to insert coercions. -/
+/-- Normalize casts in an expression using the same method as the `norm_cast` tactic. -/
 syntax (name := modCast) "mod_cast " term : term
 
 /--
--- a/src/Init/Tactics.lean
+++ b/src/Init/Tactics.lean
@@ -1054,15 +1054,21 @@ syntax (name := omega) "omega" (config)? : tactic
 /-- Implementation of `norm_cast` (the full `norm_cast` calls `trivial` afterwards). -/
 syntax (name := normCast0) "norm_cast0" (location)? : tactic
 
-/-- `assumption_mod_cast` runs `norm_cast` on the goal. For each local hypothesis `h`, it also
-normalizes `h` and tries to use that to close the goal. -/
+/-- `assumption_mod_cast` is a variant of `assumption` that solves the goal
+using a hypothesis. Unlike `assumption`, it first pre-processes the goal and
+each hypothesis to move casts as far outwards as possible, so it can be used
+in more situations.
+
+Concretely, it runs `norm_cast` on the goal. For each local hypothesis `h`, it also
+normalizes `h` with `norm_cast` and tries to use that to close the goal. -/
 macro "assumption_mod_cast" : tactic => `(tactic| norm_cast0 at * <;> assumption)
 
 /--
 The `norm_cast` family of tactics is used to normalize casts inside expressions.
-It is basically a simp tactic with a specific set of lemmas to move casts
+It is basically a `simp` tactic with a specific set of lemmas to move casts
 upwards in the expression.
-Therefore it can be used more safely as a non-terminating tactic.
+Therefore even in situations where non-terminal `simp` calls are discouraged (because of fragility),
+`norm_cast` is considered safe.
 It also has special handling of numerals.
 
 For instance, given an assumption
@@ -1076,25 +1082,24 @@ writing `norm_cast at h` will turn `h` into
 h : a + b < 10
 ```
 
-You can also use `exact_mod_cast`, `apply_mod_cast`, `rw_mod_cast`
-or `assumption_mod_cast`.
-Writing `exact_mod_cast h` and `apply_mod_cast h` will normalize the goal and
-`h` before using `exact h` or `apply h`.
-Writing `assumption_mod_cast` will normalize the goal and for every
-expression `h` in the context it will try to normalize `h` and use
+There are also variants of `exact`, `apply`, `rw`, and `assumption` that
+work modulo `norm_cast` - in other words, they apply `norm_cast` to make
+them more flexible. They are called `exact_mod_cast`, `apply_mod_cast`,
+`rw_mod_cast`, and `assumption_mod_cast`, respectively.
+Writing `exact_mod_cast h` and `apply_mod_cast h` will normalize casts
+in the goal and `h` before using `exact h` or `apply h`.
+Writing `assumption_mod_cast` will normalize casts in the goal and, for
+every hypothesis `h` in the context, it will try to normalize casts in `h` and use
 `exact h`.
 `rw_mod_cast` acts like the `rw` tactic but it applies `norm_cast` between steps.
 
-See also `push_cast`, for move casts inwards.
-
-The implementation and behavior of the `norm_cast` family is described in detail at
-<https://lean-forward.github.io/norm_cast/norm_cast.pdf>.
+See also `push_cast`, which moves casts inwards rather than lifting them outwards.
 -/
 macro "norm_cast" loc:(location)? : tactic =>
   `(tactic| norm_cast0 $[$loc]? <;> try trivial)
 
 /--
-`push_cast` rewrites the expression to move casts toward the leaf nodes.
+`push_cast` rewrites the goal to move casts inward, toward the leaf nodes.
 This uses `norm_cast` lemmas in the forward direction.
 For example, `↑(a + b)` will be written to `↑a + ↑b`.
 It is equivalent to `simp only with push_cast`.
@@ -1110,9 +1115,6 @@ begin
   push_cast [int.add_zero] at h2,
 end
 ```
-
-The implementation and behavior of the `norm_cast` family is described in detail at
-<https://lean-forward.github.io/norm_cast/norm_cast.pdf>.
 -/
 syntax (name := pushCast) "push_cast" (config)? (discharger)? (&" only")?
   (" [" (simpStar <|> simpErase <|> simpLemma),* "]")? (location)? : tactic
@@ -1171,7 +1173,7 @@ syntax normCastLabel := &"elim" <|> &"move" <|> &"squash"
 
 /--
 The `norm_cast` attribute should be given to lemmas that describe the
-behaviour of a coercion in regard to an operator, a relation, or a particular
+behaviour of a coercion with respect to an operator, a relation, or a particular
 function.
 
 It only concerns equality or iff lemmas involving `↑`, `⇑` and `↥`, describing the behavior of
@@ -1204,8 +1206,9 @@ Lemmas tagged with `@[norm_cast]` are classified into three categories: `move`, 
 and to cancel them from both sides of an equation or relation. It uses `squash` lemmas to clean
 up the result.
 
-Occasionally you may want to override the automatic classification.
-You can do this by giving an optional `elim`, `move`, or `squash` parameter to the attribute.
+It is typically not necessary to specify these categories, as `norm_cast` lemmas are
+automatically classified by default. The automatic classification can be overridden by
+giving an optional `elim`, `move`, or `squash` parameter to the attribute.
 
 ```lean
 @[simp, norm_cast elim] lemma nat_cast_re (n : ℕ) : (n : ℂ).re = n := by
@@ -1213,9 +1216,6 @@ You can do this by giving an optional `elim`, `move`, or `squash` parameter to t
 ```
 
 Don't do this unless you understand what you are doing.
-
-A full description of the tactic, and the use of each lemma category, can be found at
-<https://lean-forward.github.io/norm_cast/norm_cast.pdf>.
 -/
 syntax (name := norm_cast) "norm_cast" (ppSpace normCastLabel)? (ppSpace num)? : attr
 
--- a/src/Init/TacticsExtra.lean
+++ b/src/Init/TacticsExtra.lean
@@ -64,7 +64,7 @@ macro_rules
     | n+1 => `(tactic| ($seq:tacticSeq); iterate $(quote n) $seq:tacticSeq)
 
 /--
-Rewrite with the given rules and normalize casts between steps.
+Rewrites with the given rules, normalizing casts prior to each step.
 -/
 syntax "rw_mod_cast" (config)? rwRuleSeq (location)? : tactic
 macro_rules
@@ -74,12 +74,12 @@ macro_rules
     `(tactic| ($[$tacs]*))
 
 /--
-Normalize the goal and the given expression, then close the goal with exact.
+Normalize casts in the goal and the given expression, then close the goal with `exact`.
 -/
 macro "exact_mod_cast " e:term : tactic => `(tactic| exact mod_cast ($e : _))
 
 /--
-Normalize the goal and the given expression, then apply the expression to the goal.
+Normalize casts in the goal and the given expression, then `apply` the expression to the goal.
 -/
 macro "apply_mod_cast " e:term : tactic => `(tactic| apply mod_cast ($e : _))
 
--- a/src/Lean/Elab/Tactic/NormCast.lean
+++ b/src/Lean/Elab/Tactic/NormCast.lean
@@ -9,6 +9,9 @@ import Lean.Elab.ElabRules
 
 /-!
 # The `norm_cast` family of tactics.
+
+A full description of the tactic, and the use of each theorem category, can be found at
+<https://arxiv.org/abs/2001.10594>.
 -/
 namespace Lean.Elab.Tactic.NormCast
 open Lean Meta Simp NormCast
@@ -16,7 +19,7 @@ open Lean Meta Simp NormCast
 -- TODO: trace name consistency
 builtin_initialize registerTraceClass `Tactic.norm_cast
 
-/-- Prove `a = b` using the given simp set. -/
+/-- Proves `a = b` using the given simp set. -/
 def proveEqUsing (s : SimpTheorems) (a b : Expr) : MetaM (Option Simp.Result) := do
   let go : SimpM (Option Simp.Result) := do
     let a' ← Simp.simp a
@@ -27,19 +30,19 @@ def proveEqUsing (s : SimpTheorems) (a b : Expr) : MetaM (Option Simp.Result) :=
     (go (← Simp.mkDefaultMethods).toMethodsRef
       { simpTheorems := #[s], congrTheorems := ← Meta.getSimpCongrTheorems }).run' {}
 
-/-- Prove `a = b` by simplifying using move and squash lemmas. -/
+/-- Proves `a = b` by simplifying using move and squash lemmas. -/
 def proveEqUsingDown (a b : Expr) : MetaM (Option Simp.Result) := do
   withTraceNode `Tactic.norm_cast (return m!"{exceptOptionEmoji ·} proving: {← mkEq a b}") do
   proveEqUsing (← normCastExt.down.getTheorems) a b
 
-/-- Construct the expression `(e : ty)`. -/
+/-- Constructs the expression `(e : ty)`. -/
 def mkCoe (e : Expr) (ty : Expr) : MetaM Expr := do
   let .some e' ← coerce? e ty | failure
   return e'
 
 /--
-Check if an expression is the coercion of some other expression,
-and if so return that expression.
+Checks whether an expression is the coercion of some other expression,
+and if so returns that expression.
 -/
 def isCoeOf? (e : Expr) : MetaM (Option Expr) := do
   if let Expr.const fn .. := e.getAppFn then
@@ -49,8 +52,8 @@ def isCoeOf? (e : Expr) : MetaM (Option Expr) := do
   return none
 
 /--
-Check if an expression is a numeral in some type,
-and if so return that type and the natural number.
+Checks whether an expression is a numeral in some type,
+and if so returns that type and the natural number.
 -/
 def isNumeral? (e : Expr) : Option (Expr × Nat) :=
   -- TODO: cleanup, and possibly remove duplicate
@@ -65,9 +68,19 @@ def isNumeral? (e : Expr) : Option (Expr × Nat) :=
 /--
 This is the main heuristic used alongside the elim and move lemmas.
 The goal is to help casts move past operators by adding intermediate casts.
-An expression of the shape: op (↑(x : α) : γ) (↑(y : β) : γ)
-is rewritten to:            op (↑(↑(x : α) : β) : γ) (↑(y : β) : γ)
-when (↑(↑(x : α) : β) : γ) = (↑(x : α) : γ) can be proven with a squash lemma
+An expression of the shape:
+```
+op (↑(x : α) : γ) (↑(y : β) : γ)
+```
+is rewritten to:
+```
+op (↑(↑(x : α) : β) : γ) (↑(y : β) : γ)
+```
+when
+```
+(↑(↑(x : α) : β) : γ) = (↑(x : α) : γ)
+```
+can be proven with a squash lemma
 -/
 def splittingProcedure (expr : Expr) : MetaM Simp.Result := do
   let Expr.app (Expr.app op x ..) y .. := expr | return {expr}
@@ -140,7 +153,7 @@ partial def upwardAndElim (up : SimpTheorems) (e : Expr) : SimpM Simp.Step := do
   return Simp.Step.visit r
 
 /--
-If possible, rewrite `(n : α)` to `(Nat.cast n : α)` where `n` is a numeral and `α ≠ ℕ`.
+If possible, rewrites `(n : α)` to `(Nat.cast n : α)` where `n` is a numeral and `α ≠ ℕ`.
 Returns a pair of the new expression and proof that they are equal.
 -/
 def numeralToCoe (e : Expr) : MetaM Simp.Result := do
@@ -236,14 +249,12 @@ def evalNormCast0 : Tactic := fun stx => do
         (← (← getMainGoal).getNondepPropHyps).forM normCastHyp
   | _ => throwUnsupportedSyntax
 
-@[inherit_doc Lean.Parser.Tactic.Conv.normCast,
-  builtin_tactic Lean.Parser.Tactic.Conv.normCast]
+@[builtin_tactic Lean.Parser.Tactic.Conv.normCast]
 def evalConvNormCast : Tactic :=
   open Elab.Tactic.Conv in fun _ => withMainContext do
     applySimpResult (← derive (← getLhs))
 
-@[inherit_doc Lean.Parser.Tactic.pushCast,
-  builtin_tactic pushCast]
+@[builtin_tactic pushCast]
 def evalPushCast : Tactic := fun stx => do
   let { ctx, simprocs, dischargeWrapper } ← withMainContext do
     mkSimpContext (simpTheorems := pushCastExt.getTheorems) stx (eraseLocal := false)
--- a/src/Lean/Meta/Tactic/NormCast.lean
+++ b/src/Lean/Meta/Tactic/NormCast.lean
@@ -35,7 +35,7 @@ def getSimpArgs (e : Expr) : MetaM (Array Expr) := do
         args := args.push a
     return args
 
-/-- Count how many coercions are at the top of the expression. -/
+/-- Counts how many coercions are at the head of the expression. -/
 partial def countHeadCoes (e : Expr) : MetaM Nat := do
   if let Expr.const fn .. := e.getAppFn then
     if let some info ← getCoeFnInfo? fn then
@@ -43,7 +43,7 @@ partial def countHeadCoes (e : Expr) : MetaM Nat := do
         return (← countHeadCoes (e.getArg! info.coercee)) + 1
   return 0
 
-/-- Count how many coercions are inside the expression, including the top ones. -/
+/-- Counts how many coercions are inside the expression, including the head ones. -/
 partial def countCoes (e : Expr) : MetaM Nat :=
   lambdaTelescope e fun _ e => do
     if let Expr.const fn .. := e.getAppFn then
@@ -55,7 +55,7 @@ partial def countCoes (e : Expr) : MetaM Nat :=
           return coes
     return (← (← getSimpArgs e).mapM countCoes).foldl (·+·) 0
 
-/-- Count how many coercions are inside the expression, excluding the top ones. -/
+/-- Counts how many coercions are inside the expression, excluding the head ones. -/
 def countInternalCoes (e : Expr) : MetaM Nat :=
   return (← countCoes e) - (← countHeadCoes e)
 
@@ -95,15 +95,15 @@ initialize pushCastExt : SimpExtension ←
     The `push_cast` simp attribute uses `norm_cast` lemmas \
     to move casts toward the leaf nodes of the expression."
 
-/--  The `norm_cast` attribute stores three simp sets. -/
+/--  The `norm_cast` attribute stores a simp set for each of the three types of `norm_cast` lemma. -/
 structure NormCastExtension where
-  /-- A simp set which lifts coercion arrows to the top level. -/
+  /-- A simp set which lifts coercions to the top level. -/
   up : SimpExtension
-  /-- A simp set which pushes coercion arrows to the leaves. -/
+  /-- A simp set which pushes coercions to the leaves. -/
   down : SimpExtension
   /-- A simp set which simplifies transitive coercions. -/
   squash : SimpExtension
-  deriving Inhabited
+deriving Inhabited
 
 /-- The `norm_cast` extension data. -/
 builtin_initialize normCastExt : NormCastExtension ← pure {
@@ -112,26 +112,27 @@ builtin_initialize normCastExt : NormCastExtension ← pure {
   squash := ← mkSimpExt (decl_name% ++ `squash)
 }
 
-/-- `addElim decl` adds `decl` as an `elim` lemma to the cache. -/
+/-- `addElim decl` adds `decl` as an `elim` lemma to be used by `norm_cast`. -/
 def addElim (decl : Name)
     (kind := AttributeKind.global) (prio := eval_prio default) : MetaM Unit :=
   addSimpTheorem normCastExt.up decl (post := true) (inv := false) kind prio
 
-/-- `addMove decl` adds `decl` as a `move` lemma to the cache. -/
+/-- `addMove decl` adds `decl` as a `move` lemma to be used by `norm_cast`. -/
 def addMove (decl : Name)
     (kind := AttributeKind.global) (prio := eval_prio default) : MetaM Unit := do
   addSimpTheorem pushCastExt decl (post := true) (inv := false) kind prio
   addSimpTheorem normCastExt.up decl (post := true) (inv := true) kind prio
   addSimpTheorem normCastExt.down decl (post := true) (inv := false) kind prio
 
-/-- `addSquash decl` adds `decl` as a `squash` lemma to the cache. -/
+/-- `addSquash decl` adds `decl` as a `squash` lemma to be used by `norm_cast`. -/
 def addSquash (decl : Name)
     (kind := AttributeKind.global) (prio := eval_prio default) : MetaM Unit := do
   addSimpTheorem pushCastExt decl (post := true) (inv := false) kind prio
   addSimpTheorem normCastExt.squash decl (post := true) (inv := false) kind prio
   addSimpTheorem normCastExt.down decl (post := true) (inv := false) kind prio
 
-/-- `addInfer decl` infers the label of `decl` and adds it to the cache.
+/-- `addInfer decl` infers the label of `decl` (`elim`, `move`, or `squash`) and arranges for it to
+be used by `norm_cast`.
 
 * elim lemma:   LHS has 0 head coes and ≥ 1 internal coe
 * move lemma:   LHS has 1 head coe and 0 internal coes,    RHS has 0 head coes and ≥ 1 internal coes
--- a/src/Init/Data/Array/Subarray.lean
+++ b/src/Init/Data/Array/Subarray.lean
@@ -143,6 +143,7 @@ def toSubarray (as : Array α) (start : Nat := 0) (stop : Nat := as.size) : Suba
      else
        { as := as, start := as.size, stop := as.size, h₁ := Nat.le_refl _, h₂ := Nat.le_refl _ }
 
+@[coe]
 def ofSubarray (s : Subarray α) : Array α := Id.run do
   let mut as := mkEmpty (s.stop - s.start)
   for a in s do
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -1814,6 +1814,8 @@ structure Fin (n : Nat) where
   /-- If `i : Fin n`, then `i.2` is a proof that `i.1 < n`. -/
   isLt : LT.lt val n
 
+attribute [coe] Fin.val
+
 theorem Fin.eq_of_val_eq {n} : ∀ {i j : Fin n}, Eq i.val j.val → Eq i j
   | ⟨_, _⟩, ⟨_, _⟩, rfl => rfl
 
--- a/tests/lean/delabProjectionApp.lean
+++ b/tests/lean/delabProjectionApp.lean
@@ -47,13 +47,16 @@ set_option pp.structureProjections false
 
 end
 
-structure Fin' extends Fin 5
+structure Fin0 where
+  val : Nat
 
-structure Fin'' (n : Nat) extends Fin n
+structure Fin' extends Fin0
+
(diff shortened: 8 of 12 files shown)


leodemoura:
@nomeata The new test file is a good example of what this PR is fixing: https://github.com/leanprover/lean4/blob/e7cc454161017598c7bf5d472401de54ae90765e/tests/lean/evalLeak.lean

src/Init/Data/Cast.lean
```diff
@@ -0,0 +1,101 @@
...
+import Init.Data.Nat.Basic
+import Init.Data.Int.Basic
+
+/-!
+# `NatCast` and `IntCast`
+
+We introduce the typeclass `NatCast R` for a type `R` with a "canonical homomorphism" `Nat → R`.
+The typeclass carries the data of the function, but no required axioms.
+
+This typeclass was introduced to a uniform simp normal form for such morphisms.
```
david-christiansen, line 16:
```suggestion
This typeclass provides a uniform `simp` normal form for such morphisms.
```
The existing text wasn't grammatical as far as I could see - but I think it's also better to talk about what it does, rather than why it came to be in the past, in the primary position in reference docs.

david-christiansen: Or is this the ported library note? In that case, I think it's better to leave the text mostly as is, but indicate this fact higher up. Then I'd say:
```suggestion
This typeclass was introduced to support a uniform `simp` normal form for such morphisms.
```

src/Init/Data/Cast.lean
```diff
@@ -0,0 +1,101 @@
...
+
+Without such a typeclass, we would have coercions such as `Int.ofNat`,
+but also later in Mathlib the generic coercion from `Nat` into any semiring (including `Int`),
+and we would need to use `simp` to move between them.
+However `simp` lemmas expressed using a non-normal form on the LHS would then not fire.
+
+In order to avoid this, we introduce this typeclass as early as possible,
+ensure it is a higher priority coercion that built-in coercions such as `Int.ofNat`,
```
david-christiansen, line 24: Now that this is being upstreamed, what about deleting `Int.ofNat` and using only this one? That would simplify things a bit.

src/Init/Data/Cast.lean
```diff
@@ -0,0 +1,101 @@
...
+  /-- The canonical map `Nat → R`. -/
+  protected natCast : Nat → R
+
+instance : NatCast Nat where natCast n := n
+instance : NatCast Int where natCast n := Int.ofNat n
+
+/--
+Canonical homomorphism from `Nat` to a type `R`.
+
+It contains just the function, with no axioms.
```
david-christiansen, line 60: Even though the class doesn't impose proof obligations as axioms, it would be nice to state the properties that are expected of an implementation in this docstring. Just saying "canonical homomorphism" won't help a lot of readers who are wondering if they should implement it.

src/Init/Data/Cast.lean
```diff
@@ -0,0 +1,101 @@
...
+
+The prototypical example is `Int.ofNat`.
+
+In Mathlib there will be such a homomorphism whenever `R` is an additive monoid with a `1`.
```
david-christiansen, line 64:
```suggestion
```
I don't think our docs should contain forward references to other projects/libraries - we have enough difficulties keeping them internally consistent!

src/Init/Data/Cast.lean
```diff
@@ -0,0 +1,101 @@
...
+@[coe, reducible, match_pattern] protected def Nat.cast {R : Type u} [NatCast R] : Nat → R :=
+  NatCast.natCast
+
+-- see the notes about coercions into arbitrary types in the module doc-string
+instance [NatCast R] : CoeTail Nat R where coe := Nat.cast
+
+-- see the notes about coercions into arbitrary types in the module doc-string
+instance [NatCast R] : CoeHTCT Nat R where coe := Nat.cast
+
+/-- This instance is needed to ensure that `instCoeNatInt` from core is not used. -/
```
david-christiansen, line 75:
```suggestion
/-- This instance is needed to ensure that `instCoeNatInt` is not used. -/
```

src/Init/Data/Cast.lean
```diff
@@ -0,0 +1,101 @@
...
+instance : IntCast Int where intCast n := n
+
+/--
+Canonical homomorphism from `Nat` to a type `R`.
+
+It contains just the function, with no axioms.
+
+The prototypical example is `Int.ofNat`.
+
+In Mathlib there will be such a homomorphism whenever `R` is an additive group with a `1`.
```
david-christiansen, line 92:
```suggestion
```
Forward references to downstream projects seems like a bad idea to me - I'd like to avoid the maintenance headache of keeping it all consistent!

src/Init/Data/Cast.lean
```diff
@@ -0,0 +1,101 @@
...
+-- see the notes about coercions into arbitrary types in the module doc-string
+instance [NatCast R] : CoeTail Nat R where coe := Nat.cast
+
+-- see the notes about coercions into arbitrary types in the module doc-string
+instance [NatCast R] : CoeHTCT Nat R where coe := Nat.cast
+
+/-- This instance is needed to ensure that `instCoeNatInt` from core is not used. -/
+instance : Coe Nat Int where coe := Nat.cast
+
+/-- Type class for the canonical homomorphism `Int → R`. -/
```
david-christiansen, line 78:
```suggestion
/-- The canonical homomorphism `Int → R`. -/
```
We know it's a type class :)

I think the docstring for the class should include the properties expected of instances, even though we're not imposing proof obligations. Which structure is preserved by the homomorphism?

src/Init/Data/Cast.lean
```diff
@@ -0,0 +1,101 @@
...
+  protected intCast : Int → R
+
+instance : IntCast Int where intCast n := n
+
+/--
+Canonical homomorphism from `Nat` to a type `R`.
+
+It contains just the function, with no axioms.
+
+The prototypical example is `Int.ofNat`.
```
david-christiansen, line 90: I think this should be moved to the class's docstring

src/Init/Data/Cast.lean
```diff
@@ -0,0 +1,101 @@
...
+class IntCast (R : Type u) where
+  /-- The canonical map `Int → R`. -/
+  protected intCast : Int → R
+
+instance : IntCast Int where intCast n := n
+
+/--
+Canonical homomorphism from `Nat` to a type `R`.
+
+It contains just the function, with no axioms.
```
david-christiansen, line 88:
```suggestion
Apply the canonical homomorphism from `Int` to a type `R` from an `IntCast R` instance.
```

src/Init/Notation.lean
```diff
@@ -476,6 +476,9 @@ instance : Coe Syntax (TSyntax `rawStx) where
 /-- `with_annotate_term stx e` annotates the lexical range of `stx : Syntax` with term info for `e`. -/
 scoped syntax (name := withAnnotateTerm) "with_annotate_term " rawStx ppSpace term : term
 
+/-- Term elaborator which uses the expected type to insert coercions. -/
```
david-christiansen, line 479: Is this intended for users? I think I would find it quite unhelpful in a docstring. Doesn't up-arrow already do this?

src/Init/Tactics.lean
```diff
@@ -894,6 +894,73 @@ The tactic `nomatch h` is shorthand for `exact nomatch h`.
...
+/-- Implementation of `norm_cast` (the full `norm_cast` calls `trivial` afterwards). -/
+syntax (name := normCast0) "norm_cast0" (location)? : tactic
+
+/-- `assumption_mod_cast` runs `norm_cast` on the goal. For each local hypothesis `h`, it also
+normalizes `h` and tries to use that to close the goal. -/
+macro "assumption_mod_cast" : tactic => `(tactic| norm_cast0 at * <;> assumption)
+
+/--
+The `norm_cast` family of tactics is used to normalize casts inside expressions.
+It is basically a simp tactic with a specific set of lemmas to move casts
```
david-christiansen, line 907:
```suggestion
It is basically a `simp` tactic with a specific set of lemmas to move casts
```

src/Init/Tactics.lean
```diff
@@ -894,6 +894,73 @@ The tactic `nomatch h` is shorthand for `exact nomatch h`.
...
+upwards in the expression.
+Therefore it can be used more safely as a non-terminating tactic.
```
david-christiansen, line 909: I don't understand this sentence - can you elaborate?

src/Init/Tactics.lean
```diff
@@ -894,6 +894,73 @@ The tactic `nomatch h` is shorthand for `exact nomatch h`.
...
+It also has special handling of numerals.
+
+For instance, given an assumption
+```lean
+a b : ℤ
+h : ↑a + ↑b < (10 : ℚ)
```
david-christiansen, line 915:
```suggestion
a b : Int
h : ↑a + ↑b < (10 : Rational)
```
(but we don't have `Rational`, right? can this example be made to work with things that ship with Lean?)

src/Init/Tactics.lean
```diff
@@ -894,6 +894,73 @@ The tactic `nomatch h` is shorthand for `exact nomatch h`.
...
+```
+
+writing `norm_cast at h` will turn `h` into
+```lean
+h : a + b < 10
+```
```
david-christiansen, line 921: I think it's easier to understand with all the binders here too:
```suggestion
```lean
a b : Int
h : a + b < 10
```
```

File: src/Init/Data/Cast.lean (hunk 2 above)
Comment (digama0) [reply src/Init/Data/Cast.lean:24]: We can't delete `Int.ofNat`, it's one of the constructors of `Int`

File: src/Init/Tactics.lean [hunk 15]
```diff
@@ -894,6 +894,73 @@ The tactic `nomatch h` is shorthand for `exact nomatch h`.
...
+
+You can also use `exact_mod_cast`, `apply_mod_cast`, `rw_mod_cast`
+or `assumption_mod_cast`.
+Writing `exact_mod_cast h` and `apply_mod_cast h` will normalize the goal and
+`h` before using `exact h` or `apply h`.
+Writing `assumption_mod_cast` will normalize the goal and for every
+expression `h` in the context it will try to normalize `h` and use
+`exact h`.
```
david-christiansen, line 929:
```suggestion
Writing `exact_mod_cast h` and `apply_mod_cast h` will normalize casts
in the goal and `h` before using `exact h` or `apply h`.
Writing `assumption_mod_cast` will normalize casts in the goal and, for
every hypothesis `h` in the context, it will try to normalize casts in `h` and use
`exact h`.
```
This is extremely pedantic :-) But the term "normalize" is so overloaded that I want to make sure readers don't think it's performing full reduction or running `simp` or something.

src/Init/Tactics.lean
```diff
@@ -894,6 +894,73 @@ The tactic `nomatch h` is shorthand for `exact nomatch h`.
...
+`rw_mod_cast` acts like the `rw` tactic but it applies `norm_cast` between steps.
+
+See also `push_cast`, for move casts inwards.
```
david-christiansen, line 932:
```suggestion
See also `push_cast`, which moves casts inwards rather than lifting them outwards.
```

src/Init/Tactics.lean
```diff
@@ -894,6 +894,73 @@ The tactic `nomatch h` is shorthand for `exact nomatch h`.
...
+
+The implementation and behavior of the `norm_cast` family is described in detail at
+<https://lean-forward.github.io/norm_cast/norm_cast.pdf>.
```
david-christiansen, line 935: Can we get permission to mirror this at lean-lang.org so we can ensure availability?

src/Init/Data/Cast.lean
digama0: I am concerned about this general direction, referring to mathlib in lean is not / should not be illegal. We should discuss this on Zulip.

src/Init/Tactics.lean
david-christiansen: And perhaps the link to the implementation is better as a comment in the source, rather than as part of hovers in editors and the like.

src/Init/Tactics.lean
```diff
@@ -894,6 +894,73 @@ The tactic `nomatch h` is shorthand for `exact nomatch h`.
...
+-/
+macro "norm_cast" loc:(location)? : tactic =>
+  `(tactic| norm_cast0 $[$loc]? <;> try trivial)
+
+/--
+`push_cast` rewrites the expression to move casts toward the leaf nodes.
```
david-christiansen, line 941:
```suggestion
`push_cast` rewrites the goal to move casts inward, toward the leaf nodes.
```

src/Init/Tactics.lean
```diff
@@ -894,6 +894,73 @@ The tactic `nomatch h` is shorthand for `exact nomatch h`.
...
+
+```lean
+example (a b : ℕ) (h1 : ((a + b : ℕ) : ℤ) = 10) (h2 : ((a + b + 0 : ℕ) : ℤ) = 10) :
+  ((a + b : ℕ) : ℤ) = 10 :=
+begin
+  push_cast,
+  push_cast at h1,
+  push_cast [int.add_zero] at h2,
+end
+```
```
david-christiansen, line 956: Can this example be rewritten to work in Lean 4, using things only in this repo, and perhaps even added as a test case?

src/Init/Tactics.lean
```diff
@@ -894,6 +894,73 @@ The tactic `nomatch h` is shorthand for `exact nomatch h`.
...
+<https://lean-forward.github.io/norm_cast/norm_cast.pdf>.
+-/
+macro "norm_cast" loc:(location)? : tactic =>
+  `(tactic| norm_cast0 $[$loc]? <;> try trivial)
+
+/--
+`push_cast` rewrites the expression to move casts toward the leaf nodes.
+This uses `norm_cast` lemmas in the forward direction.
+For example, `↑(a + b)` will be written to `↑a + ↑b`.
+It is equivalent to `simp only with push_cast`.
```
david-christiansen, line 944: Do we have a similar property for `norm_cast`? If not, why not?

src/Init/Tactics.lean
```diff
@@ -894,6 +894,73 @@ The tactic `nomatch h` is shorthand for `exact nomatch h`.
...
+  ((a + b : ℕ) : ℤ) = 10 :=
+begin
+  push_cast,
+  push_cast at h1,
+  push_cast [int.add_zero] at h2,
+end
+```
+
+The implementation and behavior of the `norm_cast` family is described in detail at
+<https://lean-forward.github.io/norm_cast/norm_cast.pdf>.
```
david-christiansen, line 959: I think this is perhaps better as a source comment. If it's a docstring, I think we should mirror this PDF (with permission) at lean-lang.org so we can ensure availability.

src/Init/Tactics.lean
```diff
@@ -941,6 +1008,61 @@ If there are several with the same priority, it is uses the "most recent one". E
 ```
 -/
 syntax (name := simp) "simp" (Tactic.simpPre <|> Tactic.simpPost)? (ppSpace prio)? : attr
+
+
+/-- The possible `norm_cast` kinds: `elim`, `move`, or `squash`. -/
```
david-christiansen, line 1013: It'd be great if this docstring had a forward reference to the `norm_cast` docstring that describes them.

src/Init/Tactics.lean
```diff
@@ -941,6 +1008,61 @@ If there are several with the same priority, it is uses the "most recent one". E
...
+syntax normCastLabel := &"elim" <|> &"move" <|> &"squash"
+
+/--
+The `norm_cast` attribute should be given to lemmas that describe the
+behaviour of a coercion in regard to an operator, a relation, or a particular
```
david-christiansen, line 1018:
```suggestion
behaviour of a coercion with respect to an operator, a relation, or a particular
```
This phrasing sounds more natural in my ears, but this might just be my North American background. Feel free to reject if that's the case.

src/Init/Tactics.lean
```diff
@@ -941,6 +1008,61 @@ If there are several with the same priority, it is uses the "most recent one". E
...
+function.
+
+It only concerns equality or iff lemmas involving `↑`, `⇑` and `↥`, describing the behavior of
```
david-christiansen, line 1021: Do we have all of these in core now?

src/Init/Tactics.lean
```diff
@@ -941,6 +1008,61 @@ If there are several with the same priority, it is uses the "most recent one". E
...
+@[norm_cast] theorem coe_int_denom (n : ℤ) : (n : ℚ).denom = 1
+
+@[norm_cast] theorem cast_id : ∀ n : ℚ, ↑n = n
+
+@[norm_cast] theorem coe_nat_add (m n : ℕ) : (↑(m + n) : ℤ) = ↑m + ↑n
+
+@[norm_cast] theorem cast_coe_nat (n : ℕ) : ((n : ℤ) : α) = n
+
+@[norm_cast] theorem cast_one : ((1 : ℚ) : α) = 1
+```
```
david-christiansen, line 1038: These examples should be modified to work in core, and ideally also added as a test case!

src/Init/Tactics.lean
```diff
@@ -941,6 +1008,61 @@ If there are several with the same priority, it is uses the "most recent one". E
...
+
+Lemmas tagged with `@[norm_cast]` are classified into three categories: `move`, `elim`, and
+`squash`. They are classified roughly as follows:
+
+* elim lemma:   LHS has 0 head coes and ≥ 1 internal coe
+* move lemma:   LHS has 1 head coe and 0 internal coes,    RHS has 0 head coes and ≥ 1 internal coes
+* squash lemma: LHS has ≥ 1 head coes and 0 internal coes, RHS has fewer head coes
```
david-christiansen, line 1045:
```suggestion
* `elim` lemma:   the LHS does not have `coe` at its head, and contains ≥ 1 non-head `coe`
* `move` lemma:  the LHS has `coe` at its head, and no internal occurrences, while the RHS is not headed by `coe` and contains ≥ 1 internal `coe`s
* `squash` lemma: LHS has ≥ 1 `coe`s at its head and 0 internal `coe`s, RHS has fewer head `coe`s

In other words, `elim` lemmas are used to eliminate occurrences of `coe`, `move` lemmas are used to push `coe`s inwards, and `squash` lemmas are used to simplify stacked coercions.
```
If this is inaccurate (e.g. "coe" in the original may have referred to all three types of coercions mentioned earlier), let me know and I'll suggest alternate phrasing.

src/Init/Tactics.lean
```diff
@@ -941,6 +1008,61 @@ If there are several with the same priority, it is uses the "most recent one". E
...
+
+`norm_cast` uses `move` and `elim` lemmas to factor coercions toward the root of an expression
+and to cancel them from both sides of an equation or relation. It uses `squash` lemmas to clean
+up the result.
+
+Occasionally you may want to override the automatic classification.
+You can do this by giving an optional `elim`, `move`, or `squash` parameter to the attribute.
```
david-christiansen, line 1052:
```suggestion
It is typically not necessary to specify these categories, as `norm_cast` lemmas are
automatically classified by default. The automatic classification can be overridden by
giving an optional `elim`, `move`, or `squash` parameter to the attribute.
```

src/Init/Tactics.lean
```diff
@@ -941,6 +1008,61 @@ If there are several with the same priority, it is uses the "most recent one". E
...
+
+```lean
+@[simp, norm_cast elim] lemma nat_cast_re (n : ℕ) : (n : ℂ).re = n := by
+  rw [← of_real_nat_cast, of_real_re]
+```
```
david-christiansen, line 1057: This example should be rewritten to work with core Lean 4

src/Init/TacticsExtra.lean
```diff
@@ -41,4 +41,24 @@ macro_rules
   | `(tactic| if%$tk $c then%$ttk $pos else%$etk $neg) =>
     expandIfThenElse tk ttk etk pos neg fun pos neg => `(if h : $c then $pos else $neg)
 
+/--
+Rewrite with the given rules and normalize casts between steps.
```
david-christiansen, line 45:
```suggestion
Rewrites with the given rules, normalizing casts prior to each step.
```
Indicative mood, and as I read the code, it's not just between, but also at the beginning.

src/Init/TacticsExtra.lean
```diff
@@ -41,4 +41,24 @@ macro_rules
...
+-/
+syntax "rw_mod_cast" (config)? rwRuleSeq (location)? : tactic
+macro_rules
+  | `(tactic| rw_mod_cast $[$config]? [$rules,*] $[$loc]?) => do
+    let tacs ← rules.getElems.mapM fun rule =>
+      `(tactic| (norm_cast at *; rw $[$config]? [$rule] $[$loc]?))
+    `(tactic| ($[$tacs]*))
+
+/--
+Normalize the goal and the given expression, then close the goal with exact.
```
david-christiansen, line 55:
```suggestion
Normalize casts in the goal and the given expression, then close the goal with `exact`.
```

src/Init/TacticsExtra.lean
```diff
@@ -41,4 +41,24 @@ macro_rules
...
+-/
+macro "exact_mod_cast " e:term : tactic => `(tactic| exact mod_cast ($e : _))
+
+/--
+Normalize the goal and the given expression, then apply the expression to the goal.
```
david-christiansen, line 60:
```suggestion
Normalize casts in the goal and the given expression, then `apply` the expression to the goal.
```

src/Init/Notation.lean
david-christiansen: OK, having read more of this PR I get what's going on a bit more. What about:

```suggestion
/-- Normalize casts in an expression using the same method as the `norm_cast` tactic. -/
```

src/Lean/Elab/Tactic/NormCast.lean
```diff
@@ -0,0 +1,254 @@
...
+/-!
+# The `norm_cast` family of tactics.
+-/
+namespace Lean.Elab.Tactic.NormCast
+open Lean Meta Simp NormCast
+
+-- TODO: trace name consistency
+builtin_initialize registerTraceClass `Tactic.norm_cast
+
+/-- Prove `a = b` using the given simp set. -/
```
david-christiansen, line 19:
```suggestion
/-- Proves `a = b` using the given simp set. -/
```

src/Lean/Elab/Tactic/NormCast.lean
```diff
@@ -0,0 +1,254 @@
...
+  let go : SimpM (Option Simp.Result) := do
+    let a' ← Simp.simp a
+    let b' ← Simp.simp b
+    unless ← isDefEq a'.expr b'.expr do return none
+    a'.mkEqTrans (← b'.mkEqSymm b)
+  withReducible do
+    (go (← Simp.mkDefaultMethods).toMethodsRef
+      { simpTheorems := #[s], congrTheorems := ← Meta.getSimpCongrTheorems }).run' {}
+
+/-- Prove `a = b` by simplifying using move and squash lemmas. -/
```
david-christiansen, line 30:
```suggestion
/-- Proves `a = b` by simplifying using move and squash lemmas. -/
```

src/Lean/Elab/Tactic/NormCast.lean
```diff
@@ -0,0 +1,254 @@
...
+def proveEqUsingDown (a b : Expr) : MetaM (Option Simp.Result) := do
+  withTraceNode `Tactic.norm_cast (return m!"{exceptOptionEmoji ·} proving: {← mkEq a b}") do
+  proveEqUsing (← normCastExt.down.getTheorems) a b
+
+/-- Construct the expression `(e : ty)`. -/
```
david-christiansen, line 35:
```suggestion
/-- Constructs the expression `(e : ty)`. -/
```

src/Lean/Elab/Tactic/NormCast.lean
```diff
@@ -0,0 +1,254 @@
...
+def mkCoe (e : Expr) (ty : Expr) : MetaM Expr := do
+  let .some e' ← coerce? e ty | failure
+  return e'
+
+/--
+Check if an expression is the coercion of some other expression,
+and if so return that expression.
```
david-christiansen, line 42:
```suggestion
Checks whether an expression is the coercion of some other expression,
and if so returns that expression.
```

src/Lean/Elab/Tactic/NormCast.lean
```diff
@@ -0,0 +1,254 @@
...
+def isCoeOf? (e : Expr) : MetaM (Option Expr) := do
+  if let Expr.const fn .. := e.getAppFn then
+    if let some info ← getCoeFnInfo? fn then
+      if e.getAppNumArgs == info.numArgs then
+        return e.getArg! info.coercee
+  return none
+
+/--
+Check if an expression is a numeral in some type,
+and if so return that type and the natural number.
```
david-christiansen, line 53:
```suggestion
Checks whether an expression is a numeral in some type,
and if so returns that type and the natural number.
```

src/Lean/Elab/Tactic/NormCast.lean
```diff
@@ -0,0 +1,254 @@
...
+    some (α, n)
+  else
+    none
+
+/--
+This is the main heuristic used alongside the elim and move lemmas.
+The goal is to help casts move past operators by adding intermediate casts.
+An expression of the shape: op (↑(x : α) : γ) (↑(y : β) : γ)
+is rewritten to:            op (↑(↑(x : α) : β) : γ) (↑(y : β) : γ)
+when (↑(↑(x : α) : β) : γ) = (↑(x : α) : γ) can be proven with a squash lemma
```
david-christiansen, line 70:
````suggestion
An expression of the shape:
```
op (↑(x : α) : γ) (↑(y : β) : γ)
```
is rewritten to:
```
op (↑(↑(x : α) : β) : γ) (↑(y : β) : γ)
```
when
```
(↑(↑(x : α) : β) : γ) = (↑(x : α) : γ)
```
can be proven with a squash lemma
````
Should be easier to read when rendered.

File: src/Lean/Elab/Tactic/NormCast.lean [hunk 38]
```diff
@@ -0,0 +1,254 @@
...
+partial def upwardAndElim (up : SimpTheorems) (e : Expr) : SimpM Simp.Step := do
+  let r ← withDischarger prove do
+    Simp.rewrite? e up.post up.erased (tag := "squash") (rflOnly := false)
+  let r := r.getD { expr := e }
+  let r ← r.mkEqTrans (← splittingProcedure r.expr)
+  if r.expr == e then return Simp.Step.done {expr := e}
+  return Simp.Step.visit r
+
+/--
+If possible, rewrite `(n : α)` to `(Nat.cast n : α)` where `n` is a numeral and `α ≠ ℕ`.
```
david-christiansen, line 143:
```suggestion
If possible, rewrites `(n : α)` to `(Nat.cast n : α)` where `n` is a numeral and `α ≠ ℕ`.
```

src/Lean/Elab/Tactic/NormCast.lean
```diff
@@ -0,0 +1,254 @@
...
+import Lean.Elab.Tactic.Conv.Simp
+import Lean.Elab.ElabRules
+
+/-!
+# The `norm_cast` family of tactics.
+-/
+namespace Lean.Elab.Tactic.NormCast
+open Lean Meta Simp NormCast
+
+-- TODO: trace name consistency
```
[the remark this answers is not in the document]
digama0: We need an issue to track normalization of option names (in `set_option`) as well as trace scopes (which also appear as option names after `trace.`). Currently there are a bunch of incompatible conventions in use

src/Lean/Meta/Tactic/NormCast.lean
```diff
@@ -0,0 +1,162 @@
...
+-/
+inductive Label
+  /-- elim lemma: LHS has 0 head coes and ≥ 1 internal coe -/
+  | elim
+  /-- move lemma: LHS has 1 head coe and 0 internal coes,
+  RHS has 0 head coes and ≥ 1 internal coes -/
+  | move
+  /-- squash lemma: LHS has ≥ 1 head coes and 0 internal coes, RHS has fewer head coes -/
+  | squash
+  deriving DecidableEq, Repr, Inhabited
```
david-christiansen, line 25: Can we arrange there to be a single source of truth for this info? E.g. with careful use of `inherit_doc`?

src/Lean/Meta/Tactic/NormCast.lean
```diff
@@ -0,0 +1,162 @@
...
+  match ← mkCongrSimp? e.getAppFn with
+  | none => return e.getAppArgs
+  | some {argKinds, ..} =>
+    let mut args := #[]
+    for a in e.getAppArgs, k in argKinds do
+      if k matches .eq then
+        args := args.push a
+    return args
+
+/-- Count how many coercions are at the top of the expression. -/
```
david-christiansen, line 38:
```suggestion
/-- Counts how many coercions are at the top of the expression. -/
```

src/Lean/Meta/Tactic/NormCast.lean
```diff
@@ -0,0 +1,162 @@
...
+partial def countHeadCoes (e : Expr) : MetaM Nat := do
+  if let Expr.const fn .. := e.getAppFn then
+    if let some info ← getCoeFnInfo? fn then
+      if e.getAppNumArgs >= info.numArgs then
+        return (← countHeadCoes (e.getArg! info.coercee)) + 1
+  return 0
+
+/-- Count how many coercions are inside the expression, including the top ones. -/
```
david-christiansen, line 46:
```suggestion
/-- Counts how many coercions are inside the expression, including the top ones. -/
```

src/Lean/Meta/Tactic/NormCast.lean
```diff
@@ -0,0 +1,162 @@
...
+    if let Expr.const fn .. := e.getAppFn then
+      if let some info ← getCoeFnInfo? fn then
+        if e.getAppNumArgs >= info.numArgs then
+          let mut coes := (← countHeadCoes (e.getArg! info.coercee)) + 1
+          for i in [info.numArgs:e.getAppNumArgs] do
+            coes := coes + (← countCoes (e.getArg! i))
+          return coes
+    return (← (← getSimpArgs e).mapM countCoes).foldl (·+·) 0
+
+/-- Count how many coercions are inside the expression, excluding the top ones. -/
```
david-christiansen, line 58:
```suggestion
/-- Counts how many coercions are inside the expression, excluding the head ones. -/
```

src/Lean/Meta/Tactic/NormCast.lean
```diff
@@ -0,0 +1,162 @@
...
+
+/-- The `push_cast` simp attribute. -/
+initialize pushCastExt : SimpExtension ←
+  registerSimpAttr `push_cast "\
+    The `push_cast` simp attribute uses `norm_cast` lemmas \
+    to move casts toward the leaf nodes of the expression."
+
+/--  The `norm_cast` attribute stores three simp sets. -/
+structure NormCastExtension where
+  /-- A simp set which lifts coercion arrows to the top level. -/
```
david-christiansen, line 100:
```suggestion
  /-- A simp set which lifts coercions to the top level. -/
```

src/Lean/Meta/Tactic/NormCast.lean
```diff
@@ -0,0 +1,162 @@
...
+  up : SimpExtension
+  /-- A simp set which pushes coercion arrows to the leaves. -/
```
david-christiansen, line 102:
```suggestion
  /-- A simp set which pushes coercions to the leaves. -/
```

src/Lean/Meta/Tactic/NormCast.lean
```diff
@@ -0,0 +1,162 @@
...
+        norm_cast: badly shaped shaped squash lemma, \
+        rhs must have fewer head coes than lhs{indentExpr ty}"
+
+/-- The `push_cast` simp attribute. -/
+initialize pushCastExt : SimpExtension ←
+  registerSimpAttr `push_cast "\
+    The `push_cast` simp attribute uses `norm_cast` lemmas \
+    to move casts toward the leaf nodes of the expression."
+
+/--  The `norm_cast` attribute stores three simp sets. -/
```
david-christiansen, line 98:
```suggestion
/--  The `norm_cast` attribute stores a simp set for each of the three types of `norm_cast` lemma. -/
```

src/Lean/Meta/Tactic/NormCast.lean
```diff
@@ -0,0 +1,162 @@
...
+structure NormCastExtension where
+  /-- A simp set which lifts coercion arrows to the top level. -/
+  up : SimpExtension
+  /-- A simp set which pushes coercion arrows to the leaves. -/
+  down : SimpExtension
+  /-- A simp set which simplifies transitive coercions. -/
+  squash : SimpExtension
+  deriving Inhabited
```
david-christiansen, line 106:
```suggestion
deriving Inhabited
```

src/Lean/Meta/Tactic/NormCast.lean
```diff
@@ -0,0 +1,162 @@
...
+
+/-- The `norm_cast` extension data. -/
+builtin_initialize normCastExt : NormCastExtension ← pure {
+  up := ← mkSimpExt (decl_name% ++ `up)
+  down := ← mkSimpExt (decl_name% ++ `down)
+  squash := ← mkSimpExt (decl_name% ++ `squash)
+}
+
+/-- `addElim decl` adds `decl` as an `elim` lemma to the cache. -/
```
david-christiansen, line 115: Is this a cache? It seems to me like it's something else.

```suggestion
/-- `addElim decl` adds `decl` as an `elim` lemma to be used by `norm_cast`. -/
```

src/Lean/Meta/Tactic/NormCast.lean
```diff
@@ -0,0 +1,162 @@
...
+def addElim (decl : Name)
+    (kind := AttributeKind.global) (prio := eval_prio default) : MetaM Unit :=
+  addSimpTheorem normCastExt.up decl (post := true) (inv := false) kind prio
+
+/-- `addMove decl` adds `decl` as a `move` lemma to the cache. -/
```
david-christiansen, line 120:
```suggestion
/-- `addMove decl` adds `decl` as a `move` lemma to be used by `norm_cast`. -/
```

src/Lean/Meta/Tactic/NormCast.lean
```diff
@@ -0,0 +1,162 @@
...
+def addMove (decl : Name)
+    (kind := AttributeKind.global) (prio := eval_prio default) : MetaM Unit := do
+  addSimpTheorem pushCastExt decl (post := true) (inv := false) kind prio
+  addSimpTheorem normCastExt.up decl (post := true) (inv := true) kind prio
+  addSimpTheorem normCastExt.down decl (post := true) (inv := false) kind prio
+
+/-- `addSquash decl` adds `decl` as a `squash` lemma to the cache. -/
```
david-christiansen, line 127:
```suggestion
/-- `addSquash decl` adds `decl` as a `squash` lemma to be used by `norm_cast`. -/
```

src/Lean/Meta/Tactic/NormCast.lean
```diff
@@ -0,0 +1,162 @@
...
+  addSimpTheorem pushCastExt decl (post := true) (inv := false) kind prio
+  addSimpTheorem normCastExt.squash decl (post := true) (inv := false) kind prio
+  addSimpTheorem normCastExt.down decl (post := true) (inv := false) kind prio
+
+/-- `addInfer decl` infers the label of `decl` and adds it to the cache.
+
+* elim lemma:   LHS has 0 head coes and ≥ 1 internal coe
+* move lemma:   LHS has 1 head coe and 0 internal coes,    RHS has 0 head coes and ≥ 1 internal coes
+* squash lemma: LHS has ≥ 1 head coes and 0 internal coes, RHS has fewer head coes
+-/
```
david-christiansen, line 139: Can we get a single source of truth for these descriptions?

src/Lean/Meta/Tactic/NormCast.lean
```diff
@@ -0,0 +1,162 @@
...
+  addSimpTheorem normCastExt.down decl (post := true) (inv := false) kind prio
+
+/-- `addSquash decl` adds `decl` as a `squash` lemma to the cache. -/
+def addSquash (decl : Name)
+    (kind := AttributeKind.global) (prio := eval_prio default) : MetaM Unit := do
+  addSimpTheorem pushCastExt decl (post := true) (inv := false) kind prio
+  addSimpTheorem normCastExt.squash decl (post := true) (inv := false) kind prio
+  addSimpTheorem normCastExt.down decl (post := true) (inv := false) kind prio
+
+/-- `addInfer decl` infers the label of `decl` and adds it to the cache.
```
david-christiansen, line 134:
```suggestion
/-- `addInfer decl` infers the label of `decl` (`elim`, `move`, or `squash`) and arranges for it to
be used by `norm_cast`.
```

src/Init/Data/Cast.lean
david-christiansen: Hah, that's right! Thanks :-) Perhaps it can stop being a coercion, then, once we have this merged?

src/Init/Data/Cast.lean
david-christiansen: I talked this over a bit more with @semorrison, and now I have a bit more insight into the context.

How about something like this revision that includes a bit more info?
```suggestion

This class and `IntCast` exist to allow different libraries with their own types that can be notated as natural numbers to have consistent `simp` normal forms without needing to create coercion simplification sets that are aware of all combinations. Libraries should make it easy to work with `NatCast` where possible. For instance, in Mathlib there will be such a homomorphism (and thus a `NatCast R` instance) whenever `R` is an additive monoid with a `1`.
```

src/Init/Data/Cast.lean
david-christiansen: (my concern here is keeping things as simple as possible - if this would be too simple, no problem, but I think it's worth asking)

src/Lean/Elab/Tactic/NormCast.lean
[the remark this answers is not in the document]
david-christiansen: OK, sounds like it's out of scope for this PR. Do you want to create it, or should I? I'm happy to, but you've got lots of context on it!

src/Init/Data/Cast.lean
semorrison: I certainly appreciate there's a problem using identifiers and filenames from downstream projects (documentation maintenance nightmare!), but that's not really an issue here. I'm pretty happy with David's rewording here.

src/Init/Data/Cast.lean
semorrison: Yes, I'm hoping we can do that. But it will happen after this PR. I've added it to the Std todo list.

src/Init/Data/Cast.lean
semorrison:
```suggestion
It contains just the function, with no axioms. In practice, the target type will likely have a (semi)ring structure, and this homomorphism should be a ring homomorphism.
```

src/Init/Data/Cast.lean
semorrison:
```suggestion
/--
The canonical homomorphism `Int → R`.
In most use cases `R` will have a ring structure and this will be a ring homomorphism.
-/
```

src/Init/Tactics.lean
semorrison, line 909:
```suggestion
Therefore even in situations where non-terminal `simp` calls are discouraged (because of fragility),
`norm_cast` is considered safe.
```

src/Init/Tactics.lean
semorrison: I've asked Rob and Paul-Nicolas.

src/Init/Data/Cast.lean
david-christiansen: @digama0 We certainly don't want referring to Mathlib to be illegal! But it's important that we can provide as consistent an experience as possible and that we can flexibly move forward with important changes, making as efficient as possible use of our limited time. The goal here is not to avoid Mathlib, it's to both explain things in a way that non-Mathlib users can understand, to explain how others might use this part of the language, and to avoid fragile external references that may be broken by refactorings or API changes carried out by others. I think this reworded version addresses all these concerns, and also does a better job explaining why the code is here.

src/Init/Tactics.lean
semorrison: They responded with an arxiv URL, which is even better. We're using that now.

semorrison: and I've moved the links to the module docs.

src/Init/Tactics.lean
```diff
@@ -941,6 +1008,61 @@ If there are several with the same priority, it is uses the "most recent one". E
...
+
+Don't do this unless you understand what you are doing.
+
+A full description of the tactic, and the use of each lemma category, can be found at
+<https://lean-forward.github.io/norm_cast/norm_cast.pdf>.
```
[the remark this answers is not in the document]
semorrison: Per request: https://github.com/leanprover/lean4/issues/3332

src/Lean/Elab/Tactic/NormCast.lean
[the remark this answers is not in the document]
semorrison: @david-christiansen, tracking issue at https://github.com/leanprover/lean4/issues/3333

PR #3323, opened 2024-02-13
fix: `#eval` command was leaking auxiliary declarations into the environment

leodemoura:


nomeata:
This means one cannot use `#eval` anymore to, for example, bring the generated splitter lemmas into scope?

Makes sense, of course; maybe some tests will need another mechanism (`elab` maybe).

leodemoura:
> This means one cannot use #eval anymore to, for example, bring the generated splitter lemmas into scope?

Yes, you still use `#eval <expr>` for that, but the auxiliary declarations created to elaborate `<expr>` will not leak into the environment. The declarations created while executing `<expr>` will be in the resulting environment.

leodemoura:
@nomeata The new test file is a good example of what this PR is fixing: https://github.com/leanprover/lean4/blob/e7cc454161017598c7bf5d472401de54ae90765e/tests/lean/evalLeak.lean

nomeata:
Ah, thanks for the explanation!

PR #3326, opened 2024-02-14
feat: delaboration collapses parent projections 

kmill:
When projection functions are delaborated, intermediate parent projections are no longer printed. For example, rather than pretty printing as `o.toB.toA.x` with these `toB` and `toA` parent projections, it pretty prints as `o.x`.

This feature is being upstreamed from mathlib.

### DIFF
--- a/src/Lean/PrettyPrinter/Delaborator/Builtins.lean
+++ b/src/Lean/PrettyPrinter/Delaborator/Builtins.lean
@@ -798,7 +798,7 @@ Delaborates an application of a projection function, for example `Prod.fst p` as
 Collapses intermediate parent projections, so for example rather than `o.toB.toA.x` it produces `o.x`.
 
 Does not delaborate projection functions from classes, since the instance parameter is implicit;
-we would rather see `default` rather than `instInhabitedNat.default`.
+we would rather see `default` than `instInhabitedNat.default`.
 -/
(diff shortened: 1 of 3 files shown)


src/Lean/PrettyPrinter/Delaborator/Builtins.lean
```diff
@@ -793,23 +793,50 @@ def delabProj : Delab := do
   let idx := Syntax.mkLit fieldIdxKind (toString (idx + 1));
   `($(e).$idx:fieldIdx)
 
-/-- Delaborate a call to a projection function such as `Prod.fst`. -/
+/--
+Delaborates an application of a projection function, for example `Prod.fst p` as `p.fst`.
+Collapses intermediate parent projections, so for example rather than `o.toB.toA.x` it produces `o.x`.
+
+Does not delaborate projection functions from classes, since the instance parameter is implicit;
+we would rather see `default` rather than `instInhabitedNat.default`.
```
semorrison, line 801:
```suggestion
we would rather see `default` than `instInhabitedNat.default`.
```

semorrison (approved):
This has been pretty thoroughly exercised by Mathlib. It will also (hopefully) help me fix some Mathlib tests failing on the nightly-testing branch after all our changes, so I am keen to have this in.

PR #3328, opened 2024-02-14
feat: support idents in auto tactics

digama0:
This is still experimental, but it implements identifier support in auto tactics "in the obvious way". It also converts `quoteAutoTactic` to generate Expr directly instead of going via syntax (this doesn't have any effect other than increasing compile cost AFAICT).

### DIFF
--- a/tests/lean/run/autoparam.lean
+++ b/tests/lean/run/autoparam.lean
@@ -8,4 +8,8 @@ f x y
 def f2 (x y : Nat) (h : x = y := by exact rfl) : Nat :=
 x + x
 
+def f3 (x y : Nat) (h : x = y := by exact Eq.refl x) : Nat :=
+x + x
+
 #check fun x => f2 x x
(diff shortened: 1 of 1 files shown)


semorrison:
What's the status of this PR? It would be good to have a test showing the changed behaviour. It's clear there's enthusiasm for this one, so it would be nice to get in.

tests/lean/run/autoparam.lean
```diff
@@ -4,3 +4,8 @@ x + x
 
 def g (x y z : Nat) (h : x = y) : Nat :=
 f x y
+
+def f2 (x y : Nat) (h : x = y := by exact rfl) : Nat :=
```
eric-wieser, line 8: is `by exact Eq.refl x` legal here too? If so, can you add a test for it?

eric-wieser (approved):
Thanks!

PR #3329, opened 2024-02-14
fix: `savePanelWidgetInfo` on `@[builtin_widget_module]`

Kha:


src/Lean/Widget/UserWidget.lean
```diff
@@ -212,10 +213,12 @@ def erasePanelWidget [Monad m] [MonadEnv m] (h : UInt64) : m Unit := do
 /-- Save the data of a panel widget which will be displayed whenever the text cursor is on `stx`.
 `hash` must be `hash (toModule c).javascript`
 where `c` is some global constant annotated with `@[widget_module]`. -/
-def savePanelWidgetInfo [Monad m] [MonadEnv m] [MonadError m] [MonadInfoTree m]
-    (hash : UInt64) (props : StateM Server.RpcObjectStore Json) (stx : Syntax) : m Unit := do
+def savePanelWidgetInfo (hash : UInt64) (props : StateM Server.RpcObjectStore Json) (stx : Syntax) :
+    CoreM Unit := do
```
Kha, line 217: @Vtec234 Do you see any downside in specializing this (and others?) to `CoreM`? I feel like we sometimes went overboard with monad generalization.

src/Lean/Widget/UserWidget.lean
```diff
@@ -212,10 +213,12 @@ def erasePanelWidget [Monad m] [MonadEnv m] (h : UInt64) : m Unit := do
...
   let env ← getEnv
-  let some (id, _) := moduleRegistry.getState env |>.find? hash
+  let builtins ← builtinModulesRef.get
+  let some id :=
+    (builtins.find? hash |>.map (·.1)) <|> (moduleRegistry.getState env |>.find? hash |>.map (·.1))
     | throwError s!"No widget module with hash {hash} registered"
   pushInfoLeaf <| .ofUserWidgetInfo { id, javascriptHash := hash, props, stx }
```
Kha, line 223: @Vtec234 Am I seeing it correctly that `id` is essentially used only for the deprecated `UserWidgetDefinition`? Is there documentation on why we are using the hash and not `id` for identifying modules?

Vtec234: Technically `id` is still used in `vscode-lean4` to produce a React UUID, but not in any capacity besides that afaicr. The reason why I switched to using the hash is that it is a value accessible at runtime at the object-level, whereas `id`s can only be produced at the meta-level. So if I am given some variable of type `Module`, I can still map it to a hash UUID, but not to a `id`-UUID as it need not be a constant. Of course we could alternatively have added a `id := by exact decl_name%` field to `Module`; this just felt less error-prone, and it will complain if you store the same code twice as a small bonus.

src/Lean/Widget/UserWidget.lean
Vtec234: I don't immediately see any downsides.

Vtec234 (approved):

PR #3335, opened 2024-02-14
chore: upstream Std.Data.List.Init.Basic

kim-em:


semorrison:
I've made a generic tracking issue for ensuring we eventually provide tail-recursive versions of all List functions:

https://github.com/leanprover/lean4/issues/3336

PR #3339, opened 2024-02-15
chore: upstream set notation

joehendrix:
This upstream Std Set notation except for [set literals](https://github.com/leanprover/std4/blob/1b4e6926f03a1f76c16b00c7b6d4891788738a24/Std/Classes/SetNotation.lean#L115-L131).

### DIFF
--- a/tests/lean/run/macro.lean
+++ b/tests/lean/run/macro.lean
@@ -1,3 +1,5 @@
+namespace Test
+
 abbrev Set (α : Type) := α → Prop
 axiom setOf {α : Type} : (α → Prop) → Set α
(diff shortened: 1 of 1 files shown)


digama0:
> This upstream Std Set notation except for [set literals](https://github.com/leanprover/std4/blob/1b4e6926f03a1f76c16b00c7b6d4891788738a24/Std/Classes/SetNotation.lean#L115-L131).

I think it might be particularly useful to upstream set literals, because it's easier to make them cooperate with structure instance notation if they are defined together. But it's complicated enough that it would be good to have in a separate PR anyway.

PR #3340, opened 2024-02-15
chore: upstream Std.Classes.LawfulMonad (except SatisfiesM)

kim-em:


src/Init/Data/List.lean
```diff
@@ -7,3 +7,4 @@ prelude
 import Init.Data.List.Basic
 import Init.Data.List.BasicAux
 import Init.Data.List.Control
+import Init.Data.List.Lemmas
```
semorrison, line 10:
```suggestion
```

PR #3347, opened 2024-02-15
chore: upstream NatCast and IntCast

joehendrix:
This upstreams NatCast and IntCast alone independent of norm_cast in #3322.

This will allow more efficiently upstreaming parts of Std.Data.Int relevant for omega.


joehendrix:
!bench

src/Lean/Server/Utils.lean
```diff
@@ -117,7 +117,7 @@ def publishDiagnostics (m : DocumentMeta) (diagnostics : Array Lsp.Diagnostic) (
     method := "textDocument/publishDiagnostics"
     param  := {
       uri         := m.uri
-      version?    := m.version
+      version?    := .some m.version
```
joehendrix, line 120:
```suggestion
      version?    := some m.version
```

PR #3348, opened 2024-02-15
chore: set literal notation

leodemoura:


src/Lean/Server/Rpc/Deriving.lean
```diff
@@ -57,7 +57,7 @@ private def deriveStructureInstance (indVal : InductiveVal) (params : Array Expr
       enc a := return toJson { $[$encInits],* : RpcEncodablePacket }
       dec j := do
         let a : RpcEncodablePacket ← fromJson? j
-        return { $[$decInits],* }
+        return $(⟨structInst.raw⟩)
```
Kha, line 60:
```suggestion
        return { $decInits:structInstField,* }
```
should also work I believe

Kha (approved):

PR #3367, opened 2024-02-16
chore: upstream omega

kim-em:


src/Init/Omega/Int.lean
```diff
@@ -0,0 +1,174 @@
...
+prelude
+import Init.Data.Int.Order
+
+/-!
+# Lemmas about `Nat` and `Int` needed internally by `omega`.
+
+These statements are useful for constructing proof expressions,
+but unlikely to be widely useful, so are inside the `Std.Tactic.Omega` namespace.
+
+If you do find a use for them, please move them into the appropriate file and namespace!
```
eric-wieser, line 15: A lot of these lemmas already exist downstream in mathlib / std, right? Can you add a sentence indicating that for many users, they can find the lemmas there?

src/Lean/Elab/Tactic/Omega/MinNatAbs.lean
```diff
@@ -0,0 +1,139 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Scott Morrison
+-/
+
+/-!
+# `List.nonzeroMinimum`, `List.minNatAbs`, `List.maxNatAbs`
+
+`List.minNatAbs` computes the minimum non-zero absolute value of a `List Int`.
+This is not generally useful outside of the implementation of the `omega` tactic,
+so we keep it in the `Std/Tactic/Omega` directory
```
eric-wieser, line 12: Presumably this should no longer refer to `Std/` in this way (I haven't checked any other files for Std references)

src/Init/Omega/Int.lean
semorrison: I don't think they do, actually. I initially wrote these when `omega` lived in Mathlib! (on a branch)

PR #3370, opened 2024-02-16
feat: infoview.maxTraceChildren

Kha:
Incrementally unveil trace children for excessively large nodes to improve infoview rendering time, adjust particularly chatty `simp.ground` trace to make use of it.

### DIFF
--- a/src/Lean/Widget/InteractiveDiagnostic.lean
+++ b/src/Lean/Widget/InteractiveDiagnostic.lean
@@ -91,6 +91,15 @@ private inductive EmbedFmt
 
 private abbrev MsgFmtM := StateT (Array EmbedFmt) IO
 
+/--
+Number of trace node children to display by default in the info view in order to prevent slowdowns
+from rendering.
+-/
+register_option infoview.maxTraceChildren : Nat := {
+  defValue := 50
+  descr := "Number of trace node children to display by default"
+}
+
 open MessageData in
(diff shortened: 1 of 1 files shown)


src/Lean/Meta/Tactic/Simp/Rewrite.lean
```diff
@@ -387,9 +387,10 @@ def simpGround : Simproc := fun e => do
   if ctx.simpTheorems.isErased (.decl declName) then return .continue
   -- Matcher applications should have been reduced before we get here.
   if (← isMatcher declName) then return .continue
-  trace[Meta.Tactic.Simp.ground] "seval: {e}"
-  let r ← seval e
-  trace[Meta.Tactic.Simp.ground] "seval result: {e} => {r.expr}"
+  let r ← withTraceNode `Meta.Tactic.simp.ground (fun
```
Kha, line 390: @leodemoura Are you happy with this nesting adjustment? This PR only affects the display of nested traces as on the top level they can be mixed with other messages, which would effectively require a completely separate implementation.

Kha: In action after fixing the `withTraceNode` bug: 
image

leodemoura: Yes.

leodemoura (approved):

PR #3372, opened 2024-02-16
fix: liasolver benchmark bug introduced by #3364

joehendrix:
This fixes a rounded division/mod bug introduced by the change in semantics from Int.div to Int.mod in #3364.

### DIFF
--- a/tests/bench/liasolver.lean
+++ b/tests/bench/liasolver.lean
@@ -14,13 +14,16 @@ namespace Int
     if b = 0 then
       return 0
     let mut div := a / b
-    let rest := a / b
+    let rest := a % b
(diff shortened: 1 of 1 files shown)


joehendrix:
!bench

joehendrix:
I want to do a bit more testing to confirm the semantics is preserved by this change.

joehendrix:
!bench

joehendrix:
To validate that I haven't changed things, I ran tests confirming the new and old roundedDiv function haven't diverged on positive and negative numerator and divisor.  There is a special case that affects divide by zero, but that's essentially a bug fix.

PR #3374, opened 2024-02-16
perf: add prelude to all Lean modules

hargoniX:
As discussed with @joehendrix and @Kha this removes the automatic `Init` import in all modules in the `Lean` namespace.

This change was created in 3 steps:
1. Running mathlib's shake tool on the `Lean` modules and applying its sugggestions. Note that I configured it to not care about overeagerly imported modules from `Lean` itself so this change remains somewhat atomic.
2. After shake did not find things anymore I started adding `prelude` to all remaining modules and imported `Init` modules that fixed errors if necessary. I'm not 100% confident that I always imported the least modules necessary but everything is working properly now.
3. A single test that inspects the declaration range of a function failed as its declaration range changed -> fixed by copying the new range.

To confirm that it has the desired effect I edited `Init.lean` and reran the build. This only resulted in a rebuild of `Lake` with no `Lean` module being affected anymore.

Should we document that we now require modules in Lean to work with `prelude` and a minimal(ish) set of imports from `Init` somewhere?


Kha:
!bench

leodemoura:
> Should we document that we now require modules in Lean to work with prelude and a minimal(ish) set of imports from Init somewhere?

Yes.

Kha:
> Should we document that we now require modules in Lean to work with prelude and a minimal(ish) set of imports from Init somewhere?

I don't know where a good place for this would be, I hope the copying-from-existing-modules people should already be doing for the copyright header will take care of that. If we're concerned about people forgetting, a CI check against importing `Init` in `Lean` should not be too difficult.

PR #3375, opened 2024-02-16
feat: apply app unexpanders for all prefixes of an application

kmill:
Before, app unexpanders would only be applied to entire applications. However, some notations produce functions, and these functions can be given additional arguments. The solution so far has been to write app unexpanders so that they can take an arbitrary number of additional arguments. However, as reported in [this Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/pretty.20printer.20bug/near/420662236), this leads to misleading hover information in the Infoview. For example, while `HAdd.hAdd f g 1` pretty prints as `(f + g) 1`, hovering over `f + g` shows `f`. There is no way to fix the situation from within an app unexpander; the expression position for `HAdd.hAdd f g` is absent, and app unexpanders cannot register TermInfo.

This commit changes the app delaborator to try running app unexpanders on every prefix of an application, from longest to shortest prefix. For efficiency, it is careful to only try this when app delaborators do in fact exist for the head constant, and it also ensures arguments are only delaborated once. Then, in `(f + g) 1`, the `f + g` gets TermInfo registered for that subexpression, making it properly hoverable.

The app delaborator is also refactored, and there are some bug fixes:
- app unexpanders only run when `pp.explicit` is false
- trailing parameters in under-applied applications are now only considered up to reducible & instance transparency, which lets, for example, optional arguments for `IO`-valued functions to be omitted. (`IO` is a reader monad, so it's hiding a pi type)
- app unexpanders will no longer run for delaborators that use `withOverApp`
- auto parameters now always pretty print, since we are not verifying that the provided argument equals the result of evaluating the tactic

Furthermore, the `notation` command has been modified to generate an app unexpander that relies on the app delaborator's new behavior.

The change to app unexpanders is reverse-compatible, but it's recommended to update `@[app_unexpander]`s in downstream projects so that they no longer handle overapplication themselves.

### DIFF
--- a/tests/lean/301.lean.expected.out
+++ b/tests/lean/301.lean.expected.out
@@ -1,10 +1,7 @@
 301.lean:1:9-1:17: error: missing cases:
 (Nat.succ _)
 301.lean:1:21-1:24: error: type mismatch
-  «»
-    («» fun x =>
-      match x with
-      | 0 => 0)
+  «» «»
 has type
   Nat → Nat : Type
 but is expected to have type


tests/lean/lcnfTypes.lean
```diff
@@ -119,6 +119,8 @@ def testMono (declName : Name) : MetaM Unit := do
   let mono ← LCNF.toMonoType base
   IO.println s!"{declName} : {← ppExpr mono}"
 
+-- Universe levels are erased, and when the delaborator sees type incorrect applications it always inserts a `@`.
+-- This is why one sees `@List Ty` rather than `List Ty` in the following tests.
```
Kha, line 123: Is there anything we could do against this?

Kha: Ah, there is an explicit `pp.explicit` in the test, so I suppose this regression is expected

kmill: It turns out that `match` missing pattern errors also lack universe levels, so this helps with that as well.

PR #3377, opened 2024-02-16
chore: make server completion predicate not private

joehendrix:
This renames and exports a predicate used in the server implementation for deciding if a name is suitable for auto-completion.

I'd like `exact?` and related tactics that suggest names to use the same logic as the completion since the goal seems to be the same -- neither VSCode nor those tactics should suggest results to users that shouldn't be used directly in code.


nomeata:
Looking forward to using that in Loogle as well

src/Lean/Meta/CompletionName.lean
```diff
@@ -0,0 +1,47 @@
...
+
+This exports a predicate for checking whether a name should be made visible
+in auto-completion and other tactics.  It uses a blacklist environment extension
+to enable names in an environment to be specifically hidden.
+-/
+namespace Lean.Server.Completion
+
+builtin_initialize completionBlackListExt : TagDeclarationExtension ← mkTagDeclarationExtension
+
+end Lean.Server.Completion
```
joehendrix, line 20: I have to keep this extension in `Lean.Server.Completion` for tests to pass, but I thought that through an educated guess.  How could one move the namespace of this extension?

src/Lean/Meta/CompletionName.lean
```diff
@@ -0,0 +1,41 @@
+/-
```
mhuisi, line 1: I feel like this file name doesn't accurately describe what's in it. If this is to live in Lean.Meta, it should have a more general name than "completion" or "blacklist w.r.t. completion". Unfortunately, I can't think of anything better than something like "UserFacingName", either.

PR #3380, opened 2024-02-17
feat: make loose fvars pretty print as `_fvar.123` instead of `_uniq.123`

kmill:
Loose fvars are never supposed to be pretty printed, but having them print with "fvar" in the name can help with debugging broken tactics and elaborators.

Metaprogramming users often do not realize at first that `_uniq.???` in pretty printing output refers to fvars not in the current local context.

### DIFF
--- a/src/Lean/PrettyPrinter/Delaborator/Builtins.lean
+++ b/src/Lean/PrettyPrinter/Delaborator/Builtins.lean
@@ -24,13 +24,13 @@ def unfoldMDatas : Expr → Expr
 
 @[builtin_delab fvar]
 def delabFVar : Delab := do
-let Expr.fvar fvarId ← getExpr | unreachable!
-try
-  let l ← fvarId.getDecl
-  maybeAddBlockImplicit (mkIdent l.userName)
-catch _ =>
-  -- loose free variable, use internal name
-  maybeAddBlockImplicit <| mkIdent fvarId.name
+  let Expr.fvar fvarId ← getExpr | unreachable!
+  try
+    let l ← fvarId.getDecl
+    maybeAddBlockImplicit (mkIdent l.userName)
+  catch _ =>
+    -- loose free variable, use internal name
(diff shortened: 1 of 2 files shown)


kmill:
Start of a Zulip discussion: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.E2.9C.94.20Using.20information.20from.20.60LocalContext.60.20to.20remove.20.60_uniq.60.2E/near/422031285

src/Lean/PrettyPrinter/Delaborator/Builtins.lean
```diff
@@ -24,13 +24,13 @@ def unfoldMDatas : Expr → Expr
...
-catch _ =>
-  -- loose free variable, use internal name
-  maybeAddBlockImplicit <| mkIdent fvarId.name
+  let Expr.fvar fvarId ← getExpr | unreachable!
+  try
+    let l ← fvarId.getDecl
+    maybeAddBlockImplicit (mkIdent l.userName)
+  catch _ =>
+    -- loose free variable, use internal name
+    maybeAddBlockImplicit <| mkIdent (fvarId.name.replacePrefix `_uniq `_fvar)
```
nomeata, line 33: Would it make sense to generate the name as `_fvar` in the first place? Or are there unique names around that aren't free variables?

kmill: I looked into that briefly while creating the PR, and `_uniq` is coming from the name generator configuration. That generator is used directly by `Lean.mkFreshFVarId` and `Lean.mkFreshMVarId`. The `mvar` delaborator also replaces `_uniq` like this.

It would be nice if the generated names could say `fvar` and `mvar` for debugging purposes.

nomeata: ok, if there is precedent for this replacement idiom, then all is well

PR #3383, opened 2024-02-18
fix: exposeRelevantUniverses was inserting the whole expression into itself

kmill:
This bug appeared in c9db8619f114769e4a741caf22ca8442a04a024d


kmill:
This was discovered while working on #3375

leodemoura (approved):

PR #3398, opened 2024-02-19
fix: occurence check in `mkInjectiveTheoremTypeCore?`

### LINKED ISSUE #3386: `genInjectivity` failure
### Prerequisites

### Description

In some cases, Lean fails to generate injectivity lemmas for inductive types

### Context

The issue was first discovered in [lean-reflection](https://github.com/Kiiyya/lean-reflection/blob/6846a0670183ecc523ca6df698475bfdfb85be95/Reflection/Mutual.lean#L52)

### Steps to Reproduce
MWE:
```lean
inductive Tyₛ : Type (u+1)
| SPi : (T : Type u) -> (T -> Tyₛ) -> Tyₛ

inductive Tmₛ.{u} :  Tyₛ.{u} -> Type (u+1)
| app : Tmₛ (.SPi T A) -> (arg : T) -> Tmₛ (A arg)
```

**Expected behavior:** No error

**Actual behavior:** Error:
```
application type mismatch
  Tyₛ.SPi T A
argument has type
  T✝ → Tyₛ
but function has type
  (T → Tyₛ) → Tyₛ
```

### Versions

4.6.0-rc1
Web

### Additional Information

The generated injectivity lemma looks like this:
```lean
theorem Tmₛ.app.inj {T : Type u} {A : T → Tyₛ} {a : Tmₛ (Tyₛ.SPi T A)} {arg : T} {T_1 : Type u} {a_1 : Tmₛ (Tyₛ.SPi T_1 A)} :
  Tmₛ.app a arg = Tmₛ.app a_1 arg →
    T = T_1 ∧ HEq a a_1 := fun x => Tmₛ.noConfusion x fun T_eq A_eq a_eq arg_eq => eq_of_heq a_eq 
```

The error happens because T is duplicated where it shouldn't be. In order for the injectivity lemma to be as general as possible, Lean duplicates any argument which doesn't occur in the end type of the constructor (here, in `Tmₛ (A arg)`). However, this is not precise enough, as some terms of that type may still depend on the variable: here, `T` doesn't appear in that type, but appears in the type of `A`. A solution would be to also check the occurence of variables inside the types of free variables appearing in the end type.

### Impact

arthur-adjedj:
Closes #3386

Currently, when generating the signature of an injectivity lemma for a certain constructor `c : forall xs, Foo a_1 ... a_n`, `mkInjectiveTheoremTypeCore?` will differentiate between variables which are bound to stay the same between the two equal values (i.e inductive indices), and non-fixed ones. To do that, the function currently checks whether a variable `x ∈ xs` appears in the final co-domain `Foo a_1 ... a_n` of the constructor. This condition isn't enough however. As shown in the linked issue, the codomain may also depend on variables which appears in the type of free vars contained in `Foo a_1 ... a_n`, but not in the term itself. This PR fixes the issue by also checking the types of any free variable occuring in the final codomain, so as to ensure injectivity lemmas are well-typed.

### DIFF
--- a/src/Lean/Meta/Injective.lean
+++ b/src/Lean/Meta/Injective.lean
@@ -133,7 +133,6 @@ private def mkInjectiveTheorem (ctorVal : ConstructorVal) : MetaM Unit := do
     | return ()
   let value ← mkInjectiveTheoremValue ctorVal.name type
   let name := mkInjectiveTheoremNameFor ctorVal.name
-  trace[Meta.injective] "theorem {name} : {type} := {value}"
   addDecl <| Declaration.thmDecl {
     name
     levelParams := ctorVal.levelParams
@@ -165,7 +164,6 @@ private def mkInjectiveEqTheorem (ctorVal : ConstructorVal) : MetaM Unit := do
     | return ()
   let value ← mkInjectiveEqTheoremValue ctorVal.name type
   let name := mkInjectiveEqTheoremNameFor ctorVal.name
-  trace[Meta.injective] "theorem {name} : {type} := {value}"
   addDecl <| Declaration.thmDecl {
     name
     levelParams := ctorVal.levelParams


arthur-adjedj:
awaiting-review

src/Lean/Util/FindExpr.lean
```diff
@@ -56,6 +56,18 @@ def find? (p : Expr → Bool) (e : Expr) : Option Expr :=
 def occurs (e : Expr) (t : Expr) : Bool :=
   (t.find? fun s => s == e).isSome
 
+def findM? [Monad m] (p : Expr → m Bool) (e : Expr) : m (Option Expr) := do
```
semorrison, line 59: I'm hesitant to merge this as I'm not sure whether this should also receive a `PtrSet`-backed efficient implementation.

semorrison: Once #3970 lands, could you update this PR and ping me again?

arthur-adjedj: @semorrison The PR is now rebased to `master`.

semorrison:
!bench

PR #3413, opened 2024-02-20
feat: custom error recovery in parser

david-christiansen:
Adds a simple error-recovery mechanism to Lean's parser, similar to those used in other combinator parsing libraries.

Lean itself isn't very amenable to error recovery with this mechanism, as it requires global knowledge of the grammar in question to write recovery rules that don't break backtracking or `<|>`. I only found a few opportunities.

But for DSLs, this is really important. In particular, Verso parse errors interacted very badly with Lean parse errors in a way that required frequent "restart file" commands, but this mechanism allows me to both recover from Verso parse errors and to have Lean skip the rest of the file rather than repeatedly trying to parse it as Lean commands.

### DIFF
--- a/src/Lean/Parser/Module.lean
+++ b/src/Lean/Parser/Module.lean
@@ -63,9 +63,9 @@ private partial def mkErrorMessage (c : InputContext) (pos : String.Pos) (stk : 
 where
   -- Error recovery might lead to there being some "junk" on the stack
   lastTrailing (s : SyntaxStack) : Option Substring :=
-    if s.size < 1 then none
-    else if let .original (trailing := trailing) .. := s.back.getTailInfo then some trailing
-    else lastTrailing s.pop
+    s.toSubarray.findSomeRevM? (m := Id) fun stx =>
+      if let .original (trailing := trailing) .. := stx.getTailInfo then pure (some trailing)
+        else none
 
 def parseHeader (inputCtx : InputContext) : IO (Syntax × ModuleParserState × MessageLog) := do
   let dummyEnv ← mkEmptyEnvironment


david-christiansen:
!bench

src/Lean/Parser/Command.lean
```diff
@@ -67,7 +81,7 @@ def declModifiers (inline : Bool) := leading_parser
   optional («partial» <|> «nonrec»)
 /-- `declId` matches `foo` or `foo.{u,v}`: an identifier possibly followed by a list of universe names -/
 def declId           := leading_parser
-  ident >> optional (".{" >> sepBy1 ident ", " >> "}")
+  ident >> optional (".{" >> sepBy1 (recover ident (recoverUntil (fun c => c.isWhitespace || c ∈ [',', '}']))) ", " >> "}")
```
Kha, line 84: I find it a bit confusing that both of these have `recover` in their name but different roles

src/Lean/Parser/Command.lean
```diff
@@ -17,6 +17,20 @@ namespace Parser
...
+private def recoverUntil (pred : Char → Bool) : Parser where
+  fn :=
+    andthenFn
+      (takeUntilFn pred)
+      (takeWhileFn Char.isWhitespace)
+
+/-- Skip input until the next whitespace character (used for error recovery for certain tokens) -/
+private def recoverUntilWs : Parser := recoverUntil Char.isWhitespace
+
+/-- Skip input until the next whitespace character (used for error recovery for certain tokens) -/
```
Kha, line 30: C&P docstring

src/Lean/Parser/Module.lean
```diff
@@ -52,14 +52,20 @@ private def mkErrorMessage (c : InputContext) (s : ParserState) (e : Parser.Erro
...
     pos := c.fileMap.toPosition pos
     endPos := c.fileMap.toPosition <$> endPos?
     keepFullRange := true
     data := toString e }
+where
+  -- Error recovery might lead to there being some "junk" on the stack
+  lastTrailing (s : SyntaxStack) : Option Substring :=
+    if s.size < 1 then none
+    else if let .original (trailing := trailing) .. := s.back.getTailInfo then some trailing
+    else lastTrailing s.pop
```
Kha, line 68: I think this is `s.toSubarray.findRev? (fun | .original (trailing := trailing) .. => some trailing | _ => none)`?

src/Lean/Parser/Module.lean
```diff
@@ -69,7 +75,7 @@ def parseHeader (inputCtx : InputContext) : IO (Syntax × ModuleParserState × M
   let stx := if s.stxStack.isEmpty then .missing else s.stxStack.back
   match s.errorMsg with
   | some errorMsg =>
-    let msg := mkErrorMessage inputCtx s errorMsg
+    let msg := mkErrorMessage inputCtx s.pos s.stxStack errorMsg
     pure (stx, { pos := s.pos, recovering := true }, { : MessageLog }.add msg)
```
Kha, line 79: Let's make sure we don't ignore any recovery errors here even if we don't have any right now.

src/Lean/Parser/Module.lean
```diff
@@ -103,6 +109,9 @@ partial def parseCommand (inputCtx : InputContext) (pmctx : ParserModuleContext)
     let pos' := pos
     let p := andthenFn whitespace topLevelCommandParserFn
     let s := p.run inputCtx pmctx (getTokenTable pmctx.env) { cache := initCacheForInput inputCtx.input, pos }
+    -- save errors from sub-recoveries
+    for (rpos, rstk, recovered) in s.recoveredErrors do
+      messages := messages.add <| mkErrorMessage inputCtx rpos rstk recovered
```
Kha, line 114: Should this be guarded by `unless recovering && ignore` as well?

tests/lean/parserRecovery.lean
```diff
@@ -0,0 +1,88 @@
+import Lean
+
```
Kha, line 2: Should have a module docstring; https://github.com/leanprover/lean4/blob/master/doc/dev/testing.md#writing-good-tests

Kha (commented):
The parser framework changes look great. For the grammar changes it would be nice to have a less manual approach as discussed but this is a good start.

tests/lean/parserRecovery.lean
david-christiansen: You're right!

src/Lean/Parser/Module.lean
david-christiansen: No, because the array contains syntax rather than sourceinfo. It would work to map `getTailInfo` over it first, but it seems wasteful in a strict language.

This does work:
```
    s.toSubarray.findSomeRevM? (m := Id) fun stx =>
      if let .original (trailing := trailing) .. := stx.getTailInfo then pure (some trailing)
        else none
```

src/Lean/Parser/Command.lean
david-christiansen: Good point! Fixed.

src/Lean/Parser/Command.lean
david-christiansen: Fixed - thanks

src/Lean/Parser/Module.lean
david-christiansen: I don't think so. The idea here is that if any errors were recovered from while parsing the command, they need to end up in the message log. But `recovering` says that the parser had failed to parse the command, and it's attempting to resynchronize. The guarded case below is there to ignore errors that occur as a part of the outer loop's recovery process, but this code is intended to save errors that occurred as part of the "ordinary" us of the parser.

Does that make sense?

src/Lean/Parser/Module.lean
david-christiansen: I just pushed it. I'm not sure that the code is any easier to read, but de gustibus non disputandum est :)

src/Lean/Parser/Module.lean
david-christiansen: Should be good now!

src/Lean/Parser/Module.lean
Kha: I'm not sure I follow, we don't know whether we're in the "ordinary" case unless we use that specific conditional. But it seems unlikely that `getPos?` would be empty if we did recover from errors during the parse so in practice it should be okay.

Kha (approved):

src/Lean/Parser/Module.lean
david-christiansen: I mean, passing the tests and seeming to work is necessary, but not sufficient. There could totally be an obscure bug if we don't understand why this is where it is, so it's worth hashing out.

The way it's written right now, every invocation of the top-level command parser may end up in the following states:
1. No errors at all - then the `for` executes 0 times, no errors to report. ✅ 
2. A successful parse, but errors were recovered from - these belong in the message log, and the elaborator will bump into some `missing`s, but we don't need to go into recovery mode at the top level.
3. An unsuccessful parse, whether or not errors were recovered - in this case the recovered errors need reporting, and the outer loop needs to go into recovery mode

I think this is what's going on here. But it sounds like we're talking past each other, or one or both of us has misunderstood something. Can you say more about why it should perhaps be guarded?

Kha: In the end, we recover from every parse error, so "recovered errors" is a slightly confusing description to me - I would expect either all or none of `s.recoveredErrors` and `s.errorMsg` to be reported for a single command parser invocation, which is not the case in your current code. But as I said, I can't think of a practical issue here - it is hard to imagine that we would ever want to use `recover` without already having consumed at least one token, which should lead to `ignore` being false. So I don't have a clear recommendation except "looked potentially inconsistent with existing code to me at first glance".

PR #3414, opened 2024-02-20
fix: improve error message when termination argument is too dependent

### LINKED ISSUE #2260: `termination_by` fails with dependent matching
The well-founded recursion module gives a surprising error with some dependent pattern matches:
```lean
def Option.attach (o : Option α) : Option { x : α // o = x } :=
  match o with
  | some x => some ⟨x, rfl⟩
  | none => none

def foo (xs : List α) : Option (List α) := do
  if h : xs.length < 10 then
    pure (xs.take 2)
  else
    have : sizeOf (xs.drop 5) < sizeOf xs := sorry
    let ⟨res, h⟩ ← .attach (foo (xs.drop 5))
    pure (res ++ res)
termination_by foo => xs

/-
type mismatch
  xs
has type
  List α : Type u_1
but is expected to have type
  ?β : Sort ?u.938
-/
```

Weirdly enough, it complains that the termination measure does not type check.  The crux is the `let ⟨res, h⟩ ← .attach (foo (xs.drop 5))` line; it works just fine with `let res ← .attach (foo (xs.drop 5))` or `let res ← foo (xs.drop 5)`.

nomeata:
this may help users when they face #2260

fixes #2260

### DIFF
--- a/src/Lean/Elab/PreDefinition/WF/Rel.lean
+++ b/src/Lean/Elab/PreDefinition/WF/Rel.lean
@@ -68,7 +68,7 @@ def elabWFRel (preDefs : Array PreDefinition) (unaryPreDefName : Name) (fixedPre
       let mvarId ← unpackUnary preDef fixedPrefixSize mvarId d element
       mvarId.withContext do
         let errorMsgHeader? := if preDefs.size > 1 then
-          "The termination argument types differ for the different functions, or depends on the " ++
+          "The termination argument types differ for the different functions, or depend on the " ++
           "function's varying parameters. Try using `sizeOf` explicitly:\nThe termination argument"
         else
           "The termination argument depends on the function's varying parameters. Try using " ++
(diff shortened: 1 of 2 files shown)


src/Lean/Elab/PreDefinition/WF/Rel.lean
```diff
@@ -67,7 +67,14 @@ def elabWFRel (preDefs : Array PreDefinition) (unaryPreDefName : Name) (fixedPre
     for (d, mvarId) in subgoals, element in wf, preDef in preDefs do
       let mvarId ← unpackUnary preDef fixedPrefixSize mvarId d element
       mvarId.withContext do
+        let errorMsgHeader? := if preDefs.size > 1 then
+          "The termination argument types differ for the different functions, or depends on the " ++
```
david-christiansen, line 71: Subject-verb agreement:
```suggestion
          "The termination argument types differ for the different functions, or depend on the " ++
```

nomeata: So I had it right initially (https://github.com/leanprover/lean4/pull/3414/commits/bddb56d69bebee3ad6fadcf55942b058d89ed8a1) 🤦‍♂️

PR #3419, opened 2024-02-20
chore: upstream norm_cast attributes and tests

kim-em:
Unfortunately, the test file is broken, and so far I can't work out why.

Standalone repro (which maybe shouldn't work on master, because of missing attributes, but I hope will work here):
```
example (an bn : Nat) : (an : Int) ≠ (bn : Int) ↔ an ≠ bn := by norm_cast
```
is failing with unsolved goals.

So far my attempt to understand what has changed is running
```
import Std.Tactic.NormCast
import Std.Tactic.ShowTerm
import Std.Data.Int.Init.Lemmas

example (an : Nat) : (an : Int) ≠ 0 ↔ an ≠ 0 := by show_term norm_cast
```
on Std's `main` branch. This prints
```exact Eq.mpr (id (congrArg (fun x => ¬x ↔ ¬an = 0) Std.Data.Int.Init.Lemmas._auxLemma.3)) Iff.rfl```
which is unfortunate: the trailing `.3` in the name makes it hard to see what this lemma is. (I know @kmill? had a tool for inspecting these numeric named declarations, but I can't find it.)

At least this gives a clue that it is a lemma in Std.Data.Init.Init.Lemmas that is missing. However I'd verified that the 5 `@[norm_cast]` attributes there are faithfully reproduced in this PR.

### DIFF
--- a/src/Init/SimpLemmas.lean
+++ b/src/Init/SimpLemmas.lean
@@ -84,6 +84,7 @@ theorem dite_congr {_ : Decidable b} [Decidable c]
   | inr h => rw [dif_neg h]; subst b; rw [dif_neg h]; exact h₃ h
 
 @[simp] theorem ne_eq (a b : α) : (a ≠ b) = ¬(a = b) := rfl
+norm_cast_add_elim ne_eq
 @[simp] theorem ite_true (a b : α) : (if True then a else b) = a := rfl
 @[simp] theorem ite_false (a b : α) : (if False then a else b) = b := rfl
 @[simp] theorem dite_true {α : Sort u} {t : True → α} {e : ¬ True → α} : (dite True t e) = t True.intro := rfl

### CODE REVIEW

**Comment by kmill** [comment]
@semorrison Here might be what I shared before:
```lean
import Lean

open Lean Meta Elab

/-- Whether the lemma has a name of the form produced by `Lean.Meta.mkAuxLemma`. -/
def Lean.Name.isAuxLemma (n : Name) : Bool := n matches .num (.str _ "_auxLemma") _

/-- Unfold all lemmas created by `Lean.Meta.mkAuxLemma`. These end in `_auxLemma.nn` where `nn` is a number. -/
def Lean.Meta.unfoldAuxLemmas (e : Expr) : MetaM Expr := do
  deltaExpand e Lean.Name.isAuxLemma
```

PR #3421, opened 2024-02-20
fix: split `libInit_shared` out of `libleanshared`

Kha:
Avoids hitting the Windows limit on symbols per shared library soon


Kha:
!bench

Kha:
!bench

PR #3422, opened 2024-02-20
feat: improve error messages and docstring for `decide` tactic

kmill:
The `decide` tactic produces error messages that users find to be obscure. Now:
1. If the `Decidable` instance reduces to `isFalse`, it reports that `decide` failed because the proposition is false.
2. If the `Decidable` instance fails to reduce, it explains what proposition it failed for, and it shows the reduced `Decidable` instance rather than the `Decidable.decide` expression. That expression tends to be less useful since it shows the unreduced `Decidable` argument (plus it's a lot longer!)

Examples:
```lean
example : 1 ≠ 1 := by decide
/-
tactic 'decide' proved that the proposition
  1 ≠ 1
is false
-/

opaque unknownProp : Prop

open scoped Classical in
example : unknownProp := by decide
/-
tactic 'decide' failed for proposition
  unknownProp
since its 'Decidable' instance reduced to
  Classical.choice ⋯
rather than to the 'isTrue' constructor.
-/
```

When reporting the error, `decide` only shows the whnf of the `Decidable` instance. In the future we could consider having it reduce all decidable instances present in the term, which can help with determining the cause of failure (this was explored in 8cede580690faa5ce18683f168838b08b372bacb).

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -81,6 +81,8 @@ v4.7.0 (development in progress)
   In both cases, `h` is applicable because `simp` does not index f-arguments anymore when adding `h` to the `simp`-set.
   It's important to note, however, that global theorems continue to be indexed in the usual manner.
 
+* Improved the error messages produced by the `decide` tactic. [#3422](https://github.com/leanprover/lean4/pull/3422)
+
 Breaking changes:
 * `Lean.withTraceNode` and variants got a stronger `MonadAlwaysExcept` assumption to
   fix trace trees not being built on elaboration runtime exceptions. Instances for most elaboration


kmill:
@david-christiansen The `decide` tactic has had a confusing error message when it fails to prove the goal, and here's an attempt to improve the situation, and the documentation as well while we're here.

There was [recent Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/custom.20messages.20for.20tactic.20failure/near/422433922) about `decide`. There, there was a request to make `decide` synthesize a reason for the failure, but that doesn't seem to be possible to do in a satisfying amount of generality, unless we had a parallel version of `Decidable` whose constructors contained a reason for the success or failure. This is an idea I had that [made it into LSpec](https://github.com/lurk-lab/LSpec/blob/1d3536f96e964bd57405f18c41c2adbaf4f257e0/LSpec/LSpec.lean#L28-L32).

src/Lean/Elab/Tactic/ElabTerm.lean
```diff
@@ -372,10 +372,23 @@ private def preprocessPropToDecide (expectedType : Expr) : TermElabM Expr := do
...
+      throwError "\
+        tactic 'decide' proved that the proposition\
+        {indentExpr expectedType}\n\
+        is false"
+    unless r.isAppOf ``isTrue do
+      throwError "\
+        tactic 'decide' failed for proposition\
+        {indentExpr expectedType}\n\
+        since its 'Decidable' instance did not reduce to the 'isTrue' constructor:\
+        {indentExpr r}"
```
david-christiansen, line 389: I find this phrasing ever so slightly confusing - what about this?
```suggestion
        since its 'Decidable' instance did not reduce to the 'isTrue' constructor.\n\
        It reduced to:\
        {indentExpr r}"
```
Otherwise it sounds to me like the expression is the constructor, instead of what we got instead.

src/Lean/Parser/Tactic.lean
```diff
@@ -66,13 +66,55 @@ doing a pattern match. This is equivalent to `fun` with match arms in term mode.
...
-by well founded recursion, since this requires reducing proofs.
-```
+/--
+`decide` attempts to prove a goal of type `p` by synthesizing an instance
+of `Decidable p` and then reducing that instance. If it reduces to `isTrue h`,
+then `h` is used to close the goal.
+
+Limitations:
+- The goal is not allowed to have local variables or metavariables.
+  If there are local variables, you can try to `revert` them first.
```
david-christiansen, line 76: Slight pedantic wording thing:
```suggestion
- The goal is not allowed to contain local variables or metavariables.
  If there are local variables, you can try to `revert` them first.
```

Do we expect there to be enough understanding of `revert` in the audience for this? Or should it be a bit more explanatory, like:
```suggestion
- The goal is not allowed to refer to local variables or contain metavariables.
  The `revert` tactic can transform local variable references into bindings in the
  goal, which may allow `decide` to succeed.
```

src/Lean/Parser/Tactic.lean
```diff
@@ -66,13 +66,55 @@ doing a pattern match. This is equivalent to `fun` with match arms in term mode.
...
+- Because this uses kernel reduction to evaluate the term, it may not work in the
+  presence of definitions by well founded recursion, since this requires reducing proofs.
```
david-christiansen, line 78:
```suggestion
- Because this uses kernel reduction to evaluate the term, `Decidable` instances defined
   by well-founded recursion may not work, because evaluating them requires reducing proofs.
```

david-christiansen:
This looks like a nice improvement!

I think we should look into adding reasons as optional arguments. If we do it right, we might not even break anything :-)

src/Lean/Parser/Tactic.lean
kmill: I did some rewording and added an extra comment about `Eq.rec` reduction being an issue. (I don't understand it completely, but I think it's because it takes reducing propositions, which the kernel avoids).

src/Lean/Elab/Tactic/ElabTerm.lean
```diff
@@ -372,10 +372,24 @@ private def preprocessPropToDecide (expectedType : Expr) : TermElabM Expr := do
...
+        tactic 'decide' proved that the proposition\
+        {indentExpr expectedType}\n\
+        is false"
+    unless r.isAppOf ``isTrue do
+      throwError "\
+        tactic 'decide' failed for proposition\
+        {indentExpr expectedType}\n\
+        since its 'Decidable' instance reduced to\
+        {indentExpr r}\n\
+        rather than to the 'isTrue' constructor."
```
kmill, line 390: @david-christiansen What do you think of this version?

src/Lean/Parser/Tactic.lean
```diff
@@ -66,13 +66,60 @@ doing a pattern match. This is equivalent to `fun` with match arms in term mode.
...
+
+Limitations:
+- The target is not allowed to contain local variables or metavariables.
+  If there are local variables, you can try first using the `revert` tactic with these local variables
+  to move them into the target, which may allow `decide` to succeed.
+- Because this uses kernel reduction to evaluate the term, `Decidable` instances defined
+  by well-founded recursion might not work, because evaluating them requires reducing proofs.
+  The kernel can also get stuck reducing `Decidable` instances with `Eq.rec` terms for rewriting propositions.
+  These can appear for instances defined using the `rw` and `simp` tactics for example --
+  make sure to use definitions such as `decidable_of_iff` instead.
```
david-christiansen, line 82:
```suggestion
  These can appear for instances defined using the `rw` and `simp` tactics, for example, so
  make sure to use definitions such as `decidable_of_iff` instead.
```

I think this is a bit nicer style.

But in any case, a dash here should be an em dash, which should be done via Unicode because Markdown doesn't support TeX-style dash rewriting, and even if it did, `--` is an en dash.

kmill: Right, `--` is commonly used as a substitute for an em dash, and if this were TeX I would have used `---`. I'm happy to substitute a Unicode symbol.

I don't think using the conjunction "so" would be right here, since the command "make sure [...]" is a separate (yet related) thought, with "instead" referring to the example tactics as well as anything that introduces `Eq.rec`. The first part is not in itself the reason for the directive.

david-christiansen: I now am not sure where our communication is going wrong, but we're not understanding each other! That's a sign that the text could be simpler, I think. What about something like:
```suggestion
  These can appear for instances defined using tactics (such as `rw` and `simp`). To avoid this,
  use definitions such as `decidable_of_iff` instead.
```

david-christiansen: (and I've seen the TeX convention of `--` being an en dash in enough contexts that I really think the Unicode is the right solution here)

david-christiansen: I've seen the TeX convention used in documents that were plainly not intended to be TeX. Though not for some years.

I think we can conclude that double-dashing conventions are varied and confusing, highly dependent on where one came from, and so Unicode is the way forward :-)

david-christiansen (approved):

leodemoura:
Rebased

PR #3428, opened 2024-02-21
chore: remove mkAppN macro in omega

kim-em:


nomeata:
In light of the discussion in https://lean-fro.zulipchat.com/#narrow/stream/398861-general/topic/Idioms.20for.20matching.20.60Expr.60 even the `mkApp6` macro probably causes code bloat? But it’s just too damn practical in this case…

PR #3431, opened 2024-02-21
perf: link lean statically on Unix 

Kha:
A trade-off between speed and download size


Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
Mostly obsoleted by #3529

PR #3432, opened 2024-02-21
feat: functional induction

nomeata:
This adds the concept of **functional induction** to lean.

Derived from the definition of a (possibly mutually) recursive function, a **functional
induction principle** is tailored to proofs about that function. For example from:

```
def ackermann : Nat → Nat → Nat
  | 0, m => m + 1
  | n+1, 0 => ackermann n 1
  | n+1, m+1 => ackermann n (ackermann (n + 1) m)
derive_functional_induction ackermann
```
we get
```
ackermann.induct (motive : Nat → Nat → Prop) (case1 : ∀ (m : Nat), motive 0 m)
  (case2 : ∀ (n : Nat), motive n 1 → motive (Nat.succ n) 0)
  (case3 : ∀ (n m : Nat), motive (n + 1) m → motive n (ackermann (n + 1) m) → motive (Nat.succ n) (Nat.succ m))
  (x x : Nat) : motive x x
```

At the moment, the user has to ask for the functional induction principle explicitly using
```
derive_functional_induction ackermann
```

The module docstring of `Lean/Meta/Tactic/FunInd.lean` contains more details on the
design and implementation of this command.

More convenience around this (e.g. a `functional induction` tactic) will follow eventually.

This PR includes a bunch of `PSum`/`PSigma` related functions in the `Lean.Tactic.FunInd`
namespace. I plan to move these to `PackArgs`/`PackMutual` afterwards, and do some cleaning 
up as I do that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,7 +140,8 @@ jobs:
                 "shell": "msys2 {0}",
                 "CMAKE_OPTIONS": "-G \"Unix Makefiles\" -DUSE_GMP=OFF",
                 // for reasons unknown, interactivetests are flaky on Windows
-                "CTEST_OPTIONS": "--repeat until-pass:2",
+                // also, the liasolver test hits “too many exported symbols”
+                "CTEST_OPTIONS": "--repeat until-pass:2 -E 'leanbenchtest_liasolver.lean'",
                 "llvm-url": "https://github.com/leanprover/lean-llvm/releases/download/15.0.1/lean-llvm-x86_64-w64-windows-gnu.tar.zst",
                 "prepare-llvm": "../script/prepare-llvm-mingw.sh lean-llvm*",
                 "binary-check": "ldd"


src/Lean/Meta/Basic.lean
```diff
@@ -1347,6 +1347,16 @@ private def withNewMCtxDepthImp (allowLevelAssignments : Bool) (x : MetaM α) :
   finally
     modify fun s => { s with mctx := saved.mctx, postponed := saved.postponed }
 
+/--
+Removes `fvarId` from the local context, and replaces occurrences of it with `e`.
+It is the responsibility of the caller to ensure that `e` is well-typed in the context
+of any occurrence of `fvarId`.
```
david-christiansen, line 1353: This operation sounds like "substitution" to me, but you're not using that terminology. Why not?

src/Lean/Meta/Match/MatchEqs.lean
```diff
@@ -109,8 +109,10 @@ def unfoldNamedPattern (e : Expr) : MetaM Expr := do
   - `type` is the resulting type for `altType`.
 
   We use the `mask` to build the splitter proof. See `mkSplitterProof`.
+
+  This can be used to use the alternative of a match expression in its splitter.
 -/
-partial def forallAltTelescope (altType : Expr) (altNumParams numDiscrEqs : Nat)
+partial def forallAltTelescope {α} (altType : Expr) (altNumParams numDiscrEqs : Nat)
```
david-christiansen, line 115: We don't typically use this style inside Lean
```suggestion
partial def forallAltTelescope (altType : Expr) (altNumParams numDiscrEqs : Nat)
```

src/Lean/Meta/Match/MatcherApp/Transform.lean
```diff
@@ -156,4 +156,219 @@ def refineThrough? (matcherApp : MatcherApp) (e : Expr) :
   catch _ =>
     return none
 
+
+/--
+Given `n` and a type `α₁ → α₂ → ... → αₙ → Sort u`, returns the types `α₁, α₂, ..., αₙ`.
```
david-christiansen, line 161:
```suggestion
Given `n` and a type `α₁ → α₂ → ... → αₙ → Sort u`, returns the types `α₁, α₂, ..., αₙ`. An error
is thrown if there are not at least `n` argument types.
```

src/Lean/Meta/Match/MatcherApp/Transform.lean
```diff
@@ -156,4 +156,219 @@ def refineThrough? (matcherApp : MatcherApp) (e : Expr) :
...
+    let Expr.forallE _ α β _ ← pure type | throwError "expected {n} arguments, got {i}"
+    if β.hasLooseBVars then throwError "unexpected dependent type"
+    ts := ts.push α
+    type := β
+  return ts
+
+/--
+Performs a possibly type-changing transformation to a `MatcherApp`.
+
+* `onParams` is run on each parameter and discrimitant
```
david-christiansen, line 180: I don't know the term "discrimitant", and my dictionary doesn't either. Is this specialized technical vocabulary, or should it be "scrutinee"?

src/Lean/Meta/Match/MatcherApp/Transform.lean
```diff
@@ -156,4 +156,219 @@ def refineThrough? (matcherApp : MatcherApp) (e : Expr) :
...
+* `onMotive` runs on the body of the motive, and is passed the motive parameters
+  (same number as discrs)
```
david-christiansen, line 182: What's a "discr"? Same as last time, or not? It's good to be consistent with abbreviations.

src/Lean/Meta/Match/MatcherApp/Transform.lean
```diff
@@ -156,4 +156,219 @@ def refineThrough? (matcherApp : MatcherApp) (e : Expr) :
...
+Given a `MatcherApp`, replaces the motive with one that is inferred from the actual types of the
+alternatives.
+
+For example, given
+```
+(match (motive := Nat → Unit → ?) n with
+ 0 => 1
+ _ => true) ()
+```
+(for any `?`; this will be ignored) will give this type
```
david-christiansen, line 312:
```suggestion
(for any `?`; the motive's result universe will be ignored) will give this type
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+  (case2 : ∀ (n : Nat), motive n 1 → motive (Nat.succ n) 0)
+  (case3 : ∀ (n m : Nat), motive (n + 1) m → motive n (ackermann (n + 1) m) → motive (Nat.succ n) (Nat.succ m))
+  (x x : Nat) : motive x x
+```
+
+## Spec
+
+The functional induction principle
+
+* Take the same fixed parameters as the function.
```
david-christiansen, line 40:
```suggestion
* Takes the same fixed parameters as the function.
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+```
+we get
+```
+ackermann.induct (motive : Nat → Nat → Prop) (case1 : ∀ (m : Nat), motive 0 m)
+  (case2 : ∀ (n : Nat), motive n 1 → motive (Nat.succ n) 0)
+  (case3 : ∀ (n m : Nat), motive (n + 1) m → motive n (ackermann (n + 1) m) → motive (Nat.succ n) (Nat.succ m))
+  (x x : Nat) : motive x x
+```
+
+## Spec
```
david-christiansen, line 36:
```suggestion
## Specification
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+
+The functional induction principle
+
+* Take the same fixed parameters as the function.
+* The motive takes the same non-fixed parameter as the original function.
```
david-christiansen, line 41:
```suggestion
* The motive takes the same non-fixed parameters as the original function.
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+* For each branch of the original function, there is a case in the induction principle.
+  Here branch roughly corresponds to tail-call positions: Branches of top-level
+  if-then-else and matches.
```
david-christiansen, line 44:
```suggestion
* For each branch of the original function, there is a case in the induction principle.
  Here "branch" roughly corresponds to tail-call positions: branches of top-level
  `if`-`then`-`else` and `match` expressions.
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+* The local context of the branch (e.g. the condition of an if-then-else) is provided as assumptions
+  in the case.
```
david-christiansen, line 46:
```suggestion
  in the corresponding induction case.
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+* In addition, for every recursive call in that branch, an induction hypothesis asserting the motive
+  for the arguments of the recursive call is provided.
+* If the recursive call is under binder and it, or its proof of termination, on the the bound,
+  values, these become assumptions on the inductive hypothesis.
```
david-christiansen, line 50: This doesn't make sense to me. My best guess is:
```suggestion
* If the recursive call is under binder and it, or its proof of termination, depend on the the
  bound values, then these become assumptions on the inductive hypothesis.
```
Is that right?

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+import Lean.Meta.Check
+import Lean.Meta.Tactic.Cleanup
+import Lean.Meta.Injective -- for elimOptParam
+import Lean.Elab.PreDefinition.WF.Eqns
+import Lean.Elab.PreDefinition.WF.PackMutual
+import Lean.Elab.Command
+
+/-!
+This module contains code to derive, from the definition of a (possibly mutual) recursive function,
+a **functional induction principle** tailored to proofs about that function. For example from:
```
david-christiansen, line 19: As I understand it, this is really relevant for functions defined by well-founded recursion, rather than structural recursion, but this fact isn't state here. Making this front-and-center will also explain the later comment that it expects a `fix` in the sufficiently-normalized definition.

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+
+## Implementation overview
+
+For a non-mutual, unary function `foo` (or else for the `_unary` function), we
+
+1. expect its definition, possibly after some `whnf`’ing, to be of the form
+   ```
+   def foo := fun x₁ … xₙ (y : a) => fix (fun y' oldIH => body) y
+   ```
+   where `xᵢ…` are the fixed prefix and `y` is parameter of the function.
```
david-christiansen, line 62:
```suggestion
   where `xᵢ…` are the fixed prefix and `y` is the function's argument.
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+
+2. From this structure we derive the type of the `motive`, and start assembling the induction
```
david-christiansen, line 64: This seems more like an ordinary use of the term, rather than a reference to the name in the code below.
```suggestion
2. From this structure we derive the type of the motive, and start assembling the induction
```
If I read wrong, then do this instead:
```suggestion
2. From this structure we derive the type of `motive`, and start assembling the induction
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+   principle:
+   ```
+   def foo.induct := fun x₁ … xₙ (motive : (y : a) → Prop) =>
+    fix (fun y' newIH => T[body])
+   ```
+
+3. The first phase, transformation `T1[body]` (implemented in) `buildInductionBody`,
+   mirrors the branching structure of `foo`, i.e. replicates `ite` and matcher applications,
```
david-christiansen, line 72: What about `dite`?

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+   while adjusting their motive. It also unfolds call to `oldIH` and collects induction hypotheses
```
david-christiansen, line 73:
```suggestion
   while adjusting their motive. It also unfolds calls to `oldIH` and collects induction hypotheses
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+   in conditions (see below).
+
+   In particular when translating a `match`, it is prepared to see the idiom
+   as introduced by `mkFix` via `Lean.Meta.MatcherApp.addArg?`, which refines the type of `oldIH`
+   throught the match. The transformation will replace `oldIH` with `newIH` here.
```
david-christiansen, line 78:
```suggestion
   throughout the match. The transformation will replace `oldIH` with `newIH` here.
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+   def foo.induct := fun x₁ … xₙ (motive : (y : a) → Prop) =>
+    fix (fun y' newIH => T[body])
+   ```
+
+3. The first phase, transformation `T1[body]` (implemented in) `buildInductionBody`,
+   mirrors the branching structure of `foo`, i.e. replicates `ite` and matcher applications,
+   while adjusting their motive. It also unfolds call to `oldIH` and collects induction hypotheses
+   in conditions (see below).
+
+   In particular when translating a `match`, it is prepared to see the idiom
```
david-christiansen, line 76:
```suggestion
   In particular, when translating a `match` it is prepared to recognize the idiom
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+   as introduced by `mkFix` via `Lean.Meta.MatcherApp.addArg?`, which refines the type of `oldIH`
+   throught the match. The transformation will replace `oldIH` with `newIH` here.
+   ```
+        T[(match (motive := fun oldIH => …) y with | … => fun oldIH' => body) oldIH]
+    ==> (match (motive := fun newIH => …) y with | … => fun newIH' => T[body]) newIH
+   ```
+
+4. When a tail position (no more branching) is found, function `buildInductionCase` assembles the
+   type of the case: A fresh `MVar` asserts the current goal, unwanted values from the local context
```
david-christiansen, line 85:
```suggestion
   type of the case: a fresh `MVar` asserts the current goal, unwanted values from the local context
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+   are cleared, and the current `body` is frisked for induction hypotheses using `collectIHs`,
```
david-christiansen, line 86: What does "frisked" mean here? I only know the term as searching someone's clothes for hidden weapons, which seems unlikely to be what was meant!

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+6. The function `foldCalls` replaces calls to `oldIH` with calls to `foo` that
+   make sense to the user.
+
+   At the end of this transformation, no mention of `oldIH` must remain.
+
+7. After this construction, the MVars introduced by `buildInductionCase` are turned into parameters.
+
+The resulting term then becomes `foo.induct` at its inferred type.
+
+If `foo` is not unary and/or part of a mutual reduction, then the induction theorem for `foo._unary`
```
david-christiansen, line 114: Readers may not know about `_unary` - the first mention of it should briefly summarize what it is.

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+-/
+
+set_option autoImplicit false
+
+namespace Lean.Tactic.FunInd
+
+open Lean Elab Meta
+
+/-- Opens the body of a lambda, _without_ putting the free variable into the local context.
+This is used when replacing that paramters with a different expression.
```
david-christiansen, line 147:
```suggestion
This is used when replacing parameters with different expressions.
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+This way it will not be picked up by metavariables.
+-/
+def removeLamda {α} (e : Expr) (k : FVarId → Expr →  MetaM α) : MetaM α := do
+  let .lam _n _d b _bi := ← whnfD e | throwError "removeLamda: expected lambda, got {e}"
+  let x ← mkFreshFVarId
+  let b := b.instantiate1 (.fvar x)
+  k x b
+
+-- Replace calls to oldIH back to calls to the original function. At the end,
+-- oldIH better be unused
```
david-christiansen, line 157:
```suggestion
/-- Replace calls to oldIH back to calls to the original function. At the end, if `oldIH` occurs, an error is thrown. -/
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+      go s.mvarId s.fields[1]!.fvarId! (ys.push s.fields[0]!)
+    else
+      let ys := ys.push (mkFVar y)
+      mvarId.assign (alt.beta ys)
+  go mvar.mvarId! y #[]
+  instantiateMVars mvar
+
+/--
+Given expression `e` with type `(x : A) → (y : B[x]) → … → (z : D[x,y]) → R`
+return an expression of type `(x : A ⊗' B ⊗' … ⊗' D) → R`.
```
david-christiansen, line 664: Indicative mood
```suggestion
returns an expression of type `(x : A ⊗' B ⊗' … ⊗' D) → R`.
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+  go : List Expr → Array Expr → MetaM α
+  | [], acc => k acc
+  | t::ts, acc => do
+    let name := s!"{name}{acc.size+1}"
+    withLocalDecl name .default t fun x => do
+      go ts (acc.push x)
+
+
+/--
+Given expression `e` of type `(x : a ⊗' b ⊕' c ⊗' d) → e[x]`, returns expression of type
```
david-christiansen, line 702: This isn't `e`'s type, is it?

I generally don't understand what this docstring is indicating.

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+      else
+        k e
+
+
+-- Adapted from PackMutual: TODO: Compare and unify
+/--
+  Combine/pack the values of the different definitions in a single value
+  `x` is `PSum`, and we use `PSum.casesOn` to select the appropriate `preDefs.value`.
+  See: `packMutual`.
+  Remark: this method does not replace the nested recursive `preDefValues` applications.
```
david-christiansen, line 734:
```suggestion

  Remark: this method does not replace the nested recursive `preDefValues` applications.
```
Separate paragraphs seems useful here

david-christiansen:
And now a bit of velocipede storage facility color determination: isn't `derive_induction` a very broad keyword for something fairly specific? What if the syntax were `deriving fun induction for f` instead? This doesn't add keywords, but it is more specific. Or even `deriving functional induction for f` with `functional` being non-reserved?

src/Lean/Meta/Tactic/FunInd.lean
nomeata: I didn't think frisking is reservd for weapons.
searched through or combed are maybe less flowery phrasings :-)

david-christiansen:
Thanks for leaving detailed docs, they were generally easy to follow and will be a big benefit to anyone encountering this in the future!

nomeata:
The command keyword is certainly up for discussion.

In Coq a multi word command would be quite idiomatic, but is it so in Lean?

We envision that in the future, when we have a `functional induction` tactic, the explicit command will actually be rarely used, so happy to use something more verbose and explicit.

I also wonder if “derive” should be used when there is no type class around. What other verb could we use. `generate`? `define`?

src/Lean/Meta/Match/MatcherApp/Transform.lean
nomeata: Scrurtinee seems to be mostly GHC lingo, and `discrs` is used in the lean code base. 🤷‍♂️

src/Lean/Meta/Basic.lean
nomeata: No deep reason, just out of consistency with existing function names and doc strings: https://leanprover-community.github.io/mathlib4_docs/Lean/Expr.html#Lean.Expr.replaceFVar

src/Lean/Meta/Tactic/FunInd.lean
nomeata: Down the road we might extend it to to structurally recursive or possibly even non-recursive functions, but for now you are right.

```suggestion
This module contains code to derive, from the definition of a (possibly mutual) well-founded
recursive function, a **functional induction principle** tailored to proofs about that
function. For example from:
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+* Mutual recursion is supported and results in multiple motives.
+
+
+## Implementation overview
+
+For a non-mutual, unary function `foo` (or else for the `_unary` function), we
+
+1. expect its definition, possibly after some `whnf`’ing, to be of the form
+   ```
+   def foo := fun x₁ … xₙ (y : a) => fix (fun y' oldIH => body) y
```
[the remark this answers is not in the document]
nomeata:
```suggestion
   def foo := fun x₁ … xₙ (y : a) => WellFounded.fix (fun y' oldIH => body) y
```

src/Lean/Meta/Tactic/FunInd.lean
nomeata: Parameter is more correct than argument here, isn't it?

src/Lean/Meta/Match/MatcherApp/Transform.lean
nomeata:
```suggestion
* `onParams` is run on each parameter and discriminant
```

src/Lean/Meta/Match/MatcherApp/Transform.lean
nomeata:
```suggestion
  (one for each `MatcherApp.discrs`)
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+
+* Take the same fixed parameters as the function.
+* The motive takes the same non-fixed parameter as the original function.
+* For each branch of the original function, there is a case in the induction principle.
+  Here branch roughly corresponds to tail-call positions: Branches of top-level
+  if-then-else and matches.
+* The local context of the branch (e.g. the condition of an if-then-else) is provided as assumptions
+  in the case.
+* In addition, for every recursive call in that branch, an induction hypothesis asserting the motive
+  for the arguments of the recursive call is provided.
```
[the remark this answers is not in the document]
nomeata: Could, but given that this is the point of the exercise, it seems to deserve its own bullet.

src/Lean/Meta/Tactic/FunInd.lean
nomeata:
```suggestion
   mirrors the branching structure of `foo`, i.e. replicates `dite` and matcher applications,
```
(In fact, there is no `ite` here because `Fix.lean` replaces `ite` with `dite` in the body of recursive functions.)

src/Lean/Meta/Tactic/FunInd.lean
nomeata:
```suggestion
   are cleared, and the current `body` is combed for induction hypotheses using `collectIHs`,
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,855 @@
...
+      d ← mkLambdaFVars #[x] d
+      d ← mkAppOptM ``PSigma #[some (← inferType x), some d]
+    return (d, codomain)
+  withLocalDecl `x .default d fun x => do
+    let value ← mkPSigmaNCasesOn x.fvarId! codomain e
+    mkLambdaFVars #[x] value
+
+/--
+Given type `(a ⊗' b ⊕' c ⊗' d) → e`, brings `a → b → e` and `c → d → e`
+into scope and passes them to the contiuation.
```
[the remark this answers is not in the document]
nomeata:
```suggestion
into scope as fresh local declarations and passes their FVars to the contiuation.
```

david-christiansen:
I like `derive` here because it minimizes keyword capture. But it is a bit strange - it'll be good when the tactic lazily creates the induction principle. We do already have `deriving instance X for Y`, which I tried to follow in the suggestion.

I guess the direction that this hints at is having "deriving X ..." commands be a little code-generation DSL that makes various things get built. I don't know if we want to go that way more generally.

nomeata:
Ok, then `derive` is fine. Between
```
derive_functional_induction foo
derive functional induction for foo
```
I think the former is more lean-y. We also say `termination_by` and `not `termination by` etc., don't we?

david-christiansen:
If we're not hooking into the `deriving _ for _` syntax, then `generate` or `derive` seem equally good to me.

Looking forward to the tactic - this will be a big step up for verified programming!

src/Lean/Meta/Match/MatcherApp/Transform.lean
david-christiansen: Consistency is good here!

"Scrutinee" is also what we called it in Idris 1, but there were plenty of GHC-isms in that code as well, like "zonk".

david-christiansen: Though "-ant" is ["a person or thing performing or causing the stated action"](https://dictionary.cambridge.org/dictionary/english/ant?q=-ant). But the thing we're talking about here is not that which is causing discrimination to occur, it's that which is being subject to discrimination. "Discriminee" isn't in any English dictionaries that I could find, but it follows the regular pattern of "-ee" pointing at the object of a verb.

src/Lean/Meta/Tactic/FunInd.lean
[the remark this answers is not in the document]
david-christiansen: What about indenting it then? Two bullets suggests that it's independent, which it isn't really.

src/Lean/Meta/Tactic/FunInd.lean
david-christiansen: Now to really split hairs so thin that a microscope can hardly see them: a well-founded recursive function is, in my mind, just a total recursive function. What about "from the definition of a recursive function (or mutually recursive functions) defined by well-founded recursion"?

src/Lean/Meta/Tactic/FunInd.lean
david-christiansen: Ah, good point.

What about "and `y` is the parameter of the function" then? There should be a definite article.

src/Lean/Meta/Match/MatcherApp/Transform.lean
nomeata: Indeed googling shows up Coq docs that (occasionally? rarely?) use of `discriminee`, but never discriminant.

But yet:
```
~/build/lean/lean4 $ git grep discriminant|wc -l
84
~/build/lean/lean4 $ git grep discriminee|wc -l
0
```

I don’t mind changing that, but if we do, then everywhere separate of this PR.

Do we need a little “glossary and nomenclature” document somewhere where we can keep track of such decisions?

src/Lean/Meta/Tactic/FunInd.lean
david-christiansen: Oh, "frisked" could work here, the sudden metaphor just threw me off and made me think it was a mistake rather than intentional. For me, "comb" has more a connotation of "setting in place" than "removing the dirt", although the latter certainly does work in the case of lice or sand. 

What about "the current `body`'s induction hypotheses are confiscated by `collectIHs`" if you want the metaphor?

But I think the metaphor was jarring in a body of text that's otherwise straightforward and technical, so I'd reduce the fun here if anything.

src/Lean/Meta/Tactic/FunInd.lean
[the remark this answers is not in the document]
nomeata: Rewrote completely, no bullets, and move the more important facts (induction hyptheses) first.

src/Lean/Meta/Tactic/FunInd.lean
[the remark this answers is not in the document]
david-christiansen: Much better!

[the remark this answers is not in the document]
david-christiansen: Even better would be "the continuation `k`".

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -36,20 +36,23 @@ ackermann.induct (motive : Nat → Nat → Prop) (case1 : ∀ (m : Nat), motive
...
+For every recursive call in that branch, an induction hypothesis asserting the
+motive for the arguments of the recursive call is provided.
+If the recursive call is under binder and it, or its proof of termination,
+depend on the the bound values, then these become assumptions on the inductive
+hypothesis.
+
+Additionally, the local context of the branch (e.g. the condition of an
+if-then-else) is provided as assumptions in the corresponding induction case.
+
+Mutual recursion is supported and results in multiple motives.
```
david-christiansen: This is a nice improvement!

src/Lean/Meta/Tactic/FunInd.lean
nomeata:
```suggestion
   where `xᵢ…` are the fixed parameter prefix and `y` is the varying parameter of
   the function.
```

leodemoura:
Consider the following example
```
inductive Vector (α : Type u) : Nat → Type u
  | nil : Vector α 0
  | cons : α → Vector α n → Vector α (n+1)

def Vector.zip : Vector α n → Vector β n → Vector (α × β) n
  | nil, nil => nil
  | cons a as, cons b bs => cons (a, b) (zip as bs)
termination_by as bs => sizeOf as

derive_functional_induction Vector.zip
```
We get the following induction principle
```
Vector.zip.induct.{u_1, u_2} {α : Type u_1} (motive : (x : Nat) → (x_1 : Type u_2) → Vector α x → Vector x_1 x → Prop)
  (case1 : ∀ (fst : Type u_2), motive 0 fst Vector.nil Vector.nil)
  (case2 :
    ∀ (fst : Type u_2) (n : Nat) (a : α) (as : Vector α n) (b : fst) (bs : Vector fst n),
      motive n fst as bs → motive (n + 1) fst (Vector.cons a as) (Vector.cons b bs))
  (x : Nat) (x : Type u_2) (x : Vector α x) (x : Vector x x) : motive x x x x
```
`β` has been abstracted although it is a parameter and is fixed.

src/Lean/Meta/Match/MatcherApp/Transform.lean
timotree3: "scrutinee" is also used in the Rust reference: https://doc.rust-lang.org/reference/expressions/match-expr.html#match-expressions

leodemoura:
The induction principle for the following. function has three cases as expected
```

def g (i : Nat) (as : Array (Fin 2)) : Nat :=
  if _  : i < as.size then
    have : as.size - (i+2) < as.size - i := by omega
    if as[i] = 0 then
      g (i+1) as + 1
    else
      g (i+2) as + 2
  else
    0
termination_by as.size - i

derive_functional_induction g
#check g.induct
```
But, the induction principle for the following function has only 2. 
```
def f (i : Nat) (as : Array (Fin 2)) : Nat :=
  if _  : i < as.size then
    have : as.size - (i+2) < as.size - i := by omega
    match as[i] with
    | 0 => f (i+1) as + 1
    | 1 => f (i+2) as + 2
  else
    0
termination_by as.size - i

derive_functional_induction f
#check f.induct
```

leodemoura:
@nomeata Pushed a new test with mutual recursion + dependent types. The induction principle looks correct to me, but the variable names look confusing. For example, on `case8`, it would be great to have `∀ (α : Type) (results : List α) (t1 t2 : Finite),` instead of `∀ (fst : Type) (snd : List fst) (t1 t2 : Finite),`.

leodemoura:
The following example produces a cryptic error message
```
def takeWhile (p : α → Bool) (as : Array α) : Array α :=
  go 0 #[]
where
  go (i : Nat) (r : Array α) : Array α :=
    if h : i < as.size then
      let a := as.get ⟨i, h⟩
      if p a then
        go (i+1) (r.push a)
      else
        r
    else
      r
  termination_by as.size - i

derive_functional_induction takeWhile -- Cryptic error message

derive_functional_induction takeWhile.go -- Works as expected
```

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -0,0 +1,863 @@
...
+  | .sort .. | .lit .. | .const .. | .mvar .. | .bvar .. =>
+    unreachable! -- cannot contain free variables, so early exit above kicks in
+
+  | .fvar _ =>
+    throwError "collectIHs: could not collect recursive calls, unsaturated application of old induction hypothesis"
+
+def withLetDecls {α} (vals : Array Expr) (k : Array FVarId → MetaM α) (i : Nat := 0) : MetaM α := do
+  if h : i < vals.size then
+    let e := vals[i]
+    withLetDecl s!"IH{i+1}" e (← inferType e) fun a =>
```
leodemoura, line 365: "IH" => "ih"

nomeata:
Thanks for the review and additional tests! I’ll have a closer look tomorrow.

---

Re `Vector.zip`: The problem already starts with `Vector.zip` itself, where `β` comes after the first varying parameter and thus isn’t part of the fixed prefix:
```
Vector.zip._unary.{u_1, u_2} {α : Type u_1} (_x✝ : (n : Nat) ×' (β : Type u_2) ×' (_ : Vector α n) ×' Vector β n) :
  Vector (α × _x✝.2.1) _x✝.1
```
The induction principle follows the structure of the function definition, so I’d say that part is correct, and the issue needs to be addressed earlier. Not sure if `PreDefinition.WF` should reorder arguments, or if autoImplicit parameters should be inserted earlier, or if we simply expect people to be explicit in the argument order in cases like this.

---

As for the variable names, I expect that some (if not all) unfortunate names will improve with my planned refactoring of the argument packing/unpacking infrastructure. Is it ok to merge this with suboptimal names first and refine later, or should I try harder already here?

nomeata:
> The induction principle for the following. function has three cases as expected
…
> But, the induction principle for the following function has only 2.

I know how to fix that, will do that tomorrow.

It isn’t always clear that more cases are always better, and the heuristic of “split tail-call matches but not other matches” is a bit arbitrary – it’s not actually technically needed, as I initially thought, but it probably often does the right thing. We’ll learn more when people are starting to use it.

nomeata:
> > The induction principle for the following. function has three cases as expected
> > …
> > But, the induction principle for the following function has only 2.
> 
> I know how to fix that, will do that tomorrow.

Ah, I remember, it’s not so easy. Consider this example:
```
def with_match_tailrec : Nat → Nat
  | 0 => 0
  | n+1 =>
    match n % 2 with
    | 0 => with_match_tailrec n
    | _ => with_match_tailrec n
```

with the current code (extended in the obvious way), we’d get
```
info: with_match_tailrec.induct (motive : Nat → Prop) (case1 : motive 0)
  (case2 case3 : ∀ (n : Nat), motive n → motive (Nat.succ n)) (x : Nat) : motive x
```

That’s not helpful. Only if the user writes `h : n % 2` in the match, they get the more helpful
```
info: with_match_tailrec.induct (motive : Nat → Prop) (case1 : motive 0)
  (case2 : ∀ (n : Nat), n % 2 = 0 → motive n → motive (Nat.succ n))
  (case3 : ∀ (n x : Nat), (x = 0 → False) → n % 2 = x → motive n → motive (Nat.succ n)) (x : Nat) : motive x
-
```

I believe at the moment we don’t have the infrastructure to add these discriminant assumptions afterwards. Maybe `forallAltTelescope` could learn to create them (optionally)… let’s see.

nomeata:
Ok, I have a plan: The translation should manually add `scrut = constr a b c` hypotheses to any scrutinee that doesn't already have a user-written `h :` one, or that isn't refined in the goal.

nomeata:
Plan executed. Excerpt from the module docs:

>  In addition, the information gathered from the match is preserved, so that when performing the
>  proof by induction, the user can reliably enter the right case. To achieve this
> 
>  * the matcher is replaced by its splitter, which brings extra assumptions into scope when
>    patterns are overlapping
>  * simple discriminants that are mentioned in the goal (i.e plain parameters) are instantiated
>    in the code.
>  * for discriminants that are not instantiated that way, equalities connecting the discriminant
>    to the instantiation are added (just as if the user wrote `match h : x with …`)


It’s pushing a bit on the complexity budget, but I agree it is the right thing to do.

PR #3435, opened 2024-02-21
feat: bv_omega, a simple BitVec->Nat preprocessor + omega

kim-em:
```
example (x y z : BitVec 8)
    (hx : x >>> 1 < 16) (hy : y < 16) (hz : z = x + 2 * y) : z ≤ 64 := by
  bv_omega

example (x : BitVec 8) (hx : (x + 1) <<< 1 = 4) : x = 1 ∨ x = 129 := by
  bv_omega

example (x y : BitVec 64) (_ : x < (y.truncate 32).zeroExtend 64) :
    ~~~x > (1#64 <<< 63) := by
  bv_omega
```

### DIFF
--- a/src/Init/Data/BitVec/Basic.lean
+++ b/src/Init/Data/BitVec/Basic.lean
@@ -75,7 +75,7 @@ theorem isLt (x : BitVec w) : x.toNat < 2^w := x.toFin.isLt
 
 /-- Theorem for normalizing the bit vector literal representation. -/
 -- TODO: This needs more usage data to assess which direction the simp should go.
-@[simp] theorem ofNat_eq_ofNat : @OfNat.ofNat (BitVec n) i _ = .ofNat n i := rfl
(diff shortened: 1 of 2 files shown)


src/Init/Tactics.lean
```diff
@@ -1088,6 +1088,13 @@ Currently, all of these are on by default.
 -/
 syntax (name := omega) "omega" (config)? : tactic
 
+/--
+`bv_omega` is `omega` with an additional preprocessor that turns statements about `BitVec` into statements about `Nat`.
+Currently the preprocessor is implemented as `try simp only [bv_toNat] at *`.
+`bv_toNat` is a `@[simp]` attribute that you can (cautiously) add to more theorems.
+-/
+macro "bv_omega" : tactic => `(tactic| (try simp only [bv_toNat] at *) <;> omega)
```
nomeata, line 1096: Instead of `try`, this could use `{ config := { failIfUnchanged := false }}`. Is there any reason to prefer the latter over the former? It’s more verbose, but it seems less roundabout than failing first and then recovering. (Maybe it simply doesn’t matter.)

PR #3436, opened 2024-02-21
feat: setup simp lemmas: 'msb -> getLsb -> decide ...'

bollu:
This is a follow up to 'https://github.com/leanprover/std4/pull/645', where the simp lemmas were requested: https://github.com/leanprover/std4/pull/645#issuecomment-1944862251

---

Note that @semorrison  asked to use `(Fin.last _)` to index. Now that we use a `Nat` to index `msb` , the pattern `(Fin.last _)` would not have the width be automatically inferred. Therefore, I've changed the definitions to use `Nat` for indexing.

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -130,7 +130,7 @@ theorem msb_eq_getLsb_last (x : BitVec w) :
   · simp [BitVec.eq_nil x]
   · simp
 
-theorem getLsb_last (x : BitVec (Nat.succ w)) :
+@[simp] theorem getLsb_last (x : BitVec (w + 1)) :
     x.getLsb w = decide (2 ^ w ≤ x.toNat) := by
   simp only [Nat.zero_lt_succ, decide_True, getLsb, Nat.testBit, Nat.succ_sub_succ_eq_sub,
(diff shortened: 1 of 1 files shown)


bollu:
awaiting-review

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -123,9 +123,18 @@ private theorem lt_two_pow_of_le {x m n : Nat} (lt : x < 2 ^ m) (le : m ≤ n) :
 
 /-! ### msb -/
 
-theorem msb_eq_decide (x : BitVec (Nat.succ w)) : BitVec.msb x = decide (2 ^ w ≤ x.toNat) := by
-  simp only [BitVec.msb, getMsb, Nat.zero_lt_succ,
-    decide_True, getLsb, Nat.testBit, Nat.succ_sub_succ_eq_sub,
+@[simp]
```
semorrison, line 126: I'm sorry if I'm contradicting myself, but I'd prefer this not be a simp lemma (for now). Just add it to the simp call in `msb_eq_decide`.

semorrison:
Oh `getLsb_last` is a good simp lemma, you didn't need to remove that!

I'm just going to push that back, and merge now.

PR #3442, opened 2024-02-21
fix: expression tree elaborator for relations now localizes error messages to the LHS or RHS

kmill:
Added `withRef` when processing the LHS or RHS. Without this, in an expression such as `true = ()` the entire expression would be highlighted with "type mismatch, `()` has type `Unit` but is expected to have type `Bool`". Now the error is localized to `()`.

This behavior was pointed out [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/error.20location.20bug/near/422665805).


Kha:
Test please? :)

PR #3445, opened 2024-02-21
feat: show basic properties of BitVec multiplication

alexkeizer:
Show that multiplication of bitvectors is associative and commutative, and show that it has 1#w as identity (both on the left and right).


alexkeizer:
@joehendrix I'd like your opinion on this PR: we're proving commutativity and associativity of bitvector multiplication by appealing to these properties of `Fin` multiplication. But, not in a way that relies on the definition of `BitVec.mul` being `Fin.mul`, we instead use the `toFin_mul` lemma, so this proof still works after #3430.

Is this an acceptable level of coupling between `BitVec` and `Fin`, since it's just an implementation detail of the proof, or would you prefer we stay away from `Fin` entirely?

joehendrix:
@alexkeizer Looks good to me.  I have no concerns with relying on `Fin` in lemmas when it is useful.

PR #3447, opened 2024-02-22
chore: upstream exact? and apply? from Std

joehendrix:
This is still a draft PR, but includes the core exact? and apply? tactics.

Still need to convert to builtin syntax and test on Std.

### DIFF
--- a/src/Lean/Meta/Tactic/LibrarySearch.lean
+++ b/src/Lean/Meta/Tactic/LibrarySearch.lean
@@ -162,8 +162,8 @@ def mkImportFinder : IO CandidateFinder := do
 
 end IncDiscrTreeFinder
 
-initialize registerTraceClass `Tactic.stdLibrarySearch
-initialize registerTraceClass `Tactic.stdLibrarySearch.lemmas
+initialize registerTraceClass `Tactic.librarySearch
+initialize registerTraceClass `Tactic.librarySearch.lemmas
 
 /-- State for resolving imports -/
 private def LibSearchState := IO.Ref (Option CandidateFinder)
@@ -294,7 +294,7 @@ otherwise the full list of subgoals is returned.
 private def librarySearchLemma (cfg : ApplyConfig) (act : List MVarId → MetaM (List MVarId))
     (allowFailure : MVarId → MetaM Bool) (cand : Candidate)  : MetaM (List MVarId) := do
   let ((goal, mctx), (name, mod)) := cand
-  withTraceNode `Tactic.stdLibrarySearch (return m!"{emoji ·} trying {name} with {mod} ") do
+  withTraceNode `Tactic.librarySearch (return m!"{emoji ·} trying {name} with {mod} ") do
     setMCtx mctx
     let lem ← mkLibrarySearchLemma name mod
     let newGoals ← goal.apply lem cfg
@@ -342,7 +342,7 @@ private def librarySearch' (goal : MVarId)
     (allowFailure : MVarId → MetaM Bool)
     (leavePercentHeartbeats : Nat) :
     MetaM (Option (Array (List MVarId × MetavarContext))) := do
-  withTraceNode `Tactic.stdLibrarySearch (return m!"{librarySearchEmoji ·} {← goal.getType}") do
+  withTraceNode `Tactic.librarySearch (return m!"{librarySearchEmoji ·} {← goal.getType}") do
   profileitM Exception "librarySearch" (← getOptions) do
   let importFinder ← do
         let r := ext.getState (←getEnv)


src/Lean/Elab/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,86 @@
...
+      let g ← g.withContext (instantiateMVars (.mvar g))
+      return required.all fun e => e.occurs g
+    match ← librarySearch goal tactic allowFailure with
+    -- Found goal that closed problem
+    | none =>
+      addExactSuggestion tk (← instantiateMVars (mkMVar mvar)).headBeta
+    -- Found suggestions
+    | some suggestions =>
+      if requireClose then throwError
+        "`std_exact?` could not close the goal. Try `std_apply?` to see partial suggestions."
```
david-christiansen, line 48:
```suggestion
        "`exact?` could not close the goal. Try `apply?` to see partial suggestions."
```

src/Lean/Elab/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,86 @@
...
+      reportOutOfHeartbeats `library_search tk
+      for (_, suggestionMCtx) in suggestions do
+        withMCtx suggestionMCtx do
+          addExactSuggestion tk (← instantiateMVars (mkMVar mvar)).headBeta (addSubgoalsMsg := true)
+      if suggestions.isEmpty then logError "std_apply? didn't find any relevant lemmas"
```
david-christiansen, line 53:
```suggestion
      if suggestions.isEmpty then logError "apply? didn't find any relevant lemmas"
```

david-christiansen: Or even
```suggestion
      if suggestions.isEmpty then logError "exact? didn't find any relevant lemmas"
```

src/Init/Tactics.lean
```diff
@@ -1280,6 +1280,12 @@ a lemma from the list until it gets stuck.
 syntax (name := applyRules) "apply_rules" (config)? (&" only")? (args)? (using_)? : tactic
 end SolveByElim
 
+/-- Syntax for `exact?` -/
```
david-christiansen, line 1283: It would be nice if this docstring described what `exact?` does! What about something like:
```suggestion
/-- Searches all libraries in scope for a definition that can solve the goal using `exact` -/
```

What does the `using` option do here?

src/Init/Tactics.lean
```diff
@@ -1280,6 +1280,12 @@ a lemma from the list until it gets stuck.
...
+syntax (name := exact?) "exact?" (" using " (colGt ident),+)? : tactic
+
+/-- Syntax for `apply?` -/
```
david-christiansen, line 1286:
```suggestion
/-- Searches all libraries in scope for relevant lemmas that can be used to refine the goal with `apply` -/
```

src/Lean/Elab/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,86 @@
...
+
+
+-- TODO: implement the additional options for `library_search` from Lean 3,
+-- in particular including additional lemmas
+-- with `std_exact? [X, Y, Z]` or `std_exact? with attr`.
+
+-- For now we only implement the basic functionality.
+-- The full syntax is recognized, but will produce a "Tactic has not been implemented" error.
+
+/-- Implementation of the `exact?` tactic. -/
```
david-christiansen, line 23: See suggestions on syntax for a docstring that's a bit more useful.

Here, it'd be good to see docs for the meaning of each argument.

src/Lean/Elab/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Lean.Elab.LibrarySearch
+
+open Lean Meta LibrarySearch
+open Elab Tactic TryThis
+
+
+-- TODO: implement the additional options for `library_search` from Lean 3,
+-- in particular including additional lemmas
+-- with `std_exact? [X, Y, Z]` or `std_exact? with attr`.
```
david-christiansen, line 18:
```suggestion
-- with `exact? [X, Y, Z]` or `exact? with attr`.
```

src/Lean/Elab/Tactic/LibrarySearch.lean
david-christiansen: Does this code path know which tactic we're in?

src/Lean/Elab/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,86 @@
...
+  let goal ← mkFreshExprMVar expectedType
+  let (_, introdGoal) ← goal.mvarId!.intros
+  introdGoal.withContext do
+    let tactic := fun exfalso g => solveByElim []  (maxDepth := 6) exfalso g
+    if let some suggestions ← librarySearch introdGoal tactic then
+      reportOutOfHeartbeats `library_search tk
+      for suggestion in suggestions do
+        withMCtx suggestion.2 do
+          addTermSuggestion tk (← instantiateMVars goal).headBeta
+      if suggestions.isEmpty then logError "std_exact? didn't find any relevant lemmas"
```
david-christiansen, line 80:
```suggestion
      if suggestions.isEmpty then logError "exact? didn't find any relevant lemmas"
```

src/Lean/Meta/Basic.lean
```diff
@@ -1067,6 +1067,23 @@ partial def withNewLocalInstances (fvars : Array Expr) (j : Nat) : n α → n α
...
   map2MetaM (fun k => forallTelescopeImp type k) k
 
+/--
+Given a monadic function `F` that takes a type and a term of that type and produces a new term,
+lifts this to the monadic function that opens a `∀` telescope, applies `F` to the body,
+and then builds the lambda telescope term for the new term.
+-/
+def mapForallTelescope' (F : Expr → Expr → MetaM Expr) (forallTerm : Expr) : MetaM Expr := do
+  forallTelescope (← inferType forallTerm) fun xs ty => do
+    mkLambdaFVars xs (← F ty (mkAppN forallTerm xs))
```
david-christiansen, line 1077:
```suggestion
/--
Given a monadic function `f` that takes a type and a term of that type and produces a new term,
lifts this to the monadic function that opens a `∀` telescope, applies `f` to the body,
and then builds the lambda telescope term for the new term.
-/
def mapForallTelescope' (f : Expr → Expr → MetaM Expr) (forallTerm : Expr) : MetaM Expr := do
  forallTelescope (← inferType forallTerm) fun xs ty => do
    mkLambdaFVars xs (← f ty (mkAppN forallTerm xs))
```

src/Lean/Meta/Basic.lean
```diff
@@ -1067,6 +1067,23 @@ partial def withNewLocalInstances (fvars : Array Expr) (j : Nat) : n α → n α
...
+
+/--
+Given a monadic function `F` that takes a term and produces a new term,
+lifts this to the monadic function that opens a `∀` telescope, applies `F` to the body,
+and then builds the lambda telescope term for the new term.
+-/
+def mapForallTelescope (F : Expr → MetaM Expr) (forallTerm : Expr) : MetaM Expr := do
+  mapForallTelescope' (fun _ e => F e) forallTerm
```
david-christiansen, line 1085:
```suggestion
/--
Given a monadic function `f` that takes a term and produces a new term,
lifts this to the monadic function that opens a `∀` telescope, applies `f` to the body,
and then builds the lambda telescope term for the new term.
-/
def mapForallTelescope (f : Expr → MetaM Expr) (forallTerm : Expr) : MetaM Expr := do
  mapForallTelescope' (fun _ e => f e) forallTerm
```

src/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,866 @@
...
+-/
+import Lean.Meta.CompletionName
+import Lean.Meta.DiscrTree
+
+/-!
+# Lazy Discrimination Tree
+
+This file defines a new type of discrimination tree optimized for
+rapidly population of imported modules for use in tactics.  It uses a
+lazy initialization strategy.
```
david-christiansen, line 14:
```suggestion
This file defines a new type of discrimination tree optimized for
rapid population of imported modules for use in tactics.  It uses a
lazy initialization strategy.
```

src/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,866 @@
...
+
+-- This namespace contains definitions copied from Lean.Meta.DiscrTree.
+namespace MatchClone
+
+/--
+Discrimination tree key.
+-/
+private inductive Key where
+    /-- Constant -/
+  | const : Name → Nat → Key
```
david-christiansen, line 35: Most of the Lean codebase puts the docstring for a constructor after the `|`

src/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,866 @@
...
+
+instance : Hashable Key := ⟨Key.hash⟩
+
+end Key
+
+private def tmpMVarId : MVarId := { name := `_discr_tree_tmp }
+private def tmpStar := mkMVar tmpMVarId
+
+/--
+  Return true iff the argument should be treated as a "wildcard" by the discrimination tree.
```
david-christiansen, line 64:
```suggestion
  Returns true iff the argument should be treated as a "wildcard" by the discrimination tree.
```

src/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,866 @@
...
+    So, we index implicit arguments if they are types.
+    This setting seems sensible for simplification theorems such as:
+    ```
+    forall (x y : Unit), (@Eq Unit x y) = true
+    ```
+    If we ignore the implicit argument `Unit`, the `DiscrTree` will say it is a candidate
+    simplification theorem for any equality in our goal.
+
+  Remark: if users have problems with the solution above, we may provide a `noIndexing` annotation,
+  and `ignoreArg` would return true for any term of the form `noIndexing t`.
```
david-christiansen, line 91:
```suggestion
  - Proofs are ignored because of proof irrelevance. It doesn't make sense to try to
    index their structure.

  - Instance implicit arguments (e.g., `[Add α]`) are ignored because they are "morally"
    canonical. Moreover, we may have many definitionally equal terms floating around.
    Example: `Ring.hasAdd Int Int.isRing` and `Int.hasAdd`.

  - Implicit arguments (e.g., `{α : Type}`) are not ignored. It may seem that they
    should be, since users don't "see" them,  and may not even understand why
    some simplification rule is not firing.
    However, in type class resolution, there are instances such as `Decidable (@Eq Nat x y)`,
    where `Nat` is an implicit argument. Thus, the path
    ```
    Decidable -> Eq -> * -> * -> * -> [Nat.decEq]
    ```
    would be added to the discrimination tree IF the implicit `Nat` argument were ignored.
    This would be BAD since **ALL** decidable equality instances would be in the same path.
    So, implicit arguments are indexed if they are types.
    This setting seems sensible for simplification theorems such as:
    ```
    forall (x y : Unit), (@Eq Unit x y) = true
    ```
    If we ignore the implicit argument `Unit`, the `DiscrTree` will say it is a candidate
    simplification theorem for any equality in our goal.

  Remark: if users have problems with the solution above, we may provide a `noIndexing` annotation,
  and `ignoreArg` would return true for any term of the form `noIndexing t`.
```

But this docstring seems to include details of `simp` - is it copied from the other discrtree? If so, it'd be good to check that this makes sense still.

src/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,866 @@
...
+    MetaM (Array Expr)
+  | i, .app f a, todo => do
+    if (← ignoreArg a i infos) then
+      pushArgsAux infos (i-1) f (todo.push tmpStar)
+    else
+      pushArgsAux infos (i-1) f (todo.push a)
+  | _, _, todo => return todo
+
+/--
+  Return true if `e` is one of the following
```
david-christiansen, line 115:
```suggestion
  Returns `true` if `e` is one of the following
```
Indicative mood

src/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,866 @@
...
+        return 0
+      else
+        failure
+    | _ => failure
+
+private def isNatType (e : Expr) : MetaM Bool :=
+  return (← whnf e).isConstOf ``Nat
+
+/--
+  Return true if `e` is one of the following
```
david-christiansen, line 161:
```suggestion
  Returns `true` if `e` is one of the following
```

src/Lean/Meta/LazyDiscrTree.lean
```diff
@@ -0,0 +1,866 @@
...
+  if fName == ``Nat.add && e.getAppNumArgs == 2 then
+    return isNumeral e.appArg!
+  else if fName == ``Add.add && e.getAppNumArgs == 4 then
+    if (← isNatType (e.getArg! 0)) then return isNumeral e.appArg! else return false
+  else if fName == ``HAdd.hAdd && e.getAppNumArgs == 6 then
+    if (← isNatType (e.getArg! 1)) then return isNumeral e.appArg! else return false
+  else
+    return fName == ``Nat.succ && e.getAppNumArgs == 1
+
+/--
```
david-christiansen, line 178:
```suggestion
/-
```
This is more internals docs than API docs, IMO

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+solving the current goal
+(subgoals are solved using `solveByElim`).
+
+```
+example : x < x + 1 := std_exact?%
+example : Nat := by std_exact?
+```
+
+These functions will likely lose their `std_` prefix once
+we are ready to replace the corresponding implementations in Mathlib.
```
david-christiansen, line 26:
````suggestion
This file defines tactics `exact?` and `apply?`,
(formerly known as `library_search`)
and a term elaborator `exact?%`
that tries to find a lemma
solving the current goal
(subgoals are solved using `solveByElim`).

```
example : x < x + 1 := exact?%
example : Nat := by exact?
```
````

File: src/Lean/Meta/Tactic/LibrarySearch.lean [hunk 18]
```diff
@@ -0,0 +1,397 @@
...
+-/
+
+
+namespace Lean.Meta.LibrarySearch
+
+open SolveByElim
+
+/-- Shortcut for calling `solveByElim`. -/
```
david-christiansen, line 34: How is this a shortcut for calling itself?

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+def solveByElim (required : List Expr) (exfalso : Bool) (goals : List MVarId) (maxDepth : Nat) := do
+  -- There is only a marginal decrease in performance for using the `symm` option for `solveByElim`.
+  -- (measured via `lake build && time lake env lean test/librarySearch.lean`).
```
david-christiansen, line 37: Where is this file?

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+    match type.getAppFnArgs with
+    | (``Iff, #[lhs, rhs]) => do
+      let tree ← addPath config tree rhs (name, DeclMod.mp)
+      let tree ← addPath config tree lhs (name, DeclMod.mpr)
+      return tree
+    | _ =>
+      return tree
+
+/--
+Constructs an discriminator tree from the current environment.
```
david-christiansen, line 95:
```suggestion
Constructs an discrimination tree from the current environment.
```

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+/--
+LibrarySearch has an extension mechanism for replacing the function used
+to find candidate lemmas.
+-/
+@[reducible]
+def CandidateFinder := Expr → MetaM (Array (Name × DeclMod))
+
+namespace DiscrTreeFinder
+
+/-- Add a path to a discrimination tree.-/
```
david-christiansen, line 72:
```suggestion
/-- Adds a path to a discrimination tree. -/
```

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+def buildImportCache (config : WhnfCoreConfig) : MetaM (DiscrTree (Name × DeclMod)) := do
+  let profilingName := "apply?: init cache"
+  -- Sort so lemmas with longest names come first.
+  let post (A : Array (Name × DeclMod)) :=
+        A.map (fun (n, m) => (n.toString.length, n, m)) |>.qsort (fun p q => p.1 > q.1) |>.map (·.2)
+  profileitM Exception profilingName (← getOptions) do
+    (·.mapArrays post) <$> (← getEnv).constants.map₁.foldM (init := {}) (updateTree config)
+
+/--
+Return matches from local constants.
```
david-christiansen, line 106:
```suggestion
Returns matches from local constants.
```

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+
+N.B. The efficiency of this could likely be considerably improved by caching in environment
+extension.
+-/
```
david-christiansen, line 110:
```suggestion
-/
/-
N.B. The efficiency of this could likely be considerably improved by caching in environment
extension.
-/
```

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+def localMatches (config : WhnfCoreConfig) (ty : Expr) : MetaM (Array (Name × DeclMod)) := do
+  let locals ← (← getEnv).constants.map₂.foldlM (init := {}) (DiscrTreeFinder.updateTree config)
+  pure <| (← locals.getMatch  ty config).reverse
+
+/--
+Candidate finding function that uses strict discrimination tree for resolution.
```
david-christiansen, line 116:
```suggestion
Candidate-finding function that uses a strict discrimination tree for resolution.
```

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+end DiscrTreeFinder
+
+namespace IncDiscrTreeFinder
+
+open LazyDiscrTree (InitEntry createImportedEnvironment)
+
+/--
+The maximum number of constants an individual task performed.
+
+The value below was picked because it roughly correponded to 50ms of work on the machine this was
```
david-christiansen, line 131:
```suggestion
The value was picked because it roughly correponded to 50ms of work on the machine this was
```
Because docstrings are mostly consumed outside the context of the source code

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+  pure <| (← importTree.getMatch ty config).reverse
+
+end DiscrTreeFinder
+
+namespace IncDiscrTreeFinder
+
+open LazyDiscrTree (InitEntry createImportedEnvironment)
+
+/--
+The maximum number of constants an individual task performed.
```
david-christiansen, line 129:
```suggestion
The maximum number of constants an individual task may perform.
```

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+    let a := #[e]
+    if e.key == .const ``Iff 2 then
+      let a := a.push (←e.mkSubEntry 0 (name, DeclMod.mp))
+      let a := a.push (←e.mkSubEntry 1 (name, DeclMod.mpr))
+      pure a
+    else
+      pure a
+
+/--
+Candidate finding function that uses strict discrimination tree for resolution.
```
david-christiansen, line 150:
```suggestion
Candidate-finding function that uses a strict discrimination tree for resolution.
```

But if I'm reading right (hard to do in GitHub), this should really read
```suggestion
Candidate-finding function that uses a lazy discrimination tree for resolution.
```

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+  let last :=
+        if x.size > n then
+          (x.extract n x.size).map f
+        else
+          (y.extract n y.size).map g
+  pure $ res ++ last
+
+
+/--
+An exception Id that indicates further speculation on candidate lemmas should stop
```
david-christiansen, line 233:
```suggestion
An exception ID that indicates further speculation on candidate lemmas should stop
```
We typically lowercase things in Lean APIS that would otherwise be all-caps, like "JSON" or "HTML", but the English should still use the all-caps

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+and current results returned.
```
david-christiansen, line 234:
```suggestion
and current results should be returned.
```

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+    let newType ← instantiateMVars  (← symmGoal.getType)
+    let l2 ← searchFn newType
+    let symmMCtx ← getMCtx
+    let symmGoalCtx := (symmGoal, symmMCtx)
+    setMCtx coreMCtx
+    pure $ interleaveWith (coreGoalCtx, ·) l1 (symmGoalCtx, ·) l2
+  else
+    pure $ l1.map (coreGoalCtx, ·)
+
+private def emoji (e:Except ε α) := if e.toBool then checkEmoji else crossEmoji
```
david-christiansen, line 277:
```suggestion
private def emoji (e : Except ε α) := if e.toBool then checkEmoji else crossEmoji
```

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+  | .none => pure lem
+  | .mp => mapForallTelescope (fun e => mkAppM ``Iff.mp #[e]) lem
+  | .mpr => mapForallTelescope (fun e => mkAppM ``Iff.mpr #[e]) lem
+
+/--
+Try applying the given lemma (with symmetry modifier) to the goal,
+then try to close subsequent goals using `solveByElim`.
+If `solveByElim` succeeds, we return `[]` as the list of new subgoals,
+otherwise the full list of subgoals.
+-/
```
david-christiansen, line 292: Indicative:
```suggestion
/--
Tries to apply the given lemma (with symmetry modifier) to the goal,
then tries to close subsequent goals using `solveByElim`.
If `solveByElim` succeeds, `[]` is returned as the list of new subgoals,
otherwise the full list of subgoals is returned.
-/
```

src/Lean/Meta/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,397 @@
...
+  let candidates ← librarySearchSymm searchFn goal
+  let cfg : ApplyConfig := { allowSynthFailures := true }
+  let act := fun cand => do
+      if ←shouldAbort then
+        abortSpeculation
+      librarySearchLemma cfg tactic allowFailure cand
+  tryOnEach act candidates
+
+/--
+Try to solve the goal either by:
```
david-christiansen, line 369:
```suggestion
Tries to solve the goal either by:
```

src/Lean/Util/Heartbeats.lean
```diff
@@ -0,0 +1,59 @@
...
+up basic definitions, most of which Lean already has "built in knowledge" about,
+so it is important that they be set up in exactly this way. (For example, Lean will
+use `PUnit` in the desugaring of `do` notation, or in the pattern match compiler.)
+
+-/
+
+namespace Lean
+
+/--
+Count the number of heartbeats used during a monadic function.
```
david-christiansen, line 22:
```suggestion
Counts the number of heartbeats used during a monadic function.
```

src/Lean/Util/Heartbeats.lean
```diff
@@ -0,0 +1,59 @@
...
+def withHeartbeats [Monad m] [MonadLiftT BaseIO m] (x : m α) : m (α × Nat) := do
+  let start ← IO.getNumHeartbeats
+  let r ← x
+  let finish ← IO.getNumHeartbeats
+  return (r, finish - start)
+
+/-- Return the current `maxHeartbeats`. -/
+def getMaxHeartbeats : CoreM Nat := do pure <| (← read).maxHeartbeats
+
+/-- Return the current `initHeartbeats`. -/
```
david-christiansen, line 38:
```suggestion
/-- Returns the current `initHeartbeats`. -/
```

src/Lean/Util/Heartbeats.lean
```diff
@@ -0,0 +1,59 @@
...
+def getInitHeartbeats : CoreM Nat := do pure <| (← read).initHeartbeats
+
+/-- Return the remaining heartbeats available in this computation. -/
```
david-christiansen, line 41:
```suggestion
/-- Returns the remaining heartbeats available in this computation. -/
```

src/Lean/Util/Heartbeats.lean
```diff
@@ -0,0 +1,59 @@
...
+def getRemainingHeartbeats : CoreM Nat := do
+  pure <| (← getMaxHeartbeats) - ((← IO.getNumHeartbeats) - (← getInitHeartbeats))
+
+/--
+Return the percentage of the max heartbeats allowed
```
david-christiansen, line 46:
```suggestion
Returns the percentage of the max heartbeats allowed
```

src/Init/Tactics.lean
joehendrix:
```suggestion
/--
Search environment for definitions or theorems that can refine the goal using `apply`
with conditions resolved when possible with `solve_by_elim`.

The optional `using` clause provides identifiers in the local context that must be
used when closing the goal.
-/
```

src/Lean/Elab/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,86 @@
...
+open Lean Meta LibrarySearch
+open Elab Tactic TryThis
+
+
+-- TODO: implement the additional options for `library_search` from Lean 3,
+-- in particular including additional lemmas
+-- with `std_exact? [X, Y, Z]` or `std_exact? with attr`.
+
+-- For now we only implement the basic functionality.
+-- The full syntax is recognized, but will produce a "Tactic has not been implemented" error.
```
joehendrix, line 21:
```suggestion
```

src/Lean/Elab/Tactic/LibrarySearch.lean
joehendrix:
```suggestion
      if suggestions.isEmpty then logError "exact?# didn't find any relevant lemmas"
```

joehendrix: This command also needs testing.

src/Lean/Elab/Tactic/LibrarySearch.lean
joehendrix: It happens to be `apply?` due to the `requresClose` check above.  I documented and clarified.

src/Lean/Elab/Tactic/LibrarySearch.lean
```diff
@@ -0,0 +1,82 @@
...
+      if suggestions.isEmpty then logError "apply? didn't find any relevant lemmas"
+      admitGoal goal
+
+@[builtin_tactic Lean.Parser.Tactic.exact?]
+def evalExact : Tactic := fun stx => do
+  let `(tactic| exact? $[using $[$required],*]?) := stx
+        | throwUnsupportedSyntax
+  exact? (← getRef) required true
+
+
```
joehendrix, line 56:
```suggestion
```

PR #3450, opened 2024-02-22
feat: omega uses b^(e+1) = b^e*b when b constant

bollu:
This is very helpful when dealing with bitvectors, where a case analysis on the bitwidth leaves one with hypotheses of the form `x<2^(Nat.succ w)`.

Design decisions I am unsure about:
- Is creating a helper `succ?` the correct way to match on the exponent `e+1`?
- I'm not certain why the prior call to `Int.ofNat_pow` also checked that the exponent was a ground natural. I removed this, since we now explicitly handle cases where the exponent is a term of the form `e+1`.


bollu:
awaiting-review

semorrison:
I would like support for:

```
example (x : Int) (e : Nat) : x % 2^e ≤ x % 2^(e+1) := by omega
```

semorrison:
> I would like support for:
> 
> ```
> example (x : Int) (e : Nat) : x % 2^e ≤ x % 2^(e+1) := by omega
> ```

Ah, sorry, I am wrong, this is actually quite a bit harder, and out of scope for this PR.

PR #3452, opened 2024-02-22
feat: add BitVec.ult/ule_ofNat

bollu:
This supersedes https://github.com/leanprover/std4/pull/664, and adds lemmas that mirror `ofNat_{add,sub}_ofNat` for ult, ule.


bollu:
awaiting-review

joehendrix:
I don't think either of these should be simp rules.  There is active work on integrating SAT solving into Lean, and in the ground case, I think it'd be useful to keep the Bitvec types around.

bollu:
They're no longer `simp` lemmas.

bollu:
awaiting-review

Kha:
@bollu @semorrison Is this still desirable?

PR #3457, opened 2024-02-22
feat: add `BitVec.getLsb_concat`

alexkeizer:
First (baby)-step to a `concat`-based `bitblast`: a characterization of `concat` in terms of `getLsb`.

The proof might benefit slightly from a `toNat_concat` lemma, but I wasn't sure what the normal form there should be, so I avoided it.

### DIFF
--- a/src/Init/Data/Bool.lean
+++ b/src/Init/Data/Bool.lean
@@ -223,10 +223,10 @@ theorem toNat_le_one (c:Bool) : c.toNat ≤ 1 := by
 theorem toNat_lt (b : Bool) : b.toNat < 2 :=
   Nat.lt_succ_of_le (toNat_le_one _)
 
-@[simp] theorem decide_toNat_eq_zero (b : Bool) : decide (b.toNat = 0) = !b := by
-  cases b <;> rfl
-@[simp] theorem decide_toNat_eq_one (b : Bool) : decide (b.toNat = 1) = b := by
-  cases b <;> rfl
+@[simp] theorem toNat_eq_zero (b : Bool) : b.toNat = 0 ↔ b = false := by
+  cases b <;> simp
+@[simp] theorem toNat_eq_one (b : Bool) : b.toNat = 1 ↔ b = true := by
+  cases b <;> simp
 
 end Bool
 


alexkeizer:
awaiting-review

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -421,6 +421,17 @@ theorem truncate_succ (x : BitVec w) :
     have j_lt : j.val < i := Nat.lt_of_le_of_ne (Nat.le_of_succ_le_succ j.isLt) j_eq
     simp [j_eq, j_lt]
 
+/-! ### concat -/
+
+@[simp] theorem getLsb_concat (x : BitVec w) (b : Bool) (i : Nat) :
+    (concat x b).getLsb i = match i with
+      | 0 => b
+      | i+1 => x.getLsb i := by
```
semorrison, line 429: I'm skeptical of this as a simp lemma with the match on the RHS.

The @[simp] lemmas should be for `getLsb_concat_zero` and `getLsb_concat_succ`.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -421,6 +421,17 @@ theorem truncate_succ (x : BitVec w) :
...
+  simp only [concat, getLsb, toNat_append, toNat_ofBool, Nat.testBit_or, Nat.shiftLeft_eq]
+  cases i
+  · simp [Nat.mod_eq_of_lt b.toNat_lt]
+  · simp [Nat.div_eq_of_lt b.toNat_lt]
```
semorrison, line 433:
```suggestion
theorem getLsb_concat (x : BitVec w) (b : Bool) (i : Nat) :
    (concat x b).getLsb i = if i = 0 then b else x.getLsb (i - 1) := by
  simp only [concat, getLsb, toNat_append, toNat_ofBool, Nat.testBit_or, Nat.shiftLeft_eq]
  cases i
  · simp [Nat.mod_eq_of_lt b.toNat_lt]
  · simp [Nat.div_eq_of_lt b.toNat_lt]

@[simp] theorem getLsb_concat_zero : (concat x b).getLsb 0 = b := by
  simp [getLsb_concat]

@[simp] theorem getLsb_concat_succ : (concat x b).getLsb (i + 1) = x.getLsb i := by
  simp [getLsb_concat]
```

src/Init/Data/Bool.lean
```diff
@@ -220,6 +220,14 @@ def toNat (b:Bool) : Nat := cond b 1 0
 theorem toNat_le_one (c:Bool) : c.toNat ≤ 1 := by
   cases c <;> trivial
 
+theorem toNat_lt (b : Bool) : b.toNat < 2 :=
+  Nat.lt_succ_of_le (toNat_le_one _)
```
semorrison, line 224: I would prefer consistent names here, but I think `toNat_lt` is correct, so will rename the other one.

src/Init/Data/Bool.lean
```diff
@@ -220,6 +220,14 @@ def toNat (b:Bool) : Nat := cond b 1 0
...
+
+@[simp] theorem decide_toNat_eq_zero (b : Bool) : decide (b.toNat = 0) = !b := by
+  cases b <;> rfl
+@[simp] theorem decide_toNat_eq_one (b : Bool) : decide (b.toNat = 1) = b := by
+  cases b <;> rfl
```
semorrison, line 229: These are more atomic:
```
@[simp] theorem toNat_eq_zero (b : Bool) : b.toNat = 0 ↔ b = false := by
  cases b <;> simp
@[simp] theorem toNat_eq_one (b : Bool) : b.toNat = 1 ↔ b = true := by
  cases b <;> simp
```

PR #3459, opened 2024-02-22
fix: `zetaDelta := false` regression

leodemoura:
See new test. It is a mwe for an issue blocking Mathlib.

### DIFF
--- a/tests/lean/structInstExtraEta.lean.expected.out
+++ b/tests/lean/structInstExtraEta.lean.expected.out
@@ -1,9 +1,9 @@
 def b : B :=
-let src := a;
-{ toA := src }
+let __src := a;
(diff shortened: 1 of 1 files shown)


kmill:
Just recording here that it turns out that it turns out mathlib was relying on the sources in structure instance notation introducing local instances.

We're testing out a temporary workaround to have the spread syntax introduce let bindings itself, using a special let binder that changes the kind to be an implementation detail after introducing any local instances. (Spread syntax is that `{...; __ := a; __ := b}` is equivalent to `{a, b with ...}`. It's nice because it works with `where` syntax.) https://github.com/leanprover-community/mathlib4/commit/77711ed99e5eabb25ab7ddc5a80f3816846df9a5

PR #3460, opened 2024-02-22
fix: auto-completion bugs and performance

### LINKED ISSUE #3228: Completion in def in namespace can add unnecessary namespace components
```lean
def A.B1.foo := 1
open A B1

namespace A
#check foo
def c2 : Nat := foo
```
Completion in the `def` but not in `#check` completes to `B1.foo`. Both the surrounding namespace and the nesting A.B1 seem to be necessary to reproduce.

mhuisi:
This PR addresses several performance issues in the auto-completion implementation. It also fixes a number of smaller bugs related to auto-completion.

In a file with `import Mathlib`, the performance of various kinds of completions has improved as follows:
- Completing `C`: 49000ms -> 1400ms
- Completing `Cat`: 14300ms -> 1000ms
- Completing `x.` for `x : Nat`: 3700ms -> 220ms
- Completing `.` for an expected type of `Nat`: 11000ms -> 180ms

The following bugs have been fixed as well:
- VS Code never used our custom completion order. Now, the server fuzzy completion score decides the order that completions appear in.
- Dot auto-completion for private types did not work at all. It does now.
- Completing `.<identifier>` (where the expected type is used to infer the namespace) did not filter by the expected type and instead displayed all matching constants in the respective namespace. Now, it uses the expected type for filtering. Note that this is not perfect because sub-namespaces are technically correct completions as well (e.g. `.Foo.foobar`). Implementing this is future work.
- Completing `.` was often not possible at all. Now, as long as the `.` is not used in a bracket (where it may be used for the anonymous lambda feature, e.g. `(. + 1)`), it triggers the correct completion.
-  Fixes #3228.
- The auto-completion in `#check` commands would always try to complete identifiers using the full declaration name (including namespaces) if it could be resolved. Now it simply uses the identifier itself in case users want to complete this identifier to another identifier.

## Details

Regarding completion performance, I have more ideas on how to improve it further in the future.

Other changes:
- The feature that completions with a matching expected type are sorted to the top of the server-side ordering was removed. This was never enabled in VS Code because it would use its own completion item order and when testing it I found it to be more confusing than useful.
- In the server-side ordering, we would always display keywords at the top of the list. They are now displayed according to their fuzzy match score as well.

The following approaches have been used to improve performance:
- Pretty-printing the type for every single completion made up a significant amount of the time needed to compute the completions. We now do not pretty-print the type for every single completion that is offered to the user anymore. Instead, the language server now supports `completionItem/resolve` requests to compute the type lazily when the user selects a completion item.
  - Note that we need to keep the amount of properties that we compute in a resolve request to a minimum. When the server receives the resolve request, the document state may have changed from the state it was in when the initial auto-completion request was received. LSP doesn't tell us when it will stop sending resolve requests, so we cannot keep this state around, as we would have to keep it around forever.
    LSP's solution for this dilemma is to have servers send all the state they need to compute a response to a resolve request to the client as part of the initial auto completion response (which then sends it back as part of the resolve request), but this is clearly infeasible for all real language servers where the amount of state needed to resolve a request is massive.
    This means that the only practical solution is to use the current state to compute a response to the resolve request, which may yield an incorrect result. This scenario can especially occur when using LiveShare where the document is edited by another person while cycling through available completions. 
- Request handlers can now specify a "header caching handler" that is called after elaborating the header of a file. Request handlers can use this caching handler to compute caches for information stored in the header. The auto-completion uses this to pre-compute non-blacklisted imported declarations, which in turn allow us to iterate only over non-blacklisted imported declarations where we would before iterate over all declarations in the environment. This is significant because blacklisted declarations make up about 4/5 of all declarations.
- Dot completion now looks up names modulo private prefixes to figure out whether a declaration is in the namespace of the type to the left of the dot instead of first stripping the private prefix from the name and then comparing it. This has the benefit that we do not need to scan the full name in most cases.

This PR also adds a couple of regression tests for fixed bugs, but *no benchmarks*. We will add these in the future when we add proper support for benchmarking server interaction sessions to our benchmarking architecture.

All tests that were broken by producing different completion output (empty `detail` field, added `sortText?` and `data?` fields) have been manually checked by me to be still correct before replacing their expected output.

### DIFF
--- a/src/Lean/Meta/CompletionName.lean
+++ b/src/Lean/Meta/CompletionName.lean
@@ -34,16 +34,16 @@ Checks whether a given name is internal due to something other than `_private`.
 Correctly deals with names like `_private.<SomeNamespace>.0.<SomeType>._sizeOf_1` in a private type
 `SomeType`, which `n.isInternal && !isPrivateName n` does not.
 -/
-private def isNonPrivateInternalName : Name → Bool
-  | n@(.str p s) => s.get 0 == '_' && n != privateHeader || isNonPrivateInternalName p
-  | .num p _ => isNonPrivateInternalName p
+private def isInternalNameModuloPrivate : Name → Bool
+  | n@(.str p s) => s.get 0 == '_' && n != privateHeader || isInternalNameModuloPrivate p
+  | .num p _ => isInternalNameModuloPrivate p
   | _       => false
 
 /--
 Return true if name is blacklisted for completion purposes.
 -/
 private def isBlacklisted (env : Environment) (declName : Name) : Bool :=
-  isNonPrivateInternalName declName
+  isInternalNameModuloPrivate declName
   || isAuxRecursor env declName
   || isNoConfusion env declName
   || isRecCore env declName
--- a/src/Lean/Server/Completion.lean
+++ b/src/Lean/Server/Completion.lean
@@ -39,19 +39,7 @@ for that item, again containing the `data?` field provided by the server.
 inductive CompletionIdentifier where
   | const (declName : Name)
   | fvar  (id       : FVarId)
-
-instance : ToJson CompletionIdentifier where
-  toJson
-    | .const declName => Json.mkObj [("const", toJson declName)]
-    | .fvar id        => Json.mkObj [("fvar", toJson id)]
-
-instance : FromJson CompletionIdentifier where
-  fromJson? j := do
-    if let .ok declName := j.getObjValAs? Name "const" then
-      return .const declName
-    if let .ok id := j.getObjValAs? FVarId "fvar" then
-      return .fvar id
-    .error "invalid CompletionItemData"
+  deriving FromJson, ToJson
 
 /--
 `CompletionItemData` that also contains a `CompletionIdentifier`.
--- a/src/Lean/Elab/BuiltinTerm.lean
+++ b/src/Lean/Elab/BuiltinTerm.lean
@@ -159,7 +159,7 @@ private def mkTacticMVar (type : Expr) (tacticCode : Syntax) : TermElabM Expr :=
   elabTerm stx[1] (mkNoImplicitLambdaAnnotation <$> expectedType?)
 
 @[builtin_term_elab Lean.Parser.Term.cdot] def elabBadCDot : TermElab := fun stx expectedType? => do
-  if let `(.) := stx then
+  if stx[0].getAtomVal == "." then
     -- Users may input bad cdots because they are trying to auto-complete them using the expected type
     addCompletionInfo <| CompletionInfo.dotId stx .anonymous (← getLCtx) expectedType?
   throwError "invalid occurrence of `·` notation, it must be surrounded by parentheses (e.g. `(· + 1)`)"
--- a/tests/lean/interactive/1265.lean.expected.out
+++ b/tests/lean/interactive/1265.lean.expected.out
@@ -8,7 +8,7 @@
    {"params":
     {"textDocument": {"uri": "file:///1265.lean"},
      "position": {"line": 0, "character": 51}},
-    "id": {"const": "Lean.TSyntax.expandInterpolatedStr"}}},
+    "id": {"const": {"declName": "Lean.TSyntax.expandInterpolatedStr"}}}},
   {"sortText": "1",
    "label": "getChar",
    "kind": 3,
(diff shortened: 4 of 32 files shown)


src/Lean/Server/Completion.lean
```diff
@@ -15,62 +15,141 @@ import Lean.Meta.Tactic.Apply
...
+    | .const declName => Json.mkObj [("const", toJson declName)]
+    | .fvar id        => Json.mkObj [("fvar", toJson id)]
+
+instance : FromJson CompletionIdentifier where
+  fromJson? j := do
+    if let .ok declName := j.getObjValAs? Name "const" then
+      return .const declName
+    if let .ok id := j.getObjValAs? FVarId "fvar" then
+      return .fvar id
+    .error "invalid CompletionItemData"
```
Kha, line 49: Don't we have `deriving` handlers for that?

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -609,18 +624,76 @@ partial def handleWaitForDiagnostics (p : WaitForDiagnosticsParams)
...
-  registerLspRequestHandler "textDocument/typeDefinition"       TextDocumentPositionParams (Array LocationLink)    (handleDefinition GoToKind.type)
-  registerLspRequestHandler "textDocument/documentHighlight"    DocumentHighlightParams    DocumentHighlightResult handleDocumentHighlight
-  registerLspRequestHandler "textDocument/documentSymbol"       DocumentSymbolParams       DocumentSymbolResult    handleDocumentSymbol
-  registerLspRequestHandler "textDocument/semanticTokens/full"  SemanticTokensParams       SemanticTokens          handleSemanticTokensFull
-  registerLspRequestHandler "textDocument/semanticTokens/range" SemanticTokensRangeParams  SemanticTokens          handleSemanticTokensRange
-  registerLspRequestHandler "textDocument/foldingRange"         FoldingRangeParams         (Array FoldingRange)    handleFoldingRange
-  registerLspRequestHandler "$/lean/plainGoal"                  PlainGoalParams            (Option PlainGoal)      handlePlainGoal
-  registerLspRequestHandler "$/lean/plainTermGoal"              PlainTermGoalParams        (Option PlainTermGoal)  handlePlainTermGoal
```
Kha, line 624: This makes me sad, though a code formatter would have done the same...
Actually, why are these explicit type parameters to begin with? With implicits I'd hope these could become one-liners again

mhuisi: I don't want to break this API just so that these handlers fit into one line. FWIW, I don't mind the explicit types.

src/Lean/Server/Completion.lean
mhuisi: I genuinely wasn't aware that it works for inductives now.

src/Lean/Meta/CompletionName.lean
```diff
@@ -29,11 +29,21 @@ builtin_initialize completionBlackListExt : TagDeclarationExtension ← mkTagDec
 def addToCompletionBlackList (env : Environment) (declName : Name) : Environment :=
   completionBlackListExt.tag env declName
 
+/--
+Checks whether a given name is internal due to something other than `_private`.
+Correctly deals with names like `_private.<SomeNamespace>.0.<SomeType>._sizeOf_1` in a private type
+`SomeType`, which `n.isInternal && !isPrivateName n` does not.
+-/
+private def isNonPrivateInternalName : Name → Bool
```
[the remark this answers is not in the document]
mhuisi: I don't think that name is much better, but I'm happy to use it instead.

src/Lean/Server/FileWorker/RequestHandling.lean
tydeu: As an alternative, in Alloy, I use [type class instances](https://github.com/tydeu/lean4-alloy/blob/5ae885f01218145ac2aac9bc2dd66b99dbb686d3/Alloy/Util/Server/Methods.lean) to infer types for methods. That allows me to keep the registrations on [one line each](https://github.com/tydeu/lean4-alloy/blob/5ae885f01218145ac2aac9bc2dd66b99dbb686d3/Alloy/C/Server.lean#L13-L20) and even automatically infer types when [a handler is provided as a lambda](https://github.com/tydeu/lean4-alloy/blob/5ae885f01218145ac2aac9bc2dd66b99dbb686d3/Alloy/C/Server/Worker.lean#L114-L120).

Kha: > I don't want to break this API

Reasonable

PR #3463, opened 2024-02-22
chore: CI: flag `Lean` modules not using `prelude`

Kha:


Kha:
should probably be moved into a separate file

PR #3473, opened 2024-02-23
fix: `isDefEq` handling of delayed-assigned mvars when `withAssignableSyntheticOpaque` is `true`

thorimur:
(In progress; fully experimental.)


thorimur:
WIP

PR #3474, opened 2024-02-23
feat: `BitVec` int lemmas

joehendrix:
This introduces lemma support for BitVec.ofInt/BitVec.toInt as well as lemmas upstreamed from Std and Mathlib for reasoning about emod and bmod.

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -154,7 +154,8 @@ theorem toNat_ge_of_msb_true {x : BitVec n} (p : BitVec.msb x = true) : x.toNat 
     simp [BitVec.msb, BitVec.getMsb] at p
   | n + 1 =>
     simp [BitVec.msb_eq_decide] at p
-    omega
+    simp only [Nat.add_sub_cancel]
+    exact p
 
 /-! ### cast -/
 


src/Init/Data/Int/DivMod.lean
```diff
@@ -158,4 +158,38 @@ instance : Div Int where
...
+Balanced mod (and balanced div) are a division and modulus pair such
+that `b * (Int.bdiv a b) + Int.bmod a b = a` and `b/2 ≤ Int.bmod a b <
+b/2` for all `a : Int` and `b > 0`.
+
+This is used in Omega as well as signed bitvectors.
+-/
+
+/--
+Balanced mod, taking values in the range [- m/2, (m - 1)/2].
+-/
```
semorrison, line 173: Needs rewriting!

src/Init/Data/Int/Basic.lean
```diff
@@ -325,6 +325,19 @@ protected def pow (m : Int) : Nat → Int
 instance : HPow Int Nat Int where
   hPow := Int.pow
 
+protected
+theorem pow_succ (b : Int) (n : Nat) : b ^ (n+1) = b^n * b := by rfl
```
semorrison, line 329:
```suggestion
protected theorem pow_succ (b : Int) (n : Nat) : b ^ (n+1) = b^n * b := by rfl
```

src/Init/Data/Int/Basic.lean
```diff
@@ -325,6 +325,19 @@ protected def pow (m : Int) : Nat → Int
...
+
+theorem powNatCast (b n : Nat) : (@Nat.cast Int _ b) ^ n = (b^n : Nat) := by
+  match n with
+  | 0 => rfl
+  | n + 1 =>
+    simp only [Int.pow_succ, powNatCast b n, Nat.pow_succ]
+    rfl
+
+theorem powOfNat (b n : Nat) : (@OfNat.ofNat Int b _) ^ n = (b^n : Nat) :=
+  powNatCast b n
```
semorrison, line 339: I think this should be called `ofNat_pow`?

src/Init/Data/Int/Basic.lean
```diff
@@ -325,6 +325,19 @@ protected def pow (m : Int) : Nat → Int
 instance : HPow Int Nat Int where
   hPow := Int.pow
 
+protected
+theorem pow_succ (b : Int) (n : Nat) : b ^ (n+1) = b^n * b := by rfl
+
+theorem powNatCast (b n : Nat) : (@Nat.cast Int _ b) ^ n = (b^n : Nat) := by
```
semorrison, line 331: and this one `nat_cast_pow`?

src/Init/Data/Int/Basic.lean
semorrison: Hmm, actually not sure, I think a lot of our `ofNat_X` lemmas in the `Int` namespace need renaming, and probably should be `nat_cast_X` or `cast_X`.

src/Init/Data/Int/Basic.lean
semorrison: pow_succ has also been introduced in a later file in the meantime! (I'm inclined to believe it belongs later than here, this file should only have the minimal theorems needed for definitions.)

semorrison:
I pushed some very minor changes to proofs. Looks good to me otherwise.

PR title should be `feat: ...`

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -148,6 +148,15 @@ theorem msb_eq_getLsb_last (x : BitVec w) :
...
 
+theorem toNat_ge_of_msb_true {x : BitVec n} (p : BitVec.msb x = true) : x.toNat ≥ 2^(n-1) := by
+  match n with
+  | 0 =>
+    simp [BitVec.msb, BitVec.getMsb] at p
+  | n + 1 =>
+    simp [BitVec.msb_eq_decide] at p
+    omega
+
+
```
joehendrix, line 159:
```suggestion
```

PR #3480, opened 2024-02-23
feat: generalize msb_eq_decide to also handle the zero width case

bollu:
Note that this is a strict generalization of the previous statemens of `getLsb_last` and `msb_eq_decide` that worked for bitwidths `>= 1`.

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -147,6 +147,9 @@ theorem msb_eq_getLsb_last (x : BitVec w) :
       · have : BitVec.toNat x < 2^w + 2^w := by simpa [Nat.pow_succ, Nat.mul_two] using x.isLt
         omega
 
+@[bv_toNat] theorem getLsb_succ_last (x : BitVec (w + 1)) :
+    x.getLsb w = decide (2 ^ w ≤ x.toNat) := getLsb_last x
+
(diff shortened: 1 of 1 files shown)


src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -133,19 +133,21 @@ theorem msb_eq_getLsb_last (x : BitVec w) :
   · simp [BitVec.eq_nil x]
   · simp
 
-@[bv_toNat] theorem getLsb_last (x : BitVec (w + 1)) :
-    x.getLsb w = decide (2 ^ w ≤ x.toNat) := by
-  simp only [Nat.zero_lt_succ, decide_True, getLsb, Nat.testBit, Nat.succ_sub_succ_eq_sub,
+@[bv_toNat] theorem getLsb_last (x : BitVec w) :
+    x.getLsb (w-1) = decide (2 ^ (w-1) ≤ x.toNat) := by
```
semorrison, line 137: Hmm, while this is mathematically more general, it is syntactically not. Can we keep the original statement, maybe named `getLsb_succ_last`, as a corollary of the new statement?

bollu: I've added the other lemma tagged with `bv_toNat` as requested. Luckily, this is defeq, as `(w - 1) + 1 =def= w`, it's cheap to add.

bollu:
awaiting-review

PR #3487, opened 2024-02-24
feat: add lemmas about `BitVec.concat` and bitwise ops

alexkeizer:
Show how the various bitwise ops (`and`, `or`, `not`, and `xor`) distribute over `concat`.

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -467,18 +467,18 @@ theorem getLsb_concat (x : BitVec w) (b : Bool) (i : Nat) :
 @[simp] theorem getLsb_concat_succ : (concat x b).getLsb (i + 1) = x.getLsb i := by
   simp [getLsb_concat]
 
-theorem not_concat (x : BitVec w) (b : Bool) : ~~~(concat x b) = concat (~~~x) !b := by
+@[simp] theorem not_concat (x : BitVec w) (b : Bool) : ~~~(concat x b) = concat (~~~x) !b := by
   ext i; cases i using Fin.succRecOn <;> simp [*, Nat.succ_lt_succ]
 
-theorem concat_or_concat (x y : BitVec w) (a b : Bool) :
+@[simp] theorem concat_or_concat (x y : BitVec w) (a b : Bool) :
     (concat x a) ||| (concat y b) = concat (x ||| y) (a || b) := by
   ext i; cases i using Fin.succRecOn <;> simp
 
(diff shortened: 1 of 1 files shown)


semorrison:
> What is the thinking behind this orientation of the rules?

I think this is the right direction for bitblasting via `concat`. (I still expect it will be easier to bitblast via `forall` / `getLsb`).

You repeatedly:
* use cases to replace BitVec variables with `concat`
* use lemmas like these to push operations into the concat
* use `congr` to get a goal about smaller bitvecs, and a boolean goal.

Notes:
* This approach to bitblasting may be slow, because `cases` is expensive compared to using the simplifier.
* Surely `cons` is more useful than `concat` for bitblasting? Maybe this is not obvious. In any case, it would be nice to develop the cons lemmas in parallel to the concat lemmas.

joehendrix:
I also realized when I started looking at this, I initially scanned `concat` as `append`.  I think I should have stuck with the `push` name in retrospect.

If this is mainly used for bitblasting, perhaps it should belong to a dedicated simpset.  One question with these lemmas is what to do about terms like `concat x a = y`.

Is it advantageous to simplify, or is this a normal form?

alexkeizer:
> What is the thinking behind this orientation of the rules?

The main motivation is indeed bitblasting, but it also seems generally applicable. The idea being that the lemma "unfolds" the bitwise operation for a single bit, so that the effect of each operation on the concatenated bit becomes just the Boolean operation. I don't feel super strongly about it, though, so would be happy to put it in a separate simpset, say `bv_concat`?

> Surely cons is more useful than concat for bitblasting? Maybe this is not obvious. In any case, it would be nice to develop the cons lemmas in parallel to the concat lemmas.

The new bit added by `concat` is the least significant bit. I think of bitwise addition as operating from the least significant to the most significant bit, so bitblasting via `concat` seemed the natural way to do it. Nonetheless, we should have the `cons` lemmas as well. I'll open a PR for those after we're agreed on the direction of the current lemmas, and whether to make them part of the default simp-set, so I can mirror those choices for the `cons`-lemmas.

> One question with these lemmas is what to do about terms like concat x a = y.
> Is it advantageous to simplify, or is this a normal form?

Not sure how we'd simplify that. 
I'd say we let that be the normal form; further simplification can be done by doing `cases y` first

semorrison:
Interesting: I think it's wrong to think about addition as operating from lsb to msb: because you need to know the carry term, the natural lemmas run from the msb down. In any case, that is beyond the scope of this PR.

I think the direction of these simp lemmas is good. Let's keep them in the main simp set for now.

PR #3489, opened 2024-02-24
fix: make omission syntax a builtin syntax

### LINKED ISSUE #3476: [error when printing message: unknown constant '«term⋯»'] when working on Init
### Description

when working on file `Init/Data/Array/Basic.lean` in core, I often can’t see the proof state because of
```
[error when printing message: unknown constant '«term⋯»'] 
```

I assume this is related to @kmill’s https://github.com/leanprover/lean4/pull/3241 and https://github.com/leanprover/lean4/pull/3201.

Indeed with
```
set_option pp.proofs.threshold 10000
```
I see the goals.

Not sure if this is something we just have to live with when working on `Init`, or if it can be improved.

### Versions

89490f648a6f41c3f43724a52f6ac61c6072eb4a

### Impact

kmill:
When editing core Lean, the `pp.proofs` feature causes goal states to fail to display in the Infoview, instead showing only "error when printing message: unknown constant '«term⋯»'". This PR moves the `⋯` syntax from Init.NotationExtra to Lean.Elab.BuiltinTerm

It also makes it so that `⋯` elaborates as `_` while logging a warning, rather than throwing an error, which should be somewhat more friendly when copy/pasting from the Infoview.

Closes #3476

### DIFF
--- a/src/Init/NotationExtra.lean
+++ b/src/Init/NotationExtra.lean
@@ -181,7 +181,7 @@ syntax (name := calcTactic) "calc" calcSteps : tactic
     match tail with
     | `([])      => `([$x])
     | `([$xs,*]) => `([$x, $xs,*])
-    --| `(⋯)       => `([$x, $tail]) -- Unexpands to `[x, y, z, ⋯]` for `⋯ : List α`
+    | `(⋯)       => `([$x, $tail]) -- Unexpands to `[x, y, z, ⋯]` for `⋯ : List α`
     | _          => throw ()
   | _ => throw ()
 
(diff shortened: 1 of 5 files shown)


Kha:
#3491 fixes the update-stage0 issue

kmill:
Thanks @Kha for fixing that! I was getting deep into the boostrapping configuration trying to figure out what I was doing wrong.

Kha (approved):
Looks good but for a PR containing an update-stage0, you will need to make sure that every commit follows the convention as we'll rebase-merge it

kmill:
@Kha Done

PR #3492, opened 2024-02-24
feat: add intMax

bollu:


bollu:
awaiting-review

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -599,4 +599,20 @@ protected theorem lt_of_le_ne (x y : BitVec n) (h1 : x <= y) (h2 : ¬ x = y) : x
   simp
   exact Nat.lt_of_le_of_ne
 
+/- ! ### intMax -/
+
+/-- A bitvector when interpreted as an integer is the least value: -2^(w-1) when w > 0-/
+def intMax (w : Nat) : BitVec w  := (2^w - 1)#w
+
+theorem intMax_getLsb_eq (w : Nat) : (intMax w).getLsb i = decide (i < w) := by
```
semorrison, line 607: Should be `getLsb_intMax`, also below.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -599,4 +599,20 @@ protected theorem lt_of_le_ne (x y : BitVec n) (h1 : x <= y) (h2 : ¬ x = y) : x
   simp
   exact Nat.lt_of_le_of_ne
 
+/- ! ### intMax -/
+
+/-- A bitvector when interpreted as an integer is the least value: -2^(w-1) when w > 0-/
```
semorrison, line 604:
```suggestion
/-- A bitvector when interpreted as an integer is the least value: -2^(w-1) when w > 0-/
```
Garbled?

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -599,4 +599,20 @@ protected theorem lt_of_le_ne (x y : BitVec n) (h1 : x <= y) (h2 : ¬ x = y) : x
...
+def intMax (w : Nat) : BitVec w  := (2^w - 1)#w
+
+theorem intMax_getLsb_eq (w : Nat) : (intMax w).getLsb i = decide (i < w) := by
+  simp [intMax, getLsb]
+
+theorem intMax_toNat_eq : (intMax w).toNat = if w = 0 then 0 else 2^w - 1 := by
```
semorrison, line 610: No `if` required, 2^w - 1 is correct always.

src/Init/Data/BitVec/Lemmas.lean
bollu: Thanks, fixed.

bollu:
awaiting-review

src/Init/Data/BitVec/Lemmas.lean
bollu: done.

joehendrix (approved):

PR #3494, opened 2024-02-25
feat: syntax and delaborator for delayed assignment metavariables

kmill:
It's often confusing how `fun x y => ?foo` elaborates as `fun x y => ?m.123 x y`, with no way to see how `?m.123` is connected to `?foo`. This PR adds syntax for delayed assignment metavariables, where the elaborated result pretty prints as `fun x y => ?foo(x, y)` instead. Hovering over the expression still shows the underlying `?m.123 x y` term, and hovering over `foo` shows whatever `foo` might be currently assigned to, if anything. The syntax can be turned off with `set_option pp.mvars.delayed false`.

This also adds an elaborator for `?foo(x := a, y := b)` syntax to create a new delayed assignment metavariable from a metavariable, making the pretty printed syntax elaboratable. In addition to being for making pretty printing round trip, it is filling in a missing feature, since, as a user, it's impossible to use synthetic holes in incompatible local contexts without being able to specify delayed assignments. For example, with the `?foo` above we can then later refer to `fun x y => ?foo(x, y)` (where `?foo(x, y)` is short for `?foo(x := x, y := y)`), even though the `x` and `y` binders are totally new fvars.

The syntax is chosen to (1) use parentheses with no whitespace since that is reserved for new syntax, (2) to resemble structure instance notation, and (3) to be something like the Lean version of lambda calculus substitution (and in any case, square brackets are already used by GetElem).

### DIFF
--- a/src/Lean/Elab/BuiltinTerm.lean
+++ b/src/Lean/Elab/BuiltinTerm.lean
@@ -100,6 +100,41 @@ private def elabOptLevel (stx : Syntax) : TermElabM Level :=
         else
           throwError "synthetic hole has already been defined with an incompatible local context"
 
+@[builtin_term_elab «delayedAssignedMVar»] def elabDelayedAssignedMVar : TermElab := fun stx expectedType? =>
+  match stx with
+  | `(?$arg($[$idents $[:= $terms]?],*)) => do
+    let mvar ←
+      if arg.raw.isIdent then
+        let userName := arg.raw.getId
+        match (← getMCtx).findUserName? userName with
+        | some mvarId => pure <| Expr.mvar mvarId
+        | none        => mkFreshExprMVar expectedType? .syntheticOpaque userName
+      else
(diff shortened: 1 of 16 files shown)


ericrbg-harmonic:
Will we be able to limit the scope of the metavariable in this way? e.g. `?foo(x)` from your example above

kmill:
@ericrbg-harmonic It lets you write `?foo(x)` to assign just a value for `x`, but to be clear `y` is still in its context and you can only do this if `?foo`'s `y` is still a local variable in the current scope.

digama0:
I am concerned about the size of this expression, as I often see these metavariables in printouts and I generally don't actually care about the values of the arguments and only the fact that it is a metavariable (possibly an unassignable metavariable) is relevant to me. Maybe `?m(...)` should be printed instead under some pp option? Maybe it should also be default-enabled?

PR #3500, opened 2024-02-26
fix: C++ exceptions across shared libraries on Linux

Kha:
Server interruptions in C++ started to fail after #3421


Kha:
Manual test successful

PR #3531, opened 2024-02-28
fix: `get_elem_tactic_trivial` regression

leodemoura:


src/Init/Tactics.lean
```diff
@@ -1430,9 +1430,25 @@ where `i < arr.size` is in the context) and `simp_arith` and `omega`
...
+[simp] theorem getElem_pop (a : Array α) (i : Nat) (hi : i < a.pop.size) :
+    a.pop[i] = a[i]'(Nat.lt_of_lt_of_le (a.size_pop ▸ hi) (Nat.sub_le _ _)) :=
+```
+There is a proof embedded in the right-hand-side, and we want it to be just `hi`.
+If `omega` is used to "fill" this proof, we will have a more complex proof term that
+cannot be inferred by unification.
+Note that users can accidentaly break this IF they add new `macro_rules` for `get_elem_tactic_trivial`.
+
+TODO: Implement priorities for `macro_rules`.
+TODO: Ensure we have a **high-priority** macro_rules for `get_elem_tactic_trivial` which is just `assumption`.
```
nomeata, line 1447: There are more rules added to `get_elem_tactic_trivial` elsewhere…

Until we have priorities, should we make it a two-stage process, where `get_elem_tactic_aux` is either `assumption` or `get_elem_tactic_trivial`, so that `assumption` really happens first reliably?

Or we simply add `assumption` to `get_elem_tactic` below

leodemoura: Nice. I pushed this change.

PR #3552, opened 2024-03-01
fix: load references asynchronously

mhuisi:
In v4.6.0, there was a significant regression in initial server startup performance because the .ilean files got bigger in #3082 and we load the information stored in all .ilean files synchronously when the server starts up. 

This PR makes this loading asynchronous. The trade-off is that requests that are issued right after the initial server start when the references are not fully loaded yet may yield incomplete results.

Benchmark for this in a separate PR soon after this one.

### DIFF
--- a/src/Lean/Server/Watchdog.lean
+++ b/src/Lean/Server/Watchdog.lean
@@ -904,7 +904,7 @@ def findWorkerPath : IO System.FilePath := do
 /--
 Starts loading .ileans present in the search path asynchronously in an IO task.
 This ensures that server startup is not blocked by loading the .ileans.
-In return, while the .ileans are being loaded, users will omly get incomplete
+In return, while the .ileans are being loaded, users will only get incomplete
 results in requests that need references.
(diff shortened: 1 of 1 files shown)


src/Lean/Server/Watchdog.lean
```diff
@@ -901,23 +901,33 @@ def findWorkerPath : IO System.FilePath := do
...
-    catch _ =>
-      -- could be a race with the build system, for example
-      -- ilean load errors should not be fatal, but we *should* log them
-      -- when we add logging to the server
-      pure ()
-  return refs
+/--
+Starts loading .ileans present in the search path asynchronously in an IO task.
+This ensures that server startup is not blocked by loading the .ileans.
+In return, while the .ileans are being loaded, users will omly get incomplete
```
Kha, line 907:
```suggestion
In return, while the .ileans are being loaded, users will only get incomplete
```

Kha (approved):

PR #3553, opened 2024-03-01
fix: `rename_i` in macro

Kha:
Correct comparison of lexical scopes


leodemoura:
@Kha We need a `rebase` here.

PR #3560, opened 2024-03-01
doc: small improvements to docstrings for `let` and `have` tactics

kmill:


david-christiansen (approved):

kmill:
@david-christiansen I did a little bit of work on the docstrings for `have` and `let` here. It would be good to have a standard for how a tactic docstring should be organized, and how I've imagined it is

```quote
Short high-level description of tactic

* syntax variant one
* syntax variant two

Other brief notes you should know about the tactic, like related tactics.

# Examples

some illuminating examples

# Limitations

explanations of what you shouldn't expect from this tactic

# See also

other things to look into
```

That's just a quick idea jotted down, and really I've got the Mathematica documentation style in mind, which is worth emulating.

PR #3562, opened 2024-03-02
feat: generalize simp normal form testing code and validate `Nat`/`Int` normal forms

joehendrix:
This is still a work in progress.


src/Init/Data/Int/Lemmas.lean
```diff
@@ -14,6 +14,7 @@ open Nat
 
 /-! ## Definitions of basic functions -/
 
+
```
joehendrix, line 17:
```suggestion
```

Kha:
Assuming out of date

Issue #3567, opened 2024-03-02
RFC: lean/lake should continue building files downstream of an elaboration error


digama0:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to mathlib4 or std4.

### Description

When performing fixes to mathlib or other large projects after a change to lean or other foundational library, it can be inconvenient to have to iterate several times with CI or a local build finding errors, fixing them, and then finding new errors in downstream files. Lean will already report multiple errors when they occur in a single file, but the cross file case is currently not handled, and fixing this requires changes to both lean and lake.

### Context

This has come up many times, but the nightly fixes for this month's release have been particularly susceptible to this because of many back-and-forths with new lean changes that fix and break many things in mathlib. (I suggested this briefly [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/tryAtEachStep.3A.20run.20a.20tactic.20on.20every.20step.20in.20your.20proofs/near/424372118) and it appeared well received; I expect that the team would be on board with this general direction, although there are a few design decisions to be made about how to surface it and what to make the default behavior.)

### Additional Information

The lean side of this is easy to implement; for the lake side requires a bit of discussion on how best to continue after an error without stopping the build or swallowing the error silently.

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

semorrison:
Yes, some implementation of this would be super useful when getting Mathlib building on new versions.

PR #3568, opened 2024-03-02
feat: `--continue-on-error` to produce .olean file even if elaboration fails

digama0:
Implementation of the lean side of #3567. Passing `--continue-on-error` to lean will cause it to produce an `.olean` file even if the file had an error, but it will still return error code 1 so lake will still not continue after it.

@tydeu How do you think lake should handle this? The issue is that because lean returns a nonzero error code (and I think it would be misleading for it not to), lake will immediately stop the running thread (`compileO` calls `proc` which calls `error`) and all dependent targets will also fail (IIUC). Should `proc` take a `continueOnError` flag?

### DIFF
--- a/src/util/shell.cpp
+++ b/src/util/shell.cpp
@@ -220,6 +220,7 @@ static void display_help(std::ostream & out) {
     std::cout << "  --deps             just print dependencies of a Lean input\n";
     std::cout << "  --print-prefix     print the installation prefix for Lean and exit\n";
     std::cout << "  --print-libdir     print the installation directory for Lean's built-in libraries and exit\n";
+    std::cout << "  --continue-on-error create olean file even if elaboration produces errors\n";
     std::cout << "  --profile          display elaboration/type checking time for each definition/theorem\n";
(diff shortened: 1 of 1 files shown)


digama0:
> One other issue to consider is how partial build should be factor into Lake's traces. That is, module should be rebuilt in the future, so Lake somehow needs to invalidate the trace such it that it will trigger a rebuild.

I actually disagree with this to some extent: It should be fine for lake to cache the failure and assume that re-elaborating the file assuming up to date inputs should produce exactly the same failure. The only complication is that lake would need to buffer and replay the output instead of re-running the compiler. (This is also what `cargo` does when you re-run a failed build without any source changes.) Changes that would make the build succeed should already be handled by the usual invalidation mechanisms - either the file or its dependencies should need to be changed, and then it will be rebuilt as normal.

Kha:
> The only complication is that lake would need to buffer and replay the output instead of re-running the compiler.

To some lesser degree this is already desirable for non-error diagnostics in successful builds

Kha:
drafting pending RFC acceptance

PR #3573, opened 2024-03-02
fix: `simp?` suggests generated equations lemma names

### LINKED ISSUE #3547: simp? suggests generated equations lemma names
Consider this:

```
import Std

def foo : Nat → Nat 
  | 0 => 0
  | n+1 => foo n
decreasing_by decreasing_tactic

def foo2 : foo 2 = 0 := by
  simp? [foo]
  -- suggests: simp only [foo._eq_2, foo._eq_1]
  -- but that yields: unknown constant 'foo._eq_2'
```

These `foo._eq_6` lemmas  are created on demand when you use `simp [foo]` or `rw [foo]`; from then on they are available. But if you just mention them manually like this, they may not exist yet.

Therefore, `simp?` should not suggest them. Instead it should suggest `simp only [foo]`.

### Versions

4.6.0  on live.lean-lang.org

### Impact

leodemoura:
closes #3547

### DIFF
--- a/src/Lean/Meta/Tactic/Simp/Types.lean
+++ b/src/Lean/Meta/Tactic/Simp/Types.lean
@@ -259,7 +259,7 @@ def getSimpCongrTheorems : SimpM SimpCongrTheorems :=
 
 def recordSimpTheorem (thmId : Origin) : SimpM Unit := do
   /-
-  If `thmId` is an equational theorem (e.g., `foo._eq_1`), we shound record `foo` instead.
+  If `thmId` is an equational theorem (e.g., `foo._eq_1`), we should record `foo` instead.
   See issue #3547.
   -/
   let thmId ← match thmId with


src/Lean/Meta/Tactic/Simp/Types.lean
```diff
@@ -256,7 +257,22 @@ def getSimpCongrTheorems : SimpM SimpCongrTheorems :=
 @[inline] def withDischarger (discharge? : Expr → SimpM (Option Expr)) (x : SimpM α) : SimpM α :=
   savingCache <| withReader (fun r => { MethodsRef.toMethods r with discharge? }.toMethodsRef) x
 
-def recordSimpTheorem (thmId : Origin) : SimpM Unit :=
+def recordSimpTheorem (thmId : Origin) : SimpM Unit := do
+  /-
+  If `thmId` is an equational theorem (e.g., `foo._eq_1`), we shound record `foo` instead.
```
nomeata, line 262:
```suggestion
  If `thmId` is an equational theorem (e.g., `foo._eq_1`), we should record `foo` instead.
```

Issue #3458, opened 2024-02-22
Incorrect promotion from index to paramater


arthur-adjedj:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to mathlib4 or std4.

### Description

Some inductive types with indices get their indices promoted to parameters in cases where it shouldn't, leading to kernel-level errors afterwards

### Context

The issue was first discovered [here](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/bug.3A.20nonsensical.20disapperarence.20of.20kernel.20error.20message)

### Steps to Reproduce
MWE:
```lean
inductive All {I : Type u} (P : I → Type v) : List I → Type (max u v) where
  | cons : P x → All P xs → All P (x :: xs)

inductive Iμ {I : Type u}  : I → Type (max u v) where
  | mk : (i : I)  → All Iμ [] → Iμ i
```

**Expected behavior:** No error

**Actual behavior:** The index of `Iμ` gets promoted to a parameter during elaboration, leading to the following error:
```
arg #6 of '_nested.All_1.cons' contains a non valid occurrence
```

### Versions

nightly-2024-02-21
Ubuntu

### Additional Information

While there are some workarounds to avoid the promotion of indices to parameters, Lean would benefit from having an option to turn that feature off altogether.

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

PR #3590, opened 2024-03-04
feat: add `inductive.autoPromoteIndices` option

arthur-adjedj:
This PR partly addresses #3458, by adding an option `autoPromoteIndices` to turn off the promotion of fixed indices to parameters. The actual fix for the issue is in a separate PR #3591.

Because nested inductive datatypes parameters cannot contain local variables, it is often desirable for a fixed index to not be promoted, as to allow free variables in that place. See example in `3458_1.lean`


arthur-adjedj:
awaiting-review

PR #3594, opened 2024-03-04
doc: Don't say "Reference implementation" in doc strings

girving:
The docstring is for the user of the function, who doesn't need to care that it is a reference implementation.  Instead, make the "reference implementation" bit a comment in the body of the reference implementation.


digama0:
Hm, I was hoping you would put actual documentation on the functions instead...

girving:
I'm happy to just remove the comments, but I don't want to write detailed replacement doc strings at the moment.  Let me know if the former is sufficient; otherwise I can retract this one.

semorrison:
Yeah, doesn't seem like a positive change.

PR #3612, opened 2024-03-05
chore: response file to avoid arg limits in lean static lib build

tydeu:


Kha:
refined version at https://github.com/leanprover/lean4/pull/3601/commits/47da35d7ce01498651300f138899302b12a38970

PR #3616, opened 2024-03-06
feat: better support for reducing `Nat.rec`

### LINKED ISSUE #3022: `Nat.add n 0` expressions appear when unfolding with `simp` (but not with `rw`)
### Description

When using `unfold` or `simp` to unfold a recursive function, the expression `Nat.add n 0` appears inside the body of the unfolded function rather than `n`. In the following example, you can uncomment the `unfold` or `simp` lines to see the given goal state. In contrast, `rw` results in the wanted state.

```lean
def foo : Nat → Nat
  | 0 => 2
  | n + 1 => foo n

example (n : Nat) : foo (n + 1) = 2 := by
  --unfold foo      -- ⊢ foo (Nat.add n 0) = 2
  --simp only [foo] -- ⊢ foo (Nat.add n 0) = 2
  --rw [foo]        -- ⊢ foo n = 2
  sorry
```

**Expected behavior:** In each of these three commented out lines, the goal state becomes `⊢ foo n = 2`.

**Actual behavior:** Only the `rw` tactic results in the expected goal state.

### Context

When using a custom eliminator in the `induction` or `cases` tactic for `Nat` that uses `0` and `n + 1` rather than `Nat.zero` and `Nat.succ n`, expressions such as `foo (n + 1)` occur. If this is a proof about `foo`, then it is natural to unfold it. However, using `unfold` or `simp` results in expressions with `Nat.add n 0`. Eliminating these requires further application of lemmas such as `Nat.add_eq` and `add_zero` to clean up the expression.

```lean
protected def Nat.recAux {motive : Nat → Sort _}
    (zero : motive 0) (succ : ∀ n, motive n → motive (n+1)) : (t : Nat) → motive t
  | 0 => zero
  | _+1 => succ _ (Nat.recAux zero succ _)

example (n : Nat) : foo n = 2 := by
  induction n using Nat.recAux with
  | zero => rfl
  | succ n ih =>
    --unfold foo      -- ⊢ foo (Nat.add n 0) = 2
    --simp only [foo] -- ⊢ foo (Nat.add n 0) = 2
    --rw [foo]        -- ⊢ foo n = 2
    exact ih
```

This is an obstruction to the "beautiful `induction` for `Nat`" project. We do not want users to see `Nat.add` expressions.

### Versions

MacOS arm64
```
#eval Lean.versionString -- "4.4.0-rc1"
```

### Additional Information

### Impact

leodemoura:
closes #3022

With this commit, given the declaration
```
def foo : Nat → Nat
  | 0 => 2
  | n + 1 => foo n
```
when we unfold `foo (n+1)`, we now obtain `foo n` instead of `foo (Nat.add n 0)`.

### DIFF
--- a/src/Lean/Meta/CtorRecognizer.lean
+++ b/src/Lean/Meta/CtorRecognizer.lean
@@ -5,6 +5,7 @@ Authors: Leonardo de Moura
 -/
 prelude
 import Lean.Meta.LitValues
+import Lean.Meta.Offset
 
 namespace Lean.Meta
 
@@ -64,9 +65,17 @@ def constructorApp? (e : Expr) : MetaM (Option (ConstructorVal × Array Expr)) :
 
 /--
 Similar to `constructorApp?`, but on failure it puts `e` in WHNF and tries again.
+It also `isOffset?`
 -/
 def constructorApp'? (e : Expr) : MetaM (Option (ConstructorVal × Array Expr)) := do
-  if let some r ← constructorApp? e then
+  if let some (e, k) ← isOffset? e then
+    if k = 0 then
+      return none
+    else
+      let .ctorInfo val ← getConstInfo ``Nat.succ | return none
+      if k = 1 then return some (val, #[e])
+      else return some (val, #[mkNatAdd e (toExpr (k-1))])
+  else if let some r ← constructorApp? e then
     return some r
   else
     constructorApp? (← whnf e)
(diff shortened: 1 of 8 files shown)


src/Lean/Meta/CtorRecognizer.lean
```diff
@@ -64,9 +65,17 @@ def constructorApp? (e : Expr) : MetaM (Option (ConstructorVal × Array Expr)) :
 
 /--
 Similar to `constructorApp?`, but on failure it puts `e` in WHNF and tries again.
+It also `isOffset?`
```
eric-wieser, line 68: I assume this sentence is incomplete?

PR #3617, opened 2024-03-06
feat: guard_msgs to escapes trailing newlines

### LINKED ISSUE #3571: guard_msgs and trailing space
Consider this code:
```
#guard_msgs in
run_meta
  logInfo "Foo \nbar\nbaz"
```
The code action will add
```
/--
info: Foo 
bar
baz
-/
```
but then saving this file with default VSCode settings removes the extra space, turn it into
```
/--
info: Foo
bar
baz
-/
```
and then `#guard_msgs` complains again.

Maybe code shouldn’t produce messages with final newlines, but some tools (in particular `addSuggestion`) may, so it would be nice to handle this gracefully.

Possible fixes:

 * `#guard_msgs` could strip trailing spaces
 * `#guard_msgs` could highlight trailing spaces, by replacing it with `_`, or some unicode symbol indicating a placeholder (`⎵`).

### Versions

4.7.0-nightly-2024-03-01

### Impact

nomeata:
This makes trailing whitespace visible and protectes them against
trimming by the editor, by appending the symbol ⏎ to such a line (and
also to any line that ends with such a symbol, to avoid ambiguities in
the case the message already had that symbol).

(Only the code action output / docstring parsing is affected; the error message as sent
to the InfoView is unaffected.)

Fixes #3571

### DIFF
--- a/src/Lean/Elab/GuardMsgs.lean
+++ b/src/Lean/Elab/GuardMsgs.lean
@@ -73,9 +73,28 @@ structure GuardMsgFailure where
   res : String
 deriving TypeName
 
+/--
+Makes trailing whitespace visible and protectes them against trimming by the editor, by appending
+the symbol ⏎ to such a line (and also to any line that ends with such a symbol, to avoid
+ambiguities in the case the message already had that symbol).
+-/
+def revealTrailingWhitespace (s : String) : String :=
+  s.replace "⏎\n" "⏎⏎\n" |>.replace "\t\n" "\t⏎\n" |>.replace " \n" " ⏎\n"
+
+/- The inverse of `revealTrailingWhitespace` -/
+def removeTrailingWhitespaceMarker (s : String) : String :=
+  s.replace "⏎\n" "\n"
+
+/--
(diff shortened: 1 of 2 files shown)


nomeata:
@kmill, would you be interested in having a look if this approach is reasonable?

kmill:
I'm a bit uncertain about using a new symbol (`⏎`) for line breaks, but I've tried to come up with other solutions and there doesn't seem to be anything better.

1. `⏎` is unlikely to be used in messages, so we won't see very much of the `⏎⏎` encoding.
2. `⏎` is a good graphical representation for the concept.

Maybe @digama0 has thoughts about this? I approve conditional on Mario's approval (or if Mario conditionally approves on my approval).

nomeata:
Thanks!

Fractional reviewing :-)

digama0:
Yes, I'm fine with this solution. I don't think we will see much of this symbol anyway, because we want to avoid trailing whitespace altogether, but at least this won't cause heisenbugs that trigger only on save.

PR #3620, opened 2024-03-06
feat: MatcherApp.transform: Try to preserve alt’s variable name

nomeata:
this makes the ugly `fst`/`snd` variable names in the functional
induction principles go away.

Ironically I thought in order to fix these name, I should touch the
mutual/n-ary argument packing code used for well-founded recursion, and
embarked on a big refactor/rewrite of that code, only to find that at
least this particular instance of the issue was somewhere else. Hence
breaking this into its own PR; the refactoring will follow (and will
also improve some other variable names.)

### DIFF
--- a/src/Lean/Meta/Match/MatcherApp/Transform.lean
+++ b/src/Lean/Meta/Match/MatcherApp/Transform.lean
@@ -179,7 +179,7 @@ def arrowDomainsN (n : Nat) (type : Expr) : MetaM (Array Expr) := do
 /--
 Sets the user name of the FVars in the local context according to the given array of names.
 
-If they differe in size the shorter size wins.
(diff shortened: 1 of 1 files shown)


src/Lean/Meta/Match/MatcherApp/Transform.lean
```diff
@@ -176,6 +176,17 @@ def arrowDomainsN (n : Nat) (type : Expr) : MetaM (Array Expr) := do
     type := β
   return ts
 
+/--
+Sets the user name of the FVars in the local context according to the given array of names.
+
+If they differe in size the shorter size wins.
```
nomeata, line 182:
```suggestion
If they differ in size the shorter size wins.
```

PR #3621, opened 2024-03-06
refactor: ArgsPacker

nomeata:
This introduces the `ArgsPacker` module and abstraction, to replace the exising `PackDomain`/`PackMutual` code. The motivation was that we now have more uses besides `Fix.lean` (`GuessLex` and `FunInd`), and the code was spread in various places.

The goals are

 * consistent function naming withing the the `PSigma` handling, the `PSum` handling, and the combined interface
 * avoid taking a type apart just based on the `PSigma`/`PSum` nesting, to be robust in case the user happens to be using `PSigma`/`PSum` somewhere. Therefore, always pass an `arity` or `numFuncs` or `varNames` around.
 *  keep all the `PSigma`/`PSum` encoding logic contained within one module (`ArgsPacker`), and keep that module independent of its users (so no `EqnInfos` visible here).
 * pick good variable names when matching on a packed argument
 * the unary function now is either called `fun1._unary` or `fun1._mutual`, never `fun1._unary._mutual`.

This file has less heavy dependencies than `PackMutual` had, so build parallelism is improved as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,8 @@ jobs:
                 // exclude seriously slow tests
                 "CTEST_OPTIONS": "-E 'interactivetest|leanpkgtest|laketest|benchtest'"
               },
-              {
+              // TODO: suddenly started failing in CI
+              /*{
                 "name": "Linux fsanitize",
                 "os": "ubuntu-latest",
                 "quick": false,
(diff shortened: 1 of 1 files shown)


nomeata:
(includes #3620 and should be merged afterwards.)

nomeata:
This is done, I think. Nice improvements to the variable names shown by `termination_by?`.
I’ll wait for the next nightly-with-mathlib for some extra testing though.

src/Lean/Meta/ArgsPacker.lean
```diff
@@ -0,0 +1,534 @@
...
+```
+(x : A ⊕' B) → (match x with | .inl x => R₁[x] | .inr R₂[z]
+```
+if the codomains are dependent, or
+```
+(x : A ⊕' B) → R
+```
+if they are all the same.
+
+-/
```
nomeata, line 323: TODO: This heuristic isn't great. The caller knows whether they require a dependent or a non-depenent arrow, so this should be two functions, or an option.

PR #3622, opened 2024-03-06
fix: stray dbgTraceVal in trace children elision

Kha:


Kha:
Unfortunately not automatically testable, see test file

PR #3625, opened 2024-03-06
fix: make `delabConstWithSignature` avoid using inaccessible names

### LINKED ISSUE #2846: Signatures displayed by #check should not show daggered names
### Description

[Some time ago](https://github.com/leanprover/lean4/pull/1943), `#check` and hovers were set to show signatures of functions rather  than their types. In other words, `#check String.get` shows `String.get (s : String) (p : String.Pos) : Char` rather than `String.get : String -> String.Pos -> Char`. This is generally a good idea - `String.replace (s pattern replacement : String) : String`is far more useful than `String.replace : String -> String -> String -> String`, after all.

But it falls down when arguments are not given explicit names in the signature. Then, internal daggered names are shown, eg:
```
#check String.append
String.append (a✝ : String) (a✝¹ : String) : String
``` 

This is confusing. While we should probably ensure that all functions in core and Std name their arguments, users will be writing anonymous names. Witness:
```
def length : List α → Nat
  | [] => 0
  | _ :: xs => length xs + 1

#check length
length.{u_1} {α : Type u_1} (a✝ : List α) : Nat
```

### Context

Real users have encountered this. We should fix it.

### Versions
At least `Lean (version 4.3.0, commit 734ce1ef2f62, Release)`.

### Potential solutions

Idris has [a pragma to associate name hints with types](https://idris2.readthedocs.io/en/latest/reference/pragmas.html#name). Then, when Idris needs to cook up a name for something (e.g. in a code action), it will devise a locally-unique name based on that hint. So we could write something like:
```
@[names xs, ys, zs]
inductive List ...
```
and then get output like
```
#check length
length.{u_1} {α : Type u_1} (xs : List α) : Nat
```
if the user hadn't already specified something. Likewise,
```
def doStuff : List a -> List a -> List a -> List a -> List a ...

#check doStuff
length.{u_1} {α : Type u_1} (xs ys zs xs' : List α) : Nat

def xs : List Nat := []
#check length
length.{u_1} {α : Type u_1} (ys : List α) : Nat
```

 ### Impact

kmill:
The `delabConstWithSignature` delaborator is responsible for pretty printing constants with a declaration-like signature, with binders, a colon, and a type. This is used by the `#check` command when it is given just an identifier.

It used to accumulate binders from pi types indiscriminately, but this led to unfriendly behavior. For example, `#check String.append` would give
```
String.append (a✝ : String) (a✝¹ : String) : String
```
with inaccessible names. These appear because `String.append` is defined using patterns, so it never names these parameters.

Now the delaborator stops accumulating binders once it reaches an inaccessible name, and for example `#check String.append` now gives
```
String.append : String → String → String
```
We do not synthesize names for the sake of enabling binder syntax because the binder names are part of the API of a function — one can use `(arg := ...)` syntax to pass arguments by name. The delaborator also now stops accumulating binders once it reaches a parameter with a name already seen before — we then rely on the main delaborator to provide that parameter with a fresh name when pretty printing the pi type.

As a special case, instance parameters with inaccessible names are included as binders, pretty printing like `[LT α]`, rather than relegating them (and all the remaining parameters) to after the colon. It would be more accurate to pretty print this as `[inst✝ : LT α]`, but we make the simplifying assumption that such instance parameters are generally used via typeclass inference. Likely `inst✝` would not directly appear in pretty printer output, and even if it appears in a hover, users can likely figure out what is going on. (We may consider making such `inst✝` variables pretty print as `‹LT α›` or `infer_instance` in the future, to make this more consistent.)

Something we note here is that we do not do anything to make sure parameters that can be used as named arguments actually appear named after the colon (nor do we assure that the names are the correct names). For example, one sees `foo : String → String → String` rather than `foo : String → (baz : String) → String`. We can investigate this later if it is wanted.

We also give `delabConstWithSignature` a `universes` flag to enable turning off pretty printing universe levels parameters.

Closes #2846

### DIFF
--- a/src/Lean/PrettyPrinter/Delaborator/Builtins.lean
+++ b/src/Lean/PrettyPrinter/Delaborator/Builtins.lean
@@ -1130,7 +1130,7 @@ where
   delabParamsAux (bindingNames : NameSet) (idStx : Ident) (groups : TSyntaxArray ``bracketedBinder) (curIds : Array Ident) := do
     let e@(.forallE n d e' i) ← getExpr | unreachable!
     let bindingNames := bindingNames.insert n
-    let stxN ← annotateCurPos (mkIdent n)
+    let stxN := mkIdent n
     let curIds := curIds.push ⟨stxN⟩
     if shouldGroupWithNext bindingNames e e' then
       withBindingBody n <| delabParamsAux bindingNames idStx groups curIds


nomeata:
Nice!

It’s probably fine, but maybe add some examples to the test file where there are named parameters after an inaccessible name:

def foo : String → (needle : String) → String := fun _ y => yo
def bar : String → (n : Nat) → Fin (n+1) := fun _ n => ⟨0, Nat.zero_lt_succ n⟩

kmill:
What do we think about inaccessible instance argument pretty printing as nameless binders?
```
def bar' [LT α] (x : α) : x < x := sorry
#check bar'
-- bar'.{u_1} {α : Type u_1} [LT α] (x : α) : x < x
```
If I follow the rule that inaccessible arguments must go after the colon, then pretty much all cases of instance arguments would have them coming after the colon, which is less than ideal.

If I omit the name, then the problem is that when that name appears in hovers (or in the pretty printed result itself, as unlikely as that might be) then you can't see which binder it's associated to. But, if I don't omit the name, then we have inaccessible names visible.

I think this is a good compromise printing them without their inaccessible name, and before the binder.

tests/lean/run/2846.lean
```diff
@@ -0,0 +1,74 @@
...
+#guard_msgs in #check foo'
+
+/-!
+Named argument after inaccesible name, still stays after the colon.
+But, it does not print using named pi notation since this is not a dependent type.
+-/
+def foo'' : String → (needle : String) → String := fun _ yo => yo
+
+/-- info: foo'' : String → String → String -/
+#guard_msgs in #check foo''
```
kmill, line 56: @nomeata With your first example, `needle` doesn't appear as a named argument because it's not a dependent type. With some more work, we could make it appear, but everything after the `:` doesn't have a reliable name, so I'm not sure if that's worth doing.

nomeata: Could you show those names that don’t have a macro scope (`.hasMacroScopes`)? (Just blindly guessing here, I don't know my way around these things yet).

kmill: Everything after the colon is pretty printed with the usual delaborator, so this would take some coordination with that.

If there were a pretty printer option that could cause pi types to print dependently, then this delaborator could annotate the pi types along the spine of this expression with that option, the ones that don't have macro scopes, before handing it off to be delaborated.

nomeata: Thanks for the explanation, then this is out of scope for this PR

PR #3631, opened 2024-03-07
chore: pr-release: fix MESSAGE calculation

nomeata:
previously there was a logic bug where the check if a std tag exists
would override the result of whether a mathlib tag exists (both wrote
`MESSAGE` in both cases).

Now we only check for the mathlib tag if we found the std tag.


semorrison:
Sorry, missed this, and have already merged #3700.

PR #3632, opened 2024-03-07
feat: thread initialization for reverse FFI

Kha:
Makes it possible to properly allocate and free thread-local runtime resources for threads not started by Lean itself

### DIFF
--- a/doc/dev/ffi.md
+++ b/doc/dev/ffi.md
@@ -111,6 +111,15 @@ if (lean_io_result_is_ok(res)) {
 lean_io_mark_end_initialization();
 ```
 
+In addition, any other thread not spawned by the Lean runtime itself must be initialized for Lean use by calling
+```c
+void lean_initialize_thread();
+```
+and should be finalized in order to free all thread-local resources by calling
+```c
+void lean_finalize_thread();
+```
+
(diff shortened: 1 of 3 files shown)


digama0:
Wondering if you noticed #2464 ...

Kha:
@digama0 Oh no, I didn't remember that one... I'd like to start small driven by practical needs, e.g. `lean_finalize*` seems unobjectionable to me but thread finalizers are a bit more fringe and I don't even know whether post finalizers will stay around

digama0:
There is more info on the [Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Expose.20thread.20management.20routines.20to.20FFI.3F/near/387373357). The basic motivation is that FFI should be able to do all the same lifecycle things that lean does so that it can correctly initialize lean code when lean is not running the show. This has in fact stopped some FFI projects of mine with no recourse. As an example where thread finalizers are needed, suppose there is a lean object stored in some thread local state which is not managed by lean. This object will not be cleaned up properly when the thread goes down and may corrupt lean if the same CPU is reused for another thread.

If we get rid of `post_finalize` then they would also disappear from this interface; I don't think there is any stability promised in that regard.

PR #3637, opened 2024-03-08
fix: handle delayed-assigned mvars in `isDefEq` safely

### LINKED ISSUE #3393: "Unknown free variable" during isDefEq for delayed assignments when synthetic opaque assignable
### Description

When synthetic opaque metavariables are assignable, it is possible to get an "unknown free variable" error from isDefEq.

### Context

[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Interesting.20.60withAssignableSyntheticOpaque.60.20.60isDefEq.60.20bug/near/400471552) where this was reported.

### Steps to Reproduce

This reports "unknown free variable '_fvar.2067'" at `refine`:
```lean
example : (fun _ : Nat => 2) = (fun _ => 2) := by
  refine (?_ : (fun n : Nat => ?e) = _)
  rfl
```

**Expected behavior:** This succeeds, or at least gives a type mismatch error.

### Versions

```lean
#eval Lean.versionString -- "4.6.0-rc1"
```

### Additional Information

It appears that `Lean.Meta.expandDelayedAssigned?` is causing the type of a metavariable to leak an fvar from its local context. In `mkAppRange (mkMVar mvarIdPending) fvars.size tArgs.size tArgs`, the metavariable `mvarIdPending` might have a type that's not compatible with the current local context. In the above example, `?e` has a type that depends on `n`.

### Impact

thorimur:
Currently, unknown fvars can be introduced during `isDefEq` when synthetic opaque metavariables are assignable and appear in a delayed assignment. Specifically, if `?m.23` is delayed-assigned to `?a`, we currently replace `?m.23 as bs` with `?a bs`. However, if the type of `?a` depends on the new fvars in its context, we've now introduced unknown fvars to `isDefEq`. Notably, this happens when elaborating expressions like `fun x => ?a` which leave the type of `?a` unspecified (and therefore dependent on `x` via another delayed-assigned mvar).

This PR checks that either the context of `?a` is compatible with the local context, or attempts to clear the new fvars from `?a`'s context if `?a`'s context is strictly larger than the current one. Otherwise, `isDefEq` now fails.

---

Closes #3393

### DIFF
--- a/tests/lean/withAssignableSyntheticOpaqueBug.lean
+++ b/tests/lean/withAssignableSyntheticOpaqueBug.lean
@@ -10,15 +10,14 @@ def tst : MetaM Unit := do
   lambdaTelescope e fun _ b => do
     let m' := b.getAppFn
     assert! m'.isMVar
-    let some { mvarIdPending .. } ← getDelayedMVarAssignment? m'.mvarId!
-      | IO.println f!"m' is not delayed-assigned"
-    IO.println (mvarIdPending == m.mvarId!)
-    IO.println (← mvarIdPending.isDelayedAssigned)
-    IO.println (← getExprMVarAssignment? m'.mvarId!)
     assert! (← withAssignableSyntheticOpaque <| isDefEq b (mkNatLit 0))
     IO.println (← getExprMVarAssignment? m'.mvarId!)
+    let some m ← getExprMVarAssignment? m.mvarId!
+      | IO.println f!"{m} is not assigned"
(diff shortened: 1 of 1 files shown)


semorrison:
It looks like the Mathlib failure is genuinely caused by this PR, right?

thorimur:
Yes, I believe this one's for real. :) Contrast with #3743, which built mathlib around the same time.

PR #3638, opened 2024-03-08
fix: eta-expanded instances at `SynthInstance.lean`

leodemoura:
Remark: this commit removes the `jason1.lean` test. Motivation: It breaks all the time due to changes we make, and it is not clear anymore what it is testing.

### DIFF
--- a/src/Lean/Meta/SynthInstance.lean
+++ b/src/Lean/Meta/SynthInstance.lean
@@ -323,7 +323,7 @@ def getSubgoals (lctx : LocalContext) (localInsts : LocalInstances) (xs : Array 
 
 /--
 Similar to `mkLambdaFVars`, but ensures result is eta-reduced.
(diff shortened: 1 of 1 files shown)


src/Lean/Meta/SynthInstance.lean
```diff
@@ -321,6 +321,26 @@ def getSubgoals (lctx : LocalContext) (localInsts : LocalInstances) (xs : Array
     subgoals := inst.synthOrder.map (mvars[·]!) |>.toList
   }
 
+/--
+Similar to `mkLambdaFVars`, but ensures result is eta-reduced.
+For example, support `e` is the local variable `inst x y`, and `xs` is `#[x, y]`, then
```
nomeata, line 326:
```suggestion
For example, suppose `e` is the local variable `inst x y`, and `xs` is `#[x, y]`, then
```

PR #3641, opened 2024-03-09
feat: `modifyM`/`modifyGetM` for `IO.Ref`

tydeu:
Adds `IO.Ref.modfyM` / `IO.Ref.modifyGetM` which are monadic variants of `IO.Ref.modify` (`ST.Prim.Ref.modify`) / `IO.Ref.modifyGet` (`ST.Prim.Ref.modifyGet`). Like them, it uses `ST.Prim.Ref.take` to atomically modify the reference via busy-waiting, but allow users to perform `BaseIO` actions within the critical section. This is safe because `BaseIO` / `ST σ α` actions do not throw exceptions.  Without these utilities, to modify a reference while preforming I/O, one needs to use `IO.Mutex`. Thus, this provides a lightweight alternative for users.

The outstanding question here is whether this is good idea or if users *should* be expected to use `IO.Mutex` if they wish to modify a reference while performing I/O.

### DIFF
--- a/src/Init/System/IO.lean
+++ b/src/Init/System/IO.lean
@@ -740,6 +740,9 @@ instance : MonadLift (ST IO.RealWorld) BaseIO := ⟨id⟩
 def mkRef (a : α) : BaseIO (IO.Ref α) :=
   ST.mkRef a
 
+@[inline] def Ref.modifyM (r : Ref α) (f : α → BaseIO α) : BaseIO Unit := ST.Prim.Ref.modifyM r f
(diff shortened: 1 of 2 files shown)


mhuisi:
One potential issue is that you can get deadlocks using `modify` this way, right?

tydeu:
@mhuisi Could you elaborate? I do not see how this introduces more possibility of deadlock than `IO.Ref.modify` and/or `IO.Mutex`. The example I can think of is that if you wait on a thread within the `modifyM` that is also waiting to modify the reference. But the same issue appears with `IO.Mutex`, so it does not seem to surprising.

PR #3647, opened 2024-03-10
feat: allow `noncomputable unsafe` definitions

tydeu:
Enables the combination of `noncomputable unsafe` to be used for definitions. Outside of pure theory, `noncomputable` is also useful to prevent Lean from compiling a definition which will be implemented with external code later. Such definitions may also wish to be marked `unsafe` if they  perform morally impure or memory-unsafe functions.

### DIFF
--- a/src/Lean/Elab/MutualDef.lean
+++ b/src/Lean/Elab/MutualDef.lean
@@ -68,8 +68,6 @@ private def check (prevHeaders : Array DefViewElabHeader) (newHeader : DefViewEl
     throwError "'partial' theorems are not allowed, 'partial' is a code generation directive"
   if newHeader.kind.isTheorem && newHeader.modifiers.isNoncomputable then
     throwError "'theorem' subsumes 'noncomputable', code is not generated for theorems"
-  if newHeader.modifiers.isNoncomputable && newHeader.modifiers.isUnsafe then
-    throwError "'noncomputable unsafe' is not allowed"
   if newHeader.modifiers.isNoncomputable && newHeader.modifiers.isPartial then
     throwError "'noncomputable partial' is not allowed"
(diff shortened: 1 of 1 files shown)


tydeu:
For future reference, here is some additional info on the motivation behind this PR. [Alloy](https://github.com/tydeu/lean4-alloy) allow users to external (C) implementations to definitions via Lean code. This is done via metaprogramming utilities, but is essentially equivalent to a definition followed an `attribute [extern]` command. For example:

```lean
alloy c extern def myAdd (x y : UInt32) : UInt32 := {
  return x + y;
}
```

is essentially equivalent to:

```lean
noncomputable opaque myAdd (x y : UInt32) : UInt32
attribute [extern "alloy_myAdd"] myAdd
alloy c section
LEAN_EXPORT uint32_t alloy_myAdd ( uint32_t x , uint32_t y ) {
  return x + y;
}
end
```

Libraries which use Alloy may define functions that perform actions that are morally impure or memory-unsafe. Thus, the library creators would like to mark the definition `unsafe`. However, passing this `unsafe` through Alloy's macros results in a `noncomputable unsafe` definition, which Lean currently forbids. For instance, using the above example:

```lean
alloy c extern unsafe def myAdd (x y : UInt32) : UInt32
```

would produce:

```lean
noncomputable unsafe opaque myAdd (x y : UInt32) : UInt32
```

which is forbidden. The goal of this PR is to allow such a use case.

---

The use of `noncomputable` in the Alloy's definitions could be avoided by adding the `@[extern]` attribute to the definition when it is declared. However, this would require some special casing in Alloy, as it also wants to support providing implementations for already-existing definitions (e.g., constructors or projections). Thus, for consistency, it always produces the implementation separately from the definition.

PR #3648, opened 2024-03-10
refactor: simplify `Array.findIdx?` code

### LINKED ISSUE #3646: RFC: Simplify `Array.findIdx?`
### Proposal

In Data/Array/Basic.lean

I would like to simplify `Array.findIdx?`, and propose replacing its [code](https://github.com/leanprover/lean4/blob/1d3ef577c28b6298f4edf8aef86b23ca3ba99b66/src/Init/Data/Array/Basic.lean#L457-L476) with the following

```lean
@[inline]
def findIdx? {α : Type u} (as : Array α) (p : α → Bool) : Option Nat :=
  let rec loop (j : Nat) :=
    if h : j < as.size then
      if p as[j] then some j else loop (j + 1)
    else none
    termination_by as.size - j
  loop 0
```
I haven't changed the structure of the executable code. It was using two Nat's and some invariant with, I think, the intent to avoid wf recursion, (maybe because it's old code ?). 

  - **User Experience**: How does this feature improve the user experience?

It would be easier to reason about `Array.findIdx?`.

  - **Beneficiaries**: Which Lean users and projects benefit most from this feature/change?

In std, there is a proof_wanted [here](https://github.com/leanprover/std4/blob/f3447c3732c9d6e8df3bdad78e5ecf7e8b353bbc/Std/Data/Array/Lemmas.lean#L726),  the proof of which might turn out to be easier to write, with the proposed change.

  - **Maintainability**: Will this change streamline code maintenance or simplify its structure?
 
This change would simplify `Array.findIdx?`

### Community Feedback

I created a thread [here](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/findIdx.3F.20and.20indexOf.3F)

### Impact

michelsol:
This shortens `Array.findIdx?` code, by using termination_by (and well-founded recursion) instead of a structural recursion trick, with the intent to make it more proof friendly. 

One motivation is that it makes it easier to write a proof that `Array.findIdx?` and `List.findIdx?` are equivalent. Furthermore, this will be useful to prove that more complex functions are equivalent.

Closes #3646

### DIFF
--- a/src/Init/Data/Array/Basic.lean
+++ b/src/Init/Data/Array/Basic.lean
@@ -456,24 +456,12 @@ def findRev? {α : Type} (as : Array α) (p : α → Bool) : Option α :=
 
 @[inline]
 def findIdx? {α : Type u} (as : Array α) (p : α → Bool) : Option Nat :=
-  let rec loop (i : Nat) (j : Nat) (inv : i + j = as.size) : Option Nat :=
-    if hlt : j < as.size then
-      match i, inv with
-      | 0, inv => by
-        apply False.elim
-        rw [Nat.zero_add] at inv
-        rw [inv] at hlt
-        exact absurd hlt (Nat.lt_irrefl _)
-      | i+1, inv =>
-        if p as[j] then
-          some j
-        else
-          have : i + (j+1) = as.size := by
-            rw [← inv, Nat.add_comm j 1, Nat.add_assoc]
-          loop i (j+1) this
-    else
-      none
-  loop as.size 0 rfl
+  let rec loop (j : Nat) :=
+    if h : j < as.size then
+      if p as[j] then some j else loop (j + 1)
+    else none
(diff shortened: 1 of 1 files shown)


semorrison:
@michelsol, we would need to see convincing evidence there is no possible performance regression from this.

semorrison:
!bench

digama0:
I would expect the new version to perform better than the original one, because it has an unnecessary fuel parameter which will be there in the compiled code. It has the look of something that was written either before a good well founded recursion mechanism was available, or in order to avoid well founded recursion for other reasons (e.g. defeq reduction in the kernel), but I think we have since decided not to worry about this in Array functions given the proliferation of well founded recursive definitions.

michelsol:
Sorry I wasn't aware of the bench command. Is the benchmark report good enough or is there something else I'm expected to do?

PR #3655, opened 2024-03-12
feat: add option `tactic.customEliminators` to be able to turn off custom eliminators for `induction` and `cases`

kmill:
This was suggested by Scott Morrison to be able to help projects adjust to `Nat` having built-in custom eliminators.


semorrison:
I think we can merge after resolving the conflict in RELEASES.md. It would be good to have this in 4.8.

kmill:
@semorrison Updated

PR #3656, opened 2024-03-12
fix: jump to correct definition when names overlap

### LINKED ISSUE #1170: Go to definition jumps to incorrect file
If there are two different files that contain declarations with the same name, the go to definition feature will sometimes jump to the declaration in the wrong file. This seems to happen consistently if the offending files are all open in a particular editor group (in tabs). 

Example: If I make a file `x.lean`

```
structure User where
  name : String
```

Then another file `y.lean`

```
structure User where
  name : Unit

example : User -> User := id
```

Right clicking on the `User` in the `example` declaration and selecting go to definition will take me to the declaration in `x.lean`.

I'm using vscode 1.67,  v0.0.75 of the extension, and  lean 4 nightly-2022-05-19.

mhuisi:
Fixes #1170.

This PR adds the module name to `RefIdent` in order to distinguish conflicting names from different files. This also fixes related issues in find-references or the call hierarchy feature.
It also adds some docstrings and stylistically refactors a bunch of code.

### DIFF
--- a/tests/lean/run/944.lean
+++ b/tests/lean/run/944.lean
@@ -12,7 +12,7 @@ open Lsp
 def identOf : Info → Option (RefIdent × Bool)
   | .ofTermInfo ti => match ti.expr with
     | .const n .. => some (.const `anonymous n, ti.isBinder)
-    | .fvar id .. => some (.fvar `anonymous `anonymous id, ti.isBinder)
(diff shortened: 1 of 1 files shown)


mhuisi:
!bench

m4lvin:
Thank you so much, this will make my current project a lot easier to work with :-)

PR #3660, opened 2024-03-12
fix: toUInt64LE! and toUInt64BE! are swapped

### LINKED ISSUE #3657: Are ByteArray.toInt64LE! and ByteArray.toInt64BE! reversed?
### Prerequisites

### Description

I suspect that the little-endian and big-endian ByteArray functions have their names backwards. The little-endian function seems to convert a ByteArray to a UInt64 by taking the leftmost byte as most significant, and vice-versa.

### Steps to Reproduce

It should be easy to check, but here's a minimal reproduction, showing how the current functions seem to get things wrong when reading an ELF binary. Requires a 64 bit ELF binary - I used `/bin/true`.

```
def test: IO Unit := do
  let bytes ← IO.FS.readBinFile ↑"/bin/true"
  if h : bytes.size > 256 then do
    IO.println bytes[0x5] -- e_ident[EI_DATA], 1 means little endian, 2 means bigendian
    let ph_off := bytes.extract 0x20 0x28 --program header offset in a 64 bit binary, should be 64.
    IO.println ph_off
    IO.println ph_off.toUInt64LE!
    IO.println ph_off.toUInt64BE!

#eval test 
```

**Expected behavior:**

I would expect the code above to show toUInt64LE! returning 64 for bytes [0x20,0x28) on a little-endian binary, and toUInt64BE! to give a 64 for that range on a big-endian binary.

**Actual behavior:** 

Seems to be the reverse

### Versions

4.7.0-rc1
Ubuntu 22.04.1

### Impact

digama0:
fixes #3657

These functions are mostly not used by lean itself, but it does affect two occurrences of `ByteArray.toUInt64LE! <$> IO.getRandomBytes 8` which I left as is instead of switching them to use `toUInt64BE!` to preserve behavior; but they are random bytes anyway seeded by the OS so it's unlikely any use of them depending on particular values was sound to begin with.


semorrison:
@Vtec234, just checking if you wanted to look at this? I'm happy with this.

semorrison:
CI had failed, but I think just transiently.

semorrison:
Nope, not transient. Let me merge master and see if that helps.

PR #3661, opened 2024-03-13
chore: rename automatically generated equational theorems

leodemoura:
cc @nomeata


nomeata:
This renaming implicitly says “these theorems are no longer internals, you can use them”, right? 

In that case I think we should also have a nicer way of making sure they exist reliability. If we can't afford to generate them always because it's too expensive for definitions nobody wants to prove anything about, can we add an attribute maybe?

(Doesn't have to be part of this PR of course.)

PR #3670, opened 2024-03-14
feat: upstream apply helper tactics from Mathlib

kim-em:
These are used in Mathlib's `congr!` and `convert` tactics, which will be upstreamed soon.

### DIFF
--- a/src/Lean/Meta/Tactic/Apply.lean
+++ b/src/Lean/Meta/Tactic/Apply.lean
@@ -282,8 +282,8 @@ def propext (mvarId : MVarId) : MetaM MVarId := do
     -- Avoid applying `propext` if the target is not an equality of `Prop`s.
     -- We don't want a unification specializing `Sort*` to `Prop`.
     let tgt ← withReducible mvarId.getType'
-    let some (ty, _, _) := tgt.eq? | failure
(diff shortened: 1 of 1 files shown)


src/Lean/Meta/Tactic/Apply.lean
```diff
@@ -256,4 +261,68 @@ def _root_.Lean.MVarId.nthConstructor
...
+Try to convert an `Eq` into an `Iff` by applying `propext`.
+If successful, then returns then new goal, otherwise returns the original `MVarId`.
+-/
+def propext (mvarId : MVarId) : MetaM MVarId := do
+  let res ← observing? do
+    -- Avoid applying `propext` if the target is not an equality of `Prop`s.
+    -- We don't want a unification specializing `Sort*` to `Prop`.
+    let tgt ← withReducible mvarId.getType'
+    let some (ty, _, _) := tgt.eq? | failure
+    guard ty.isProp
```
kmill, line 286: `Expr.isProp` looks for the exact expression `Sort 0`, so this should probably be changed to
```suggestion
    let some (_, x, _) := tgt.eq? | failure
    guard <| ← Meta.isProp x
```

Issue #3302, opened 2024-02-11
RFC: A single rfl tactic implementation


nomeata:
I noticed (while #3299 passed by) that `rfl` produces a rather unhelpful error message:
```
import Std
example (P : Prop) : P := by rf
```
gives
```
type mismatch
  HEq.rfl
has type
  HEq ?m.10 ?m.10 : Prop
but is expected to have type
  P : Prop
```

This is not a good UX for a tactic that early users will be exposed to early (and also not great later on).

Looking at what happens I found that there are four different implementations registered (using `macro`/`macro_rules`) for `rfl`:

* Wrapping [`Lean.MVarId.refl`](https://leanprover-community.github.io/mathlib4_docs/Lean/Meta/Tactic/Refl.html#Lean.MVarId.refl), for `Eq`.
* exact `HEq.rfl`
* exact `Iff.rfl`
* [https://leanprover-community.github.io/mathlib4_docs/Std/Tactic/Relation/Rfl.html#Lean.MVarId.applyRfl](`Lean.MVarId.applyRfl`), which supports any relation with a suitable `@[refl]` lemma.

Unfortunately, it seems that using `macro_rules` to extend tactics doesn’t give great control over which error message is shown if all fail. 

Assuming we want `@[refl]` in core, I propose that, after upstreaming `Std.Tactic.Relation.Rfl`, we back the `rlf` tactic by a single function that combines `Lean.MVarId.refl` and `Lean.MVarId.applyRfl`, possibly still with a special path for `Eq` (should that be needed, TBD), and with good error messages for the various failure modes.

With a suitable `@[refl]` attribute, this should allow us to remove the extra `exact ….rfl` macro rules.

### Community Feedback

This was escalated from a [discussion on zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/rfl.20tactic.20without.20a.20macro).

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others benefit from the changes in this proposal being added, please ask them to add :+1: to it.

nomeata:
Having a closer look. One issue is that the order of parameters in `HEq` makes it ineligible to the `@[refl]` machinery:
```
HEq.refl.{u} {α : Sort u} (a : α) : @HEq α a α a
```
(it would have to be `@HEq α α a a`).

[...]

PR #3671, opened 2024-03-14
feat: upstream Std's rfl tactic

kim-em:
This allows tagging lemmas with `@[refl]`, that will then by used by `rfl`.

This is preparatory to upstreaming Mathlib's `convert` tactic.


nomeata:
Once this is merged we can tackle https://github.com/leanprover/lean4/issues/3302

PR #3673, opened 2024-03-14
feat: upstream `congr!` from Mathlib

kim-em:
This is a preliminary step to upstreaming `convert`.

We should later think about combining this with the standard `congr`.

There is still a builtin problem here, that you can not use `congr! (config := ...)` without `import Lean.Meta.Tactic.Congr!`. We need to move the config structure into e.g. `Init/MetaTypes.lean`, and then an update-stage0, but I haven't got it to work.

This is stacked on top of #3671 and #3672, and I'll wait for those to merge.


kmill:
Oh wow, I wasn't expecting `congr!` to be upstreamed.  That should be fine, but I've been thinking of simplifying it now that we've now seen it in action for a while in mathlib. (And I'm looking forward to `congr` one day being able to do everything `congr!` can do!)

By the way, I have an open PR to add that LawfulBEq instances are equal: https://github.com/leanprover-community/mathlib4/pull/11179

semorrison:
> Oh wow, I wasn't expecting to see that `congr!` was being upstreamed! That should be fine, but I've been thinking of simplifying it now that we've now seen it in action for a while in mathlib. (And I'm looking forward to `congr` one day being able to do everything `congr!` can do!)
> 
> By the way, I have an open PR to add that LawfulBEq instances are equal: [leanprover-community/mathlib4#11179](https://github.com/leanprover-community/mathlib4/pull/11179)

@kmill, I don't think it is super urgent, so if you think there is good reason to delay moving it to allow for a refactor, I think we can do that. I'm happy with moving it as is and updating later, however.

kmill:
I don't have any good reason to delay merging it, and I'm not sure when I'd get around to refactoring it, so carrying on with upstreaming seems fine to me. (Thanks!)

nomeata (commented):
Is there any hope that we can have a single `congr` tactic in lean, so that users don't have to learn about the difference? What would happen if this `congr!` were the only one?

semorrison:
We decided not to proceed at this point. @kmill is refactoring the Mathlib implementation in the meantime.

PR #3676, opened 2024-03-14
refactor: clean up public API around `Array.eraseIdx`

timotree3:
- Removes the public definitions `Array.eraseIdxAux` and `Array.eraseIdxSzAux` which were implementation details.
  - Motivation: `Array.eraseIdxAux` and `Array.eraseIdxSzAux` were clearly not intended to remain public, but simply making them private would make it inconvenient to unfold them when writing proofs in Std.
- Adds documentation comments to the public `Array.eraseIdx`-related definitions which remain.
- Removes `Array.eraseIdx'` which was just `Array.feraseIdx` wrapped in a subtype and adds `Array.size_feraseIdx` to prove the subtype property as a standalone theorem.

Co-Authored-By: Daniel Windham <daniel@atlascomputing.org>

### DIFF
--- a/src/Init/Data/Array/Basic.lean
+++ b/src/Init/Data/Array/Basic.lean
@@ -760,9 +760,6 @@ theorem size_feraseIdx (a : Array α) (i : Fin a.size) : (a.feraseIdx i).size = 
 def eraseIdx (a : Array α) (i : Nat) : Array α :=
   if h : i < a.size then a.feraseIdx ⟨i, h⟩ else a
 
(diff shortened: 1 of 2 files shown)


Seppel3210:
BTW porting over the lemma from my std4 PR leanprover/std4#690 was pretty easy: https://github.com/Seppel3210/std4/commit/109c3ce97e182813a166c9b55418d9a599a01bb6
So this is definitely bug free :smile: 

I also think this makes the API nicer

Seppel3210 (approved):

Issue #2867, opened 2023-11-12
DiscrTree doesn't match lemmas with `OfNat.ofNat *` to nat literal keys


timotree3:
### Prerequisites

* [ ] Put an X between the brackets on this line if you have done all of the following:
    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to mathlib4 or std4.

### Description

When a `@[simp]` lemma is `∀ (n : Nat), P (OfNat.ofNat n)`, `simp` fails to apply it to targets like `P 2` due to a bug in `DiscrTree`.

### Context

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/writing.20lemmas.20about.20.60ofNat.60/near/401399369)

This has been requiring us to add `no_index` around every explicit invocation of `OfNat.ofNat`. e.g. https://github.com/leanprover-community/mathlib4/pull/8317

### MWE

```lean
def MyProp (n : Nat) : Prop := False

-- DiscrTree key: [MyProp, OfNat.ofNat, Nat, *, *]
theorem myProp_ofNat (n : Nat) : MyProp (OfNat.ofNat n) := sorry

example : MyProp 0 := by simp [myProp_ofNat] -- fails. should succeed
example (n : Nat) : MyProp (OfNat.ofNat n) := by simp [myProp_ofNat] -- succeeds

-- DiscrTree key: [MyProp, *]
theorem myProp_ofNat' (n : Nat) : MyProp (no_index (OfNat.ofNat n)) := sorry
example : MyProp 0 := by simp [myProp_ofNat'] -- succeeds
```

**Expected behavior:** `simp` should be able to solve `MyProp 0`

**Actual behavior:** `simp` makes no progress

### Versions

4.3.0-rc1

### Additional Information

My diagnosis of the issue is that in `DiscrTree`, we are stripping off invocations of `OfNat.ofNat` on numerals, but not on variables.

Code stripping it off of numerals:
https://github.com/leanprover/lean4/blob/8cfcf7ce61bba328ed57d97bb12cf1a0fc40a0b9/src/Lean/Meta/DiscrTree.lean#L167-L173

It seems like the possible solutions are:
[...]

Issue #2916, opened 2023-11-20
dsimp doesn't pack raw `nat_lit`s into `OfNat.ofNat` invocations


timotree3:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to mathlib4 or std4.

### Description

Currently, `simp` has special logic to normalize `nat_lit n` in the target into `@OfNat.ofNat Nat (nat_lit n) _`. This conversion is a defeq, so it should be supported by `dsimp`.

```lean
set_option pp.coercions false -- Show `OfNat.ofNat` when present for clarity

example : nat_lit 2 = sorry := by
  -- ⊢ 2 = sorryAx Nat
  simp only
  -- ⊢ OfNat.ofNat 2 = sorryAx Nat
  -- (as expected)

example : nat_lit 2 = sorry := by
  dsimp only -- dsimp made no progress
  -- expected: same result as simp

-- simp delegates to dsimp for subexprs with forward dependencies
example (α : Nat → Type) (f : (n : Nat) → α n): f (nat_lit 2) = sorry := by
  simp only -- simp made no progress
  -- expected: ⊢ f (OfNat.ofNat 2) = sorryAx Nat
```

### Context

This has particular implications for any meta code which uses `simp` to produce theorem statements. Generally it is an invariant that declaration statements should only contain nat literals wrapped in `OfNat.ofNat`, and it is desirable that running `simp` on the statement guarantees this property, even in the presence of forward dependencies. For a concrete example, due to this bug, the code which generates simplification equations for well-founded definitions produces lemmas with LHSs containing raw nat literals:

```lean
/- Key ingredients:
  - Well-founded recursion (so equation lemmas are even generated)
  - Matching on a nat literal pattern (`0` below)
  - The conclusion type depends on the nat discriminant (so `dsimp` is used instead of `simp`)
  -/
@[simp] def wellFounded (n : Nat): { r : Nat // r = n } :=
  match n with
  | 0 => ⟨0, rfl⟩
  | n + 1 =>
    let _ := wellFounded (id n)
    ⟨n + 1, rfl⟩

#check wellFounded._eq_1
-- wellFounded._eq_1 : wellFounded 0 = { val := OfNat.ofNat 0, property := wellFounded.proof_3 }
--                                 ^ raw nat lit `0`
```

[...]

PR #3684, opened 2024-03-14
fix: remove special handling of numerals in `DiscrTree`

timotree3:
Fixes https://github.com/leanprover/lean4/issues/2867 by removing special handling of `OfNat.ofNat` numerals from `DiscrTree`.

The special handling was added in 2b8e55c2f19e6599128bc7e57cb7221df5693b4e to fix a bug represented in the test `tests/lean/run/bugNatLitDiscrTree.lean`, but with https://github.com/leanprover/lean4/issues/2916 fixed, that test case passes without the special handling.

Some special handling of `Nat.zero` and `Nat.succ` remains to ensure that keys continue to match when unification replaces constants of the form `Nat.zero...succ.succ` with numerals.

One notable implication of the removal of this special handling is that it means that instances of `OfNat MyType n` for specific natural numbers `n` should be written as `OfNat MyType (nat_lit n)`. See [tests/run/rewrites.lean](https://github.com/leanprover/lean4/pull/3684/files#diff-2cc9ca7f3a3c467c6c2dfc61f2d32867b5f087b1c4ef9721ecfc73a8ba72567eR39) for an example of this modification being needed.

Since this PR changes the persisted discrimination tree format, an `update stage0` commit is necessary to make tests pass. (Otherwise oleans produced by the stage0 compiler with old discrimination tree data are used alongside the new discrimination tree algorithm.)


timotree3:
Blocked by #3686 which would fix the only remaining failing test.

timotree3:
awaiting-review

tests/lean/run/discrTreeNumeral.lean
```diff
@@ -0,0 +1,144 @@
...
+
+/-- A base instance for `P Nat.zero` and a step instance for `P n → P (n.succ)` can combine to reach any
+  sequence of `Nat.zero...succ.succ`.
+
+  Because the kernel has special handling for unifying `Nat.succ`,
+  this requires that they can also combine to reach any numeral.
+
+  For example, to reach `P (Nat.zero.succ.succ)`,
+  the first recursive step contains the unification problem `P (Nat.zero.succ.succ) =?= P (?n.succ)`,
+  which the kernel solves by setting `?n := 1`, *not* `?n := Nat.zero.succ`.
```
timotree3, line 113: It just occurred to me to test this for `nat_lit 0`, and lo and behold, unification has the same surprising behavior of setting the metavariable to `1`:

```lean
set_option pp.natLit true in
#check (show (nat_lit 0).succ.succ = Nat.succ _ from rfl)
-- Nat.succ (Nat.succ (nat_lit 0)) = Nat.succ 1
```

If we replace `Nat.zero` with `nat_lit 0` in this test case, we get a new test case which fails after this PR:
```lean
class MyClass (n : Nat) : Prop where

instance myClass_succ [MyClass n] : MyClass (n.succ) := ⟨⟩

instance myClass_nat_lit_zero : MyClass (nat_lit 0) := ⟨⟩

example : MyClass (nat_lit 0).succ.succ :=
  inferInstance -- failed to synthesize instance
```

Making this instance path work as expected would require a significantly more complicated fix to the discrtree code (namely, normalize `nat_nit n` to `OfNat.ofNat (nat_lit n)`, except when already inside `OfNat.ofNat`), and the behavior after the current PR is easy to work around (add an instance for `0` instead of or in addition to `nat_lit 0`). Would it be acceptable to leave addressing this case as future work?

timotree3 (commented):
I believe this PR is ready, although I'm flagging one shortcoming which I hope can be addressed in future work.

timotree3:
@leodemoura I would love if you would review this PR when you get the chance. I think you are effectively the code owner for the discrimination tree.

david-christiansen:
> One notable implication of the removal of this special handling is that it means that instances of `OfNat MyType n` for specific natural numbers `n` should be written as `OfNat MyType (nat_lit n)`. See [tests/run/rewrites.lean](https://github.com/leanprover/lean4/pull/3684/files#diff-2cc9ca7f3a3c467c6c2dfc61f2d32867b5f087b1c4ef9721ecfc73a8ba72567eR39) for an example of this modification being needed.

This seems like a significant usability regression from my perspective. If failing to decorate a term in one spot with a special modifier leads to various tools degrading in their performance or not working well, but doesn't throw an error, then I suspect we'll mostly end up with various tools degrading in their performance.

Is there a way to fix this elsewhere?

timotree3:
> This seems like a significant usability regression from my perspective.

You're right. Getting an error message `failed to synthesize instance OfNat MyType 0` when you have an instance of `OfNat MyType 0` (but not `OfNat MyType (nat_lit 0)`) would be a very confusing user experience. If we did think it was preferable from a language design point of view for the two spellings of this instance to have differing precise semantics, then we should at least improve this error message. One easy way would be to set `pp.natLit true` by default so that the error message is `failed to synthesize instance OfNat MyType (nat_lit 0)` instead.

timotree3:
Upon further reflection, I am realizing that the following might be a helpful guiding principle for discrimination tree behavior:

If `e1` and `e2` can be unified at reducible transparency, then queries at key `e2` should return values inserted at key `e1` in a discrimination tree.

Given this specification, let's evaluate the status quo and this PR on three key examples:

```lean
/-! For each reducible defeq, test that instances for the LHS apply to the RHS -/

class MyClass (n : Nat) : Prop where

def myClass_zero : MyClass 0 where
def myClass_ofNat (n : Nat) : MyClass (OfNat.ofNat n) where

#check show 0 = nat_lit 0 by with_reducible rfl -- OK (reducibly defeq)

attribute [local instance] myClass_zero in
#synth MyClass (nat_lit 0)
-- Status quo: OK (instance applied)
-- This PR: ERROR (instance didn't apply)

#check show OfNat.ofNat _ = 0 by with_reducible rfl -- OK (reducibly defeq)

attribute [local instance] myClass_ofNat in
#synth MyClass 0
-- Status quo: ERROR (instance didn't apply)
-- This PR: OK (instance applied)

#check show OfNat.ofNat _ = nat_lit 0 by with_reducible rfl -- ERROR (not reducibly defeq)

attribute [local instance] myClass_ofNat in
#synth MyClass (nat_lit 0)
-- Status quo: ERROR (instance didn't apply)
-- This PR: ERROR (instance didn't apply)
```

As you can see, the key change with this PR is that the discrimination tree now fails to honor the reducible defeq `0 = nat_lit 0`, but successfully honors the reducible defeq `OfNat.ofNat _ = 0`.

I've included the third example to highlight that the current behavior of reducible defeq is actually non-transitive: We have `OfNat.ofNat _ = 0` and `0 = nat_lit 0`, but not `OfNat.ofNat _ = nat_lit 0`. This non-transitivity is rather directly the cause of the existing bug: The discrimination tree eagerly normalizes by the equality `0 = nat_lit 0`, and then fails to unify `OfNat.ofNat _` with `nat_lit 0`. Fixing the behavior of reducible defeq will naturally determine the correct fix for the discrimination tree.

I think there are two reasonable approaches to fixing this:
- Demote `OfNat.ofNat` to a regular old semireducible function.
  - This means embracing the distinction between `0` (AKA `OfNat.ofNat (nat_lit 0)` and `nat_lit 0` and removing the reducible defeq `0 = nat_lit 0`.
- Embrace `OfNat.ofNat` as a reducible function and make `@OfNat.ofNat Nat x (instOfNatNat _) = x` a reducible defeq.

This PR would move the discrimination tree a step towards the former approach of demoting `OfNat.ofNat` to a normal function, although it does not finish the job by actually updating the behavior of reducible defeq.

Now that we're discussing it though, I think the latter approach might have some serious benefits. Namely, it would mean that you (almost?) never have to mentally distinguish between `nat_lit n` and `n` as a user, effectively demoting `nat_lit` to an implementation detail rather than a language feature.

What do you think @david-christiansen?

david-christiansen:
I don't feel like I have enough context to evaluate the technical trade-offs of the concrete proposals - sorry!

But I do know that explaining where and why `nat_lit` was required was extremely difficult while working on Functional Programming in Lean, and this actually led to it not being mandatory when writing instances. I would really like to maintain the property that users don't have to think about it, treating it (as you said) as an internal implementation detail.

david-christiansen:
(and thank you for the thorough consideration!)

semorrison:
Without implying that it is sufficient, a Mathlib adaptation branch would be necessary for progress here.

PR #3688, opened 2024-03-15
feat: additional collection operations

joehendrix:
This is a work-in-progress PR to ensure consistency of operations on collection data structures in Lean, specifically, `RBMap`, `NameMap` and `HashMap`.

Changes include
* `Repr` instances as needed.
* Renames `find*` operations to `get` for consistency with `GetElem`.
* `map` operations

This is stacked on top of #3694 and that should be merged first.

### DIFF
--- a/src/Lean/Data/AssocList.lean
+++ b/src/Lean/Data/AssocList.lean
@@ -54,17 +54,34 @@ def mapVal (f : β → δ) : AssocList α β → AssocList α δ
   | nil        => nil
   | cons k v t => cons k (f v) (mapVal f t)
 
-def findEntry? [BEq α] (a : α) : AssocList α β → Option (α × β)
+/-- Auxiliary for `List.reverse`. `List.reverseAux l r = l.reverse ++ r`, but it is defined directly. -/
+def reverseAux : AssocList α β → AssocList α β → AssocList α β
+  | nil,   r => r
+  | cons k v l, r => reverseAux l (cons k v r)
+
+def reverse (as : AssocList α β) : AssocList α β := reverseAux as nil
+
+@[inline]
+def mapValM {m : Type u → Type v} [Monad m] {α : Type u} {β γ : Type _} (f : β → m γ) (as : AssocList α β) : m (AssocList α γ) :=
+  let rec @[specialize] loop
+    | nil,         bs => pure bs.reverse
+    | cons k v as, bs => do loop as (cons k (← f v) bs)
+  loop as nil
+
+def getEntry? [BEq α] (a : α) : AssocList α β → Option (α × β)
   | nil         => none
   | cons k v es => match k == a with
     | true  => some (k, v)
-    | false => findEntry? a es
+    | false => getEntry? a es
 
-def find? [BEq α] (a : α) : AssocList α β → Option β
+def get? [BEq α] (a : α) : AssocList α β → Option β
   | nil         => none
   | cons k v es => match k == a with
     | true  => some v
-    | false => find? a es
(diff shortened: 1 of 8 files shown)


src/Lean/Data/AssocList.lean
```diff
@@ -54,17 +54,34 @@ def mapVal (f : β → δ) : AssocList α β → AssocList α δ
...
+
+def getEntry? [BEq α] (a : α) : AssocList α β → Option (α × β)
   | nil         => none
   | cons k v es => match k == a with
     | true  => some (k, v)
-    | false => findEntry? a es
+    | false => getEntry? a es
 
-def find? [BEq α] (a : α) : AssocList α β → Option β
+def get? [BEq α] (a : α) : AssocList α β → Option β
```
digama0, line 77: I don't like this naming because it sounds too much like a function that returns the n-th element of the list. `find` has more of a search/lookup connotation.

src/Lean/Data/HashMap.lean
```diff
@@ -140,11 +150,16 @@ def erase [BEq α] [Hashable α] (m : HashMapImp α β) (a : α) : HashMapImp α
...
+@[inline] def map (f : β → γ) (m : HashMapImp α β) : HashMapImp α γ :=
+  { size := m.size, buckets := m.buckets.mapVal f }
+end
+
+inductive WellFormed {α : Type u} [BEq α] [Hashable α] : {β : Type v} → HashMapImp α β → Prop where
+  | mkWff          : ∀{β : Type v} n, WellFormed (@mkHashMapImp α β n)
   | insertWff      : ∀ m a b, WellFormed m → WellFormed (insert m a b |>.1)
   | insertIfNewWff : ∀ m a b, WellFormed m → WellFormed (insertIfNew m a b |>.1)
   | eraseWff       : ∀ m a,   WellFormed m → WellFormed (erase m a)
+  | mapWff         : ∀(f : β → γ) m, WellFormed m → WellFormed (map f m)
```
digama0, line 162: This doesn't work, it has universe issues. I've lost track of whether this code is now also used in Std.HashMap, but we can't do this in that case. For `Lean` it doesn't really matter since we can probably set all the universes to 0 anyway.

src/Lean/Data/AssocList.lean
tydeu: @digama0 one problem with using `find` for this distinction is that is already used for e.g. `Array` to denote the operation that searches for the element matching a predicate, not one indexes by key.

digama0: Isn't that one called `findP`?

tydeu: @digama0 no, they are [`List.find?`](https://github.com/leanprover/lean4/blob/0b01ceb3bb3334fd6c2f3b24996646046722c62a/src/Init/Data/List/Basic.lean#L337) and [`Array.find?`](https://github.com/leanprover/lean4/blob/0b01ceb3bb3334fd6c2f3b24996646046722c62a/src/Init/Data/Array/Basic.lean#L436), respectively. The convention of a search by function is carried through to `findSome?`, `findRev?`, `findIdx?` as well. Conversely, the reverse lookup of an index by value for an array is called [`Array.getIdx?`](https://github.com/leanprover/lean4/blob/0b01ceb3bb3334fd6c2f3b24996646046722c62a/src/Init/Data/Array/Basic.lean#L478).

joehendrix:
I'm closing this since the `GetElem` changes are in #3694 while the HashMap and RBMap changes will be done in independent PRs.

PR #3693, opened 2024-03-15
fix: deadlock in `IO.Promise.resolve`

Kha:


Kha:
As found in the incrementality PR

PR #3694, opened 2024-03-15
chore: extend GetElem with getElem! and getElem?

joehendrix:
This makes changes to the `GetElem` class so that it does not lead to unnecessary overhead in container like `RBMap`.

The changes are to:
1. Make `getElem?` and `getElem!` part of the `GetElem` class so they can be overridden in instances.
2. Introduce a `LawfulGetElem` class that contains correctness theorems for  `getElem?` and `getElem!` using the original definitions.
3. Reorganize definitions (e.g, by moving `GetElem` out of `Init.Prelude`) so that the `GetElem` changes are feasible.
4. Provide `LawfulGetElem` instances to complement all existing `GetElem` instances in Lean core.

To reduce the size of the PR, this doesn't do the work of providing new `GetElem` instances for `RBMap`, `HashMap` etc.  That will be done in a separate PR (#3688) that depends on this.

### DIFF
--- a/tests/lean/lcnfTypes.lean.expected.out
+++ b/tests/lean/lcnfTypes.lean.expected.out
@@ -3,9 +3,9 @@ mkConstTuple : {α : Type u_1} → α → Nat → ◾
 Fin.add : {n : Nat} → Fin ◾ → Fin ◾ → Fin ◾
 Vec.cons : {α : Type u} → {n : Nat} → α → Vec α ◾ → Vec α ◾
 Eq.rec : {α : Sort u_1} → {a : α} → {motive : α → ◾ → Sort u} → motive ◾ ◾ → {a : α} → ◾ → motive ◾ ◾
-GetElem.getElem : {cont : Type u} →
+GetElem.getElem : {coll : Type u} →
   {idx : Type v} →
-    {elem : Type w} → {dom : cont → idx → Prop} → [self : GetElem cont idx elem ◾] → cont → idx → ◾ → elem
+    {elem : Type w} → {valid : coll → idx → Prop} → [self : GetElem coll idx elem ◾] → coll → idx → ◾ → elem
 Term.constFold : {ctx : List Ty} → {ty : Ty} → Term ◾ ◾ → Term ◾ ◾
 Term.denote : {ctx : List Ty} → {ty : Ty} → Term ◾ ◾ → HList ◾ ◾ → ◾
 HList.get : {α : Type u_1} → {β : α → Type u_2} → {is : List α} → {i : α} → HList β ◾ → Member ◾ ◾ → β ◾


src/Init/GetElem.lean
```diff
@@ -0,0 +1,108 @@
...
+  * `arr[i]?`: returns `none` if the side goal is false
+  * `arr[i]'h`: uses `h` to prove the side goal
+  -/
+  getElem (xs : cont) (i : idx) (h : dom xs i) : elem
+
+  getElem! [Inhabited elem] (xs : cont) (i : idx) [Decidable (dom xs i)] : elem :=
+    if h : _ then getElem xs i h else outOfBounds
+
+  getElem? (xs : cont) (i : idx) [Decidable (dom xs i)] : Option elem :=
+    if h : _ then some (getElem xs i h) else none
```
tydeu, line 47: I would suggest defining `getElem!` in terms of `getElem?`. This allows implementers which need a specialized `getElem?` (e.g., the map use case) to get `getElem!` for free. It is also currently the canonical implementation for such maps.

```suggestion
  getElem? (xs : cont) (i : idx) [Decidable (dom xs i)] : Option elem :=
    if h : _ then some (getElem xs i h) else none

  getElem! [Inhabited elem] (xs : cont) (i : idx) [Decidable (dom xs i)] : elem :=
    match getElem? xs i h with | some e => e | none => outOfBounds
```

joehendrix: The current `getElem!` was not implemented this way, but I've rebased off `nightly-with-mathlib` so we can see if this approach has downstream consequences.

src/Init/GetElem.lean
```diff
@@ -0,0 +1,113 @@
...
+  * `arr[i]?`: returns `none` if the side goal is false
+  * `arr[i]'h`: uses `h` to prove the side goal
+  -/
+  getElem (xs : cont) (i : idx) (h : dom xs i) : elem
+
+  getElem? (xs : cont) (i : idx) [Decidable (dom xs i)] : Option elem :=
+    if h : _ then some (getElem xs i h) else none
+
+  getElem! [Inhabited elem] (xs : cont) (i : idx) [Decidable (dom xs i)] : elem :=
+    match getElem? xs i h with | some e => e | none => outOfBounds
```
tydeu, line 52:
```suggestion
    match getElem? xs i with | some e => e | none => outOfBounds
```

Oops! Typo'd.

joehendrix:
!bench

src/Init/GetElem.lean
```diff
@@ -0,0 +1,168 @@
...
+of `GetElem cont idx elem dom`. Here `elem` is the type of `xs[i]`, while
+`dom` is whatever proof side conditions are required to make this applicable.
+For example, the instance for arrays looks like
+`GetElem (Array α) Nat α (fun xs i => i < xs.size)`.
+
+The proof side-condition `dom xs i` is automatically dispatched by the
+`get_elem_tactic` tactic, which can be extended by adding more clauses to
+`get_elem_tactic_trivial`.
+-/
+class GetElem (cont : Type u) (idx : Type v) (elem : outParam (Type w)) (dom : outParam (cont → idx → Prop)) where
```
david-christiansen, line 25: While we're in the process of rearranging things, should we consider new names for `cont` and `dom`? My head always maps `cont` to "continuation" rather than "container". Perhaps `coll` or `container` or `collection`? Likewise, while `dom` as "domain" makes sense, `valid` seems even easier to interpret to me.

src/Init/GetElem.lean
```diff
@@ -0,0 +1,168 @@
...
+  * `arr[i] : α` where `arr : Array α` and `i : Nat` or `i : USize`:
+    does array indexing with no bounds check and a proof side goal `i < arr.size`.
+  * `l[i] : α` where `l : List α` and `i : Nat`: index into a list,
+    with proof side goal `i < l.length`.
+  * `stx[i] : Syntax` where `stx : Syntax` and `i : Nat`: get a syntax argument,
+    no side goal (returns `.missing` out of range)
+
+  There are other variations on this syntax:
+  * `arr[i]`: proves the proof side goal by `get_elem_tactic`
+  * `arr[i]!`: panics if the side goal is false
```
david-christiansen, line 42:
```suggestion
  * `arr[i]!`: panics if the side goal is false using `getElem!` in this class
```

david-christiansen: But I think actually it just expands to `getElem!`, which is intended to panic when the value isn't there, right? That is, the side goal is not checked anymore? In that case,

```suggestion
  * `arr[i]!`: invokes `GetElem.getElem!`, which panics if the index is not found
```

src/Init/GetElem.lean
```diff
@@ -0,0 +1,168 @@
...
+  * `arr[i]?`: returns `none` if the side goal is false
```
david-christiansen, line 43:
```suggestion
  * `arr[i]?`: expands to `GetElem.getElem?`, which returns `none` if the index is not found
```

src/Lean/Data/HashMap.lean
```diff
@@ -212,6 +212,8 @@ def insertIfNew (m : HashMap α β) (a : α) (b : β) : HashMap α β × Option
 instance : GetElem (HashMap α β) α (Option β) fun _ _ => True where
```
david-christiansen, line 212: Is this the right instance? I had thought that the plan was to make `getElem!` and `getElem?` overridable so that they could do something smarter than "check the side goal, then grab the element", which would allow the element type to efficiently be `β`, so that it would work like array indexing. Did I misunderstand?

tydeu: @david-christiansen The `Option β` instance already pre-existed this PR. Based on this PR's description, I believe the plan is to add the desired `β` instance in a follow-up PR and leave this just as a simple refactor.

david-christiansen: My bad :) Thanks!

src/Init/GetElem.lean
```diff
@@ -0,0 +1,171 @@
...
+  important implementations:
+  * `arr[i] : α` where `arr : Array α` and `i : Nat` or `i : USize`:
+    does array indexing with no bounds check and a proof side goal `i < arr.size`.
+  * `l[i] : α` where `l : List α` and `i : Nat`: index into a list,
+    with proof side goal `i < l.length`.
+  * `stx[i] : Syntax` where `stx : Syntax` and `i : Nat`: get a syntax argument,
+    no side goal (returns `.missing` out of range)
+
+  There are other variations on this syntax:
+  * `arr[i]`: proves the proof side goal by `get_elem_tactic`
```
semorrison, line 44: Slightly confusing here, to have the syntax already described listed under "other variations". I spent a moment looking for the typo!

semorrison (approved):

src/Init/GetElem.lean
joehendrix: Made changes

src/Init/GetElem.lean
joehendrix: Thanks.  I've tried to clarify

PR #3699, opened 2024-03-16
feat: make `registerLabelAttr` consistent with `registerTagAttribute`

thorimur:
This PR modifies the recently-upstreamed `LabelAttribute`s to be consistent with `TagAttribute`s. Specifically, we modify `registerLabelAttr` to
* allow a `validate` argument
* allow an `AttributeApplicationTime` argument
* change the default `applicationTime` to `.afterTypeChecking`
* use the same argument order as `registerTagAttribute`

This also modifies some docstrings to point users to these features.


semorrison:
Just confirming that this is still a draft. The changes do seem nice to have!

thorimur:
@semorrison Apologies, I've been inactive for the past few months! :) If this is still desired I'm happy to finalize it and make it a real PR.

PR #3700, opened 2024-03-16
chore: fix combined CI for mathlib

kim-em:
Previously, if there was a `nightly-testing-YYYY-MM-DD` tag at Std, but not Mathlib, we were erroneously proceeding with Mathlib CI, and hence using a probably-broken version of Mathlib.


nomeata:
Subsumes https://github.com/leanprover/lean4/pull/3631 I presume?

semorrison:
Oh, sorry, completely missed #3631.

PR #3714, opened 2024-03-19
feat: apply_rfl tactic: handle Eq, HEq, better error messages

nomeata:
This implements the first half of #3302: It improves the extensible
`apply_rfl` tactic (the one that looks at `refl` attributes, part of
the `rfl` macro) to

* Check itself and ahead of time that the lhs and rhs are defEq, and give
  a nice consistent error message when they don't (instead of just passing on
  the less helpful error message from `apply Foo.refl`), and using the 
  machinery that `apply` uses to elaborate expressions to highlight diffs
  in implicit arguments.

* Also handle `Eq` and `HEq` (built in) and `Iff` (using the attribute)

Care is taken that, as before, the current transparency setting affects
comparing the lhs and rhs, but not the reduction of the relation

So before we had

```lean
opaque P : Nat → Nat → Prop
@[refl] axiom P.refl (n : Nat) : P n n

/--
error: tactic 'apply' failed, failed to unify
  P ?n ?n
with
  P 42 23
⊢ P 42 23
-/
#guard_msgs in
example : P 42 23 := by apply_rfl

opaque withImplicitNat {n : Nat} : Nat

/--
error: tactic 'apply' failed, failed to unify
  P ?n ?n
with
  P withImplicitNat withImplicitNat
⊢ P withImplicitNat withImplicitNat
-/
#guard_msgs in
example : P (@withImplicitNat 42) (@withImplicitNat 23) := by apply_rfl
```

and with this PR the messages we get are

```
error: tactic 'apply_rfl' failed, The lhs
  42
is not definitionally equal to rhs
  23
⊢ P 42 23
```
resp.
```
error: tactic 'apply_rfl' failed, The lhs
  @withImplicitNat 42
is not definitionally equal to rhs
  @withImplicitNat 23
⊢ P withImplicitNat withImplicitNat
```

A test file checks the various failure modes and error messages.

I believe this `apply_rfl` can serve as the only implementation of
`rfl`, which would then complete #3302, and actually expose these improved
error messages to the user. But as that seems to require a
non-trivial bootstrapping dance, it’ll be separate.

### DIFF
--- a/src/Init/Core.lean
+++ b/src/Init/Core.lean
@@ -823,6 +823,8 @@ theorem iff_iff_implies_and_implies {a b : Prop} : (a ↔ b) ↔ (a → b) ∧ (
 protected theorem Iff.rfl {a : Prop} : a ↔ a :=
   Iff.refl a
 
+macro_rules | `(tactic| rfl) => `(tactic| exact Iff.rfl)
+
 theorem Iff.of_eq (h : a = b) : a ↔ b := h ▸ Iff.rfl
 
 theorem Iff.trans (h₁ : a ↔ b) (h₂ : b ↔ c) : a ↔ c :=
--- a/tests/lean/run/5359.lean
+++ b/tests/lean/run/5359.lean
@@ -16,8 +16,4 @@ variable {α : Nat → Type u} [∀ i, LE (α i)]
 instance : L (S α) Nat α := sorry
 instance : LE (S α) := ⟨fun f g ↦ ∀ i, f i ≤ g i⟩
 
-example : ∀ {a b : S α}, L.coe a ≤ L.coe b ↔ a ≤ b := by
-  -- at some point, rfl would, for Iff.rfl, use `exact` which would enable implicit lambdas
-  fail_if_success rfl
-  -- how, one has to intro first
-  intros; rfl
+example : ∀ {a b : S α}, L.coe a ≤ L.coe b ↔ a ≤ b := by rfl


src/Lean/Meta/Tactic/Rfl.lean
```diff
@@ -45,31 +45,66 @@ initialize registerBuiltinAttribute {
...
+  let t ← whnfR <|← instantiateMVars <|← goal.getType
+  if t.getAppNumArgs < 2 then
+    throwError "rfl can only be used on binary relations, not{indentExpr (← goal.getType)}"
+
+  -- Special case HEq here as it has a different argument order.
+  if t.isAppOfArity ``HEq 4 then
+    let gs ← goal.applyConst ``HEq.refl
+    unless gs.isEmpty do
+      throwError MessageData.tagged `Tactic.unsolvedGoals <| m!"unsolved goals\n{
+        goalsToMessageData gs}"
```
digama0, line 71: this seems like it should be a function somewhere

src/Lean/Meta/Tactic/Rfl.lean
```diff
@@ -45,31 +45,66 @@ initialize registerBuiltinAttribute {
...
+    let s ← saveState
+    let mut ex? := none
+    for lem in ← (reflExt.getState (← getEnv)).getMatch rel reflExt.config do
+      try
+        let gs ← goal.apply (← mkConstWithFreshMVarLevels lem)
+        if gs.isEmpty then return () else
+          throwError MessageData.tagged `Tactic.unsolvedGoals <| m!"unsolved goals\n{
+            goalsToMessageData gs}"
+      catch e =>
+        ex? := ex? <|> (some (← saveState, e)) -- stash the first failure of `apply`
```
digama0, line 101: I think this is a case of unfortunate lifting, ideally `saveState` would be called only if `ex?` is none.

nomeata: I noticed as well, but didn’t want to change too much initially.

```suggestion
        unless ex2.isNone do
          ex? := some (← saveState, e) -- stash the first failure of `apply`
```

nomeata:
Ah, interesting: The `eq_refl` tactic reduces the goal with the current transparency (`whnf`) before it checks if it is an `Eq`, while the extensible `apply_rfl` tactic reduces the goal only with reducible transparency (`whnfR`) before consulting the the `@[refl]` lemmas. This PR currently tries to do the `whnfR` same in both cases, but of course that breaks existing proofs.

So I should adjust this code to do these things as well (reduce with `whnf`, check if it is `Eq, `Iff` or `HEq`, handle that, else re-reduce with `whnfR`, and handle that) to not have to chance down breakages.

nomeata:
Speedcenter results insignificant: http://speed.lean-fro.org/mathlib4/compare/f60b59ba-cc12-4fc8-889c-64af4af5b420/to/88f7233a-a865-471e-99e7-4ead3df9debf

PR #3715, opened 2024-03-19
chore: test results as job summary

Kha:
Tired of scrolling through and parsing the test output myself

### DIFF
--- a/doc/flake.nix
+++ b/doc/flake.nix
@@ -27,7 +27,7 @@
         src = inputs.mdBook;
         cargoDeps = drv.cargoDeps.overrideAttrs (_: {
           inherit src;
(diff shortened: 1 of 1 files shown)


nomeata:
This will only help with `full-ci`, I guess, unless we find a way to extract that file from a (failed) nix build.

Kha:
@nomeata It does now list the failing tests but actually showing the stdout output would require some ctest-specific changes to the action

image

nomeata:
Well, it is an improvement, so merge and maybe iterate there?

We could additionally configure ctest to not print successful tests in the build log. Advantage: Errors are easier to find. Downside: no progress indication when looking at live logs.

Kha:
> Well, it is an improvement, so merge and maybe iterate there?

Agreed

> We could additionally configure ctest to not print successful tests in the build log. Advantage: Errors are easier to find. Downside: no progress indication when looking at live logs.

Too bad `--progress` doesn't seem to work in Actions

PR #3716, opened 2024-03-19
refactor: rfl tactic: do not use Kernel.isDefEq

nomeata:
Sebastian mentioned that the use of the kernel defeq was to work around
a performance issue that was fixed since. Let's see if we can do
without.


nomeata:
Ugh, I was at most expecting performance differences, but this seems to actually break proofs in mathlib (in `Mathlib/NumberTheory/ADEInequality.lean` and `Mathlib/GroupTheory/CommutingProbability.lean` according to https://github.com/leanprover-community/mathlib4/actions/runs/8342372573).

Not sure if this is worth tracking down at the moment, although we should eventally if we want lean to be able to work without C++ code outside the (then optional?) checker.

arthur-adjedj:
This particular error (as well as most of the other ones possibly) is due to the fact that the kernel doesn't know about transparency modes and reduces everything it can, whereas the default transparency mode in the elaborator doesn't unfold `@[irreducible]`  constants, In this particular case, I believe [`Rat.mul`](https://leanprover-community.github.io/mathlib4_docs/Std/Data/Rat/Basic.html#Rat.mul) to be the culprit of the failure of defeq.

Apart from this difference, I believe the elaborator and the kernel should in theory behave the same way on terms with no meta-variables.

nomeata:
Currently it only uses the kernel in transparency mode default and all
```
    return (← getTransparency) matches TransparencyMode.default | TransparencyMode.all
```
So you are saying that it’s a bug that currently `TransparencyMode.default` is included here, and if this check were only be true with `TransparencyMode.all`, then this PR would likely go through?

arthur-adjedj:
I do not know if this behaviour was originally intended, or if it is a bug, but I'd say that this is the cause of the errors appearing in Mathlib with this PR. This is one of the reasons why `by rfl` and `rfl` currently behave differently:
```lean
@[irreducible]
def foo := 2

example : foo = 2 := rfl --fails
example : foo = 2 := by rfl --works
```

semorrison:
@nomeata, I wanted to look at what happened in Mathlib, but after `full-ci` I'm mystified by the build failures.

nomeata:
Oh, looks like fallout from #3717, I'll have to pull master into the PR.

nomeata:
@semorrison all yours to test!

semorrison:
I couldn't get Mathlib going at all. Would you be able to rebase onto `nightly-with-mathlib`? (I tried and failed. :-)

nomeata:
I have rebased and reopened this as #3772; I guess the `lean-pr-testing-3716` branch easily gets confused when the Lean PR merges master or rebases, because it will not rebase the testing branch accordingly.

PR #3717, opened 2024-03-19
chore: replace shell.nix with a devShell in flake.nix

nomeata:
as a side effect this pins the “old nixpkgs” revision used by CI for release builds. 
(Not that that old branch is likely to change a lot…)


nomeata:
@Kha, this also touches some of our release machinery in `ci.nix`, so maybe good if you have a critical look, as you know the intricacies of the various builds better.

Kha (approved):

PR #3718, opened 2024-03-19
refactor: back `rfl` tactic primarily via `apply_rfl`

nomeata:
building upon #3714, this (almost) implements the second half of #3302.

The main effect is that we now get a better error message when `rfl` fails. For
```lean
example : n+1+m = n + (1+m) := by rfl
```
instead of the wall of text
```
The rfl tactic failed. Possible reasons:
- The arguments of the relation are not equal.
Try using the reflexivity lemma for your relation explicitly, e.g. `exact Eq.refl _` or
`exact HEq.rfl` etc.
n m : Nat
⊢ n + 1 + m = n + (1 + m)
```
we now get
```
error: tactic 'rfl' failed, the left-hand side
  n + 1 + m
is not definitionally equal to the right-hand side
  n + (1 + m)
n m : Nat
⊢ n + 1 + m = n + (1 + m)
```

Unfortunately, because of very subtle differences in semantics (which transparency setting is used when reducing the goal and whether the “implicit lambda” feature applies) I could not make this simply the only `rfl` implementation. So `rfl` remains a macro and is still expanded to `eq_refl` (difference transparency setting) and `exact Iff.rfl` and `exact HEq.rfl` (implicit lambda) to not break existing code. This can be revised later, so this still closes: #3302.

A user might still be puzzled *why* to terms are not defeq. Explaining that better (“reduced to… and reduces to… etc.”) would also be great, but that’s not specific to `rfl`, so better left for some other time.

### DIFF
--- a/tests/lean/run/4644.lean
+++ b/tests/lean/run/4644.lean
@@ -11,9 +11,9 @@ def check_sorted [x: LE α] [DecidableRel x.le] (a: Array α): Bool :=
   sorted_from_var a 0
 
 /--
-error: tactic 'rfl' failed, The lhs
+error: tactic 'rfl' failed, the left-hand side
   check_sorted #[0, 3, 3, 5, 8, 10, 10, 10]
-is not definitionally equal to rhs
+is not definitionally equal to the right-hand side
   true
 ⊢ check_sorted #[0, 3, 3, 5, 8, 10, 10, 10] = true
 -/
--- a/tests/lean/run/wfirred.lean
+++ b/tests/lean/run/wfirred.lean
@@ -31,9 +31,9 @@ example : foo (n+1) = foo n := rfl
 
 -- also for closed terms
 /--
-error: tactic 'rfl' failed, The lhs
+error: tactic 'rfl' failed, the left-hand side
   foo 0
-is not definitionally equal to rhs
+is not definitionally equal to the right-hand side
   0
 ⊢ foo 0 = 0
 -/
@@ -42,9 +42,9 @@ example : foo 0 = 0 := by rfl
 
 -- It only works on closed terms:
 /--
-error: tactic 'rfl' failed, The lhs
+error: tactic 'rfl' failed, the left-hand side
   foo (n + 1)
-is not definitionally equal to rhs
+is not definitionally equal to the right-hand side
   foo n
 n : Nat
 ⊢ foo (n + 1) = foo n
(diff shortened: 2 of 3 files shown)


nomeata:
(Likely needs a stage0 update after 846300038f0e692fe08ea83d410f96c233624489)

nomeata:
I am a bit confused now. After a stage0 update, so that
```
@[builtin_tactic Lean.Parser.Tactic.applyRfl] def evalApplyRfl
```
kicks in I would have expected that with
```
syntax (name := applyRfl) "rfl" : tactic
```
the new code is used reliably.

Nevertheless
```
theorem add_le_iff_le_sub (a b c : Int) : a + b ≤ c ↔ a ≤ c - b := by
  conv =>
    lhs
    rw [← Int.add_zero c, ← Int.sub_self (-b), Int.sub_eq_add_neg, ← Int.add_assoc, Int.neg_neg,
      Int.add_le_add_iff_right]
  -- rfl here works, so why does the rfl within `conv` no work?
```
fails with 
```
unsolved goals
abc: Int
⊢ a ≤ c + -b ↔ a ≤ c - b
```

It seems to me that somehow the 
```
   evalTactic (← `(tactic| try rfl))
```
in 
```
@[builtin_tactic Lean.Parser.Tactic.Conv.conv] def evalConv : Tactic := fun stx => do
```
is broken.

Maybe I can work around that, do another stage0 update, and then it works…

nomeata:
I still see mathlib failures

```
 Error: ./././Mathlib/Analysis/NormedSpace/Star/ContinuousFunctionalCalculus/Instances.lean:223:35: error: type mismatch
  HEq.rfl
has type
  HEq ?m.240013 ?m.240013 : Prop
but is expected to have type
  IsSelfAdjoint 2 : Prop
```
which I did not expect. It seems that the `macro_rules` are duplicated in std4, proposing a fix at https://github.com/leanprover/std4/pull/702.

nomeata:
Now that’s a perf benchmark with lots of changes for a change:
http://speed.lean-fro.org/mathlib4/compare/74d6d12b-3209-44e1-b7d2-7265a9de7635/to/ef55fa22-f97d-4d12-8940-a2430e8a147d

Many modules benefit, a few regress, one badly (+213.3B instructions). Overall good (Build instructions -1.823 %). I have a hunch where we are doing extra work now and an idea of how to improve.

PR #3719, opened 2024-03-19
feat: upstream rw? tactic

joehendrix:
This updates the rw? tactic from Mathlib to use lazy discriminator trees and upstreams it.

### DIFF
--- a/src/Lean/Meta/Tactic/Rewrites.lean
+++ b/src/Lean/Meta/Tactic/Rewrites.lean
@@ -129,20 +129,18 @@ structure RewriteResult where
   weight : Nat
   /-- The result from the `rw` tactic. -/
   result : Meta.RewriteResult
-  /-- Pretty-printed result. -/
-  -- This is an `Option` so that it can be computed lazily.
-  ppResult? : Option String
   /-- The metavariable context after the rewrite.
   This needs to be stored as part of the result so we can backtrack the state. -/
   mctx : MetavarContext
+  rfl? : Bool
 
 /-- Update a `RewriteResult` by filling in the `rfl?` field if it is currently `none`,
 to reflect whether the remaining goal can be closed by `with_reducible rfl`. -/
-def RewriteResult.computeRfl (r : RewriteResult) : MetaM Bool := do
+def computeRfl (mctx : MetavarContext) (res : Meta.RewriteResult) : MetaM Bool := do
   try
-    withoutModifyingState <| withMCtx r.mctx do
+    withoutModifyingState <| withMCtx mctx do
       -- We use `withReducible` here to follow the behaviour of `rw`.
-      withReducible (← mkFreshExprMVar r.result.eNew).mvarId!.applyRfl
+      withReducible (← mkFreshExprMVar res.eNew).mvarId!.applyRfl
       -- We do not need to record the updated `MetavarContext` here.
       pure true
   catch _e =>
@@ -151,11 +149,8 @@ def RewriteResult.computeRfl (r : RewriteResult) : MetaM Bool := do
 /--
 Pretty print the result of the rewrite.
 -/
-def RewriteResult.ppResult (r : RewriteResult) : MetaM String :=
-  if let some pp := r.ppResult? then
-    return pp
-  else
-    return (← ppExpr r.result.eNew).pretty
+private def RewriteResult.ppResult (r : RewriteResult) : MetaM String :=
+  return (← ppExpr r.result.eNew).pretty
 
 
 /-- Should we try discharging side conditions? If so, using `assumption`, or `solve_by_elim`? -/
@@ -184,21 +179,19 @@ def rwLemma (ctx : MetavarContext) (goal : MVarId) (target : Expr) (side : SideC
     let some result ← some <$> goal.rewrite target expr symm <|> pure none
       | return none
     if result.mvarIds.isEmpty then
-      return some {
-        expr,
-        symm,
-        weight,
-        result,
-        ppResult? := none,
-        mctx := ← getMCtx
-      }
+      let mctx ← getMCtx
+      let rfl? ← computeRfl mctx result
+      return some { expr, symm, weight, result, mctx, rfl? }
     else
       -- There are side conditions, which we try to discharge using local hypotheses.
-      match (← match side with
+      let discharge ←
+        match side with
         | .none => pure false
-        | .assumption => ((result.mvarIds.mapM fun m => m.assumption) >>= fun _ => pure true) <|> pure false
-        | .solveByElim => (solveByElim result.mvarIds >>= fun _ => pure true) <|> pure false) with
-      | false => return none
+        | .assumption => ((fun _ => true) <$> result.mvarIds.mapM fun m => m.assumption) <|> pure false
+        | .solveByElim => (solveByElim result.mvarIds >>= fun _ => pure true) <|> pure false
+      match discharge with
+      | false =>
+        return none
       | true =>
         -- If we succeed, we need to reconstruct the expression to report that we rewrote by.
         let some expr := rewriteResultLemma result | return none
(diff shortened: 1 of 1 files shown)


src/Lean/Elab/Tactic/Rewrites.lean
```diff
@@ -0,0 +1,70 @@
...
+        throwError "Could not find any lemmas which can rewrite the hypothesis {← f.getUserName}"
+      for r in results do withMCtx r.mctx do
+        Tactic.TryThis.addRewriteSuggestion tk [(r.expr, r.symm)]
+          r.result.eNew (loc? := .some (.fvar f)) (origSpan? := ← getRef)
+      if let some r := results[0]? then
+        setMCtx r.mctx
+        let replaceResult ← goal.replaceLocalDecl f r.result.eNew r.result.eqProof
+        replaceMainGoal (replaceResult.mvarId :: r.result.mvarIds)
+    -- See https://github.com/leanprover/lean4/issues/2150
+    do withMainContext do
```
semorrison, line 55:
```suggestion
    do
```
That issue has been fixed now.

src/Lean/Meta/Tactic/Rewrites.lean
```diff
@@ -0,0 +1,378 @@
...
+  /-- The result from the `rw` tactic. -/
+  result : Meta.RewriteResult
+  /-- Pretty-printed result. -/
+  -- This is an `Option` so that it can be computed lazily.
+  ppResult? : Option String
+  /-- The metavariable context after the rewrite.
+  This needs to be stored as part of the result so we can backtrack the state. -/
+  mctx : MetavarContext
+
+/-- Update a `RewriteResult` by filling in the `rfl?` field if it is currently `none`,
```
semorrison, line 139: I see you've removed the cache for the `computeRfl` result in `RewriteResult`. I presume that's reasonable, but could you explain? This doc-string will need updating.

src/Lean/Meta/Tactic/Rewrites.lean
```diff
@@ -0,0 +1,378 @@
...
+/-- Shortcut for calling `solveByElim`. -/
+def solveByElim (goals : List MVarId) (depth : Nat := 6) : MetaM PUnit := do
+  -- There is only a marginal decrease in performance for using the `symm` option for `solveByElim`.
+  -- (measured via `lake build && time lake env lean test/librarySearch.lean`).
+  let cfg : SolveByElimConfig := { maxDepth := depth, exfalso := false, symm := true }
+  let ⟨lemmas, ctx⟩ ← mkAssumptionSet false false [] [] #[]
+  let [] ← SolveByElim.solveByElim cfg lemmas ctx goals
+    | failure
+
+def rwLemma (ctx : MetavarContext) (goal : MVarId) (target : Expr) (side : SideConditions := .solveByElim) (lem : Expr ⊕ Name) (symm : Bool) (weight : Nat) : MetaM (Option RewriteResult) :=
```
semorrison, line 176:
```suggestion
def rwLemma (ctx : MetavarContext) (goal : MVarId) (target : Expr) (side : SideConditions := .solveByElim) 
    (lem : Expr ⊕ Name) (symm : Bool) (weight : Nat) : MetaM (Option RewriteResult) :=
```
That's a long line even by Lean's standards. :-)

src/Lean/Meta/Tactic/Rewrites.lean
```diff
@@ -0,0 +1,378 @@
...
+            ppResult? := none,
+            mctx := ← getMCtx
+          }
+
+/--
+Find keys which match the expression, or some subexpression.
+
+Note that repeated subexpressions will be visited each time they appear,
+making this operation potentially very expensive.
+It would be good to solve this problem!
```
semorrison, line 220: We could potentially address this now. I think the right approach would be to generalize/duplicate `Lean.Util.FindExpr` to allow accumulating a value rather than just returning a Bool.

Up to you (or I can take a look).

src/Lean/Meta/Tactic/Rewrites.lean
```diff
@@ -0,0 +1,378 @@
...
+
+/--
+Find lemmas which can rewrite the goal.
+
+This core function returns a monadic list, to allow the caller to decide how long to search.
+See also `rewrites` for a more convenient interface.
+-/
+-- We need to supply the current `MetavarContext` (which will be reused for each lemma application)
+-- because `MLList.squash` executes lazily,
+-- so there is no opportunity for `← getMCtx` to record the context at the call site.
```
semorrison, line 250:
```suggestion
```

src/Lean/Meta/Tactic/Rewrites.lean
```diff
@@ -0,0 +1,378 @@
...
+      args.foldlM (fun acc arg => return acc ++ (← getSubexpressionMatches op (← inferType arg)))
+        (← getSubexpressionMatches op body).reverse)
+  | _ =>
+    let init := ((← op e).reverse)
+    e.foldlM (init := init) (fun a f => return a ++ (← getSubexpressionMatches op f))
+
+/--
+Find lemmas which can rewrite the goal.
+
+This core function returns a monadic list, to allow the caller to decide how long to search.
```
semorrison, line 245:
```suggestion
```

src/Lean/Meta/Tactic/Rewrites.lean
```diff
@@ -0,0 +1,378 @@
...
+        deduped := deduped.push (l, s, w)
+        backward := backward.insert l
+    else
+      if ¬ forward.contains l then
+        deduped := deduped.push (l, s, w)
+        forward := forward.insert l
+
+  trace[Tactic.rewrites.lemmas] m!"Candidate rewrite lemmas:\n{deduped}"
+
+  -- Lift to a monadic list, so the caller can decide how much of the computation to run.
```
semorrison, line 280:
```suggestion
```

src/Lean/Meta/Tactic/Rewrites.lean
```diff
@@ -0,0 +1,378 @@
...
+  let hyps := hyps.map fun ⟨hyp, symm, weight⟩ => (Sum.inl hyp, symm, weight)
+  let lemmas := deduped.map fun ⟨lem, symm, weight⟩ => (Sum.inr lem, symm, weight)
+  pure <| hyps ++ lemmas
+
+structure FinRwResult where
```
semorrison, line 285: What is the `Fin` in the name referring to?

tests/lean/run/rewrites.lean
```diff
@@ -0,0 +1,126 @@
+attribute [refl] Eq.refl
+
+private axiom test_sorry : ∀ {α}, α
+
+-- To see the (sorted) list of lemmas that `rw?` will try rewriting by, use:
+-- set_option trace.Tactic.rewrites.lemmas true
+
+/--
+info: Try this: rw [@List.map_append]
```
semorrison, line 9: Ah --- I was just noticing this while otherwise happily using `rw_search` to repair some proofs in Mathlib. There is a pretty unnecessary `@` at the front of each lemma name here, because of the way we've going from `Name` to `Expr`. Do you see a way to avoid having this initial `@`?

src/Lean/Meta/Tactic/Rewrites.lean
joehendrix: How much of a problem has this been in practice?

src/Lean/Meta/Tactic/Rewrites.lean
joehendrix: As far as I can tell, both the `rfl?` nor `ppResult?` fields cached values that were only read once, as such it didn't seem worth caching them.

Are there users of this datatype downstream where caching would be useful?  I can re-add if needed.

tests/lean/run/rewrites.lean
nomeata: Blind guess: use https://leanprover-community.github.io/mathlib4_docs/Std/Lean/Delaborator.html#Lean.ppConst in the right place?

src/Lean/Meta/Tactic/Rewrites.lean
joehendrix: This was a bad refactor

tests/lean/run/rewrites.lean
joehendrix: @nomeata Thanks.  I just tested it and it seems to work by upstreaming and inserting into `Tactic.TryThis.addRewrite suggestion`.   I'll make a separate PR for it since it's largely orthogonal to this work.

PR #3723, opened 2024-03-21
chore: remove @[inline] from BitVec.cast

kim-em:
One of the main reasons to have `BitVec.cast` is to ensure we don't see strange Eq.rec goals with strange motives, and instead try to simplify away `BitVec.cast` by combining them with other BitVec operations. Having it be @[inline] means that unfolding definitions using it immediately produces less controllable `Eq.rec` terms.

### DIFF
--- a/src/Init/Data/BitVec/Basic.lean
+++ b/src/Init/Data/BitVec/Basic.lean
@@ -371,7 +371,7 @@ end relations
 section cast
 
 /-- `cast eq i` embeds `i` into an equal `BitVec` type. -/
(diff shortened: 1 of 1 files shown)


digama0:
Are you sure about this explanation? `@[inline]` only affects the compiler, not the elaborator (which would be looking for `@[reducible]` instead). Putting `@[inline]` on a definition should have no effect on tactic behavior, and since this is a compiler no-op we definitely want to inline it at runtime.

semorrison:
Indeed. I will investigate further!

PR #3725, opened 2024-03-21
fix: ignore unused alternatives in `Ord` derive handler

### LINKED ISSUE #3706: deriving an ord instance fails with redundant alternative
### Prerequisites

### Description

deriving an ord instance fails on a GADT style inductive data type

### Context

https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Redundant.20alternative.20in.20deriving.20Ord/near/427262837

### Steps to Reproduce

Create a lean file with the following

```
inductive Ty where
  | int
  | bool
inductive Expr : Ty → Type where
  | a : Expr .int
  | b : Expr .bool
deriving Ord
```

**Expected behavior:**
Ord instance is created and no error messages

**Actual behavior:**
The `deriving Ord` line has the error `redundant alternative`

### Versions

[Output of `#eval Lean.versionString` or of `lean --version` in the folder that the issue occured in]  4.7.0-rc2
[OS version] NixOS 24.05pre-git (Uakari) x86_64

arthur-adjedj:
Closes #3706

This derive handler's implementation is very similar to `BEq`'s, which already ignores unused alternative so as to work correctly on indexed inductive types. This PR simply implements the same solution as the one present in [`BEq.lean`](https://github.com/leanprover/lean4/blob/2c15cdda044e77bb8c3937c63501850790e60dc6/src/Lean/Elab/Deriving/BEq.lean#L94).

After some tests, it doesn't seem like any other derive handler present in Core suffers from the same issue (though some handlers don't work on indexed inductives for other reasons).

### DIFF
--- a/src/Lean/Elab/Deriving/Ord.lean
+++ b/src/Lean/Elab/Deriving/Ord.lean
@@ -84,6 +84,7 @@ def mkMutualBlock (ctx : Context) : TermElabM Syntax := do
   for i in [:ctx.typeInfos.size] do
     auxDefs := auxDefs.push (← mkAuxFunction ctx i)
   `(mutual
+     set_option match.ignoreUnusedAlts true
      $auxDefs:command*
     end)
 
--- a/tests/lean/run/3706.lean
+++ b/tests/lean/run/3706.lean
@@ -1,0 +1,13 @@
+/--Ensure that `deriving Ord` works on indexed inductive types.-/
+inductive Ty where
+  | int
+  | bool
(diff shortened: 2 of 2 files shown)


arthur-adjedj:
awaiting-review

arthur-adjedj:
wip

arthur-adjedj:
WIP

arthur-adjedj:
awaiting-review

nomeata (approved):
If it’s good for BEq, then surely also for Ord. Thanks!

PR #3727, opened 2024-03-21
feat: add BitVec.flattenList and lemmas

kim-em:
This PR adds `BitVec.flattenList`, which concatenates a list of bitvectors of a common width into a single bitvector, together with lemmas describing its bits: `getLsbD_flattenList` and `getMsbD_flattenList` compute an individual bit in terms of the corresponding element of the list, and `extractLsb_flattenList` describes extracting a contiguous range that falls within a single element. For efficient execution, `flattenList` is replaced at runtime via `@[csimp]` with a divide-and-conquer implementation costing `O(n * L * log L)` rather than the `O(n * L²)` of a naive left fold (≈900x faster at a million elements), while keeping `O(log L)` recursion depth so it remains stack-safe.


src/Init/Data/Nat/Lemmas.lean
```diff
@@ -800,6 +801,53 @@ theorem mul_add_mod (m x y : Nat) : (m * x + y) % m = y % m := by
...
+theorem mod_eq_sub_mul_div (a b : Nat) : a % b = a - b * (a / b) := by
+  have := Nat.mod_add_div a b
+  have := Nat.mul_div_le a b
+  omega
+
+theorem lt_mul_div_self_add {x k : Nat} (h : 0 < k) : x < k * (x / k) + k := by
+  rw [← Int.ofNat_lt]
+  simpa using Int.lt_mul_ediv_self_add (by omega)
+
+theorem mul_sub_mod {x n p : Nat} (h : x < n * p) : (n * p - (x + 1)) % n = n - (x % n + 1) := by
```
semorrison, line 813: @fpvandoorn gave a different proof at https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/Nat.2Emul_sub_mod/near/428107094; perhaps should refactor to use that.

src/Init/Data/BitVec/Basic.lean
```diff
@@ -575,6 +575,17 @@ def replicate : (i : Nat) → BitVec w → BitVec (w*i)
...
+/-! ### flatten -/
+
+/-- Flatten a list of bitvectors into one bitvector. -/
+protected def flatten {n : Nat} (xs : List (BitVec n)) : BitVec (n * xs.length) :=
+  match xs with
+  | [] => 0#0
+  | x :: rest =>
+    have h : n + n * List.length rest = n * List.length (x :: rest) := by
+      simp [List.length_cons, Nat.mul_add, Nat.add_comm]
+    BitVec.cast h (x ++ (BitVec.flatten rest))
```
joehendrix, line 587: I think this should either be `List.flattenBV` or `flattenList`.

I also think this should be made tail recursive.

src/Init/Data/Nat/Lemmas.lean
semorrison: I've now included a shorter proof by Junyan Xu.

src/Init/Data/BitVec/Basic.lean
semorrison: I've renamed to `flattenList`.

I'm a little skeptical of making it tail-recursive. Of course, it's doable, but comes at a time cost, and I'm not sure that long lists of bitvectors arise in practice.

PR #3738, opened 2024-03-22
feat: functional induction for structural recursion

nomeata:
This extends `derive_functional_induction` to work with structural recursion as well.

It produces the less general, more concrete induction rule where the induction hypothesis is
specialized for every argument of the recursive call, not just the the one that the function
is recursing on.

Care is taken so that the induction principle and it's motive take the arguments in the same
order as the original function.

While I was it, also makes sure that the order of the cases in the induction principle matches
the order of recursive calls in the function better.

### DIFF
--- a/src/Lean/Meta/Tactic/FunInd.lean
+++ b/src/Lean/Meta/Tactic/FunInd.lean
@@ -346,7 +346,12 @@ partial def collectIHs (is_wf : Bool) (fn : Expr) (oldIH newIH : FVarId) (e : Ex
     if let some (e', args) ← isPProdProjWithArgs oldIH newIH e then
       let args' ← args.mapM (foldCalls is_wf fn oldIH newIH)
       let ihs ← args.concatMapM (collectIHs is_wf fn oldIH newIH)
-      let e' := mkAppN e' args'
+      let t ← whnf (← inferType e')
+      let arity ← forallTelescopeReducing t fun xs t' => do
+        unless t'.getAppFn.isFVar do -- we expect an application of the `motive` FVar here
+          throwError m!"Unexpected type {t} of {e}: Reduced to application of {t'.getAppFn}"
+        pure xs.size
+      let e' := mkAppN e' args'[:arity]
       let eTyp ← inferType e'
       -- The inferred type that comes out of motive projections has beta redexes
       let eType' := eTyp.headBeta
--- a/tests/lean/run/funind_structural.lean
+++ b/tests/lean/run/funind_structural.lean
@@ -124,6 +124,7 @@ error: Function idAcc is defined in a way not supported by functional induction,
 #guard_msgs in
 derive_functional_induction idAcc
 
+namespace TreeExample
 
 inductive Tree (β : Type v) where
   | leaf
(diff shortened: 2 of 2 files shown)


nomeata:
It seems to be working, but it needs some serious refactoring.

leodemoura:
In the following example, the generated principle has only 2 cases instead of 4. If we force `Tree.insert` to be defined using well-founded recursion (e.g., using `termination_by t`), we get 4 cases as expected.

```lean
inductive Tree (β : Type v) where
  | leaf
  | node (left : Tree β) (key : Nat) (value : β) (right : Tree β)

def Tree.insert (t : Tree β) (k : Nat) (v : β) : Tree β :=
  match t with
  | leaf => node leaf k v leaf
  | node left key value right =>
    if k < key then
      node (left.insert k v) key value right
    else if key < k then
      node left key value (right.insert k v)
    else
      node left k v right

derive_functional_induction Tree.insert
/-
Tree.insert.induct.{u_1} {β : Type u_1} (motive : Tree β → Nat → β → Prop)
  (case1 : ∀ (k : Nat) (v : β), motive Tree.leaf k v)
  (case2 :
    ∀ (k : Nat) (v : β) (left : Tree β) (key : Nat) (value : β) (right : Tree β),
      motive left k v → motive right k v → motive (left.node key value right) k v)
  (t : Tree β) (k : Nat) (v : β) : motive t k v
-/
#check Tree.insert.induct
```

leodemoura:
The following example fails with a type error
```lean
inductive HList {α : Type v} (β : α → Type u) : List α → Type (max u v)
  | nil  : HList β []
  | cons : β i → HList β is → HList β (i::is)

inductive Member : α → List α → Type
  | head : Member a (a::as)
  | tail : Member a bs → Member a (b::bs)

def HList.get : HList β is → Member i is → β i
  | .cons a as, .head => a
  | .cons _ as, .tail h => as.get h

inductive Ty where
  | nat
  | fn : Ty → Ty → Ty

@[reducible] def Ty.denote : Ty → Type
  | nat    => Nat
  | fn a b => a.denote → b.denote

inductive Term : List Ty → Ty → Type
  | var   : Member ty ctx → Term ctx ty
  | const : Nat → Term ctx .nat
  | plus  : Term ctx .nat → Term ctx .nat → Term ctx .nat
  | app   : Term ctx (.fn dom ran) → Term ctx dom → Term ctx ran
  | lam   : Term (.cons dom ctx) ran → Term ctx (.fn dom ran)
  | let   : Term ctx ty₁ → Term (.cons ty₁ ctx) ty₂ → Term ctx ty₂

def Term.denote : Term ctx ty → HList Ty.denote ctx → ty.denote
  | .var h,     env => env.get h
  | .const n,   _   => n
  | .plus a b,  env => a.denote env + b.denote env
  | .app f a,   env => f.denote env (a.denote env)
  | .lam b,     env => fun x => b.denote (.cons x env)
  | .let a b,   env => b.denote (.cons (a.denote env) env)

derive_functional_induction Term.denote
```

nomeata:
Thanks for testing, will have a look!

PR #3741, opened 2024-03-22
fix: include NUL characters on calls to getLine

### LINKED ISSUE #3546: null characters break `readFile`/`getLine`
### Prerequisites

### Description

Null characters break `IO.FS.readFile`  /  `IO.FS.Handle.getLine`. The null character will trim lines to the null character because of the C implementation of `getLine`. The implementation uses the null-terminated output of `fgets` rather than the buffer to and uses  the null-terminated C string `lean_mk_string` instead of `lean_mk_string_from_bytes` to construct the string. This *is* mentioned in the C docstring of `getLine` (but not in the Lean one). Nonetheless, this is an undesirable behavior.

https://github.com/leanprover/lean4/blob/0362fcea699e55b966c897b240c520f579da7a35/src/runtime/io.cpp#L464-L495

### Steps to Reproduce

```lean
def test : IO Unit := do
  let tmpFile := "readLineBug.tmp"
  let s := "foo\u0000bar"
  IO.FS.writeFile tmpFile s
  IO.println <| ← IO.FS.readFile tmpFile

#eval test -- foo
```

**Expected behavior:**

`IO.FS.readFile` should read a whole file as a `String` including null characters.

**Actual behavior:**

`IO.FS.readFile` does not properly read lines including null characters.

### Impact

axionbuster:
Fixes #3546 ("null characters break `readFile`/`getLine`").

This draft pull request showcases an implementation of `getLine` that would not truncate on the first occurrence of the NUL character (`\0`) and properly seek the newline character (assuming text mode) in case users expect `getLine` to preserve NUL characters in text input.

However, I was concerned that some parts of the Lean runtime might expect NUL-terminated strings. I am new to the codebase, so I'm not sure.

The repeated calls to `ungetc` may negatively affect performance. I'm not sure about this approach, either. I'm open to discussions.

### DIFF
--- a/src/Init/Data/String/Extra.lean
+++ b/src/Init/Data/String/Extra.lean
@@ -24,6 +24,12 @@ def toNat! (s : String) : Nat :=
 @[extern "lean_string_from_utf8_unchecked"]
 opaque fromUTF8Unchecked (a : @& ByteArray) : String
 
+/--
+  Decide whether the byte array `a` represents a proper UTF-8 encoded string.
+-/
+@[extern "lean_bytearray_is_utf8"]
+opaque isUtf8 (a : @& ByteArray) : Bool
+
 /-- Convert the given `String` to a [UTF-8](https://en.wikipedia.org/wiki/UTF-8) encoded byte array. -/
 @[extern "lean_string_to_utf8"]
 opaque toUTF8 (a : @& String) : ByteArray
--- a/src/Init/System/IO.lean
+++ b/src/Init/System/IO.lean
@@ -430,14 +430,13 @@ partial def Handle.readBinToEnd (h : Handle) : IO ByteArray := do
 
 partial def Handle.readToEnd (h : Handle) : IO String := do
   let s ← h.readBinToEnd
-  -- FIXME: Validate UTF-8.
-  return String.fromUTF8Unchecked <| ByteArray.mk (
-    if System.Platform.isWindows then
+  let s := if System.Platform.isWindows then
       -- Remove carriage return (0x0d).
-      Array.filter (λ c => c != UInt8.mk 0x0d) (s.data)
+      ByteArray.mk <| Array.filter (λ c => c != UInt8.mk 0x0d) (s.data)
     else
-      s.data
-  )
+      s
+  -- TODO: Better panic?
+  if String.isUtf8 s then return String.fromUTF8Unchecked s else panic! "invalid UTF-8"
 
 def readBinFile (fname : FilePath) : IO ByteArray := do
   let h ← Handle.mk fname Mode.read
--- a/src/runtime/object.cpp
+++ b/src/runtime/object.cpp
@@ -1618,6 +1618,28 @@ extern "C" LEAN_EXPORT obj_res lean_string_from_utf8_unchecked(b_obj_arg a) {
     return lean_mk_string_from_bytes(reinterpret_cast<char *>(lean_sarray_cptr(a)), lean_sarray_size(a));
 }
 
+extern "C" LEAN_EXPORT uint8_t lean_bytearray_is_utf8(b_obj_arg a) {
+    auto s = reinterpret_cast<char const *>(lean_sarray_cptr(a));
+    size_t const n = lean_sarray_size(a);
+    for (size_t i = 0; i < n;) {
+        // Old value of `i`.
+        auto const j = i;
+        // Apply the following assumptions:
+        //  - If j + 1 == i ("i is incremented by 1") then
+        //      s[j] is either ASCII or does not begin a (valid) code point ("invalid").
+        //  - If s[j] is invalid then it's not ASCII.
+        //  - If invalid, i is incremented by 1.
+        //  - If ASCII, i is incremented by 1.
+        // Thus,
+        //  - s[j] is invalid if and only if i is incremented by 1 and not ASCII.
+        next_utf8(s, n, i);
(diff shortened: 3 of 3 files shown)


axionbuster:
The test has apparently stalled; I cannot stop it.

axionbuster:
@joehendrix hello. A stalled CI test has been running for a few hours. I thought that I had to let someone know.

axionbuster:
It passes all the tests. The test that failed on my computer was apparently known to fail on macOS.

Kha:
Thanks for your contribution but fgetc seems unlikely to make for a satisfactory solution. We should first analyze what Rust does, probably.

axionbuster:
> Thanks for your contribution but fgetc seems unlikely to make for a satisfactory solution. We should first analyze what Rust does, probably.

The way it's done in Rust is similar to the original code: In the original code in the Lean library, a bulk of text input is first fetched in a buffer (done by `fgets`). The C library will cut off the buffer where it fits the buffer's capacity, or the last character in the buffer contains the NUL character or the line feed (`\n`) character. Many buffers are copied and then concatenated to assemble a contiguous string representing a line.

The difference between the Rust and C implementations is that the Rust library does not specially treat the NUL character. On the other hand, in C, the NUL character is used as a sentry to mark the end of the string; in other words, it believes that the stream input is NUL terminated. In addition, it does not otherwise indicate the number of bytes read.

To accomplish the goal of preserving NUL characters in the input string, I considered a couple of approaches:

1. Use `fread` and rewind the buffer manually using `fseek`: Ignore NUL characters (`fread`). Otherwise, if a line feed is detected, advance the buffer by the difference between the index of the earliest line feed character and the actual length of the buffer. Unfortunately, if the file is opened in "text mode," then `fseek` cannot be used (except with the unmodified values returned from a call to `ftell`). If the file is not on disk, then `fseek` is meaningless. Also, for portability on Windows, the file must be either opened in text mode (to convert carriage return-line feed to line feed) or the compound string `\r\n` must be searched for in the buffer, including in the case where `\r` and `\n` are split into disjoint buffers.
2. Use `fread` and rewind the buffer using `ungetc`: `ungetc` works for all C files (including devices and streams), but it may only be called at most once after a successful read or up to an implementation-defined limit (typically very short), whichever is larger.
3. Use `fgets`. If it returns a string with precisely one byte (it should be the NUL character, right?), append that single-character string and try again. I haven't given this approach much thought until now. I could test it.
4. Use `fgetc` many times. I was worried that this approach would be slow, but, to my surprise, it wasn't ... unbearable (on my local computer against the standard input stream). Maybe (?) I can assume that the stream is buffered by the C/C++ library (or maybe not). Though the PR check ran slower than the others, it could have been an hour-of-day effect when GitHub's computers were congested. I'll have to see if I can run any relevant benchmarks.
5. Implement a buffered I/O system. Probably a no-go.

Thank you for your feedback, and don't hesitate to tell me what you think.

Kha:
Thanks for the analysis. Since the original issue is about `readFile`, I believe the right course of action is to implement `readFile` independently of `getLine`, which it really should be anyway.

axionbuster:
I understand. The implementation of `readFile` (specifically, `readToEnd`) calls `getLine` repeatedly to construct a buffer for the file, which I thought should be reformed anyway.

axionbuster:
While I was reading the `IO.lean` source file, I realized I had several questions:

https://github.com/leanprover/lean4/blob/466ef74ccc024e334e9c78d9e008d3f7612f9ccf/src/Init/System/IO.lean#L422-L460

- Why are there different functions for text and binary modes? In the C library, the difference is included in the file handle itself, and users do not need to distinguish between using one type of function over another.
- Would it make sense to tie the ability to translate line endings to the operating system running the program in `lines`? Assuming the interface is independent of the C library, shouldn't the user be able to choose between the two approaches?

Right now, there seem to be text and binary mode functions solely because of line endings. Perhaps it would be more appropriate to use binary mode functions only, read the input, and then translate the line endings in bulk according to the platform. If the input buffer is clipped between `\r` and `\n` on Windows, one more byte must be read to determine it. Then, there's the issue: "What if there's an `\r` but no paired `\n`, whether due to EOF or not?"

Another issue is this: "Would there ever be a case where a stream could EOF but resume later?" If the answer is "yes," then, lest buffered I/O be implemented in Lean itself, either (1) the file must be opened in text mode in the C library or (2) any unpaired `\r` must be discarded, to avoid the tragic case of EOF happening right in between `\r` and `\n` and that case being untranslated. If the answer is "no" today, it could become "yes" tomorrow. Would this be a behavior change, or is it not relevant?

I am considering the following steps:

- Keep using my current implementation of `getLine` because the old implementation is "incorrect" (if preserving NUL characters is required), and I haven't found a better approach that doesn't require a more significant change. I also had trouble trying to run the benchmarks.
- Modify the bulk text input and output functions to use the binary file primitives underneath. If the platform is Windows, translate the line endings. Otherwise, do not process the buffer any further.
- While translating the line endings on Windows, uniformly discard `\r`.

Any feedback would be appreciated.

Kha:
We never use text mode: d85a4f5495a60a528befca9a2e77d461c2fbf918. We have two functions for reading raw binary data and for valid Unicode text.

axionbuster:
> We never use text mode: [d85a4f5](https://github.com/leanprover/lean4/commit/d85a4f5495a60a528befca9a2e77d461c2fbf918). We have two functions for reading raw binary data and for valid Unicode text.

I get it. (And sorry for the late reply.) Here, I've made some changes to my plan and drafted a UTF-8 validation function.

PR #3746, opened 2024-03-23
fix: make generalized field notation for abbreviation types handle optional parameters

### LINKED ISSUE #3745: Field notation for abbreviations fails if the argument is an optParam
### Description

If the receiver of generalized field notation is an optional parameter then it fails to resolve the field notation for type abbreviations.

This is because it's using `consumeMData` rather than `cleanupAnnotations` in `Lean.Elab.Term.typeMatchesBaseName`.

### Context

This is not an important bug, and I doubt it is affecting anyone, but it's easy to fix. Type abbreviations *are* used for overriding methods for field notation in mathlib, but not while using optional parameters.

### Steps to Reproduce

```lean
structure A

abbrev B := A

def A.x (_ : A) := 1
def B.x (_ : B) := 2

def A.y (_ : A) := 1
def B.y (_ : B := {}) := 2

#eval (show A from {}).x -- 1
#eval (show B from {}).x -- 2
#eval (show A from {}).y -- 1
#eval (show B from {}).y -- fails, should be 2
/-
invalid field notation, function 'B.y' does not have argument
with type (B ...) that can be used, it must be explicit
or implicit with a unique name
-/
```

### Versions

4.7.0-rc2
macOS 14.4

### Impact

kmill:
Closes #3745

### DIFF
--- a/src/Lean/Elab/App.lean
+++ b/src/Lean/Elab/App.lean
@@ -1149,7 +1149,7 @@ private partial def mkBaseProjections (baseStructName : Name) (structName : Name
 private def typeMatchesBaseName (type : Expr) (baseName : Name) : MetaM Bool := do
   if baseName == `Function then
     return (← whnfR type).isForall
-  else if type.consumeMData.isAppOf baseName then
+  else if type.cleanupAnnotations.isAppOf baseName then
     return true
   else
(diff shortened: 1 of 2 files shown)


nomeata (approved):
LGTM

PR #3747, opened 2024-03-23
feat: make `rcases` use the custom `Nat` eliminator

kmill:
As a special case, makes the `rcases` machinery use `Nat.casesAuxOn` so that goal states see `0` and `n + 1` rather than `Nat.zero` and `Nat.succ n`. This is a followup to enabling custom eliminators for `cases` and `induction`.

This doesn't use custom eliminators in general since `rcases` uses `Lean.MVarId.cases`, which is completely different from what `cases` and `induction` use.

### DIFF
--- a/src/Lean/Meta/Tactic/Cases.lean
+++ b/src/Lean/Meta/Tactic/Cases.lean
@@ -249,7 +249,7 @@ def cases (mvarId : MVarId) (majorFVarId : FVarId) (givenNames : Array AltVarNam
         else
           let s₁ ← generalizeIndices mvarId majorFVarId
           trace[Meta.Tactic.cases] "after generalizeIndices\n{MessageData.ofGoal s₁.mvarId}"
-          let s₂ ← inductionCasesOn s₁.mvarId s₁.fvarId givenNames ctx (useNatCasesAuxOn := useNatCasesAuxOn)
+          let s₂ ← inductionCasesOn s₁.mvarId s₁.fvarId givenNames ctx
           let s₂ ← elimAuxIndices s₁ s₂
           unifyCasesEqs s₁.numEqs s₂
   catch ex =>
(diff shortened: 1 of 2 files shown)


digama0:
This seems a bit specialized. I think it would be better to just use the default custom eliminator for the type in `MVarId.cases` (why is it "completely different from what `cases` does" in the first place?)

kmill:
@digama0 Good, you saw this, I was going to ping you to take a look.

This is part of the "nat induction beautification" project. Leo was OK with custom induction principles for Nat in particular, and for `induction`/`cases`, the more general solution went through because it was easy.

There are two issues with `Lean.MVarId.cases`:
1. It is not able to analyze a custom induction principle, like what's done in `Lean.Meta.getElimExprInfo`, so there's more work that would need to be done if it's not just special casing `Nat.casesOn` (as `Nat.casesAuxOn` is defeq, so there's no harm in swapping it out).
2. It turns out enabling this custom induction principle breaks the equation lemma generator, which is why it's behind a flag. It really seems to be meant for doing cases to get pristine constructors.

Probably what would be better is to modify `rcases` to use the `induction`/`cases` framework than to use `Lean.MVarId.cases`, but that's not packaged up in a useable form. In the short term, having a special case for `Nat` would finish nat induction beautification.

I'm not sure about the right implementation for `rcases`, but I created this PR with one possibility to start that conversation.

leodemoura (approved):

digama0:
> Probably what would be better is to modify `rcases` to use the `induction`/`cases` framework than to use `Lean.MVarId.cases`, but that's not packaged up in a useable form. In the short term, having a special case for `Nat` would finish nat induction beautification.

I think that `MVarId.cases` should "use the `induction`/`cases` framework" to the extent possible. But as a short term measure I agree with the changes in this PR (we should have some issue or TODO to avoid losing track of this though).

PR #3751, opened 2024-03-23
feat: hovering over omission term shows reason for omission

kmill:
This avoids printing the entire docstring for `⋯` when hovering over it, which is rather long, and instead it gives a brief reason for omission and what option to set to pretty print the omitted term.

### DIFF
--- a/src/Lean/PrettyPrinter/Delaborator/Basic.lean
+++ b/src/Lean/PrettyPrinter/Delaborator/Basic.lean
@@ -237,8 +237,8 @@ inductive OmissionReason
   | proof
 
 def OmissionReason.toString : OmissionReason → String
-  | deep => "This `⋯` denotes a term omitted due to its depth (see option `pp.deepTerms`)."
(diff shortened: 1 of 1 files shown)


src/Lean/PrettyPrinter/Delaborator/Basic.lean
```diff
@@ -232,12 +232,21 @@ where
...
   }
 
-def addOmissionInfo (pos : Pos) (stx : Syntax) (e : Expr) : DelabM Unit := do
+inductive OmissionReason
+  | deep
+  | proof
+
+def OmissionReason.toString : OmissionReason → String
+  | deep => "This `⋯` denotes a term omitted due to its depth (see option `pp.deepTerms`)."
+  | proof => "This `⋯` denotes an omitted proof (see option `pp.proofs`)."
```
mhuisi, line 241: Even more brief: "Term omitted due to its depth (see option `pp.deepTerms`)" 
I think it should be obvious from the symbol that `⋯` is some kind of omitted term.

mhuisi (approved):

PR #3753, opened 2024-03-24
fix: actually catch the error code from MoveFileEx

eric-wieser:
A user on Zulip [reported seeing an error code of "no error"](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/.60lake.20update.60.20broken.20on.20Windows.20.28.3F.29/near/429134334) here.

### DIFF
--- a/src/runtime/io.cpp
+++ b/src/runtime/io.cpp
@@ -789,19 +789,20 @@ extern "C" LEAN_EXPORT obj_res lean_io_rename(b_obj_arg from, b_obj_arg to, lean
     // with the unix-like OSs
     bool ok = MoveFileEx(string_cstr(from), string_cstr(to), MOVEFILE_REPLACE_EXISTING) != 0;
     if (!ok) {
-        errno = GetLastError();
+        // TODO: actually produce the right type of IO error
+        return io_result_mk_error((sstream()
+            << "failed to rename '" << string_cstr(from) << "' to '" << string_cstr(to) << "': " << GetLastError()).str());
     }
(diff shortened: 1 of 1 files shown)


src/runtime/io.cpp
```diff
@@ -788,6 +788,9 @@ extern "C" LEAN_EXPORT obj_res lean_io_rename(b_obj_arg from, b_obj_arg to, lean
     // so we have to call the underlying windows API directly to get behavior consistent
     // with the unix-like OSs
     bool ok = MoveFileEx(string_cstr(from), string_cstr(to), MOVEFILE_REPLACE_EXISTING) != 0;
+    if (!ok) {
+        errno = GetLastError();
```
eric-wieser, line 792: Actually, I think some conversion is probably needed here...

tydeu:
@eric-wieser  Thanks for finding this! Unfortunately, as you surmised, `GetLastErrror` does not directly map to `errno`, so some additional conversion is necessary to properly fix this error.

src/runtime/io.cpp
Kha: See other uses of `GetLastError`, we usually use the string overload of `io_result_mk_error` for lack of a conversion function

eric-wieser: In the longer run perhaps it's worth copying [`errmap` from CPython](https://github.com/python/cpython/blob/main/PC/errmap.h#L2)?

Kha (approved):

PR #3756, opened 2024-03-24
chore: upstream `Nat.binaryRec`

astrainfinita:
This was a PR for Std, but everything in `Std.Data.Nat.Bitwise` has been moved to Init.

`Nat.binaryRec`, `Nat.binaryRec'` and `Nat.binaryRecFromOne` replace `Nat.div2Induction`, which is noncomputable and more difficult to use. These definitions have been modified to be simpler and faster than the current version in Mathlib.

We also need `Nat.bit` because it is part of the type of `Nat.binaryRec` and we need some lemmas about it in this PR. If we do not introduce this definition, we'll have to use `cond b` everywhere. We cannot use `Bool.toNat` here because we have the import chain `Init.Data.Nat.Bitwise.Basic` -> `Init.Data.Fin.Basic` -> `Init.Data.Array.Basic` -> `Init.Data.Array.Basic` -> `Init.Data.Array.Subarray` -> `Init.NotationExtra` -> `Init.BinderPredicates` -> `Init.Data.Bool`.

---

Was https://github.com/leanprover/std4/pull/314.

[Zulip](https://leanprover.zulipchat.com/#narrow/stream/144837-PR-reviews/topic/std4.23314)

mathlib adaptations: https://github.com/leanprover-community/mathlib4/pull/12419

### DIFF
--- a/src/Init/Data/Bool.lean
+++ b/src/Init/Data/Bool.lean
@@ -342,7 +342,7 @@ theorem and_or_inj_left_iff :
 /-! ## toNat -/
 
 /-- convert a `Bool` to a `Nat`, `false -> 0`, `true -> 1` -/
-def Bool.toNat (b : Bool) : Nat := cond b 1 0
+def toNat (b:Bool) : Nat := cond b 1 0
 
 @[simp] theorem toNat_false : false.toNat = 0 := rfl
 


src/Init/Data/Nat/Bitwise/Basic.lean
```diff
@@ -71,13 +71,100 @@ theorem shiftRight_eq_div_pow (m : Nat) : ∀ n, m >>> n = m / 2 ^ n
...
+/-- `bodd n` returns `true` if `n` is odd-/
+def bodd (n : Nat) : Bool :=
+  (1 &&& n) != 0 -- `1 &&& n` is faster than `n &&& 1` for big `n`. This may change in the future.
+
+/-- `div2 n = ⌊n/2⌋` the greatest integer smaller than `n/2`-/
+def div2 (n : Nat) : Nat :=
+  n >>> 1
+
+/-- `bit b` appends the digit `b` to the binary representation of
+  its natural number input. -/
```
semorrison, line 83: Should be clearer that it is adding `b` at the "little end".

src/Init/Data/Nat/Bitwise/Basic.lean
```diff
@@ -71,13 +71,100 @@ theorem shiftRight_eq_div_pow (m : Nat) : ∀ n, m >>> n = m / 2 ^ n
     rw [shiftRight_add, shiftRight_eq_div_pow m k]
     simp [Nat.div_div_eq_div_mul, ← Nat.pow_succ, shiftRight_succ]
 
+/-- `bodd n` returns `true` if `n` is odd-/
+def bodd (n : Nat) : Bool :=
+  (1 &&& n) != 0 -- `1 &&& n` is faster than `n &&& 1` for big `n`. This may change in the future.
+
+/-- `div2 n = ⌊n/2⌋` the greatest integer smaller than `n/2`-/
+def div2 (n : Nat) : Nat :=
+  n >>> 1
```
semorrison, line 80: If it possible to just do without these two definitions?

semorrison:
Please include in the PR description an explanation of why this should be moved to Lean. (Not just a link to zulip, which doesn't really explain either.)

negiizhao:
awaiting-review

negiizhao:
!bench

src/Init/Data/Nat/Bitwise/Basic.lean
```diff
@@ -8,6 +8,9 @@ import Init.Data.Nat.Basic
 import Init.Data.Nat.Div
 import Init.Coe
 
+/-- convert a `Bool` to a `Nat`, `false => 0`, `true => 1` -/
+def Bool.toNat (b : Bool) : Nat := cond b 1 0
+
```
semorrison, line 13: Why have you moved this? This seems the wrong file.

semorrison:
> Please include in the PR description an explanation of why this should be moved to Lean. (Not just a link to zulip, which doesn't really explain either.)

I don't see any response to this request so far. Why do you want to move `Nat.bit` up to Lean?

negiizhao:
awaiting-review

src/Init/Data/Nat/Bitwise/Basic.lean
```diff
@@ -71,13 +71,92 @@ theorem shiftRight_eq_div_pow (m : Nat) : ∀ n, m >>> n = m / 2 ^ n
...
+  1 &&& (m >>> n) != 0
+
+theorem shiftRight_one (n) : n >>> 1 = n / 2 := rfl
+
+theorem mod_two_eq_zero_or_one (n : Nat) : n % 2 = 0 ∨ n % 2 = 1 :=
+  match n % 2, @Nat.mod_lt n 2 (by decide) with
+  | 0, _ => .inl rfl
+  | 1, _ => .inr rfl
+
+@[simp] theorem one_land_eq_mod_two (n : Nat) : 1 &&& n = n % 2 := by
```
semorrison, line 97: Should be `one_and_eq_mod_two` for consistency?

src/Init/Data/Nat/Bitwise/Lemmas.lean
```diff
@@ -88,51 +98,37 @@ theorem testBit_to_div_mod {x : Nat} : testBit x i = decide (x / 2^i % 2 = 1) :=
...
-      simp_all
+  induction x using binaryRecFromOne with
+  | z₀ => exact absurd rfl xnz
+  | z₁ => exact ⟨0, rfl⟩
+  | f b n n0 hyp =>
+    obtain ⟨i, h⟩ := hyp n0
+    refine ⟨i + 1, ?_⟩
+    rwa [testBit_succ, bit_div_two]
 
 theorem ne_implies_bit_diff {x y : Nat} (p : x ≠ y) : ∃ i, testBit x i ≠ testBit y i := by
```
semorrison, line 118: This could be renamed (with a deprecation) to `exists_testBit_ne_of_ne`.

src/Init/Data/Nat/Bitwise/Lemmas.lean
```diff
@@ -88,51 +98,37 @@ theorem testBit_to_div_mod {x : Nat} : testBit x i = decide (x / 2^i % 2 = 1) :=
...
 
+@[simp] theorem _root_.Bool.toNat_testBit_zero (b : Bool) : b.toNat.testBit 0 = b := by
+  cases b <;> rfl
+
 theorem toNat_testBit (x i : Nat) :
     (x.testBit i).toNat = x / 2 ^ i % 2 := by
   rw [Nat.testBit_to_div_mod]
   rcases Nat.mod_two_eq_zero_or_one (x / 2^i) <;> simp_all
 
 theorem ne_zero_implies_bit_true {x : Nat} (xnz : x ≠ 0) : ∃ i, testBit x i := by
```
semorrison, line 109: `exists_testBit_eq_true_of_ne_zero`? or just `exists_testBit`?

src/Init/Data/Nat/Bitwise/Lemmas.lean
```diff
@@ -146,25 +142,18 @@ theorem eq_of_testBit_eq {x y : Nat} (pred : ∀i, testBit x i = testBit y i) :
...
-        simpa using jp.right
+    | 0 =>
+      simp only [zero_le, true_and]
+      exact ne_zero_implies_bit_true (ne_of_gt (Nat.lt_of_succ_le p))
+    | n+1 =>
+      obtain ⟨i, h⟩ := hyp (mul_two_le_bit.mp p)
+      refine ⟨i + 1, ?_⟩
+      rwa [Nat.add_le_add_iff_right, testBit_succ, bit_div_two]
 
 theorem testBit_implies_ge {x : Nat} (p : testBit x i = true) : x ≥ 2^i := by
```
semorrison, line 158: Hmm, all these `implies` in the names are bad. They predate this PR of course. Maybe we should deal with them separately.

semorrison:
I'm still uncertain about whether it is a good idea to move `bit` up to Lean, just for the sake of removing `div2induction`.

Would you consider splitting this into a PR that does all the cleanup that you do here *without* adding `bit` and `Nat.binaryRec`, and then once we have a more minimal PR we can try to get some other opinions?

semorrison:
(This is not urgent at this point while we're still deciding what to do, but certainly for this PR to be merged we'll need a successful mathlib build, which I think hasn't occurred yet.)

negiizhao:
We had https://github.com/leanprover-community/mathlib4/pull/12419 2 weeks ago. Should I force-push to lean-pr-testing-3756?

src/Init/Data/Nat/Bitwise/Basic.lean
negiizhao: We have `Nat.land_bit` `Nat.land_comm` `Nat.land_assoc` `Nat.testBit_land` in Mathlib now.

semorrison:
> We had [leanprover-community/mathlib4#12419](https://github.com/leanprover-community/mathlib4/pull/12419) 2 weeks ago. Should I force-push to lean-pr-testing-3756?

Yes, you can keep pushing commits to lean-pr-testing-3756 to get it working.

src/Init/Data/Nat/Bitwise/Basic.lean
semorrison: Maybe those should all be changed too. :-) I'm more concerned about internal consistency in lean4 here.

negiizhao:
> Would you consider splitting this into a PR that does all the cleanup that you do here _without_ adding `bit` and `Nat.binaryRec`, and then once we have a more minimal PR we can try to get some other opinions?

I split the change of `Nat.testBit` to #4188, could you give some comments?

PR #3758, opened 2024-03-25
feat: add `set_equation_theorems` command to override equation theorems

kmill:
**Note:** This is just an exploration of how to support what mathlib's `@[eqns]` does, to help decide on next steps. Mario has pointed out that another solution would be to provide hooks for overriding what `simp`/`rw` uses for the equation lemmas for a definition.

This implementation does not do any consistency checks on the provided equation theorems.

(Some [Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/428973-nightly-testing/topic/.2311437.2C.20adaptations.20for.20.60nightly-2024-03-16.60/near/427286041))

### DIFF
--- a/src/Lean/Meta/Eqns.lean
+++ b/src/Lean/Meta/Eqns.lean
@@ -137,12 +137,27 @@ private partial def alreadyGenerated? (declName : Name) : MetaM (Option (Array N
     return none
 
 /--
+Equation theorems to use instead of the generated ones.
+-/
+builtin_initialize userEqnsExt : MapDeclarationExtension (Array Name) ← mkMapDeclarationExtension
+
+/--
+Adds the given equation theorems to the user equations extension.
+-/
+def setUserEqns (declName : Name) (eqns : Array Name) : CoreM Unit := do
+  registerEqnThms declName eqns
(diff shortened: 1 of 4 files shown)


src/Lean/Meta/Eqns.lean
```diff
@@ -136,6 +136,18 @@ private partial def alreadyGenerated? (declName : Name) : MetaM (Option (Array N
...
+/--
+Equation theorems to use instead of the generated ones.
+-/
+builtin_initialize userEqnsExt : MapDeclarationExtension (Array Name) ← mkMapDeclarationExtension
+
+/--
+Adds the given equation theorems to the user equations extension.
+-/
+def setUserEqns (declName : Name) (eqns : Array Name) : CoreM Unit := do
+  registerEqnThms declName eqns
```
kmill, line 148: Oops, there's a bug here. The point of calling this is to clear the cached equation theorems, but `registerEqnThms` does not expect being called on a declaration that has already had equation theorems registered. The issue is that `mapInv` is still going to say that the previous equation theorems are equation theorems, making `isEqnThm?` provide incorrect results.

PR #3759, opened 2024-03-25
chore: ToJson/FromJson Sum

kim-em:


nomeata (commented):
That's a dangerous instance, isn't it, if the constituents types have overlapping JSON representation?

Wouldn't be a tagged representation safer?

tydeu:
@nomeata There are cases were a union type like this is useful to accurately encode a type in a JSON specification. However, I do agree that using this for `Sum` may not be the best. Maybe adding this to a separate `def JUnion := Sum` with documentation highlighting that it does not round-trip would work instead.

digama0:
It should be possible to round-trip in the majority of cases, basically all of the cases where `JUnion` is safely usable in the first place, and certainly any case where you will see the type in JSON specs. (The bad case would be things like `JUnion Nat Nat`.)

PR #3762, opened 2024-03-25
fix: functional induction: preseve order of cases better

nomeata:
by passing an explicit array of metavariable around, instead of relying
on `getMVarsNoDelayed`, which may return them in unexpected order.

### DIFF
--- a/src/Lean/Meta/Match/MatcherApp/Transform.lean
+++ b/src/Lean/Meta/Match/MatcherApp/Transform.lean
@@ -309,8 +309,7 @@ def transform
         splitterNumParams in matchEqns.splitterAltNumParams,
         origAltType in origAltTypes,
         altType in altTypes do
-      let alt' ←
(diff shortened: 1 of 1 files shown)


src/Lean/Meta/Match/MatcherApp/Transform.lean
```diff
@@ -290,7 +309,8 @@ def transform (matcherApp : MatcherApp)
         splitterNumParams in matchEqns.splitterAltNumParams,
         origAltType in origAltTypes,
         altType in altTypes do
-      let alt' ← Match.forallAltTelescope origAltType (numParams - numDiscrEqs) 0 fun ys _eqs args _mask _bodyType => do
+      let alt' ←
+        forallAltTelescope' origAltType (numParams - numDiscrEqs) 0 fun ys args => do
```
nomeata, line 313:
```suggestion
      let alt' ← forallAltTelescope' origAltType (numParams - numDiscrEqs) 0 fun ys args => do
```

PR #3763, opened 2024-03-25
feat: in conv tactic, use `try with_reducibe rfl`

nomeata:
The `conv` tactic tries to close “trivial” goals after itself. As of now, it uses
`try rfl`, which means it can close goals that are only trivial after reducing with
default transparency. This is suboptimal

 * this can require a fair amount of unfolding, and possibly slow down the proof
   a lot. And the user cannot even prevent it.
 * it does not match what `rw` does, and a user might expect the two to behave the
   same.

So this PR changes it to `with_reducible rfl`, matching `rw`’s behavior.

I considered `with_reducible eq_refl` to only solve trivial goals that involve equality,
but not other relations (e.g. `Perm xs xs`), but a discussion on mathlib pointed out
that it’s expected and desirable to solve more general reflexive goals:

https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Closing.20after.20.60rw.60.2C.20.60conv.60.3A.20.60eq_refl.60.20instead.20of.20.60rfl.60/near/429851605

### DIFF
--- a/src/Init/Omega/Int.lean
+++ b/src/Init/Omega/Int.lean
@@ -137,11 +137,13 @@ theorem add_le_iff_le_sub (a b c : Int) : a + b ≤ c ↔ a ≤ c - b := by
     lhs
     rw [← Int.add_zero c, ← Int.sub_self (-b), Int.sub_eq_add_neg, ← Int.add_assoc, Int.neg_neg,
       Int.add_le_add_iff_right]
+  try rfl -- stage0 update TODO: Change this to rfl or remove
 
 theorem le_add_iff_sub_le (a b c : Int) : a ≤ b + c ↔ a - c ≤ b := by
   conv =>
     lhs
     rw [← Int.neg_neg c, ← Int.sub_eq_add_neg, ← add_le_iff_le_sub]
+  try rfl -- stage0 update TODO: Change this to rfl or remove
 
 theorem add_le_zero_iff_le_neg (a b : Int) : a + b ≤ 0 ↔ a ≤ - b := by
   rw [add_le_iff_le_sub, Int.zero_sub]


nomeata:
We certainly have test cases that fail, like
```
example : id (fun x => 0 + x) = id := by
  conv =>
    lhs
    arg 1
    ext y
    rw [Nat.zero_add]
```
failing with
```
unsolved goals
⊢ (id fun y => y) = id
```
Not unreasonable – this is only true when unfolding `id`.

Or another case
```
theorem len_2 (a b : α) (bs : List α) : len (a::b::bs) = 1 + len (b::bs) := by
  conv => lhs; unfold len
```


I expect that mathlib will be have many more of these, where at last I don’t dare to make a judgement call if that was an intentional use of defeq, or an accidential one.

digama0:
does `rfl` still work to close the goal in the example? If so then I think this is an okay mitigation for any issues that arise in mathlib.

nomeata:
Maybe I should first try to switch to `eq_refl`, unless we expect conv to handle non-Eq-goals here?

nomeata:
> does `rfl` at the end of the `conv` block still work to close the goal in the example? If so then I think this is an okay mitigation for any issues that arise in mathlib.

It helps to put `rfl` (or something like `exact Iff.rfl`) after the `conv` block.

I’m on the fence. It’s convenient not having to write `rfl` or `exact Iff.rfl` afterwards. OTOH, why is `conv` special-casing `rfl` as a “convenience closing tactic”? Why not “`trivial`” or something else? It seems a bit undisciplined.

It’s a bit different for `conv_lhs` – there the goal is a relation, and after converting the LHS one might expect a reflexive goal to be solved. But for the general `conv` tactic there is nothing special about goals that are relation.

@digama0, what do you think, is less magic in `conv` good or bad here?

PR #3768, opened 2024-03-26
feat: `#print` command shows structure fields

### LINKED ISSUE #3644: RFC: have `#print` display structure fields
### Proposal

The `#print` command currently treats structures just like any other inductive type. For example:
```lean
structure A (α β) where 
  a : α
  b : β 

/-
inductive A.{u_1, u_2} : Sort u_1 → Sort u_2 → Sort (max (max 1 u_1) u_2)
number of parameters: 2
constructors:
A.mk : {α : Sort u_1} → {β : Sort u_2} → α → β → A α β
-/
#print A
```
Unfortunately, this omits useful information about the structure type `A`, in particular the name and types of the fields. Indeed, there is no easy way to find the names the fields of a structure at this time.

I propose changing that default behavior as follows:
```lean
/-
structure A.{u_1, u_2} : Sort u_1 → Sort u_2 → Sort (max (max 1 u_1) u_2)
number of parameters: 2
constructor:
A.mk : {α : Sort u_1} → {β : Sort u_2} → α → β → A α β
fields:
a : α
b : β
-/
#print A
```
I propose the following spec for the info:

* The output of the `#print` command for structures should extend that for inductive except that the word "inductive" should be replaced by "structure" or "class" and the word "constructors" should be replaced by "constructor".
* There should be an additional line "fields:" followed, on separate lines, by the names of each structure field followed by the type of the associated projection, omitting the type parameters and the "self" parameter.
* Private fields should be prefixed by the "private" keyword.
* The fields should be listed in the order of their indices so that numerical projections can be identified as well.

The rationale is ensure that no information is lost from the current behavior and to keep the info simple. Considerations were given for adding an option for printing flattened subobject fields and other display features. However, these were rejected on the basis that these would not fit with the current utilitarian design of the `#print` command in Lean 4. That information could be obtained by multiple uses of the `#print` command. Additional features such as this should be tested in Std or Mathlib before adding to Lean 4.

The WIP implementation in #3768  shows that the PR footprint would be limited to Lean.Elab.Print, with no additional infrastructure. This supports that the maintenance cost should be minimal compared to user benefits.

### Context

The `#print` command in Lean 3 did print structure fields and their types.

The lack of support for structures in the `#print` command has been mentioned several times on Zulip. Here is a recent example: <https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/improving.20.23print.20structure/near/422209781>

This RFC comes from WIP [std4#672](https://github.com/leanprover/std4/pull/672), which is currently on hold pending the outcome of this RFC. Note that this PR implements features which are _not_ part of this RFC.

### Impact

### Edits

Mar 28: Revised spec after @kmill comments. The spec now omits type parameters and "self" parameter from the field types.

fgdorais:
See RFC #3644 for a discussion of design choices.

Closes #3644


src/Lean/Elab/Print.lean
```diff
@@ -65,8 +65,36 @@ private def printInduct (id : Name) (levelParams : List Name) (numParams : Nat)
     m := m ++ Format.line ++ ctor ++ " : " ++ cinfo.type
   logInfo m
 
+open Meta in
+private def printStructOrClass (id : Name) (levelParams : List Name) (numParams : Nat) (type : Expr)
```
semorrison, line 69: Perhaps just `printStructure`? (No need to abbreviate `Structure`, unless it's being done consistently elsewhere.)

tests/lean/printStruct.lean
```diff
@@ -0,0 +1,28 @@
+/-!
+  Test #print command for structures and classes
+-/
+
+/- Structure -/
+#print Prod
```
semorrison, line 6: Could we update this test file to use `#guard_msgs`? It is the future. :-)

fgdorais: Indeed it is! I'm very happy to update the tests!

fgdorais:
awaiting-review

PR #3771, opened 2024-03-26
feat: failing macros to show error from first registered rule

### LINKED ISSUE #3770: RFC: macros should show last error, not first
### Proposal

Consider this:
```
macro "foo" : tactic => `(tactic|fail "first")
macro_rules | `(tactic|foo) => `(tactic|exact 1)
macro_rules | `(tactic|foo) => `(tactic|fail "middle")
macro_rules | `(tactic|foo) => `(tactic|exact 2)
macro_rules | `(tactic|foo) => `(tactic|fail "last")

def what_is_foo : Nat := by foo

/-- info: 2 -/
#guard_msgs in
#eval what_is_foo


macro "bar" : tactic => `(tactic|fail "first")
macro_rules | `(tactic|bar) => `(tactic|fail "middle")
macro_rules | `(tactic|bar) => `(tactic|fail "last")

/--
error: last
⊢ Nat
-/
#guard_msgs in
def what_is_bar : Nat := by bar
```

We see that the `macro` machinery tries the later registered macro first, and `what_is_foo` becomes `2`. This is good: It means that one can (partially) override macros later in the development.

But we see that the error message we get if all macros fail is from the macro that was tried first (register last). I propose that we change this to reporting the failure from the one that was tried last (registered first):

One main purpose of overriding tactic macros is to make an extensible tactic (e.g. `rfl`, `decreasing_trivial`, etc.) This typically starts with a general purpose tactics, and then special ones are added later. The current logic means that the users sees the error message from some arbitrary specialized tactic, which is often puzzling. The maybe best and most pressing example for this is #3302:
```
example (P : Prop) : P := by rfl
```
gives (depending on lean version and imports)
```
type mismatch
  HEq.rfl
has type
  HEq ?m.10 ?m.10 : Prop
but is expected to have type
  P : Prop
```

Alone the fact that the error message may depend on the input order (if independent modules both extend the tactic) shows that this is not great behavior.

If instead the error message were always from the first registered / last tried tactic, then it would be stable under extension. Moreover it would allow the idiom of always putting a `fail` tactic first with a good user-facing error message, like
```
macro "reffl" : tactic => `(tactic|fail "reffl failed. This means that the goal either is not a reflexive relation, or the arguments of the relation are not equal. Try using the reflexivitiy lemma for your relation explicitly, e.g. `exact Eq.rfl` or `exact Iff.rfl`. ")
macro_rules | `(tactic|reffl) => `(tactic|exact Eq.refl _)
macro_rules | `(tactic|reffl) => `(tactic|exact Iff.refl _)

example : true = false := by reffl
```

I also expect that the implementation is actually simpler if the last tried macro’s output is reported, rather than preserving the first tired macro’s output.

(NB: The idea of tactics and commands offering more insight would help here as well: A macro could offer a verbose mode that shows the tactics that were tried in turn, and each invocation's messages; this can be useful to better understand why a macro failed, or when it succeeded why it used the rule it did.)

### Impact

nomeata:
fixes #3770

Also start `rfl` with a `fail` message that is hopefully more helpful
than what we get now (see updated test output). This would be a cheaper
way to address #3302 without changing the implementation of rfl (as
tried in #3714).

### DIFF
--- a/src/Init/Tactics.lean
+++ b/src/Init/Tactics.lean
@@ -368,7 +368,7 @@ for new reflexive relations.
 Remark: `rfl` is an extensible tactic. We later add `macro_rules` to try different
 reflexivity theorems (e.g., `Iff.rfl`).
 -/
-macro "rfl" : tactic => `(tactic| fail "The rfl tactic failed. Possible reasions:
+macro "rfl" : tactic => `(tactic| fail "The rfl tactic failed. Possible reasons:
 - The goal is not a reflexive relation (neither `=` nor a relation with a @[refl] lemma).
 - The arguments of the relation are not equal.
 Try using the reflexivitiy lemma for your relation explicitly, e.g. `exact Eq.rfl`.")


src/Init/Tactics.lean
```diff
@@ -365,8 +368,12 @@ for new reflexive relations.
 Remark: `rfl` is an extensible tactic. We later add `macro_rules` to try different
 reflexivity theorems (e.g., `Iff.rfl`).
 -/
-macro "rfl" : tactic => `(tactic| eq_refl)
+macro "rfl" : tactic => `(tactic| fail "The rfl tactic failed. Possible reasions:
+- The goal is not a reflexive relation (neither `=` nor a relation with a @[refl] lemma)
```
nomeata, line 372:
```suggestion
- The goal is not a reflexive relation (neither `=` nor a relation with a @[refl] lemma).
```

src/Init/Tactics.lean
```diff
@@ -365,8 +368,12 @@ for new reflexive relations.
 Remark: `rfl` is an extensible tactic. We later add `macro_rules` to try different
 reflexivity theorems (e.g., `Iff.rfl`).
 -/
-macro "rfl" : tactic => `(tactic| eq_refl)
+macro "rfl" : tactic => `(tactic| fail "The rfl tactic failed. Possible reasions:
```
nomeata, line 371:
```suggestion
macro "rfl" : tactic => `(tactic| fail "The rfl tactic failed. Possible reasons:
```

PR #3772, opened 2024-03-26
refactor: rfl tactic: do not use Kernel.isDefEq

nomeata:
Sebastian mentioned that the use of the kernel defeq was to work around
a performance issue that was fixed since. Let's see if we can do
without.

This is also a semantic change: Ground terms (no free vars, no mvars) are reduced at 
“all” transparency even if the the transparency setting is default. This was the case
even before 03f6b87647ae79cdca92c5de87a84603eba78b06 switched to the kernel defeq
checking for performance. It seems that this is rather surprising behavior from the user
point of view. The fallout on batteries and mathlib is rather limited, only a few
`rfl` proofs seem to have (inadvertently or not) have relied on this.

The speedcenter reports no significant regressions on core or mathlib.

### DIFF
--- a/tests/lean/run/4644.lean
+++ b/tests/lean/run/4644.lean
@@ -10,9 +10,17 @@ termination_by a.size - i
 def check_sorted [x: LE α] [DecidableRel x.le] (a: Array α): Bool :=
   sorted_from_var a 0
 
--- works (because `rfl` of closed terms resorts to kernel defeq, see #3772)
-example: check_sorted #[0, 3, 3, 5, 8, 10, 10, 10] := by
-  rfl
+/--
+error: The rfl tactic failed. Possible reasons:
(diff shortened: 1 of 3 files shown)


nomeata:
No significant perf changes on mathlib: http://speed.lean-fro.org/mathlib4/compare/60f80410-4f54-47fd-b9f1-a653d7f9d920/to/3ce2b676-acda-428f-9af4-480e2f8e7649

nomeata:
Good 🟢 . Will update the adaption branches to check for new surprises, and then merge.

nomeata:
Ok, double checked the speed center that there are no significant regressions:
http://speed.lean-fro.org/mathlib4/compare/8c347d50-8923-404f-9e41-daa5fd4ab2f7/to/dfab2686-177b-44a1-afc4-b23a55bcce45

PR #3773, opened 2024-03-26
chore: add refl attribute to Eq.refl

kim-em:
Without this attribute, `rw?` doesn't work.

### DIFF
--- a/src/Init/Prelude.lean
+++ b/src/Init/Prelude.lean
@@ -278,6 +278,8 @@ inductive Eq : α → α → Prop where
   equality type. See also `rfl`, which is usually used instead. -/
   | refl (a : α) : Eq a a
 
+attribute [refl] Eq.refl
+
 /-- Non-dependent recursor for the equality type. -/
 @[simp] abbrev Eq.ndrec.{u1, u2} {α : Sort u2} {a : α} {motive : α → Sort u1} (m : motive a) {b : α} (h : Eq a b) : motive b :=
   h.rec m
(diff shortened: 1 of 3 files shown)


nomeata:
Now `rfl` will try to solve it with equality twice, I fear. Not horrible, but kinda unsatisfying… #3714  could help, not sure though if I’ll follow through with that.

nomeata:
Is it this line where we need this?

https://github.com/leanprover/lean4/blob/e8a2786d6d4defa29f8991eb03e1dd975311e5ae/src/Lean/Elab/Tactic/Rewrites.lean#L66

But `rfl` should do equality as well… so probably somewhere else?

Ah, maybe it’s this line:

https://github.com/leanprover/lean4/blob/e8a2786d6d4defa29f8991eb03e1dd975311e5ae/src/Lean/Meta/Tactic/Rewrites.lean#L143

Could this just use the `rfl` tactic as well, using `evalTactic`, as a stop-gap measure?

Or else run [Lean.MVarId.refl](https://leanprover-community.github.io/mathlib4_docs/Lean/Meta/Tactic/Refl.html#Lean.MVarId.refl) first and `catch` if it fails to run `applyRfl` second, at least until #3714 is fixed.

joehendrix:
@nomeata That second place looks like the right place.

It strikes me as a little strange to have two extension mechanisms for extending `rfl`.  Should we try to consolidate?

With the current system, I think the right approach is to use the elaborator `evalTactic (← `(tactic| try rfl))` approach so that both kinds of user extension mechanisms have a chance to trigger.

nomeata:
My attempt at consolidating is in https://github.com/leanprover/lean4/pull/3714, and I plan to return to that.


Until then, evalTactic is certainly a reasonable choice.

joehendrix:
I looked into this, and `evalTactic (← ``(tactic| rfl))` would change the current behavior of `rw?` to use `rfl` in a context with `default` transparency rather than `reducible`.  Given there's a test case for that, it seems like that's not desired.

I think using the attribute here is reasonable while we have two mechanisms like this.  `applyRfl` looks up which refl lemma to use in a discriminator tree, so `rfl` will only check twice the term is a `Eq`.

nomeata:
Looks like #3783 subsumes this?

PR #3792, opened 2024-03-27
fix: handle dependent fields when `deriving BEq`

### LINKED ISSUE #3740: BEq fails on structures with dependent types.
### Prerequisites

### Description
Deriving a BEq instance like so:
```lean
structure BVBit where
  {w : Nat}
  idx : Fin w
  deriving BEq
```
leads to an error:
```
type mismatch
  b✝
has type
  Fin b✝¹ : Type
but is expected to have type
  Fin a✝¹ : Type
```
Presumably because the BEq derive handler does not account for dependent types
in this fashion.

### Steps to Reproduce
Run the above code

**Expected behavior:** we get a BEq instance

**Actual behavior:** we get a (rather undescriptive) error

### Versions
"4.7.0-rc2"

### Impact

arthur-adjedj:
Closes #3740

If a field `x : xType` appears in the type of another field `y : f xType`, it's generally not possible to derive a faithful `BEq` instance because the two `y` fields won't necessarily have definitionally equal types. However, such a problem can be avoided whenever the BEq on `xType` is Lawful. This PR uses `eq_of_beq` whenever `xType` appears in another fvar's types, and unifies the dependent types as needed.

### DIFF
--- a/src/Init/Data/Nat/Linear.lean
+++ b/src/Init/Data/Nat/Linear.lean
@@ -173,13 +173,13 @@ instance : LawfulBEq PolyCnstr where
   eq_of_beq {a b} h := by
     cases a; rename_i eq₁ lhs₁ rhs₁
     cases b; rename_i eq₂ lhs₂ rhs₂
-    have h : eq₁ == eq₂ && lhs₁ == lhs₂ && rhs₁ == rhs₂ := h
+    have h : eq₁ == eq₂ && (lhs₁ == lhs₂ && rhs₁ == rhs₂) := h
     simp at h
-    have ⟨⟨h₁, h₂⟩, h₃⟩ := h
+    have ⟨h₁, h₂, h₃⟩ := h
     rw [h₁, h₂, h₃]
   rfl {a} := by
     cases a; rename_i eq lhs rhs
-    show (eq == eq && lhs == lhs && rhs == rhs) = true
+    show (eq == eq && (lhs == lhs && rhs == rhs)) = true
     simp [LawfulBEq.rfl]
 
 def PolyCnstr.mul (k : Nat) (c : PolyCnstr) : PolyCnstr :=


timotree3:
One downside of this derive handler is that the code it produces using `Eq.rec` sometimes can't be reduced by the kernel, while a hand-written implementation might use easier to reduce type-specific operations like `Fin.cast`. For this to be a problem, it requires that `(x == y) = true` is a defeq but `x = y` is not, otherwise  `Eq.rec` successfully reduces. A likely way this could happen is with quotients:
```lean
def ZMod5 : Type := Quot (· % 5 = · % 5)

abbrev ZMod5.val : ZMod5 → Nat :=
  Quot.lift (· % 5) sorry

instance : BEq ZMod5 where
  beq x y := x.val == y.val

instance : LawfulBEq ZMod5 := sorry

structure MyStruct where
  x : ZMod5
  y : Fin x.val

-- This is approximately what would be generated by `deriving BEq` after this PR
def instBEqGenerated : BEq MyStruct where
  beq x y :=
    if h : x.1 == y.1 then
      let x2' : Fin y.1.val := (eq_of_beq h) ▸ x.2
      x2' == y.2
    else
      false

def instBEqPreferred : BEq MyStruct where
  beq x y :=
    if h : x.1 == y.1 then
      let x2' : Fin y.1.val := Fin.cast (eq_of_beq h) x.2
      x2' == y.2
    else
      false

#reduce instBEqPreferred.beq ⟨Quot.mk _ 6, 0⟩ ⟨Quot.mk _ 11, 0⟩ -- true
#reduce instBEqGenerated.beq ⟨Quot.mk _ 6, 0⟩ ⟨Quot.mk _ 11, 0⟩ -- _ (Nat.rec _ _ (⋯ ▸ 0))
```

That said, it doesn't seem feasible in the short term to generate code using functions like `Fin.cast`, so I don't think this concern should block the PR. In the long term, this could be solved with a typeclass `EqFunctor F where mapEq : x = y → F x → F y` .

semorrison:
LGTM. Could you mark this as ready for review if you're happy to proceed with this?

arthur-adjedj:
As it turns out, some parts of the code rely on the reduction behaviour of the derived instances, and these issues only arise when compiling on stage2, which I didn't try locally until now. Most of these are easy to correct, but I would rather not make this fix a breaking change, and I don't have much time to allocate to fixing this PR right now. I'll mark this as WIP and ping you back once I get this in a better state. Sorry for the trouble.

arthur-adjedj:
WIP

arthur-adjedj:
@semorrison 
I have managed to get the same reduction behavior for derived instances as was previously present. This implied reordering the appearance of the construction of `rhs`, and removing the initial `true` value where it wasn't necessary. The only difference left between this and the previous implementation is that the chaining of `&&` operations is now done from left to right instead of right to left (see [src/Init/Data/Nat/Linear.lean](https://github.com/leanprover/lean4/pull/3792/files#diff-62085398b7f9a0be4685bea9ea6211c25e3a3ba3e25ebb2c7a55ca1df87668f9), where `eq₁ == eq₂ && lhs₁ == lhs₂ && rhs₁ == rhs₂` becomes `eq₁ == eq₂ && (lhs₁ == lhs₂ && rhs₁ == rhs₂)`). This only breaks one place of the codebase, and Batteries/Mathlib seems to compile correctly. Lastly, this update requires an update to stage0, which I believe implies some complications ?

semorrison:
Oops, there are stage0 changes since `nightly-with-mathlib`. Repeating, but rebasing onto `master`.

semorrison:
Could you confirm that I correctly removed the `WIP` label?

arthur-adjedj:
Ah yes, not putting this as `awaiting-review` was an oversight on my end.

PR #3793, opened 2024-03-27
feat: `lake lean`

tydeu:
Adds a `lake lean` CLI command that builds the imports of a Lean file and then elaborates it via `lean` with the workspace's configuration (i.e., adding the `moreLeanArgs` / `leanOptions` of the root `package` and Lake's environment).


tydeu:
This can address second-half of #3793 (elaborating Lean files outside the package).

PR #3795, opened 2024-03-28
fix: lake: check normalized packages directory path before rename

### LINKED ISSUE #3502: Lake update error: could not rename package on Windows
### Description

Due to file path separator difference on windows, lake update comment cannot rename or access .lake/packages correctly, raising "could not rename packages directory (.\.lake/packages -> .\.lake\packages): no error (error code: 0)"

### Context

Error when trying to obtain new package (doc-gen4)

### Steps to Reproduce

"lake update" on windows

### Versions

[Lean 4.4.0]
[Windows 11]

tydeu:
Normalize the relative packages directory paths in the pre-rename check to avoid renames if the difference in paths is only in the path separators. Also adds a log message on rename.

Closes #3502.


tydeu:
This builds on #2546 to fix a Windows error in Lake that has occurring often [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/error.3A.20***.3A.20could.20not.20rename.20packages.20directory/near/429855907) lately.

PR #3798, opened 2024-03-28
feat: add `pp.mvars` and `pp.mvars.withType`

### LINKED ISSUE #3781: RFC: pp.showNumericNames
### Proposal

I wonder if it would be useful to have a `pp.showNumericNames` option, `true` by default, that if off prints `?m.1235` as simply `?m` or maybe `?m._`.

The concrete use case is that I tried to include this in the test suite:
```
/--
error: invalid field notation, type is not of the form (C ...) where C is a constant
  takeWhile
has type
  (?m.165402 → Bool) → Array ?m.165402 → Array ?m.165402
-/
#guard_msgs in
#check Errors.takeWhile.induct
```
and the error message would quickly go out of sync.

(Of course it’s not a great error message, and I wonder whether lean couldn’t count metavariables afresh for each command, but that’s not the point).

I imagine if I could add
```
set_option pp.showNumericNames false in
```
here then this would help.

(The option name isn't particularly carefully chosen yet.)

### Impact

kmill:
* Setting `pp.mvars` to false causes metavariables to pretty print as `?_`.
* Setting `pp.mvars.withType` to true causes metavariables to pretty print with type ascriptions.

Motivation: when making tests, it is inconvenient using `#guard_msgs` when there are metavariables, since the unique numbering is subject to change.

This feature does not use `⋯` omissions since a metavariable is already in a sense an omitted term. If repeated metavariables do not appear in an expression, there is a chance that a term pretty printed with `pp.mvars` set to false can still elaborate to the correct term, unlike for other omissions.

(In the future we could consider an option that pretty prints uniquely numbered metavariables as `?m✝`, `?m✝¹`, `?m✝²`, etc. to be able to tell them apart, at least in the same pretty printed expression. It would take care to make sure that these names are stable across different hovers.)

Closes #3781

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -69,6 +69,12 @@ v4.8.0 (development in progress)
   to enable pretty printing function applications using generalized field notation (defaults to true).
   Field notation can be disabled on a function-by-function basis using the `@[pp_nodot]` attribute.
 
+* Added options `pp.mvars` (default: true) and `pp.mvars.withType` (default: false).
+  When `pp.mvars` is false, metavariables pretty print as `?_`,
+  and when `pp.mvars.withType` is true, metavariables pretty print with a type ascription.
+  These can be set when using `#guard_msgs` to make tests not rely on the unique ids assigned to anonymous metavariables.
+  [#3798](https://github.com/leanprover/lean4/pull/3798).
+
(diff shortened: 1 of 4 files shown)


nomeata (approved):
LGTM

PR #3806, opened 2024-03-29
fix: typo in `withSetOptionIn`

thorimur:
When using `withSetOptionIn` on syntax `set_option ... in <command>`, recurse into command syntax (`stx[2]`) instead of the syntax `in` (`stx[1]`).

---

Demonstration of `stx[1]` vs. `stx[2]`:
```lean
import Lean

def stx := (Lean.Unhygienic.run `(set_option trace.debug true in #print foo)).raw

#eval stx[1] -- Lean.Syntax.atom (Lean.SourceInfo.none) "in"
#eval stx[2] -- `#print` command syntax
```

### DIFF
--- a/tests/lean/withSetOptionIn.lean
+++ b/tests/lean/withSetOptionIn.lean
@@ -17,15 +17,23 @@ elab "#trace_debug_foo" : command => do trace[debug] "foo"
 /-- Elab `cmd` using `withSetOptionIn`. -/
 elab "#test " cmd:command : command => withSetOptionIn elabCommand cmd
 
-/-! ## Controls -/
+/-! ## Controls
 
-#trace_debug_foo -- no trace
+Ensure that `#trace_debug_foo` works as expected.
+-/
 
-set_option trace.debug true in #trace_debug_foo -- trace
+#guard_msgs in
+#trace_debug_foo
+
+/-- info: [debug] foo -/
+#guard_msgs in
+set_option trace.debug true in #trace_debug_foo
 
 /-! ## Test
 
 Should trace `[debug] foo`, and not log the error "unexpected command 'in'".
 -/
 
+/-- info: [debug] foo -/
+#guard_msgs in
 #test set_option trace.debug true in #trace_debug_foo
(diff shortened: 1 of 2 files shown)


digama0:
This should have a test, you can either create a file saying exactly what you wrote in the PR or add that to an existing (relevant) test.

thorimur:
awaiting-review

tests/lean/withSetOptionIn.lean
```diff
@@ -0,0 +1,32 @@
+import Lean
+
+/-!
+# `withSetOptionIn`
+
+This test checks that `withSetOptionIn` recurses into the command syntax (`stx[2]`) in
+`set_option ... in <cmd>`.
+
+Prior to #3806, `withSetOptionIn` erroneously recursed into the syntax `in` (`stx[1]`).
```
adomani, line 9: Since the PR description already mentions this, maybe this line can be removed?

tests/lean/withSetOptionIn.lean
```diff
@@ -0,0 +1,32 @@
...
+-/
+
+open Lean Elab Command
+
+/-- Trace `foo` when `set_option trace.debug true`. -/
+elab "#trace_debug_foo" : command => do trace[debug] "foo"
+
+/-- Elab `cmd` using `withSetOptionIn`. -/
+elab "#test " cmd:command : command => do
```
adomani, line 18:
```suggestion
elab "#test " cmd:command : command =>
```
I think that the `do` is superfluous here.

adomani (approved):
Thanks for digging into this!

tests/lean/withSetOptionIn.lean
thorimur: I'd like the info to be immediately available to anyone reading the test without having to go to github! :)

tests/lean/withSetOptionIn.lean
thorimur: Ah right, nice! :) I also moved the body onto the same line for brevity.

PR #3808, opened 2024-03-30
fix: replace unary Nat.succ simp rules with simprocs

joehendrix:
This removes simp attributes from `Nat.succ.injEq` and `Nat.succ_sub_succ_eq_sub` to replace them with simprocs.  This is because any reductions involving `Nat.succ` has a high risk of leading proof performance problems when dealing with even moderately large numbers.

Here are a couple examples that will both report a maximum recursion depth error currently.  These examples are fixed by this PR.

```
example : (123456: Nat) = 12345667 := by
  simp

example (x : Nat) (p : x = 0) : 1000 - (x + 1000) = 0 := by
  simp
```

### DIFF
--- a/src/Init/Data/Nat/Basic.lean
+++ b/src/Init/Data/Nat/Basic.lean
@@ -574,8 +574,7 @@ theorem eq_zero_or_eq_succ_pred : ∀ n, n = 0 ∨ n = succ (pred n)
   | 0 => .inl rfl
   | _+1 => .inr rfl
 
-@[deprecated Nat.succ.injEq]
-theorem succ_inj' : succ a = succ b ↔ a = b := ⟨succ.inj, congrArg _⟩
+theorem succ_inj' : succ a = succ b ↔ a = b := (Nat.succ.injEq a b).to_iff
 
 theorem succ_le_succ_iff : succ a ≤ succ b ↔ a ≤ b := ⟨le_of_succ_le_succ, succ_le_succ⟩
 


digama0:
This seems like it would be better addressed by finding a way for that expression not to count as having the form `succ x - succ y`, either in this particular simp lemma or in the discrimination tree in general. Disabling it may be the best short term option, but this seems like a regression.

joehendrix:
I'll take a stab at a simproc that covers this case, but I don't think this should be a reason to touch discriminator tree patterns (e.g., the logic in `isNatOffset`).  I expect that would have big downstream effects.

src/Init/Data/Nat/Basic.lean
```diff
@@ -248,11 +248,12 @@ theorem lt_succ_of_le {n m : Nat} : n ≤ m → n < succ m := succ_le_succ
 
 @[simp] protected theorem sub_zero (n : Nat) : n - 0 = n := rfl
 
-@[simp] theorem succ_sub_succ_eq_sub (n m : Nat) : succ n - succ m = n - m := by
+theorem succ_sub_succ_eq_sub (n m : Nat) : succ n - succ m = n - m := by
   induction m with
   | zero      => exact rfl
   | succ m ih => apply congrArg pred ih
 
+
```
joehendrix, line 256:
```suggestion
```

src/Init/Data/Nat/Bitwise/Lemmas.lean
```diff
@@ -271,15 +272,16 @@ theorem testBit_two_pow_sub_succ (h₂ : x < 2 ^ n) (i : Nat) :
   induction i generalizing n x with
   | zero =>
     match n with
-    | 0 => simp
+    | 0 =>
+      simp [succ_sub_succ_eq_sub]
```
joehendrix, line 276:
```suggestion
    | 0 => simp [succ_sub_succ_eq_sub]
```

eric-wieser:
> * Mathlib branch [lean-pr-testing-3808](https://github.com/leanprover-community/mathlib4/compare/nightly-testing...lean-pr-testing-3808) built against this PR, but linting failed. (2024-04-04 21:47:46) [View Log](https://github.com/leanprover-community/mathlib4/actions/runs/8561215659)

@semorrison, this is misleading; the CI job "built", but the "build" step of the CI job failed.

src/Lean/Meta/Tactic/Simp/BuiltinSimprocs/Nat.lean
```diff
@@ -68,4 +65,264 @@ builtin_dsimproc [seval] isValue ((OfNat.ofNat _ : Nat)) := fun e => do
...
+  | const (n : Nat)
+  /-- denotes `e + o` where `o` is expression definitionally equal to `n` -/
+  | offset (e o : Expr) (n : Nat)
+
+/- Attempt to parse a `NatOffset` from an expression-/
+private partial def NatOffset.asOffset (e : Expr) : Meta.SimpM (Option (Expr × Nat)) := do
+  if e.isAppOfArity ``HAdd.hAdd 6 then
+    let inst := e.appFn!.appFn!.appArg!
+    unless inst.isAppOfArity ``instHAdd 2 do return none
+    unless inst.appArg!.isAppOfArity ``instAddNat 0 do return none
```
eric-wieser, line 80: This is going to fail if the instance is actually `AddMonoid.toAdd` from mathlib, which seems very likely to occur.

eric-wieser: A small repro:
```lean
import Mathlib.Data.Nat.Basic

example : 1 + 2 = 3 := by
  rewrite [add_comm]
  -- goal is no longer about `instAddNat`
  sorry
  ```

joehendrix: Could you make an example showing this?  Using `import Mathlib` is fine.

eric-wieser: @joehendrix: I just went back in time and posted one :)

eric-wieser: Note that this was fixed in mathlib in https://github.com/leanprover-community/mathlib/pull/18129

Vierkantor: For reference, the pattern of matching operators by name and typeclasses by defeq is called "keyed matching". (I learned this name from [Reflexive tactics for algebra, revisited by Kazuhiko Sakaguchi](https://arxiv.org/abs/2202.04330), Section 3.2.)

semorrison:
> > * Mathlib branch [lean-pr-testing-3808](https://github.com/leanprover-community/mathlib4/compare/nightly-testing...lean-pr-testing-3808) built against this PR, but linting failed. (2024-04-04 21:47:46) [View Log](https://github.com/leanprover-community/mathlib4/actions/runs/8561215659)
> 
> @semorrison, this is misleading; the CI job "built", but the "build" step of the CI job failed.

Indeed, now that the linter runs even when the build fails, we need to change the logic.

PR #3819, opened 2024-04-01
fix: loose bound variables at ACLt 

### LINKED ISSUE #3705: `simp` failing with `unexpected bound variable`
Initially reported as https://github.com/leanprover-community/mathlib4/issues/11357, minimisation below due to @mattrobball:

```
structure MonoidHom (M : Type _) (N : Type _) [Mul M] [Mul N] where
  toFun : M → N
  map_mul' : ∀ x y, toFun (x * y) = toFun x * toFun y

class CommMagma (G : Type _) extends Mul G where
  mul_comm : ∀ a b : G, a * b = b * a

set_option quotPrecheck false
infixr:25 " →*' " => MonoidHom

instance [Mul M] [Mul N] : CoeFun (M →*' N) (fun _ => M → N) where
  coe := MonoidHom.toFun

open CommMagma

-- -- this instance needed
instance MonoidHom.commMonoid [Mul M] [Mul N] :
    CommMagma (M →*' N) where
  mul := fun f g => { toFun := fun x => f x * g x, map_mul' := sorry }
  mul_comm := sorry

example {M} [Mul M] [Mul G] [Pow G Int] :
    let zpow : Int → (M →*' G) → (M →*' G) := fun n f => { toFun := fun x => f x ^ n, map_mul' := sorry }
    ∀ (n : Nat) (a : M →*' G), zpow (Int.ofNat (Nat.succ n)) a = a * zpow (Int.ofNat n) a := by
  simp only [Int.ofNat_eq_coe] -- commenting out this line makes simp loop
  simp only [mul_comm] -- unexpected bound variable 2 
  sorry
```

leodemoura:
Closes #3705 

This PR also fixes a performance issue at `ACLt` also exposed by example at #3705

### DIFF
--- a/src/Lean/Meta/ACLt.lean
+++ b/src/Lean/Meta/ACLt.lean
@@ -50,7 +50,7 @@ mutual
    - We ignore metadata.
    - We ignore universe parameterst at constants.
 -/
-unsafe def main (a b : Expr) (mode : ReduceMode := .none) : MetaM Bool :=
+partial def main (a b : Expr) (mode : ReduceMode := .none) : MetaM Bool := do
   lt a b
 where
   reduce (e : Expr) : MetaM Expr := do
@@ -66,7 +66,9 @@ where
       | .none => return e
 
   lt (a b : Expr) : MetaM Bool := do
-    if ptrAddrUnsafe a == ptrAddrUnsafe b then
+    if a == b then
+      -- We used to have an "optimization" using only pointer equality.
+      -- This was a bad idea, `==` is often much cheaper than `acLt`.
       return false
     -- We ignore metadata
     else if a.isMData then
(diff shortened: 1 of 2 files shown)


src/Lean/Meta/ACLt.lean
```diff
@@ -66,7 +66,9 @@ where
       | .none => return e
 
   lt (a b : Expr) : MetaM Bool := do
-    if ptrAddrUnsafe a == ptrAddrUnsafe b then
+    if a == b then
+      -- We used to have an "optimization" using only pointer equality.
+      -- This was a bad idea, `==` is often much cheaper than `acLt`.
```
nomeata, line 71: A reader might think that the pointer equality optimization was a bad idea (and not the “only”)

```suggestion
      -- This was insufficient, as `==` is often much cheaper than `acLt`.
```

src/Lean/Meta/ACLt.lean
```diff
@@ -84,6 +86,16 @@ where
     else
       lt a₂ b₂
 
+  getParamsInfo (f : Expr) (numArgs : Nat) : MetaM (Array ParamInfo) := do
+    -- Ensure `f` does not have loose bound variables. This may happen in
```
nomeata, line 90:
```suggestion
    -- Ensure `f` does not have loose bound variables. This may happen
```

PR #3820, opened 2024-04-01
fix: do not lift `(<- ...)` over pure if-then-else

### LINKED ISSUE #3713: <- lifting over if produces incorrect code
### Prerequisites

### Description

<- can lift over if statements, which makes it easy to write non-performant or even wrong code when using branches in monadic code.

### Context

While I have seen this issue before, I still got bitten by it recently and spent several hours debugging a very unintuitive error that 
I think this (being aware of the issue and still falling in to it) points to the fact that this behaviour is unintuitive and should be changed or at least linted against by default.
Cf, some examples of this being discovered and fixed repeatedly across lean + mathlib https://github.com/leanprover/lean4/pull/1987, https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/.60.E2.86.90.60.20lifting.20over.20.60if.60.20footgun/near/379342271, https://github.com/leanprover-community/mathlib4/pull/6199, https://github.com/leanprover-community/mathlib4/pull/6197#pullrequestreview-1551151463

### Steps to Reproduce

```
import Lean
open Lean  Elab Meta

def somethingBad : MetaM Nat := do
  IO.println "oh no"
  return 1

#eval show MetaM Unit from do
  let t := if false then ← somethingBad else 9
```

**Expected behavior:** t should be 9  and nothing should be printed

**Actual behavior:** t is 9 and oh no is printed as `somethingBad` is executed anyway

### Versions

All

### Impact

leodemoura:
Now, only `(<- ...)`s occurring in the condition of a pure if-then-else are lifted.
That is, `if (<- foo) then ... else ...` is ok, but `if ... then (<- foo) else ...` is not. See #3713 

closes #3713 

This PR also adjusts this repo. Note that some of the `(<- ...)` were harmless since they were just accessing some
read-only state.

### DIFF
--- a/src/Lean/Compiler/IR/EmitLLVM.lean
+++ b/src/Lean/Compiler/IR/EmitLLVM.lean
@@ -147,7 +147,7 @@ def callLeanRefcountFn (builder : LLVM.Builder llvmctx)
     (delta : Option (LLVM.Value llvmctx) := Option.none) : M llvmctx Unit := do
   let fnName :=  s!"lean_{kind}{if checkRef? then "" else "_ref"}{if delta.isNone then "" else "_n"}"
   let retty ← LLVM.voidType llvmctx
-  let argtys := if delta.isNone then #[← LLVM.voidPtrType llvmctx] else #[← LLVM.voidPtrType llvmctx, ← LLVM.size_tType llvmctx]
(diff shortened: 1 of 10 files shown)


alexjbest:
Thank you Leo! I got hit by this again recently so I am very happy to see it fixed

PR #3824, opened 2024-04-02
fix: `simp only` should break `Char` literals

### LINKED ISSUE #3686: `simp only` breaks char literals
### Prerequisites

### Description

`Char` literals like `'a'` are notation for `Char.ofNat (nat_lit x)` where `x` is the character's unicode scalar value. `simp only` rewrites `Char.ofNat (nat_lit x)` to `Char.ofNat (OfNat.ofNat (nat_lit x))`, breaking the character literals in the process.

### Context

This bug is the cause of the one failing test in #3684.

### MWE

```lean
set_option pp.coercions false -- Show `OfNat.ofNat` for clarity
set_option pp.natLit true -- Show `nat_lit` for clarity

example : 'a' = sorry := by
-- ⊢ 'a' = sorryAx Char
  simp only
-- ⊢ Char.ofNat (OfNat.ofNat (nat_lit 97)) = sorryAx Char

-- Interestingly, `simp` without `only` behaves as expected:
example : 'a' = sorry := by
-- ⊢ 'a' = sorryAx Char
  simp -- error (as expected): simp made no progress
```

**Expected behavior:** `simp only` makes no progress as `'a'` should already be in simp-normal form.

**Actual behavior:** `simp only` rewrites `'a'` to `Char.ofNat 97`.

### Versions

4.7.0-rc2

### Additional Information

`simp` without `only` has the expected behavior. My guess is that this is because the simpproc [Char.isValue](https://leanprover-community.github.io/mathlib4_docs/Lean/Meta/Tactic/Simp/BuiltinSimprocs/Char.html#Char.isValue) stops the simplifier from entering the literal.

One possible fix would be to make sure the `Char.isValue` simpproc fires even with `simp only` (or `dsimp only`). This is currently done for `OfNat` and `OfScientific` with special handling [in `dsimpImpl`](https://github.com/leanprover/lean4/blob/4e3a8468c351531235ba5cffd5566b7283b7a8ca/src/Lean/Meta/Tactic/Simp/Main.lean#L445) and [in `simpApp`](https://github.com/leanprover/lean4/blob/4e3a8468c351531235ba5cffd5566b7283b7a8ca/src/Lean/Meta/Tactic/Simp/Main.lean#L565)

### Impact

leodemoura:
closes #3686

### DIFF
--- a/src/Lean/Meta/Tactic/Simp/Main.lean
+++ b/src/Lean/Meta/Tactic/Simp/Main.lean
@@ -54,6 +54,10 @@ def foldRawNatLit (e : Expr) : SimpM Expr := do
 def isOfScientificLit (e : Expr) : Bool :=
   e.isAppOfArity ``OfScientific.ofScientific 5 && (e.getArg! 4).isRawNatLit && (e.getArg! 2).isRawNatLit
 
+/-- Return true if `e` is of the form `Char.ofNat n` where `n` is a kernel Nat literals. -/
+def isCharLit (e : Expr) : Bool :=
+  e.isAppOfArity ``Char.ofNat 1 && e.appArg!.isRawNatLit
+
 private def reduceProjFn? (e : Expr) : SimpM (Option Expr) := do
   matchConst e.getAppFn (fun _ => pure none) fun cinfo _ => do
     match (← getProjectionFnInfo? cinfo.name) with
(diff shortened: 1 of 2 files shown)


nomeata:
I assume the title is lying :-)

PR #3832, opened 2024-04-04
fix: `String.splitOn` bug

### LINKED ISSUE #3829: Definition of `String.splitOn` is incorrect
### Description

```lean
#eval "012".splitOn "12" -- output: ["0", ""]
#eval "007".splitOn "07" -- ["007"]
```

The value of the second expression above should be `["0", ""]`.

### Context

I was trying to prove [`String.splitOn_of_valid`](https://github.com/leanprover/std4/pull/719/files#diff-b2fb96bb7d0273dbe3a81a67bb9b432c089139ac50e3e0b3578f40cac0de2a51R662-R672) in `Std.Data.List.Lemmas`, but I realized that the current definition of `String.splitOn` is flawed. See https://github.com/leanprover/std4/pull/719.

I think @digama0's definition of [`List.splitOnList`](https://github.com/leanprover/std4/pull/719/files#diff-da0f75af783a1dd2e2f8d5bd692a63a1d3b96cec323e322a21db342852d09c6eR97-R112) in [`Std.Data.List.SplitOnList`](https://github.com/leanprover/std4/pull/719/files#diff-da0f75af783a1dd2e2f8d5bd692a63a1d3b96cec323e322a21db342852d09c6e) is correct, as can be seen below:

```lean
import Std.Data.List.SplitOnList

#eval ("012".1.splitOnList "12".1).map String.mk -- output: ["0", ""]
#eval ("007".1.splitOnList "07".1).map String.mk -- ["0", ""]
```

Discussions on the Lean Zulip chat: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/current.20definition.20of.20.60String.2EsplitOn.60.20is.20incorrect/near/430930535

### Version

4.7.0-rc2

### Impact

digama0:
Fixes #3829. As reported on Zulip (both [recently](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/current.20definition.20of.20.60String.2EsplitOn.60.20is.20incorrect/near/430930535) and [a year ago](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/should.20we.20redefine.20.60String.2EsplitOnAux.60.3F/near/365899332)), `String.splitOn` has a bug when dealing with separators of more than one character (which are luckily rare). The code change here is very small, replacing a `i` with `i - j`, but it makes termination more complex so that's where the rest of the line count goes.

### DIFF
--- a/src/Init/Data/String/Basic.lean
+++ b/src/Init/Data/String/Basic.lean
@@ -245,6 +245,16 @@ termination_by s.endPos.1 - i.1
 @[specialize] def split (s : String) (p : Char → Bool) : List String :=
   splitAux s p 0 0 []
 
+/--
+Auxiliary for `splitOn`. Preconditions:
+* `sep` is not empty
+* `b <= i` are indexes into `s`
+* `j` is an index into `sep`, and not at the end
+
+It represents the state where we have currently parsed some split parts into `r` (in reverse order),
+`b` is the beginning of the string / the end of the previous match of `sep`, and the first `j` bytes
+of `sep` match the bytes `i-j .. i` of `s`.
+-/
 def splitOnAux (s sep : String) (b : Pos) (i : Pos) (j : Pos) (r : List String) : List String :=
   if s.atEnd i then
     let r := (s.extract b i)::r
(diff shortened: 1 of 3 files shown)


Kha (approved):
Looks good but could you add a short doc on the meaning of b/i/j before we forget again? The current code is quite opaque

digama0:
I also added a doc comment to `splitOn`. (Why are all of these functions still undocumented? I went through all the list functions some time ago but I guess we don't have anyone on docs duty.)

Kha:
Perfect!

PR #3833, opened 2024-04-04
doc: update release checklist based on experience with 4.7.0

david-christiansen:
@semorrison, does this include all the answers to the questions I asked in our thread? I think so!

### DIFF
--- a/doc/dev/release_checklist.md
+++ b/doc/dev/release_checklist.md
@@ -40,7 +40,6 @@ We'll use `v4.6.0` as the intended release version as a running example.
         If they depend on `main` or `master`, don't change this; you've just updated the dependency, so it will work and be saved in the manifest
       - Run `lake update`
       - The PR title should be "chore: bump toolchain to v4.6.0".
-      - Since the `v4.6.0` release should be functionally identical to the last release candidate, which the repository should already be on, this PR is a no-op besides changing the toolchain, the Lakefile, and the manifest
     - Merge the PR once CI completes.
     - Create the tag `v4.6.0` from `master`/`main` and push it.
     - Merge the tag `v4.6.0` into the `stable` branch and push it.


doc/dev/release_checklist.md
```diff
@@ -34,48 +34,77 @@ We'll use `v4.6.0` as the intended release version as a running example.
...
-      which the repository should already be on, this PR is a no-op besides changing the toolchain.
-    - Once this is merged, create the tag `v4.6.0` from `master`/`main` and push it.
-    - Merge the tag `v4.6.0` into the stable branch.
+    - Make a PR to `master`/`main` changing the toolchain to `v4.6.0`
+      - Update the toolchain file
+      - In the Lakefile, if there are dependencies on specific version tags of dependencies that you've already pushed as part of this process, update them to the new tag.
+        If they depend on `main` or `master`, don't change this; you've just updated the dependency, so it will work and be saved in the manifest
+      - Run `lake update`
+      - The PR title should be "chore: bump toolchain to v4.6.0".
+      - Since the `v4.6.0` release should be functionally identical to the last release candidate, which the repository should already be on, this PR is a no-op besides changing the toolchain, the Lakefile, and the manifest
```
Kha, line 43: As we've seen in the case of Mathlib, "no-op" is not necessarily true since we might have done an arbitrary dependency jump with potential breaking changes, no?

doc/dev/release_checklist.md
```diff
@@ -21,7 +21,7 @@ We'll use `v4.6.0` as the intended release version as a running example.
   - Reconcile discrepancies in the `v4.6.0` section,
     usually via copy and paste and a commit to `releases/v4.6.0`.
 - `git tag v4.6.0`
-- `git push origin v4.6.0`
+- `git push $REMOTE v4.6.0`, where `$REMOTE` is the upstream Lean repository
```
tydeu, line 24:
```suggestion
- `git push $REMOTE v4.6.0`, where `$REMOTE` is the upstream Lean repository (e.g., `origin`, `upstream`)
```

Providing an example seems useful.

doc/dev/release_checklist.md
david-christiansen: Good point :) This should be changed.

PR #3837, opened 2024-04-06
perf: improve heuristic at `isDefEq`

kim-em:
Adds a test case containing a minimization of a Mathlib slowdown from #3807, and a patch to isDefEq to fix the slowdown.

Prior to #3807, the declaration `exists_algHom_adjoin_of_splits'''` at the end of the file would take around 16,000 heartbeats. After #3807 it took around 210,000 heartbeats.

### DIFF
--- a/src/Lean/Meta/ExprDefEq.lean
+++ b/src/Lean/Meta/ExprDefEq.lean
@@ -1173,6 +1173,35 @@ private def isDefEqLeftRight (fn : Name) (t s : Expr) : MetaM LBool := do
   trace[Meta.isDefEq.delta.unfoldLeftRight] fn
   toLBoolM <| Meta.isExprDefEqAux t s
 
+/-- Helper predicate for `tryHeuristic`. -/
+private def isNonTrivialRegular (info : DefinitionVal) : MetaM Bool := do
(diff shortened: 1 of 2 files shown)


semorrison:
Curiously, deleting `setOf` in both

```
def range (f : ι → α) : Set α := setOf fun x => ∃ y, f y = x
```
and
```
def image (f : α → β) (s : Set α) : Set β := setOf fun b => ∃ a ∈ s, f a = b
```
reduces the heartbeat count to around 80,000.

semorrison:
These are the outputs of
```
set_option trace.Meta.isDefEq true in
exact (congrFun (congrArg (fun g : L' →ₐ[F] K => (g : L' → K)) hφ) _).trans (congrArg f <| AlgEquiv.symm_apply_apply _ _)
```
(i.e. the step that is timing out) both before and after #3807. Be warned that `after.txt` is about 800mb (and even that is truncated as it doesn't seem to finish).
[after.txt.gz](https://github.com/leanprover/lean4/files/14892841/after.txt.gz)
[before.txt.gz](https://github.com/leanprover/lean4/files/14892844/before.txt.gz)

PR #3838, opened 2024-04-06
fix: offset typeclass checking in simp rules

### LINKED ISSUE #3836: Simp offset checking fails with non-canonical instances.
The logic for checking if a term is an offset seems to do a syntactic check that prevents simp rules from firing.

This appears to be a regression introduced by #3614.  Prior to that, the example below succeeds, but after #3614 lean returns `error: simp made no progress`.

```lean
class AddCommMagma (G : Type u) extends Add G where
  /-- Addition is commutative in an commutative additive magma. -/
  protected add_comm : ∀ a b : G, a + b = b + a

instance : AddCommMagma Nat where
  add_comm := Nat.add_comm

example (n : Nat) (p : n = 3) : 1 + n = 1 + 3 := by
  rw [AddCommMagma.add_comm 1, AddCommMagma.add_comm 1]
  simp only [Nat.succ.injEq]
  exact p
```

joehendrix:
This changes how Nat typeclass checks in offset terms from syntactic equality to definitional equality with "instances" transparency.

This may have a negative performance penalty in `isOffset?`, but it should be small in common cases since the relevant instances are small terms.

This closes #3836

### DIFF
--- a/src/Lean/Expr.lean
+++ b/src/Lean/Expr.lean
@@ -2022,17 +2022,46 @@ def mkEM (p : Expr) : Expr := mkApp (mkConst ``Classical.em) p
 /-- Return `p ↔ q` -/
 def mkIff (p q : Expr) : Expr := mkApp2 (mkConst ``Iff) p q
 
+/-! Constants for Nat typeclasses. -/
+namespace Nat
+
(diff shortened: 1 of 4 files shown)


joehendrix:
!bench

PR #3846, opened 2024-04-08
doc: `lean.h`

marcusrossel:
This PR adds documentation comments for a few functions in `lean.h`. This was previously discussed at https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Structure.20in.20FFI/near/431985775

I think it would be good if someone knowledgable (@digama0?) could check that the content of the comments is actually correct.

### DIFF
--- a/src/include/lean/lean.h
+++ b/src/include/lean/lean.h
@@ -454,7 +454,7 @@ LEAN_EXPORT void lean_dealloc(lean_object * o);
 /*
 An object `o` satisfies `lean_is_ctor` if it is a member of an inductive type, barring the exceptions
 listed at https://lean-lang.org/lean4/doc/dev/ffi.html#translating-types-from-lean-to-c.
-One notable exception is that inductive types where all constructors have no parameters (also called _enumeration types_) is represented more efficiently, so their values do not satisfy `lean_is_ctor`. 
+One notable exception is that inductive types where all constructors have no parameters (also called _enumeration types_) are represented more efficiently, so their values do not satisfy `lean_is_ctor`. 
 Behind the scenes, structure types are inductive types with a single constructor (whose parameters are the structure's fields), so values of structure types typically satisfy `lean_is_ctor`.
 
 Accessing the fields of such an object `o` works differently depending on the type of the field.
@@ -535,7 +535,6 @@ static inline void lean_set_non_heap_header_for_big(lean_object * o, unsigned ta
 
 /*
 Returns the number of non-scalar parameters of a given ctor-object. 
-
 For example, let `o` be a value of the following structure in Lean:
 
 ```
@@ -573,10 +572,7 @@ static inline lean_object * lean_alloc_ctor(unsigned tag, unsigned num_objs, uns
 }
 
 /*
-Returns the `i`th non-scalar parameter of a given ctor-object. Notably,
-the index `i` does not index the entire list of parameters of a given 
-constructor, but rather only the non-scalar ones.
-
+Returns the `i`th non-scalar parameter of a given ctor-object. 
 For example, let `o` be a value of the following structure in Lean:
 
 ```
@@ -594,6 +590,9 @@ Then
 * `lean_ctor_get(o, 2)` returns the object representing `e`
 
 To get the number of non-scalar parameters use `lean_ctor_num_objs`.
+
+For a more complete description of how inductive types' values are layed out,
+see https://lean-lang.org/lean4/doc/dev/ffi.html#inductive-types.
 */
 static inline b_lean_obj_res lean_ctor_get(b_lean_obj_arg o, unsigned i) {
     assert(i < lean_ctor_num_objs(o));
@@ -624,38 +623,8 @@ static inline size_t lean_ctor_get_usize(b_lean_obj_arg o, unsigned i) {
 
 /*
 Given a ctor-object, returns a scalar parameter of type `uint8_t` at a given offset. 
-
-The offset does *not* represent an index into the list of parameters of an inductive 
-type constructor, but is instead only a byte offset. To access a specific constructor 
-parameter, construct the correct byte offset as follows:
-
-All scalar parameters of a ctor-object are represented in a contiguous block following
-the non-scalar parameters. Thus, the base offset required to access a scalar parameter
-of a ctor-object `o` is `lean_ctor_num_objs(o) * sizeof(void*)` (see the documentation 
-for `lean_ctor_num_objs` for more details). Then, to locate the desired scalar parameter, 
-add the sizes of all scalar parameters preceding the target parameter. For example, 
-let `o` be a value of the following structure in Lean:
-
-```
-structure S where
-  a : String
-  b : Bool
-  c : UInt64
-  e : Array Nat
-  d : UInt8
-```
-
-Then 
-* the base offset is `base_offset := 3 * sizeof(void*)`.
-* as `b` is the first scalar parameter, its offset is exactly the `base_offset`.
-* the offset for `c` is `base_offset + sizeof(uint8_t)`, where `sizeof(uint8_t)` 
-  is the number of bytes used by the preceding parameter `b`. 
-* the offset for `d` is `base_offset + sizeof(uint8_t) + sizeof(uint64_t)` where
-  `sizeof(uint8_t) + sizeof(uint64_t)` are the number of bytes used by the preceeding
-  parameters `b` and `c`.
-
-Thus, `lean_ctor_get_uint8(o, 3 * sizeof(void*) + sizeof(uint8_t) + sizeof(uint64_t))`
-returns the value of `d`.
+For a description of how to compute the correct offset for accessing a specific parameter,
+see https://lean-lang.org/lean4/doc/dev/ffi.html#inductive-types.
 */
 static inline uint8_t lean_ctor_get_uint8(b_lean_obj_arg o, unsigned offset) {
     assert(offset >= lean_ctor_num_objs(o) * sizeof(void*));
@@ -664,7 +633,8 @@ static inline uint8_t lean_ctor_get_uint8(b_lean_obj_arg o, unsigned offset) {
 
 /*
 Given a ctor-object, returns a scalar parameter of type `uint16_t` at a given offset. 
-For an explanation of the offset, see the documentation for `lean_ctor_get_uint8`.
+For a description of how to compute the correct offset for accessing a specific parameter,
+see https://lean-lang.org/lean4/doc/dev/ffi.html#inductive-types.
 */
 static inline uint16_t lean_ctor_get_uint16(b_lean_obj_arg o, unsigned offset) {
     assert(offset >= lean_ctor_num_objs(o) * sizeof(void*));
@@ -673,7 +643,8 @@ static inline uint16_t lean_ctor_get_uint16(b_lean_obj_arg o, unsigned offset) {
 
 /*
 Given a ctor-object, returns a scalar parameter of type `uint32_t` at a given offset. 
-For an explanation of the offset, see the documentation for `lean_ctor_get_uint8`.
+For a description of how to compute the correct offset for accessing a specific parameter,
+see https://lean-lang.org/lean4/doc/dev/ffi.html#inductive-types.
 */
 static inline uint32_t lean_ctor_get_uint32(b_lean_obj_arg o, unsigned offset) {
     assert(offset >= lean_ctor_num_objs(o) * sizeof(void*));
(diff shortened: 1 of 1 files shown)


src/include/lean/lean.h
```diff
@@ -451,6 +451,19 @@ static inline void lean_dec(lean_object * o) { if (!lean_is_scalar(o)) lean_dec_
 /* Just free memory */
 LEAN_EXPORT void lean_dealloc(lean_object * o);
 
+/*
+An object `o` satisfies `lean_is_ctor` if it is a member of an inductive type, barring the exceptions
+listed at https://lean-lang.org/lean4/doc/dev/ffi.html?highlight=ffi#translating-types-from-lean-to-c.
```
david-christiansen, line 456:
```suggestion
listed at https://lean-lang.org/lean4/doc/dev/ffi.html#translating-types-from-lean-to-c.
```
The highlight parameter was leading to "sufficient" having a highlight, which I don't think we want here.

src/include/lean/lean.h
```diff
@@ -451,6 +451,19 @@ static inline void lean_dec(lean_object * o) { if (!lean_is_scalar(o)) lean_dec_
...
+A notable example of an exception are inductive types where all constructors have no parameters (enumeration types).
```
david-christiansen, line 457: The singular/plural isn't really working here. What about:
```suggestion
One notable exception is that inductive types where all constructors have no parameters (also called _enumeration types_) is represented more efficiently, so their values do not satisfy `lean_is_ctor`. 
```

src/include/lean/lean.h
```diff
@@ -451,6 +451,19 @@ static inline void lean_dec(lean_object * o) { if (!lean_is_scalar(o)) lean_dec_
 /* Just free memory */
 LEAN_EXPORT void lean_dealloc(lean_object * o);
 
+/*
+An object `o` satisfies `lean_is_ctor` if it is a member of an inductive type, barring the exceptions
+listed at https://lean-lang.org/lean4/doc/dev/ffi.html?highlight=ffi#translating-types-from-lean-to-c.
+A notable example of an exception are inductive types where all constructors have no parameters (enumeration types). 
+A notable example of types which *do* satisfy `lean_is_ctor` are structure types which are represented
+as inductive types with a single constructor (whose parameters are the structure's fields).
```
david-christiansen, line 459: This is phrased as if it's an exception, but it really isn't, given the way the elaborator treats structures. What about:
```suggestion
Behind the scenes, structure types are inductive types with a single constructor (whose parameters are the structure's fields), so values of structure types typically satisfy `lean_is_ctor`.
```

src/include/lean/lean.h
```diff
@@ -521,6 +534,23 @@ static inline void lean_set_non_heap_header_for_big(lean_object * o, unsigned ta
...
+
+For example, let `o` be a value of the following structure in Lean:
+
+```
+structure S where
+    a : String
+    b : String
+    c : Bool
+    d : UInt64
+    e : Array Nat
```
david-christiansen, line 548: More typical formatting is:
```suggestion
structure S where
  a : String
  b : String
  c : Bool
  d : UInt64
  e : Array Nat
```

src/include/lean/lean.h
```diff
@@ -543,6 +573,29 @@ static inline lean_object * lean_alloc_ctor(unsigned tag, unsigned num_objs, uns
...
```
david-christiansen, line 589:
```suggestion
structure S where
  a : String
  b : String
  c : Bool
  d : UInt64
  e : Array Nat
```

src/include/lean/lean.h
```diff
@@ -570,26 +623,77 @@ static inline size_t lean_ctor_get_usize(b_lean_obj_arg o, unsigned i) {
...
+add the sizes of  all scalar parameters preceeding the target parameter. For example, 
+let `o` be a value of the following structure in Lean:
+
+```
+structure S where
+    a : String
+    b : Bool
+    c : UInt64
+    e : Array Nat
+    d : UInt8
```
david-christiansen, line 646:
```suggestion
structure S where   a : String
  b : Bool
  c : UInt64
  e : Array Nat
  d : UInt8
```

src/include/lean/lean.h
```diff
@@ -570,26 +623,77 @@ static inline size_t lean_ctor_get_usize(b_lean_obj_arg o, unsigned i) {
     return *((size_t*)(lean_ctor_obj_cptr(o) + i));
 }
 
+/*
+Given a ctor-object, returns a scalar parameter of type `uint8_t` at a given offset. 
+
+The offset does *not* represent an index into the list of parameters of an inductive 
+type constructor, but is instead only a byte offset. To access a specific constructor 
+parameter, construct the correct byte offset as follows.
```
david-christiansen, line 631:
```suggestion
parameter, construct the correct byte offset as follows:
```

src/include/lean/lean.h
```diff
@@ -570,26 +623,77 @@ static inline size_t lean_ctor_get_usize(b_lean_obj_arg o, unsigned i) {
...
+
+The offset does *not* represent an index into the list of parameters of an inductive 
+type constructor, but is instead only a byte offset. To access a specific constructor 
+parameter, construct the correct byte offset as follows. 
+
+All scalar parameters of a ctor-object are represented in a contiguous block following
+the non-scalar parameters. Thus, the base offset required to access a scalar parameter
+of a ctor-object `o` is `lean_ctor_num_objs(o) * sizeof(void*)` (cf. the documentation 
+for `lean_ctor_num_objs` for more details). Then, to locate the desired scalar parameter, 
+add the sizes of  all scalar parameters preceeding the target parameter. For example,
```
david-christiansen, line 637:
```suggestion
add the sizes of all scalar parameters preceeding the target parameter. For example, 
```

src/include/lean/lean.h
```diff
@@ -570,26 +623,77 @@ static inline size_t lean_ctor_get_usize(b_lean_obj_arg o, unsigned i) {
...
+    c : UInt64
+    e : Array Nat
+    d : UInt8
+```
+
+Then 
+* the base offset is `base_offset := 3 * sizeof(void*)`.
+* as `b` is the first scalar parameter, its offset is exactly the `base_offset`.
+* the offset for `c` is `base_offset + sizeof(uint8_t)`, where `sizeof(uint8_t)` 
+  is the number of bytes used by the preceeding parameter `b`.
```
david-christiansen, line 653:
```suggestion
  is the number of bytes used by the preceding parameter `b`. 
```

Also, where should readers go to find out how much space is used by `Bool` and thus `b`?

src/include/lean/lean.h
```diff
@@ -570,26 +623,77 @@ static inline size_t lean_ctor_get_usize(b_lean_obj_arg o, unsigned i) {
...
+returns the value of `d`.
+*/
 static inline uint8_t lean_ctor_get_uint8(b_lean_obj_arg o, unsigned offset) {
     assert(offset >= lean_ctor_num_objs(o) * sizeof(void*));
     return *((uint8_t*)((uint8_t*)(lean_ctor_obj_cptr(o)) + offset));
 }
 
+/*
+Given a ctor-object, returns a scalar parameter of type `uint16_t` at a given offset. 
+For an explanation of the offset, cf. the documentation for `lean_ctor_get_uint8`.
```
david-christiansen, line 668:
```suggestion
For an explanation of the offset, see the documentation for `lean_ctor_get_uint8`.
```

src/include/lean/lean.h
```diff
@@ -570,26 +623,77 @@ static inline size_t lean_ctor_get_usize(b_lean_obj_arg o, unsigned i) {
...
+structure S where
+    a : String
+    b : Bool
+    c : UInt64
+    e : Array Nat
+    d : UInt8
+```
+
+Then 
+* the base offset is `base_offset := 3 * sizeof(void*)`.
```
digama0, line 650: This should be `2 * sizeof(void*)` I think, since `a` and `e` (why is `e` before `d` in the example?) are the non-scalars.

src/include/lean/lean.h
```diff
@@ -570,26 +623,77 @@ static inline size_t lean_ctor_get_usize(b_lean_obj_arg o, unsigned i) {
...
+For an explanation of the offset, cf. the documentation for `lean_ctor_get_uint8`.
+*/
 static inline uint16_t lean_ctor_get_uint16(b_lean_obj_arg o, unsigned offset) {
     assert(offset >= lean_ctor_num_objs(o) * sizeof(void*));
     return *((uint16_t*)((uint8_t*)(lean_ctor_obj_cptr(o)) + offset));
 }
 
+/*
+Given a ctor-object, returns a scalar parameter of type `uint32_t` at a given offset. 
+For an explanation of the offset, cf. the documentation for `lean_ctor_get_uint8`.
```
david-christiansen, line 677:
```suggestion
For an explanation of the offset, see the documentation for `lean_ctor_get_uint8`.
```

src/include/lean/lean.h
```diff
@@ -570,26 +623,77 @@ static inline size_t lean_ctor_get_usize(b_lean_obj_arg o, unsigned i) {
...
+*/
 static inline uint32_t lean_ctor_get_uint32(b_lean_obj_arg o, unsigned offset) {
     assert(offset >= lean_ctor_num_objs(o) * sizeof(void*));
     return *((uint32_t*)((uint8_t*)(lean_ctor_obj_cptr(o)) + offset));
 }
 
+/*
+Given a ctor-object, returns a scalar parameter of type `uint64_t` at a given offset. 
+For an explanation of the offset, cf. the documentation for `lean_ctor_get_uint8`.
```
david-christiansen, line 686:
```suggestion
For an explanation of the offset, see the documentation for `lean_ctor_get_uint8`.
```

src/include/lean/lean.h
```diff
@@ -570,26 +623,77 @@ static inline size_t lean_ctor_get_usize(b_lean_obj_arg o, unsigned i) {
...
+*/
 static inline uint64_t lean_ctor_get_uint64(b_lean_obj_arg o, unsigned offset) {
     assert(offset >= lean_ctor_num_objs(o) * sizeof(void*));
     return *((uint64_t*)((uint8_t*)(lean_ctor_obj_cptr(o)) + offset));
 }
 
+/*
+Given a ctor-object, returns a scalar parameter of type `float` at a given offset. 
+For an explanation of the offset, cf. the documentation for `lean_ctor_get_uint8`.
```
david-christiansen, line 695:
```suggestion
For an explanation of the offset, see the documentation for `lean_ctor_get_uint8`.
```

src/include/lean/lean.h
david-christiansen: Oh, that is tricky! My eyes bounced right off that.

src/include/lean/lean.h
```diff
@@ -570,26 +623,77 @@ static inline size_t lean_ctor_get_usize(b_lean_obj_arg o, unsigned i) {
...
+Then 
+* the base offset is `base_offset := 3 * sizeof(void*)`.
+* as `b` is the first scalar parameter, its offset is exactly the `base_offset`.
+* the offset for `c` is `base_offset + sizeof(uint8_t)`, where `sizeof(uint8_t)` 
+  is the number of bytes used by the preceeding parameter `b`. 
+* the offset for `d` is `base_offset + sizeof(uint8_t) + sizeof(uint64_t)` where
+  `sizeof(uint8_t) + sizeof(uint64_t)` are the number of bytes used by the preceeding
+  parameters `b` and `c`.
+
+Thus, `lean_ctor_get_uint8(o, 3 * sizeof(void*) + sizeof(uint8_t) + sizeof(uint64_t))`
```
digama0, line 658: This does not take alignment into account. I actually don't know whether these fields are aligned (my independent reimplementation of the layout algorithm [here](https://github.com/digama0/oleandump/blob/acc2ef6ad12dba9edf28ad3c24ab50da9e95f40f/OLeanDump/Object.lean#L90-L135) doesn't seem to do alignment), but the text should be edited to assert this one way or another.

digama0:
From a documentation / explanation point of view, I'm not sure about the idea of spreading the layout algorithm over these individual functions. I think the best approach would be to have the layout algorithm in one place (e.g. `ffi.md`), and then have these functions just talk about how they relate to the lean data model, assuming the details of inductive type <-> layout are worked out.

digama0:
I added docs for the layout algorithm in #3915. This should take some of the pressure off to try to explain the details in individual accessors.

src/include/lean/lean.h
marcusrossel: https://lean-lang.org/lean4/doc/dev/ffi.html#translating-types-from-lean-to-c explains that `Bool` is represented as a `uint8_t`.

src/include/lean/lean.h
marcusrossel: I like the phrasing better, but now there's a weird transition from the previous sentence. So perhaps we can just remove this sentence entirely?

marcusrossel:
I've removed my (incorrect) description of layout algorithm now, instead linking to @digama0's section in the Lean Manual.

marcusrossel:
*ping*

Kha:
@marcusrossel Thanks for the PR and sorry for the long wait. We took a look at the PR in the triage team meeting and decided that we would like to keep documentation on this level concentrated in ffi.md so it doesn't go out of sync. Also, `lean_is_ctor` and `lean_ctor_num_objs` are not relevant for FFI. Thus I'm going to close this PR as all relevant information should already exist in ffi.md.

PR #3847, opened 2024-04-08
feat: omega: more helpful error messages

nomeata:
while trying to help a user who was facing an unhelpful
```
omega did not find a contradiction:
[0, 0, 0, 0, 1, -1] ∈ [1, ∞)
[0, 0, 0, 0, 0, 1] ∈ [0, ∞)
[0, 0, 0, 0, 1] ∈ [0, ∞)
[1, -1] ∈ [1, ∞)
[0, 0, 0, 1] ∈ [0, ∞)
[0, 1] ∈ [0, ∞)
[1] ∈ [0, ∞)
[0, 0, 0, 1, 1] ∈ [-1, ∞)
```
I couldn’t resist and wrote a pretty-printer for these problem that
shows the linear combination as such, and includes the recognized atoms.
This is especially useful since oftem `omega` failures stem from failure
to recognize atoms as equal. In this case, we now get:

```
omega-failure.lean:19:2-19:7: error: omega could not prove the goal:
a possible counterexample may satisfy the constraints
  d - e ≥ 1
  e ≥ 0
  d ≥ 0
  a - b ≥ 1
  c ≥ 0
  b ≥ 0
  a ≥ 0
  c + d ≥ -1
where
 a := ↑(sizeOf xs)
 b := ↑(sizeOf x)
 c := ↑(sizeOf x.fst)
 d := ↑(sizeOf x.snd)
 e := ↑(sizeOf xs)
```
and this might help the user make progress (e.g. by using `case x`
first, and investingating why `sizeOf xs` shows up twice)

### DIFF
--- a/tests/lean/omega-failure.lean
+++ b/tests/lean/omega-failure.lean
@@ -10,8 +10,11 @@ example (x : Nat) (y : Int) : x < y := by omega
 
 example (x y : Int) : 5 < x ∧ x < 10 → y > 0 := by omega
 
-theorem sizeOf_snd_lt_sizeOf_list {α : Type u} {β : Type v} [SizeOf α] [SizeOf β] {x : α × β} {xs : List (α × β)} :
-  x ∈ xs → sizeOf x.snd < 1 + sizeOf xs
+-- this used to fail with y = x, but then omega got better, so now there are unrelated x and y
+-- to make omega fail
+theorem sizeOf_snd_lt_sizeOf_list {α : Type u} {β : Type v} [SizeOf α] [SizeOf β]
+  {x y : α × β} {xs : List (α × β)} :
+  y ∈ xs → sizeOf x.snd < 1 + sizeOf xs
 := by
   intro h
   have := List.sizeOf_lt_of_mem h
--- a/tests/lean/omega-failure.lean.expected.out
+++ b/tests/lean/omega-failure.lean.expected.out
@@ -19,23 +19,21 @@ a possible counterexample may satisfy the constraints
 where
  a := x
  b := y
-omega-failure.lean:19:2-19:7: error: omega could not prove the goal:
+omega-failure.lean:22:2-22:7: error: omega could not prove the goal:
 a possible counterexample may satisfy the constraints
-  d - e ≥ 1
-  e ≥ 0
   d ≥ 0
   a - b ≥ 1
   c ≥ 0
+  a - d ≤ -1
   b ≥ 0
   a ≥ 0
   c + d ≥ -1
 where
  a := ↑(sizeOf xs)
- b := ↑(sizeOf x)
+ b := ↑(sizeOf y)
  c := ↑(sizeOf x.fst)
  d := ↑(sizeOf x.snd)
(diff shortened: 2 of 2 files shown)


nomeata:
Draft status to indicate that I don't consider the code finished, but would like to hear if I should go ahead with this, and what changes, if any, you’d like to see for the output.

(I wonder if such non-trivial error messages should be constructed lazily, using `ofPPFormat` somehow, so that `try omega` does not get slower.)

nomeata:
I was expecting this to be somewhat expensive in the presence of `try omega`, but the speedcenter thinks it’s fine:

http://speed.lean-fro.org/mathlib4/compare/a93fa055-e732-498b-8c1f-fd5ab4f48b05/to/fe837d0c-f16b-4edb-981d-f63ef9508ed5

So I guess we can merge this as it is (and maybe use lazy error message construction here once we have it.)

src/Lean/Elab/Tactic/Omega/Frontend.lean
```diff
@@ -526,6 +526,59 @@ def cases₂ (mvarId : MVarId) (p : Expr) (hName : Name := `h) :
...
+      let mask ← mentioned p.constraints
+      return m!"a possible counterexample may satisfy the constraints\n" ++
+        m!"{prettyConstraints p.constraints}\nwhere\n" ++
+        (← prettyAtoms mask)
+  else
+    -- formatErrorMessage should not be used in this case
+    return "it is trivially solvable"
+where
+  var (n : Nat) : String :=
+    "x" ++ ((toString (n+1)).map fun c => .ofNat (c.toNat - '0'.toNat + '₀'.toNat))
```
nomeata, line 546: Will extract that as `Nat.toSubscript`, otherwise this is probably good enough as a first iteration

david-christiansen:
This is really good stuff :-)

But are subscripted `x`s the easiest to read? Using different letters, perhaps resorting to subscripts when exhausting a name supply, would be easier to distinguish at a glace, I think. Eg:
```
omega-failure.lean:19:2-19:7: error: omega could not prove the goal:
a possible counterexample may satisfy the constraints
  d - e ≥ 1
  e ≥ 0
  d ≥ 0
  a - b ≥ 1
  c ≥ 0
  b ≥ 0
  a ≥ 0
  c + d ≥ -1
where
  a := ↑(sizeOf xs)
  b := ↑(sizeOf x)
  c := ↑(sizeOf x.fst)
  d := ↑(sizeOf x.snd)
  e := ↑(sizeOf xs)
```

Otherwise we may end up squinting, especially when working on laptops.

This example is not perfect - It's weird using beginning-of-alphabet for variables rather than constants, and the naming could use late-alphabet letters. But I do think separate names is good.

In either case, I would hope that the name selection avoids things already in scope, which is needed with subscripted variables anyway, as users might do that. And I noticed that there was no x_3 atom in the original message - is that because the atom was unconstrained?

nomeata:
As much as I 🤎 unicode, I fear you are right.

nomeata:
@david-christiansen, I implemented your idea, I think. What do you think of that?

PR #3850, opened 2024-04-09
feat: complete Int div/mod simprocs

joehendrix:
This PR introduces complete simprocs for all the Int versions of div/mod, and makes some small refactoring of Int lemmas and library_search.

### DIFF
--- a/src/Init/Data/Int/Order.lean
+++ b/src/Init/Data/Int/Order.lean
@@ -590,7 +590,7 @@ theorem lt_add_one_iff {a b : Int} : a < b + 1 ↔ a ≤ b := Int.add_le_add_iff
 @[simp] theorem succ_ofNat_pos (n : Nat) : 0 < (n : Int) + 1 :=
   lt_add_one_iff.mpr (ofNat_zero_le _)
 
-theorem ofNat_not_neg (n : Nat) : ¬((n : Int) < 0) :=
+theorem not_ofNat_neg (n : Nat) : ¬((n : Int) < 0) :=
   Int.not_lt.mpr (ofNat_zero_le ..)
 
 theorem le_add_one {a b : Int} (h : a ≤ b) : a ≤ b + 1 :=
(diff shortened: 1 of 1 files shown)


src/Init/Data/Int/Order.lean
```diff
@@ -584,7 +586,10 @@ theorem add_one_le_iff {a b : Int} : a + 1 ≤ b ↔ a < b := .rfl
 theorem lt_add_one_iff {a b : Int} : a < b + 1 ↔ a ≤ b := Int.add_le_add_iff_right _
 
 @[simp] theorem succ_ofNat_pos (n : Nat) : 0 < (n : Int) + 1 :=
-  lt_add_one_iff.2 (ofNat_zero_le _)
+  lt_add_one_iff.mpr (ofNat_zero_le _)
+
+theorem ofNat_not_neg (n : Nat) : ¬((n : Int) < 0) :=
```
jcommelin, line 591: Should this be called
```suggestion
theorem not_ofNat_neg (n : Nat) : ¬((n : Int) < 0) :=
```

src/Init/Data/Int/Order.lean
```diff
@@ -799,6 +804,12 @@ protected theorem lt_add_of_neg_lt_sub_right {a b c : Int} (h : -b < a - c) : c
 protected theorem neg_lt_sub_right_of_lt_add {a b c : Int} (h : c < a + b) : -b < a - c :=
   Int.lt_sub_left_of_add_lt (Int.sub_right_lt_of_lt_add h)
 
+protected theorem add_lt_iff (a b c : Int) : a + b < c ↔ a < -b + c := by
+  rw [← Int.add_lt_add_iff_left (-b), Int.add_comm (-b), Int.add_neg_cancel_right]
+
+protected theorem sub_lt_iff (a b c : Int) : a - b < c ↔ a < b + c :=
```
jcommelin, line 810: Here I would have expected
```suggestion
protected theorem sub_lt_iff (a b c : Int) : a - b < c ↔ a < c + b :=
```

jcommelin:
What is the small refactor of library_search? I only see the new tests.

semorrison:
> What is the small refactor of library_search? I only see the new tests.

I'll remove that commit from the PR, I think.

src/Init/Data/Int/Order.lean
semorrison: Sure. Done. (It's apparently unused, in any case.)

src/Init/Data/Int/Order.lean
semorrison: Easy enough to fix the proof, so sure.

There are way too many lemmas in this file, given that we now have `omega`, but that's a story for another day.

semorrison:
Mathlib is fixed on `lean-pr-testing-3850` (verified locally).

Issue #2164, opened 2023-03-23
Meaningful fields for `Subarray`


fgdorais:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

The current [`Subarray` structure](https://github.com/leanprover/lean4/blob/158d58f3c3a2a9b120a037acf8b3f0a0cd4ae573/src/Init/Data/Array/Subarray.lean#LL11-L16) has cryptic field names:
```lean
structure Subarray (α : Type u)  where
  as : Array α
  start : Nat
  stop : Nat
  h₁ : start ≤ stop
  h₂ : stop ≤ as.size
```
I propose the following alternative with more intuitive field names:
```lean
structure Subarray (α : Type u) extends Array α where
  start : Nat
  stop : Nat
  valid : start ≤ stop ∧ stop ≤ toArray.size
```
I have a potential PR candidate implementing this: <https://github.com/fgdorais/lean4/tree/subarray>

fgdorais:
An alternative with a simpler validity condition would be:
```lean
structure Subarray (α : Type u) extends Array α where
  start : Nat
  size  : Nat
  valid : start + size ≤ toArray.size

abbrev Subarray.stop (as : Subarray) : Nat := as.start + as.size
```

PR #3851, opened 2024-04-09
chore: rename fields of Subarray to follow Lean conventions

david-christiansen:


src/Init/Data/Array/Subarray.lean
```diff
@@ -9,25 +9,34 @@ import Init.Data.Array.Basic
...
+  stop_le_array_size : stop ≤ array.size
+
+@[deprecated]
+def Subarray.as (s : Subarray α) : Array α := s.array
+
+@[deprecated]
+def Subarray.h₁ (s : Subarray α) : s.start ≤ s.stop := s.start_le_stop
+
+@[deprecated]
+def Subarray.h₂ (s : Subarray α) : s.stop ≤ s.array.size := s.stop_le_array_size
```
eric-wieser, line 25:
```suggestion
@[deprecated Subarray.array]
abbrev Subarray.as (s : Subarray α) : Array α := s.array

@[deprecated Subarray.start_le_stop]
def Subarray.h₁ (s : Subarray α) : s.start ≤ s.stop := s.start_le_stop

@[deprecated Subarray.stop_le_array_size]
def Subarray.h₂ (s : Subarray α) : s.stop ≤ s.as.size := s.stop_le_array_size
```
so that the deprecation tells you what to use instead

david-christiansen: Thanks! I was trying to figure this one out.

eric-wieser (approved):

src/Init/Data/Array/Subarray.lean
```diff
@@ -9,25 +9,34 @@ import Init.Data.Array.Basic
...
 
 structure Subarray (α : Type u)  where
-  as : Array α
+  array : Array α
   start : Nat
   stop : Nat
-  h₁ : start ≤ stop
-  h₂ : stop ≤ as.size
+  start_le_stop : start ≤ stop
+  stop_le_array_size : stop ≤ array.size
```
digama0, line 16:
```suggestion
  stop_le_size : stop ≤ array.size
```
this still seems sufficiently precise and is less annoyingly long

david-christiansen:
OK, Mathlib builds against https://github.com/leanprover/std4/pull/733 and https://github.com/leanprover-community/aesop/pull/122 locally for me.

src/Init/Data/Array/Subarray.lean
digama0: Oh, yeah that is awkward. I wonder if we could change the name to avoid confusion (i.e. `Subarray.length`)? I think we will want names for both of these quantities eventually (e.g. in lemmas), so it would be good to think about what names are clear and distinct. `length` and `size`? `size` and `fullSize`? `extent` and `maxStop`? Or just use `t.array.size`.

As a pedantic note, prefix naming convention would yield `stop_le_size_array`.

david-christiansen: On reflection, I expect that this projection will not be used so often (mostly I expect users to be using the API here, or writing meaningful operators that preserve the invariants - borne out by the very small changes needed for Std and Aesop and the fact that Mathlib itself needed none), so I'll go with explicitness over concision.

Thanks!

fgdorais:
Closes #2164

PR #3853, opened 2024-04-09
fix: omega: ignore levels in canonicalizer

### LINKED ISSUE #3848: omega canonicalizer vs. universe expressions
At the time of writing, `omega` does not use full defEq when recognizing atoms, but since #3639 the aim is to consider atoms equal if they are definitionally equal, but differ only in *implicit* parameters.

Universe parameters are similarly implicit to the user, and cause great confusion in a proof like
```
theorem sizeOf_snd_lt_sizeOf_list
  {α : Type u} {β : Type v} [SizeOf α] [SizeOf β] {x : α × β} {xs : List (α × β)} :
  x ∈ xs → sizeOf x.snd < 1 + sizeOf xs
:= by
  intro h
  have h1 := List.sizeOf_lt_of_mem h
  have h2 : sizeOf x = 1 + sizeOf x.1 + sizeOf x.2 := rfl
  cases x; dsimp at *
  omega
```
where the user sees
```
omega did not find a contradiction:
[0, 0, 1] ∈ [0, ∞)
[0, 0, 0, 1] ∈ [0, ∞)
[1, -1, -1] ∈ [2, ∞)
[0, 1] ∈ [0, ∞)
[0, 0, 1, -1] ∈ [1, ∞)
[1] ∈ [0, ∞)
```
when they expect `omega` to solve it.

To debug the problem I had to come up with #3847 *and* turn on `set_option pp.universes true`, and now we see what goes wrong:
```
omega could not prove the goal:
a possible counterexample may satisfy the constraints
x₃ ∈ [0, ∞)
x₄ ∈ [0, ∞)
x₁ - x₂ - x₃ ∈ [2, ∞)
x₂ ∈ [0, ∞)
x₃ - x₄ ∈ [1, ∞)
x₁ ∈ [0, ∞)
where
x₁ := ↑(sizeOf.{(max u v) + 1} xs)
x₂ := ↑(sizeOf.{u + 1} fst✝)
x₃ := ↑(sizeOf.{v + 1} snd✝)
x₄ := ↑(sizeOf.{max (u + 1) (v + 1)} xs)
```

One can work around the problem by changing the lemma statement as such
```
theorem sizeOf_snd_lt_sizeOf_list
  {α : Type u} {β : Type v} [SizeOf α] [SizeOf β] {x : α × β} {xs : List (α × β)} :
  x ∈ xs → sizeOf x.snd < 1 + sizeOf.{(max u v)+1} xs
```
but arguably one shouldn't have to.

### Versions

0aa68312b643abce8f80be35274c306c3fc1601b

### Impact

nomeata:
fixes #3848

### DIFF
--- a/src/Lean/Meta/Canonicalizer.lean
+++ b/src/Lean/Meta/Canonicalizer.lean
@@ -84,8 +84,8 @@ private partial def mkKey (e : Expr) : CanonM Key := do
     let key ← match e with
       | .sort .. | .fvar .. | .bvar .. | .lit .. =>
         pure { e := (← shareCommon e) }
-      | .const n _us =>
-        pure { e := (← shareCommon (.const n [])) }
+      | .const n ls =>
+        pure { e := (← shareCommon (.const n (List.replicate ls.length levelZero))) }
       | .mvar .. =>
         -- We instantiate assigned metavariables because the
         -- pretty-printer also instantiates them.


src/Lean/Meta/Canonicalizer.lean
```diff
@@ -82,8 +82,10 @@ private partial def mkKey (e : Expr) : CanonM Key := do
     return key
   else
     let key ← match e with
-      | .sort .. | .fvar .. | .bvar .. | .const .. | .lit .. =>
+      | .sort .. | .fvar .. | .bvar .. | .lit .. =>
         pure { e := (← shareCommon e) }
+      | .const n us =>
+        pure { e := (← shareCommon (.const n (us.map .normalize))) }
```
nomeata, line 88: Actually, it's just a cache key, just drop the level 
```suggestion
        pure { e := (← shareCommon (.const n [])) }
```

src/Lean/Meta/Canonicalizer.lean
```diff
@@ -82,8 +82,10 @@ private partial def mkKey (e : Expr) : CanonM Key := do
     return key
   else
     let key ← match e with
-      | .sort .. | .fvar .. | .bvar .. | .const .. | .lit .. =>
+      | .sort .. | .fvar .. | .bvar .. | .lit .. =>
         pure { e := (← shareCommon e) }
+      | .const n us =>
```
nomeata, line 87:
```suggestion
      | .const n _us =>
```

PR #3857, opened 2024-04-10
doc: add docstring for Nat.gcd

kim-em:


src/Init/Data/Nat/Gcd.lean
```diff
@@ -10,6 +10,15 @@ import Init.RCases
...
 
+/--
+Computes the greatest common divisor of two natural numbers.
+
+This reference implementation via the Euclidean algorithm
+is overridden in both the kernel and the compiler to efficiently
+evaluate using the "bignum" representation (see `Nat`).
+The definition provided here is the logical model
+(and it is soundness-critical that they coincide).
+-/
```
digama0, line 21: Code examples would be useful, incl. edge cases.

digama0 (approved):

src/Init/Data/Nat/Gcd.lean
semorrison: Added, thanks.

PR #3860, opened 2024-04-10
doc: docstrings for Alternative

nomeata:


src/Init/Control/Basic.lean
```diff
@@ -20,8 +20,21 @@ def Functor.discard {f : Type u → Type v} {α : Type u} [Functor f] (x : f α)
 
 export Functor (discard)
 
+/--
+An `Alternative` functor is an `Applicative` functor that has a notion of “emptyness” or “failure”
+and a binary operation `<|>` with a notion of “collecting valures” or “left-most success”.
```
digama0, line 25:
```suggestion
and a binary operation `<|>` with a notion of “collecting values” or “left-most success”.
```

src/Init/Control/Basic.lean
```diff
@@ -20,8 +20,21 @@ def Functor.discard {f : Type u → Type v} {α : Type u} [Functor f] (x : f α)
...
+
+Important instances include
+* `Option`, where `failure := .none` and `<|>` returns the left-most `.some`.
+* Parser combinators typically provide an `Applicative` instance for error-handling and
+  backtracking.
```
digama0, line 30: As highlighted in [this thread](https://leanprover.zulipchat.com/#narrow/stream/239415-metaprogramming-.2F-tactics/topic/Tauto.20infinite.20loop/near/430695245), lean monads are somewhat inconsistent about whether `try catch` and/or `<|>` backtracks some or all of the state. Probably not much can be done on the docstring here other than to flag that the behavior varies and the user should consult the documentation for the relevant monad.

digama0 (approved):

src/Init/Control/Basic.lean
nomeata: How about
```suggestion
  backtracking; see the respective monad’s documentation for details.
```

nomeata: Next try:

```suggestion
  backtracking.
  
See the respective monad documentation for the precise semantics, e.g. which side effects 
are backtracked.
```

src/Init/Control/Basic.lean
```diff
@@ -20,8 +20,21 @@ def Functor.discard {f : Type u → Type v} {α : Type u} [Functor f] (x : f α)
 
 export Functor (discard)
 
+/--
+An `Alternative` functor is an `Applicative` functor that has a notion of “emptyness” or “failure”
```
david-christiansen, line 24: Spelling:
```suggestion
An `Alternative` functor is an `Applicative` functor that has a notion of “emptiness” or “failure”
```

But "has a notion" in my ears means "suddenly wants to". What about:
```suggestion
An `Alternative` functor is an `Applicative` functor that can "fail" or be "empty"
```

src/Init/Control/Basic.lean
```diff
@@ -20,8 +20,21 @@ def Functor.discard {f : Type u → Type v} {α : Type u} [Functor f] (x : f α)
...
+and a binary operation `<|>` with a notion of “collecting values” or “left-most success”.
```
david-christiansen, line 25:
```suggestion
and a binary operation `<|>` that “collects values” or finds the “left-most success”.
```

src/Init/Control/Basic.lean
david-christiansen: An `Alternative` is not necessarily a monad, and the docs are often not as informative as they could be on this point.

Another way to say this would be something like:

Error recovery and state can interact subtly. For example, the implementation of `Alternative` for `OptionT (StateT \sigma Id)` keeps modifications made to the state while recovering from failure, while`StateT \sigma (OptionT Id)` discards them.

src/Init/Control/Basic.lean
```diff
@@ -20,8 +20,21 @@ def Functor.discard {f : Type u → Type v} {α : Type u} [Functor f] (x : f α)
...
+
+Important instances include
+* `Option`, where `failure := .none` and `<|>` returns the left-most `.some`.
+* Parser combinators typically provide an `Applicative` instance for error-handling and
+  backtracking.
+-/
+-- NB: List instance is in mathlib. Once upstreamed, add
+-- * `List`, where `failure` is the empty list and `<|>` concatenates.
 class Alternative (f : Type u → Type v) extends Applicative f : Type (max (u+1) v) where
+  /-- See `Alternative`. -/
```
david-christiansen, line 35: I don't think this is the hover I'd like to see.

```suggestion
  /--
  Produces an empty collection or recoverable failure.  The `<|>` operator collects values or recovers
  from failures.
  -/
```

src/Init/Control/Basic.lean
```diff
@@ -20,8 +20,21 @@ def Functor.discard {f : Type u → Type v} {α : Type u} [Functor f] (x : f α)
...
   failure : {α : Type u} → f α
+  /-- See `Alternative`. Can be writen using `<|>`.  -/
```
david-christiansen, line 37:
```suggestion
  /--
  Depending on the `Alternative` instance, collects values or recovers from `failure`s by
  returning the leftmost success.
  -/
```

src/Init/Control/Basic.lean
```diff
@@ -30,9 +43,15 @@ variable {f : Type u → Type v} [Alternative f] {α : Type u}
 
 export Alternative (failure)
 
+/--
+If the proposition `p` holds, does nothing, else fails (using `failure`).
```
david-christiansen, line 47:
```suggestion
If the proposition `p` is true, does nothing, else fails (using `failure`).
```

src/Init/Control/Basic.lean
nomeata: Are you sure that there shouldn't be at least a mention of `Alternative` here, as the place to look for more details?

src/Init/Control/Basic.lean
nomeata: Did you intentionally drop the reference to the binary operator syntax?

src/Init/Control/Basic.lean
david-christiansen: Yes, there should be!

src/Init/Control/Basic.lean
david-christiansen: The dangers of commenting before coffee...  The syntax should definitely be there.

nomeata:
```suggestion
  /--
  Depending on the `Alternative` instance, collects values or recovers from `failure`s by
  returning the leftmost success. Can be written using the `<|>` operator syntax.
  -/
```

src/Init/Control/Basic.lean
nomeata: Next try:

```suggestion
  backtracking.
  
Error recovery and state can interact subtly. For example, the implementation of `Alternative` for `OptionT (StateT σ Id)` keeps modifications made to the state while recovering from failure, while `StateT σ (OptionT Id)` discards them.
```

src/Init/Control/Basic.lean
nomeata:
```suggestion
  /--
  Produces an empty collection or recoverable failure.  The `<|>` operator collects values or recovers
  from failures. See `Alternative` for more details.
  -/
```

PR #3863, opened 2024-04-10
doc: add docstring to add_decl_doc

kim-em:


src/Lean/Parser/Command.lean
```diff
@@ -277,6 +277,26 @@ def initializeKeyword := leading_parser
...
+
+Parent projections in structures are an example of this:
+```
+structure Triple (α β γ : Type) extends Prod α β where
+  thrd : γ
+
+/-- Extracts the first two projections of a triple. -/
+add_decl_doc Triple.toProd
+```
+-/
```
digama0, line 299:
```suggestion

Doc comments can only be added to declarations declared in the same module.
-/
```

digama0 (approved):

src/Lean/Parser/Command.lean
david-christiansen: What about:
```suggestion

Documentation can only be added to declarations in the same module.
-/
```

PR #3867, opened 2024-04-10
doc: docstrings for List.mapM and friends

nomeata:


src/Init/Data/List/Control.lean
```diff
@@ -40,24 +40,54 @@ Finally, we rarely use `mapM` with something that is not a `Monad`.
...
 
+/--
+Applies the applicative action `f` on every element in the list, left-to-right, and returns the list of
+results.
+
+NB: If `m` is also a `Monad`, then using `mapM` can be more efficient.
+
+See `List.forA` for the variant that discards the results.
+See `List.mapM` for the variant that works with `Monad`.
+-/
```
digama0, line 65:
```suggestion

**Warning**: this function is not tail-recursive, meaning that it may fail with stack overflow on long lists.
-/
```

digama0 (approved):

src/Init/Data/List/Control.lean
nomeata: Is this related to the
```
NB: If `m` is also a `Monad`, then using `mapM` can be more efficient.
```
comment or independent?

digama0: Independent, although I'm not sure it's even possible to write this in a tail-recursive way. It's much easier to get a `bind` chain to have the desired form for tail recursion than a `seq` chain.

nomeata: Thanks!

src/Init/Data/List/Control.lean
```diff
@@ -40,24 +40,56 @@ Finally, we rarely use `mapM` with something that is not a `Monad`.
...
+/--
+Applies the applicative action `f` on every element in the list, left-to-right, and returns the list of
+results.
+
+NB: If `m` is also a `Monad`, then using `mapM` can be more efficient.
+
+See `List.forA` for the variant that discards the results.
+See `List.mapM` for the variant that works with `Monad`.
+
+**Warning**: this function is not tail-recursive, meaning that it may fail with stack overflow on long lists.
```
david-christiansen, line 66:
```suggestion
**Warning**: this function is not tail-recursive, meaning that it may fail with a stack overflow on long lists.
```

src/Init/Data/List/Control.lean
```diff
@@ -71,15 +103,27 @@ def filterAuxM {m : Type → Type v} [Monad m] {α : Type} (f : α → m Bool) :
     let b ← f h
     filterAuxM f t (cond b (h :: acc) acc)
 
+/--
+Applies the monadic predicate `f` on every element in the list, left-to-right, and returns those
```
david-christiansen, line 107: Should we rename `f` to `p`?

src/Init/Data/List/Control.lean
```diff
@@ -71,15 +103,27 @@ def filterAuxM {m : Type → Type v} [Monad m] {α : Type} (f : α → m Bool) :
...
+elements `x` for which `f x` returns `true`.
+-/
 @[inline]
 def filterM {m : Type → Type v} [Monad m] {α : Type} (f : α → m Bool) (as : List α) : m (List α) := do
   let as ← filterAuxM f as []
   pure as.reverse
 
+/--
+Applies the monadic predicate `f` on every element in the list, right-to-left, and returns those
```
david-christiansen, line 116: Rename to `p`?

src/Init/Data/List/Control.lean
```diff
@@ -71,15 +103,27 @@ def filterAuxM {m : Type → Type v} [Monad m] {α : Type} (f : α → m Bool) :
...
+elements `x` for which `f x` returns `true`.
+-/
 @[inline]
 def filterRevM {m : Type → Type v} [Monad m] {α : Type} (f : α → m Bool) (as : List α) : m (List α) :=
   filterAuxM f as.reverse []
```
nomeata, line 121: Tip: If possible select all lines that may be affected by a change, then one can use the github suggestion feature to do it.

```suggestion
Applies the monadic predicate `p` on every element in the list, right-to-left, and returns those
elements `x` for which `p x` returns `true`.
-/
@[inline]
def filterRevM {m : Type → Type v} [Monad m] {α : Type} (p : α → m Bool) (as : List α) : m (List α) :=
  filterAuxM p as.reverse []
```

src/Init/Data/List/Control.lean
```diff
@@ -71,15 +103,27 @@ def filterAuxM {m : Type → Type v} [Monad m] {α : Type} (f : α → m Bool) :
...
     filterAuxM f t (cond b (h :: acc) acc)
 
+/--
+Applies the monadic predicate `f` on every element in the list, left-to-right, and returns those
+elements `x` for which `f x` returns `true`.
+-/
 @[inline]
 def filterM {m : Type → Type v} [Monad m] {α : Type} (f : α → m Bool) (as : List α) : m (List α) := do
   let as ← filterAuxM f as []
   pure as.reverse
```
nomeata, line 113:
```suggestion
/--
Applies the monadic predicate `p` on every element in the list, left-to-right, and returns those
elements `x` for which `p x` returns `true`.
-/
@[inline]
def filterM {m : Type → Type v} [Monad m] {α : Type} (p : α → m Bool) (as : List α) : m (List α) := do
  let as ← filterAuxM p as []
  pure as.reverse
```

david-christiansen (approved):

PR #3869, opened 2024-04-10
doc: Leo-Henrik retreat doc

hargoniX:
Part of the retreat Hackathon.

### DIFF
--- a/src/Init/WF.lean
+++ b/src/Init/WF.lean
@@ -48,6 +48,7 @@ If a relation is `WellFounded`, it does not allow for an infinite descent along 
 
 If the arguments of the recursive calls in a function definition decrease according to
 a well founded relation, then the function terminates.
+Well-founded relations are sometimes called _Artinian_ or said to satisfy the “descending chain condition”.
 -/
 inductive WellFounded {α : Sort u} (r : α → α → Prop) : Prop where
   | intro (h : ∀ a, Acc r a) : WellFounded r


src/Lean/Meta/Basic.lean
```diff
@@ -301,6 +301,44 @@ structure Context where
...
+- A `Cache` for chaching results of the key `MetaM` operations.
+
+The key operations provided by `MetaM` are:
+- `inferType`, it attempts to automatically infer the type of a given expression.
+- `whnf`, it reduces an expression to the point where the outermost part is no longer reducible
+  but the inside may still contain unreduced expression.
+- `isDefEq`, it determines whether two expressions are definitionally equal, possibly assigning
+  meta variables in the process.
+- `forallTelescope` and `lambdaTelescope`, they make it possible to automatically move into
+  (nested) binders while updating the local context.
```
digama0, line 320:
```suggestion
- `inferType`, which attempts to automatically infer the type of a given expression.
- `whnf`, which reduces an expression to the point where the outermost part is no longer reducible
  but the inside may still contain unreduced expression.
- `isDefEq`, which determines whether two expressions are definitionally equal, possibly assigning
  meta variables in the process.
- `forallTelescope` and `lambdaTelescope`, which make it possible to automatically move into
  (nested) binders while updating the local context.
```

src/Init/Data/Repr.lean
```diff
@@ -13,11 +13,24 @@ open Sum Subtype Nat
 
 open Std
 
+/--
+A typeclass that specifies the standard way of turning values of some type into `Format`.
+
+When rendered this `Format` should be as close as possible to something that can be parsed as the
```
david-christiansen, line 19: "As close as possible" is a bit tricky :-) Is it a bug if it doesn't round-trip?

src/Init/Data/Repr.lean
```diff
@@ -13,11 +13,24 @@ open Sum Subtype Nat
...
+input value.
+-/
 class Repr (α : Type u) where
+  /--
+  Turn a value of type `α` into `Format` at a given precedence. The precedence value can be used
+  to avoid parenthesis if they are not necessary.
```
david-christiansen, line 25:
```suggestion
  to avoid parentheses if they are not necessary.
```
Where can the user go to find out what the reasonable values of the `Nat` are, and how they use this `Nat` in practice to add parentheses? It might also be good to refer to useful helpers here, like `addAppParen`.

src/Init/WF.lean
```diff
@@ -9,7 +9,18 @@ import Init.Data.Nat.Basic
...
+`Acc` is the accessibility predicate. Given some relation `r` (e.g. `<`) and a value `x`,
+`Acc r x` means that `x` is accessible through `r`:
+
+`x` is accessible if there exists no infinite sequence `... < y₂ < y₁ < y₀ < x`.
+-/
 inductive Acc {α : Sort u} (r : α → α → Prop) : α → Prop where
+  /--
+  A value is accessible if for all `y` such that `r y x`, `y` is also accessible.
+  Note that if there exists no `y` such that `r y x`, `x` is accessible, such an `x` is called a
+  base case.
```
david-christiansen, line 22:
```suggestion
  Note that if there exists no `y` such that `r y x`, then `x` is accessible. Such an `x` is called a
  _base case_.
```

src/Init/Data/Repr.lean
digama0: I don't think this Nat is used very much, it's nothing like the global precedence system used with `notation` et al. Most of the uses are very local and I've only ever seen 1 or 2 different precedences used.

src/Init/WF.lean
```diff
@@ -31,6 +42,13 @@ def inv {x y : α} (h₁ : Acc r x) (h₂ : r y x) : Acc r y :=
 
 end Acc
 
+/--
+A relation `r` is `WellFounded` if all elements of `α` are accessible within `r`.
+If a relation is `WellFounded`, it does not allow for an infinite descent along the relation.
+
+This is used to prove that a recursive function, where the arguments of the recursive calls
+decrease according to a well founded relation, terminates.
```
david-christiansen, line 50:
```suggestion
If the arguments of the recursive calls in a function definition decrease according to
a well founded relation, then the function terminates. 
```

src/Lean/Meta/Basic.lean
```diff
@@ -301,6 +301,44 @@ structure Context where
     Note that we do not cache results at `whnf` when `canUnfold?` is not `none`. -/
   canUnfold?        : Option (Config → ConstantInfo → CoreM Bool) := none
 
+/--
+The `MetaM` monad is a core component of Lean's metaprogramming framework, facilitating the
+construction and manipulation, of expressions (`Lean.Expr`) within Lean.
```
david-christiansen, line 306:
```suggestion
construction and manipulation of expressions (`Lean.Expr`) within Lean.
```

src/Lean/Meta/Basic.lean
```diff
@@ -301,6 +301,44 @@ structure Context where
...
+
+It builds on top of `CoreM` and additionally provides:
+- A `LocalContext` for managing free variables.
+- A `MetavarContext` for managing meta variables.
```
david-christiansen, line 310:
```suggestion
- A `MetavarContext` for managing metavariables.
```

src/Lean/Meta/Basic.lean
```diff
@@ -301,6 +301,44 @@ structure Context where
...
+- A `Cache` for chaching results of the key `MetaM` operations.
```
david-christiansen, line 311:
```suggestion
- A `Cache` for caching results of the key `MetaM` operations.
```

src/Init/Data/Repr.lean
hargoniX: Leo mentioned that it is "the moral obligation" but it is not strictly considered a bug.

src/Init/WF.lean
```diff
@@ -31,6 +42,13 @@ def inv {x y : α} (h₁ : Acc r x) (h₂ : r y x) : Acc r y :=
 
 end Acc
 
+/--
+A relation `r` is `WellFounded` if all elements of `α` are accessible within `r`.
+If a relation is `WellFounded`, it does not allow for an infinite descent along the relation.
+
+If the arguments of the recursive calls in a function definition decrease according to
+a well founded relation, then the function terminates.
+-/
```
nomeata, line 51:
```suggestion

A well-founded relation is also called Artinian, and has the “descending chain condition”.
-/
```

src/Init/Data/Repr.lean
david-christiansen: OK, this formulation is good then :)

src/Init/WF.lean
david-christiansen: Which users would be helped by this addition? It seems to me to be a side remark, but I also don't know the whole audience intimately.

digama0: It's helpful to use names from the literature in docstrings so that if someone searches for "descending chain condition" they will find this definition. There is no need to draw attention to these words, they are basically just keywords for SEO.

david-christiansen: There's always a balance between the SEO aspect and the sheer intimidation of having lots of big words that few people understand. I'm not sure where that balance is here, for half our target audience; I'm inclined to trust you if you think the benefits outweigh the costs in this case.

david-christiansen: I've got nothing against highfalutin' book-talk when it's helpful - but there is a kind of self-parody effect that it'd be good to avoid. Like I said - I'm inclined to trust you if you think this balance works out here for the mathematician audience.

If we do want:

> additional wording around it to make it clear that if you don't know what it's talking about you can safely ignore the remark

then we could say

```suggestion
Well-founded relations are sometimes called _Artinian_ or said to satisfy the “descending chain condition”.
-/
```

PR #3870, opened 2024-04-10
doc: describe all `simp` configuration options

kmill:
Co-authored by Marc Huisinga, with input from Leo.

### DIFF
--- a/src/Init/MetaTypes.lean
+++ b/src/Init/MetaTypes.lean
@@ -77,7 +77,7 @@ See also `Lean.Meta.Simp.neutralConfig`.
 structure Config where
   /--
   The maximum number of subexpressions to visit when performing simplification.
-  The default is large number, but not infinite.
+  The default is 100000.
   -/
   maxSteps          : Nat  := defaultMaxSteps
   /--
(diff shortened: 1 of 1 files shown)


src/Init/MetaTypes.lean
```diff
@@ -68,45 +68,115 @@ namespace Simp
...
+/--
+The configuration for `simp`.
+Passed to `simp` using, for example, the `simp (config := {contextual := true})` syntax.
+
+See also `Lean.Meta.Simp.neutralConfig`.
+-/
 structure Config where
+  /--
+  The maximum number of subexpressions to visit when performing simplification.
+  The default is large number, but not infinite.
```
digama0, line 80:
```suggestion
  The default is 100000.
```
Giving the actual number helps with giving people a sense of e.g. it's order of magnitude.

digama0:
It will probably be a lot more work than what you have already done here (thank you all for this BTW!), but it would also be useful to have examples where a user might plausibly want to turn on each option and demonstrate the difference, because the specification of the flags can be somewhat opaque if you don't know simp / type theory internals.

digama0 (approved):

david-christiansen:
Yes, the examples definitely needs to be in the new reference! In the meantime, I think it's useful to have them here "below the fold", but it's OK not to as well.

david-christiansen (approved):
This is a very nicely written set of docs!

PR #3871, opened 2024-04-10
doc: Classical.choice

Kha:


digama0 (approved):

src/Init/Classical.lean
```diff
@@ -15,6 +15,11 @@ namespace Classical
 noncomputable def indefiniteDescription {α : Sort u} (p : α → Prop) (h : ∃ x, p x) : {x // p x} :=
   choice <| let ⟨x, px⟩ := h; ⟨⟨x, px⟩⟩
 
+/--
+Given that there exists an element satisfying `p`, returns one such element.
```
david-christiansen, line 19: I stumbled a bit over "returns" here, which sounds very much like computation to me! I don't have a suggestion for a better word off the top of my head, though.

fgdorais: denotes? selects?

khoek: chooses? :)

nomeata: is logically equivalent to an arbitrary such element

david-christiansen: I like the last two suggestions a lot here

david-christiansen: What about:
```suggestion
Given that there exists an element of `α` that satisfies `p`, the result of `choose` is logically equivalent to an
arbitrary such element.
```

This makes the non-computational nature of it very apparent.  We're also out in ultra-nitpick land here :)

digama0: I'm not sure "the result of `choose` is logically equivalent to an arbitrary such element" is a well formed thing to say. I think I know the theorem it's alluding to but this is a really misleading way to say it. I think you might want to get some mathematician eyes on this

david-christiansen: Better to have the slight odor of computation than to write something that's incorrect :) I think that the discussion on this has now exceeded the very minor downsides of the word "returns", and am supportive of a merge as-is

fgdorais: Using "logically equivalent" that way is sketchy and doesn't quite make sense for the technical meaning of logical equivalence. I like: "Given that there exists an element of `α` satisfying `p`, `choose p` denotes an otherwise arbitrary such element."

semorrison:
Let's stick a fork in this one and call it done.

PR #3876, opened 2024-04-10
perf: faster Nat.repr implementation in C

nomeata:
`Nat.repr` was implemented by generating a list of `Chars`, each created by a 10-way if-then-else. This can cause significant slow down in some particular use cases.

Now `Nat.repr` is `implemented_by` a faster implementation that uses C++’s `std::to_string` on small numbers (< USize.size) and maintains an array of pre-allocated strings for the first 128 numbers.

The handling of big numbers (≥ USize.size) remains as before.

### DIFF
--- a/src/Init/Data/Repr.lean
+++ b/src/Init/Data/Repr.lean
@@ -103,16 +103,34 @@ instance {p : α → Prop} [Repr α] : Repr (Subtype p) where
 
 namespace Nat
 
-def digitChars : Array Char :=
-  -- Array syntax not available yet
-  Array.mk ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f']
+/-
+We have pure functions for calculating the decimal representation of a `Nat` (`toDigits`), but also
+a fast variant that handles small numbers (`USize`) via C code (`lean_string_of_usize`).
+-/
 
-def digitChar (n : Nat) : Char := digitChars.getD n '*'
+def digitChar (n : Nat) : Char :=
+  if n = 0 then '0' else
+  if n = 1 then '1' else
+  if n = 2 then '2' else
+  if n = 3 then '3' else
+  if n = 4 then '4' else
+  if n = 5 then '5' else
+  if n = 6 then '6' else
+  if n = 7 then '7' else
+  if n = 8 then '8' else
+  if n = 9 then '9' else
+  if n = 0xa then 'a' else
+  if n = 0xb then 'b' else
+  if n = 0xc then 'c' else
+  if n = 0xd then 'd' else
+  if n = 0xe then 'e' else
+  if n = 0xf then 'f' else
+  '*'
 
 def toDigitsCore (base : Nat) : Nat → Nat → List Char → List Char
   | 0,      _, ds => ds
   | fuel+1, n, ds =>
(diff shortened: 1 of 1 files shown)


nomeata:
!bench

src/Init/Data/Repr.lean
```diff
@@ -103,29 +103,16 @@ instance {p : α → Prop} [Repr α] : Repr (Subtype p) where
...
-  if n = 0xa then 'a' else
-  if n = 0xb then 'b' else
-  if n = 0xc then 'c' else
-  if n = 0xd then 'd' else
-  if n = 0xe then 'e' else
-  if n = 0xf then 'f' else
-  '*'
+def digitChars : Array Char :=
+  -- Array syntax not available yet
+  Array.mk ['0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f']
```
digama0, line 108: I would assume that doing character arithmetic is faster than this. (And if omega is available here then it should also be able to handle the subgoals.)

fgdorais: How much of this is in user space? Nobody purposefully uses bases 7, 11, or 13 except in demo code. If this is not intended for general use and performance is an issue, then it makes sense to further specialize to the regularly used bases (2, 8, 10, 16). ([Base 64 encoding](https://en.wikipedia.org/wiki/Base64) is a totally different ballpark.)

nomeata: All good points. You could also allocate a ByteArray the right size, write the bytes mutably, and turn it into a string, to avoid going through list.

nomeata: I just learned from @hargoniX  that `Nat.toString` shows up significantly in the profile for writing out DIMACS file, so now there is a good reason for me to nerd around here.

I plan to add a small microbenchmark for `Nat.toString`, and then see how well it works to calculate number of digits, allocate a bytearray, fill it linearly, convert to string. Maybe also maintain a static array for the string representation of short numbers (up to 128 or something like that), so that we can never allocate for these.

nomeata: Although this is all not so easy due to bootstrapping issues: `Nat.repr` is a dependency of the module that `ByteArray` is defined in. And there is also no function yet for “initialize zero-filled ByteArray of size n”. Not so low-hanging after all.

nomeata:
!bench

nomeata:
!bench

nomeata:
!bench

nomeata:
!bench

nomeata:
!bench

nomeata:
!bench

nomeata:
!bench

nomeata:
Impact on mathlib: http://speed.lean-fro.org/mathlib4/compare/a93fa055-e732-498b-8c1f-fd5ab4f48b05/to/4457e6c3-ba82-4273-b390-de8345e68c5f

hargoniX (approved):
LGTM, the only thing I'm wondering about is the stddev on our metrics for the `nat_repr` benchmark, how reliable are they?

nomeata:
Don't know. Is that something we should look at now, or when it seems to become a problem?

hargoniX:
The latter I think, it would only be an issue if we end up improving this even more.

PR #3877, opened 2024-04-11
feat: binary recursive implementation of List.mapA

digama0:
Inspired by https://github.com/leanprover/lean4/pull/3867#discussion_r1559543688 . After playing with this function a bit more, I was able to confirm that it's not really possible to implement it tail recursively in most monads (note that tail-recursiveness depends on the monad itself), because `seq` is not tail recursive in most monads and you have to stack up at least `n-1` of them to reduce a list of length `n`. However, we can do the next best thing which is to use a balanced tree of `seq` applications. I have confirmed that this will evaluate
```lean
#eval List.mapA (m := StateT Nat Id) pure (List.range 10000) |>.run 1
```
without stack overflow, unlike the original implementation, but it still deserves a `!bench` because the binary reduction has overhead and it may be worthwhile to switch over to naive recursion below some threshold.

### DIFF
--- a/src/Init/Data/List/Control.lean
+++ b/src/Init/Data/List/Control.lean
@@ -6,6 +6,7 @@ Author: Leonardo de Moura
 prelude
 import Init.Control.Basic
 import Init.Data.List.Basic
+import Init.Data.Nat.Log2
 
 namespace List
 universe u v w u₁ u₂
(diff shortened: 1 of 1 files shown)


david-christiansen:
!bench

digama0:
(note: the benchmark is useless because `List.mapA` is unused in core)

semorrison:
Could you add a doc-string explaining at least that the complicated implementation is for performance, and ideally explaining why tail-recursiveness is impossible and that this is "the next best thing"?

srghma:
@digama0 
> not really possible to implement it tail recursively in most monads

what are these monads? maybe for these monads possible to create

```lean
class CanImplementTailRecursivenessFor m where
  proof_x : ...
```

and then write in docstring for `List.mapA` - "hey, if You want List.mapA to work faster for Your monad then proove/add class CanImplementTailRecursivenessFor for it"

and if monad has `CanImplementTailRecursivenessFor` then we enable optimization using `@[csimp]`

kim-em:

PR #3881, opened 2024-04-11
perf: fix linearity issue in insertIfNew

hargoniX:
This fixes a linearity isssue in `insertIfNew`. As `insertIfNew` is used in `Lean.finalizeImport` we expect this to improve performance.


hargoniX:
!bench

PR #3883, opened 2024-04-12
feat: whitespace and message ordering configurations for `#guard_msgs`

kmill:
Adds options to control whitespace normalization and message ordering in `#guard_msgs`.

Examples:
1. `#guard_msgs (whitespace := lax)` ignores differences in whitespace completely.
2. `#guard_msgs (whitespace := exact)` requires an exact match for whitespace (after trimming).
3. `#guard_msgs (ordering := sorted)` sorts the list of messages, to make it insensitive to message order.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -61,6 +61,11 @@ v4.8.0 (development in progress)
   to enable pretty printing function applications using generalized field notation.
   Field notation can now be disabled function-by-function using the `@[pp_nodot]` attribute.
 
+* The `#guard_msgs` command now has options to change whitespace normalization and sensitivity to message ordering.
+  For example, `#guard_msgs (whitespace := lax) in cmd` collapses whitespace before checking messages,
+  and `#guard_msgs (ordering := sorted) in cmd` sorts the messages in lexicographic order before checking.
+  PR [#3883](https://github.com/leanprover/lean4/pull/3883).
+
 Breaking changes:
 
 * Automatically generated equational theorems are now named using suffix `.eq_<idx>` instead of `._eq_<idx>`, and `.def` instead of `._unfold`. Example:


src/Init/Notation.lean
```diff
@@ -552,10 +552,13 @@ except that it doesn't print an empty diagnostic.
 -/
 syntax (name := runMeta) "run_meta " doSeq : command
 
-/-- Element that can be part of a `#guard_msgs` specification. -/
-syntax guardMsgsSpecElt := &"drop"? (&"info" <|> &"warning" <|> &"error" <|> &"all")
+/-- Element of a `#guard_msgs` specification. -/
+syntax guardMsgsSpecElt :=
+  (&"drop"? (&"info" <|> &"warning" <|> &"error" <|> &"all"))
+  <|> (&"whitespace" " := " (&"exact" <|> &"normalized" <|> &"lax"))
+  <|> (&"ordering" " := " (&"exact" <|> &"sorted"))
```
digama0, line 559: I think this is a bit too complex to be a single `syntax`, this kind of thing is really painful to match against because none of the subexpressions are named. I think it would be better to have the three <|> alternatives have their own name, as well as the arguments to `whitespace` and `ordering`. Moreover, I suspect that this is related to the issue you have with the syntax pattern not matching.

digama0 (approved):

digama0:
The code action also needs to be modified to take the `ordering := sorted` setting into account, or else it will suggest a docstring that will fail the command.

src/Init/Notation.lean
kmill: I ended up not splitting out the `syntax` for the arguments to `whitespace` and `ordering`. I almost did, to be able to split out the parsing functions, but I think there's about the same or even less syntax parsing code this way.

It would be nice to get to the bottom of that syntax match issue though. In principle it should work.

kmill:
@digama0 It's now about ready to come out of draft state. If you approve, then I'll go and create a PR for the first commit ([bc75d48](https://github.com/leanprover/lean4/pull/3883/commits/bc75d483207fc3094c939c2a021170a7d6239202)), then update stage0, and then this PR will just be for the third commit.

It turns out the code action didn't need special handling since the sorted version is passed into the info tree. I tested it by using it to create the docstrings in the test file.

src/Init/Notation.lean
digama0: Well I'm thinking not just about the code for this function itself but also downstream code that is manipulating these syntaxes, and which may not need to match on every part of it.

kmill: Do you expect any downstream code to match on these syntaxes? Happy to do it if there's a plausible application.

The only application I can think of is some other command that generates `#guard_msgs` commands with dynamic configurations, but I can't think of what that command would be for. I can only imagine a `macro` with a fixed configuration.

src/Lean/Elab/GuardMsgs.lean
```diff
@@ -40,31 +40,55 @@ inductive SpecResult
...
-      return p
-    | _ => throwErrorAt spec "Invalid #guard_msgs specification"
-  else
-    return fun _ => .check
+    CommandElabM (WhitespaceMode × MessageOrdering × (Message → SpecResult)) := do
+  let elts ←
+    if let some spec := spec? then
+      match spec with
+      | `(guardMsgsSpec| ($[$elts:guardMsgsSpecElt],*)) => pure elts
+      | _ => throwErrorAt spec "Invalid #guard_msgs specification"
```
digama0, line 68:
```suggestion
      | _ => throwUnsupportedSyntax
```
When syntax patterns don't match, it's usually not that helpful to throw errors about it, since it's generally the result of a parse error which produced a partial syntax tree. You can just use `throwUnsupportedSyntax` to kick it back to the main system in this case.

src/Init/Notation.lean
digama0: Mathport does this kind of thing frequently, for all sorts of commands and syntaxes. Reformatting tools and migration tools will be reading and writing most command syntaxes, and I often had to work around upstream code that overused `("a" <|> "b" <|> "c") "+" ("x" <|> "y" <|> "z")` which is really painful to match.

kmill:
@digama0 The staging business is all worked out now in this PR, and it's set to trigger a stage0 update after it merges.

digama0:
Okay, LGTM. Let's see if merging works...

Kha:
> #guard_msgs (ordering := sorted) sorts the list of messages, to make it insensitive to message order.

Should this be considered a workaround for #2335?

kmill:
@Kha Yes, though it goes a little further since, for example, a tactic might spawn multiple threads and report multiple messages nondeterministically at the same source position.

PR #3893, opened 2024-04-12
chore: use `Ordering.then` in `deriving Ord`

digama0:
This should improve the performance of the deriving a bit since it doesn't have to generate so many matchers. The main motivation though is to make it easier to prove properties about the expression by using more standard functions. The generated implementation should end up the same, since `Ordering.then` is `@[macro_inline]`.


digama0:
!bench

nomeata (approved):

PR #3896, opened 2024-04-13
feat: hover / go-to-def for `attribute` cmd

digama0:
`attribute [attr] foo` was missing a hover on `foo`.

### DIFF
--- a/src/Lean/Elab/Declaration.lean
+++ b/src/Lean/Elab/Declaration.lean
@@ -353,14 +353,14 @@ def elabMutual : CommandElab := fun stx => do
     for builtin simprocs.
     -/
     let declNames ←
-       try
-         realizeGlobalConst ident
-       catch _ =>
-         let name := ident.getId.eraseMacroScopes
-         if (← Simp.isBuiltinSimproc name) then
-           pure [name]
-         else
-           throwUnknownConstant name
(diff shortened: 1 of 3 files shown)


src/Lean/Elab/Declaration.lean
```diff
@@ -353,15 +353,21 @@ def elabMutual : CommandElab := fun stx => do
...
-       try
-         realizeGlobalConst ident
-       catch _ =>
-         let name := ident.getId.eraseMacroScopes
-         if (← Simp.isBuiltinSimproc name) then
-           pure [name]
-         else
-           throwUnknownConstant name
-    let declName ← ensureNonAmbiguous ident declNames
+      try some <$> realizeGlobalConst ident
```
nomeata, line 356: Don't we have `realizeGlobalConstWithInfo` or the like?

digama0: The FIXME in this code is because we are using `realizeGlobalConst` even though sometimes it's not a valid declaration but a simproc instead. So some of the code from these functions is inlined here to ensure the error handling is done right.

I think it would be possible to use `realizeGlobalConstWithInfos` here and only do the ambiguity handling separately, although it has slightly different error behavior.

PR #3899, opened 2024-04-13
fix: throwError when name action does not define requested name

nomeata:
(no test, it’s tedious with our test suite because it would need two files, and really shouldn’t be user-visible anyways.)


nomeata:
Looks like the current behavior is roughly intended, e.g. with `foo.eq_2` where `foo` has less than 2 equations it will hit this code path. Maybe not worth doing anything here for now.

PR #3903, opened 2024-04-14
feat: make frontend normalize line endings to LF

kmill:
To eliminate parsing differences between Windows and other platforms, the frontend now normalizes all CRLF line endings to LF, like [in Rust](https://github.com/rust-lang/rust/issues/62865).

Effects:
- This makes Lake hashes be faithful to what Lean sees (Lake already normalizes line endings before computing hashes).
- Docstrings now have normalized line endings. In particular, this fixes `#guard_msgs` failing multiline tests for Windows users using CRLF.
- Now strings don't have different lengths depending on the platform. Before this PR, the following theorem is true for LF and false for CRLF files.
```lean
example : "
".length = 1 := rfl
```

Note: the normalization will take `\r\r\n` and turn it into `\r\n`. In the elaborator, we reject loose `\r`'s that appear in whitespace. Rust instead takes the approach of making the normalization routine fail. They do this so that there's no downstream confusion about any `\r\n` that appears.

Implementation note: the LSP maintains its own copy of a source file that it updates when edit operations are applied. We are assuming that edit operations never split or join CRLFs. If this assumption is not correct, then the LSP copy of a source file can become slightly out of sync. If this is an issue, there is some discussion [here](https://github.com/leanprover/lean4/pull/3903#discussion_r1592930085).


src/Init/Data/String/Extra.lean
```diff
@@ -98,4 +98,32 @@ def removeLeadingSpaces (s : String) : String :=
   let n := findLeadingSpacesSize s
   if n == 0 then s else removeNumLeadingSpaces n s
 
+/--
+Replaces each `\r\n` with `\n` to normalize line endings,
+but does not validate that there are no isolated `\r` characters.
+It is an optimized version of `String.replace text "\r\n" "\n"`.
+-/
+def crlfToLf (text : String) : String :=
```
mhuisi, line 106: Idle thought: Whenever I read code like this with indexing proofs, I wonder whether we need better mechanisms to separate these kinds of proofs from the code they are used in.

src/Init/Data/String/Extra.lean
```diff
@@ -98,4 +98,32 @@ def removeLeadingSpaces (s : String) : String :=
   let n := findLeadingSpacesSize s
   if n == 0 then s else removeNumLeadingSpaces n s
 
+/--
```
mhuisi, line 101: Some tests for `crlfToLf` would be nice

digama0: Note that this is not a new function, it is moving from lake, where it has been involved in hash calculations for some time. So we have some evidence that this works to normalize e.g. the files in mathlib when checking them out on windows / linux, because otherwise `lake exe cache` would break.

src/Lean/Server/Utils.lean
```diff
@@ -125,7 +125,7 @@ def applyDocumentChange (oldText : FileMap) : (change : Lsp.TextDocumentContentC
   | TextDocumentContentChangeEvent.rangeChange (range : Range) (newText : String) =>
     replaceLspRange oldText range newText
   | TextDocumentContentChangeEvent.fullChange (newText : String) =>
-    newText.toFileMap
+    newText.crlfToLf.toFileMap
```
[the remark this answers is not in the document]
kmill: The `replaceLspRange` case made me decide not to incorporate `crlfToLf` into `toFileMap`. We only want to normalize the line endings of the new text, and it's worth making it explicit where we're normalizing.

There's still something unsatisfactory however, which is that if somehow a `\r\n` is split and re-merged by edit operations, then we can get into a situation where there's an unnormalized `\r\n`. I'm not sure if this is an issue that could crop up in practice.

If we're not happy with leaving it in the current form, here are some solutions that came to mind:
1. Have a second FileMap with the unnormalized text. Make sure LSP operations edit that, and then re-normalize it to create the actual FileMap used by elaboration each time.
2. Do that, but accelerate it somehow. There's probably some fancy data structure supporting inserting-and-normalizing.
3. Make the LSP throw an error when it detects a loose `\r`. It's not clear how to get this information back to the user in a nice way.

src/Init/Data/String/Extra.lean
kmill: It's possible to separate out proofs with `decreasing_by`:
```lean
def crlfToLf (text : String) : String :=
  go "" 0 0
where
  go (acc : String) (accStop pos : String.Pos) : String :=
    if h : text.atEnd pos then
      -- note: if accStop = 0 then acc is empty
      if accStop = 0 then text else acc ++ text.extract accStop pos
    else
      let c := text.get' pos h
      let pos' := text.next' pos h
      if h' : ¬ text.atEnd pos' ∧ c == '\r' ∧ text.get pos' == '\n' then
        let acc := acc ++ text.extract accStop pos
        go acc pos' (text.next' pos' h'.1)
      else
        go acc accStop pos'
  termination_by text.utf8ByteSize - pos.byteIdx
  decreasing_by
    decreasing_with
      have k := Nat.gt_of_not_le <| mt decide_eq_true h
      exact Nat.sub_lt_sub_left k (Nat.lt_trans (String.lt_next text pos) (String.lt_next _ _))
    decreasing_with
      have k := Nat.gt_of_not_le <| mt decide_eq_true h
      exact Nat.sub_lt_sub_left k (String.lt_next _ _)
      decreasing_tactic
```

src/Init/Data/String/Extra.lean
kmill: Still, I added some tests to make sure that it does what we think it does in some corner cases.

tydeu:
Now that we have position termination proofs in core (e.g., `termination_by text.utf8ByteSize - pos.byteIdx`), it may be worth supporting these automatically through `decreasing_trivial`. For example, I did this over in my Partax library ([code](https://github.com/tydeu/lean4-partax/blob/master/Partax/String.lean)).

src/lake/Lake/Util/Newline.lean
```diff
@@ -0,0 +1,10 @@
+/-
+Copyright (c) 2023 Mario Carneiro. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Mario Carneiro
+-/
+
+namespace Lake
+
+/-- This is the same as `String.replace text "\r\n" "\n"`, but more efficient. -/
+@[inline, deprecated] partial def crlf2lf (text : String) : String := text.crlfToLf
```
tydeu, line 10: FYI, you could instead remove the `Lake.Util.Newline` import from `src/lake/tests/toml/Test.lean`.

Kha (approved):

src/Lean/Server/Utils.lean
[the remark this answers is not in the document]
mhuisi: The current state seems fine. I'm more worried about accidentally forgetting `crlfToLf` somewhere in the future, but I agree that this is not that important and explicitly seeing the normalization in the code is helpful, too.

mhuisi (approved):

PR #3908, opened 2024-04-14
doc: update `norm_cast` and `push_cast` documentation

kmill:


kmill:
@david-christiansen This could be improved more still, but my primary goal was just "let's make it not wrong" and do some incremental improvements on top of that.

An obvious question for future improvements is answering "what's a cast exactly?" I don't know the answer, beyond "it's a coercion that these tactics have any cast lemmas about".

src/Init/Tactics.lean
```diff
@@ -1125,57 +1125,75 @@ normalizes `h` with `norm_cast` and tries to use that to close the goal. -/
...
-`push_cast` rewrites the goal to move casts inward, toward the leaf nodes.
+`push_cast` rewrites the goal to move certain coercions (*casts*) inward, toward the leaf nodes.
 This uses `norm_cast` lemmas in the forward direction.
 For example, `↑(a + b)` will be written to `↑a + ↑b`.
-It is equivalent to `simp only with push_cast`.
-It can also be used at hypotheses with `push_cast at h`
-and with extra simp lemmas with `push_cast [int.add_zero]`.
+- `push_cast` moves casts inward in the goal.
+- `push_cast at h` moves casts inward in the hypothesis `h`.
+It can be used with extra simp lemmas with, for example, `push_cast [int.add_zero]`.
```
digama0, line 1171:
```suggestion
It can be used with extra simp lemmas with, for example, `push_cast [Int.add_zero]`.
```

semorrison:
Better than what we had, so let's merge.

PR #3910, opened 2024-04-15
feat: display auto-implicits as inlay hints

Kha:


Kha:
image

Prototype developed with @nomeata at FRO retreat

TODO:
- [ ] test usefulness/annoyance factor
- [ ] test reactivity, especially during editing in long files
- [ ] support more than `def`
- [ ] fix `def` missing syntactic binders
- [ ] review, refine, and move introduced types, copy over missing fields and docs from LSP spec

digama0:
> do we need a specific opt-out in the vscode-lean4 or is the generic one sufficient?

It definitely needs an opt out (I personally dislike inlay hints because of the effect they have on the cursor), but I think it should be possible to opt out only of certain kinds of inlay hints (assuming we get more in the future)

mhuisi:
Closing as this PR was superseded by #6768.

PR #3911, opened 2024-04-15
chore: remove @ from rw? suggestions, and enable hover on constants in #check

kim-em:
* Replaces the unused `Lean.PrettyPrinter.ppConst` with
`MessageData.ofConst` (which similarly avoids an unnecessary `@`) and that further generates a hover for the constant

* Uses this in `TryThis.addRewriteSuggestion`, so that `rw?` suggestions don't have unnecessary `@`s.

* Add `MessageData.signature`, as a wrapper around `PrettyPrinter.signature`, using the same machinery to generate hovers for constants, improving the hover behaviour in #check so that we get second order pop-up for constants in the signature. (Not sure how to write tests for second order hovers, so there is no test for this.)

### DIFF
--- a/src/Lean/PrettyPrinter.lean
+++ b/src/Lean/PrettyPrinter.lean
@@ -109,11 +109,11 @@ Uses the `pp.tagAppFns` option to annotate constants with terminfo, which is nec
 -/
 def ofFormatWithInfos
     (fmt : MetaM FormatWithInfos)
-    (noContext : Thunk Format := "<no context, could not generate MessageData>") : MessageData :=
+    (noContext : Unit → Format := fun _ => "<no context, could not generate MessageData>") : MessageData :=
   .ofPPFormat
   { pp := fun
       | some ctx => ctx.runMetaM <| withOptions (pp.tagAppFns.set · true) fmt
-      | none => return noContext.get }
+      | none => return noContext () }
 
 /-- Pretty print a const expression using `delabConst` and generate terminfo.
 This function avoids inserting `@` if the constant is for a function whose first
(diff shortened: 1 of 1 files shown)


digama0 (approved):

src/Lean/PrettyPrinter.lean
```diff
@@ -101,4 +98,37 @@ unsafe def registerParserCompilers : IO Unit := do
...
+
+open Lean PrettyPrinter Delaborator
+
+/--
+Turns a `MetaM FormatWithInfos` into a `MessageData` using `.ofPPFormat` and running the monadic value in the given context.
+Uses the `pp.tagAppFns` option to annotate constants with terminfo, which is necessary for seeing the type on mouse hover.
+-/
+def ofFormatWithInfos
+    (fmt : MetaM FormatWithInfos)
+    (noContext : Thunk Format := "<no context, could not generate MessageData>") : MessageData :=
```
kmill, line 112: I think one-shot lazy values usually use `Unit -> _` instead of `Thunk _`, and I *believe* `Unit -> _` is a bit more efficient, but I've never measured it.
```suggestion
    (noContext : Unit -> Format := fun _ => "<no context, could not generate MessageData>") : MessageData :=
```

kmill (commented):
In the PR message:

>  improving the hover behaviour in #check.

Could you mention what's being improved in the PR message? I seem to be able to hover over everything except for the constant itself, so I'm guessing it's that this lets you hover over the constant? That makes some sense, for being able to see the docstring, even if you don't need to see the type.

src/Lean/PrettyPrinter.lean
semorrison: Fixed!

PR #3912, opened 2024-04-15
feat: show diffs when #guard_msgs fails

david-christiansen:
Adds the ability to show a diff when `guard_msgs` fails, using the histogram diff algorithm pioneered in jgit. This algorithm tends to produce more user-friendly diffs, but it can be quadratic in the worst case. Empirically, the quadratic case of this implementation doesn't seem to be slow enough to matter for messages smaller than hundreds of megabytes, but if it's ever a problem, we can mitigate it the same way jgit does by falling back to Myers diff.

See lean/run/guard_msgs.lean in the tests directory for some examples of its output.


src/Init/Data/Array/Subarray/Split.lean
```diff
@@ -0,0 +1,68 @@
...
+  }
+  (pre, post)
+
+/--
+Removes the first `i` elements of the subarray. If there are `i` or fewer elements, the resulting
+subarray is empty.
+-/
+def drop (arr : Subarray α) (i : Nat) : Subarray α where
+  array := arr.array
+  start := if arr.start + i ≤ arr.stop then arr.start + i else arr.stop
```
digama0, line 49:
```suggestion
  start := min (arr.start + i) arr.stop
```

src/Init/Data/Array/Subarray/Split.lean
```diff
@@ -0,0 +1,68 @@
...
+  stop_le_array_size := arr.stop_le_array_size
+
+/--
+Keeps only the first `i` elements of the subarray. If there are `i` or fewer elements, the resulting
+subarray is empty.
+-/
+def take (arr : Subarray α) (i : Nat) : Subarray α where
+  array := arr.array
+  start := arr.start
+  stop := if i ≤ arr.stop - arr.start then arr.start + i else arr.stop
```
digama0, line 61:
```suggestion
  stop := min (arr.start + i) arr.stop
```

src/Init/Data/Array/Subarray/Split.lean
david-christiansen: Thanks! I did it in the editor so I could fix the proofs too.

src/Init/Data/Array/Subarray/Split.lean
david-christiansen: Thanks! I did it in the editor so I could fix the proofs too.

david-christiansen:
@nomeata you've asked about this a few times. Does this PR cover your use case?

semorrison:
Sounds great to me, but maybe should be 3 separate PRs (diffs, subarrays, guard_msgs)? No objection if someone else wants to merge as is.

semorrison:
(Sorry, wrong button!)

RELEASES.md
```diff
@@ -99,6 +99,10 @@ v4.8.0 (development in progress)
   and `#guard_msgs (ordering := sorted) in cmd` sorts the messages in lexicographic order before checking.
   PR [#3883](https://github.com/leanprover/lean4/pull/3883).
 
+* The `#guard_msgs` command now supports showing a diff between the expected and actual outputs. This feature is currently
+  disabled by default, but can be enabled with `set_option guard_msgs.diff true`. Depending on user feedback, this option
+  may default to `true` in a future version of Lean.
```
nomeata, line 104: Why not on by default? If it's off I expect few people will ever benefit from it, simply because they don't know about it.

We can still turn it off by default if it's not well received.

digama0: For both this and in expected type mismatch messages, I think both views are useful, but I don't think a `set_option` is the best way to surface it to users, a toggle button in the info view would be better.

david-christiansen: I agree that an infoview toggle would be better, but it's not achievable in the time I have right now.

Another reason to have it off by default right now is that the practicality of something like this is hard to assess in the abstract, or with contrived examples. I plan to have it on for a couple of months, and I'm sure a bunch of little papercuts will surface that I can fix.

I'm already expecting that I'll want a threshold of similarity that triggers the diff rendering (e.g. if less than X% of lines are edited, show a diff, otherwise don't), but I'd rather base that kind of thing on real experience by myself and others.

The plan is to make it on by default once it makes people happy enough, but I'd rather have the experiment be opt-in to start with.

nomeata: Fair enough!

david-christiansen:
OK, I think this is both good enough to be useful and unlikely to get substantially better in the next little while - I'll merge, then dogfood for a while to make it really nice and polished

PR #3913, opened 2024-04-15
feat: BaseIO.asThunk

nomeata:
this adds
```
asThunk (f : BaseIO α) : BaseIO (Thunk α)
```
with `unsafeInterleaveIO`-like semantiks: The `IO` action will not be
executed until the first use of `Thunk.get`.

This is a samewhat unsavory operation: Whether `f` is executed now
depends on when and whether the pure `Thunk.get` is performed.
Nevertheless it seemed to be the least painful way to achieve lazy
calcuation of MessageData (see #3234), in particular because Lean's
positivity checker will allow recursion of `MessageData` through
`Thunk`, but not through `IO`.

### DIFF
--- a/tests/compiler/io_as_thunk.lean
+++ b/tests/compiler/io_as_thunk.lean
@@ -1,6 +1,6 @@
 def p (s: String) : BaseIO String := do
   match ← (IO.println s).toBaseIO with
-  | .ok _ => pure s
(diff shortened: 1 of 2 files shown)


nomeata:
Far to big a hammer for what we want to do, closing for now.

PR #3915, opened 2024-04-15
doc: layout algorithm

digama0:
The layout algorithm, while somewhat finicky, is (unfortunately) necessary for C code to interface with lean structures. This adds a (AFAIK) complete description of the layout algorithm, including a worked example large enough to make it possible to reconstruct the whole decision diagram.

### DIFF
--- a/doc/dev/ffi.md
+++ b/doc/dev/ffi.md
@@ -70,8 +70,8 @@ The memory order of the fields is derived from the types and order of the fields
 Within each group the fields are ordered in declaration order. **Warning**: Trivial wrapper types still count toward a field being treated as non-scalar for this purpose.
 
 * To access fields of the first kind, use `lean_ctor_get(val, i)` to get the `i`th non-scalar field.
-* To access usize fields, use `lean_ctor_get_usize(val, n+i)` to get the `i`th usize field and `n` is the total number of fields of the first kind.
-* To access scalar fields, use `lean_ctor_get_uintN(val, off)` or `lean_ctor_get_usize(val, off)` as appropriate. Here `off` is the byte offset of the field in the structure, starting at `n*sizeof(void*)` where `n` is the number of fields of the first two kinds.
+* To access `USize` fields, use `lean_ctor_get_usize(val, n+i)` to get the `i`th usize field and `n` is the total number of fields of the first kind.
+* To access other scalar fields, use `lean_ctor_get_uintN(val, off)` or `lean_ctor_get_usize(val, off)` as appropriate. Here `off` is the byte offset of the field in the structure, starting at `n*sizeof(void*)` where `n` is the number of fields of the first two kinds.
 
 For example, a structure such as
 ```lean


doc/dev/ffi.md
```diff
@@ -53,10 +53,59 @@ In the case of `@[extern]` all *irrelevant* types are removed first; see next se
...
+
+For inductive types which are in the fallback `lean_object *` case above and not trivial constructors, the type is stored as a `lean_ctor_object`, and `lean_is_ctor` will return true. A `lean_ctor_object` stores the constructor index in the header, and the fields are stored in the `m_objs` portion of the object.
+
+The memory order of the fields is derived from the types and order of the fields in the declaration. They are ordered as follows:
+
+* Non-scalar fields stored as `lean_object *`
+* Fields of type `USize`
+* Other scalar fields, in decreasing order by size
+
+Within each group the fields are ordered in declaration order. **Warning**: Trivial wrapper types still count toward a field being treated as non-scalar for this purpose.
```
digama0, line 70: I did not know this until I sat down to document it. I'm pretty sure this is a bug.

Kha:
If we're documenting access, shouldn't we also document construction?

doc/dev/ffi.md
Kha: So what happens with the stored value, is it boxed or extended (what happens with a trivial UInt64 wrapper on 32 bit?)?

digama0:
The example with access functions was mainly for demonstration purposes, to show the way offsets are used. Hopefully pointing to the existence of setter functions and saying "these are used similarly" should be sufficient.

doc/dev/ffi.md
digama0: It's boxed. Basically it's the same as the fallback encoding for e.g. the elements of `Array UInt32`, which stores its values in `lean_object*` slots and uses `lean_box` and `lean_unbox` to get at the values.

We should probably say something about how types like `Array UInt32` work here...

digama0: As for UInt64 on 32 bit, I'm actually not sure, I was also wondering about that. The current scheme of putting `usize, u64, u32, u16, u8` in that order and not using any padding seems to assume that `usize` has the largest alignment, but I don't think this is true on 32 bit unless u64 is also 32 bit aligned, which I think is not the case on at least some of the supported arches. From what I can tell of the code though there is no guard against this condition, the offsets are still as documented here and the u64s will be unaligned in this case.

doc/dev/ffi.md
```diff
@@ -53,10 +53,59 @@ In the case of `@[extern]` all *irrelevant* types are removed first; see next se
...
+* `S.usize_1` - `lean_ctor_get_usize(val, 3)`
+* `S.usize_2` - `lean_ctor_get_usize(val, 4)`
+* `S.sc64_1` - `lean_ctor_get_uint64(val, sizeof(void*)*5)`
+* `S.sc64_2` - `lean_ctor_get_float(val, sizeof(void*)*5 + 8)`
+* `S.sc64_3` - `lean_ctor_get_uint64(val, sizeof(void*)*5 + 16)`
+* `S.sc32_1` - `lean_ctor_get_uint32(val, sizeof(void*)*5 + 24)`
+* `S.sc16_1` - `lean_ctor_get_uint16(val, sizeof(void*)*5 + 28)`
+* `S.sc16_2` - `lean_ctor_get_uint16(val, sizeof(void*)*5 + 30)`
+* `S.sc8_1` - `lean_ctor_get_uint8(val, sizeof(void*)*5 + 32)`
+* `S.sc8_2` - `lean_ctor_get_uint8(val, sizeof(void*)*5 + 33)`
```
eric-wieser, line 107: Should we have a `set_option trace.compiler.ffi_accessors` option to dump this information for user structures?

doc/dev/ffi.md
```diff
@@ -53,10 +53,59 @@ In the case of `@[extern]` all *irrelevant* types are removed first; see next se
...
+
+* Non-scalar fields stored as `lean_object *`
+* Fields of type `USize`
+* Other scalar fields, in decreasing order by size
+
+Within each group the fields are ordered in declaration order. **Warning**: Trivial wrapper types still count toward a field being treated as non-scalar for this purpose.
+
+* To access fields of the first kind, use `lean_ctor_get(val, i)` to get the `i`th non-scalar field.
+* To access usize fields, use `lean_ctor_get_usize(val, n+i)` to get the `i`th usize field and `n` is the total number of fields of the first kind.
+* To access scalar fields, use `lean_ctor_get_uintN(val, off)` or `lean_ctor_get_usize(val, off)` as appropriate. Here `off` is the byte offset of the field in the structure, starting at `n*sizeof(void*)` where `n` is the number of fields of the first two kinds.
```
Kha, line 74:
```suggestion
* To access other scalar fields, use `lean_ctor_get_uintN(val, off)` or `lean_ctor_get_usize(val, off)` as appropriate. Here `off` is the byte offset of the field in the structure, starting at `n*sizeof(void*)` where `n` is the number of fields of the first two kinds.
```

doc/dev/ffi.md
```diff
@@ -53,10 +53,59 @@ In the case of `@[extern]` all *irrelevant* types are removed first; see next se
...
+The memory order of the fields is derived from the types and order of the fields in the declaration. They are ordered as follows:
+
+* Non-scalar fields stored as `lean_object *`
+* Fields of type `USize`
+* Other scalar fields, in decreasing order by size
+
+Within each group the fields are ordered in declaration order. **Warning**: Trivial wrapper types still count toward a field being treated as non-scalar for this purpose.
+
+* To access fields of the first kind, use `lean_ctor_get(val, i)` to get the `i`th non-scalar field.
+* To access usize fields, use `lean_ctor_get_usize(val, n+i)` to get the `i`th usize field and `n` is the total number of fields of the first kind.
```
Kha, line 73:
```suggestion
* To access `USize` fields, use `lean_ctor_get_usize(val, n+i)` to get the `i`th usize field and `n` is the total number of fields of the first kind.
```

Kha (approved):

doc/dev/ffi.md
digama0: opened as a separate issue: #4041

PR #3917, opened 2024-04-15
doc: docstrings on binder types, make sure hovers work

kmill:
Moved `ppGroup` inside the `leading_parser`s for all the binder types so that hovering works. Improved the docstrings.


nomeata (commented):
Thanks for the docstrings, very helpful already!

Kha (approved):

PR #3919, opened 2024-04-15
doc: docstrings for List.rotateRight/Left and example for List.partitionM

austinletson:


src/Init/Data/List/BasicAux.lean
```diff
@@ -288,6 +306,15 @@ def mapMono (as : List α) (f : α → α) : List α :=
 Monadic generalization of `List.partition`.
 
 This uses `Array.toList` and which isn't imported by `Init.Data.List.Basic`.
+```
+def posOrNeg? (x : Int) : Option Bool :=
```
austinletson, line 310: Is there a better/canonical example which is used for monadic functions?

I also considered,
```lean
def posOrNeg (x : Int) : Except String Bool :=
  if x > 0 then pure true
  else if x < 0 then pure false
  else throw "Zero is not positive or negative
```

src/Init/Data/List/BasicAux.lean
```diff
@@ -148,6 +148,15 @@ def getLastD : (as : List α) → (fallback : α) → α
   | [],   a₀ => a₀
   | a::as, _ => getLast (a::as) (fun h => List.noConfusion h)
 
+/--
+`O(n)`. Rotates the elements of `xs` to the left such that the element at
+`xs[i]` rotates to `xs[(i - n) % l.length]`.
+
+`n` defaults to `1`.
```
austinletson, line 155: Is it necessary to specify the default value of `n` given that the [docs will already include `(n : optParam Nat 1)`](https://leanprover-community.github.io/mathlib4_docs/Init/Data/List/BasicAux.html#List.rotateLeft)? 

2024-04-15-18:18:22-screenshot

src/Init/Data/List/BasicAux.lean
digama0: You can use `IO` and `println!`

src/Init/Data/List/BasicAux.lean
digama0: I would say no... The main places people see these doc strings are docgen and hovers, but the type signature including the optParam shows up in both places in this case.

src/Init/Data/List/BasicAux.lean
austinletson: What about this instead of the `Option Bool` example?
```lean
def isEven (x : Int) : IO Bool := do
  println! s!"Checking if {x} is even"
  pure (x % 2 == 0)

partitionM isEven [1, 2] = do
  println! "Checking if 1 is even"
  println! "Checking if 2 is even"
  pure ([2], [1])
```

src/Init/Data/List/BasicAux.lean
austinletson: Sounds good. I removed the note about the default value in both rotate functions

david-christiansen: That makes sense - we can assume that readers are seeing the docstring in the context of the signature.

src/Init/Data/List/BasicAux.lean
david-christiansen: While this equality might hold, I think we'll have readers who aren't used to seeing things that way. I think the `Except` version is a better example, though the `IO` one could be an OK followup.

src/Init/Data/List/BasicAux.lean
```diff
@@ -288,6 +302,15 @@ def mapMono (as : List α) (f : α → α) : List α :=
 Monadic generalization of `List.partition`.
 
 This uses `Array.toList` and which isn't imported by `Init.Data.List.Basic`.
+```
+def posOrNeg? (x : Int) : Option Bool :=
+  if x > 0 then true
```
david-christiansen, line 307: The use of the coercion here can make the code harder to follow for new users. What about replacing them with `pure` to make the monad more clear, if you keep this version over `Except`?

src/Init/Data/List/BasicAux.lean
digama0: Maybe it could be written not as an equality but just "this is printed when you `#eval` it"

austinletson: I like the `Except` version because `Except` is incorporated into the predicate logic. With two examples the docstring can illustrate how `partitionM` handles both successful partitions and exceptions:

```lean
partitionM posOrNeg [-1, 2, 3] = Except.ok ([2, 3], [-1])
partitionM posOrNeg [0, 2, 3] = Except.error "Zero is not positive or negative"
```

Additionally, here is an alternative `IO` example:

```lean
def isEven (x : Int) : IO Bool := do  
  println! s!"Checking if {x} is even"  
  pure (x % 2 == 0)

-- prints "Checking if 1 is even\nChecking if 2 is even\n"
#eval partitionM isEven [1, 2] => pure ([2], [1])
```

I would lean toward `Except` (or both if we want two examples).

src/Init/Data/List/BasicAux.lean
austinletson: Good point. If I keep this example, I will make that change. I am going to wait to update until the above conversation is resolved.

src/Init/Data/List/BasicAux.lean
david-christiansen: I think this `Except` example is great. No need to block this contribution on re-planning everything - let's look at the other functions on a case-by-case basis rather than deciding on an example monad _a priori_.

austinletson: Got it. I updated the proposed changes with the `Except` example.

PR #3920, opened 2024-04-16
feat: equivalence of bit-vector negation and bitblasted negation

mhk119:


src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -159,4 +159,35 @@ theorem add_eq_adc (w : Nat) (x y : BitVec w) : x + y = (adc x y false).snd := b
...
   rw [← add_not_self x, BitVec.add_comm, add_sub_cancel]
 
+/-! ### Negation -/
+
+/-- Bitwise 1's complement implemented via `iunfoldr`. -/
+abbrev bit_not (x : BitVec w) : BitVec w :=
+  ((iunfoldr fun (i : Fin w) c => (c, !(x.getLsb i))) ()).snd
+
+/-- Bitwise 2's complement implemented via `iunfoldr`. -/
+abbrev bit_neg (x : BitVec w) : BitVec w := (adc (bit_not x) (BitVec.ofNat w 1) false).snd
```
semorrison, line 169: Could you explain looking further forward what you would like to use these results for? In isolation this PR is introducing two new definitions and proving they are equal to existing ones. I'd like to understand the application before increasing the API surface in this way.

mhk119: Yes, this was raised in [a previous PR](https://github.com/leanprover/std4/pull/563#discussion_r1467183951) too. Here we use `abbrev` instead of `def` to circumvent adding new definitions. However, if this is undesirable then we can certainly go with [@fgdorais's comment](https://github.com/leanprover/std4/pull/563#discussion_r1467183951).

These definitions aren't only meant to be used by the SMT solver in the bit-blasting stage. The `abbrev` makes the theorem statements shorter.

semorrison: Let's go with François' suggestion, please.

PR #3925, opened 2024-04-16
fix: do not send as many semantic token refresh requests

### LINKED ISSUE #3879: semantic token polling causes non-trivial idle CPU usage
I observed that a VS Code in the background (or different virtual desktop) with a fully elaborated file still consumes a non-trivial amount of CPU on my machine (~10%).

Based on `strace` and `perf top`, the issue seems to be that the periodic semantic token updates are relatively costly even if nothing new is done.

### Versions

892bfe2c5fdd964405d94f0adb0e178fe2b485af

### Impact

mhuisi:
Fixes #3879.

Making semantic token requests fast is still in progress.

### DIFF
--- a/src/Lean/Server/FileWorker.lean
+++ b/src/Lean/Server/FileWorker.lean
@@ -73,6 +73,11 @@ structure WorkerContext where
   maxDocVersionRef     : IO.Ref Int
   freshRequestIdRef    : IO.Ref Int
   /--
+  Channel that receives a message for every a `$/lean/fileProgress` notification, indicating whether
+  the notification suggests that the file is currently being processed.
+  -/
+  chanIsProcessing     : IO.Channel Bool
+  /--
   Diagnostics that are included in every single `textDocument/publishDiagnostics` notification.
   -/
   stickyDiagnosticsRef : IO.Ref StickyDiagnostics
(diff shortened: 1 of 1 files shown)


src/Lean/Server/FileWorker.lean
```diff
@@ -374,6 +381,9 @@ section Initialization
           -- note that because of `server.reportDelayMs`, we cannot simply set `maxDocVersion` here
           -- as that would allow outdated messages to be reported until the delay is over
         o.writeLspMessage msg |>.catchExceptions (fun _ => pure ())
+        if let .notification "$/lean/fileProgress" (some params) := msg then
+          if let some (params : LeanFileProgressParams) := fromJson? (toJson params) |>.toOption then
+            chanIsProcessing.send (! params.processing.isEmpty)
```
Kha, line 386: Could be done in `reportSnapshots` without parsing? I do hope no-one else is sending out file progress notifs

Kha (approved):

src/Lean/Server/FileWorker.lean
mhuisi: What if the version for the `fileProgress` notification is out-of-date?

Kha: That shouldn't be a big concern as cancellation should prevent the reporter from doing that for any relevant amount of time but it's fine like this then

PR #3928, opened 2024-04-16
feat: make binder predicates use strict-implicit binders

kmill:
This PR is so far an experiment to see the impact of this change.

This would probably need to be augmented with a feature like `h (_ := 2)` or `h @(2)` to be able to pass in individual implicit arguments without doing `@h 2`.


kmill:
This isn't a good idea — the value is what it's quantifying over, so it shouldn't be implicit.

PR #3929, opened 2024-04-16
feat: apply’s error message should show implicit arguments as needed

### LINKED ISSUE #3232: “failed to unify” message should not print identical types

nomeata:
luckily the necessary functionality already exists in the form of
`addPPExplicitToExposeDiff`. But it is not cheap, and we should not run this code
when the error message isn’t shown, so we should do this lazily.

We already had `MessageData.ofPPFormat` to assemble the error message lazily, but it 
was restricted to returning `FormatWithInfo`, a data type that doesn’t admit a nice 
API to compose more complex messages (like `Format` or `MessageData` has; an attempt to 
fix that is in #3926).

Therefore we split the functionality of `.ofPPFormat` into `.ofFormatWithInfo` and `.ofLazy`,
and use `.ofLazy` to compute the more complex error message of `apply`.

Fixes #3232.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -144,6 +144,18 @@ fact.def :
 
 * `Option.toMonad` has been renamed to `Option.getM` and the unneeded `[Monad m]` instance argument has been removed.
 
+* The `MessageData.ofPPFormat` constructor has been removed.
+  Its functionality has been split into two:
+
+  - for lazy structured messages, please use `MessageData.lazy`;
+  - for embedding `Format` or `FormatWithInfos`, use `MessageData.ofFormatWithInfos`.
+
+  An example migration can be found in [#3929](https://github.com/leanprover/lean4/pull/3929/files#diff-5910592ab7452a0e1b2616c62d22202d2291a9ebb463145f198685aed6299867L109).
+
+* The `MessageData.ofFormat` constructor has been turned into a function.
+  If you need to inspect `MessageData`,
+  you can pattern-match on `MessageData.ofFormatWithInfos`.
+
 v4.7.0
 ---------
 


tests/lean/issue3232.lean
```diff
@@ -0,0 +1,11 @@
+inductive foo {n : Nat} : Prop
+
+-- Check that the error message will print the types with implicit arguments shownx
```
david-christiansen, line 3:
```suggestion
-- Check that the error message will print the types with implicit arguments shown
```

src/Lean/Message.lean
```diff
@@ -78,11 +78,25 @@ inductive MessageData where
     Example: an inspector that tries to find "definitional equality failures" may look for the tag "DefEqFailure". -/
   | tagged            : Name → MessageData → MessageData
   | trace (data : TraceData) (msg : MessageData) (children : Array MessageData)
-  deriving Inhabited
+  /-- Lazy message data production. The `Dymamic` is expected to be `MessageData`, we
```
david-christiansen, line 81:
```suggestion
  /-- Lazy message data production. The `Dynamic` is expected to be `MessageData`, we
```

src/Lean/Message.lean
```diff
@@ -78,11 +78,25 @@ inductive MessageData where
...
+  use this to work around the positivity restriction. -/
```
david-christiansen, line 82: Typo and comma splice:
```suggestion
  /-- Lazy message data production. The `Dynamic` is expected to be `MessageData`, which
  is a workaround for the positivity restriction. -/
```

src/Lean/Message.lean
```diff
@@ -78,11 +78,25 @@ inductive MessageData where
...
+  .ofLazy (fun ctx => Dynamic.mk <$> f ctx) hasSyntheticSorry
+
+/-- Lazily formatted text with Info annotations -/
+def ofPPFormat (f : PPFormat) : MessageData :=
+  ofLazy (hasSyntheticSorry := f.hasSyntheticSorry) fun ppctx => do
+    let md ← f.pp ppctx
+    return Dynamic.mk (MessageData.ofFormatWithInfos md)
+
+/-- Determines whether the message contains any content.
+This considers lazily generated messages (`.ofLazy`) to be nonempty.
```
david-christiansen, line 98: Why is this the right spec here? What do we use this check for? A quick comment to that effect would be nice.

src/Lean/Message.lean
```diff
@@ -94,7 +108,10 @@ def isEmpty : MessageData → Bool
   | _ => false
 
 variable (p : Name → Bool) in
-/-- Returns true when the message contains a `MessageData.tagged tag ..` constructor where `p tag` is true. -/
+/-- Returns true when the message contains a `MessageData.tagged tag ..` constructor where `p tag`
+is true.
+This does not descend into lazily generated subtress (`.ofLazy`).
```
david-christiansen, line 113: Why is `false` the right approximation here? I would think that lazy messages absolutely could contain tags, and I'd expect that this was used to make some operation go faster by ignoring them.

src/Lean/Message.lean
```diff
@@ -166,6 +183,11 @@ partial def formatAux : NamingContext → Option MessageDataContext → MessageD
     msg := f!"{msg} {(← formatAux nCtx ctx header).nest 2}"
     let children ← children.mapM (formatAux nCtx ctx)
     return .nest 2 (.joinSep (msg::children.toList) "\n")
+  | nCtx, ctx?,      ofLazy pp _             => do
+    let dyn ← pp (ctx?.map (mkPPContext nCtx))
+    let some msg := dyn.get? MessageData
+      | throw <| IO.userError "MessageData.ofLazy: expected MessageData in Dynamic"
```
david-christiansen, line 189: I would expect this to panic. Why is it an IO exception?

david-christiansen: It might also be nice, in case this ever does happen, to include something like `expected MessageData in Dynamic but got '{dyn.typeName}'`, but we can always put that in if it's ever a problem

src/Lean/Meta/Tactic/Apply.lean
```diff
@@ -50,8 +51,18 @@ def getExpectedNumArgs (e : Expr) : MetaM Nat := do
   let (numArgs, _) ← getExpectedNumArgsAux e
   pure numArgs
 
-private def throwApplyError {α} (mvarId : MVarId) (eType : Expr) (targetType : Expr) : MetaM α :=
-  throwTacticEx `apply mvarId m!"failed to unify{indentExpr eType}\nwith{indentExpr targetType}"
+private def throwApplyError {α} (mvarId : MVarId) (eType : Expr) (targetType : Expr) : MetaM α := do
+  let explanation := MessageData.lazy
+    (f := fun
+      | none =>
+        return f!"{eType}\nwith{targetType}"
```
Vtec234, line 58:
```suggestion
        return f!"{indentExpr eType}\nwith{indentExpr targetType}"
```
Perhaps?

nomeata: That only exists for MessageData not Format. But the whole case is a bit bogus and hopefully dead code, we should always have a context here, I hope. Maybe I'll make this an error and see if anyone hits this code path.

src/Lean/Message.lean
nomeata: Good question, let’s see. This is only used in 
```
def throwTacticEx (tacticName : Name) (mvarId : MVarId) (msg : MessageData) : MetaM α :=
  if msg.isEmpty then
    throwError "tactic '{tacticName}' failed\n{mvarId}"
  else
    throwError "tactic '{tacticName}' failed, {msg}\n{mvarId}"
```
and almost all calls to `throwTacticEx` has a manifestly non-empty error message. Exceptions are for example
```
       let some mvarIds ← splitTarget? mvarId | Meta.throwTacticEx `split mvarId ""
```
I’m inclined to side-step the issue by adding a proper message to any `Meta.throwTacticEx` invocation. Or, if that is tricky for some reason, change the type to `msg : Option MessageData := none` and use proper types for optional parameters.

src/Lean/Message.lean
nomeata: Also a good question. Note that this function doesn't check for just any tag, but for specific ones. Typical uses are

```
      -- discard elaboration errors, except for a few important and unlikely misleading ones, on
      -- parse error
      msgs := ⟨msgs.msgs.filter fun msg =>
        msg.data.hasTag (fun tag => tag == `Elab.synthPlaceholder ||
          tag == `Tactic.unsolvedGoals || (`_traceMsg).isSuffixOf tag)⟩
```
…
```
def _root_.Lean.MessageData.isDeprecationWarning (msg : MessageData) : Bool :=
  msg.hasTag (· == ``deprecatedAttr)
```
…
```
/-- Returns true if this is a message produced by the unused variable linter.
This is used to give these messages an additional "faded" style in the editor. -/
def MessageData.isUnusedVariableWarning (msg : MessageData) : Bool :=
  msg.hasTag (· == Linter.linter.unusedVariables.name)
```

I assume that these tags are always somewhere near the root of the message, and not inside a `.lazy` constructor, e.g. when created like this:
```
def logLint [Monad m] [MonadLog m] [AddMessageContext m] [MonadOptions m]
    (linterOption : Lean.Option Bool) (stx : Syntax) (msg : MessageData) : m Unit :=
  let disable := m!"note: this linter can be disabled with `set_option {linterOption.name} false`"
  logWarningAt stx (.tagged linterOption.name m!"{msg}\n{disable}")
```

One could imagine more verbose types to track such invariants, but I’d shy away from such a refactoring (I believe the data type already has such invariants, e.g. that certain constructors are always inside an `.withContext`).

Maybe good enough just documenting this assumption here?

```suggestion

This does not descend into lazily generated subtress (`.ofLazy`); message tags
of interest(like those added by `logLinter`) are expected to be near the root
of the `MessageData`, and not hidden inside `.ofLazy`.
```

src/Lean/Message.lean
nomeata: Did the latter in #3951, so this issue will simply disappear.

src/Lean/Message.lean
```diff
@@ -62,8 +62,8 @@ structure TraceData where
 inductive MessageData where
   /-- Eagerly formatted text. We inspect this in various hacks, so it is not immediately subsumed by `ofPPFormat`. -/
   | ofFormat          : Format → MessageData
-  /-- Lazily formatted text. -/
-  | ofPPFormat        : PPFormat → MessageData
+  /-- Formatted text with info annotations -/
+  | ofFormatWithInfos : FormatWithInfos → MessageData
```
Vtec234, line 66: It is plausible that the comment on `ofFormat` no longer applies, and `ofFormat` now *is* subsumed, i.e., all the `MessageData`-inspecting functions could as well inspect `ofFormatWithInfos` instead.

src/Lean/Message.lean
```diff
@@ -78,12 +78,38 @@ inductive MessageData where
...
+-/
+def lazy (f : PPContext → IO MessageData) (hasSyntheticSorry : MetavarContext → Bool) : MessageData :=
+  .ofLazy (hasSyntheticSorry := hasSyntheticSorry) fun ctx? => do
+    let msg ← match ctx? with
+      | .none => pure (.ofFormat "(invalid MessageData.lazy, missing context)")
+      | .some ctx => f ctx
+    return Dynamic.mk msg
+
+/-- Lazily formatted text with Info annotations -/
+def ofPPFormat (f : PPFormat) : MessageData :=
```
Vtec234, line 100: `PPFormat` was introduced to unify various lazy formatters, and is hardly used in the codebase. As this PR introduces a more general lazy constructor, I suggest that `ofPPFormat` and `PPFormat` be removed to clean up the API.

src/Lean/Message.lean
```diff
@@ -78,12 +78,38 @@ inductive MessageData where
     Example: an inspector that tries to find "definitional equality failures" may look for the tag "DefEqFailure". -/
   | tagged            : Name → MessageData → MessageData
   | trace (data : TraceData) (msg : MessageData) (children : Array MessageData)
-  deriving Inhabited
+  /-- Lazy message data production. The `Dynamic` is expected to be `MessageData`, which
+  is a workaround for the positivity restriction. -/
```
Vtec234, line 82:
```suggestion
  /-- A lazy message.
  The provided thunk will not be run until it is about to be displayed.
  This can save computation in cases where the message may never be seen,
  e.g. when nested inside a collapsed trace.
  The `Dynamic` value is expected to be a `MessageData`,
  which is a workaround for the positivity restriction. -/
```

Vtec234 (approved):
Hi! I have re-fetched the pretty-printing infrastructure into my mental cache and I think I agree that this is the right-est solution. Just to confirm though, why did you discard #3926? Is it because that still doesn't allow for lazily-produced `MessageData`, only `FormatWithInfos`, and furthermore introduced yet another appendable message type?

src/Lean/Message.lean
Vtec234: I have now carried out this experiment [here](https://github.com/Vtec234/lean4/tree/msgdata-refactor). The fallout in core appears minimal.

nomeata: Thanks, looks good! Do you want to push that to my branch?

nomeata:
> Hi! I have re-fetched the pretty-printing infrastructure into my mental cache and I think I agree that this is the right-est solution.

Thanks! It's good to have a second pair of eyes!

> Just to confirm though, why did you discard #3926? Is it because that still doesn't allow for lazily-produced `MessageData`, only `FormatWithInfos`, and furthermore introduced yet another appendable message type?

Yes! The way `FormarWithInfo` works makes appending then rather tedious (shifting indices), and it's not just appending but also string interpolation that would have to be duplicated. It didn't seem desirable to have three similar data structures around.

src/Lean/Message.lean
Vtec234: Done!

nomeata: Thanks! Now mathlib fails, and searching for the error message on the phone is annoying. But probably nothing worrying and we can merge?

Vtec234:
> Now mathlib fails, and searching for the error message on the phone is annoying. But probably nothing worrying and we can merge?

I have made adaptations [to batteries](https://github.com/leanprover-community/batteries/commits/lean-pr-testing-3929) and [to mathlib4](https://github.com/leanprover-community/mathlib4/commits/lean-pr-testing-3929), and written a changelog. If these build, I think we should be good.

But, it would still be good to benchmark as `MessageData` is used quite ubiquitously.

Vtec234:
!bench

nomeata:
Mathlib seems to be happy as well, thanks. Any last thoughts before I press the button?

Vtec234:
LGTM!

nomeata:
Thanks for your help getting this over the finishing line.

We should probably now make the `omega` error lazy. I'm curious if that will show up in mathlib.

PR #3930, opened 2024-04-17
feat: `isTty`

tydeu:
Adds `IO.FS.Handle.isTty` to check whether a handle is a Windows console or Unix terminal. Also adds an `isTty` field to `IO.FS.Stream`, so that this can be checked on, e.g., `stdout`.

### DIFF
--- a/src/Init/System/IO.lean
+++ b/src/Init/System/IO.lean
@@ -311,7 +311,8 @@ Note that EOF does not actually close a stream, so further reads may block and r
   -/
   getLine : IO String
   putStr  : String → IO Unit
-  isatty  : BaseIO Bool
+  /-- Returns true if a stream refers to a Windows console or Unix terminal. -/
+  isTty   : BaseIO Bool
   deriving Inhabited
 
 open FS
@@ -362,7 +363,7 @@ Will succeed even if no lock has been acquired.
 @[extern "lean_io_prim_handle_unlock"] opaque unlock (h : @& Handle) : IO Unit
 
 /-- Returns true if a handle refers to a Windows console or Unix terminal. -/
-@[extern "lean_io_prim_handle_isatty"] opaque isatty (h : @& Handle) : BaseIO Bool
+@[extern "lean_io_prim_handle_is_tty"] opaque isTty (h : @& Handle) : BaseIO Bool
 
 @[extern "lean_io_prim_handle_flush"] opaque flush (h : @& Handle) : IO Unit
 /-- Rewinds the read/write cursor to the beginning of the handle. -/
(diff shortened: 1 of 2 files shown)


src/runtime/io.cpp
```diff
@@ -391,13 +394,40 @@ extern "C" LEAN_EXPORT obj_res lean_io_prim_handle_unlock(b_obj_arg h, obj_arg /
...
 #endif
 
+/* Handle.isatty : (@& Handle) → BaseIO Bool */
+extern "C" LEAN_EXPORT obj_res lean_io_prim_handle_isatty(b_obj_arg h, obj_arg /* w */) {
+    FILE * fp = io_get_handle(h);
+#ifdef LEAN_WINDOWS
+    /*
+    On Windows, there are two approaches for detecting a console.
+    1)  _isatty(_fileno(fp)) != 0
+        This checks whether the the file descriptor is is a *character device*,
```
Kha, line 404:
```suggestion
        This checks whether the file descriptor is a *character device*,
```

src/runtime/io.cpp
```diff
@@ -391,13 +394,40 @@ extern "C" LEAN_EXPORT obj_res lean_io_prim_handle_unlock(b_obj_arg h, obj_arg /
...
+        not just a terminal (unlike Unix's isatty). Thus, it produces a false
+        positive in some edge cases (such as NUL).
+        https://stackoverflow.com/q/3648711
+    2)  GetConsoleMode(win_handle(fp), &mode) != 0
+        Errors if the handle is not a console. Unfortunately, this produces
+        a false negative for a terminal emulator like MSYS/Cygwin's Mintty,
+        which is not implemented as a Windows-recognized console on
+        old Windows versions (e.g., pre-Windows 10, pre-ConPTY).
```
Kha, line 412: I am not sure why this caveat is needed when we explicitly do not support those platforms :) . But it can't hurt.

src/Init/System/IO.lean
```diff
@@ -360,6 +361,9 @@ Will succeed even if no lock has been acquired.
 -/
 @[extern "lean_io_prim_handle_unlock"] opaque unlock (h : @& Handle) : IO Unit
 
+/-- Returns true if a handle refers to a Windows console or Unix terminal. -/
+@[extern "lean_io_prim_handle_isatty"] opaque isatty (h : @& Handle) : BaseIO Bool
```
Kha, line 365: Since the underlying function is platform-dependent anyway, why not a more standard name like `isTty` or `isTerminal`? https://docs.rs/termion/latest/termion/fn.is_tty.html

src/runtime/io.cpp
tydeu: Good point! I think it is worth keeping in mind, though, in case a similar issue emerges in the future.

PR #3933, opened 2024-04-17
feat: rm partial / bounds checks in Array.qsort

digama0:
With `omega` in core, it is now more feasible to perform the necessary proofs to eliminate the `partial` from `Array.qsort`, as well as the many uses of `Array.get!` and `Array.swap!`. (This is also a perfect example of the kind of code that would benefit from the proposed [`Vector`](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/RFC.3A.20Arrays.20with.20fixed.20length) API.)

I wrote this code over a year ago, but the stance regarding proofs in lean core has shifted so I think this is now more likely to be accepted. This is in any case blocking some proofs in Std which are built on top of uses of Array.qsort.

### DIFF
--- a/src/Init/Data/Array/QSort.lean
+++ b/src/Init/Data/Array/QSort.lean
@@ -5,42 +5,58 @@ Authors: Leonardo de Moura
 -/
 prelude
 import Init.Data.Array.Basic
+import Init.Omega
 
 namespace Array
--- TODO: remove the [Inhabited α] parameters as soon as we have the tactic framework for automating proof generation and using Array.fget
 
-def qpartition (as : Array α) (lt : α → α → Bool) (lo hi : Nat) : Nat × Array α :=
-  if h : as.size = 0 then (0, as) else have : Inhabited α := ⟨as[0]'(by revert h; cases as.size <;> simp)⟩ -- TODO: remove
-  let mid := (lo + hi) / 2
-  let as  := if lt (as.get! mid) (as.get! lo) then as.swap! lo mid else as
-  let as  := if lt (as.get! hi)  (as.get! lo) then as.swap! lo hi  else as
-  let as  := if lt (as.get! mid) (as.get! hi) then as.swap! mid hi else as
-  let pivot := as.get! hi
-  let rec loop (as : Array α) (i j : Nat) :=
-    if h : j < hi then
-      if lt (as.get! j) pivot then
-        let as := as.swap! i j
-        loop as (i+1) (j+1)
+@[inline] def qpartition (as : {as : Array α // as.size = n})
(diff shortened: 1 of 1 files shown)


digama0:
!bench

digama0:
!build

hargoniX:
Do you know why this gets the `stdlib` benchmark slower by a clearly significant margin? This seems rather counter intuitive to me. Are the proofs in `Array.qsort` taking so long?

src/Init/Data/Array/QSort.lean
```diff
@@ -5,42 +5,58 @@ Authors: Leonardo de Moura
...
-      else
-        let as := sort as low mid
-        sort as (mid+1) high
+@[inline] def qsort (as : Array α) (lt : α → α → Bool) (low := 0) (high := as.size - 1) : Array α :=
+  let rec @[specialize] sort {n} (as : {as : Array α // as.size = n})
+      (lo : Nat) (hi : Fin n) : {as : Array α // as.size = n} :=
+    if h : lo < hi.1 then
+      let ⟨as, mid, (_ : lo ≤ mid), _⟩ :=
+        qpartition as lt ⟨lo, Nat.lt_trans h hi.2⟩ hi (Nat.le_of_lt h)
+      let as := sort as lo ⟨mid - 1, by omega⟩
```
collares, line 50: Not necessarily relevant for this PR, but one simple improvement to quicksort (as explained in the Sedgewick paper) is to recurse on the smallest subproblem first. This ensures a stack depth of `O(log n)` because the other call will be in tail position.

digama0:
The proof is actually suspiciously slow (the definition of `qpartition` is about 2 seconds). I think it might actually be the compiler too; this code contains three if statements at the beginning and an `@[inline]` and this seems to hit a relatively bad case for the compiler (there were other variations of this code that behaved even worse). The generated code itself seems to be fine though.

This code is slightly suboptimal, in that `n` is passed as an argument to `qpartition` even though it is redundant, and likewise `lo` and `hi` are passed unnecessarily to `qpartition.loop`. My original version of this code used some tricks to eliminate this, at the cost of making the proof significantly more obtuse; I decided that this was not a good tradeoff for maintenance reasons and submitted the simpler version here. Ideally, this would be solved by marking these parameters as "ghost" / not computationally relevant.

digama0:
(The version that was previously benched had a sharing bug. Unsure if that is related to the stdlib slowdown.)

Kha:
Marking draft pending further benchmarking (build time and run time)

kim-em:
Oh, sorry, I missed this, and did it independently. (Additionally, using Vector to clean up a bit.)

I also have most of proof that the result is sorted, on a branch.

PR #3939, opened 2024-04-17
feat: lean CLI option to print messages as JSON

tydeu:
Adds a `--json` option to the `lean` CLI. When used, the Lean frontend will print messages as JSON objects using the default `ToJson` encoding for the `Message` structure. This allows consumers (such as Lake) to handle Lean output in a more intelligent, well-structured way.

`Message` has been refactored into `BaseMessage`, `Message`, and `SerialMessage` to enable deriving `ToJson`/ `FromJson` instances automatically for `BaseMessage` / `SerialMessage`.  `SerialMessage` is a `Message` with its `MessageData` eagerly serialized to a `String`.

### DIFF
--- a/src/Lean/Language/Basic.lean
+++ b/src/Lean/Language/Basic.lean
@@ -140,7 +140,10 @@ export ToSnapshotTree (toSnapshotTree)
   preorder. -/
 @[specialize] partial def SnapshotTree.forM [Monad m] (s : SnapshotTree)
     (f : Snapshot → m PUnit) : m PUnit := do
-  match s with | mk element children => f element; children.forM (·.get.forM f)
+  match s with
+  | mk element children =>
+    f element
+    children.forM (·.get.forM f)
 
 /--
   Option for printing end position of each message in addition to start position. Used for testing
--- a/src/Lean/Message.lean
+++ b/src/Lean/Message.lean
@@ -239,8 +239,6 @@ namespace SerialMessage
 @[inline] def toMessage (msg : SerialMessage) : Message :=
   {msg with data := msg.data}
 
-instance : Coe SerialMessage Message := ⟨SerialMessage.toMessage⟩
-
 protected def toString (msg : SerialMessage) (includeEndPos := false) : String := Id.run do
   let mut str := msg.data
   let endPos := if includeEndPos then msg.endPos else none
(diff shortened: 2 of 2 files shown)


src/Lean/Language/Basic.lean
```diff
@@ -135,32 +135,36 @@ class ToSnapshotTree (α : Type) where
   toSnapshotTree : α → SnapshotTree
 export ToSnapshotTree (toSnapshotTree)
 
+/-- Runs a tree of snapshots to conclusion, incrementally performing `f` on each snapshot in tree preorder. -/
+@[specialize] partial def SnapshotTree.forM [Monad m] (s : SnapshotTree)
+    (f : Snapshot → m PUnit) : m PUnit := do
+  match s with | mk element children => f element; children.forM (·.get.forM f)
+
```
tydeu, line 142: @Kha I added this method to make `runAndReport` and `getAll` one-liners and to avoid a potential linearity issue with the projections. As this is stylistic refactor, I wanted to highlight it to you (in case you do not approve).

src/Lean/Message.lean
```diff
@@ -216,24 +222,55 @@ structure Message where
...
+
+end SerialMessage
+
+namespace Message
+
+@[inline] def serialize (msg : Message) : IO SerialMessage := do
+  return {msg with data := ← msg.data.toString}
+
+protected def toString (msg : Message) (includeEndPos := false) : IO String := do
+  -- Remark: We inline here to avoid a copy in the case were `msg` is shared
```
Kha, line 267: I don't understand this comment. A copy of what? Should we care, what use case are we talking about that could make this relevant?

src/Lean/Language/Basic.lean
Kha: The function great, just the formatting of the body is unusual. I would definitely write this as four lines, not one.

Kha (changes requested):

src/Lean/Message.lean
```diff
@@ -216,24 +222,55 @@ structure Message where
...
+serialized and is thus appropriate for use in pure contexts where the effectful
+`MessageData.toString` cannot be used. -/
+abbrev SerialMessage := BaseMessage String
+
+namespace SerialMessage
+
+@[inline] def toMessage (msg : SerialMessage) : Message :=
+  {msg with data := msg.data}
+
+instance : Coe SerialMessage Message := ⟨SerialMessage.toMessage⟩
```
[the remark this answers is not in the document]
tydeu: No, but this is potentially useful for Lean consumers of the JSON messages as they can report them back as normal messages. Since it sounds like we are unsure of this direction, I will remove the coercion.

src/Lean/Message.lean
tydeu: If `msg` is shared, we can just replace the data with its serialized version before applying `toString`. However, If the `msg` is shared, serializing it will require constructing (and allocating) and new `BaseMessage` with the serialized data. This will then be immediately deconstructed by the `toString` function. With the inline, we avoid this temporary allocation (as the serialization and destruction occur in the same function and thus Lean is smart enough to optimize it away).

Kha (approved):

PR #3940, opened 2024-04-18
feat: incremental elaboration of definition headers, bodies, and tactics

Kha:
Extends Lean's incremental reporting and reuse between commands into various steps inside declarations:
* headers and bodies of each (mutual) definition/theorem
* `theorem ... := by` for each contained tactic step, including recursively inside supported combinators currently consisting of
  * `·` (cdot), `case`, `next`
  * `induction`, `cases`
  * macros such as `next` unfolding to the above

Recording 2024-05-10 at 11 07 32

*Incremental reuse* means not recomputing any such steps if they are not affected by a document change. *Incremental reporting* includes the parts seen in the recording above: the progress bar and messages. Other language server features such as hover etc. are *not yet* supported incrementally, i.e. they are shown only when the declaration has been fully processed as before.

### DIFF
--- a/src/Lean/Message.lean
+++ b/src/Lean/Message.lean
@@ -326,6 +326,10 @@ structure MessageLog where
 namespace MessageLog
 def empty : MessageLog := {}
 
+@[deprecated "renamed to `unreported`; direct access should in general be avoided in favor of \
+using `MessageLog.toList/toArray`"]
+def msgs : MessageLog → PersistentArray Message := unreported
+
 def hasUnreported (log : MessageLog) : Bool :=
   !log.unreported.isEmpty
 


Kha:
#3636 was accidentally closed without merging

src/Lean/Elab/MutualDef.lean
```diff
@@ -68,6 +64,8 @@ private def check (prevHeaders : Array DefViewElabHeader) (newHeader : DefViewEl
     throwError "'partial' theorems are not allowed, 'partial' is a code generation directive"
   if newHeader.kind.isTheorem && newHeader.modifiers.isNoncomputable then
     throwError "'theorem' subsumes 'noncomputable', code is not generated for theorems"
+  if newHeader.modifiers.isNoncomputable && newHeader.modifiers.isUnsafe then
+    throwError "'noncomputable unsafe' is not allowed"
```
tydeu, line 68: A merge at some point appears to have gone wrong.

tydeu: Oh, looking further, the are other changes not incorporate here either. I presume you have just not merged master in a while. Sorry for the noise!

Kha: No, you were completely right! Thanks for flagging.

Kha:
Drafted until the base is changed back to master just before merge

PR #3942, opened 2024-04-18
fix: add instances to make ac_rfl work out of the box

TwoFX:
Previously the `ac_rfl` tactic was only really usable when depending on mathlib. With these instances, `ac_rfl` can deal with the various operations defined in Lean.


src/Init/Data/Nat/Lcm.lean
```diff
@@ -54,6 +59,7 @@ Nat.dvd_antisymm
     (Nat.dvd_trans (dvd_lcm_left m n) (dvd_lcm_left (lcm m n) k))
     (lcm_dvd (Nat.dvd_trans (dvd_lcm_right m n) (dvd_lcm_left (lcm m n) k))
       (dvd_lcm_right (lcm m n) k)))
+instance : Std.Associative lcm := ⟨lcm_assoc⟩
```
eric-wieser, line 62: It looks like `Associative gcd` was missed here

PR #3951, opened 2024-04-19
refactor: make throwTacticEx parameter `msg` optional

nomeata:
previously, the empty `MessageData` (`m!""`) was used to indicate “no
message”, and `throwTacticEx` would format the message differently then.
But the semantics of `MessageData.isEmpty` isn't entirely clear in the
presence of lazy message data (e.g. `.ofPPFormat`).

So to avoid wondering what `isEmpty` should do there, let's simply use
an optional argument to `throwTacticEx` and get rid of
`MessageData.isEmpty`.

### DIFF
--- a/src/Lean/Elab/Tactic/Split.lean
+++ b/src/Lean/Elab/Tactic/Split.lean
@@ -21,12 +21,12 @@ open Meta
       throwErrorAt stx[2] "'split' tactic failed, select a single target to split"
     if simplifyTarget then
       liftMetaTactic fun mvarId => do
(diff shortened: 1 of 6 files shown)


nomeata:
A quick grep through std4 and mathlib indicates that they don’t rely on the `""` behavior, so this doesn't affect them.

semorrison:
Looks like the lean-pr-testing-3951 branches are good to go, so I've put this on the queue and will hopefully be able to merge those branches tonight.

PR #3957, opened 2024-04-20
fix: lake: package duplication in workspace

tydeu:
Fixes a bug where packages that appeared multiple times in the dependency tree would be duplicated in the workspace (and in manifests). Added a regression test for this to prevent this from happening again in the future.

This was first reported in l[eanprover/mathlib4#12250](https://github.com/leanprover-community/mathlib4/pull/12258#discussion_r1571834509).


semorrison (approved):
LGTM. I'd like to have this in by v4.8.0-rc1.

semorrison:
Haven't seen that failure before! Adding `full-ci` to help debug.

PR #3958, opened 2024-04-20
feat: UTF-8 string validation

digama0:
Previously, there was a function `opaque fromUTF8Unchecked : ByteArray -> String` which would convert a list of bytes into a string, but as the name implies it does not validate that the string is UTF-8 before doing so and as a result it produces unsound results in the compiler (because the lean model of `String` indirectly asserts UTF-8 validity). This PR replaces that function by
```lean
opaque validateUTF8 (a : @& ByteArray) : Bool

opaque fromUTF8 (a : @& ByteArray) (h : validateUTF8 a) : String
```
so that while the function is still "unchecked", we have a proof witness that the string is valid. To recover the original, actually unchecked version, use `lcProof` or other unsafe methods to produce the proof witness.

Because this was the only `ByteArray -> String` conversion function, it was used in several places in an unsound way (e.g. reading untrusted input from IO and treating it as UTF-8). These have been replaced by `fromUTF8?` or `fromUTF8!` as appropriate.

### DIFF
--- a/tests/lean/run/utf8英語.lean
+++ b/tests/lean/run/utf8英語.lean
@@ -4,10 +4,13 @@ def check_eq {α} [BEq α] [Repr α] (tag : String) (expected actual : α) : IO 
       s!"assertion failure \"{tag}\":\n  expected: {repr expected}\n  actual:   {repr actual}"
 
 def DecodeUTF8: IO Unit := do
-  let cs := String.toList "Hello, 英語!"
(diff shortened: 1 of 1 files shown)


nomeata (approved):
Looks good to me, thanks!

 
Should we have some tests for the validator? Is there maybe an “official” set of test inputs that we can use in a test?

PR #3960, opened 2024-04-20
fix: don't use modulo for UInt upcasting

digama0:
This makes `(v : UInt8).toUInt16.toNat = (v : UInt8).toNat` a defeq, which simplifies proofs.

### DIFF
--- a/src/Init/Data/UInt/Basic.lean
+++ b/src/Init/Data/UInt/Basic.lean
@@ -103,7 +103,7 @@ def UInt16.shiftLeft (a b : UInt16) : UInt16 := ⟨a.val <<< (modn b 16).val⟩
 @[extern "lean_uint16_to_uint8"]
 def UInt16.toUInt8 (a : UInt16) : UInt8 := a.toNat.toUInt8
 @[extern "lean_uint8_to_uint16"]
-def UInt8.toUInt16 (a : UInt8) : UInt16 := a.toNat.toUInt16
+def UInt8.toUInt16 (a : UInt8) : UInt16 := ⟨a.val, Nat.lt_trans a.1.2 (by decide)⟩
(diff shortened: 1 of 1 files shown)


nomeata:
(Removed from queue due to mathlib failures; were they expected?)

digama0:
Yes, they are expected. Should I make a bump PR?

digama0:
Just checked, it is from theorems in Std like this:
```lean
@[simp] theorem UInt8.toUInt16_toNat (x : UInt8) : x.toUInt16.toNat = x.toNat :=
  Nat.mod_eq_of_lt (Nat.lt_of_lt_of_le x.toNat_lt (by decide))

@[simp] theorem UInt8.toUInt32_toNat (x : UInt8) : x.toUInt32.toNat = x.toNat :=
  Nat.mod_eq_of_lt (Nat.lt_of_lt_of_le x.toNat_lt (by decide))

@[simp] theorem UInt8.toUInt64_toNat (x : UInt8) : x.toUInt64.toNat = x.toNat :=
  Nat.mod_eq_of_lt (Nat.lt_of_lt_of_le x.toNat_lt (by decide))
```
which are exactly the ones that are rfl after this PR. So it should be easily fixable.

nomeata:
If you say it's fine I can merge, although it doesn't hurt to have the fixes ready ahead of time, by pushing to the lean-pr-testing branches. Then whoever updates nightly-testing has less to think about, don't they?

semorrison:
I did encounter the errors on Std's nightly-testing before reading this thread, but found it pretty quickly.

I think a good option, when not having a `lean-pr-testing-NNNN` branch ready (or even when!) is to post on zulip's `nightly-testing` stream saying that a breakage is expected (and, if available, a quick description of the fix).

PR #3961, opened 2024-04-20
feat: add model implementation for UTF8 enc/dec

digama0:
- [x] Depends on: #3958 
- [x] Depends on: #3960

This makes the UTF-8 encode and decode functions have lean definitions, so that we can prove properties about them downstream.

### DIFF
--- a/src/Init/Data/String/Extra.lean
+++ b/src/Init/Data/String/Extra.lean
@@ -26,7 +26,7 @@ def utf8DecodeChar? (a : ByteArray) (i : Nat) : Option Char := do
     guard (c1 &&& 0xc0 == 0x80)
     let r := ((c &&& 0x1f).toUInt32 <<< 6) ||| (c1 &&& 0x3f).toUInt32
     guard (0x80 ≤ r)
-    -- FIXME: this check is not needed
+    -- TODO: Prove h from the definition of r once we have the necessary lemmas
     if h : r < 0xd800 then some ⟨r, .inl h⟩ else none
   else if c &&& 0xf0 == 0xe0 then
     let c1 ← a[i+1]?
@@ -37,7 +37,7 @@ def utf8DecodeChar? (a : ByteArray) (i : Nat) : Option Char := do
       ((c1 &&& 0x3f).toUInt32 <<< 6) |||
       (c2 &&& 0x3f).toUInt32
     guard (0x800 ≤ r)
-    -- FIXME: the `r < 0x110000` check is not needed
+    -- TODO: Prove `r < 0x110000` from the definition of r once we have the necessary lemmas
     if h : r < 0xd800 ∨ 0xdfff < r ∧ r < 0x110000 then some ⟨r, h⟩ else none
   else if c &&& 0xf8 == 0xf0 then
     let c1 ← a[i+1]?


semorrison:
Could you add a basic test file using `test_extern`, e.g. starting with

```
import Lean.Util.TestExtern

instance : BEq ByteArray where
  beq x y := x.data == y.data

test_extern String.toUTF8 "$£€𐍈"
```
and then similarly for the other functions?

semorrison (approved):

semorrison:
LGTM, but I'm not the one to merge changes to UTF8 handling!

tests/lean/run/utf8英語.lean
```diff
@@ -1,3 +1,47 @@
...
+instance : BEq ByteArray where
+  beq x y := x.data == y.data
+
+test_extern String.toUTF8 ""
+test_extern String.toUTF8 "\x00"
+test_extern String.toUTF8 "$£€𐍈"
+
+macro "test_extern'" t:term " => " v:term : command =>
+  `(test_extern $t
+    #guard $t == $v)
```
nomeata, line 12: Can we have this next to `test_extern` with a more helpful name? Maybe
`guard_extern`? Or simply `test_extern` and using `=>` to enable the additional guard?

src/Init/Data/String/Extra.lean
```diff
@@ -17,13 +17,69 @@ def toNat! (s : String) : Nat :=
...
+  let c ← a[i]?
+  if c &&& 0x80 == 0 then
+    some ⟨c.toUInt32, .inl (Nat.lt_trans c.1.2 (by decide))⟩
+  else if c &&& 0xe0 == 0xc0 then
+    let c1 ← a[i+1]?
+    guard (c1 &&& 0xc0 == 0x80)
+    let r := ((c &&& 0x1f).toUInt32 <<< 6) ||| (c1 &&& 0x3f).toUInt32
+    guard (0x80 ≤ r)
+    -- FIXME: this check is not needed
+    if h : r < 0xd800 then some ⟨r, .inl h⟩ else none
```
nomeata, line 30: Why is it not? Why is it still there then?

tests/lean/run/utf8英語.lean
nomeata, line 12: Should we have this next to `test_extern` with a more helpful name? Maybe
`guard_extern`? Or simply use `test_extern` and look for the `=>` syntax to enable the additional guard?

nomeata (commented):
Good stuff, minor comments only.

src/Init/Data/String/Extra.lean
digama0: it's possible to prove from the previous line that `r` is at most 2^11, which is less than 0xd800. But proving those kind of facts about bitwise ops is not easy, especially in this very early file.

tests/lean/run/utf8英語.lean
digama0: I tried to sneak this feature request in https://github.com/leanprover/lean4/pull/3075#issuecomment-2068637377 . It's a separate PR though in any case.

src/Init/Data/String/Extra.lean
nomeata: Ah, now I get it. How about

```suggestion
    -- We could prove h from the definition of r, but it is too tedious,
    -- so for now we just check it again.
    if h : r < 0xd800 then some ⟨r, .inl h⟩ else none
```

digama0:
```suggestion
    -- TODO: Prove h from the definition of r once we have the necessary lemmas
    if h : r < 0xd800 then some ⟨r, .inl h⟩ else none
```

src/Init/Data/String/Extra.lean
```diff
@@ -17,13 +17,69 @@ def toNat! (s : String) : Nat :=
...
+  else if c &&& 0xf0 == 0xe0 then
+    let c1 ← a[i+1]?
+    let c2 ← a[i+2]?
+    guard (c1 &&& 0xc0 == 0x80 && c2 &&& 0xc0 == 0x80)
+    let r :=
+      ((c &&& 0x0f).toUInt32 <<< 12) |||
+      ((c1 &&& 0x3f).toUInt32 <<< 6) |||
+      (c2 &&& 0x3f).toUInt32
+    guard (0x800 ≤ r)
+    -- FIXME: the `r < 0x110000` check is not needed
```
nomeata, line 40:
```suggestion
    -- TODO: Prove h from the definition of r once we have the necessary lemmas
```

nomeata (approved):
Looks good. `test_extern` improvement can be generalized in a separate PR. Happy to merge if you are happy with this.

src/Init/Data/String/Extra.lean
digama0:
```suggestion
    -- TODO: Prove `r < 0x110000` from the definition of r once we have the necessary lemmas
```

PR #3962, opened 2024-04-20
fix: solveByElim would add symm hypotheses to local context and make impossible-to-elaborate terms

### LINKED ISSUE #3922: Interaction with `symm` attribute makes `apply?` nearly unusable
### Prerequisites

### Description

If there are hypotheses in the context involving relations marked with the `symm` attribute, `apply?` does not work as expected, degrading its usefulness substantially:
* its suggestions contain expresions like `(?_ (id (r.symm h₁)))`, which produce an error when used to replace the call to `apply?`, so the resulting code has to be fixed manually;
* the list of resulting subgoals contains many spurious entries (usually of the type `h → g` when `g` is already present as a subgoal);
* a side effect of the latter is that it blows up the amount of text in the infoview with the result that VSCode reacts extremely slowly to keystrokes/mouse wheel when scrolling. **EDIT:** This seems to have been exacerbated by a different problem that was fixed in [vscode-lean#432](https://github.com/leanprover/vscode-lean4/pull/432).

### Context

Since equality is a `symm` relation, this issue occurs whenever one tries to use `apply?` when there are equality hypotheses around (which is very common).

See [this Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/apply.3F.20and.20the.20.22symm.22.20attribute/near/433120818) (in particular the poll there) and also [this one](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/apply.3F.20suggested.20something.20that.20led.20to.20an.20error/near/432933179).

### Steps to Reproduce

Here is an MWE showing the effect of marking a relation `symm` on `apply?`:
```lean
-- set up a binary relation
axiom r : Nat → Nat → Prop

-- that is symmetric
theorem r.symm {a b : Nat} : r a b → r b a := sorry

-- and has some other property
theorem r.trans {a b c : Nat} : r a b → r b c → r a c := sorry

-- now try `apply?`
example (a b c : Nat) (h₁ : r b a) (h₂ : r b c) : r c a := by
  apply?
/-
Try this: refine r.trans ?_ ?_
Remaining subgoals:
⊢ Nat
⊢ r c ?b
⊢ r ?b a

Try this: refine r.symm ?_
Remaining subgoals:
⊢ r a c

which looks fine.
-/

-- now attach the `symm` attribute to `r.symm`
attribute [symm] r.symm

-- and repeat
example (a b c : Nat) (h₁ : r b a) (h₂ : r b c) : r c a := by
  -- `exact?` → `exact?` could not close the goal.
  -- I think that attaching `symm` *should* enable `exact?` to come up with
  -- `exact r.trans h₂.symm h₁` or `exact (r.trans h₁.symm h₂).symm`
  apply?
/-
Try this: refine r.trans (?_ (id (r.symm h₁))) (?_ (id (r.symm h₁)))
Remaining subgoals:
⊢ r a b → Nat
⊢ r a b → r c (?m.239 ⋯)
⊢ r a b → r (?m.239 ⋯) a
⊢ r c b → Nat
⊢ Nat
⊢ r c b → r c (?m.239 ⋯)
⊢ r c (?m.239 ⋯)
⊢ r c b → r (?m.239 ⋯) a
⊢ r (?m.239 ⋯) a

Try this: refine r.symm (r.trans (?_ (id (r.symm h₁))) (?_ (id (r.symm h₁))))
Remaining subgoals:
⊢ r a b → Nat
⊢ r a b → r a (?m.331 ⋯)
⊢ r a b → r (?m.331 ⋯) c
⊢ r c b → Nat
⊢ Nat
⊢ r c b → r a (?m.331 ⋯)
⊢ r a (?m.331 ⋯)
⊢ r c b → r (?m.331 ⋯) c
⊢ r (?m.331 ⋯) c

Try this: refine r.symm (?_ (id (r.symm h₁)))
Remaining subgoals:
⊢ r a b → r a c
⊢ r c b → r a c
⊢ r a c

Try this: refine r.symm (r.symm (?_ (id (r.symm h₁))))
Remaining subgoals:
⊢ r a b → r c a
⊢ r c b → r c a

... and clicking on any of these results in an error of the kind
  function expected at
    ?m.5099
  term has type
    ?m.5098
-/
```

**Expected behavior:**
`apply?` should make suggestions (including applying symmetry lemmas) that work without manual editing when clicked on (in particular, without `?_` in function positions) and give the actual list of subgoals generated.

**Actual behavior:**
In the presence of `symm` relations in the context, `apply?` produces suggestions involving `?_` in function positions (hence leading to an error when transferred to the code) and bloats the list of remaining subgoals with redundant statements. See [here](https://leanprover.zulipchat.com/#narrow/stream/113489-new-members/topic/apply.3F.20suggested.20something.20that.20led.20to.20an.20error/near/433036405) for a particulary egregious example (this is just the first suggestion shown).

### Versions

4.7.0 / Linux (my laptop). 4.7.0 on https://live.lean-lang.org

### Impact

This bug makes `apply?` almost unusable in most real-life situations, which would otherwise be a very helpful tool.
See [this poll](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/apply.3F.20and.20the.20.22symm.22.20attribute/near/433303853).

kmill:
Rather than adding symm hypotheses to the local context, it now adds them to the list of hypotheses derived from the local context.

This is not ideal for performance reasons, but it at least closes #3922.

In the future, solveByElim could maintain its own cache of facts that it updates whenever it does intro.

### DIFF
--- a/src/Lean/Meta/Tactic/SolveByElim.lean
+++ b/src/Lean/Meta/Tactic/SolveByElim.lean
@@ -68,7 +68,7 @@ We use this in `apply_rules` and `apply_assumption` where backtracking is not ne
 -/
 def applyFirst (cfg : ApplyConfig := {}) (transparency : TransparencyMode := .default)
     (lemmas : List Expr) (g : MVarId) : MetaM (List MVarId) := do
-  (←applyTactics cfg transparency lemmas g).head
+  (← applyTactics cfg transparency lemmas g).head
 
 open Lean.Meta.Tactic.Backtrack (BacktrackConfig backtrack)
 
@@ -168,7 +168,7 @@ applied to the instantiations of the original goals, fails or returns `false`.
 def testPartialSolutions (cfg : SolveByElimConfig := {}) (test : List Expr → MetaM Bool) : SolveByElimConfig :=
   { cfg with
     proc := fun orig goals => do
-      let .true ← test (← orig.mapM fun m => m.withContext do instantiateMVars (.mvar m)) | failure
+      guard <| ← test (← orig.mapM fun m => m.withContext do instantiateMVars (.mvar m))
       cfg.proc orig goals }
 
 /--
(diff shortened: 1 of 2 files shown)


kmill:
@semorrison I took a stab at this issue. It's likely a performance regression, but it's a big usability improvement, so maybe it's ok in the meantime?

semorrison:
Adding `full-ci` so I can look at Mathlib.

semorrison:
Mathlib is easy enough to fix (pushed to lean-pr-testing-3962). Agree we need to fix this despite the possible performance problems.

Issue #1707, opened 2022-10-08
PANIC at Option.get! Init.Data.Option.BasicAux:16:14: value is none


dwrensha:
Lean panics on the input `def y := c.5�` where � signifies the non-utf8 byte `0xff`.

```
$ echo "ZGVmIHkgOj0gYy41/wo=" | base64 -d > panic.lean
$ cat panic.lean
def y := c.5�
$ ~/src/lean4/build/release/stage1/bin/lean --version
Lean (version 4.0.0, commit e3ec468e3bfe, Release)
$ ~/src/lean4/build/release/stage1/bin/lean panic.lean
PANIC at Option.get! Init.Data.Option.BasicAux:16:14: value is none
backtrace:
/home/dwrensha/src/lean4/build/release/stage1/bin/../lib/lean/libleanshared.so(+0x32f432c)[0x7ff39ccf432c]
/home/dwrensha/src/lean4/build/release/stage1/bin/../lib/lean/libleanshared.so(lean_panic_fn+0xc4)[0x7ff39ccf4694]
/home/dwrensha/src/lean4/build/release/stage1/bin/../lib/lean/libleanshared.so(l___private_Lean_Elab_App_0__Lean_Elab_Term_elabAppFn+0x9f3)[0x7ff39be219a3]
/home/dwrensha/src/lean4/build/release/stage1/bin/../lib/lean/libleanshared.so(l___private_Lean_Elab_App_0__Lean_Elab_Term_elabAppAux+0x160)[0x7ff39be32440]
/home/dwrensha/src/lean4/build/release/stage1/bin/../lib/lean/libleanshared.so(l___private_Lean_Elab_App_0__Lean_Elab_Term_elabAtom+0x134)[0x7ff39be35fb4]
/home/dwrensha/src/lean4/build/release/stage1/bin/../lib/lean/libleanshared.so(lean_apply_9+0x961)[0x7ff39cd0df81]
/home/dwrensha/src/lean4/build/release/stage1/bin/../lib/lean/libleanshared.so(l___private_Lean_Elab_Term_0__Lean_Elab_Term_elabUsingElabFnsAux___lambda__1+0x174)[0x7ff39b4edaf4]
/home/dwrensha/src/lean4/build/release/stage1/bin/../lib/lean/libleanshared.so(lean_apply_7+0xd1e)[0x7ff39cd0aa2e]
[...]

Issue #1690, opened 2022-10-05
Lean fails to terminate on non-utf8 input


dwrensha:
I'm seeing `lean` fail to terminate on the input `def foo :=� `, where `�` signifies the non-utf8 byte 0xf3.
```
$ echo ZGVmIGZvbyA6PfMg | base64 -d > hang.lean
$ xxd hang.lean 
00000000: 6465 6620 666f 6f20 3a3d f320            def foo :=. 
$ ~/.elan/toolchains/leanprover--lean4---nightly-2022-10-05/bin/lean hang.lean 
... [hangs forever]
```
I expect `lean` to terminate on this input, ideally with some error about non-utf8 input.

I tried running some `lean` nightly back to mid January 2021, and none of them terminated within a reasonable amount of time.

(I'm running on Linux.)

Issue #301, opened 2021-01-29
Stack overflow on an input file that is not valid UTF-8


TwoFX:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

Here is a hex dump of an input file that causes a stack overflow when fed to the `lean` executable:
```
00000000: 6465 6620 8828 783a 3d66 756e 2030 3d3e  def .(x:=fun 0=>
00000010: 3029 6e3a 3dff 2080 0a                   0)n:=. ..
```

### Steps to Reproduce
The input file can be downloaded [here](https://github.com/leanprover/lean4/files/5891754/test.txt) or recovered from the hex dump above via `xxd -r test.lean.xxd > test.lean` after saving the hex dump in a file called `test.lean.xxd`.

**Expected behavior:** An error message

**Actual behavior:**
```
$ lean test.lean

Stack overflow detected. Aborting.
Aborted
```

**Reproduces how often:** 100%

### Versions
```
Lean (version 4.0.0-nightly-2021-01-29, commit 5a02c4facc06, Release)
```
Ubuntu 20.04

PR #3963, opened 2024-04-20
fix: validate UTF-8 at C++ -> Lean boundary

digama0:
Continuation of #3958. To ensure that lean code is able to uphold the invariant that `String`s are valid UTF-8 (which is assumed by the lean model), we have to make sure that no lean objects are created with invalid UTF-8. #3958 covers the case of lean code creating strings via `fromUTF8Unchecked`, but there are still many cases where C++ code constructs strings from a `const char *` or `std::string` with unclear UTF-8 status.

To address this and minimize accidental missed validation, the `(lean_)mk_string` function is modified to validate UTF-8. The original function is renamed to `mk_string_unchecked`, with several other variants depending on whether we know the string is UTF-8 or ASCII and whether we have the length and/or utf8 char count on hand. I reviewed every function which leads to `mk_string` or its variants in the C code, and used the appropriate validation function, defaulting to `mk_string` if the provenance is unclear.

This PR adds no new error handling paths, meaning that incorrect UTF-8 will still produce incorrect results in e.g. IO functions, they are just not causing unsound behavior anymore. A subsequent PR will handle adding better error reporting for bad UTF-8.

### DIFF
--- a/src/runtime/object.cpp
+++ b/src/runtime/object.cpp
@@ -2046,9 +2046,10 @@ extern "C" LEAN_EXPORT obj_res lean_string_utf8_set(obj_arg s, b_obj_arg i0, uin
     std::string tmp;
     push_unicode_scalar(tmp, c);
     std::string new_s = string_to_std(s);
+    usize len = lean_string_len(s);
     dec(s);
     new_s.replace(i, get_utf8_char_size_at(new_s, i), tmp);
-    return lean_mk_string_unchecked(new_s.data(), new_s.size(), lean_string_len(s));
+    return lean_mk_string_unchecked(new_s.data(), new_s.size(), len);
 }
 
 extern "C" LEAN_EXPORT uint64 lean_string_hash(b_obj_arg s) {


nomeata (approved):

digama0:
!bench

digama0:
(disabled auto-merge to run a benchmark first)

digama0:
The affected tests are regression tests from #301, #1707, #1690, all of which appear to have been strange behaviors of lean caused by invalid UTF-8. The new behavior is for the C code to throw an unspecific error before even calling lean code; this is inevitable because the lean parser/elaborator assumes it can store the file in a `String` so we can't really use most of lean's error handling.

src/include/lean/lean.h
```diff
@@ -2064,11 +2066,11 @@ static inline uint8_t lean_version_get_is_release(lean_obj_arg _unit) {
 }
 
 static inline lean_obj_res lean_version_get_special_desc(lean_obj_arg _unit) {
-    return lean_mk_string(LEAN_SPECIAL_VERSION_DESC);
+    return lean_mk_ascii_string(LEAN_SPECIAL_VERSION_DESC);
```
nomeata, line 2069: I’d probably lean towards using `mk_utf8_string` here. It’s probably not performance critical, and maybe some German developer with a ü in their name builds lean from a branch with `v0.7.8-Günnar`….
```suggestion
    return lean_mk_utf8_string(LEAN_SPECIAL_VERSION_DESC);
```

src/include/lean/lean.h
```diff
@@ -2064,11 +2066,11 @@ static inline uint8_t lean_version_get_is_release(lean_obj_arg _unit) {
...
 }
 
 static inline lean_obj_res lean_system_platform_target(lean_obj_arg _unit) {
-    return lean_mk_string(LEAN_PLATFORM_TARGET);
+    return lean_mk_ascii_string(LEAN_PLATFORM_TARGET);
```
nomeata, line 2073:
```suggestion
    return lean_mk_utf8_string(LEAN_PLATFORM_TARGET);
```
Unlikely to matter, but better safe than sorry.

src/runtime/platform.cpp
```diff
@@ -40,7 +40,7 @@ extern "C" LEAN_EXPORT uint8 lean_system_platform_emscripten(obj_arg) {
 #endif
 }
 
-extern "C" object * lean_get_githash(obj_arg) { return lean_mk_string(LEAN_GITHASH); }
+extern "C" object * lean_get_githash(obj_arg) { return lean_mk_ascii_string(LEAN_GITHASH); }
```
nomeata, line 43: Also unlikely, but since the source is out of our (local) control, better safe than sorry.
```suggestion
extern "C" object * lean_get_githash(obj_arg) { return lean_mk_utf8_string(LEAN_GITHASH); }
```

src/include/lean/lean.h
digama0: well, the "better safe than sorry" option is `lean_mk_string`. If we have any reason to believe that the data isn't necessarily in the subset it appears to be in, we should just validate it.

nomeata: Indeed, after I wrote this and was away from the computer I paused wondered if my suggestion is bogus.

It’s not entirely bogus, though: If the there is utf8 in the TARGET, then `mk_ascii_string` will produce an invalid length. So if `LEAN_SPECIAL_VERSION_DESC` etc. happens to be utf8, then `mk_utf8_string` is better than `mk_ascii_string`, but not as safe as a validating function.

I agree that it’s worth changing the `lean_mk_{utf8,utf8}_string` function names to include `unchecked to avoid future bugs.

src/include/lean/lean.h
```diff
@@ -991,6 +991,8 @@ static inline size_t lean_string_byte_size(lean_object * o) { return sizeof(lean
 /* instance : inhabited char := ⟨'A'⟩ */
 static inline uint32_t lean_char_default_value() { return 'A'; }
 LEAN_EXPORT lean_obj_res lean_mk_string_from_bytes(char const * s, size_t sz);
+LEAN_EXPORT lean_obj_res lean_mk_ascii_string(char const * s);
```
Kha, line 994: Are we using this in any performance-critical spot?

src/include/lean/lean.h
Kha: Agreed on better names

src/runtime/object.cpp
```diff
@@ -1611,9 +1611,23 @@ extern "C" LEAN_EXPORT object * lean_mk_string_from_bytes(char const * s, size_t
 }
 
 extern "C" LEAN_EXPORT object * lean_mk_string(char const * s) {
+    size_t len = strlen(s);
+    if (validate_utf8((const uint8_t *)s, len)) {
```
Kha, line 1615: Why is this a separate step, can't it be combined with `utf8_strlen`? Do we even need the mk_string/mk_utf8_string distinction in that case?

src/runtime/object.cpp
```diff
@@ -1611,9 +1611,23 @@ extern "C" LEAN_EXPORT object * lean_mk_string_from_bytes(char const * s, size_t
...
+        return lean_mk_string_from_bytes(s, len);
+    } else {
+        return lean_mk_string_core("", 0, 0);
```
Kha, line 1618: Is this really a sensible fallback? I matches the panic value but there is no message. Shouldn't this raise an exception if done as part of I/O?

src/include/lean/lean.h
nomeata: I introduced that function for Nat.repr

src/runtime/object.cpp
digama0: Not really. I was planning to put in a follow-up PR which does something like rust's `String::from_utf8_lossy` with replacement characters. In most cases it's not really a good idea to rely on this for error handling, and instead code should call `validate` separately and throw IO errors etc as appropriate.

src/runtime/object.cpp
digama0: We can do `utf8_strlen` at the same time as validation, but I think there are still use cases for "I know this is UTF8", in particular when doing the check beforehand so as to give a custom error message. Also, from the looks of it `utf8_strlen` will be optimized to a faster function than `validate_utf8`, so there is likely some performance advantage to using only `utf8_strlen` when there are reasons to believe the input is already utf8. This happens in `String.set` for example (although actually in that example we can do even better since we know what the length is already).

src/runtime/object.cpp
Kha: I think it would be good to elaborate on this plan before going forward. It would also be helpful to mention such relevant plans in the PR description :) .

If we want to get to the point where no-one, especially not Lean APIs, depends on this branch, should it become a hard assertion?

digama0: I have a revision in the works which addresses many of the review comments. I'll post that and then we can talk about how we want to proceed here.

digama0:
I pushed a new version of the PR. Summary of changes:

* `lean_mk_string_unchecked(const char * s, size_t sz, size_t utf8_chars)` was previously `lean_mk_string_core`
  * Use this when we have both the length and utf8_length values available and also know the input is valid UTF-8.
* `lean_mk_string_from_bytes_unchecked(const char * s, size_t sz)` was `lean_mk_string_from_bytes`
  * Use this when we have the length and know the input is valid UTF-8, but need to compute the utf8_chars.
* `lean_mk_string_from_bytes(const char * s, size_t sz)` does validation now, and if validation fails it converts invalid UTF-8 sequences to `U+FFFD` instead of just returning the empty string.
  * Use this when we have a (char*, len) pair but do not know it encodes valid UTF-8
* `lean_mk_string(const char * s)` is a wrapper around `lean_mk_string_from_bytes`
  * Use this when we have char*, not necessarily UTF-8 and need `strlen` to get the length
* `lean_string_from_utf8` was renamed to `lean_string_from_utf8_unchecked` 
  * This is the implementation of `String.fromUTF8`, you should probably not call it directly. I added `_unchecked` to the name because the C version doesn't validate, even though the lean version takes a validity proof witness as an argument.
* `lean_mk_ascii_string_unchecked(char const * s)` was `lean_mk_ascii_string`
  * Use this when you know the string is valid ASCII but need `strlen` to get the length.
* `lean::mk_ascii_string_unchecked(std::string const & s)` was `lean::mk_ascii_string`
  * Use this when you know the string is valid ASCII and have a `std::string` (which counts the length).

Other (non-API) changes:

* The code generator produces calls to `lean_mk_string_unchecked` now instead of `lean_mk_string_from_bytes`, since we don't want to validate string literals, plus we can calculate the utf8_chars in advance.
* ` bool validate_utf8(uint8_t const * str, size_t size, size_t & pos, size_t & i)` also returns the byte position of the first bad character, which can be used for better error reporting in addition to its use here as a fast path for the replacement character algorithm.
* I removed the special error handling for invalid UTF-8 in lean files. These cases, in addition to the IO uses of `mk_string`, are handled by the replacement algorithm, meaning that lean will complain in the bad UTF-8 tests with a "more normal" error saying that `U+FFFD` is not a token; if you stick it inside a comment or guillemets it will be accepted like any other unicode character. This is not necessarily a good choice, but at least it's a less damaging default behavior than erasing the whole string. You can however do separate validation, by calling `validate_utf8` and `lean_mk_string_unchecked` instead of `lean_mk_string`.

nomeata:
Sounds like a nice and thorough cleanup, thanks a lot!

I'd lean towards rejecting invalid utf8 in lean files (not replacing, not using the empty string). Who knows what kind of surprises will occur when people accidentally have invalid utf8 and lean does something, but not what they expect to.

Kha:
I'm happy as well modulo below -

> I'd lean towards rejecting invalid utf8 in lean files (not replacing, not using the empty string)

Agreed, and more generally I/O functions should at least by default throw an error when decoding fails. I'm pretty sure this is the default in all the big stdlibs out there as well, no?

digama0:
I agree, but can we agree that this is an improvement on the status quo and do that in a follow up PR? Handling UTF-8 errors properly means having new kinds of error reporting functions we don't have currently, as well as touching a bunch more code.

Kha:
Sure, I just want to make sure it is documented and discussed that this is a partial solution in that way. That is very relevant context for reviewing.

Kha (approved):

nomeata:
Sounds like as soon as the PR description reflects the current state I can press the button?

digama0:
PR description updated

nomeata:
Linux-debug fails with 
```
1658/1862 Test #1661: leanruntest_subset.lean ...................................***Failed    1.07 sec
../../common.sh: line 42: 48489 Aborted                 (core dumped) LEAN_BACKTRACE=0 "$@" 2>&1
     48490 Done                    | perl -pe 's/(\?(\w|_\w+))\.[0-9]+/\1/g' > "$f.produced.out"
Unexpected return code 134 executing 'lean -Dlinter.all=false subset.lean'; expected 0. Output:
LEAN ASSERTION VIOLATION
File: /home/runner/work/lean4/lean4/build/stage1/include/lean/lean.h
Line: 473
lean_is_string(o)
(C)ontinue, (A)bort/exit, (S)top/trap
```

Kha:
@digama0 Do you have time to look into the assertion failure?

digama0:
Sorry for the delay, the assertion failure should be fixed now.

PR #3967, opened 2024-04-22
chore: add dates to @[deprecated] attributes

kim-em:

PR #3968, opened 2024-04-22
feat: improve @[deprecated] attr

digama0:
Complement to #3967 , adds a `(since := "<date>")` field to `@[deprecated]` so that metaprogramming code has access to the deprecation date for e.g. bulk removals. Also adds `@[deprecated "deprecation message"]` to optionally replace the default text "`{declName}` has been deprecated, use `{newName}` instead".

### DIFF
--- a/doc/examples/Certora2022/ex8.lean
+++ b/doc/examples/Certora2022/ex8.lean
@@ -4,16 +4,16 @@ def ack : Nat → Nat → Nat
   | 0,   y   => y+1
   | x+1, 0   => ack x 1
   | x+1, y+1 => ack x (ack (x+1) y)
-termination_by ack x y => (x, y)
+termination_by x y => (x, y)
 
 def sum (a : Array Int) : Int :=
   let rec go (i : Nat) :=
-     if i < a.size then
+     if _ : i < a.size then
         a[i] + go (i+1)
      else
         0
+  termination_by a.size - i
   go 0
-termination_by go i => a.size - i
 
 set_option pp.proofs true
 #print sum.go
--- a/doc/examples/ICERM2022/ctor.lean
+++ b/doc/examples/ICERM2022/ctor.lean
@@ -4,43 +4,42 @@ open Lean Meta
 
 def ctor (mvarId : MVarId) (idx : Nat) : MetaM (List MVarId) := do
   /- Set `MetaM` context using `mvarId` -/
-  withMVarContext mvarId do 
+  mvarId.withContext do
     /- Fail if the metavariable is already assigned. -/
-    checkNotAssigned mvarId `ctor
+    mvarId.checkNotAssigned `ctor
     /- Retrieve the target type, instantiateMVars, and use `whnf`. -/
-    let target ← getMVarType' mvarId
+    let target ← mvarId.getType'
     let .const declName us := target.getAppFn
       | throwTacticEx `ctor mvarId "target is not an inductive datatype"
     let .inductInfo { ctors, .. } ← getConstInfo declName
       | throwTacticEx `ctor mvarId "target is not an inductive datatype"
     if idx = 0 then
-      throwTacticEx `ctor mvarId "invalid index, it must be > 0"  
+      throwTacticEx `ctor mvarId "invalid index, it must be > 0"
     else if h : idx - 1 < ctors.length then
-      apply mvarId (.const ctors[idx - 1] us)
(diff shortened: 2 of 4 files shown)


semorrison:
Thanks. Could you add two tests for `tests/lean/deprecated.lean` for the alt-text (it's fine to use `#guard_msgs` there rather than having to update expected.out) and validating that `(since := ...)` is accepted?

semorrison:
I've pulled #3967 off the queue and will update it once this is in.

digama0:
The manual is failing and I have no idea how to track down errors in that build step...

digama0:
@Kha It appears that alectryon is failing because it misses a [bug fix](https://github.com/cpitclaudel/alectryon/commit/55340d4e61a54ba2c55a67fa3b57873193c6db02) not present on the [lean4 fork of alectryon](https://github.com/insightmind/alectryon/tree/lean4). (There is likely another error behind this one, but the one shown in the [log](https://github.com/leanprover/lean4/actions/runs/8778666534/job/24085482573) seems to be this issue.)

Kha:
Can we somehow survive for a little longer without touching Alectryon?

digama0:
Could there be some documentation for how to build the manual? All I see is some nix stuff I don't understand and a broken error message due to a bug in an abandoned dependency. I don't even know what sources are being built, it's loading some SHAs from the nix cache?

src/Init/Notation.lean
```diff
@@ -492,9 +492,12 @@ The attribute `@[deprecated]` on a declaration indicates that the declaration
...
 existing code. It may be removed in a future version of the library.
 
-`@[deprecated myBetterDef]` means that `myBetterDef` is the suggested replacement.
+* `@[deprecated myBetterDef]` means that `myBetterDef` is the suggested replacement.
+* `@[deprecated myBetterDef "use myBetterDef instead"]` allows customizing the deprecation message.
+* `@[deprecated (since := "2024-04-21")]` records when the deprecation was first applied.
 -/
-syntax (name := deprecated) "deprecated" (ppSpace ident)? : attr
+syntax (name := deprecated) "deprecated" (ppSpace ident)? (ppSpace str)?
+    (" (" &"since" " := " str ")")? : attr
```
eric-wieser, line 500: Can you declare `since` as a separate syntax so that it gets its own docstring (which can recommend the date format)

Kha:
> error: builder for '/nix/store/wgi78yspl1dxsx0bkf020v24cbm5fzfh-examples.ICERM2022.ctor.md.drv' failed with exit code 1;

Let's please just fix all deprecation warnings in ICERM2022/ctor.lean and then document everything about the manual when it's redone in Verso

digama0:
> Let's please just fix all deprecation warnings in ICERM2022/ctor.lean and then document everything about the manual when it's redone in Verso

How did you manage to deduce that this was the problem? That was the issue I was having, the error message is completely opaque.

src/Init/Notation.lean
digama0: I think it is better to have this documentation on `deprecated`, because otherwise people won't necessarily know that the clause exists in the first place. The example in `deprecated` gives at least a bare hint of what the date format should be.

Kha:
My bad, I thought the line I quoted would clarify how I found it but I guess it's still pretty opaque. If you look at the failing log (which I can't see anymore because of the rebase), you will see that Nix store path I quoted mentioned. If you ignore the hash and file system parts, the relevant part is `examples.ICERM2022.ctor.md` - this is a mangled string of the "build step" (derivation) name that ultimately is derived from the output file name, examples/ICERM2022/ctor.md. Does that make sense? Why it's the deprecations specifically I have no idea but looking at the file and considering the PR at hand that was the only logical conclusion.

> Otherwise it should just be removed from the build?

If it continues to be an issue, we should just downgrade these examples to basic non-interactive code blocks, yes. But this was the first (perplexing) failure in that part for a long while.

digama0:
> Why it's the deprecations specifically I have no idea but looking at the file and considering the PR at hand that was the only logical conclusion.

Yes, I don't understand this either. It does appear that the deprecations are responsible, in the sense that removing the deprecated uses fixed the build, but the code here doesn't touch deprecation reporting at all, and that file contains no deprecated attributes. My best guess is that there are some stage0 shenanigans going on whereby the lean compiler is either reading a deprecation with the wrong syntax or reading an olean deprecation entry as the old type or vice versa, resulting in a failure of the deprecation linter to report errors correctly.

PR #3970, opened 2024-04-22
feat: monadic generalization of FindExpr

kim-em:
Not certain this is a good idea. Motivated by code duplication introduced in #3398.


semorrison:
!bench

leodemoura:
This generalization is problematic. For example, the monad may have state, and this state may influence the result of `p`. The reference implementation does not match the actual one because of the `checkVisited` cache.

PR #3971, opened 2024-04-22
chore: prevent stage0 changes via the merge queue

nomeata:
these need manual rebase merges by an admin, so lets prevent accidential
merges via the squashing merge queue.

### DIFF
--- a/.github/workflows/check-stage0.yml
+++ b/.github/workflows/check-stage0.yml
@@ -1,7 +1,7 @@
 name: Check for stage0 updates on the merge queue
 
 # We only want to run this on merge_group checks, but github uses the same
-# list of requires checks for PR → queu and queue → master, so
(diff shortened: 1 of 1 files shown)


nomeata:
@kha This is the first line of defense against squshed stage0 updates. Needs to also be added to the required checks in the branch protection rules.

.github/workflows/check-stage0.yml
```diff
@@ -0,0 +1,27 @@
+name: Check for stage0 updates on the merge queue
+
+# We only want to run this on merge_group checks, but github uses the same
+# list of requires checks for PR → queu and queue → master, so
```
Kha, line 4:
```suggestion
# list of requires checks for PR → queue and queue → master, so
```

PR #3973, opened 2024-04-23
feat: use `rfl` proofs of `Iff` in `dsimp`

thorimur:
Retrying #2679 from scratch. WIP.


thorimur:
release-ci

nomeata:
Ah, here is another attempt at this. Same question: what's the status of this?

thorimur:
> Ah, here is another attempt at this. Same question: what's the status of this?

See [this zulip thread](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/using.20.60Iff.60.20rfl.20proofs.20in.20.60dsimp.60)! The long and short of it is that this causes *lots* of breakages, and asking people to try `-dsimp` when their `simp` call doesn't work seems unacceptable to me. I haven't had much time in the past couple months, but I posted a few thoughts on ways forward I'm planning to explore in that thread.

PR #3974, opened 2024-04-23
feat: `exact?%`: do not report suggestions which do not close the goal

bustercopley:
This makes `exact?%` behave like `by exact?` rather than `by apply?`.

If the underlying function `librarySearch` finds a suggestion which closes the goal, use it (and add a code action). Otherwise log an error and use `sorry`. The error is either
```text
`exact?%` didn't find any relevant lemmas
```
or
```text
`exact?%` could not close the goal. Try `by apply` to see partial suggestions.
```

---

[Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Useful.20term.20elaborators/near/434863856)


semorrison:
I would prefer to simply delete `apply?%` (or turn it into something that just throws an error and suggests using `exact?%`), and change the failure message in `exact?%` to "Try `by apply?` to see partial suggestions".

It looks like you have a conflict and will need to rebase.

bustercopley:
Force push: rebased.

Thanks! I deleted `apply?%` (the name didn't exist before this PR). I don't think a stub would be much use for discoverability, as users will find `exact?` and `apply?` before `exact?%` and `apply?%`.

bustercopley:
awaiting review

(FOA Kim (@semorrison))

bustercopley:
awaiting-review

(trying again, to adjust the labels)

PR #3975, opened 2024-04-23
chore: upstream Std material from Data/List|Array/Init

kim-em:
See proposal on [zulip](https://leanprover.zulipchat.com/#narrow/stream/348111-std4/topic/upstreaming.20of.20List.2FArray.20material/near/434879041); I won't merge this until there's a chance for discussion there.


digama0 (approved):

semorrison:
I've narrowed the CODEOWNERS changes to just the files modified here. Ping me on zulip if there are other files containing earlier upstreamed material that we should add back in.

PR #3976, opened 2024-04-23
feat: Implement Functor for Array

alok:
Addresses https://leanprover-community.github.io/archive/stream/270676-lean4/topic/Functor.20Array.html and mirrors https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/List.20Functor/near/290370886

### DIFF
--- a/src/Init/Data/Array/Basic.lean
+++ b/src/Init/Data/Array/Basic.lean
@@ -501,6 +501,9 @@ def toList (as : Array α) : List α :=
 def toListAppend (as : Array α) (l : List α) : List α :=
   as.foldr List.cons l
 
+instance : Functor Array where
+  map := Array.map
+
 instance {α : Type u} [Repr α] : Repr (Array α) where
   reprPrec a _ :=
     let _ : Std.ToFormat α := ⟨repr⟩


semorrison:
I'm a little skeptical this is useful without some lemmas, despite there not being any for `List Functor`. Perhaps just the lemma that says `f <$> as = as.map f`, in case someone needs to rewrite by it to escape `Functor` land?

semorrison:
Can we have the `LawfulFunctor` instance as the same time?

semorrison:
Could you please rebase on `nightly-with-mathlib`, so CI will run against Std/Mathlib, and we can see if it has any downstream effects?

semorrison:
> working on adding lawful functor to array, did it for list. my main difficulty is that proving stuff for list is easier because of the inductive def, but array doesn't lend itself to that sort of recursive reasoning. tips?

If you're proving that two arrays are equal, you might try using that `Array.data` is injective, and then have a goal about lists (which might already be true via simp lemmas and the corresponding fact about lists).

How about post on zulip with an example of what you're trying to do? Easier to give advice there than here.

alok:
@ellipsis-dev

PR #3978, opened 2024-04-23
feat: make `pp.mvars false` pretty print universe mvars as `_`

kmill:
Suggestion on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/.23guard_msgs.20variant.3A.20don't.20care.20about.20whitespace/near/434906526)


semorrison:
Looks great, thanks!

PR #3980, opened 2024-04-23
feat: make `Level -> MessageData` coercion respect `pp.mvars`

kmill:
Adds `ppLevel` to the `PPFns` extension so that the coercion can pass the pretty printing context (including the `pp.mvars` option setting) to the `Level` formatter.


kmill:
@semorrison There were some extra places where level metavariables could be printed without respecting `pp.mvars`, and this PR should have caught the rest.

semorrison (approved):
LGTM

Kha (approved):

PR #3991, opened 2024-04-25
perf: use with_reducible in special-purpose decreasing_trivial macros

nomeata:
Because of the last-added-tried-first rule for macros, all the special
purpose `decreasing_trivial` rules are tried for most recursive
definitions out there, and because they use `apply` and `assumption`
with default transparency may cause some definitoins to be unfolded over
and over again.

A quick test with one of the functions in the leansat project shows that
elaboration time goes down from 600ms to 375ms when using
```
decreasing_by all_goals decreasing_with with_reducible decreasing_trivial
```
instead of
```
decreasing_by all_goals decreasing_with decreasing_trivial
```

This change uses `with_reducible` in most of these macros.

This means that these tactics will no longer work when the
relations/definitions they look for is hidden behind a definition.
This affected in particular `Array.sizeOf_get`, which now has a companion `sizeOf_getElem`.

In addition, there were three tactics using `apply` to apply Nat-related lemmas
that we now expect `omega` to solve. We still need them when building `Init` modules
that don’t have access to `omega`, but they now live in `decreasing_trivial_pre_omega`,
meant to be only used internally.

### DIFF
--- a/src/Init/Data/Array/Basic.lean
+++ b/src/Init/Data/Array/Basic.lean
@@ -31,6 +31,7 @@ def ofFn {n} (f : Fin n → α) : Array α := go 0 (mkEmpty n) where
   go (i : Nat) (acc : Array α) : Array α :=
     if h : i < n then go (i+1) (acc.push (f ⟨i, h⟩)) else acc
 termination_by n - i
+decreasing_by simp_wf; decreasing_trivial_pre_omega
 
 /-- The array `#[0, 1, ..., n - 1]`. -/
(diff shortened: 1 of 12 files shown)


nomeata:
http://speed.lean-fro.org/mathlib4/compare/d444b71b-0228-43f5-951d-975ddabbc565/to/77bd3441-1d67-4527-9a0c-188a17a83b8f 🥳

digama0 (approved):

PR #3992, opened 2024-04-26
chore: CI: pin macos-13

Kha:
macos-latest changed to arm64. It should not be hard to switch our setup to cross-compiling x64 instead of arm64 but let's get master green again first.


nomeata:
That unfortunately still failed

PR #3995, opened 2024-04-26
fix: mark well-founded definitions irreducible

### LINKED ISSUE #3988: Subst syntax unfolds recursive definition
This is something I have seen today during office hours, and is plausible the reason for significant slow downs.

Consider
```
def foo : Nat → Nat
  | 0 => 0
  | (n +1) => foo n + 1
termination_by n => n

example (h : n = m) (h2 : foo n > 42) : foo (m + 1) > 42 :=
  set_option trace.Meta.isDefEq true in
  (h ▸ h2)
```
The trace shows something like
```
[Meta.isDefEq] ❌ foo (n + 1) > 42 =?= foo n > 42 ▼
  [] ❌ foo (n + 1) =?= foo n ▼
    [] ❌ n + 1 =?= n ▶
    [] ❌ WellFounded.fix foo.proof_1
          (fun a a_1 =>
            (match (motive :=
                (x : Nat) → ((y : Nat) → (invImage (fun a => a) instWellFoundedRelation).1 y x → Nat) → Nat) a with
              | 0 => fun x => 0
              | Nat.succ n => fun x => x n ⋯ + 1)
              a_1)
          (n +
            1) =?= WellFounded.fix foo.proof_1
          (fun a a_1 =>
            (match (motive :=
                (x : Nat) → ((y : Nat) → (invImage (fun a => a) instWellFoundedRelation).1 y x → Nat) → Nat) a with
              | 0 => fun x => 0
              | Nat.succ n => fun x => x n ⋯ + 1)
              a_1)
          n ▶
  [] ❌ 42 < foo (n + 1) =?= 42 < foo n ▶
```
and I wonder if this means that `▶` is using `isDefEq` at the wrong reducibility setting.

Wrapping this in `by  with_reducible exact` avoids going near `WellFounded.fix`, as expected.

Maybe unrelated, but it seems it tries the same equality twice, ordered differently:
```
[Meta.isDefEq] ❌ foo (n + 1) > 42 =?= foo n > 42 ▶

[Meta.isDefEq] ❌ foo n > 42 =?= foo (n + 1) > 42 ▶
```

### Versions

4.7.0

### Impact

nomeata:
we keep running into examples where working with well-founded recursion
is slow becuase defeq checks (which are all over the place, including
failing ones that are back-tracked) unfold well-founded definitions.

The definition of a function defined by well-founded recursion should be
an implementation detail that should only be peeked inside by the
equation generator and the functional induction generator.

We now mark the mutual recursive function as irreducible, and use
`withAtLeastTransparency .all` when producing the equations.

This fixes #3988

### DIFF
--- a/tests/lean/run/casesRec.lean
+++ b/tests/lean/run/casesRec.lean
@@ -8,7 +8,7 @@ def f (x : Nat) : Nat := by
 
 #eval f 10
 
-example : f x.succ = 2 * f x := rfl
+example : f x.succ = 2 * f x := by rw [f]; rfl
 end Ex1
 
 namespace Ex2
--- a/tests/lean/run/defaultEliminator.lean
+++ b/tests/lean/run/defaultEliminator.lean
@@ -22,7 +22,7 @@ termination_by (x, y)
 
 example (x y : Nat) : f x y > 0 := by
   induction x, y with
-  | zero_zero => decide
+  | zero_zero => simp [f]
   | succ_zero x ih => simp [f, ih]
   | zero_succ y ih => simp [f, ih]
   | succ_succ x y ih => simp [f, ih]
--- a/tests/lean/run/lazylistThunk.lean
+++ b/tests/lean/run/lazylistThunk.lean
@@ -25,7 +25,7 @@ in the list, ignoring delays
 
 theorem length_toList (l : LazyList α) : l.toList.length = l.length := by
   match l with
-  | nil => rfl
+  | nil => simp [length_toList]
   | cons a as => simp [length_toList as]
   | delayed as => simp [length_toList as.get]
 
--- a/tests/lean/run/splitIssue.lean
+++ b/tests/lean/run/splitIssue.lean
@@ -15,23 +15,21 @@ termination_by l => l.length
 decreasing_by
   all_goals sorry
 
-theorem len_nil : len ([] : List α) = 0 := by
-  simp [len]
-
--- The `simp [len]` above generated the following equation theorems for len
+-- The equational theorems are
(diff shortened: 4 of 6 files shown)


nomeata:
So here is what breaks:


* A few tests ([`casesRec`](https://github.com/leanprover/lean4/blob/master/tests/lean/run/casesRec.lean), [`defaultEliminator`](https://github.com/leanprover/lean4/blob/master/tests/lean/run/defaultEliminator.lean),…) have `decide` or `rfl` proofs that had to be replaced by `rw [f]; rfl` or `simp [f]`, these look benign.

* [Test case for](https://github.com/leanprover/lean4/blob/master/tests/lean/run/2389.lean) #2389  uses well-founded recursion to define a `Prop` and expects that to unfold during typechecking.

* In mathlib a bunch of proofs near well-founded definitions cannot use `rfl` in the base case, but have to use `simp` or `unfold` or so; nothing too exciting.

* It seems thought that in mathlib, [`Nat.strongRec'`](https://github.com/leanprover-community/mathlib4/blob/3d892b71de5b062b8adc7126a4666f8b5db64362/Mathlib/Data/Nat/Defs.lean#L906-L907) is expected to reduce is a problem; this is used in https://github.com/leanprover-community/mathlib4/blob/3d892b71de5b062b8adc7126a4666f8b5db64362/Mathlib/CategoryTheory/Filtered/Small.lean#L79 it seems. I tried to somehow fix these proofs, but got lost.

nomeata:
Starting a second attempt at #4061, using a fresh PR to get fresh adaption branches.

PR #3997, opened 2024-04-26
experiment: split PHashMap / PHashMapSized

digama0:
More experimentation per #3046. This tries to solve the performance issue without changing the implementation, with two observations:

* Most code doesn't actually care to track the size, and need not incur the cost for it
* Of those that do, in many cases we know in advance whether we will find the value or not because we checked it immediately before.

So we introduce `PHashMapSized` as the old `PHashMap` and `PHashMap` for the old `PHashMap.Node`, and the Sized variant we have `size` and `isEmpty` functions but also has `insert`/`insertNew`/`replace` instead of just `insert` which does all three.


digama0:
Subsumed by #4503

PR #3998, opened 2024-04-26
feat:  `lake install`

### LINKED ISSUE #3423: feature request: `lake install`
Many Lean projects are development tools, and should not need to be upstream dependencies in order to use them in a project. Instead we need a `lake install` command, that registers a program as installed, and then provides a cache of binaries for that program, automatically building on requested toolchains as needed.

There are *many* projects that would make good use of this. `shake`, `graph`, and `LeanCopilot` being the more obvious ones.

tydeu:
Creates a new CLI command `lake install` which adds a package to Lake's new toolchain-local package set. These packages are stored in a `lake-packages` directory in the current Elan toolchain and are available to all packages configured with Lake on that toolchain. This allows users to "install" Lean tools such as executables and facets which will then usable in every package without specifying said tool as a dependency.  For example, [doc-gen4](https://github.com/leanprover/doc-gen4) or [import-graph](https://github.com/leanprover-community/import-graph) would be useful tools to install.

**Technical Note:** Installed packages are loaded into the workspace after the current package's dependency tree. This means that if the same package is both installed and a dependency, the dependency version will appear in the workspace. Inversely, this means that Lake will first look into installed packages for executables / libraries before looking at the current package. For example, if two separate packages, one installed and one local, have a `graph` executable, Lake will run the installed one on `lake exe graph`. 

Closes #3423. This change does not include the feature to automatically install the package on a toolchain switch because Lake does not currently have the version information to do this well. Namely, it needs a toolchain to compatible package version mapping to figure out what version to install.

### DIFF
--- a/src/lake/Lake/Load/Main.lean
+++ b/src/lake/Lake/Load/Main.lean
@@ -218,8 +218,6 @@ def Workspace.updateAndMaterialize
   (ws : Workspace) (leanOpts : Options := {})
   (toUpdate : NameSet := {}) (reconfigure := true)
 : LogIO Workspace := do
-  let ws ← ws.loadInstalledDeps leanOpts
-  let startIdx := ws.packages.size
   let (ws, entries) ← StateT.run (s := mkNameMap PackageEntry) do
     -- Use manifest versions of root packages that should not be updated
     match (← Manifest.load ws.manifestFile |>.toBaseIO) with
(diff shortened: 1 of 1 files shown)


semorrison:
Could you clarify what:
> This means that if the same package is both installed and a dependency, the dependency version will appear in the workspace.

means?

semorrison:
Isn't this

> For example, if two separate packages, one installed and one local, have a `graph` executable, Lake will run the installed one on `lake exe graph`.

the opposite of the desirable behaviour (i.e. that local configuration overrides global configuration)?

semorrison:
> This change does not include the feature to automatically install the package on a toolchain switch because Lake does not currently have the version information to do this well.

But we have the information to do it badly, don't we, by checking for tags matching the toolchain? :-)

wvhulle:
Hey what happened to this PR? Do you need any help? I was going to get started on this as well.

tydeu:
@wvhulle One of the major use cases for this at the time ended up switching to a different approach, thus this was put on the backburner as a result. There were also some design decisions we were not confident in. All in all, this is not yet a priority feature.

wvhulle:
The ability to build binaries locally and call the from anywhere seems a powerful accelerator for new language learners.

PR #3999, opened 2024-04-27
chore: lake: error on package name mismatch + std special case

tydeu:
Lake now errors instead of warns on a mismatch between a package name and what is required as. This avoids sometimes confusing downstream errors. Also, this change provides additional information for errors that may be caused by the upcoming Std rename.

### DIFF
--- a/src/lake/Lake/Load/Main.lean
+++ b/src/lake/Lake/Load/Main.lean
@@ -133,6 +133,21 @@ private def resolveDepsAcyclic
     let cycle := cycle.map (s!"  {·}")
     error s!"dependency cycle detected:\n{"\n".intercalate cycle}"
 
+def stdMismatchError (newName : String) (rev : String) :=
+s!"the 'std' package has been renamed to '{newName}' and moved to the
+'leanprover-community' organization; downstream packages which wish to
+update to the new std should replace
+
+  require std from
+    git \"https://github.com/leanprover/std4\"{rev}
+
+in their Lake configuration file with
+
+  require {newName} from
+    git \"https://github.com/leanprover-community/{newName}\"{rev}
+
+"
+
 /--
 Rebuild the workspace's Lake manifest and materialize missing dependencies.
 
@@ -192,10 +207,11 @@ def Workspace.updateAndMaterialize
           let depPkg ← liftM <| loadDepPackage dep leanOpts reconfigure
           if depPkg.name ≠ dep.name then
             if dep.name = .mkSimple "std" then
-              logError s!"the 'std' package has been renamed to '{depPkg.name}'; \
-                users should switch packages depending on new versions of std/{depPkg.name} \
-                to 'require {depPkg.name}' instead of 'require std' and also update the \
-                GitHub URL accordingly"
+              let rev :=
+                match dep.manifestEntry with
+                | .git (inputRev? := some rev) .. => s!" @ {repr rev}"
+                | _ => ""
+              logError (stdMismatchError depPkg.name.toString rev)
             try
               IO.FS.removeDirAll depPkg.dir -- cleanup
             catch e =>


src/lake/Lake/Load/Main.lean
```diff
@@ -189,7 +189,18 @@ def Workspace.updateAndMaterialize
...
-            logWarning s!"{pkg.name}: package '{depPkg.name}' was required as '{dep.name}'"
+            if dep.name = .mkSimple "std" ∧ depPkg.name = .mkSimple "boost" then
+              logError "what was the 'std' package has been renamed to 'boost'; \
+                users should switch packages depending on new versions of std/boost \
+                to 'require boost' instead of 'require std' and also update the \
+                GitHub URL accordingly"
+            if !Platform.isWindows then
+              IO.FS.removeDirAll depPkg.dir -- cleanup
+            else
+              -- Deleting git repositories via IO.FS.removeDirAll does not work reliably on Windows
```
eric-wieser, line 200: I assume this is because the vscode process is holding onto a file handle somewhere? I would expect this to work fine from the command line if vscode is not open

tydeu: @eric-wieser The removal works from the command (e.g., with `rm`), so I don't think that is the problem. I believe the problem is that Lean uses `std::remove` in the C++ code for `removeFile`, which doesn't play nice with Windows. Rust [instead](https://doc.rust-lang.org/std/fs/fn.remove_file.html) uses `unlink` on Unix and `DeleteFile` on Windows

eric-wieser: A cursory google suggests that `std::remove` should largely work fine; the issue arises only if you try to delete the file while you (or something else) has a handle on the process.

Is there any reason not to call `IO.FS.removeDirAll` here anyway on the off-chance it does work, and catch the error on windows?

src/lake/Lake/Load/Main.lean
```diff
@@ -189,7 +189,18 @@ def Workspace.updateAndMaterialize
           -- Load the package
           let depPkg ← liftM <| loadDepPackage dep leanOpts reconfigure
           if depPkg.name ≠ dep.name then
-            logWarning s!"{pkg.name}: package '{depPkg.name}' was required as '{dep.name}'"
+            if dep.name = .mkSimple "std" then
+              logError s!"what was the 'std' package has been renamed to '{depPkg.name}'; \
+                users should switch packages depending on new versions of std/{depPkg.name} \
```
semorrison, line 194: Why the `src/{...}`? I don't understand what the `/` is signifying.

tydeu: This translates to e.g. `std/batteries`. It is just a way of referring to the renamed package in a hopefully non-confusing way.

semorrison: Ah, nope, that was very confusing. :-)

Suggested text:
````
The `std` package has been renamed to `batteries`, and moved to the `leanprover-community` organization.

Downstream packages should replace
```
require std from git "https://github.com/leanprover/std4" @ "main"
```
in their `lakefile.lean` with
```
require batteries from git "https://github.com/leanprover-community/batteries" @ "main"
```
````

(Yes, I realize not everyone will be using this repo URL, or this branch --- such users are experts and can work out how to adapt.)

tydeu: @semorrison Are you just using the ` ``` ` syntax for clarity? These errors are not formatted in any way. As such, maybe an indent would be clearer?

tydeu: Added a version of your suggestion to the PR. Thanks!

semorrison (approved):

PR #4001, opened 2024-04-27
doc: add docstrings and usage examples in `Init.Data.String.Basic`

austinletson:
Add docstrings  and usage examples for `String.length`, `.push`, `.append`, `.get?`, `.set`, `.modyify`, and `.next`. Update docstrings and add usage examples for `String.toList`, `.get`, and `.get!`.

### DIFF
--- a/src/Init/Data/String/Basic.lean
+++ b/src/Init/Data/String/Basic.lean
@@ -179,8 +179,7 @@ the result is unspecified.
 
 Examples:
 * `"abc".next ⟨1⟩ = String.Pos.mk 2`
-* `"L∃∀N".next ⟨1⟩ = String.Pos.mk 4`
-`'∃'` is a multi-byte character
+* `"L∃∀N".next ⟨1⟩ = String.Pos.mk 4`, since `'∃'` is a multi-byte UTF-8 character
 
 Cases where the result is unspecified:
 * `"abc".next ⟨3⟩`, since `3 = s.endPos`


austinletson:
awaiting-review

src/Init/Data/String/Basic.lean
```diff
@@ -24,23 +24,41 @@ instance : LT String :=
...
 def length : (@& String) → Nat
   | ⟨s⟩ => s.length
 
-/-- The internal implementation uses dynamic arrays and will perform destructive updates
-   if the String is not shared. -/
+/--
+Pushes a character onto the end of a string.
+
+The internal implementation uses dynamic arrays and will perform destructive updates
+if the String is not shared.
```
nomeata, line 40:
```suggestion
if the string is not shared.
```
(I know it said String before, but think lower case is more appropriate)

src/Init/Data/String/Basic.lean
```diff
@@ -59,9 +77,13 @@ def utf8GetAux : List Char → Pos → Pos → Char
...
-  Return character at position `p`. If `p` is not a valid position
-  returns `(default : Char)`.
-  See `utf8GetAux` for the reference implementation.
+Returns the character at position `p` of a string.
+
+If `p` is not a valid position, returns `(default : Char)`.
+
+See `utf8GetAux` for the reference implementation.
+* `"abc".get ⟨1⟩ = 'b'`
+* `"abc".get ⟨3⟩ = (default : Char) = 'A'`
```
nomeata, line 86: Worth including an example that shows an invalid position within the string, due to multi-byte characters?

src/Init/Data/String/Basic.lean
```diff
@@ -89,13 +120,34 @@ def utf8SetAux (c' : Char) : List Char → Pos → Pos → List Char
   | c::cs, i, p =>
     if i = p then (c'::cs) else c::(utf8SetAux c' cs (i + c) p)
 
+/--
+Replaces the character at a specified position in a string with a new character.
+
+If the position is invalid, returns the string unchanged.
+* `"abc".set ⟨1⟩ 'B' = "aBc"`
+* `"abc".set ⟨3⟩ 'D' = "abc"`
+-/
```
nomeata, line 129:
```suggestion

If both the replacement character and the replaced character are ASCII characters and the string is not shared, destructive updates are used.
-/
```

nomeata (commented):
Thanks!

src/Init/Data/String/Basic.lean
```diff
@@ -84,6 +84,7 @@ If `p` is not a valid position, returns `(default : Char)`.
 See `utf8GetAux` for the reference implementation.
 * `"abc".get ⟨1⟩ = 'b'`
 * `"abc".get ⟨3⟩ = (default : Char) = 'A'`
+* `"L∃∀N".get ⟨2⟩ = (default : Char) = 'A' -- invalid UTF-8 position as '∃' is a mutli-byte character`
```
nomeata, line 87: Ha, nice example :-D

nomeata:
I'm happy. I'd like to give David a chance to have a look (or decline to having a look).

src/Init/Data/String/Basic.lean
austinletson: Good point. I have added another example to `String.get` variants and `String.set` to show this case.

src/Init/Data/String/Basic.lean
```diff
@@ -89,13 +123,38 @@ def utf8SetAux (c' : Char) : List Char → Pos → Pos → List Char
...
+
+If `p` is an invalid position, returns the string unchanged.
+* `abc.modify ⟨1⟩ Char.toUpper = "aBc"`
+* `abc.modify ⟨3⟩ Char.toUpper = "abc"`
+-/
 def modify (s : String) (i : Pos) (f : Char → Char) : String :=
   s.set i <| f <| s.get i
 
+/--
+Returns the next position in a string after `p` given that `p` is a valid position.
```
austinletson, line 151: @nomeata  I updated the docstring for `next` after considering the multi-byte char cases more carefully. It seems to me that `p` being a valid position (lying on a UTF-8 boundary) is an expectation of this function. 

In the case where `p` doesn't lie on a UTF-8 byte boundary. For instance, `"L∃∀N".next ⟨2⟩ = 3` since `"L∃∀N".get ⟨2⟩ = (default : Char) = 'A'`.

There is also the separate `next'` which requires `(h : ¬ s.atEnd p)` but it doesn't include the multi-byte char case.

What do you think?

austinletson: Much better wording. Thanks!

I also added the case where `p = s.endPos` as it also results in an invalid position.

src/Init/Data/String/Basic.lean
```diff
@@ -24,23 +24,41 @@ instance : LT String :=
 instance decLt (s₁ s₂ : @& String) : Decidable (s₁ < s₂) :=
   List.hasDecidableLt s₁.data s₂.data
 
+/--
+Returns the length of a string.
```
david-christiansen, line 28:
```suggestion
Returns the length of a string in Unicode code points.
```

src/Init/Data/String/Basic.lean
```diff
@@ -24,23 +24,41 @@ instance : LT String :=
...
+* `"".length = 0`
+* `"abc".length = 3`
```
david-christiansen, line 30: Generally, bulleted lists of examples should be a separate paragraph and have some indication that this is what they are, eg:

```suggestion

Examples:
* `"".length = 0`
* `"abc".length = 3`
```

In this particular case, something that shows that it's code points, rather than bytes or grapheme clusters, would be useful, but not essential.

src/Init/Data/String/Basic.lean
```diff
@@ -24,23 +24,41 @@ instance : LT String :=
...
   | ⟨s⟩ => s.length
 
-/-- The internal implementation uses dynamic arrays and will perform destructive updates
-   if the String is not shared. -/
+/--
+Pushes a character onto the end of a string.
+
+The internal implementation uses dynamic arrays and will perform destructive updates
+if the string is not shared.
+* `"abc".push 'd' = "abcd"`
```
david-christiansen, line 41:
```suggestion

Example: `"abc".push 'd' = "abcd"`
```

src/Init/Data/String/Basic.lean
```diff
@@ -24,23 +24,41 @@ instance : LT String :=
...
 def push : String → Char → String
   | ⟨s⟩, c => ⟨s ++ [c]⟩
 
-/-- The internal implementation uses dynamic arrays and will perform destructive updates
-   if the String is not shared. -/
+/--
+Appends two strings.
+
+The internal implementation uses dynamic arrays and will perform destructive updates
+if the String is not shared.
```
david-christiansen, line 51:
```suggestion
if the string is not shared.
```

src/Init/Data/String/Basic.lean
```diff
@@ -24,23 +24,41 @@ instance : LT String :=
...
+* `"abc".append "def" = "abcdef"`
```
david-christiansen, line 52:
```suggestion

Example: `"abc".append "def" = "abcdef"`
```

src/Init/Data/String/Basic.lean
```diff
@@ -24,23 +24,41 @@ instance : LT String :=
...
+-/
 @[extern "lean_string_append"]
 def append : String → (@& String) → String
   | ⟨a⟩, ⟨b⟩ => ⟨a ++ b⟩
 
-/-- O(n) in the runtime, where n is the length of the String -/
+/--
+`O(s.length)`. Converts a string to a list of characters.
+* `"abc".toList = ['a', 'b', 'c']`
```
david-christiansen, line 60:
```suggestion

Example: `"abc".toList = ['a', 'b', 'c']`
```

src/Init/Data/String/Basic.lean
```diff
@@ -24,23 +24,41 @@ instance : LT String :=
...
+if the String is not shared.
+* `"abc".append "def" = "abcdef"`
+-/
 @[extern "lean_string_append"]
 def append : String → (@& String) → String
   | ⟨a⟩, ⟨b⟩ => ⟨a ++ b⟩
 
-/-- O(n) in the runtime, where n is the length of the String -/
+/--
+`O(s.length)`. Converts a string to a list of characters.
```
david-christiansen, line 59: The description of what it does seems more important than how long it takes.

Also, a brief reason for the complexity is useful, eg:
```suggestion
 Converts a string to a list of characters.
 
 Even though the logical model of strings is as a structure that wraps a list of characters, this operation takes time and space linear in the length of the string, because the compiler uses an optimized representation as dynamic arrays.
```

src/Init/Data/String/Basic.lean
```diff
@@ -59,9 +77,14 @@ def utf8GetAux : List Char → Pos → Pos → Char
   | c::cs, i, p => if i = p then c else utf8GetAux cs (i + c) p
 
 /--
-  Return character at position `p`. If `p` is not a valid position
-  returns `(default : Char)`.
-  See `utf8GetAux` for the reference implementation.
+Returns the character at position `p` of a string.
+
+If `p` is not a valid position, returns `(default : Char)`.
```
david-christiansen, line 82:
```suggestion
Returns the character at position `p` of a string. If `p` is not a valid position, returns `(default : Char)`.
```
I think this makes sense as a single paragraph, because both sentences describe the return value.

src/Init/Data/String/Basic.lean
```diff
@@ -59,9 +77,14 @@ def utf8GetAux : List Char → Pos → Pos → Char
...
+
+See `utf8GetAux` for the reference implementation.
+* `"abc".get ⟨1⟩ = 'b'`
+* `"abc".get ⟨3⟩ = (default : Char) = 'A'`
```
david-christiansen, line 86:
```suggestion

Examples:
* `"abc".get ⟨1⟩ = 'b'`
* `"abc".get ⟨3⟩ = (default : Char) = 'A'`
```

src/Init/Data/String/Basic.lean
```diff
@@ -59,9 +77,14 @@ def utf8GetAux : List Char → Pos → Pos → Char
...
+* `"L∃∀N".get ⟨2⟩ = (default : Char) = 'A' -- invalid UTF-8 position as '∃' is a mutli-byte character`
```
david-christiansen, line 87:
```suggestion

Positions can also be invalid if a byte index points into the middle of a multi-byte UTF-8 character. For example,`"L∃∀N".get ⟨2⟩ = (default : Char) = 'A'`.
```

src/Init/Data/String/Basic.lean
```diff
@@ -72,12 +95,23 @@ def utf8GetAux? : List Char → Pos → Pos → Option Char
   | [],    _, _ => none
   | c::cs, i, p => if i = p then c else utf8GetAux? cs (i + c) p
 
+/--
+Returns the character at position `p`.
+
+If `p` is not a valid position, returns `none`.
+* `"abc".get? ⟨1⟩ = some 'b'`
+* `"abc".get? ⟨3⟩ = none`
+* `"L∃∀N".get? ⟨2⟩ = none -- invalid UTF-8 position as '∃' is a mutli-byte character`
```
david-christiansen, line 104:
```suggestion

Examples:
* `"abc".get? ⟨1⟩ = some 'b'`
* `"abc".get? ⟨3⟩ = none`

Positions can also be invalid if a byte index points into the middle of a multi-byte UTF-8 character. For example, `"L∃∀N".get? ⟨2⟩ = none`
```

src/Init/Data/String/Basic.lean
```diff
@@ -72,12 +95,23 @@ def utf8GetAux? : List Char → Pos → Pos → Option Char
...
+-/
 @[extern "lean_string_utf8_get_opt"]
 def get? : (@& String) → (@& Pos) → Option Char
   | ⟨s⟩, p => utf8GetAux? s 0 p
 
 /--
-  Similar to `get`, but produces a panic error message if `p` is not a valid `String.Pos`.
+Similar to `get`, but produces a panic error message if `p` is not a valid `String.Pos`.
```
david-christiansen, line 111:
```suggestion
Returns the character at position `p` of a string. If `p` is not a valid position, returns `(default : Char)` and produces a panic error message.
```
Just to keep it a bit more self-contained and make the hovers more useful.

src/Init/Data/String/Basic.lean
```diff
@@ -72,12 +95,23 @@ def utf8GetAux? : List Char → Pos → Pos → Option Char
...
+* `"abc".get! ⟨1⟩ = 'b'`
+* `"abc".get! ⟨3⟩ => panic!`
+* `"L∃∀N".get! ⟨2⟩ => panic! -- invalid UTF-8 position as '∃' is a mutli-byte character`
```
david-christiansen, line 114:
```suggestion

Examples:
* `"abc".get! ⟨1⟩ = 'b'`
* `"abc".get! ⟨3⟩` panics

Positions can also be invalid if a byte index points into the middle of a multi-byte UTF-8 character. For example,
`"L∃∀N".get! ⟨2⟩` panics.
```
The idea here is to keep the code in the back-ticks syntactically valid.

src/Init/Data/String/Basic.lean
```diff
@@ -89,13 +123,38 @@ def utf8SetAux (c' : Char) : List Char → Pos → Pos → List Char
   | c::cs, i, p =>
     if i = p then (c'::cs) else c::(utf8SetAux c' cs (i + c) p)
 
+/--
+Replaces the character at a specified position in a string with a new character.
+
+If the position is invalid, returns the string unchanged.
```
david-christiansen, line 129:
```suggestion
/--
Replaces the character at a specified position in a string with a new character. If the position is invalid, the string is returned unchanged.

If both the replacement character and the replaced character are ASCII characters and the string is not shared, destructive updates are used.
```

david-christiansen: I think it makes sense to move this here, because it's still describing the behavior. With this structure, the first paragraph is a functional spec, and the second describes relevant nonfunctional properties.

src/Init/Data/String/Basic.lean
```diff
@@ -89,13 +123,38 @@ def utf8SetAux (c' : Char) : List Char → Pos → Pos → List Char
...
+* `"abc".set ⟨1⟩ 'B' = "aBc"`
+* `"abc".set ⟨3⟩ 'D' = "abc"`
+* `"L∃∀N".set ⟨1⟩ 'X' = "L∃∀N" -- invalid position as '∃' is a mutli-byte character`
```
david-christiansen, line 132:
```suggestion

Examples:
* `"abc".set ⟨1⟩ 'B' = "aBc"`
* `"abc".set ⟨3⟩ 'D' = "abc"`
* `"L∃∀N".set ⟨1⟩ 'X' = "LX∀N"`

Because `'∃'` is a multi-byte character, the byte index `2` is an invalid position, so `"L∃∀N".set ⟨2⟩ 'X' = "L∃∀N"`.
```

Note that position 1 is just fine in that string here.

src/Init/Data/String/Basic.lean
```diff
@@ -89,13 +123,38 @@ def utf8SetAux (c' : Char) : List Char → Pos → Pos → List Char
...
+
+If both the replacement character and the replaced character are ASCII characters and the string is not shared, destructive updates are used.
```
david-christiansen, line 134:
```suggestion
```

src/Init/Data/String/Basic.lean
```diff
@@ -89,13 +123,38 @@ def utf8SetAux (c' : Char) : List Char → Pos → Pos → List Char
...
+-/
 @[extern "lean_string_utf8_set"]
 def set : String → (@& Pos) → Char → String
   | ⟨s⟩, i, c => ⟨utf8SetAux c s 0 i⟩
 
+/--
+Applies `f` to the character at position `p` in the string `s`.
```
david-christiansen, line 141:
```suggestion
Replaces the character at position `p` in the string `s` with the result of applying `f` to that character. If `p` is an invalid position, the string is returned unchanged.
```

src/Init/Data/String/Basic.lean
```diff
@@ -89,13 +123,38 @@ def utf8SetAux (c' : Char) : List Char → Pos → Pos → List Char
...
+
+If `p` is an invalid position, returns the string unchanged.
```
david-christiansen, line 143:
```suggestion
Examples:
```

src/Init/Data/String/Basic.lean
```diff
@@ -89,13 +123,38 @@ def utf8SetAux (c' : Char) : List Char → Pos → Pos → List Char
...
+* `abc.modify ⟨1⟩ Char.toUpper = "aBc"`
+* `abc.modify ⟨3⟩ Char.toUpper = "abc"`
+-/
 def modify (s : String) (i : Pos) (f : Char → Char) : String :=
   s.set i <| f <| s.get i
 
+/--
+Returns the next position in a string after position `p`.
+
+If `p` is not a valid position or `p = s.endPos`, the result is unspecified.
```
david-christiansen, line 153:
```suggestion
Returns the next position in a string after position `p`. If `p` is not a valid position or `p = s.endPos`, the result is unspecified.
```

src/Init/Data/String/Basic.lean
```diff
@@ -89,13 +123,38 @@ def utf8SetAux (c' : Char) : List Char → Pos → Pos → List Char
...
+* `"abc".next ⟨1⟩ = String.Pos.mk 2`
+* `"abc".next ⟨3⟩ = String.Pos.mk 4`
+* `"L∃∀N".next ⟨1⟩ => String.Pos.mk 4 -- '∃' is a mutli-byte character`
```
david-christiansen, line 156: These examples could use a bit more context. The first one is showing the expected functioning of the function, and the other two are the two exceptional cases, but this is not very clear from the description.

david-christiansen (changes requested):
Thanks! This is really useful!

There's a few things that I'd like to see happen before we merge:

1. Please indicate examples explicitly, rather than just including a bulleted list. I suggested a format here.
2. The paragraph structure of the text should ideally group related sentences together, and the first paragraph (which will show up in hovers) should describe the function (that is, which inputs go to which outputs). Subsequent paragraphs can then cover things like efficiency, data representation, and usage notes (e.g. time complexity).
3. If possible, it's best when code is valid Lean syntax, because our medium-term goal is to have Lean actually parse and validate examples.

I made suggestions that I think would satisfy all these concerns - what do you think of them?

src/Init/Data/String/Basic.lean
austinletson: Good catch here. 

I updated this example to `"L∃∀N".set ⟨4⟩ 'X' = "L∃XN"` to show setting a character which occurs after another multi-byte character.

src/Init/Data/String/Basic.lean
austinletson: I split the examples for String.next into two sections: "Examples" and "Cases where the result is unspecified." I have also included explanation of the confusing cases.

src/Init/Data/String/Basic.lean
austinletson: I added `"L∃∀N".length = 4` to differentiate between code points and bytes.

austinletson:
> Thanks! This is really useful!
> 
> There's a few things that I'd like to see happen before we merge:
> 
> 1. Please indicate examples explicitly, rather than just including a bulleted list. I suggested a format here.
> 2. The paragraph structure of the text should ideally group related sentences together, and the first paragraph (which will show up in hovers) should describe the function (that is, which inputs go to which outputs). Subsequent paragraphs can then cover things like efficiency, data representation, and usage notes (e.g. time complexity).
> 3. If possible, it's best when code is valid Lean syntax, because our medium-term goal is to have Lean actually parse and validate examples.
> 
> I made suggestions that I think would satisfy all these concerns - what do you think of them?

Thank you @david-christiansen for your feedback and suggestions!

I think all the suggestions are improvements. I made a batch commit to add them to the PR.

Additionally, I updated the examples for `String.length` and `String.next` in response to your suggestions (see my replies for details). I also went through and made some minor updates to align with the above feedback (as well as remove white space and add new lines).

src/Init/Data/String/Basic.lean
```diff
@@ -89,13 +143,51 @@ def utf8SetAux (c' : Char) : List Char → Pos → Pos → List Char
...
   s.set i <| f <| s.get i
 
+/--
+Returns the next position in a string after position `p`. If `p` is not a valid position or `p = s.endPos`,
+the result is unspecified.
+
+Examples:
+* `"abc".next ⟨1⟩ = String.Pos.mk 2`
+* `"L∃∀N".next ⟨1⟩ = String.Pos.mk 4`
+`'∃'` is a mutli-byte character
```
david-christiansen, line 183: I've seen this typo a few times:

```suggestion
`'∃'` is a multi-byte character
```

src/Init/Data/String/Basic.lean
```diff
@@ -89,13 +143,51 @@ def utf8SetAux (c' : Char) : List Char → Pos → Pos → List Char
...
+
+Cases where the result is unspecified:
+* `"abc".next ⟨3⟩ = String.Pos.mk 4`
+Since `3 = s.endPos`, the result is an invalid position
+* `"L∃∀N".next ⟨2⟩ = String.Pos.mk 3`
+Since `2` points into the middle of a multi-byte UTF-8 character the result is an invalid position
```
david-christiansen, line 189: This seems like something that will read a bit strangely in hovers, with a capital letter in the middle of the bullet point. And furthermore, if we're saying the result is unspecified, then I think it mixes messages to give concrete results.

What about this?
```suggestion
* `"abc".next ⟨3⟩`, since `3 = s.endPos`
* `"L∃∀N".next ⟨2⟩`, since `2` points into the middle of a multi-byte UTF-8 character
```

david-christiansen (commented):
This is looking very good! I think it's almost ready to go. Two final comments.

Thank you very much!

src/Init/Data/String/Basic.lean
austinletson: Including concrete results for the unspecified cases was bugging me as well. This is much better.

austinletson: I also aligned the formatting and wording of the second example under "Examples" to match these:

* `"L∃∀N".next ⟨1⟩ = String.Pos.mk 4`, since `'∃'` is a multi-byte UTF-8 character

src/Init/Data/String/Basic.lean
austinletson: I only see this instance. Maybe the others got fixed in other suggestions?

austinletson:
I addressed the two final comments and rebased off of master to try and fix the mathlib builds.

semorrison:
Don't worry about the Mathlib build for this one, there's been a CI hiccup due to the Std -> Batteries rename. Normal service should resume tomorrow.

david-christiansen:
Thanks again!

PR #4002, opened 2024-04-27
feat: make fixF irreducible, but allow fix to unroll once

nomeata:
After trying to make recursive definitions irreducible (#3995, breaks too
much), and then making `WellFounded.fix` irreducible (still breaks too
much), in this variant we define `fix` as
```
fix F x = F x (fun y _ => fixF F y (apply hwf y))
```
with an extra layer on `F` on the outside, and mark `fixF` as
irreducible.

This means that

* a wf-recursively defined function is defeq to its right-hand side for
  for the base case. This allows the majority of existing `rfl` proofs
  about wf-recursive functions to go through.

* `fix F = fix G` if `F = G`; this kind of reasoning shows up
  in mathlib, where
  [OrdinalApprox.gfpApprox](https://leanprover-community.github.io/mathlib4_docs/Mathlib/SetTheory/Ordinal/FixedPointApproximants.html#OrdinalApprox.gfpApprox)
  needs to be defeq to the similarly defined `lfpApprox` for certain
  arguments.

Maybe this is a decent sweet-spot.

### DIFF
--- a/tests/lean/run/simpDiag.lean
+++ b/tests/lean/run/simpDiag.lean
@@ -20,11 +20,6 @@ example : f (x + 50) = f x := by
   set_option diagnostics true in
   simp [f_eq, q_eq]
 
-example : f (x + 50) = f x := by
-  set_option diagnostics true in
-  simp [f_eq, q_eq]
-
-
 def ack : Nat → Nat → Nat
   | 0,   y   => y+1
   | x+1, 0   => ack x 1


nomeata:
Mathlib branch [lean-pr-testing-4002](https://github.com/leanprover-community/mathlib4/compare/nightly-testing...lean-pr-testing-4002) is down to a few files missing, with errors that I couldn’t fix easily within
```
././././Mathlib/CategoryTheory/ComposableArrows.lean (aesop_cat stopped working in one instance)
././././Mathlib/RingTheory/WittVector/FrobeniusFractionField.lean
././././Mathlib/CategoryTheory/Filtered/Small.lean
```

Not sure if I should push harder, or get help to finish this before deciding on the fate of this PR, or leave it at that and (should we end up taking the present path) just throw it over the wall. But I don’t like the idea of throwing it over the wall, and I would like to see the mathlib report…  so maybe I should ask for experts on these files if they want to help with `lean-pr-testing-4002` if it turns out to not be too hard. Anyways, not today…

nomeata:
By deleting some files I coudln’t fix I was able to obtain a speedcenter report:
http://speed.lean-fro.org/mathlib4/compare/7f5b6acd-65fc-499b-914e-cebdb4597821/to/462f3fa6-ee87-40cd-b772-f9977ad3f0e3

Significant changes are

* Mathlib.Computability.Halting — instructions -12.9B
* Mathlib.NumberTheory.PythagoreanTriples — instructions -41.4B 

and otherwise nothing worth mentioning. As the example in https://github.com/leanprover/lean4/issues/4051 shows preventing wf functions from reducing altogether (#3995) might be the bigger win.

PR #4004, opened 2024-04-27
perf: `isDefEqProj`

leodemoura:


semorrison:
(Added `full-ci` so I get a usable toolchain before the nightly lands.)

PR #4006, opened 2024-04-27
fix: proposition fields must be theorems

### LINKED ISSUE #2575: Proposition fields result in `def`s not `theorem`s
### Prerequisites

### Description

When declaring a structure, all the projections are generated as `def`s even though morally they should sometimes be thought of as theorems. Specifically, fields of type `h : P` where `P : Prop` would be more natural as `theorem`s, as this matches the intended use of `theorem` elsewhere.

### Context

This impacts the `docBlame` linter in Std, which declares that `def`s should have docstrings (because their type alone doesn't describe their function) but `theorem`s need not.
For now, we are having to work around this for the special case of projections in std4#269.

### Steps to Reproduce

```lean
structure AtLeastThirtySeven where
  val : Nat
  le : 37 ≤ val

theorem AtLeastThirtySeven.lt (x : AtLeastThirtySeven) : 36 < x.val := x.le
```

**Expected behavior:** `AtLeastThirtySeven.le` and `AtLeastThirtySeven.lt` should both be `theorem`s (or at least, there should be some way to request that they both be).

**Actual behavior:** `AtLeastThirtySeven.le` is a `def`

### Versions

leanprover/lean4:v4.1.0-rc1

### Impact

leodemoura:
closes #2575

### DIFF
--- a/src/Lean/Elab/StructInst.lean
+++ b/src/Lean/Elab/StructInst.lean
@@ -821,7 +821,9 @@ partial def reduce (structNames : Array Name) (e : Expr) : MetaM Expr := do
     | some r => reduce structNames r
     | none   => return e.updateProj! (← reduce structNames b)
   | .app f .. =>
-    match (← reduceProjOf? e structNames.contains) with
+    -- Recall that proposition fields are theorems. Thus, we must set transparency to .all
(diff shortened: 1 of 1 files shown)


eric-wieser:
Thanks!

PR #4023, opened 2024-04-29
chore: minimization of timeout in Mathlib.Algebra.Module.Submodule.Localization

kim-em:
This is a minimization of an `isDefEq` timeout from Mathlib.Algebra.Module.Submodule.Localization,
where it is reasonably fast with `set_option backward.isDefEq.lazyWhnfCore false`,
but very slow with the default `isDefEq` algorithm.

### DIFF
--- a/tests/lean/run/localization.lean
+++ b/tests/lean/run/localization.lean
@@ -1012,7 +1012,7 @@ def Submodule.toLocalizedQuotient' : M ⧸ M' →ₗ[R] N ⧸ M'.localized' S p 
 set_option maxHeartbeats 100000 in
 set_option backward.isDefEq.lazyWhnfCore false in
 theorem Submodule.toLocalizedQuotient'_mk₁ (x : M) :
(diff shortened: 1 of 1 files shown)


tests/lean/run/localization.lean
```diff
@@ -0,0 +1,1036 @@
...
+end Mathlib.Algebra.Quotient
+
+section Mathlib.GroupTheory.Coset
+
+variable {α : Type _}
+
+namespace QuotientAddGroup
+
+variable [AddGroup α] (s : AddSubgroup α)
+
```
mattrobball, line 753:
```suggestion
instance [Add α] : VAdd α α := 
  { vadd := Add.add }

def leftRel : Setoid α :=
  VAdd.orbitRel s α
```
- makes this very fast 
- explains why making `op` irreducible speeds things up 
- fixes defeq abuse 

The instance is just so you can switch out `s.op` for `s`.

leodemoura:
We found the performance issue and fixed at [#4085](https://github.com/leanprover/lean4/pull/4085)

PR #4025, opened 2024-04-29
refactor: Use withReducible mkEqnTypes in WF.mkEqns

nomeata:
it seems to help with problems that arise in #4002, so I am curious
about its effects (if any) on its own.

It could lead to
 * faster `mkEqnTypes`, if in the well-founded case the `tryURefl` never
   hits (in which case we should probaby skip that altogether)
 * more fine-grained equational lemams, if the `tryURefl` did trigger


nomeata:
Whooaa, the perf changes are all over the place, but many regressions:
http://speed.lean-fro.org/mathlib4/compare/d444b71b-0228-43f5-951d-975ddabbc565/to/7736775e-e4c2-4bae-ae8c-e63c111a36d9.
So probably not the right thing to do.

nomeata:
Subsumed by #4047

PR #4032, opened 2024-04-30
chore: modernize build instructions

Kha:
Use `cmake --preset`, adjust and document parallelism settings

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -417,7 +417,7 @@ jobs:
         run: |
           ulimit -c unlimited # coredumps
           # clean rebuild in case of Makefile changes
-          make -C build update-stage0 && rm -rf ./stage* && make -C build -j$NPROC
+          make -C build update-stage0 && rm -rf build/stage* && make -C build -j$NPROC
         if: matrix.name == 'Linux' && needs.configure.outputs.quick == 'false'
       - name: CCache stats
         run: ccache -s


doc/make/index.md
```diff
@@ -17,39 +17,27 @@ Platform-Specific Setup
...
 ```bash
-git clone https://github.com/leanprover/lean4 --recurse-submodules
+git clone https://github.com/leanprover/lean4
 cd lean4
-mkdir -p build/release
-cd build/release
-cmake ../..
-make
+cmake --preset release
+cmake --build --preset release -j$(nproc)
```
digama0, line 26: This looks a lot harder to remember than `make` or `make -j4` which is what I usually use.

Kha: Yes, I've simplified that step and adjusted the parallelism description. I still think it makes sense to give the instructions that give the fastest build for the majority of users first. It's just insane that neither make nor cmake provide a built-in way to do so.

kmill:
While we're editing doc/main/index.md, could we add an explanation to the top for who these instructions are for? At a Lean school I taught at, one of the participants first inclinations as a hardcore Linux user was to get started with Mathematics in Lean by building Lean from source, which immediately caused issues, so it would be good to at least include a pointer to the quickstart. Here's a draft:

> These are instructions to set up a working development environment for those who wish to make changes to Lean itself. It is part of the [Development Guide](doc/dev/index.md).

> We strongly suggest that new users instead follow the [Quickstart](doc/quickstart.md) to get started using Lean, since this sets up an environment that can automatically manage multiple Lean toolchain versions, which is necessary when working within the Lean ecosystem.

PR #4037, opened 2024-04-30
feat: lemmas to simplify equalities with `Option`-typed dependent if-then-else

### LINKED ISSUE #4013: Adding new lemmas `dite_some_none_eq_some` and `dite_some_none_eq_none`
There are `ite_some_none_eq_none` and  `dite_some_none_eq_some` in `Init.ByCases`. But I would like to add the 'dite' versions to the same file, as proved below:
```
theorem dite_some_none_eq_none {P : Prop} [Decidable P] {α : Type*} {x : P → α} :
    (if h : P then some (x h) else none) = none ↔ ¬P := by
  simp only [dite_eq_right_iff]

@[simp] theorem dite_some_none_eq_some {P : Prop} [Decidable P] {α : Type*} {x : P → α} {y : α} :
    (if h : P then some (x h) else none) = some y ↔ ∃ h : P, x h = y := by
  split <;> simp_all
```

I've never contributed directly to this repository, so I thought I'd open an issue here to make sure this is the correct place to add the lemmas.

wupr:
Closes #4013.

Add `dite_some_none_eq_none` and `dite_some_none_eq_some`, analogous to the existing `ite_some_none_eq_none` and `ite_some_none_eq_some`.


nomeata:
Hmm, looks like this breaks mathlib, this might need diagnosing

wupr:
It looks like the problem could be with the use of a non-final `simp` in Mathlib. But I think I need to merge the new commits in `nightly-with-mathlib` first.

wupr:
The errors was thrown in `Mathlib/Data/FinEnum.lean` at the third line below:
```
instance Subtype.finEnum [FinEnum α] (p : α → Prop) [DecidablePred p] : FinEnum { x // p x } :=
  ofList ((toList α).filterMap fun x => if h : p x then some ⟨_, h⟩ else none)
    (by rintro ⟨x, h⟩; simp; exists x; simp [*])
```
But the proof can be simplified as follows now:
```
instance Subtype.finEnum [FinEnum α] (p : α → Prop) [DecidablePred p] : FinEnum { x // p x } :=
  ofList ((toList α).filterMap fun x => if h : p x then some ⟨_, h⟩ else none)
    (by simp)
```
There may be other problems like this and I'm trying to compile Mathlib locally to identify them.

Where should I push the fix? To the `lean-pr-testing-4037` branch on Mathlib?

semorrison:
Changes to Mathlib look fine, so lets proceed!

PR #4047, opened 2024-05-01
refactor: do not try rfl in mkEqnTypes in WF.mkEqns

nomeata:
when dealing with well-founded recursive definitions, `tryURefl` isn't
going to be that useful and possibly slow. So disable that code path
when doing well-founded recursion.

(This is a variant of #4025 where I tried using `with_reducible` to
limit the impact of slow unfolding, but if we can get away with
disabling it complete, then even better.)


nomeata:
!bench

nomeata:
Mathlib perf results: http://speed.lean-fro.org/mathlib4/compare/d263df69-867e-4d8e-b1b0-3f3ff0c077c9/to/029cb476-33f2-4227-84be-3ad9be09f0b8
Not significant changes.

PR #4052, opened 2024-05-02
feat: validate reducibility attribute setting

leodemoura:
and new option `set_option allowUnsafeReductibility true` to override validation.

### DIFF
--- a/src/Lean/ReducibilityAttrs.lean
+++ b/src/Lean/ReducibilityAttrs.lean
@@ -95,8 +95,8 @@ register_builtin_option allowUnsafeReducibility : Bool := {
 }
 
 private def validate (declName : Name) (status : ReducibilityStatus) (attrKind : AttributeKind) : CoreM Unit := do
-  let suffix := "use `set_option allowUnsafeReductibility true` to override reducibility status validation"
-  unless allowUnsafeReductibility.get (← getOptions) do
+  let suffix := "use `set_option allowUnsafeReducibility true` to override reducibility status validation"
+  unless allowUnsafeReducibility.get (← getOptions) do
     match (← getConstInfo declName) with
     | .defnInfo _ =>
       let statusOld := getReducibilityStatusCore (← getEnv) declName
--- a/tests/lean/run/reducibilityAttrValidation.lean
+++ b/tests/lean/run/reducibilityAttrValidation.lean
@@ -2,7 +2,7 @@ def f (x : Nat) := x + 1
 
 /--
 error: failed to set `[semireducible]` for `f`, declarations are `[semireducible]` by default
-use `set_option allowUnsafeReductibility true` to override reducibility status validation
+use `set_option allowUnsafeReducibility true` to override reducibility status validation
 -/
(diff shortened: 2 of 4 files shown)


src/Lean/ReducibilityAttrs.lean
```diff
@@ -65,15 +70,75 @@ def setReducibilityStatusCore (env : Environment) (declName : Name) (status : Re
 private def setReducibilityStatusImp (env : Environment) (declName : Name) (status : ReducibilityStatus) : Environment :=
   setReducibilityStatusCore env declName status .global .anonymous
 
+/-
+TODO: it would be great if we could distinguish between the following two situations
```
tydeu, line 74: Could `ReducibilityHints.abbrev` be used to distinguish between these cases?

leodemoura: Unfortunately, it does not help. The key issue here is that the current attribute management system does not tell us whether an attribute is being set at a declaration time or not.

tydeu: True, having the attribute know whether it is applied at declaration time would be the best solution.  However, I was proposing a potential half-baked solution assuming that was too complicated for now. 

If I understand correctly, `abbrev` is just a ``@[reducible, inline] def`` with `ReducibilityHints.abbrev`. Thus, the primary use case for a declaring `@[reducible] def` over an `abbrev`  is when the definition is meant to be reducible but not inlined. In such a case,  `@[noinline] abbrev` could serve the same purpose. Thus, to avoid the potential for applying `@[reducible]` after a declaration, we could forbid `@[reducible]` from being applied to anything except an `abbrev`. This would avoid the need to make the declaration time distinction.

src/Lean/ReducibilityAttrs.lean
```diff
@@ -65,15 +70,75 @@ def setReducibilityStatusCore (env : Environment) (declName : Name) (status : Re
...
+def foo := ...
+...
+attribute [reducible] foo
+```
+
+Reason: the second one is problematic if user has add simp theorems or TC instances that include `foo`.
+Recall that the discrimination trees unfold `[reducible]` declarations while indexing new entries.
+-/
+
+register_builtin_option allowUnsafeReductibility : Bool := {
```
digama0, line 92:
```suggestion
register_builtin_option allowUnsafeReducibility : Bool := {
```
The same typo occurs in the PR description

src/Lean/ReducibilityAttrs.lean
leodemoura: @tydeu This is a good suggestion. Then, we would not even have the attribute `[reducible]`. It seems there are very few instances of `@[reducible] def` in Mathlib. So, the impact would be minimal. @semorrison What do you think? We can use a different PR for this change.

Kha: Fwiw, adding a flag to `AttributeImpl.add` that tells us whether this is a post-hoc application seems like a simple (apart from downstream breakage) and generally very useful change to me

PR #4053, opened 2024-05-02
feat: add `seal` and `unseal` commands

leodemoura:


nomeata:
Do we want/need the `seal … in $c:command` variant, and maybe also the `seal … in` variant for terms and tactics?

leodemoura:
> Do we want/need the `seal … in $c:command` variant, and maybe also the `seal … in` variant for terms and tactics?

We already have it. See new test:
```lean
seal f in
example : f x = x + 1 := rfl
```

nomeata:
Ah, neat! I didn't expect that after scrolling through the macro definition. Neat.

Kha:
@nomeata `in` is a generic command combinator

nomeata:
But I assume it won’t work scoped around terms and tactics for free, does it?
(Not urgently needed).

I’m a heavy user of `unseal` already: 
https://github.com/leanprover-community/mathlib4/compare/nightly-testing...lean-pr-testing-4061

PR #4055, opened 2024-05-02
fix: remove `Subtype.instInhabited`

kmill:
It has had a long history going back [10 years](https://github.com/leanprover-community/lean/commit/3afad10a7294fd183f66cc9d4723511b799ac4ff#diff-4e22e2bb74f004d2ff7cdabcb5c01429abbc906e20befe2517679e257b4387e4R41), but its time has come to an end since this instance is never applicable.


kmill:
!bench

PR #4059, opened 2024-05-03
feat: upstream lemmas about basic List/Array operations

kim-em:
This PR upstreams lemmas about List/Array operations already defined in Lean from std/batteries.

Happy to take suggestions about increasing or decreasing scope.

### DIFF
--- a/tests/lean/run/rwRegression.lean
+++ b/tests/lean/run/rwRegression.lean
@@ -1,9 +1,5 @@
 namespace Array
 
-@[simp] theorem getElem_pop (a : Array α) (i : Nat) (hi : i < a.pop.size) :
-    a.pop[i] = a[i]'(Nat.lt_of_lt_of_le (a.size_pop ▸ hi) (Nat.sub_le _ _)) :=
-  sorry
-
 theorem ex {as : Array α} (h : i < size as)  (hlt: i < size (pop as)) :
     as[i] = (pop as)[i] := by
   rw [getElem_pop] -- should close the goal, proof should be found by unification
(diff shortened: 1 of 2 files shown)


src/Init/Data/List/Lemmas.lean
```diff
@@ -735,4 +737,976 @@ theorem minimum?_eq_some_iff [Min α] [LE α] [anti : Antisymm ((· : α) ≤ ·
...
+
+@[simp] theorem getElem?_eq_get? (l : List α) (i : Nat) : l[i]? = l.get? i := by
+  simp only [getElem?]; split
+  · exact (get?_eq_get ‹_›).symm
+  · exact (get?_eq_none.2 <| Nat.not_lt.1 ‹_›).symm
+
+/--
+If one has `get l i hi` in a formula and `h : l = l'`, one can not `rw h` in the formula as
+`hi` gives `i < l.length` and not `i < l'.length`. The theorem `get_of_eq` can be used to make
+such a rewrite, with `rw (get_of_eq h)`.
```
digama0, line 1274:
```suggestion
such a rewrite, with `rw [get_of_eq h]`.
```

digama0:
For this and future moves, would it be possible to follow the same protocol we used during the port, and have the initial commit be a plain copy/paste job so that the diffs needed to make it compile stand out more clearly?

semorrison:
> For this and future moves, would it be possible to follow the same protocol we used during the port, and have the initial commit be a plain copy/paste job so that the diffs needed to make it compile (and any additional changes) stand out more clearly?

Ah, yes, sorry, I know you've asked this before. I'll try to remember. It's a bit difficult to know where exactly to make that commit here: I initially copy-pasted in the entire files that I was drawing from, and then whittled down as dependencies weren't available or I decided I didn't want something upstreamed.

digama0:
As long as the diff isn't a complete mess, I think it's okay if the first commit includes the entirety of the affected files and hence subsequent commits contain a lot of red

Issue #4051, opened 2024-05-02
Elaboration of explicit term proof significantly slower than equivalent tactic proof


somombo:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to mathlib4 or std4.

### Description

I have an example of a term based proof (in a recursive context) being about 200 times slower to elaborate than by simply adding "`by exact` ..." (or "by apply ...") to the front of the same term (see [mwe](https://leanprover-community.github.io/mwe.html) below).

### Context
_**See [Swap-Perm](https://github.com/somombo/swaps-perm) project which formally defines permutations of arrays and proves related theorems.**_

I am trying to prove a basic theorem about an extension to the definition of `Array.swap` that allows for making several swaps all at once. My extension is called `Array.swaps` and accepts a list of pairs of indices instead of just a single pair of indices.

My first attempt at the theorem in question (`Array.swaps_cancel` below) was a entirely a term based recursive proof. The offending line of code was inductive hypothesis (See
https://github.com/somombo/swaps-perm/blob/b4a05c1775ad1d32eadc07b635c10a25323a1815/SwapsPerm.lean#L100-L101)

### Steps to Reproduce
Consider the following [mwe](https://leanprover-community.github.io/mwe.html):

```lean
def Array.swaps (a : Array α) : List (Fin a.size × Fin a.size) → Array α
  | [] => a
  | (i, j) :: ijs =>
    have : (a.swap i j).size = a.size := a.size_swap _ _

    swaps (a.swap i j) (ijs.map (fun p => ⟨⟨p.1.1, this.symm ▸ p.1.2⟩, ⟨p.2.1, this.symm ▸ p.2.2⟩⟩))
termination_by l => l.length

set_option trace.profiler true in
theorem Array.swaps_cancel (a : Array α) (l : List (Fin a.size × Fin a.size)) : a.swaps (l ++ l.reverse) = a :=
  match l with
  | [] => sorry
  | c :: cs =>

    have h : a.size = (a.swaps [c]).size := sorry

    have ih1 : ((a.swaps [c]).swaps ((h ▸ cs) ++ (h ▸ cs).reverse)) = (a.swaps [c]) := swaps_cancel (a.swaps [c]) (h ▸ cs) -- FIXME: takes approx 5s
[...]

PR #4061, opened 2024-05-03
feat: well-founded definitions irreducible by default

nomeata:
we keep running into examples where working with well-founded recursion
is slow because defeq checks (which are all over the place, including
failing ones that are back-tracked) unfold well-founded definitions.

The definition of a function defined by well-founded recursion should be
an implementation detail that should only be peeked inside by the
equation generator and the functional induction generator.

We now mark the mutual recursive function as irreducible (if the user did not
set a flag explicitly), and use `withAtLeastTransparency .all` when producing
the equations.

Proofs can be fixed by using rewriting, or – a bit blunt, but nice for adjusting 
existing proofs – using `unseal` (a.k.a. `attribute [local semireducible]`).

Mathlib performance does not change a whole lot:
http://speed.lean-fro.org/mathlib4/compare/08b82265-75db-4a28-b12b-08751b9ad04a/to/16f46d5e-28b1-41c4-a107-a6f6594841f8
Build instructions -0.126 %, four modules with significant instructions decrease.

To reduce impact, these definitions were changed:

 * `Nat.mod`, to make `1 % n` reduce definitionally, so that `1` as a `Fin 2` literal
   works nicely. Theorems with larger `Fin` literals tend to need a `unseal Nat.modCore`
   https://github.com/leanprover/lean4/pull/4098
 * `List.ofFn` rewritten to be structurally recursive and not go via `Array.ofFn`:
   https://github.com/leanprover-community/batteries/pull/784

Alternative designs explored were

 * Making `WellFounded.fix` irreducible. 
 
   One benefit is that recursive functions with equal definitions (possibly after
   instantiating fixed parameters) are defeq; this is used in mathlib to relate
   [`OrdinalApprox.gfpApprox`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/SetTheory/Ordinal/FixedPointApproximants.html#OrdinalApprox.gfpApprox) with `.lfpApprox`.
   
   But the downside is that one cannot use `unseal` in a
   targeted way, being explicit in which recursive function needs to be reducible here.

   And in cases where Lean does unwanted unfolding, we’d still unfold the recursive
   definition once to expose `WellFounded.fix`, leading to large terms for often no good
   reason.

 * Defining `WellFounded.fix` to unroll defintionally once before hitting a irreducible
   `WellFounded.fixF`. This was explored in #4002. It shares most of the ups and downs
   with the previous variant, with the additional neat benefit that function calls that
   do not lead to recursive cases (e.g. a `[]` base case) reduce nicely. This means that
   the majority of existing `rfl` proofs continue to work.

Issue #4051, which demonstrates how badly things can go if wf recursive functions can be
unrolled, showed that making the recursive function irreducible there leads to noticeably
faster elaboration than making `WellFounded.fix` irreducible; this is good evidence that
the present PR is the way to go. 

This fixes https://github.com/leanprover/lean4/issues/3988

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -9,9 +9,15 @@ assignees: ''
 
 ### Prerequisites
 
-* [ ] Put an X between the brackets on this line if you have done all of the following:
-    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
-    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to Mathlib or Batteries.
+Please put an X between the brackets as you perform the following steps:
+
+* [ ] Check that your issue is not already filed:
+      https://github.com/leanprover/lean4/issues
+* [ ] Reduce the issue to a minimal, self-contained, reproducible test case.
+      Avoid dependencies to Mathlib or Batteries.
+* [ ] Test your test case against the latest nightly release, for example on
+      https://live.lean-lang.org/#project=lean-nightly
+      (You can also use the settings there to switch to “Lean nightly”)
 
 ### Description
 
@@ -33,8 +39,8 @@ assignees: ''
 
 ### Versions
 
-[Output of `#eval Lean.versionString` or of `lean --version` in the folder that the issue occured in]
-[OS version]
+[Output of `#eval Lean.versionString`]
+[OS version, if not using live.lean-lang.org.]
 
 ### Additional Information
 
--- a/src/Init/Data/Array/Subarray.lean
+++ b/src/Init/Data/Array/Subarray.lean
@@ -22,7 +22,7 @@ abbrev Subarray.as (s : Subarray α) : Array α := s.array
 theorem Subarray.h₁ (s : Subarray α) : s.start ≤ s.stop := s.start_le_stop
 
 @[deprecated Subarray.stop_le_array_size]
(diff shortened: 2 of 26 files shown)


nomeata:
Whoohoo, mathlib is green!

After having discussed this with Leo, I will move forward with this approach.

We can reduce the impact on mathlib (seen [here](https://github.com/leanprover-community/mathlib4/compare/nightly-testing...lean-pr-testing-4061)) by rewriting some functions using structural recursion (I’m looking at you, `Array.ofFn.go` and `Nat.modCore`), but that can happen in separate PRs.

Another possible way to reduce the impact a bit is if the `decides` tactics reduces more aggressively by default.

digama0:
The PR says "by default". Is there a way to change the default for a given definition? Is slapping `@[semireducible]` on the definition sufficient?

nomeata:
The latter, see https://github.com/leanprover-community/mathlib4/compare/nightly-testing...lean-pr-testing-4061#diff-fd2963a2e28f1375d7ef28b499c216ed5b86dceabd79c37154edf5f6bdb62295R82

semorrison:
Perhaps rebase now that #4053 is in a nightly? Not essential.

semorrison (approved):

nomeata:
No need to rebase for that, the diff will simply disappear if there is no conflict.

nomeata:
I plan to merge this once 
https://github.com/leanprover/lean4/pull/4098
and
https://github.com/leanprover-community/batteries/pull/784
have landed.

I should trigger a stage0 update right afterwards, else it won’t have an effect on the released stdlib.

PR #4066, opened 2024-05-05
fix: mainModuleName should use srcSearchPath

digama0:
As [reported on Zulip](https://leanprover.zulipchat.com/#narrow/stream/341532-lean4-dev/topic/Find.20references.20broken.20in.20lean.20core/near/437051935). The `mainModuleName` was being set incorrectly when browsing lean core sources, resulting in failure of cross-file server requests like "Find References". Because the `srcSearchPath` is generated asynchronously, we store it as a `Task Name` which is resolved some time before the header is finished parsing. (I don't think the `.get` here will ever block, because the srcSearchPath will be ready by the time the initial command snap is requested.)

### DIFF
--- a/src/Lean/Elab/Frontend.lean
+++ b/src/Lean/Elab/Frontend.lean
@@ -154,9 +154,9 @@ def runFrontend
 
     return (s.commandState.env, !s.commandState.messages.hasErrors)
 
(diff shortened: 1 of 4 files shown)


digama0:
Yes, LGTM.

mhuisi (approved):
LGTM

PR #4070, opened 2024-05-05
refactor: subst notation try to run isDefEq only once

nomeata:
there is already a `isDefEq` in `ensureHasType` and previously was
another one after rewriting in `hType`. If we just always rewrite in
`hType`, then we may get away with just one `isDefEq`.


nomeata:
Not really a good idea, it can to rewriting twice where only once is desired.

PR #4074, opened 2024-05-06
refactor: Array.feraseIdx: avoid have in definition

nomeata:
otherwise it remains in the equational theorem and may cause the
“unused have linter” to trigger. By moving the proof into
`decreasing_by`, the equational theorems are unencumbered by termination
arguments.

see also https://github.com/leanprover/std4/pull/690#issuecomment-2095378609


digama0:
What are the chances that this is fixed in the equation compiler so that it doesn't generate terms that violate the unused have linter?

PR #4076, opened 2024-05-06
To int eq broken

tobiasgrosser:
Test if the CI fails in case of broken proofs


tobiasgrosser:
Close as this should have gone to my personal fork.

PR #4083, opened 2024-05-06
refactor: lake: manifest semver & code cleanup

tydeu:
Switches the manifest format to use `major.minor.patch` semantic versions. Major version increments indicate breaking changes (e.g., new required fields and semantic changes to existing fields). Minor version increments (after `0.x`) indicate backwards-compatible extensions (e.g., adding optional fields, removing fields). This change is backwards compatible. Lake will still successfully read old manifest with numeric versions. It will treat the numeric version `N` as semantic version `0.N.0`. Lake will also accept manifest versions with `-` suffixes (e.g., `x.y.z-foo`) and then ignore the suffix.

This change also includes the general cleanup/refactoring of the manifest code and data structures that was part of #3174.

### DIFF
--- a/src/lake/Lake/Load/Manifest.lean
+++ b/src/lake/Lake/Load/Manifest.lean
@@ -50,7 +50,7 @@ That is, Lake ignores the `-` suffix.
 -/
 @[inline] def Manifest.version : LeanVer := v!"1.0.0"
 
-/-- Manifest version 6 package entry. For backwards compatibility. -/
+/-- Manifest version `0.6.0` package entry. For backwards compatibility. -/
 inductive PackageEntryV6
 | path (name : Name) (opts : NameMap String) (inherited : Bool) (dir : FilePath)
 | git (name : Name) (opts : NameMap String) (inherited : Bool) (url : String) (rev : String)


src/lake/tests/manifest/lake-manifest-v1.0.0.json
```diff
@@ -0,0 +1,20 @@
+{"version": "1.0.0",
```
Vtec234, line 1: Syntactic comment, but for somebody just looking at the manifest, this field is easy to confuse with the package version. But I suppose that was already the case while using sequential versioning.

semorrison:
Just a minor comment about PR labelling: I think you might be overusing the "draft" status, or at least I don't think anything should ever be both a draft and `will-merge-soon`, which have opposite implications for reviewing urgency!

tydeu:
> I don't think anything should ever be both a draft and will-merge-soon,

@semorrison That was just me forgetting to mark it ready for review. 🤦‍♂️

src/lake/tests/manifest/lake-manifest-v1.0.0.json
tydeu: True, that is certainly something to keep in mind. I don't think I have good solution for that now in this PR, though.

PR #4087, opened 2024-05-07
Incr tactic rc1

Kha:
#3940 merged into 4.8.0-rc1 for testing


Kha:
Temporarily changed the base of #3940 instead to keep things simple and in one place

PR #4088, opened 2024-05-07
chore: lake: tweak hovers for `family_def` et al

digama0:
This makes the `leanArts` in `library_data leanArts : BuildJob Unit` get a hover for the generated axiom. It also simplifies the `quoteFrom` function so that it delaborates properly by using a name literal (which elaborates to `mkStr1`, `mkStr2` etc) instead of a `mkStr` application.


tydeu:
LGTM! Thanks!

PR #4085, opened 2024-05-07
perf: issue at `binop%` and `binrel%` elaborators

leodemoura:
This issue was affecting several Mathlib files.

@mattrobball Found where the `[nonassignable]`s were coming from. Thanks for helping to diagnose this performance issue.

PR #4092, opened 2024-05-07
perf: issue at `binop%` and `binrel%` elaborators

leodemoura:
This issue was affecting several Mathlib files.

@mattrobball @semorrison This is a different solution for the issue. The comment at `Extra.lean` describes the new solution and documents the new issues found with the previous one.

closes #4085

### DIFF
--- a/src/Lean/Elab/Extra.lean
+++ b/src/Lean/Elab/Extra.lean
@@ -273,7 +273,33 @@ where
            match (← get).max? with
            | none     => modify fun s => { s with max? := type }
            | some max =>
-             unless (← withNewMCtxDepth <| isDefEqGuarded max type) do
+             /-
+              Remark: Previously, we used `withNewMCtxDepth` to prevent metavariables in `max` and `type` from being assigned.
+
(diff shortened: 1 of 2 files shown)


leodemoura:
Rebased on top of `nightly-with-mathlib`

mattrobball:
Mathlib master builds locally with this cherry-picked on top of v4.8.0-rc1 without any changes. Anecdotally, it seems quite snappy.

mattrobball:
leanprover-community/mathlib4#12742 has a benchmark run using the above. Overall -1.1T build instructions and no individual files suffered significant regressions but the wall clock was up +37%. I think the latter is anomaly. Will re-run.

mattrobball:
Wall clock was an anomaly. Last run was neutral.

semorrison:
(The automatic Mathlib run didn't happen because of a CI breakage induced by the std -> batteries rename. Hopefully fixed going forward by https://github.com/leanprover-community/batteries/pull/785.)

PR #4097, opened 2024-05-07
feat: `IO.TaskState`

tydeu:
Adds `IO.getTaskState` which returns the state of a `Task` in the Lean runtime's task manager. The `TaskState` inductive has 3 constructors: `waiting`, `running`, and `finished`. The `waiting`  constructor encompasses the waiting and queued states within the C task object documentation, because the task object does not provide a low cost way to distinguish these different forms of waiting. Furthermore, it seems unlikely for consumers to wish to distinguish between these internal states. The `running` constructor encompasses both the running and promised states in C docs. While not ideal, the C implementation does not provide a way to distinguish between a running `Task` and a waiting `Promise.result` (they both have null closures).

### DIFF
--- a/tests/lean/run/taskState.lean
+++ b/tests/lean/run/taskState.lean
@@ -1,34 +1,25 @@
-import Lean.Elab.ElabRules
-
-open Lean
-
-elab "file_pos%" : term => return toExpr <| (← getFileMap).toPosition (← getRefPos)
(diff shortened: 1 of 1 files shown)


Kha (approved):
Using `Promise` we could add a test case for at least waiting and finished

PR #4098, opened 2024-05-07
refactor: make 1 % n reduce without well-founded recursion

nomeata:
this is in preparation for #4061. Once that lands, `1 % 42 = 1` will no
longer hold definitionally (at least not without an ungly `unseal
Nat.modCore in` around). This affects mathlib in a few places,
essentially every time a `1 : Fin (n+1)` literal is written.

So this extends the existing special case for `0 % n = 0` to `1 % n`.


src/Init/Data/Nat/Div.lean
```diff
@@ -82,22 +82,35 @@ decreasing_by apply div_rec_lemma; assumption
...
   | 0, _ => 0
-  | x@(_ + 1), y => Nat.modCore x y
+  | 1, 0 => 0
+  | 1, 1 => 0
+  | 1, (_+2) => 1
+  | x@(_ + 2), y => Nat.modCore x y
 
 instance instMod : Mod Nat := ⟨Nat.mod⟩
 
+unseal Nat.modCore in
```
nomeata, line 98: Not needed
```suggestion
```

PR #4109, opened 2024-05-08
chore: bug template: point to live.lean-lang.org

nomeata:
https://live.lean-lang.org/#project=lean-nightly now allows users to
play around with the latest lean nightly, and it seems prudent to ask
them to test bug reports, if possible, there, and not just with whatever
release they use.

Also reformatted the descriptions to look well in a text area. Users
will not see this as rendered markdown, but as plain text.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -11,7 +11,7 @@ assignees: ''
 
 Please put an X between the brackets as you perform the following steps:
 
-* [ ] Check that your issue is not already files:
+* [ ] Check that your issue is not already filed:
       https://github.com/leanprover/lean4/issues
 * [ ] Reduce the issue to a minimal, self-contained, reproducible test case.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -9,9 +9,15 @@ assignees: ''
 
 ### Prerequisites
 
-* [ ] Put an X between the brackets on this line if you have done all of the following:
-    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
-    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to Mathlib or Batteries.
+Please put an X between the brackets as you perform the following steps:
+
+* [ ] Check that your issue is not already files:
```
Kha, line 14:
```suggestion
* [ ] Check that your issue is not already filed:
```

Kha (approved):

Kha:
Good idea!

PR #4113, opened 2024-05-08
feat: propagate `maxHeartbeats` to kernel

leodemoura:


leodemoura:
@Kha Do you understand why the Nix Linux build is failing? I don't use Nix.

PR #4118, opened 2024-05-09
chore: CI: optimize code under Linux Debug

Kha:


Kha:
Well it's fast, exactly as fast as the standard release build. So I suspect that `-UNDEBUG` did not actually do anything, but not willing right now to reproduce and debug locally.

PR #4120, opened 2024-05-09
fix: take `synthPendingDepth` into account when caching TC results

leodemoura:
@semorrison This is for debugging purposes only. It contains only the TC cache fix. If #4119 does not fix the performance issues in Mathlib. We can use this one to check whether the issue is due to this commit or not.


semorrison:
Closing in favour of #4119.

PR #4124, opened 2024-05-10
chore: add warning comments about `findM?`

kim-em:


semorrison:
Pinging @arthur-adjedj, just FYI.

arthur-adjedj:
Thank you for the ping, I didn't take that into consideration when writing that function.
FYI, `occursOrInType` could be rewritten with the non-monadic `find` by passing the LocalContext to the function and fetching the type of fvars there. Doing so might be the better option compared to keeping `findM` in the codebase.

semorrison:
> Thank you for the ping, I didn't take that into consideration when writing that function. FYI, `occursOrInType` could be rewritten with the non-monadic `find` by passing the LocalContext to the function and fetching the type of fvars there. Doing so might be the better option compared to keeping `findM` in the codebase.

If you'd be interested in doing that, I'd love to merge it. :-) `findM?` seems like a scary footgun to leave around.

arthur-adjedj:
Right. I'm not available to work on this now, but will make an according PR to fix `occursOrInType` and to remove the monadic find by the end of the day.

semorrison:
Obsolete given #4125.

PR #4137, opened 2024-05-11
fix: hovers on binders with metavariables

### LINKED ISSUE #4078: Rpc Error when hovering variable
### Prerequisites

### Description

In the following code:
```lean
import Lean.Data.HashMap

open Lean

inductive Decl (α : Type) where
  | atom (idx : α)
  deriving BEq, Hashable

def Cache (α : Type) := Unit

structure CacheHit (decl : Decl α) where
  idx : Nat

def Cache.find (cache : Cache α) (decl : Decl α) : (CacheHit decl) :=
  sorry

theorem denote_mkAtom_cached {cache : Cache α} {hit} (h : cache.find (.atom v) = hit) :
    hit = hit := by
  sorry
```
Hovering `hit` at `{hit}` yields 
```
Error updating: Error fetching goals: Rpc error: InternalError: unknown metavariable '?_uniq.927'. Try again.
```
In the infoview. Furthermore the hover itself prints:
```
hit : failed to pretty print expression (use 'set_option pp.rawOnError true' for raw representation)
```
However hovering on `hit` in the theorem statement `hit = hit` does reveal a proper hover.

### Steps to Reproduce

1. Copy code into editor
2. Hover over `{hit}`
3. :boom: 

**Expected behavior:** The hover should be the same at the binder and the equation

**Actual behavior:** A broken hover and seemingly something wrong with the LSP as well.

### Versions
4.8.0-rc1

### Impact

nomeata:
this fixes #4078

### DIFF
--- a/tests/lean/interactive/4078.lean
+++ b/tests/lean/interactive/4078.lean
@@ -1,14 +1,17 @@
-def foo1 (i : ∀ α, List α) : List β := i β
+def foo1 {β} (i : ∀ α, List α) : List β := i β
+                  --^ textDocument/hover
+
+def foo2 (i : ∀ α, List α) : List β := i β
               --^ textDocument/hover
 
-def foo2 (i : (α : _) → List α) : List β := i β
+def foo3 (i : (α : _) → List α) : List β := i β
               --^ textDocument/hover
 
-def foo3 (i : (α : id _) → List α) : List β := i β
+def foo4 (i : (α : id _) → List α) : List β := i β
               --^ textDocument/hover
 
-def foo4 (i : (α : Type _) → List α) : List β := i β
+def foo5 (i : (α : Type _) → List α) : List β := i β
               --^ textDocument/hover
 
-def foo5 (i : (α : Type 0) → List α) : List β := i β
+def foo6 (i : (α : Type 0) → List α) : List β := i β
               --^ textDocument/hover
--- a/tests/lean/interactive/4078.lean.expected.out
+++ b/tests/lean/interactive/4078.lean.expected.out
@@ -1,20 +1,17 @@
 {"textDocument": {"uri": "file:///4078.lean"},
- "position": {"line": 0, "character": 16}}
+ "position": {"line": 0, "character": 20}}
 {"range":
- {"start": {"line": 0, "character": 16}, "end": {"line": 0, "character": 17}},
+ {"start": {"line": 0, "character": 20}, "end": {"line": 0, "character": 21}},
  "contents": {"value": "```lean\nα : Type ?u\n```", "kind": "markdown"}}
 {"textDocument": {"uri": "file:///4078.lean"},
  "position": {"line": 3, "character": 16}}
 {"range":
- {"start": {"line": 3, "character": 14}, "end": {"line": 3, "character": 30}},
- "contents":
- {"value":
-  "```lean\nType (?u + 1)\n```\n***\nThe dependent arrow. `(x : α) → β` is equivalent to `∀ x : α, β`, but we usually\nreserve the latter for propositions. Also written as `Π x : α, β` (the \"Pi-type\")\nin the literature. ",
-  "kind": "markdown"}}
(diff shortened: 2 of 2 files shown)


nomeata:
Mostly as a learning experience, I wanted to be see if I can fix a bug this part of the code based. I wonder if this is really the right fix, because I don’t see much `withSaveInfoContext` in this module, so happy to be educated! @kmill maybe?

src/Lean/Elab/Binders.lean
```diff
@@ -211,8 +211,9 @@ private partial def elabBinderViews (binderViews : Array BinderView) (fvars : Ar
       let id := binderView.id.getId
       let kind := kindOfBinderName id
       withLocalDecl id binderView.bi type (kind := kind) fun fvar => do
-        addLocalVarInfo binderView.ref fvar
-        loop (i+1) (fvars.push (binderView.id, fvar))
+        withSaveInfoContext do
```
kmill, line 214: I don't know this part of the code base very well (in particular, when `withSaveInfoContext` is needed), but it seems to me that the error actually has something to do with autoimplicits. Over in `Lean.Elab.Term.withAutoBoundImplicit` when the new autoimplicit is declared, the state is restored, but there's nothing recording the new state in the info trees. Maybe `withSaveInfoContext` should go there instead?

One piece of evidence for this is that there's no error when hovering over `α` when autoImplicits isn't activated:
```lean
def oops {β} (i : ∀ α, List α) : List β := i β
```

nomeata: Good `catch`, let’s `try`.

(Unrelatedly, but given that backtracking implementation of auto bound implicit, I wonder how easily that can increase elaboration time significantly.)

nomeata: Nope, 
```diff
diff --git a/src/Lean/Elab/Term.lean b/src/Lean/Elab/Term.lean
index ee1014f944..476d807dda 100644
--- a/src/Lean/Elab/Term.lean
+++ b/src/Lean/Elab/Term.lean
@@ -1533,7 +1533,8 @@ partial def withAutoBoundImplicit (k : TermElabM α) : TermElabM α := do
               s.restore
               withLocalDecl n .implicit (← mkFreshTypeMVar) fun x =>
                 withReader (fun ctx => { ctx with autoBoundImplicits := ctx.autoBoundImplicits.push x } ) do
-                  loop (← saveState)
+                  withSaveInfoContext do
+                    loop (← saveState)
             | none   => throw ex
       loop (← saveState)
   else
```
does not help.

kmill: It seems to be that there's stale info because `s.restore` doesn't restore infotrees. Assuming we want *full* backtracking, then maybe this is the solution?
```
partial def withAutoBoundImplicit (k : TermElabM α) : TermElabM α := do
  let flag := autoImplicit.get (← getOptions)
  if flag then
    withReader (fun ctx => { ctx with autoBoundImplicit := flag, autoBoundImplicits := {} }) do
      let rec loop (s : SavedState) : TermElabM α := withIncRecDepth do
        checkSystem "auto-implicit"
        try
          withSaveInfoContext k
        catch
          | ex => match isAutoBoundImplicitLocalException? ex with
            | some n =>
              -- Restore state, declare `n`, and try again
              s.restore (restoreInfo := true)
              withLocalDecl n .implicit (← mkFreshTypeMVar) fun x =>
                withReader (fun ctx => { ctx with autoBoundImplicits := ctx.autoBoundImplicits.push x } ) do
                  loop (← saveState)
            | none   => throw ex
      loop (← saveState)
  else
    k
```

kmill: In further testing, doing `s.restore (restoreInfo := true)` is enough and `withSaveInfoContext` isn't necessary.

nomeata: Feel free to push here and take over, or open a new PR

nomeata: Trying that in https://github.com/leanprover/lean4/pull/4192

PR #4145, opened 2024-05-13
refactor: let Nat.mod reduce more

nomeata:
this refined upon #4098 and makes `Nat.mod` reduce on even more
literals. The key observation that I missed earlier is that `if m ≤ n`
reduces better than `if n < m`.

Also see discussion at
https://github.com/leanprover-community/mathlib4/pull/12853#discussion_r1597798308


nomeata:
@semorrison WDYT of this way to handle the `Fin`-literal issues?

nomeata:
I’ll merge this so that it’s in the next nightly. Certainly better than #4098.

PR #4149, opened 2024-05-13
fix: FunInd: Introduce HEq where necessary and no equality on Props

### LINKED ISSUE #4146: FunInd chokes on pattern match with dependent targets
[Lean.Meta.MatcherApp.transform](https://leanprover-community.github.io/mathlib4_docs/Lean/Meta/Match/MatcherApp/Transform.html#Lean.Meta.MatcherApp.transform), used by `FunInd`, chokes upon a dependent `match` as is produced by

```
def bar (n : Nat) : Bool :=
  if h : n = 0 then
    true
  else
    have m+1 := n
    bar m
termination_by n

/--
error: failed to transform matcher, type error when constructing new pre-splitter motive:
  bar.match_1
    (fun n h_1 =>
      h = h_1 → (∀ (y : Nat), (invImage (fun x => x) instWellFoundedRelationOfSizeOf).1 y n → motive y) → motive n)
    x h
---
error: application type mismatch
  h✝ = h
argument
  h
has type
  ¬n✝ = 0 : Prop
but is expected to have type
  ¬x = 0 : Prop
-/
#check bar.induct
```

The `have` syntax is desugared to
```
def baz (n : Nat) : Bool :=
  if h : n = 0 then
    true
  else
    match n, h with
    | m+1, _ => baz m
termination_by n
#check baz.induct
```

And also affects other types, e.g.
```
def foo [DecidableEq α] (xs : List α) : Bool :=
  if h : xs = [] then
    true
  else
    have _::ys := xs
    foo ys
termination_by xs
#check foo.induct
```

### Versions

4.9.0-nightly-2024-05-11

### Impact

nomeata:
this fixes #4146

The resulting functional induction principles generated when there is a
`match` in the function definition are far from pretty and contain
suprious copies of local definitions at their non-refined or refined
type. I expect that sometimes these may be useful, so I am hesitant to
remove them too aggressively, so for now it’s better to have too many
assumptions than too few, or even a failure (as in #4146).

### DIFF
--- a/tests/lean/run/issue4146.lean
+++ b/tests/lean/run/issue4146.lean
@@ -1,0 +1,59 @@
+set_option linter.unusedVariables false
+
+def bar (n : Nat) : Bool :=
+  if h : n = 0 then
+    true
+  else
+    match n with -- NB: the elaborator adds `h` as an discriminant
+    | m+1 => bar m
+termination_by n
+
+-- set_option pp.match false
+-- #print bar
+-- #check bar.match_1
+-- #print bar.induct
+
+-- NB: The induction theorem has both `h` in scope, as its original type mentioning `x`,
+-- and a refined `h` mentioning `m+1`.
+-- The former is redundant here, but will we always know that?
+-- No HEq betwen the two `h`s due to proof irrelevance
+
+/--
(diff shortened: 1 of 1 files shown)


nomeata:
Now contained in https://github.com/leanprover/lean4/pull/5364

PR #4150, opened 2024-05-13
perf: globally cache intermediate type class results

JovanGerb:
I seem to have messed up the git stuff, so I made a new identical PR: #4152

See Zulip discussion [here](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Ways.20to.20speed.20up.20Mathlib/near/438126466)

This is a PR to determine whether this idea indeed results in faster type class search.


nomeata:
If this PR is created just to get the CI test, but you don't want people to look at and comment on the idea and implementation yet, using a “draft PR” makes it even more clear.

PR #4151, opened 2024-05-13
fix: add missing registerTraceClass `Meta.synthInstance.instances

JovanGerb:
The trace class Meta.synthInstance.answer isn't registered, so it can't be used.

I set `inherited := true`, because I think it is a useful trace to have. In particular it tells you when an instance has been found that has a too large size. This is very useful information.


JovanGerb:
Is there a way to automatically commit the changes that are required in the test output files, due to the extra trace?

semorrison:
> Is there a way to automatically commit the changes that are required in the test output files, due to the extra trace?

https://lean-lang.org/lean4/doc/dev/testing.html#fixing-tests (but please note the request for caution!)

JovanGerb:
awaiting-review

PR #4158, opened 2024-05-14
refactor: DiscrTree: return more specific results first

nomeata:
(just an experiment for now, an I gotta run, proper text following later.)

### DIFF
--- a/tests/lean/run/heapSort.lean
+++ b/tests/lean/run/heapSort.lean
@@ -125,7 +125,7 @@ def extractMax {lt} (self : BinaryHeap α lt) : Option α × BinaryHeap α lt :=
 
 theorem size_pos_of_max {lt} {self : BinaryHeap α lt} (e : self.max = some x) : 0 < self.size :=
   Decidable.of_not_not fun h: ¬ 0 < self.1.size => by
-    simp [BinaryHeap.max, Array.get?, h, - List.length_pos] at e
+    simp [BinaryHeap.max, Array.get?, h] at e
 
 /-- `O(log n)`. Equivalent to `extractMax (self.insert x)`, except that extraction cannot fail. -/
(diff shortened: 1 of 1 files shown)


nomeata:
Noting down an analysis of one test failure, mostly for my own reference:

In `heapSort.lean`, 
```
theorem size_pos_of_max {lt} {self : BinaryHeap α lt} (e : self.max = some x) : 0 < self.size :=
  Decidable.of_not_not fun h: ¬ 0 < self.1.size => by
    simp [BinaryHeap.max, Array.get?, h] at e
```
breaks. After unfolding `BinaryHeap.max` and `Array.get?` we have rewritten to
```
(if h : 0 < self.arr.size then some self.arr[0] else none) = some x
```
Originally, now the local assumption
```
[Meta.Tactic.simp.rewrite] h:1000, 0 < self.arr.size ==> False 
```
is applied. But with the reordering in this PR, for some reason
```
[Meta.Tactic.simp.rewrite] @List.length_pos:1000, 0 < self.arr.size ==> self.arr.data ≠ [] 
```
is used. Yay for non-confluent rewrite systems.

I did not expect my change to affect this; `h: ¬ 0 < self.1.size` seems to be more specific than `length_pos {l : List α} : 0 < length l ↔ l ≠ []`, but maybe the DiscTree key is actually not what I expect here.

Unrelatedly: Not sure if `length_pos` really is a good simp lemma.

nomeata:
!bench

nomeata:
!bench

nomeata:
!bench

nomeata:
I’ll pursue another idea first: Attaching explicit priorities to the equational lemmas. Will close this once I have a new PR up.

nomeata:
Closing in favor of #4434

PR #4166, opened 2024-05-14
doc: add docstrings and examples for String functions

austinletson:
Add docstrings, usage examples, and doc tests for `String.prev`, `.front`, `.back`, `.atEnd`.

Improve docstring examples for `String.next` based on discussion examples for `String.prev`.

### DIFF
--- a/src/Init/Data/String/Basic.lean
+++ b/src/Init/Data/String/Basic.lean
@@ -178,8 +178,9 @@ Returns the next position in a string after position `p`. If `p` is not a valid 
 the result is unspecified.
 
 Examples:
-* `"abc".next ⟨1⟩ = String.Pos.mk 2`
-* `"L∃∀N".next ⟨1⟩ = String.Pos.mk 4`, since `'∃'` is a multi-byte UTF-8 character
+Given `def abc := "abc"` and `def lean := "L∃∀N"`,
+* `abc.get (0 |> abc.next) = 'b'`
+* `lean.get (0 |> lean.next |> lean.next) = '∀'`
 
 Cases where the result is unspecified:
 * `"abc".next ⟨3⟩`, since `3 = s.endPos`
@@ -201,9 +202,10 @@ Returns the position in a string before a specified position, `p`. If `p = ⟨0�
 If `p` is not a valid position, the result is unspecified.
 
 Examples:
-* `"abc".prev ⟨2⟩ = String.Pos.mk 1`
-* `"abc".prev ⟨0⟩ = String.Pos.mk 0`
-* `"L∃∀N".prev ⟨4⟩ = String.Pos.mk 1`, since `'∃'` is a multi-byte UTF-8 character
+Given `def abc := "abc"` and `def lean := "L∃∀N"`,
+* `abc.get (abc.endPos |> abc.prev) = 'c'`
+* `lean.get (lean.endPos |> lean.prev |> lean.prev |> lean.prev) = '∃'`
+* `"L∃∀N".prev ⟨3⟩` is unspecified, since byte 3 occurs in the middle of the multi-byte character `'∃'`.
 -/
 @[extern "lean_string_utf8_prev"]
 def prev : (@& String) → (@& Pos) → Pos
@@ -235,11 +237,11 @@ points to the end of a string. Otherwise, returns `false`.
 
 Examples:
 Given `def abc := "abc"` and `def lean := "L∃∀N"`,
-* `0 |> abc.next |> abc.next |> abc.atEnd = false`
-* `0 |> abc.next |> abc.next |> abc.next |> abc.next |> abc.atEnd = true`
-* `0 |> lean.next |> lean.next |> lean.next |> lean.next |> lean.atEnd = true`
+* `(0 |> abc.next |> abc.next |> abc.atEnd) = false`
+* `(0 |> abc.next |> abc.next |> abc.next |> abc.next |> abc.atEnd) = true`
+* `(0 |> lean.next |> lean.next |> lean.next |> lean.next |> lean.atEnd) = true`
 
-Because "L∃∀N" contains multi-byte characters, `lean.next (lean.next 0)` is not equal to `abc.next (abc.next 0)`.
+Because `"L∃∀N"` contains multi-byte characters, `lean.next (lean.next 0)` is not equal to `abc.next (abc.next 0)`.
 -/
 @[extern "lean_string_utf8_at_end"]
 def atEnd : (@& String) → (@& Pos) → Bool
--- a/tests/lean/run/string.lean
+++ b/tests/lean/run/string.lean
@@ -1,14 +1,17 @@
 /-!
 # Examples from documentation added in https://github.com/leanprover/lean4/pull/4166
 -/
+def abc : String := "abc"
+def lean : String := "L∃∀N"
 
-#guard "abc".prev ⟨2⟩ = String.Pos.mk 1
-#guard "abc".prev ⟨0⟩ = String.Pos.mk 0
-#guard "L∃∀N".prev ⟨4⟩ = String.Pos.mk 1
+#guard abc.get (0 |> abc.next) = 'b'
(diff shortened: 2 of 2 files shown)


src/Init/Data/String/Basic.lean
```diff
@@ -196,16 +196,48 @@ def utf8PrevAux : List Char → Pos → Pos → Pos
...
   get s (prev s s.endPos)
 
+/--
+Returns `true` if a specified position is greater than or equal to the position which
+points to the end of a string. Otherwise, returns `false`.
+
+Examples:
+* `"abc".atEnd ⟨2⟩ = false`
+* `"abc".atEnd ⟨3⟩ = true `
+* `"L∃∀N".atEnd ⟨8⟩ = true `
```
semorrison, line 239:
```suggestion
* `"abc".atEnd ⟨3⟩ = true`
* `"L∃∀N".atEnd ⟨8⟩ = true`
```

src/Init/Data/String/Basic.lean
```diff
@@ -196,16 +196,48 @@ def utf8PrevAux : List Char → Pos → Pos → Pos
...
 @[extern "lean_string_utf8_prev"]
 def prev : (@& String) → (@& Pos) → Pos
   | ⟨s⟩, p => if p = 0 then 0 else utf8PrevAux s 0 p
 
+/--
+Returns the first character in `s`. If `s = ""`, returns `(default : Char)`.
+
+Examples:
+* `"abc".front = 'a'`
+* `"".front = (default : Char) = 'A'`
```
semorrison, line 217: This example is confusing, since you can't actually chain `=` like that in Lean!
```suggestion
* `"".front = (default : Char)`
```

src/Init/Data/String/Basic.lean
```diff
@@ -196,16 +196,48 @@ def utf8PrevAux : List Char → Pos → Pos → Pos
...
+-/
 def front (s : String) : Char :=
   get s 0
 
+/--
+Returns the last character in `s`. If `s = ""`, returns `(default : Char)`.
+
+Examples:
+* `"abc".back = 'c'`
+* `"".back = (default : Char) = 'A'`
```
semorrison, line 227:
```suggestion
* `"".back = (default : Char)`
```

src/Init/Data/String/Basic.lean
```diff
@@ -196,16 +196,48 @@ def utf8PrevAux : List Char → Pos → Pos → Pos
...
     if i' = p then i else utf8PrevAux cs i' p
 
+/--
+Returns the position in a string before a specified position, `p`. If `p = ⟨0⟩`, returns `0`.
+If `p` is not a valid position, the result is unspecified.
+
+Examples:
+* `"abc".prev ⟨2⟩ = String.Pos.mk 1`
+* `"abc".prev ⟨0⟩ = String.Pos.mk 0`
+* `"L∃∀N".prev ⟨4⟩ = String.Pos.mk 1`, since `'∃'` is a multi-byte UTF-8 character
```
david-christiansen, line 206: These examples all worry me, because they seem to suggest that users should be counting bytes and creating a `String.Pos` from scratch. Additionally, the use of two separate notations for `String.Pos.mk` risks throwing off some users, and I think the `String` API is more likely to attract new users than many other areas. Finally, it shows off an unspecified result, which I think is better left unwritten.

I also think that byte 4 is valid, based on:
```
#eval let i := ⟨0⟩; let str := "L∃∀N"; str.next i
```
yielding 1 and
```
#eval let i := ⟨0⟩; let str := "L∃∀N"; str.next <| str.next i
```
yielding 4.

What about:
```suggestion
Examples:
* `"abc".prev ⟨2⟩ = ⟨1⟩`
* `"abc".prev ⟨0⟩ = ⟨0⟩`
* `"L∃∀N".prev ⟨3⟩` is unspecified, since `'∃'` is a multi-byte UTF-8 character and byte 3 is in the middle of it
```

src/Init/Data/String/Basic.lean
david-christiansen: I think it even makes sense to leave off the `'A'` entirely here, because there's nothing really special about that being the default `Char` - we could just as easily make it be `'∃'` someday.

One argument for including it is that it could provide a clue to users who are getting some weird data, though.

src/Init/Data/String/Basic.lean
david-christiansen, line 239: Byte counts still feel like they're setting a bad example here. What about:
```suggestion
Given `def abc := "abc"` and `def lean := "L∃∀N"`,
* `0 |> abc.next |> abc.next |> abc.atEnd = false`
* `0 |> abc.next |> abc.next |> abc.next |> abc.next |> abc.atEnd = true `
* `0 |> lean.next |> lean.next |> lean.next |> lean.next |> lean.atEnd = true `

Because "L∃∀N" contains multi-byte characters, `lean.next (lean.next 0)` is not equal to `abc.next (abc.next 0)`.
```

src/Init/Data/String/Basic.lean
austinletson: > I also think that byte 4 is valid, based on:

 I intended for `"L∃∀N".prev ⟨4⟩ = String.Pos.mk 1` to illustrate a valid result in the case where the string contains multi-byte characters. I didn't include an unspecified example here (unlike [`String.next`](https://github.com/leanprover/lean4/blob/master/src/Init/Data/String/Basic.lean#L184)) but can definitely add one.

More generally, I think your concerns regarding byte counts from scratch is spot on. What about the following modification modeled after your suggestion for `String.atEnd` to encourage more natural usage:

```suggestion
Examples:
Given `def abc := "abc"` and `def lean := "L∃∀N"`,
* `abc.endPos |> abc.prev = ⟨2⟩`
* `lean.endPos |> lean.prev |> lean.prev |> lean.prev = ⟨1⟩`
* `"L∃∀N".prev ⟨3⟩` is unspecified, since `'∃'` is a multi-byte UTF-8 character and byte 3 is in the middle of it
```

The other advantage here is demonstrating that if you call `String.prev` with a constructed `String.pos`, instead of with an iterative pattern, you run the risk of unspecified results.

Here is an alternative version with `String.get` which is more explicit at the cost of complicating the example code:
```suggestion
Examples:
Given `def abc := "abc"` and `def lean := "L∃∀N"`,
* `abc.get <| abc.endPos |> abc.prev = 'c'`
* `lean.get <| lean.endPos |> lean.prev |> lean.prev |> lean.prev = '∃'`
* `"L∃∀N".prev ⟨3⟩` is unspecified, since `'∃'` is a multi-byte UTF-8 character and byte 3 is in the middle of it
```

What do you think?

src/Init/Data/String/Basic.lean
austinletson: Updated. This is much more clear.

austinletson:
@semorrison Thank you for adding the tests! After all the threads are resolved, I will go back and update the test to reflect the final docstrings.

src/Init/Data/String/Basic.lean
david-christiansen: I like it, except for the mixing of `<|` and `|>` - the precedence of them is not necessarily obvious. What about parens around the chains of `|>` instead?

austinletson: :+1: I changed these examples to use parens instead of `<|`.

Additionally, I converted the examples for the similar `String.next` which was documented a few weeks ago in #4001 to this iterative style instead of constructing positions.

austinletson:
I updated the tests based on the updated examples.

Let me know if there is any additional areas for improvement. If not, this should be ready to go!

src/Init/Data/String/Basic.lean
david-christiansen: This looks great, thanks!

david-christiansen (approved):
This looks really good now!

Sorry for the long turnaround, this was a holiday weekend in Denmark.

austinletson:
Thanks, David! I hope you had a nice holiday!

Issue #2670, opened 2023-10-11
Simp regressions probably related to transparency


PatrickMassot:
### Prerequisites

* [C] Put an X between the brackets on this line if you have done all of the following:
    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to mathlib4 or std4.

### Description

Many `simp` proofs from Lean 3 no longer work, probably because the discrimination tree that is necessary for improved performance does not allow seeing through definitional equality. I think this is worth reporting anyway.

### Context

The specific examples come from [this Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/simp.20.5BX.5D.20fails.2C.20rw.20.5BX.5D.20works/near/390538357)

### Steps to Reproduce

```lean
example {α β : Type} {f : α × β → β → β} (h : ∀ p : α × β, f p p.2 = p.2)
  (a : α) (b : β) : f (a, b) b = b := by
  -- simp [h] -- fails but worked in Lean 3
  rw [h] -- works

example {α β : Type} {f : α × β → β → β}
  (a : α) (b : β) (h : f (a,b) (a,b).2 = (a,b).2) : f (a, b) b = b := by
  -- simp [h] -- fails but worked in Lean 3
  rw [h] -- works
```

**Expected behavior:** 

The `simp` call closes the goal as they did in Lean 3.

**Actual behavior:** 

`simp` is not working but `rw` works.

### Versions

Lean (version 4.2.0-rc1, commit https://github.com/leanprover/lean4/commit/a62d2fd4979671b76b8ab13ccbe4fdf410ec0d9d, Release)
Ubuntu 22.04

### Impact

In isolation this is not a problem, but when porting project it can make a complicated simp call involving a lot of lemmas very painful to debug. And the ported output then requires alternating calls to `simp only` and `rw`.

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

leodemoura:
Yes, it is due to discriminations trees. We can disable the discrimination tree indexing for a subterm `t` using `no_index t`. Example:
```lean
example {α β : Type} {f : α × β → β → β} (h : ∀ p : α × β, f p (no_index p.2) = p.2)
[...]

PR #4171, opened 2024-05-15
feat: test for simp problem

kim-em:
This will fail unless we have a fix!

### DIFF
--- a/tests/lean/run/bimon.lean
+++ b/tests/lean/run/bimon.lean
@@ -1,0 +1,728 @@
+
+/-!
+This is a minimization of a problem in Mathlib where a simp lemma `foo` would not fire,
+but variants:
+* `simp [(foo)]`
+* `simp [foo.{v₁ + 1}]`
+* `simp [foo']`, where a `no_index` is added in the statement
+all work.
+-/
+
+section Mathlib.Data.Opposite
+
+universe v u
+
+variable (α : Sort u)
+
+structure Opposite :=
+  op ::
+  unop : α
+
+notation:max α "ᵒᵖ" => Opposite α
+
+end Mathlib.Data.Opposite
+
+section Mathlib.Combinatorics.Quiver.Basic
+
+open Opposite
+
+universe v v₁ v₂ u u₁ u₂
+
+class Quiver (V : Type u) where
+  Hom : V → V → Sort v
+
+infixr:10 " ⟶ " => Quiver.Hom
+
+structure Prefunctor (V : Type u₁) [Quiver.{v₁} V] (W : Type u₂) [Quiver.{v₂} W] where
+  obj : V → W
+  map : ∀ {X Y : V}, (X ⟶ Y) → (obj X ⟶ obj Y)
+namespace Quiver
+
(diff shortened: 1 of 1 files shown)


leodemoura:
Copying my Zulip message here to make sure we find it in the future.

I investigated the examples in this issue, and I now understand what is going on. It is ugly :( Here are my findings:

- The issue lies in the discrimination tree. `foo` is not even tried when using `simp [foo]`.

- `simp` treats non-constant arguments differently. When processing a non-constant, `simp` will index function arguments as wildcards, mimicking Lean 3's behavior. This approach was motivated by issue #2670. The idea is that a non-constant is essentially a "local theorem," and its arguments may be messy. For example, given `(h : f (a,b) (a,b).2 = (a,b).2)`, if we use `simp [h]` and index the argument `(a,b).2`, it will produce a bad index.

- `simp` has special support for identifiers that correspond to constants. `simp [foo]` is shorthand for `simp [@foo]`. Thus, the left-hand-side arguments of `foo` are indexed. The following index is added to the discrimination tree: `[Quiver.Hom.unop, *, *, *, *, Opposite.op, Quiver.Hom, *, *, Opposite.0, *, Opposite.0, *, *]`.

- In `simp [(foo)]`, metavariables are created for `foo`'s implicit arguments, and it is **not** elaborated as a constant, but as a local theorem. Thus, the index generated is: `[Quiver.Hom.unop, *, *, *, *, *]`.

- In `simp [foo']`, `foo'` is processed as a constant, **but** the theorem statement for `foo'` uses a `no_index`, resulting again in the index is: `[Quiver.Hom.unop, *, *, *, *, *]`.

- In `simp [foo.{v+1}]`, `foo.{v+1}` is not an identifier, and `simp`'s support for treating it as `@foo.{v+1}` does not kick in. Metavariables for implicit arguments are again added, the result is not a constant, and again the index `[Quiver.Hom.unop, *, *, *, *, *]` is generated. Note that `simp [@foo.{v+1}]` fails, as it is elaborated as a constant. Providing the universe is not directly helping; it just disables the code that treats identifiers like `foo` as `@foo` in `simp` arguments. I acknowledge this is horrible and counterintuitive.

This is all quite messy, and it is not clear how to proceed. I see the some options:

- Create a new diagnostic section for `simp` arguments that were never used, and show their index. 

- Allow users to instruct `simp` to use Lean 3 style indexing for constants such as `foo`. This would result in the simpler index `[Quiver.Hom.unop, *, *, *, *, *]`. We could even suggest this option for arguments that have a non-trivial index and were not used. Right now, `simp [(foo)]` is a very contrived way to accomplish that, but binder annotations are lost in the process.

nomeata:
> Create a new diagnostic section for simp arguments that were never used, and show their index.

Related: Just yesterday I was wondering if we could have a linter that warns about unused `simp` arguments. I assume it has to be a linter, and not a warning that simp can emit, in case the `simp` invocation is used multi times (e.g. after `<;>`). (I'd guess that simp can put in the info tree whether it used the argument, or even how often, and then the linter can aggregate over multiple entries from multiple runs of the tactic.)

PR #4172, opened 2024-05-15
fix: List.length_pos no @[simp]

nomeata:
in #4158 I was experimenting with a change to the simplifier that
affectes the order in which lemmas were tried, and of course it breaks
proofs all over the place whenever we have a non-confluent simp set.

Among the first breakages encountered, a large fraction was due to
`simp` rewriting with `List.length_pos  : 0 < length l ↔ l ≠ []`.

This does not strike me a as a good simp lemma: If `l` is a manifest
constructor, the simplifier will reduce `length` and solve it anyways,
and if it isn't then an inequality usually isn’t very simp friendly. It
is also highly non-confluent with any kind of `length`-lemma we might
have.

This therefore removes it from the standard simp set.

### DIFF
--- a/src/Init/Data/List/Lemmas.lean
+++ b/src/Init/Data/List/Lemmas.lean
@@ -777,7 +777,6 @@ theorem exists_cons_of_length_succ :
     ∀ {l : List α}, l.length = n + 1 → ∃ h t, l = h :: t
   | _::_, _ => ⟨_, _, rfl⟩
 
-@[simp]
 theorem length_pos {l : List α} : 0 < length l ↔ l ≠ [] :=
   Nat.pos_iff_ne_zero.trans (not_congr length_eq_zero)
 


TwoFX:
For the record, another argument why this might be an undesirable simp lemma is that due to `Array.size` being reducible, `0 < d.size` for `d : Array α` simplifies to `¬d.data = []` with this lemma, which I found quite surprising when I first encountered it after leanprover-community/batteries#769 was merged.

nomeata:
Oh, that’s even worse thatn what I saw!

Also thanks for digging out this PR. So this is a recent change? Then I am more confident about proposing to undo it.

I wish we could just require the default simp set to be locally confluent, with a [checker](https://github.com/nomeata/lean-simplc). It would give us more than gut feeling and anecdotal evidence when evaluating simp set changes!

nomeata:
Ok, and even mathlib builds. Should be straight forward then.

semorrison (approved):

PR #4175, opened 2024-05-15
chore: use #guard_msgs in run tests

kim-em:
Many of our tests in `tests/lean/run/` produce output from `#eval` (or `#check`) statements, that is then ignored.

This PR tries to capture all the useful output using `#guard_msgs`. I've only done a cursory check that the output is still sane --- there is a chance that some "unchecked" tests have already accumulated regressions and this just cements them! 

In the other direction, I did identify two rotten tests:
* a minor one in `setStructInstNotation.lean`, where a comment says `Set Nat`, but `#check` actually prints `?_`. Weird?
* `CompilerProbe.lean` is generating empty output, apparently indicating that something is broken, but I don't know the signficance of this file.

In any case, I'll ask about these elsewhere.

(This started by noticing that a recent `grind` test file had an untested `trace_state`, and then got carried away.)

### DIFF
--- a/tests/lean/run/conv1.lean
+++ b/tests/lean/run/conv1.lean
@@ -131,13 +131,8 @@ example (x y : Nat) (f : Nat → Nat → Nat) (g : Nat → Nat) (h₁ : ∀ z, f
   trace_state
   simp [h₁, h₂]
 
+set_option linter.unusedVariables false
 /--
-warning: unused variable `y`
-note: this linter can be disabled with `set_option linter.unusedVariables false`
----
-warning: unused variable `h₂`
-note: this linter can be disabled with `set_option linter.unusedVariables false`
----
 info: x y : Nat
 f : Nat → Nat → Nat
 g : Nat → Nat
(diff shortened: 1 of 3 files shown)


nomeata:
Thanks a lot!

Once that is taken care of, I’d like to propose that a `run` test with any output is considered failing, to avoid mistakes.

(And then we could actually remove special handling of `run` and treat an absent `.expected` file as indicating that the expected output is empty. Maybe with an opt-in escape hatch for “really ignore this output, even though I cannot use `#guard_msg` for it.)

semorrison:
> Once that is taken care of, I’d like to propose that a `run` test with any output is considered failing

We're not quite there yet. There are many `#check` statements that I though it was pointless to wrap in `#guard_msgs`. The output here is irrelevant, the point in the test file is that it typechecks.

There are also many cases where a test either #prints a large statement, or looks a IR output, or long simp/defeq/etc trace output... So there is more thought and cleanup required before we can silence tests.

I guess the intermediate solution is to split `run` into the quiet ones and the noisy ones.

nomeata:
I'd the point of the check command is not it's output, you can make that explicit using `#guard_msgs() in`, can't you?

digama0:
> I'd the point of the check command is not it's output, you can make that explicit using `#guard_msgs() in`, can't you?

I think you mean `#guard_msgs(drop info)`

PR #4177, opened 2024-05-15
feat: simp to still work even if one simp arg does not work

nomeata:
this fixes a usability paper cut that just annoyed me. When editing a
larger simp proof, I usually want to see the goal state after the simp,
and this is what I see while the `simp` command is complete. But then,
when I start typing, and necessarily type incomplete lemma names, that
error makes `simp` do nothing again and I see the original goal state.
In fact, if a prefix of the simp theorem name I am typing is a valid
identifier, it jumps even more around.

With this PR, using `logException`, I still get the red squiggly lines
for the unknown identifer, but `simp` just ignores that argument and
still shows me the final goal. Much nicer.

I also demoted the message for `[-foo]` when `foo` isn’t `simp` to a
warning and gave it the correct `ref`.

See it in action here: (in the middle, when you suddenly see the terminal,
I am switching lean versions.)

https://github.com/leanprover/lean4/assets/148037/8cb3c563-1354-4c2d-bcee-26dfa1005ae0

### DIFF
--- a/src/Lean/Elab/Util.lean
+++ b/src/Lean/Elab/Util.lean
@@ -210,7 +210,7 @@ def logException [Monad m] [MonadLog m] [AddMessageContext m] [MonadOptions m] [
       logError m!"internal exception: {name}"
 
 /--
-If `x` throws an exception, catch it, turn it into a log message (using `logException`).
+If `x` throws an exception, catch it and turn it into a log message (using `logException`).
 -/
 def withLogging [Monad m] [MonadLog m] [MonadExcept Exception m] [AddMessageContext m] [MonadOptions m] [MonadLiftT IO m]
     (x : m Unit) : m Unit := do


src/Lean/Elab/Tactic/Simp.lean
```diff
@@ -201,7 +202,7 @@ def elabSimpArgs (stx : Syntax) (ctx : Simp.Context) (simprocs : Simp.SimprocsAr
               arg[0][0].getKind == ``Parser.Tactic.simpPost
           let inv  := !arg[1].isNone
           let term := arg[2]
-          match (← resolveSimpIdTheorem? term) with
+          match ← resolveSimpIdTheorem? term with
```
leodemoura, line 205: Could you please keep the parentheses here? This is the style we use in the core repo. We write nested actions `<- f a` as `(<- f a)`.

nomeata: Ah, then I have to revise my code elsewhere. I have been omitting it since the syntax doesn't require it in `if` and `match` and `unless`, and I tend to not use unnecessary parentheses.

```suggestion
          match (← resolveSimpIdTheorem?) term with
```

src/Lean/Elab/Tactic/Simp.lean
```diff
@@ -201,7 +202,7 @@ def elabSimpArgs (stx : Syntax) (ctx : Simp.Context) (simprocs : Simp.SimprocsAr
               arg[0][0].getKind == ``Parser.Tactic.simpPost
           let inv  := !arg[1].isNone
           let term := arg[2]
-          match (← resolveSimpIdTheorem? term) with
+          match (← resolveSimpIdTheorem?) term with
```
nomeata, line 205:
```suggestion
          match (← resolveSimpIdTheorem? term) with
```

src/Lean/Elab/Util.lean
```diff
@@ -209,10 +209,14 @@ def logException [Monad m] [MonadLog m] [AddMessageContext m] [MonadOptions m] [
...
 
+/--
+If `x` throws an exception, catch it, turn it into a log message (using `logException`).
+-/
 def withLogging [Monad m] [MonadLog m] [MonadExcept Exception m] [AddMessageContext m] [MonadOptions m] [MonadLiftT IO m]
-    (x : m Unit) : m Unit := do
+    (x : m Unit) : m Unit :=
   try x catch ex => logException ex
 
+
```
nomeata, line 219:
```suggestion
```

src/Lean/Elab/Util.lean
```diff
@@ -209,8 +209,11 @@ def logException [Monad m] [MonadLog m] [AddMessageContext m] [MonadOptions m] [
       let name ← id.getName
       logError m!"internal exception: {name}"
 
+/--
+If `x` throws an exception, catch it, turn it into a log message (using `logException`).
+-/
 def withLogging [Monad m] [MonadLog m] [MonadExcept Exception m] [AddMessageContext m] [MonadOptions m] [MonadLiftT IO m]
-    (x : m Unit) : m Unit := do
+    (x : m Unit) : m Unit :=
```
nomeata, line 216:
```suggestion
    (x : m Unit) : m Unit := do
```

david-christiansen:
This has also bothered me - thanks for addressing it!

nomeata:
Do you feel confident enough about the implementation to supplement my confidence and approve :-)

PR #4178, opened 2024-05-15
feat: bitblasting theorems about unsigned bitvector inequalities

alexkeizer:
This PR adds theorems that relate unsigned bitvector comparisons `BitVec.ult` and `BitVec.ule` to `BitVec.carry`. These lemmas are a prerequisite to bit-blasting these comparisons in LeanSAT.


alexkeizer:
awaiting-review

PR #4179, opened 2024-05-15
feat: getLsb_sshiftRight

bollu:
In the course of the development, I grabbed facts about right shifting over integers [from `mathlib4`](https://github.com/leanprover-community/mathlib4/blob/master/Mathlib/Data/Int/Bitwise.lean).

The core proof strategy is to perform a case analysis of the msb:
- If `msb = false`, then `sshiftRight = ushiftRight`.
- If `msb = true`. then `x >>>s  i = ~~~(~~~(x >>>u i))`. The double negation introduces the high `1` bits that one expects of the arithmetic shift.

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -660,10 +660,10 @@ theorem sshiftRight_eq_of_msb_false {x : BitVec w} {s : Nat} (h : x.msb = false)
   have hxbound : 2 * x.toNat < 2 ^ w := (BitVec.msb_eq_false_iff_two_mul_lt x).mp h
   simp only [hxbound, ↓reduceIte, Int.natCast_shiftRight, Int.ofNat_eq_coe, ofInt_natCast,
     toNat_ofNat, toNat_ushiftRight]
-  have hxshift : x.toNat >>> s < 2 ^ w := by
+  replace hxbound : x.toNat >>> s < 2 ^ w := by
     rw [Nat.shiftRight_eq_div_pow]
     exact Nat.lt_of_le_of_lt (Nat.div_le_self ..) x.isLt
-  apply Nat.mod_eq_of_lt hxshift
+  apply Nat.mod_eq_of_lt hxbound
 
 /--
 If the msb is `true`, the arithmetic shift right equals negating,


src/Init/Data/Int/Bitwise/Lemmas.lean
```diff
@@ -0,0 +1,48 @@
+/-
+Copyright (c) 2023 Siddharth Bhat. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Siddharth Bhat, Jeremy Avigad
+-- https://github.com/leanprover-community/mathlib4/blob/12b9e3064e35636d35f0c349bf2be29fb9b75bf2/Mathlib/Data/Int/Bitwise.lean#L490-L490
```
bollu, line 5: How precisely should I tag the fact that these were taken from mathlib and adapted here?

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -629,6 +633,111 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+    rcases hmsb : x.msb with rfl | rfl <;> try rfl
+    exfalso
+    have hcontra := msb_eq_true_iff_large (x := x).mp hmsb
+    omega
+
+/-- Equation theorem for 'Int.sub' when both arguments are 'negSucc' -/
+theorem Int.toNat_sub_toNat_eq_negSucc_ofLt {n m : Nat} (hlt : n < m) :
+  (n : Int) - (m : Int) = (Int.negSucc (m - 1 - n)) := by
+  rw [Int.negSucc_eq] -- TODO: consider adding this to omega cleanup set.
+  omega
```
bollu, line 678: I wonder what the canonical way to achieve this is, I was forced to write a local theorem to make progess in massaging arithmetic.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -629,6 +633,111 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+
+/-- Equation theorem for 'Int.mod' -/
+private theorem Int.mod_negSucc_eq (m : Nat) (n : Int) :
+  (Int.negSucc m) % n = Int.subNatNat (Int.natAbs n) (Nat.succ (m % Int.natAbs n)) := rfl
```
bollu, line 682: I was surprised that I needed to write such a theorem, I think I'm missing either the requisite lemma, or a tactic incantation to perform this unfolding for me?

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -629,6 +633,111 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+
+theorem sshiftRight_eq {x : BitVec n} {i : Nat} :
+    x.sshiftRight i = BitVec.ofInt n (x.toInt >>> i) := by
+  apply BitVec.eq_of_toInt_eq
+  simp [BitVec.sshiftRight]
+
+/-- The MSB of a bitvector is `true` iff its value interpreted as a natural number in 2s complement
+  is larger than 2^(half the bitwidth). -/
+theorem msb_eq_true_iff_large {x : BitVec w}
+    : x.msb = true ↔ 2 * x.toNat ≥ 2^w  := by
```
tobiasgrosser, line 646:
```suggestion
theorem msb_eq_true_iff_large {x : BitVec w} :
    x.msb = true ↔ 2 * x.toNat ≥ 2^w  := by
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -629,6 +633,111 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+  · intro h
+    rw [BitVec.msb_eq_decide]
+    simp
+    omega
+
+/-- The MSB of a bitvector is `false` iff its value is smaller than 2^(half the bitwidth).
+  This is the contrapositive of `BitVec.msb_eq_true_iff_large`.
+  TODO: this can be massively cleaned up via `contrapose`, but unfortunately this is in mathli  -/
+theorem msb_eq_false_iff_small {x : BitVec w}
+    : x.msb = false ↔ 2 * x.toNat < 2^w  := by
```
tobiasgrosser, line 663:
```suggestion
theorem msb_eq_false_iff_small {x : BitVec w} :
    x.msb = false ↔ 2 * x.toNat < 2^w  := by
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -629,6 +633,111 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+theorem getLsb_sshiftRight (x : BitVec w) (s i : Nat) :
+  getLsb (x.sshiftRight s) i =  (!decide (w ≤ i) && if s + i < w then x.getLsb (s + i) else x.msb) := by
+  rcases hmsb:x.msb with rfl | rfl
+  · simp [sshiftRight_eq_of_msb_false hmsb]
+    by_cases hi : i ≥ w <;> simp [hi]
+    · apply getLsb_ge
+      omega
+    · intros hlsb
+      apply BitVec.lt_of_getLsb _ _ hlsb
+  · by_cases hi:(i ≥ w)
```
tobiasgrosser, line 742:
```suggestion
  · by_cases hi : i ≥ w
```

src/Init/Data/Int/Bitwise/Lemmas.lean
```diff
@@ -0,0 +1,48 @@
...
+namespace Int
+
+theorem shiftRight_eq (n : Int) (s : Nat) : n >>> s = Int.shiftRight n s := rfl
+@[simp]
+theorem shiftRight_ofNat (n s : Nat) : (n : Int) >>> s = Int.ofNat (n >>> s) := rfl
+theorem natCast_shiftRight (n s : Nat) : ((↑n) : Int) >>> s = n >>> s := rfl
+
+@[simp]
+theorem shiftRight_negSucc (m : Nat) (n : Nat) :
+  -[m+1] >>> n = -[m >>>n +1] := rfl
```
tobiasgrosser, line 20:
```suggestion
    -[m+1] >>> n = -[m >>>n +1] := rfl
```

src/Init/Data/Int/Bitwise/Lemmas.lean
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Int
+
+theorem shiftRight_eq (n : Int) (s : Nat) : n >>> s = Int.shiftRight n s := rfl
+@[simp]
+theorem shiftRight_ofNat (n s : Nat) : (n : Int) >>> s = Int.ofNat (n >>> s) := rfl
+theorem natCast_shiftRight (n s : Nat) : ((↑n) : Int) >>> s = n >>> s := rfl
+
+@[simp]
+theorem shiftRight_negSucc (m : Nat) (n : Nat) :
```
tobiasgrosser, line 19:
```suggestion
theorem shiftRight_negSucc (m n : Nat) :
```

src/Init/Data/Int/Bitwise/Lemmas.lean
```diff
@@ -0,0 +1,48 @@
...
+
+theorem Int.shiftRight_shiftRight (i : Int) (m n : Nat) :
+    i >>> m >>> n = i >>> (m + n) := by
+  cases i
+  case ofNat i =>
+    simp [natCast_shiftRight, Nat.shiftRight_add]
+  case negSucc i =>
+    simp [Int.shiftRight_negSucc, Nat.shiftRight_add]
+
+theorem shiftRight_eq_div_pow (m : Int) (n : Nat) : m >>> n = m / ((((2 : Nat) ^ n) : Nat) : Int) := by
```
tobiasgrosser, line 30:
```suggestion
theorem shiftRight_eq_div_pow (m : Int) (n : Nat) :
    m >>> n = m / ((((2 : Nat) ^ n) : Nat) : Int) := by
```

tobiasgrosser (commented):
Some style comments.

src/Init/Data/Int/Bitwise/Lemmas.lean
```diff
@@ -0,0 +1,49 @@
...
+  rcases m
+  case ofNat m =>
+    simp only [Int.ofNat_eq_coe, shiftRight_eq, Int.shiftRight, Nat.shiftRight_eq_div_pow,
+      ofNat_ediv, natCast_pow, Nat.cast_ofNat_Int]
+  case negSucc m =>
+    rw [Int.shiftRight_negSucc]
+    rw [negSucc_ediv]
+    rw [Nat.shiftRight_eq_div_pow]
+    . norm_cast
+    · norm_cast
```
tobiasgrosser, line 41:
```suggestion
    · norm_cast
    · norm_cast
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -629,6 +633,119 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+  constructor
+  · intros h
+    by_cases hx : (2 * x.toNat ≥ 2^w) <;> try omega
+    exfalso
+    simp_all [msb_eq_true_iff_large (x := x).mpr (by omega)]
+  · intros h
+    rcases hmsb : x.msb with rfl | rfl <;> try rfl
+    exfalso
+    have hcontra := msb_eq_true_iff_large (x := x).mp hmsb
+    omega
```
alexkeizer, line 673: Is this not a straightforward corollary of `msb_eq_true_iff_large`?

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -629,6 +633,119 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+
+theorem sshiftRight_eq {x : BitVec n} {i : Nat} :
+    x.sshiftRight i = BitVec.ofInt n (x.toInt >>> i) := by
+  apply BitVec.eq_of_toInt_eq
+  simp [BitVec.sshiftRight]
+
+/-- The MSB of a bitvector is `true` iff its value interpreted as a natural number in 2s complement
+  is larger than 2^(half the bitwidth). -/
+theorem msb_eq_true_iff_large {x : BitVec w} :
+    x.msb = true ↔ 2 * x.toNat ≥ 2^w  := by
```
alexkeizer, line 646: I would prefer a more "mechanical" name, like `msb_eq_true_iff_toNat_ge`. It's not immediately clear what "large" or "small" means in this context (without looking at the theorem statement)

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -629,6 +633,119 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+    have hxbound' : ¬ (2 * x.toNat < 2 ^ (w + 1)) := by omega
+    simp only [hxbound', ↓reduceIte, toNat_ofInt, toNat_not, toNat_ushiftRight]
+    rw [Int.toNat_sub_toNat_eq_negSucc_ofLt (by omega)]
+    rw [Int.shiftRight_negSucc]
+    rw [Int.mod_negSucc_eq]
+    simp only [Int.natAbs_ofNat, Nat.succ_eq_add_one]
+    rw [Int.subNatNat_of_le]
+    simp only [Int.toNat_ofNat]
+    rw [Nat.mod_eq_of_lt]
+    omega
```
alexkeizer, line 723:
```suggestion
    rw [Nat.mod_eq_of_lt (by omega)]
```
Currenlty, it's not super clear `omega` is closing a subgoal

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -629,6 +633,119 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+/-- If the msb is `true`, the arithmetic shift right equals negating, then logical shifting right, then negating again.
+  The double negation preserves the lower bits that have been shifted,
+  and the outer negation ensures that the high bits are '1'. -/
+theorem sshiftRight_eq_of_msb_true {x : BitVec w} {s : Nat} (h : x.msb = true) :
+    (x.sshiftRight s) = ~~~((~~~x) >>> s) := by
+  rcases w with rfl | w
+  · apply BitVec.eq_of_toNat_eq
+    simp
+  · apply BitVec.eq_of_toNat_eq
+    rw [BitVec.sshiftRight_eq]
```
alexkeizer, line 711:
```suggestion
  apply BitVec.eq_of_toNat_eq
  rcases w with rfl | w
  · simp
  · rw [BitVec.sshiftRight_eq]
```

src/Init/Data/BitVec/Lemmas.lean
semorrison: `msb_eq_true_iff_two_mul_ge`

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -629,6 +634,98 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
   unfold getLsb ; simp
 
+/-! ### sshiftRight -/
+
+theorem sshiftRight_eq {x : BitVec n} {i : Nat} :
+    x.sshiftRight i = BitVec.ofInt n (x.toInt >>> i) := by
+  apply BitVec.eq_of_toInt_eq
+  simp [BitVec.sshiftRight]
+
+/-- The MSB of a bitvector is `false` iff its value is smaller than 2^(half the bitwidth). -/
```
semorrison, line 644:
```suggestion
/-- The MSB of a bitvector is `false` iff its value is smaller than 2^(bitwidth - 1). -/
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -629,6 +634,98 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+    x.msb = false ↔ 2 * x.toNat < 2^w  := by
+  cases w <;> simp [Nat.pow_succ, Nat.mul_comm _ 2, msb_eq_decide]
+
+/-- The MSB of a bitvector is `true` iff its value interpreted as a natural number in 2s complement
+  is larger than 2^(half the bitwidth). -/
+theorem msb_eq_true_iff_large {x : BitVec w} :
+    x.msb = true ↔ 2 * x.toNat ≥ 2^w  := by
+  simp [show x.msb = true ↔ ¬(x.msb = false) by simp, msb_eq_false_iff_small]
+
+/-- Equation theorem for 'Int.sub' when both arguments are 'negSucc' -/
```
semorrison, line 655: Comment seems wrong?

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -629,6 +634,98 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+theorem Int.toNat_sub_toNat_eq_negSucc_ofLt {n m : Nat} (hlt : n < m) :
+  (n : Int) - (m : Int) = (Int.negSucc (m - 1 - n)) := by
```
semorrison, line 657: I am very skeptical of introducing `Int.negSucc` into the proof at all. Can you explain why? Even better, can you avoid it?

semorrison: Also, these lemmas should be further down, immediately before the theorem that uses them.

semorrison: e.g. I would try to keep negSucc out of it, but wrap the fact critical fact `Int.shiftRight_negSucc` into a `Int.shiftRight_sub_of_lt`.

bollu: @semorrison that's fair enough, I'm happy to refactor it. 

when I write `Int.shiftRight_sub_of_lt` in `Init/Data/Int/Bitwise.lean`, I imagine I would still need the (private) theorem `Int.toNat_sub_toNat_eq_negSucc_ofLt `. Is there a different way to arrive at the same result that I'm missing? If not, I'd prefer it lives somewhere that helps the next person who stumbles upon the same need.

src/Init/Data/BitVec/Lemmas.lean
bollu: renamed accordingly.

src/Init/Data/BitVec/Lemmas.lean
bollu: @alexkeizer Note that the proof that `Nat.mod_eq_of_lt` is correct is the *larger* piece below:

```lean
    · rw [Nat.shiftRight_eq_div_pow]
      suffices (2 ^ (w + 1) - 1 - x.toNat) / 2 ^ s < 2 ^ (w + 1) - 1 by omega
      apply Nat.lt_of_le_of_lt (Nat.div_le_self _ _) (by omega)
```

--- 

I tried to clean the implementation by using a more structured proof style

```lean
    rw [Nat.mod_eq_of_lt]
    · omega
    · rw [Nat.shiftRight_eq_div_pow]
      suffices (2 ^ (w + 1) - 1 - x.toNat) / 2 ^ s < 2 ^ (w + 1) - 1 by omega
      apply Nat.lt_of_le_of_lt (Nat.div_le_self _ _) (by omega)
```

src/Init/Data/BitVec/Lemmas.lean
bollu: Resolved, since you cleaned this up!

src/Init/Data/BitVec/Lemmas.lean
bollu: @semorrison actually, on second thought, I prefer the current implementation: The lemma name `Int.shiftRight_sub_of_lt` does not convey the full extent of the theorem statement, doesn't it? The equivalent `sshiftRight` theorem would be: 

```lean
diff --git a/src/Init/Data/Int/Bitwise/Lemmas.lean b/src/Init/Data/Int/Bitwise/Lemmas.lean
index c4b29d23ce..01ade45c2e 100644
--- a/src/Init/Data/Int/Bitwise/Lemmas.lean
+++ b/src/Init/Data/Int/Bitwise/Lemmas.lean
@@ -19,7 +19,21 @@ theorem natCast_shiftRight (n s : Nat) : ((↑n) : Int) >>> s = n >>> s := rfl
 theorem shiftRight_negSucc (m n : Nat) :
     -[m+1] >>> n = -[m >>>n +1] := rfl

+/-- Equation theorem for 'Int.sub' when the arguments are `.ofNat` and
+  the result is known to be negative. -/
+private theorem toNat_sub_toNat_eq_negSucc_ofLt {n m : Nat} (hlt : n < m) :
+  (n : Int) - (m : Int) = (Int.negSucc (m - 1 - n)) := by
+  rw [Int.negSucc_eq] -- TODO: consider adding this to omega cleanup set.
+  omega
+
+/-- shiftRight of a subtraction is evaluated when
+  the result of the subtraction is negative. -/
+theorem toNat_sub_toNat_shiftRight_eq_ofLt {m n i: Nat} (h : m < n) :
+  ((m : Int) - (n : Int)) >>> i = -[((n - 1 - m) >>> i) +1] := by
+    rw [toNat_sub_toNat_eq_negSucc_ofLt (by omega)]
+    rw [shiftRight_negSucc]
```

but this theorem seems *far too specialized* for my taste --- I much prefer the status quo, where the proof factors as two steps: first evaluate the subtraction, then evaluate the shift, knowing that the subtraction was a negative number. 

Maybe I misunderstood the refactor you were suggesting?

bollu:
awaiting-review

src/Init/Data/BitVec/Lemmas.lean
semorrison: Okay, well, I haven't tried to understand the rest of your proof yet so I may be wrong here. I had in mind a theorem that didn't even have an `Int.negSucc` on the RHS. That seems really smelly --- `Int.negSucc` is necessary when setting up the basics, but it is an implementation detail and should never appear in arithmetic lemmas about `Int`.

bollu: Hopefully, here's a diff that's more tasteful:

```diff
 /-- If the msb is `true`, the arithmetic shift right equals negating, then logical shifting right, then negating again.
   The double negation preserves the lower bits that have been shifted,
   and the outer negation ensures that the high bits are '1'. -/
@@ -693,16 +694,17 @@ theorem sshiftRight_eq_of_msb_true {x : BitVec w} {s : Nat} (h : x.msb = true) :
     have hxbound : (2 * x.toNat ≥ 2 ^ (w + 1)) := BitVec.msb_eq_true_iff_two_mul_ge.mp h
     have hxbound' : ¬ (2 * x.toNat < 2 ^ (w + 1)) := by omega
     simp only [hxbound', ↓reduceIte, toNat_ofInt, toNat_not, toNat_ushiftRight]
-    rw [Int.toNat_sub_toNat_eq_negSucc_ofLt (by omega)]
-    rw [Int.shiftRight_negSucc]
+    rw [← Int.subNatNat_eq_coe]
+    rw [Int.subNatNat_of_lt (by omega)]
+    simp
     rw [Int.mod_negSucc_eq]
-    simp only [Int.natAbs_ofNat, Nat.succ_eq_add_one]
+    simp [Int.natAbs_ofNat, Nat.succ_eq_add_one]
     rw [Int.subNatNat_of_le (by omega)]
     simp only [Int.toNat_ofNat]
     rw [Nat.mod_eq_of_lt]
-    · omega
+    · rw [Nat.sub_right_comm]
+      omega
     · rw [Nat.shiftRight_eq_div_pow]
-      suffices (2 ^ (w + 1) - 1 - x.toNat) / 2 ^ s < 2 ^ (w + 1) - 1 by omega
       apply Nat.lt_of_le_of_lt (Nat.div_le_self _ _) (by omega)
```

I switched to using `subNatNat` to rewrite the statement as I wanted. It feels obscure to me, but I guess that's just knowing the API :)

semorrison: Does look better to me. Sorry this was only a superficial comment. I'll try to look closely later tonight.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -629,6 +634,89 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+  apply BitVec.eq_of_toInt_eq
+  simp [BitVec.sshiftRight]
+
+/-- The MSB of a bitvector is `false` iff its value is smaller than 2^(bitwidth - 1). -/
+theorem msb_eq_false_iff_two_mul_lt {x : BitVec w} :
+    x.msb = false ↔ 2 * x.toNat < 2^w  := by
+  cases w <;> simp [Nat.pow_succ, Nat.mul_comm _ 2, msb_eq_decide]
+
+/-- The MSB of a bitvector is `true` iff its value interpreted as a natural number in 2s complement
+  is larger than 2^(half the bitwidth). -/
```
alexkeizer, line 650:
```suggestion
  is larger than 2^(bitwidth - 1). -/
```

src/Init/Data/Int/Bitwise/Lemmas.lean
```diff
@@ -0,0 +1,59 @@
...
+-/
+prelude
+import Init.Data.Nat.Bitwise.Lemmas
+import Init.Data.Int.Bitwise
+
+namespace Int
+
+theorem shiftRight_eq (n : Int) (s : Nat) : n >>> s = Int.shiftRight n s := rfl
+@[simp]
+theorem shiftRight_ofNat (n s : Nat) : (n : Int) >>> s = Int.ofNat (n >>> s) := rfl
```
semorrison, line 14: The LHS is identical to the LHS of `natCast_shiftRight`, and the RHS unnecessarily introduces non-simp-normal-form. Please delete and use `natCast_shiftRight` instead.

src/Init/Data/Int/Bitwise/Lemmas.lean
```diff
@@ -0,0 +1,59 @@
...
+theorem natCast_shiftRight (n s : Nat) : ((↑n) : Int) >>> s = n >>> s := rfl
+
+@[simp]
+theorem shiftRight_negSucc (m n : Nat) :
+    -[m+1] >>> n = -[m >>>n +1] := rfl
```
semorrison, line 19: `>>>` is an infix operator, so this should be named `negSucc_shiftRight`.

But my general object remains, I'd really prefer to mention `negSucc` as little as possible. :-(

src/Init/Data/Int/Bitwise/Lemmas.lean
semorrison: I think as long as you copied across the authors field (✅ ), you're fine. It wouldn't hurt to have a module doc:

```
/-!
# Lemmas about bitwise operations on `Int`

Some of these lemmas have been adapted from existing material in Mathlib.
-/

File: src/Init/Data/Int/Bitwise/Lemmas.lean [hunk 21]
```diff
@@ -0,0 +1,59 @@
...
+    rw [toNat_sub_toNat_eq_negSucc_ofLt (by omega)]
+    rw [shiftRight_negSucc]
+
+theorem shiftRight_shiftRight (i : Int) (m n : Nat) :
+    i >>> m >>> n = i >>> (m + n) := by
+  rcases i with i | i <;>
+    simp [natCast_shiftRight, Int.shiftRight_negSucc, Nat.shiftRight_add]
+
+theorem shiftRight_eq_div_pow (m : Int) (n : Nat) :
+    m >>> n = m / ((((2 : Nat) ^ n) : Nat) : Int) := by
```
semorrison, line 41:
```suggestion
    m >>> n = m / ((2 ^ n) : Nat) := by
```
identical meaning.

src/Init/Data/Int/Bitwise/Lemmas.lean
```diff
@@ -0,0 +1,59 @@
...
+    simp only [Int.ofNat_eq_coe, shiftRight_eq, Int.shiftRight, Nat.shiftRight_eq_div_pow,
+      ofNat_ediv, natCast_pow, Nat.cast_ofNat_Int]
+  case negSucc m =>
+    rw [Int.shiftRight_negSucc]
+    rw [negSucc_ediv]
+    rw [Nat.shiftRight_eq_div_pow]
+    · norm_cast
+    · norm_cast
+      apply Nat.pow_pos
+      omega
```
semorrison, line 53: How about just
```
  simp only [shiftRight_eq, Int.shiftRight, Nat.shiftRight_eq_div_pow]
  split
  · simp
  · rw [negSucc_ediv _ (by norm_cast; exact Nat.pow_pos (Nat.zero_lt_two))]
    rfl
```

src/Init/Data/Int/Bitwise/Lemmas.lean
```diff
@@ -0,0 +1,59 @@
...
+  the result of the subtraction is negative. -/
+theorem toNat_sub_toNat_shiftRight_eq_ofLt {m n i: Nat} (h : m < n) :
+  ((m : Int) - (n : Int)) >>> i = -[((n - 1 - m) >>> i) +1] := by
+    rw [toNat_sub_toNat_eq_negSucc_ofLt (by omega)]
+    rw [shiftRight_negSucc]
+
+theorem shiftRight_shiftRight (i : Int) (m n : Nat) :
+    i >>> m >>> n = i >>> (m + n) := by
+  rcases i with i | i <;>
+    simp [natCast_shiftRight, Int.shiftRight_negSucc, Nat.shiftRight_add]
```
semorrison, line 38:
```suggestion
  simp only [shiftRight_eq, Int.shiftRight]
  cases i <;> simp [Nat.shiftRight_add]
```
Then you don't need `shiftRight_negSucc` at all.

src/Init/Data/Int/Bitwise/Lemmas.lean
```diff
@@ -0,0 +1,59 @@
...
+  omega
+
+/-- shiftRight of a subtraction is evaluated when
+  the result of the subtraction is negative. -/
+theorem toNat_sub_toNat_shiftRight_eq_ofLt {m n i: Nat} (h : m < n) :
+  ((m : Int) - (n : Int)) >>> i = -[((n - 1 - m) >>> i) +1] := by
+    rw [toNat_sub_toNat_eq_negSucc_ofLt (by omega)]
+    rw [shiftRight_negSucc]
+
+theorem shiftRight_shiftRight (i : Int) (m n : Nat) :
```
semorrison, line 35: This equality should be flipped and renamed to `shiftRight_add` for consistency with `Nat`.

bollu: This breaks consistency with [`Nat.shiftLeft_shiftLeft`](https://github.com/leanprover/lean4/blob/873ef2d894af80d8fc672e35f7e28bae314a1f6f/src/Init/Data/Nat/Lemmas.lean#L788-L790), however.

src/Init/Data/Int/Bitwise/Lemmas.lean
bollu: Nice, I've added this cleanup.

src/Init/Data/Int/Bitwise/Lemmas.lean
bollu: @semorrison I must admit, I don't understand the reason for shying away from writing this `simp` lemma here: It's building the basic theory of `>>>`, and this is taken [straight from mathlib](https://github.com/leanprover-community/mathlib4/blob/c9ee63a2a9d4eeb41607c930473a224b081dc2c9/Mathlib/Data/Int/Bitwise.lean#L490-L490)

bollu: I've fixed the name.

src/Init/Data/Int/Bitwise/Lemmas.lean
bollu: Done. Switched to `natCast_shiftRight`

bollu:
awaiting-review

src/Init/Data/Int/Bitwise/Lemmas.lean
semorrison: Sigh. :-) I think the flipped version is slightly more likely to be useful? Since we're going to be inconsistent either way, my slight preference is
* flip this one
* express my everlasting gratitude when you make a separate PR renaming Nat.shiftLeft_shiftLeft (with a deprecation warning).

But it's also fine to just leave as is.

src/Init/Data/Int/Bitwise/Lemmas.lean
semorrison: I just don't want people using `Int.negSucc` when they don't need to, because it's a slippery slope of ever increasing API surface area. It's meant to be an internal implementation detail of `Int`. What if we decided to change the definition of `Int` to pairs of `Nat`s modulo `(a,b) ~ (a+c, b+c)`? (Obviously we won't...)

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -641,6 +645,78 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+/-- Equation theorem for 'Int.mod' -/
+private theorem Int.mod_negSucc_eq (m : Nat) (n : Int) :
+  (Int.negSucc m) % n = Int.subNatNat (Int.natAbs n) (Nat.succ (m % Int.natAbs n)) := rfl
+
+/-- if the msb is false, the arithmetic shift right equals logical shift right -/
+theorem sshiftRight_eq_of_msb_false {x : BitVec w} {s : Nat} (h : x.msb = false) :
+    (x.sshiftRight s) = x >>> s := by
+  apply BitVec.eq_of_toNat_eq
+  rw [BitVec.sshiftRight_eq]
+  rw [BitVec.toInt_eq_toNat_cond]
```
semorrison, line 664: Combine these.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -641,6 +645,78 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+  have hxbound : 2 * x.toNat < 2 ^ w := (BitVec.msb_eq_false_iff_two_mul_lt x).mp h
+  simp only [hxbound, ↓reduceIte, Int.natCast_shiftRight, Int.ofNat_eq_coe, ofInt_natCast,
+    toNat_ofNat, toNat_ushiftRight]
+  rw [Nat.mod_eq_of_lt]
+  rw [Nat.shiftRight_eq_div_pow]
+  suffices x.toNat / 2^s ≤ x.toNat by
+    apply Nat.lt_of_le_of_lt this x.isLt
+  apply Nat.div_le_self
```
semorrison, line 672: Can you introduce a `have` statement, so the logical slow at the first `rw` is clear? (It is not immediately clear without inspecting the goal view that the first rw closes the goal and replaces it with a side condition.)

semorrison, line 672:
```suggestion
    exact Nat.lt_of_le_of_lt (Nat.div_le_self _ _) x.isLt
```
The `suffices` seems an unnecessary detour here.

src/Init/Data/Int/Bitwise/Lemmas.lean
```diff
@@ -0,0 +1,51 @@
...
+  rw [Int.negSucc_eq] -- TODO: consider adding this to omega cleanup set.
+  omega
+
+/-- shiftRight of a subtraction is evaluated when
+  the result of the subtraction is negative. -/
+theorem toNat_sub_toNat_shiftRight_eq_ofLt {m n i: Nat} (h : m < n) :
+  ((m : Int) - (n : Int)) >>> i = -[((n - 1 - m) >>> i) +1] := by
+    rw [toNat_sub_toNat_eq_negSucc_ofLt (by omega)]
+    rw [negSucc_shiftRight]
+
```
semorrison, line 33: These are no longer used, right?

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -641,6 +645,78 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+  · rw [BitVec.sshiftRight_eq]
+    rw [BitVec.toInt_eq_toNat_cond]
+    have hxbound : (2 * x.toNat ≥ 2 ^ (w + 1)) := (BitVec.msb_eq_true_iff_two_mul_ge x).mp h
+    have hxbound' : ¬ (2 * x.toNat < 2 ^ (w + 1)) := by omega
+    simp only [hxbound', ↓reduceIte, toNat_ofInt, toNat_not, toNat_ushiftRight]
+    rw [← Int.subNatNat_eq_coe]
+    rw [Int.subNatNat_of_lt (by omega)]
+    simp only [Nat.pred_eq_sub_one, Int.negSucc_shiftRight]
+    rw [Int.mod_negSucc_eq]
+    simp [Int.natAbs_ofNat, Nat.succ_eq_add_one]
```
semorrison, line 692: This should be a `simp only`.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -641,6 +645,78 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+    rw [Int.subNatNat_of_le (by omega)]
+    simp only [Int.toNat_ofNat]
+    rw [Nat.mod_eq_of_lt]
```
semorrison, line 695: In fact, these can all be `rw`, and combined into a single `rw` statement.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -641,6 +645,78 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+  and the outer negation ensures that the high bits are '1'. -/
+theorem sshiftRight_eq_of_msb_true {x : BitVec w} {s : Nat} (h : x.msb = true) :
+    (x.sshiftRight s) = ~~~((~~~x) >>> s) := by
+  apply BitVec.eq_of_toNat_eq
+  rcases w with rfl | w
+  · simp
+  · rw [BitVec.sshiftRight_eq]
+    rw [BitVec.toInt_eq_toNat_cond]
+    have hxbound : (2 * x.toNat ≥ 2 ^ (w + 1)) := (BitVec.msb_eq_true_iff_two_mul_ge x).mp h
+    have hxbound' : ¬ (2 * x.toNat < 2 ^ (w + 1)) := by omega
```
semorrison, line 686:
```suggestion
    have hxbound : (2 * x.toNat ≥ 2 ^ (w + 1)) := (BitVec.msb_eq_true_iff_two_mul_ge x).mp h
    replace hxbound : ¬ (2 * x.toNat < 2 ^ (w + 1)) := by omega
```
is perhaps slightly clearer, so it's obvious that you no longer need the first bound.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -641,6 +645,78 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+    rw [Int.subNatNat_of_le (by omega)]
+    simp only [Int.toNat_ofNat]
+    rw [Nat.mod_eq_of_lt]
+    · rw [Nat.sub_right_comm]
+      omega
+    · rw [Nat.shiftRight_eq_div_pow]
+      apply Nat.lt_of_le_of_lt (Nat.div_le_self _ _) (by omega)
+
+theorem getLsb_sshiftRight (x : BitVec w) (s i : Nat) :
+  getLsb (x.sshiftRight s) i =  (!decide (w ≤ i) && if s + i < w then x.getLsb (s + i) else x.msb) := by
```
semorrison, line 702:
```suggestion
  getLsb (x.sshiftRight s) i = (!decide (w ≤ i) && if s + i < w then x.getLsb (s + i) else x.msb) := by
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -641,6 +645,78 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+  rcases hmsb:x.msb with rfl | rfl
```
semorrison, line 703:
```suggestion
  rcases hmsb : x.msb with rfl | rfl
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -641,6 +645,78 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+    · simp only [hi, decide_False, Bool.not_false, Bool.true_and, Bool.iff_and_self,
+      decide_eq_true_eq]
+      intros hlsb
+      apply BitVec.lt_of_getLsb _ _ hlsb
+  · by_cases hi : i ≥ w
+    · simp [hi]
+    · simp only [sshiftRight_eq_of_msb_true hmsb, getLsb_not, getLsb_ushiftRight, Bool.not_and,
+      Bool.not_not, hi, decide_False, Bool.not_false, Bool.if_true_right, Bool.true_and,
+      Bool.and_iff_right_iff_imp, Bool.or_eq_true, Bool.not_eq_true', decide_eq_false_iff_not,
+      Nat.not_lt, decide_eq_true_eq] <;> omega
```
semorrison, line 718: Please don't use `<;>` unless there are actually multiple goals. This should just be `omega` on a separate line.

semorrison, line 718: Indent the line wrapped `simp only` lemmas.

semorrison: (and elsewhere)

src/Init/Data/BitVec/Lemmas.lean
semorrison, line 702:
```suggestion
    getLsb (x.sshiftRight s) i =  (!decide (w ≤ i) && if s + i < w then x.getLsb (s + i) else x.msb) := by
```

semorrison, line 702: Let's limit the line length to 100.

src/Init/Data/BitVec/Lemmas.lean
semorrison: No, I think you need it. I've finally convinced myself that this proof is actually hard. :-)

Can you move this theorem back up to some `Int` file?

semorrison:
Many stylistic comments, but I'm now satisfied that the main proof really requires this much talking about `Int.negSucc`! Hopefully we can merge soon.

tobiasgrosser:
awaiting-author

src/Init/Data/Int/Bitwise/Lemmas.lean
bollu: Renamed and flipped equality to `Int.shiftRight_add`, will address the rest in a follow up commit.

src/Init/Data/Int/Bitwise/Lemmas.lean
bollu: Deleted dead lemmas.

src/Init/Data/BitVec/Lemmas.lean
bollu: Done.

bollu: Done.

src/Init/Data/BitVec/Lemmas.lean
bollu: Done.

src/Init/Data/BitVec/Lemmas.lean
bollu: Done.

src/Init/Data/BitVec/Lemmas.lean
bollu: Done.

bollu:
awaiting-review

PR #4188, opened 2024-05-16
perf: faster `Nat.testBit`

astrainfinita:
`1 &&& n` is faster than `n &&& 1` for big `n`.

---

### DIFF
--- a/src/Init/Data/Nat/Bitwise/Lemmas.lean
+++ b/src/Init/Data/Nat/Bitwise/Lemmas.lean
@@ -61,9 +61,9 @@ noncomputable def div2Induction {motive : Nat → Sort u}
   simp
 
 @[simp] theorem one_and_eq_mod_two (n : Nat) : 1 &&& n = n % 2 := by
-  match Nat.decEq n 0 with
-  | isTrue n0 => subst n0; decide
-  | isFalse n0 =>
+  if n0 : n = 0 then
+    subst n0; decide
+  else
     simp only [HAnd.hAnd, AndOp.and, land]
     cases mod_two_eq_zero_or_one n with | _ h => simp [bitwise, n0, h]
 


negiizhao:
awaiting-review

src/Init/Data/Nat/Bitwise/Basic.lean
```diff
@@ -78,6 +78,8 @@ of a number.
 -/
 
 /-- `testBit m n` returns whether the `(n+1)` least significant bit is `1` or `0`-/
-def testBit (m n : Nat) : Bool := (m >>> n) &&& 1 != 0
+def testBit (m n : Nat) : Bool :=
+  -- `1 &&& n` is faster than `n &&& 1` for big `n`. This may change in the future.
+  1 &&& (m >>> n) != 0
```
alexkeizer, line 83: Do you have numbers to confirm this claim?
Is it faster in kernel reduction or for compiled code? If it really is faster, doesn't the compiler just optimize it?

negiizhao:
```lean
def test (x : Nat) : Nat := 1 &&& x
def test' (x : Nat) : Nat := x &&& 1

def run_test (n : Nat) (x : Nat) : IO Unit := do
  let mut t := 0
  for _ in [0 : n] do
    t := test x

def run_test' (n : Nat) (x : Nat) : IO Unit := do
  let mut t := 0
  for _ in [0 : n] do
    t := test' x

def n := 10^5
def x := 11^1000000

#eval timeit "1 &&& x" (run_test n x) -- 41ms
#eval timeit "x &&& 1" (run_test' n x) -- 1.5s
```

src/Init/Data/Nat/Bitwise/Lemmas.lean
```diff
@@ -60,21 +60,17 @@ noncomputable def div2Induction {motive : Nat → Sort u}
...
-    simp only [HAnd.hAnd, AndOp.and, land] at andz
+@[simp] theorem one_and_eq_mod_two (n : Nat) : 1 &&& n = n % 2 := by
+  match Nat.decEq n 0 with
+  | isTrue n0 => subst n0; decide
+  | isFalse n0 =>
     simp only [HAnd.hAnd, AndOp.and, land]
-    unfold bitwise
-    cases mod_two_eq_zero_or_one x with | _ p =>
-      simp [xz, p, andz, one_div_two, mod_eq_of_lt]
+    cases mod_two_eq_zero_or_one n with | _ h => simp [bitwise, n0, h]
```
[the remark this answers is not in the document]
negiizhao: The commutativity of `&&&` is in Mathlib.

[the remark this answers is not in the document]
semorrison: Yes, I don't think we can just drop `and_one_is_mod` without a deprecation.

Perhaps move commutativity of `&&&` up?

src/Init/Data/Nat/Bitwise/Basic.lean
alexkeizer: That's a surprisingly big difference, almost suspiciously so. I only know enough about micro-benchmarking to know results can sometimes be misleading, so I tried playing with various variations (e.g., sprinkling `no_inline`) but everything points towards `1 &&& x` indeed being significantly faster.

Admittedly, In my original comment I was only thinking about "small" (up-to 64 bit) numbers, is this `x &&& 1` vs `1 &&& x` a known behaviour of gmp (which I assume what is going on here, with the big numbers given in your testcase)?

src/Init/Data/Nat/Bitwise/Lemmas.lean
```diff
@@ -60,6 +60,13 @@ noncomputable def div2Induction {motive : Nat → Sort u}
   unfold bitwise
   simp
 
+@[simp] theorem one_and_eq_mod_two (n : Nat) : 1 &&& n = n % 2 := by
+  match Nat.decEq n 0 with
+  | isTrue n0 => subst n0; decide
+  | isFalse n0 =>
+    simp only [HAnd.hAnd, AndOp.and, land]
+    cases mod_two_eq_zero_or_one n with | _ h => simp [bitwise, n0, h]
```
alexkeizer, line 68: Would this proof work with `if`, as in:
```suggestion
  if n0 : n = 0 then
    subst n0; decide
  else
    simp only [HAnd.hAnd, AndOp.and, land]
    cases mod_two_eq_zero_or_one n with | _ h => simp [bitwise, n0, h]
```
This matches the proof strategy of `and_one_is_mod` and is IMO much more readable

src/Init/Data/Nat/Bitwise/Basic.lean
```diff
@@ -78,6 +78,8 @@ of a number.
 -/
 
 /-- `testBit m n` returns whether the `(n+1)` least significant bit is `1` or `0`-/
-def testBit (m n : Nat) : Bool := (m >>> n) &&& 1 != 0
+def testBit (m n : Nat) : Bool :=
+  -- `1 &&& n` is faster than `n &&& 1` for big `n`. This may change in the future.
```
alexkeizer, line 82: I'm not sure we need this comment in the code, but it would be good to add it to the PR description as explanation for why this change was made.
Presumably the only thing that might change in the future is that `n &&& 1` will be similarly optimized, not that suddenly `n &&& 1` will be significantly faster than `1 &&& n`, so the "this may change in the future" part seems unneeded?

alexkeizer (commented):
I'm not knowledgable enough about micro-benchmarks to say whether the given example doesn't have any weird inlining or other behaviour that might interfere with the numbers, but the difference does seem so significant that it makes sense to switch over.

> Maybe we could have a much faster C implementation in the future.
For 64 bit or lower, bitwise `and` and shifting already have special cased code generation. I guess for larger numbers you could optimize indexing much more by using knowledge of the gmp representation, but we'd need some good evidence that the extra maintenance burden is worth the benefit. In any case, that's not a discussion relevant to this PR, so maybe just take that comment out of the PR description.

Otherwise, I've got a minor comment about the proof style but the rest LGTM

PR #4192, opened 2024-05-16
fix: hovers on binders with metavariables

### LINKED ISSUE #4078: Rpc Error when hovering variable

nomeata:
this fixes #4078. It is an alternative fix to the one in #4137, suggested
by @kmill.

Incidentially, it makes the unused variable linter better. My theory is that
if we don’t reset the info when backtracking, the binder shows up more than
once in the info tree, and then it is considered “used”, although there are
just multiple binders.

### DIFF
--- a/tests/lean/1018unknowMVarIssue.lean.expected.out
+++ b/tests/lean/1018unknowMVarIssue.lean.expected.out
@@ -13,10 +13,6 @@ a : α
         • [.] α : some Sort.{?_uniq} @ ⟨7, 13⟩-⟨7, 14⟩
         • α : Type @ ⟨7, 13⟩-⟨7, 14⟩
       • a (isBinder := true) : α @ ⟨7, 9⟩-⟨7, 10⟩
-      • α : Type @ ⟨7, 13⟩-⟨7, 14⟩ @ Lean.Elab.Term.elabIdent
-        • [.] α : some Sort.{?_uniq} @ ⟨7, 13⟩-⟨7, 14⟩
-        • α : Type @ ⟨7, 13⟩-⟨7, 14⟩
-      • a (isBinder := true) : α @ ⟨7, 9⟩-⟨7, 10⟩
       • Fam2 α β : Type 1 @ ⟨7, 21⟩-⟨7, 29⟩ @ Lean.Elab.Term.elabApp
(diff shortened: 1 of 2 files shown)


nomeata:
The changes to the test output looks reasonable, in one case we see that the info tree no longer contains entries from a back-tracked attempt at elaboration.

nomeata:
Interesting, this makes the unused-variable-linter better (see https://github.com/leanprover-community/batteries/pull/802

kmill (approved):
This looks good to me (auto implicits should act as if they were always there, right?), though it'd be good to get another opinion since I'm not completely sure why `restore` doesn't restore info trees to begin with.

Kha:
> I'm not completely sure why `restore` doesn't restore info trees to begin with.

I don't know. It might be good to double-check and document the need for this parameter but that shouldn't hold back this PR.

Kha (approved):

Kha:
IIRC the original intent was to retain information from overloading but it seems some of that was switched to `restoreInfo := true`

PR #4199, opened 2024-05-17
feat: `Nat.shiftLeft_shiftRight`

alexkeizer:
Show that shifting a natural number left and then shifting right by the same amount is a no-op.

I originally proved this in a different PR, ended up not needing the fact after all, but it still seemed like a generally useful simp lemma to have.


tobiasgrosser (approved):
This looks great.

alexkeizer:
awaiting-review

leodemoura (approved):

PR #4200, opened 2024-05-17
feat: characterize `BitVec.toInt` in terms of `BitVec.msb`

alexkeizer:
This PR extracts `msb_eq_false_iff_two_mul_lt` and `msb_eq_true_iff_two_mul_ge` from #4179, and uses them to prove a theorem that characterizes `BitVec.toInt` in terms of `BitVec.msb`. This lemma will be useful to prove a bit-blasting theorem for `BitVec.slt` and `BitVec.sle`.

Also cleans up an existing proof (`toInt_eq_toNat_cond `), which turns out to be provable by `rfl`.

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -230,7 +230,7 @@ theorem msb_eq_false_iff_two_mul_lt (x : BitVec w) : x.msb = false ↔ 2 * x.toN
   cases w <;> simp [Nat.pow_succ, Nat.mul_comm _ 2, msb_eq_decide]
 
 theorem msb_eq_true_iff_two_mul_ge (x : BitVec w) : x.msb = true ↔ 2 * x.toNat ≥ 2^w := by
-  simp [show x.msb = true ↔ ¬(x.msb = false) by simp, msb_eq_false_iff_two_mul_lt]
+  simp [← Bool.ne_false_iff, msb_eq_false_iff_two_mul_lt]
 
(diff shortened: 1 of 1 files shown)


alexkeizer:
awaiting-review

leodemoura (approved):

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -222,9 +223,21 @@ theorem toInt_eq_toNat_cond (i : BitVec n) :
...
-  unfold BitVec.toInt
-  split <;> omega
+        (i.toNat : Int) - (2^n : Nat) :=
+  rfl
+
+theorem msb_eq_false_iff_two_mul_lt (x : BitVec w) : x.msb = false ↔ 2 * x.toNat < 2^w := by
+  cases w <;> simp [Nat.pow_succ, Nat.mul_comm _ 2, msb_eq_decide]
+
+theorem msb_eq_true_iff_two_mul_ge (x : BitVec w) : x.msb = true ↔ 2 * x.toNat ≥ 2^w := by
+  simp [show x.msb = true ↔ ¬(x.msb = false) by simp, msb_eq_false_iff_two_mul_lt]
```
semorrison, line 233:
```suggestion
  simp [← Bool.ne_false_iff, msb_eq_false_iff_two_mul_lt]
```

semorrison: Despite me recently giving you an example of `show ...` inside a `simp`, it's probably worth avoiding where possible!

PR #4205, opened 2024-05-17
chore: fix the `MonadStore` type classes, with semiOutParam

JovanGerb:
The type class `MonadStore1` and friends have an outParam, which should not be an outParam, because there are multiple possible values for this parameter. At this function [fetchOrCreate](https://github.com/leanprover/lean4/blob/1382e9fbc4877820570d99f8a1226081a4a41750/src/lake/Lake/Load/Main.lean#L196C49-L196C63), there are multiple stacked `StateT` monad transformers that each give a different instance to `MonadStore1`. It is an implementation detail of type class synthesis which instance is found. This particular type class synthesis fails when the unused instance `Lake.instMonadStore1OfMonadDStoreOfFamilyOut` is set to a lower priority, because then the synthesis order happens to go differently, so the wrong instance is found.

Replacing the outParam with a semiOutParam solves this issue. Thus, we make a new type class `MonadStore1Of`, which is the same, but with a semiOutParam. This follows the design of `MonadState` and `MonadStateOf`.

However, then it turns out that the instance cannot anymore be synthesised.

There are two instances for `MonadStore1`:
```
instance [MonadDStore κ β m] : MonadStore1 k (β k) m
instance [MonadDStore κ β m] [FamilyOut β k α] : MonadStore1 k α m 
```
The first one is problematic during unification, especially when `β` should be instantiated as a constant function. We make the second one sufficient by adding an instance for the general type family:
```
/-- The general type family -/
instance (priority := low) : FamilyDef Fam a (Fam a) where
  family_key_eq_type := rfl
```
So then we can get rid of the first instance.

### DIFF
--- a/src/lake/Lake/Util/StoreInsts.lean
+++ b/src/lake/Lake/Util/StoreInsts.lean
@@ -27,5 +27,5 @@ instance [Monad m] : MonadStore Name α (StateT (NameMap α) m) :=
   inferInstanceAs (MonadStore _ _ (StateT (RBMap ..) _))
 
 @[inline] instance [MonadDStore κ β m] [t : FamilyOut β k α] : MonadStore1Of k α m where
-  fetch? := t.family_key_eq_type ▸ fetch? k
+  fetch? := cast (by rw [t.family_key_eq_type]) <| fetch? (m := m) k
   store a := store k <| cast t.family_key_eq_type.symm a


JovanGerb:
awaiting-review

src/lake/Lake/Util/Store.lean
```diff
@@ -3,33 +3,46 @@ Copyright (c) 2022 Mac Malone. All rights reserved.
...
 
+@[inline] instance [MonadDStore κ β m] [t : FamilyOut β k α] : MonadStore1Of k α m where
+  fetch? := t.family_key_eq_type ▸ MonadDStore.fetch? k
+  store a := MonadDStore.store k <| cast t.family_key_eq_type.symm a
+
 instance [MonadLift m n] [MonadDStore κ β m] : MonadDStore κ β n where
-  fetch? k := liftM (m := m) <| fetch? k
-  store k a := liftM (m := m) <| store k a
+  fetch? k := liftM (m := m) <| MonadDStore.fetch? k
+  store k a := liftM (m := m) <| MonadDStore.store k a
```
tydeu, line 42: Why does this need qualification? What makes it so the exported `fetch?` / `store` do not work here?

tests/lean/MonadStore.lean
tydeu, line 1: While I appreciate the initiative to add tests, I am not sure this PR needs one. Tests for Lake data types do not go in `tests/lean` and `src/lake/tests` does not currently have API tests like this. So, I would just suggest removing this test.

src/lake/Lake/Util/Store.lean
```diff
@@ -3,33 +3,46 @@ Copyright (c) 2022 Mac Malone. All rights reserved.
...
+  fetch? := MonadStore1Of.fetch? k
+  store := MonadStore1Of.store k
 
 /-- A monad equipped with a dependently typed key-object store. -/
-class MonadDStore (κ : Type u) (β : outParam $ κ → Type v) (m : Type v → Type w) where
+class MonadDStore (κ : Type u) (β : semiOutParam $ κ → Type v) (m : Type v → Type w) where
   fetch? : (key : κ) → m (Option (β key))
   store : (key : κ) → β key → m PUnit
 
-instance [MonadDStore κ β m] : MonadStore1 k (β k) m where
```
tydeu, line 20: As I mentioned in the Zulip, we don't always want to use the `Family` instance (due to it requiring a `cast`).  Why did you remove this instance (and, I imagine, add the general type family instance)?

Also, from a conceptual level, there should not be a reason why `MonadStore` (which is just about equipped a monad with a key-value store) should depend on the type family implementation. That is, replace this instance with the family is conceptually equivalent to doing the same to `MonadState`, `MonadReader`, etc.

tydeu (commented):
Thanks for your work on this! ❤️ The `MonadStore(1)`/`MonadStore(1)Of` split looks great! However, I still have some outstanding questions on the design of the instances, along with a few minor organizational notes.

src/lake/Lake/Util/Store.lean
JovanGerb: Oh, I think there was an error for me because my build didn't have the additional type family instance, so there should be no need for the full name

src/lake/Lake/Util/Store.lean
JovanGerb: Mario Carneiro said that it was compiled to a no-op, so that there would be no overhead. But I'll happily to go back, replacing the general instance of `FamilyDef` with the constant instance of `FamilyDef`, and adding this `MonadStore1` instance back in.

The problem here is that this instance doesn't interact well with unification, and for that I'm "misusing" the `FamilyDef`.

src/lake/Lake/Util/Store.lean
```diff
@@ -3,33 +3,46 @@ Copyright (c) 2022 Mac Malone. All rights reserved.
...
-  fetch? := MonadDStore.fetch? k
-  store o := MonadDStore.store k o
-
 /-- A monad equipped with a key-object store. -/
 abbrev MonadStore κ α m := MonadDStore κ (fun _ => α) m
 
+@[inline] instance [MonadDStore κ β m] [t : FamilyOut β k α] : MonadStore1Of k α m where
```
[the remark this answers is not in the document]
JovanGerb: Having no instance at all for `MonadStoreOf` in this file seemed a bit sad, but with the other change that will no longer be the case.

src/lake/Lake/Util/Store.lean
tydeu: > Mario Carneiro said that it was compiled to a no-op, so that there would be no overhead. 

Yes. The concern I've heard of has to do with reasoning about things with casts. Admittedly, that is not likely relevant for Lake use of this class at the moment, but is still probably something worth avoiding if possible,

tydeu: > The problem here is that this instance doesn't interact well with unification, and for that I'm "misusing" the `FamilyDef`.

I am still confused as to what exactly is the problem here. Is it that unification has a hard time solving for  `β` and the `FamilyDef` is helping it along? That is, `FamilyDef` actually is just generally useful for solving these dependently-typed unification problems? If so, to make it more generically useful and get around the potential problem with `cast`, we could move `ofFamily` and `toFamily` into the type class itself and make them avoid a `cast` in the general/constant type family case.

JovanGerb: Yes, the issue arises because `β` is a higher order metavariable that can be inferred via `semiOutParam`s. Unification has a hard time solving for `β`, and in the type family instance it doesn't need to solve for `β` via unification, and instead `β` is inferred as a `semiOutParam` from `MonadDStore`.

JovanGerb: A different solution to this problem is to completely get rid of `MonadStore1`, and instead use `MonadDStore`. I don't actually know why we needed `MonadStore1` in the first place 😅

src/lake/Lake/Util/StoreInsts.lean
```diff
@@ -13,19 +13,19 @@ namespace Lake
...
-  store k a :=  modify (·.insert k a)
+  store k a := modify (·.insert k a)
 
 instance [Monad m] : MonadStore Name α (StateT (NameMap α) m) :=
   inferInstanceAs (MonadStore _ _ (StateT (RBMap ..) _))
 
-@[inline] instance [MonadDStore κ β m] [t : FamilyOut β k α] : MonadStore1 k α m where
-  fetch? := cast (by rw [t.family_key_eq_type]) <| fetch? (m := m) k
+@[inline] instance [MonadDStore κ β m] [t : FamilyOut β k α] : MonadStore1Of k α m where
+  fetch? := t.family_key_eq_type ▸ fetch? k
```
tydeu, line 30: I just noticed this, but why did you restyle this? I think consistently using `cast` over other proof tools  when needed makes more sense.

JovanGerb: ok, I'll revert

tydeu:
Okay, everything looks good to me! Thank you the time you spent on this PR and explaining the situation to me! ❤️

PR #4223, opened 2024-05-19
chore: decrease priority for `MonadError`

### LINKED ISSUE #4212: RFC: type class synthesis order when `outParam` has multiple possible values
### Proposal

Classes such as `MonadState`, `MonadExcept`, `MonadReader` have an outParam, and they also come with a version with a `semiOutParam`, namely `MonadStateOf`, `MonadExceptOf` and `MonadReaderOf`. The value that gets picked for the `outParam` depends on the order in which instances are applied during type class resolution. This can be influenced by setting priorities on the instances, but this is not currently done.  The convention for monad transformers tends to be that the outermost transformer is the one that gives the outParam instance and programmers rely on this a lot.

### Bad instance example
```
example (e : ε) : ExceptT ε MetaM Unit := throw e -- error
```
gives an error, because the instance generated for `MonadExcept` is the one that comes from the `CoreM` instance. So, we could increase the priority of the instance `MonadExcept ε (ExceptT ε)`. This would make it much more convenient to use `ExceptT` in this scenario.

Similarly, I propose to give a high instance priority to `MonadState σ (StateT σ)`, `MonadReader ρ (ReaderT ρ)` etc.. These won't have any direct effect as far as I know of, but will make sure that no future instances will accidentally mess with the expected behaviour of these classes.

original discussion on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Ways.20to.20speed.20up.20Mathlib/near/439319211)

JovanGerb:
I decreased the priority of the `extends` instances of `MonadError`, because this type class has no global instances at all (meaning that these instances are rarely useful), and these instances commonly lead to long type class searches with results that are more complicated than necessary.

A result of this is that the `MonadError` instance generated when multiple instances are available is the one corresponding to the outer-most monad transformer. The current behaviour is that the `MonadError` instance is tried first: after the change, `MonadExcept _ (ExceptT Nat MetaM)` gets `Nat` as the error type instead of `Exception`. Similarly for `MonadExcept _ (OptionT MetaM)` , which gets error type `Unit`. In practice you would never use this error instance of `OptionT`, so I've removed this instance.

Making this change is a bit dangerous, because any `try ... catch _ => ...` block could change it's meaning while still passing through the type checker. I've found only one case of this, and added the `: Exception` annotation.

closes #4212

### DIFF
--- a/tests/lean/run/macro.lean
+++ b/tests/lean/run/macro.lean
@@ -40,7 +40,7 @@ syntax "⋃ " term ", " term : term
 macro_rules
 | `(⋃ $b, $r) => `(Union {$b:term | $r})
 
-/-- info: Union (setOf fun x => x = x) : Set ?m.4501 -/
+/-- info: Union (setOf fun x => x = x) : Set ?m.4491 -/
 #guard_msgs in
 #check ⋃ x,              x = x
 
@@ -48,7 +48,7 @@ macro_rules
 #guard_msgs in
 #check ⋃ (x : Set Unit), x = x
 
-/-- info: Union (setOf fun x => mem x univ ∧ x = x) : Set ?m.4538 -/
+/-- info: Union (setOf fun x => mem x univ ∧ x = x) : Set ?m.4528 -/
 #guard_msgs in
 #check ⋃ x ∈ univ,       x = x
 


semorrison:
Do you have evidence there is a significant performance impact?

JovanGerb:
!bench

JovanGerb:
@semorrison Yes, I've run `!bench` on mathlib, and many files in `Batteries`, and `Mathlib.Tactic` have sped up. For example

Batteries.Tactic.NoMatch | 2,467 mld. | -118,849 mln. | -4,818 %
Batteries.Tactic.Case   | 7,719 mld. | -252,05 mln. | -3,265 %
Mathlib.Tactic.Congr! | 15,574 mld. | -386,562 mln. | -2,482 %

Also introducing a new notation has become faster: for example `Mathlib.Data.Bracket` has sped up by 1,341%.

PR #4224, opened 2024-05-20
fix: missing `withIncRecDepth` and `unifyEqs?` and add support for offsets at `unifyEq?`

### LINKED ISSUE #4219: Stackoverflow when using `cases` on numeric expressions
### Prerequisites

### Description

The following code causes lean to crash with a stack overflow:

```lean
example (h : 20001 = 20000) : False := by
  cases h
```

### Steps to Reproduce

1. Run the code above in the lean web editor

**Expected behavior:** Either success, or a message from the lean server telling me that I exceeded the recursion depth.

**Actual behavior:** Segfault from the lean server, which has overflowed the stack due to recursion in `Lean.instantiateExprMVars`.

### Versions

* 4.9.0-nightly-2024-05-19
* 4.8.0rc1

### Additional Information

I get the expected recursion depth error with the smaller test of
```
example (h : 2000 = 2001) : False := by
  cases h
```

The summary here is the same as in https://github.com/leanprover/lean4/issues/4144#issuecomment-2107257336; though in that case the recursion was a bug. Here, the recursion is expected, but the mechanism of the crash is not. Again, inserting a call to `withIncRecDepth` would relax the segfault to a monadic exception, but doing so requires a larger import refactor.

### Impact

leodemoura:
Given `h` with type `x + k = y + k'` (or `h : k = k')`, `cases h` produced a proof of size linear in `min k k'`. `isDefEq` has support for offset, but `unifyEq?` did not have it, and a stack overflow occurred while processing the resulting proof. This PR fixes this issue.

closes #4219

### DIFF
--- a/src/Init/Data/Nat/Linear.lean
+++ b/src/Init/Data/Nat/Linear.lean
@@ -714,4 +714,10 @@ theorem Expr.eq_of_toNormPoly_eq (ctx : Context) (e e' : Expr) (h : e.toNormPoly
   simp [Expr.toNormPoly, Poly.norm] at h
   assumption
 
-end Nat.Linear
+end Linear
+
+def elimOffset {α : Sort u} (a b k : Nat) (h₁ : a + k = b + k) (h₂ : a = b → α) : α := by
+  simp_arith at h₁
+  exact h₂ h₁
+
+end Nat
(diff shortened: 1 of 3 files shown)


leodemoura:
Merging this one. @semorrison Reported that Mathlib issue is easy to fix.

PR #4226, opened 2024-05-20
chore: remove `#guard_msgs` from tests that rely on pointer equality

leodemoura:
Motivation: improve CI stability.


nomeata:
This undoes part of @semorrison’s #4175, right? Maybe it’s worth leaving a comment why this test can’t use `#guard_msgs`, so that next time someone cleans this up it won’t break.

Also, nobody looks at the output of `run` tests anyways, and any changes are unteste, so maybe it would be clearer to use `#guard_msgs(drop info)` here to make this clear? (Still hoping we can have all `run` test be quiet, so that any unexpected output appearing will be flagged.)

PR #4228, opened 2024-05-20
fix: add inc for double resets

anfelor:
@bollu and I also ran into the [double reset bug](https://github.com/leanprover/lean4/pull/4028) a while ago (sorry for not reporting this earlier -- we didn't have a good MWE and put it off at first). However, we fixed it slightly differently: The reference counting can still be sound if there are two reset instructions on the same variable, but then an `inc` has to be generated before the first `reset`.

The underlying reason for the bug is thus that Figure 5 of https://pp.ipd.kit.edu/uploads/publikationen/ullrich19counting.pdf has a wrong `reset` case, where the `O^+` function is not called on `reset`s, even though the variable to be reset should be owned (compare to the `Ty-Reset` rule of https://lean-lang.org/papers/beans_appendix.pdf). In contrast, Koka used to call `O^+` in this case before we switched to drop-guided reuse analysis.

This PR fixes the issue by calling `O^+` on `reset`. This might make the fix in #4028 easier to maintain since it prevents segfaults even if there are two `reset` instructions on the same variable.


Kha:
@anfelor Thank you for the PR! After discussion, we would like to ensure the invariant that we truly reset a variable at most once, mirroring the Ty-Reset rule. It would be good to have an IR checker for this at some point!

PR #4229, opened 2024-05-20
feat: define rotateLeft/Right with modulo

bollu:
This ensures that rotateLeft/Right behave correctly even when the rotation amount is larger than the bitwidth.

This shall be followed up with `getLsb` theorems for rotations for LeanSAT. 

We choose to write `aux` definitions since it is cleaner to reason about the `aux` theorems with the assumption that `rotation-amount < bit-width`, followed by auxiliary lemmas that link the behavior of rotation to the canonical case when `rotation-amount < bit-width`.

Proof strategy we will execute based on these definitions: [Link to proof of `getLsb_rotateLeft`](https://github.com/opencompl/lean4/blob/a0b18ec0f4ff9af488ced980fc31b2da81137315/src/Init/Data/BitVec/Lemmas.lean#L1129-L1204)

### DIFF
--- a/src/Init/Data/BitVec/Basic.lean
+++ b/src/Init/Data/BitVec/Basic.lean
@@ -535,7 +535,7 @@ instance {n} : HShiftLeft  (BitVec m) (BitVec n) (BitVec m) := ⟨fun x y => x <
 instance {n} : HShiftRight (BitVec m) (BitVec n) (BitVec m) := ⟨fun x y => x >>> y.toNat⟩
 
 /-- Auxiliary function for `rotateLeft`, which does not take into account the case where
-the rotateion amount is greater than the bitvector width. -/
+the rotation amount is greater than the bitvector width. -/
 def rotateLeftAux (x : BitVec w) (n : Nat) : BitVec w :=
   x <<< n ||| x >>> (w - n)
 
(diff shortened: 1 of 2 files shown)


bollu:
awaiting-review

src/Init/Data/BitVec/Basic.lean
```diff
@@ -534,6 +534,11 @@ def sshiftRight (a : BitVec n) (s : Nat) : BitVec n := .ofInt n (a.toInt >>> s)
 instance {n} : HShiftLeft  (BitVec m) (BitVec n) (BitVec m) := ⟨fun x y => x <<< y.toNat⟩
 instance {n} : HShiftRight (BitVec m) (BitVec n) (BitVec m) := ⟨fun x y => x >>> y.toNat⟩
 
+/-- Auxiliary function for `rotateLeft`, which does not take into account the case where
+the rotateion amount is greater than the bitvector width. -/
```
tobiasgrosser, line 538:
```suggestion
the rotation amount is greater than the bitvector width. -/
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1043,4 +1043,32 @@ theorem toNat_intMax_eq : (intMax w).toNat = 2^w - 1 := by
...
 
+/-! # Rotate Left -/
+
+/-- rotateLeft is invariant under `mod` by the bitwidth. -/
+@[simp]
+theorem rotateLeft_mod_eq_rotateLeft {x : BitVec w} {r : Nat} :
+    x.rotateLeft (r % w) = x.rotateLeft r := by
+  simp only [rotateLeft, Nat.mod_mod]
+
+/-- `rotateRight` equals the bit fiddling definition of `rotateRightAux` when the rotation amount is
```
tobiasgrosser, line 1054:
```suggestion
/-- `rotateLeft` equals the bit fiddling definition of `rotateLeftAux` when the rotation amount is
```

tobiasgrosser (approved):

PR #4233, opened 2024-05-20
feat: some string `simproc`s

leodemoura:
For the SSFT24 summer school.


leodemoura:
Merging since Mathlib was built successfully.

PR #4238, opened 2024-05-21
fix: docstring in `Attributes.lean`

znssong:
Changing document string in `Attributes.lean`, in order to consistent with code in `Lean.Parser.Attr`.

### DIFF
--- a/src/Lean/Attributes.lean
+++ b/src/Lean/Attributes.lean
@@ -68,7 +68,6 @@ def registerBuiltinAttribute (attr : AttributeImpl) : IO Unit := do
   We have the following ones:
   ```
   @[builtin_attr_parser] def simple     := leading_parser ident >> optional (ppSpace >> (priorityParser <|> ident))
-  /- Remark, We can't use `simple` for `class`, `instance`, `export`, and `macro` because they are keywords. -/
   @[builtin_attr_parser] def «macro»    := leading_parser "macro " >> ident
   @[builtin_attr_parser] def «export»   := leading_parser "export " >> ident
   ```


src/Lean/Util/PtrSet.lean
```diff
@@ -21,7 +21,7 @@ unsafe instance : BEq (Ptr α) where
 /--
 Set of pointers. It is a low-level auxiliary datastructure used for traversing DAGs.
 -/
-unsafe def PtrSet (α : Type) :=
+unsafe abbrev PtrSet (α : Type) :=
```
semorrison, line 24: This seems unrelated?

src/Lean/Attributes.lean
```diff
@@ -67,11 +67,10 @@ def registerBuiltinAttribute (attr : AttributeImpl) : IO Unit := do
   Helper methods for decoding the parameters of builtin attributes that are defined before `Lean.Parser`.
   We have the following ones:
   ```
-  @[builtin_attr_parser] def simple     := leading_parser ident >> optional ident >> optional priorityParser
-  /- We can't use `simple` for `class`, `instance`, `export` and `macro` because they are  keywords. -/
-  @[builtin_attr_parser] def «class»    := leading_parser "class"
-  @[builtin_attr_parser] def «instance» := leading_parser "instance" >> optional priorityParser
+  @[builtin_attr_parser] def simple     := leading_parser ident >> optional (ppSpace >> (priorityParser <|> ident))
+  /- Remark, We can't use `simple` for `class`, `instance`, `export`, and `macro` because they are keywords. -/
```
semorrison, line 71:
```suggestion
```

src/Lean/Attributes.lean
```diff
@@ -67,11 +67,10 @@ def registerBuiltinAttribute (attr : AttributeImpl) : IO Unit := do
...
   @[builtin_attr_parser] def «macro»    := leading_parser "macro " >> ident
+  @[builtin_attr_parser] def «export»   := leading_parser "export " >> ident
   ```
   Note that we need the parsers for `class`, `instance`, and `macros` because they are keywords.
```
semorrison, line 75:
```suggestion
  Note that we need the parsers for `class`, `instance`, `export` and `macros` because they are keywords.
```

src/Lean/Util/PtrSet.lean
meow-sister: Oh, sorry. I did another modification that needed to write in a new PR.

meow-sister: Now I fixed this. I moved the `PtrSet` modification to another branch.

PR #4242, opened 2024-05-21
feat: add missing theorems for `+ 1` and `- 1` normal form

markusschmaus:
`Nat.succ_eq_add_one` and `Nat.pred_eq_sub_one` are now simp lemmas. For theorems about `Nat.succ` or `Nat.pred` without corresponding theorem for `+ 1` or `- 1`, this adds the corresponding theorem.

### DIFF
--- a/src/Init/Data/Nat/Basic.lean
+++ b/src/Init/Data/Nat/Basic.lean
@@ -271,7 +271,7 @@ theorem not_add_one_le_zero (n : Nat) : ¬ n + 1 ≤ 0 := nofun
 
 theorem not_add_one_le_self : (n : Nat) → ¬ n + 1 ≤ n := Nat.not_succ_le_self
 
-theorem zero_lt_add_one (n : Nat) : 0 < n + 1 := Nat.zero_lt_succ n
+theorem add_one_pos (n : Nat) : 0 < n + 1 := Nat.zero_lt_succ n
 
 theorem succ_sub_succ_eq_sub (n m : Nat) : succ n - succ m = n - m := by
   induction m with


markusschmaus:
This comment will track theorems involving `Nat.succ`, `Nat.pred`, and `Nat.zero` and whether there is an analogue for `+ 1`, `- 1`, and `0`.

[x] `Nat.pred_succ` (simp): `Nat.sub_add_cancel` (simp), not exactly the same, but should be sufficient
[x] `Nat.succ_eq_one_add`: `Nat.add_comm`, not exactly the same, but should be sufficient
[ ] [`Nat.succ_le_of_lt`](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Nat/Basic.html#Nat.succ_le_of_lt): only `ENat.add_one_le_of_lt` available
[ ] [`Nat.pred_lt`](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Nat/Basic.html#Nat.pred_lt): [`Nat.sub_lt`](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Nat/Basic.html#Nat.sub_lt) is almost the same, but probably worth adding one for `sub_one`

src/Init/Data/Nat/Basic.lean
```diff
@@ -750,12 +824,21 @@ theorem succ_pred {a : Nat} (h : a ≠ 0) : a.pred.succ = a := by
   | zero => contradiction
   | succ => rfl
 
+theorem add_one_add_one {a : Nat} (h : a ≠ 0) : a - 1 + 1 = a := by
```
semorrison, line 827: Name.

src/Init/Data/Nat/Basic.lean
```diff
@@ -750,12 +824,21 @@ theorem succ_pred {a : Nat} (h : a ≠ 0) : a.pred.succ = a := by
...
+  | zero => contradiction
+  | succ => rfl
+
 theorem succ_pred_eq_of_pos : ∀ {n}, 0 < n → succ (pred n) = n
   | _+1, _ => rfl
 
 theorem sub_one_add_one_eq_of_pos : ∀ {n}, 0 < n → (n - 1) + 1 = n
   | _+1, _ => rfl
 
+theorem eq_zero_or_eq_add_one_sub_one : ∀ {n}, n = 0 ∨ n = (n - 1) + 1
```
semorrison, line 838: Name.

markusschmaus: I'm not sure what the name should be, this is completely analogous to `Nat.eq_zero_or_eq_succ_pred`, which follows the usual mathematical application `(· + 1) ((· - 1) n)`, which is the opposite of how it's written `n - 1 + 1`.

markusschmaus: [Nat.sub_one_add_one_eq_of_pos](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Nat/Basic.html#Nat.sub_one_add_one_eq_of_pos) follows the same logic

markusschmaus: I see that [Nat.sub_one_add_one_eq_of_pos](https://leanprover-community.github.io/mathlib4_docs/Init/Data/Nat/Basic.html#Nat.sub_one_add_one_eq_of_pos) follows the other logic, so I will use it as well.

semorrison: Yes, when naming we read arithmetic from left to right, rather than making up functions to apply. :-)

markusschmaus:
I think I've covered everything in Init.Data.Nat with a few exceptions. `List.length_cons`, `Nat.add_div_right`, and `Nat.add_div_left` are simp-lemmas which introduce `succ`, they should go into a separate PR. Two other minor exceptions are `Nat.div_eq_of_lt_le1` and `Nat.mul_sub_div` primarily due to unclear naming. There are some more in Init.Data.Int, but I think that this also should be a separate PR.

CI keeps failing, is this because I changed to doc-string for `Nat`? Should I revert this? Or is this just some infrastructure trouble which will get resolved independently?

semorrison:
There are four test failures, and they are caused by this PR. Have a go at fixing them, please, but feel free to ping if you get stuck.

semorrison:
For such a potentially disruptive change, we'll also need a successful Mathlib build. You can push changes to the `lean-pr-testing-4242` branch at Mathlib, and results will be reported back here. (If you don't have write access there, please ask.)

markusschmaus:
I think I understand what is happening.

```
[04:30 (00.000366)] lean-test-stage1> -  "```lean\nNat.succ (n : Nat) : Nat\n```\n***\nThe successor function on natural numbers, `succ n = n + 1`.\nThis is one of the two constructors of `Nat`. \n***\n*import Init.Prelude*",
[04:30 (00.001027)] lean-test-stage1> +  "```lean\nNat.succ (n : Nat) : Nat\n```\n***\nThe successor function on natural numbers, `succ n = n + 1`.\nThis is one of the two constructors of `Nat`. Using `succ n` should usually\nbe avoided in favor of `n + 1`, in order to remain compatible with the simp\nnormal form 
```

I changed the doc-string of `Nat.succ` to make clear that `+ 1` is the preferred choice. But `hover.lean.expected.out` still contains the old doc-string.

Is my change to the doc-string good? Should I revert to the prior doc-string? Or do you have a better suggestion?

If the doc-string is good, I would just copy-paste the output as it currently is produced into `hover.lean.expected.out`.

markusschmaus:
I fixed some failing tests. Keeping the doc-string as is and changing the tests.

The remaining failures also show up in PR #4263, so I wait until they are resolved there, in the hopes it will resolve them here as well.

markusschmaus:
I think I still need to re-trigger the Mathlib CI. I'm trying to figure out how this process works.

markusschmaus:
In the other `lean-pr-testing-` branches it's always [leanprover-community-mathlib4-bot](https://github.com/leanprover-community/mathlib4/commits?author=leanprover-community-mathlib4-bot) who triggers CI, so I don't think I can do this.

markusschmaus:
I'm starting to understand how Mathlib CI works

markusschmaus:
Fixing the CI took me a while, but with a very small change to a `guard_msgs` test (379ac44), Mathlib now builds locally. Assuming the CI goes through as well, what is the next step? Should I create a PR for Mathlib for this change?

markusschmaus:
Why doesn't the Mathlib CI trigger?

semorrison:
@markusschmaus, could you provide a link showing the diff required at Mathlib? (i.e. the diff from the head of lean-pr-testing-4242 back to where it departed from `nightly-testing`). I'd like to be able to review that at the same time.

src/Init/Data/Nat/Basic.lean
```diff
@@ -112,6 +112,10 @@ attribute [simp] Nat.zero_le
 
 /-! # Helper Bool relation theorems -/
 
+theorem ble_add_one_eq_true :
+    {n m : Nat} → (ble n m) = true → (ble n (m + 1)) = true :=
+  Nat.ble_succ_eq_true
+
```
semorrison, line 118: I don't think we should be adding `add_one` analogues of `succ` theorems that are only being used in bootstrapping. I think this is a good example of one. Please be particularly skeptical of making copies of lemmas from Init.Prelude. These will just add pointless API surface.

src/Init/Data/Nat/Basic.lean
```diff
@@ -370,19 +393,40 @@ theorem le_or_eq_of_le_succ {m n : Nat} (h : m ≤ succ n) : m ≤ n ∨ m = suc
        have : succ m ≤ succ n := succ_le_of_lt this
        Or.inl (le_of_succ_le_succ this))
 
+theorem le_or_eq_of_le_add_one {m n : Nat} (h : m ≤ succ n) : m ≤ n ∨ m = n + 1 :=
```
semorrison, line 396: There's still a succ in the hypothesis?

semorrison: Can't you just prove this from `le_or_eq_of_le_succ` rather than repeating the proof?

src/Init/Data/Nat/Basic.lean
```diff
@@ -537,9 +581,14 @@ protected theorem le_iff_lt_or_eq {n m : Nat} : n ≤ m ↔ n < m ∨ n = m :=
 
 protected theorem lt_succ_iff : m < succ n ↔ m ≤ n := ⟨le_of_lt_succ, lt_succ_of_le⟩
 
+protected theorem lt_add_one_iff' : m < n + 1 ↔ m ≤ n := ⟨le_of_lt_succ, lt_succ_of_le⟩
```
semorrison, line 584: Why is there a prime here? I don't think `Nat.lt_add_one_iff` otherwise exists.

src/Init/Data/Nat/Basic.lean
```diff
@@ -571,12 +620,18 @@ attribute [simp] zero_lt_succ
 
 theorem succ_ne_self (n) : succ n ≠ n := Nat.ne_of_gt (lt_succ_self n)
 
+theorem add_one_ne_self (n) : n + 1 ≠ n := Nat.ne_of_gt (lt_succ_self n)
+
 theorem succ_le : succ n ≤ m ↔ n < m := .rfl
 
+theorem add_one_le_iff' : n + 1 ≤ m ↔ n < m := .rfl
```
semorrison, line 627: Similarly here.

src/Init/Data/Nat/Basic.lean
```diff
@@ -740,6 +809,9 @@ theorem not_eq_zero_of_lt (h : b < a) : a ≠ 0 := by
 theorem pred_lt' {n m : Nat} (h : m < n) : pred n < n :=
   pred_lt (not_eq_zero_of_lt h)
 
+theorem sub_one_lt' {n m : Nat} (h : m < n) : n - 1 < n :=
```
semorrison, line 812: Perhaps should be called `sub_one_lt_of_lt`. Prefer to only use primes for close variants of the unprimed theorem, not just to resolve a name conflict.

src/Init/Data/Nat/Basic.lean
```diff
@@ -974,16 +1075,24 @@ protected theorem sub_eq_iff_eq_add {c : Nat} (h : b ≤ a) : a - b = c ↔ a =
...
+  cases n with
+  | zero   => simp
+  | succ n =>
+    rw [Nat.add_sub_cancel, add_one_mul, Nat.add_sub_cancel]
 
 theorem mul_pred_right (n m : Nat) : n * pred m = n * m - n := by
   rw [Nat.mul_comm, mul_pred_left, Nat.mul_comm]
 
+theorem mul_sub_one_right (n m : Nat) : n * (m - 1) = n * m - n := by
+  rw [Nat.mul_comm, mul_sub_one_left, Nat.mul_comm]
```
semorrison, line 1095: These theorems are all named incorrectly to begin with. Can you change to `sub_one_mul` and `mul_sub_one`. You could rename the pred theorems too, but please use `@[deprecated (since := "2024-06-01")] abbrev mul_pred_left := @pred_mul`, etc.

src/Init/Data/Nat/Basic.lean
markusschmaus: It already exists in Mathlib with this name, I believe it should be moved to core, but I didn't know the process, so I added it with the primed name as a place holder. Knowing a little bit more now, I guess the process is to delete that theorem from [lean-pr-testing-4242](https://github.com/leanprover-community/mathlib4/compare/nightly-testing...lean-pr-testing-4242).

src/Init/Data/Nat/Basic.lean
markusschmaus: This also exists in Mathlib with the same name.

src/Init/Data/Nat/Basic.lean
markusschmaus: Sounds good to me, I can do the name change to the `pred_lt'` as well.

src/Init/Data/Nat/Basic.lean
markusschmaus: 👍 The other lemmas look all very reasonable to me, `Nat.zero_lt_succ` even has a an equivalent `+ 1` version in Mathlib ( [Nat.add_one_pos](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Init/Data/Nat/Basic.html#Nat.add_one_pos))

markusschmaus:
> @markusschmaus, could you provide a link showing the diff required at Mathlib? (i.e. the diff from the head of lean-pr-testing-4242 back to where it departed from `nightly-testing`). I'd like to be able to review that at the same time.

I'm not sure what you mean, [lean-pr-testing-4242](https://github.com/leanprover-community/mathlib4/compare/nightly-testing...lean-pr-testing-4242) shows the diff at the bottom. Do you want me to create a Mathlib PR? The only meaningful change is the one in `test/LibrarySearch/basic.lean`. But there will be two more changes if I move the theorems from Mathlib to core as I mention above.

src/Init/Data/Nat/Basic.lean
semorrison: That's right!

semorrison:
> > @markusschmaus, could you provide a link showing the diff required at Mathlib? (i.e. the diff from the head of lean-pr-testing-4242 back to where it departed from `nightly-testing`). I'd like to be able to review that at the same time.
> 
> I'm not sure what you mean, [lean-pr-testing-4242](https://github.com/leanprover-community/mathlib4/compare/nightly-testing...lean-pr-testing-4242) shows the diff at the bottom. Do you want me to create a Mathlib PR? The only meaningful change is the one in `test/LibrarySearch/basic.lean`. But there will be two more changes if I move the theorems from Mathlib to core as I mention above.

That link is all I was looking for. Sorry, could have made it myself!

semorrison:
Would you mind clicking resolve on conversations above that you're sure have been resolved satisfactorily?

semorrison:
Okay, I think I am mostly happy. It would be great if you could get a `builds-mathlib` label again (it's possible that would require rebasing onto `nightly-with-mathlib` again, and then rebasing or merging Mathlib's `lean-pr-testing-4242` so it builds off the corresponding `nightly-testing-YYYY-MM-DD`).

Or at least assure me that once this lands, merging `lean-pr-testing-4242` into `nightly-testing` will go smoothly!

markusschmaus:
[Nat.add_one_pos](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Init/Data/Nat/Basic.html#Nat.add_one_pos)
new: Nat.zero_lt_add_one[Nat.add_one_pos](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Init/Data/Nat/Basic.html#Nat.add_one_pos) already exists in Mathlib, I created the same theorem in core with the name `Nat.zero_lt_add_one` with a name analogous to `Nat.zero_lt_succ`. First, what is the correct name? Second, should I remove the theorem in Mathlib with a deprecated `abbrev`?

semorrison:
Let's go with the `_pos` name.

markusschmaus:
Everything is green. What's next?

semorrison:
Sorry, I'd lost track of this one. Reviewing again now. (I ran into problems where this PR would have helped, and was surprised it hadn't landed, oops!)

PR #4247, opened 2024-05-22
chore: set up procedure for preparing release notes

kmill:
We are switching to a new system for preparing release notes.
* Release notes will be compiled when creating a release candidate from all the commits that are part of that release.
* PRs can include suggestions for release notes in PR messages. Please use language such as "release notes" and "breaking changes" to call attention to the suggestions. Release notes are user-centric rather than developer-centric.
* For more complicated release notes, these can be put into the `releases_drafts` folder.

This solves an issue where PRs that include release notes can, when merged, have those notes appear under the wrong Lean version, since they might have been created before a release but not merged until after. It also solves merge conflicts due to multiple PRs updating the release notes.


digama0:
Where have the release notes for 4.8.0-rc2 gone? They aren't in the `releasees_drafts/4.8.0.md` file as advertised.

I think it is important to have release notes also for release candidates, as this is the first place to look for breaking changes when updating mathlib or a project depending on it.

kmill:
@digama0 They're in the release branch in RELEASES.md. I deleted them here because something went wrong with how they were merged into master, and I was going to copy them back in a future step.

semorrison:
Oops, conflict to resolve.

kmill:
@semorrison I guess case-in-point for contention and weird merges! #3940 ended up somehow adding another v4.8.0 heading on `master` 🙃

PR #4250, opened 2024-05-22
chore: suggest autofix when #guard_msgs errors

kim-em:
Not certain this is a good idea, I'll let reviewers evaluate.

(I'm a bit worried about encouraging people to blindly update the message!)

Suggested on [zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/.23guard_msgs.20variant.3A.20don't.20care.20about.20whitespace/near/440040526).


nomeata (commented):
I don't know if it's a good idea to have editor specific instructions in the info view, nvim users will be left out, it's odd in verso output and is noise in a test suite log file. Feels like a laying violation.

semorrison:
nvim users feeling left out is a feature, not a bug, but yes, I agree overall.

PR #4252, opened 2024-05-22
fix: `case` tactic in macros

leodemoura:
We must erase macro scopes for tags in `case` as we do in `cases .. with ..` and `induction .. with ..`.


leodemoura:
Merging since the Mathlib build succeeded, the failures happened during the test script.

PR #4257, opened 2024-05-23
feat: getLsb_{rotateLeft, rotateRight}

bollu:
These will be used by LeanSAT for bitblasting rotations by constant distances. 

We first reduce the case when the rotation amount is larger than the width to the case where the rotation amount is less than the width (`x.rotateLeft/Right r = x.rotateLeft/Right (r%w)`).

Then, we case analyze on the low bits versus the high bits of the rotation, where we prove equality by extensionality.

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -1123,9 +1123,9 @@ theorem getLsb_rotateLeftAux_of_geq {x : BitVec w} {r : Nat} {i : Nat} (hi : i �
 /-- When `r < w`, we give a formula for `(x.rotateRight r).getLsb i`. -/
 theorem getLsb_rotateLeft_of_le {x : BitVec w} {r i : Nat} (hr: r < w) :
     (x.rotateLeft r).getLsb i =
-      if i < r
-      then x.getLsb (w - r + i)
-      else decide (i < w) && x.getLsb (i - r) := by
+      cond (i < r)
+      (x.getLsb (w - r + i))
+      (decide (i < w) && x.getLsb (i - r)) := by
   · rw [rotateLeft_eq_rotateLeftAux_of_lt hr]
     by_cases h : i < r
     · simp [h, getLsb_rotateLeftAux_of_le h]
@@ -1134,9 +1134,9 @@ theorem getLsb_rotateLeft_of_le {x : BitVec w} {r i : Nat} (hr: r < w) :
 @[simp]
 theorem getLsb_rotateLeft {x : BitVec w} {r i : Nat}  :
     (x.rotateLeft r).getLsb i =
-      if i < r % w
-      then x.getLsb (w - (r % w) + i)
-      else decide (i < w) && x.getLsb (i - (r % w)) := by
+      cond (i < r % w)
+      (x.getLsb (w - (r % w) + i))
+      (decide (i < w) && x.getLsb (i - (r % w))) := by
   rcases w with ⟨rfl, w⟩
   · simp
   · rw [← rotateLeft_mod_eq_rotateLeft, getLsb_rotateLeft_of_le (Nat.mod_lt _ (by omega))]
@@ -1204,9 +1204,9 @@ theorem rotateRight_mod_eq_rotateRight {x : BitVec w} {r : Nat} :
 /-- When `r < w`, we give a formula for `(x.rotateRight r).getLsb i`. -/
 theorem getLsb_rotateRight_of_le {x : BitVec w} {r i : Nat} (hr: r < w) :
     (x.rotateRight r).getLsb i =
-      if i < w - r
-      then x.getLsb (r + i)
-      else decide (i < w) && x.getLsb (i - (w - r)) := by
+      cond (i < w - r)
+      (x.getLsb (r + i))
+      (decide (i < w) && x.getLsb (i - (w - r))) := by
   · rw [rotateRight_eq_rotateRightAux_of_lt hr]
     by_cases h : i < w - r
     · simp [h, getLsb_rotateRightAux_of_le h]
@@ -1215,9 +1215,9 @@ theorem getLsb_rotateRight_of_le {x : BitVec w} {r i : Nat} (hr: r < w) :
 @[simp]
 theorem getLsb_rotateRight {x : BitVec w} {r i : Nat} :
     (x.rotateRight r).getLsb i =
-      if i < w - (r % w)
-      then x.getLsb ((r % w) + i)
-      else decide (i < w) && x.getLsb (i - (w - (r % w))) := by
+      cond (i < w - (r % w))
+      (x.getLsb ((r % w) + i))
+      (decide (i < w) && x.getLsb (i - (w - (r % w)))) := by
   rcases w with ⟨rfl, w⟩
   · simp
   · rw [← rotateRight_mod_eq_rotateRight, getLsb_rotateRight_of_le (Nat.mod_lt _ (by omega))]


src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1069,8 +1069,153 @@ theorem rotateLeft_eq_rotateLeftAux_of_lt {x : BitVec w} {r : Nat} (hr : r < w)
...
+    Intuitively, grab the full width (7), then move the marker `|` by `r` to the right `(-2)`
+    Then, access the bit at `i` from the right `(+i)`.
+ -/
+theorem getLsb_rotateLeftAux_of_le {x : BitVec w} {r : Nat} {i : Nat} (hi : i < r) :
+    (x.rotateLeftAux r).getLsb i = x.getLsb (w - r + i) := by
+  rw [rotateLeftAux, getLsb_or, getLsb_ushiftRight]
+  suffices (x <<< r).getLsb i = false by
+    simp only [getLsb_shiftLeft, Bool.or_iff_right_iff_imp, Bool.and_eq_true, decide_eq_true_eq,
+      Bool.not_eq_true', decide_eq_false_iff_not, Nat.not_lt, and_imp]
+    omega
```
alexkeizer, line 1098: I'm not sure if all the lemmas given here are in the default simp-set, but if they are:
```suggestion
    simp; omega
```
Is considered an acceptable use of unsqueezed `simp` (Idem ditto for the next 3 lines)

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1069,8 +1069,153 @@ theorem rotateLeft_eq_rotateLeftAux_of_lt {x : BitVec w} {r : Nat} (hr : r < w)
...
+theorem getLsb_rotateLeftAux_of_geq {x : BitVec w} {r : Nat} {i : Nat} (hi : i ≥ r) :
+    (x.rotateLeftAux r).getLsb i = (decide (i < w) && x.getLsb (i - r)) := by
+  rw [rotateLeftAux, getLsb_or]
+  suffices (x >>> (w - r)).getLsb i = false by
+    rw [this]
+    simp only [getLsb_shiftLeft, Bool.or_false, hi]
+    have hiltr : decide (i < r) = false := by
+      simp [hi]
+    rw [hiltr]
+    simp
```
alexkeizer, line 1130: You can golf this proof by folding the various rewrite steps into a single `simp`. I'll leave it up to your sense of aesthetics as to whether this is actually better.
```suggestion
    have hiltr : decide (i < r) = false := by
      simp [hi]
    simp [getLsb_shiftLeft, Bool.or_false, hi, hiltr, this]
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1069,8 +1069,153 @@ theorem rotateLeft_eq_rotateLeftAux_of_lt {x : BitVec w} {r : Nat} (hr : r < w)
...
+    (x.rotateLeft r).getLsb i =
+      if i < r
+      then x.getLsb (w - r + i)
+      else decide (i < w) && x.getLsb (i - r) := by
+  · rw [rotateLeft_eq_rotateLeftAux_of_lt hr]
+    by_cases h : i < r
+    · simp [h, getLsb_rotateLeftAux_of_le h]
+    · rw [getLsb_rotateLeftAux_of_geq]
+      simp only [Bool.and_eq_true, decide_eq_true_eq, h, ↓reduceIte, Bool.decide_eq_true]
+      omega
```
alexkeizer, line 1146: Using `omega` to conclude `i >= r` from `\lnot (i < r)` feels a bit overkill
```suggestion
    · simp [h, getLsb_rotateLeftAux_of_geq <| Nat.ge_of_not_lt h]
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1069,8 +1069,153 @@ theorem rotateLeft_eq_rotateLeftAux_of_lt {x : BitVec w} {r : Nat} (hr : r < w)
...
+theorem getLsb_rotateRightAux_of_geq {x : BitVec w} {r : Nat} {i : Nat} (hi : i ≥ w - r) :
+    (x.rotateRightAux r).getLsb i = (decide (i < w) && x.getLsb (i - (w - r))) := by
+  rw [rotateRightAux, getLsb_or]
+  suffices (x >>> r).getLsb i = false by
+    simp only [this, getLsb_shiftLeft, Bool.false_or]
+    by_cases hiw : i < w
+    · simp only [hiw, decide_True, Bool.true_and, Bool.and_iff_right_iff_imp, Bool.not_eq_true',
+      decide_eq_false_iff_not, Nat.not_lt]
+      omega
+    · simp only [hiw, decide_False, Bool.false_and]
```
alexkeizer, line 1214:
```suggestion
    by_cases hiw : i < w 
    <;> simp [hiw, hi]
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1083,4 +1228,27 @@ theorem rotateRight_mod_eq_rotateRight {x : BitVec w} {r : Nat} :
...
+    (x.rotateRight r).getLsb i =
+      if i < w - r
+      then x.getLsb (r + i)
+      else decide (i < w) && x.getLsb (i - (w - r)) := by
+  · rw [rotateRight_eq_rotateRightAux_of_lt hr]
+    by_cases h : i < w - r
+    · simp [h, getLsb_rotateRightAux_of_le h]
+    · rw [getLsb_rotateRightAux_of_geq]
+      simp only [h, ↓reduceIte]
+      omega
```
alexkeizer, line 1242:
```suggestion
    · simp [h, getLsb_rotateRightAux_of_geq <| Nat.le_of_not_lt h]
```

alexkeizer (approved):
I've golved some of the proofs, but otherwise the proof strategy LGTM

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1069,8 +1069,145 @@ theorem rotateLeft_eq_rotateLeftAux_of_lt {x : BitVec w} {r : Nat} (hr : r < w)
     x.rotateLeft r = x.rotateLeftAux r := by
   simp only [rotateLeft, Nat.mod_eq_of_lt hr]
 
+
+/-- Accessing bits in `x.rotateLeft r` the range `[0, r)` is equal to
+    accesing bits `x` in the range `[w - r, w)`,
```
tobiasgrosser, line 1074:
```suggestion
    accessing bits `x` in the range `[w - r, w)`,
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1069,8 +1069,145 @@ theorem rotateLeft_eq_rotateLeftAux_of_lt {x : BitVec w} {r : Nat} (hr : r < w)
...
+      then x.getLsb (w - (r % w) + i)
+      else decide (i < w) && x.getLsb (i - (r % w)) := by
+  rcases w with ⟨rfl, w⟩
+  · simp
+  · rw [← rotateLeft_mod_eq_rotateLeft, getLsb_rotateLeft_of_le (Nat.mod_lt _ (by omega))]
+
 /-! ## Rotate Right -/
 
+/-- Accessing bits in `x.rotateRight r` the range `[0, w-r)` is equal to
+    accesing bits `x` in the range `[r, w)`,
```
tobiasgrosser, line 1156:
```suggestion
    accessing bits `x` in the range `[r, w)`,
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1069,8 +1069,145 @@ theorem rotateLeft_eq_rotateLeftAux_of_lt {x : BitVec w} {r : Nat} (hr : r < w)
...
+    (x.rotateRightAux r).getLsb i = x.getLsb (r + i) := by
+  rw [rotateRightAux, getLsb_or, getLsb_ushiftRight]
+  suffices (x <<< (w - r)).getLsb i = false by
+    simp only [this, Bool.or_false]
+  simp only [getLsb_shiftLeft, Bool.and_eq_false_imp, Bool.and_eq_true, decide_eq_true_eq,
+    Bool.not_eq_true', decide_eq_false_iff_not, Nat.not_lt, and_imp]
+  omega
+
+/-- Accessing bits in `x.rotateRight r` the range `[w-r, w)` is equal to
+    accesing bits `x` in the range `[0, r)`,
```
tobiasgrosser, line 1183:
```suggestion
    accessing bits `x` in the range `[0, r)`,
```

bollu:
awaiting-review

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1069,8 +1069,143 @@ theorem rotateLeft_eq_rotateLeftAux_of_lt {x : BitVec w} {r : Nat} (hr : r < w)
     x.rotateLeft r = x.rotateLeftAux r := by
   simp only [rotateLeft, Nat.mod_eq_of_lt hr]
 
+
+/-- Accessing bits in `x.rotateLeft r` the range `[0, r)` is equal to
+    accessing bits `x` in the range `[w - r, w)`,
+    given by `i : [0, r) ↦ w - r + i : [w - r, w)`.
```
semorrison, line 1075:
```suggestion
/--
Accessing bits in `x.rotateLeft r` the range `[0, r)` is equal to
accessing bits `x` in the range `[w - r, w)`,
given by `i : [0, r) ↦ w - r + i : [w - r, w)`.
```
and below. Life is too short to maintain indenting. :-)

semorrison: To be honest, I found this explanation (particularly that it uses lots of non-Lean notation!) less helpful than the theorem statement. I like examples in doc-strings, but this one needs a rewrite!

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1069,8 +1069,143 @@ theorem rotateLeft_eq_rotateLeftAux_of_lt {x : BitVec w} {r : Nat} (hr : r < w)
...
+  simp only [getLsb_ushiftRight]
+  apply getLsb_ge
+  omega
+
+/-- When `r < w`, we give a formula for `(x.rotateRight r).getLsb i`. -/
+theorem getLsb_rotateLeft_of_le {x : BitVec w} {r i : Nat} (hr: r < w) :
+    (x.rotateLeft r).getLsb i =
+      if i < r
+      then x.getLsb (w - r + i)
+      else decide (i < w) && x.getLsb (i - r) := by
```
semorrison, line 1135: I wonder (no strong opinion) if it would be better to use `cond (decide (i < r) ...` rather than `if` here, to stay in the pure `Bool` fragment. I don't think it matter a lot, and would perhaps be less readable.

semorrison:
Otherwise, I'm happy with the statements and proofs.

src/Init/Data/BitVec/Lemmas.lean
bollu: Fixed!

src/Init/Data/BitVec/Lemmas.lean
bollu: I updated to use `cond`, since it feels cleaner to me as well: it's less jarring, since it's now all uniformly low level.

bollu:
@semorrison all comments addressed.

PR #4258, opened 2024-05-23
doc: add docstrings for `dsimp` configuration

kmill:
The dsimp configuration is a subset of the simp configuration, and so it's a matter of copying the docstrings.


Kha:
Should the simp config inherit from the dsimp config?

david-christiansen (approved):

PR #4260, opened 2024-05-23
feat: make `<num>#<term>` bitvector literal notation global

leodemoura:
The bitvector literal notation (e.g., `0x1#4`) is currently scoped, but it is used when pretty-printing. This has led to confusion among users who find that the notation does not work in their code unless they include `open BitVec`.

Users are often puzzled when they try to use the bitvector literal notation directly in their code, only to find it doesn't work without explicitly opening `BitVec`. This behavior seems counterintuitive and has been a source of frustration. Additionally, this notation is popular among users for its compact and expressive representation.

- Alternative Notations:
  - Using polymorphic numerals (e.g., `1`) loses valuable bitvector size information.
  - Using the verbose form (e.g., `(1 : BitVec 4)`) is cumbersome and less readable.

@kmill and @semorrison suggested the updated syntax definition for bitvector literals:

```lean
scoped syntax:max num noWs "#" noWs term:max : term
macro_rules | `($i:num#$n) => `(BitVec.ofNat $n $i)
```

This change ensures that:
- There is no lexical conflict with existing Mathlib notations, particularly with cardinality and Finset notations.

- The notation remains intuitive and easy to use.

- Users will no longer need to remember to open `BitVec` to use the notation, making it more intuitive and less error-prone.

- Pretty-printing BitVec-heavy goals, such as those in the SSFT24 tutorial, will be more readable and less verbose.

- We can still write `0#n` to denote `BitVec.ofNat n 0`


leodemoura:
Note: can't use `nightly-with-mathlib` due to recent CI changes.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -138,21 +138,21 @@ theorem ofBool_eq_iff_eq : ∀(b b' : Bool), BitVec.ofBool b = BitVec.ofBool b'
...
 -- Remark: we don't use `[simp]` here because simproc` subsumes it for literals.
 -- If `x` and `n` are not literals, applying this theorem eagerly may not be a good idea.
 theorem getLsb_ofNat (n : Nat) (x : Nat) (i : Nat) :
-  getLsb (x#n) i = (i < n && x.testBit i) := by
+  getLsb (BitVec.ofNat n x) i = (i < n && x.testBit i) := by
   simp [getLsb, BitVec.ofNat, Fin.val_ofNat']
 
 @[simp, deprecated toNat_ofNat (since := "2024-02-22")]
-theorem toNat_zero (n : Nat) : (0#n).toNat = 0 := by trivial
+theorem toNat_zero (n : Nat) : (BitVec.ofNat n 0).toNat = 0 := by trivial
```
semorrison, line 151: Are these changes intentional? It seems fine to leave the notation in place for literals!

I can change them back if you like.

leodemoura: @semorrison Are you suggesting we use the following variant?
```
syntax:max num noWs "#" noWs term:max : term
```
That is, the bitvector size can still be a term. If it does not conflict with Mathlib notation, I am happy with the change, and I think it is a good one.

leodemoura:
@semorrison This PR is blocked by `BitVec` theorems in Mathlib. My understanding is that the Mathlib maintainers agreed that `BitVec` theorems should not be in Mathlib, thereby reducing the amount of cross-repo coordination.

eric-wieser:
I pushed a commit to the testing branch in mathlib that fixes the build; though it seems that independently there are unrelated CI issues regarding noisy lake output, which are causing the tests to fail.

semorrison:
@leodemoura, this only went green after I added a stage0 update, so it couldn't be merged via the queue. I rebase-merged it just now.

PR #4269, opened 2024-05-24
fix: accidental ownership with specialization

### LINKED ISSUE #4240: Missing inferred borrow annotation related to arrays
### Prerequisites

### Description

In the code snippet
```lean
class MyClass (α : Type u) where

instance : MyClass Nat where

inductive MyOption (α : Type u) where
  | none
  | some (key : α)

namespace MyOption

def isSomeWithInstance [MyClass α] : MyOption α → Bool
  | none => false
  | some _ => true

def isSome : MyOption α → Bool
  | none => false
  | some _ => true

end MyOption

def isSomeWithInstanceGen [MyClass α] (m : { m : Array (MyOption α) // 0 < m.size }) : Bool :=
  (m.1.uget 0 m.2).isSomeWithInstance

def isSomeWithInstanceNat (m : { m : Array (MyOption Nat) // 0 < m.size }) : Bool :=
  (m.1.uget 0 m.2).isSomeWithInstance

def isSomeWithInstanceNatEx (m : Array (MyOption Nat)) (h : 0 < m.size) : Bool :=
  (m.uget 0 h).isSomeWithInstance

def isSomeWithInstanceGenL [MyClass α] (m : { m : List (MyOption α) // 0 < m.length }) : Bool :=
  m.1.get ⟨0, m.2⟩ |>.isSomeWithInstance

def isSomeWithInstanceNatL (m : { m : List (MyOption Nat) // 0 < m.length }) : Bool :=
  m.1.get ⟨0, m.2⟩ |>.isSomeWithInstance

def isSomeWithInstanceNatExL (m : List (MyOption Nat)) (h : 0 < m.length) : Bool :=
  m.get ⟨0, h⟩ |>.isSomeWithInstance

def isSomeGen (m : { m : Array (MyOption α) // 0 < m.size }) : Bool :=
  (m.1.uget 0 m.2).isSome

def isSomeNat (m : {m : Array (MyOption Nat) // 0 < m.size }) : Bool :=
  (m.1.uget 0 m.2).isSome

def isSomeNatEx (m : Array (MyOption Nat)) (h : 0 < m.size) : Bool :=
  (m.uget 0 h).isSome
```

I would expect all nine definitions to borrow the array/list, but only eight of them do so. The odd one out is `isSomeWithInstanceNat` (the second one). If `MyClass` is changed to live in `Prop`, then all nine borrow the array/list.

### Context

This bug affects `Lean.HashMap.contains` (and presumably all other read operations, but I haven't checked this in detail), leading to unnecessary reference counting operations on every hashmap access. Note that the issue only becomes visible when looking at specialized IR for the operation. The IR generated for the polymorphic versions of the functions is fine.

### Steps to Reproduce

1. Run the above code and check the traced IR for borrow annotations

**Expected behavior:** The array/list is always borrowed

**Actual behavior:** The array is not borrowed in `isSomeWithInstanceNat`.

### Versions

4.9.0-nightly-2024-05-21 on live.lean-lang.org

### Additional Information

### Impact

Kha:
The old compiler tried to avoid borrowing that could inhibit mutual TCO, but the implementation was wrong and in an unnecessary way:
* `ownVar` is called but it can only mark variables of the current function, so the wrong variable was marked
* there was no mutual recursion as the additional declaration came from specialization
* the old compiler only supports TCO for self-calls anyway

Thus we restrict the borrow inhibition to self-calls. More extensive TCO could be considered in the new compiler.

Fixes #4240

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -582,11 +582,11 @@ if(${STAGE} GREATER 0 AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "Emscripten")
     message(FATAL_ERROR "src/lake does not exist. Please check out the Lake submodule using `git submodule update --init src/lake`.")
   endif()
 
-  #add_custom_target(lake ALL
-  #  WORKING_DIRECTORY ${LEAN_SOURCE_DIR}
-  #  DEPENDS leanshared
-  #  COMMAND $(MAKE) -f ${CMAKE_BINARY_DIR}/stdlib.make Lake
-  #  VERBATIM)
+  add_custom_target(lake ALL
+    WORKING_DIRECTORY ${LEAN_SOURCE_DIR}
+    DEPENDS leanshared
+    COMMAND $(MAKE) -f ${CMAKE_BINARY_DIR}/stdlib.make Lake
+    VERBATIM)
 endif()
 
 if(PREV_STAGE)


Kha:
!bench

Kha:
!bench

Kha:
We deemed the parser-specific slowdown (which effectively must be random chance given that the removed heuristic did not do what it intended to do) acceptable in exchange for a clear fix to the code and a modest overall stdlib boost.

PR #4271, opened 2024-05-24
feat: fat static libraries with all transitive dependencies

bollu:
Adapted from [this zulip thread titled 'reverse FFI: building a "fat" static library'](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/reverse.20FFI.3A.20building.20a.20.22fat.22.20static.20library.3F/near/440460506)

--- 

The motivation is to enable more reverse-FFI use-cases. End users should be able to build a single static library that can be linked against. This library should include all (transitive) dependencies of the Lean library. This enables easy static linking.

My personal use-case is to link a lean+mathlib+batteries library against a rust project. This is accomplished by building a static library as proposed, and linking against this in the `build.rs` build script.

---

The larger part of the PR is to add an example that showcases how to use it in a new example in a library with a dependency, which has been adapted from the already existing example of `reverse-ffi`. We add a new test that checks that reverse FFI does indeed include transitive dependencies by grepping for the `_initialize_` function of the dependency.

### DIFF
--- a/src/lake/Lake/Build/Library.lean
+++ b/src/lake/Lake/Build/Library.lean
@@ -76,7 +76,7 @@ def LeanLib.staticExportFacetConfig : LibraryFacetConfig staticExportFacet :=
 @[specialize] protected def LeanLib.buildStaticFat
 (self : LeanLib) : FetchM (BuildJob FilePath) := do
     withRegisterJob s!"{self.name}:static.fat" do
-    let mods ← self.rootModules.concatMapM fun mod => do
+    let mods ← (← self.modules.fetch).concatMapM fun mod => do
       return (← mod.transImports.fetch).push mod
     let oJobs ← mods.concatMapM fun mod =>
       mod.nativeFacets (shouldExport := true) |>.mapM fun facet => fetch <| mod.facet facet.name


bollu:
awaiting-review

src/lake/Lake/Config/LeanLib.lean
```diff
@@ -64,6 +64,10 @@ The names of the library's root modules
 @[inline] def staticExportLibFile (self : LeanLib) : FilePath :=
   self.pkg.nativeLibDir / self.staticLibFileName.addExtension "export"
 
+/-- The path to the static fat library in the package's `libDir`. -/
+@[inline] def staticFatLibFile (self : LeanLib) : FilePath :=
+  self.pkg.nativeLibDir / nameToStaticLib s!"{self.config.libName}Fat"
```
tydeu, line 69:
```suggestion
  self.pkg.nativeLibDir / self.staticLibFileName.addExtension "fat"
```

I would be inclined to give this extension `.fat` like the `export` and `noexport` variants. When Is there a reason you avoided that approach?

src/lake/Lake/Build/Library.lean
```diff
@@ -73,6 +73,20 @@ def LeanLib.staticFacetConfig : LibraryFacetConfig staticFacet :=
 def LeanLib.staticExportFacetConfig : LibraryFacetConfig staticExportFacet :=
   mkFacetJobConfig (LeanLib.recBuildStatic · true)
 
+@[specialize] protected def LeanLib.buildStaticFat
+(self : LeanLib) : FetchM (BuildJob FilePath) := do
+    withRegisterJob s!"{self.name}:static.fat" do
+    let mods ← self.rootModules.concatMapM fun mod => do
```
tydeu, line 79: This should be using `self.modules.fetch`, not `rootModules` directly.

src/lake/Lake/Build/Library.lean
```diff
@@ -73,6 +73,20 @@ def LeanLib.staticFacetConfig : LibraryFacetConfig staticFacet :=
...
+      return (← mod.transImports.fetch).push mod
+    let oJobs ← mods.concatMapM fun mod =>
+      mod.nativeFacets (shouldExport := true) |>.mapM fun facet => fetch <| mod.facet facet.name
```
tydeu, line 82: Should the fat library really have exported symbols? Exported symbols are generally only for dynamic libraries. Statically linking a library will cause symbol resolution regardless of export.

src/lake/Lake/Build/Library.lean
```diff
@@ -73,6 +73,20 @@ def LeanLib.staticFacetConfig : LibraryFacetConfig staticFacet :=
 def LeanLib.staticExportFacetConfig : LibraryFacetConfig staticExportFacet :=
   mkFacetJobConfig (LeanLib.recBuildStatic · true)
 
+@[specialize] protected def LeanLib.buildStaticFat
+(self : LeanLib) : FetchM (BuildJob FilePath) := do
+    withRegisterJob s!"{self.name}:static.fat" do
+    let mods ← self.rootModules.concatMapM fun mod => do
+      return (← mod.transImports.fetch).push mod
```
tydeu, line 80: To avoid aborting the entire build process on ill-formed imports, this be wrapped in a `try ... catch` block like the following example from `Module`:

https://github.com/leanprover/lean4/blob/fe0cb97c5d4d43393728ddd885415bc5bfb4b929/src/lake/Lake/Build/Module.lean#L111-L112

src/lake/Lake/Build/Library.lean
```diff
@@ -73,6 +73,20 @@ def LeanLib.staticFacetConfig : LibraryFacetConfig staticFacet :=
...
+    let oJobs ← mods.concatMapM fun mod =>
+      mod.nativeFacets (shouldExport := true) |>.mapM fun facet => fetch <| mod.facet facet.name
+    let libFile := self.staticFatLibFile
+    let lib ← buildStaticLib libFile oJobs
+    return lib
```
tydeu, line 85:
```suggestion
    buildStaticLib libFile oJobs
```

tydeu (commented):
I have finally reviewed this PR. My apologies for taking so long! 🙏 

There are some elements that need discussion and potential rework -- see the line-by-line comments for details. However, overall, I think you are on the right track, and I would be happy to see this PR merged!

digama0:
I think it is a bit misleading to call this a "fat library", and suggest using the term "statically linked" instead. "fat binaries" are the usual term for binaries which simultaneously contain code from multiple architectures (e.g. x86 and ARM), with some header information so that they can be correctly run in multiple environments. The technique is fairly rarely used; a notable recent application of the technique is in Apple binaries during the ARM transition period. See also [Fat binary](https://en.wikipedia.org/wiki/Fat_binary).

src/lake/Lake/Config/LeanLib.lean
bollu: @tydeu I didn't want to hamper the default behaviour of `ld` searching for `-lfoo` as `libfoo.a`, so I preferred to keep the `.a` as the rightmost extension.

tydeu:
@digama0 

> I think it is a bit misleading to call this a "fat library", and suggest using the term "statically linked" instead.

Statically linked is too general for this. The standard static library Lake produces is also statically linked (as would be any static library). Furthermore, it is perfectly normal for a static library to not contain  symbols from other (public) libraries, so it would weird to use the general term for the currently name "fat" version. 

Maybe we could call this a "standalone" static library instead of a "fat" one?

src/lake/Lake/Config/LeanLib.lean
tydeu: @bollu My thought process is that I wouldn't expect a user to be shipping all 3 different variants of the static library but instead just distribute the appropriate one as `libfoo.a`.  Internally, in Lake, the full name can be easily used without concern. 

One problem with having a suffix just for this version of the library is that it can conflict with a library name ending in a normal library ending with the same suffix. For example, a Lean library named `BigFat`'s normal static library would conflict with the Lean library `Big`'s fat static library. As this would be a very confusing issue for a user to debug, it would probably be best to ensure all static libraries have non-overlapping suffixes (or none) to avoid this.

src/lake/Lake/Build/Library.lean
```diff
@@ -73,6 +73,19 @@ def LeanLib.staticFacetConfig : LibraryFacetConfig staticFacet :=
...
 
+@[specialize] protected def LeanLib.buildStaticFat
+(self : LeanLib) : FetchM (BuildJob FilePath) := do
+    withRegisterJob s!"{self.name}:static.fat" do
+    let mods ← self.rootModules.concatMapM fun mod => do
+      return (← mod.transImports.fetch).push mod
+    let oJobs ← mods.concatMapM fun mod =>
+      mod.nativeFacets (shouldExport := true) |>.mapM fun facet => fetch <| mod.facet facet.name
+    let libFile := self.staticFatLibFile
+    buildStaticLib libFile oJobs
```
tydeu, line 84:
```suggestion
    buildStaticLib self.staticFatLibFile oJobs
```

Sorry for not catching this on the first review, but it seems stylistically best to do this all one one line.

src/lake/Lake/Build/Library.lean
```diff
@@ -73,6 +73,18 @@ def LeanLib.staticFacetConfig : LibraryFacetConfig staticFacet :=
 def LeanLib.staticExportFacetConfig : LibraryFacetConfig staticExportFacet :=
   mkFacetJobConfig (LeanLib.recBuildStatic · true)
 
+@[specialize] protected def LeanLib.buildStaticFat
+(self : LeanLib) : FetchM (BuildJob FilePath) := do
+    withRegisterJob s!"{self.name}:static.fat" do
+    let mods ← self.rootModules.concatMapM fun mod => do
+      return (← mod.transImports.fetch).push mod
```
tydeu, line 80:
```suggestion
    let mods ← self.modules.fetch
    let mods ← 
      try 
        mods.concatMapM fun mod => do
          return (← mod.transImports.fetch).push mod
      catch errPos =>
        return Job.error (← takeLogFrom errPos)
```

Here is a possible adaption incorporating the changes I suggested regarding `self.modules.fetch` and the `try .. catch` block. (I have not type checked this, so there may be a typo.)

tydeu:
```suggestion
    let mods ← (← self.modules.fetch).concatMapM fun mod => do
      return (← mod.transImports.fetch).push mod
```


Due to some touchups on master, the `try ... catch` block is no longer needed.

PR #4276, opened 2024-05-25
perf: use `hasExprMvar` in type class search

JovanGerb:
I'm wondering if this change has a significant effect on Mathlib, because there might be lots of universe metavariables flying around there that we don't care about.

---

Closes #0000 (`RFC` or `bug` issue number fixed by this PR, if any)


semorrison:
The spurious Mathlib failure from lake being noisy in tests should hopefully be resolved by tomorrow.

PR #4301, opened 2024-05-28
feat: add a linter for local vars that clash with their constructors

david-christiansen:
This came up when watching new Lean users in a class situation. A number of them were confused when they omitted a namespace on a constructor name, and Lean treated the variable as a pattern that matches anything.

For example, this program is accepted but may not do what the user thinks:
```
inductive Tree (α : Type) where
  | leaf
  | branch (left : Tree α) (val : α) (right : Tree α)

def depth : Tree α → Nat
  | leaf => 0
```
Adding a `branch` case to `depth` results in a confusing message.

With this linter, Lean marks `leaf` with:
```
Local variable 'leaf' resembles constructor 'Tree.leaf' - write '.leaf' (with a dot) or 'Tree.leaf' to use the constructor.
note: this linter can be disabled with `set_option linter.constructorNameAsVariable false`
```

Additionally, the error message that occurs when invalid names are applied in patterns now suggests similar names. This means that:
```
def length (list : List α) : Nat :=
  match list with
  | nil => 0
  | cons x xs => length xs + 1
```
now results in the following warning on `nil`:
```
warning: Local variable 'nil' resembles constructor 'List.nil' - write '.nil' (with a dot) or 'List.nil' to use the constructor.
note: this linter can be disabled with `set_option linter.constructorNameAsVariable false`
```

and error on `cons`:
```
invalid pattern, constructor or constant marked with '[match_pattern]' expected

Suggestion: 'List.cons' is similar
```

The list of suggested constructors is generated before the type of the pattern is known, so it's less accurate, but it truncates the list to ten elements to avoid being overwhelming. This mostly comes up with `mk`.

### DIFF
--- a/src/Lean/Data/Lsp/LanguageFeatures.lean
+++ b/src/Lean/Data/Lsp/LanguageFeatures.lean
@@ -333,8 +333,8 @@ def SemanticTokenType.names : Array String :=
     "event", "method", "macro", "modifier", "comment", "string", "number",
     "regexp", "operator", "decorator", "leanSorryLike"]
 
-def SemanticTokenType.toNat (type : SemanticTokenType) : Nat :=
-  type.toCtorIdx
+def SemanticTokenType.toNat (tokenType : SemanticTokenType) : Nat :=
+  tokenType.toCtorIdx
 
 -- sanity check
 -- TODO: restore after update-stage0
--- a/src/Lean/Linter/ConstructorAsVariable.lean
+++ b/src/Lean/Linter/ConstructorAsVariable.lean
@@ -1,0 +1,83 @@
+/-
+Copyright (c) 2024 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: David Thrane Christiansen
+-/
+prelude
+import Lean.Elab.Command
+import Lean.Linter.Util
+
+set_option linter.missingDocs true
+
+namespace Lean.Linter
+
+open Lean Elab Command
+open Lean.Linter (logLint)
+
+/--
+A linter that warns when bound variable names are the same as constructor names for their types,
+modulo namespaces.
+ -/
+register_builtin_option linter.constructorNameAsVariable : Bool := {
+  defValue := true,
+  descr := "enable the linter that warns when bound variable names are nullary constructor names"
+}
+
+/--
+Reports when bound variables' names overlap with constructor names for their type. This is to warn
+especially new users that they have built a pattern that matches anything, rather than one that
+matches a particular constructor. Use `linter.constructorNameAsVariable` to disable.
+-/
+def constructorNameAsVariable : Linter where
+  run cmdStx := do
+    let some cmdStxRange := cmdStx.getRange?
+      | return
+
+    let infoTrees := (← get).infoState.trees.toArray
+    let warnings : IO.Ref (Lean.HashMap String.Range (Syntax × Name × Name)) ← IO.mkRef {}
+
+    for tree in infoTrees do
+      tree.visitM' (preNode := fun ci info _ => do
+        match info with
+        | .ofTermInfo ti =>
+          match ti.expr with
+          | .fvar id .. =>
+            let some range := info.range? | return
+            if (← warnings.get).contains range then return
+            let .original .. := info.stx.getHeadInfo | return
+            if ti.isBinder then
+              -- This is a local variable declaration.
+              let some ldecl := ti.lctx.find? id | return
+              -- Skip declarations which are outside the command syntax range, like `variable`s
+              -- (it would be confusing to lint these), or those which are macro-generated
+              if !cmdStxRange.contains range.start || ldecl.userName.hasMacroScopes then return
+              let opts := ci.options
+              -- we have to check for the option again here because it can be set locally
+              if !linter.constructorNameAsVariable.get opts then return
+              if let n@(.str .anonymous s) := info.stx.getId then
+                -- Check whether the type is an inductive type, and get its constructors
+                let ty ←
+                  if let some t := ti.expectedType? then pure t
+                  else ti.runMetaM ci (Meta.inferType ti.expr)
+                let ty ← ti.runMetaM ci (instantiateMVars ty >>= Meta.whnf)
+                if let .const tn _ := ty.getAppFn' then
+                  if let some (.inductInfo i) := (← getEnv).find? tn then
+                    for c in i.ctors do
+                      -- Only warn when the constructor has 0 fields. Pattern variables can't be
+                      -- confused with constructors that want arguments.
+                      if let some (.ctorInfo ctorInfo) := (← getEnv).find? c then
+                        if ctorInfo.numFields > 0 then continue
+                      if let .str _ cn := c then
+                        if cn == s then
+                          warnings.modify (·.insert range (info.stx, n, c))
+            else pure ()
+          | _ => pure ()
+        | _ => pure ())
+
+    -- Sort the outputs by position
+    for (_range, declStx, userName, ctorName) in (← warnings.get).toArray.qsort (·.1.start < ·.1.start) do
+      logLint linter.constructorNameAsVariable declStx <|
+        m!"Local variable '{userName}' resembles constructor '{ctorName}' - " ++
+        m!"write '.{userName}' (with a dot) or '{ctorName}' to use the constructor."
+
+builtin_initialize addLinter constructorNameAsVariable
--- a/src/Lean/Meta/Tactic/AC/Main.lean
+++ b/src/Lean/Meta/Tactic/AC/Main.lean
@@ -101,11 +101,10 @@ def buildNormProof (preContext : PreContext) (l r : Expr) : MetaM (Lean.Expr × 
 where
   mkContext (α : Expr) (u : Level) (vars : Array Expr) : MetaM (Array Bool × Expr) := do
     let arbitrary := vars[0]!
-    let zero := mkLevelZeroEx ()
-    let plift := mkApp (mkConst ``PLift [zero])
-    let pliftUp := mkApp2 (mkConst ``PLift.up [zero])
-    let noneE tp   := mkApp  (mkConst ``Option.none [zero]) (plift tp)
-    let someE tp v := mkApp2 (mkConst ``Option.some [zero]) (plift tp) (pliftUp tp v)
+    let plift := mkApp (mkConst ``PLift [.zero])
+    let pliftUp := mkApp2 (mkConst ``PLift.up [.zero])
+    let noneE tp   := mkApp  (mkConst ``Option.none [.zero]) (plift tp)
+    let someE tp v := mkApp2 (mkConst ``Option.some [.zero]) (plift tp) (pliftUp tp v)
     let vars ← vars.mapM fun x => do
       let isNeutral :=
         let isNeutralClass := mkApp3 (mkConst ``LawfulIdentity [u]) α preContext.op x
--- a/releases_drafts/varCtorNameLint.md
+++ b/releases_drafts/varCtorNameLint.md
@@ -1,0 +1,45 @@
+A new linter flags situations where a local variable's name is one of
+the argumentless constructors of its type. This can arise when a user either
+doesn't open a namespace or doesn't add a dot or leading qualifier, as
+in the following:
+
+````
+inductive Tree (α : Type) where
+  | leaf
+  | branch (left : Tree α) (val : α) (right : Tree α)
+
+def depth : Tree α → Nat
+  | leaf => 0
+````
+
+With this linter, the `leaf` pattern is highlighted as a local
+variable whose name overlaps with the constructor `Tree.leaf`.
+
+The linter can be disabled with `set_option linter.constructorNameAsVariable false`.
+
+Additionally, the error message that occurs when a name in a pattern that takes arguments isn't valid now suggests similar names that would be valid. This means that the following definition:
+
+```
+def length (list : List α) : Nat :=
+  match list with
+  | nil => 0
+  | cons x xs => length xs + 1
+```
+
+now results in the following warning:
+
+```
+warning: Local variable 'nil' resembles constructor 'List.nil' - write '.nil' (with a dot) or 'List.nil' to use the constructor.
+note: this linter can be disabled with `set_option linter.constructorNameAsVariable false`
+```
+
+and error:
+
+```
+invalid pattern, constructor or constant marked with '[match_pattern]' expected
+
+Suggestion: 'List.cons' is similar
+```
+
+
+#4301
--- a/src/Lean/Data/SMap.lean
+++ b/src/Lean/Data/SMap.lean
@@ -74,6 +74,12 @@ def forM [Monad m] (s : SMap α β) (f : α → β → m PUnit) : m PUnit := do
   s.map₁.forM f
   s.map₂.forM f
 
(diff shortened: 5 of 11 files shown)


david-christiansen:
CC @leodemoura - this is what we talked about yesterday

david-christiansen:
This is a draft until I fix the linter warnings during bootstrapping - but I think the linter itself is ready for review

src/Lean/Linter/ConstructorAsVariable.lean
```diff
@@ -0,0 +1,79 @@
...
+                          warnings.modify (·.insert range (info.stx, n, c))
+            else pure ()
+          | _ => pure ()
+        | _ => pure ())
+
+    -- Sort the outputs by position
+    for (_range, declStx, userName, ctorName) in (← warnings.get).toArray.qsort (·.1.start < ·.1.start) do
+      logLint linter.constructorNameAsVariable declStx <|
+        m!"Local variable '{userName}' resembles constructor '{ctorName}' - " ++
+        m!"missing namespace qualifier or 'open' statement?"
```
nomeata, line 77: How about 
```suggestion
        m!"did you want to write .{userName} (with a dot)?"
```
or so? This gives the (likely confused and frustrated) user some immediate and likely to work fix without requiring them to go off and read about possibly new concepts (`open`), and nudges then towards the recommended (at least by y Leo style). Could even become a quick fix later…

src/Lean/Data/Lsp/LanguageFeatures.lean
```diff
@@ -333,8 +333,8 @@ def SemanticTokenType.names : Array String :=
     "event", "method", "macro", "modifier", "comment", "string", "number",
     "regexp", "operator", "decorator", "leanSorryLike"]
 
-def SemanticTokenType.toNat (type : SemanticTokenType) : Nat :=
```
mhuisi, line 336: Should this new linter be limited to situations where denoting a constructor makes sense?

src/Lean/Linter/ConstructorAsVariable.lean
david-christiansen: I think this is a clear improvement, thanks! And the bit about the namespace qualifier seems redundant with the explicit namespace written in the error message.

src/Lean/Data/Lsp/LanguageFeatures.lean
david-christiansen: I don't think so, but it's a good question to ask.

There's two reasons:

 1. For new Lean users, it may not be clear where a constructor can or can't be denoted by a name. Having the linter fire may help make it clear that this is a variable.
 2. For less-new Lean users, avoiding this lint can make it more difficult to introduce bugs during refactoring. While I've been going through trying to make the compiler avoid this lint, there's been a few cases where renaming the argument to the function but skipping its use sites results in a program that still type checks, because the function is defined in the datatype's namespace and the constructor in question is nullary. I'm being careful while I do the renaming, and using "rename symbol", but it seems useful to have code in general that's robust in the face of incomplete renamings.

What do you think?

src/Lean/Linter/ConstructorAsVariable.lean
david-christiansen: I want to tweak the wording a bit - many of the target users didn't **want** to write `.name`, because this is their first encounter with the syntax, and I don't want the text to seem presumptuous.

src/Lean/Data/Lsp/LanguageFeatures.lean
mhuisi: There are plenty of examples in the diff of this PR where people felt like the constructor name was the most natural name to use (both because it is known to be the constructor and by accident) :-)

david-christiansen: > Maybe using `fvar` to bind an `Expr` known to be a `.fvar n`

This is the single most common thing that would need changing in this diff, followed by `mvar` for the same reason, and then `alt` for `Alt`.

I could add an additional setting for constructors that we don't want to warn about, or I could do an update-stage0 followed by disabling this linter in those files. Those would work fine as well - I also don't have a hugely strong opinion on this, so long as new users are getting a good experience.

david-christiansen: Though whether it's firing in places where constructor patterns are allowed is a bit of a red herring for this kind of situation, because we have a fair number of instances of `let fvar := ...`

david-christiansen: OK, I'm going to aim for less disturbance right now, and then we can have the discussion later. I'll add an option to disable it for certain constructors, set the default to false, update-stage0, then set the default to true with all of the cases that we frequently use set to exceptions. Then we can take it case by case, and not all in one ginormous diff.

src/Lean/Linter/ConstructorAsVariable.lean
david-christiansen: New wording:
> warning: Local variable 'y' resembles constructor 'A.y' - write '.y' (with a dot) or 'A.y' to use the constructor.

david-christiansen:
Simulating the next PR post-stage0-update is showing that there's something I don't understand about attributes and bootstrapping, so I'm marking draft until the followup PR also works locally.

david-christiansen:
OK, I locally have a followup change that enables and cleanly suppresses the warning after an update-stage0. I think this is ready to go.

tests/lean/byCasesMetaM.lean
```diff
@@ -6,8 +6,8 @@ open Lean in
 open Lean.Meta in
 def test : MetaM Unit := do
   let type := (← getConstInfo ``ex).type
-  let mvar ← mkFreshExprMVar type
-  let (#[p, q, h], mvarId) ← mvar.mvarId!.introNP 3 | throwError "unexpected"
+  let mv ← mkFreshExprMVar type
```
tydeu, line 9: This suggests to me that the linter should probably not fire on pattern matches with only one arm / alternative.

david-christiansen: I would tend to disagree here - the example code in the PR text can lead a new user without much experience to think that their function is complete, because there's no red squigglies to give the hint that a case was forgotten.

In any case, the updated version of the PR code now has a mechanism for excepting certain constructors from the lint, and it's going to be applied to `mvar`, among others, so we won't need to change this style if we don't want to.

david-christiansen: (I should revert these test changes in anticipation of that, I suppose)

tydeu: @david-christiansen 

> I would tend to disagree here - the example code in the PR text can lead a new user without much experience to think that their function is complete, because there's no red squigglies to give the hint that a case was forgotten.

In the `Tree` PR example, the unused variables linter fires, so there is still a hint that something is amiss. I would also argue that even a novice user would be unlikely to leave the `depth` function as is, because it still looks to produce a constant depth of `0`. On the flip side, this certainly may be less clear in more complex examples. However, I think the unused variables linter is likely still a good enough hint there.

> In any case, the updated version of the PR code now has a mechanism for excepting certain constructors from the lint, and it's going to be applied to `mvar`, among others, so we won't need to change this style if we don't want to.

Marking constructors seems like an incredibly invasive way to handle this problem. The `mvar`/`fvar` pattern of naming a variable in `let` after a constructor when the object in question is known to be that alternative seems like it will be common in user code . Needing to mark all of these constructors `allow_name_as_variable` seems burdensome (and possibly quite difficult if the type is defined in another package).

david-christiansen: Thanks for the feedback!

The unused variables linter isn't quite enough for situations where users are struggling. I've also seen this pattern, though not at the recent event (the examples/labs didn't include enumeration-looking things):
```
inductive Day where
  | monday
  | tuesday
  | wednesday

def num : Day → Nat
  | monday => 0
  | tuesday => 1
  | wednesday => 2
```
The last two patterns are marked "redundant alternative", and because the definition contains an error, the unused variables linter doesn't fire at all. This linter suggests a `.` before all three patterns, even in the presence of the error message, which can plausibly unstick those who are still struggling with the namespace system.

> Needing to mark all of these constructors allow_name_as_variable seems burdensome (and possibly quite difficult if the type is defined in another package).

Depending on the situation, it's also reasonable to `set_option linter.constructorNameAsVariable false`. All style linters have a cost - this one seems to benefit new users at the cost of experts, which seems like an OK prioritization to me.

In the Lean codebase, only a very few datatypes regularly use constructor names as variables, and newly-written code seems unlikely to have a hard time avoiding these lints if they don't want to set the option.


I think your concern could also be alleviated by having this linter only fire when the constructor has arity 0. In practice, it doesn't fire for constructors like `Tree.branch`, which instead say that the pattern is invalid: `invalid pattern, constructor or constant marked with '[match_pattern]' expected`. What if this linter was for arity-0 constructors only, and then that error message additionally checked for a similar constructor name and then made the same suggestion?

tydeu: @david-christiansen I agree that this linter should fire in that example. Note that my suggestion was only to not fire in matches with a single branch / arm and your example has 3. 😉 

>  What if this linter was for arity-0 constructors only, and then that error message additionally checked for a similar constructor name and then made the same suggestion?

I think this heuristic is a fine alternative solution that could cover most instances of the problem I raised as well, 😄 

---

Below are some additional comments  I had, but they are not really that important to progressing this issue, so feel free to skip reading them.

> Depending on the situation, it's also reasonable to `set_option linter.constructorNameAsVariable false`. All style linters have a cost - this one seems to benefit new users at the cost of experts

The ability to get around this with an option suggests to me that this is likely to be helpful for more advanced users to fix mistakes and/or disable it when they know there is none. Novice users are likely to get confused if/when this triggers in places were they probably don't even realize a pattern match is occurring. When I first saw this error case, I was confused as to why this linter was firing on a basic `let` declaration. Only after looking at it for while did it strike me that all `let` declarations are technically pattern matches. I imagine a novice user would be entirely bamboozled.

> In the Lean codebase, only a very few datatypes regularly use constructor names as variables, and newly-written code seems unlikely to have a hard time avoiding these lints if they don't want to set the option.

I really think you are underestimating how easy it is to stumble into this problem with a `let`. It is quite natural to write a `let` declaration with a constructor name for a type with smart constructors (e.g., `Expr`, `Syntax`, `Json`) For example, `let obj := Json.mkObj`  or `let atom := Syntax.mkAtom` would trigger the linter as-is. However, it could be argued that such names are not very good and thus are a code smell all on their own. However, I think that depends on the context. If the function is dealing with a single generic `obj`, `atom`, this can be quite natural to write and easy to understand.

Separately, one of the reasons I think these examples don't appear more in core outside of specialist types like `Expr` and `Syntax` is that there are surprisingly few complex inductive types in core. Structure-like inductives, abstract combinators (e.g., `Option`, `Except`, `Sum`) and arity 0 enumerations (e.g., `Ordering`, kinds, modes, flags, etc.) make up the lion's share of inductive types in core.

david-christiansen: This is useful, thanks for helping me refine it. I'll update the linter to fire on unary constructors only, and the error message above as well. I think that'll serve new users at least as well, with less of an imposition on experts.

david-christiansen: OK, looking at the error message in question, I don't think it's feasible to do type-based constructor name suggestions. I'll see if just listing out all the overloads of the name that would be eligible as patterns gives a good enough experience.

david-christiansen: I think it works pretty well, so long as not *all* the eligible overloads are listed!

Writing:
```
def length (list : List α) : Nat :=
  match list with
  | nil => 0
  | cons x xs => length xs + 1
```
now results in the following warning on `nil`:
```
warning: Local variable 'nil' resembles constructor 'List.nil' - write '.nil' (with a dot) or 'List.nil' to use the constructor.
note: this linter can be disabled with `set_option linter.constructorNameAsVariable false`
```

and error on `cons`:
```
invalid pattern, constructor or constant marked with '[match_pattern]' expected

Suggestion: 'List.cons' is similar
```

What do you think?

src/Lean/Data/Lsp/LanguageFeatures.lean
david-christiansen: This has now been changed so that the linter only fires for nullary constructors that overlap bound variable names of the same type - only one constructor was hit by this in all of Lean.

At the same time, I also updated the error message for non-constructor things being applied in patterns so that it suggests valid names if any exist. This should also catch similar misunderstandings in those cases.

Now it doesn't need to be selectively disabled, which is a good sign.

tests/lean/byCasesMetaM.lean
tydeu: Looks good to me!

src/Lean/Data/Lsp/LanguageFeatures.lean
tydeu: @david-christiansen 

>  only one constructor was hit by this in all of Lean.

I am curious what the one example was?

david-christiansen: Turns out I was wrong - there was this one, and then also https://github.com/leanprover/lean4/pull/4301/files#diff-ec4d9d868e0d4a5702d821884ef44382eb312b8de8fab76db01879ee97fead78L104-R108 .

So two places - not too terrible, I suppose, and I think both are easier to read with this change.

src/Lean/Meta/Tactic/AC/Main.lean
```diff
@@ -101,11 +101,11 @@ def buildNormProof (preContext : PreContext) (l r : Expr) : MetaM (Lean.Expr ×
...
-    let zero := mkLevelZeroEx ()
-    let plift := mkApp (mkConst ``PLift [zero])
-    let pliftUp := mkApp2 (mkConst ``PLift.up [zero])
-    let noneE tp   := mkApp  (mkConst ``Option.none [zero]) (plift tp)
-    let someE tp v := mkApp2 (mkConst ``Option.some [zero]) (plift tp) (pliftUp tp v)
+    let lvlZero := mkLevelZeroEx ()
+    let plift := mkApp (mkConst ``PLift [lvlZero])
+    let pliftUp := mkApp2 (mkConst ``PLift.up [lvlZero])
+    let noneE tp   := mkApp  (mkConst ``Option.none [lvlZero]) (plift tp)
+    let someE tp v := mkApp2 (mkConst ``Option.some [lvlZero]) (plift tp) (pliftUp tp v)
```
tydeu, line 108:
```suggestion
    let plift := mkApp (mkConst ``PLift [.zero])
    let pliftUp := mkApp2 (mkConst ``PLift.up [.zero])
    let noneE tp   := mkApp  (mkConst ``Option.none [.zero]) (plift tp)
    let someE tp v := mkApp2 (mkConst ``Option.some [.zero]) (plift tp) (pliftUp tp v)
```

I think this can be simplified further to just use the constructor directly. `mkLevelZeroEx` is iexport of the constructor for use in C code. There is no reason for it have been used it here.  Thus, this serves to further demonstrate the utility of this lint. 😄

david-christiansen: I saw that while reading the code - figured Chesterton's Fence meant I shouldn't tweak it. Let's see if the tests still pass with this suggestion committed - it would be surprising if this broke something, after all!

david-christiansen:
Rebase and squash to try to fix CI failure

PR #4305, opened 2024-05-29
doc: explain the borrow syntax

eric-wieser:
Obviously a link to the web docs isn't ideal, but having hovers available on the symbol is much better than nothing.

### DIFF
--- a/src/Lean/Parser/Term.lean
+++ b/src/Lean/Parser/Term.lean
@@ -451,7 +451,7 @@ def withAnonymousAntiquot := leading_parser
 /-- 
 Indicates that an argument to a function marked `@[extern]` is borrowed.
 
-Being borrowed only affects the way C code interacts with the function. From the perspective of Lean, this annotation has no effect. It similarly has no effect on functions not marked `@[extern]`.
+Being borrowed only affects the ABI and runtime behavior of the function when compiled or interpreted. From the perspective of Lean's type system, this annotation has no effect. It similarly has no effect on functions not marked `@[extern]`.
 
 When a function argument is borrowed, the function does not consume the value. This means that the function will not decrement the value's reference count or deallocate it, and the caller is responsible for doing so.
 


src/Lean/Parser/Term.lean
```diff
@@ -448,6 +448,9 @@ def withAnonymousAntiquot := leading_parser
 @[builtin_term_parser] def «trailing_parser» := leading_parser:leadPrec
   "trailing_parser" >> optExprPrecedence >> optExprPrecedence >> ppSpace >> termParser
 
+/-- Marks a function argument or return value as borrowed.
+
+See https://lean-lang.org/lean4/doc/dev/ffi.html#borrowing -/
```
eric-wieser, line 453: https://lean-lang.org/lean4/doc/dev/ffi.html#borrowing doesn't actually mention borrowed return values, but I assume they do exist given

https://github.com/leanprover/lean4/blob/8bbb015a973904d00d89cb844920d49c9cacbef5/src/runtime/object.h#L24

src/Lean/Parser/Term.lean
```diff
@@ -448,6 +448,9 @@ def withAnonymousAntiquot := leading_parser
 @[builtin_term_parser] def «trailing_parser» := leading_parser:leadPrec
   "trailing_parser" >> optExprPrecedence >> optExprPrecedence >> ppSpace >> termParser
 
+/-- Marks a function argument or return value as borrowed.
```
TwoFX, line 451: This should mention that the borrowing annotation only has an effect for functions with `@[extern]`. It is ignored on all other functions.

eric-wieser: Done. Also fixed a docstring that made an incorrect claim about `export`.

src/Lean/Parser/Term.lean
```diff
@@ -448,6 +448,10 @@ def withAnonymousAntiquot := leading_parser
 @[builtin_term_parser] def «trailing_parser» := leading_parser:leadPrec
   "trailing_parser" >> optExprPrecedence >> optExprPrecedence >> ppSpace >> termParser
 
+/-- Marks a function argument of an `@[extern]`ed function as borrowed.
```
david-christiansen, line 451:
```suggestion
/-- Marks a function argument of an function as borrowed. This only has an effect if the function is made available through the FFI using `@[extern]`.
```

I think the use of the attribute is a bit confusing for two reasons: the syntactic noise takes a moment to parse, and "externed" is also not an ordinary word.

src/Lean/Parser/Term.lean
```diff
@@ -448,6 +448,10 @@ def withAnonymousAntiquot := leading_parser
...
+
+See https://lean-lang.org/lean4/doc/dev/ffi.html#borrowing;
```
david-christiansen, line 453: The linked section isn't all that long, perhaps it should be summarized here instead?

eric-wieser: My thinking was mainly that the linked section had a lot of remarks about the current state of development, and so duplicating the information increased the chance of things going out of sync. I can duplicate it here verbatim if you'd prefer; though if you would prefer to reword it, perhaps it's best for you to push directly to this PR.

digama0: I think it should be explained in more detail inline than this. The comment as is adds no information besides the word "borrowed". I think the important information for users is:

* This is an optimization hint and does not change the semantics of the code
* This is only relevant for @[extern] functions and does nothing unless it appears in a parameter type (AFAIK there is no borrowing annotation for return types, even though `b_obj_res` exists)
* This changes the ABI for the function to take an argument as "borrowed", meaning that it does not consume the value (or lower the RC value by 1 if it is shared)

src/Lean/Parser/Term.lean
```diff
@@ -448,6 +448,10 @@ def withAnonymousAntiquot := leading_parser
...
+currently this annotation is ignored on `@[export]`ed functions and return values. -/
```
david-christiansen, line 454: Is this revision OK?

```suggestion
/-- 
Indicates that an argument to a function marked `@[extern]` is borrowed.

Being borrowed only affects the way C code interacts with the function. From the perspective of Lean, this annotation has no effect. It similarly has no effect on functions not marked `@[extern]`.

When a function argument is borrowed, the function does not consume the value. This means that the function will not decrement the value's reference count or deallocate it, and the caller is responsible for doing so.

Please see https://lean-lang.org/lean4/doc/dev/ffi.html#borrowing for a complete description.
-/
```

src/Lean/Parser/Term.lean
david-christiansen: I think this makes sense - I used that as a starting point for a suggested revision below. What do you think?

src/Lean/Parser/Term.lean
tydeu: @david-christiansen I would suggest a slight change:

```suggestion
/-- 
Indicates that an argument to a function is borrowed.

Being borrowed only affects how the compiler interacts with the function. From the perspective of Lean, this annotation has no effect. For the current compiler, it similarly has no effect on functions not marked `@[extern]`.

When a function argument is borrowed, the function does not consume the value. This means that the function will not decrement the value's reference count or deallocate it, and the caller is responsible for doing so.

Please see https://lean-lang.org/lean4/doc/dev/ffi.html#borrowing for a complete description.
-/
```

There are two things to note here. First, the borrow annotation does not just effect C code, it also effects IR. It is even possible to provide IR directly to an `@[extern]` definition without external C code. For example, I do this in [Lyre](https://github.com/tydeu/lyre). Second, user code can make use of the borrow annotation for more than just `@[extern]`, so I would suggest describing its general semantic meaning first, and then the way it is used by core code second.

digama0: > Second, user code can make use of the borrow annotation for more than just @[extern].

No, AFAIK the annotation does not work on user code that is not @[extern] (for incidental reasons, even though the semantics still make sense on user code). Do you have evidence to the contrary?

tydeu: @digama0 Sorry, I was not clear. By "user code", I meant metaprogramming user code. The borrow annotation is recorded in the IR, and user code interacting with the IR (or with term syntax) can make use of that information. As you say, the semantics can still make sense in user code and some user-defined metaprogramming can take advantage of that. Thus, it seems worth to describe the semantic meaning first then describe the limitations of the core implementation.

digama0: To be fair, that's just saying that `@&` is an annotation and therefore leaves a trace in the term. The exact same thing is true about all other annotations, and user code can do whatever it wants with annotations. Unless there are specific well known standard metaprograms that also do something with this annotation, I think it is best to stick to the description of what its actual effects are, since the part about it being an annotation which user code (potentially, but not in practice) could do something with is implicit and obviously the case.

eric-wieser:
FWIW, this has come up [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/What's.20.40.26.20.3F/near/448397959) again.

src/Lean/Parser/Term.lean
eric-wieser: > The comment as is adds no information besides the word "borrowed".

This is the difference between hovering over the symbol in vscode showing nothing, and hovering over it showing a word that can be searched on Zulip.

@david-christiansen, would you be happy to adopt this PR? I'm unsure how to address your concerns simultaneously with Mac and Mario's, and would at this point be happy with any non-empty docstring here.

src/Lean/Parser/Term.lean
eric-wieser: I've taken David's suggestion here. Mac, if you want your version, can you discuss this with David and push as you see fit?

src/Lean/Parser/Term.lean
```diff
@@ -448,6 +448,15 @@ def withAnonymousAntiquot := leading_parser
 @[builtin_term_parser] def «trailing_parser» := leading_parser:leadPrec
   "trailing_parser" >> optExprPrecedence >> optExprPrecedence >> ppSpace >> termParser
 
+/-- 
+Indicates that an argument to a function marked `@[extern]` is borrowed.
+
+Being borrowed only affects the way C code interacts with the function. From the perspective of Lean, this annotation has no effect. It similarly has no effect on functions not marked `@[extern]`.
```
Kha, line 454:
```suggestion
Being borrowed only affects the ABI and runtime behavior of the function when compiled or interpreted. From the perspective of Lean's type system, this annotation has no effect. It similarly has no effect on functions not marked `@[extern]`.
```

Kha (approved):

PR #4312, opened 2024-05-31
feat: change List.length_cons to use + 1 normal form instead of succ

markusschmaus:
`List.length_cons` still used `Nat.succ` which is no longer a simp normal form. This changes the theorem to use `+ 1` instead.


markusschmaus:
This is a follow up PR to #4242.

PR #4314, opened 2024-05-31
chore: remove @[simp] from bind_eq_some

kim-em:


semorrison:
I have the Mathlib fixes, but on an apparently-inaccessible-via-ssh machine right now. I'll finish this once I'm home next week.

PR #4317, opened 2024-05-31
doc: updated build instructions for mac os X

jtristan:
Edit to the platform specific setup for mac os X. The installation of llvm does not accept the options --with-clang and --with-asan anymore.


semorrison:
Modern homebrew does not allow these options. I think homebrew's llvm has clang and asan by default anyway, but haven't managed to confirm as I'm travelling with my antique laptop.

PR #4319, opened 2024-05-31
perf: optimize `isClassQuickConst?`

JovanGerb:
`isClassQuick?` is used as a quicker version of `isClassExpensive?`, returning `.undef` when it encounters an unfoldable constant. It checks this in the global transparency setting. But `isClassExpensive?` runs in the `reducible` transparency setting. I fixed it so that `isClassQuick?` does this check in the `reducible` setting as well.

In this refactor I've also deleted the `private` functions `getConstTemp?` and `getDefInfoTemp`, which aren't used anywhere else. 

And I moved `isClassImp?` outside of the mutual block it was in.

This leads to a 0.1% reduction in build instructions for Mathlib

### DIFF
--- a/src/Lean/Meta/Basic.lean
+++ b/src/Lean/Meta/Basic.lean
@@ -991,11 +991,12 @@ private def isClassQuickConst? (constName : Name) : MetaM (LOption Name) := do
     return .some constName
   else
     match (← getEnv).find? constName with
-    | some info =>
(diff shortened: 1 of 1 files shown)


JovanGerb:
awaiting-review

Kha:
Closing as per reasoning in #4379

PR #4321, opened 2024-06-02
chore: deprecate shiftLeft_shiftLeft, shiftRight_shiftRight

bollu:
As discussed previously (https://github.com/leanprover/lean4/pull/4179#discussion_r1615183093), we deprecate the two functions in favour of `shift{Left, Right}_add`.


bollu:
awaiting-review

PR #4325, opened 2024-06-03
feat: ship cadical

Kha:


Kha:
Let's first see if it builds at all

PR #4336, opened 2024-06-04
doc: correct reference to `repeat'` tactic

frangio:
Batteries is no longer needed for `repeat'` since it was upstreamed in #3305.


frangio:
Closing in favor of https://github.com/leanprover/lean4/pull/4337 and https://github.com/leanprover/lean4/pull/4338.

PR #4349, opened 2024-06-04
fix: `split` (for `if`-expressions) should work on non-propositional goals

### LINKED ISSUE #4313: `split` fails for non-propositional goals
### Prerequisites

### Description

The current implementation of `split` splits on `p ∨ ¬p`, where `p` is the condition of an `ite`. This fails if the goal is non-propositional since `Or.elim` can only eliminate into `Prop`. MWE:

```lean
example {A B : Type} {p : Prop} [Decidable p] (h : if p then A else B) :
    Sum A B := by
  split at h
  /-
  tactic 'cases' failed, nested error:
  tactic 'induction' failed, recursor 'Or.casesOn' can only eliminate into Prop
  AB: Type
  p: Prop
  inst✝: Decidable p
  h: if p then A else B
  hByCases✝: p ∨ ¬p
  ⊢ A ⊕ B
  -/
```

### Versions

`"4.9.0-nightly-2024-05-31"`

### Additional Information

I believe the implementation strategy suggested in #2414 would also fix this issue.

### Impact

leodemoura:
Remark: when splitting an `if-then-else` term, the subgoals now have tags `isTrue` and `isFalse` instead of `inl` and `inr`.
closes #4313

### DIFF
--- a/src/Init/Data/AC.lean
+++ b/src/Init/Data/AC.lean
@@ -146,8 +146,8 @@ theorem Context.evalList_mergeIdem (ctx : Context α) (h : ContextInformation.is
     | nil =>
       simp [mergeIdem, mergeIdem.loop]
       split
-      case inl h₂ => simp [evalList, h₂, h.1, EvalInformation.evalOp]
-      rfl
+      next h₂ => simp [evalList, h₂, h.1, EvalInformation.evalOp]
+      next => rfl
     | cons z zs =>
       by_cases h₂ : x = y
       case pos =>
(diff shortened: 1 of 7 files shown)


src/Lean/Meta/Tactic/SplitIf.lean
```diff
@@ -68,23 +68,24 @@ private def discharge? (numIndices : Nat) (useDecide : Bool) : Simp.Discharge :=
 def mkDischarge? (useDecide := false) : MetaM Simp.Discharge :=
   return discharge? (← getLCtx).numIndices useDecide
 
-/-- Return the condition of an `if` expression to case split. -/
-partial def findIfToSplit? (e : Expr) : Option Expr :=
+/-- Return the conditiong and decidable instance of an `if` expression to case split. -/
```
digama0, line 71:
```suggestion
/-- Return the condition and decidable instance of an `if` expression to case split. -/
```

leodemoura:
@semorrison The `split` on `if-then-else`-terms now creates subgoals with tags `isTrue` and `isFalse`. Reason: it is internally performing `cases` on a `Decidable p` instead of an `Or p (Not p)`. In principle, we could rename the tags to `inl` and `inr`, but I think the new tags are more informative than the old ones: `inl` and `inr`. Some proofs will break, but the impact on Mathlib should be small since they have their own `split_if` tactic.

PR #4351, opened 2024-06-05
perf: lake: fix `LogIO` inling/lifting

tydeu:
The current manner of lifting `LogIO` into `CliM` produces excessive specializations (due to a nested inlined `forM`). There was also a bug where `IO` was lifted into `CliM` via `LogIO` rather than directly through `MainM`.


tydeu:
!bench

PR #4330, opened 2024-06-03
Simple String lemma

jtristan:
# Read this section before submitting

* Ensure your PR follows the [External Contribution Guidelines](https://github.com/leanprover/lean4/blob/master/CONTRIBUTING.md).
* Please make sure the PR has excellent documentation and tests. If we label it `missing documentation` or `missing tests` then it needs fixing!
* Include the link to your `RFC` or `bug` issue in the description.
* If the issue does not already have approval from a developer, submit the PR as draft.
* The PR title/description will become the commit message. Keep it up-to-date as the PR evolves.
* If you rebase your PR onto `nightly-with-mathlib` then CI will test Mathlib against your PR.
* You can manage the `awaiting-review`, `awaiting-author`, and `WIP` labels yourself, by writing a comment containing one of these labels on its own line.
* Remove this section, up to and including the `---` before submitting.

---

Closes #0000 (`RFC` or `bug` issue number fixed by this PR, if any)

PR #4354, opened 2024-06-05
chore: upstream basic String lemmas

kim-em:


src/Init/Data/String/Basic.lean
```diff
@@ -1018,5 +1020,145 @@ def decapitalize (s : String) :=
...
+
+-- `toSubstring'` is just a synonym for `toSubstring` without the `@[inline]` attribute
+-- so for proving can be unfolded.
+attribute [simp] toSubstring'
+
+theorem singleton_eq (c : Char) : singleton c = ⟨[c]⟩ := rfl
+
+@[simp] theorem data_singleton (c : Char) : (singleton c).data = [c] := rfl
+
+@[simp] theorem append_nil (s : String) : s ++ "" = s := ext (List.append_nil _)
```
TwoFX, line 1145: Currently the lemmas in Batteries and mathlib are inconsistent about whether the empty string should be called `nil` or `empty` in lemma names (see [Loogle](https://loogle.lean-lang.org/?q=String%2C+%22empty%22) and [Loogle](https://loogle.lean-lang.org/?q=String%2C+"nil")). This seems to be the first case where Init has to take a stand on this (except for `String.isEmpty`, which doesn't count because `List` also has `isEmpty` even though the empty list is called `nil`).

My guess would be that most users would try `empty` before `nil`, if they even think of `nil` at all. #4330 also uses `empty`.

semorrison: Good catch. I'm strongly in favour of `empty`.

digama0 (approved):

PR #4357, opened 2024-06-05
chore: unify `String.csize : Nat` and `Char.utf8Size : UInt32` as `Char.size : Nat`

kim-em:
It seems:
* there was no actual need for the UInt32 valued version
* downstream we were getting duplicative lemmas about both
* so lets reduce the API surface area!

If anyone would prefer the remaining function is still called `Char.utf8Size` I will happily change it. (`size` is hopefully still unambiguous, and it's helpful to rename here so we can give a deprecation warning that explains the type signature change.)

### DIFF
--- a/src/Init/Data/String/Basic.lean
+++ b/src/Init/Data/String/Basic.lean
@@ -1141,14 +1141,14 @@ theorem add_eq (p₁ p₂ : Pos) : p₁ + p₂ = ⟨p₁.byteIdx + p₂.byteIdx�
 
 theorem sub_eq (p₁ p₂ : Pos) : p₁ - p₂ = ⟨p₁.byteIdx - p₂.byteIdx⟩ := rfl
 
-@[simp] theorem addChar_byteIdx (p : Pos) (c : Char) : (p + c).byteIdx = p.byteIdx + csize c := rfl
+@[simp] theorem addChar_byteIdx (p : Pos) (c : Char) : (p + c).byteIdx = p.byteIdx + c.utf8Size := rfl
 
-theorem addChar_eq (p : Pos) (c : Char) : p + c = ⟨p.byteIdx + csize c⟩ := rfl
+theorem addChar_eq (p : Pos) (c : Char) : p + c = ⟨p.byteIdx + c.utf8Size⟩ := rfl
 
-theorem zero_addChar_byteIdx (c : Char) : ((0 : Pos) + c).byteIdx = csize c := by
+theorem zero_addChar_byteIdx (c : Char) : ((0 : Pos) + c).byteIdx = c.utf8Size := by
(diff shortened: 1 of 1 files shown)


src/Init/Data/Char/Lemmas.lean
```diff
@@ -22,13 +22,18 @@ protected theorem le_total (a b : Char) : a ≤ b ∨ b ≤ a := UInt32.le_total
 protected theorem lt_asymm {a b : Char} (h : a < b) : ¬ b < a := UInt32.lt_asymm h
 protected theorem ne_of_lt {a b : Char} (h : a < b) : a ≠ b := Char.ne_of_val_ne (UInt32.ne_of_lt h)
 
-theorem utf8Size_pos (c : Char) : 0 < c.utf8Size := by
-  simp only [utf8Size]
-  repeat (split; decide)
-  decide
+/-- `Char.size` now returns a `Nat`, rather than a `UInt32`. -/
+@[deprecated Char.size (since := "2024-06-04")] abbrev utf8Size := Char.size
```
tydeu, line 26: It seems like it would still be a good idea to return a `UInt32` from this for now to avoid breakages. The deprecation warning I think should be more than enough.

semorrison: Done.

src/Init/Data/Char/Lemmas.lean
```diff
@@ -22,13 +22,24 @@ protected theorem le_total (a b : Char) : a ≤ b ∨ b ≤ a := UInt32.le_total
 protected theorem lt_asymm {a b : Char} (h : a < b) : ¬ b < a := UInt32.lt_asymm h
 protected theorem ne_of_lt {a b : Char} (h : a < b) : a ≠ b := Char.ne_of_val_ne (UInt32.ne_of_lt h)
 
-theorem utf8Size_pos (c : Char) : 0 < c.utf8Size := by
-  simp only [utf8Size]
-  repeat (split; decide)
-  decide
+/-- `Char.size` now returns a `Nat`, rather than a `UInt32`. -/
```
digama0, line 25: the comment is now incorrect

digama0 (approved):

src/Init/Data/Char/Lemmas.lean
semorrison: I think it was technically correct, as the doc-string is talking about `Char.size`, but I've clarified, thanks.

Kha:
> If anyone would prefer the remaining function is still called `Char.utf8Size` I will happily change it

Yes please, "size" is too ambiguous for the subtleness that is Unicode.

semorrison:
> > If anyone would prefer the remaining function is still called `Char.utf8Size` I will happily change it
> 
> Yes please, "size" is too ambiguous for the subtleness that is Unicode.

Done!

PR #4366, opened 2024-06-05
chore: add error message for failing elabNumLit

### LINKED ISSUE #4365: Improve error messages for elabNumLit
### Prerequisites

### Description

The elaboration error message in the MWE below is a bit cryptic.

### Context

The error occured to a Lean user by accidentally providing `1` as a function argument in the wrong position, where something of type `A : Prop` was expected.

[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/cryptic.20error/near/442799857)

### Steps to Reproduce

```lean
import Lean

#check (1 : ∀ (n : Nat), True) -- error: invalid universe level, ?u.5 is not greater than 0
```

**Expected behavior:** The error message should state that it's not possible to parse a numeric literal of this type. See suggestion PR #4366

**Actual behavior:** The error message only talks about universe levels, which isn't immediately obvious to the untrained eye what it means.

### Versions

4.8.0-rc2

### Impact

joneugster:
Closes #4365

### DIFF
--- a/src/Lean/Elab/BuiltinTerm.lean
+++ b/src/Lean/Elab/BuiltinTerm.lean
@@ -191,7 +191,7 @@ private def mkFreshTypeMVarFor (expectedType? : Option Expr) : TermElabM Expr :=
     | none     => throwIllFormedSyntax
   let typeMVar ← mkFreshTypeMVarFor expectedType?
   let some u ← getDecLevel? typeMVar
-    | throwError m!"Tried to elab numeric literal{indentD m!"{stx}"}\nof type {indentD m!"{typeMVar} : {(← inferType typeMVar)}"}\nbut universe level of{indentD (← inferType typeMVar)}\nis not greater than 0."
+    | throwError m!"Tried to elab numeric literal{indentD m!"{stx}"}\nof type{indentD m!"{typeMVar} : {(← inferType typeMVar)}"}\nbut universe level of{indentD (← inferType typeMVar)}\nis not greater than 0."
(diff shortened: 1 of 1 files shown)


src/Lean/Elab/BuiltinTerm.lean
```diff
@@ -190,7 +190,8 @@ private def mkFreshTypeMVarFor (expectedType? : Option Expr) : TermElabM Expr :=
     | some val => pure val
     | none     => throwIllFormedSyntax
   let typeMVar ← mkFreshTypeMVarFor expectedType?
-  let u ← getDecLevel typeMVar
+  let some u ← getDecLevel? typeMVar
+    | throwError s!"Cannot elab numeral. Universe level of expected type must be greater than 0."
```
kmill, line 194: The error message of `getDecLevel?` includes the universe level, and I think we shouldn't lose that:
```suggestion
  let u' ← getLevel typeMVar
  let some u ← decLevel? u'
    | throwError s!"Universe level of expected type must be greater than 0, but it is {u'}."
```
Alternatively:
```suggestion
  let some u ← getDecLevel? typeMVar
    | throwError s!"Expected type is{indentD (← inferType typeMVar)}\nbut universe level of expected type must be greater than 0."
```

joneugster: Thanks for the suggestion @kmill ! I particularly like the second approach, which I extended a bit. What do you think:

```
Tried to elab numeric literal
  1
of type 
  Nat → True : Prop
but universe level of
  Prop
is not greater than 0.
```

I think the piece of information I was missing here previously is that the error is about the "numeric literal" `1`.

leodemoura:
Thanks for the PR, but it has been subsumed by #4368

PR #4367, opened 2024-06-05
chore: fix typo and incorrect name in doc

b-mehta:
Fixes typo "reflexivitiy" to "reflexivity", and changes `exact Eq.rfl` to `exact rfl`, since `Eq.rfl` does not exist. (Perhaps this should be `Eq.refl _` instead?)


leodemoura:
Thanks for the PR.
Pending: broken test.

Issue #2452, opened 2023-08-23
RFC: change variable inclusion mechanism


sgouezel:
The current Lean 4 mechanism for including variables is the following: When starting the proof of a theorem (or the body of a definition, it doesn't make a difference), all variables in the context are included and may be used in the proof. When the proof is over, only those variables that have indeed been used are kept for the statement of the theorem (i.e., all the other ones are discarded). This looks nice, but we have faced several issues in mathib 4 because of this:

* if in the context there is `variable {n : ℕ} (hn : 1 < n) (h'n : 2 < n)` and a theorem using `hn` but not `h'n`, then `induction n` will also pull `h'n` in the assumptions of the theorem. So will `subst`. The solution is to `clear h'n` before the induction even though `h'n` should not be in sight anywhere.
* When cleaning up a proof, say by using lemma B instead of lemma A in the proof, new typeclass assumptions may be pulled in by lemma B and weaken the initial statement. All this will go unnoticed (no linter can help here).
* In files which are heavy on variables or typeclass assumptions (say files on vector bundles), the goal view is cluttered by dozens of lines of assumptions which are not relevant for the lemma being currently proved, making it really hard to use.
* Refactors in a file may change the *statements* of theorems in another file without breaking anything: if the proof of a theorem involves a typeclass `classA α`, and in the context there are assumptions `classB α` and `classC α`, both of them implying `classA α`, then the above mechanism will only keep `classB α` or `classC α`, and the one which will be kept depends on the details of the path chosen by typeclass inference. Therefore, adjusting instance priorities somewhere else may change which one will appear in the assumptions of the theorem. 

No linter can guard against this lack of robustness. Therefore, most mathlib folks have been convinced that a less clever but more predictable behavior would be an improvement over the current situation. After several discussions on Zulip, the following scheme has been suggested:
* Include for a lemma or a def all the variables that are mentioned explicitly in the statement of the lemma (not its body, even if it is in term mode, contrary to Lean 3)
* Also include all the typeclass assumptions about these variables which are in the context (named or unnamed, contrary to Lean 3)
* No automatic removal of unused typeclass assumptions at the end of the proof: Better let a linter signal it, and let the user fix it by hand (to avoid non-robustness issues pointed out in the discussion above).
[...]

PR #4370, opened 2024-06-06
fix: `variable` must execute pending tactics and elaboration problems

### LINKED ISSUE #2226: `def` fails without error (`by` in `variable`)
In this (definitely contrived) example, the `foo` declaration fails, as evidenced by the fact that it is not in the environment, but no error is reported:
```lean
variable (A : Nat) (B : by skip)
def foo := A = B -- no error is reported

#print foo -- unknown constant 'foo'
```
Using `def foo : Prop` instead causes the declaration to succeed and report that it uses `sorry`.

### LINKED ISSUE #3214: #synth silently succeeds if given nonsense
### Prerequisites

### Description

The following code emits no diagnostics:
```lean
class Foo (α : Type)

variable {α : Type} (s : Missing α)

#synth Foo s
```

### Context

[Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.23synth.20is.20ignoring.20me/near/417649900)

### Steps to Reproduce

1. Run the above code

**Expected behavior:** `#synth` should either report an info message or a failure message

**Actual behavior:** Nothing happens.

### Versions

4.5.0-rc1

### Impact

leodemoura:
closes #2226
closes #3214

### DIFF
--- a/src/Lean/Elab/BuiltinCommand.lean
+++ b/src/Lean/Elab/BuiltinCommand.lean
@@ -229,7 +229,7 @@ private def replaceBinderAnnotation (binder : TSyntax ``Parser.Term.bracketedBin
 @[builtin_command_elab «variable»] def elabVariable : CommandElab
   | `(variable $binders*) => do
     -- Try to elaborate `binders` for sanity checking
-    runTermElabM fun _ => Term.withAutoBoundImplicit <|
+    runTermElabM fun _ => Term.withSynthesize <| Term.withAutoBoundImplicit <|
       Term.elabBinders binders fun _ => pure ()
     for binder in binders do
       let binders ← replaceBinderAnnotation binder
--- a/tests/lean/run/2226.lean
+++ b/tests/lean/run/2226.lean
@@ -1,0 +1,13 @@
+/--
+error: unsolved goals
+A : Nat
+⊢ Sort ?u.3
+-/
+#guard_msgs in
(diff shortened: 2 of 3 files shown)


leodemoura:
@semorrison This PR may affect Mathlib. If there are too many issues there, we should just skip it and fix the issues above when we address RFC #2452

semorrison:
@leodemoura, the fallout in Mathlib is manageable. The only change is that we can no longer change the explicitness of a variable, and in the same line introduce a new variable that depends on it.

(I'm not sure if that is an intended or avoidable effect of this PR!)

semorrison:
Mathlib should build now. Diff is at https://github.com/leanprover-community/mathlib4/commit/026f1b9315c058e9c0305bb832d8fc658634867e for reference.

eric-wieser:
> The only change is that we can no longer change the explicitness of a variable, and in the same line introduce a new variable that depends on it.

This was often a footgun in mathlib anyway, as sometimes it resulted in duplicate hypotheses.

leodemoura:
> > The only change is that we can no longer change the explicitness of a variable, and in the same line introduce a new variable that depends on it.
> 
> This was often a footgun in mathlib anyway, as sometimes it resulted in duplicate hypotheses.

Thanks for feedback. So, I will go ahead and merge the PR.

PR #4371, opened 2024-06-06
refactor: lake: simplify load code

tydeu:
Simplifies the Lake dependency resolution code. Largely split from #3998.


tydeu:
!bench

tydeu:
!bench

PR #4372, opened 2024-06-06
perf: linearity in HashMap.(insert|erase)

TwoFX:
The speedcenter doesn't seem to care much, but in [my benchmark](https://github.com/TwoFX/HashMap/blob/2e04b4f8440dd5900e555d83209cf0c084c18abc/Hashmap/Benchmark/InsertReplace.lean) for replace-heavy workloads, the fixed code is about 40% faster.

### DIFF
--- a/src/Lean/Data/HashMap.lean
+++ b/src/Lean/Data/HashMap.lean
@@ -112,7 +112,7 @@ def expand [Hashable α] (size : Nat) (buckets : HashMapBucket α β) : HashMapI
     let ⟨i, h⟩ := mkIdx (hash a) buckets.property
     let bkt    := buckets.val[i]
     if bkt.contains a then
-      -- make sure `bkt` is used linearly in the follwing call to `replace`
+      -- make sure `bkt` is used linearly in the following call to `replace`
       let buckets' := buckets.update i .nil h
       (⟨size, buckets'.update i (bkt.replace a b) (by simpa [buckets'])⟩, true)
     else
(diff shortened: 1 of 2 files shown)


TwoFX:
!bench

TwoFX:
!bench

leodemoura:
@TwoFX Great catch. Have you considered using `Array.modify`?

TwoFX:
@leodemoura I have, but I think it makes the code less clear in this case (because we already have `bkt`, so it's confusing to bind another name to it in the lambda passed to `modify`) and according to my [benchmark](https://github.com/TwoFX/HashMap/blob/bfabba0e6b2e19f7da7ce840189d3cbb11bfd586/Hashmap/Benchmark/InsertReplace.lean#L141) there is no statistically significant difference in speed between the two variants. I'm happy to change it if you prefer the `modify` version (though note that this will make the diff larger because I would have to add a new definition `HashMapBucket.modify`).

leodemoura:
> @leodemoura I have, but I think it makes the code less clear in this case (because we already have `bkt`, so it's confusing to bind another name to it in the lambda passed to `modify`) and according to my [benchmark](https://github.com/TwoFX/HashMap/blob/bfabba0e6b2e19f7da7ce840189d3cbb11bfd586/Hashmap/Benchmark/InsertReplace.lean#L141) there is no statistically significant difference in speed between the two variants. I'm happy to change it if you prefer the `modify` version (though note that this will make the diff larger because I would have to add a new definition `HashMapBucket.modify`).

Great. No need to change. Just wanted to know if you considered. I trust your judgment.

Issue #4288, opened 2024-05-26
RFC: pointless constant `outParam` in `MonadExcept` instance


JovanGerb:
### Proposal

https://github.com/leanprover/lean4/blob/f05a82799a01569edeb5e2594cd7d56282320f9e/src/Init/Prelude.lean#L2980

Get rid of this `outParam` which doesn't do anything at all. `OutParam` only has a meaning in a `class`, not in an `instance`.

Having it here is confusing.

Issue #4287, opened 2024-05-26
RFC: unused function in `SynthInstance`


JovanGerb:
### Proposal

Remove this unused function:

https://github.com/leanprover/lean4/blob/f05a82799a01569edeb5e2594cd7d56282320f9e/src/Lean/Meta/SynthInstance.lean#L197-L198

closed by #4277

Issue #4285, opened 2024-05-26
trace messages for type class resolution instantiate metavariables that they shouldn't


JovanGerb:
### Description

The type class trace messages sometimes show assignments to metavariables that they shouldn't

### Context

When using
```
set_option trace.Meta.synthInstance true
```

### Steps to Reproduce
write
```
set_option trace.Meta.synthInstance true

#synth HAdd Nat Nat _
```

**Expected behavior:**
trace shows
```
[Meta.synthInstance] ✅ HAdd Nat Nat ?m.4 ▼
  [] new goal HAdd Nat Nat _tc.0 ▶
  [] ✅ apply @instHAdd to HAdd Nat Nat ?m.5 ▶
  [] ✅ apply instAddNat to Add Nat ▶
  [resume] propagating Add Nat to subgoal Add Nat of HAdd Nat Nat Nat ▶
  [] result instHAdd 
```

**Actual behavior:**
trace shows
```
[Meta.synthInstance] ✅ HAdd ℕ ℕ ℕ ▼
  [] new goal HAdd ℕ ℕ _tc.0 ▶
  [] ✅ apply @instHAdd to HAdd ℕ ℕ ℕ ▶
  [] ✅ apply instAddNat to Add ℕ ▶
  [resume] propagating Add ℕ to subgoal Add ℕ of HAdd ℕ ℕ ℕ ▶
  [] result instHAdd 
```

### Versions

4.8.0

`isDefEq` faces the same problem, but does it correctly
It can be fixed using `withTraceNodeBefore`, or `withMCtx mctxBefore do addMessageContext`
I fix this here #4277.

nomeata:
JFTR: One might expect that
```
set_option pp.instantiateMVars false
```
could fix this, but it doesn’t.

Issue #4283, opened 2024-05-26
RFC: `instantiateMVars` is unnecessary to use in trace messages.


JovanGerb:
### Proposal

Trace messages automatically come with the current `MetavarContext`, so there is no need to also call `instantiateMVars`. For example:
https://github.com/leanprover/lean4/blob/f05a82799a01569edeb5e2594cd7d56282320f9e/src/Lean/Meta/SynthInstance.lean#L374
https://github.com/leanprover/lean4/blob/f05a82799a01569edeb5e2594cd7d56282320f9e/src/Lean/Meta/SynthInstance.lean#L415-L417

In the second example, `abstractMVars` also doesn't need `instantiateMVars`, because it already runs it on its input, so here instantiating metavariables is also redundant.

This also appears twice in this file:

https://github.com/leanprover/lean4/blob/f05a82799a01569edeb5e2594cd7d56282320f9e/src/Lean/Elab/Match.lean#L1152

Kha:
Closing as per reasoning in the PR

nomeata:
Small addendum, in case someone stumbles over this: The `Lean.Expr` data type has a computed field (`Lean.Data`) that stores whether the expression has `MVars`, so an extra `instantiateMVar` is very cheap and usually not worth the bother to fret about.

PR #4379, opened 2024-06-06
perf: deduplicate `← instantiateMVars (← inferType mvar)` in TC search

### LINKED ISSUE #4284: RFC: in TC resolution, `← instantiateMVars (← inferType mvar)` is computed 2 or 3 times over for every new goal
### Proposal

These three lines do this same computation:

https://github.com/leanprover/lean4/blob/f05a82799a01569edeb5e2594cd7d56282320f9e/src/Lean/Meta/SynthInstance.lean#L286
https://github.com/leanprover/lean4/blob/f05a82799a01569edeb5e2594cd7d56282320f9e/src/Lean/Meta/SynthInstance.lean#L243
https://github.com/leanprover/lean4/blob/f05a82799a01569edeb5e2594cd7d56282320f9e/src/Lean/Meta/SynthInstance.lean#L474

I propose to do it once, and pass this result as an argument to all these functions.

JovanGerb:
Closes #4284

### DIFF
--- a/src/Lean/Meta/SynthInstance.lean
+++ b/src/Lean/Meta/SynthInstance.lean
@@ -238,28 +238,20 @@ def getInstances (type : Expr) : MetaM (Array Instance) := do
       trace[Meta.synthInstance.instances] result.map (·.val)
       return result
 
-def mkGeneratorNode? (key mvar : Expr) : MetaM (Option GeneratorNode) := do
-  let mvarType  ← inferType mvar
-  let mvarType  ← instantiateMVars mvarType
-  let instances ← getInstances mvarType
-  if instances.isEmpty then
-    return none
-  else
-    let mctx ← getMCtx
-    return some {
-      mvar, key, mctx, instances
-      typeHasMVars := mvarType.hasMVar
-      currInstanceIdx := instances.size
-    }
-
 /--
   Create a new generator node for `mvar` and add `waiter` as its waiter.
   `key` must be `mkTableKey mctx mvarType`. -/
-def newSubgoal (mctx : MetavarContext) (key : Expr) (mvar : Expr) (waiter : Waiter) : SynthM Unit :=
-  withMCtx mctx do withTraceNode' `Meta.synthInstance do
-    match (← mkGeneratorNode? key mvar) with
-    | none      => pure ((), m!"no instances for {key}")
-    | some node =>
+def newSubgoal (key mvar mvarType : Expr) (waiter : Waiter) : SynthM Unit :=
+  withTraceNode' `Meta.synthInstance do
+    let instances ← getInstances mvarType
+    if instances.isEmpty then
+      return ((), m!"no instances for {key}")
+    else
+      let node := {
+        mvar, key, instances
+        mctx := ← getMCtx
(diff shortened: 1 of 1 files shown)


nomeata:
!bench

leodemoura:
Dear @JovanGerb ,

We have noticed that you have been submitting many PRs, and it is great to see your enthusiasm. Thank you for filing issues and proposing fixes. We are always looking for new contributors.
We want to share some guidelines based on our previous experiences. All successful external Lean contributors have started by making uncontroversial and small contributions. This approach helps build trust between the contributor and the core development team.
We strongly believe that critical modules that have been modified by multiple contributors become more challenging to maintain. For example, the type class resolution module has already been updated by Daniel, Gabriel, and myself. Daniel was the original author and is no longer part of the core team. Gabriel is a seasoned Lean developer who carefully studied this delicate module before making contributions.
We don't want to discourage you, but we would greatly appreciate it if you read our contribution guidelines and started by making contributions to simpler modules.
We’d encourage you avoid core modules and functionality initially, and maybe focus on low risk contributions (error messages, leaf tactics). Contributing to Mathlib is also a good way to build experience and reputation. We understand your eagerness to help with lean, and sympathize, but experience shows that it works better for everyone involved to take it slowly before, eventually maybe, touching core files.

Best wishes,
Leo and Sebastian (@Kha)

JovanGerb:
Dear @leodemoura and @Kha,

Thank you for your kind message. I can undestand that making significant changes to core Lean requires trust and careful consideration. And the benefits of the change need to outweigh the cost of worsened maintanability.
I think this applies most to my suggested changes in type class synthesis caching (#4272) and the fix for eta-unreduced metavariable assignments (#4387).
Then there are some less disruptive changes such as fixing the implementation of `isClassQuickConst?` (#4319) and this PR, which seem to me to not negatively impact maintainability (although I'm not the right person to judge this) and which still give a noticeable improvement. There is also the fix for metavariable assignments in `trace.Meta.synthInstance` (#4285), which improves the trace messages. The fix at #4282 might improve performance, but I haven't checked this.
Finally there are changes that don't impact the performance or user experience, and only improve readability of the code itself, such as removing redundant parts from the code (#4388, #4288, #4287 and #4283). Are such changes welcome or are they considered to be too insignificant?
In #4277 I made many independent changes to type class synthesis, which I know is against the contribution guidelines, but I did it to see how the performance would improve by doing all those changes together. It now turns out that the improvement is already captured by the change in this PR. I can make separate PRs for the other parts of that PR if that is helpful.
Finally I'd like to say that I really like Lean both as a programming language and as a theorem prover, and that I'm happy to be in a position where I can help make it even better.

Best whishes,
Jovan

Kha:
Thanks for your understanding. Maintainability of the code is one important aspect, but reviewing work is another one I would like to highlight. For a critical module like TC synth, there are basically no uncontroversial changes because every change must be reviewed with the full context of the module in mind, which usually means there is at most one potential reviewer and that person likely does not have that context ready in their mind right now.

> Are such changes welcome or are they considered to be too insignificant?

So even positive but insignificant changes to such modules can be distractive and will be considered very low priority. For significant changes, as Leo wrote one should start at less critical places.

PR #4382, opened 2024-06-06
perf: remove `@[inline]` from `NameMap.find?`

tydeu:
This `@[inline]` causes Lean to respecialize `RBMap.find?` to `NameMap` at each call site of `NameMap.find?`, creating lots of unnecessary duplicate IR.

### DIFF
--- a/src/Lean/Data/NameMap.lean
+++ b/src/Lean/Data/NameMap.lean
@@ -27,7 +27,7 @@ def insert (m : NameMap α) (n : Name) (a : α) := RBMap.insert m n a
 
 def contains (m : NameMap α) (n : Name) : Bool := RBMap.contains m n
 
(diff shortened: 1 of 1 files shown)


tydeu:
@Kha, do you have any idea why [interactive test 1403](https://github.com/leanprover/lean4/blob/master/tests/lean/interactive/1403.lean) would be [failing with a mutex error](https://github.com/leanprover/lean4/actions/runs/9408001999/job/25915008356?pr=4382#step:10:4092):

```
Fatal glibc error: pthread_mutex_lock.c:94 (___pthread_mutex_lock): assertion failed: mutex->__data.__owner == 0
```
This seems like it could just be some generally flaky issue with the server. I can't imagine how this change could cause this.

tydeu:
!bench

PR #4383, opened 2024-06-06
fix: don't duplicate goals in `rw`

### LINKED ISSUE #4381: `rw` duplicates goals for metavariables appearing in the rewritten expression
### Prerequisites

### Description

When rewriting a term that contains metavariables, this metavariable is added as a new goal in the tactic state, even if it already is a goal. This clutters the tactic state and is confusing.

### Context

Question on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Rewriting.20the.20goal.20with.20metavariables.20duplicates.20goals/near/442515619)

### Steps to Reproduce

The example provided on Zulip is:
```
example : ∀ d g, d = g → exists x : Nat, x = d := by
  intros d g H1
  constructor
  rewrite [H1,←H1,H1,←H1,H1]
  -- rotate_left; exact g; rfl
```

**Expected behavior:** [Clear and concise description of what you expect to happen]

The goal
```
dg: Nat
H1: d = g
⊢ Nat
```
should still appear only once after the rewrites

**Actual behavior:** [Clear and concise description of what actually happens]

The goal appears 6 times after the 5 rewrites.

### Versions

4.7.0 (works on any version)

### Impact

JovanGerb:
Closes #4381

### DIFF
--- a/tests/lean/rwDuplicateGoals.lean
+++ b/tests/lean/rwDuplicateGoals.lean
@@ -1,0 +1,15 @@
+
+/--
+error: unsolved goals
+case h
+h : 3 = 3
+⊢ ?w ≠ 3
+
+case w
+⊢ Nat
+-/
+#guard_msgs in
+example : ∃ n : Nat, (3=3) → n ≠ 3 := by
+  constructor
+  intro h
+  rw [h]


leodemoura:
@JovanGerb Thanks for creating the issue and submitting the PR. However, we will fix the issue using #4385.

PR #4388, opened 2024-06-07
chore: remove redundant `if-else` in `isDefEqQuickOther`

JovanGerb:
I removed a redundant `if tFn.isMVar || sFn.isMVar then ... else return LBool.undef` in the `else` clause of 
```
if !tFn.isMVar && !sFn.isMVar then
  return LBool.undef
else
```


JovanGerb:
awaiting-review

PR #4392, opened 2024-06-07
feat: relate BitVec.signExtend to truncate

bollu:
This adds helper lemmas to relate sign extension to truncation, and as a corollary shows that sign extension to the same width is a no-op.


tobiasgrosser (approved):

bollu:
awaiting-review

PR #4400, opened 2024-06-07
feat: switching List lookup normal forms to `L[n]` and `L[n]?`

kim-em:
This is presumably going to have significant breakage downstream.


semorrison:
I have Batteries working against this. Next up, Mathlib?

semorrison:
I have a local Mathlib build now (finally!), although I expected the `simpNF` linter will still have some complaints.

semorrison:
!bench

PR #4402, opened 2024-06-07
feat: lake: reliably cache logs and hashes

### LINKED ISSUE #2751: lake does not reliably regenerate .hash files
If you delete the .hash files, lake should regenerate them, but it only seems to regenerate `olean.hash` and not `ilean.hash` and `c.hash` files, even with the `-R` flag.

tydeu:
Moves the cached log into the trace file (no more `.log.json`). This means logs are no longer cached on fatal errors and this ensures that an out-of-date log is not associated with an up-to-date trace. Separately, `.hash` file generation was changed to be more reliable as well.  `.hash` files are deleted as part of the build and always regenerate with `--rehash`.

Closes #2751.

### DIFF
--- a/src/lake/Lake/Build/Common.lean
+++ b/src/lake/Lake/Build/Common.lean
@@ -67,7 +67,7 @@ as the point of comparison instead.
     return false
 
 /--
-Checks wether `info` is up-to-date, and runs `build` to recreate it if not.
+Checks whether `info` is up-to-date, and runs `build` to recreate it if not.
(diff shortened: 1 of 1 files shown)


tydeu:
!bench

semorrison:
Verified on `lean-pr-testing-4402`, but needed an upgrade to leantar, provided by @digama0 and merged in Mathlib at https://github.com/leanprover-community/mathlib4/pull/13808

PR #4404, opened 2024-06-08
chore: fix typo and incorrect name in doc

b-mehta:
Fixes typo "reflexivitiy" to "reflexivity", and changes exact Eq.rfl to exact rfl, since Eq.rfl does not exist.

(I got something confused wrt the bot message on #4367 and accidentally closed that one, so making this one instead, which I think satisfies the requirements it wanted.)

### DIFF
--- a/tests/lean/run/wfirred.lean
+++ b/tests/lean/run/wfirred.lean
@@ -39,7 +39,7 @@ error: The rfl tactic failed. Possible reasons:
 - The goal is not a reflexive relation (neither `=` nor a relation with a @[refl] lemma).
 - The arguments of the relation are not equal.
 Try using the reflexivity lemma for your relation explicitly, e.g. `exact Eq.refl _` or
-`exact HEq.rfl`
+`exact HEq.rfl` etc.
 n : Nat
 ⊢ foo (n + 1) = foo n
 -/


src/Init/Tactics.lean
```diff
@@ -371,7 +371,7 @@ reflexivity theorems (e.g., `Iff.rfl`).
 macro "rfl" : tactic => `(tactic| case' _ => fail "The rfl tactic failed. Possible reasons:
 - The goal is not a reflexive relation (neither `=` nor a relation with a @[refl] lemma).
 - The arguments of the relation are not equal.
-Try using the reflexivitiy lemma for your relation explicitly, e.g. `exact Eq.rfl`.")
+Try using the reflexivity lemma for your relation explicitly, e.g. `exact rfl`.")
```
nomeata, line 374:
```suggestion
Try using the reflexivity lemma for your relation explicitly, e.g. `exact rfl` or `exact HEq.rfl` etc.")
```

tests/lean/run/wfirred.lean
```diff
@@ -38,7 +38,7 @@ example : foo 0 = 0 := by rfl
 error: The rfl tactic failed. Possible reasons:
 - The goal is not a reflexive relation (neither `=` nor a relation with a @[refl] lemma).
 - The arguments of the relation are not equal.
-Try using the reflexivitiy lemma for your relation explicitly, e.g. `exact Eq.rfl`.
+Try using the reflexivity lemma for your relation explicitly, e.g. `exact rfl`.
```
nomeata, line 41:
```suggestion
Try using the reflexivity lemma for your relation explicitly, e.g. `exact rfl` or `exact HEq.rfl` etc.
```

nomeata (commented):
Thanks for the typo fix.

It would be helpful if we had `Eq.rfl`, so that it follows the naming convention of all the other relations which have both `Foo.refl` and `Foo.rfl`, and where `Foo.rfl` has the argument implicit. After all the message suggests to use the right one for your relation. Just writing `rfl` is confusing here, also with the tactic of the same name. Maybe we should introduce `Eq.rfl`. Until then maybe list two possible lemmas (there aren't many in core, unfortunately.)

WDYT?

b-mehta:
I agree that writing `Eq.rfl` would be better here, but since it doesn't exist, I guessed that it would be unreasonable to add a new theorem in such a low-level file as part of a typo-fix only PR. How about `Eq.refl _` for now, with a follow-up adding `Eq.rfl` in a separate PR?

Could you advise how to fix the CI issue? If I do the suggested rebase command then it doesn't allow me to push.

nomeata:
I don't think there is any rebasing to do. Once the test suite works (maybe the expected output isn't quite right yet) we can merge this.

leodemoura:
Waiting test to be fixed.

PR #4417, opened 2024-06-10
feat: BitVec.twoPow and lemmas, toward bitblasting multiplication for LeanSAT

bollu:
We add a new definition `BitVec.twoPow w i` to represent `(1#w <<< i)`.
This expression is used to test bits when building the multiplication bitblaster.

Patch 1/?, being peeled from https://github.com/opencompl/lean4/pull/6.

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -1381,7 +1381,7 @@ theorem getLsb_rotateRight {x : BitVec w} {r i : Nat} :
 theorem toNat_twoPow (w : Nat) (i : Nat) : (twoPow w i).toNat = 2^i % 2^w := by
   rcases w with rfl | w
   · simp [Nat.mod_one]
-  · simp [twoPow, toNat_shiftLeft]
+  · simp only [twoPow, toNat_shiftLeft, toNat_ofNat]
     have h1 : 1 < 2 ^ (w + 1) := Nat.one_lt_two_pow (by omega)
     rw [Nat.mod_eq_of_lt h1, Nat.shiftLeft_eq, Nat.one_mul]
 


src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1374,4 +1374,63 @@ theorem getLsb_rotateRight {x : BitVec w} {r i : Nat} :
...
+
+theorem BitVec.toNat_twoPow (w : Nat) (i : Nat) : (twoPow w i).toNat = 2^i % 2^w := by
+  rcases w with rfl | w
+  · simp [Nat.mod_one]
+  · simp [twoPow, toNat_shiftLeft]
+    have hone : 1 < 2 ^ (w + 1) := by
+      rw [show 1 = 2^0 by simp[Nat.pow_zero]]
+      exact Nat.pow_lt_pow_of_lt (by omega) (by omega)
+    simp [Nat.mod_eq_of_lt hone, Nat.shiftLeft_eq]
+
```
tobiasgrosser, line 1435:
```suggestion
```

This statement is already added further above.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1374,4 +1374,63 @@ theorem getLsb_rotateRight {x : BitVec w} {r i : Nat} :
...
+    have h1 : 1 < 2 ^ (w + 1) := Nat.one_lt_two_pow (by omega)
+    rw [Nat.mod_eq_of_lt h1]
+    rw [Nat.shiftLeft_eq, Nat.one_mul]
+
+@[simp]
+theorem getLsb_twoPow (i j : Nat) : (twoPow w i).getLsb j = ((i < w) && (i = j)) := by
+  rcases w with rfl | w
+  · simp only [twoPow, BitVec.reduceOfNat, Nat.zero_le, getLsb_ge, Bool.false_eq,
+    Bool.and_eq_false_imp, decide_eq_true_eq, decide_eq_false_iff_not]
+    omega
```
tobiasgrosser, line 1393: AFAIU, `simp; omega` is permitted, so you may be able to reduce the arguments of the simp if you drop the only. This also reappears further down.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1374,4 +1374,63 @@ theorem getLsb_rotateRight {x : BitVec w} {r i : Nat} :
   · simp
   · rw [← rotateRight_mod_eq_rotateRight, getLsb_rotateRight_of_le (Nat.mod_lt _ (by omega))]
 
+/- ## twoPow -/
+
+@[simp]
```
tobiasgrosser, line 1379:
```suggestion
@[simp, bv_toNat]
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1374,4 +1374,63 @@ theorem getLsb_rotateRight {x : BitVec w} {r i : Nat} :
...
+theorem toNat_twoPow (w : Nat) (i : Nat) : (twoPow w i).toNat = 2^i % 2^w := by
+  rcases w with rfl | w
+  · simp [Nat.mod_one]
+  · simp [twoPow, toNat_shiftLeft]
+    have h1 : 1 < 2 ^ (w + 1) := Nat.one_lt_two_pow (by omega)
+    rw [Nat.mod_eq_of_lt h1]
+    rw [Nat.shiftLeft_eq, Nat.one_mul]
```
tobiasgrosser, line 1386:
```suggestion
    rw [Nat.mod_eq_of_lt h1, Nat.shiftLeft_eq, Nat.one_mul]
```

src/Init/Data/BitVec/Lemmas.lean
bollu: CC @semorrison about this style nit.

semorrison: Yes, `simp; omega` is fine if you prefer, but I'm happy either way here.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1375,4 +1375,53 @@ theorem getLsb_rotateRight {x : BitVec w} {r i : Nat} :
...
+    · by_cases hi : Nat.testBit 1 (j - i)
+      · obtain hi' := Nat.testBit_one_eq_true_iff_self_eq_zero.mp hi
+        have hij : j = i := by omega
+        simp_all
+      · have hij : i ≠ j := by
+          intro h; subst h
+          simp at hi
+        simp_all
+
+theorem and_twoPow_eq_getLsb (x : BitVec w) (i : Nat) :
```
semorrison, line 1408: I think the name should just be `and_twoPow_eq`.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1375,4 +1375,53 @@ theorem getLsb_rotateRight {x : BitVec w} {r i : Nat} :
   · simp
   · rw [← rotateRight_mod_eq_rotateRight, getLsb_rotateRight_of_le (Nat.mod_lt _ (by omega))]
 
+/- ## twoPow -/
+
+@[simp, bv_toNat]
+theorem toNat_twoPow (w : Nat) (i : Nat) : (twoPow w i).toNat = 2^i % 2^w := by
+  rcases w with rfl | w
+  · simp [Nat.mod_one]
+  · simp [twoPow, toNat_shiftLeft]
```
semorrison, line 1384: Nonterminal simp.

semorrison:
Mostly happy, but:

Convince me that this needs a name, and shouldn't just be written out as its definition in all the statements.

The proofs are less trivial than I expected (they are fine), and I think this justifies giving a name, but I just want to push back a little so we don't accumulate named entities needlessly!

bollu:
@semorrison That's somewhat of a hard ask :) I abstracted it precisely because dealing with the raw definitions was unwieldy.

I've been mulling over this a little, and I wonder what the 'correct' approach is. The crux is that I want to break a bitvector into the bitwise-or of different `1 <<< i`'s, and having a name for this `1 <<< i` object makes things easy because:

1. It conceptually isolates the core idea of each induction.
1. It makes naming lemmas easier, as they now talk about `twoPow`.


--- 

I wonder what you recommend in such a situation. I also wonder how `mathlib` handles such scenarios!

semorrison:
No, I think your justification is satisfactory! Let's get the minor suggestions sorted out, and I can merge.

src/Init/Data/BitVec/Lemmas.lean
bollu: Chose `simp; omega`/

bollu:
I've fixed all nits now.

bollu:
awaiting-review

PR #4420, opened 2024-06-10
fix: occurs check at metavariable types

### LINKED ISSUE #4405: isDefEq can construct cyclic mvars
As per my current, possibly incomplete, understanding, a fair number of functions related to MVars assume that their assignments is acyclic, and will crash and burn otherwise (e.g. with an uncatchable stackoverflow). This is by design because developer should either use low-level functions like `MVarId.assign` and know what they are doing, or else use safer functions like `isDefEq`, which supposedly check for and prevent cycles.

But it seems that even using just `isDefEq` I can accidentally build bad assignments:
```
import Lean.Elab.Command
open Lean Meta

run_meta do
  let mvarIdNat ← mkFreshExprMVar (.some (.const ``Nat []))
  let mvarIdFin ← mkFreshExprMVar (.some (.app (.const `Fin []) mvarIdNat))
  -- mvarIdNat.assign (.app (.const ``Fin.val []) mvaridFin))
  let b ← isDefEq mvarIdNat (mkApp2 (.const ``Fin.val []) mvarIdNat mvarIdFin)
  logInfo m!"Defeq?: {b}" -- prints true
  -- Now mvaridNat occurs in its own type

  -- This will stack overflow
  -- let r ← abstractMVars mvarIdFin (levels := false)
  -- logInfo m!"{r.expr}"  
```

Are the checks are incomplete, or am I doing something else wrong here?

(Hint for other developers who have to debug failures due to potentially bad mvar assignments: I wrote a [function `checkMVarsCycles`](https://github.com/nomeata/lean-simplc/blob/03c747c1ab1584f9a496542750bc66b02cf860d7/Simplc/MVarCycles.lean) that traverses the assignment and throws the cycle pretty-printed.)

### Versions

4.10.0-nightly-2024-06-08

### Impact

leodemoura:
closes #4405

### DIFF
--- a/src/Lean/Meta/ExprDefEq.lean
+++ b/src/Lean/Meta/ExprDefEq.lean
@@ -929,6 +929,29 @@ partial def check
 
 end CheckAssignmentQuick
 
+/--
+Auxiliary function used at `typeOccursCheckImp`.
+Given `type`, it tries to eliminate "dependencies". For example, suppose we are trying to
+perform the assignment `?m := f (?n a b)` where
+```
+?n : let k := g ?m; A -> h k ?m -> C
+```
+If we just perform occurs check `?m` at the type of `?n`, we get a failure, but
+we claim these occurrences are ok because the type `?n a b : C`.
+In the example above, `typeOccursCheckImp` invokes this function with `n := 2`.
+Note that we avoid using `whnf` and `inferType` at `typeOccursCheckImp` to minimize the
+performance impact of this extra check.
+
+See test `typeOccursCheckIssue.lean` for an example where this refinement is needed.
+The test is derived from a Mathlib file.
+-/
+private partial def skipAtMostNumBinders (type : Expr) (n : Nat) : Expr :=
+  match type, n with
+  | .forallE _ _ b _, n+1 => skipAtMostNumBinders b n
+  | .mdata _ b,       n   => skipAtMostNumBinders b n
+  | .letE _ _ v b _,  n   => skipAtMostNumBinders (b.instantiate1 v) n
+  | type,             _   => type
+
 /-- `typeOccursCheck` implementation using unsafe (i.e., pointer equality) features. -/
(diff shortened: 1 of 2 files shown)


src/Lean/Meta/Basic.lean
```diff
@@ -1900,6 +1900,10 @@ abbrev isDefEqGuarded (t s : Expr) : MetaM Bool :=
 def isDefEqNoConstantApprox (t s : Expr) : MetaM Bool :=
   approxDefEq <| isDefEq t s
 
+/-- Shorthand for `isDefEq (mkMVar mvarId) val` -/
+def _root_.Lean.MVarId.checkedAssign (mvarId : MVarId) (val : Expr) : MetaM Bool :=
```
nomeata, line 1904: I'd expect this to return `Unit`

leodemoura: `throwError` or `panic` if `isDefEq` fails?

nomeata: I am never quite sure what’s the right one to use. I’d lean towards `throwError` because `panic` is confusing if things then continue to run and it may make sense to run a chunk of `MetaM` code and catch all such errors. But one could also argue that then the programmer should use the `Bool` variant? What would you expect?

leodemoura: It am also not sure. I used `Bool` because our current use-cases for the `isDefEq (mkMVar mvarId) v` idiom are all testing whether the assignment failed or not. For debugging programming mistakes where the assignment **must** not fail, I am inclined to have another function that returns `Unit`, and panics if the assignment fails. It is a personal preference because I debug panics using `LEAN_ABORT_ON_PANIC=1 lldb stage1/bin/lean`

leodemoura: I will merge the PR as-is since it fixes the bug, but let's continue the discussion. Please feel free to submit a PR with APIs that are convenient for you.

PR #4422, opened 2024-06-11
fix: (rcases) handle synthetic placeholders in targets

### LINKED ISSUE #4331: rcases leaking subgoals out of have
### Prerequisites

### Description

Using `rcases` on a term containing a hole leads to subgoals leaking out of the current `have`.

### Context

This was [discussed on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/rcases.20subgoals.20leaking.20out.20of.20have). I came across this while refactoring something in the sphere eversion project.

### Steps to Reproduce

My original minimization was:

```lean
example : True := by
  have : ∃ k, 5 = 2 + k := by
    -- Using apply with a hole create a new goal right here
    -- apply Nat.exists_eq_add_of_le (?_ : 2 ≤ 5)
    -- Using rcases teleports the new goal outside of the have block
    rcases Nat.exists_eq_add_of_le (?_ : 2 ≤ 5) with ⟨k, hk⟩
    exact ⟨k, hk⟩
  trivial -- This trivial solves 2 ≤ 5
  trivial -- this one takes care of True
```

Kyle came up with a variant where the hole is not directly inside `rcases` but in a `let`:

```lean
example : True := by
  have : ∃ k, 5 = 2 + k :=
    let h := Nat.exists_eq_add_of_le (?_ : 2 ≤ 5)
    by
      rcases h with ⟨k, hk⟩
      exact ⟨k, hk⟩
  trivial -- this one takes care of True
  trivial -- This trivial solves 2 ≤ 5
```

**Expected behavior:** [Clear and concise description of what you expect to happen]

The new subgoal should be inside the `have`.

**Actual behavior:** [Clear and concise description of what actually happens]

The new subgoal leaks outside the `have`.

### Versions

Found with 4.8.0, reproduced in 4.9.0-nightly-2024-06-02 on live.lean-lang.org.

### Impact

llllvvuu:
Calls `getMVarsNoDelayed` on the targets and adds them to the goals list.

Closes #4331

### DIFF
--- a/tests/lean/4331.lean
+++ b/tests/lean/4331.lean
@@ -1,0 +1,12 @@
+example : True := by
+  have : ∃ k, 5 = 2 + k := by
+    rcases Nat.exists_eq_add_of_le (?_ : 2 ≤ 5) with ⟨k, hk⟩
+    exact ⟨k, hk⟩
+    · decide
+  exact trivial
+
+example (f : (n : Nat) → n = 1 → ∃ m, n = m) : True := by
+  let n : Nat := 1
+  obtain ⟨_, _⟩ := f n ?g1
+  · exact trivial
+  · exact rfl


llllvvuu:
awaiting-review

llllvvuu:
wip

llllvvuu:
It looks like there are places in mathlib that rely on the "leaking" behavior of #4331. I guess I should make a mathlib patch to go with this?

semorrison:
@llllvvuu, I've merged your Mathlib PR into `lean-pr-testing-4422`. In general, it's easier to just push commits directly to such branches, rather than going through a PR process.

PR #4424, opened 2024-06-11
fix: bad 'unknown package' error message

### LINKED ISSUE #4419: `lean_exe` with `srcDir` duplicates build artifacts of `Main.lean` in two places
### Prerequisites

### Description

Using `srcDir` in a executable like
```
lean_exe my_command
  srcDir := "MyFolder"
  root := `Main
  [...]
```

causes the build artifacts of `Main` to be stored at `.lake/packages/MyPackage/.lake/build/lib/` instead of `.lake/packages/MyPackage/.lake/build/lib/MyFolder/` which has negativ implications:

* building `Main.lean` through another mean later (e.g. a `lean_lib`) will cause all built files to be duplicated, e.g. `Main.olean` and `MyFolder/Main.olean`
* It seems impossible to call  ``Lean.importModules `MyFolder.Main`` (or ``Lean.findOLean `MyFolder.Main``) unless `Main` is built through this second, separate mean.

### Context

[Minimalised Reproduction](https://github.com/joneugster/testSrcDir2)

[Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/making.20a.20lean.20executable.20available.20in.20a.20project/near/443792753)

### Steps to Reproduce

Follow the README of the [Minimalised Reproduction](https://github.com/joneugster/testSrcDir2) repo.

**Expected behavior:**

directly calling `lake exe testsrcdir` should succeed.

**Actual behavior:**

Instead one needs to call `lake build testSrcDir` first to build the second copy of `Main` before the command succeeds.

### Versions

v4.8.0

Linux, albeit probably not relevant

### Impact

Kha:
This message is older than corresponding, better checks in Lake and vscode-lean4

Fixes #4419

### DIFF
--- a/src/lake/tests/badImport/test.sh
+++ b/src/lake/tests/badImport/test.sh
@@ -13,7 +13,7 @@ LAKE=${LAKE:-../../.lake/build/bin/lake}
 # https://github.com/leanprover/lean4/issues/3809
 
 # Test importing a nmissing module from outside the workspace
-($LAKE build +X 2>&1 && exit 1 || true) | grep --color -F "X.lean:1:0: unknown package 'Y'"
+($LAKE build +X 2>&1 && exit 1 || true) | grep --color -F "X.lean:1:0: unknown module prefix 'Y'"
 $LAKE setup-file ./X.lean Y # Lake ignores the file (the server will error)
 # Test importing onself
 ($LAKE build +Lib.A 2>&1 && exit 1 || true) | grep --color -F "A.lean: module imports itself"


src/Lean/Util/Path.lean
```diff
@@ -108,16 +108,10 @@ partial def findOLean (mod : Name) : IO FilePath := do
...
-    let rec maybeThisOne dir := do
-      if ← (dir / pkg).isDir then
-        return some s!"\nYou might need to open '{dir}' as a workspace in your editor"
-      if let some dir := dir.parent then
-        maybeThisOne dir
-      else
-       return none
-    if let some msg' ← maybeThisOne (← IO.currentDir) then
-      msg := msg ++ msg'
+    let mut msg := s!"unknown package '{pkg}'
```
tydeu, line 111:
```suggestion
    let mut msg := s!"unknown module prefix '{pkg}'
```

This can help prevent confusion with Lake's notion of "package".

PR #4425, opened 2024-06-11
fix: redundant and wrong case in `forEachModuleInDir`

Kha:
Calling `f` on `dir.lean` for each `dir` is unnecessary as we call it on each file anyway and the check was wrong because it ignored `entry.root`.


Kha:
@tydeu Honestly, because it made writing the docstring easier. I didn't want to worry about hypothetical cases at that point.

PR #4433, opened 2024-06-12
feat: default pp if pp expr/syntax/level without context

### LINKED ISSUE #4432: `m!"{expr}"` doesn't work as of 4.9.0-rc1
### Prerequisites

### Description

As of 4.9.0-rc1, specifically the change to `MessageData.ofExpr` in https://github.com/leanprover/lean4/commit/9f6bbfa1066adcdf4900e985641a1b944bc77516, the following no longer works:

```lean
import Lean

def main : IO Unit := do
  let foo := Lean.mkConst `foo
  let bar ← m!"{foo}".toString
  IO.println s!"Expr: {foo}"
  IO.println s!"MessageData: {bar}"

#eval main
```

Feel free to close if this is intended behavior.

If not, `MessageData.ofExpr` (and similar functions) could be changed back to something like this:.
```lean
def ofExpr (e : Expr) : MessageData :=
  .ofLazy (fun ctx? => do
            let msg ← ofFormatWithInfos <$> match ctx? with
              | .none => pure (format (toString e))
              | .some ctx => ppExprWithInfos ctx e
            return Dynamic.mk msg)
          (fun mctx => instantiateMVarsCore mctx e |>.1.hasSyntheticSorry)
```

### Context

Batteries' `lake exe runLinter` is now printing "(invalid MessageData.lazy, missing context)" which is showing up in Batteries and Mathlib CI.

https://github.com/leanprover-community/batteries/pull/838

### Steps to Reproduce

Try the above code in 4.9.0-rc1 vs 4.8.0.

**Expected behavior:** Prints "MessageData: foo"

**Actual behavior:** Prints "MessageData: (invalid MessageData.lazy, missing context)"

### Versions

4.9.0-rc1

### Impact

llllvvuu:
This restores the behavior prior to https://github.com/leanprover/lean4/commit/9f6bbfa1066adcdf4900e985641a1b944bc77516 for `MessageData.ofSyntax` `MessageData.ofExpr`, and `MessageData.ofLevel` while staying within the new `.ofLazy` paradigm.

Also adds some documentation to help developers understand the missing context issue.

Closes #4432

### DIFF
--- a/src/Lean/Message.lean
+++ b/src/Lean/Message.lean
@@ -403,11 +403,13 @@ def indentExpr (e : Expr) : MessageData :=
   indentD e
 
 class AddMessageContext (m : Type → Type) where
-  /-- Without context, a `MessageData` object may be be missing information
-      (e.g. hover info) for pretty printing, or may print an error. Hence,
-      `addMessageContext` should be called on all constructed `MessageData`
-      (e.g. via `m!`) before taking it out of context (e.g. leaving `MetaM` or
-      `CoreM`). -/
+  /-- 
+  Without context, a `MessageData` object may be be missing information
+  (e.g. hover info) for pretty printing, or may print an error. Hence,
+  `addMessageContext` should be called on all constructed `MessageData`
+  (e.g. via `m!`) before taking it out of context (e.g. leaving `MetaM` or
+  `CoreM`).
+  -/
   addMessageContext : MessageData → m MessageData
 
 export AddMessageContext (addMessageContext)


llllvvuu:
Probably a judgment call here, I think it's also fine to require a context. In which case, maybe the right move is to beef up the error messages accordingly.

nomeata:
More discussion at https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/MessageData.20without.20context/near/444163996

src/Lean/Message.lean
```diff
@@ -385,6 +403,11 @@ def indentExpr (e : Expr) : MessageData :=
   indentD e
 
 class AddMessageContext (m : Type → Type) where
+  /-- Without context, a `MessageData` object may be be missing information
+      (e.g. hover info) for pretty printing, or may print an error. Hence,
+      `addMessageContext` should be called on all constructed `MessageData`
+      (e.g. via `m!`) before taking it out of context (e.g. leaving `MetaM` or
+      `CoreM`). -/
```
nomeata, line 410:
```suggestion
  /-- 
  Without context, a `MessageData` object may be be missing information
  (e.g. hover info) for pretty printing, or may print an error. Hence,
  `addMessageContext` should be called on all constructed `MessageData`
  (e.g. via `m!`) before taking it out of context (e.g. leaving `MetaM` or
  `CoreM`).
  -/
```

PR #4434, opened 2024-06-12
feat: use priorities to ensure simp applies eqational lemmas in order

### LINKED ISSUE #4173: simp should try more specific lemmas first
Consider the following code
```
set_option autoImplicit false 

inductive DOW where
 | monday | tuesday | wednesday | thursday | friday | saturday | sunday

 def DOW.isWorkday (n : Nat) (d : DOW) : Bool :=
  match d with
   | monday | tuesday | wednesday | thursday | friday => true
   | _ => match n with | 0 => false | n+1 => d.isWorkday n

example : DOW.friday.isWorkday 0 := by
  set_option trace.Meta.Tactic.simp true in
  simp  [DOW.isWorkday]

/-- info: DOW.isWorkday.eq_5 (n : Nat) : DOW.isWorkday n DOW.friday = true -/
#guard_msgs in
#check DOW.isWorkday.eq_5
/--
info: DOW.isWorkday.eq_6 (d : DOW) (x_1 : d = DOW.monday → False) (x_2 : d = DOW.tuesday → False)
  (x_3 : d = DOW.wednesday → False) (x_4 : d = DOW.thursday → False) (x_5 : d = DOW.friday → False) :
  DOW.isWorkday 0 d = false
-/
#guard_msgs in
#check DOW.isWorkday.eq_6
```

The recursion in `DOW.isWorkday` is there so that we get equational theorems; with #3983 this would apply to a non-recursive function as well.

The lemma `DOW.isWorkday.eq_5` should solve this rather quickly. But as the trace reveals, `simp` tries the rather more complicated `DOW.isWorkday.eq_6` first!

(In fact, it does something even weirder that I don’t fully understand, unfolding the definition and then discharging `d = DOW.thursday` etc. but using them as rewrite rules? There may be another issue lurking here.)

A possible fix, explored in #4158, might be to tweak the DiscrTree to return more specific lemmas first, but there might be other approaches (e.g. ordering all lemmas returned from the discrtree somehow), hence opening this issue.

nomeata:
This assigns priorities to the equational lemmas so that more specific ones
are tried first before a possible catch-all with possible side-conditions.

We assign very low priorities to match the simplifiers behavior when unfolding
a definition, which happens in `simpLoop`’ `visitPreContinue` after applying
rewrite rules.

Definitions with more than 100 equational theorems will use priority 1 for all
but the last (a heuristic, not perfect).

fixes #4173, to some extent.

### DIFF
--- a/src/Init/Tactics.lean
+++ b/src/Init/Tactics.lean
@@ -1459,6 +1459,7 @@ have been simplified by using the modifier `↓`. Here is an example
 ```
 
 When multiple simp theorems are applicable, the simplifier uses the one with highest priority.
+The equational theorems of function are applied at very low priority (100 and below).
 If there are several with the same priority, it is uses the "most recent one". Example:
 ```lean
 @[simp high] theorem cond_true (a b : α) : cond true a b = a := rfl
--- a/src/Lean/Meta/Tactic/Simp/SimpTheorems.lean
+++ b/src/Lean/Meta/Tactic/Simp/SimpTheorems.lean
@@ -468,7 +468,23 @@ def SimpTheorems.addDeclToUnfold (d : SimpTheorems) (declName : Name) : MetaM Si
     let mut d := d
     for h : i in [:eqns.size] do
       let eqn := eqns[i]
-      d ← SimpTheorems.addConst d eqn (prio := 999 - i)
+      /-
+      We assign priorities to the equational lemmas so that more specific ones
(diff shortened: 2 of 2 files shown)


nomeata:
TODO: Think about the right priority area, and document simp priorities somewhere.

nomeata:
What, no broken mathlib? Surely that can only mean that my change has somehow no effect ;-)

PR #4438, opened 2024-06-12
doc: documenting Char and upstreaming extensionality from batteries

jtristan:
* Basic documentation for characters
* Upstreamed two extensionality theorems from batteries

### DIFF
--- a/doc/char.md
+++ b/doc/char.md
@@ -9,5 +9,3 @@ Syntactically, character literals are enclosed in single quotes.
 ```
 
 Characters are ordered and can be decidably compared using the relational operators `=`, `<`, `≤`, `>`, `≥`.
-
-Type `Char` is not empty.


src/Init/Data/Char/Lemmas.lean
```diff
@@ -31,6 +31,11 @@ theorem utf8Size_eq (c : Char) : c.utf8Size = 1 ∨ c.utf8Size = 2 ∨ c.utf8Siz
   rw [Char.ofNat, dif_pos]
   rfl
 
+theorem Char.ext : {a b : Char} → a.val = b.val → a = b
```
semorrison, line 34:
```suggestion
@[ext] theorem Char.ext : {a b : Char} → a.val = b.val → a = b
```
We should preserve the `ext` attribute here. (This enables use of the `ext` tactic to automatically apply any such lemmas.)

doc/char.md
```diff
@@ -1 +1,13 @@
...
+
+Syntactically, character literals are enclosed in single quotes.
+```lean
+#eval 'a' -- 'a'
+#eval '∀' -- '∀'
+```
+
+Characters are ordered and can be decidably compared using the relational operators `=`, `<`, `≤`, `>`, `≥`.
+
+Type `Char` is not empty.
```
semorrison, line 13:
```suggestion
```
I would remove this, unless I'm missing something here.

semorrison:
The CI failure is not your fault: unfortunately this PR landed while we were in the middle of switching over our CI machine provider. I *think* that rebasing on `master` should result in CI re-running successfully.

PR #4439, opened 2024-06-12
feat: configuration DSL touchups

### LINKED ISSUE #3385: `lake init` should not write imports with French quotes unnecessarily.
`lake init foo` creates `Foo.lean` containing `import «Foo».Basic`, which is quite confusing, and unnecessary.

Could we avoid confusing new users and teaching them unhelpful habits by not using French quotes unless necessary.

tydeu:
This incorporates many general Lake DSL changes from #2439 and adds some new related changes.

* Rework configuration names (e.g., `package <name>`)
  * String literals ca now be used instead of identifiers for names.
  * The name syntax is now optional and can instead be set via the `name` field.
* Avoid French quotes in `lake new` / `lake init` templates (except in `lean_lib` names). This is not done for `lean_lib` because it needs a proper identifier for its root. It could use a string and reparse it as an identifier, but this seems liable to produce confusion.
* The `exe` templates now names it main module `Main` like the `std` template.
* Improve `math` template error if `lean-toolchain` fails to download.
* Lake now logs a warning rather than an error on unknown configuration fields. This increases the Lake DSL's cross-version compatibility.

Closes #3385.

### DIFF
--- a/nix/bootstrap.nix
+++ b/nix/bootstrap.nix
@@ -170,7 +170,7 @@ rec {
           ln -sf ${lean-all}/* .
         '';
         buildPhase = ''
-          ctest --output-junit test-results.xml --output-on-failure -E 'leancomptest_(doc_example|foreign)|leanlaketest_init' -j$NIX_BUILD_CORES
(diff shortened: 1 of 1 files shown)


tydeu:
Many of the elements of this PR are ported from #2439 (along with some new additions).

PR #4447, opened 2024-06-13
feat: add `#discr_tree_key` command and `discr_tree_key` tactic

mattrobball:
Adds a command and tactic to print the `Array <| DiscrTree.Key` for equalities helping the user to debug perceived `simp` failures.

### DIFF
--- a/src/Lean/Elab/Tactic/DiscrTreeKey.lean
+++ b/src/Lean/Elab/Tactic/DiscrTreeKey.lean
@@ -19,18 +19,18 @@ private def mkKey (e : Expr) (simp : Bool) : MetaM (Array Key) := do
   let type ← whnfR type
   if simp then
     if let some (_, lhs, _) := type.eq? then
-      DiscrTree.mkPath lhs simpDtConfig
+      mkPath lhs simpDtConfig
     else if let some (lhs, _) := type.iff? then
-      DiscrTree.mkPath lhs simpDtConfig
+      mkPath lhs simpDtConfig
     else if let some (_, lhs, _) := type.ne? then
-      DiscrTree.mkPath lhs simpDtConfig
+      mkPath lhs simpDtConfig
     else if let some p := type.not? then
       match p.eq? with
       | some (_, lhs, _) =>
-        DiscrTree.mkPath lhs simpDtConfig
-      | _ => DiscrTree.mkPath p simpDtConfig
+        mkPath lhs simpDtConfig
+      | _ => mkPath p simpDtConfig
     else
-      DiscrTree.mkPath type simpDtConfig
+      mkPath type simpDtConfig
   else
     mkPath type {}
 


nomeata:
I see likely unrelated changes in the diff, is that intentional?

mattrobball:
No, due to a lack of sleep. Fixed now. Thanks

src/Lean/Elab/Tactic/DiscrTreeKey.lean
```diff
@@ -0,0 +1,51 @@
...
+  | some (_, lhs, _) => mkPath lhs simpDtConfig
+  | none => throwError "unexpected kind of 'simp' theorem{indentExpr type}"
+
+@[builtin_command_elab Lean.Parser.discrTreeKeyCmd]
+def evalDiscrTreeKeyCmd : CommandElab := fun stx => do
+  Command.liftTermElabM <| do
+    match stx with
+    | `(command| #discr_tree_key $t:term) => do
+      if let `($id:ident) := t then
+        let info ← getConstInfo id.getId
```
nomeata, line 30: Should this be 
```suggestion
        let info ← getConstInfo (← realizeGlobalConstNoOverloadWithInfo id)
```
so that this resolves the name properly with namespaces and such?

src/Init/Notation.lean
```diff
@@ -687,6 +687,11 @@ syntax (name := checkSimp) "#check_simp " term "~>" term : command
 -/
 syntax (name := checkSimpFailure) "#check_simp " term "!~>" : command
 
+/--
+#discr_tree_key t prints the discrimination tree key for `t`.
```
nomeata, line 691: It seems you command has two modes, worth documenting here. Maybe something like
```suggestion
#discr_tree_key t prints the discrimination tree key for the term `t`, or if `t`
is an identifier, for its type.
```

mattrobball:
awaiting-review

src/Init/Notation.lean
```diff
@@ -687,6 +687,12 @@ syntax (name := checkSimp) "#check_simp " term "~>" term : command
 -/
 syntax (name := checkSimpFailure) "#check_simp " term "!~>" : command
 
+/--
+#discr_tree_key t prints the discrimination tree key for the term `t`, or if `t`
+is an identifier, for its type.
```
nomeata, line 692: I misread the code earlier, and the examples clarify the semantics when you pass a term that is not the identifier. So maybe
```suggestion
`#discr_tree_key i` prints the discrimination tree key for the theorem with name `i` (which should be an equality).

`#discr_tree_key t` prints the discrimination tree key for the type `t` (which should be the type of a simp theorem).
```

It feels a bit weird that you have to give the RHS even though it’s ignored. I would have expected a syntax like
```
#discr_tree_key List.map f (List.map g l)
```
or
```
#discr_tree_key (f : a -> b) (g : b -> c) (l : List a) : List.map f (List.map g l)
```
to get the key for an expression. Note that DiscrTrees are useful outside of simp (instance search for example).

This also indicates that maybe the other form, where you give an identifier of a simp theorem, should be its own command
```
#discr_tree_key_of_simp_lemma List.filter_map
```

nomeata (commented):
Sorry for the salami-sliced feedback, but I don’t always see everything ahead of times

src/Lean/Elab/Tactic/DiscrTreeKey.lean
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Lean.Elab.Tactic.DiscrTreeKey
+
+open Lean.Meta DiscrTree
+open Lean.Elab.Tactic
+open Lean.Elab.Command
+
+private def mkKey (e : Expr) : MetaM (Array Key) := do
+  let (_, _, type) ← withReducible <| forallMetaTelescopeReducing e
+  let type ← whnfR type
```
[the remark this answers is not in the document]
lecopivo: It is not good idea to return `type` from a telescope as it will end up with fvars that are not in the context anymore. However, I'm not sure if it matters for `DiscrTree.mkPath` or not.

I would write it as
```
private def mkKey (e : Expr) : MetaM (Array Key) :=
  withReducible <| forallTelescopeReducing e fun _ type => do
    let type ← whnfR type
    match type.eq? with
    | some (_, lhs, _) => mkPath lhs simpDtConfig
    | none => throwError "unexpected kind of 'simp' theorem{indentExpr type}"
```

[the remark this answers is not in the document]
nomeata: Ah, yes, of course, keep everything in the `forallTelescopeReducing`. My bad.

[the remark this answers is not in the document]
mattrobball: I did notice some dropped fvars in the examples so I will try this out. Thanks!

src/Init/Notation.lean
mattrobball: I agree. One of the main tripping points I've seen is instance synthesis. So I would certainly want something to handle that also. I started with this to be more incremental but I don't think there is much work to get this functionality. 

What other functionality do you think might be useful? Identification of where a matching succeeds or fails would be good but might be better left for the future.

src/Lean/Elab/Tactic/DiscrTreeKey.lean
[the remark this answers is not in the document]
mattrobball: We end up with loose `bvar`'s without using `forAllMetaTelescopingReducing` to convert them into metavariables. Using this seems existing pattern, see `Lean.Meta.Simp.mkSimpTheoremCore` (private). So the output with this is consistent with the output of `set_option diagnostics`.

src/Init/Notation.lean
mattrobball: I separated them into two commands and tactics mainly because there is no `config` syntax in `Init.Notation`. 

I couldn't get the anti-quotation syntax quite right to unify the two branches for each elaboration function. Suggestions welcome.

src/Init/Notation.lean
```diff
@@ -687,6 +687,27 @@ syntax (name := checkSimp) "#check_simp " term "~>" term : command
...
 syntax (name := checkSimpFailure) "#check_simp " term "!~>" : command
 
+/--
+`#discr_tree_key  t` prints the discrimination tree keys for the type of a term `t` or an identifier.
+It uses the default configuration for generating keys.
+
+For example,
+```
+#discr_tree_key (∀ {a n : Nat}, bar a (OfNat.ofNat n) = default)
+-- @Eq Nat (bar _ (@OfNat.ofNat Nat _ _)) (@default Nat _)
```
nomeata, line 697: Just to drive down the point that `discr_tree_key` does not need an equality, maybe


```suggestion
#discr_tree_key (∀ {a n : Nat}, bar a (OfNat.ofNat n))
-- bar _ (@OfNat.ofNat Nat _ _)
```

Are the parentheses needed?

src/Init/Notation.lean
```diff
@@ -687,6 +687,27 @@ syntax (name := checkSimp) "#check_simp " term "~>" term : command
 -/
 syntax (name := checkSimpFailure) "#check_simp " term "!~>" : command
 
+/--
+`#discr_tree_key  t` prints the discrimination tree keys for the type of a term `t` or an identifier.
+It uses the default configuration for generating keys.
```
nomeata, line 692: I’m still confused as to whether the command gives you the key for the type given, or for the type of the term given. The doc string says the latter (“type of a term `t`”), but the example indicates the former. So probably.

```suggestion
`#discr_tree_key  t` prints the discrimination tree keys for a term `t` (or, if it is a single identifier, the type of that constant).
It uses the default configuration for generating keys.
```

src/Init/Tactics.lean
```diff
@@ -933,6 +933,20 @@ example : False ∨ True := by
 -/
 syntax (name := dbgTrace) "dbg_trace " str : tactic
 
+/--
+Similar to `#discr_tree_key` command, `discr_tree_key t?` prints the discrimination tree keys
+for the type of a term `t` or an identifier if `t` is specified.
+Otherwise, it prints the keys for the main goal.
+It uses the default configuration for generating keys.
```
nomeata, line 940: How often do you need this tactic, compared to just copy’n’pasting the goal? I’m a bit worried about polluting the default tactic namespace (and the auto completion, where it will affect beginners) with relatively obscure debugging tools.

src/Lean/Elab/Tactic/DiscrTreeKey.lean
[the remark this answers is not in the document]
nomeata: Odd, but fine.

src/Lean/Elab/Tactic/DiscrTreeKey.lean
```diff
@@ -0,0 +1,82 @@
...
+private def getType (t : TSyntax `term) : TermElabM Expr := do
+  if let `($id:ident) := t then
+    if let some ldecl := (← getLCtx).findFromUserName? id.getId then
+      return ldecl.type
+    else
+      let info ← getConstInfo (← realizeGlobalConstNoOverloadWithInfo id)
+      return info.type
+  else
+  let e ← Term.elabTerm t none
+  return e
```
nomeata, line 36:
```suggestion
private def getType (t : TSyntax `term) : TermElabM Expr := do
  if let `($id:ident) := t then
    if let some ldecl := (← getLCtx).findFromUserName? id.getId then
      return ldecl.type
    else
      let info ← getConstInfo (← realizeGlobalConstNoOverloadWithInfo id)
      return info.type
  else
    Term.elabTerm t none
```

src/Lean/Elab/Tactic/DiscrTreeKey.lean
```diff
@@ -0,0 +1,82 @@
...
+open Lean.Elab.Tactic
+open Lean.Elab.Command
+
+private def mkKey (e : Expr) (simp : Bool) : MetaM (Array Key) := do
+  let (_, _, type) ← withReducible <| forallMetaTelescopeReducing e
+  let type ← whnfR type
+  if simp then
+    match type.eq? with
+    | some (_, lhs, _) => mkPath lhs simpDtConfig
+    | none => throwError "unexpected kind of 'simp' theorem{indentExpr type}"
```
nomeata, line 23:
```suggestion
    match type.eq? with
    | some (_, lhs, _) => mkPath lhs simpDtConfig
    | none => throwError "unexpected kind of 'simp' theorem{indentExpr type}"
```

Simp theorems can have other forms, I think. They support `Iff` lemmas. And other lemmas can be `simp` as well, they are implicit converted to `… = True`.  Not sure if you want to support that. See the private `preprocess` used by `mkSimpTheoremsFromConst`.

If you plan to not support that, maybe leave a comment that this is intentional.

src/Init/Notation.lean
mattrobball: They come from `Lean.Meta.DiscrTree.keysAsPattern` so that would need changing. Personally, I think they aid readability.

src/Lean/Elab/Tactic/DiscrTreeKey.lean
mattrobball: Good point. I'll look into this

src/Init/Tactics.lean
mattrobball: We can downstream it to batteries. @semorrison asked to put things here so perhaps she can weigh in. Either is fine with me.

nomeata: If it's useful (and not just added because we can) no complaints. But also happy to hear what @semorrison says; this isn't my particular area of expertise anyways.

semorrison:
@mattrobball is there meant to be an update to stage0 included in this PR? If so, can you please rewrite the git history so that that these updates happen in a single commit, nothing else happens in that commit, and the commit message is exactly "chore: update stage0".

However I'm hoping that changes to stage0 are actually *not* needed, and you can remove them from this PR.

src/Init/Notation.lean
```diff
@@ -687,6 +687,27 @@ syntax (name := checkSimp) "#check_simp " term "~>" term : command
...
+For example,
+```
+#discr_tree_key (∀ {a n : Nat}, bar a (OfNat.ofNat n))
+-- bar _ (@OfNat.ofNat Nat _ _)
+
+#discr_tree_simp_key Nat.add_assoc
+-- @HAdd.hAdd Nat Nat Nat _ (@HAdd.hAdd Nat Nat Nat _ _ _) _
+```
+
+`#discr_tree_sim_key` is similar to `#discr_tree_key`, but it assumes the underlying type is
```
semorrison, line 703:
```suggestion
`#discr_tree_simp_key` is similar to `#discr_tree_key`, but it assumes the underlying type is
```

src/Init/Tactics.lean
semorrison: I think the command is sufficient, and we don't need the tactic.

semorrison (approved):
Looking forward to having this in!

mattrobball:
> @mattrobball is there meant to be an update to stage0 included in this PR? If so, can you please rewrite the git history so that that these updates happen in a single commit, nothing else happens in that commit, and the commit message is exactly "chore: update stage0".
> 
> However I'm hoping that changes to stage0 are actually _not_ needed, and you can remove them from this PR.

I got caught in some type of toolchain purgatory and kinda flailed around until I got a working state. I will try to extract myself.

mattrobball:
Fixed my issue (of course silly). 

Changes: 
- removed tactics
- added support for all simp preprocessing steps except conjunction (for simplicity)
- moved test to `run` and improved context and documentation

mattrobball:
awaiting-review

src/Init/Notation.lean
```diff
@@ -703,6 +703,27 @@ syntax (name := checkSimp) "#check_simp " term "~>" term : command
...
+an equality and prints the left-hand side using the `simp` configuration for generating keys.
```
nomeata, line 720:
```suggestion
`#discr_tree_simp_key` is similar to `#discr_tree_key`, but treats the underlying type
as one of a simp lemma, i.e. transforms it into an equality and produces the key of the
left-hand side.
```

src/Lean/Elab/Tactic/DiscrTreeKey.lean
```diff
@@ -0,0 +1,65 @@
...
+      DiscrTree.mkPath lhs simpDtConfig
+    else if let some p := type.not? then
+      match p.eq? with
+      | some (_, lhs, _) =>
+        DiscrTree.mkPath lhs simpDtConfig
+      | _ => DiscrTree.mkPath p simpDtConfig
+    else
+      DiscrTree.mkPath type simpDtConfig
+  else
+    mkPath type {}
```
nomeata, line 35:
```suggestion
  if simp then
    if let some (_, lhs, _) := type.eq? then
      mkPath lhs simpDtConfig
    else if let some (lhs, _) := type.iff? then
      mkPath lhs simpDtConfig
    else if let some (_, lhs, _) := type.ne? then
      mkPath lhs simpDtConfig
    else if let some p := type.not? then
      match p.eq? with
      | some (_, lhs, _) =>
        mkPath lhs simpDtConfig
      | _ => mkPath p simpDtConfig
    else
      mkPath type simpDtConfig
  else
    mkPath type {}
```

tests/lean/run/discrTreeKey.lean
```diff
@@ -0,0 +1,79 @@
...
+
+open Nat List
+
+/-!
+  We can produce `simp` keys for theorems of the form `=`, `≠`, and `↔` by supplying the name
+  of the declaration.
+-/
+
+#check zero_le
+#discr_tree_simp_key zero_le
```
nomeata, line 19: Now that this is moved to `run/`, presumably you intended to put
```
#guard_msgs in
```
before each command whose output you want to test in this file?

nomeata (changes requested):
Very nice, almost there!

tests/lean/run/discrTreeKey.lean
mattrobball: I'm no longer sure what I'm testing other than it completely breaking. 

I think it's fine if the format of output changes so I don't want to constrain that. 

This seems more like documentation now.

nomeata: I’d include the expected output of each command in this test, using
```
/-- … -/
#guard_msgs in
#discr_tree_simp_key zero_le
```
(I hope you know about the code action that will update the expected output, and its key combination, so that this is not too much work.)

This way, if someone possibly inadvertently changes how the keys are calculated, this test will notify them. And if it is an intentional change, they can easily update it.


But I see your point that then this test isn’t testing your command, it’s now testing the key calculation (with the help of your command). Fair enough, I'll merge this; we can add the test output later if we really want that.

nomeata (approved):

PR #4448, opened 2024-06-13
fix: display deprecation warning for simp lemmas on the lemma name

### LINKED ISSUE #4452: `deprecated` produces its message on `simp`
### Prerequisites

### Description

If a theorem is deprecated and used in `simp` or `dsimp`, the deprecation warning is attached to `simp`, rather than to the declaration itself.

Note: the warning *message* is correct, it is simply logged at the incorrect syntax node.

### Context

This arose in the context of automatic replacement of deprecated declarations.  See also [this Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.60deprecate.60.20at.20the.20correct.20node.20in.20.60simp.60).

### Steps to Reproduce

```lean
@[deprecated]
theorem hi : True := .intro

example : True := by
  -- the warning is on `simp`
  simp [
    hi  -- could the warning be logged at the syntax node of `hi`?
  ]
```
**Expected behavior:** The warning is on `hi`, not on `simp`

**Actual behavior:** The warning is on `simp`, not on `hi`.

### Versions

`#eval Lean.versionString`: "4.9.0-rc1"
OS version: live.lean-lang.org, Linux

### Additional Information

#4448 is already a PR to fix this issue.

### Impact

joneugster:
Closes #4452

### DIFF
--- a/src/Lean/Elab/Tactic/Simp.lean
+++ b/src/Lean/Elab/Tactic/Simp.lean
@@ -204,7 +204,7 @@ def elabSimpArgs (stx : Syntax) (ctx : Simp.Context) (simprocs : Simp.SimprocsAr
                 arg[0][0].getKind == ``Parser.Tactic.simpPost
             let inv  := !arg[1].isNone
             let term := arg[2]
-            match (← resolveSimpIdTheorem? term) with
+            match (← withRef term do resolveSimpIdTheorem? term) with
             | .expr e  =>
               let name ← mkFreshId
               thms ← addDeclToUnfoldOrTheorem thms (.stx name arg) e post inv kind
--- a/src/Lean/Elab/Term.lean
+++ b/src/Lean/Elab/Term.lean
@@ -1769,7 +1769,7 @@ private def mkConsts (candidates : List (Name × List String)) (explicitLevels :
     return (const, projs) :: result
 
 def resolveName (stx : Syntax) (n : Name) (preresolved : List Syntax.Preresolved) (explicitLevels : List Level) (expectedType? : Option Expr := none) : TermElabM (List (Expr × List String)) := do
-  withRef stx <| addCompletionInfo <| CompletionInfo.id stx stx.getId (danglingDot := false) (← getLCtx) expectedType?
+  addCompletionInfo <| CompletionInfo.id stx stx.getId (danglingDot := false) (← getLCtx) expectedType?
   if let some (e, projs) ← resolveLocalName n then
     unless explicitLevels.isEmpty do
       throwError "invalid use of explicit universe parameters, '{e}' is a local"


src/Lean/Elab/Tactic/Simp.lean
```diff
@@ -204,7 +204,7 @@ def elabSimpArgs (stx : Syntax) (ctx : Simp.Context) (simprocs : Simp.SimprocsAr
                 arg[0][0].getKind == ``Parser.Tactic.simpPost
             let inv  := !arg[1].isNone
             let term := arg[2]
-            match (← resolveSimpIdTheorem? term) with
+            match (← withRef arg do resolveSimpIdTheorem? term) with
```
tydeu, line 207:
```suggestion
            match (← withRef term do resolveSimpIdTheorem? term) with
```

It seems more reasonable to use the `term` itself rather than the whole `arg` (which  potentially includes the inversion arrow).

adomani:
In the [Zulip chat](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.60deprecate.60.20at.20the.20correct.20node.20in.20.60simp.60) it was suggested to add a test.  Would this (updated to the correct message) work?
```
import Lean.Elab.Command

@[deprecated]
theorem hi : True := .intro

open Lean Elab.Command
elab "msgs_lines " cmd:command : command => do
  elabCommand cmd
  let msgs := (← get).messages.unreported.toList
  if let [m] := msgs then logInfo m!"{(m.pos, m.endPos)}"

/--
warning: `hi` has been deprecated
---
info: (⟨20, 2⟩, some (⟨22, 3⟩))
-/
#guard_msgs in
msgs_lines
example : True := by
  simp [
    hi
  ]
```

semorrison:
> In the [Zulip chat](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.60deprecate.60.20at.20the.20correct.20node.20in.20.60simp.60) it was suggested to add a test. Would this (updated to the correct message) work?


Haha, I like it. The traditional way is to do it like the tests in `tests/lean/interactive/`.

joneugster:
@semorrison I managed to get the tests correctly :) I don't think I can remove the `awaiting-author` label, so I assume the right thing is to ping you instead?

src/Lean/Elab/Tactic/Simp.lean
joneugster: Overlooked the second suggestion, but just addressed that now, see below.

src/Lean/Elab/Term.lean
```diff
@@ -1769,7 +1769,7 @@ private def mkConsts (candidates : List (Name × List String)) (explicitLevels :
     return (const, projs) :: result
 
 def resolveName (stx : Syntax) (n : Name) (preresolved : List Syntax.Preresolved) (explicitLevels : List Level) (expectedType? : Option Expr := none) : TermElabM (List (Expr × List String)) := do
-  addCompletionInfo <| CompletionInfo.id stx stx.getId (danglingDot := false) (← getLCtx) expectedType?
+  withRef stx <| addCompletionInfo <| CompletionInfo.id stx stx.getId (danglingDot := false) (← getLCtx) expectedType?
```
joneugster, line 1772: I think this is the right place to add `withRef`:


We need it for `mkConsts` used here  in `resolveName`, `mkConsts` but and which doesn't take syntax anymore.

nomeata:
It seems that this was not fixed by #4484, please re-open if I am confused. Thanks for helping, though!

joneugster:
Absolutely, thanks for the fix!

adomani:
I also tried the web-server and I can still see the old `simp`-underlining.  Would you mind also adding `dsimp` to the test, just to make sure that the same modifications fix both?  Thanks!

joneugster:
> I also tried the web-server and I can still see the old `simp`-underlining.

Did you try the "Nightly" version of the webeditor? (go to settings and change the project) It seems to be as intended:
[MWE on web editor nightly](https://live.lean-lang.org/#project=lean-nightly&code=import%20Lean.Elab.Command%0A%0A%40%5Bdeprecated%5D%0Atheorem%20hi%20%3A%20True%20%3A%3D%20.intro%0A%0Aopen%20Lean%20Elab.Command%0Aelab%20%22msgs_lines%20%22%20cmd%3Acommand%20%3A%20command%20%3D%3E%20do%0A%20%20elabCommand%20cmd%0A%20%20let%20msgs%20%3A%3D%20(%E2%86%90%20get).messages.unreported.toList%0A%20%20if%20let%20%5Bm%5D%20%3A%3D%20msgs%20then%20logInfo%20m!%22%7B(m.pos%2C%20m.endPos)%7D%22%0A%0A%2F--%0Awarning%3A%20%60hi%60%20has%20been%20deprecated%0A---%0Ainfo%3A%20(%E2%9F%A820%2C%202%E2%9F%A9%2C%20some%20(%E2%9F%A822%2C%203%E2%9F%A9))%0A-%2F%0A%23guard_msgs%20in%0Amsgs_lines%0Aexample%20%3A%20True%20%3A%3D%20by%0A%20%20simp%20%5B%0A%20%20%20%20hi%0A%20%20%5D)

adomani:
Ah, no, I did not know that this was possible!  I can confirm that on the "nightly", the issue that I had raised is resolved, both for `simp` and for `dsimp`!

I am not really able to comment whether this is the "right" solution, but I would be very happy if this behaviour became standard!  Thanks!

adomani:
Oh, I realize now that you are saying that this *already is* the behaviour, it just has not trickled down to Mathlib!  Anyway, this is great, thanks!

joneugster:
exactly, I assume that means it'll be in `v4.10.0` and the corresponding `rc`s

PR #4459, opened 2024-06-14
feat: incremental `next` and tactic `if`

Kha:


src/Init/TacticsExtra.lean
```diff
@@ -31,15 +31,19 @@ private def expandIfThenElse
       pure (hole, #[case])
   let (posHole, posCase) ← mkCase thenTk pos `(?pos)
   let (negHole, negCase) ← mkCase elseTk neg `(?neg)
-  `(tactic| (open Classical in refine%$ifTk $(← mkIf posHole negHole); $[$(posCase ++ negCase)]*))
+  `(tactic| ((open Classical in refine%$ifTk $(← mkIf posHole negHole)); $[$(posCase ++ negCase)]*))
```
Kha, line 34: @leodemoura Without the parentheses, this used to open Classical in the case branches as well, which seems unintended. Do you agree? I can add a test case for that if so.

nomeata: Maybe add a test case also if it is intended :-)

PR #4464, opened 2024-06-16
chore: use ' instead of brackets for quotes

JovanGerb:
Bracket used for quoting an expression, such as (C ...) in
```
invalid {...} notation, {kind} type is not of the form (C ...){indentExpr type}
```
have been replaced by 'C ...'. This makes the messages clearer.

### DIFF
--- a/src/Lean/Attributes.lean
+++ b/src/Lean/Attributes.lean
@@ -332,7 +332,7 @@ unsafe def mkAttributeImplOfConstantUnsafe (env : Environment) (opts : Options) 
   | some info =>
     match info.type with
     | Expr.const `Lean.AttributeImpl _ => env.evalConst AttributeImpl opts declName
-    | _ => throw ("unexpected attribute implementation type at '" ++ toString declName ++ "' (`AttributeImpl` expected")
+    | _ => throw ("unexpected attribute implementation type at '" ++ toString declName ++ "' `AttributeImpl` expected")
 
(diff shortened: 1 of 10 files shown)


nomeata (commented):
It's good to be consistent, but probably subjective which variant is clearer. I'd be curious what @david-christiansen will say.

Also, do we maybe use back ticks in some places as well?

JovanGerb:
Yeah, I wasn't sure about backticks vs single quotes. I went for normal quotes because they seem to be used more than backticks. I don't know of any difference in the meaning between the two.

JovanGerb:
Here both quotes are even used in the same error:
https://github.com/leanprover/lean4/blob/0448e3f4ea1b2c476fc79aeee5d407971a1cd9e2/src/lake/Lake/Load/Package.lean#L31

nomeata:
My impression, from looking at the lines of code in the diffs, is that single ticks are used to refer to a _names_ and parentheses or backticks to quote _expressions_. Such semantic differentiation might actually be intended and useful, but then should be documented for the other developers so that we can apply them consistently.

leodemoura:
> It's good to be consistent, but probably subjective which variant is clearer. I'd be curious what @david-christiansen will say.
> 
> Also, do we maybe use back ticks in some places as well?

Me too. I feel like we are not using quotes and backticks consistently in error messages and doc-strings. I am also curious about what @david-christiansen will say. We should document the decision, and starting enforcing from now on.

PR #4467, opened 2024-06-17
fix: `dsimp` missing theorems for literals

leodemoura:


leodemoura:
@semorrison @jcommelin It looks like a minor and expected breakage.

semorrison:
Yes, and now fixed on `lean-pr-testing-4467`, as strict improvements!

PR #4471, opened 2024-06-17
feat: Nat lemmas about `+1`

kim-em:


semorrison:
Unfortunately we can't make the simproc for explico

PR #4472, opened 2024-06-17
feat: lemmas about List.filter

kim-em:
This upstreams some lemmas from Batteries, giving dumbed-down proofs, as I do not (yet?) want to move up `List.Sublist`.

### DIFF
--- a/src/Init/Data/List/Lemmas.lean
+++ b/src/Init/Data/List/Lemmas.lean
@@ -799,6 +799,18 @@ theorem length_filterMap_le (f : α → Option β) (l : List α) :
   apply length_filter_le
 
 @[simp]
+theorem filterMap_length_eq_length {l} :
+    (filterMap f l).length = l.length ↔ ∀ a ∈ l, (f a).isSome := by
+  induction l with
+  | nil => simp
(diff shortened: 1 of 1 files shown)


src/Init/Data/List/Lemmas.lean
```diff
@@ -710,6 +710,35 @@ theorem filter_cons :
...
+  | nil => simp
+  | cons a l ih =>
+    simp only [filter_cons, length_cons, succ_eq_add_one]
+    split
+    · simp only [length_cons, succ_eq_add_one]
+      exact Nat.succ_le_succ ih
+    · exact Nat.le_trans ih (Nat.le_add_right _ _)
+
+@[simp]
+theorem filter_eq_self {l} : filter p l = l ↔ ∀ a ∈ l, p a := by
```
nomeata, line 725: It's probably there because it was there in batteries, and that's fine, but I'm curious if you'd consider this a good default simp lemma if it was proposed new?


It seems a rather large reasoning step (non-linear pattern, rewriting the goal in very different terms), so I'm not sure and am curious about your design thoughts here.

PR #4477, opened 2024-06-17
feat: expose flags for the bundled C compiler

tydeu:
Expose the C compiler and linker flags used with the bundled compiler (clang) to Lean code. This is needed to skip the use of `leanc` in Lake.


tydeu:
@Kha This is the first part of what discussed regarding remove the `leanc` dependency from Lake.

Kha (approved):

PR #4478, opened 2024-06-17
fix: lake: remove module dynlib from platform-independent trace

tydeu:
Fixes a bug where Lake incorrectly included the module dynlib in a platform-independent trace. It was incorrectly excluded only external native libraries from the trace. Also adds a test.


tydeu:
!bench

PR #4480, opened 2024-06-17
fix: and improve `#reduce` commmand

### LINKED ISSUE #4465: Reducing Char.ofNat exceeds the recursion depth
### Prerequisites

* [ ] Check that your issue is not already filed:

### Description

`#reduce Char.ofNat (nat_lit 0)` fails with `maximum recursion depth has been reached`

### Steps to Reproduce

1. Run the above

**Expected behavior:** Does not crash

**Actual behavior:** Crashes

### Versions

[Output of `#eval Lean.versionString`]
[OS version, if not using live.lean-lang.org.]

### Additional Information

### Impact

leodemoura:
closes #4465

- `#reduce` does not reduce proofs and types by default anymore. 
- `#reduce (types := true)` can be used to reduce types.
- `#reduce (proof := true)` can be used to reduce proofs.
- Commit also adds better doc-string for `#reduce`.

### DIFF
--- a/src/Init/Notation.lean
+++ b/src/Init/Notation.lean
@@ -559,10 +559,20 @@ set_option linter.missingDocs false in
 syntax guardMsgsFilterSeverity := &"info" <|> &"warning" <|> &"error" <|> &"all"
 
 /--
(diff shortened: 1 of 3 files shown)


eric-wieser:
Thanks!

PR #4482, opened 2024-06-17
fix: discrepancy in the elaborators for `theorem`, `def`, and `example`

### LINKED ISSUE #4398: simp doesn't rewrite because of unassigned metavariables
### Prerequisites

### Description

The following code fails:
```lean
set_option trace.Meta.Tactic.simp true
example {ι : Type _} (n m : ι) (h : n = m) : m = n := by
  simp only [id h] -- no progress
  simp only [h] -- works
```
The trace message is
```
[Meta.Tactic.simp.rewrite] id h:1000, has unassigned metavariables after unification 
```
Note that the unassigned metavariable is the universe metavariable of `ι`

**Expected behavior:** `simp only [id h]` should succeed, just like `simp only [h]` already does.

### Impact

leodemoura:
When the type of a definition or example is a proposition,
we should elaborate on them as we elaborate on theorems.
This is particularly important for examples that are often
used in educational material.

Recall that when elaborating theorem headers, we convert unassigned
universe metavariables into universe parameters. The motivation is
that the proof of a theorem should not influence its statement.
However, before this commit, this was not the case for definitions and
examples when their type was a proposition. This discrepancy often confused users.

Additionally, we considered extending the above behavior whenever
the type of a definition is provided. That is, we would keep the
current behavior only if `: <type>` was omitted in a definition.
However, this proved to be too restrictive.
For example, the following instance in `Core.lean` would fail:
```
instance {α : Sort u} [Setoid α] : HasEquiv α :=
  ⟨Setoid.r⟩
```
and we would have to write instead:
```
instance {α : Sort u} [Setoid α] : HasEquiv.{u, 0} α :=
  ⟨Setoid.r⟩
```
There are other failures like this in the core, and we assume many more in Mathlib.

closes #4398

@semorrison @jcommelin: what do you think?

### DIFF
--- a/tests/lean/autoPPExplicit.lean.expected.out
+++ b/tests/lean/autoPPExplicit.lean.expected.out
@@ -5,4 +5,4 @@ argument
 has type
   Prop : Type
 but is expected to have type
-  α : Sort ?u
+  α : Sort u_1


semorrison:
This would be great. I'll take a look at Mathlib.

semorrison:
A typical example of an unhelpful error message is:
```
example (X : TopCat) (F : Presheaf CommRingCat X)
    (h : Presheaf.IsSheaf (F ⋙ (forget CommRingCat))) :
    F.IsSheaf :=
(isSheaf_iff_isSheaf_comp (forget CommRingCat) F).mpr h
```
which fails with
```
type mismatch
  (isSheaf_iff_isSheaf_comp (forget CommRingCat) ?m.3480).mpr ?m.3884
has type
  ?m.3480.IsSheaf : Prop
but is expected to have type
  F.IsSheaf : Prop
```
giving no clue that the fix is:
```
example (X : TopCat.{u₁}) (F : Presheaf CommRingCat.{u₁} X)
    (h : Presheaf.IsSheaf (F ⋙ (forget CommRingCat))) :
    F.IsSheaf :=
(isSheaf_iff_isSheaf_comp (forget CommRingCat) F).mpr h
```

semorrison:
It turns out that that class of inscrutable error was actually very rare in Mathlib.

Instead I found *many* instances where authors had implicitly constrained universes via the RHS of the `:=` in ways that were confusing and impossible to see from the type signature. I'm going to backport all the Mathlib fixes to master regardless of whether/when we merge this, but +1 from me!

PR #4488, opened 2024-06-18
chore: @[simp] List.getElem?_eq_getElem

kim-em:
This is often helpful, and Mathlib doesn't mind.


semorrison:
Mathlib is remarkably unconcerned. I needed to change a few proof of deprecated (or perhaps-should-be-deprecated) theorems, but otherwise it is fine.

PR #4490, opened 2024-06-18
feat: more infrastructure for tactic documentation

david-christiansen:
This is the groundwork for a tactic index in generated documentation, as there was in Lean 3. There are a few challenges to getting this to work well in Lean 4:
 * There's no natural notion of *tactic identity* - a tactic may be specified by multiple syntax rules (e.g. the pattern-matching version of `intro` is specified apart from the default version, but both are the same from a user perspective)
 * There's no natural notion of *tactic name* - here, we take the pragmatic choice of using the first keyword atom in the tactic's syntax specification, but this may need to be overridable someday.
 * Tactics are extensible, but we don't want to allow arbitrary imports to clobber existing tactic docstrings, which could become unpredictable in practice.

For tactic identity, this PR introduces the notion of a *tactic alternative*, which is a `syntax` specification that is really "the same as" an existing tactic, but needs to be separate for technical reasons. This provides a notion of tactic identity, which we can use as the basis of a tactic index in generated documentation. Alternative forms of tactics are specified using a new `@[tactic_alt IDENT]` attribute, applied to the new tactic syntax. It is an error to declare a tactic syntax rule to be an alternative of another one that is itself an alternative. Documentation hovers now take alternatives into account, and display the docs for the canonical name.

*Tactic tags*, created with the `register_tactic_tag` command, specify tags that may be applied to tactics. This is intended to be used by doc-gen and Verso. Tags may be applied using the `@[tactic_tag TAG1 TAG2 ...]` attribute on a canonical tactic parser, which may be used in any module to facilitate downstream projects introducing tags that apply to pre-existing tactics. Tags may not be removed, but it's fine to redundantly add them. The collection of tags, and the tactics to which they're applied, can be seen using the `#print tactic tags` command.

*Extension documentation* provides a structured way to document extensions to tactics. The resulting documentation is gathered into a bulleted list at the bottom of the tactic's docstring. Extensions are added using the `tactic_extension TAC` command. This can be used when adding new interpretations of a tactic via `macro_rules`, when extending some table or search index used by the tactic, or in any other way. It is a command to facilitate its flexible use with various extension mechanisms.


src/Lean/Parser/Tactic/Doc.lean
```diff
@@ -0,0 +1,217 @@
...
+def allTags [Monad m] [MonadEnv m] : m (List Name) := do
+  let env ← getEnv
+  let mut found : NameSet := {}
+  for (tag, _) in knownTacticTagExt.getState env do
+    found := found.insert tag
+  for arr in tacticAliasExt.toEnvExtension.getState env |>.importedEntries do
+    for (tag, _) in arr do
+      found := found.insert tag
+  pure (found.toArray.qsort (·.toString < ·.toString) |>.toList)
+
```
robertylewis, line 129:
```suggestion

elab "#print_tactic_tags" : command => do
  logInfo <| "\n".intercalate <| (← allTags).map toString

```

This, or some variant, will be a necessity for anyone adding docs -- gotta see what tags are available!

robertylewis (commented):
My hotel wifi died and I don't want to get a core dev setup going on phone data, so I'm not 100% sure these code suggestions are right. But this looks fantastic, should do the job very nicely!

src/Lean/Parser/Tactic/Doc.lean
```diff
@@ -0,0 +1,217 @@
...
+                if count == 0 then m!"(no tags defined)"
+                else if count == 1 then
+                  m!"(expected {MessageData.joinSep (all.map name) ", "})"
+                else if count < 10 then
+                  m!"(expected one of {MessageData.joinSep (all.map name) ", "})"
+                else
+                  m!"(expected one of {
+                    MessageData.joinSep (all.take 10 |>.map name) ", " ++ ", ..."
+                  })"
+              m!"(expected one of {suggestions})"
```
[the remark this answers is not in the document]
david-christiansen: Good catch, thanks!

src/Lean/Parser/Tactic/Doc.lean
david-christiansen: I added a fancier version that also shows user-facing names, docstrings for tags, and the tactics that are included in them (under `#print tactic tags`). Thanks!

david-christiansen:
@robertylewis I think this should meet your needs now - please give it a whirl at let me know! I don't want to merge before you've had a chance to validate that it supports the workflow that you're looking for.

robertylewis (approved):
I've tested the doc workflow on the mathlib branch and it works perfectly!

As written on Zulip: one change from the Lean 3 infrastructure is that this doesn't inherently distinguish between tactics, commands, and attributes. But this can be simulated with tags and the frontend can mimic the Lean 3 behavior.

david-christiansen:
Also as written on Zulip: I think we should use this only for tactics for now, and avoid overdesigning too early. A tactic index is more important than an attribute index. Then we can see whether this should be generalized to a mechanism that works for all those (and is further user-extensible) or whether we should just create the ones we need - I'd like to work through a few examples of each category to see how much they overlap in practice.

david-christiansen:
One further question: the term "alias" seems a bit poorly chosen. Perhaps "alternate" would make more sense?

robertylewis:
Or "alternative"?

david-christiansen:
I like that, but I'll use `alt` to make the attribute name a bit shorter (`@[tactic_alt theRealTactic]`).

david-christiansen:
!bench

david-christiansen:
!bench

david-christiansen:
@Kha - what do you think? Is that performance hit too severe?

src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -85,7 +89,11 @@ def handleHover (p : HoverParams)
       let stxDoc? ← match stack? with
         | some stack => stack.findSomeM? fun (stx, _) => do
           let .node _ kind _ := stx | pure none
-          return (← findDocString? snap.env kind).map (·, stx.getRange?.get!)
+          -- If the tactic is an alternative form, get the docs for the canonical version
+          let kind := alternativeOfTactic snap.env kind |>.getD kind
+          let exts := getTacticExtensionString snap.env kind
+          let docStr := (← findDocString? snap.env kind).map (· ++ exts)
+          return docStr.map (·, stx.getRange?.get!)
```
Kha, line 96: Wouldn't it be easier to copy the docstring to the alternative, like `inherit_doc` does?

david-christiansen: Because the set of extensions is open-ended, we'd need to also either do the same work of resolving alternatives to get them, or we'd need to copy them over both when extensions are added and when alternatives are added, which strikes me as a trickier invariant to maintain.

Is the concern here that the table lookup will cause latency issues?

Kha: Ah, it makes sense for extensions of course. I was mostly just worried about adding such specific code to two places in the server. Can we factor it out, perhaps even put it directly below `findDocString?`?

david-christiansen: That in particular leads to circular module imports, unfortunately, but factoring it out is a good idea.

I'll go make a module that can be downstream of things that are themselves downstream of the docstring environment extension, and upstream of the server, to put this into.

PR #4493, opened 2024-06-18
fix: discrepancy `theorem` vs `example` 

### LINKED ISSUE #4398: simp doesn't rewrite because of unassigned metavariables

leodemoura:
When the type of an `example` is a proposition,
we should elaborate on them as we elaborate on theorems.
This is particularly important for examples that are often
used in educational material.

Recall that when elaborating theorem headers, we convert unassigned
universe metavariables into universe parameters. The motivation is
that the proof of a theorem should not influence its statement.
However, before this commit, this was not the case for examples when their type was a proposition.
This discrepancy often confused users.

Additionally, we considered extending the above behavior to definitions when 
1- When their type is a proposition. However, it still caused disruption in Mathlib. 
2- When their type is provided. That is, we would keep the current behavior only if `: <type>` was omitted. This would make the elaborator for `def` much closer to the one for `theorem`, but it proved to be too restrictive.
For example, the following instance in `Core.lean` would fail:
```
instance {α : Sort u} [Setoid α] : HasEquiv α :=
  ⟨Setoid.r⟩
```
and we would have to write instead:
```
instance {α : Sort u} [Setoid α] : HasEquiv.{u, 0} α :=
  ⟨Setoid.r⟩
```
There are other failures like this in the core, and we assume many more in Mathlib.

closes #4398
closes #4482 Remark: PR #4482 implements option 1 above. We may consider it again in the future.

### DIFF
--- a/src/Lean/Elab/MutualDef.lean
+++ b/src/Lean/Elab/MutualDef.lean
@@ -846,7 +846,10 @@ private def levelMVarToParamHeaders (views : Array DefView) (headers : Array Def
   let rec process : StateRefT Nat TermElabM (Array DefViewElabHeader) := do
     let mut newHeaders := #[]
     for view in views, header in headers do
-      if ← pure view.kind.isTheorem <||> isProp header.type then
+      -- Remark: we should consider using `pure view.kind.isTheorem <||> isProp header.type`, and
+      -- also handle definitions. We used the following approach because it is less disruptive to Mathlib.
+      -- Moreover, the type of most definitions are not propositions anyway.
+      if ← pure view.kind.isTheorem <||> (pure view.kind.isExample <&&> isProp header.type) then
         newHeaders ←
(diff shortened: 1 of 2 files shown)


semorrison:
Mathlib should be working now (it was an `example` which really was "wrong": from the statement it looks like it was more universe polymorphic than the proof actually allowed).

PR #4501, opened 2024-06-19
doc: more detailed docstring for PersistentEnvExtension

david-christiansen:
Describes the intended modes of use, potential performance tradeoffs, and data representation in more detail.

### DIFF
--- a/src/Lean/Environment.lean
+++ b/src/Lean/Environment.lean
@@ -430,7 +430,7 @@ An environment extension with support for storing/retrieving entries from a .ole
  - σ is the actual state.
 
 For most extensions, α and β coincide. `α` and ‵β` do not coincide for extensions where the data
-used to update the state contains elements which cannot be stored in files (for example, closures) .
+used to update the state contains elements which cannot be stored in files (for example, closures).
 
 During elaboration of a module, state of type `σ` can be both read and written. When elaboration is
 complete, the state of type `σ` is converted to serialized state of type `Array α` by


src/Lean/Environment.lean
```diff
@@ -423,22 +423,50 @@ structure ImportM.Context where
...
+The most typical pattern for using `PersistentEnvExtension` is to set `σ` to a datatype such as
+`NameMap` that efficiently tracks data for the current module. Then, in `exportEntriesFn`, this type
+is converted to an array of pairs, sorted by the key. Given `ext : PersistentEnvExtension α β σ` and
+`env : Environment`, the complete array imported entries sorted by module index can be obtained
+using `(ext.toEnvExtension.getState env).importedEntries`. To query the extension for some name `n`,
+first use `env.getModuleIdxFor? n`. If it returns `none`, look up `n` in the current module's state
+(the `NameMap`). If it returns `some idx`, use `ext.getModuleEntries env idx` to get an array, and
+query it using `Array.binSearch`. This imposes a constraint that the extension can only track
+metadata declared in the same module as the definition to which it applies; relaxing this
+restriction can make queries slower due to needing to search _all_ modules.
```
Kha, line 456:
```suggestion
query it using `Array.binSearch`. This imposes a constraint that the extension can only track
metadata declared in the same module as the definition to which it applies; relaxing this
restriction can make queries slower due to needing to search _all_ modules, in which case
it is usually better to initialize the state of type `σ` once from all imported entries and choose
a more efficient search datastructure for it.
```
The suggestions on loading vs. not loading entries are a bit scattered between the two paragraphs (I guess I made it worse).

Kha (changes requested):

src/Lean/Environment.lean
david-christiansen: I actually don't know if this is true - it will presumably depend on the expected number of entries and the expected frequency of queries. For infrequently-queried data that isn't so large, it might be better to avoid the overhead of building the search structure on every module load, and just do N binary searches, right?

Kha: Sure, it could go either way. Which often is hard to estimate as the metaprogram author. Let's just list both alternatives.

src/Lean/Environment.lean
```diff
@@ -423,22 +423,51 @@ structure ImportM.Context where
...
-   `α` and ‵β` do not coincide for extensions where the data used to update the state contains, for example,
-   closures which we currently cannot store in files. -/
+/--
+An environment extension with support for storing/retrieving entries from a .olean file.
+ - α is the type of the entries that are stored in .olean files.
+ - β is the type of values used to update the state.
+ - σ is the actual state.
+
+For most extensions, α and β coincide. `α` and ‵β` do not coincide for extensions where the data
+used to update the state contains, for example, closures which we currently cannot store in files.
```
tydeu, line 433:
```suggestion
used to update the state contains elements which cannot be stored in files (for example, closures) .
```

The "for example" here feels like it breaks up the flow, and this seems like a good time to suggest a change.

david-christiansen: Thanks!

src/Lean/Environment.lean
```diff
@@ -423,22 +423,51 @@ structure ImportM.Context where
...
-   `α` and ‵β` do not coincide for extensions where the data used to update the state contains, for example,
-   closures which we currently cannot store in files. -/
+/--
+An environment extension with support for storing/retrieving entries from a .olean file.
+ - α is the type of the entries that are stored in .olean files.
+ - β is the type of values used to update the state.
+ - σ is the actual state.
+
+For most extensions, α and β coincide. `α` and ‵β` do not coincide for extensions where the data
+used to update the state contains elements which cannot be stored in files (for example, closures) .
```
tydeu, line 433:
```suggestion
used to update the state contains elements which cannot be stored in files (for example, closures).
```

Oops, stray space typo on my part

Kha (approved):

david-christiansen:
Thanks for the helpful feedback, @Kha and @tydeu :-)

PR #4502, opened 2024-06-19
chore: fix linter errors

JovanGerb:
The linters in Batteries can be used to spot mistakes in Lean. See the message on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Go-to-def.20on.20typeclass.20fields.20and.20type-dependent.20notation/near/442613564). These are the different linters with errors:

- unusedArguments:
  There are many unused instance arguments, especially a redundant `[Monad m]` is very common
- checkUnivs:
  There was a problem with universes in a definition in `Init.Control.StateCps`. I fixed it by adding a `variable` statement for the implicit arguments in the file.
- defLemma:
  many proofs are written as `def` instead of `theorem`, most notably `rfl`. Because `rfl` is used as a match pattern, it must be a def. Is this desirable?
  The keyword `abbrev` is sometimes used for an alias of a theorem, which also results in a def. I would want to replace it with the `alias` keyword to fix this, but it isn't available.
- dupNamespace:
  I fixed some of these, but left `Tactic.Tactic` and `Parser.Parser` as they are as these seem intended.
- unusedHaveSuffices:
  I cleaned up a few proofs with unused `have` or `suffices`
- explicitVarsOfIff:
  I didn't fix any of these, because that would be a breaking change.
- simpNF:
  I didn't fix any of these, because I think that requires knowing the intended simplification order.

### DIFF
--- a/tests/lean/run/meta3.lean
+++ b/tests/lean/run/meta3.lean
@@ -64,7 +64,7 @@ info: [Meta.debug] (Add.add => (node
     (* => (node #[5]))
     (Nat.add => (node (0 => (node (20 => (node #[3]))))))
 [Meta.debug] #[5, 1]
-[Meta.debug] Add.add ?m.4906 ?m.4906
+[Meta.debug] Add.add ?m.4899 ?m.4899
 [Meta.debug] #[5]
 [Meta.debug] #[5, 1, 4, 2]
 [Meta.debug] #[1, 4, 2, 5, 3]


src/Init/WF.lean
```diff
@@ -268,7 +268,7 @@ section
 variable {α : Type u} {β : Type v}
 variable {ra  : α → α → Prop} {rb  : β → β → Prop}
 
-def lexAccessible {a : α} (aca : Acc ra a) (acb : (b : β) → Acc rb b) (b : β) : Acc (Prod.Lex ra rb) (a, b) := by
```
eric-wieser, line 271: Isn't there some reduction-related reason that `Acc` results should be `def`s`?

digama0: Yes. Several definitions here need to reduce so that well founded recursive definitions will compute. (I discovered these are load-bearing when working on lean4lean, it is reduced inside lean core itself.)

eric-wieser: So does this change need to be reverted, or are these not such definitions?

JovanGerb: None of the definitions marked as `@[reducible]` were modified, so it seems fine to me

PR #4504, opened 2024-06-19
chore: Remove simp from `Option.elim`, replace with individal simp lemmas

BoltonBailey:
This PR removes the `simp` attribute from `Option.elim` and adds it to two related simp lemmas, `Option.elim_none` and `Option.elim_some`.

This PR comes from some discussion [here](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/optionEquivLeft_apply.20simp/near/438321459) about `simps!` feeling too aggressive in unfolding this lemma.

### DIFF
--- a/src/Init/Data/Option/Basic.lean
+++ b/src/Init/Data/Option/Basic.lean
@@ -123,9 +123,6 @@ def merge (fn : α → α → α) : Option α → Option α → Option α
   | some x, _, f => f x
   | none, y, _ => y
 
-@[simp] theorem elim_none (b) (f : α → β) : none.elim b f = b := rfl
-@[simp] theorem elim_some (a) (b) (f : α → β) : (some a).elim b f = f a := rfl
-
 /-- Extracts the value `a` from an option that is known to be `some a` for some `a`. -/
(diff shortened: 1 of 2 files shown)


BoltonBailey:
This rebase command does not seem to affect the PR in any way. I am not really sure what to do to be able to see if and how this breaks mathlib.

leodemoura:
@BoltonBailey Have tried to squash your commits before executing the command? Another option is to use 
```
git reset --hard origin/nightly-with-mathlib
git cherry-pick 1c29870c1aa
git cherry-pick 4554124b873
```

BoltonBailey:
`origin`, which in this case is my fork of this repo, does not have a `nightly-with-mathlib` branch, so I don't think that will work. I could try the "squash"ing option, but it seems like it would require me to learn something called "interactive rebasing".

semorrison:
> `origin`, which in this case is my fork of this repo, does not have a `nightly-with-mathlib` branch, so I don't think that will work. I could try the "squash"ing option, but it seems like it would require me to learn something called "interactive rebasing".

Replace `origin` with the name of the upstream referring to this repository?

PR #4511, opened 2024-06-20
perf: Micro-optimization for (8/16/32/64)-bit-uint log2

qawbecrdtey:
This applies faster code for computing `lean_uintX_log2`, where `X` is 8, 16, 32, or 64.
This is slightly modified version of [the answer from the stack overflow](https://stackoverflow.com/a/11398748).
Quickbench([1](https://quick-bench.com/q/dmluXWuPPFBZTq5T30l8T3er8Eg), [2](https://quick-bench.com/q/aNeXk0AxuTSZ9gmvQIM8Hw0rbV4)) shows at most 3 times improvement in performance, for the case `X = 64`.

There is also a function `lean_usize_log2` in the same file, but I haven't modified it yet due to the specification of `size_t`.

### DIFF
--- a/src/include/lean/lean.h
+++ b/src/include/lean/lean.h
@@ -1694,7 +1694,7 @@ static inline uint16_t lean_uint16_log2(uint16_t a) {
     /* Multiplying the magic number 0x0F2D to `2 ^ n - 1` for `n = 0` to `15`
        maps its significant `4` bits from `0` to `15`.
        `table` satisfies `n` -> `m = 2 ^ n - 1` -> `(m * 0x0F2D) mod 2 ^ 16` -- table -> `n`. */
-    return (16 - (uint16_t)(a * 0x0F2D)) & 0xF;
+    return table[(uint16_t)(a * 0x0F2D) >> 12];
 }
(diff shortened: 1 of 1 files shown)


collares:
Is it clear that a lookup table is faster than `std::bit_width(x) - 1`? LLVM's implementation does bisection to find the smallest y such that `x >> y == 0`.

qawbecrdtey:
> Is it clear that a lookup table is faster than `std::bit_width(x) - 1`? LLVM's implementation does bisection to find the smallest y such that `x >> y == 0`.

`std::bitwidth` requires c++20, and for some reason it seemed like I could only use the features from at most C99. I'm not sure that bisection method will be faster, and that needs benchmarking.

collares:
Sorry, I misspoke! It uses `__builtin_clz` and `__builtin_clzll` for uint32 and uint64.

qawbecrdtey:
> Sorry, I misspoke! It uses `__builtin_clz` and `__builtin_clzll` for uint32 and uint64 (have to be careful about undefined behavior when `x == 0`)

Are those available in all compiler? If that's the case your suggestion should be better, but if not then it will increase the line of code drastically..

Kha:
I'm closing this as we can't maintain such magic code. The specific motivation is also unclear.

PR #4512, opened 2024-06-20
fix: missing unboxing in interpreter when loading initialized value

### LINKED ISSUE #4457: `initialize` corrupts scalar values
### Prerequisites

### Description

`initialize` always stores booleans as `true`

### Steps to Reproduce

1. Add `initialize definitelyFalse : Bool ← pure false` to `InitializeBug.lean`
2. `#eval definitelyFalse` in a file that starts `import InitializeBug`, and open it in VSCode

**Expected behavior:** Output is definitely `false`

**Actual behavior:** Output is `true`

### Versions

4.9.0-rc1

### Impact

Kha:
Fixes #4457


collares:
Minor remark: This is the same as (the main part of) #2236.

PR #4514, opened 2024-06-20
fix: naming convention for `UInt` lemmas

### LINKED ISSUE #4513: Some uint lemmas do not follow correct naming conventions
### Prerequisites

### Description

The `UInt` lemmas `zero_toNat`, `div_toNat`, `mod_toNat`, `modn_toNat` reverse the normal convention where `toNat` should come first.

### Impact

fgdorais:
Closes #4513

### DIFF
--- a/src/Std/Data/DHashMap/Internal/Index.lean
+++ b/src/Std/Data/DHashMap/Internal/Index.lean
@@ -48,7 +48,7 @@ cf. https://github.com/leanprover/lean4/issues/4157
   ⟨(scrambleHash hash).toUSize &&& (sz.toUSize - 1), by
     -- Making this proof significantly less painful will be a good test for our USize API
     by_cases h' : sz < USize.size
-    · rw [USize.and_toNat, ← USize.ofNat_one, USize.sub_toNat_of_le, USize.toNat_ofNat_of_lt]
+    · rw [USize.and_toNat, ← USize.ofNat_one, USize.toNat_sub_of_le, USize.toNat_ofNat_of_lt]
       · refine Nat.lt_of_le_of_lt Nat.and_le_right (Nat.sub_lt h ?_)
         rw [USize.toNat_ofNat_of_lt]
         · exact Nat.one_pos


src/Init/Data/UInt/Lemmas.lean
```diff
@@ -45,10 +45,10 @@ protected theorem ne_of_val_ne {a b : $typeName} (h : a.val ≠ b.val) : a ≠ b
 open $typeName (ne_of_val_ne) in
 protected theorem ne_of_lt {a b : $typeName} (h : a < b) : a ≠ b := ne_of_val_ne (Fin.ne_of_lt h)
 
-@[simp] protected theorem zero_toNat : (0 : $typeName).toNat = 0 := Nat.zero_mod _
-@[simp] protected theorem mod_toNat (a b : $typeName) : (a % b).toNat = a.toNat % b.toNat := Fin.mod_val ..
-@[simp] protected theorem div_toNat (a b : $typeName) : (a / b).toNat = a.toNat / b.toNat := Fin.div_val ..
-@[simp] protected theorem modn_toNat (a : $typeName) (b : Nat) : (a.modn b).toNat = a.toNat % b := Fin.modn_val ..
+@[simp] protected theorem toNat_zero : (0 : $typeName).toNat = 0 := Nat.zero_mod _
+@[simp] protected theorem toNat_mod (a b : $typeName) : (a % b).toNat = a.toNat % b.toNat := Fin.mod_val ..
```
eric-wieser, line 49: You could argue that `Fin.mod_val` is backwards too, and that this change now makes them inconsistent.

fgdorais: I agree but I think that's another PR since these are actually used (e.g. here).

semorrison:
@fgdorais could you please add `@[deprecated (since := "2024-06-23")] abbrev foo := @bar` for these? Happy to merge after that.

fgdorais:
Deprecation completed. I kept these out of the theorem generating macro since I suppose they should eventually be removed. Let me know if you would rather have the deprecations inside the macro.

PR #4515, opened 2024-06-20
fix: calling programs with spaces on Windows

Kha:
This entire API has been carefully optimized for maximum pain output

### DIFF
--- a/src/runtime/process.cpp
+++ b/src/runtime/process.cpp
@@ -238,7 +238,7 @@ static obj_res spawn(string_ref const & proc_name, array_ref<string_ref> const &
     bool bSuccess = CreateProcess(
         // Passing `program` here should be more robust, but would require adding a `.exe` extension
         // and searching through `PATH` where necessary
-        NULL, //program.c_str(),
+        NULL,
         const_cast<char *>(command.c_str()), // command line
         NULL,                                // process security attributes
         NULL,                                // primary thread security attributes
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -126,6 +126,81 @@ jobs:
             let large = ${{ github.repository == 'leanprover/lean4' }};
             let matrix = [
               {
+                // portable release build: use channel with older glibc (2.27)
+                "name": "Linux LLVM",
+                "os": "ubuntu-latest",
+                "release": false,
+                "check-level": 2,
+                "shell": "nix develop .#oldGlibc -c bash -euxo pipefail {0}",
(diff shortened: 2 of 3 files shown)


src/runtime/process.cpp
```diff
@@ -236,7 +236,9 @@ static obj_res spawn(string_ref const & proc_name, array_ref<string_ref> const &
 
     // Create the child process.
     bool bSuccess = CreateProcess(
-        program.c_str(),
+        // Passing `program` here should be more robust, but would require adding a `.exe` extension
+        // where necessary
+        NULL, //program.c_str(),
```
tydeu, line 241: According to the [docs](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa), this argument also does look in `PATH` (hence even `git.exe` would not work).

Kha: I assume you forgot a negation there

Kha:
Well, *something* broke, but it seems more likely to be a Unicode problem with Lake (I think we're still not setting the code page to UTF-8) or something on the way to it.

tests/pkg/spaces/test.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+rm -rf .lake/build
+lake exe "«path with spaces»"
```
tydeu, line 4:
```suggestion
lake exe "path with spaces"
```

I think the French quotes here are causing the problem. I skip the part of Lake's `init` test that uses French quotes because I have found they are not properly passed to Lake from the CLI.

https://github.com/leanprover/lean4/blob/073b2cfc83dbb385548b5e5851e69d0713a578dc/src/lake/tests/init/test.sh#L124-L128

tydeu: I believe this is an argument encoded issue, as this example program I think demonstrates:

```
$ echo 'def main (as : List String) : IO Unit := IO.println (repr as)' | lean --stdin --run "«path with spaces»"
[]

Comment (Kha) [reply tests/pkg/spaces/test.sh]: I think that's a different `--run` bug, at least on Linux. Could you try
```
echo 'def main (as : List String) : IO Unit := IO.println (repr as)' | lean --stdin --run '«path with spaces»' '«path with spaces»'
```
?

Kha: #4601

Kha:
Well, now the test isn't broken even without escaping. It might be some weird edge case like "`CreateProcess` with `A\B C.exe "an arg"` works, but only if `A\B` does not exist by itself" :shrug: .

Kha:
Hah, successfully failed the test. Now to add back the fix.

> error: unspecified system_category error (error code: 193)

PR #4518, opened 2024-06-20
refactor: lake: more robust trace reading

tydeu:
The recent change of the trace format exposed some unexpected issues with Lake's tracing handling. This aims to fix that. 

Lake will now perform a rebuild if the trace file is invalid/unreadable. However, it will still fall back to modification times if the trace file is missing. Also, Lake is now backwards compatible with the previous pure numeric traces (and tolerates the absence of a `log` field in the JSON trace).

### DIFF
--- a/src/lake/Lake/Build/Common.lean
+++ b/src/lake/Lake/Build/Common.lean
@@ -5,6 +5,7 @@ Authors: Mac Malone
 -/
 import Lake.Config.Monad
 import Lake.Build.Actions
+import Lake.Util.JsonObject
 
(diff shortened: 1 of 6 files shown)


digama0:
@semorrison FYI, this is (IMO) high prio to get into 4.9.0, it fixes a [backward compatibility issue](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/ltar.20error.20aften.20lean.234402/near/445227132) which currently afflicts people updating to 4.9.0-rcX from 4.8.0.

PR #4521, opened 2024-06-21
feat: lemmas about List.map

kim-em:


src/Init/Data/List/Lemmas.lean
```diff
@@ -33,6 +33,17 @@ For each `List` operation, we would like theorems describing the following, when
 
 Of course for any individual operation, not all of these will be relevant or helpful, so some judgement is required.
 
+General principles for `simp` normal forms for `List` operations:
+* Arithmetic operations are "light", so e.g. we prefer to simplify `(L.drop i).drop j` to `L.drop (i + j)`,
```
nomeata, line 37: Should this be `j + i`? Rationale: where would a constructor (`succ`) be for this to reduce. If L is headed by a cons, and `i` is headed by a `succ`, then the LHS can reduce, but the RHS not, unless you swap them around. It looks odd, but that's what we get for defining addition by recursion on the right argument …

src/Init/Data/List/Lemmas.lean
```diff
@@ -33,6 +33,17 @@ For each `List` operation, we would like theorems describing the following, when
...
+  rather than the other way round.
+* Function compositions are "light", so we prefer to simplify `(L.map f).map g` to `L.map (g ∘ f)`.
+* We generally prefer to move "more basic operations inwards", so for example
+  `List.map_reverse : (L.reverse.map f) = (L.map f).reverse` is a `simp` lemma,
+  because `List.map` is a more basic operation than `reverse`
+  (and there is more chance that we "know how to calculate `L.map f`")
+* We try to avoid non-linear left hand sides (i.e. with subexpressions appearing multiple times),
+  but this is only a weak preference.
+
 -/
```
nomeata, line 47: Maybe a bullet point that conversion functions (e.g. `toArray`) move inwards most aggressively, so that if an expression can be converted completely, it should so?

src/Init/Data/List/Lemmas.lean
```diff
@@ -707,33 +741,105 @@ theorem get_map (f : α → β) {l n} :
...
+  cases l
+  · simp
+  · rw [getLast?_eq_getLast, getLast?_eq_getLast, getLast_map] <;> simp
+
+@[simp] theorem getLastD_map (f : α → β) (l : List α) (a : α) : getLastD (map f l) (f a) = f (getLastD l a) := by
+  cases l
+  · simp
+  · rw [getLastD_eq_getLast?, getLastD_eq_getLast?, getLast?_map] <;> simp
+
+@[simp] theorem map_append (f : α → β) : ∀ l₁ l₂, map f (l₁ ++ l₂) = map f l₁ ++ map f l₂ := by
```
nomeata, line 833: This duplicates `f`. I still think this lemma should be simp, but maybe we can give a rationale in the design section above that duplicating higher order argument (mapped functions, predicates) is ok, because they are often small?

src/Init/Data/List/Lemmas.lean
```diff
@@ -1202,6 +1343,16 @@ theorem eq_replicate {a : α} {n} {l : List α} :
...
    fun ⟨e, al⟩ => e ▸ eq_replicate_of_mem al⟩
 
+theorem map_eq_replicate_iff {l : List α} {f : α → β} {b : β} :
+    l.map f = replicate l.length b ↔ ∀ x ∈ l, f x = b := by
+  simp [eq_replicate]
+
+@[simp] theorem map_const (l : List α) (b : β) : map (Function.const α b) l = replicate l.length b :=
+  map_eq_replicate_iff.mpr fun _ _ => rfl
+
+@[simp] theorem map_const' (l : List α) (b : β) : map (fun _ => b) l = replicate l.length b :=
```
nomeata, line 1353: Looks very useful, but I assume this will now be tried for every and each `List.map` application out there, due to the discrtree structure, will it? Should we worry about the cost?

src/Init/Data/List/Lemmas.lean
```diff
@@ -1421,7 +1574,7 @@ theorem take_left : ∀ l₁ l₂ : List α, take (length l₁) (l₁ ++ l₂) =
 theorem take_left' {l₁ l₂ : List α} {n} (h : length l₁ = n) : take n (l₁ ++ l₂) = l₁ := by
   rw [← h]; apply take_left
 
-theorem map_take (f : α → β) :
+@[simp] theorem map_take (f : α → β) :
```
nomeata, line 1577: Is map simpler or take? Hard to say; the RHS applies f to elements that are thrown away. Maybe it doesn't matter much, but maybe the section describing the simp setup could have a list of all functions ordered by how the simp set will sort them?

nomeata (commented):
Great to see the design written down, this will help a lot in the future.

src/Init/Data/List/Lemmas.lean
nomeata: Here is a possible rationale: `map` has a higher chance to simplify with list producer functions, so it's worth moving it in.

In general maybe it's more likely for things to happen further in, so functions with the ”better” simp set, i.e. more rules, like simp, should be moved inside.

src/Init/Data/List/Lemmas.lean
semorrison: Yes, this is terrible, we can't do that! :-)

src/Init/Data/List/Lemmas.lean
semorrison: Yes, that's good. I better review for this problem elsewhere.

(Maybe one could feel less gross here by writing `List.drop i (List.drop j L) = List.drop (i + j) L`...)

semorrison: Oh, and in fact this is actually how the statement is. I just misquoted it in the module doc.

nomeata: I keep confusing myself always with this issue, sorry.

I guess if the proof doesn't need `add_comm` then it's right :-D

src/Init/Data/List/Lemmas.lean
nomeata: But then we also cannot have `map (fun x => x) l = l` either, can we, which would be sad.

Maybe we need to add support for lambdas to the discr tree…

src/Init/Data/List/Lemmas.lean
semorrison: I think this is another example of "prefer moving functions inwards past 'structural' transformations", on the principle that the function will eventually hit the elements of the list and compute. I've updated the docs to reflect this.

src/Init/Data/List/Lemmas.lean
nomeata: In light of this discussion, should Length.list_append have the RHS commuted, so that `a.length` is in the “active” position?

PR #4522, opened 2024-06-21
chore: move @[simp] from pred_le to sub_one_le

kim-em:
(We already have a simp lemma unfolding `pred` to `· - 1`.)

### DIFF
--- a/tests/lean/guessLexTricky.lean
+++ b/tests/lean/guessLexTricky.lean
@@ -54,7 +54,7 @@ decreasing_by
   try -- the need for `try` here is fishy
       -- the proof with explicit `termination_by` does not need it, so it should not throw
       -- GuessLex off, but without `try` it does
-      -- This appeared after #4522, which made Nat.sub_one_le a simp lemma
+      -- This appeared after #4522, which made Nat.sub_le a simp lemma
     search_lex solve
       | decreasing_trivial
       | apply Nat.bitwise_rec_lemma; assumption


semorrison:
@nomeata, could you assist me on this one? `tests/lean/guessLexTricky.lean` appears to be the only thing failing, and I can't work out what is going wrong.

src/Init/Data/Nat/Basic.lean
```diff
@@ -278,10 +278,13 @@ theorem succ_sub_succ_eq_sub (n m : Nat) : succ n - succ m = n - m := by
   | zero      => exact rfl
   | succ m ih => apply congrArg pred ih
 
-@[simp] theorem pred_le : ∀ (n : Nat), pred n ≤ n
+theorem pred_le : ∀ (n : Nat), pred n ≤ n
   | zero   => Nat.le.refl
   | succ _ => le_succ _
 
+@[simp] theorem sub_one_le (n : Nat) : n - 1 ≤ n :=
+  pred_le n
```
nomeata, line 286: Isn't this lemma subsumed by `Nat.sub_le`, which could just be made `@[simp]`?

nomeata:
Pushed a work-around

tests/lean/guessLexTricky.lean
```diff
@@ -39,32 +39,32 @@ macro_rules | `(tactic|search_lex $ts:tacticSeq) => `(tactic| (
...
+-- termination_by (y, 0, 1)
 decreasing_by
   simp_wf
-  search_lex solve
-    | decreasing_trivial
-    | apply Nat.bitwise_rec_lemma; assumption
+  try -- the need for `try` here is fishy
+      -- the proof with explicit `termination_by` does not need it, so it should not throw
+      -- GuessLex off, but without `try` it does
+      -- This appeared after #4522, which made Nat.sub_one_le a simp lemma
```
semorrison, line 57:
```suggestion
      -- This appeared after #4522, which made Nat.sub_le a simp lemma
```

PR #4538, opened 2024-06-23
fix: make import resolution case-sensitive on all platforms

Kha:


src/lake/Lake/Build/Module.lean
```diff
@@ -31,6 +31,10 @@ Recursively parse the Lean files of a module and its imports
 building an `Array` product of its direct local imports.
 -/
 def Module.recParseImports (mod : Module) : FetchM (Array Module) := do
+  -- Make sure we fail reading the file independently of the FS case sensitivity
+  unless (← Lean.moduleExists mod.lib.srcDir "lean" mod.name) do
+    logError s!"file {mod.leanFile} of module {mod.name} does not exist"
+    failure
```
Kha, line 37: @tydeu Does this look reasonable?

tydeu: It should use `error` rather than `logError; failure`. I would also make the error a bit more concise as the module name is already part of the caption and file name. Otherwise, looks great!

```suggestion
  -- Make sure we fail reading the file independently of the FS case sensitivity
  unless (← Lean.moduleExists mod.lib.srcDir "lean" mod.name) do
    error s!"module source file not found: {mod.leanFile}"
```

tydeu (approved):

PR #4539, opened 2024-06-23
feat: @[simp] Nat.add_assoc

kim-em:
Just an experiment for now.


semorrison:
Not a good idea, I think. Perhaps if `Nat.add_assoc` were the other way round, but that is even more disruptive.

PR #4542, opened 2024-06-23
feat: termination_by structural

### LINKED ISSUE #3909: RFC: `termination_by structurally x`
### Proposal

Right now if the user writes a recursive function without annotation, Lean tries

 * structural recursion on each argument
 * well-founded recursion with GuessLex

If the user wants to be explicit about well-founded recursion, they can use `termination_by x`. But they have no way of indicating that they want to use structural recursion on a specific argument. 

Right now the above search procedure is not slow, so the need isn’t high, but once we have structural recursion on mutually recursive functions, the first step might consider many combinations, and it would be good if the user can just declare their intent.

Furthermore, if the user can be specific, they get a less noisy error message when the derecursifying fails.

Initial proposal for the syntax is
```
termination_by structurally x      -- if x is bound before the colon
termination_by structurally x => x -- if x is bound after the colon
```
in analogy to the existing `termination_by`. This way, no new “definition trailing keyword” is introduced.

It can also be extended to more variants in the future, such as `termination_by tail-recursion` (which isn’t really “termination”, sigh.)

Once we have this, then `termination_by?` should *not* force well-founded recursion and should, if lean detects structural recursion, suggest the corresponding `termination_by structurally …` annotation.

### Community Feedback

(none yet)

### Impact

### LINKED ISSUE #4546: `termination_by?` code action unergonomic for mutual recursion
The `termination_by?` command suffix shows the inferred termination argument to the user and offers  a “Try this” command to replace it. It works fine for non-mutual recursive function, but for mutual recursive functions, there is a problem.

Consider
```
mutual
  def foo  : Nat → Nat
  | 0 => 0
  | n+1 => bar n
  termination_by? -- Try this: termination_by x1 => x1

  def bar : Nat → Nat
  | 0 => 0
  | n+1 => foo n
end
```

If the user clicks on the `Try this` action, they get a broken file
```
mutual
  def foo  : Nat → Nat
  | 0 => 0
  | n+1 => bar n
  termination_by x1 => x1

  def bar : Nat → Nat -- Missing `termination_by`; this function is mutually
                      -- recursive with foo, which has a `termination_by` clause.
  | 0 => 0
  | n+1 => foo n
end
```
because (right now) either every or no function of a clique can have a `termination_by` annotation.

Possible ways forward:

 * Show the inferred termination measure for every function, and offer a code action that replaces 
   all of them at once.

   This requires extending the Try this functionality.

 * If some functions have a `termination_by` and others don’t, ignore the `termination_by`  (with a warning), so that the user can write 
    
    ```
    mutual
      def foo  : Nat → Nat
      | 0 => 0
      | n+1 => bar n
      termination_by x1 => x1 -- Warning: ignored until every function has one
    
      def bar : Nat → Nat
      | 0 => 0
      | n+1 => foo n
      termination_by?
    end
    ```
    and still have GuessLex running and be able to click on `termination_by?`.

 * If some functions have a `termination_by` and others don’t, do not complain, run GuessLex as usual, and afterwards check that the inferred measure matches the given one.
  
   So `termination_by` can be used to assert the inferred measure and document it, and only if all functions in a clique have one, one gets the benefit of not running GuessLex at all.

 * If some functions have a `termination_by` and others don’t, do not complain, and run GuessLex in a way that extends the given termination argument to the other functions. (I don’t think this is easily possible or worth it, just listing for completeness.)

So far nobody complained, so this isn’t very pressing.

### Versions

4.8.0

### Impact

nomeata:
This implements the `termination_by structural` syntax proposed in #3909.

I went with `termination_by structural` over, say,
`termination_by (config := {method := .structural})` mainly because it was
easier to get going (otherwise I’d have to look into how to define recursive
parsers, as `Parser.config` depends on `term` and `termination_by` is part of
term. But also because I find it more ergonomic and aesthetic as a user.
But syntax can still change.

The `termination_by?` syntax will no longer force well-founded recursion,
and instead the inferred `termination_by structurally` annotation will be shown
if structural termination is possible.

While I was it, this fixes #4546 the easy way (log errors about but otherwise
ignore incomplete `termination_by` sets for mutual recursion). Maybe we get
multiple replacements (#4551), but even then this this good behavior.

Involves a bit of shuffling around `TerimationHints` (now validated for a
clique already by `PreDefinition.main`) and `TerminationArguments` (now lifted
out of the `WF` namespace, and a bit simplified).

Fixes #3909

### DIFF
--- a/tests/lean/run/terminationByStructurally.lean
+++ b/tests/lean/run/terminationByStructurally.lean
@@ -71,7 +71,7 @@ def ackermann2 (n m : Nat) := match n, m with
 termination_by structural m
 
 /--
-error: The termination argument of a structural recursive function must be one of the parameters 'n', but
+error: The termination argument of a structurally recursive function must be one of the parameters 'n', but
   id n + 1
 isn't one of these.
 -/
--- a/tests/lean/termination_by.lean.expected.out
+++ b/tests/lean/termination_by.lean.expected.out
@@ -17,8 +17,8 @@ Try this: termination_by x1 => x1
 termination_by.lean:79:2-79:27: error: Incomplete set of `termination_by` annotations:
 This function is mutually with Test.f, Test.h and Test.i, which do not have a `termination_by` clause.
 The present clause is ignored.
-termination_by.lean:101:2-101:27: error: Invalid `termination_by`; this function is mutually recursive with Test2.f, which is marked as `termination_by structurally` so this one also needs to be marked `structurally`.
-termination_by.lean:120:2-120:40: error: Invalid `termination_by`; this function is mutually recursive with Test3.f, which is not marked as `structurally` so this one cannot be `structurally` either.
-termination_by.lean:144:2-144:40: error: Invalid `termination_by`; this function is mutually recursive with Test4.f, which is not marked as `structurally` so this one cannot be `structurally` either.
-termination_by.lean:159:2-159:21: error: Invalid `decreasing_by`; this function is marked as structural recursive, so no explicit termination proof is needed.
+termination_by.lean:101:2-101:27: error: Invalid `termination_by`; this function is mutually recursive with Test2.f, which is marked as `termination_by structural` so this one also needs to be marked `structural`.
+termination_by.lean:120:2-120:38: error: Invalid `termination_by`; this function is mutually recursive with Test3.f, which is not marked as `structural` so this one cannot be `structural` either.
+termination_by.lean:144:2-144:38: error: Invalid `termination_by`; this function is mutually recursive with Test4.f, which is not marked as `structural` so this one cannot be `structural` either.
+termination_by.lean:159:2-159:21: error: Invalid `decreasing_by`; this function is marked as structurally recursive, so no explicit termination proof is needed.
 termination_by.lean:149:0-165:3: error: structural recursion does not handle mutually recursive functions


nomeata:
More on my preferrence of `termination_by structurally` over `termination_by (config := {method := .structural})`:

As a user, these “record terms syntax as part of tactic syntax” feels very clumsy and heavy (not helped by the rather verbose `(config := ` and `)`). This gives me the impression that setting options is meant for advanced users, for corner cases, not for “idiomatic lean”.

I get this impression for example from the fact the simp syntax offers `simp only` rather than `simp (config := {onlyGivenTheorems := true)`, because `simp only` is a common, basic use case , while other simp options (`memoize`, `failIfUnchanged`) are clearly more advanced, and appear to be more aimed at users wrapping up the `simp` tactic within other user-facing tactic, or maybe even only intended for use from the `MetaM` level.

Maybe if the concrete syntax was a bit less heavy (no `config :=`, no need for `:= true` for boolean flags) I’d come to a different conclusion.

src/Lean/Parser/Term.lean
```diff
@@ -583,24 +583,31 @@ letrec we need them here already.
...
 
 /--
-Specify a termination argument for well-founded termination:
+Specify a termination argument for recursive functions.
 ```
 termination_by a - b
 ```
 indicates that termination of the currently defined recursive function follows
-because the difference between the the arguments `a` and `b`.
+because the difference between the the arguments `a` and `b` decreases.
```
richardkiss, line 591:
```suggestion
because the difference between the arguments `a` and `b` decreases.
```

Issue #3643, opened 2024-03-10
Unexpected behavior of `local ext` and `scoped ext`


fgdorais:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to mathlib4 or std4.

### Description

The `@[local ext]`/`@[scoped ext]` variants don't work as expected when applied to structures. 

* The `.ext` and `.ext_iff` are unconditionally generated, which fails if the lemmas were already generated by an unrelated `@[local ext]` in a a dependency, for example. 

* The generated `.ext` lemma is unexpectedly marked plain `@[ext]` even when `@[local ext]` or `@[scoped ext]` is used. For example, one would expect `@[scoped ext] structure A ...` to work the same as:
```lean
structure A ...
@[scoped ext] protected theorem A.ext ...
```
* The generated `.ext` lemma has a sub-optimal signature. For the simple structure `structure A (α) where a : α` the generated extensionality lemma is `A.ext.{u_1} {α : Sort u_1} (x y : A α) (a : x.a = y.a) : x = y` but one would expect `x` and `y` to be implicit parameters. (Even the `Init.Ext` file has manually generated ext lemmas to work around this.)

### Context

These bugs were noticed in the discussion of [std4#611](https://github.com/leanprover/std4/pull/611). This issue is an expanded version of issue [std4#618](https://github.com/leanprover/std4/issues/618), migrated here since the `ext` tactic was moved from `Std` to core.

### Steps to Reproduce

Here are three self-contained examples illustrating the issues.

1. Applying @[ext] to a structure should not attempt to regenerate existing extensionality lemmas if they already exist.

```lean
@[local ext] structure S (α β : Type _) where 
  a : α
  b : β

/- unexpectedly fails
'S.ext' has already been declared
-/
attribute [local ext] S

/- works as expected -/
attribute [local ext] S.ext
```

2. The `@[local ext]` attribute should basically only generate the extensionality lemmas with no attributes.

```lean
-- Local example:

/- works -/
@[local ext] structure S1 (α β : Type _) where 
  a : α
  b : β

/- generated ext lemma
S1.ext.{u_1, u_2} {α : Type u_1} {β : Type u_2} (x y : S1 α β) (a : x.a = y.a) (b : x.b = y.b) : x = y
-/
#check S1.ext

[...]

PR #4543, opened 2024-06-23
feat: make `@[ext]` derive `ext_iff` theorems from user `ext` theorems

kmill:
This PR refactors the 'ext' attribute and implements the following features:
- The 'local' and 'scoped' attribute kinds are now usable.
- The attribute realizes the `ext`/`ext_iff` lemmas when they do not already exist, rather than always generating them. This is useful in conjunction with `@[local ext]`.
- Adding `@[ext]` to a user ext lemma now realizes an `ext_iff` lemma as well; formerly this was only for structures. The name of the generated `ext_iff` theorem for a user `ext` theorem named `A.B.myext` is `A.B.myext_iff`. If this process leads to an error, the user can write `@[ext (iff := false)]` to disable this feature.

Breaking changes:
- Now the "x" and "y" term arguments to the realized `ext` and `ext_iff` lemmas are implicit.
- Now the realized `ext` and `ext_iff` lemmas are protected.

Bootstrapping notes:
- There are a few `ext_iff` lemmas to address after the next stage0 update.

Closes https://github.com/leanprover/lean4/issues/3643

Suggested by Floris [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/.22Missing.20Tactics.22.20list/near/446267660).

### DIFF
--- a/src/Init/Data/Char/Lemmas.lean
+++ b/src/Init/Data/Char/Lemmas.lean
@@ -31,10 +31,13 @@ theorem utf8Size_eq (c : Char) : c.utf8Size = 1 ∨ c.utf8Size = 2 ∨ c.utf8Siz
   rw [Char.ofNat, dif_pos]
   rfl
 
-@[ext] theorem ext : {a b : Char} → a.val = b.val → a = b
+-- TODO(kmill): remove ext_iff and apply [ext] to ext directly
+protected theorem ext : {a b : Char} → a.val = b.val → a = b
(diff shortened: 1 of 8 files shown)


fpvandoorn (commented):
I thought this was a Mathlib PR, so I started reviewing it. It looks good to me. The only surprising thing to me is that the macro in `Init/Ext` relies on the macro in `Lean/Elab/Tactic/Ext`, but I don't know if that is an actual issue.

PR #4552, opened 2024-06-24
feat: suggest fixes for potentially misspelled identifiers

david-christiansen:
This PR uses a heuristic based on the Levenshtein distance to suggest corrections to potentially-misspelled identifiers, along with quickfix code actions.

It's not yet ready for merging. TODO:
 * [ ] Make sure it's not too slow in a truly huge context
 * [ ] ~~Apply the technique to a few related error messages, like `unknown constant`, so there's a consistent UX~~
 * [x] Add tests for the code action aspect
 * [x] Documentations and explanations

### DIFF
--- a/src/Lean/Elab/NameSuggestions.lean
+++ b/src/Lean/Elab/NameSuggestions.lean
@@ -9,6 +9,8 @@ import Lean.Elab.InfoTree
 import Lean.Message
 import Lean.Util.EditDistance
 
+set_option linter.missingDocs true
+
 open Lean.EditDistance
 
 namespace Lean.Elab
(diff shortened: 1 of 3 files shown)


nomeata:
Lean is growing up :-)

david-christiansen:
It looks like adding it to "unknown constant" messages will take a bit more thought, just because most of those are thrown where there's no infotree available. I think this is a usability improvement as-is, so I think the remaining thing to do prior to merge is see how it fares performance-wise on all Mathlib.

david-christiansen:
!bench

david-christiansen:
I won't have time to look at this this week, but the performance issue definitely needs investigating before it gets merged.

PR #4560, opened 2024-06-25
feat: remove Decidable instances from GetElem

kim-em:


src/Init/GetElem.lean
```diff
@@ -82,9 +82,19 @@ class LawfulGetElem (cont : Type u) (idx : Type v) (elem : outParam (Type w))
    (dom : outParam (cont → idx → Prop)) [ge : GetElem cont idx elem dom] : Prop where
 
   getElem?_def (c : cont) (i : idx) [Decidable (dom c i)] :
-    c[i]? = if h : dom c i then some (c[i]'h) else none := by intros; eq_refl
+      c[i]? = if h : dom c i then some (c[i]'h) else none := by
+    intros
+    try simp only [getElem?, reduceDIte]
```
TwoFX, line 87: This should be `reduceDite` I think.

src/Lean/Data/HashMap.lean
```diff
@@ -222,6 +222,8 @@ def insertIfNew (m : HashMap α β) (a : α) (b : β) : HashMap α β × Option
 
 instance : GetElem (HashMap α β) α (Option β) fun _ _ => True where
   getElem m k _ := m.find? k
+  getElem? m k := some (m.find? k) -- Not really sure what should go here. We're meant to be returning an `Option (Option β)`.
```
david-christiansen, line 225: Isn't this a sign that the instance doesn't quite make sense here? I'd expect there to be an `Elem` predicate, and that the result of `getElem` would not be an `Option`.

david-christiansen: (I get the pragmatic reason why we don't have that right now, of course)

semorrison: Yes, @TwoFX will replace this with something more sensible in the big HashMap refactor.

src/Init/GetElem.lean
semorrison: The capitalisation changed at some point, this is correct now.

david-christiansen:
Unfortunately, it doesn't seem to let me leave suggestions for lines that weren't changed. But here's a suggestion for an updated docstring to match the code better:

```
The class `GetElem coll idx elem valid` implements lookup notation,
specifically `xs[i]`, `xs[i]?`, `xs[i]!`, and `xs[i]'p`.

The types `coll`, `idx`, and `elem` are the collection, the index, and the
element types. A single collection may support lookups with multiple index
types. The relation `valid` determines when the index is guaranteed to be
valid; lookups of valid indices are guaranteed not to fail.

For example, an instance for arrays looks like
`GetElem (Array α) Nat α (fun xs i => i < xs.size)`. In other words, given an
array `xs` and a natural number `i`, `xs[i]` will return an `α` when `valid xs i`
holds, which is true when `i` is less than the size of the array. `Array`
additionally supports indexing with `USize`. Because the bounds are
checked at compile time, no runtime check is required.

Given `xs[i]` with `xs : coll` and `i : idx`, Lean looks for an instance of
`GetElem coll idx elem valid` and uses this to infer the type of the return
value `elem` and side conditions `valid` required to ensure `xs[i]` yields
a valid value of type `elem`. The tactic `get_elem_tactic` is
invoked to prove validity automatically. The `xs[i]'p` notation uses the
proof `p` to satisfy the validity condition, but it is often easier to place the
proof in the context using `have`, because `get_elem_tactic` tries
`assumption`.

`xs[i]?` and `xs[i]!` do not impose a proof obligation; the former returns
an `Option elem`, with `none` signalling that the value isn't present, and
the latter returns `elem` but panics if the value isn't there, returning
`default : elem` based on the `Inhabited elem` instance.

The proof side-condition `valid xs i` is automatically dispatched by the
`get_elem_tactic` tactic; this tactic can be extended by adding more clauses to
`get_elem_tactic_trivial` using `macro_rules`.

Important instances include:
  * `arr[i] : α` where `arr : Array α` and `i : Nat` or `i : USize`: does array
    indexing with no bounds check and a proof side goal `i < arr.size`.
  * `l[i] : α` where `l : List α` and `i : Nat`: index into a list, with proof
    side goal `i < l.length`.
  * `stx[i] : Syntax` where `stx : Syntax` and `i : Nat`: get a syntax argument,
    no side goal (returns `.missing` out of range)

-/
class GetElem (coll : Type u) (idx : Type v) (elem : outParam (Type w))
              (valid : outParam (coll → idx → Prop)) where
  /--
  The syntax `arr[i]` gets the `i`'th element of the collection `arr`. If there
  are proof side conditions to the application, they will be automatically
  inferred by the `get_elem_tactic` tactic.
  -/
```

What do you think?

nomeata:
Good! Should we exclude `Syntax` from the docstring? The docstring is most useful to relative beginners of Lean, and I expect they are confused or at least distracted by meta-programming concepts. We can hopefully add HashMap instead soon!

david-christiansen:
I agree!

semorrison:
Thanks, I've adopted the doc-string, with a few minor changes.

src/Init/GetElem.lean
```diff
@@ -7,54 +7,75 @@ prelude
...
+  getElem? (xs : coll) (i : idx) : Option elem
 
-  getElem! [Inhabited elem] (xs : coll) (i : idx) [Decidable (valid xs i)] : elem :=
-    match getElem? xs i with | some e => e | none => outOfBounds
+  /--
+  The syntax `arr[i]!` gets the `i`'th element of the collection `arr`,
+  if it is present, and otherwise panics at runtime and returns the `default` term
+  from `Inhabited elem`.
+  -/
+  getElem! [Inhabited elem] (xs : coll) (i : idx) : elem
```
tydeu, line 78: Can we retain the default for `getElem!` since it just uses `getElem?` (and thus does not need the `Decidable` instance if `getElem?` does not have it).

src/Init/Data/Array/Basic.lean
```diff
@@ -59,6 +59,8 @@ def uget (a : @& Array α) (i : USize) (h : i.toNat < a.size) : α :=
 
 instance : GetElem (Array α) USize α fun xs i => i.toNat < xs.size where
   getElem xs i h := xs.uget i h
+  getElem? xs i  := if h : i.toNat < xs.size then some (xs.uget i h) else none
```
tydeu, line 62: Instead of copying the decidable default to each instance, can we have a helper (e.g., `decidableGetElem?`) defined in `GetElem` that is used in these instances?

src/Init/GetElem.lean
semorrison: Yes, indeed, the default value for `getElem!` in terms of `getElem?` works nicely. I've updated that now.

src/Init/Data/Array/Basic.lean
semorrison: The problem is that the right hand side of `getElem` would still need to be repeated inside the call the `decidableGetElem?`.

I think we can perhaps resolve these default field difficulties by splitting `GetElem` in `GetElemBase` (with just `getElem), and extending it. I'll try.

semorrison: Works nicely, I think.

tydeu: @semorrison Looks good! Given that `GetElemBase` defines `getElem`, I think it would be best to give it the `GetElem` name and name the other something like `OptGetElem`, `GetElem?`, or `GetElem?!`.

src/Init/GetElem.lean
```diff
@@ -66,6 +38,74 @@ macro_rules | `($x[$i]) => `(getElem $x $i (by get_elem_tactic))
...
+
+Important instances include:
+  * `arr[i] : α` where `arr : Array α` and `i : Nat` or `i : USize`: does array
+    indexing with no runtime bounds check and a proof side goal `i < arr.size`.
+  * `l[i] : α` where `l : List α` and `i : Nat`: index into a list, with proof
+    side goal `i < l.length`.
+
+-/
+class GetElem (coll : Type u) (idx : Type v) (elem : outParam (Type w))
+    (valid : outParam (coll → idx → Prop)) extends GetElemBase coll idx elem valid where
```
tydeu, line 90:
```suggestion
class GetElem (coll : Type u) (idx : Type v) (elem : outParam (Type w)) where
```

With the `getElemOfDecidable` instances, this class could avoid extending `GetElemBase`.

src/Init/Data/Array/Subarray.lean
```diff
@@ -44,10 +44,10 @@ def get (s : Subarray α) (i : Fin s.size) : α :=
    exact Nat.add_lt_of_lt_sub this
   s.array[s.start + i.val]
 
-instance : GetElem (Subarray α) Nat α fun xs i => i < xs.size where
+instance : GetElemBase (Subarray α) Nat α fun xs i => i < xs.size where
   getElem xs i h := xs.get ⟨i, h⟩
 
-instance : LawfulGetElem (Subarray α) Nat α fun xs i => i < xs.size where
+instance : GetElem (Subarray α) Nat α fun xs i => i < xs.size where
```
tydeu, line 50: This instance is no longer necessary with `getElemOfDecidable`.

src/Init/Data/ByteArray/Basic.lean
```diff
@@ -49,15 +49,15 @@ def get! : (@& ByteArray) → (@& Nat) → UInt8
 def get : (a : @& ByteArray) → (@& Fin a.size) → UInt8
   | ⟨bs⟩, i => bs.get i
 
-instance : GetElem ByteArray Nat UInt8 fun xs i => i < xs.size where
+instance : GetElemBase ByteArray Nat UInt8 fun xs i => i < xs.size where
   getElem xs i h := xs.get ⟨i, h⟩
 
-instance : LawfulGetElem ByteArray Nat UInt8 fun xs i => i < xs.size where
+instance : GetElem ByteArray Nat UInt8 fun xs i => i < xs.size where
```
tydeu, line 55: This instance is no longer necessary with `getElemOfDecidable`.

src/Init/Data/ByteArray/Basic.lean
```diff
@@ -49,15 +49,15 @@ def get! : (@& ByteArray) → (@& Nat) → UInt8
...
 
-instance : GetElem ByteArray USize UInt8 fun xs i => i.val < xs.size where
+instance : GetElemBase ByteArray USize UInt8 fun xs i => i.val < xs.size where
   getElem xs i h := xs.uget i h
 
-instance : LawfulGetElem ByteArray USize UInt8 fun xs i => i.val < xs.size where
+instance : GetElem ByteArray USize UInt8 fun xs i => i.val < xs.size where
```
tydeu, line 60: This instance is no longer necessary with `getElemOfDecidable`.

tydeu (commented):
Below are a few `GetElem` instances that can be removed because of `getElemOfDecidable` which slipped through.

src/Init/Data/Array/Subarray.lean
semorrison: This is the base one. It's `GetElem?` that's not needed anymore!

src/Init/GetElem.lean
semorrison: There's no dependent parameter here?

Removing the extends just means we have to repeat ourselves in stating lemmas below, having separate instances for `GetElem` and `GetElem?`. I'm unconvinced.

semorrison:
> Below are a few `GetElem` instances that can be removed because of `getElemOfDecidable` which I imagine just slipped through.

The rename had already happened, so these are actually the ones that are needed!

src/Init/Data/Array/Lemmas.lean
```diff
@@ -220,7 +220,7 @@ theorem getElem?_len_le (a : Array α) {i : Nat} (h : a.size ≤ i) : a[i]? = no
 theorem getD_get? (a : Array α) (i : Nat) (d : α) :
   Option.getD a[i]? d = if p : i < a.size then a[i]'p else d := by
   if h : i < a.size then
-    simp [setD, h, getElem?]
+    simp [setD, h, getElem?, decidableGetElem?]
```
TwoFX, line 223: I haven't tried it, but I think better than unfolding would be
```suggestion
    simp [setD, h, getElem?_def]
```
because that is why we have the `LawfulGetElem` class.

src/Init/GetElem.lean
```diff
@@ -66,6 +38,77 @@ macro_rules | `($x[$i]) => `(getElem $x $i (by get_elem_tactic))
...
+-/
+class GetElem? (coll : Type u) (idx : Type v) (elem : outParam (Type w))
+    (valid : outParam (coll → idx → Prop)) extends GetElem coll idx elem valid where
+  /--
+  The syntax `arr[i]?` gets the `i`'th element of the collection `arr`,
+  if it is present (and wraps it in `some`), and otherwise returns `none`.
+  -/
+  getElem? : coll → idx → Option elem := by
+    intro xs i
+    exact decidableGetElem? xs i
```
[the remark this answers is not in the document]
semorrison: I guess not!

TwoFX (approved):
LGTM!

tydeu (approved):
Looks great to me as well!

PR #4569, opened 2024-06-26
fix: interrupt exception was swallowed by some `tryCatchRuntimeEx` uses

Kha:
This appears to have been a semantic merge conflict between #3940 and #4129. The effect on the language server is that if two edits are sufficiently close in time to create an interrupt, some elaboration steps like `simp` may accidentally catch the exception when it is triggered during their execution, which makes incrementality assume that elaboration of the body was successful, which can lead to incorrect reuse, presenting the interrupted state to the user with symptoms such as "uses sorry" without accompanying errors and incorrect lints.


Kha:
@leodemoura I'm going ahead with merging this because of its critical priority but happy for feedback on the new docstrings.

PR #4571, opened 2024-06-27
feat: bitVec shiftLeft recurrences for bitblasting

bollu:
```lean 
@[simp]
theorem shiftLeftRec_zero (x : BitVec w₁) (y : BitVec w₂) :
    shiftLeftRec x y 0 = x <<< (y &&& twoPow w₂ 0)  := by
  simp [shiftLeftRec]

@[simp]
theorem shiftLeftRec_succ (x : BitVec w₁) (y : BitVec w₂) :
    shiftLeftRec x y (n + 1) =
      (shiftLeftRec x y n) <<< (y &&& twoPow w₂ (n + 1)) := by
  simp [shiftLeftRec]

theorem shiftLeftRec_eq (x : BitVec w₁) (y : BitVec w₂) (n : Nat) (hn : n + 1 ≤ w₂) :
  shiftLeftRec x y n = x <<< (y.truncate (n + 1)).zeroExtend w₂ := by
```

These theorems are used for bitblasting shiftLeft in LeanSAT.

### DIFF
--- a/src/Init/Data/BitVec/Bitblast.lean
+++ b/src/Init/Data/BitVec/Bitblast.lean
@@ -119,7 +119,7 @@ theorem carry_width {x y : BitVec w} :
 /--
 If `x &&& y = 0`, then addition does not overflow, and thus `(x + y).toNat = x.toNat + y.toNat`.
 -/
-theorem toNat_add_eq_toNat_add_toNat_of_and_eq_zero {x y : BitVec w} (h : x &&& y = 0#w) :
+theorem toNat_add_of_and_eq_zero {x y : BitVec w} (h : x &&& y = 0#w) :
     (x + y).toNat = x.toNat + y.toNat := by
   rw [toNat_add]
   apply Nat.mod_eq_of_lt
@@ -364,7 +364,7 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
 
 The theorem `shiftLeft_eq_shiftLeftRec` proves the equivalence of `(x <<< y)` and `shiftLeftRec`.
 
-Together with equations `shiftLeftRec_zero`, `shiftLeftRec_succ`, 
+Together with equations `shiftLeftRec_zero`, `shiftLeftRec_succ`,
 this allows us to unfold `shiftLeft` into a circuit for bitblasting.
  -/
 def shiftLeftRec (x : BitVec w₁) (y : BitVec w₂) (n : Nat) : BitVec w₁ :=
@@ -392,7 +392,7 @@ theorem shiftLeft_or_of_and_eq_zero {x : BitVec w₁} {y z : BitVec w₂}
     (h : y &&& z = 0#w₂) :
     x <<< (y ||| z) = x <<< y <<< z := by
   rw [← add_eq_or_of_and_eq_zero _ _ h,
-    shiftLeft_eq', toNat_add_eq_toNat_add_toNat_of_and_eq_zero h]
+    shiftLeft_eq', toNat_add_of_and_eq_zero h]
   simp [shiftLeft_add]
 
 /--
@@ -403,7 +403,7 @@ theorem shiftLeftRec_eq {x : BitVec w₁} {y : BitVec w₂} {n : Nat} :
   induction n generalizing x y
   case zero =>
     ext i
-    simp only [shiftLeftRec_zero, twoPow_zero, Nat.reduceAdd, truncate_one_eq_ofBool_getLsb]
+    simp only [shiftLeftRec_zero, twoPow_zero, Nat.reduceAdd, truncate_one]
     suffices (y &&& 1#w₂) = zeroExtend w₂ (ofBool (y.getLsb 0)) by simp [this]
     ext i
     by_cases h : (↑i : Nat) = 0
@@ -415,7 +415,7 @@ theorem shiftLeftRec_eq {x : BitVec w₁} {y : BitVec w₂} {n : Nat} :
     by_cases h : y.getLsb (n + 1)
     · simp only [h, ↓reduceIte]
       rw [zeroExtend_truncate_succ_eq_zeroExtend_truncate_or_twoPow_of_getLsb_true h,
-        shiftLeft_or_eq_shiftLeft_shiftLeft_of_and_eq_zero]
+        shiftLeft_or_of_and_eq_zero]
       simp
     · simp only [h, false_eq_true, ↓reduceIte, shiftLeft_zero']
       rw [zeroExtend_truncate_succ_eq_zeroExtend_truncate_of_getLsb_false (i := n + 1)]


src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -235,4 +249,142 @@ theorem sle_eq_carry (x y : BitVec w) :
     x.sle y = !((x.msb == y.msb).xor (carry w y (~~~x) true)) := by
   rw [sle_eq_not_slt, slt_eq_not_carry, beq_comm]
 
+/-## shiftLeft recurrence -/
```
semorrison, line 252: Like other section titles, please!

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -235,4 +249,142 @@ theorem sle_eq_carry (x y : BitVec w) :
...
+    shiftLeftRec x y 0 = x <<< (y &&& twoPow w₂ 0)  := by
+  simp [shiftLeftRec]
+
+@[simp]
+theorem shiftLeftRec_succ (x : BitVec w₁) (y : BitVec w₂) :
+    shiftLeftRec x y (n + 1) =
+      (shiftLeftRec x y n) <<< (y &&& twoPow w₂ (n + 1)) := by
+  simp [shiftLeftRec]
+
+-- | TODO: should this be a simp-lemma? Probably not.
```
semorrison, line 278: Let's say no for now, but I think this is worth reconsidering. It would reduce to API surface area for verification, and make shift by a bitvec a "programming only" concept.

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -235,4 +249,142 @@ theorem sle_eq_carry (x y : BitVec w) :
...
+theorem shiftLeft_zero' (x : BitVec w) :
+    x <<< (0#w₂) = x := by
+  simp [shiftLeft_eq']
+
+@[simp]
+theorem getLsb_ofNat_one (w i : Nat) :
+    (1#w).getLsb i = (decide (i = 0) && decide (i < w)) := by
+  rcases w with rfl | w
+  · simp;
+  · simp [getLsb]
```
semorrison, line 293: non-terminal, and below

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -290,7 +290,7 @@ theorem zeroExtend_truncate_succ_eq_zeroExtend_truncate_add_twoPow (x : BitVec w
         simp [hik', hik'']
   · ext k
     simp
-    omega
+    by_cases hi : x.getLsb i <;> simp [hi] <;> omega
```
alexkeizer, line 293: This feels like it could be shortened to:
```suggestion
    cases x.getLsb i <;> simp <;> omega
```

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,101 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
+    by_cases hi : i = 0
+    · simp [hi]
+    · simp [hi]
+      intros _; simp [testBit, shiftRight_eq_div_pow];
+      suffices 1 / 2^i = 0 by simp [this]
+      apply Nat.div_eq_of_lt;
+      exact Nat.one_lt_two_pow_iff.mpr hi
+
+theorem shiftLeft_or_eq_shiftLeft_shiftLeft_of_and_eq_zero {x : BitVec w} {y z : BitVec w₂}
+    (h : y &&& z = 0#w₂) (h' : y.toNat + z.toNat < 2^w₂):
```
alexkeizer, line 367: Isn't `h'` implied by `h`? Given that `y` and `z` have no bits in common, addition should be equal to bitwise or, and bitwise or cannot overflow

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,101 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
+  induction n generalizing x y
+  case zero =>
+    ext i
+    simp only [shiftLeftRec_zero, twoPow_zero_eq_one, Nat.reduceAdd, truncate_one_eq_ofBool_getLsb]
+    have heq : (y &&& 1#w₂) = zeroExtend w₂ (ofBool (y.getLsb 0)) := by
+      ext i
+      by_cases h : (↑i : Nat) = 0 <;> simp [h, Bool.and_comm]
+    simp [heq]
+  case succ n ih =>
+    simp
```
alexkeizer, line 387: Another non-terminal simp!

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,101 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
+    by_cases h : y.getLsb (n + 1) <;> simp [h]
```
alexkeizer, line 388:
```suggestion
    cases h : y.getLsb (n + 1) <;> simp [h]
```

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,101 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
+            · apply Nat.mod_lt
+              · apply Nat.pow_pos (by decide)
+            · apply Nat.pow_le_pow_of_le_right (by decide) (by omega)
+          · simp
+            apply Nat.pow_le_pow_of_le_right (by decide) (by omega)
+    · rw [ih (hn := by omega)]
+      rw [zeroExtend_truncate_succ_eq_zeroExtend_truncate_of_getLsb_false (i := n + 1)]
+      simp [h]
+
+theorem shiftLeft_eq_shiftLeft_rec (x : BitVec ℘) (y : BitVec w₂) :
```
alexkeizer, line 420:
```suggestion
theorem shiftLeft_eq_shiftLeftRec (x : BitVec w₁) (y : BitVec w₂) :
```

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,101 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
+
+/--
+A recurrence that describes shifting by an arbitrary bitvector
+as shifting by a constant amount.
+ -/
+def shiftLeftRec (x : BitVec w₁) (y : BitVec w₂) (n : Nat) : BitVec w₁ :=
+  let shiftAmt := (y &&& (twoPow w₂ n))
+  match n with
+  | 0 => x <<< shiftAmt
+  | n + 1 => (shiftLeftRec x y n) <<< shiftAmt
```
alexkeizer, line 339: From a quick glance, it seems `shiftLeftRec x y n` computes "shifting `x` left by the first `n` bits of `y`", i.e:
  `shiftLeftRec x y n = x <<< (y.toNat % 2^n)`
I think it'd be helpful to at least write this direct form in the docstring

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,101 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
+    x <<< (y ||| z) = x <<< y <<< z := by
+  simp [← add_eq_or_of_and_eq_zero _ _ h, shiftLeft_eq', shiftLeft_add,
+    toNat_add, Nat.mod_eq_of_lt h']
+
+theorem getLsb_shiftLeft' {x : BitVec w} {y : BitVec w₂} {i : Nat} :
+    (x <<< y).getLsb i = (decide (i < w) && !decide (i < y.toNat) && x.getLsb (i - y.toNat)) := by
+  simp [shiftLeft_eq', getLsb_shiftLeft]
+
+theorem shiftLeftRec_eq {x : BitVec w₁} {y : BitVec w₂} {n : Nat} (hn : n + 1 ≤ w₂) :
+  shiftLeftRec x y n = x <<< (y.truncate (n + 1)).zeroExtend w₂ := by
```
alexkeizer, line 377:
```suggestion
    shiftLeftRec x y n = x <<< (y.truncate (n + 1)).zeroExtend w₂ := by
```

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,101 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
+    (h : y &&& z = 0#w₂) (h' : y.toNat + z.toNat < 2^w₂):
+    x <<< (y ||| z) = x <<< y <<< z := by
+  simp [← add_eq_or_of_and_eq_zero _ _ h, shiftLeft_eq', shiftLeft_add,
+    toNat_add, Nat.mod_eq_of_lt h']
+
+theorem getLsb_shiftLeft' {x : BitVec w} {y : BitVec w₂} {i : Nat} :
+    (x <<< y).getLsb i = (decide (i < w) && !decide (i < y.toNat) && x.getLsb (i - y.toNat)) := by
+  simp [shiftLeft_eq', getLsb_shiftLeft]
+
+theorem shiftLeftRec_eq {x : BitVec w₁} {y : BitVec w₂} {n : Nat} (hn : n + 1 ≤ w₂) :
```
alexkeizer, line 376: Do you need `hn`? I'd expect the equality to be true even for `n > w2`, since the use of `twoPow` would make the `shiftAmt` 0 for higher values of `n`

src/Init/Data/BitVec/Bitblast.lean
bollu: @alexkeizer and I discussed this. This needs a similar theorem on the Nat level to get rid of the assumption. I think it's worth removing the assumption, but in a follow up PR. I've added a note accordingly.

src/Init/Data/BitVec/Bitblast.lean
bollu: Done.

src/Init/Data/BitVec/Bitblast.lean
bollu: I disagree, `by_cases` gives us control over the proof state here.

src/Init/Data/BitVec/Bitblast.lean
bollu: This simplifies the case analysis inside the induction, where we use ` n + 1 ≤ w₂` to simplify our intermediate modules. While I believe the inequality succeeds for higher values, it would need a case split on `n > w₂`, which does not seem useful for the terminal goal of bitblasting-friendly lemmas.

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,112 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
+theorem getLsb_ofNat_one {w i : Nat} :
+    (1#w).getLsb i = (decide (i = 0) && decide (i < w)) := by
+  rcases w with rfl | w
+  · simp
+  · simp only [getLsb, toNat_ofNat, testBit_mod_two_pow]
+    by_cases hi : i = 0
+    · simp [hi]
+    · simp only  [hi, decide_False, Bool.false_and,
+        and_eq_false_imp, decide_eq_true_eq]
+      intros _; simp [testBit, shiftRight_eq_div_pow];
```
tobiasgrosser, line 362: None terminal simp

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,110 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
   · simp
   · omega
 
+/-## shiftLeft recurrence for bitblasting -/
+
+/--
+A recurrence that describes shifting by an arbitrary bitvector
+as shifting by a constant amount.
```
alexkeizer, line 333:
```suggestion
as shifting by a constant amount.

This computation is equivalent to `x <<< (y.truncate (n + 1)).zeroExtend w₂`,
as witnessed by the `shiftLeftRec_eq` theorem further down.
We phrase it recursively to simplify other proofs
```

alexkeizer (commented):
LGTM

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,114 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
   · simp
   · omega
 
+/-## shiftLeft recurrence for bitblasting -/
```
semorrison, line 329:
```suggestion
/-! ## shiftLeft recurrence for bitblasting -/
```

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,114 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
+
+@[simp]
+theorem shiftLeftRec_succ {x : BitVec w₁} {y : BitVec w₂} :
+    shiftLeftRec x y (n + 1) =
+      (shiftLeftRec x y n) <<< (y &&& twoPow w₂ (n + 1)) := by
+  simp [shiftLeftRec]
+
+@[simp]
+theorem getLsb_ofNat_one {w i : Nat} :
+    (1#w).getLsb i = (decide (i = 0) && decide (i < w)) := by
```
semorrison, line 358: Does this belong in this file? Seems unlikely. Can it just be named `getLsb_one`?

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,114 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
+  · simp
+  · simp only [getLsb, toNat_ofNat, testBit_mod_two_pow]
+    by_cases hi : i = 0
+    · simp [hi]
+    · simp only  [hi, decide_False, Bool.false_and,
+        and_eq_false_imp, decide_eq_true_eq]
+      intros _
+      simp only [testBit, shiftRight_eq_div_pow, one_and_eq_mod_two, bne_eq_false_iff_eq]
+      suffices 1 / 2^i = 0 by simp [this]
+      apply Nat.div_eq_of_lt (Nat.one_lt_two_pow_iff.mpr hi)
```
semorrison, line 369: This proof is clearly too long for such a statement.

It should be:
```
  rw [← twoPow_zero_eq_one, getLsb_twoPow]
```
Except:
* `twoPow_zero_eq_one` should be renamed `twoPow_zero` (apologies I didn't catch this earlier).
* you still need to flip some `&&` and `=` to make this proof work. Ideally you will change the statement here to make it easy to prove.

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,114 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
+
+/--
+If `(y &&& z = 0)`, then shifting by `y ||| z` is the same as shifting by `y` and then by `z`.
+Note that the hypothesis `h'` is implied by `h : y &&& z = 0#w`,
+but we choose to take the additional hypothesis and leave proving it
+as an implication of `h` for a follow up PR.
```
semorrison, line 375: This is an easy consequence of `add_eq_or_of_and_eq_zero`, no? It's just a line or two, easier to write it than to write this doc-string!

semorrison: Particularly given that below you write a 20 line proof of this in a special case??

semorrison: Oops, I see, it's not implied by that.

semorrison, line 375: I don't think this should be too bad. Just prove inductively that `carry i x y false = false` given `x &&& y = 0`.

e.g.

```
theorem carry_of_and_eq_zero (h : x &&& y = 0) : carry i x y false = false := by
  induction i with
  | zero => simp
  | succ i ih => 
    replace h := congrArg (·.getLsb i) h
    simp_all [carry_succ]
```

comes right after `carry_succ`.

Seems easier that the proof in the special case you have to give for this hypothesis later. 

```
theorem carry_width {x y : BitVec w} : carry w x y c = decide (x.toNat + y.toNat + c.toNat ≥ 2^w) := sorry
```
should make your life easy, to turn the goal about `x.toNat + y.toNat` into something about `carry`.

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +326,114 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
 
+/-## shiftLeft recurrence for bitblasting -/
+
+/--
+A recurrence that describes shifting by an arbitrary bitvector
+as shifting by a constant amount.
+
+This computation is equivalent to `x <<< (y.truncate (n + 1)).zeroExtend w₂`,
+as witnessed by the `shiftLeftRec_eq` theorem further down.
+We phrase it recursively to simplify other proofs
```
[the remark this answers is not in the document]
semorrison: e.g. the doc-string of `shiftLeftRec_eq` is by itself a better doc-string here! I would suggest starting with that, and *then* giving the explanation you want about why this is helpful for bitblasting.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -626,6 +632,10 @@ theorem shiftLeft_zero_eq (x : BitVec w) : x <<< 0 = x := by
   apply eq_of_toNat_eq
   simp
 
+@[simp]
+theorem zero_shiftLeft_eq (n : Nat) : (0#w) <<< n = 0 := by
```
semorrison, line 636:
```suggestion
theorem zero_shiftLeft (n : Nat) : (0#w) <<< n = 0 := by
```

semorrison: I guess rename the previous lemma while you're at it.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -691,6 +701,20 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
     (x <<< n) <<< m = x <<< (n + m) := by
   rw [shiftLeft_add]
 
+/-! ### shiftLeft reductions from BitVec to Nat -/
+
+theorem shiftLeft_eq' {x : BitVec w} {y : BitVec w₂} :
+  x <<< y = x <<< y.toNat := by rfl
```
semorrison, line 707: Can this just be a simp lemma, and then we never prove anything about shifting by a BitVec?

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1471,6 +1495,27 @@ theorem mul_twoPow_eq_shiftLeft (x : BitVec w) (i : Nat) :
       apply Nat.pow_dvd_pow 2 (by omega)
     simp [Nat.mul_mod, hpow]
 
+@[simp]
+theorem and_twoPow_eq_getLsb {x : BitVec w} {i : Nat} :
```
semorrison, line 1499: Rename to `and_twoPow`, and also add `twoPow_and`.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1471,6 +1495,27 @@ theorem mul_twoPow_eq_shiftLeft (x : BitVec w) (i : Nat) :
...
+    x &&& (twoPow w i) = if x.getLsb i then twoPow w i else 0#w := by
+  ext j
+  simp only [getLsb_and, getLsb_twoPow]
+  by_cases hj : i = j <;> by_cases hx : x.getLsb i <;> simp_all
+
+theorem BitVec.toNat_twoPow {w : Nat} {i : Nat} : (twoPow w i).toNat = 2^i % 2^w := by
```
semorrison, line 1505: This has two `BitVec`s in the name. If you remove one, you'll find this already exists.

src/Init/Data/BitVec/Bitblast.lean
bollu: Thanks for pointing out the (much cleaner) proof strategy. Adapted.

src/Init/Data/BitVec/Bitblast.lean
bollu: Thanks, this is incredibly clean. I've implemented this now. The hint about using `carry` to prove this escaped me!

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -531,6 +537,12 @@ theorem and_assoc (x y z : BitVec w) :
   ext i
   simp [Bool.and_assoc]
 
+theorem and_comm (x y : BitVec w) :
+    x &&& y = y &&& x := by
+  ext i
+  simp [Bool.and_comm]
+
+
 /-! ### xor -/
```
tobiasgrosser, line 546:
```suggestion
  simp [Bool.and_comm]

/-! ### xor -/
```

Only one empty line needed between theorems.

src/Init/Data/BitVec/Bitblast.lean
bollu: Thanks to Kim's suggested cleanup, this is no longer necessary!

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -436,6 +436,12 @@ theorem zeroExtend_ofNat_one_eq_ofNat_one_of_lt {v w : Nat} (hv : 0 < v) :
   have hv := Nat.testBit_one_eq_true_iff_self_eq_zero.mp hi₁
   omega
 
+/-- Truncating to width 1 produces a bitvector equal to the the least significant bit. -/
```
tobiasgrosser, line 439: Drop the repeated 'the'

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +357,77 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
+theorem shiftLeftRec_succ {x : BitVec w₁} {y : BitVec w₂} :
+    shiftLeftRec x y (n + 1) = (shiftLeftRec x y n) <<< (y &&& twoPow w₂ (n + 1)) := by
+  simp [shiftLeftRec]
+
+/--
+If `y &&& z = 0`, `x <<< (y ||| z) = x <<< y <<< z`.
+This follows as `y &&& z = 0` implies `y ||| z = y + z`,
+and thus `x <<< (y ||| z) = x <<< (y + z) = x <<< y <<< z`.
+-/
+theorem shiftLeft_or_eq_shiftLeft_shiftLeft_of_and_eq_zero {x : BitVec w} {y z : BitVec w₂}
```
AtticusKuhn, line 390: Stylistically, it's not a big thing, but maybe you could label the widths as `w1` and `w2` instead of `w` and `w2`.

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -98,6 +98,37 @@ theorem carry_succ (i : Nat) (x y : BitVec w) (c : Bool) :
...
+  simp [carry]
+
+/--
+If `x &&& y = 0`, then addition does not overflow, and thus `(x + y).toNat = x.toNat + y.toNat`.
+-/
+theorem toNat_add_eq_toNat_add_toNat_of_and_eq_zero {x y : BitVec w} (h : x &&& y = 0#w) :
+    (x + y).toNat = x.toNat + y.toNat := by
+  rw [toNat_add]
+  apply Nat.mod_eq_of_lt
+  suffices ¬ decide (x.toNat + y.toNat + (false : Bool).toNat ≥ 2^w) by
```
AtticusKuhn, line 126: A minor nitpick, but is the type annotation nessecary? You could probably write `(x.toNat + y.toNat + false.toNat ≥ 2^w)` instead (or at least such code does not produce an error when I run it). I guess it's a stylistic choice.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1471,6 +1508,14 @@ theorem mul_twoPow_eq_shiftLeft (x : BitVec w) (i : Nat) :
       apply Nat.pow_dvd_pow 2 (by omega)
     simp [Nat.mul_mod, hpow]
 
+theorem twoPow_zero {w : Nat} : twoPow w 0 = 1#w := by
+  apply eq_of_toNat_eq
+  simp
+
+@[simp]
+theorem getLsb_one {w i : Nat} : (1#w).getLsb i = (decide (0 < w) && decide (0 = i)) := by
```
AtticusKuhn, line 1516: This is a minor stylistic point, but I think it would look cleaner to write
`(1#w).getLsb i = decide (0 < w ∧ 0 = i)` instead of `(decide (0 < w) && decide (0 = i))`.
I guess it's a matter of preference, so it's up to you.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -691,6 +706,22 @@ theorem shiftLeft_shiftLeft {w : Nat} (x : BitVec w) (n m : Nat) :
...
+theorem shiftLeft_eq' {x : BitVec w} {y : BitVec w₂} : x <<< y = x <<< y.toNat := by rfl
+
+@[simp]
+theorem shiftLeft_zero' {x : BitVec w} : x <<< 0#w₂ = x := by simp
+
+theorem shiftLeft_shiftLeft' {x y z : BitVec w} :
+    x <<< y <<< z = x <<< (y.toNat + z.toNat) := by
+  simp [shiftLeft_add]
+
+theorem getLsb_shiftLeft' {x : BitVec w} {y : BitVec w₂} {i : Nat} :
```
AtticusKuhn, line 721: See my comment above, but I think that it would look more consistent to call the widths `w1` and `w2` instead of `w` and `w2`.

src/Init/Data/BitVec/Lemmas.lean
bollu: @AtticusKuhn we write it this way because it's the simp normal form.

src/Init/Data/BitVec/Bitblast.lean
bollu: done.

src/Init/Data/BitVec/Lemmas.lean
bollu: done.

src/Init/Data/BitVec/Lemmas.lean
bollu: Done.

bollu:
awaiting-review

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +357,78 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
 
+/-! ## shiftLeft recurrence for bitblasting -/
+
+/--
+`shiftLeftRec x y n` shifts `x` to the left by the first `n` bits of `y`.
+
+The theorem `shiftLeft_eq_shiftLeftRec` proves the equivalence of `(x <<< y)` and `shiftLeftRec`.
+
+This in conjecton with equations `shiftLeftRec_zero`, `shiftLeftRec_succ`
+allows us to unfold `shiftLeft` into a circuit for bitblasting.
```
alexkeizer, line 368:
```suggestion
Together with equations `shiftLeftRec_zero`, `shiftLeftRec_succ`, 
this allows us to unfold `shiftLeft` into a circuit for bitblasting.
```

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -98,6 +98,37 @@ theorem carry_succ (i : Nat) (x y : BitVec w) (c : Bool) :
...
+
+/-- The final carry bit when computing `x + y + c` is `true` iff `x.toNat + y.toNat + c.toNat ≥ 2^w`. -/
+theorem carry_width {x y : BitVec w} :
+    carry w x y c = decide (x.toNat + y.toNat + c.toNat ≥ 2^w) := by
+  simp [carry]
+
+/--
+If `x &&& y = 0`, then addition does not overflow, and thus `(x + y).toNat = x.toNat + y.toNat`.
+-/
+theorem toNat_add_eq_toNat_add_toNat_of_and_eq_zero {x y : BitVec w} (h : x &&& y = 0#w) :
```
semorrison, line 122: `toNat_add_of_and_eq_zero` will suffice: if there's only one sensible RHS, it doesn't need to be described in the name.

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -326,4 +357,78 @@ theorem getLsb_mul (x y : BitVec w) (i : Nat) :
...
+theorem shiftLeftRec_succ {x : BitVec w₁} {y : BitVec w₂} :
+    shiftLeftRec x y (n + 1) = (shiftLeftRec x y n) <<< (y &&& twoPow w₂ (n + 1)) := by
+  simp [shiftLeftRec]
+
+/--
+If `y &&& z = 0`, `x <<< (y ||| z) = x <<< y <<< z`.
+This follows as `y &&& z = 0` implies `y ||| z = y + z`,
+and thus `x <<< (y ||| z) = x <<< (y + z) = x <<< y <<< z`.
+-/
+theorem shiftLeft_or_eq_shiftLeft_shiftLeft_of_and_eq_zero {x : BitVec w₁} {y z : BitVec w₂}
```
semorrison, line 391:
```suggestion
theorem shiftLeft_or_of_and_eq_zero {x : BitVec w₁} {y z : BitVec w₂}
```
similarly.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -436,6 +436,12 @@ theorem zeroExtend_ofNat_one_eq_ofNat_one_of_lt {v w : Nat} (hv : 0 < v) :
   have hv := Nat.testBit_one_eq_true_iff_self_eq_zero.mp hi₁
   omega
 
+/-- Truncating to width 1 produces a bitvector equal to the least significant bit. -/
+theorem truncate_one_eq_ofBool_getLsb {x : BitVec w} :
```
semorrison, line 440:
```suggestion
theorem truncate_one {x : BitVec w} :
```

semorrison:
Besides the renamings, and Alex's comment suggestion this now looks good to go!

PR #4574, opened 2024-06-27
feat: Unicode tables generation tool

jtristan:
Tool to generate Unicode property tables using the skip list data structure. The skip list, which is the data structure used in Rust, allows to create tables with very small memory footprint. The code generator is currently specialized for the numeric table, which we used as a first use case, but the skip list is general. The table search could be faster, but we leave this for future work. There is also a program to verify the table that we could use in the CI.

### DIFF
--- a/src/Init/Data/Char/UnicodeSkipList.lean
+++ b/src/Init/Data/Char/UnicodeSkipList.lean
@@ -61,7 +61,7 @@ structure UnicodePropertyTable where
 Search a character in the outer `runs` array and return a range in
 `offsets` where the character will be found and the length of the run.
 -/
-def searchRuns (table : UnicodePropertyTable) (c : Char) : Nat × Range := Id.run do
+private def searchRuns (table : UnicodePropertyTable) (c : Char) : Nat × Range := Id.run do
   let codepoint := c.toNat
   let mut i := 0
   for run in table.runs do
@@ -81,7 +81,7 @@ Search a character in the inner `offsets` array by keeping track
 of how many characters of the prefix sum of ranges' length and
 use the index parity to decide if the character has the property.
 -/
-def searchOffsets (table : UnicodePropertyTable) (c : Char) (range : Range) (pfs : Nat) : Bool := Id.run do
+private def searchOffsets (table : UnicodePropertyTable) (c : Char) (range : Range) (pfs : Nat) : Bool := Id.run do
   let codepoint := c.toNat
   let mut i := 0
   let mut prefixSum := pfs
@@ -93,6 +93,7 @@ def searchOffsets (table : UnicodePropertyTable) (c : Char) (range : Range) (pfs
       prefixSum := prefixSum + (table.offsets[j]!).toNat
   return i % 2 = 1
 
+/-- Search a character in a Unicode skip list table -/
 def search (table : UnicodePropertyTable) (c : Char) : Bool :=
   let (pfs, range) := searchRuns table c
   searchOffsets table c range pfs


src/Init/Data/Char/Tables.lean
```diff
@@ -0,0 +1,45 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jean-Baptiste Tristan
+-/
+import Init.Data.Char.UnicodeSkipList
+
```
Kha, line 7: We should put a comment here referencing how the data is generated, as well as to `Char.unicodeVersion`

src/Init/Data/String/Extra.lean
```diff
@@ -229,4 +229,19 @@ where
...
+  if s.isNatHex then
+    some <| s.foldl (fun n c =>  n*16 + (if c.isDigit then c.toNat - '0'.toNat else 10 + (c.toNat - 'A'.toNat))) 0
+  else
+    none
+
+def toNatHex! (s : String) : Nat :=
+  if s.isNatHex then
+    s.foldl (fun n c =>  n*16 + (if c.isDigit then c.toNat - '0'.toNat else 10 + (c.toNat - 'A'.toNat))) 0
+  else
+    panic! "Nat in hexadecimal expected"
```
TwoFX, line 245: You could consider trying to reduce duplication by doing something like
```suggestion
def toNatHex (s : String) (_h : s.isNatHex) : Nat :=
  s.foldl (fun n c =>  n*16 + (if c.isDigit then c.toNat - '0'.toNat else 10 + (c.toNat - 'A'.toNat))) 0

def toNatHex? (s : String) : Option Nat :=
  if h : s.isNatHex then s.toNatHex h else none

def toNatHex! (s : String) : Nat :=
  if h : s.isNatHex then s.toNatHex h else panic! "Nat in hexadecimal expected
```

script/unicode/lakefile.lean
```diff
@@ -0,0 +1,15 @@
...
+  -- add package configuration options here
+
+lean_lib «Unicode» where
+  -- add library configuration options here
+
+@[default_target]
+lean_exe "generate_tables" where
+  root := `Generation
+
+lean_exe "verify_tables" where
```
TwoFX, line 14: Depending on how exactly CI builds this, we could consider adding
```suggestion
@[default_target]
lean_exe "verify_tables" where
```
so that we catch compile errors right when we introduce them even if the verify script isn't actually executed.

script/unicode/Verification.lean
```diff
@@ -0,0 +1,49 @@
...
+  let mut failed := false
+  let table := numericTable
+  let referenceTable := referenceTable ucd property
+  for i in Range.mk 0 1114112 1 do
+    let c := Char.ofNat i
+    let ref := referenceSearch referenceTable c
+    let candidate := search table c
+    if ref ≠ candidate then
+      failed := true
+      println s!"{c.toNat} {c} {ref} {candidate}"
```
TwoFX, line 24: I think it would be good to provide a slighly more useful error message here, so that it's easier to tell at a glance what the problem is if CI fails.

script/unicode/Unicode/Parse.lean
```diff
@@ -0,0 +1,63 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jean-Baptiste Tristan
+-/
+import Unicode.Unicode
+
+open System IO FilePath Process FS Std
+
+def unicodeUrl : String := "https://www.unicode.org/Public/UCD/latest/ucd/"
```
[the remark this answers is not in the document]
jtristan: I think it is a good idea, let me propose something along those lines.

script/unicode/Unicode/Reference.lean
```diff
@@ -0,0 +1,17 @@
...
+-/
+import Unicode.Unicode
+
+def referenceTable (ucd : List UnicodeData) (property : UnicodeData → Bool) : List Nat :=
+  (ucd.filter property).map (fun ucdc => ucdc.codepoint)
+
+def referenceSearch (table : List Nat) (c : Char) : Bool :=
+  table.contains c.toNat
+
+@[simp]
```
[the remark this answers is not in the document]
jtristan: No because the search ends up reading the table from a file. Since there are only up to about 1.2 million characters, I wrote the verify program to ensure that the implementation is correct "by reflection".

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,52 @@
...
+-/
+prelude
+import Init.Data.Range
+
+open Std
+
+namespace Char.UnicodeSkipList
+
+structure UnicodePropertyTable where
+  runs : Array UInt32
```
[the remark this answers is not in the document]
jtristan: They do. This approach currently works for all the basic tables would implement. I will make sure the code fails gracefully if that were not the case. In such event, we would need to use more bits for the outer table, or, if it is a good tradeoff, use a unicode trie.

src/Init/Data/Char/Extra.lean
```diff
@@ -0,0 +1,14 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jean-Baptiste Tristan
+-/
+import Init.Data.Char.UnicodeSkipList
```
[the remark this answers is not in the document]
jtristan: I ended up fixing it before I saw this message.

script/unicode/Unicode/Parse.lean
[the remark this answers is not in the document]
jtristan: I now use the database based on a choice of major/minor/update version and generate the unicodeVersion constant accordingly.

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,52 @@
...
+  let mut prefixSum := pfs
+  for j in range do
+    if codepoint < prefixSum + (table.offsets.get! j).toNat then
+      i := j
+      break
+    else
+      prefixSum := prefixSum + (table.offsets.get! j).toNat
+  return i % 2 = 1
+
+def search (table : UnicodePropertyTable) (c : Char) : Bool :=
```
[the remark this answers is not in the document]
jtristan: I wrote some comments. Does that help?

script/unicode/Unicode/Parse.lean
```diff
@@ -0,0 +1,63 @@
...
+  else if s = "Cf" then pure <| .Other .Cf
+  else if s = "Cs" then pure <| .Other .Cs
+  else if s = "Co" then pure <| .Other .Co
+  else if s = "Cn" then pure <| .Other .Cn
+  else throw s!"Unknown General Category: {s}"
+
+def loadUnicodeData (file : FilePath) : ExceptT String IO (List UnicodeData) := do
+  let content : String ← readFile file
+  let content : List String := content.splitOn "\n"
+  let mut data := []
```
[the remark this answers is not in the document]
jtristan: Can you explain why? Given how sparse the codepoints are, I don't see why using an Array would make a big difference.

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,148 @@
...
+  let mut index := 0
+  let mut indices := [0]
+  for gap in gaps do
+    index := index + 1
+    if gap ≥ 256 then
+      indices := index * 2^21 :: indices
+  return indices.reverse
+
+/-
+Computer the list of codepoint that mark the beginning of a run
```
TwoFX, line 90:
```suggestion
Compute the list of codepoints that mark the beginning of a run
```

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,148 @@
...
+    | none, true =>
+      -- Open a range
+      let newRange : Range := { start := code , stop := code }
+      rangeOpt := some newRange
+    | none, false => ()
+
+  return ranges
+
+/-
+Turn the list or ranges into a gap encoding.
```
TwoFX, line 58:
```suggestion
Turn the list of ranges into a gap encoding.
```

src/Init/Data/String/Extra.lean
```diff
@@ -229,4 +229,16 @@ where
       have k := Nat.gt_of_not_le <| mt decide_eq_true h
       exact Nat.sub_lt_sub_left k (String.lt_next _ _)
 
+def isNatHex (s : String) : Bool :=
```
TwoFX, line 232: These should all get a docstring

src/Init/Data/Char/UnicodeVersion.lean
```diff
@@ -0,0 +1,14 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jean-Baptiste Tristan
+-/
+-- DO NOT EDIT: file generated using the scripts/unicode tool
+prelude
+import Init.Data.Nat.Basic
+
+namespace Char
+
+def unicodeVersion : Nat × Nat × Nat := (15,1,0)
```
TwoFX, line 12: This should get a docstring.

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,148 @@
...
+        rangeOpt := some {r with stop := code}
+      else
+        -- Hidden gap
+        let completedRange : Range := { start := r.start , stop := r.stop + 1 }
+        let newRange : Range := { start := code , stop := code }
+        rangeOpt := some newRange
+        ranges := completedRange :: ranges
+    | some r, false =>
+      -- Close the range
+      -- Cannot use code for range end as their may be a jump in codepoints
```
TwoFX, line 45:
```suggestion
      -- Cannot use code for range end as there may be a jump in codepoints
```

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,148 @@
...
+  let mut rangeOpt : Option Range := none
+  let mut ranges := []
+  for datapoint in ucd do
+    let code := datapoint.codepoint
+    let prop := property datapoint
+    match rangeOpt, prop with
+    | some r, true =>
+      if r.stop + 1 = code then
+        -- Extend the range
+        rangeOpt := some {r with stop := code}
```
TwoFX, line 36:
```suggestion
        rangeOpt := some { r with stop := code }
```

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,148 @@
...
+/-
+The following code creates the unicode skip list data structure for
+a given property. To understand it at a high-level, we provide an
+explanation in Init/Data/Char/UnicodeSkipList.lean
+-/
+
+/-
+Break down the sequence of codepoints into a sequence of ranges,
+alternating between ranges that satisfy the property and ranges that
+do not.
```
TwoFX, line 24: I don't think this is true? To me the code looks like it only returns the ranges that satisfy the property?

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,148 @@
...
+      let completedRange : Range := { start := r.start , stop := r.stop + 1 }
+      rangeOpt := none
+      ranges := completedRange :: ranges
+    | none, true =>
+      -- Open a range
+      let newRange : Range := { start := code , stop := code }
+      rangeOpt := some newRange
+    | none, false => ()
+
+  return ranges
```
TwoFX, line 55: This looks like it fails if the last datapoint has the property, because then there is still a pending range in `rangeOpt` that hasn't been added to `ranges` yet.

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,148 @@
...
+-/
+
+/-
+Break down the sequence of codepoints into a sequence of ranges,
+alternating between ranges that satisfy the property and ranges that
+do not.
+-/
+def explicitRanges (ucd : List UnicodeData) (property : UnicodeData → Bool) : List Range := Id.run do
+  let mut rangeOpt : Option Range := none
+  let mut ranges := []
```
TwoFX, line 28: I think it's quite confusing that a range in `rangeOpt` is a closed interval while a range in `ranges` is a half-open interval. `Std.Range` is always meant to be a half-open interval, so I think it would be good if your code used it that way.

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,148 @@
...
+    if gap ≥ 256 then
+      prefixSums := prefixSum :: prefixSums
+  return prefixSums.reverse
+
+def largeOffsetEncoding (indices prefixSums : List Nat) : Array UInt32 :=
+  let prefixSums := prefixSums ++ [1114112]
+  ((indices.zip prefixSums).map (fun (idx,pf) => (idx + pf).toUInt32)).toArray
+
+def calculateTable (ucd : List UnicodeData) (property : UnicodeData → Bool) : UnicodePropertyTable :=
+  let ranges := (explicitRanges ucd property)
```
TwoFX, line 107:
```suggestion
  let ranges := explicitRanges ucd property
```

TwoFX (commented):
Take my comments about spelling with a grain of salt, I'm not a native speaker.

script/unicode/Unicode/UnicodeSkipList.lean
jtristan: I confirm that both kind of ranges are represented (e.g. for the numeric table, the first range of 48 values or any range marked as 0). It is crucial since determining if a codepoint has the property is based on the parity of the index of the range.

script/unicode/Unicode/UnicodeSkipList.lean
jtristan: They are ranges under construction.

script/unicode/Unicode/Reference.lean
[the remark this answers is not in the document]
jtristan: Done

script/unicode/Unicode/UnicodeSkipList.lean
jtristan: In that code, the ranges are under construction, so they are half-open with respect to the previous iteration that finished, and closed with respect to the current one. Perhaps I should not use ranges has an intermediate data structure, but I think the resulting code would be more complicated. I'd rather not change it, but it is your call.

script/unicode/Unicode/UnicodeSkipList.lean
jtristan: I added some comments to clarify what is going on and what the last case does nothing (and by doing nothing, it just continues to the next codepoint while in a range of codepoints that do not satisfy the property).

src/Init/Data/Char/UnicodeSkipList.lean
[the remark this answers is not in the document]
jtristan: Done.

script/unicode/Unicode/Parse.lean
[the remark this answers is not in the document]
jtristan: Done

script/unicode/Unicode/UnicodeSkipList.lean
jtristan: I updated the comments. That property is true, but only later.

script/unicode/Verification.lean
```diff
@@ -0,0 +1,49 @@
...
+  let msg := if failed then "failed" else "succeeded"
+  println s!"Verification {msg} in {((← monoMsNow) - time).toFloat / 1000} seconds"
```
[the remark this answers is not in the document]
jtristan: Since the function currently doesn't have have any explanation, I will delay using it until later.

script/unicode/Unicode/UnicodeSkipList.lean
TwoFX: To give an example of what I mean, I ran
```
#eval (explicitRanges #[UnicodeData.mk "1" 1 (.Letter .Lu)] (fun _ => True)).length
```
and it came out to `0`. I think this is incorrect, I would have expected a list of length one to be returned.

src/Init/Data/Char/Basic.lean
```diff
@@ -15,6 +16,9 @@ Note that values in `[0xd800, 0xdfff]` are reserved for [UTF-16 surrogate pairs]
 
 namespace Char
 
+/-- Maximum valid Unicode codepoint -/
+def max : Nat := 1114111
```
TwoFX, line 20: This seems to be unused. In any case, I think `Char.max` should be of type `Char`. Also, declarations with a generic name like `max` should be marked as `protected` so that saying `open Char` won't make `Char.max` accessible as just `max` (which would be very confusing).

script/unicode/Verification.lean
```diff
@@ -0,0 +1,48 @@
...
+-/
+import Init.Data.Char.UnicodeSkipList
+import Init.Data.Char.Tables
+import Unicode.Reference
+import Unicode.Parse
+
+open System IO FilePath Process FS Std Char.UnicodeSkipList
+
+def compareTables (ucd : Array UnicodeData) (property : UnicodeData → Bool) : IO Bool := do
+
```
TwoFX, line 14:
```suggestion
```

script/unicode/Verification.lean
```diff
@@ -0,0 +1,48 @@
...
+  let mut failed := false
+  let table := numericTable
+  let referenceTable := referenceTable ucd property
+  for i in [:1114112] do
```
TwoFX, line 18: Perhaps you introduced `Char.max` in order to use it here?

script/unicode/README.md
```diff
@@ -0,0 +1,5 @@
+# Unicode
+
+This tool is used to generate Unicode tables for fast and memory efficient lookup of character properties. It uses a skip list (inspired from Rust).
```
TwoFX, line 3: Is there some page you could link to that talks about the Rust implementation?

script/unicode/Unicode/Reference.lean
```diff
@@ -0,0 +1,16 @@
...
+-/
+import Unicode.Unicode
+
+def referenceTable (ucd : Array UnicodeData) (property : UnicodeData → Bool) : Array Nat :=
+  (ucd.filter property).map fun ucdc => ucdc.codepoint
+
+def referenceSearch (table : Array Nat) (c : Char) : Bool :=
+  table.contains c.toNat
+
+def reference (ucd : Array UnicodeData) (property : UnicodeData → Bool) (c : Char) : Bool :=
```
TwoFX, line 14: I think this is unused? It's probably best to delete it because calling it is probably never a good idea given that it generated the entire table from scratch each time.

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,157 @@
...
+      -- a range of codepoints that do have the property, so continue
+      ()
+
+  return ranges
+
+/-
+Turn the list of ranges into a gap encoding.
+-/
+def mergeRanges (ranges : List Range) : List Nat := Id.run do
+  let flat := ranges.foldl (init := []) fun acc => fun range => range.start :: range.stop :: acc
```
TwoFX, line 70:
```suggestion
  let flat := ranges.foldl (init := []) fun acc range => range.start :: range.stop :: acc
```

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,157 @@
...
+  content := content ++ "Released under Apache 2.0 license as described in the file LICENSE.\n"
+  content := content ++ "Authors: Jean-Baptiste Tristan\n"
+  content := content ++ "-/\n"
+  content := content ++ "-- DO NOT EDIT: file generated using the scripts/unicode tool\n"
+  content := content ++ "prelude\n"
+  content := content ++ "import Init.Data.Char.UnicodeSkipList\n"
+  content := content ++ "import Init.Data.Array.BasicAux\n"
+  content := content ++ "\n"
+  content := content ++ "namespace Char.UnicodeSkipList\n"
+  content := content ++ s!"instance {property}Table : UnicodePropertyTable where\n"
```
TwoFX, line 139: This seems to have slipped through after a previous review:
```suggestion
  content := content ++ s!"def {property}Table : UnicodePropertyTable where\n"
```

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,157 @@
...
+      -- We are in a range of codepoints that satisfy a property and the
+      -- new codepoint also satisfies the property. Two possibilities:
+      if r.stop + 1 = code then
+        -- If the new codepoint is contiguous to the range
+        -- Extend the range
+        rangeOpt := some { r with stop := code }
+      else
+        -- There's a range of unallocated codepoints
+        -- So we need two create a new range
+        let completedRange : Range := { start := r.start , stop := r.stop + 1 }
```
TwoFX, line 43:
```suggestion
        let completedRange : Range := { start := r.start, stop := r.stop + 1 }
```
(and elsewhere in this and other files)

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,157 @@
...
+  let gaps := mergeRanges ranges
+  let offsets := offsets gaps
+  let indices := indices gaps
+  let prefixSums := prefixSums gaps
+  let runs := largeOffsetEncoding indices prefixSums
+  { runs, offsets }
+
+def writeTable (property : String) (table : UnicodePropertyTable) : IO Unit := do
+  let workingDir : FilePath ← currentDir
+  let f : FilePath := join workingDir <| System.mkFilePath ["..","..","src","Init","Data","Char","Tables.lean"]
```
TwoFX, line 126:
```suggestion
  let f : FilePath := join workingDir <| System.mkFilePath ["..", "..", "src", "Init", "Data", "Char", "Tables.lean"]
```
(and elsewhere in this and other files)

src/shell/CMakeLists.txt
```diff
@@ -75,6 +75,17 @@ FOREACH(T ${LEANRUNTESTS})
...
+# LEAN UNICODE TESTS
+file(GLOB LEANRUNTESTS "${LEAN_SOURCE_DIR}/../script/unicode/Verification.lean")
+FOREACH(T ${LEANRUNTESTS})
+  if(NOT T MATCHES "\\.#")
+    GET_FILENAME_COMPONENT(T_NAME ${T} NAME)
+    add_test(NAME "leanruntest_${T_NAME}"
+             WORKING_DIRECTORY "${LEAN_SOURCE_DIR}/../script/unicode"
+             COMMAND bash -c "${TEST_VARS} ./test.sh ${T_NAME}")
+  endif()
+ENDFOREACH(T)
```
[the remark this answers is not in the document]
jtristan: The proposed code doesn't work. Also, starting with the next PR, I will have more than one test so I will need the directory part of the test.

src/Init/Data/Char/Tables.lean
```diff
@@ -0,0 +1,48 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jean-Baptiste Tristan
+-/
+-- DO NOT EDIT: file generated using the scripts/unicode tool
+prelude
+import Init.Data.Char.UnicodeSkipList
+import Init.Data.Array.BasicAux
+
+namespace Char.UnicodeSkipList
+instance numericTable : UnicodePropertyTable where
```
TwoFX, line 12: Please regenerate the table after your latest change to the generation script changing `instance` to `def`.

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,100 @@
...
+`offsets` record the length of the ranges. Large ranges of codepoints
+that do not satisfy a property appear in the offsets as 0.
+
+A key property of the inner array is that the index of range tells us
+whether the property holds for that range or not. Therefore, to find
+whether a character has a property, we first use the `runs` array
+to locate where to search in the `offsets` array. We can then locate the
+character in the offsets array to determine the parity of its position.
+
+The 32 bit integers in the outer array contain encode the following
```
TwoFX, line 39:
```suggestion
The 32 bit integers in the outer array encode the following
```

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,100 @@
...
+information. The 11 most significants bits are an index into the inner
+array, pointing to the start of a run. The 21 least significant bits
+point to the codepoint one past the run.
+
+As an example, consider some property that would hold for the set of
+codepoints: {100, 101, 103, 104, 105, 108, 109, 2000, 2001, 2002, 2004}
+
+The offsets would be: [100,5,2,2,0,4]
```
TwoFX, line 47: Is this correct? I would have expected something like `[100,2,1,3,2,2,0,3,1,1]` here. Am I misunderstanding what the offsets array represents?

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,100 @@
...
+  let mut i := 0
+  for run in table.runs do
+    let prefixSum := run.toNat % 2^21
+    if codepoint < prefixSum then
+      break
+    i := i + 1
+  let idx := i
+  let codepointStart := if idx = 0 then 0 else table.runs[idx - 1]!.toNat % 2^21
+  let rangeStart := table.runs[idx]!.toNat / 2^21
+  let rangeStop := if idx + 1 = table.runs.size then table.offsets.size else (table.runs[idx + 1]!).toNat / 2^21
```
TwoFX, line 75:
```suggestion
  let rangeStop := if idx + 1 = table.runs.size then table.offsets.size else table.runs[idx + 1]!.toNat / 2^21
```

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,100 @@
...
+  for j in range do
+    if codepoint < prefixSum + (table.offsets[j]!).toNat then
+      i := j
+      break
+    else
+      prefixSum := prefixSum + (table.offsets[j]!).toNat
+  return i % 2 = 1
+
+def search (table : UnicodePropertyTable) (c : Char) : Bool :=
+  let (pfs,range) := searchRuns table c
```
TwoFX, line 97:
```suggestion
  let (pfs, range) := searchRuns table c
```

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,166 @@
...
+  let mut prefixSums := []
+  for gap in gaps do
+    prefixSum := prefixSum + gap
+    if gap ≥ 256 then
+      prefixSums := prefixSum :: prefixSums
+  return prefixSums.reverse
+
+def largeOffsetEncoding (indices prefixSums : List Nat) : Array UInt32 :=
+  let prefixSums := prefixSums ++ [Char.max.val.toNat + 1]
+  ((indices.zip prefixSums).map fun (idx,pf) => (idx + pf).toUInt32).toArray
```
TwoFX, line 122:
```suggestion
  ((indices.zip prefixSums).map fun (idx, pf) => (idx + pf).toUInt32).toArray
```

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,100 @@
...
+Authors: Jean-Baptiste Tristan
+-/
+prelude
+import Init.Data.Range
+
+open Std
+
+namespace Char.UnicodeSkipList
+
+/-
```
TwoFX, line 13: Are you aware of the various types of comments we have in Lean? We have `/- ... -/` for normal comments, `/-! ... -/` for module documentation that will be rendered in the online documentation, and `/-- ... -/` for declaration docstrings which appear in the online documentation and also when hovering over the declaration in VS code. For an example, you can look at https://github.com/leanprover/lean4/blob/master/src/Init/Prelude.lean and the corresponding online documentation https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html. Most well-documented files have a module doc in the beginning, right after the imports and before everything else (so before the `open Std` in this case) and then potentially additional module docs for dividing large files into sections and explaining details relevant to that section. The comment you have here should almost certainly be a module doc, and you could consider turning the comments on the functions below into docstrings.

script/unicode/Unicode/FetchDatabase.lean
```diff
@@ -0,0 +1,39 @@
...
+  content := content ++ "Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.\n"
+  content := content ++ "Released under Apache 2.0 license as described in the file LICENSE.\n"
+  content := content ++ "Authors: Jean-Baptiste Tristan\n"
+  content := content ++ "-/\n"
+  content := content ++ "-- DO NOT EDIT: file generated using the scripts/unicode tool\n"
+  content := content ++ "prelude\n"
+  content := content ++ "import Init.Data.Nat.Basic\n"
+  content := content ++ "\n"
+  content := content ++ "namespace Char\n\n"
+  content := content ++ s!"def unicodeVersion : Nat × Nat × Nat := ({major},{minor},{update})\n"
```
TwoFX, line 37: I think this declaration should have a docstring.

TwoFX (commented):
Looking very good now, this should be my last round of comments. Would appreciate a review from @Kha on the CMake changes.

src/Init/Data/Char/UnicodeSkipList.lean
jtristan: You are right, I mixed up your example and mine.

src/Init/Data/Char/UnicodeSkipList.lean
jtristan: ACK for module doc. I'll add a docstring to search. I did not make the comment for searchRuns and searchOffsets docstrings because ideally, I wouldn't even want those to appear at all in the doc, they're not meant to be called directly by an end user.

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,166 @@
...
+explanation in Init/Data/Char/UnicodeSkipList.lean
+
+Note that it is implicitly assuming that `ucd` is not any value
+of type `Array UnicodeData` but the result of parsing the UnicodeData.txt
+file. Likewise, we are implicitly assuming that `property` is not an
+arbitrary Boolean-valued function but rather one that checks for some
+specific attribute.
+
+In case where such implicit assumptions are violated, the verifier
+will reject the table was incorrectly created.
```
semorrison, line 27:
```suggestion
In case where such implicit assumptions are violated, the verifier
will reject the table was incorrectly created.
```
Hard to understand here. Perhaps "will reject the table, reporting that it was incorrectly created"?

script/unicode/Unicode/UnicodeSkipList.lean
semorrison: Sorry, still catching up here, and haven't seen an answer to @TwoFX last query here.

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,101 @@
...
+
+/-!
+The unicode skip list data structure is used to represent the set of characters
+that have some property. The data structure is precomputed using the
+tool in script/unicode.
+
+We use a variant of the skip list data structure that takes advantage
+of the distribution of character properties for the common properties of
+interest. Considering the sequence of codepoints that satisfy a given
+property, they tend to appear is small ranges that are clustered. These
```
semorrison, line 17:
```suggestion
property, they tend to appear as small ranges that are clustered. These
```
(or perhaps `in`, not sure which was intended).

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,101 @@
...
+
+A key property of the inner array is that the index of range tells us
+whether the property holds for that range or not. Therefore, to find
+whether a character has a property, we first use the `runs` array
+to locate where to search in the `offsets` array. We can then locate the
+character in the offsets array to determine the parity of its position.
+
+The 32 bit integers in the outer array encode the following
+information. The 11 most significants bits are an index into the inner
+array, pointing to the start of a run. The 21 least significant bits
```
semorrison, line 37:
```suggestion
information. The 11 most significants bits are an index into the `offsets``
array, pointing to the start of a run. The 21 least significant bits
```

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,101 @@
...
+that do not satisfy a property appear in the offsets as 0.
+
+A key property of the inner array is that the index of range tells us
+whether the property holds for that range or not. Therefore, to find
+whether a character has a property, we first use the `runs` array
+to locate where to search in the `offsets` array. We can then locate the
+character in the offsets array to determine the parity of its position.
+
+The 32 bit integers in the outer array encode the following
+information. The 11 most significants bits are an index into the inner
```
semorrison, line 36:
```suggestion
information. The 11 most significants bits are an index into the `offsets`
```

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,101 @@
...
+array, pointing to the start of a run. The 21 least significant bits
+point to the codepoint one past the run.
```
semorrison, line 38:
```suggestion
contain the codepoint one past the end of that run.
```
I think "contain" is better than "point to", as it is the literal value, not an array index.

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,101 @@
...
+property, they tend to appear is small ranges that are clustered. These
+clusters are separated from each other by large ranges of codepoints that
+do not satisfy the property. We refer to such clusters as runs.
+
+The data structure takes advantage of this by representing the set
+using an array of 32 bit integers (the outer array that allows to skip)
+and an array of 8 bit integers. The integers of the outer array
+contains enough information to let us locate in which run a codepoint
+belong, and we call it `runs`. The bytes of the inner array, which we call
+`offsets` record the length of the ranges. Large ranges of codepoints
```
semorrison, line 26: Nothing in the text here explains that the numbers in `offsets` are alternately the lengths of sequences satisfying the property and lengths of sequences not satisfying the property: this is only apparent when we get to the example.

This also makes the reference to "parity" in the next paragraph impossible to interpret until later.

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,101 @@
...
+The 32 bit integers in the outer array encode the following
+information. The 11 most significants bits are an index into the inner
+array, pointing to the start of a run. The 21 least significant bits
+point to the codepoint one past the run.
+
+As an example, consider some property that would hold for the set of
+codepoints: {100, 101, 103, 104, 105, 108, 109, 2000, 2001, 2002, 2004}
+
+The offsets would be: [100,2,1,3,2,2,0,3,1,1,0]
+Before packing into a 32 bit integer, the runs would be: [(0,2000),(5,1114112)]
```
semorrison, line 44: I remain confused here. The initial `100` in `offsets` is effectively saying "the first 100 codepoints do not have this property". How does this work if the first codepoint is larger than 8 bits?

semorrison, line 44: I think it would be good if the example above already included something about the last 21 bit number always being `1114112` (that is true, right?)

semorrison: (Perhaps a 2nd example is warranted here?)

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,101 @@
...
+contains enough information to let us locate in which run a codepoint
+belong, and we call it `runs`. The bytes of the inner array, which we call
+`offsets` record the length of the ranges. Large ranges of codepoints
+that do not satisfy a property appear in the offsets as 0.
+
+A key property of the inner array is that the index of range tells us
+whether the property holds for that range or not. Therefore, to find
+whether a character has a property, we first use the `runs` array
+to locate where to search in the `offsets` array. We can then locate the
+character in the offsets array to determine the parity of its position.
```
semorrison, line 33:
```suggestion
character in the `offsets` array to determine the parity of its position.
```

src/Init/Data/Char/UnicodeSkipList.lean
semorrison, line 44: The `5` here is `[2,1,3,2,2].length`, right? It is pretty confusing (and not really explained) that the `100` is not included in the "length of the run".

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,101 @@
...
+namespace Char.UnicodeSkipList
+
+structure UnicodePropertyTable where
+  runs : Array UInt32
+  offsets : Array UInt8
+  deriving Repr, Inhabited, Nonempty
+
+/-
+Search a character in the outer `runs` array and return a range in
+`offsets` where the character will be found and the length of the run.
```
semorrison, line 62: This comment suggests that the `Nat` being returned is the length of the run, but it is actually the initial codepoint with the property in this cluster, right?

src/Init/Data/Char/UnicodeSkipList.lean
```diff
@@ -0,0 +1,101 @@
...
+private def searchOffsets (table : UnicodePropertyTable) (c : Char) (range : Range) (pfs : Nat) : Bool := Id.run do
+  let codepoint := c.toNat
+  let mut i := 0
+  let mut prefixSum := pfs
+  for j in range do
+    if codepoint < prefixSum + (table.offsets[j]!).toNat then
+      i := j
+      break
+    else
+      prefixSum := prefixSum + (table.offsets[j]!).toNat
```
semorrison, line 93: You could reduce duplication here by moving `prefixSum :=` before the `if`.

src/Init/Data/Char/UnicodeVersion.lean
```diff
@@ -0,0 +1,25 @@
+/-
+Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jean-Baptiste Tristan
+-/
+-- DO NOT EDIT: file generated using the scripts/unicode tool
```
semorrison, line 6:
```suggestion
-- DO NOT EDIT: file generated by running `lake exe generate_tables` in `scripts/unicode/`.
```

src/Init/Data/Char/UnicodeVersion.lean
```diff
@@ -0,0 +1,25 @@
...
+prelude
+import Init.Data.Nat.Basic
+
+namespace Char
+
+/--
+The `unicodeVersion` definition gives Lean's current version of Unicode in format (major,minor,update). 
+
```
semorrison, line 14: Let's get rid of the excess new lines and indenting.

src/Init/Data/Char/Basic.lean
```diff
@@ -5,6 +5,7 @@ Author: Leonardo de Moura
 -/
 prelude
 import Init.Data.UInt.Basic
+import Init.Data.Char.UnicodeVersion
```
semorrison, line 8: Perhaps better to import this at `Init/Data/Char`, as it is not actually needed here.

src/Init/Data/Char/UnicodeSkipList.lean
semorrison: Or alternatively, I think this should be `6`??

script/unicode/README.md
```diff
@@ -0,0 +1,5 @@
...
+The program is meant to be run when we want to update to a newer version of Unicode. The code generation program can be run as `lake exe generate_tables`.
+The generated table can be verified as `lake exe verify_tables`.
```
semorrison, line 5: Please include instructions that `FetchDatabase` needs to be updated for the desired Unicode version.

script/unicode/Unicode/FetchDatabase.lean
```diff
@@ -0,0 +1,46 @@
...
+  content := content ++ "prelude\n"
+  content := content ++ "import Init.Data.Nat.Basic\n"
+  content := content ++ "\n"
+  content := content ++ "namespace Char\n\n"
+  content := content ++ "/--\n"
+  content := content ++ "The `unicodeVersion` definition gives Lean's current version of Unicode in format (major,minor,update). \n
+                         New versions of Unicode are released regularly and subsequently all methods \n
+                         in the standard library depending on Unicode are updated. Therefore the behavior \n
+                          of some `Char` and `String` methods and the value of this constant changes \n
+                          over time. *This is not considered to be a breaking change.*"
```
semorrison, line 42:
```suggestion
  content := content ++ "The `unicodeVersion` definition gives Lean's current version of Unicode in format (major,minor,update). \n\
                         New versions of Unicode are released regularly and subsequently all methods \n\
                         in the standard library depending on Unicode are updated. Therefore the behavior \n\
                          of some `Char` and `String` methods and the value of this constant changes \n\
                          over time. *This is not considered to be a breaking change.*"
```

semorrison: this should prevent the unwanted indenting

script/unicode/Unicode/FetchDatabase.lean
```diff
@@ -0,0 +1,46 @@
...
+
+def download  (url : String) (file : FilePath) : IO Output := do
+  if (¬ (← file.pathExists)) then
+    output { cmd := "curl", args := #["-s", "-S", "-f", "-o", file.toString, "-L", url] }
+  else pure { exitCode := 0, stdout := "", stderr := "" }
+
+def writeUnicodeVersion : IO Unit := do
+  let workingDir : FilePath ← currentDir
+  let f : FilePath := join workingDir <| System.mkFilePath ["..", "..", "src", "Init", "Data", "Char", "UnicodeVersion.lean"]
+  let mut content := ""
```
semorrison, line 26: In fact you can write the entire thing as a single string using the `\` trick from the suggestion below.

script/unicode/Unicode/Parse.lean
```diff
@@ -0,0 +1,56 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jean-Baptiste Tristan
+-/
+import Unicode.Unicode
+import Unicode.FetchDatabase
+
+open System IO FilePath Process FS Std
+
+def mkGeneralCategory (s : String) : Except String GeneralCategory := do
+  if s = "Lu" then pure <| .Letter Letter.Lu
```
semorrison, line 12:
```suggestion
  if s = "Lu" then pure <| .Letter .Lu
```

script/unicode/Unicode/Parse.lean
```diff
@@ -0,0 +1,56 @@
...
+  else if s = "Ll" then pure <| .Letter .Ll
+  else if s = "Lt" then pure <| .Letter .Lt
```
semorrison, line 14: A `match` would be prettier than the nested if, but not really necessary.

src/Init/Data/Char/UnicodeSkipList.lean
semorrison, line 44: Let's have this example, and any other example if another is added, also run as an executable test somewhere.

(Both that constructing the skip list from the list of codepoints would produce this answer, and that compare the search function with the reference on this skiplist.)

script/unicode/Unicode/UnicodeSkipList.lean
```diff
@@ -0,0 +1,166 @@
...
+    | some r, true =>
+      -- We are in a range of codepoints that satisfy a property and the
+      -- new codepoint also satisfies the property. Two possibilities:
+      if r.stop + 1 = code then
+        -- If the new codepoint is contiguous to the range
+        -- Extend the range
+        rangeOpt := some { r with stop := code }
+      else
+        -- There's a range of unallocated codepoints
+        -- So we need two create a new range
```
semorrison, line 51:
```suggestion
        -- So we need to create a new range
```

script/unicode/lean-toolchain
```diff
@@ -0,0 +1 @@
+lean4
```
semorrison, line 1: This will rely on having a toolchain override called `lean4` set up, right? I guess this is what we need for now, but it would be nice if we can change this to a real toolchain later.

src/Init/Data/Char/Tables.lean
```diff
@@ -0,0 +1,48 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Jean-Baptiste Tristan
+-/
+-- DO NOT EDIT: file generated using the scripts/unicode tool
+prelude
+import Init.Data.Char.UnicodeSkipList
+import Init.Data.Array.BasicAux
+
+namespace Char.UnicodeSkipList
+def numericTable : UnicodePropertyTable where
```
semorrison, line 12: Could we include in this initial test also a property that is first true beyond 256?

script/unicode/lean-toolchain
jtristan: What toolchain do you recommend?

script/unicode/Unicode/UnicodeSkipList.lean
jtristan: We had a discussion offline. This example has neither a valid unicode database nor a property for which we may want to construct a table, so indeed, the code is not meaningful.

jtristan:
Thanks to the feedback, I'm wondering if I should either use a less memory efficient but less "ad hoc" data structure like a unicode trie or perhaps use a representation with pointers and without a contiguous offset array. Going back to the drawing board for now (WIP).

script/unicode/lean-toolchain
Kha: I think we can just leave out this file. Developers are expected to have their own elan configuration, such as `lean4` in a root `lean-toolchain`.

Issue #837, opened 2021-12-01
rfl not working for pattern matching on indexed inductive types


imbrem:
### Prerequisites

* [X] Put an X between the brackets on this line if you have done all of the following:
    * Checked that your issue isn't already [filed](https://github.com/leanprover/lean4/issues).
    * Reduced the issue to a self-contained, reproducible test case.

### Description

Unless I'm doing something very wrong, `rfl` does not seem to be working properly for functions defined by pattern matching on indexed inductive types, even for every simple functions

### Steps to Reproduce

The following compiles file
```lean
inductive RawWk: Type 0 where
  | id: RawWk
  | step: RawWk -> RawWk
  | lift: RawWk -> RawWk

def raw_wk_comp (ρ: RawWk) (σ: RawWk): RawWk :=
  match ρ, σ with
    | RawWk.id, σ => σ
    | RawWk.step ρ, σ => RawWk.step (raw_wk_comp ρ σ)
    | RawWk.lift ρ, RawWk.id => RawWk.lift ρ
    | RawWk.lift ρ, RawWk.step σ => RawWk.step (raw_wk_comp ρ σ)
    | RawWk.lift ρ, RawWk.lift σ => RawWk.lift (raw_wk_comp ρ σ)

theorem raw_wk_comp_id_id {ρ: RawWk}: raw_wk_comp RawWk.id ρ = ρ := rfl
```
but, when indexing the inductive type as follows
```lean
inductive Wk: Nat -> Nat -> Type 0 where
  | id: Wk n n
  | step: Wk m n -> Wk (Nat.succ m) n
  | lift: Wk m n -> Wk (Nat.succ m) (Nat.succ n)

def wk_comp {n m l: Nat} (ρ: Wk n m) (σ: Wk m l): Wk n l :=
  match n, m, l, ρ, σ with
    | _, _, _, Wk.id, σ => σ
    | _, _, _, Wk.step ρ, σ => Wk.step (wk_comp ρ σ)
    | _, _, _, Wk.lift ρ, Wk.id => Wk.lift ρ
    | _, _, _, Wk.lift ρ, Wk.step σ => Wk.step (wk_comp ρ σ)
    | _, _, _, Wk.lift ρ, Wk.lift σ => Wk.lift (wk_comp ρ σ)

theorem wk_comp_id_id {ρ: Wk m n}: wk_comp Wk.id ρ = ρ := rfl
```
I get an error

**Expected behavior:** this should type-check (it works in Agda)

[...]

PR #4575, opened 2024-06-27
feat: structural mutual recursion

nomeata:
This adds support for mutual structural recursive functions.

For now this is opt-in: The functions must have a `termination_by structural …` annotation (new since #4542) for this to work:

```lean
mutual
inductive A
  | self : A → A
  | other : B → A
  | empty
inductive B
  | self : B → B
  | other : A → B
  | empty
end

mutual
def A.size : A → Nat
  | .self a => a.size + 1
  | .other b => b.size + 1
  | .empty => 0
termination_by structural x => x
def B.size : B → Nat
  | .self b => b.size + 1
  | .other a => a.size + 1
  | .empty => 0
termination_by structural x => x
end
```

The recursive functions don’t have to be in a one-to-one relation to a set of mutually recursive inductive data types. It is possible to ignore some of the types:

```lean
def A.self_size : A → Nat
  | .self a => a.self_size + 1
  | .other _ => 0
  | .empty => 0
termination_by structural x => x
```

or have more than one function per argument type:

```lean
  def isEven : Nat → Prop
    | 0 => True
    | n+1 => ¬ isOdd n
  termination_by structural x => x

  def isOdd : Nat → Prop
    | 0 => False
    | n+1 => ¬ isEven n
  termination_by structural x => x
```

This does not include

 * Support for nested inductive data types or nested recursion
 * Inferring mutual structural recursion in the absence of `termination_by`.
 * Functional induction principles for these.
 * Mutually recursive functions that live in different universes. This may be possible,
   maybe after beefing up the `.below` and `.brecOn` functions; we can look into this some
   other time, maybe when there are concrete use cases.

### DIFF
--- a/src/Lean/Elab/PreDefinition/Structural/FunPacker.lean
+++ b/src/Lean/Elab/PreDefinition/Structural/FunPacker.lean
@@ -1,5 +1,5 @@
 /-
-Copyright (c) 2021 Microsoft Corporation. All rights reserved.
+Copyright (c) 2024 Lean FRO, LLC. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Joachim Breitner
 -/


src/Lean/Elab/PreDefinition/Basic.lean
```diff
@@ -183,16 +183,31 @@ def addAndCompilePartialRec (preDefs : Array PreDefinition) : TermElabM Unit :=
...
     Meta.forEachExpr e fun e => do
-      if e.isAppOf recFnName then
+      if e.getAppFn.isConst && recFnNames.contains e.getAppFn.constName! then
         throwError "unexpected occurrence of recursive application{indentExpr e}"
-    pure e
-  else
-    pure e
+
+/--
+Checks that all codomians have the same level, throws an error otherwise.
```
tobiasgrosser, line 196:
```suggestion
Checks that all codomains have the same level, throws an error otherwise.
```

src/Lean/Elab/PreDefinition/Structural/Basic.lean
```diff
@@ -24,16 +31,27 @@ structure RecArgInfo where
...
+  /-- The types mutually inductive with indName -/
+  indAll      : Array Name
+deriving Inhabited
 
 def RecArgInfo.recArgPos (info : RecArgInfo) : Nat :=
   info.fixedParams.size + info.pos
 
+/--
+If `xs` are the parameters of the functions (excluding fixed prefix), partitions them
+into indices and major argugument, and other parameters.
```
tobiasgrosser, line 43:
```suggestion
into indices and major arguments, and other parameters.
```

tests/lean/guessLexFailures.lean.expected.out
```diff
@@ -67,7 +67,7 @@ guessLexFailures.lean:59:6-59:7: error: fail to show termination for
   Mutual.g
   Mutual.h
 with errors
-structural recursion does not handle mutually recursive functions
+mutual structural recursion reqiures explicit `termination_by` clauses
```
tobiasgrosser, line 70:
```suggestion
mutual structural recursion requires explicit `termination_by` clauses
```

tests/lean/terminationFailure.lean.expected.out
```diff
@@ -2,7 +2,7 @@ terminationFailure.lean:7:2-7:3: error: fail to show termination for
   f.g
   f
 with errors
-structural recursion does not handle mutually recursive functions
+mutual structural recursion reqiures explicit `termination_by` clauses
```
tobiasgrosser, line 5:
```suggestion
mutual structural recursion requires explicit `termination_by` clauses
```

tobiasgrosser:
I am new to this, but is there a corresponding mathlib branch for this?

nomeata:
Yes, branch `lean-pr-testing-4575` on mathlib exists for that purpose! It is a bit behind, so not sure if it will have a valid cache.

Or you just create a project that doesn't use mathlib and use `break it? You can use `leanprover/lean4-pr-releases:pr-release-4575` as the lean-toolchain.

tobiasgrosser:
The build seems broken: https://github.com/leanprover-community/mathlib4/tree/lean-pr-testing-4575

nomeata:
Ah, this is due to #4632. I’ll fix manually.

nomeata:
@tobiasgrosser, took a while, but it seems we have a working mathlib branch.

tobiasgrosser:
I moved our codebase to this PR: https://github.com/opencompl/lean-mlir/pull/431.

It fails with `failed to generate equational theorem for 'Lets.denote'`.

nomeata:
It shouldn't, but I had to extend other pieces of the code, so maybe made a mistake there. You could try the mathlib tag `nightly-testing-2024-07-05` first to see if it is due to my PR or some other change.

tobiasgrosser:
The issue does not arise in `nightly-testing-2024-07-05`. I just pushed my update (https://github.com/opencompl/lean-mlir/pull/432) and will rebase the mutual-recursion PR as soon it is merged.

nomeata:
Thanks for the analysis, I’ll search for the bug early next week at the latest.

nomeata:
I have a fix/workaround for your code:
```diff
diff --git a/SSA/Core/Framework.lean b/SSA/Core/Framework.lean
index bf18ff1b..fb9db629 100644
--- a/SSA/Core/Framework.lean
+++ b/SSA/Core/Framework.lean
@@ -160,7 +160,7 @@ def Com.denoteLets : (com : Com d Γ eff ty) → (Γv : Valuation Γ) →
       body.denoteLets (V.snoc Ve) >>= fun V =>
       return V.cast (by simp [Com.outContext])

-def Lets.denote [DialectSignature d] [DialectDenote d]
+def Lets.denote [DialectSignature d] [DialectDenote d] {Γ₂}
     (lets : Lets d Γ₁ eff Γ₂) (Γ₁'v : Valuation Γ₁) : (eff.toMonad d.m <| Valuation Γ₂) :=
   match lets with
   | .nil => return Γ₁'v
```
without this, the implicit `Γ₂` is put before the `[DialectSignature d]`. That is part of the `Let`’s parameters, and with my code parameters must come before indices.

Why did it work before? In e461ea6517586f2440f9602950160244a1e7c3a3 I changed `hasBadParamDep?`, thinking I am fixing a bug, but maybe the new code is wrong or too restrictive. Will have to think about that.


Incidentially, since there is no `termination_by structural lets` clause, lean then tries to find another recursive argument and settles on `termination_by structural Γ₂`, and for that (for some reason maybe also worth investigating) the equational theorem generation fails.

nomeata:
Since this issue isn't about mutual recursion per se, let’s move it to https://github.com/leanprover/lean4/issues/4671, if you don't mind.

tobiasgrosser:
> I have a fix/workaround for your code:
> 
> ```diff
> diff --git a/SSA/Core/Framework.lean b/SSA/Core/Framework.lean
> index bf18ff1b..fb9db629 100644
> --- a/SSA/Core/Framework.lean
> +++ b/SSA/Core/Framework.lean
> @@ -160,7 +160,7 @@ def Com.denoteLets : (com : Com d Γ eff ty) → (Γv : Valuation Γ) →
>        body.denoteLets (V.snoc Ve) >>= fun V =>
>        return V.cast (by simp [Com.outContext])
> 
> -def Lets.denote [DialectSignature d] [DialectDenote d]
> +def Lets.denote [DialectSignature d] [DialectDenote d] {Γ₂}
>      (lets : Lets d Γ₁ eff Γ₂) (Γ₁'v : Valuation Γ₁) : (eff.toMonad d.m <| Valuation Γ₂) :=
>    match lets with
>    | .nil => return Γ₁'v
> ```
> 
> without this, the implicit `Γ₂` is put before the `[DialectSignature d]`. That is part of the `Let`’s parameters, and with my code parameters must come before indices.
> 
> Why did it work before? In [e461ea6](https://github.com/leanprover/lean4/commit/e461ea6517586f2440f9602950160244a1e7c3a3) I changed `hasBadParamDep?`, thinking I am fixing a bug, but maybe the new code is wrong or too restrictive. Will have to think about that.
> 
> Incidentially, since there is no `termination_by structural lets` clause, lean then tries to find another recursive argument and settles on `termination_by structural Γ₂`, and for that (for some reason maybe also worth investigating) the equational theorem generation fails.

This works. I also have the feeling that the issues I report may be a bit beyond what this PR should be about. Happy to see this merged.

nomeata:
> It fails with `failed to generate equational theorem for 'Lets.denote'` in Framework.lean. I

Just for the record: The error message you see here is due to #837: When recursing on an index, rather than the indexed type itself, then the function can be defined but lean struggles with proving the equational theorems.

I’ll keep this in mind as I extend inferring recursive arguments to mutual recursion.

tests/lean/run/structuralMutual.lean
```diff
@@ -0,0 +1,574 @@
...
+    (fun ⟨a₁,a₂⟩ a_ih => ⟨F (Tree.node ⟨a₁, a₂⟩) ⟨a_ih, PUnit.unit⟩, ⟨a_ih, PUnit.unit⟩⟩)
+    (fun _a _a_1 a_ih a_ih_1 => ⟨PUnit.unit, ⟨a_ih, ⟨a_ih_1, PUnit.unit⟩⟩⟩)
+    t).1
+
+-- Then the decrecursifier works just fine! (and FunInd too, see below)
+#guard_msgs in
+def Tree.size : Tree → Nat
+  | leaf => 0
+  | node (t₁, t₂) => t₁.size + t₂.size
+termination_by structural t => t
```
tobiasgrosser, line 354: @nomeata, we started using this patch and it has already made some proofs easier. Thank you!

I'm curious, do you plan to add support for nested mutuals as well? We have both nested mutuals as well as mutual functions that also use inductive that are not part of a mutual block and are interested in beta-test some of your features. Interesting to use are the following mutuals:


# Only Expr and Com are mutual inductive types, but HVector is also part of the mutual functions.
https://github.com/opencompl/lean-mlir/blob/97f33691aec49b7b939ec0452b835e93c0a18adc/SSA/Core/Framework.lean#L523

# Termination due to well-foundedness requires high heartbeats, but our mutual recursion is a bit complex. We would really like to use mutual recursion here, if possible.

https://github.com/opencompl/lean-mlir/blob/97f33691aec49b7b939ec0452b835e93c0a18adc/SSA/Core/Framework.lean#L1720

nomeata: I plan to add support for (possibly mutual) structural recursion over nested inductive data types, but not nested recursion.

In other words of you have `inductive Tree where | node : List Tree -> Tree` you will be able to define recursive functions over it, but you cannot use `List.map` but have to replicate that functionality in the mutual group.

tobiasgrosser: Lovely. Thank you for the update.

PR #4576, opened 2024-06-27
feat: add option `debug.byAsSorry true`

leodemoura:


src/Lean/Elab/BuiltinTerm.lean
```diff
@@ -157,9 +157,19 @@ private def mkTacticMVar (type : Expr) (tacticCode : Syntax) : TermElabM Expr :=
...
+  group    := "debug"
+  descr    := "replace `by ..` blocks with `sorry`"
+}
+
 @[builtin_term_elab byTactic] def elabByTactic : TermElab := fun stx expectedType? => do
   match expectedType? with
-  | some expectedType => mkTacticMVar expectedType stx
+  | some expectedType =>
+    if debug.byAsSorry.get (← getOptions) then
+      mkSorry expectedType false
```
nomeata, line 170: I wonder if this should maybe only happen when the type is a `Prop` (else it might be defining data and this break later) and contains no MVars (else the tactic might fill them in, and without that other parts of the file breaks.)?

leodemoura: Agreed.

PR #4583, opened 2024-06-28
feat: Std.HashMap

TwoFX:
### Preliminary PRs:

- [x] #4597 
- [x] #4599
- [x] #4600
- [x] #4602
- [x] #4603
- [x] #4604
- [x] #4605
- [x] #4607
- [x] #4627
- [x] #4629 

#### Lean

* `find?` -> `get?`/`getElem?`
* `find!` -> `get!`/`getElem!`
* `findD` -> `getD`
* `findEntry?` -> not implemented for now
* `insert'` -> `containsThenInsert` (order reversed in result)
* `insertIfNew` -> `getThenInsertIfNew?` (order reversed in result)
* `numBuckets` -> `Internal.numBuckets`
* `ofListWith` -> not implemented for now
* `Array.groupByKey` -> not implemented for now
* `merge` -> not implemented for now, but you can use `insertMany`

#### Batteries

* `modify` -> not implemented for now
* `mergeWith` -> not implemented for now
* `mergeWithM` -> not implemented for now

### DIFF
--- a/src/Std/Data/DHashMap/Basic.lean
+++ b/src/Std/Data/DHashMap/Basic.lean
@@ -42,6 +42,12 @@ and an array of buckets, where each bucket is a linked list of key-value pais. T
 is always a power of two. The hash map doubles its size upon inserting an element such that the
 number of elements is more than 75% of the number of buckets.
 
+The hash map uses `==` (provided by the `BEq` typeclass) to compare keys and `hash` (provided by
+the `Hashable` typeclass) to hash them. To ensure that the operations behave as expected, `==`
+should be an equivalence relation and `a == b` should imply `hash a = hash b` (see also the
+`EquivBEq` and `LawfulHashable` typeclasses). Both of these conditions are automatic if the BEq
+instance is lawful, i.e., if `a == b` implies `a = b`.
+
 These hash maps contain a bundled well-formedness invariant, which means that they cannot
 be used in nested inductive types. For these use cases, `Std.Data.DHashMap.Raw` and
 `Std.Data.DHashMap.Raw.WF` unbundle the invariant from the hash map. When in doubt, prefer
--- a/src/Std/Data/DHashMap/RawDef.lean
+++ b/src/Std/Data/DHashMap/RawDef.lean
@@ -26,6 +26,12 @@ inductive types. The well-formedness invariant is called `Raw.WF`. When in doubt
 over `DHashMap.Raw`. Lemmas about the operations on `Std.Data.DHashMap.Raw` are available in the
 module `Std.Data.DHashMap.RawLemmas`.
 
+The hash map uses `==` (provided by the `BEq` typeclass) to compare keys and `hash` (provided by
+the `Hashable` typeclass) to hash them. To ensure that the operations behave as expected, `==`
+should be an equivalence relation and `a == b` should imply `hash a = hash b` (see also the
+`EquivBEq` and `LawfulHashable` typeclasses). Both of these conditions are automatic if the BEq
+instance is lawful, i.e., if `a == b` implies `a = b`.
+
 This is a simple separate-chaining hash table. The data of the hash map consists of a cached size
 and an array of buckets, where each bucket is a linked list of key-value pais. The number of buckets
 is always a power of two. The hash map doubles its size upon inserting an element such that the
--- a/src/Std/Data/HashMap/Basic.lean
+++ b/src/Std/Data/HashMap/Basic.lean
@@ -39,6 +39,12 @@ and an array of buckets, where each bucket is a linked list of key-value pais. T
 is always a power of two. The hash map doubles its size upon inserting an element such that the
 number of elements is more than 75% of the number of buckets.
 
+The hash map uses `==` (provided by the `BEq` typeclass) to compare keys and `hash` (provided by
+the `Hashable` typeclass) to hash them. To ensure that the operations behave as expected, `==`
+should be an equivalence relation and `a == b` should imply `hash a = hash b` (see also the
+`EquivBEq` and `LawfulHashable` typeclasses). Both of these conditions are automatic if the BEq
+instance is lawful, i.e., if `a == b` implies `a = b`.
+
 These hash maps contain a bundled well-formedness invariant, which means that they cannot
 be used in nested inductive types. For these use cases, `Std.Data.HashMap.Raw` and
 `Std.Data.HashMap.Raw.WF` unbundle the invariant from the hash map. When in doubt, prefer
--- a/src/Std/Data/HashMap/Raw.lean
+++ b/src/Std/Data/HashMap/Raw.lean
@@ -37,6 +37,12 @@ inductive types. The well-formedness invariant is called `Raw.WF`. When in doubt
 over `HashMap.Raw`. Lemmas about the operations on `Std.Data.HashMap.Raw` are available in the
 module `Std.Data.HashMap.RawLemmas`.
 
+The hash map uses `==` (provided by the `BEq` typeclass) to compare keys and `hash` (provided by
+the `Hashable` typeclass) to hash them. To ensure that the operations behave as expected, `==`
+should be an equivalence relation and `a == b` should imply `hash a = hash b` (see also the
+`EquivBEq` and `LawfulHashable` typeclasses). Both of these conditions are automatic if the BEq
+instance is lawful, i.e., if `a == b` implies `a = b`.
+
 This is a simple separate-chaining hash table. The data of the hash map consists of a cached size
 and an array of buckets, where each bucket is a linked list of key-value pais. The number of buckets
 is always a power of two. The hash map doubles its size upon inserting an element such that the
--- a/src/Std/Data/HashSet/Basic.lean
+++ b/src/Std/Data/HashSet/Basic.lean
@@ -35,6 +35,12 @@ and an array of buckets, where each bucket is a linked list of keys. The number 
 is always a power of two. The hash set doubles its size upon inserting an element such that the
 number of elements is more than 75% of the number of buckets.
 
+The hash set uses `==` (provided by the `BEq` typeclass) to compare elements and `hash` (provided by
+the `Hashable` typeclass) to hash them. To ensure that the operations behave as expected, `==`
+should be an equivalence relation and `a == b` should imply `hash a = hash b` (see also the
+`EquivBEq` and `LawfulHashable` typeclasses). Both of these conditions are automatic if the BEq
+instance is lawful, i.e., if `a == b` implies `a = b`.
+
 These hash sets contain a bundled well-formedness invariant, which means that they cannot
 be used in nested inductive types. For these use cases, `Std.Data.HashSet.Raw` and
 `Std.Data.HashSet.Raw.WF` unbundle the invariant from the hash set. When in doubt, prefer
--- a/src/Std/Data/HashSet/Raw.lean
+++ b/src/Std/Data/HashSet/Raw.lean
@@ -37,6 +37,12 @@ inductive types. The well-formedness invariant is called `Raw.WF`. When in doubt
 over `HashSet.Raw`. Lemmas about the operations on `Std.Data.HashSet.Raw` are available in the
 module `Std.Data.HashSet.RawLemmas`.
 
+The hash set uses `==` (provided by the `BEq` typeclass) to compare elements and `hash` (provided by
+the `Hashable` typeclass) to hash them. To ensure that the operations behave as expected, `==`
+should be an equivalence relation and `a == b` should imply `hash a = hash b` (see also the
+`EquivBEq` and `LawfulHashable` typeclasses). Both of these conditions are automatic if the BEq
+instance is lawful, i.e., if `a == b` implies `a = b`.
+
 This is a simple separate-chaining hash table. The data of the hash set consists of a cached size
 and an array of buckets, where each bucket is a linked list of keys. The number of buckets
 is always a power of two. The hash set doubles its size upon inserting an element such that the


src/Std/Init/Option.lean
```diff
@@ -0,0 +1,79 @@
...
+  | some _, _ => rfl
+
+theorem some_get! [Inhabited α] : (o : Option α) → o.isSome → some (o.get!) = o
+  | some _, _ => rfl
+
+theorem get!_eq_getD_default [Inhabited α] (o : Option α) : o.get! = o.getD default := rfl
+
+def or : Option α → Option α → Option α
+  | o@(some _), _ => o
+  | none, o => o
```
semorrison, line 31: We've got by with `orElse` so far. Would this suffice for your purposes?

If not, please add a doc-string.

src/Std/Init/Nat.lean
```diff
@@ -0,0 +1,46 @@
...
+Authors: Markus Himmel
+-/
+prelude
+import Init.Data.Nat.Bitwise.Lemmas
+import Init.RCases
+
+namespace Nat
+
+theorem lt_of_eq_of_lt {n m k : Nat} : n = m → m < k → n < k :=
+  fun h₁ h₂ => h₁ ▸ h₂
```
semorrison, line 13: This is already in `_root_` for any type with `[LT]`.

src/Std/Init/Nat.lean
```diff
@@ -0,0 +1,46 @@
...
+
+theorem le_of_testBit {n m : Nat} (h : ∀ i, n.testBit i = true → m.testBit i = true) : n ≤ m := by
```
semorrison, line 15: I guess we need to work out where the split is between `Init` and `Std`, but this can be adjusted later.

src/Std/Data/HashSet/Lemmas.lean
```diff
@@ -0,0 +1,274 @@
...
+theorem size_remove_le [EquivBEq α] [LawfulHashable α] {k : α} : (m.remove k).size ≤ m.size :=
+  HashMap.size_remove_le
+
+@[simp]
+theorem fst_containsThenInsert {k : α} : (m.containsThenInsert k).1 = m.insert k :=
+  ext HashMap.fst_containsThenInsert
+
+@[simp]
+theorem snd_containsThenInsert {k : α} : (m.containsThenInsert k).2 = m.contains k :=
+  HashMap.snd_containsThenInsert
```
semorrison, line 270: I think these should be `containsThenInsert_fst`. It's not entirely sensible, but this seems to be the convention for structure projections.

src/Std/Data/HashSet/Lemmas.lean
```diff
@@ -0,0 +1,274 @@
...
+  HashMap.mem_iff_contains
+
+@[simp] theorem not_mem_empty {a : α} {c} : ¬a ∈ (empty c : HashSet α) :=
+  HashMap.not_mem_empty
+
+@[simp] theorem contains_emptyc {a : α} : (∅ : HashSet α).contains a = false :=
+  HashMap.contains_emptyc
+
+@[simp] theorem not_mem_emptyc {a : α} : ¬a ∈ (∅ : HashSet α) :=
+  HashMap.not_mem_emptyc
```
semorrison, line 189: Perhaps shuffle the order here, so each `_empty` / `_emptyc` pair is adjacent?

src/Std/Data/HashSet/Lemmas.lean
```diff
@@ -0,0 +1,274 @@
...
+prelude
+import Std.Data.HashMap.Lemmas
+import Std.Data.HashSet.Basic
+
+/-!
+# Hash set lemmas
+
+This file contains lemmas about `Std.Data.HashSet.Raw` and `Std.Data.HashSet`. Most of the lemmas require
+`EquivBEq α` and `LawfulHashable α` for the key type `α`. The easiest way to obtain these instances
+is to provide an instance of `LawfulBEq α`.
```
semorrison, line 15: Does it make sense to move all lemmas about `Raw` into a separate file? This makes it easier to direct users to the "outermost" law.

src/Std/Init/Nat.lean
TwoFX: Sorry, I should have made this more clear: my plan is to move everything in the `Std/Init` folder into `Init` as part of separate PRs that are to be merged before this one. The only reason why I haven't done this yet is that I wanted the PR up by the end of the week and I ran out of week.

src/Std/Data/HashSet/Lemmas.lean
TwoFX: I guess the question is: do you care more about having `empty`/`emptyc` pairs adjacent or having `contains`/`mem` pairs adjacent? I'm happy with either.

src/Std/Data/HashSet/Lemmas.lean
TwoFX: Sorry, I don't understand the second sentence, could you elaborate?

src/Std/Data/HashMap/Basic.lean
```diff
@@ -0,0 +1,299 @@
...
+
+instance [BEq α] [Hashable α] {m : Type w → Type w} : ForIn m (HashMap α β) (α × β) where
+  forIn m init f := m.forIn (fun a b acc => f (a, b) acc) init
+
+@[inline] def toList (m : Raw α β) : List (α × β) :=
+  DHashMap.Raw.Const.toList m.inner
+
+@[inline] def toArray (m : Raw α β) : Array (α × β) :=
+  DHashMap.Raw.Const.toArray m.inner
+
```
semorrison, line 271: I think these should be talking about HashMap, not Raw?

src/Std/Data/HashSet/Lemmas.lean
semorrison: Sorry, typo on my part (layer -> law)

I just meant that since most users will not actually need to know about `Raw`, splitting these files in two would make the code more accessible.

src/Std/Data/HashSet/Lemmas.lean
semorrison: In either case, `mem_iff_contains` needs to move up one. :-)

For the other question, I'm happy with either now, so no change required.

semorrison:
Some functions don't have any theorems (e.g. `toList`, `insertMany`, `ofList`). This is of course fine, but we should document this somewhere. Perhaps I would suggest collecting their definitions together and putting them under a `/-! ... -/` comment explaining that they don't have theorems at this point.

src/Std/Init/Option.lean
TwoFX: I think we want this in the interest of provding a very deep API for all of the basic types. Of course you can get by with `orElse`, but having `or` is just a bit more convenient sometimes. Rust has both (Java only has `orElse`).  It's very canonical, it makes `Option α` into a monoid for all `α`.

Docstrings will follow as promised :)

src/Std/Data/HashSet/Lemmas.lean
TwoFX: Yes, good idea. We also get a bit more build parallelism as an added bonus.

src/Std/Init/Option.lean
TwoFX: See #4600.

src/Std/Data/DHashMap/RawDef.lean
```diff
@@ -0,0 +1,31 @@
...
+
+set_option linter.missingDocs true
+set_option autoImplicit false
+
+universe u v
+
+namespace Std.DHashMap
+
+/-- Dependent hash maps without bundled well-formedness invariant. Suitable for use in nested
+inductive types. The well-formedness invariant is called `Raw.WF`. -/
```
david-christiansen, line 24:
```suggestion
/--
Dependent hash maps without a bundled well-formedness invariant, suitable for use in nested
inductive types. The well-formedness invariant is called `Raw.WF`. For most uses, the bundled
version `Std.DHashMap` is better.

This is a simple separate-chaining hash table. The data of the hash map consists of a cached
size and an array of buckets, where each bucket is an `AssocList α β` (which is the same as
a `List ((a : α) × β a)` but with one less level of indirection). The number of buckets is always a
power of two. The hash map doubles its size upon inserting an element such that the number of
elements is more than 75% of the number of buckets.
-/
```

src/Std/Data/DHashMap/Basic.lean
```diff
@@ -0,0 +1,230 @@
...
+
+universe u v w
+
+variable {α : Type u} {β : α → Type v} {δ : Type w} {m : Type w → Type w} [Monad m]
+
+namespace Std
+
+open DHashMap.Internal DHashMap.Internal.List
+
+/-- Dependent hash maps. -/
```
david-christiansen, line 37:
```suggestion
/--
Dependent hash maps.

This is a simple separate-chaining hash table. The data of the hash map consists of a cached
size and an array of buckets, where each bucket is an `AssocList α β` (which is the same as
a `List ((a : α) × β a)` but with one less level of indirection). The number of buckets is always a
power of two. The hash map doubles its size upon inserting an element such that the number of
elements is more than 75% of the number of buckets.

These hash maps contain a bundled well-formedness invariant, which means that they cannot
be used in nested inductive types. For these use cases, `Std.Data.DHashMap.Raw` and `Std.Data.DHashMap.Raw.WF` unbundle the invariant from the hash map. When in doubt, prefer
`DHashMap` over `DHashMap.Raw`.
-/
```

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+use the empty collection notations `∅` and `{}` to create an empty hash map with the default
+capacity. -/
+@[inline] def empty (capacity := 8) : Raw α β :=
+  (Raw₀.empty capacity).1
+
+instance : EmptyCollection (Raw α β) where
+  emptyCollection := empty
+
+/-- Insert the given mapping into the map, replacing an existing mapping for the key if there is
+one. -/
```
david-christiansen, line 56:
```suggestion
/--
Inserts the given mapping into the map, replacing an existing mapping for the key if there is
one.
-/
```

Indicative mood, plus the format we decided on for multi-line docstring comment indicator placement

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+@[inline] def insert [BEq α] [Hashable α] (m : Raw α β) (a : α) (b : β a) : Raw α β :=
+  if h : 0 < m.buckets.size then
+    (Raw₀.insert ⟨m, h⟩ a b).1
+  else m -- will never happen for well-formed inputs
+
+/-- If there is no mapping for the given key, insert the given mapping into the map. Otherwise,
+return the map unaltered. -/
```
david-christiansen, line 63:
```suggestion
/--
If there is no mapping for the given key, inserts the given mapping into the map. Otherwise,
returns the map unaltered.
-/
```

src/Std/Data/DHashMap/RawDef.lean
TwoFX: I'm not a fan of explicitly mentioning `AssocList` here, because that type is supposed to be an internal implementation detail. Would you be fine with "The data of the hash map consists of a cached
size and an array of buckets, where each bucket is a linked list of key-value pairs"?

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+@[inline] def insertIfNew [BEq α] [Hashable α] (m : Raw α β) (a : α) (b : β a) : Raw α β :=
+  if h : 0 < m.buckets.size then
+    (Raw₀.insertIfNew ⟨m, h⟩ a b).1
+  else m -- will never happen for well-formed inputs
+
+/-- Equivalent to (but potentially faster than) calling `contains` followed by `insert`. -/
```
david-christiansen, line 69:
```suggestion
/--
Checks whether a key is present in a map, and unconditionally inserts a value for the key.

Equivalent to (but potentially faster than) calling `contains` followed by `insert`.
-/
```

src/Std/Data/DHashMap/Raw.lean
TwoFX: Oh no, for some reason I remembered that we settled on the exact opposite convention. I should have double-checked. I will go over all of them and change them.

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+@[inline] def containsThenInsert [BEq α] [Hashable α] (m : Raw α β) (a : α) (b : β a) :
+    Bool × Raw α β:=
+  if h : 0 < m.buckets.size then
+    let ⟨replaced, ⟨r, _⟩⟩ := Raw₀.containsThenInsert ⟨m, h⟩ a b
+    ⟨replaced, r⟩
+  else (false, m) -- will never happen for well-formed inputs
+
+/-- Equivalent to (but potentially faster than) calling `get?` followed by `insertIfNew`. -/
```
david-christiansen, line 77:
```suggestion
/--
Checks whether a key is present in a map, and inserts a value for the key if not.

Equivalent to (but potentially faster than) calling `get?` followed by `insertIfNew`.
-/
```

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+@[inline] def getThenInsertIfNew? [BEq α] [Hashable α] [LawfulBEq α] (m : Raw α β) (a : α)
+    (b : β a) : Option (β a) × Raw α β :=
+  if h : 0 < m.buckets.size then
+    let ⟨previous, ⟨r, _⟩⟩ := Raw₀.getThenInsertIfNew? ⟨m, h⟩ a b
+    ⟨previous, r⟩
+  else (none, m) -- will never happen for well-formed inputs
+
+/-- Equivalent to (but potentially faster than) calling `contains` followed by `insertIfNew`. -/
```
david-christiansen, line 85:
```suggestion
/--
Checks whether a key is present in a map and inserts a value for the key if it was not found.

If the returned `Bool` is `true`, then the returned map is unaltered. If the `Bool` is `false`, then
the returned map has a new value inserted.

Equivalent to (but potentially faster than) calling `contains` followed by `insertIfNew`.
-/
```

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+Uses the `LawfulBEq` instance to cast the retrieved value to the correct type. -/
+@[inline] def get! [BEq α] [Hashable α] [LawfulBEq α] (m : Raw α β) (a : α) [Inhabited (β a)] :
+    β a :=
+  if h : 0 < m.buckets.size then
+    Raw₀.get! ⟨m, h⟩ a
+  else default -- will never happen for well-formed inputs
+
+/-- Removes the mapping for the given key if it exists.
+
+Uses the `LawfulBEq` instance to cast the retrieved value to the correct type. -/
```
david-christiansen, line 146:
```suggestion
/-- Removes the mapping for the given key if it exists. -/
```

I think this was a copy-paste oversight.

src/Std/Data/DHashMap/RawDef.lean
david-christiansen: Absolutely. The thought is mostly to give readers a rough cost model, and the most important part of that is that it's based on an array.

david-christiansen: (and the second-most important part is the resizing/capacity strategy)

src/Std/Data/DHashMap/Internal/Defs.lean
```diff
@@ -0,0 +1,414 @@
...
+  -- a "load factor" of 0.75 is the usual standard for hash maps
+  capacity * 4 / 3
+
+def toListModel (buckets : Array (AssocList α β)) : List ((a : α) × β a) :=
+  buckets.data.bind AssocList.toList
+
+@[inline] def computeSize (buckets : Array (AssocList α β)) : Nat :=
+  buckets.foldl (fun d b => d + b.length) 0
+
+abbrev Raw₀ (α : Type u) (β : α → Type v) :=
```
david-christiansen, line 162:
```suggestion
/--
This is an internal implementation detail of the DHashMap library and should not be used externally.
-/
abbrev Raw₀ (α : Type u) (β : α → Type v) :=
```

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+@[inline] def Const.getD [BEq α] [Hashable α] (m : Raw α (fun _ => β)) (a : α) (fallback : β) : β :=
+  if h : 0 < m.buckets.size then
+    Raw₀.Const.getD ⟨m, h⟩ a fallback
+  else fallback -- will never happen for well-formed inputs
+
+/-- Tries to retrieve the mapping for the given key, panicking if no such mapping is present. -/
+@[inline] def Const.get! [BEq α] [Hashable α] [Inhabited β] (m : Raw α (fun _ => β)) (a : α) : β :=
+  if h : 0 < m.buckets.size then
+    Raw₀.Const.get! ⟨m, h⟩ a
+  else default -- will never happen for well-formed inputs
```
david-christiansen, line 180:
```suggestion
  else default
```
Sure the empty map is well-formed, and I can call `get!` on it

TwoFX: The empty map has a non-empty buckets array consisting of empty buckets :) So on the empty map we are taking the true-branch of the `if`.

TwoFX: See also the theorem `WF.size_buckets_pos` in that file.

src/Std/Data/DHashMap/Internal/Defs.lean
TwoFX: My philosophy with these internal types was: they must live in the `Internal` folder and they must not be reachable in one step via go-to-definition from user case. So if a user finds themselves looking at this definition, they should already know that they're looking at an implementation detail. Do you still think it would make sense to annotate all the types in the `Internal` folders?

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+
+/-- Folds the given function over the mappings in the hash map in some order. -/
+@[inline] def foldlM (f : δ → (a : α) → β a → m δ) (init : δ) (b : Raw α β) : m δ :=
+  b.buckets.foldlM (fun acc l => l.foldlM f acc) init
+
+/-- Folds the given function over the mappings in the hash map in some order. -/
+@[inline] def foldl (f : δ → (a : α) → β a → δ) (init : δ) (b : Raw α β) : δ :=
+  Id.run (b.foldlM f init)
+
+/-- Folds the given function over the mappings in the hash map in some order. -/
```
david-christiansen, line 232:
```suggestion
/--
Carries out a monadic action on each mapping in the hash map in some order.
-/
```

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+    Raw₀.map f ⟨m, h⟩
+  else ∅ -- will never happen for well-formed inputs
+
+/-- Removes all mappings of the hash map for which the given function returns `false`. -/
+@[inline] def filter (f : (a : α) → β a → Bool) (m : Raw α β) : Raw α β :=
+  if h : 0 < m.buckets.size then
+    Raw₀.filter f ⟨m, h⟩
+  else ∅ -- will never happen for well-formed inputs
+
+/-- Folds the given function over the mappings in the hash map in some order. -/
```
david-christiansen, line 224:
```suggestion
/--
Monadically computes a value by folding the given function over the mappings in the hash
map in some order.
-/
```

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+@[inline] def foldlM (f : δ → (a : α) → β a → m δ) (init : δ) (b : Raw α β) : m δ :=
+  b.buckets.foldlM (fun acc l => l.foldlM f acc) init
+
+/-- Folds the given function over the mappings in the hash map in some order. -/
+@[inline] def foldl (f : δ → (a : α) → β a → δ) (init : δ) (b : Raw α β) : δ :=
+  Id.run (b.foldlM f init)
```
david-christiansen, line 230: Are these really left folds, given that the order is unspecified? Should they be called `foldM` and `fold`, respectively?

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+/-- Returns a list of all keys present in the hash map in some order. -/
+@[inline] def keys (m : Raw α β) : List α :=
+  m.foldl (fun acc k _ => k :: acc) []
+
+/-- Returns an array of all keys present in the hash map in some order. -/
+@[inline] def keysArray (m : Raw α β) : Array α :=
+  m.foldl (fun acc k _ => acc.push k) #[]
+
+/-- Returns a list of all values present in the hash map in some order. -/
+@[inline] def values {β : Type v} (m : Raw α (fun _ => β)) : List β :=
```
david-christiansen, line 271: Why is there `keysArray` but not `valuesArray`?

src/Std/Data/DHashMap/Raw.lean
TwoFX: Yes, you're right. Not sure what I was thinking.

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+  else m -- will never happen for well-formed inputs
+
+@[inline, inherit_doc Raw.insertMany] def Const.insertManyUnit [BEq α] [Hashable α] {ρ : Type w}
+    [ForIn Id ρ α] (m : Raw α (fun _ => Unit)) (l : ρ) : Raw α (fun _ => Unit) :=
+  if h : 0 < m.buckets.size then
+    (Raw₀.Const.insertManyUnit ⟨m, h⟩ l).1
+  else m -- will never happen for well-formed inputs
+
+/-- Creates a hash map from a list of mappings. -/
+@[inline] def ofList [BEq α] [Hashable α] {ρ : Type w} [ForIn Id ρ ((a : α) × β a)] (l : ρ) :
```
david-christiansen, line 295: The docstring and name don't seem to match the type signature here. Should the type be specialized?

src/Std/Data/DHashMap/Raw.lean
TwoFX: The historical answer is: `keysArray` powers `HashSet.toArray`, so it was impossible to forget it, while my main use case for `values` was specifying prop-valued invariants for proving things. I'll add `valuesArray`.

src/Std/Data/DHashMap/Raw.lean
TwoFX: We definitely want a function with that name for recognizability, even though it is just a special case of `insertMany`. So the remaining question is: if the type signature of the function can be generalized without any disadvantages, is there a reason not to do it? If I actually call the function with a list, then the compiler specializes the IR to be just as efficient as if the function had been declared for lists from the start. The only reason I see why you would want to explicitly specialize to lists here is that Loogle might find the function on a query like `List, Std.HashMap`?

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+instance [Repr α] [(a : α) → Repr (β a)] : Repr (Raw α β) where
+  reprPrec m prec := Repr.addAppParen ("Std.DHashMap.Raw.ofList " ++ reprArg m.toList) prec
+
+end Unverified
+
+section WF
+
+/-- Well-formedness predicate for hash maps. Users of `DHashMap` will not need to interact with
+this. Users of `DHashMap.Raw` will need to provide proofs of `WF` to lemmas and should use lemmas
+like `WF.empty` and `WF.insert` to show that map operations preserve well-formedness. The
```
david-christiansen, line 322: Is the lemma name always the same as the operation name? If so, it'd be good to say so here.

src/Std/Data/HashMap/Basic.lean
```diff
@@ -0,0 +1,193 @@
...
+prelude
+import Std.Data.DHashMap.Basic
+
+set_option linter.missingDocs true
+set_option autoImplicit false
+
+/-!
+# Hash maps
+
+This file develops the type `Std.Data.HashMap` of dependent hash maps.
```
david-christiansen, line 15:
```suggestion
This module develops the type `Std.Data.HashMap` of non-dependent hash maps. Dependent hash
maps are defined in `Std.Data.DHashMap`.
```

src/Std/Data/HashMap/Basic.lean
```diff
@@ -0,0 +1,193 @@
...
+nested inductive types.
+-/
+
+universe u v w
+
+variable {α : Type u} {β : Type v}
+
+namespace Std
+
+/-- Hash maps. -/
```
david-christiansen, line 33:
```suggestion
/--
Hash maps.

This is a simple separate-chaining hash table. The data of the hash map consists of a cached
size and an array of buckets, where each bucket is a linked list of key-value pairs. The number
of buckets is always a power of two. The hash map doubles its size upon inserting an element
such that the number of elements is more than 75% of the number of buckets.

These hash maps contain a bundled well-formedness invariant, which means that they cannot
be used in nested inductive types. For these use cases, `Std.Data.HashMap.Raw` and `Std.Data.HashMap.Raw.WF` unbundle the invariant from the hash map. When in doubt, prefer
`HashMap` over `HashMap.Raw`.

Dependent hash maps, in which keys may occur in their values' types, are available as
`Std.Data.DHashMap`.
-/
```

src/Std/Data/HashMap/Basic.lean
```diff
@@ -0,0 +1,193 @@
...
+structure HashMap (α : Type u) (β : Type v) [BEq α] [Hashable α] where
+  /-- Internal implementation detail of the hash map -/
+  inner : DHashMap α (fun _ => β)
```
david-christiansen, line 36: Should the constructor be made private?

TwoFX: Unfortunately, we need to call this constructor in the file `HashMap/AdditionalOperations.lean`. We can make it private for `HashMap.Raw`, `HashSet` and `HashSet.Raw`, though.

src/Std/Data/DHashMap/Internal/Defs.lean
david-christiansen: I'm not sure, to be honest.

The concern I have is that they might show up in a proof state or the infoview, and get followed from there.

On the balance, I think it's better to have it on each type, because users do all sorts of things we don't expect, and a bit of redundant communication is better than a misunderstanding. After all, the goal is to keep internals internal so that user code doesn't break; it's not to have a thing to point at and say "I told you so" if it breaks and they complain.

But this is an opinion I hold lightly.

src/Std/Data/HashMap/Basic.lean
TwoFX: Actually, this is not true, none of the constructors can be private because `Lemmas.lean`/`RawLemmas.lean` needs to refer to them.

src/Std/Data/DHashMap/Raw.lean
david-christiansen: Gotta get that linter :-)

src/Std/Data/HashMap/Basic.lean
```diff
@@ -0,0 +1,193 @@
...
+  let ⟨replaced, r⟩ := m.inner.containsThenInsert a b
+  ⟨replaced, ⟨r⟩⟩
+
+@[inline, inherit_doc DHashMap.containsThenInsertIfNew] def containsThenInsertIfNew [BEq α]
+    [Hashable α] (m : HashMap α β) (a : α) (b : β) : Bool × HashMap α β :=
+  let ⟨replaced, r⟩ := m.inner.containsThenInsertIfNew a b
+  ⟨replaced, ⟨r⟩⟩
+
+/-- Equivalent to (but potentially faster than) calling `get?` followed by `insertIfNew`. -/
+@[inline] def getThenInsertIfNew? [BEq α] [Hashable α] (m : HashMap α β) (a : α) (b : β) :
```
david-christiansen, line 66: Why is this not `@[inherit_doc]`?

src/Std/Data/HashMap/Basic.lean
```diff
@@ -0,0 +1,193 @@
...
+@[inline, inherit_doc DHashMap.Const.insertMany] def insertMany [BEq α] [Hashable α] {ρ : Type w}
+    [ForIn Id ρ (α × β)] (m : HashMap α β) (l : ρ) : HashMap α β :=
+  ⟨DHashMap.Const.insertMany m.inner l⟩
+
+@[inline, inherit_doc DHashMap.Const.insertManyUnit] def insertManyUnit [BEq α] [Hashable α]
+    {ρ : Type w} [ForIn Id ρ α] (m : HashMap α Unit) (l : ρ) : HashMap α Unit :=
+  ⟨DHashMap.Const.insertManyUnit m.inner l⟩
+
+@[inline, inherit_doc DHashMap.Const.ofList] def ofList [BEq α] [Hashable α] {ρ : Type w}
+    [ForIn Id ρ (α × β)] (l : ρ) : HashMap α β :=
```
david-christiansen, line 194: This type signature also seems to not match the name (see comment on underlying functions)

src/Std/Data/DHashMap/Raw.lean
```diff
@@ -0,0 +1,438 @@
...
+    (Raw₀.insertMany ⟨m, h⟩ l).1
+  else m -- will never happen for well-formed inputs
+
+@[inline, inherit_doc Raw.insertMany] def Const.insertMany {β : Type v} [BEq α] [Hashable α]
+    {ρ : Type w} [ForIn Id ρ (α × β)] (m : Raw α (fun _ => β)) (l : ρ) : Raw α (fun _ => β) :=
+  if h : 0 < m.buckets.size then
+    (Raw₀.Const.insertMany ⟨m, h⟩ l).1
+  else m -- will never happen for well-formed inputs
+
+@[inline, inherit_doc Raw.insertMany] def Const.insertManyUnit [BEq α] [Hashable α] {ρ : Type w}
```
david-christiansen, line 337: I think this one doesn't have a docstring to inherit.

Is this a reasonable thing to have in a public API? When should a user prefer this over a `HashSet`?

src/Std/Data/HashMap/Basic.lean
TwoFX: Ah, that would be because the docstring for `DHashMap.getThenInsertIfNew?` would talk about the `[LawfulBEq]` instance and the docstring for `DHashMap.Const.getThenInsertIfNew?` would talk about the `Const` prefix which no longer exists at the `HashMap` layer. I forgot to mention the `LawfulBEq` instance in the `DHashMap.getThenInsertIfNew?` docstring, I will fix that.

src/Std/Data/HashMap/Raw.lean
```diff
@@ -0,0 +1,255 @@
...
+universe u v w
+
+variable {α : Type u} {β : Type v}
+
+namespace Std
+
+namespace HashMap
+
+/-- Hash maps without bundled well-formedness invariant. Suitable for use in nested
+inductive types. The well-formedness invariant is called `Raw.WF`. -/
```
david-christiansen: I think a note about the representation and use case would also be good here, in parallel with `DHashMap.Raw`.

TwoFX: Already done :)

src/Std/Data/HashSet/Basic.lean
```diff
@@ -0,0 +1,154 @@
...
+set_option linter.missingDocs true
+set_option autoImplicit false
+
+universe u v
+
+variable {α : Type u}
+
+namespace Std
+
+/-- Hash sets. -/
```
david-christiansen: Please see comments on the other datatypes :)

src/Std/Data/HashSet/Basic.lean
```diff
@@ -0,0 +1,154 @@
...
+structure HashSet (α : Type u) [BEq α] [Hashable α] where
+  /-- Internal implementation detail of the hash set. -/
+  inner : HashMap α Unit
+
+namespace HashSet
+
+/-- Creates a new empty hash set. The optional parameter `capacity` can be supplied to presize the
+map so that it can hold the given number of elements without reallocating. It is also possible to
+use the empty collection notations `∅` and `{}` to create an empty hash map with the default
```
david-christiansen, line 56:
```suggestion
use the empty collection notations `∅` and `{}` to create an empty hash set with the default
```

src/Std/Data/DHashMap/Raw.lean
TwoFX: Did you mean `associate` or `associated`? My non-native english is failing me here :)

src/Std/Data/HashSet/Basic.lean
```diff
@@ -0,0 +1,154 @@
...
+
+instance [BEq α] [Hashable α] : EmptyCollection (HashSet α) where
+  emptyCollection := empty
+
+/-- Insert the given element into the set. -/
+@[inline] def insert [BEq α] [Hashable α] (m : HashSet α) (a : α) : HashSet α :=
+  ⟨m.inner.insertIfNew a ()⟩
+
+/-- Equivalent to (but potentially faster than) calling `contains` followed by `insert`. -/
+@[inline] def containsThenInsert [BEq α] [Hashable α] (m : HashSet α) (a : α) : Bool × HashSet α :=
```
david-christiansen: This is useful when the `BEq` and `Hashable` instances are lawful, but coarser than propositional equality (e.g. terms identified up to alpha-equivalence). Should there be a note to that effect here? Otherwise, readers might wonder what this operation is for.

src/Std/Data/HashSet/Basic.lean
```diff
@@ -0,0 +1,154 @@
...
+  let ⟨replaced, r⟩ := m.inner.containsThenInsertIfNew a ()
```
david-christiansen, line 75: But actually here it looks like it doesn't insert the new element. This strikes me as a footgun.

src/Std/Data/DHashMap/Raw.lean
TwoFX: Yes, done.

src/Std/Data/HashSet/Basic.lean
```diff
@@ -0,0 +1,154 @@
...
+capacity. -/
+@[inline] def empty [BEq α] [Hashable α] (capacity := 8) : HashSet α :=
+  ⟨HashMap.empty capacity⟩
+
+instance [BEq α] [Hashable α] : EmptyCollection (HashSet α) where
+  emptyCollection := empty
+
+/-- Insert the given element into the set. -/
+@[inline] def insert [BEq α] [Hashable α] (m : HashSet α) (a : α) : HashSet α :=
+  ⟨m.inner.insertIfNew a ()⟩
```
david-christiansen, line 67: Shouldn't this unconditionally insert the element?

Here's a silly example:
```
structure Mod5 where
  val : Nat 

instance : BEq Mod5 where
  beq x y := x.val % 5 == y.val % 5

instance : Hashable Mod5 where
  hash x := hash x.val
```
Putting these in a `HashSet Mod5` would mean that I could keep track of counters, but only one for each equivalence class. If I were doing this, then I would expect that inserting a value would update the value for its equivalence class, rather than being a no-op.

This is a silly way to write that code, and `HashMap (Fin 5) Nat` or just a five-tuple would much more clearly show the intention, but I hope it was illustrative. At the very least, I think there should be a prominent warning if the `BEq` will be interpreted as indicating that all elements of the `BEq`'s equivalence class are interchangeable.

src/Std/Data/DHashMap/Raw.lean
david-christiansen: Got it, thanks :)

src/Std/Data/DHashMap/Raw.lean
david-christiansen: Search is important. The other reason to want the monomorphised version is that it can help type inference and improve error messages.

If I pass a value to `ofList`, I'm presumably doing it because I believe it to be a list; the monomorphised function shares this belief with Lean. This could solve metavariables elsewhere in my program and make it typecheck, for example.

src/Std/Data/HashMap/Basic.lean
david-christiansen: Makes sense :)

src/Std/Data/HashSet/Basic.lean
TwoFX: I think this is the way it usually works with hash sets. [Rust](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.insert) and [Java](https://docs.oracle.com/en%2Fjava%2Fjavase%2F22%2Fdocs%2Fapi%2F%2F/java.base/java/util/HashSet.html#add(E)) do the same thing, except the boolean they return is flipped from how we do it.

src/Std/Data/DHashMap/Raw.lean
david-christiansen: My native English is what's failing. It should be "associated".

src/Std/Data/HashSet/Basic.lean
TwoFX: The operation you want is called `replace` in the Rust hash set and if different from `insert` which behaves like the `insert` I define. The Java hash set does not have this operation at all. I'm not saying that wanting this is unreasonable, but I think that we should keep the `insert` operation as is and add a new operation `replace` in the future if we want it.

src/Std/Data/HashSet/Basic.lean
david-christiansen: I think the docstring should mention this, then, just in case - I think there's perhaps a higher likelihood of someone trying to be clever here.

src/Std/Data/HashSet/Basic.lean
david-christiansen: That seems reasonable to me

src/Std/Data/DHashMap/Raw.lean
TwoFX: Yes, the only reason this exists is to be called by `HashSet.insertMany`. I have updated the docstring to say as much.

src/Std/Data/HashSet/Basic.lean
TwoFX: I have updated the relevant docstrings to reflect this. Do you think the updated docstrings are clear enough?

src/Std/Data/HashSet/Basic.lean
TwoFX: In my experience, it's not so unusual to have this use case of "I want to insert this, and also know whether I have just inserted this for the first time". Rust and Java certainly also support this use case in their `insert`/`add` methods and there were countless occurrences of this in the Java code base I worked on at my previous job. So I don't think we need to say more here.

src/Std/Data/DHashMap/Internal/Defs.lean
TwoFX: Okay, I've put a "nothing to see here" docstring on every definition for now.

TwoFX:
@david-christiansen Thank you very much for the detailed review! I'll give you a chance to confirm that all of your comments have been addressed and then I think it's time to get this merged.

src/Std/Data/HashSet/Basic.lean
david-christiansen: Yes! I think I would still be surprised, but then I would read the docstring, reflect, and think "ok, that makes sense" and feel good.

src/Std/Data/DHashMap/Raw.lean
david-christiansen: LGTM!

david-christiansen (approved):

david-christiansen:
This is great, thanks!

PR #3811, opened 2024-03-30
fix: nix: add platform dependent flag to lib target

lenianiva:
Closes #3810

PR #4587, opened 2024-06-29
fix: Add linking of `-lStd` back into nix build flags on darwin

lenianiva:
Adds linkage to `Std` so the build behaviour on darwin is in line with linux

I'm not sure why linking with `Std` is needed. I deleted it in the previous patch https://github.com/leanprover/lean4/pull/3811 and Lean still builds and runs. @tydeu mentioned this issue so I created this PR.


nomeata:
Maybe because `Std` is new and not used much?

lenianiva:
> Maybe because `Std` is new and not used much?

but if its needed by the lean executable why would compiling without it not give symbol not found errors?

tydeu:
@lenianiva Because [Std is currently empty](https://github.com/leanprover/lean4/tree/f31d4dc1284ad66a40cd3751ad014de4f45aca8d/src/Std). That will change soon, though.

lenianiva:
> @lenianiva Because [Std is currently empty](https://github.com/leanprover/lean4/tree/f31d4dc1284ad66a40cd3751ad014de4f45aca8d/src/Std). That will change soon, though.

will Std4 be merged into lean4?

tydeu:
@lenianiva The previous std4 was renamed to [batteries](https://github.com/leanprover-community/batteries). The core `Std` will be something different.

PR #4590, opened 2024-06-29
fix: typo hearbeats -> heartbeats

kmill:
Reported [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/is.20.60trace.2Eprofiler.2EuseHeartbeats.60.20a.20thing.3F/near/447950838).


nomeata (approved):
Do we have to worry about backward compat here, or can we just fix it? (This is probably obscure enough that we can just change it.)

PR #4595, opened 2024-06-30
feat: `Simp.Config.implicitDefEqProofs`

leodemoura:
This PR implements `Simp.Config.implicitDefEqsProofs`. When `true` (default: `true`), `simp` will **not** create a proof term for a rewriting rule associated with an `rfl`-theorem. Rewriting rules are provided by users by annotating theorems with the attribute `@[simp]`. If the proof of the theorem is just `rfl` (reflexivity), and `implicitDefEqProofs := true`, `simp` will **not** create a proof term which is an application of the annotated theorem.

The default setting does change the existing behavior. Users can use `simp -implicitDefEqProofs` to force `simp` to create a proof term for `rfl`-theorems. This can positively impact proof checking time in the kernel.

This PR also fixes an issue in the `split` tactic that has been exposed by this feature. It was looking for `split` candidates in proofs and implicit arguments. See new test for issue exposed by the previous feature.

### DIFF
--- a/src/Lean/Meta/Tactic/SplitIf.lean
+++ b/src/Lean/Meta/Tactic/SplitIf.lean
@@ -28,19 +28,28 @@ structure Context where
 
 unsafe abbrev FindM := ReaderT Context $ StateT (PtrSet Expr) MetaM
 
-private def isCandidate (env : Environment) (ctx : Context) (e : Expr) : Bool := Id.run do
-  if ctx.exceptionSet.contains e then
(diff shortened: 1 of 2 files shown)


nomeata:
I got thrown off by this feature being documented in our releases, but silently not working. https://github.com/leanprover/lean4/pull/5751 thus removes that currently misleading docstring, whenever this PR lands it should re-add the docstring.

PR #4599, opened 2024-07-01
feat: additional lemmas for Option

TwoFX:
Split from #4583

Mathlib has `isSome_map'` but calls it `isSome_map`.

### DIFF
--- a/src/Init/Data/Option/Lemmas.lean
+++ b/src/Init/Data/Option/Lemmas.lean
@@ -4,6 +4,7 @@ Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Mario Carneiro
 -/
 prelude
+import Init.Data.Option.BasicAux
 import Init.Data.Option.Instances
 import Init.Classical
 import Init.Ext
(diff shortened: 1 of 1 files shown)


src/Init/Data/Option/Lemmas.lean
```diff
@@ -145,6 +161,12 @@ theorem map_eq_some : f <$> x = some b ↔ ∃ a, x = some a ∧ f a = b := map_
 @[simp] theorem map_eq_none' : x.map f = none ↔ x = none := by
   cases x <;> simp only [map_none', map_some', eq_self_iff_true]
 
+theorem isSome_map {x : Option α} : (f <$> x).isSome = x.isSome := by
+  cases x <;> simp
+
+@[simp] theorem isSome_map' {x : Option α} : (x.map f).isSome = x.isSome := by
+  cases x <;> simp
```
semorrison, line 168: Let's not now, but I wonder if we should reverse all these names, and preserve the `'` for the lemmas that go via the monad.

PR #4600, opened 2024-07-01
feat: Option.or

TwoFX:
Split from #4583

### DIFF
--- a/src/Init/Data/Option/Basic.lean
+++ b/src/Init/Data/Option/Basic.lean
@@ -93,7 +93,7 @@ instance : OrElse (Option α) where
 
 /-- If the first argument is `some a`, returns `some a`, otherwise returns the second argument.
 This is similar to `<|>`/`orElse`, but it is strict in the second argument. -/
-@[always_inline, inline] def or : Option α → Option α → Option α
+@[always_inline, macro_inline] def or : Option α → Option α → Option α
   | some a, _ => some a
   | none,   b => b
 


src/Init/Data/Option/Basic.lean
```diff
@@ -89,6 +89,11 @@ Implementation of `OrElse`'s `<|>` syntax for `Option`.
 instance : OrElse (Option α) where
   orElse := Option.orElse
 
+/-- If the first argument is `some a`, returns `some a`, otherwise returns the second argument. -/
+@[always_inline, inline] def or : Option α → Option α → Option α
```
digama0, line 93: should this be `macro_inline`? That way the second argument will not be evaluated if not needed, similar to `cond`

src/Init/Data/Option/Basic.lean
```diff
@@ -89,6 +89,11 @@ Implementation of `OrElse`'s `<|>` syntax for `Option`.
 instance : OrElse (Option α) where
   orElse := Option.orElse
 
+/-- If the first argument is `some a`, returns `some a`, otherwise returns the second argument. -/
```
semorrison, line 92: Could the doc-string explicitly reference `orElse` (and vice versa). I worry still about people being confused about which one they should be using.

src/Init/Data/Option/Basic.lean
TwoFX: I think if this behavior is desired then `orElse` is a better choice. I have updated the docstring to say as much.

digama0: I agree, but it still seems unnecessary and inconsistent to have eager evaluation behavior when all other similar functions (like `Bool.or`) have macro inline behavior.

TwoFX: I guess we don't lose anything by making it `macro_inline`. Done.

PR #4607, opened 2024-07-01
feat: EquivBEq and LawfulHashable classes

TwoFX:
Split from #4583

There are two open questions, opinions appreciated:

- Should this material be part of `Init` or `Std`?
- Should the typeclasses be in the `Std` namespace?


src/Std/remove-after-adding-other-files
semorrison, line 1: Still to remove!

semorrison:
I think this should be in the root namespace, at least for now.

Let's keep this in Init.

TwoFX:
Okay, I have moved the classes to Init.

PR #4608, opened 2024-07-01
feat: create ci workflow on lake new/init

### LINKED ISSUE #4606: `lean-action` template for `lake new/init`
### Proposal

Add a template to `lake new/init` to allow users to automatically include a default continuous integration workflow that uses [lean-action](https://github.com/leanprover/lean-action).

With the `auto-config` improvement introduced by leanprover/lean-action#61, a call to `lean-action` with no inputs should be a reasonable default:

```yaml
name: CI

on:
  push:
    branches: ["main"]
  pull_request:
    branches: ["main"]
  workflow_dispatch:

jobs:
  build:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v4
      # uses lean action with all default input values
      - uses: leanprover/lean-action@v1-beta
 ```

Here are a few questions regarding implementation and user experience we need to align on:
- **How will the template interact with the existing templates and lake options?** Currently, users must choose one main template (std, exe, lib, math) and one file format (.lean, .toml) for the lakefile. Does adding a long option (e.g. `lake new --lean-action-workflow`) instead of a template make sense? @tydeu I thought you might be the best person to answer this.
- **How should the workflow triggers be configured by default?** I was thinking pushes and pull requests to the main branch. Alternatively,  we could keep it simple and trigger on a push it to any branch.

### Community Feedback

[link to relevant Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/lean-action/near/446192244)

Related to #3950 

### PR DESCRIPTION
**Pull request by austinletson**
Draft of adding ci workflow using lean-action on `lake new/init`

This PR is currently missing lake options for the user to control this feature.

Closes #4606

### DIFF
--- a/src/lake/Lake/CLI/Init.lean
+++ b/src/lake/Lake/CLI/Init.lean
@@ -315,7 +315,6 @@ def validatePkgName (pkgName : String) : LogIO PUnit := do
   if pkgName.toLower ∈ ["init", "lean", "lake", "main"] then
     error "reserved package name"
 
-
 def init (name : String) (tmp : InitTemplate) (lang : ConfigLang) (env : Lake.Env) (cwd : FilePath := ".") : LogIO PUnit := do
   let name ← id do
     if name == "." then

### CODE REVIEW

File: src/lake/Lake/CLI/Init.lean [hunk 1]
```diff
@@ -156,6 +156,24 @@ def mathToolchainBlobUrl : String :=
...
   "https://github.com/leanprover-community/mathlib4/blob/master/lean-toolchain"
 
+def leanActionWorkflowContents :=
+s!"name: Lean Action CI
+
+on:
+  push:
+    branches: [\"main\"]
+  pull_request:
+    branches: [\"main\"]
```
semorrison, line 166: Ooh, I don't think we can afford to be opinionated on `main` / `master`.

src/lake/Lake/CLI/Init.lean
```diff
@@ -156,6 +156,24 @@ def mathToolchainBlobUrl : String :=
 def mathToolchainUrl : String :=
   "https://github.com/leanprover-community/mathlib4/blob/master/lean-toolchain"
 
+def leanActionWorkflowContents :=
+s!"name: Lean Action CI
```
semorrison, line 160: Isn't this just a plain string? The `s!` doesn't seem necessary.

src/lake/Lake/CLI/Init.lean
austinletson: That makes sense. I removed the branch names so it will run on any push/ pull request.

src/lake/Lake/CLI/Init.lean
austinletson: Yes, `s!` was unnecessary. Removed.

src/lake/Lake/CLI/Init.lean
```diff
@@ -287,6 +303,17 @@ def validatePkgName (pkgName : String) : LogIO PUnit := do
   if pkgName.toLower ∈ ["init", "lean", "lake", "main"] then
     error "reserved package name"
 
+def createLeanActionWorkflow (dir : FilePath) : LogIO PUnit := do
+  logInfo "creating lean-action CI workflow"
+  let workflowDir := dir / ".github" / "workflows"
+  let workflowFile := workflowDir / "lean_action_ci.yml"
+  if (← workflowFile.pathExists) then
+    logWarning "lean-action CI workflow already exists"
```
tydeu, line 311:
```suggestion
    logVerbose "lean-action CI workflow already exists"
```

`lake init` / `lake new` do not generally perform step-by-step logging. However, adding some is probably not a bad idea. If we do log here, since logging it is not currently expected, I think this (and below) should use `logVerbose`. That way, it is only printed with `-v` (or on an error).

src/lake/Lake/CLI/Init.lean
```diff
@@ -287,6 +303,17 @@ def validatePkgName (pkgName : String) : LogIO PUnit := do
...
+    return
+  IO.FS.createDirAll workflowDir
+  IO.FS.writeFile workflowFile leanActionWorkflowContents
+  logInfo s!"created lean-action CI workflow at '{workflowFile}'"
```
tydeu, line 315:
```suggestion
  logVerbose s!"created lean-action CI workflow at '{workflowFile}'"
```

See above comment on `logWarning`.

src/lake/Lake/CLI/Init.lean
```diff
@@ -300,6 +327,7 @@ def init (name : String) (tmp : InitTemplate) (lang : ConfigLang) (env : Lake.En
   validatePkgName name
   IO.FS.createDirAll cwd
   initPkg cwd (stringToLegalOrSimpleName name) tmp lang env
+  createLeanActionWorkflow cwd
```
tydeu, line 330: This should be part of `initPkg` rather than duplicated in `init` and `new`.

src/lake/Lake/CLI/Init.lean
```diff
@@ -309,3 +337,4 @@ def new (name : String) (tmp : InitTemplate) (lang : ConfigLang)  (env : Lake.En
   let dir := cwd / dirName
   IO.FS.createDirAll dir
   initPkg dir name tmp lang env
+  createLeanActionWorkflow dir
```
tydeu, line 340: See above comment in `init`.

tydeu (changes requested):
Thanks! Looks pretty good to me. 🎉  I just have a few minor tweaks I would like to see.

src/lake/Lake/CLI/Init.lean
austinletson: Makes sense. Done :+1:

src/lake/Lake/CLI/Init.lean
austinletson: Ah, okay. Done :+1:

tydeu (approved):
LGTM! Give me a ping (and undraft it) when/if you are satisfied with the PR and I will merge it.

austinletson:
@tydeu Thanks for the suggestions! I tested it locally again to make sure it works as intended. Ready to merge :rocket:

PR #4612, opened 2024-07-01
feat: omega error message: normalize constraint order

nomeata:
using the order as it comes out of the `HashMap` led to annying test
suite output variations. Moreover, sorting by the canonical order leads
to messages that are probably easier to digest as a user.


nomeata:
I’ll go ahead and merge this, as this is blocking me from getting another PR past CI.

PR #4625, opened 2024-07-02
chore: make constructor-as-variable test more robust

TwoFX:
The test tended to fail every time the number of structures in `Init` changes, which turns out to be quite often.


david-christiansen (approved):
This is a good change!

Issue #4610, opened 2024-07-01
Metavariables in `have`'s type should not leak into body


Kha:
As a side effect of #4308, the tactic `have : $ty := by $tacs*` now requires all mvars in `ty` to be solved before elaborating `tacs`, which is not the case for other `have` forms but is the case for e.g. `theorem`. Everyone seems to agree that `have` should consistently behave like `theorem` in this respect, which should also help with parallelism.

Reported [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Rules.20of.20metavariable.20unification/near/448255713).

PR #4626, opened 2024-07-02
feat: consistent elaboration of tactic `have ... by`

Kha:
This implements a much smaller version of #4610: we make sure that any *tactic* `have ... by` does not allow postponing mvars in the type, but do not change any other `have`


Kha:
(seems to have created some issues with partial syntax trees in the test. Resolving this is not highest prio for me at the moment)

PR #4627, opened 2024-07-02
feat: additional lemmas for arrays

TwoFX:
Split from #4583

### DIFF
--- a/src/Init/Data/Array/TakeDrop.lean
+++ b/src/Init/Data/Array/TakeDrop.lean
@@ -9,7 +9,7 @@ import Init.Data.List.TakeDrop
 
 namespace Array
 
-theorem exists_of_update (self : Array α) (i d h) :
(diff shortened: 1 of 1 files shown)


src/Init/Data/Array/TakeDrop.lean
```diff
@@ -0,0 +1,17 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Markus Himmel
+-/
+prelude
+import Init.Data.Array.Lemmas
+import Init.Data.List.TakeDrop
+
+namespace Array
+
+theorem exists_of_update (self : Array α) (i d h) :
```
semorrison, line 12: Should this be `exists_of_uset`?

TwoFX, line 12:
```suggestion
theorem exists_of_uset (self : Array α) (i d h) :
```

PR #4637, opened 2024-07-03
feat: safe exponentiation

leodemoura:
Summary:
- Adds configuration option `exponentiation.threshold`
- An expression `b^n` where `b` and `n` are literals is not reduced by `whnf`, `simp`, and `isDefEq` if `n > exponentiation.threshold`.

Motivation: prevents system from becoming irresponsive and/or crashing without memory.

TODO: improve support in the kernel. It is using a hard-coded limit for now.

### DIFF
--- a/src/kernel/type_checker.cpp
+++ b/src/kernel/type_checker.cpp
@@ -595,6 +595,18 @@ template<typename F> optional<expr> type_checker::reduce_bin_nat_op(F const & f,
     return some_expr(mk_lit(literal(nat(f(v1.raw(), v2.raw())))));
 }
 
+#define ReducePowMaxExp 1<<24 // TODO: make it configurable
+
+optional<expr> type_checker::reduce_pow(expr const & e) {
+    expr arg1 = whnf(app_arg(app_fn(e)));
+    expr arg2 = whnf(app_arg(e));
+    if (!is_nat_lit_ext(arg2)) return none_expr();
+    nat v1 = get_nat_val(arg1);
+    nat v2 = get_nat_val(arg2);
+    if (v2 > nat(ReducePowMaxExp)) return none_expr();
+    return some_expr(mk_lit(literal(nat(nat_pow(v1.raw(), v2.raw())))));
+}
+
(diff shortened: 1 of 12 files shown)


eric-wieser:
> * if `n > exponentiation.threshold`

What do you think about using `(max (log2 a) 1) * n > exponentiation.threshold` instead? That would also prevent memory being consumed when `a` is itself large.

src/kernel/type_checker.cpp
```diff
@@ -622,7 +634,7 @@ optional<expr> type_checker::reduce_nat(expr const & e) {
         if (f == *g_nat_add) return reduce_bin_nat_op(nat_add, e);
         if (f == *g_nat_sub) return reduce_bin_nat_op(nat_sub, e);
         if (f == *g_nat_mul) return reduce_bin_nat_op(nat_mul, e);
-        if (f == *g_nat_pow) return reduce_bin_nat_op(nat_pow, e);
+        if (f == *g_nat_pow) return reduce_pow(e);
```
eric-wieser, line 637: I'm a little worried that this doesn't actually help; I think when I tried this patch, instead of:

* Lean asks GMP to compute a very large natural, using lots of CPU or perhaps running out of heap memory

The result is now:

* Lean refuses to expand `a ^ n` via nat reduction
* `a ^ n` is expanded to `a * a ^ (n - 1)`, allocating small objects and filling up the stack
* Lean refuses to expand `a ^ (n - 1)` via nat reduction
* ...
* Recursion uses lots of CPU, and hits a stack overflow

eric-wieser: Here's the smallest test I could construct:
```
import Lean
import Qq

open Qq

theorem foo : 2 ^ 100000000 = (2 ^ 10000) ^ 10000 := by
  run_tac do
    (← Lean.Elab.Tactic.getMainGoal).assign
      q(Eq.refl ((2 ^ 10000) ^ 10000))
```
Apologies for the Qq use

eric-wieser: I think this was fixed by #4934!

PR #4646, opened 2024-07-03
fix: improve `synthAppInstances`

leodemoura:
This is an auxiliary procedured used by `rw` and `apply` tactics. It synthesizes pending type class instances.
The new test contains an example where it failed. The comment at `synthAppInstances.step` explains why, and the fix.

### DIFF
--- a/src/Lean/Meta/Tactic/Apply.lean
+++ b/src/Lean/Meta/Tactic/Apply.lean
@@ -58,7 +58,7 @@ where
      {M N : outParam (Type _)} (φ : outParam (M → N))
      (X Y : outParam (Type _)) [SMul M X] [SMul N Y] [FunLike F X Y] : Prop where
   ```
-  The right-hand-side of the equation does not bind `N`. Thus, `SMul N Y` cannot
+  The left-hand-side of the equation does not bind `N`. Thus, `SMul N Y` cannot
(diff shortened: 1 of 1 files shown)


src/Lean/Meta/Tactic/Apply.lean
```diff
@@ -29,18 +29,73 @@ private def throwApplyError {α} (mvarId : MVarId) (eType : Expr) (targetType :
...
+           [SMul M X] [SMul N Y] [FunLike F X Y] [MulActionSemiHomClass F φ X Y]
+           (f : F) (c : M) (x : X) : DFunLike.coe f (c • x) = φ c • DFunLike.coe f x
+  ```
+  and `MulActionSemiHomClass` is defined as
+  ```
+  class MulActionSemiHomClass (F : Type _)
+     {M N : outParam (Type _)} (φ : outParam (M → N))
+     (X Y : outParam (Type _)) [SMul M X] [SMul N Y] [FunLike F X Y] : Prop where
+  ```
+  The right-hand-side of the equation does not bind `N`. Thus, `SMul N Y` cannot
```
nomeata, line 61: Really right hand side, or should this be left hand side?

leodemoura: Thanks. Fixed.

PR #4647, opened 2024-07-03
chore: add parallelism fallback for macOS on build

alok:
1 less thing to think about.

### DIFF
--- a/doc/make/index.md
+++ b/doc/make/index.md
@@ -27,9 +27,9 @@ Setting up a basic parallelized release build:
 git clone https://github.com/leanprover/lean4
 cd lean4
 cmake --preset release
-make -C build/release -j$(nproc || sysctl -n hw.logicalcpu)  # may still fail on some alternative shells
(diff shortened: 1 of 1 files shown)


doc/make/index.md
```diff
@@ -27,7 +27,7 @@ Setting up a basic parallelized release build:
 git clone https://github.com/leanprover/lean4
 cd lean4
 cmake --preset release
-make -C build/release -j$(nproc)  # see below for macOS
+make -C build/release -j$(nproc || sysctl -n hw.logicalcpu)  # may still fail on some alternative shells
 ```
 You can replace `$(nproc)`, which is not available on macOS and some alternative shells, with the desired parallelism amount.
```
Kha, line 32:
````suggestion
make -C build/release -j$(nproc || sysctl -n hw.logicalcpu)
```
You can replace `$(nproc || sysctl -n hw.logicalcpu)` with the desired parallelism amount.
````

PR #4658, opened 2024-07-05
feat: .below and .brecOn for nested inductive

nomeata:
We now get `.below` and `.brecOn` definitions for nested inductives.

No surprises in the implementation: the kernel already gives us suitable `.rec_1` etc. recursors, and our construction follows the structure of this recursor.


src/Lean/Meta/Constructions/BRecOn.lean
```diff
@@ -277,117 +304,127 @@ fun {α} {motive} t F_1 => (
...
+  unless indVal.isRec do return
+  if ← isPropFormerType indVal.type then return
+
+  let recName := mkRecName indName
+  let brecOnName := if ind then mkBInductionOnName indName else mkBRecOnName indName
+  mkBRecOnFromRec recName ind indVal.isReflexive indVal.numParams indVal.all.toArray brecOnName
+
+  -- If this is the first inductive in a mutual group with nested inductives,
+  -- generate the constructions for the nested inductives now
+  if indVal.all[0]! = indName then
```
tobiasgrosser, line 421:
```suggestion
  -- If this is the first inductive in a mutual group with nested inductives,
  -- generate the constructions for the nested inductives now.
  if indVal.all[0]! = indName then
```

nomeata:
After finally being able to test this in #4733 it’s good to go.

PR #4659, opened 2024-07-05
feat: `Process.tryWait`

hargoniX:
Inspired by: https://doc.rust-lang.org/stable/std/process/struct.Child.html#method.try_wait.
My use case for this API is in LeanSAT: We have a long running external process (the SAT solver) in a tactic. Due to interactivity I also need to check whether the tactic might have been cancelled yet. This means I want to periodically poll both the cancellation token as well as the process status until either of them are finished. However the current `Process.wait` API blocks so this is not possible. This new API does not block.


hargoniX:
We should probably also let this run on the Windows CI before merging if that is possible?

PR #4668, opened 2024-07-06
fix: explicitly initialize `Std` in `lean_initialize`

Kha:
Fixes the stage 2 build, which runs with `prefer_native=true`

### DIFF
--- a/src/Std/Data/DHashMap.lean
+++ b/src/Std/Data/DHashMap.lean
@@ -5,5 +5,6 @@ Authors: Markus Himmel
 -/
 prelude
 import Std.Data.DHashMap.Basic
+import Std.Data.DHashMap.RawLemmas
 import Std.Data.DHashMap.Lemmas
 import Std.Data.DHashMap.AdditionalOperations


src/Std/Data/DHashMap.lean
```diff
@@ -5,5 +5,6 @@ Authors: Markus Himmel
 -/
 prelude
 import Std.Data.DHashMap.Basic
+import Std.Data.DHashMap.RawLemmas
```
Kha, line 8: @TwoFX Just checking, was orphaning the file intentional?

TwoFX: Yes, the idea was that in the "normal" use case where you just want to use `HashMap` and not `HashMap.Raw` you can just `import Std.Data.HashMap` and import only the files that are needed for `HashMap`.

In fact, the same thing should be true for all three of `DHashMap`, `HashMap` and `HashSet`, so I'm surprised that just changing this for `DHashMap` made any difference.

TwoFX: Ah, maybe it's because `DHashMap.RawLemmas` has macros in it, unlike `HashMap.RawLemmas` and `HashSet.RawLemmas`?

Kha: Yes, exactly. It's a limitation of the bootstrapping setup, if you will. But note that a Lake build would also skip orphan files.

Perhaps the import should be moved to Std.Data instead then? Unlikely people will import that and then complain that it contains too much stuff.

TwoFX: Sounds good to me, I'll open a PR.

PR #4669, opened 2024-07-06
chore: fix "max dynamic symbols" metric

Kha:
As we do not build multiple shared libraries on non-Windows anymore, count the max exported symbols per static library instead. Unfortunately, this still does seem to match the number on Windows.


Kha (commented):
!bench

Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
!bench

PR #4674, opened 2024-07-06
feat: detailed feedback on `decide` tactic failure

kmill:
When the `decide` tactic fails, it can try to give hints about the failure:
- It tells you which `Decidable` instances it unfolded, by making use of the diagnostics feature.
- If it encounters `Eq.rec`, it gives you a hint that one of these instances was likely defined using tactics.
- If it encounters `Classical.choice`, it hints that you might have classical instances in scope.
- During this, it tries to process `Decidable.rec`s and matchers to pin blame on a particular instance that failed to reduce.

This idea comes from discussion with Heather Macbeth [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Decidable.20with.20structures/near/449409870).

### DIFF
--- a/src/Lean/Elab/Tactic/ElabTerm.lean
+++ b/src/Lean/Elab/Tactic/ElabTerm.lean
@@ -439,7 +439,7 @@ private partial def blameDecideReductionFailure (inst : Expr) : MetaM Expr := do
             m!"\n\n\
             Hint: Reduction got stuck on '▸' ({MessageData.ofConstName ``Eq.rec}), \
             which suggests that one of the '{MessageData.ofConstName ``Decidable}' instances is defined using tactics such as 'rw' or 'simp'. \
-            To avoid tactics, make use of declarations such as \
+            To avoid tactics, make use of functions such as \
             '{MessageData.ofConstName ``inferInstanceAs}' or '{MessageData.ofConstName ``decidable_of_decidable_of_iff}' \
             to alter a proposition."
           else if reason.isAppOf ``Classical.choice then
--- a/tests/lean/run/decideTactic.lean
+++ b/tests/lean/run/decideTactic.lean
@@ -74,7 +74,7 @@ After unfolding the instances 'baz' and 'instDecidableNice', reduction got stuck
   ⋯ ▸ inferInstance
 
 Hint: Reduction got stuck on '▸' (Eq.rec), which suggests that one of the 'Decidable' instances is
-defined using tactics such as 'rw' or 'simp'. To avoid tactics, make use of declarations such as
+defined using tactics such as 'rw' or 'simp'. To avoid tactics, make use of functions such as
 'inferInstanceAs' or 'decidable_of_decidable_of_iff' to alter a proposition.
 -/
 #guard_msgs in
@@ -97,7 +97,7 @@ at the 'Decidable' instance
   ⋯ ▸ inferInstance
 
 Hint: Reduction got stuck on '▸' (Eq.rec), which suggests that one of the 'Decidable' instances is
-defined using tactics such as 'rw' or 'simp'. To avoid tactics, make use of declarations such as
+defined using tactics such as 'rw' or 'simp'. To avoid tactics, make use of functions such as
 'inferInstanceAs' or 'decidable_of_decidable_of_iff' to alter a proposition.
 -/
 #guard_msgs in


src/Lean/Elab/Tactic/ElabTerm.lean
```diff
@@ -385,12 +385,48 @@ private def preprocessPropToDecide (expectedType : Expr) : TermElabM Expr := do
...
+          let unfoldMsg :=
+            if unfoldedInsts.isEmpty then
+              MessageData.nil
+            else
+              let instances := if unfoldedInsts.size == 1 then "instance" else "instances"
+              m!"after unfolding the {instances}{indentD <| MessageData.andList unfoldedInsts.toList}\n"
+          let hint :=
+            if r.isAppOf ``Eq.rec then
+              m!"\n\n\
+              Hint: reduction got stuck on '▸' (Eq.rec), which indicates that one of the instances \
```
hrmacbeth, line 409: I think it would be a little more beginner-friendly to change the wording from "instances" to "decidability instances".

hrmacbeth (commented):
This PR seems to me like a very good idea!

kmill:
@david-christiansen Would you mind taking a look at the wordings of the error messages in the test files?

One thing that might be confusing is that the expression it reports as being to blame might not be the original instance. We can't use the word "subexpression" here because it's only a subexpression after enough reduction. Instead, I went for being slightly vague with the phrase "reduction got stuck at".

tests/lean/run/decideTactic.lean
```diff
@@ -0,0 +1,102 @@
...
+since its Decidable instance
+  Classical.propDecidable unknownProp
+did not reduce to isTrue or isFalse.
+
+After unfolding the instance Classical.propDecidable, reduction got stuck at the Decidable instance
+  Classical.choice ⋯
+
+Hint: Reduction got stuck on Classical.choice, suggesting that there are "classical" instances, due
+to for example an 'open scoped Classical' command. The 'decide' tactic works by evaluating a
+decision procedure via reduction, but such classical instances cannot be evaluated.
```
david-christiansen, line 42: I think this message might be a bit hard to figure out for non-experts, and while the double-quotes make sense, I worry that they split the audiences a bit. Here's a suggestion:

```suggestion
Hint: Reduction got stuck on 'Classical.choice', which indicates that the 'Decidable' instance
depends on classical reasoning. Classical reasoning principles cannot be run as programs, and
the 'decide' tactic works by evaluating a decision procedure via reduction, which then gets
stuck. Instances that depend on classical reasoning are most often in scope due to an
'open scoped Classical' command.
```

tests/lean/run/decideTactic.lean
```diff
@@ -0,0 +1,102 @@
...
+since its Decidable instance
+  instDecidableNice
+did not reduce to isTrue or isFalse.
+
+After unfolding the instances baz and instDecidableNice, reduction got stuck at the Decidable instance
+  ⋯ ▸ inferInstance
+
+Hint: Reduction got stuck on '▸' (Eq.rec), which suggests that one of the decidability instances is
+defined using tactics such as 'rw' or 'simp'. Use definitions such as inferInstanceAs or
+decidable_of_decidable_of_iff to alter a proposition.
```
david-christiansen, line 76:
```suggestion
Hint: Reduction got stuck on '▸' (Eq.rec), which suggests that one of the 'Decidable' instances is
defined using tactics such as 'rw' or 'simp'. Instead of these tactics, use functions such as 'inferInstanceAs' or
'decidable_of_decidable_of_iff' to alter a proposition.
```
I'm not too happy with the term "function" here, but I think it's better than "definitions" (as the definition is not what's being used, really). Perhaps "operator"?

Thanks for being careful about making them all clickable :)

david-christiansen:
I have only nitpicks. This is a great improvement!

tests/lean/run/decideTactic.lean
kmill: I've got another iteration that tries to explain exactly what "classical" means in this context. I also toned down "most often in scope" since in mathlib it's probably about as common to use the `classical` tactic.

tests/lean/run/decideTactic.lean
david-christiansen: For me, "declaration" and "definition" have the same downside: a declaration/definition is that which results in a name being given a meaning, but not the name nor the meaning itself. User's aren't expected to stick `def ...` here, they're expected to invoke something that was previously defined.

This is indeed a bit nitpicky, and I think the risk of someone being misled is fairly low.

"function", then?

src/Lean/Elab/Tactic/ElabTerm.lean
```diff
@@ -423,38 +423,41 @@ private partial def blameDecideReductionFailure (inst : Expr) : MetaM Expr := do
...
             to alter a proposition."
           else if reason.isAppOf ``Classical.choice then
             m!"\n\n\
-            Hint: Reduction got stuck on {MessageData.ofConstName ``Classical.choice}, suggesting that there are \
-            \"classical\" instances, due to for example an 'open scoped Classical' command. \
-            The 'decide' tactic works by evaluating a decision procedure via reduction, but such \
-            classical instances cannot be evaluated."
+            Hint: Reduction got stuck on '{MessageData.ofConstName ``Classical.choice}', \
+            which indicates that a '{MessageData.ofConstName ``Decidable}' instance \
+            is defined using classical reasoning, proving an instance exists rather than giving a concrete construction. \
```
david-christiansen, line 449: This is really excellent phrasing, a big improvement on my suggestion!

tests/lean/run/decideTactic.lean
david-christiansen: This is really good now!

tests/lean/run/decideTactic.lean
kmill: Imagine if these were theorems instead of definitions. Would you have the same objection to "Instead of these tactics, use theorems such as 'thm1' or 'thm2'"? Definitions and theorems are fundamental notions that we need to refer to somehow (both of which are kind of declarations).

Is the analogy "definition is to function/constant as theorem is to theorem"?

tests/lean/run/2161.lean
```diff
@@ -0,0 +1,50 @@
...
+  ((mul 4 1).mul 1).mul 1 = 4
+since its 'Decidable' instance
+  instDecidableEqFoo (((mul 4 1).mul 1).mul 1) 4
+did not reduce to 'isTrue' or 'isFalse'.
+
+After unfolding the instances 'decEqFoo✝', 'instDecidableEqFoo', 'instDecidableEqNat' and
+'Nat.decEq', reduction got stuck at the 'Decidable' instance
+  match h : (((mul 4 1).mul 1).mul 1).num.beq 4 with
+  | true => isTrue ⋯
+  | false => isFalse ⋯
```
hrmacbeth, line 26: Possible future addition: add another "Hint" covering this situation (one of the definitions is marked as irreducible)?

kmill: It would be nice to have a more general system to diagnose reduction issues! Right now, I'm a little worried that this reduction blamer might mis-blame `match` discriminants. At least every Decidable instance that appears *should* be reducible, so it should be harmless. With arbitrary discriminants, there's more risk of being wrong, unless perhaps we report each discriminant that doesn't reduce to a constructor.

(Somewhat related, today [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Should.20wfrec.20defs.20require.20an.20.40.5Birreducible.5D.20annotation.3F/near/450769240) I was thinking about this problem for diagnosing type mismatches. It's a bit different, since it's for isDefEq, but isDefEq depends on whnf.)

PR #4675, opened 2024-07-07
feat: `@[pp_no_structure_instances]`

TwoFX:


TwoFX:
See also https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/making.20.60op.60.20irreducible

nomeata:
The negation in the attribute may be a bit confusing. Can we maybe have `pp_structure_instances`, on by default, and negated by `-pp_structure_instances`, akin to `-simp`? (Just a thought, not a big deal)

TwoFX:
> The negation in the attribute may be a bit confusing. Can we maybe have `pp_structure_instances`, on by default, and negated by `-pp_structure_instances`, akin to `-simp`? (Just a thought, not a big deal)

I'm not sure how I feel about this. Is there precedence for such an on-by-default attribute? I can't think of one off the top of my head. Do I understand correctly that with your suggestion one would tag the structure with `@[-pp_structure_instances]`? I have never seen that syntax used in this way, I have only ever seen `attribute [-attr] name` used after the fact.

FWIW, there is definitely precedence for the negated attributes, namely the (not entirely unrelated) `@[pp_nodot]` attribute.

nomeata:
Fair points, I'll retract my suggestion

nomeata:
Hmm, but something rubs me the wrong way about this attribute.

Would it be more elegant to have a `[pp_instances method]` attribute where the method indicates the desired pretty printing (record syntax, anonymous brackets, constructor function)? Assuming that a user who wants to have control about the pretty printing, why not give them more control?

TwoFX:
> it be more elegant to have a `[pp_instances method]` attribute where the method indicates the desired pretty printing (record syntax, anonymous brackets, constructor function)? Assuming that a user who wants to have control about the pretty printing, why not give them more control?

I like this idea in principle. One potential problem I see is that adding `pp_instances` would probably entail removing `@[pp_with_anonymous_constructor]`, which would be an otherwise avoidable backwards-incompatible change. Another interesting question would be whether this attribute would apply even if the `pp.structureInstances` setting is set to false (this is not the case for `pp_with_anonymous_constructor` today).

nomeata:
The old attribute could just become a synonym for `pp_instances anonymous`, couldn't it (possibly with a deprecation warning, now or later)

As for the interaction with `pp.structureInstances` I'd consider changing that, so that if one disables this pretty printing option, one reliably gets less pretty and more explicit terms - that seems to be the purpose of `pp.structureInstances false`, isn't it?

kmill:
Thanks for the implementation @TwoFX, but I'm going to suggest that we close this PR. There's enough discussion here where it would be better to continue it on Zulip and form an RFC before continuing. However, I also think that this particular feature, while justified for the particular use case, has limited applicability and is served well enough by app unexpanders:

```lean
structure Opposite (α : Type u) where
  op :: unop : α

/-- info: { unop := 1 } : Opposite Nat -/
#guard_msgs in
#check (Opposite.op 1)

@[app_unexpander Opposite.op]
protected def Opposite.unexpander_op : Lean.PrettyPrinter.Unexpander
  | s => pure s

/-- info: Opposite.op 1 : Opposite Nat -/
#guard_msgs in
#check (Opposite.op 1)

open Opposite

/-- info: op 1 : Opposite Nat -/
#guard_msgs in
#check (Opposite.op 1)
```

@nomeata The `pp.structureInstances` option has the same polarity as `pp.notation`. In principle neither structure instances nor notation hide information and they're just showing a term with a different point of view. This is a little bit of a lie, but, for what it's worth, they do disable themselves when, for example, `pp.explicit` is true and they would be hiding implicit arguments (at least for correctly written delaborators).

TwoFX:
@kmill Thanks! I agree that unless this use case turns out to come up much more frequently than we're currently seeing the unexpander is the best solution. I'll open a mathlib PR.

kmill:
@TwoFX I completely forgot to suggest just not using structures. Are they being used just to avoid needing to define the `unop` projection? Or is there some reduction performance consideration?
```lean
inductive Opposite (α : Type u)
  | op : α → Opposite α

def Opposite.unop {α : Type u} : Opposite α → α
  | .op x => x

/-- info: Opposite.op 1 : Opposite Nat -/
#guard_msgs in
#check (Opposite.op 1)

open Opposite

/-- info: op 1 : Opposite Nat -/
#guard_msgs in
#check (Opposite.op 1)
```

kmill:
@TwoFX I know that the `structure` command uses `Expr.proj` in the projection functions it defines, and this expression is used in some kernel-level optimization, but I don't know the exact impact, and I would guess that for one-field structures it's not worth worrying about. (By the way, isStructureLike is simply checking that the inductive type is non-recursive and has just one-constructor. The kernel doesn't use isStructure.)

The big issue in Lean 3 used to be that there was no "eta for structures":
```lean
inductive Opposite (α : Type*)
  | op : α → Opposite

def Opposite.unop {α : Type*} : Opposite α → α
  | (Opposite.op x) := x

open Opposite

variable {α : Type*}

example (x : α) : unop (op x) = x := rfl
example (x : Opposite α) : op (unop x) = x := rfl -- fails
```
However, the Lean 4 version works:
```lean
inductive Opposite (α : Type u)
  | op : α → Opposite α

def Opposite.unop {α : Type u} : Opposite α → α
  | .op x => x

open Opposite

example (x : α) : unop (op x) = x := rfl
example (x : Opposite α) : op (unop x) = x := rfl
```
It this is because eta-for-structures applies and can transform `x` into `op x.1` where `x.1` is using `Expr.proj`.

Regarding mathlib, we could consider writing a command for defining these wrapped types and handling how it pretty prints, leaving whether it's done as an `inductive` or `structure` as an implementation detail. It'd be good for uniformity.

TwoFX:
@kmill Thank you for the explanation. Having such a command would be nice, but I don't think it's urgent.

PR #4678, opened 2024-07-07
feat: lemmas for List.head and List.getLast

kim-em:


semorrison:
Not sure why merge is failing, status checks looked good here. I've merged `master` and will try again.

PR #4684, opened 2024-07-08
refactor: `InductiveVal.numNested` instead of `.isNested`

nomeata:
Before, in order to find out how many auxilary datatype are in a
mutual group of inductive with nested data type, one had to jump
through hoops like this:

```
private def numNestedInducts (indName : Name) : MetaM Nat := do
  let .inductInfo indVal ← getConstInfo indName | panic! "{indName} is an inductive"
  let .recInfo recVal ← getConstInfo (mkRecName indName) | panic! "{indName} has a recursor"
  return recVal.numMotives - indVal.all.length
```

The `InductiveVal` data structure already has `.isNested : Bool`, so it
seems to be a natural extension to beef that up to `.numNested: Nat`.

This touched kernel code.


nomeata:
(This includes #4683, diff will be prettier once that’s merged.)

Issue #4698, opened 2024-07-09
Lots of unnecessary copies of `object_ref`


legrosbuffle:
### Prerequisites

Please put an X between the brackets as you perform the following steps:

* [X ] Check that your issue is not already filed:
      https://github.com/leanprover/lean4/issues
* [ ] Reduce the issue to a minimal, self-contained, reproducible test case.
      Avoid dependencies to Mathlib or Batteries.
* [ ] Test your test case against the latest nightly release, for example on
      https://live.lean-lang.org/#project=lean-nightly
      (You can also use the settings there to switch to “Lean nightly”)

### Description

Move constructors for objects deriving from `object_ref` in lean are making a copy, probably due to the assumption that an rvalue reference is propagated as such across function calls.

Example:

```
expr(expr && other):object_ref(other)
```

should be:

```
expr(expr && other):object_ref(std::move(other))
```

As a result, there are a lot of unnecessary incref/decref calls on `object_ref`. We've measured that `expr` copying and assignment by itself uses 10% of total runtime on our workloads.

### Context

No specific context.

### Steps to Reproduce

**Expected behavior:**

`expr::expr(expr&& other)` calls `object_ref::object_ref(object_ref&&)`, which steals refcount from `other`, no incref happens.

**Actual behavior:**

`expr::expr(expr&& other)` calls `object_ref::object_ref(const object_ref&)`, which increfs `other`'s refcount. Then the callee decrefs `other`'s refcount.

### Versions

v4.7.0

### Additional Information

n/a

### Impact

Performance only.

legrosbuffle:
I have a few more :)

PR #4700, opened 2024-07-09
perf: fix implementation of move constructors and move assignment ope…

legrosbuffle:
…rators

Right now those constructors result in a copy instead of the desired move. We've measured that expr copying and assignment by itself uses around 10% of total runtime on our workloads.

See #4698 for details.

### DIFF
--- a/src/kernel/declaration.h
+++ b/src/kernel/declaration.h
@@ -63,9 +63,9 @@ class constant_val : public object_ref {
 public:
     constant_val(name const & n, names const & lparams, expr const & type);
     constant_val(constant_val const & other):object_ref(other) {}
-    constant_val(constant_val && other):object_ref(other) {}
+    constant_val(constant_val && other):object_ref(std::move(other)) {}
     constant_val & operator=(constant_val const & other) { object_ref::operator=(other); return *this; }
-    constant_val & operator=(constant_val && other) { object_ref::operator=(other); return *this; }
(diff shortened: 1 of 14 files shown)


hargoniX:
!bench

semorrison:
Mathlib benchmark is hopefully running now at https://github.com/leanprover-community/mathlib4/pull/14599.

semorrison:
> Mathlib benchmark is hopefully running now at [leanprover-community/mathlib4#14599](https://github.com/leanprover-community/mathlib4/pull/14599).

No significant changes in Mathlib build time. (This is, of course, not intended to be a representative benchmark!)

Kha:
@legrosbuffle I assume the time savings happen in kernel type checking in your case? While lean4 and mathlib4 are not bottlenecked by the kernel at the moment, we are still interested in optimizing it. Do you think you could extract a representative stand-alone test case for the bottleneck and speedup?

legrosbuffle:
> @legrosbuffle I assume the time savings happen in kernel type checking in your case?

Yes, the majority of the cost if from `src/kernel/replace_fn.cpp`.

> While lean4 and mathlib4 are not bottlenecked by the kernel at the moment, we are still interested in optimizing it. Do you think you could extract a representative stand-alone test case for the bottleneck and speedup?

It's not obvious that this is easy to reduce, but I'll see what I can do.

PR #4702, opened 2024-07-09
perf: avoid expr copies in replace_rec_fn::apply

legrosbuffle:
Those represent ~13% of the time spent in `save_result`,
even though `r` is a temporary in all cases but one.

See #4698 for details.


Kha:
!bench

PR #4708, opened 2024-07-09
chore: reorganise lemmas on list getters

kim-em:
Just reordering lemmas, sorting into subsections.


digama0:
@semorrison This PR introduced a breakage missing a deprecation, `List.getD_eq_getElem?` -> `List.getD_eq_getElem?_getD`, could you add a deprecation for this?

PR #4717, opened 2024-07-10
fix: instantiate mvars of indices before instantiating fvars

### LINKED ISSUE #3242: Unabstracted variables in indices of inductive types in mutual block
### Prerequisites

### Description

In certain cases, when elaborating inductive types, Lean fails to abstract some loose free variables in the signature of inductive types (more particularly on indices) in a mutual block, leading to an error.

### Context

Discovered initially [here](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.E2.9C.94.20Local.20quantified.20vars.20in.20inductive.20proposition/near/419322703). The issue can be avoided by using `variable` 

### Steps to Reproduce

```lean
mutual

inductive R (F: α → α → Prop) (a : α) : α → Prop where
| ind : ∀ (f: Nat → α) b, (∀ n, And₂ F a b f n) → R F a b

inductive And₂ (F: α → α → Prop) (a : α) : α → (Nat → α) → Nat → Prop where
  | mk (b : α) (f : Nat → α) (n : Nat): R F a (f n) → F (f n) b → And₂ F a b f n

end
```

**Expected behavior:** No error

**Actual behavior:** Error `(kernel) declaration has free variables 'And₂'`

### Versions

4.5.0
Web Browser

### Additional Information

The signature type of `And₂` here is: `{α : Sort u_1} → (α → α → Prop) → α → _uniq.21 → (Nat → _uniq.21) → Nat → Prop`.

The issue can be circumvented by using `variable`, the following code works:
```lean
variable {α : Type _}

mutual

inductive R (F: α → α → Prop) (a : α) : α → Prop where
  | base : R F a a
  | ind : ∀ (f: Nat → α) b, (∀ n, And₂ F a b f n) → R F a b

inductive And₂ (F: α → α → Prop) (a : α) : α → (Nat → α) → Nat → Prop where
  | mk (b : α) (f : Nat → α) (n : Nat): R F a (f n) → F (f n) b → And₂ F a b f n

end
```

### Impact

arthur-adjedj:
When elaborating the headers of mutual indexed inductive types, mvars have to be synthesized and instantiated before replacing the fvars present there. Otherwise, some fvars present in uninstantiated mvars may be missed and lead to an error later.
Closes #3242 (again)

### DIFF
--- a/src/Lean/Elab/Inductive.lean
+++ b/src/Lean/Elab/Inductive.lean
@@ -779,6 +779,14 @@ private def mkInductiveDecl (vars : Array Expr) (views : Array InductiveView) : 
         let indFVar := indFVars[i]!
         Term.addLocalVarInfo views[i]!.declId indFVar
         let r     := rs[i]!
+        /- At this point, because of `withInductiveLocalDecls`, the only fvars that are in context are the ones related to the first inductive type.
+           Because of this, we need to replace the fvars present in each inductive type's header of the mutual block with those of the first inductive.
+           However, some mvars may still be uninstantiated there, and might hide some of the old fvars.
+           As such we first need to synthesize all possible mvars at this stage, instantiate them in the header types and only
+           then replace the parameters' fvars in the header type.
+
+           See issue #3242 (`https://github.com/leanprover/lean4/issues/3242`)
+        -/
         let type  ←  instantiateMVars r.type
         let type  := type.replaceFVars r.params params
         let type  ← mkForallFVars params type


arthur-adjedj:
WIP

arthur-adjedj:
I haven't had time to make thorough tests to see if this really solves the issue for good. Knowing the issue popped up once after being already closed, I'd like to make sure this doesn't happen again. I haven't dived into how inductive headers are unified yet, and don't know if some expr mvars aren't expected to be unified later on here. If the expected behavior is that all expr mvars in the headers have already been solved at this point, then this certainly is the right solution, and I'd be happy to get this merged. 
I'm not available to investigate on this right now, but plan on looking into this in the coming days.

arthur-adjedj:
Despite multiple rebases, the CI on mathlib/batteries seems to keep on failing, and I'm not sure why. Looking at the errors, I don't think the failures are due to the changes made in that PR. I'd like to see at least one CI run work "correctly" on this to ensure no breakage was made, but otherwise, I'm confident this PR makes the right fixes, and will be marking it as ready for review.

arthur-adjedj:
awaiting-review

nomeata:
Glad it's green now. Don’t forget to update the PR description, as it will become the commit message.

nomeata:
Looks good to me, I don’t think I understand the intricacies any better than you. Would you mind leaving a comment in the code with whatever insights you have so that the next person looking at the code won’t have to re-discover it?

nomeata (approved):

PR #4729, opened 2024-07-11
feat: have IR checker suggest `noncomputable`

### LINKED ISSUE #1785: IR check failed in presence of noncomputable
### Prerequisites

### Description

Code generator sometimes doesn't correctly flag noncomputaility

### Steps to Reproduce

In this (somewhat) minimal example I believe `def bug` should be marked noncomputable, but the `IR` generator doesn't pick up on this for some reason.
```lean
namespace WellFounded

variable {α : Sort u} {C : α → Sort v} {r : α → α → Prop}

unsafe def fix'.impl (hwf : WellFounded r) (F : ∀ x, (∀ y, r y x → C y) → C x) (x : α) : C x :=
  F x fun y _ => impl hwf F y

@[implemented_by fix'.impl]
def fix' (hwf : WellFounded r) (F : ∀ x, (∀ y, r y x → C y) → C x) (x : α) : C x := hwf.fix F x

end WellFounded

namespace Nat

theorem lt_or_eq_of_le {a b : Nat} : a ≤ b → a < b ∨ a = b := sorry
section find
variable (p : Nat → Prop)

private def lbp (m n : Nat) : Prop := m = n + 1 ∧ ∀ k, k ≤ n → ¬p k

variable [DecidablePred p] (H : ∃ n, p n)

private def wf_lbp : WellFounded (lbp p) := by
  refine ⟨let ⟨n, pn⟩ := H; ?_⟩
  suffices ∀ m k, n ≤ k + m → Acc (lbp p) k from fun a => this _ _ (Nat.le_add_left _ _)
  intro m
  induction m with refine fun k kn => ⟨_, fun | _, ⟨rfl, a⟩ => ?_⟩
  | zero => exact absurd pn (a _ kn)
  | succ m IH => exact IH _ (by rw [Nat.add_right_comm]; exact kn)

protected def find_x : {n // p n ∧ ∀ m, m < n → ¬p m} :=
(wf_lbp p H).fix' (C := fun k => (∀n, n < k → ¬p n) → {n // p n ∧ ∀ m, m < n → ¬p m})
  (fun m IH al => if pm : p m then ⟨m, pm, al⟩ else
      have this : ∀ n, n ≤ m → ¬p n := fun n h =>
        (lt_or_eq_of_le h).elim (al n) fun e => by rw [e]; exact pm
      IH _ ⟨rfl, this⟩ fun n h => this n $ Nat.le_of_succ_le_succ h)
  0 fun n h => absurd h (Nat.not_lt_zero _)

protected def find : Nat := (Nat.find_x p H).1
end find
end Nat

variable (R : Type) [∀ n, OfNat R n]

open Classical

noncomputable
def char : Nat := Nat.find (fun n => (1 : R) = 1) ⟨1, rfl⟩

def bug : R := -- IR check failed at 'bug._rarg', error: unknown declaration 'char'
  match char R with
  | 0 => 1
  | _ => 0
```

**Expected behavior:** message saying `def bug` must be marked noncomputable

**Actual behavior:**  IR check failed

**Reproduces how often:**  100% but seems to need a fairly complicated example

### Versions

nightly-10-25

kmill:
Currently, `ll_infer_type` is responsible for telling the user about `noncomputable` when a definition depends on one without executable code. However, this is imperfect because type inference does not check every subexpression. This leads to errors later on that users find to be hard to interpret.

Now, `Lean.IR.checkDecls` has a friendlier error message when it encounters constants without compiled definitions, suggesting to consider using `noncomputable`. While this function is an internal IR consistency check, it is also reasonable to have it give an informative error message in this particular case. The suggestion to use `noncomputable` is limited to just unknown constants.

Some alternatives would be to either (1) create another checker just for missing constants, (2) change `ll_infer_type` to always visit every subexpression no matter if they are necessary for inferring the type, or (3) investigate whether `tests/lean/run/1785.lean` is due to a deeper issue.

Closes #1785

### DIFF
--- a/src/Lean/Compiler/IR/Checker.lean
+++ b/src/Lean/Compiler/IR/Checker.lean
@@ -50,7 +50,7 @@ def markJP (j : JoinPointId) : M Unit :=
 def getDecl (c : Name) : M Decl := do
   let ctx ← read
   match findEnvDecl' ctx.env c ctx.decls with
-  | none   => throw s!"unknown declaration '{c}'"
+  | none   => throw s!"depends on declaration '{c}', which has no executable code; consider marking definition as 'noncomputable'"
   | some d => pure d
 
 def checkVar (x : VarId) : M Unit := do
@@ -182,7 +182,7 @@ end Checker
 def checkDecl (decls : Array Decl) (decl : Decl) : CompilerM Unit := do
   let env ← getEnv
   match (Checker.checkDecl decl { env := env, decls := decls }).run' {} with
-  | .error msg => throw s!"compiler IR check failed at '{decl.name}', error: {msg}"
+  | .error msg => throw s!"failed to compile definition, compiler IR check failed at '{decl.name}'. Error: {msg}"
   | _ => pure ()
 
 def checkDecls (decls : Array Decl) : CompilerM Unit :=
--- a/tests/lean/run/1785.lean
+++ b/tests/lean/run/1785.lean
@@ -1,0 +1,21 @@
+/-!
+# Improve compiler IR check message for users when constants are not compiled
+-/
+
+/-
+This is a simplified version of the example in #1785.
+Note that the error changes if the typeclass argument is removed.
+-/
+
+noncomputable
+def char (R : Type) [∀ n, OfNat R n] : Nat := 0
+
+/--
+error: failed to compile definition, compiler IR check failed at 'bug._rarg'. Error: depends on
+declaration 'char', which has no executable code; consider marking definition as 'noncomputable'
+-/
+#guard_msgs in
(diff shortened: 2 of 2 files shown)


kmill:
@leodemoura Keven Buzzard was telling me today that he has to frequently help people with this IR compiler check issue. I'm not sure if this is the correct fix, but putting this error message is this particular place seemed potentially appropriate after studying `ll_infer_type`, `checkDecls`, and some other bits of the compiler.

If anything, this PR contains a simpler version of the test case at #1785. Note that removing the typeclass argument makes `ll_infer_type` catch the missing constant instead.

PR #4718, opened 2024-07-10
feat: infer mutual structural recursion

nomeata:
the support for mutual structural recursion (new since #4575) is extended so that Lean tries to infer it even without annotations.

* The error message when termination checking fails looks quite different now. Maybe a bit better, maybe with more room for improvements.
* If there are too many combinations (with an arbitrary cut-off) for a given argument type, it will just give up and ask the user to use `termination_by structural`.
* It is now legal to specify `termination_by structural` on not necessarily all functions of a clique; this simply restricts the combinations of arguments that Lean considers.

PR #4738, opened 2024-07-13
refactor: IndGroupInfo and IndGroupInst

nomeata:
This adds the types
* `IndGroupInfo`, a variant of `InductiveVal` with information that
   applies to a whole group of mutual inductives and
* `IndGroupInst` which extends `IndGroupInfo` with levels and parameters
   to indicate a instantiation of the group.

One purpose of this abstraction is to make it clear when a fuction
operates on a group as a whole, rather than a specific inductive within
the group.

This is extracted from #4718 and #4733 to reduce PR size and improve
bisectability.

### DIFF
--- a/src/Lean/Elab/PreDefinition/Structural/BRecOn.lean
+++ b/src/Lean/Elab/PreDefinition/Structural/BRecOn.lean
@@ -235,7 +235,6 @@ It was already checked earlier in `checkCodomainsLevel` that the functions live 
 -/
 def mkBRecOnConst (recArgInfos : Array RecArgInfo) (positions : Positions)
    (motives : Array Expr) : MetaM (Nat → Expr) := do
-  -- Get a representative recArgInfo; in particular
   let indGroup := recArgInfos[0]!.indGroupInst
   let motive := motives[0]!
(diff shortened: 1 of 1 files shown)


src/Lean/Elab/PreDefinition/Structural/BRecOn.lean
```diff
@@ -237,35 +228,40 @@ def mkBRecOnF (recArgInfos : Array RecArgInfo) (positions : Positions)
...
 Given the `motives`, figures out whether to use `.brecOn` or `.binductionOn`, pass
 the right universe levels, the parameters, and the motives.
 It was already checked earlier in `checkCodomainsLevel` that the functions live in the same universe.
 -/
 def mkBRecOnConst (recArgInfos : Array RecArgInfo) (positions : Positions)
-   (motives : Array Expr) : MetaM (Name → Expr) := do
-  -- For now, just look at the first
-  let recArgInfo := recArgInfos[0]!
+   (motives : Array Expr) : MetaM (Nat → Expr) := do
+  -- Get a representative recArgInfo; in particular
```
nomeata, line 238:
```suggestion
```

PR #4739, opened 2024-07-13
feat: make `instance` attribute check that type is a class

kmill:
* Modifies `Lean.Meta.addInstance` to check that the declaration being marked as an instance has a class as its type.
* Changes the application time of the `instance` attribute to `afterCompilation` so that auxiliary definitions (for example those produced when processing mutual WF recursion) do not themselves get the `instance` attribute.
* For the `instance` command, the error from `addInstance` is localized to the `instance` keyword rather than the entire command.

Issue reported by Arthur Adjedj [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Non-class.20instances.20don't.20fail/near/451131340).

Co-authored-by: Joachim Breitner <mail@joachim-breitner.de>

### DIFF
--- a/src/Lean/Meta/Instances.lean
+++ b/src/Lean/Meta/Instances.lean
@@ -191,6 +191,10 @@ private partial def computeSynthOrder (inst : Expr) : MetaM (Array Nat) :=
 
   return synthed
 
+/--
+Adds the `instance` attribute to `declName` with the specified attribute kind and priority.
+Throws an error if the type of the declaration is not a class.
+-/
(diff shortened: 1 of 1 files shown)


kmill:
Closing. It turns out that this implementation doesn't work because the instance attribute controls both what's an instance and what gets reduced at instance reducibility. Being stingy with applying the attribute causes some intermediate definitions to not reduce. Also, it can make instances fail to reduce properly during their own compilation!

PR #4742, opened 2024-07-15
feat: theory from LeanSAT

hargoniX:


src/Init/PropLemmas.lean
```diff
@@ -309,6 +309,11 @@ theorem not_forall_of_exists_not {p : α → Prop} : (∃ x, ¬p x) → ¬∀ x,
 @[simp] theorem exists_eq_right_right' : (∃ (a : α), p a ∧ q a ∧ a' = a) ↔ p a' ∧ q a' := by
   simp [@eq_comm _ a']
 
+@[simp] theorem exists_or_eq_left (y : α) (p : α → Prop) : ∃ x : α, x = y ∨ p x := ⟨y, .inl rfl⟩
```
TwoFX, line 312: Is there a reason why this is not
```suggestion
@[simp] theorem exists_or_eq_left (y : α) (p : Prop) : ∃ x : α, x = y ∨ p := ⟨y, .inl rfl⟩
```

hargoniX: The current version is able to deal with the case where `p` depends on the existentially bound `x`, your version would not be able to do that, e.g.:
```lean

namespace Foo
theorem exists_or_eq_left (y : α) (p : α → Prop) : ∃ x : α, x = y ∨ p x := ⟨y, .inl rfl⟩

example : ∃ x, x = 0 ∨ 0 ≤ x := by simp only [exists_or_eq_left]

end Foo

namespace Bar

theorem exists_or_eq_left (y : α) (p : Prop) : ∃ x : α, x = y ∨ p := ⟨y, .inl rfl⟩

example : ∃ x, x = 0 ∨ 0 ≤ x := by simp only [exists_or_eq_left]

end Bar
```

TwoFX: Oh right, I had the "more general" relation the wrong way around in my head :) Thank you for the explanation.

jcommelin:
LGTM, fwiw

PR #4746, opened 2024-07-15
chore: exclude more symbols to get below Windows symbol limit

Kha:
This is a short-term fix, a more principled fix is in the works


Kha:
(needs a double-stage0 update to take effect)

PR #4748, opened 2024-07-15
doc: document `Command.Scope`

grunweg:
Also extends existing definition for `getScope`/`getScopes` and clarifies that the `end` command is optional at the end of a file.

### DIFF
--- a/src/Lean/Elab/Command.lean
+++ b/src/Lean/Elab/Command.lean
@@ -39,8 +39,7 @@ structure Scope where
   currNamespace : Name := Name.anonymous
   /-- All currently `open`ed namespaces and names. -/
   openDecls     : List OpenDecl := []
-  /-- The current list of names for universe level variables to use for new declarations.
-  This is managed by the `universe` command. -/
+  /-- The current list of names for universe level variables to use for new declarations. This is managed by the `universe` command. -/
   levelNames    : List Name := []
   /--
   The current list of binders to use for new declarations.


src/Lean/Elab/Command.lean
```diff
@@ -12,17 +12,35 @@ import Lean.Language.Basic
...
-  /-- section variables -/
+  /-- All variable binders which apply in the current scope -/
   varDecls      : Array (TSyntax ``Parser.Term.bracketedBinder) := #[]
   /-- Globally unique internal identifiers for the `varDecls` -/
   varUIds       : Array Name := #[]
-  /-- noncomputable sections automatically add the `noncomputable` modifier to any declaration we cannot generate code for. -/
+  /-- Whether the current scope is (or is nested inside) a `noncomputable` `section`:
+  in this case, automatically add the `noncomputable` modifier to any declaration
+  we cannot generate code for.
+  A scope is noncomputable if it is a noncomputable section or its parent is noncomputable.
```
grunweg, line 42: The first and second sentence are somewhat redundant, as I liked both phrasings and could not decide if any one was better. Feel free to articulate a preference (or keep both).

src/Lean/Elab/Command.lean
```diff
@@ -12,17 +12,35 @@ import Lean.Language.Basic
...
 
+/--
+A `Scope` collects context information which is inherently scoped: this contains
+the current namespace and information about `open`, `universe` or `variable` declarations.
+There is always a base scope; scopes can be nested.
+
+`universe`, `open` and `variable` declarations only modify the current scope
+(and all scopes nested below it): these declarations usually *append* to the
+context in the current scope; `variable` statements can also change existing items.
+(XXX Is shadowing possible? If so, this sentence should be reworded.)
```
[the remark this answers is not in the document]
kmill: In "only the top `Scope` on the stack" sentence in my previous comment, I think we don't need this paragraph.

There's not a concept of 'closing scopes' in the internal Lean ontology (one could imagine there could be — what if `Scope` had a user-specified handler that had to be run when 'closing a scope', which could add additional declarations given all the declarations made in the current scope?)

src/Lean/Elab/Command.lean
```diff
@@ -12,17 +12,35 @@ import Lean.Language.Basic
...
+Unclosed scopes are automatically closed at the end of the current module.
+-/
 structure Scope where
   header        : String
```
kmill, line 27:
```suggestion
  /-- The component of the `namespace` or `section` that this scope is associated to.
  For example, `section a.b.c` and `namespace a.b.c` each create three scopes with headers named `a`, `b`, and `c`.
  This is used for checking the `end` command. The "base scope" has `""` as its header.  -/
  header        : String
```

src/Lean/Elab/Command.lean
```diff
@@ -12,17 +12,35 @@ import Lean.Language.Basic
...
   opts          : Options := {}
```
kmill, line 28:
```suggestion
  /-- The current state of all set options at this point in the scope. Note that this is the full current set options and *not* simply the options set while this scope has been active. -/
  opts          : Options := {}
```

src/Lean/Elab/Command.lean
```diff
@@ -12,17 +12,35 @@ import Lean.Language.Basic
...
+  /-- The current namespace: by default, this is an anonymous namespace -/
```
kmill, line 29:
```suggestion
  /-- The current namespace. The top-level namespace is represented by `Name.anonymous`. -/
```

src/Lean/Elab/Command.lean
```diff
@@ -12,17 +12,35 @@ import Lean.Language.Basic
...
   currNamespace : Name := Name.anonymous
+  /-- All currently `open`ed namespaces -/
```
kmill, line 31:
```suggestion
  /-- All currently `open`ed namespaces and names. -/
```
The `OpenDecl` type gives information about individual names too, needed to support `open MyNamespace (decl1, decl2)`.

src/Lean/Elab/Command.lean
```diff
@@ -12,17 +12,35 @@ import Lean.Language.Basic
...
   openDecls     : List OpenDecl := []
+  /-- All universe levels which apply in the current scope -/
```
kmill, line 33:
```suggestion
  /-- The current list of names for universe level variables to use for new declarations. This is managed by the `universe` command. -/
```

src/Lean/Elab/Command.lean
```diff
@@ -12,17 +12,35 @@ import Lean.Language.Basic
...
   levelNames    : List Name := []
-  /-- section variables -/
+  /-- All variable binders which apply in the current scope -/
```
kmill, line 35:
```suggestion
  /-- The current list of binders to use for new declarations. This is managed by the `variable` command.
  Each binder is represented in `Syntax` form, and it is re-elaborated within each command that uses this information.
  
  This is also used by commands, such as `#check`, to create an initial local context, even if they do not work with binders per se. -/
```

src/Lean/Elab/Command.lean
```diff
@@ -12,17 +12,35 @@ import Lean.Language.Basic
...
   varDecls      : Array (TSyntax ``Parser.Term.bracketedBinder) := #[]
   /-- Globally unique internal identifiers for the `varDecls` -/
```
kmill, line 37:
```suggestion
  /-- Globally unique internal identifiers for the `varDecls`. There is one identifier per variable introduced by the binders (recall that a binder such as `(a b c : Ty)` can produce more than one variable), and each identifier is the user-provided variable name with a macro scope.
  This is used by `TermElabM` in `Lean.Elab.Term.Context` to help with processing macros that capture these variables. -/
```

src/Lean/Elab/Command.lean
kmill: Could you make a version that doesn't talk about nesting scopes? For someone using this type, it's important that they know that none of the other scopes are looked at when deciding whether it's noncomputable, so I'd avoid talking about parent/child scopes too much, to prevent it giving the wrong idea.

Here's a suggestion to give some ideas:
```quote
Whether declarations that fail to compile should automatically receive the `noncomputable` modifier.
A scope with this flag set is created by `noncomputable section`.

Recall that a new scope inherits all values from its parent scope, so all sections and namespaces nested within a `noncomputable` section will also have this flag set.
```

kmill (commented):
Here's some feedback so far, along with some documentation for the remaining fields.

src/Lean/Elab/Command.lean
grunweg: Makes sense - I like your suggestion, and went with something very similar.

grunweg:
Thanks for the suggestions and explanations; these were very helpful. I have addressed them all.

grunweg:
I was not sure about two details of the "how to contribute":
- I made separate commits for my changes (so you can check more easily what changed)
- I aimed to wrap all comments at 100 characters (do you use semantic line wrapping, hard wrapping at 100 chars or a mixture?)

I'm very happy to follow the house style, if I know what it is. `CONTRIBUTING` and the information linked therein does not seem to mention it.

src/Lean/Elab/Command.lean
```diff
@@ -612,6 +650,9 @@ Interrupt and abort exceptions are caught but not logged.
 private def liftAttrM {α} (x : AttrM α) : CommandElabM α := do
   liftCoreM x
 
+/-- Return the stack of all currently active scopes:
+the base scope always comes last; new scopes are prepended in the front.
+In particular, the current scope is always the first element. -/
```
kmill, line 655: For one sentence docstrings you can put them on one line, but the style is that multiline ones have `/--` and `-/` on their own lines:
```suggestion
/--
Return the stack of all currently active scopes:
the base scope always comes last; new scopes are prepended in the front.
In particular, the current scope is always the first element.
-/
```
This holds for field docstrings as well. Would you mind updating `Scope`?

src/Lean/Parser/Command.lean
```diff
@@ -269,7 +269,7 @@ corresponding `end <id>` or the end of the file.
   "namespace " >> checkColGt >> ident
 /--
 `end` closes a `section` or `namespace` scope. If the scope is named `<id>`, it has to be closed
-with `end <id>`.
+with `end <id>`. Otherwise, the `end` command is optional.
```
kmill, line 272: "Otherwise" makes it seem like "in contrast to the preceding" on first reading, so I suggest changing it.
```suggestion
with `end <id>`. The `end` command is optional at the end of a file.
```

kmill (commented):
Don't worry about squashing, we'll squash merge.

Also, don't worry about wrapping to 100 exactly. There doesn't seem to be a hard limit for line length in the repository, so long as the lines are reasonable. I usually do semantic wrapping myself, but there's no requirement.

src/Lean/Parser/Command.lean
grunweg: That's better, indeed! Changed.

src/Lean/Elab/Command.lean
grunweg: Sure, done. Do you prefer `levelNames` to also be put on separate lines (for me, that's morally "one sentence")?

kmill: For `levelNames`, your options are
```
  /-- The current list of names for universe level variables to use for new declarations. This is managed by the `universe` command. -/
  levelNames    : List Name := []
```
or
```
  /--
  The current list of names for universe level variables to use for new declarations.
  This is managed by the `universe` command.
  -/
  levelNames    : List Name := []
```
(Along with some minor variations.) I'll leave the choice to you.

I shouldn't have said "one sentence" but rather "one line".

grunweg: Changed, to the first version.

kmill:
Thanks! I've submitted it to the merge queue.

PR #4755, opened 2024-07-16
feat: test minimizing slowdown from #4595

kim-em:
This is a minimization of a fairly typical slowdown from #4595 in Mathlib. The bad slowdowns seem to mostly be in declarations that generate many different `simp` proofs internally (often via `aesop`).

This test works fine on `master`, but is very slow on #4595. There's some nonlinearity going on here: the time savings coming from replacing each `simp only` by `sorry` adds up to more than the total time.


kim-em:
I've moved this directly into #4595. (I'm not entirely sure why I didn't do that in the first place.)

PR #4756, opened 2024-07-16
feat: PProd syntax (part 3)

nomeata:
reworks #4730 based on feedback from @kmill:

 * Uses `×'` for PProd
 * No syntax for MProd for now
 * Angle brackets (without nesting) for the values

### DIFF
--- a/src/Init/NotationExtra.lean
+++ b/src/Init/NotationExtra.lean
@@ -163,16 +163,6 @@ end Lean
   | `($(_) $x $y)          => `(($x, $y))
   | _                      => throw ()
 
-@[app_unexpander PProd.mk] def unexpandPProdMk : Lean.PrettyPrinter.Unexpander
-  | `($(_) $x ⟨$y, $ys,*⟩) => `(⟨$x, $y, $ys,*⟩)
-  | `($(_) $x $y)         => `(⟨$x, $y⟩)
-  | _                     => throw ()
-
-@[app_unexpander MProd.mk] def unexpandMProdMk : Lean.PrettyPrinter.Unexpander
-  | `($(_) $x ⟨$y, $ys,*⟩) => `(⟨$x, $y, $ys,*⟩)
-  | `($(_) $x $y)         => `(⟨$x, $y⟩)
-  | _                     => throw ()
-
 @[app_unexpander ite] def unexpandIte : Lean.PrettyPrinter.Unexpander
   | `($(_) $c $t $e) => `(if $c then $t else $e)
   | _                => throw ()
--- a/src/Lean/PrettyPrinter/Delaborator/Builtins.lean
+++ b/src/Lean/PrettyPrinter/Delaborator/Builtins.lean
@@ -1135,6 +1135,24 @@ def delabSigma : Delab := delabSigmaCore (sigma := true)
 @[builtin_delab app.PSigma]
 def delabPSigma : Delab := delabSigmaCore (sigma := false)
 
+-- PProd and MProd value delaborator
+-- (like pp_using_anonymous_constructor but flattening nested tuples)
+
+def delabPProdMkCore (mkName : Name) : Delab := whenNotPPOption getPPExplicit <| whenPPOption getPPNotation do
+  guard <| (← getExpr).getAppNumArgs == 4
+  let a ← withAppFn <| withAppArg delab
+  let b ← withAppArg <| delab
+  if (← getExpr).appArg!.isAppOfArity mkName 4 then
+    if let `(⟨$xs,*⟩) := b then
+      return ← `(⟨$a, $xs,*⟩)
+  `(⟨$a, $b⟩)
+
+@[builtin_delab app.PProd.mk]
+def delabPProdMk : Delab := delabPProdMkCore ``PProd.mk
+
+@[builtin_delab app.MProd.mk]
+def delabMProdMk : Delab := delabPProdMkCore ``MProd.mk
+
 partial def delabDoElems : DelabM (List Syntax) := do
   let e ← getExpr
   if e.isAppOfArity ``Bind.bind 6 then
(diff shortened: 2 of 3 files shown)


tests/lean/run/PUnit_syntax.lean
```diff
@@ -1,44 +1,32 @@
...
+/-- info: PUnit.{u} : Sort u -/
 #guard_msgs in
-#check Bool ×ₘ Nat ×ₘ List Unit
+#check PUnit
 
-/-- info: Bool ×ₘ Nat ×ₘ List Unit : Type -/
+/-- info: ⟨true, ⟨5, [()]⟩⟩ : Bool ×' Nat ×' List Unit -/
 #guard_msgs in
-#check MProd Bool (MProd Nat (List Unit))
+#check (⟨true, 5, [()]⟩ : Bool ×' Nat ×' List Unit)
```
nomeata, line 19: I was hoping that the angle brackets wouldn’t be nested here

Maybe the elaborator in
```
  if hasPPUsingAnonymousConstructorAttribute env s.induct then
    return ← withTypeAscription (cond := (← withType <| getPPOption getPPStructureInstanceType)) do
      `(⟨$[$(stx[1].getArgs)],*⟩)
```
should flatten angle brackets in the last argument, given that the input syntax allows this idiom?

Or is this too confusing if the angle brackets do not come from a type that is intended to be used nested like tuples. 

So for now this PR uses
```
@[app_unexpander PProd.mk] def unexpandPProdMk : Lean.PrettyPrinter.Unexpander
  | `($(_) $x ⟨$y, $ys,*⟩) => `(⟨$x, $y, $ys,*⟩)
  | `($(_) $x $y)         => `(⟨$x, $y⟩)
  | _                     => throw ()
```
to un-nest the brackets just for PProd.

WDYT, @kmill?

kmill: Flattening angle bracket notation has a drawback, which is that you lose the ability to hover over the flattened-in expression. That's a reason why the general anonymous constructor delaborator doesn't do it.

I know some users avoid using the flattening feature, so even for a good rule like "flatten if all constructors are for the same type" I'd hesitate pretty printing with it.

It seems fine flattening for tuple-like types though. `PProd` is also very specialized, and in my experience only used in automation, so I think we can feel free to choose to flatten here.

nomeata: Ok, and is the way I do the flattening ok? Or is there a better way, maybe one that doesn't flatten unrelated anonymous structures in the last field?

kmill: It looked OK until your last comment... I didn't think about how it doesn't know whether or not the angle bracket notation is coming from PProd.

We'd have to either be OK with this (PProd is for specialized use) or write a delaborator. The delaborator isn't hard, and I can post one in a few minutes.

nomeata: If you are inclined I would like to see that, as a learning experience, just push to this PR. But it isn't important if you'd rather leave it as it is.

kmill: I think I got it right, but there's probably an oversight still... (I almost forgot `withIncDepth` in case it's a deeply nested tuple.)
```lean
@[delab app.PProd.mk]
partial def delabPProdMk : Delab := whenNotPPOption getPPExplicit <| whenPPOption getPPNotation do
  guard <| (← getExpr).getAppNumArgs == 4
  let rec go (acc : Array Term) : DelabM Term := do
    let e ← getExpr
    if e.isAppOfArity ``PProd.mk 4 then
      let a ← withAppFn <| withAppArg delab
      withIncDepth <| withAppArg <| go (acc.push a)
    else
      let b ← delab
      `(⟨$acc,*, $b⟩)
  go #[]
```

There's also a version that can flatten expressions involving `PProd`/`Sigma`/`PSigma`. I fear this is going too far however.
```lean
/--
Delaborator for values of the "tuple-like" types `PProd`, `Sigma`, and `PSigma`.
In each case, the constructor has arity 4 and the last two arguments are the values of the tuple.
-/
@[delab app.PProd.mk, delab app.Sigma.mk, delab app.PSigma.mk]
partial def delabTupleLike : Delab := whenNotPPOption getPPExplicit <| whenPPOption getPPNotation do
  guard <| (← getExpr).getAppNumArgs == 4
  let rec go (acc : Array Term) : DelabM Term := do
    let e ← getExpr
    if e.getAppNumArgs == 4 && e.getAppFn.constName ∈ [``PProd.mk, ``Sigma.mk, ``PSigma.mk] then
      let a ← withAppFn <| withAppArg delab
      withIncDepth <| withAppArg <| go (acc.push a)
    else
      let b ← delab
      `(⟨$acc,*, $b⟩)
  go #[]
```

kmill: Oops, deeply nested tuples aren't being handled correctly. It will use the omission term on each element, but not truncate the tuple.

New attempt:
```lean
@[delab app.PProd.mk]
partial def delabPProdMk : Delab := whenNotPPOption getPPExplicit <| whenPPOption getPPNotation do
  guard <| (← getExpr).getAppNumArgs == 4
  let a ← withAppFn <| withAppArg delab
  let b ← withAppArg <| delab
  if (← getExpr).appArg!.isAppOfArity ``PProd.mk 4 then
    if let `(⟨$xs,*⟩) := b then
      return ← `(⟨$a, $xs,*⟩)
  `(⟨$a, $b⟩)
```

kmill: Thanks for asking about the delaborator. I took this opportunity to add a comment explaining the reasoning: #4764

nomeata: Great, I’ve put this in.

PR #4762, opened 2024-07-16
feat: improve `@[ext]` error message when `ext_iff` generation fails

### LINKED ISSUE #4758: actionable error message when `@[ext]` should be `@[ext (iff := false)]`
Typically when `@[ext]` fails to generate the `ext_iff` lemma because of a non-propositional argument, we get an error message just saying

```
argument inst✝ is not a proof, which is not supported
```

(e.g. I encountered this at `tensorObj₄_ext` in `Mathlib.CategoryTheory.Graded.Monoidal` on `nightly-testing`)

but the error message could usefully suggest also: "Try `@[ext (iff := false)]` to disable generation of an `ext_iff` theorem."

kmill:
Now it suggests using `@[ext (iff := false)]` to disable generating the `ext_iff` lemma.

This PR also adjusts error messages and attribute documentation. Additionally, to simplify the code now the `x` and `y` arguments can't come in reverse order (this feature was was added in the refactor #4543).

Closes #4758

### DIFF
--- a/src/Init/Ext.lean
+++ b/src/Init/Ext.lean
@@ -33,7 +33,7 @@ Registers an extensionality theorem.
 * An optional natural number argument, e.g. `@[ext 9000]`, specifies a priority for the `ext` lemma.
   Higher-priority lemmas are chosen first, and the default is `1000`.
 
-* The flag `@[ext (iff := false)]` prevents the generation of an `ext_iff` theorem.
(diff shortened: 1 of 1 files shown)


kmill:
@semorrison What do you think about these error messages? If they're good, feel free to merge, I'm stepping away from the computer for a few hours.

PR #4768, opened 2024-07-17
fix: do not ban `..` with a `.` on the next line

eric-wieser:
Without this change,
```lean
example : True := by
  refine' trivial ..
  . trivial
```
is a parse error.

### DIFF
--- a/tests/lean/run/4768.lean
+++ b/tests/lean/run/4768.lean
@@ -1,3 +1,9 @@
+/-!
+Test that `..` tokens do not break nearby `.`s.
+
+Note that this tests specifically for issues with `.` that are not present with `·`.
+-/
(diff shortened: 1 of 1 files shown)


Kha:
Why is this a draft?

eric-wieser:
It has no tests yet and the description says it should be draft if I haven't filed an issue first.

eric-wieser:
I've added the test from the description

Kha:
Thanks. Please add a modulo docstring to the test describing its purpose as mentioned in testing.md

eric-wieser:
Sorry for the delay, the docstring is now present.

eric-wieser:
awaiting-review

PR #4772, opened 2024-07-17
feat: functional induction for mutual structural recursion

nomeata:


nomeata:
Todo:
* Let-bind duplicated terms (or introduce `.mutual_brecOn`.
* Unify PProd handling in the various modules.

src/Lean/Meta/Tactic/FunInd.lean
```diff
@@ -690,6 +710,22 @@ def deriveUnaryInduction (name : Name) : MetaM Name := do
     { name := inductName, levelParams := us, type := eTyp, value := e' }
   return inductName
 
+-- TODO: put in a better position
+/--  Given `foo.mutual_induct`, defined `foo.induct`, `bar.induct` etc.  -/
```
nomeata, line 714:
```suggestion
/--
Given `foo.mutual_induct`, defined `foo.induct`, `bar.induct` etc.
Used for well-founded and structural recursion.
-/
```

PR #4774, opened 2024-07-17
feat: generalized `Parsec`

hargoniX:
For experimentation by @the-sofi-uwu.

I also have an efficient number parser in LeanSAT that I am planning to upstream after we have sufficiently bikeshed this change.


TwoFX:
By switching around the namespaces and naming you have practically guaranteed to break every single downstream user of `Lean.Parsec`. Do we have an idea how many people are affected by this? Are the error messages comprehensible so that affected users know what they need to do? Can we avoid the breakage entirely in some way?

PR #4781, opened 2024-07-18
fix: make "use `set_option diagnostics true" message conditional on current setting

kmill:
It is confusing that the message suggesting to use the `diagnostics` option is given even when the option is already set. This PR makes use of lazy message data to make the message contingent on the option being false.

It also tones down the promise that there is any diagonostic information available, since sometimes there is nothing to report.

Suggested by Johan Commelin.

### DIFF
--- a/tests/lean/run/1163.lean
+++ b/tests/lean/run/1163.lean
@@ -16,7 +16,7 @@ error: failed to synthesize
 numerals are polymorphic in Lean, but the numeral `0` cannot be used in a context where the expected type is
   Bool
 due to the absence of the instance above
-use `set_option diagnostics true` to get diagnostic information
+Additional diagnostic information may be available by using the `set_option diagnostics true` command.
 -/
 #guard_msgs in
 def x : Bool := 0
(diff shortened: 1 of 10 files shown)


src/Lean/CoreM.lean
```diff
@@ -300,8 +309,10 @@ register_builtin_option debug.moduleNameAtTimeout : Bool := {
 def throwMaxHeartbeat (moduleName : Name) (optionName : Name) (max : Nat) : CoreM Unit := do
   let includeModuleName := debug.moduleNameAtTimeout.get (← getOptions)
   let atModuleName := if includeModuleName then s!" at `{moduleName}`" else ""
-  let msg := s!"(deterministic) timeout{atModuleName}, maximum number of heartbeats ({max/1000}) has been reached\nuse `set_option {optionName} <num>` to set the limit\n{useDiagnosticMsg}"
-  throw <| Exception.error (← getRef) (MessageData.ofFormat (Std.Format.text msg))
+  throw <| Exception.error (← getRef) m!"\
+    (deterministic) timeout{atModuleName}, maximum number of heartbeats ({max/1000}) has been reached\n\
+    Use `set_option {optionName} <num>` to set the limit.\
+    {useDiagnosticMsg}"
```
eric-wieser, line 315: This change caused #5565, which probably has subtle fallout on mathlib. Can a fix make it into a 4.12.1?

PR #4786, opened 2024-07-19
fix: filter duplicate subexpressions

mhuisi:
For every parenthesized expression `(foo)`, the InfoView produces an interactive component both for `(foo)` itself and its subexpression `foo` because the corresponding `TaggedText` in the language server is duplicated as well. Both of these subexpressions have the same subexpression position and so they are identical w.r.t. interactive features. 

Removing this duplication would help reduce the size of the DOM of the InfoView and ensure that the UI for InfoView features is consistent for `(foo)` and `foo` (e.g. hovers would always highlight `(foo)`, not either `(foo)` or `foo` depending on whether the mouse cursor is on the bracket or not). It would also help resolve a bug where selecting a subexpression will yield selection highlighting both for `(foo)` and `foo`, as we use the subexpression position to identify which terms to highlight.

This PR adjusts the parenthesizer to move the corresponding info instead of duplicating it.

### DIFF
--- a/src/Lean/PrettyPrinter/Parenthesizer.lean
+++ b/src/Lean/PrettyPrinter/Parenthesizer.lean
@@ -350,15 +350,17 @@ def term.parenthesizer : CategoryParenthesizer | prec => do
   maybeParenthesize `term true wrapParens prec $
     parenthesizeCategoryCore `term prec
 where
-  /-- Wraps the term `stx` in parentheses and then copies its `SourceInfo` to the result.
-  The purpose of this is to copy synthetic delaborator positions from the `stx` node to the parentheses node,
-  which causes the info view to view both of these nodes as referring to the same expression.
-  If we did not copy info, the info view would consider the parentheses to belong to the outer term.
+  /-- Wraps the term `stx` in parentheses and then moves its `SourceInfo` to the result.
+  The purpose of this is to move synthetic delaborator positions from the `stx` node to the parentheses node,
+  which causes the info view to view the node with parentheses as referring to the parenthesized expression.
+  If we did not move info, the info view would consider the parentheses to belong to the outer term.
   Note: we do not do `withRef stx` because that causes the "(" and ")" tokens to have source info as well,
   causing the info view to highlight each parenthesis as an independent expression. -/
   wrapParens (stx : Syntax) : Syntax := Unhygienic.run do
+    let stxInfo := SourceInfo.fromRef stx
+    let stx := stx.setInfo .none
     let pstx ← `(($(⟨stx⟩)))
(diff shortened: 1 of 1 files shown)


Vtec234:
One question before thinking about the details of filtering: why do this here rather than somewhere earlier? Perhaps it should be viewed a bug in the delaborator or somewhere in the pretty-printer that it produces both of these rather than just one?

kmill:
Here's where the parenthesizer adds parentheses and copies source position from the syntax being parenthesized: https://github.com/leanprover/lean4/blob/master/src/Lean/PrettyPrinter/Parenthesizer.lean#L353-L361

The reason it doesn't delete source position from the syntax is because it's using `SourceInfo.fromRef`, which is willing to descend into the syntax to find something, and we just didn't implement the algorithm to try to correctly delete source position from the inner expression (being able to hover over the parentheses was a good enough improvement on its own). Potentially using `SourceInfo.fromRef` is incorrect here, and it ought to try to only get source position from the syntax node/atom itself, which would make deleting easy.

On the other hand, the filtering approach in the PR might be a way to address that comment in the linked code about why it doesn't use withRef — this PR would keep each parenthesis from being individually hoverable, right?

mhuisi:
> On the other hand, the filtering approach in the PR might be a way to address that comment in the linked code about why it doesn't use withRef — this PR would keep each parenthesis from being individually hoverable, right?

Hovering over the parenthesis will highlight the parentheses and the term inside of it. Hovering on the head of the term inside of it will also highlight the parentheses, whereas before it would only highlight the term, not the parentheses.

mhuisi:
I believe something like the following would also accomplish the same thing as this PR if I am not mistaken:
```
index f2509edd85..71451432f3 100644
--- a/src/Lean/PrettyPrinter/Parenthesizer.lean
+++ b/src/Lean/PrettyPrinter/Parenthesizer.lean
@@ -357,8 +357,10 @@ where
   Note: we do not do `withRef stx` because that causes the "(" and ")" tokens to have source info as well,
   causing the info view to highlight each parenthesis as an independent expression. -/
   wrapParens (stx : Syntax) : Syntax := Unhygienic.run do
+    let stxInfo := SourceInfo.fromRef stx
+    let stx := stx.setInfo .none
     let pstx ← `(($(⟨stx⟩)))
-    return pstx.raw.setInfo (SourceInfo.fromRef stx)
+    return pstx.raw.setInfo stxInfo
 
 @[builtin_category_parenthesizer tactic]
 def tactic.parenthesizer : CategoryParenthesizer | prec => do
 ```
 What do you think?

mhuisi:
(Kyle signed this off in a PM)

PR #4800, opened 2024-07-20
fix: make `elab_as_elim` eagerly elaborate arguments for parameters appearing in the types of targets

kmill:
The `elab_as_elim` elaborator eagerly elaborates arguments that can help with elaborating the motive, however it does not include the transitive closure of parameters appearing in types of parameters appearing in ... types of targets.

This leads to counter-intuitive behavior where arguments supplied to the eliminator may unexpectedly have postponed elaboration, causing motives to be type incorrect for under-applied eliminators such as the following:

```lean
class IsEmpty (α : Sort u) : Prop where
  protected false : α → False

@[elab_as_elim]
def isEmptyElim [IsEmpty α] {p : α → Sort _} (a : α) : p a :=
  (IsEmpty.false a).elim

example {α : Type _} [IsEmpty α] :
  id (α → False) := isEmptyElim (α := α)
```

The issue is that when `isEmptyElim (α := α)` is computing its motive, the value of the postponed argument `α` is still an unassignable metavariable. With this PR, this argument is now among those that are eagerly elaborated since it appears as the type of the target `a`.

This PR also contains some other fixes:
* When underapplied, does unification when instantiating foralls in the expected type.
* When overapplied, type checks the generalized-and-reverted expected type.
* When collecting targets, collects them in the correct order.

Adds trace class `trace.Elab.app.elab_as_elim`.

This is a followup to #4722, which added motive type checking but exposed the eagerness issue.

### DIFF
--- a/src/Lean/Elab/App.lean
+++ b/src/Lean/Elab/App.lean
@@ -807,7 +807,7 @@ def finalize : M Expr := do
           discrs := discrs.set! tidx x
     if let some idx := discrs.findIdx? Option.isNone then
       -- This should not happen.
-      trace[Elab.app.elab_as_elim] "Internal error, missing target at argument index {idx}"
+      trace[Elab.app.elab_as_elim] "Internal error, missing target with index {idx}"
       throwError "failed to elaborate eliminator, insufficient number of arguments"
     trace[Elab.app.elab_as_elim] "discrs: {discrs.map Option.get!}"
     let motiveVal ← mkMotive (discrs.map Option.get!) expectedType


src/Lean/Elab/App.lean
```diff
@@ -761,37 +765,58 @@ def revertArgs (args : List Arg) (f : Expr) (expectedType : Expr) : TermElabM (E
...
+    if discrs.any Option.isNone then
       for i in [idx:idx + xs.size], x in xs do
-        if (← read).elimInfo.targetsPos.contains i then
-          discrs := discrs.push x
-    let motiveVal ← mkMotive discrs expectedType
+        if let some tidx := (← read).elimInfo.targetsPos.indexOf? i then
+          discrs := discrs.set! tidx x
+    if let some idx := discrs.findIdx? Option.isNone then
+      -- This should not happen.
+      trace[Elab.app.elab_as_elim] "Internal error, missing target at argument index {idx}"
```
kmill, line 810:
```suggestion
      trace[Elab.app.elab_as_elim] "Internal error, missing target with index {idx}"
```

PR #4801, opened 2024-07-21
feat: usize for array types

### LINKED ISSUE #4654: RFC: Array.usize
### Proposal

I propose to add `Array.usize` to get the `USize` size of an array, without going through `Nat`.

Currently, to get the `USize` of an array, one needs to use `Array.size` and then `Nat.toUSize`. The latter takes the return value of `Array.size` and calculates the remainder modulo `USize.size`. This is not optimal since arrays are stored in memory and therefore cannot exceed `USize.size`.  This proposal would bypass the remainder step and return the correct array size as an `USize` scalar (which is trivial to access in the current array implementation).

Other array operations have optimizations that bypass this trivial size check for the sake of nearly matching C-language performance. Some of my code at `UnicodeBasic` and elsewhere could benefit a lot from this nearly trivial improvement to the array API.

The drawback is that this adds one more unsafe implementation. So this request needs some careful consideration.

### Impact

fgdorais:
Add efficient `usize` functions for `Array`, `ByteArray`, `FloatArray`.

This is part 1 of 2 since there is a need to update stage0 between the two parts. (See discussion below.)

Closes #4654

### DIFF
--- a/stage0/src/stdlib_flags.h
+++ b/stage0/src/stdlib_flags.h
@@ -18,3 +18,5 @@ options get_default_options() {
     return opts;
 }
 }
+
+// please update stage0


fgdorais:
It was necessary to update stage0 in order to use the `usize` functions.

fgdorais:
awaiting-review

nomeata:
I didn't expect this to need a stage0 update. Why is that so?

fgdorais:
I didn't expect that either. It's because these are `extern` definitions so there is no code for them until stage0 is updated.

nomeata:
Hmm, still somewhat strange, the the C symbol should already be there, shouldn’t it? Did you try setting `prefer_native` to true?

But I’m no expert here. I’m just bringing this up because PRs with stage0 updates need to be manually merged by (I think) Sebastian or Leo. If you’d split it in two PRs then more people can merge and it can go through the normal merge queue.

fgdorais:
I'm not an expert there either. If I don't update stage0 then calls to `usize` look for the reference implementation instead of the extern symbol, but the reference implementation is never compiled.

Yes, I read the manual. I also made sure the commits were clean. I was thinking about splitting in two but I was hoping Sebastian would say something before I did that. Either way, there needs to be a stage0 update between the two commits.

nomeata:
Right, this works of course! It’s just that with two PRs, the stage0 update can happen via the github bot. But not problem for me letting Sebastian merge this PR :-)

fgdorais:
There's a bot for that! I thought Sebastian would have to do it either way...

I will split right away then!

nomeata:
There is a bot these days. It runs automatically after a PR merge whenever it touches `stage0/src/stdlib_flags.h`, or anyone with commit rights can trigger it.

So you could even leave a comment in `stage0/src/stdlib_flags.h` of the form
```
// please update stage0
```
and it’ll happen automatically after the PR enters master (and it will nicely override this comment).

src/Init/Data/Array/Basic.lean
```diff
@@ -50,6 +50,13 @@ instance : Inhabited (Array α) where
 def singleton (v : α) : Array α :=
   mkArray 1 v
 
+/-- Low-level version of `size` that directly queries the C array object cached size.
+    While this is not provable, `usize` always returns the exact size of the array since
+    the implementation only supports arrays of size less than `USize.size`.
+-/
+@[extern "lean_array_size", simp]
+def usize (a : @& Array α) : USize := a.size.toUSize
```
nomeata, line 58: It seems that other functions from C world that are used via `extern` have a
`LEAN_EXPORT` declaration in the corresponding header file. Did you intentionally not add one of these for the new functions?

fgdorais: I think `LEAN_EXPORT` only matters for symbols exported to a shared library. That doesn't seem necessary in this case. For example, `lean_array_get_size` (which implements `Array.size`) does not have a `LEAN_EXPORT`.

nomeata: Great, thanks for checking.

nomeata (approved):

fgdorais:
Thank you so much for your help Joachim! This PR would have stayed in limbo for a long time without your enlightening comments!

PR #4802, opened 2024-07-21
feat: use usize for array types

### LINKED ISSUE #4654: RFC: Array.usize

fgdorais:
This is part 2 of 2 of #4801 (which closes #4654). That PR was split in two to allow a stage0 update between declaring the `usize` functions and using them where they are needed.


fgdorais:
awaiting-review

PR #4810, opened 2024-07-23
feat: safer #eval, and #eval!

### LINKED ISSUE #1697: `#eval` evaluates term with failing tactics
```lean
#eval show Nat from False.rec (by decide)
```

- In the server the file crashes.
- On the command line it prints `unreachable code` (original report produced `incomplete case`)

We should probably not eval terms where the elaborator fails.

Zulip: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/crash.20when.20using.20.22by.20decide.22/near/301853648

nomeata:
previously, `#eval` would happily evaluate expressions that contain
`sorry`, either explicitly or because of failing tactics. In conjunction
with operations like array access this can lead to the lean process
crashing, which isn't particularly great.

So how `#eval` will refuse to run code that (transitively) depends on
the `sorry` axiom (using the same code as `#print axioms`).

If the user really wants to run it, they can use `#eval!`.

Closes #1697

### DIFF
--- a/tests/lean/terminationFailure.lean
+++ b/tests/lean/terminationFailure.lean
@@ -8,8 +8,8 @@ where
   | 0 => 2
   | x => f x * 2
 
-#check! f
-#check! f.g
+#check f
+#check f.g
 
 #eval! f 0
 #eval! f.g 0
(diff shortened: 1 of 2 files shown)


nomeata:
Actually I’m feeling confident enough, I’ll just merge :-)

Issue #2789, opened 2023-10-30
Mixing binder updates and variable declarations gives confusing errors


eric-wieser:
### Prerequisites

* [x] Put an X between the brackets on this line if you have done all of the following:
    * Check that your issue is not already [filed](https://github.com/leanprover/lean4/issues).
    * Reduce the issue to a minimal, self-contained, reproducible test case. Avoid dependencies to mathlib4 or std4.

### Description

The `variable` command is used both to declare new variables, and to update the binder type (implicit / explicit / instance implicit) of existing variables.

Using both features at the same time can produce error messages that advise the user to make things even worse

### Context

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/unexpected.20.60variable.60.20behaviour.3F/near/360154080)

### Steps to Reproduce

Run the following code
```lean
class one (A : Nat)
class two (A : Nat) [one A] := t : Nat

variable (A : Nat) [one A]

section fails
variable {A} [two A]
/-
failed to synthesize instance
  one A
-/
end fails

section bad_advice
-- ok, the error told me to write `one A` again I guess
variable {A} [one A] [two A]

example : True := 
  -- put your cursor here; now `one A` is here twice!
  trivial

end bad_advice

section works
variable {A} [two (by exact A)]
end works

section also_works
variable {A}
variable [two A]
end also_works
```

**Expected behavior:** The `section fails` should either succeed, or it should emit an error message that does not encourage me to write `section bad_advice`. If you put your cursor at the `exact`, `A` should appear once in the goal view.
**Actual behavior:** The error message tricks me into making a mess of duplicate variables. Putting the cursor at `exact` shows `A` twice (as there are two copies!).

### Versions

4.2.0-rc4 on the web editor

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

Kha:
Fixed in #5142

PR #4814, opened 2024-07-23
feat: refined `variable` behavior in proofs

### LINKED ISSUE #2452: RFC: change variable inclusion mechanism

Kha:
Resolves #2452. For *theorems*, section variables are now conditionally included when elaborating the theorem body and sending the result to the kernel based on their use in the header; elaboration of the header is unchanged. *Definitions* as well are unchanged for now because the cost/benefit ratio is much less clear for them.

Specifically, section variables are included if they
* are directly referenced by the theorem header,
* are included via the new `include` command in the current section,
* are directly referenced by any variable included by these rules, OR
* are instance-implicit variables that reference only variables included by these rules.

For porting, a new option `deprecated.oldSectionVars` is included to locally switch back to the old behavior.

**breaking change**

### DIFF
--- a/src/Init/Data/List/Erase.lean
+++ b/src/Init/Data/List/Erase.lean
@@ -354,7 +354,7 @@ theorem erase_eq_iff [LawfulBEq α] {a : α} {l : List α} :
   rw [erase_of_not_mem]
   simp_all
 
-theorem Nodup.erase_eq_filter [BEq α] [LawfulBEq α] {l} (d : Nodup l) (a : α) : l.erase a = l.filter (· != a) := by
+theorem Nodup.erase_eq_filter [LawfulBEq α] {l} (d : Nodup l) (a : α) : l.erase a = l.filter (· != a) := by
   induction d with
   | nil => rfl
   | cons m _n ih =>
@@ -367,13 +367,13 @@ theorem Nodup.erase_eq_filter [BEq α] [LawfulBEq α] {l} (d : Nodup l) (a : α)
       simpa [@eq_comm α] using m
     · simp [beq_false_of_ne h, ih, h]
 
-theorem Nodup.mem_erase_iff [BEq α] [LawfulBEq α] {a : α} (d : Nodup l) : a ∈ l.erase b ↔ a ≠ b ∧ a ∈ l := by
+theorem Nodup.mem_erase_iff [LawfulBEq α] {a : α} (d : Nodup l) : a ∈ l.erase b ↔ a ≠ b ∧ a ∈ l := by
   rw [Nodup.erase_eq_filter d, mem_filter, and_comm, bne_iff_ne]
 
-theorem Nodup.not_mem_erase [BEq α] [LawfulBEq α] {a : α} (h : Nodup l) : a ∉ l.erase a := fun H => by
+theorem Nodup.not_mem_erase [LawfulBEq α] {a : α} (h : Nodup l) : a ∉ l.erase a := fun H => by
   simpa using ((Nodup.mem_erase_iff h).mp H).left
 
-theorem Nodup.erase [BEq α] [LawfulBEq α] (a : α) : Nodup l → Nodup (l.erase a) :=
+theorem Nodup.erase [LawfulBEq α] (a : α) : Nodup l → Nodup (l.erase a) :=
   Nodup.sublist <| erase_sublist _ _
 
 end erase
--- a/src/Init/WF.lean
+++ b/src/Init/WF.lean
@@ -68,6 +68,7 @@ noncomputable def recursion {C : α → Sort v} (a : α) (h : ∀ x, (∀ y, r y
   induction (apply hwf a) with
   | intro x₁ _ ih => exact h x₁ ih
 
+include hwf in
 theorem induction {C : α → Prop} (a : α) (h : ∀ x, (∀ y, r y x → C y) → C x) : C a :=
   recursion hwf a h
 
--- a/src/Std/Data/DHashMap/Internal/RawLemmas.lean
+++ b/src/Std/Data/DHashMap/Internal/RawLemmas.lean
@@ -20,7 +20,7 @@ open Std.DHashMap.Internal.List
 
 universe u v
 
-variable {α : Type u} {β : α → Type v} [BEq α] [Hashable α]
+variable {α : Type u} {β : α → Type v}
 
 namespace Std.DHashMap.Internal
 
@@ -41,6 +41,8 @@ theorem Raw.buckets_emptyc {i : Nat} {h} :
     (∅ : Raw α β).buckets[i]'h = AssocList.nil :=
   buckets_empty
 
+variable [BEq α] [Hashable α]
+
 @[simp]
 theorem buckets_empty {c} {i : Nat} {h} :
     (empty c : DHashMap α β).1.buckets[i]'h = AssocList.nil := by
@@ -55,7 +57,9 @@ end empty
 
 namespace Raw₀
 
+variable [BEq α] [Hashable α]
 variable (m : Raw₀ α β) (h : m.1.WF)
+set_option deprecated.oldSectionVars true
 
 /-- Internal implementation detail of the hash map -/
 scoped macro "wf_trivial" : tactic => `(tactic|
--- a/src/Std/Data/DHashMap/RawLemmas.lean
+++ b/src/Std/Data/DHashMap/RawLemmas.lean
@@ -75,6 +75,7 @@ namespace Raw
 open Internal.Raw₀ Internal.Raw
 
 variable {m : Raw α β} (h : m.WF)
+set_option deprecated.oldSectionVars true
 
 @[simp]
 theorem isEmpty_empty {c} : (empty c : Raw α β).isEmpty := by
--- a/src/Std/Data/HashMap/RawLemmas.lean
+++ b/src/Std/Data/HashMap/RawLemmas.lean
@@ -27,6 +27,7 @@ namespace Std.HashMap
 namespace Raw
 
 variable {m : Raw α β} (h : m.WF)
+set_option deprecated.oldSectionVars true
 
 private theorem ext {m m' : Raw α β} : m.inner = m'.inner → m = m' := by
   cases m; cases m'; rintro rfl; rfl
--- a/src/Std/Data/HashSet/Lemmas.lean
+++ b/src/Std/Data/HashSet/Lemmas.lean
@@ -20,7 +20,7 @@ set_option autoImplicit false
 
 universe u v
 
-variable {α : Type u} {_ : BEq α} {_ : Hashable α} [EquivBEq α] [LawfulHashable α]
+variable {α : Type u} {_ : BEq α} {_ : Hashable α}
 
 namespace Std.HashSet
 
--- a/src/Std/Data/HashSet/RawLemmas.lean
+++ b/src/Std/Data/HashSet/RawLemmas.lean
@@ -27,6 +27,7 @@ namespace Std.HashSet
 namespace Raw
 
 variable {m : Raw α} (h : m.WF)
+set_option deprecated.oldSectionVars true
 
 private theorem ext {m m' : Raw α} : m.inner = m'.inner → m = m' := by
   cases m; cases m'; rintro rfl; rfl


fpvandoorn:
> are instance-implicit variables that reference any variable included by these rules.

Should this be "are instance-implicit variables and all referenced variables are included by these rules."
For example, if I have 
```lean
variable {R : Type*} [CommRing R] {M : Type*} [AddCommGroup M] [Module R M]
```
If only one of `R`/`M` is referenced, `Module R M` should not be included.

(I would also love it if `include` supports binder updates)

Kha:
> > are instance-implicit variables that reference any variable included by these rules.
> 
> Should this be "are instance-implicit variables and all referenced variables are included by these rules."? For example, if I have

Makes sense to me, updated.

> (I would also love it if include supports binder updates)

Would `include {v}` be equivalent to `variable {v} include v`? I'm not sure yet what to think about such overloading of semantics

fpvandoorn:
Yes, those would be equivalent.

In Mathlib we are striving towards using `variable {v}` and `include v` *only* using `in` (i.e. `variable {v}` or `include v` is banned, you should solely use `variable {v} in` and `include v in`).
If we follow that rule, then the ability to write `include {v}` means that we don't need `variable {v}` anymore in Mathlib, since all its uses can be replaced by `include`.

semorrison:
> Yes, those would be equivalent.
> 
> In Mathlib we are striving towards using `variable {v}` and `include v` _only_ using `in` (i.e. `variable {v}` or `include v` is banned, you should solely use `variable {v} in` and `include v in`). If we follow that rule, then the ability to write `include {v}` means that we don't need `variable {v}` anymore in Mathlib, since all its uses can be replaced by `include`.

@fpvandoorn, I've finally started looking at Mathlib under this PR (the `lean-pr-testing-4814` branch). (It is going to be epic! :-)
 
I've already encountered a few places where is seems rather inconvenient for only using `include ... in`, rather than bounded by a section. Please see https://github.com/leanprover-community/mathlib4/commit/c3ec6bedfaddc035e24f86a44099089854a7d13c for the first two examples I encountered.

semorrison:
Re: looking at Mathlib on this PR. So far there are two classes of problems: typeclasses hypothesis being included which should not be (for which the new linter is *really* helpful), and variables with need an `include`.

Fixing the first class of problems requires either adding sections to bound `variable [...]` scopes, using `variable ... in`, or reordering results.

I propose that for anyone working on the Mathlib branch for this PR, we try to backport all the changes of the first form.

A reasonable workflow seems to be:
* As you make changes on `lean-pr-testing-4814`, any changes that use `include` (i.e. which will not work on master), you immediately stage.
* But any changes for *excluding* typeclasses hypotheses you leave unstage.
* At suitable intervals, make a commit of the stages changes on `lean-pr-testing-4814`
* Then stash the remaining changes, make a new branch off master, make a PR from it, and then merge that branch back to `lean-pr-testing-4814`.

(We have have following a similar workflow for the `byAsSorry` branch, c.f. [zulip](https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Running.20Mathlib.20under.20.60set_option.20debug.2EbyAsSorry.60) discusssion.)

This is going to be a *massive* adaptation exercise to get this working, and I will need help to get it done.

semorrison:
> > > are instance-implicit variables that reference any variable included by these rules.
> > 
> > 
> > Should this be "are instance-implicit variables and all referenced variables are included by these rules."? For example, if I have
> 
> Makes sense to me, updated.

@Kha, I think this change may have not worked as intended. In Mathlib on lean-pr-testing-4814 I am seeing:
```
section Trans

variable [IsTrans N r] (m n : M) {a b c d : N}

--  Lemmas with 3 elements.
theorem act_rel_of_rel_of_act_rel (ab : r a b) (rl : r (μ m b) c) : r (μ m a) c :=
  _root_.trans (act_rel_act_of_rel m ab) rl

theorem rel_act_of_rel_of_rel_act (ab : r a b) (rr : r c (μ m a)) : r c (μ m b) :=
  _root_.trans rr (act_rel_act_of_rel _ ab)

end Trans
```
failing with `failed to synthesize IsTrans N r`, despite both the arguments of `IsTrans` being mentioned in the theorem statement.

(I have no idea why in Mathlib the first argument of `IsTrans` is explicit rather than implicit -- I'll investigate that too...)

Let me know if you'd like a minimization of this.

semorrison:
@Kha, another bug, where typeclass arguments that are included transitively can be include with multiplicity, apparently!

Again on `lean-pr-testing-4814`, in `Mathlib.Order.Monotone`, we see:

```
variable (α) [Preorder α] [Nonempty α] [NoMinOrder α] [NoMaxOrder α]

/-- If `α` is a nonempty preorder with no minimal or maximal elements, then there exists a strictly
monotone function `f : ℤ → α`. -/
theorem exists_strictMono : ∃ f : ℤ → α, StrictMono f := by
  inhabit α
  rcases Nat.exists_strictMono' (default : α) with ⟨f, hf, hf₀⟩
  rcases Nat.exists_strictAnti' (default : α) with ⟨g, hg, hg₀⟩
  refine ⟨fun n ↦ Int.casesOn n f fun n ↦ g (n + 1), strictMono_int_of_lt_succ ?_⟩
  rintro (n | _ | n)
  · exact hf n.lt_succ_self
  · show g 1 < f 0
    rw [hf₀, ← hg₀]
    exact hg Nat.zero_lt_one
  · exact hg (Nat.lt_succ_self _)
```

resulting in the signature
```
Int.exists_strictMono.{u} (α : Type u) [Preorder α] [Preorder α] [Nonempty α] [NoMinOrder α] [NoMaxOrder α] :
  ∃ f, StrictMono f
```
where apparently `[Preorder α]` has been added twice, once because `[NoMinOrder α]` needs it and again because  `[NoMaxOrder α]` needs it.

Again, please let me know if you'd like an minimisation.

semorrison:
And another problem, @Kha! We have places in Mathlib where people have add implicit arguments that could otherwise be automatically included instance arguments, on the principle that it is faster/better to find these by unification from later arguments that to redo typeclass search.

Here's an example from Mathlib in `Mathlib.Algebra.GroupWithZero`:
```
@[simp]
lemma coe_copy {_ : MulZeroOneClass α} {_ : MulZeroOneClass β} (f : α →*₀ β) (f' : α → β) (h) :
    (f.copy f' h) = f' := rfl

lemma copy_eq {_ : MulZeroOneClass α} {_ : MulZeroOneClass β} (f : α →*₀ β) (f' : α → β) (h) :
    f.copy f' h = f := DFunLike.ext' h
```
It's perfectly okay on `master` to just omit these, and instead we get corresponding instance arguments included.

However on lean-pr-testing-4814, we get warning, because the signature becomes:
```
MonoidWithZeroHom.coe_copy.{u_2, u_3} {α : Type u_2} {β : Type u_3} [MulZeroOneClass α] [MulZeroOneClass β] :
  ∀ {x : MulZeroOneClass α} {x_1 : MulZeroOneClass β} (f : α →*₀ β) (f' : α → β) (h : f' = ⇑f), ⇑(f.copy f' h) = f'
```

I'm uncertain the best way to proceed here. Probably we ought to make the variable inclusion mechanism robust to this, and decline to include instance arguments if there are already there as implicit arguments. In this particular case it's fine to just remove them, but I don't think that is always going to be true.

Again, let me know if you want a minimization.

semorrison:
And more. :-)

We may need to pull in typeclass arguments because they mention the *type* of an included argument, not just the argument itself.

For example in Mathlib we have in `Mathlib.Algebra.Ring.Divisibility.Basic`:
```
variable [Semigroup α] [Semigroup β] {F : Type*} [EquivLike F α β] [MulEquivClass F α β] (f : F)

theorem map_dvd_iff {a b} : f a ∣ f b ↔ a ∣ b :=
  let f := MulEquivClass.toMulEquiv f
  ⟨fun h ↦ by rw [← f.left_inv a, ← f.left_inv b]; exact map_dvd f.symm h, map_dvd f⟩

theorem MulEquiv.decompositionMonoid [DecompositionMonoid β] : DecompositionMonoid α where
  primal a b c h := by
    rw [← map_dvd_iff f, map_mul] at h
    obtain ⟨a₁, a₂, h⟩ := DecompositionMonoid.primal _ h
    refine ⟨symm f a₁, symm f a₂, ?_⟩
    simp_rw [← map_dvd_iff f, ← map_mul, eq_symm_apply]
    iterate 2 erw [(f : α ≃* β).apply_symm_apply]
    exact h
```

Here `include f` on `MulEquiv.decompositionMonoid` (or just writing it as an explicit argument) doesn't suffice to pull in `[MulEquivClass F α β]`.

Making matters worse, if we just change the theorem statement to
```
theorem MulEquiv.decompositionMonoid 
    [MulEquivClass F α β] (f : F) [DecompositionMonoid β] : DecompositionMonoid α where
```
then we get the incorrect warning:
```
included section variable '[MulEquivClass F α β]' is not used in 'MulEquiv.decompositionMonoid', consider excluding it
```
I can work around this by just not making it a section variable at all and instead including it directly in both theorems here.

(As above with minimization. :-)

fpvandoorn:
> > Yes, those would be equivalent.
> > In Mathlib we are striving towards using `variable {v}` and `include v` _only_ using `in` (i.e. `variable {v}` or `include v` is banned, you should solely use `variable {v} in` and `include v in`). If we follow that rule, then the ability to write `include {v}` means that we don't need `variable {v}` anymore in Mathlib, since all its uses can be replaced by `include`.
> 
> @fpvandoorn, I've finally started looking at Mathlib under this PR (the `lean-pr-testing-4814` branch). (It is going to be epic! :-)
> 
> I've already encountered a few places where is seems rather inconvenient for only using `include ... in`, rather than bounded by a section. Please see [leanprover-community/mathlib4@c3ec6be](https://github.com/leanprover-community/mathlib4/commit/c3ec6bedfaddc035e24f86a44099089854a7d13c) for the first two examples I encountered.

We should probably not strive to do this immediately. Another thing we can consider is to have (in the long-term) no explicit variables, instead repeating those in each theorem statement. That should also reduce the number of `include`s.
But given your example we maybe want to allow having a global `include` directly after the corresponding `variable` statement. Then whenever that variable is visible it will have the include statement.

Kha:
> @Kha, another bug, where typeclass arguments that are included transitively can be included with multiplicity, apparently!
> 
> Again on `lean-pr-testing-4814`, in `Mathlib.Order.Monotone`, we see:
> 
> ```
> variable (α) [Preorder α] [Nonempty α] [NoMinOrder α] [NoMaxOrder α]
> 
> /-- If `α` is a nonempty preorder with no minimal or maximal elements, then there exists a strictly
> monotone function `f : ℤ → α`. -/
> theorem exists_strictMono : ∃ f : ℤ → α, StrictMono f := by
>   inhabit α
>   rcases Nat.exists_strictMono' (default : α) with ⟨f, hf, hf₀⟩
>   rcases Nat.exists_strictAnti' (default : α) with ⟨g, hg, hg₀⟩
>   refine ⟨fun n ↦ Int.casesOn n f fun n ↦ g (n + 1), strictMono_int_of_lt_succ ?_⟩
>   rintro (n | _ | n)
>   · exact hf n.lt_succ_self
>   · show g 1 < f 0
>     rw [hf₀, ← hg₀]
>     exact hg Nat.zero_lt_one
>   · exact hg (Nat.lt_succ_self _)
> ```
> 
> resulting in the signature
> 
> ```
> Int.exists_strictMono.{u} (α : Type u) [Preorder α] [Preorder α] [Nonempty α] [NoMinOrder α] [NoMaxOrder α] :
>   ∃ f, StrictMono f
> ```
> 
> where apparently `[Preorder α]` has been added twice, once because `[NoMinOrder α]` needs it and again because `[NoMaxOrder α]` needs it.
> 
> Again, please let me know if you'd like an minimisation.

This looks correct in that there is already a `[Preorder \a]` in context. But also #2789 comes in and prevents us from removing the duplicate assumption unless broken into two lines.
```
variable (α)
variable [Nonempty α] [NoMinOrder α] [NoMaxOrder α]
```

Kha:
> Probably we ought to make the variable inclusion mechanism robust to this, and decline to include instance arguments if there are already there as implicit arguments.

I'm a bit wary of comparing parameter types by equality (which one?). I assume one alternative would be to name the instances and use binder updates instead?

Kha:
> failing with `failed to synthesize IsTrans N r`, despite both the arguments of `IsTrans` being mentioned in the theorem statement.

Fixed



> Here `include f` on `MulEquiv.decompositionMonoid` (or just writing it as an explicit argument) doesn't suffice to pull in `[MulEquivClass F α β]`.

Fixed

semorrison:
I am rebasing on `nightly-with-mathlib` (regenerating `update-stage0` commits), in order to keep the Mathlib adaptation branch viable.

semorrison:
Sigh, the merge conflict with master occurs after `nightly-with-mathlib`. I will see if Mathlib's `nightly-testing` is compatible with `master`, and if so rebase this onto `master`.

src/Lean/Parser/Command.lean
```diff
@@ -274,30 +274,35 @@ with `end <id>`. The `end` command is optional at the end of a file.
...
+useful in particular when writing many definitions that have parameters in common (see below for an
+example).
 
 Variable declarations have the same flexibility as regular function paramaters. In particular they
 can be [explicit, implicit][binder docs], or [instance implicit][tpil classes] (in which case they
 can be anonymous). This can be changed, for instance one can turn explicit variable `x` into an
 implicit one with `variable {x}`. Note that currently, you should avoid changing how variables are
 bound and declare new variables at the same time; see [issue 2789] for more on this topic.
 
+In *theorem bodies* (i.e. proofs), variables are not included based on usage in order to ensure that
```
Kha, line 290: @david-christiansen I merged this version for now to avoid a stale update-stage0, but please feel free to refine!

PR #4817, opened 2024-07-24
fix: make `elabAsElim` aware of explicit motive arguments

### LINKED ISSUE #4347: The motive argument to `False.rec` is explicit but elaboration does not use a supplied positional argument as the motive
### Description

One normally expects the motive argument for eliminators to be implicit, but for `False.rec` it is explicit. This interacts in a confusing way with the app eliminator since this explicit argument does not make use of the supplied positional arguments.

### Context

This question has come up a couple times recently: Zulip [discussion 1](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Signature.20of.20.60False.2Erec.60/near/441354391) and [discussion 2](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Need.20of.20.40.20in.20False.2Erec/near/442638039)

### Steps to Reproduce

As one can see, the type of `False.rec` is
```lean
#check False.rec -- False.rec.{u} (motive : False → Sort u) (t : False) : motive t
```
It has an explicit argument for the motive.

However, using `False.rec` with an explicit motive throws an error:
```lean
def bar (h : False) : Nat := False.rec (fun _ => Nat) h
/-
application type mismatch
  False.rec ?m.29 fun x => Nat
argument
  fun x => Nat
has type
  ?m.31 → Type : Sort (max 2 ?u.30)
but is expected to have type
  False : Prop
-/
```
This is because it's being elaborated as an eliminator, which causes the elaborator to solve for the motive and ignore that it's a positional argument.

The expected behavior is for explicit arguments to be filled positionally from supplied arguments.

A workaround is to write `@False.rec (fun _ => Nat) h` instead to disable elaborating as an eliminator.

### Versions

4.8.0-rc2

### Discussion

These are some potential solutions:
1. Make sure that the motive argument for `False.rec` is implicit.
2. Make the ElabAsElim code check to see if the motive argument is explicit, and if it is, use a positional argument. Users can supply `_` if it is meant to be solved for. If `x : False`, then `x.rec` on its own should be the same as `False.rec _ x`.

They are not mutually exclusive, and even if the first is implemented, the second could still be useful for user `@[elab_as_elim]` declarations.

### Impact

kmill:
Some eliminators (such as `False.rec`) have an explicit motive argument. The `elabAsElim` elaborator assumed that all motives are implicit.

If the explicit motive argument is `_`, then it uses the elab-as-elim procedure, and otherwise it falls back to the standard app elaborator.

Furthermore, if an explicit elaborator is not provided, it falls back to treating the elaborator as being implicit, which is convenient for writing `h.rec` rather than `h.rec _`. Rationale: for `False.rec`, this simulates it having an implicit motive, and also motives are generally not going to be available in the expected type.

Closes #4347

### DIFF
--- a/tests/lean/run/1697.lean
+++ b/tests/lean/run/1697.lean
@@ -8,7 +8,7 @@ error: cannot evaluate expression that depends on the `sorry` axiom.
 Use `#eval!` to evaluate nevertheless (which may cause lean to crash).
 -/
 #guard_msgs in
-#eval show Nat from False.rec (by decide)
+#eval show Nat from False.elim (by decide)
 
 /--
 warning: declaration uses 'sorry'


nomeata:
Just based off the PR description: did you consider the alternative of making all motives implicit? Or is there a good reason to make them explicit it some cases?

kmill:
@nomeata Yes, and that's a good question that's worth discussing here. First, there's [Leo's comment](https://github.com/leanprover/lean4/issues/4347#issuecomment-2148585463) about how eliminators like `False.rec` come from the kernel, and so we should fix the elaborator. Second, user eliminators do sometimes have explicit motives for whatever reason.

nomeata:
Thanks, these are good points.

PR #4818, opened 2024-07-24
fix: make sure anonymous dot notation works with pi-type-valued type synonyms

### LINKED ISSUE #4761: Anonymous dot notation fails on propositions defined by `∀`
### Prerequisites

### Description

```
def Foo : Prop := ∀ a : Nat, a = a

protected theorem Foo.intro : Foo := sorry
example : Foo := .intro
```
produces
```
invalid dotted identifier notation, unknown identifier `Eq.intro` from expected type
  Foo
```
on the `example`

**Expected behavior:** Anonymous dot notation works here

**Actual behavior:** Anonymous dot notation doesn't work here

### Context

In Mathlib, anonymous dot notation on `Monotone` fails due to this bug. See eg https://github.com/leanprover-community/mathlib4/pull/13338#discussion_r1660087804.

### Versions

4.10.0-rc2

### Impact

kmill:
When resolving anonymous dot notation (`.ident x y z`), it would reduce the expected type to whnf. Now, it unfolds definitions step-by-step, even if the type synonym is for a pi type like so
```lean
def Foo : Prop := ∀ a : Nat, a = a
protected theorem Foo.intro : Foo := sorry
example : Foo := .intro
```

Closes #4761

### DIFF
--- a/src/Lean/Elab/App.lean
+++ b/src/Lean/Elab/App.lean
@@ -1357,7 +1357,7 @@ private partial def resolveDotName (id : Syntax) (expectedType? : Option Expr) :
   withForallBody expectedType fun resultType => do
     go resultType expectedType #[]
 where
-  /-- A weak version of forallTelescopeReducing that only uses whnfCore. -/
+  /-- A weak version of forallTelescopeReducing that only uses whnfCore, to avoid unfolding definitions except by `unfoldDefinition?` below. -/
   withForallBody {α} (type : Expr) (k : Expr → TermElabM α) : TermElabM α :=
     forallTelescope type fun _ body => do
       let body ← whnfCore body
--- a/tests/lean/run/4761.lean
+++ b/tests/lean/run/4761.lean
@@ -6,3 +6,12 @@ def Foo : Prop := ∀ a : Nat, a = a
 
 protected theorem Foo.intro : Foo := sorry
 example : Foo := .intro
+
+/-!
+Making sure that the type synonym `Foo'` still works.
+This used to be unfolded during `forallTelescopeReducing`,
+and now it is "manually" unfolded in the elaboration algorithm.
+-/
+example : Nat → Option Nat := .some
+def Foo' := Nat → Option Nat
+example : Foo' := .some


src/Lean/Elab/App.lean
```diff
@@ -1354,9 +1354,17 @@ private partial def resolveDotName (id : Syntax) (expectedType? : Option Expr) :
   tryPostponeIfNoneOrMVar expectedType?
   let some expectedType := expectedType?
     | throwError "invalid dotted identifier notation, expected type must be known"
-  forallTelescopeReducing expectedType fun _ resultType => do
```
nomeata, line 1357: Just curious: Could we have used `forallTelescope` if we had a `TransparencyMode` none?

Or would using `withReducible` do the right thing here! Or do we want people to use dot notation with reducible terms (I guess we do, commonly with `abbrev`, right?)

kmill: This made it mirroring how `Lean.Elab.Term.consumeImplicits` works, which also uses whnfCore and manual definition unfolding. It would be nice to have a version of forallTelescope where you can supply your own reduction function — that's something I was thinking about, but didn't want to touch that core meta code for this.

Supposing reducible transparency was ok here, a problem with withReducible is that it applies to the entire computation (plus the continuation!), where we just want to delimit it to the internal mechanism of forallTelescopeReducing.

nomeata: Yeah, two continuation-style functions don't compose well. Could be a ReducibilityMode option to `forallTelescopeReducing`. (Not important)

kmill: Yeah, it could be, but better would be reduction function rather than configuration, since we do want whnfCore here rather than whnf. Without measuring, I worry a bit about the performance impact though since the function won't be specialized to whnf, and forallTelescope(Reducing) is used everywhere.

kmill: On second thought, there's no need to adjust forallTelescopeReducingAuxAux itself, since you can use the pattern in this PR. I don't think there are really any optimizations from inserting a custom whnf function into it.

(This PR also continues on with whnfCore of the body, vs forallTelescopeReducing which gives you the unreduced body if it didn't reduce to a forall.)

PR #4821, opened 2024-07-24
fix: language server windows issues

### LINKED ISSUE #3786: server hangs when changing imports
### Description

When changing the imports of a file, Lean has to re-import the files. However, sometimes the Lean server (in VSCode) will just hang forever showing yellow bars, not making any progress for many mintutes. Reloading the window or restarting the file in VSCode will cause the Lean server to import the files correctly and be ready for use after a matter of seconds.

I have noticed this multiple times myself on Windows, and I have noticed this with various students and other people, I believe always on Windows. I *think* this is a Windows-only issue. EDIT: at least one Mac user also experiences this (see Zulip thread below).

I don't have a way to reliably reproduce this.

Possibly related: leanprover/lean4#3787

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Lean.20hangs.20when.20changing.20imports)

### Versions

`vscode-lean v0.0.135` (and many other versions)
`Lean 4.7.0-rc2` (and many other versions)
`Windows 11 Home`

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/vscode-lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

### LINKED ISSUE #3787: server out of sync with file when editing imports on Windows
### Description

On Windows, the following steps reliably cause the server to go out of sync with the file. This issue is not reproducible on Linux.

### Context

[Zulip thread](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Server.20out.20of.20sync.20with.20file)

### Steps to Reproduce

1. Download mathlib and its cache (I can reproduce on v4.2.0, but presumably also every other version)
2. Create a new file with the following contents, and wait until Lean has compiled the file
```
import Mathlib
#eval 1+1
```
3. Delete the first line all at once (e.g. with ctrl+X). On 2 separate Windows computers this does *not* cause a recompilation of the file
4. Type the letter `a` on the first line and wait until Lean shows an error
5. Remove the letter `a`. The error remains. The error remains even if you start typing new commands (e.g. `def foo := 1`), and the server is interpreting as a letter `a` floating somewhere in your file. It disappears once you write `import` again.

I experimented with importing less than `Mathlib`, and below a certain number of imports, this behavior seems to go away. (e.g. if we replace `import Mathlib` with `import Std`, then step 3 does cause a recompilation of the file.

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/vscode-lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

mhuisi:
This PR resolves two language server bugs that especially affect Windows users:
1. Editing the header could result in the watchdog not correctly restarting the file worker (#3786, #3787), which would lead to the file seemingly being processed forever. 
  - The cause of this issue was a race condition in the watchdog that was accidentally introduced as far back as #1884: In specific circumstances, the watchdog will attempt forwarding a message to the file worker after the process has exited due to a changed header, but before the file worker exiting has been noticed by the watchdog (which will then restart the file worker). In this case, the watchdog would mark the file worker as having crashed and not look at its exit code to restart the file worker, but instead treat it like a crashed file worker that will only be restarted when editing the file again. Not inspecting the exit code of the file worker when it crashed from forwarding a message from the file worker is necessary since we do not restart the file worker until another notification from the client arrives, and so we would read the same crash exit code over and over again in the main loop of the watchdog if we did not remove it from our list of file workers that we listen to.
  - This PR resolves this issue by distinguishing between "crashes when forwarding messages to the file worker" and "crashes when forwarding messages from the file worker". In the former case, we still inspect the exit code of the file worker and potentially restart it if the imports changed, whereas in the latter case, we stop inspecting the exit code of the file worker. This is correct because the latter case is exactly the one where we need to stop inspecting the exit code but where a crash cannot occur as a result of a changed header, whereas the former case is exactly the one where we still need to inspect the exit code after a crash to ensure that we restart the file worker in case it exited because the header changed.
  - At some point in the future, it would be nice to revamp the concurrency model of the watchdog entirely now that we have all those fancy concurrency primitives that were not available four years ago when the watchdog was first written. 

2. On an especially slow Windows machine, we found that starting the language server would sometimes not succeed at all because reading from the stdin pipe in the watchdog produced an EINVAL error, which was in turn caused by an NT "pipe empty" error.  
  - After lots of debugging, @Kha found that Lake accidentally passes its stdin to Git because it does not explicitly set the `stdin` field to `null` when spawning the process. 
  - Changing this fixes the issue, which suggests that Git may mutate the pipe we pass to it to be non-blocking, which then causes a "pipe empty" error in the watchdog when we also attempt to read from that same pipe.
  - I'm still very uncertain why we only saw this issue on one particularly slow machine and not across the whole eco system.

This PR also resolves an issue where we would not correctly emit messages that we received while the file worker is being restarted to the corresponding file worker after the restart.

Closes #3786, closes #3787.

### DIFF
--- a/src/Lean/Server/Watchdog.lean
+++ b/src/Lean/Server/Watchdog.lean
@@ -140,6 +140,11 @@ section FileWorker
     for ⟨id, _⟩ in pendingRequests do
       hError.writeLspResponseError { id := id, code := code, message := msg }
 
+  def queuedMsgs (fw : FileWorker) : Array JsonRpc.Message :=
+    match fw.state with
+    | .running => #[]
+    | .crashed queuedMsgs _ => queuedMsgs
+
   end FileWorker
 end FileWorker
 
@@ -413,6 +418,19 @@ section ServerM
       | return
     updateFileWorkers { fw with state := WorkerState.crashed queuedMsgs origin }
 
+  def tryDischargeQueuedMessages (uri : DocumentUri) (queuedMsgs : Array JsonRpc.Message) : ServerM Unit := do
+      let some fw ← findFileWorker? uri
+        | throwServerError "Cannot find file worker for '{uri}'."
+      let mut crashedMsgs := #[]
+      -- Try to discharge all queued msgs, tracking the ones that we can't discharge
+      for msg in queuedMsgs do
+        try
+          fw.stdin.writeLspMessage msg
+        catch _ =>
+          crashedMsgs := crashedMsgs.push msg
+      if ¬ crashedMsgs.isEmpty then
+        handleCrash uri crashedMsgs .clientToFileWorkerForwarding
+
   /-- Tries to write a message, sets the state of the FileWorker to `crashed` if it does not succeed
       and restarts the file worker if the `crashed` flag was already set. Just logs an error if
       there is no FileWorker at this `uri`.
@@ -436,17 +454,7 @@ section ServerM
       -- restart the crashed FileWorker
       eraseFileWorker uri
       startFileWorker fw.doc
-      let some newFw ← findFileWorker? uri
-        | throwServerError "Cannot find file worker for '{uri}'."
-      let mut crashedMsgs := #[]
-      -- try to discharge all queued msgs, tracking the ones that we can't discharge
-      for msg in queuedMsgs do
-        try
-          newFw.stdin.writeLspMessage msg
-        catch _ =>
-          crashedMsgs := crashedMsgs.push msg
-      if ¬ crashedMsgs.isEmpty then
-        handleCrash uri crashedMsgs .clientToFileWorkerForwarding
+      tryDischargeQueuedMessages uri queuedMsgs
     | WorkerState.running =>
       let initialQueuedMsgs :=
         if queueFailedMessage then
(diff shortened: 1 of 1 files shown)


src/Lean/Server/Watchdog.lean
```diff
@@ -955,7 +959,16 @@ section MainLoop
     let workers ← st.fileWorkersRef.get
     let mut workerTasks := #[]
     for (_, fw) in workers do
-      if let WorkerState.running := fw.state then
+      -- When the forwarding task crashes, its return value will be stuck at
+      -- `WorkerEvent.crashed _`.
+      -- We do not want to handle this event only once, not over and over again,
```
Kha, line 964:
```suggestion
      -- We want to handle this event only once, not over and over again,
```
?

src/Lean/Server/Watchdog.lean
```diff
@@ -98,7 +102,7 @@ section Utils
       that are in-flight are errored. Upon receiving the next packet for that file worker, the file
       worker is restarted and the packet is forwarded to it. If the crash was detected while writing
       a packet, we queue that packet until the next packet for the file worker arrives. -/
-    | crashed (queuedMsgs : Array JsonRpc.Message)
+    | crashed (queuedMsgs : Array JsonRpc.Message) (origin: CrashOrigin)
```
Kha, line 105:
```suggestion
    | crashed (queuedMsgs : Array JsonRpc.Message) (origin : CrashOrigin)
```

Kha (approved):

mhuisi:
(We've decided that this PR will wait until the start of the next release cycle so that we can test it ourselves for a while)

PR #4826, opened 2024-07-25
refactor: deriving DecidableEq to use `termination_by structural`

nomeata:
now that we support structural mutual recursion, I expect that every
`DecidableEq` instance be implemented using structural recursion, so
let's be explicit about it.


nomeata:
@arthur-adjedj, what do you think about this change?

arthur-adjedj:
I think ensuring good reduction on `decide` is desirable, so I'm all for this. some other instances such as `BEq` and `Hashable` might also benefit from having them forcibly be structural, but I'm not so certain it would be useful enough to impose it.

nomeata:
I guess I’ll wait until we have a `nighty-with-mathlib` and see if it survives testing against mathlib :-)

Issue #1672, opened 2022-10-01
`IndPreBelow.mkBelow` failure


leodemoura:
The function `IndPredBelow.mkBelow` fails to generate `brecOn` for the following inductive predicate.
```lean
set_option trace.Meta.IndPredBelow true in
inductive TransClosure (r : α → α → Prop) : α → α → Prop
  | extends : r a b → TransClosure r a b
  | trans_left : r a b → TransClosure r b c → TransClosure r a c
```
The trace contains
```lean
[Meta.IndPredBelow] failed to prove brecOn for TransClosure.below
    couldn't solve by backwards chaining (Lean.Meta.IndPredBelow.maxBackwardChainingDepth = 10): case a
    α : Sort u_1
    r : α → α → Prop
    motive✝ : (a a_1 : α) → TransClosure r a a_1 → Prop
    a✝³ a✝² : α
    : TransClosure r a✝³ a✝²
    : ∀ (a a_1 : α) (x : TransClosure r a a_1), TransClosure.below x → motive✝ a a_1 x
    a b c : α
    : r a b
    a✝ : TransClosure r b c
    a_ih✝ : TransClosure.below a✝
    ⊢ r a a✝³
```
Thus, we cannot use structural recursion in the following definition
```lean
def trans' {a b c} : TransClosure r a b → TransClosure r b c → TransClosure r a c
| .extends h₁       => .trans_left h₁
| .trans_left h₁ h₂ => .trans_left h₁ ∘ trans' h₂
```
This issue has been reported at 
https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Recursive.20definition.20fails/near/301780366

leodemoura:
@DanielFabian Do you have time to take a look? 
Thanks,
Leo

DanielFabian:
At first glance it looks like unprovable non-sense. So it's possible that it applied a bad tactic some steps earlier or something about the goal is wrong. I'll try to understand a bit better what is going on

DanielFabian:
here's roughly what's going on: 

We first do a few introductions, apply the IH, induction and apply constructors. Up to this point, I think it looks still ok. This we create a bunch of goals to solve:

DanielFabian:
```lean
     α : Sort u_1
     r : α → α → Prop
     motive✝ : (a a_1 : α) → Ex.TransClosure r a a_1 → Prop
     a✝² a✝¹ : α
     : Ex.TransClosure r a✝² a✝¹
[...]

PR #4827, opened 2024-07-25
fix: IndPred: track function's motive in a let binding, use withoutProofIrrelevance

### LINKED ISSUE #4751: structural recursion over inductive predicates struggle with reflexive inductives
### Prerequisites

### Description

The following code

```lean
inductive F: Prop where
  | base
  | step (fn: Nat → F)

def F.asdf: F → True
  | base => trivial
  | step f => F.asdf (f 0)
```

results in the following error

```
fail to show termination for
  F.asdf
with errors
structural recursion cannot be used:
```

### Context

This is a variation of a [very similar previous issue](https://github.com/leanprover/lean4/issues/1674). Just like previously, making F a Type rather than a Prop makes the error disappear

### Steps to Reproduce

**Expected behavior:** There are no errors, the recursion is recognized as structural recursion.

**Actual behavior:** The error is present.

### Versions

4.11.0-nightly-2024-07-12

### Impact

nomeata:
this improves support for structural recursion over inductive
*predicates* when there are reflexive arguments.

Consider
```lean
inductive F: Prop where
  | base
  | step (fn: Nat → F)

-- set_option trace.Meta.IndPredBelow.search true
set_option pp.proofs true

def F.asdf1 : (f : F) → True
  | base => trivial
  | step f => F.asdf1 (f 0)
termination_by structural f => f`
```

Previously the search for the right induction hypothesis would fail with
```
could not solve using backwards chaining x✝¹ : F
x✝ : x✝¹.below
f : Nat → F
a✝¹ : ∀ (a : Nat), (f a).below
a✝ : Nat → True
⊢ True
```

The backchaining process will try to use `a✝ : Nat → True`, but then has
no idea what to use for `Nat`.

There are three steps to fix this.

1. We let-bind the function's type before the whole process. Now the
   goal is

   ```
   funType : F → Prop := fun x => True
   x✝ : x✝¹.below
   f : Nat → F
   a✝¹ : ∀ (a : Nat), (f a).below
   a✝ : ∀ (a : Nat), funType (f a)
   ⊢ funType (f 0)
   ```

2. Next we do more aggressive unification when seeing if an assumption
   matches: Given
   ```
   g a b =?= h c d
   ```
   it continues with `a =?= c` and `b =?= d`, even if `g` is a let-bound
   variable.

3. This gives us `f 0 =?= f ?a`. In order to make progress here, we use
   `withoutProofIrrelevance`, because else `isDefEq` is happy to say
   “they are equal” without actually looking at the terms and thus
   assigning `?a := 0`.

This idea of let-binding the function's motive may also be useful for
the other recursion compilers, as it may simplify the FunInd
construction. This is to be investigated.

fixes #4751

### DIFF
--- a/src/Lean/Elab/PreDefinition/Structural/IndPred.lean
+++ b/src/Lean/Elab/PreDefinition/Structural/IndPred.lean
@@ -12,7 +12,7 @@ import Lean.Elab.PreDefinition.Structural.RecArgInfo
 namespace Lean.Elab.Structural
 open Meta
 
-private partial def replaceIndPredRecApps (recArgInfo : RecArgInfo) (motive : Expr) (e : Expr) : M Expr := do
+private partial def replaceIndPredRecApps (recArgInfo : RecArgInfo) (funType : Expr) (motive : Expr) (e : Expr) : M Expr := do
   let maxDepth := IndPredBelow.maxBackwardChainingDepth.get (← getOptions)
   let rec loop (e : Expr) : M Expr := do
     match e with
@@ -36,7 +36,8 @@ private partial def replaceIndPredRecApps (recArgInfo : RecArgInfo) (motive : Ex
         e.withApp fun f args => do
           if f.isConstOf recArgInfo.fnName then
             -- let ty ← inferType e
-            let ty := mkAppN motive args[recArgInfo.numFixed:]
+            let ty := mkAppN funType args[recArgInfo.numFixed:]
+            trace[Elab.definition.structural] "Recursive call {e} with inferred type {←inferType e} and expected type {ty}"
             unless (← isDefEq ty (← inferType e)) do
               throwError "Recursive call {e} does not have expected type {ty}"
             let main ← mkFreshExprSyntheticOpaqueMVar ty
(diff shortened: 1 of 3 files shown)


nomeata:
@DanielFabian this the idea that I mentioned, and seems to be able to handle the examples in that bug report.

I don’t fully understand yet whether I really need both `withoutProofIrrelevance` _and_ my “eager unification path”, and why the latter doesn't subsume the former for `f 0 =?= f ?a`. This part is rather crude.

I also wonder whether a full proof search is really needed for replacing the recursive arguments. When would the assumption to use be more complicated than something like
```
a✝ : ∀ (a : Nat), motive (f a)
```
i.e. the function’s recursive type, applied to indices, the target, and possibly extra arguments, where the target is a plain variable or a function application (if the constructor has a reflexive parameters).

DanielFabian:
Iirc, the motive is indeed always simple because it is completely opaque to us. In particular variables neither show up, nor are swallowed.

The only problem I had in the past was that sometimes there were multiple objects of the right type in the local context and it was important to pick the right one

DanielFabian:
I'm going to mention #1672 again, because since you're changing the backwards-chaining it's possible it'll have an impact over there, too?

nomeata:
> Iirc, the motive is indeed always simple because it is completely opaque to us

That's true for constructing `.below`, but when compiling a concrete function (like `T -> True`), the motive is `fun _ => True`, isn't it?

DanielFabian:
I'm really sorry, at this point I'm fairly hazy about the details. It's been a few years and the construction is not trivial. To jog my memory, it would be really great if you could share some of the proof states that we're seeing and where the tactics fail. Unless you want to walk me through your steps; that's also an option.

PR #4831, opened 2024-07-25
fix: structural recursion: do not check for brecOn too early

nomeata:
Due to nested recursion, we do two passes of `getRecArgInfo`: One on
each argument in isolation, to see which inductive types are around
(e.g. `Tree` and `List`), and
then we later refine/replace this result with the data for the nested
type former (the implicit `ListTree`).

If we have nested recursion through a non-recursive data type like
`Array` or `Prod` then arguemnts of these types should survive the first
phase, so that we can still use them when looking for, say, `Array
Tree`.

This was helpfully reported by @arthur-adjedj.


nomeata:
I think there is a way to not need an eventually unusable `RecArgInfo` in the first pass, with a little code restructuring. I’ll do that if the current code causes more problems in the future.

PR #4833, opened 2024-07-25
fix: make cdot anonymous function notation handle ambiguous notation

### LINKED ISSUE #4832: cdot function notation fails to macro expand properly when there is ambiguous syntax
### Description

If ambiguous notation appears inside a cdot function, then the function can end up with more arguments than expected since `expandCDot?` collects cdots from each alternative in choice nodes.

### Context

Appeared [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/why.20simple.20lambda.20notation.20fails.3F/near/454006959)

### Steps to Reproduce

In the following, the cdot function should have one argument, but it ends up with two:
```lean
def tag (_ : α) (y : α) := y
notation "f" x => tag 1 x
notation "f" x => tag "2" x
#check (String.length <| f ·)
-- fun x x_1 => (f x).length : (x : String) → ?m.4972 x → Nat
```

### Versions

nightly 7/25/2024

### Impact

kmill:
Fixes an issue where each alternative in choice nodes would get their own arguments. Now cdot function expansion is aware of choice nodes.

Also modifies the variable naming so that multi-argument functions like `(· + ·)` expand as `fun x1 x2 => x1 + x2` rather than `fun x x_1 => x + x_1`.

Closes #4832

### DIFF
--- a/src/Lean/Elab/BuiltinNotation.lean
+++ b/src/Lean/Elab/BuiltinNotation.lean
@@ -259,16 +259,17 @@ partial def hasCDot : Syntax → Bool
   Examples:
   - `· + 1` => `fun x => x + 1`
   - `f · · b` => `fun x1 x2 => f x1 x2 b` -/
-partial def expandCDot? (stx : Term) : MacroM (Option Term) := withFreshMacroScope do
+partial def expandCDot? (stx : Term) : MacroM (Option Term) := do
   if hasCDot stx then
-    let mut (newStx, binders) ← (go stx).run #[]
-    if binders.size == 1 then
-      -- It is nicer using `x` over `x1` if there's only a single binder.
-      let x1 := binders[0]!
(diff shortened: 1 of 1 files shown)


src/Lean/Elab/BuiltinNotation.lean
```diff
@@ -261,27 +261,62 @@ partial def hasCDot : Syntax → Bool
...
+  -/
+  countArgs : Syntax → MacroM Nat
+  | `(($_)) => return 0
+  | `(·) => return 1
+  | stx =>
+    match stx with
+    | .node _ k args => do
+      if k == choiceKind then
+        if args.isEmpty then
+          Macro.throwUnsupported
```
kmill, line 282:
```suggestion
          return 0
```

Issue #4090, opened 2024-05-07
RFC: let `unfold` tactic zeta reduce local definitions


kmill:
### Proposal

Right now there's a feature mismatch between `simp` and `unfold`, where if `f` is a global definition and `x` is a local definition you can write `simp only [f, x]` to unfold `f` and zeta reduce `x`, but you cannot write `unfold f x` since the tactic expects that `x` be a global definition. Users frequently ask about this on Zulip, being familiar with `unfold` but confused why they can't unfold local definitions. Recently this was brought up explicitly by Jon Eugster [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/unfold.20and.20unfold_let/near/435411452) asking if there was any technical limitation to letting `unfold` unfold local definitions, hence this RFC.

One of the workarounds for this in mathlib, to have a direct analogue of `unfold`, is the [`unfold_let`](https://github.com/leanprover-community/mathlib4/blob/9ef305b3458090295906a0bb870b46b624cbd9b2/Mathlib/Tactic/DefEqTransformations.lean#L117-L158) tactic. (Note: this tactic predates the feature allowing `simp` and `dsimp` to take local definitions.)

The proposal is to alter `unfold` to be able to handle unfolding local definitions. The implementation cost is low, and instead of resolving only global names, it would need to resolve a name in general and conditionally unfold a constant or zeta reduce a local definition depending on what kind of name it was. This would technically be a breaking change since if a local name shadowed a global one it would now be preferred.

The primary beneficiary for this change would be those learning Lean. Users tend to see local definitions and global definitions as being similar, and it's a common guess that if a tactic works with one it should work for the other. Making `unfold` work in this case will reduce frustration in the game of guess-the-tactic.

*Side note:* It's unclear whether users know that `unfold f1 f2 f3` is the same as `unfold f1; unfold f2; unfold f3`, and they might expect it to interleave unfolding of the definitions. This RFC is not making a recommendation here beyond suggesting that when the tactic docstring this detail can be mentioned.

### Community Feedback

[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/unfold.20and.20unfold_let/near/435411452)

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others benefit from the changes in this proposal being added, please ask them to add :+1: to it.

[...]

PR #4834, opened 2024-07-25
feat: let `unfold` do zeta-delta reduction of local definitions

kmill:
This is "upstreaming" mathlib's `unfold_let` tactic by incorporating its functionality into `unfold`. Now `unfold` can, in addition to unfolding global definitions, unfold local definitions. The PR also updates the `conv` version of the tactic.

An improvement over `unfold_let` is that it beta reduces unfolded local functions.

Two features not present in `unfold` are that (1) `unfold_let` with no arguments does zeta delta reduction of *all* local definitions, and also (2) `unfold_let` can interleave unfoldings (in contrast, `unfold a b c` is exactly the same as `unfold a; unfold b; unfold c`).

Closes RFC #4090


kmill:
This is still a draft. It needs
* updated tactic docstrings
* similar changes to the conv version of `unfold`
* (perhaps) an additional purely zeta delta tactic that's closer to `unfold_let`

kmill:
@semorrison Would you mind taking a look at this change to the `unfold` tactic? It'd be nice to get some eyes on it before merging.

Something I didn't get to is create a `zeta_delta` tactic (which would be a true upstreaming of `unfold_let`), but it shouldn't be very much work to define it from here.

PR #4840, opened 2024-07-26
fix: improve proof search in IndPredBelow

### LINKED ISSUE #1672: `IndPreBelow.mkBelow` failure

nomeata:
`MVarId.apply` is clever in the order it returns mvars, by putting those
arising from non-dependent arguments first. If we lean on that behavior,
this fixes #1672.

H'T to @DanielFabian, this was found together in a call.

Also improves tracing.

### DIFF
--- a/src/Lean/Meta/IndPredBelow.lean
+++ b/src/Lean/Meta/IndPredBelow.lean
@@ -305,7 +305,10 @@ where
         args.back.withApp fun ctor _ => do
         let ctorName := ctor.constName!.updatePrefix below.constName!
         let ctor := mkConst ctorName below.constLevels!
-        m.apply ctor
+        let ms ← m.apply ctor
+        trace[Meta.IndPredBelow] "new goals {ms}"
+        pure ms.reverse
+
     return mss.foldr List.append []
 
   introNPRec (m : MVarId) : MetaM MVarId := do
@@ -321,12 +324,14 @@ where
 def mkBrecOnDecl (ctx : Context) (idx : Nat) : MetaM Declaration := do
   let type ← mkType
   let indVal := ctx.typeInfos[idx]!
-  let name := indVal.name ++ .mkSimple brecOnSuffix
+  let name := mkBRecOnName indVal.name
+  let value ← proveBrecOn ctx indVal type
+  trace[Meta.IndPredBelow] "brecOn value: {value}"
   return Declaration.thmDecl {
     name := name
     levelParams := indVal.levelParams
     type := type
-    value := ←proveBrecOn ctx indVal type }
+    value := value}
 where
   mkType : MetaM Expr :=
     forallTelescopeReducing ctx.headers[idx]! fun xs _ => do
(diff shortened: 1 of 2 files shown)


nomeata:
Ok, we now get test failures because the `.brecOn` construction succeeds now (previously it was failing silently), and then the function definitions compilation fails in a way that #4827 is likely to fix. So I’ll wait here for that (or its variant #4839).

src/Lean/Meta/IndPredBelow.lean
```diff
@@ -303,10 +305,10 @@ where
         args.back.withApp fun ctor _ => do
         let ctorName := ctor.constName!.updatePrefix below.constName!
         let ctor := mkConst ctorName below.constLevels!
-        let ctorInfo ← getConstInfoCtor ctorName
-        let (mvars, _, _) ← forallMetaTelescope ctorInfo.type
-        let ctor := mkAppN ctor mvars
-        m.apply ctor
+        let ms ← m.apply ctor
+        trace[Meta.IndPredBelow] "new goals {ms}"
+        pure ms.reverse
```
nomeata, line 310: This change broke some other tests, until I added the `reverse` here; now the existing tests succeed. :shrug:

Note to my future self:

I wonder if this just means that for any order of mvars we’ll have examples where it fails, or if there is really a reason why this order is good™. I’ll have to come back to this and think about it some more.

Maybe I’ll also investigate if we avoid proof search here and build the necessary terms direct, relying on the structure of the `.ibelow` construction.

nomeata:
Obsoleted by #9995

PR #4841, opened 2024-07-26
refactor: IndPredBelow: use `apply_assumption`

nomeata:
when transforming the `match` statements in `IndPredBelow`, given a
local variable `x : T`, we need to search for `hx : T.below x`.
Previously this was done using the custom `backwardsChaining` method,
although my hypothesis is that we don’t need to chain anything here, and
can use `apply_assumption`.


nomeata:
@DanielFabian. Here too: It seems that backwards chaining is more powerful than we need here. WDYT of this change, to make it more clear that we don’t expect recursive search here? Or would you suggest to not change this code without a clear need at this point?

DanielFabian:
I'm all for simplifying and even deleting backwards chaining if we can get rid of it. I built it initially because in those days the automation features were not quite there yet and it was an easy enough tactic to build. But I'm in no way attached to it. We have our motivating examples in the unit tests, imo.

PR #4846, opened 2024-07-26
fix: wildcard `generalize` only generalizes visible theorems

### LINKED ISSUE #4845: Generalize seems to try to generalize inside the hypothesis being defined
### Prerequisites

### Description

`generalize ... at *` sometimes will try to modify the recursive hypothesis corresponding to the current theorem being defined, which may not be the expected behaviour.  It should only try to `generalize` hypotheses that it can actually modify and are visible, not implementation details. Otherwise this means that there are discrepancies between `generalize ... at *` and `generalize ... at H`, even though `H` is the only hypothesis in the context.  See the following examples:

```lean
/-! Fails because it tries to generalize the internal example hypothesis.  However, it should succeed and behave like the second example shown here. -/
example (H: 0 = 0): True := by
  generalize _H : 0 = z at *
  constructor

/-! Produces the expected behaviour. -/
example (H: 0 = 0): True := by
  generalize _H : 0 = z at H
  constructor

/-! This works as well, because it won't generalize under the binder for the internal hypothesis. -/
example (y : Nat) (H: y = 0): True := by
  generalize _H : y = z at *
  constructor
```

I have also made a tentative PR to try and fix to this.

### Context

I tried to prove a few lemmas that required inverting hypotheses, however, these needed some of the arguments to be abstracted using generalize first so that the inversion with `cases` succeeds.  By default I just tried `generalize ... at *`, however, this failed.  Instead I used `generalize ... at H` which had the expected behaviour.

### Steps to Reproduce

[Link to example in playground](https://live.lean-lang.org/#project=lean-nightly&code=%2F-!%20Fails%20because%20it%20tries%20to%20generalize%20the%20internal%20example%20hypothesis.%20%20However%2C%20it%20should%20succeed%20and%20behave%20like%20the%20second%20example%20shown%20here.%20-%2F%0Aexample%20(H%3A%200%20%3D%200)%3A%20True%20%3A%3D%20by%0A%20%20generalize%20_H%20%3A%200%20%3D%20z%20at%20*%0A%20%20constructor%0A%0A%2F-!%20Produces%20the%20expected%20behaviour.%20-%2F%0Aexample%20(H%3A%200%20%3D%200)%3A%20True%20%3A%3D%20by%0A%20%20generalize%20_H%20%3A%200%20%3D%20z%20at%20H%0A%20%20constructor%0A%0A%2F-!%20This%20works%20as%20well%2C%20because%20it%20won't%20generalize%20under%20the%20binder%20for%20the%20internal%20hypothesis.%20-%2F%0Aexample%20(y%20%3A%20Nat)%20(H%3A%20y%20%3D%200)%3A%20True%20%3A%3D%20by%0A%20%20generalize%20_H%20%3A%20y%20%3D%20z%20at%20*%0A%20%20constructor)

1. Open the link above and observe that the first `generalize` fails whereas the two others succeed.

**Expected behavior:** `generalize ... at *` and `generalize ... at H1 H2 ... HN` should behave the same if `H1 H2 ... HN` are the only visible hypothesis in the context.

**Actual behavior:** `generalize ... at *` can fail sometimes where `generalize ... at H1 H2 ... HN` would succeed, because the wildcard version will try to generalize the internal hypothesis generated by the theorem being proven.

### Versions

`4.11.0-nightly-2024-07-26`

### Impact

ymherklotz:
`generalize ... at *` sometimes will try to modify the recursive hypothesis corresponding to the current theorem being defined, which may not be the expected behaviour.  It should only try to `generalize` hypotheses that it can actually modify and are visible, not implementation details. Otherwise this means that there are discrepancies between `generalize ... at *` and `generalize ... at H`, even though `H` is the only hypothesis in the context.

This commit uses `getLocalHyps` instead of `getFVarIds` to get the current valid `FVarIds` in the context.  This uses `isImplementationDetail` to filter out `FVarIds` that are implementation details in the context and are not visible to the user and should not be manipulated by `generalize`.

Closes #4845

### DIFF
--- a/src/Lean/Elab/Tactic/Generalize.lean
+++ b/src/Lean/Elab/Tactic/Generalize.lean
@@ -30,7 +30,7 @@ open Meta
       args := args.push { hName?, expr, xName? := arg[3].getId : GeneralizeArg }
     let hyps ← match expandOptLocation stx[2] with
     | .targets hyps _ => getFVarIds hyps
-    | .wildcard => pure (← getLCtx).getFVarIds
+    | .wildcard => pure ((← getLocalHyps).map (·.fvarId!))
     let mvarId ← getMainGoal
     mvarId.withContext do
       let (_, newVars, mvarId) ← mvarId.generalizeHyp args hyps
--- a/tests/lean/4845.lean
+++ b/tests/lean/4845.lean
@@ -1,0 +1,12 @@
+/-!
+  Generalize should not try to abstract the variable from hypotheses that are
+  implementation details. -/
+
+/-!
+  In this case, generalize tries to revert the lemma being defined to generalize
+  the 0 in it. -/
+
+example : 0 = 0 → True := by
+  intro H; generalize _H : 0 = z at *
+  trace_state
+  constructor
--- a/tests/lean/4845.lean.expected.out
+++ b/tests/lean/4845.lean.expected.out
@@ -1,0 +1,4 @@
+z : Nat
+_H : 0 = z
+H : z = z
+⊢ True


ymherklotz:
If there is a preferred way to fix this, or if this is actually the expected behaviour then let me know, happy to close the PR then.

awaiting-review

kim-em:
@ymherklotz, apologies for my unreasonable delay here. This looks good. I've just rebase on `nightly-with-mathlib` so we can get some downstream testing before merging. If that goes well, I will merge.

PR #4864, opened 2024-07-30
chore: copy release notes from `releases/v4.10.0`

kmill:
This also updates the 4.9.0 release notes with backported changes.


kmill:
@semorrison I'm copying the current version of the 4.10.0 release notes now, and we can update them again if there happen to be further changes.

An oversight with the 4.9.0 process on my part is that I didn't realize there were further backported changes, and I didn't update the release notes. I figure we can just add them here (and in 4.10.0) and not worry about 4.9.0 proper?

Feel free to merge at your leisure.

PR #4873, opened 2024-07-30
feat: getLsb_replicate

bollu:
This allows bitblasting `BitVec.replicate`.

I changed the definition of `BitVec.replicate` to use `BitVec.cast` in order to make the proof smoother, since it's an easier time simplifying away terms with `BitVec.cast`.


tobiasgrosser:
Hey, this looks good to me. Let's see what @alexkeizer says, but you may be able to updraft this PR after @alexkeizer gives feedback.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1549,4 +1549,45 @@ theorem zeroExtend_truncate_succ_eq_zeroExtend_truncate_or_twoPow_of_getLsb_true
...
+  rw [Nat.mul_succ] at hhi
+  rw [Nat.sub_eq_of_eq_add]
+  suffices i / w = n by rw [← this, Nat.add_comm, Nat.div_add_mod]
+  apply Nat.div_eq_of_lt_le
+    (by rw [Nat.mul_comm]; omega)
+    (by rw [Nat.mul_comm, Nat.mul_succ]; omega)
+
+theorem getLsb_replicate {n w : Nat} (x : BitVec w) :
+    (x.replicate n).getLsb i =
+    ((decide (i < w * n)) && (x.getLsb (i % w))) := by
```
semorrison, line 1578: Excessive parentheses.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1549,4 +1549,45 @@ theorem zeroExtend_truncate_succ_eq_zeroExtend_truncate_or_twoPow_of_getLsb_true
...
+private theorem Nat.sub_mul_eq_mod_of_lt_of_le (hlo : w * n ≤ i) (hhi : i < w * (n + 1)) :
+    i - w * n = i % w := by
+  rw [Nat.mul_succ] at hhi
+  rw [Nat.sub_eq_of_eq_add]
+  suffices i / w = n by rw [← this, Nat.add_comm, Nat.div_add_mod]
+  apply Nat.div_eq_of_lt_le
+    (by rw [Nat.mul_comm]; omega)
+    (by rw [Nat.mul_comm, Nat.mul_succ]; omega)
+
+theorem getLsb_replicate {n w : Nat} (x : BitVec w) :
```
semorrison, line 1576: `@[simp]`?

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1549,4 +1549,45 @@ theorem zeroExtend_truncate_succ_eq_zeroExtend_truncate_or_twoPow_of_getLsb_true
...
+@[simp]
+theorem replicate_succ_eq {x : BitVec w} :
+    x.replicate (n + 1) =
+    (x ++ replicate n x).cast (by rw [Nat.mul_succ]; omega) := by
+  simp [replicate]
+
+/--
+If a number `w * n ≤ i < w * (n + 1)`, then `i - w * n` equals `i % w`.
+Intuitively, this is true by subtracting `w * n` from the inequality, giving
+`0 ≤ i - w * n < w`, which uniquely identifies `i % w`.
```
semorrison, line 1565: Best not to say "intuitively", when you're giving a perfectly rigorous explanation.

In fact, I would just drop these two lines.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1549,4 +1549,45 @@ theorem zeroExtend_truncate_succ_eq_zeroExtend_truncate_or_twoPow_of_getLsb_true
...
+-/
+private theorem Nat.sub_mul_eq_mod_of_lt_of_le (hlo : w * n ≤ i) (hhi : i < w * (n + 1)) :
+    i - w * n = i % w := by
+  rw [Nat.mul_succ] at hhi
+  rw [Nat.sub_eq_of_eq_add]
+  suffices i / w = n by rw [← this, Nat.add_comm, Nat.div_add_mod]
+  apply Nat.div_eq_of_lt_le
+    (by rw [Nat.mul_comm]; omega)
+    (by rw [Nat.mul_comm, Nat.mul_succ]; omega)
```
semorrison, line 1574: This theorem should be factored into two statements:
```
example (m k : Nat) : m % k = m - k * (m / k) := by sorry
-- Surely in the library? The Int version is called `Int.emod_def`. `Nat.mod_add_div` is close.
```
```
example (hlo : w * n ≤ i) (hhi : i < w * (n + 1)) : n = i / w := ...
```
Both of these should then live in `Nat/Div.lean`.

bollu: Resolved.

src/Init/Data/BitVec/Lemmas.lean
bollu: I kept the explanation, since I prefer having some text that tells me why something's true on-paper, and dropped "Intuitively".

src/Init/Data/BitVec/Lemmas.lean
bollu: Resolved.

src/Init/Data/BitVec/Lemmas.lean
bollu: `example (hlo : w * n ≤ i) (hhi : i < w * (n + 1)) : n = i / w := ...` already exists as `div_eq_of_lt_le`, so I use it directly.

bollu:
awaiting-review

src/Init/Data/Nat/Div.lean
```diff
@@ -203,6 +203,10 @@ theorem mod_add_div (m k : Nat) : m % k + k * (m / k) = m := by
   | base x y h => simp [h]
   | ind x y h IH => simp [h]; rw [Nat.mul_succ, ← Nat.add_assoc, IH, Nat.sub_add_cancel h.2]
 
+theorem mod_def (m k : Nat) : m % k = m - k * (m / k) := by
```
semorrison, line 206: Technically this is not the definition for `Nat`, but I like the parallelism with the `Int` naming, and will worry about whether this needs to be changed later.

PR #4875, opened 2024-07-30
fix: mistake in statement of `List.take_takeWhile`

kmill:
This theorem is meant to say that `List.take` and `List.takeWhile` commute.


semorrison:
Ooops! 😊

PR #4884, opened 2024-07-31
fix: export symbols needed by Verso

david-christiansen:
Verso needed a symbol that was unexported - this exposes it again.


semorrison:
No opinion on the solution, but unless someone objects I will make sure this is in in time for `v4.11.0-rc1`.

david-christiansen:
Thanks! We should do a stage0 update before rc1 as well, because otherwise Verso will only work on stage2 builds from `master` (the exported symbols are determined by the prior stage's compiler).

david-christiansen:
Stage0 update triggered.

PR #4890, opened 2024-08-01
perf: precise cache for `expr_eq_fn`

leodemoura:
This performance issue was exposed by the benchmarks at https://github.com/leanprover/LNSym/tree/proof_size_expt/Proofs/SHA512/Experiments


leodemoura:
!bench

PR #4892, opened 2024-08-01
feat: warnings for commands that do not have desired effects in particular contexts

### LINKED ISSUE #4743: `set_option in` makes local attribute ineffective
### Prerequisites

### Description

The construction

```lean
set_option ... in
attribute [local ...] ...
```

silently fails to add the attribute. I believe this is because the `set_option in` is desugared into a `section` and the attribute becomes local to that section, which is then immediately closed again.

### Steps to Reproduce

MWE:

```lean
axiom falso : False

section Works

attribute [local simp] falso

example : False := by simp

end Works

section Fails

set_option trace.debug true in
attribute [local simp] falso

example : False := by simp -- simp made no progress

end Fails
```

### Versions

`"4.11.0-nightly-2024-07-12"`

### Impact

### LINKED ISSUE #4744: Can't add local attribute for namespaced theorem
### Prerequisites

### Description

The construction

```lean
@[local ...]
theorem Foo.bar ...
```

seemingly makes the attribute local to the `Foo` namespace, which is immediately closed again. Hence the attribute is effectively not applied.

### Steps to Reproduce

MWE:

```lean
-- Fails

@[local simp]
axiom Foo.falso : False

example : False := by simp -- simp made no progress

-- Works

@[local simp]
axiom falso : False

example : False := by simp
```

I don't see a situation in which this behaviour would be useful, so the first example should imo either make the attribute local to the current section/namespace or throw an error at `@[local simp]`. There is probably a similar issue with scoped attributes; there the desired behaviour is less clear.

### Versions

"4.11.0-nightly-2024-07-12"

### Impact

kmill:
Previously, the `$cmd1 in $cmd2` combinator is implemented as `section $cmd1 $cmd2 end`, which means that if `$cmd2` has local effects (like a `@[local simp]` attribute) then the command apparently has no effects at all. Now `$cmd2` can detect it is in such a context and log a warning in such cases. Similarly, `$cmd1` can detect it is in a context where global effects are not welcome.

Also, this adds a `ref` field to `Command.Scope` to keep track of what was responsible for the scope. In the future, this syntax ref could be used to include links to `section`/`namespace` commands in `end`'s errors. For now, these are used by `end` to determine whether scopes were added by `section` or `namespace` for further validation. In particular, this is used to raise errors in constructions such as `cmd1 in namespace foo`, `cmd1 in end`, or `end in cmd2`.

Adds tests for the `end` command. Removes the unused `Lean.Elab.Command.withNamespace`.

Closes #4744, closes #4743

### DIFF
--- a/src/Lean/Elab/BuiltinCommand.lean
+++ b/src/Lean/Elab/BuiltinCommand.lean
@@ -23,7 +23,7 @@ namespace Lean.Elab.Command
      modifyEnv fun env => addMainModuleDoc env ⟨doc, range⟩
    | _ => throwErrorAt stx "unexpected module doc string{indentD stx[1]}"
 
-private def addScope (ref : Syntax) (isNewNamespace : Bool) (isNoncomputable : Bool) (header : String) (newNamespace : Name) : CommandElabM Unit := do
+private def addScope (ref : Syntax) (isNewNamespace : Bool) (isWeak : Bool) (isNoncomputable : Bool) (header : String) (newNamespace : Name) : CommandElabM Unit := do
   modify fun s =>
     let newScope :=
       { s.scopes.head! with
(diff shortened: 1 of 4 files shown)


kmill:
@Kha Here's a draft for making it so commands can see whether global or local effects are expected. It's a bit different from what we discussed since it was easier this way to take advantage of incrementality for macros.

PR #4895, opened 2024-08-01
chore: deprecate Nix-based build, remove interactive components

Kha:
Users who prefer the flake build should maintain it externally

### DIFF
--- a/.github/workflows/nix-ci.yml
+++ b/.github/workflows/nix-ci.yml
@@ -146,5 +146,3 @@ jobs:
       - name: Fixup CCache Cache
         run: |
           sudo chown -R $USER /nix/var/cache
-      - name: CCache stats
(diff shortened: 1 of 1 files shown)


lenianiva:
Are all Nix components going to be removed eventually? I can take over and create a repository for the Nix build paths

nomeata:
I expect that in the long run we certainly keep a nix shell with the build dependencies alive.

We *maybe* keep using Nix for our own CI uses, but certainly nothing that we’d want other people to import/depend upon.

So if you are motivated to maintain nix-related things in a separate repository, that would be great!

nomeata:
It's up to you if course. But it may be easier to copy all that machinery and start taking ownership in the new repository, and adjust it to your needs.

lenianiva:
I got one working. I'll try to support every major Lean version: https://github.com/lenianiva/lean4-nix

PR #4897, opened 2024-08-01
chore: wrap diagnostic results in `MessageData.trace`s

mattrobball:
Currently, the messages in the diagnostic summaries are created by appending interpolated strings. We wrap these in `.trace`'s, and the results are better formatted when expanding child nodes in the info view. Particularly, the latter diagnostic summaries remain on their own lines flush to the left instead of on the same line directly adjacent to the last child node.


leodemoura:
@mattrobball The new output looks good to me. Ready to merge?

mattrobball:
Yes, thanks.

PR #4900, opened 2024-08-02
fix: ilean loading performance

mhuisi:
This PR roughly halves the time needed to load the .ilean files by optimizing the JSON parser and the conversion from JSON to Lean data structures.

The code is optimized roughly as follows:
- String operations are inlined more aggressively
- Parsers are changed to use new `String.Iterator` functions `curr'` and `next'` that receive a proof and hence do not need to perform an additional check
- The `RefIdent` of .ilean files now uses a `String` instead of a `Name` to avoid the expensive parse step from `String` to `Name` (despite the fact that we only very rarely actually need a `Name` in downstream code)
- Instead of `List`s and `Subarray`s, the JSON to Lean conversion now directly passes around arrays and array indices to avoid redundant boxing
- Parsec's `peek?` sometimes generates redundant `Option` wrappers because the generation of basic blocks interferes with the ctor-match optimization, so it is changed to use an `isEof` check where possible
- Early returns and inline-do-blocks cause the code generator to generate new functions, which then interfere with optimizations, so they are now avoided
- Mutual defs are used instead of unspecialized passing of higher-order functions to generate faster code
- The object parser is made tail-recursive

This PR also fixes a stack overflow in `Lean.Json.compress` that would occur with long lists and adds a benchmark for the .ilean roundtrip (compressed pretty-printing -> parsing).

### DIFF
--- a/src/Lean/Data/Json/Parser.lean
+++ b/src/Lean/Data/Json/Parser.lean
@@ -61,10 +61,10 @@ partial def strCore (acc : String) : Parser String := do
 @[inline] def str : Parser String := strCore ""
 
 partial def natCore (acc : Nat) : Parser Nat := do
-  let c ← peekD '\x00'
-  if c == '\x00' then
+  if ← isEof then
     return acc
   else
+    let c ← peek!
     if '0' <= c && c <= '9' then
       skip
       let acc' := 10*acc + (c.val - '0'.val).toNat
@@ -73,10 +73,10 @@ partial def natCore (acc : Nat) : Parser Nat := do
       return acc
 
 partial def natCoreNumDigits (acc digits : Nat) : Parser (Nat × Nat) := do
-  let c ← peekD '\x00'
-  if c == '\x00' then
+  if ← isEof then
     return (acc, digits)
   else
+    let c ← peek!
     if '0' <= c && c <= '9' then
       skip
       let acc' := 10*acc + (c.val - '0'.val).toNat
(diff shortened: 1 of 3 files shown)


mhuisi:
!bench

mhuisi:
!bench

src/Init/Meta.lean
```diff
@@ -75,7 +75,7 @@ See #2572.
 opaque Internal.hasLLVMBackend (u : Unit) : Bool
 
 /-- Valid identifier names -/
-def isGreek (c : Char) : Bool :=
+@[inline] def isGreek (c : Char) : Bool :=
```
hargoniX, line 78: What's the rule for whether to inline or not inline these functions? Why is there e.g. no `inline` on `isSubScriptAlnum`?

src/Lean/Data/Json/Parser.lean
```diff
@@ -44,31 +43,46 @@ def escapedChar : Parser Char := do
...
+@[inline] def str : Parser String := strCore ""
 
-partial def natCore (acc digits : Nat) : Parser (Nat × Nat) := do
-  let some c ← peek? | return (acc, digits)
-  if '0' ≤ c ∧ c ≤ '9' then
-    skip
-    let acc' := 10*acc + (c.val.toNat - '0'.val.toNat)
-    natCore acc' (digits+1)
+partial def natCore (acc : Nat) : Parser Nat := do
+  let c ← peekD '\x00'
```
hargoniX, line 64: We might be able to get rid of this null byte hack with checking `eof?` and then peeking? I'd hope that the compiler is smart enough to make this not make a difference.

hargoniX (approved):

src/Lean/Data/Json/Parser.lean
mhuisi: Good idea. I've changed `eof?` to `isEof` to reflect our usual naming conventions and used that instead.

mhuisi:
!bench

Issue #4861, opened 2024-07-29
Timeout with large arithmetic expression involving numerals


TwoFX:
### Prerequisites

Please put an X between the brackets as you perform the following steps:

* [x] Check that your issue is not already filed:
      https://github.com/leanprover/lean4/issues
* [x] Reduce the issue to a minimal, self-contained, reproducible test case.
      Avoid dependencies to Mathlib or Batteries.
* [x] Test your test case against the latest nightly release, for example on
      https://live.lean-lang.org/#project=lean-nightly
      (You can also use the settings there to switch to “Lean nightly”)

### Description

The following code timeouts with the default `maxHeartbeats` setting. It sometimes fails at `whnf` and sometimes at `isDefEq`, depending on the specific heartbeat setting used.

```lean
theorem foo (x y z p q : Int) : False :=
  have : (1 * x ^ 1 + z ^ 1 * p) *
        (1 / 1 * p ^ 1 * x ^ 1 + 1 * q * p ^ 1 * x * z + 1 * q ^ 1 * p ^ 1 * x ^ 1 +
                            1 * q ^ 1 * p ^ 1 * x * z -
                          1 * q * p ^ 1 * y ^ 1 +
                        1 * q ^ 1 * p ^ 1 * x ^ 1 +
                      1 * q ^ 1 * p ^ 1 * x * z -
                    1 * q ^ 1 * p ^ 1 * y ^ 1 +
                  1 * q ^ 1 * p ^ 1 * x ^ 1 +
                1 * q ^ 1 * p ^ 1 * x * z -
              1 * q ^ 1 * p ^ 1 * y ^ 1 +
            1 * q ^ 1 * x ^ 1 -
          1 * q ^ 1 * p * y ^ 1) +
      z * (1 * y) *
        (-(1 / 1 * p ^ 1 * x * y) + 1 * q * p ^ 1 * z * y - 1 * q ^ 1 * p ^ 1 * x * y +
                  1 * q ^ 1 * p ^ 1 * z * y -
                1 * q ^ 1 * p ^ 1 * x * y +
              1 * q ^ 1 * p ^ 1 * z * y -
            1 * q ^ 1 * p ^ 1 * x * y +
          1 / 1 * q ^ 1 * p ^ 1 * z * y) +
    (-y ^ 1 + p * x * (1 * z) + q * (1 * z ^ 1)) *
[...]

PR #4903, opened 2024-08-02
perf: use `NatPow Int` instead of `HPow Int Nat Int`

leodemoura:
This modification improves the performance of the example in issue #4861. It no longer times out but is still expensive.

Here is the analysis of the performance issue: Given `(x : Int)`, to elaborate `x ^ 1`, a few default instances have to be tried.

First, the homogeneous instance is tried and fails since `Int` does not implement `Pow Int`. Then, the `NatPow` instance is tried, and it also fails. The same process is performed for each term of the form `p ^ 1`. There are seveal of them at #4861. After all of these fail, the lower priority default instance for numerals is tried, and `x ^ 1` becomes `x ^ (1 : Nat)`. Then, `HPow Int Nat Int` can be applied, and the elaboration succeeds. However, this process has to be repeated for every single term of the form `p ^ 1`. The elaborator tries all homogeneous `HPow` and `NatPow` instances for all `p ^ 1` terms before trying the lower priority default instance `OfNat`.

This commit ensures `Int` has a `NatPow` instance instead of `HPow Int Nat Int`. This change shortcuts the process, but it still first tries the homogeneous `HPow` instance, fails, and then tries `NatPow`. The elaboration can be made much more efficient by writing `p ^ (1 : Nat)`.


kmill:
This seems to have been an oversight in #2778, and it should have become a `NatPow Int` instance then.

Mathlib adds a default `Pow` instance for monoids, which is used instead of this `HPow Int Nat Int` instance. I suppose that hid the performance issue?

Issue #4891, opened 2024-08-01
IO.FS.readFile corrupts the contents of files containing null bytes


eric-wieser:
### Prerequisites

Please put an X between the brackets as you perform the following steps:

* [x] Check that your issue is not already filed:
      https://github.com/leanprover/lean4/issues
* [x] Reduce the issue to a minimal, self-contained, reproducible test case.
      Avoid dependencies to Mathlib or Batteries.
* [ ] Test your test case against the latest nightly release, for example on
      https://live.lean-lang.org/#project=lean-nightly
      (You can also use the settings there to switch to “Lean nightly”)

### Description

`IO.readFile` produces a string that is not contained in the original file.

### Context

This came up while trying to read `/proc/{pid}/cmdline`, which contains null bytes but is valid UTF8.

### Steps to Reproduce

1. Run the following, adjusting `path` to be a temporary file of your choice

```lean
#eval do
  let path := "/tmp/lean4-4891"
  IO.FS.writeBinFile path
    ("Hello\x00(this is a very secret message that must never be shared) world".toUTF8)
  let x ← IO.FS.readFile path
  IO.println x
```

**Expected behavior:** The secret message is visible

**Actual behavior:** The output is `Hello world`. The first 64-byte chunk ends at the the `)` character, and is discarded due to the null byte.

### Versions

4.10.0

### Additional Information

[Additional information, configuration or data that might be necessary to reproduce the issue]

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

eric-wieser:
Ah, mostly a duplicate of #3546

PR #4906, opened 2024-08-03
feat: Revamp file reading and writing

hargoniX:
This PR:
- changes the implementation of `readBinFile` and `readFile` to only require two system calls (`stat` + `read`) instead of one `read` per 1024 byte chunk.
- fixes a bug where `Handle.getLine` would get tripped up by a NUL character in the line and cut the string off. This is caused by the fact that the original implementation uses `strlen` and `lean_mk_string` which is the backer of `mk_string` does so as  well.
- fixes a bug where `Handle.putStr` and thus by extension `writeFile` would get tripped up by a NUL char in the line and cut the string off. Cause here is the use of `fputs` when a NUL char is possible.

Closes: #4891 
Closes: #3546
Closes: #3741


tests/lean/3546.lean
```diff
@@ -0,0 +1,12 @@
...
+  let firstLine := "foo\u0000bar\n"
+  let content := firstLine ++ "hello world\nbye"
+  IO.FS.writeFile tmpFile content
+  let handle ← IO.FS.Handle.mk tmpFile .read
+  let firstReadLine ← handle.getLine
+  let cond := firstLine == firstReadLine && firstReadLine.length == 8 -- paranoid
+  IO.println cond
+  IO.FS.removeFile tmpFile
+
+#eval test
```
eric-wieser, line 12: This could use `#guard_msgs`, right?

src/Init/System/IO.lean
```diff
@@ -473,22 +473,11 @@ partial def Handle.readBinToEnd (h : Handle) : IO ByteArray := do
...
-  h.readBinToEnd
-
-def readFile (fname : FilePath) : IO String := do
-  let h ← Handle.mk fname Mode.read
-  h.readToEnd
+def Handle.readToEnd (h : Handle) : IO String := do
+  let data ← h.readBinToEnd
+  match String.fromUTF8? data with
+  | some s => return s
+  | none => throw <| .userError s!"Tried to read from handle containing non UTF-8 data."
```
eric-wieser, line 480: A feature request for another time; `fromUTF8?` should return an `Except` that indicates where the non-utf8 character was, even if just as a string error message.

src/Init/System/IO.lean
```diff
@@ -594,6 +583,26 @@ end System.FilePath
...
 
+namespace FS
+
+def readBinFile (fname : FilePath) : IO ByteArray := do
+  -- Requires metadata so defined after metadata
+  let mdata ← fname.metadata
+  let size := mdata.byteSize.toUSize
+  if size > 0 then
+    let handle ← IO.FS.Handle.mk fname .read
+    handle.read mdata.byteSize.toUSize
```
Kha, line 594: Is this based on any prior implementation? What if the file is `/dev/stdin`?

Kha:
!bench

src/Init/System/IO.lean
hargoniX: This is based on Rust's [`read_to_string`](https://doc.rust-lang.org/stable/src/std/fs.rs.html#302-312):
```rs
pub fn read_to_string<P: AsRef<Path>>(path: P) -> io::Result<String> {
    fn inner(path: &Path) -> io::Result<String> {
        let mut file = File::open(path)?;
        let size = file.metadata().map(|m| m.len() as usize).ok();
        let mut string = String::new();
        string.try_reserve_exact(size.unwrap_or(0))?;
        io::default_read_to_string(&mut file, &mut string, size)?;
        Ok(string)
    }
    inner(path.as_ref())
}
```

hargoniX: As for `/dev/stdin` it does have a size:
```
λ ls -la /dev/stdin
lrwxrwxrwx. 1 root root 15 Jul 29 21:02 /dev/stdin -> /proc/self/fd/0
```
So this implementation would at least not do something completely crazy. That said I think the advice is if you are reading from something like /dev/stdin that can evolve over time the advice is to basically not use `readBinFile`/`readFile`.

eric-wieser: I don't think that's quite the same, [`default_read_to_string`](https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/src/std/io/mod.rs.html#432-447) looks like it might only use the size as a hint (and passes it as an option)

hargoniX:
!bench

src/Init/System/IO.lean
```diff
@@ -594,6 +586,28 @@ end System.FilePath
...
+  let buf ←
+    if size > 0 then
+      handle.read mdata.byteSize.toUSize
+    else
+      pure <| ByteArray.mkEmpty 0
+  handle.readBinToEndInto buf
+
+def readFile (fname : FilePath) : IO String := do
+  let data ← readBinFile fname
+  match String.fromUTF8? data with
```
TwoFX, line 605: Does this make a copy of the data?

TwoFX (commented):
LGTM

src/Init/System/IO.lean
hargoniX: Yes, unfortunately the runtime representation of `ByteArray` has the data part inlined instead of as a pointer and the layout of the `ByteArray` and `String` object are not compatible :/.

hargoniX:
Of course windows doesn't have getline and this breaks everything /o\

PR #4907, opened 2024-08-04
fix: change the definition of `PtrSet` and `PtrMap` to `abbrev`

znssong:
I changed the transparency of `PtrSet` and `PtrMap` to `abbrev`, because some usages would fail if they are `def`, for example the instance synthesize of `EmptyCollection (PtrMap Nat Nat)`.


Kha:
This type should be considered a primitive/newtype. Missing operations should be added as separate declarations.

PR #4917, opened 2024-08-05
chore: switch to Std.HashMap and Std.HashSet almost everywhere

TwoFX:
This removes every use except for `ShareCommon` which I will do as a follow-up.

This is very likely to break some users of the `Lean` API downstream.


TwoFX:
!bench

TwoFX:
!bench

PR #4933, opened 2024-08-06
feat: setup `Std.Sat` with definitions of SAT and CNF

hargoniX:
Step 1 out of approximately 7 to upstream LeanSAT.

### DIFF
--- a/src/Std/Sat/CNF/Basic.lean
+++ b/src/Std/Sat/CNF/Basic.lean
@@ -5,6 +5,7 @@ Authors: Kim Morrison
 -/
 prelude
 import Init.Data.List.Lemmas
+import Init.Data.List.Impl
 import Std.Sat.CNF.Literal
 
 namespace Std


src/Std/Sat/CNF/Basic.lean
```diff
@@ -0,0 +1,188 @@
...
+import Std.Sat.CNF.Literal
+
+namespace Std
+namespace Sat
+
+/--
+A clause in a CNF.
+
+The literal `(i, b)` is satisfied is the assignment to `i` agrees with `b`.
+-/
```
tobiasgrosser, line 17:
```suggestion
The literal `(i, b)` is satisfied if the assignment to `i` agrees with `b`.
-/
```

src/Std/Sat/CNF/Literal.lean
```diff
@@ -0,0 +1,47 @@
...
+
+instance [Hashable α] : Hashable (Literal α) where
+  hash := fun x => if x.2 then hash x.1 else hash x.1 + 1
+
+instance : HSat α (Literal α) where
+  eval := fun p l => (p l.1) = l.2
+
+instance (p : α → Bool) (l : Literal α) : Decidable (p ⊨ l) := by
+  rw [HSat.eval, instHSat]
+  exact Bool.decEq (p l.fst) l.snd
```
TwoFX, line 28: Constructing `Decidable` with tactics is a bad idea (remember #4674), so this should be
```suggestion
instance (p : α → Bool) (l : Literal α) : Decidable (p ⊨ l) :=
  inferInstanceAs (Decidable (p l.1 = l.2))
```
or use `decidable_of_decidable_of_iff`.

src/Std/Sat/CNF/Basic.lean
```diff
@@ -0,0 +1,188 @@
+/-
+Copyright (c) 2024 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Scott Morrison
```
TwoFX, line 4: Maybe double-check which name to use here.

src/Std/Sat/CNF/Basic.lean
```diff
@@ -0,0 +1,188 @@
...
+Literals are identified by members of `α`.
+-/
+abbrev CNF (α : Type u) : Type u := List (CNF.Clause α)
+
+namespace CNF
+
+/--
+Evaluating a `Clause` with respect to an assignment `f`.
+-/
+def Clause.eval (f : α → Bool) (c : Clause α) : Bool := c.any fun (i, n) => f i == n
```
TwoFX, line 32: It's a bit unfortunate that the thing that is called `c` or `g` in this file is called `f` in Sat/Basic.lean and the thing called `f` here is called `p` there.

src/Std/Sat/CNF/Basic.lean
```diff
@@ -0,0 +1,188 @@
...
+
+@[simp] theorem Clause.eval_nil (f : α → Bool) : Clause.eval f [] = false := rfl
+@[simp] theorem Clause.eval_succ (f : α → Bool) :
```
TwoFX, line 35:
```suggestion
@[simp] theorem Clause.eval_cons (f : α → Bool) :
```

src/Std/Sat/CNF/Basic.lean
```diff
@@ -0,0 +1,188 @@
...
+    Clause.eval f (i :: c) = (f i.1 == i.2 || Clause.eval f c) := rfl
+
+/--
+Evaluating a `CNF` formula with respect to an assignment `f`.
+-/
+def eval (f : α → Bool) (g : CNF α) : Bool := g.all fun c => c.eval f
+
+@[simp] theorem eval_nil (f : α → Bool) : eval f [] = true := rfl
+@[simp] theorem eval_succ (f : α → Bool) : eval f (c :: g) = (c.eval f && eval f g) := rfl
```
TwoFX, line 44:
```suggestion
@[simp] theorem eval_cons (f : α → Bool) : eval f (c :: g) = (c.eval f && eval f g) := rfl
```

src/Std/Sat/CNF/Basic.lean
```diff
@@ -0,0 +1,188 @@
...
+@[simp] theorem eval_append (f : α → Bool) (g h : CNF α) :
+    eval f (g ++ h) = (eval f g && eval f h) := List.all_append
+
+instance : HSat α (Clause α) where
+  eval assign clause := Clause.eval assign clause
+
+instance : HSat α (CNF α) where
+  eval assign cnf := eval assign cnf
+
+@[simp] theorem unsat_nil_iff_false : unsatisfiable α ([] : CNF α) ↔ False :=
```
TwoFX, line 55: I think that the normal way to write this would just be
```suggestion
@[simp] theorem not_unsatisfiable_nil : ¬unsatisfiable α ([] : CNF α) :=
```
and `simp` should still be able to apply it.

src/Std/Sat/CNF/Basic.lean
```diff
@@ -0,0 +1,188 @@
...
+@[simp] theorem sat_nil {assign : α → Bool} : assign ⊨ ([] : CNF α) ↔ True := by
+  simp [(· ⊨ ·)]
+
+@[simp] theorem unsat_nil_cons {g : CNF α} : unsatisfiable α ([] :: g) ↔ True := by
+  simp [unsatisfiable, (· ⊨ ·)]
+
+namespace Clause
+
+/--
+Literal `a` occurs in `Clause` `c`.
```
TwoFX, line 67: I think it this should be
```suggestion
Variable `a` occurs in `Clause` `c`.
```
(and similar below)

src/Std/Sat/CNF/Basic.lean
```diff
@@ -0,0 +1,188 @@
...
+
+/--
+Literal `a` occurs in `CNF` formula `g`.
+-/
+def mem (a : α) (g : CNF α) : Prop := ∃ c, c ∈ g ∧ c.mem a
+
+instance {a : α} {g : CNF α} [DecidableEq α] : Decidable (mem a g) :=
+  inferInstanceAs <| Decidable (∃ _, _)
+
+theorem any_nonEmpty_iff_exists_mem {g : CNF α} :
```
TwoFX, line 112:
```suggestion
theorem any_not_isEmpty_iff_exists_mem {g : CNF α} :
```

src/Std/Sat/CNF/Basic.lean
```diff
@@ -0,0 +1,188 @@
...
+  . intro h
+    rcases h with ⟨lit, clause, ⟨hclause1, hclause2⟩⟩
+    exists clause
+    constructor
+    . assumption
+    . cases hclause2 with
+      | inl hl => exact Exists.intro _ hl
+      | inr hr => exact Exists.intro _ hr
+
+@[simp] theorem not_mem_cons : (¬ ∃ a, mem a g) ↔ ∃ n, g = List.replicate n [] := by
```
TwoFX, line 133:
```suggestion
@[simp] theorem not_exists_mem : (¬ ∃ a, mem a g) ↔ ∃ n, g = List.replicate n [] := by
```

src/Std/Sat/CNF/RelabelFin.lean
```diff
@@ -0,0 +1,139 @@
...
+  simp only [mem] at hlit
+  rcases h with ⟨_, hbar⟩
+  cases hlit
+  all_goals
+    have := hbar (lit, _) (by assumption)
+    omega
+
+theorem Clause.maxLiteral_eq_some_of_mem (c : Clause Nat) (h : mem l c)
+    : ∃ maxLit, c.maxLiteral = some maxLit := by
+  dsimp[mem] at h
```
TwoFX, line 34:
```suggestion
  dsimp [mem] at h
```

src/Std/Sat/CNF/RelabelFin.lean
```diff
@@ -0,0 +1,139 @@
...
+  cases h <;> rename_i h
+  all_goals
+    have h1 := List.ne_nil_of_mem h
+    have h2 := not_congr <| @List.maximum?_eq_none_iff _ (c.map (·.1)) _
+    simp[← Option.ne_none_iff_exists', h1, h2, maxLiteral]
```
TwoFX, line 39:
```suggestion
    simp [← Option.ne_none_iff_exists', h1, h2, maxLiteral]
```

TwoFX (commented):
I've been thinking a bit more about where to put this stuff. Do we envision any users of this material outside of LeanSAT? If not, then putting this in `Lean/Elab/Tactic` instead of `Std` might be a better idea.

src/Std/Sat/Basic.lean
```diff
@@ -0,0 +1,141 @@
...
+import Init.NotationExtra
+
+namespace Std
+namespace Sat
+
+/--
+For variables of type `α` and formulas of type `β`, `HSat.eval a f` is meant to determine whether
+a formula `f` is true under assignment `a`.
+-/
+class HSat (α : Type u) (β : Type v) :=
```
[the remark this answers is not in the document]
hargoniX: I did consider this as well but that's not possible. Later in LeanSAT there is an instance:
```
instance {n : Nat} : HSat (PosFin n) (Array Assignment) where
  eval := fun p arr => ∀ i : PosFin n, ¬(hasAssignment (¬p i) arr[i.1]!)
```
in the LRAT checker. This instance will be rejected because `n` cannot be inferred from the non out param `Array Assignment`.

src/Std/Sat/CNF/Basic.lean
```diff
@@ -0,0 +1,188 @@
...
+@[simp] theorem Clause.eval_nil (a : α → Bool) : Clause.eval a [] = false := rfl
+@[simp] theorem Clause.eval_cons (a : α → Bool) :
+    Clause.eval a (i :: c) = (a i.1 == i.2 || Clause.eval a c) := rfl
+
+/--
+Evaluating a `CNF` formula with respect to an assignment `a`.
+-/
+def eval (a : α → Bool) (f : CNF α) : Bool := f.all fun c => c.eval a
+
+@[simp] theorem eval_nil (a : α → Bool) : eval a [] = true := rfl
```
TwoFX, line 43: Currently you develop half of the theory about clauses and CNFs in terms of `⊨`/`Unsatisfiable` and the other half in terms of the two `eval` methods. This is generally considered to be subobtimal design, because it makes the API difficult to discover and requires knowing when to magically unfold using `simp [(· ⊨ ·)]` to jump from one layer to the next (which then magically has completely different simp normal forms!). It would be much nicer if there was only one theorem `eval_def` for each of the `eval`s and all of the theory (like `eval_nil,` `eval_cons`, and also downstream things like `eval_relabel` are stated in terms of `⊨` instead of `eval`.

src/Std/Sat/CNF/Basic.lean
```diff
@@ -0,0 +1,189 @@
...
+@[simp] theorem eval_cons (a : α → Bool) : eval a (c :: f) = (c.eval a && eval a f) := rfl
+
+@[simp] theorem eval_append (a : α → Bool) (f1 f2 : CNF α) :
+    eval a (f1 ++ f2) = (eval a f1 && eval a f2) := List.all_append
+
+def sat (a : α → Bool) (f : CNF α) : Prop := eval a f = true
```
TwoFX, line 49:
```suggestion
def Sat (a : α → Bool) (f : CNF α) : Prop := eval a f = true
```

src/Std/Sat/CNF/Basic.lean
```diff
@@ -0,0 +1,189 @@
...
+def unsat (f : CNF α) : Prop := ∀ a, eval a f = false
+
+theorem sat_def (a : α → Bool) (f : CNF α) : sat a f ↔ (eval a f = true) := by rfl
+theorem unsat_def (f : CNF α) : unsat f ↔ (∀ a, eval a f = false) := by rfl
+
+
+@[simp] theorem not_unsatisfiable_nil : ¬unsat ([] : CNF α) :=
```
TwoFX, line 56:
```suggestion
@[simp] theorem not_unsat_nil : ¬unsat ([] : CNF α) :=
```

src/Std/Sat/CNF/RelabelFin.lean
```diff
@@ -0,0 +1,139 @@
...
+    f.relabel fun i =>
+      if w : i < n then
+        -- This branch will always hold
+        ⟨i, w⟩
+      else
+        ⟨0, numLiterals_pos h.choose_spec⟩
+  else
+    List.replicate f.length []
+
+theorem unsat_relabelFin {f : CNF Nat} :
```
TwoFX, line 116: This looks like a good `@[simp]` lemma?

TwoFX (approved):

PR #4946, opened 2024-08-07
doc: Added make doc alternative to `nproc` on MacOS

ydewit:
Just a small addition to include `sysctl -n hw.physicalcpu` as the alternative to `nproc` on MacOS.


Kha:
Duplicate of #4647

PR #4950, opened 2024-08-07
feat: revamp file IO, this time Windows compatible

hargoniX:
This implements a naive version of `getline` because Windows does not have `getline`. Given the fact that `FILE` has buffered IO, calling `fgetc` in a loop is not as big of a performance hazard as it might seem at first glance.

The proper solution to this would of course be to have our own buffered IO so we are fully in charge of the buffer. In this situation we could check the entire buffer for a newline at once instead of char by char. However that is not going to happen for the near future so I propose we stay with this implementation. If reading individual lines of a file does truly end up being the performance bottle neck we have already won^^.


hargoniX:
The failing check is a race condition introduced by a different bug on master. The key point of this PR: Make bootstrapping and in general Lean compile and run on all architectures again. is achieved so I'll merge.

PR #4953, opened 2024-08-08
feat: `Std.Sat.AIG`

hargoniX:
Step 2/~7 in upstreaming LeanSAT.

### DIFF
--- a/src/Std/Sat/AIG/CachedGates.lean
+++ b/src/Std/Sat/AIG/CachedGates.lean
@@ -26,30 +26,24 @@ def mkNotCached (aig : AIG α) (gate : Ref aig) : Entrypoint α :=
   let res := aig.mkConstCached true
   let aig := res.aig
   let constRef := res.ref
-  aig.mkGateCached
-    { lhs := 
-      { ref := constRef
-        inv := false }
-      rhs := 
-      { ref := gate.cast <| by
-          intros
-          apply LawfulOperator.le_size_of_le_aig_size (f := mkConstCached)
-          omega
-        inv := true } }
+  aig.mkGateCached {
+    lhs := {
+      ref := constRef
+      inv := false
+    }
+    rhs := {
+      ref := gate.cast <| by
+        intros
+        apply LawfulOperator.le_size_of_le_aig_size (f := mkConstCached)
+        omega
+      inv := true
+    }
+  }
 
 @[inline]
 def BinaryInput.asGateInput {aig : AIG α} (input : BinaryInput aig) (linv rinv : Bool) :
     GateInput aig :=
-  {
-    lhs := {
-      ref := input.lhs
-      inv := linv
-    },
-    rhs := {
-      ref := input.rhs
-      inv := rinv
-    }
-  }
+  { lhs := { ref := input.lhs, inv := linv }, rhs := { ref := input.rhs, inv := rinv } }
 
 /--
 Create an and gate in the input AIG. This uses the builtin cache to enable automated subterm
@@ -69,8 +63,7 @@ def mkOrCached (aig : AIG α) (input : BinaryInput aig) : Entrypoint α :=
   let res := aig.mkConstCached true
   let aig := res.aig
   let constRef := res.ref
-  aig.mkGateCached
-    {
+  aig.mkGateCached {
       lhs := {
         ref := constRef
         inv := false
@@ -103,20 +96,19 @@ def mkXorCached (aig : AIG α) (input : BinaryInput aig) : Entrypoint α :=
   let res := aig.mkGateCached rinput
   let aig := res.aig
   let aux2Ref := res.ref
-  aig.mkGateCached
-    {
-      lhs := {
-        ref := aux1Ref.cast <| by
-          simp (config := { zetaDelta := true }) only
-          apply LawfulOperator.le_size_of_le_aig_size (f := mkGateCached)
-          omega
-        inv := true
-      },
-      rhs := {
-        ref := aux2Ref
-        inv := true
-      }
+  aig.mkGateCached {
+    lhs := {
+      ref := aux1Ref.cast <| by
+        simp (config := { zetaDelta := true }) only
+        apply LawfulOperator.le_size_of_le_aig_size (f := mkGateCached)
+        omega
+      inv := true
+    },
+    rhs := {
+      ref := aux2Ref
+      inv := true
     }
+  }
 
 /--
 Create an equality gate in the input AIG. This uses the builtin cache to enable automated subterm
@@ -136,20 +128,19 @@ def mkBEqCached (aig : AIG α) (input : BinaryInput aig) : Entrypoint α :=
   let res := aig.mkGateCached rinput
   let aig := res.aig
   let aux2Ref := res.ref
-  aig.mkGateCached
-    {
-      lhs := {
-        ref := aux1Ref.cast <| by
-          simp (config := { zetaDelta := true }) only
-          apply LawfulOperator.le_size_of_le_aig_size (f := mkGateCached)
-          omega
-        inv := true
-      },
-      rhs := {
-        ref := aux2Ref
-        inv := true
-      }
+  aig.mkGateCached {
+    lhs := {
+      ref := aux1Ref.cast <| by
+        simp (config := { zetaDelta := true }) only
+        apply LawfulOperator.le_size_of_le_aig_size (f := mkGateCached)
+        omega
+      inv := true
+    },
+    rhs := {
+      ref := aux2Ref
+      inv := true
     }
+  }
 
 /--
 Create an implication gate in the input AIG. This uses the builtin cache to enable automated subterm
@@ -163,21 +154,20 @@ def mkImpCached (aig : AIG α) (input : BinaryInput aig) : Entrypoint α :=
   let res := aig.mkConstCached true
   let aig := res.aig
   let constRef := res.ref
-  aig.mkGateCached
-    {
-      lhs := {
-        ref := constRef
-        inv := false
-      },
-      rhs := {
-        ref := auxRef.cast <| by
-          intros
-          simp (config := { zetaDelta := true }) only
-          apply LawfulOperator.le_size_of_le_aig_size (f := mkConstCached)
-          omega
-        inv := true
-      }
+  aig.mkGateCached {
+    lhs := {
+      ref := constRef
+      inv := false
+    },
+    rhs := {
+      ref := auxRef.cast <| by
+        intros
+        simp (config := { zetaDelta := true }) only
+        apply LawfulOperator.le_size_of_le_aig_size (f := mkConstCached)
+        omega
+      inv := true
     }
+  }
 
 end AIG
 
--- a/src/Std/Sat/AIG/RelabelNat.lean
+++ b/src/Std/Sat/AIG/RelabelNat.lean
@@ -187,11 +187,11 @@ def addAtom {decls : Array (Decl α)} {hidx} (state : State α decls idx) (a : �
   match hmap : state.map[a]? with
   | some _ =>
     { state with
-        inv2 := by
-          apply Inv2.oldAtom
-          . exact state.inv2
-          . assumption
-          . assumption
+      inv2 := by
+        apply Inv2.oldAtom
+        . exact state.inv2
+        . assumption
+        . assumption
     }
   | none =>
     {
@@ -215,10 +215,10 @@ def addConst {decls : Array (Decl α)} {hidx} (state : State α decls idx) (b : 
     (h : decls[idx]'hidx = .const b) :
     State α decls (idx + 1) :=
   { state with
-      inv2 := by
-        apply Inv2.const
-        . exact state.inv2
-        . assumption
+    inv2 := by
+      apply Inv2.const
+      . exact state.inv2
+      . assumption
   }
 
 /--
@@ -228,10 +228,10 @@ def addGate {decls : Array (Decl α)} {hidx} (state : State α decls idx) (lhs r
     (linv rinv : Bool) (h : decls[idx]'hidx = .gate lhs rhs linv rinv) :
     State α decls (idx + 1) :=
   { state with
-      inv2 := by
-        apply Inv2.gate
-        . exact state.inv2
-        . assumption
+    inv2 := by
+      apply Inv2.gate
+      . exact state.inv2
+      . assumption
   }
 
 /--
@@ -353,9 +353,10 @@ namespace Entrypoint
 
 def relabelNat' (entry : Entrypoint α) : (Entrypoint Nat × HashMap α Nat) :=
   let res := entry.aig.relabelNat'
-  let entry := { entry with
-      aig := res.fst,
-      ref.hgate := by simp [entry.ref.hgate, res]
+  let entry :=
+  { entry with
+    aig := res.fst,
+    ref.hgate := by simp [entry.ref.hgate, res]
   }
   (entry, res.snd)
 
@@ -366,8 +367,8 @@ identifiers.
 -/
 def relabelNat (entry : Entrypoint α) : Entrypoint Nat :=
   { entry with
-      aig := entry.aig.relabelNat
-      ref.hgate := by simp [entry.ref.hgate]
+    aig := entry.aig.relabelNat
+    ref.hgate := by simp [entry.ref.hgate]
   }
 
 /--
--- a/src/Std/Sat/AIG/RefVec.lean
+++ b/src/Std/Sat/AIG/RefVec.lean
@@ -27,13 +27,13 @@ def cast' {aig1 aig2 : AIG α} (s : RefVec aig1 len)
         → ∀ {i : Nat} (h : i < len), s.refs[i]'(by have := s.hlen; omega) < aig2.decls.size) :
     RefVec aig2 len :=
   { s with
-      hrefs := by
-        intros
-        apply h
-        · intros
-          apply s.hrefs
-          assumption
-        · assumption
+    hrefs := by
+      intros
+      apply h
+      · intros
+        apply s.hrefs
+        assumption
+      · assumption
   }
 
 @[inline]
--- a/src/Std/Sat/AIG/Relabel.lean
+++ b/src/Std/Sat/AIG/Relabel.lean
@@ -155,8 +155,8 @@ namespace Entrypoint
 
 def relabel (r : α → β) (entry : Entrypoint α) : Entrypoint β :=
   { entry with
-      aig := entry.aig.relabel r
-      ref.hgate := by simp [entry.ref.hgate]
+    aig := entry.aig.relabel r
+    ref.hgate := by simp [entry.ref.hgate]
   }
 
 @[simp]


src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,475 @@
...
+-/
+import Std.Data.HashMap
+import Std.Data.HashSet
+
+namespace Std
+namespace Sat
+
+/-!
+This module contains the basic definitions for an AIG (And-Inverter Graph) in the style of AIGNET,
+as described in https://arxiv.org/pdf/1304.7861.pdf section 3. It consists of an AIG definition,
```
tobiasgrosser, line 14:
```suggestion
This module contains the basic definitions for an AIG (And Inverter Graph) in the style of AIGNET,
as described in https://arxiv.org/pdf/1304.7861.pdf section 3. It consists of an AIG definition,
```

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,475 @@
...
+    . apply aig.inv <;> assumption
+    . injections
+      have := lhs.ref.hgate
+      have := rhs.ref.hgate
+      omega
+  ⟨{ aig with decls, inv, cache }, ⟨g, by simp [g, decls]⟩⟩
+
+/--
+Add a new input node to the AIG in `aig`. Note that his version is only meant for proving,
+for production purposes use `AIG.mkAtomCached` and equality theorems to this one.
```
tobiasgrosser, line 440:
```suggestion
Add a new input node to the AIG in `aig`. Note that this version is only meant for proving,
for production purposes use `AIG.mkAtomCached` and equality theorems to this one.
```

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,475 @@
...
+
+@[inline]
+def GateInput.cast {aig1 aig2 : AIG α} (input : GateInput aig1)
+    (h : aig1.decls.size ≤ aig2.decls.size) :
+    GateInput aig2 :=
+  { input with lhs := input.lhs.cast h, rhs := input.rhs.cast h }
+
+/--
+Build an AIG gate in `aig`. Note that his version is only meant for proving,
+for production purposes use `AIG.mkGateCached` and equality theorems to this one.
```
tobiasgrosser, line 419:
```suggestion
Build an AIG gate in `aig`. Note that this version is only meant for proving,
for production purposes use `AIG.mkGateCached` and equality theorems to this one.
```

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,475 @@
...
+    simp only [decls] at *
+    simp only [Array.get_push] at h2
+    split at h2
+    . apply aig.inv <;> assumption
+    . contradiction
+  ⟨{ decls, inv, cache }, ⟨g, by simp [g, decls]⟩⟩
+
+/--
+Build a constant node in `aig`. Note that his version is only meant for proving,
+for production purposes use `AIG.mkConstCached` and equality theorems to this one.
```
tobiasgrosser, line 457:
```suggestion
Build a constant node in `aig`. Note that his version is only meant for proving,
for production purposes use `AIG.mkConstCached` and equality theorems to this one.
```
```suggestion
Build a constant node in `aig`. Note that this version is only meant for proving,
for production purposes use `AIG.mkConstCached` and equality theorems to this one.
```

src/Std/Sat/AIG/CNF.lean
```diff
@@ -0,0 +1,718 @@
...
+  -/
+  hmark : ∀ (lhs rhs : Nat) (linv rinv : Bool) (idx : Nat) (hbound : idx < aig.decls.size)
+            (_hmarked : marks[idx] = true) (heq : aig.decls[idx] = .gate lhs rhs linv rinv),
+              marks[lhs]'(by have := aig.inv idx lhs rhs linv rinv hbound heq; omega) = true
+                ∧
+              marks[rhs]'(by have := aig.inv idx lhs rhs linv rinv hbound heq; omega) = true
+  /--
+  Relate satisfiability results about our produced CNF to satisfiability results about the AIG that
+  we are processing. The intuition for this is: If a node is marked, its CNF (and all required
+  children CNFs according to `hmark`) are already part of the current CNF. Thus the current CNF is
```
tobiasgrosser, line 150:
```suggestion
  we are processing. The intuition for this is: if a node is marked, its CNF (and all required
  children CNFs according to `hmark`) are already part of the current CNF. Thus the current CNF is
```

src/Std/Sat/AIG/CachedGatesLemmas.lean
```diff
@@ -0,0 +1,268 @@
...
+-/
+
+namespace Std
+namespace Sat
+
+namespace AIG
+
+/--
+Encoding not as boolean expression in AIG form.
+-/
```
tobiasgrosser, line 22: `Encoding of not as boolean expression` or `Encode not as boolean expression`

The same comment applies below.

src/Std/Sat/AIG/If.lean
```diff
@@ -0,0 +1,280 @@
...
+      ⟧
+        =
+      if ⟦aig, discr, assign⟧ then
+        ⟦aig, lhs.get idx hidx1, assign⟧
+      else
+        ⟦aig, rhs.get idx hidx1, assign⟧ := by
+  intro idx hidx1 hidx2
+  generalize hgo : go aig curr hcurr discr lhs rhs s = res
+  unfold go at hgo
+  dsimp at hgo
```
tobiasgrosser, line 231: Non-terminal simp?

src/Std/Sat/AIG/If.lean
```diff
@@ -0,0 +1,280 @@
...
+  ∀ (idx : Nat) (hidx : idx < w),
+    ⟦(ite aig input).aig, (ite aig input).stream.get idx hidx, assign⟧
+      =
+    if ⟦aig, input.discr, assign⟧ then
+      ⟦aig, input.lhs.get idx hidx, assign⟧
+    else
+      ⟦aig, input.rhs.get idx hidx, assign⟧ := by
+  intro idx hidx
+  unfold ite
+  dsimp
```
tobiasgrosser, line 270: Non-terminal simp, or does this rule not apply for dsimp?

src/Std/Sat/AIG/If.lean
```diff
@@ -0,0 +1,280 @@
...
+  rw [ite.denote_go]
+  omega
+
+
```
tobiasgrosser, line 274:
```suggestion

```

src/Std/Sat/AIG/RefStream.lean
```diff
@@ -0,0 +1,188 @@
...
+  ⟩
+
+@[simp]
+theorem get_push_ref_eq (s : RefStream aig len) (ref : AIG.Ref aig) :
+    (s.push ref).get len (by omega) = ref := by
+  have := s.hlen
+  simp [get, push, ← this]
+
+-- This variant exists because it is sometimes hard to rewrite properly with DTT
+theorem get_push_ref_eq' (s : RefStream aig len) (ref : AIG.Ref aig) (idx : Nat)
```
tobiasgrosser, line 75:
```suggestion
-- This variant exists because it is sometimes hard to rewrite properly with DTT.
theorem get_push_ref_eq' (s : RefStream aig len) (ref : AIG.Ref aig) (idx : Nat)
```

src/Std/Sat/AIG/RefStreamOperator/Map.lean
```diff
@@ -0,0 +1,239 @@
...
+  termination_by len - idx
+
+theorem map.go_le_size {aig : AIG α} (idx : Nat) (hidx) (s : RefStream aig idx)
+    (input : RefStream aig len) (f : (aig : AIG α) → Ref aig → Entrypoint α)
+    [LawfulOperator α Ref f] [LawfulMapOperator α f] :
+    aig.decls.size ≤ (go aig idx hidx s input f).aig.decls.size := by
+  unfold go
+  split
+  . next h =>
+    dsimp
```
tobiasgrosser, line 86: Non-terminal simp?

src/Std/Sat/AIG/RefStreamOperator/Map.lean
```diff
@@ -0,0 +1,239 @@
...
+  le_size := by intros; apply map_le_size
+  decl_eq := by intros; apply map_decl_eq
+
+namespace map
+
+theorem go_get_aux {aig : AIG α} (curr : Nat) (hcurr : curr ≤ len) (s : RefStream aig curr)
+    (input : RefStream aig len) (f : (aig : AIG α) → Ref aig → Entrypoint α)
+    [LawfulOperator α Ref f] [LawfulMapOperator α f] :
+    -- The hfoo here is a trick to make the dependent type gods happy
+    ∀ (idx : Nat) (hidx : idx < curr) (hfoo),
```
tobiasgrosser, line 136:
```suggestion
    -- The hfoo here is a trick to make the dependent type gods happy.
    ∀ (idx : Nat) (hidx : idx < curr) (hfoo),
```

src/Std/Sat/AIG/RefStreamOperator/Map.lean
```diff
@@ -0,0 +1,239 @@
...
+      (go aig curr hcurr s input f).stream.get idx (by omega)
+        =
+      (s.get idx hidx).cast hfoo := by
+  intro idx hidx
+  generalize hgo : go aig curr hcurr s input f = res
+  unfold go at hgo
+  split at hgo
+  . dsimp at hgo
```
tobiasgrosser, line 144: Non-terminal dsimp? There are more, but I won't list them again.

src/Std/Sat/AIG/RelabelNat.lean
```diff
@@ -0,0 +1,385 @@
...
+  . omega
+  · apply ofAIG.Inv2
+
+end State
+end RelabelNat
+
+def relabelNat' (aig : AIG α) : (AIG Nat × HashMap α Nat) :=
+  let map := RelabelNat.State.ofAIG aig
+  let aig := aig.relabel fun x =>
+    -- The none branch never gets hit, we prove this below
```
tobiasgrosser, line 306:
```suggestion
    -- The none branch never gets hit, we prove this below.
```

src/Std/Sat/AIG/RelabelNat.lean
```diff
@@ -0,0 +1,385 @@
...
+      aig := entry.aig.relabelNat
+      ref.hgate := by simp[entry.ref.hgate]
+  }
+
+/--
+`relabelNat` preserves unsatisfiablility.
+-/
+theorem relabelNat_unsat_iff {entry : Entrypoint α} [Nonempty α] :
+    (entry.relabelNat).Unsat ↔ entry.Unsat:= by
+  simp [relabelNat, Unsat]
```
tobiasgrosser, line 378: Non terminal simp.

tobiasgrosser (commented):
Very nicely prepared patch and a big step forward.  🥳  I left some stylistic comments.

src/Std/Sat/AIG/If.lean
hargoniX: `dsimp` only applies definitional reductions if not provided with arguments as far as I know?. Its job here is to reduce the tons of lets that are in the term to a single let free term that I can rewrite on.

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,475 @@
...
+  -/
+  decls : Array (AIG.Decl α)
+  /--
+  The `Decl` cache, valid with respect to `decls`.
+  -/
+  cache : AIG.Cache α decls
+  /--
+  In order to be a valid AIG, `decls` must form a DAG.
+  -/
+  inv : AIG.IsDag α decls
```
bollu, line 220: Quick question: why is this called `inv`? An explanation would be nice. Is it `inv` for invariant? I would have expected something like `hdag`.

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,475 @@
...
+  refs : Array Nat
+  hlen : refs.size = length
+  hrefs : ∀ (h : i < length), refs[i] < aig.decls.size
+
+/--
+A sequence of references bundled with their AIG. In some sense a generalization of `Entrypoint`.
+-/
+structure RefStreamEntry (α : Type) [DecidableEq α] [Hashable α] [DecidableEq α] (length : Nat) where
+  aig : AIG α
+  stream : RefStream aig length
```
bollu, line 332: Can `length` be renamed to `w` for consistency, or will this break too much downstream? :)

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,475 @@
...
+      have := h2 x _ _ _ _ h1 h3
+      let lval := go lhs decls assign (by omega) h2
+      let rval := go rhs decls assign (by omega) h2
+      xor lval linv && xor rval rinv
+
+scoped syntax "⟦" term ", " term "⟧" : term
+scoped syntax "⟦" term ", " term ", " term "⟧" : term
+
+macro_rules
+| `(⟦$entry, $assign⟧) => `(denote $assign $entry )
```
bollu, line 369:
```suggestion
| `(⟦$entry, $assign⟧) => `(denote $assign $entry)
```

stray whitespace?

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,475 @@
...
+| `(⟦$aig, $ref, $assign⟧) => `(denote $assign (Entrypoint.mk $aig $ref))
+
+@[app_unexpander AIG.denote]
+def unexpandDenote : Lean.PrettyPrinter.Unexpander
+  | `($(_) {aig := $aig, start := $start, inv := $hbound} $assign) =>
+    `(⟦$aig, ⟨$start, $hbound⟩, $assign⟧)
+  | `($(_) $entry $assign) => `(⟦$entry, $assign⟧)
+  | _ => throw ()
+
+def UnsatAt (aig : AIG α) (start : Nat) (h : start < aig.decls.size) : Prop :=
```
bollu, line 379:
```suggestion
/--
The denotation of the sub-DAG in the `aig` at node `start` is false for all assignments.
-/
def UnsatAt (aig : AIG α) (start : Nat) (h : start < aig.decls.size) : Prop :=
```

src/Std/Sat/AIG/CNF.lean
```diff
@@ -0,0 +1,718 @@
...
+          split at hmarked
+          . simp_all
+          . have := cache.inv.hmark lhs rhs linv rinv idx hbound hmarked heq
+            simp [Array.getElem_set, this]
+        . intro assign heval idx hbound hmarked
+          rw [Array.getElem_set] at hmarked
+          split at hmarked
+          . next heq =>
+            simp at heq
+            simp [heq] at htip heval
```
bollu, line 287: These look like non terminal simps.

src/Std/Sat/AIG/CNF.lean
```diff
@@ -0,0 +1,718 @@
...
+            simp [heval, denote_idx_const htip]
+          . next heq =>
+            simp at heval
```
bollu, line 290: non-terminal simp?

bollu: I think the pattern of 

```
simp at heq
simp [heq] at htip heval
```

occurs in every proof that shows that `inv` is preserved. and they all appear to be non-terminal to me.

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,475 @@
...
+structure ShiftTarget (aig : AIG α) (w : Nat) where
+  stream : AIG.RefStream aig w
+  distance : Nat
+
+structure ArbitraryShiftTarget (aig : AIG α) (m : Nat) where
+  n : Nat
+  target : AIG.RefStream aig m
+  distance : AIG.RefStream aig n
+
+structure ExtendTarget (aig : AIG α) (newWidth : Nat) where
```
TwoFX, line 342: You're defining a lot of structures in this file, some of which aren't used in this PR. Are they used later in the development? For some of them I'm also wondering whether bundling this data together is really worth it.

src/Std/Sat/AIG/If.lean
TwoFX: I think `dsimp only at hgo` is the way to go here.

bollu (commented):
Thanks for the super clean implementation! I've read upon `CNF.lean`, and will get to the others in a bit.

src/Std/Sat/AIG/Basic.lean
hargoniX: Yes it's the invariant for the AIG

src/Std/Sat/AIG/If.lean
```diff
@@ -0,0 +1,278 @@
...
+where
+  go {w : Nat} (aig : AIG α) (curr : Nat) (hcurr : curr ≤ w) (discr : Ref aig)
+      (lhs rhs : RefStream aig w) (s : RefStream aig curr) : RefStreamEntry α w :=
+    if hcurr:curr < w then
+      let input := ⟨discr, lhs.get curr hcurr, rhs.get curr hcurr⟩
+      let res := mkIfCached aig input
+      let aig := res.aig
+      let ref := res.ref
+      have := by
+        apply LawfulOperator.le_size (f := mkIfCached)
```
TwoFX, line 105: I have to say I don't love this pattern of shadowing a variable, reaching a goal that cannot be written down and then doing some arcane invocation that just happens to have the right type. I would prefer something like
```suggestion
      have : aig.decls.size ≤ res.aig.decls.size :=
        LawfulOperator.le_size (f := mkIfCached) ..
```

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,475 @@
...
+  { input with lhs := input.lhs.cast h, rhs := input.rhs.cast h }
+
+/--
+Build an AIG gate in `aig`. Note that this version is only meant for proving,
+for production purposes use `AIG.mkGateCached` and equality theorems to this one.
+-/
+def mkGate (aig : AIG α) (input : GateInput aig) : Entrypoint α :=
+  let lhs := input.lhs
+  let rhs := input.rhs
+  let g := aig.decls.size
```
TwoFX, line 426: This applies to many functions, but why do you have so many `let`s for things that are then only used once downstream? I don't think it really improves readability much.

src/Std/Sat/AIG/LawfulStreamOperator.lean
```diff
@@ -0,0 +1,91 @@
...
+import Std.Sat.AIG.RefStream
+
+namespace Std
+namespace Sat
+
+namespace AIG
+
+variable {α : Type} [Hashable α] [DecidableEq α]
+
+-- TODO: Find a way to merge this with LawfulOperator that preserves nice automation
```
TwoFX, line 16: Another TODO that should be resolved in some way.

src/Std/Sat/AIG/Lemmas.lean
```diff
@@ -0,0 +1,292 @@
...
+    ⟦aig, ⟨idx, hidx⟩, assign1⟧ = ⟦aig, ⟨idx, hidx⟩, assign2⟧ := by
+  apply denote_idx_trichotomy
+  . intro b heq
+    simp [denote_idx_const heq]
+  . intro a heq
+    simp only [denote_idx_atom heq]
+    apply h
+    rw [mem_def]
+    rw [← heq]
+    rw [Array.mem_def]
```
TwoFX, line 281: These three can be combined into one.

src/Std/Sat/AIG/Basic.lean
hargoniX: Yes, these are all used in the bit blasting part. They are mostly bundled because I need my functions to be a lawful operator which only takes one argument.

src/Std/Sat/AIG/Cached.lean
```diff
@@ -0,0 +1,134 @@
...
+    have := input.rhs.ref.hgate
+    let decl := .gate lhs rhs linv rinv
+    match cache.get? decl with
+    | some hit =>
+      ⟨⟨decls, cache, inv⟩, ⟨hit.idx, hit.hbound⟩⟩
+    | none =>
+      /-
+      Here we implement the constant propagating subset of:
+      https://fmv.jku.at/papers/BrummayerBiere-MEMICS06.pdf
+      TODO: rest of the table
```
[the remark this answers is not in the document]
hargoniX: Yes.

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,475 @@
...
+      (idx : Fin decls.size) : String :=
+    match decls[idx] with
+    | Decl.const b => s!"{idx} [label=\"{b}\", shape=box];"
+    | Decl.atom i => s!"{idx} [label=\"{i}\", shape=doublecircle];"
+    | Decl.gate _ _ _ _ => s!"{idx} [label=\"{idx} ∧\",shape=trapezium];"
+
+/--
+A stream of references into `aig`. This is the `AIG` analog of `BitVec`.
+-/
+structure RefStream (aig : AIG α) (length : Nat) where
```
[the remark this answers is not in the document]
hargoniX: No particular reason, do you have an alternative naming suggestion?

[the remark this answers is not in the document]
hargoniX: If I had `Array Ref` I would have to iterate over the thing every time that the underlying AIG changes and apply `Ref.cast`. This would unnecessarily be `O(n)` and we could of course `implementedBy` this away but that would increase the trust required by LeanSAT. With it done like this the underlying AIG can change and we can update the proof for free all the time.

src/Std/Sat/AIG/If.lean
hargoniX: The shadowing of the variable is a stylistic choice such that I don't violate the linearity on `AIG` ever, simply because I can only ever reference one `AIG` at a time anyways. This is used basically all throughout the bitblaster as well to ensure linearity by convention. Do you have an alternative idea for a style that makes these linearity mistakes hard to do?

src/Std/Sat/AIG/Basic.lean
hargoniX: In this particular function there is no precise reason. In the functions that work on AIGs see my remarks about shadowing etc.

src/Std/Sat/AIG/Basic.lean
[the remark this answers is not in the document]
TwoFX: `Vector` or `Vec` maybe?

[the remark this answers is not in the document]
TwoFX: I see, thanks for the explanation.

src/Std/Sat/AIG/Basic.lean
TwoFX: Makes sense, thanks for the explanation.

src/Std/Sat/AIG/LawfulStreamOperator.lean
TwoFX: Yes, I think you can just remote this TODO.

src/Std/Sat/AIG/If.lean
TwoFX: Okay, you have convinced me.
```suggestion
      have := LawfulOperator.le_size (f := mkIfCached) ..
```
should also work I think.

src/Std/Sat/AIG/Basic.lean
TwoFX: Ah, I see, thanks. No need to change anything for now.

src/Std/Sat/AIG/If.lean
semorrison: Could you write something, perhaps in the module-doc, explaining this?

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+  -/
+  | const (b : Bool)
+  /--
+  An input node to the circuit.
+  -/
+  | atom (idx : α)
+  /--
+  An AIG gate with configurable input nodes and polarity. `l` and `r` are the
+  input node indices while `linv` and `rinv` say whether there is an inverter on
+  the left or right input.
```
semorrison, line 37:
```suggestion
  the left and right inputs, respectively.
```

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+as described in https://arxiv.org/pdf/1304.7861.pdf section 3. It consists of an AIG definition,
+a description of its semantics and basic operations to construct nodes in the AIG.
+-/
+
+variable {α : Type} [Hashable α] [DecidableEq α]
+
+namespace AIG
+
+/--
+A circuit node declaration. These are not recursive but instead contain indices into an `AIG`, with inputs indexed by `α`.
```
semorrison, line 23: Why are you calling these "node declarations" rather than just "nodes"?

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+  the left or right input.
+  -/
+  | gate (l r : Nat) (linv rinv : Bool)
+  deriving Hashable, Repr, DecidableEq, Inhabited
+
+
+/--
+A `Cache` is valid with respect to a list of declarations iff:
+Whenever `cache.find? decl` returns an index into `xs : Array Decl`, `xs[index] = decl`.
+This predicate limits a `HashMap Decl Nat` to this behavior.
```
semorrison, line 46:
```suggestion
A `Cache` is valid with respect to an array of declarations if
whenever `cache.find? decl` returns an index into `xs : Array Decl`, `xs[index] = decl`.
This predicate limits a `HashMap Decl Nat` to this behavior.
```

(When giving definitions of terms in prose, "iff" is implied.)

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+  input node indices while `linv` and `rinv` say whether there is an inverter on
+  the left or right input.
+  -/
+  | gate (l r : Nat) (linv rinv : Bool)
+  deriving Hashable, Repr, DecidableEq, Inhabited
+
+
+/--
+A `Cache` is valid with respect to a list of declarations iff:
+Whenever `cache.find? decl` returns an index into `xs : Array Decl`, `xs[index] = decl`.
```
semorrison, line 45:
```suggestion
A `Cache` is valid with respect to an array of declarations if
whenever `cache.find? decl` returns an index into `xs : Array Decl`, `xs[index] = decl`.
```

(When giving definitions of terms in prose, "iff" is implied.)

src/Std/Sat/AIG/Basic.lean
semorrison, line 46:
```suggestion
`Cache.WF xs` is a predicate asserting that a `cache : HashMap (Decl α) Nat`
is a valid lookup cache for `xs : Array (Decl α)`, that is,
whenever `cache.find? decl` returns an index into `xs : Array Decl`, `xs[index] = decl`.
```

(My concern here is that your original docstring starts by talking about `Cache`, which hasn't otherwise been mentioned yet.)

semorrison: Maybe explicitly say that it doesn't need to be complete? Something like
"It need not be complete, in the sense that `cache.find? decl = none` is allowed even when `decl \mem xs`."

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+  -/
+  | empty : WF decls {}
+  /--
+  Given a `cache`, valid with respect to some `decls`, we can extend the `decls` without
+  extending the cache and remain valid.
+  -/
+  | push_id (h : WF decls cache) : WF (decls.push decl) cache
+  /--
+  Given a `cache`, valid with respect to some `decls`, we can extend the `decls`
+  and the `cache` at the same time with the same values and remaind valid.
```
semorrison, line 60:
```suggestion
  and the `cache` at the same time with the same values and remain valid.
```

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+-/
+@[irreducible]
+def Cache.empty (decls : Array (Decl α)) : Cache α decls := ⟨{}, WF.empty⟩
+
+@[inherit_doc Cache.WF.push_id, irreducible]
+def Cache.noUpdate (cache : Cache α decls) : Cache α (decls.push decl) :=
+  ⟨cache.val, Cache.WF.push_id cache.property⟩
+
+/-
+We require the `decls` as an explicit attribute because we use `decls.size` so accidentally mutating
```
semorrison, line 81:
```suggestion
We require the `decls` as an explicit argument because we use `decls.size` so accidentally mutating
```

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+    idx < decls.size := by
+  rcases c with ⟨cache, hcache⟩
+  induction hcache with
+  | empty => simp at hfound
+  | push_id wf ih =>
+    specialize ih hfound
+    simp
+    omega
+  | push_cache wf ih =>
+    rename_i decl'
```
semorrison, line 111: I guess this is actually uglier? (I am always sad about `rename_i`.)

```suggestion
  | @push_cache _ _ decl' wf ih =>
```

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+    simp only [HashMap.getElem?_insert] at hfound
+    match heq:decl == decl' with
```
semorrison, line 113:
```suggestion
    match heq : decl == decl' with
```

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+    | true =>
+      simp only [beq_iff_eq] at heq
+      simp only [heq, beq_self_eq_true, ↓reduceIte, Option.some.injEq] at hfound
+      simp
+      omega
+    | false =>
+      simp only [BEq.symm_false heq, Bool.false_eq_true, ↓reduceIte] at hfound
+      specialize ih hfound
+      simp
+      omega
```
semorrison, line 123: Perhaps:
```suggestion
    simp only [HashMap.getElem?_insert, beq_iff_eq] at hfound
    split at hfound <;> rename_i h
    · subst h
      simp_all
    · specialize ih hfound
      simp
      omega
```

src/Std/Sat/AIG/Basic.lean
semorrison: Also below.

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+  -/
+  match hfound:cache.val[decl]? with
+  | some hit =>
+    some ⟨hit, Cache.get?_bounds _ _ hfound, Cache.get?_property _ _ hfound⟩
+  | none => none
+
+/--
+An `Array Decl` is a Direct Acyclic Graph (DAG) if a gate at index `i` only points to nodes with index lower than `i`.
+-/
+def IsDag (α : Type) (decls : Array (Decl α)) : Prop :=
```
semorrison, line 190: Maybe `IsDAG`? Don't really mind.

semorrison: ("Dag" is a word in Australian English. :-)

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+  it either in `whnf` or in the kernel. This causes *huge* performance issues in practice.
+  The function can still be fully verified as all the proofs we need are in `CacheHit`.
+  -/
+  match hfound:cache.val[decl]? with
+  | some hit =>
+    some ⟨hit, Cache.get?_bounds _ _ hfound, Cache.get?_property _ _ hfound⟩
+  | none => none
+
+/--
+An `Array Decl` is a Direct Acyclic Graph (DAG) if a gate at index `i` only points to nodes with index lower than `i`.
```
semorrison, line 188: This uses "gate" and "node", but probably should say "node" both times?

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+-/
+theorem IsDag.empty {α : Type} : IsDag α #[] := by
+  intro i lhs rhs linv rinv h
+  simp only [Array.size_toArray, List.length_nil] at h
+  omega
+
+end AIG
+
+/--
+An And Inverter Graph (AIG α) together with a cache for subterm sharing.
```
semorrison, line 205:
```suggestion
An And Inverter Graph (`AIG α`) together with a cache for subterm sharing.
```
but maybe just drop the parenthetical?

src/Std/Sat/AIG/Basic.lean
semorrison: I had to click "show resolved" because I had exactly the same confusion. :-) Perhaps worth renaming.

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+/--
+The atom `a` occurs in `aig`.
+-/
+def Mem (a : α) (aig : AIG α) : Prop := (.atom a) ∈ aig.decls
+
+instance : Membership α (AIG α) where
+  mem := Mem
+
+/--
+A reference to a node within an AIG. This is the `AIG` analog of `Bool`.
```
semorrison, line 237: I don't understand the intent of "This is the `AIG` analog of `Bool`."

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+      (idx : Fin decls.size) : String :=
+    match decls[idx] with
+    | Decl.const b => s!"{idx} [label=\"{b}\", shape=box];"
+    | Decl.atom i => s!"{idx} [label=\"{i}\", shape=doublecircle];"
+    | Decl.gate _ _ _ _ => s!"{idx} [label=\"{idx} ∧\",shape=trapezium];"
+
+/--
+A stream of references into `aig`. This is the `AIG` analog of `BitVec`.
+-/
+structure RefStream (aig : AIG α) (w : Nat) where
```
semorrison, line 337: My vote is for not using "stream" here.

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+  refs : Array Nat
+  hlen : refs.size = w
+  hrefs : ∀ (h : i < w), refs[i] < aig.decls.size
+
+/--
+A sequence of references bundled with their AIG. In some sense a generalization of `Entrypoint`.
```
semorrison, line 343:
```suggestion
A sequence of references bundled with their AIG. In some sense a generalization of `Entrypoint`.
```
Just drop "in some sense", or explain the sense. Otherwise it just costs reader time to work it out.

src/Std/Sat/AIG/Basic.lean
semorrison: But I think we should nevertheless have doc-strings, even if just "an X bundled with Y".

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+  distance : AIG.RefStream aig n
+
+structure ExtendTarget (aig : AIG α) (newWidth : Nat) where
+  w : Nat
+  stream : AIG.RefStream aig w
+
+/--
+Evaluate an `AIG.Entrypoint` using some assignment for atoms.
+-/
+def denote (assign : α → Bool) (entry : Entrypoint α)  : Bool :=
```
semorrison, line 365:
```suggestion
def denote (assign : α → Bool) (entry : Entrypoint α) : Bool :=
```

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+    | .const b => b
+    | .atom v => assign v
+    | .gate lhs rhs linv rinv =>
+      have := h2 h1 h3
+      let lval := go lhs decls assign (by omega) h2
+      let rval := go rhs decls assign (by omega) h2
+      xor lval linv && xor rval rinv
+
+/--
+Denotation of an `AIG` at a specific `Ref`.
```
semorrison, line 381:
```suggestion
Denotation of an `AIG` at a specific `Entrypoint`.
```

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+-/
+scoped syntax "⟦" term ", " term "⟧" : term
+
+/--
+Denotation of an `AIG` at a specific `Ref` with the `Ref` being constructed on the fly.
```
semorrison, line 386: Maybe update too?

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+-/
+structure Fanin (aig : AIG α) where
+  /--
+  The node we are referring to.
+  -/
+  ref : Ref aig
+  /--
+  Whether the node is inverted
+  -/
+  inv : Bool
```
semorrison, line 424: cf. complaint about using `inv` above with a different meaning.

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+The `Ref.cast` equivalent for `Fanin`.
+-/
+@[inline]
+def Fanin.cast {aig1 aig2 : AIG α} (fanin : Fanin aig1)
+    (h : aig1.decls.size ≤ aig2.decls.size) :
+    Fanin aig2 :=
+  { fanin with ref := fanin.ref.cast h }
+
+/--
+The input type for creating AIG and gates from scratch.
```
semorrison, line 436: Not clear what "from scratch" means.

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+The `Ref.cast` equivalent for `GateInput`.
+-/
+@[inline]
+def GateInput.cast {aig1 aig2 : AIG α} (input : GateInput aig1)
+    (h : aig1.decls.size ≤ aig2.decls.size) :
+    GateInput aig2 :=
+  { input with lhs := input.lhs.cast h, rhs := input.rhs.cast h }
+
+/--
+Build an AIG gate in `aig`. Note that this version is only meant for proving,
```
semorrison, line 452:
```suggestion
Add a new and inverter gate to the AIG in `aig`. Note that this version is only meant for proving,
```

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,506 @@
...
+  have inv := by
+    intro i lhs rhs linv rinv h1 h2
+    simp only [Array.get_push] at h2
+    split at h2
+    . apply aig.inv <;> assumption
+    . contradiction
+  ⟨{ decls, inv, cache }, ⟨g, by simp [decls]⟩⟩
+
+/--
+Build a constant node in `aig`. Note that this version is only meant for proving,
```
semorrison, line 488:
```suggestion
Add a new constant node to `aig`. Note that this version is only meant for proving,
```

src/Std/Sat/AIG/CNF.lean
```diff
@@ -0,0 +1,721 @@
...
+
+@[simp]
+theorem projectRightAssign_property :
+    (projectRightAssign assign) x hx = (assign <| .inr ⟨x, hx⟩) := by
+  simp [projectRightAssign]
+
+/--
+Given an atom assignment, produce an assignment that will always satisfy the CNF generated by our
+Tseitin transformation. This is done by combining the atom assignment with an assignment for the
+auxiliary variables, that just evaluates the AIG at the corresponding node.
```
semorrison, line 120:
```suggestion
Tseitin transformation. This is done by combining the atom assignment with an assignment for the
auxiliary variables, that just evaluating the AIG at the corresponding node.
```

src/Std/Sat/AIG/CNF.lean
```diff
@@ -0,0 +1,721 @@
...
+      omega
+    | inr =>
+      dsimp only [inj] at h
+      congr
+      omega
+
+/--
+The node that we started CNF conversion at will always be marked as visited in the CNF cache.
+-/
+theorem toCNF.go_marked :
```
semorrison, line 656: Being excessively obsessive about naming, this could just be `go_marks`. :-)

src/Std/Sat/AIG/CNF.lean
```diff
@@ -0,0 +1,721 @@
...
+Connect SAT results about the CNF to SAT results about the AIG.
+-/
+theorem toCNF.go_as_denote (aig : AIG Nat) (start) (h1) (assign1) :
+    ((⟦aig, ⟨start, h1⟩, assign1⟧ && (go aig start h1 (.empty aig)).val.eval (cnfSatAssignment aig assign1)) = sat?)
+      →
+    (⟦aig, ⟨start, h1⟩, assign1⟧ = sat?) := by
+  intro h
+  have := go_sat aig start h1 assign1 (.empty aig)
+  simp only [State.Sat, CNF.sat_def] at this
+  simpa [this] using h
```
semorrison, line 680: The logic here is a bit weird? The first fact here is the interesting one, then it's just logic puzzle?
```
theorem toCNF.go_as_denote' (aig : AIG Nat) (start) (h1) (assign1) :
    ⟦aig, ⟨start, h1⟩, assign1⟧ → (go aig start h1 (.empty aig)).val.eval (cnfSatAssignment aig assign1) := by
  have := go_sat aig start h1 assign1 (.empty aig)
  simp only [State.Sat, CNF.sat_def] at this
  simp [this] 

/--
Connect SAT results about the CNF to SAT results about the AIG.
-/
theorem toCNF.go_as_denote (aig : AIG Nat) (start) (h1) (assign1) :
    ((⟦aig, ⟨start, h1⟩, assign1⟧ && (go aig start h1 (.empty aig)).val.eval (cnfSatAssignment aig assign1)) = sat?)
      →
    (⟦aig, ⟨start, h1⟩, assign1⟧ = sat?) := by
  have := go_as_denote' aig start h1 assign1
  by_cases CNF.eval (cnfSatAssignment aig assign1) (go aig start h1 (State.empty aig)).val.cnf <;> simp_all
```

src/Std/Sat/AIG/Cached.lean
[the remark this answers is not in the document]
semorrison: Maybe make clear whether this is part of the current progression of PRs, or work to be done later?

src/Std/Sat/AIG/CachedGates.lean
```diff
@@ -0,0 +1,189 @@
...
+
+namespace AIG
+
+variable {α : Type} [Hashable α] [DecidableEq α]
+
+/--
+Create a not gate in the input AIG. This uses the builtin cache to enable automated subterm sharing.
+-/
+def mkNotCached (aig : AIG α) (gate : Ref aig) : Entrypoint α :=
+  -- ¬x = true && invert x
```
semorrison, line 25:
```suggestion
```

src/Std/Sat/AIG/CachedGates.lean
```diff
@@ -0,0 +1,189 @@
...
+          intros
+          apply LawfulOperator.le_size_of_le_aig_size (f := mkConstCached)
+          omega
+        inv := true
+      }
+    }
+
+@[inline]
+def BinaryInput.asGateInput {aig : AIG α} (input : BinaryInput aig) (linv rinv : Bool)
+    : GateInput aig :=
```
semorrison, line 46:
```suggestion
def BinaryInput.asGateInput {aig : AIG α} (input : BinaryInput aig) (linv rinv : Bool) :
    GateInput aig :=
```

src/Std/Sat/AIG/CachedGates.lean
```diff
@@ -0,0 +1,189 @@
...
+        inv := false
+      }
+      rhs := {
+        ref := gate.cast <| by
+          intros
+          apply LawfulOperator.le_size_of_le_aig_size (f := mkConstCached)
+          omega
+        inv := true
+      }
+    }
```
semorrison, line 42:
```suggestion
    { lhs := 
      { ref := constRef
        inv := false }
      rhs := 
      { ref := gate.cast <| by
          intros
          apply LawfulOperator.le_size_of_le_aig_size (f := mkConstCached)
          omega
        inv := true } }
```
I guess we don't have a definitive style rule on this. Mathlib does it this way...

src/Std/Sat/AIG/CachedLemmas.lean
```diff
@@ -0,0 +1,340 @@
...
+Copyright (c) 2024 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Henrik Böving
+-/
+import Std.Sat.AIG.Cached
+
+/-!
+This module contains the theory of the cached AIG node creation functions.
+It is mainly concerned with proving lemmas about the denotational semantics of the gate functions
+in different scenarios. In particular reductions to the semantics of the non cached versions.
```
semorrison, line 11:
```suggestion
in different scenarios, in particular reductions to the semantics of the non cached versions.
```

src/Std/Sat/AIG/LawfulOperator.lean
```diff
@@ -0,0 +1,155 @@
...
+    -/
+    size_le : decls1.size ≤ decls2.size
+    /--
+    The prefix and the other array must agree on all elements up until the bound of the prefix
+    -/
+    idx_eq : ∀ idx (h : idx < decls1.size), decls2[idx]'(by omega) = decls1[idx]'h
+
+/--
+If `decls1` is a prefix of `decls2` and we start evaluating `decls2` at an
+index in bounds of `decls1` we can evluate at `decls1`.
```
semorrison, line 34:
```suggestion
index in bounds of `decls1` we can evaluate at `decls1`.
```

src/Std/Sat/AIG/LawfulOperator.lean
```diff
@@ -0,0 +1,155 @@
...
+      congr 2
+      . apply denote.go_eq_of_IsPrefix
+        assumption
+      . apply denote.go_eq_of_IsPrefix
+        assumption
+termination_by start
+
+variable {α : Type} [Hashable α] [DecidableEq α]
+
+@[inherit_doc denote.go_eq_of_IsPrefix ]
```
semorrison, line 68:
```suggestion
@[inherit_doc denote.go_eq_of_IsPrefix]
```

semorrison:
❤️

src/Std/Sat/AIG/Cached.lean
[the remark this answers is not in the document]
hargoniX: Work to be done later.

src/Std/Sat/AIG/CachedGates.lean
TwoFX: FWIW, I looked at this a few weeks ago and came to the conclusion that Lean uses the original style quite consistently, while mathlib exclusively uses the style you suggested. So I believe that it would be more consistent to format it the way Henrik had it originally.

src/Std/Sat/AIG/Basic.lean
hargoniX: `Bool` is a single bit and just like that an `Ref` is a single bit but encoded with the AIG. Similarly `BitVec` is a collection of bits and `RefVec` is a collection of bits that are encoded within the AIG.

src/Std/Sat/AIG/CachedGates.lean
hargoniX: I'm happy to change it either way, just need a clear decision^^.

src/Std/Sat/AIG/LawfulOperator.lean
```diff
@@ -0,0 +1,155 @@
...
+    /--
+    The prefix and the other array must agree on all elements up until the bound of the prefix
+    -/
+    idx_eq : ∀ idx (h : idx < decls1.size), decls2[idx]'(by omega) = decls1[idx]'h
+
+/--
+If `decls1` is a prefix of `decls2` and we start evaluating `decls2` at an
+index in bounds of `decls1` we can evaluate at `decls1`.
+-/
+theorem denote.go_eq_of_IsPrefix (decls1 decls2 : Array (Decl α)) (start : Nat) {hdag1} {hdag2}
```
TwoFX, line 36: This should be called `denote.go_eq_of_isPrefix` (the rule is that in theorem names, each component should start with a lowercase letter). Please check that you adhere to this rule everywhere.

src/Std/Sat/AIG/Basic.lean
```diff
@@ -0,0 +1,510 @@
...
+-/
+
+variable {α : Type} [Hashable α] [DecidableEq α]
+
+namespace AIG
+
+/--
+A circuit node. These are not recursive but instead contain indices into an `AIG`, with inputs indexed by `α`.
+-/
+inductive Decl (α : Type) where
```
TwoFX, line 25: I don't think this needs changing, but is there a reason why you are not universe-polymorphic here?

src/Std/Sat/AIG/CachedGates.lean
TwoFX: I think there is no clear consensus in the existing Lean codebase at the moment (including mixing different styles within the same style), so I think in the interest of making progress I suggest leaving things as you have them now and not losing too much time on this as long as they adhere to one of these styles. However, neither style should have an opening brace on its own line (so [this](https://github.com/leanprover/lean4/pull/4953/files#diff-86e39419197a4029f16533cec6041c2f9c1d08882e2e4007609f78a06b27620cR72-R73) is not okay), and within a declaration you should stick to one style (so [this](https://github.com/leanprover/lean4/pull/4953/files#diff-05f9943490c5a32907d1aec86f0c49bc9002579f8c383a574f8afaf6463a0fb8R29-R37) is not okay).

TwoFX (approved):
We could probably spend more time nitpicking here and there but I don't think there is much point, so let's get this merged after these final remarks.

src/Std/Sat/AIG/Basic.lean
hargoniX: No particular reason.

PR #4956, opened 2024-08-08
fix: export more symbols needed by Verso

david-christiansen:
This enables the Verso LSP server extensions to work.


david-christiansen:
The Windows CI didn't build stages 2 or 3. I'm going to temporarily push a stage0 update to make sure the symbol limit hasn't been exceeded.

PR #4963, opened 2024-08-08
feat: link LibUV

TwoFX:


TwoFX:
@ydewit Yes, the ultimate goal is support for non-blocking IO. Note however that this PR only links against LibUV and it might take a very long time until we actually start building anything on top of it, so if you have need for asynchronous IO right now then it probably makes sense to explore alternatives such as [lean-libuv](https://github.com/the-sofi-uwu/lean-libuv/tree/socket-fix) for now.

hargoniX (commented):
We should probably add to the docs page on building Lean that you need to have libuv installed right? Alternatively we could do something similar to the CaDiCal PR and declare libuv as an external project that gets downloaded and compiled by our build system directly.

TwoFX:
> We should probably add to the docs page on building Lean that you need to have libuv installed right? Alternatively we could do something similar to the CaDiCal PR and declare libuv as an external project that gets downloaded and compiled by our build system directly.

Thanks, I've updated the build instructions. We don't want to build libuv as part of our build system, but use either system libraries (when building manually) or take static libraries from Nix (when creating release versions via CI), just as we do it with GMP.

Kha (approved):

PR #4966, opened 2024-08-09
chore: constructorNameAsVariable linter respects linter.all

kim-em:
This PR ensures `linter.all` disables `constructorNameAsVariable`.

The issue was discovered by @eric-wieser while investigating a quote4 issue.

This seems like an easy mistake to make when setting up a new linter, and perhaps we need a better structure to make it easy to do the right thing.


semorrison:
Hmm, the test file works fine interactively, where `linter.all` is not set, but I can't see how to test this properly without upstreaming `unset_option` from Mathlib.

david-christiansen:
Modulo the test issue, LGTM.

I agree about a better structure here, but I'm not sure what it should be - might as well fix the bug in the meantime.

Thanks!

david-christiansen (approved):

david-christiansen:
I suppose for test purposes, it would work fine to inline the part of `unset_option` that does the job into just the test file here.

src/Lean/Linter/ConstructorAsVariable.lean
```diff
@@ -53,7 +58,7 @@ def constructorNameAsVariable : Linter where
               if !cmdStxRange.contains range.start || ldecl.userName.hasMacroScopes then return
               let opts := ci.options
               -- we have to check for the option again here because it can be set locally
-              if !linter.constructorNameAsVariable.get opts then return
+              if !getLinterValue linter.constructorNameAsVariable opts then return
```
eric-wieser, line 61: Any reason you use getLinterConstructorNameAsVariable at the top of the function but inline it here? I'd lean towards not having `getLinterConstructorNameAsVariable` at all, but if you really want it then it would be good to use consistently.

kim-em: Fixed

PR #4978, opened 2024-08-10
perf: inline `lean_inc_ref_cold`

Kha:
The body is a single instruction

### DIFF
--- a/src/include/lean/lean.h
+++ b/src/include/lean/lean.h
@@ -23,6 +23,7 @@ Author: Leonardo de Moura
 #define LEAN_USING_STD using namespace std; /* NOLINT */
 extern "C" {
 #else
+#include <stdatomic.h>
 #define  LEAN_USING_STD
 #endif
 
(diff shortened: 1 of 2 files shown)


Kha (commented):
!bench

Kha:
!bench

Kha:
Platform comparison: https://godbolt.org/z/aGdqqKf8a. The odd one out is arm64, it needs `-march=v8.1a` to generate short code. Which should be okay, but also this is a Tier 2 platform.

Kha:
An interesting independent observation is that mingw apparently defaults to -fno-omit-frame-pointer, for this function alone that should have significant impact before this PR.

Kha:
Continuing the investigation into `lean_dec_ref`, it gets a bit more interesting: https://godbolt.org/z/n6vbWTYP4. The current version is already 9 (x64)/8 (aarch64) instructions. Inlining the atomic adds 0 instructions to the likely branch `rc > 1` and 4/6 instructions otherwise. A slight variation adds 1 unlikely conditional jump to the likely branch and 1/4 instructions otherwise.

Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
!bench

Kha:
The above is the speedup remaining after fixing #5004.

Some more data: on aarch64 macOS, leanshared does grow by 1.8% as it's two instructions there usually (+0.3M instructions out of 14.8M). Curiously the .a files are about equally smaller, because of less relocation information? We might want to check if there is any performance impact from this. But this also reminds me that this optimization may be much more important on Windows where this is otherwise a cross-DLL call through the Global Offset Table!

zwarich:
I just tested locally and got a 1% reduction in `libsharedlean.dylib` size. This is using Apple Clang 16, which apparently corresponds to public LLVM 17.

zwarich:
!bench

zwarich:
Still a 1.3% reduction in code size on ARM macOS.

ghost:
Mathlib CI status ([docs](https://leanprover-community.github.io/contribute/tags_and_branches.html)):
- ❗ Batteries/Mathlib CI will not be attempted unless your PR branches off the `nightly-with-mathlib` branch. Try `git rebase 7845af3105f1df43dadded4e1ef6f210853c03b3 --onto d3dda9f6d4428a906c096067ecb75e432afc4615`. You can force Mathlib CI using the `force-mathlib-ci` label. (2025-06-27 13:22:19)
- ❗ Batteries/Mathlib CI will not be attempted unless your PR branches off the `nightly-with-mathlib` branch. Try `git rebase bd16c0f87d978193d2988fdd09bf906fd424a57e --onto d3dda9f6d4428a906c096067ecb75e432afc4615`. You can force Mathlib CI using the `force-mathlib-ci` label. (2025-06-27 15:33:30)

zwarich:
!bench

PR #4994, opened 2024-08-12
perf: destructive update at `expr_eq_fn`

leodemoura:
This is an experiment. It currently only performs the destructive update at applications.


leodemoura:
!bench

leodemoura:
Remark: the optimization should check whether a child was considered equal but binder information is different because `CompareBinderInfo == false`.

PR #4996, opened 2024-08-12
fix: reduce default max depth for ext tactic

david-christiansen:
The prior default of 1000000 could not be achieved in practice, because the stack would overflow after around 5000 recursive invocations. This meant that a poorly-chosen @[ext] lemma could crash Lean.

Talking to Mathlib users, it seems that 10 would be a very large number in practice, so a default limit of 100 should not change successful uses. But it does make it much easier to diagnose and recover from poor choices of @[ext] lemmas.


david-christiansen:
A bit of context: while working on #4970, Lean was mysteriously crashing while I tried to update the test suite. It turned out that there was a stack overflow, but on my machine it was terminating with `SIGBUS` - hence #4971. This gave an uninformative, scary error message.

Hopefully this and #4971 will save future experimenters from having to fire up `gdb` or `lldb`!

eric-wieser:
> it seems that 10 would be a very large number in practice

I've needed 10-20 a few times, but I agree that 100 is probably safe.

Regarding the sigbus; should this tactic also call `Lean.withIncRecDepth` to ensure that any stack overflow is caught as a runtime error rather than a process crash?

PR #4999, opened 2024-08-12
feat: Adding some instances for `PtrSet` and `PtrMap`

znssong:
We added some instances for type `PtrSet` and `PtrMap`.


Kha:
Hi, we took a look at this PR in the triage team meeting and ultimately decided that we would prefer unsafe definitions not to be hidden behind notations. You can define these instances locally if you prefer them.

PR #5003, opened 2024-08-12
chore: upgrade cache action to silence warnings

hargoniX:
According to the release notes of cache this should not break anything as they merely upgraded the node version.


nomeata (approved):

Kha:
Can we get `save-always: true` pls? :)

PR #5004, opened 2024-08-12
perf: avoid MT marking environment in language processor

Kha:
#4976 moved resolution of a promise to an earlier point, but that led to object being marked MT earlier, so we need to move the code that minimizes those objects earlier too to revert the performance regression.


Kha:
!bench

Kha:
!bench

PR #5006, opened 2024-08-12
doc: update user widget manual

Vtec234:
Updates the user widget manual to account for more recent changes. One issue is that the samples no longer work on https://live.lean-lang.org/ because it uses an outdated version of the `@leanprover/infoview` NPM package. They work on https://lean.math.hhu.de/ and in recent versions of the VSCode extension.

### DIFF
--- a/doc/examples/widgets.lean
+++ b/doc/examples/widgets.lean
@@ -4,15 +4,18 @@ open Lean Widget
 /-!
 # The user-widgets system
 
-Proving and programming are inherently interactive tasks. Lots of mathematical objects and data
-structures are visual in nature. *User widgets* let you associate custom interactive UIs with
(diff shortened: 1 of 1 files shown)


src/Lean/Widget/UserWidget.lean
```diff
@@ -78,10 +78,7 @@ builtin_initialize moduleRegistry : ModuleRegistry ←
     toArrayFn     := fun es => es.toArray
   }
 
-/--
-Registers `[builtin_widget_module]` and `[widget_module]` and binds the latter's implementation
-(used for creating the obsolete `[widget]` alias below).
- -/
+/-- Registers a widget module. Its type must implement `Lean.Widget.ToModule`. -/
```
Vtec234, line 81: Moved implementation detail below because the docstring is what's shown in hovers.

PR #5007, opened 2024-08-13
feat: express `BitVec.extractLsb'` in terms of `extractLsb`

alexkeizer:
Adds a lemma to rewrite `BitVec.extractLsb'` to `extractLsb` plus a cast. 
Note that `extractLsb'` with a length of 0 returns `BitVec 0`, while `extractLsb` will never return an empty bitvector (because of the `+ 1` in it `hi - lo + 1`). Hence, this lemma needs a side condition that the length is non-zero.

Also adds `getLsb_extractLsb'`

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -472,7 +472,6 @@ protected theorem extractLsb_ofNat (x n : Nat) (hi lo : Nat) :
 @[simp] theorem extractLsb_toNat (hi lo : Nat) (x : BitVec n) :
   (extractLsb hi lo x).toNat = (x.toNat >>> lo) % 2^(hi-lo+1) := rfl
 
-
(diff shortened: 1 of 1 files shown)


src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -472,10 +472,19 @@ protected theorem extractLsb_ofNat (x n : Nat) (hi lo : Nat) :
 @[simp] theorem extractLsb_toNat (hi lo : Nat) (x : BitVec n) :
   (extractLsb hi lo x).toNat = (x.toNat >>> lo) % 2^(hi-lo+1) := rfl
 
+
+@[simp] theorem getLsb_extractLsb' (start len : Nat) (x : BitVec n) (i : Nat) :
```
tobiasgrosser, line 476:
```suggestion
@[simp] theorem getLsb_extractLsb' (start len : Nat) (x : BitVec n) (i : Nat) :
```

PR #5013, opened 2024-08-13
feat: upstream LeanSAT's bitblaster

hargoniX:
Step 3/~7 in upstreaming LeanSAT.

A few thoughts:
- Why is this not in `Std.Sat`? LeanSAT's bitblaster operates on a limited internal language. For example it has no idea that signed comparision operators even exist. This is because it relies on a normalization pass before being given the goal. For this reason I would not classify the bitblaster as an API that we should publicly advertise at this abstraction level
- Sometimes I slightly rebuild parts of the LawfulOperator infrastructure for operators that work non-tail-recursively. This is because they do not return an `Entrypoint` but instead an `ExtendingEntrypoint` in order to even be defined in the first place (casting Ref's and all that). Given the fact that this barely happens and I never actually commit to rebuilding the full API I'm hoping that this is indeed a fine decision?
- The single explicit `decreasing_by` that has a simp only which *almost* looks like `simp_wf` is missing a singular lemma from `simp_wf` because it doesn't terminate otherwise.
- I am not using functional induction because it basically always fails at some generalization step, that is also the reason that there is lots of explicit `generalize` and manually recursive proofs.


src/Lean/Elab/Tactic/BvDecide/BitBlast/BVExpr/BitBlast/Lemmas/Add.lean
```diff
@@ -0,0 +1,264 @@
...
+    lhs
+    rhs
+    rhs
+    rw [LawfulOperator.denote_mem_prefix (f := mkXorCached) (h := input.rhs.hgate)]
+
+theorem mkFullAdder_denote_mem_prefix (aig : AIG α) (input : FullAdderInput aig) (start : Nat)
+    (hstart) :
+    ⟦
+      (mkFullAdder aig input).aig,
+      ⟨start, by apply Nat.lt_of_lt_of_le; exact hstart; apply FullAdderOutput.hle⟩,
```
TwoFX, line 97:
```suggestion
      ⟨start, Nat.lt_of_lt_of_le hstart (FullAdderOutput.hle _)⟩,
```

src/Lean/Elab/Tactic/BvDecide/BitBlast/BVExpr/BitBlast/Lemmas/Add.lean
```diff
@@ -0,0 +1,264 @@
...
+        rw [AIG.LawfulOperator.denote_mem_prefix (f := mkFullAdderOut)]
+        rw [hleft, hright, hcin]
+        simp [atLeastTwo_eq_halfAdder]
+      . omega
+  . omega
+termination_by w - curr
+
+end blastAdd
+
+theorem blastAdd_denote_eq (aig : AIG α) (lhs rhs : BitVec w) (assign : α → Bool)
```
TwoFX, line 235: Technically many theorems here have incorrect names, for example this one should be called `denote_blastAdd`.

src/Lean/Elab/Tactic/BvDecide/BitBlast/BVExpr/BitBlast/Impl/Extract.lean
```diff
@@ -0,0 +1,74 @@
...
+    AIG.RefVecEntry α newWidth :=
+  let ⟨input, hi, lo, hnew⟩ := target
+  let res := aig.mkConstCached false
+  let aig := res.aig
+  let falseRef := res.ref
+  let input := input.cast <| AIG.LawfulOperator.le_size (f := AIG.mkConstCached) ..
+  if h : lo ≤ hi then
+    ⟨aig, go input lo falseRef 0 (by omega) .empty⟩
+  else
+    have : 1 = newWidth  := by omega
```
TwoFX, line 40:
```suggestion
    have : 1 = newWidth := by omega
```

src/Lean/Elab/Tactic/BvDecide/BitBlast/BVExpr/BitBlast/Impl/Extract.lean
```diff
@@ -0,0 +1,74 @@
...
+  go {aig : AIG α} {w : Nat} (input : AIG.RefVec aig w) (lo : Nat) (falseRef : AIG.Ref aig)
+      (curr : Nat) (hcurr : curr ≤ newWidth) (s : AIG.RefVec aig curr) :
+      AIG.RefVec aig newWidth :=
+  if h : curr < newWidth then
+    let nextRef := input.getD (lo + curr) falseRef
+    let s := s.push nextRef
+    go input lo falseRef (curr + 1) (by omega) s
+  else
+    have : curr = newWidth := by omega
+    this ▸ s
```
TwoFX, line 54: Something like `RefVec.castLen` would probably have been cleaner, but if this works then :shrug:

src/Lean/Elab/Tactic/BvDecide/BitBlast/BVExpr/BitBlast/Impl/ShiftLeft.lean
```diff
@@ -0,0 +1,218 @@
...
+structure TwoPowShiftTarget (aig : AIG α) (w : Nat) where
+  n : Nat
+  lhs : AIG.RefVec aig w
+  rhs : AIG.RefVec aig n
+  pow : Nat
+
+def twoPowShift (aig : AIG α) (target : TwoPowShiftTarget aig w) : AIG.RefVecEntry α w :=
+  let ⟨n, lhs, rhs, pow⟩ := target
+  if h : pow < n then
+    let res := blastShiftLeftConst aig ⟨lhs, (2 ^ pow) % 2^n⟩
```
TwoFX, line 111: I don't understand what's happening here. Isn't `(2^pow)%(2^n)` always equal to `2^pow`?

src/Lean/Elab/Tactic/BvDecide/BitBlast/BVExpr/BitBlast/Impl/Extract.lean
hargoniX: I tried to write this theorem but it wouldn't type check when using quite literally the same proof strategy. I tried to get it to work for a while but eventually this solution seemed easier.

src/Lean/Elab/Tactic/BvDecide/BitBlast/BVExpr/BitBlast/Lemmas/Add.lean
```diff
@@ -0,0 +1,264 @@
...
+        (h := by
+          apply LawfulOperator.lt_size_of_lt_aig_size (f := mkXorCached)
+          apply Ref.hgate
+        )
+    ]
+  conv =>
+    lhs
+    rhs
+    rhs
+    rw [LawfulOperator.denote_mem_prefix (f := mkXorCached) (h := input.rhs.hgate)]
```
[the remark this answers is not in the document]
hargoniX: It does type check slower for me but I agree it looks much nicer :+1:

src/Lean/Elab/Tactic/BvDecide/BitBlast/BVExpr/Basic.lean
```diff
@@ -0,0 +1,444 @@
...
+  /--
+  Bitwise or.
+  -/
+  | or
+  /--
+  Bitwise xor.
+  -/
+  | xor
+  /--
+  Addition.
```
tobiasgrosser, line 61: Not sure what best practice is here, but I typically terminate `sentences` with `.` and expressions that lack a verb remain without terminator.

src/Lean/Elab/Tactic/BvDecide/BitBlast/BVExpr/Basic.lean
```diff
@@ -0,0 +1,444 @@
...
+  bv: BitVec w
+
+/--
+The notion of variable assignments for `BVExpr`.
+-/
+abbrev Assignment := List PackedBitVec
+
+/--
+Obtaining the value of a `BVExpr.var` from an `Assignment`.
+-/
```
tobiasgrosser, line 267:
```suggestion
Get the value of a `BVExpr.var` from an `Assignment`.
-/
```

src/Lean/Elab/Tactic/BvDecide/BitBlast/BVExpr/BitBlast/Impl/Add.lean
```diff
@@ -0,0 +1,247 @@
...
+
+end FullAdderInput
+
+
+def mkFullAdderOut (aig : AIG α) (input : FullAdderInput aig) : AIG.Entrypoint α :=
+  -- let subExpr = XOR lhs rhs
+  -- out = XOR subExpr cin
+  -- subExpr is shared in `mkFullAdderOut` and `mkFullAdderCarry`
+  -- Due to automated subterm sharing in the AIG we don't have to make this explicitly sure
+  let ⟨lhs, rhs, cin⟩ := input
```
tobiasgrosser, line 63:
```suggestion
  -- Due to automated subterm sharing in the AIG we don't have to ensure this explicitly.
  let ⟨lhs, rhs, cin⟩ := input
```

src/Lean/Elab/Tactic/BvDecide/BitBlast/BVExpr/BitBlast/Impl/Add.lean
```diff
@@ -0,0 +1,247 @@
...
+    rw [AIG.LawfulOperator.decl_eq]
+    apply AIG.LawfulOperator.lt_size_of_lt_aig_size
+    assumption
+
+def mkFullAdderCarry (aig : AIG α) (input : FullAdderInput aig) : AIG.Entrypoint α :=
+  -- let subExpr = XOR lhs rhs
+  -- cout = OR (AND subExpr cin) (AND lhs rhs)
+  -- subExpr is shared in `mkFullAdderOut` and `mkFullAdderCarry`
+  -- Due to automated subterm sharing in the AIG we don't have to make this explicitly sure
+  let ⟨lhs, rhs, cin⟩ := input
```
tobiasgrosser, line 91:
```suggestion
  -- Due to automated subterm sharing in the AIG we don't have to ensure this explicitly.
  let ⟨lhs, rhs, cin⟩ := input
```

src/Lean/Elab/Tactic/BvDecide/BitBlast/BVExpr/BitBlast/Impl/Pred.lean
```diff
@@ -0,0 +1,99 @@
...
+    let aig := res.aig
+    let rhsRefs := res.vec
+    let lhsRefs := lhsRefs.cast <| AIG.LawfulVecOperator.le_size (f := BVExpr.bitblast) ..
+    match op with
+    | .eq => mkEq aig ⟨lhsRefs, rhsRefs⟩
+    | .ult => mkUlt aig ⟨lhsRefs, rhsRefs⟩
+  | .getLsb expr idx =>
+    /-
+    Note: This blasts the entire rexpression up to `w` despite only needing it up to `idx`.
+    However the vast majority of operations are interested in all bits so the API is currently
```
tobiasgrosser, line 39:
```suggestion
    Note: This blasts the entire expression up to `w` despite only needing it up to `idx`.
    However the vast majority of operations are interested in all bits so the API is currently
```

tobiasgrosser (commented):
Just some minor comments. Feel free to ignore, especially the one about the terminating `.`.

src/Lean/Elab/Tactic/BVDecide/Bitblast/BVExpr/Circuit/Impl/Var.lean
```diff
@@ -0,0 +1,86 @@
...
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Henrik Böving
+-/
+prelude
+import Lean.Elab.Tactic.BVDecide.Bitblast.BVExpr.Basic
+import Std.Sat.AIG.CachedGatesLemmas
+import Std.Sat.AIG.LawfulVecOperator
+
+/-!
+This module contains the implementation of a bitblaster for symbolic `BitVec` values.
```
TwoFX, line 12: Optional suggestion: the code would have been slightly easier for me to navigate if every file that corresponds to an operation on `BitVec` lived in a subfolder `Impl/Operations` so that "special" files like `Expr.lean` and `Var.lean` stand out more.

src/Lean/Elab/Tactic/BVDecide/Bitblast/BVExpr/Circuit/Lemmas.lean
```diff
@@ -0,0 +1,44 @@
...
+
+theorem bitblast.go_eval_eq_eval (expr : BVLogicalExpr) (aig : AIG BVBit) (assign : BVExpr.Assignment) :
+    ⟦ofBoolExprCached.go aig expr BVPred.bitblast, assign.toAIGAssignment⟧ = expr.eval assign := by
+  induction expr generalizing aig with
+  | const => simp [ofBoolExprCached.go]
+  | literal => simp [ofBoolExprCached.go]
+  | not expr ih => simp [ofBoolExprCached.go, ih]
+  | gate g lhs rhs lih rih => cases g <;> simp [ofBoolExprCached.go, Gate.eval, lih, rih]
+
+theorem bitblast_denote_eq (expr : BVLogicalExpr) (assign : BVExpr.Assignment) :
```
TwoFX:
```suggestion
theorem denote_bitblast (expr : BVLogicalExpr) (assign : BVExpr.Assignment) :
```

src/Lean/Elab/Tactic/BVDecide/Bitblast/BVExpr/Circuit/Lemmas/Add.lean
```diff
@@ -0,0 +1,245 @@
...
+        rw [AIG.LawfulOperator.denote_mem_prefix (f := mkFullAdderOut)]
+        rw [hleft, hright, hcin]
+        simp [atLeastTwo_eq_halfAdder]
+      . omega
+  . omega
+termination_by w - curr
+
+end blastAdd
+
+theorem blastAdd_denote_eq (aig : AIG α) (lhs rhs : BitVec w) (assign : α → Bool)
```
TwoFX:
```suggestion
theorem denote_blastAdd (aig : AIG α) (lhs rhs : BitVec w) (assign : α → Bool)
```
I'll stop now, you get the idea.

src/Lean/Elab/Tactic/BVDecide/Bitblast/BVExpr/Circuit/Lemmas/Add.lean
```diff
@@ -0,0 +1,245 @@
...
+      (input : BinaryRefVec aig w)
+      (hleft : ∀ (idx : Nat) (hidx : idx < w), ⟦aig, input.lhs.get idx hidx, assign⟧ = lhs.getLsb idx)
+      (hright : ∀ (idx : Nat) (hidx : idx < w), ⟦aig, input.rhs.get idx hidx, assign⟧ = rhs.getLsb idx) :
+      ∀ (idx : Nat) (hidx : idx < w),
+          ⟦(blastAdd aig input).aig, (blastAdd aig input).vec.get idx hidx, assign⟧
+            =
+          (lhs + rhs).getLsb idx := by
+  intro idx hidx
+  rw [BitVec.getLsb_add]
+  . rw [← hleft idx hidx]
```
TwoFX, line 222: Took me a while to notice you're using `.` here instead of `·`. I'm pretty sure that the dominant style in `Init` and `Lean` (and definitely in `Std`) is to use the center dot.

TwoFX (approved):
LGTM

PR #5018, opened 2024-08-13
fix: report info trees on cmdline for .ileans

Kha:
In #4976, I forgot that we do need info trees eventually on the cmdline for .ilean generation. Unfortunately, not reporting them incrementally would require an API change, so let's see what the impact of incremental reporting is


Kha:
Merging as incrementally marking the environment as persistent is likely the way to go in any case. Further performance investigations pending.

PR #5022, opened 2024-08-13
chore: ci.yaml: build MacOS Aarch64 release for PRs by default

nomeata:
should make https://github.com/leanprover-community/mathlib4/pull/13301
unnecessary, which has a fair number of bad side-effects


mattrobball:
Thanks!

PR #5025, opened 2024-08-13
chore: pr-release: pass `--retry` to `curl`

nomeata:
Since https://github.com/curl/curl/pull/4465 curl adheres to the
`Retry-After` header, so maybe this fixes the issues with
```
jq: error (at <stdin>:5): Cannot index string with string "body"
```
that sometimes make this workflow fail.


nomeata:
This did not fix the issue, I fear: https://github.com/leanprover/lean4/actions/runs/10415672920/job/28846565313

PR #5030, opened 2024-08-14
chore: remove dead code in Lake.Util.Compare

kim-em:
While exploring refactors of `List.lt` I ran into errors here, in code that is entirely unused. Propose cleaning up to get things out of my way!


semorrison:
@tydeu, I'm going to merge this now, but ping me if I've removed anything you still want for later and we can discuss.

PR #5036, opened 2024-08-14
feat: improve unused section variable warning

Kha:
See https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Opt.20out.20of.20.22included.20section.20variable.20is.20not.20used.22.20linter


src/Lean/Elab/MutualDef.lean
```diff
@@ -399,18 +404,19 @@ private def elabFunValues (headers : Array DefViewElabHeader) (vars : Array Expr
...
               let var := if varDecl.userName.hasMacroScopes && varDecl.binderInfo.isInstImplicit then
-                m!"[{varDecl.type}]".group
+                m!"{indentD m!"[{varDecl.type}]"}{Format.line}".group
               else
-                var
-              logWarningAt header.ref m!"included section variable '{var}' is not used in \
-                '{header.declName}', consider excluding it"
+                m!" '{var}' "
+              Linter.logLint linter.unusedSectionVars header.ref
+                m!"included section variable{var}is unused in theorem, consider restructuring \
```
grunweg, line 418: Is the change from `header.declName` to `theorem` intentional? At least the error message itself being much less useful. (Or is the reasoning that the linter underlining a particular place will make up for this? In this case, I'd at least write "this theorem".)

grunweg (commented):
Thanks for writing this! I have one small wording suggestion.

src/Lean/Elab/MutualDef.lean
Kha: Yes, the warning position is the indicator as properly referring to the theorem name at this point is not possible

digama0:
Some suggestions:
* Could this be just one warning for all the variables instead of a separate warning for each variable? This also makes the next bullet easier...
* The warning should come with an explicit copy-pasteable suggestion on how to use `include` and `omit` in a way that will suppress the warning. (Keep in mind that this linter will probably be many people's first encounter with this command, so they won't know what the syntax is.)
* Can you even use `include` to suppress the warning? My investigations on zulip suggested no, it complains even if you explicitly use `include` or `include in` on an unused variable. I think this should not be the case, the warning should only trigger on variables that are neither `include`d nor `omit`ted.

src/Lean/Elab/MutualDef.lean
grunweg: I see. Can you write "this theorem" then?
```suggestion
                m!"included section variable{var}is unused in this theorem, consider restructuring \

PR #5039, opened 2024-08-14
feat: List.foldlRecOn

hargoniX:
As discussed with @semorrison, feel free to do whatever to the branch.

### DIFF
--- a/src/Init/Data/List/Lemmas.lean
+++ b/src/Init/Data/List/Lemmas.lean
@@ -750,22 +750,22 @@ Prove a proposition about the result of `List.foldl`,
 by proving it for the initial data,
 and the implication that the operation applied to any element of the list preserves the property.
 
-The motive `C` can take values in `Sort _`, so this may be used to construct data,
+The motive can take values in `Sort _`, so this may be used to construct data,
 as well as to prove propositions.
 -/
-def foldlRecOn {C : β → Sort _} : ∀ (l : List α) (op : β → α → β) (b : β) (_ : C b)
-    (_ : ∀ (b : β) (_ : C b) (a : α) (_ : a ∈ l), C (op b a)), C (List.foldl op b l)
+def foldlRecOn {motive : β → Sort _} : ∀ (l : List α) (op : β → α → β) (b : β) (_ : motive b)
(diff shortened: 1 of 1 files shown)


src/Init/Data/List/Lemmas.lean
```diff
@@ -2554,4 +2554,17 @@ theorem all_eq_not_any_not (l : List α) (p : α → Bool) : l.all p = !l.any (!
     (l.insert a).all f = (f a && l.all f) := by
   simp [all_eq]
 
+def foldlRecOn {C : β → Sort _} (l : List α) (op : β → α → β) (b : β) (hb : C b)
```
eric-wieser, line 2557:
```suggestion
@[elab_as_elim]
def foldlRecOn {C : β → Sort _} (l : List α) (op : β → α → β) (b : β) (hb : C b)
```
perhaps `C` should also be renamed to `motive`

semorrison: Err, doesn't work.

semorrison:
@hargoniX, is this okay?

hargoniX:
Yes this works with LeanSAT! Feel free to merge.

src/Init/Data/List/Lemmas.lean
```diff
@@ -745,6 +745,58 @@ theorem foldr_hom (f : β₁ → β₂) (g₁ : α → β₁ → β₁) (g₂ :
...
 
+/--
+Prove a proposition about the result of `List.foldl`,
+by proving it for the initial data,
+and the implication that the operation applied to any element of the list preserves the property.
+
+The motive `C` can take values in `Sort _`, so this may be used to construct data,
+as well as to prove propositions.
+-/
+def foldlRecOn {C : β → Sort _} : ∀ (l : List α) (op : β → α → β) (b : β) (_ : C b)
```
eric-wieser, line 756: Repeating my comment from the other thread (sorry about the `elab_as_elim` not working), isn't the convention in lean4 to use `motive` and not `C`?

semorrison: I've never been happy with `motive`, as most new users find it more confusing than a generic name, but I guess that's a change to consider separately. Done.

eric-wieser: Thanks!

PR #5040, opened 2024-08-14
perf: mark entire reported info tree as persistent

Kha:
As we can definitely not free it until .ilean generation at the very end


Kha:
!bench

PR #5041, opened 2024-08-14
feat: mutual recursion: allow common prefix up to alpha-equivalence

nomeata:
@arthur-adjedj was very confused when a mutually recursive definition
didn't work as expected, and the reason was that he used different names
for the fixed parameters.

It seems plausible to simply allow that and calculate the fixed-prefix
up to alpha renaming.

It does mean, though, that, for example, termination proof goals will
mention the names as used by the first function. But probably better
than simply failing. And we could even fix that later (by passing down the
actual names, and renmaing the variables in the context of the mvar,
depending on the “current function”) should it bother our users.

### DIFF
--- a/tests/lean/run/recCommonPrefixAlpha.lean
+++ b/tests/lean/run/recCommonPrefixAlpha.lean
@@ -15,16 +15,6 @@ end Structural
 
 namespace WF
 
--- NB: The proof goal for `f2` mentions `α`, not `β`. Could be fixed in theory if we really care
-
-/--
-info: α : Type
-head✝ : α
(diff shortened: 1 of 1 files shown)


arthur-adjedj:
To give some details: I encountered the issue when looking at code generated by #3160. The way in which nested deriving is implemented there does not always lead to having the exact same variable names between each definition in a mutual block. I could work around this if needed, but I think it could be desirable for common-prefixes to be determined up to alpha-equivalence.

nomeata:
@arthur-adjedj, this got green light.

PR #5054, opened 2024-08-15
feat: explain `reduce` steps in `trace.Debug.Meta.Tactic.simp`

nomeata:


nomeata:
I found this helpful to understand what’s going on, probably worth having in general.

Issue #3983, opened 2024-04-24
RFC: consistent (fine-grained?) equational lemmas


nomeata:
In this issue I’d like to flesh out and describe our plans for how Lean should generate equational lemmas.

## Status quo

The status quo is unsatisfactory in a few ways, at least according to my current understanding:

* The logic for creating the equational lemmas (in `Lean.Elab.Eqns.mkEqnTypes`) is basically “split until it holds by `rfl`, and only near recursive calls”. This is somewhat hard to predict for users, and may (potentially, not sure) lead to different results depending on how the recursion is implemented.

* Because it splits as little as possible, we may get equational lemmas that have “big” case expressions on the right. When they are used by the simplifier to reduce applications to constructors, the term will become large, only to be made smaller again because of the case-of-constructor. More fine-grained equational lemmas would simplify more efficiently.

* The equational lemmas do not necessarily correspond to the cases of the functional induction theorem, violating the principle of least surprise.

* Equational lemmas do not really exist for non-recursive functions (they are always just the unfolding lemma), but would sometimes be useful there as well. Especially now that we have “lazyily defined theorems” by way of reserved names, 

* Sometimes the equational lemmas are suitable for `dsimp`, sometimes they are not.

* Right now, for structural recursion, first the equational lemmas are generated, and then the unfolding lemma, while for well-founded recursion, first the unfolding lemma is derived from `fix_eq`, and then the equational lemmas follow.

### Splitting heuristics

We want uniform equational lemams, so the same heuristics is used for non-recursive, structurally recursive and well-founded recursive lemmas. In particular

* We don’t use `rfl` to stop early.
* We split matches even if there is no recursive call on the right hand side.

This should split more than before, which should imply that where previously an equational lemma was by `rfl`, it will still be by `rfl`, and where possible, we do use `rfl` in the proof to make them `dsimp` lemmas. (TODO: not quite true: there are some `matches` where splitting them introduces assumptions, and breaks the `rfl`-property. Should we not split those?)

This splitting does not necessarily yield consistency with the `foo.induct` lemma. This should be fine as long as the induction cases are more specialized than the equations, i.e. in each induction case there is one equational lemma that will apply.

### Simp API

What does it mean to write `simp [f]`?

Ideally, the user can understand the semantics of `simp [f]` as if `f` was expanded to a list of “normal” theorems, and nothing else behaves differently.

[...]

PR #5055, opened 2024-08-15
feat: unify equational theorems between wf and structural recursion

nomeata:
by removing the `tryRefl` variation between the two.

Part of #3983


nomeata:
This was easier than expected.

PR #5059, opened 2024-08-15
feat: create CACHEDIR.TAG file in .lake directory

### LINKED ISSUE #4998: RFC: Exclude `.lake` directory from backups with `CACHEDIR.TAG`
### Proposal

The `.lake/` directory contains remote sources and build artifacts for a project. Hence its contents can be automatically recreated from the rest of the project source. Such directories should be marked with a [`CACHEDIR.TAG`](https://bford.info/cachedir/) file. This allows automatic backup tools (e.g. Restic, Borg and GNU tar) to ignore the contents of the directory when instructed (e.g. `--exclude-caches` flag on `tar`).

The only effect on users should be that their automatic backups may get smaller as the contents of `.lake/` directory get excluded. In the rare case that files should be recovered from backups, the user needs to recompile their Lean projects. I would argue that this is a relatively minor inconvenience, and if this is a problem, the user should not have excluded caches from backups in the first place.

Other programming languages are already doing this. For example, Rust is excluding both [`target/`](https://github.com/rust-lang/cargo/pull/8378) and [`~/.cargo/{git,registry}/`](https://github.com/rust-lang/cargo/pull/10553) directories.

### Community Feedback

The feature was first proposed in [the Lean Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Exclude.20.2Elake.20directory.20from.20backups.20with.20CACHEDIR.2ETAG/near/456856463). It got green light from @semorrison and @tydeu.

### Impact

henkkuli:
Create CACHEDIR.TAG file in .lake directory when it is first created. The file does not get recreated if removed if .lake directory is not removed, too. While technically against the spec, this is how Rust does it, too. Moreover, this simplifies the implementation a lot.

This change also introduces a test to check that the file gets cretead but not recreated. This test needs to change e.g. if we decide that CACHEDIR.TAG should always be recreated.

---

Closes #4998

### DIFF
--- a/src/lake/Lake/Load/Lean/Elab.lean
+++ b/src/lake/Lake/Load/Lean/Elab.lean
@@ -158,6 +158,15 @@ structure ConfigTrace where
   deriving ToJson, FromJson
 
 /--
+Define the content of the CACHEDIR.TAG file according to the specification
+http://www.brynosaurus.com/cachedir/
+-/
+def cachedirTagContent : String := "Signature: 8a477f597d28d172789f06886806bc55
+# This file is a cache directory tag created by lake.
+# For information about cache directory tags, see:
+#\thttp://www.brynosaurus.com/cachedir/"
+
+/--
 Import the `.olean` for the configuration file if `reconfigure` is not set and
 an up-to-date one exists (i.e., one with matching configuration and on the same
 toolchain). Otherwise, elaborate the configuration and save it to the `.olean`.
@@ -259,6 +268,8 @@ def importConfigFile (cfg : LoadConfig) : LogIO Environment := do
     validateTrace <| ← IO.FS.Handle.mk traceFile .read
   else
     IO.FS.createDirAll cfg.lakeDir
+    -- Create CACHEDIR.TAG file only when .lake directory is first created
+    IO.FS.writeFile (cfg.lakeDir / "CACHEDIR.TAG") cachedirTagContent
     match (← IO.FS.Handle.mk traceFile .writeNew |>.toBaseIO) with
     | .ok h =>
       h.lock; elabConfig h cfg.lakeOpts
--- a/src/lake/tests/cachedirtag/test.sh
+++ b/src/lake/tests/cachedirtag/test.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env bash
+
+# Test that CACHEDIR.TAG file is correctly created in .lake directory.
+#
+# Removing CACHEDIR.TAG without removing .lake directory and then rebuilding
+# does not currently recreate CACHEDIR.TAG. This is an artifact of the current
+# implementation and is technically against the spec:
+#
+# > The application should also regenerate the cache directory tag if it
+# > disappears --
+
+set -euxo pipefail
+
+rm -rf hello
+
+unamestr=`uname`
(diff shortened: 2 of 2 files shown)


henkkuli:
Marking this as non-draft as I noticed that @tydeu has initially approved the idea on Zulip.

src/lake/Lake/Load/Lean/Elab.lean
```diff
@@ -259,6 +268,8 @@ def importConfigFile (cfg : LoadConfig) : LogIO Environment := do
     validateTrace <| ← IO.FS.Handle.mk traceFile .read
   else
     IO.FS.createDirAll cfg.lakeDir
+    -- Create CACHEDIR.TAG file only when .lake directory is first created
+    IO.FS.writeFile (cfg.lakeDir / "CACHEDIR.TAG") cachedirTagContent
```
tydeu, line 272: This is not really what this this does. If a user deletes the `CACHEDIR.TAG` (because they wish the directory to be cached) and then, unrelatedly, delete the configuration trace (e.g., due to an incomplete configuration), this will regenerate the `CACHEDIR.TAG` (which would be quite surprising behavior).

Now that creating the lake directory is something worth doing independently, I suggest refactoring out the `cfg.lakeDir` creation here to the top of this function (outside all the locking). That is, just add something like this to the top and remove it from here:

```lean
if !(<- cfg.lakeDir.pathExists) then
  IO.FS.createDirAll cfg.lakeDir
  -- Create CACHEDIR.TAG file only when .lake directory is first created
  IO.FS.writeFile (cfg.lakeDir / "CACHEDIR.TAG") cachedirTagContent
```

src/lake/tests/cachedirtag/test.sh
```diff
@@ -0,0 +1,52 @@
...
+    test -f "$1"
+    # According to spec, the first 43 bytes are fixed: https://bford.info/cachedir/
+    test "`head -c43 "$1"`" = "Signature: 8a477f597d28d172789f06886806bc55"
+}
+
+LAKE=${LAKE:-../../.lake/build/bin/lake}
+
+$LAKE new hello
+test ! -e hello/.lake
+# Building craetes CACHEDIR.TAG
```
tydeu, line 33:
```suggestion
# Building creates CACHEDIR.TAG
```

src/lake/tests/cachedirtag/test.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+
+# Test that CACHEDIR.TAG file is correctly created in .lake directory.
+#
+# Removing CACHEDIR.TAG without removing .lake directory and then rebuilding
+# does not currently recreate CACHEDIR.TAG. This is an artifact of the current
+# implementation and is technically against the spec:
+#
+# > The application should also regenerate the cache directory tag if it
+# > disappears --
```
tydeu, line 10: This is an odd spec requirement that does not seem desirable. A user may wish to remove the tag and cache the directory and Lake should not override that.

tydeu (changes requested):
Looks pretty good to me! I have one major suggested change, but, nonetheless, great job! I really like the detailed test. 😄

kim-em:
@henkkuli, would you be able to address @tydeu's review comments above? It would be nice to have this in.

PR #5062, opened 2024-08-15
fix: Do not overwrite existing signal handlers

eric-wieser:
Such handlers can come from address sanitizers and similar. When combined with #4971, this forward-ports https://github.com/rust-lang/rust/commit/676b9bc477dfe58971b7df9df4e3a053bb187dee / rust-lang/rust#69685

### DIFF
--- a/src/runtime/stack_overflow.cpp
+++ b/src/runtime/stack_overflow.cpp
@@ -106,7 +106,7 @@ void initialize_stack_overflow() {
     for (auto signum : {SIGSEGV, SIGBUS}) {
         struct sigaction action;
         memset(&action, 0, sizeof(struct sigaction));
-        sigaction(SIGSEGV, nullptr, &action);
+        sigaction(signum, nullptr, &action);
         // Configure our signal handler if one is not already set.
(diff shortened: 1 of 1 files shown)


src/runtime/stack_overflow.cpp
```diff
@@ -102,12 +103,17 @@ void initialize_stack_overflow() {
...
-    struct sigaction action;
-    memset(&action, 0, sizeof(struct sigaction));
-    action.sa_flags = SA_SIGINFO | SA_ONSTACK;
-    action.sa_sigaction = segv_handler;
-    sigaction(SIGSEGV, &action, nullptr);
-    sigaction(SIGBUS, &action, nullptr);
+    for (auto signum : {SIGSEGV, SIGBUS}) {
+        struct sigaction action;
+        memset(&action, 0, sizeof(struct sigaction));
+        sigaction(SIGSEGV, nullptr, &action);
```
Kha, line 109:
```suggestion
        sigaction(signum, nullptr, &action);
```
?

eric-wieser: Sorry, that was clumsy!

PR #5065, opened 2024-08-16
feat: upstream and update `#where` command

kmill:
This command comes from Lean 3, which I had previously ported and contributed to Batteries (née Std). In this new version, `#where` produces actual command Syntax for all features of a top-level scope (rather than splicing together strings), and it also now reports included variables.


kmill:
@semorrison Here's an upstreamed `#where`!

PR #5069, opened 2024-08-16
feat: upstream List.Perm

kim-em:


semorrison:
@TwoFX, could you please sanity check what I've done in HashMap? I've ripped out the internal `Perm` implementation, replacing it with the new "official" one. (Both were in scope; the alternative would have been lots of disambiguation; if you'd prefer that either tell me to do it, or do it yourself! :-)

TwoFX (approved):

PR #5074, opened 2024-08-16
feat: import LeanSAT LRAT

hargoniX:
This PR imports LeanSAT's LRAT module as step 4/~6 (step 7 could go after I did some refactorings to import this) of the LeanSAT upstreaming. It is the last large component, after this only the LRAT parser and the reflection tactic that hooks everything up to the meta level remains. In particular it is the last component that contains notable proofs, yay!

Again a few remarks:
1. Why is this not in `Std`? I'm not quite sure whether it should be there. At the current level of code/proof quality we can certainly not import the checker itself into `Std` but maybe having the data type as well as the trimming algorithm there might be of interested? I'm hoping that as we refactor the checker in the future its quality will be high enough to be also put into `Std`. At this point we would have a full AIG -> CNF -> LRAT verification pipeline in `Std` for everyone to use. One additional blocker in this is that we cannot provide the parsers for the format in `Std` as of today because `Parsec` is still in `Lean` so that would also have to change.
2. There do exist two abstraction levels to make sure we can swap out the LRAT implementation at any time:
    - The public interface is just all files in the top level `LRAT` directory. It basically only contains the LRAT format itself, the checker + soundness proof and the trimming algorithm. As long as we don't need to change their API (which we shouldn't have to I think) we can always swap out the entire `Internal` directory without breaking anything else in LeanSAT.
    - The `Internal` module itself contains another layer of abstraction in the form of the `Formula` class. This allows us to swap out the most complex component in `Internal` as well, without having to touch any of the infrastructure that is built around it either.
3. I mostly performed stylistic cleanups on the `Internal` module. In my experience over upgrading to many nightlies during the course of LeanSAT development, I have gotten these proofs cleaned up to the point, where they only break if we change the `List` or `Array` proof API significantly. Given that we are currently in the process of stabilizing it I'm hoping that these proofs do not have to be touched anymore unless we do something crazy. All of the custom theory that the LRAT component developed around various basic data types has been upstreamed into Lean over the course of various other PRs.
4. If there are some simple tricks that we can pull off to increase the code / proof quality in `Internal` and in particular `Internal.Formula` (this module is not for the light-hearted Lean reviewer) I'm all for it. Otherwise the best course of action to provide LeanSAT to our users soon would probably be to merge it as is and do a cut + rewrite at one of the two interface points described above.

### DIFF
--- a/src/Lean/Elab/Tactic/BVDecide/LRAT/Checker.lean
+++ b/src/Lean/Elab/Tactic/BVDecide/LRAT/Checker.lean
@@ -28,27 +28,20 @@ def check (lratProof : Array IntAction) (cnf : CNF Nat) : Bool :=
   let internalFormula := CNF.convertLRAT cnf
   let lratProof := lratProof.toList
   let lratProof := lratProof.map (intActionToDefaultClauseAction _)
-  let lratProof : List { act // WellFormedAction act } :=
+  let lratProof :=
     lratProof.filterMap
       (fun actOpt =>
         match actOpt with
         | none => none
-        | some (LRAT.Action.addEmpty id rupHints) =>
-          some ⟨LRAT.Action.addEmpty id rupHints, by simp only [WellFormedAction]⟩
-        | some (LRAT.Action.addRup id c rupHints) =>
-          some ⟨LRAT.Action.addRup id c rupHints, by simp only [WellFormedAction]⟩
-        | some (LRAT.Action.del ids) =>
-          some ⟨LRAT.Action.del ids, by simp only [WellFormedAction]⟩
+        | some (LRAT.Action.addEmpty id rupHints) => some (LRAT.Action.addEmpty id rupHints)
+        | some (LRAT.Action.addRup id c rupHints) => some (LRAT.Action.addRup id c rupHints)
+        | some (LRAT.Action.del ids) => some (LRAT.Action.del ids)
         | some (LRAT.Action.addRat id c pivot rupHints ratHints) =>
-          if h : pivot ∈ Clause.toList c then
-            some ⟨
-              LRAT.Action.addRat id c pivot rupHints ratHints,
-              by simp [WellFormedAction, Clause.limplies_iff_mem, h]
-            ⟩
+          if pivot ∈ Clause.toList c then
+            some (LRAT.Action.addRat id c pivot rupHints ratHints)
           else
             none
       )
-  let lratProof := lratProof.map Subtype.val
   let checkerResult := lratChecker internalFormula lratProof
   checkerResult = .success
 
@@ -69,10 +62,20 @@ theorem check_sound (lratProof : Array IntAction) (cnf : CNF Nat) :
       _
       (by
         intro action h
-        simp only [List.mem_map, List.mem_filterMap] at h
+        simp only [Array.toList_eq, List.filterMap_map, List.mem_filterMap, Function.comp_apply] at h
         rcases h with ⟨WellFormedActions, _, h2⟩
-        rw [← h2]
-        exact WellFormedActions.property)
+        split at h2
+        . contradiction
+        . simp only [Option.some.injEq] at h2
+          simp [← h2, WellFormedAction]
+        . simp only [Option.some.injEq] at h2
+          simp [← h2, WellFormedAction]
+        . simp only [Option.some.injEq] at h2
+          simp [← h2, WellFormedAction]
+        . simp only [ite_some_none_eq_some] at h2
+          rcases h2 with ⟨hleft, hright⟩
+          simp [WellFormedAction, hleft, ← hright, Clause.limplies_iff_mem]
+      )
       h1
   apply CNF.unsat_of_convertLRAT_unsat
   assumption


src/Lean/Elab/Tactic/BVDecide/LRAT/Internal/Convert.lean
```diff
@@ -0,0 +1,171 @@
...
+    exact CNF.Clause.mem_lrat_of_mem clause hlit1 h
+  · simp_all
+
+/--
+Convert a `CNF Nat` with a certain maximum variable number into the `DefaultFormula`
+format for usage with `bv_decide`'s `LRAT.Internal`.
+
+Notably this:
+1. Increments all variables as DIMACS variables start at 1 instead of 0
+2. Adds a leading `none` clause. This clause *must* be persistet as the LRAT cecker wants to have
```
semorrison, line 115: s/persistet/present/?

s/cecker/checker/

semorrison:
!bench

semorrison:
(I'm not going to complain whatever `!bench` says, just want to know how much extra build time we're in for!)

src/Lean/Elab/Tactic/BVDecide/LRAT/Internal/LRATCheckerSound.lean
```diff
@@ -0,0 +1,153 @@
...
+import Lean.Elab.Tactic.BVDecide.LRAT.Internal.Actions
+
+namespace Lean.Elab.Tactic.BVDecide
+namespace LRAT
+namespace Internal
+
+open LRAT Result Formula Clause Std Sat
+
+theorem addEmptyCaseSound [DecidableEq α] [Clause α β] [Entails α σ] [Formula α β σ] (f : σ)
+    (f_readyForRupAdd : ReadyForRupAdd f) (rupHints: Array Nat)
```
TwoFX, line 18:
```suggestion
    (f_readyForRupAdd : ReadyForRupAdd f) (rupHints : Array Nat)
```

src/Lean/Elab/Tactic/BVDecide/LRAT/Internal/Formula/RupAddSound.lean
```diff
@@ -0,0 +1,834 @@
...
+  simp only [formulaEntails_def, List.all_eq_true, decide_eq_true_eq]
+  intro c' c'_in_fc
+  rw [insert_iff] at c'_in_fc
+  rcases c'_in_fc with c'_eq_c | c'_in_f
+  · rw [c'_eq_c]
+    exact insertRup_entails_hsat f f_readyForRupAdd c p pf h
+  · simp only [formulaEntails_def, List.all_eq_true, decide_eq_true_eq] at pf
+    exact pf c' c'_in_f
+
+theorem insertRupUnits_preserves_AssignmentsInvariant {n : Nat} (f : DefaultFormula n) (f_readyForRupAdd : ReadyForRupAdd f)
```
TwoFX, line 236:
```suggestion
theorem insertRupUnits_preserves_assignmentsInvariant {n : Nat} (f : DefaultFormula n) (f_readyForRupAdd : ReadyForRupAdd f)
```
(or, if you want to go by the naming convention, `assignmentInvariant_insertRupUnits`).

src/Lean/Elab/Tactic/BVDecide/LRAT/Internal/Formula/RupAddResult.lean
```diff
@@ -0,0 +1,1353 @@
...
+    constructor
+    · rfl
+    · intro l l_in_nil
+      simp only [List.find?, List.not_mem_nil] at l_in_nil
+  have h_inductive (i : Fin rupHints.size) (acc : Array Assignment × CNF.Clause (PosFin n) × Bool × Bool)
+    (ih : motive i.1 acc) := confirmRupHint_preserves_invariant f f_assignments_size rupHints i acc ih
+  rcases Array.foldl_induction motive h_base h_inductive with ⟨_, h⟩
+  exact h
+
+theorem derivedLits_nodup {n : Nat} (f : DefaultFormula n)
```
TwoFX, line 1109:
```suggestion
theorem nodup_derivedLits {n : Nat} (f : DefaultFormula n)
```

src/Lean/Elab/Tactic/BVDecide/LRAT/Internal/Formula/RupAddResult.lean
```diff
@@ -0,0 +1,1353 @@
...
+  simp only [performRupCheck]
+  rw [Array.foldl_eq_foldl_data]
+  have hb : (f.assignments, ([] : CNF.Clause (PosFin n)), false, false).1.size = f.assignments.size := rfl
+  have hl (acc : Array Assignment × CNF.Clause (PosFin n) × Bool × Bool) (hsize : acc.1.size = f.assignments.size)
+    (id : Nat) (_ : id ∈ rupHints.data) : (confirmRupHint f.clauses acc id).1.size = f.assignments.size := by
+    have h := confirmRupHint_preserves_assignments_size f.clauses acc.1 acc.2.1 acc.2.2.1 acc.2.2.2 id
+    rw [h, hsize]
+  exact List.foldlRecOn rupHints.data (confirmRupHint f.clauses) (f.assignments, [], false, false) hb hl
+
+def derivedLits_invariant {n : Nat} (f : DefaultFormula n)
```
TwoFX, line 757:
```suggestion
def DerivedLitsInvariant {n : Nat} (f : DefaultFormula n)
```

src/Lean/Elab/Tactic/BVDecide/LRAT/Internal/Formula/RupAddResult.lean
```diff
@@ -0,0 +1,1353 @@
...
+theorem clearUnit_foldl_preserves_size {α : Type u} (assignments : Array Assignment)
+    (f : Array Assignment → α → Array Assignment) (f_preserves_size :
+    ∀ arr : Array Assignment, ∀ a : α, (f arr a).size = arr.size) (l : List α) :
+  Array.size (List.foldl f assignments l) = Array.size assignments := by
+  have hb : assignments.size = assignments.size := rfl
+  have hl (assignments' : Array Assignment) (hsize : assignments'.size = assignments.size) (a : α) (_ : a ∈ l) :
+    (f assignments' a).size = assignments.size := by rw [f_preserves_size assignments' a, hsize]
+  exact List.foldlRecOn l f assignments hb hl
+
+def clear_insert_induction_motive {n : Nat} (f : DefaultFormula n) (assignments_size : f.assignments.size = n)
```
TwoFX, line 484:
```suggestion
def ClearInsertInductionMotive {n : Nat} (f : DefaultFormula n) (assignments_size : f.assignments.size = n)
```

src/Lean/Elab/Tactic/BVDecide/LRAT/Internal/Formula/RatAddSound.lean
```diff
@@ -0,0 +1,659 @@
...
+    split at ratAddSuccess
+    · simp at ratAddSuccess
+    · split at ratAddSuccess
+      · simp at ratAddSuccess
+      · split at ratAddSuccess
+        · simp at ratAddSuccess
+        · split at ratAddSuccess
+          · simp at ratAddSuccess
+          · next performRatCheck_fold_success =>
+            simp only [Bool.not_eq_false] at performRatCheck_fold_success
```
TwoFX, line 649: The soundness proof for RAT add is shaped like a RAT tail.

src/Lean/Elab/Tactic/BVDecide/LRAT/Internal/Clause.lean
```diff
@@ -0,0 +1,418 @@
...
+
+instance [Clause α β] : Inhabited β where
+  default := empty
+
+end Clause
+
+/--
+The `DefaultClause` structure is primarily a list of literals. The additional field `nodupkey` is
+included to ensure that `not_tautology` is provable (which is needed to prove `insertRup_entails_hsat`
+and `insertRat_entails_hsat` in `LRAT.Formula.RupAddSound.lean` and `LRAT.Formula.RatAddSound.lean`).
```
TwoFX, line 84: I think you want either `LRAT/Formula/Internal/RupAddSound.lean` or `LRAT.Formula.Internal.RupAddSound` here?

src/Lean/Elab/Tactic/BVDecide/LRAT/Internal/Clause.lean
```diff
@@ -0,0 +1,418 @@
...
+
+instance : ToString (DefaultClause n) where
+  toString c := s!"{c.clause.reverse}"
+
+namespace DefaultClause
+
+def toList (c : DefaultClause n) : CNF.Clause (PosFin n) := c.clause
+
+theorem not_tautology (c : DefaultClause n) (l : Literal (PosFin n)) :
+    ¬ l ∈ toList c ∨ ¬Literal.negate l ∈ toList c := by
```
TwoFX, line 110: Inconsistent spacing around the not here, I think the usual style is to omit the space.

src/Lean/Elab/Tactic/BVDecide/LRAT/Internal/Clause.lean
```diff
@@ -0,0 +1,418 @@
...
+be more than one.
+
+In principle, one could combine `nodupkey` and `nodup` to instead have one additional field that
+stipulates that `∀ l1 : PosFin numVarsSucc, ∀ l2 : PosFin numVarsSucc, l1.1 ≠ l2.1`. This would work
+just as well, and the only reason that `DefaultClause` is structured in this manner is that the
+`nodup` field was only included in a later stage of the verification process when it became clear
+that it was needed.
+-/
+@[ext] structure DefaultClause (numVarsSucc : Nat) where
+  clause : CNF.Clause (PosFin numVarsSucc)
```
TwoFX, line 97: Just out of curiosity: have you considered using `Array` instead of `List` here? Do you think it would impact performance at all?

src/Lean/Elab/Tactic/BVDecide/LRAT/Internal/CNF.lean
```diff
@@ -0,0 +1,134 @@
...
+end Clause
+
+namespace Formula
+
+theorem sat_iff_forall [Clause α β] [Entails α σ] [Formula α β σ] (p : α → Bool) (f : σ) :
+    p ⊨ f ↔ ∀ c : β, c ∈ toList f → p ⊨ c := by
+  simp only [(· ⊨ ·), formulaEntails_def p f]
+  simp only [List.all_eq_true, decide_eq_true_eq]
+
+theorem limplies_of_insert [Clause α β] [Entails α σ] [Formula α β σ] {c : β} {f : σ} :
```
TwoFX, line 110:
```suggestion
theorem limplies_insert [Clause α β] [Entails α σ] [Formula α β σ] {c : β} {f : σ} :
```

src/Lean/Elab/Tactic/BVDecide/LRAT/Internal/Actions.lean
```diff
@@ -0,0 +1,97 @@
...
+Since `IntAction` is a convenient parsing target and `DefaultClauseAction` is a useful Action type
+for working with default clauses, an expected workflow pattern is to parse an external LRAT proof
+into a list of `IntAction`s, and then use this function to convert that list of `IntAction`s to
+`DefaultClauseAction`s.
+
+This function returns an `Option` type so that `none` can be returned if converting from the
+`IntAction` to `DefaultClauseAction` fails. This can occur if any of the literals in the `IntAction`
+are 0 or ≥ n.
+-/
+def intActionToDefaultClauseAction (n : Nat) : IntAction → Option (DefaultClauseAction n)
```
TwoFX, line 64: Optional suggestion: should this be named `IntAction.toDefaultClauseAction`?

src/Lean/Elab/Tactic/BVDecide/LRAT/Actions.lean
```diff
@@ -0,0 +1,44 @@
...
+used by the checker as input and should be considered the parsing target for LRAT proofs.
+-/
+
+namespace Lean.Elab.Tactic.BVDecide
+namespace LRAT
+
+open Std.Sat
+
+/-- `β` is for the type of a clause, `α` is for the type of variables -/
+inductive Action (β : (Type v)) (α : (Type w))
```
TwoFX, line 21: Did using `u` for universes go out of fashion?
```suggestion
inductive Action (β : Type v) (α : Type w)
```

src/Lean/Elab/Tactic/BVDecide/LRAT/Actions.lean
```diff
@@ -0,0 +1,44 @@
+/-
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Josh Clune
+-/
+prelude
+import Std.Sat.CNF
+
+/-!
+This module contains the definition of the LRAT format as a type `Action`, that is polymorphic over
```
TwoFX, line 10: Since this file is where the LRAT format is defined I think it makes sense to link to some resource on the topic here for people who are curious. Something like https://www.cs.utexas.edu/~marijn/publications/lrat.pdf or whatever exposition you think is good and matches what we're doing here.

TwoFX (commented):
Probably makes sense to make one quick pass over the naming; I left some examples. Apart from that I guess there's not much point in spending too much time on the internal stuff as long as the performance of the `LRAT.check` function is good.

src/Lean/Elab/Tactic/BVDecide/LRAT/Internal/Clause.lean
hargoniX: I did think about it and I'm not quite sure. There are operations that delete literals from a clause, in this situation we would have to copy the entire clause instead of just cheaply deleting one node. So it would definitely need some evaluation as to which variant is better.

TwoFX: If we're happy with the performance at the moment let's just leave it as it is, and if we need to revisit performance later we can keep this in mind as one thing to try.

src/Lean/Elab/Tactic/BVDecide/LRAT/Checker.lean
```diff
@@ -0,0 +1,81 @@
...
+      _
+      (by
+        intro action h
+        simp only [List.mem_map, List.mem_filterMap] at h
+        rcases h with ⟨WellFormedActions, _, h2⟩
+        rw [← h2]
+        exact WellFormedActions.property)
+      h1
+  apply CNF.unsat_of_convertLRAT_unsat
+  assumption
```
TwoFX, line 78: It appears to me that you can make this slightly more efficient by folding all of the list traversals into one, for example like this:
```suggestion
def check (lratProof : Array IntAction) (cnf : CNF Nat) : Bool :=
  let internalFormula := CNF.convertLRAT cnf
  let lratProof := lratProof.toList
  let lratProof : List (Action (DefaultClause (cnf.numLiterals + 1)) (PosFin (cnf.numLiterals + 1))) :=
    lratProof.filterMap
      (fun actOpt => Option.map (Subtype.val (p := WellFormedAction)) <|
        match intActionToDefaultClauseAction _ actOpt with
        | none => none
        | some (LRAT.Action.addEmpty id rupHints) =>
          some ⟨LRAT.Action.addEmpty id rupHints, by simp only [WellFormedAction]⟩
        | some (LRAT.Action.addRup id c rupHints) =>
          some ⟨LRAT.Action.addRup id c rupHints, by simp only [WellFormedAction]⟩
        | some (LRAT.Action.del ids) =>
          some ⟨LRAT.Action.del ids, by simp only [WellFormedAction]⟩
        | some (LRAT.Action.addRat id c pivot rupHints ratHints) =>
          if h : pivot ∈ Clause.toList c then
            some ⟨
              LRAT.Action.addRat id c pivot rupHints ratHints,
              by simp [WellFormedAction, Clause.limplies_iff_mem, h]
            ⟩
          else
            none
      )
  -- let lratProof := lratProof.map Subtype.val
  let checkerResult := lratChecker internalFormula lratProof
  checkerResult = .success

open Lean.Elab.Tactic.BVDecide.LRAT.Internal in
/--
If the `check` functions succeeds on `lratProof` and `cnf` then the `cnf` is unsatisfiable.
-/
theorem check_sound (lratProof : Array IntAction) (cnf : CNF Nat) :
    check lratProof cnf → cnf.Unsat := by
  intro h1
  unfold check at h1
  simp only [decide_eq_true_eq] at h1
  have h2 :=
    lratCheckerSound
      _
      (by apply CNF.convertLRAT_readyForRupAdd)
      (by apply CNF.convertLRAT_readyForRatAdd)
      _
      (by
        intro action h
        simp only [List.mem_filterMap] at h
        rcases h with ⟨intAct, -, h3⟩
        have help : ∀ {a : { act : Action (DefaultClause (cnf.numLiterals + 1)) (PosFin (cnf.numLiterals + 1)) // WellFormedAction act }},
          WellFormedAction a.1 := fun {a} => a.2
        split at h3
        all_goals try split at h3
        all_goals try (simp at h3; done)
        all_goals
        · rw [Option.map_some', Option.some_inj] at h3
          rw [← h3]
          exact help)
      h1
  apply CNF.unsat_of_convertLRAT_unsat
  assumption
```
Probably it would also be possible to move the proofs into the theorem entriely.

TwoFX (approved):

PR #5081, opened 2024-08-18
feat: add Bitvec.ofInt_ofNat

tobiasgrosser:
We use `no_index` to work around special-handling of `OfNat.ofNat` in `DiscrTree`, which has been reported as an issue in https://github.com/leanprover/lean4/issues/2867 and is currently in the process of being fixed in https://github.com/leanprover/lean4/pull/3684. As the potential fix seems non-trivial and might need some time to arrive in-tree, we meanwhile add the `no_index` keyword to the problematic subterm.


tobiasgrosser:
I am currently debugging why `no_index` is needed.

tobiasgrosser:
I have two versions: one fails, and the other one works. Advice on what is going on would be appreciated. I would be OK with the `no_index` version but would prefer to add one sentence explaining why it is needed.

```lean
@[simp] theorem ofInt_ofNat (w n : Nat) :
    BitVec.ofInt w (no_index (OfNat.ofNat n)) = BitVec.ofNat w n :=
  rfl

@[simp] theorem ofInt_ofNat_no_noindex (w n : Nat) :
    BitVec.ofInt w (OfNat.ofNat n) = BitVec.ofNat w n :=
  rfl
```

```lean
example : BitVec.ofInt w 1 = 1#w + 0#w := by
  simp only [ofInt_ofNat] -- this one works
  simp only [BitVec.add_zero]

example : BitVec.ofInt w 1 = 1#w + 0#w := by
  simp only [ofInt_ofNat_no_noindex] -- this one fails
  simp only [BitVec.add_zero]
```

tobiasgrosser:
@urkud kindly provide the relevant context to understand the `no_index` issue. I am now convinced that `no_index` is needed at the moment and put the relevant information into the PR summary at the top. This PR is now ready to review.

tobiasgrosser:
awaiting-review

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -303,6 +303,9 @@ theorem toInt_ofNat {n : Nat} (x : Nat) :
 @[simp] theorem ofInt_natCast (w n : Nat) :
   BitVec.ofInt w (n : Int) = BitVec.ofNat w n := rfl
 
+@[simp] theorem ofInt_ofNat (w n : Nat) :
+  BitVec.ofInt w (no_index (OfNat.ofNat n)) = BitVec.ofNat w n := rfl
```
eric-wieser, line 307:
```suggestion
  BitVec.ofInt w (no_index (OfNat.ofNat n)) = BitVec.ofNat w (OfNat.ofNat n) := rfl
```
otherwise this can leave an unprotected raw nat literal.

tobiasgrosser: Nice. Thank you for your feedback. This is super helpful.

PR #5087, opened 2024-08-18
doc: mention that `inferType` does not ensure type correctness

kmill:
This also adds links to the implementations of `whnf` and `inferType` to make it easier to navigate this part of the code base.

### DIFF
--- a/src/Lean/Meta/Basic.lean
+++ b/src/Lean/Meta/Basic.lean
@@ -578,12 +578,15 @@ See `Lean.Meta.whnfImp` for the implementation.
 /--
 Returns the inferred type of the given expression. Assumes the expression is type-correct.
 
-The type inference algorithm does not do general type checking,
-which is instead handled by the `Lean.Meta.check` algorithm.
+The type inference algorithm does not do general type checking.
 Type inference only looks at subterms that are necessary for determining an expression's type,
 and as such if `inferType` succeeds it does *not* mean the term is type-correct.
 If an expression is sufficiently ill-formed that it prevents `inferType` from computing a type,
 then it will fail with a type error.
+
+For typechecking during elaboration, see `Lean.Meta.check`.
+(Note that we do not guarantee that the elaborator typechecker is as correct or as efficient as
+the kernel typechecker. The kernel typechecker is invoked when a definition is added to the environment.)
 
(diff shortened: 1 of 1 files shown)


src/Lean/Meta/Basic.lean
```diff
@@ -560,16 +560,31 @@ def useEtaStruct (inductName : Name) : MetaM Bool := do
...
+The result may contain subexpressions that have not been reduced.
+
+See `Lean.Meta.whnfImp` for the implementation.
+-/
 @[extern 6 "lean_whnf"] opaque whnf : Expr → MetaM Expr
-/-- Returns the inferred type of the given expression, or fails if it is not type-correct. -/
+/--
+Returns the inferred type of the given expression.
+
+The type inference algorithm assumes the expression is type-correct for efficiency,
```
hargoniX, line 581: Could you maybe provide an example for a term that it might fail on? This claim probably sounds quite fancy if you are used to the academic notion of "if type inference succeeds the type it returns is right"

kmill: Thanks, I've added some examples of both accidental success and necessary failure.

src/Lean/Meta/Basic.lean
```diff
@@ -560,16 +560,69 @@ def useEtaStruct (inductName : Name) : MetaM Bool := do
...
+See `Lean.Meta.whnfImp` for the implementation.
+-/
 @[extern 6 "lean_whnf"] opaque whnf : Expr → MetaM Expr
-/-- Returns the inferred type of the given expression, or fails if it is not type-correct. -/
+/--
+Returns the inferred type of the given expression. Assumes the expression is type-correct.
+
+The type inference algorithm does not do general type checking,
+which is instead handled by the `Lean.Meta.check` algorithm.
+Type inference only looks at subterms that are necessary for determining an expression's type,
```
leodemoura, line 583: It may be worth mentioning that we do not guarantee that `Lean.Meta.check` is as efficient and complete as the check implemented in the kernel.

kmill: Thanks, I've moved the mention of `Lean.Meta.check` to a new paragraph with a comment about the kernel.

PR #5088, opened 2024-08-19
feat: add `Coe (α → Bool) (α  → Prop)` and `Coe (α → α → Bool) (α → α → Prop)`

kim-em:
I'm not sure if these are a terrible idea. They would be very convenient if they are not problematic...

Let's test and benchmark first and see where we stand.


semorrison:
!bench

semorrison:
This does seem to slow Mathlib down appreciably, unfortunately, so I'll find another way.

PR #5089, opened 2024-08-19
chore: bump olean version

digama0:
This is a belated fix for an issue introduced in #4265, which turned on GMP bignum support on macos (which is an ABI breaking change).


eric-wieser (approved):

zwarich:
It looks like this was bumped in #5144.

PR #5091, opened 2024-08-19
feat: add debug.sorryProps

kim-em:
Just trying something out ... Let's sorry every Prop!

### DIFF
--- a/src/Lean/Elab/Term.lean
+++ b/src/Lean/Elab/Term.lean
@@ -1682,7 +1682,7 @@ Otherwise, it throws the error.
 -/
 def elabTermEnsuringType (stx : Syntax) (expectedType? : Option Expr) (catchExPostpone := true) (implicitLambda := true) (errorMsgHeader? : Option String := none) : TermElabM Expr := do
   if let some expectedType := expectedType? then
-    if !expectedType.hasMVar && (← pure (debug.sorryProps.get (← getOptions)) <&&> isProp expectedType) then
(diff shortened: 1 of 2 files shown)


src/Lean/Elab/Term.lean
```diff
@@ -1675,6 +1681,9 @@ If `errToSorry` is true, then if coercion insertion fails, this function returns
 Otherwise, it throws the error.
 -/
 def elabTermEnsuringType (stx : Syntax) (expectedType? : Option Expr) (catchExPostpone := true) (implicitLambda := true) (errorMsgHeader? : Option String := none) : TermElabM Expr := do
+  if let some expectedType := expectedType? then
+    if (← pure (debug.sorryProps.get (← getOptions))) && !expectedType.hasMVar && (← isProp expectedType) then
+      return ← mkSorry expectedType false
```
kmill, line 1686: For `hasMVar`, there could be an `instantiateMVars` to avoid some false negatives.
```suggestion
  if debug.sorryProps.get (← getOptions) then
    if let some expectedType := expectedType? then
      let expectedType ← instantiateMVars expectedType
      if ← pure !expectedType.hasMVar <&&> isProp expectedType then
        return ← mkSorry expectedType false
```

leodemoura:
@semorrison Lean often postpones the elaboration of sub-terms, type class instances, and similar constructs, resulting in metavariables being scattered throughout terms and their types. This behavior makes it difficult for users to predict when this flag will be effective. Do you see this as a potential issue? What are your thoughts?

PR #5092, opened 2024-08-19
feat: List.mergeSort

kim-em:
Defines `mergeSort`, a naive stable merge sort algorithm, replaces it via a `@[csimp]` lemma with something faster at runtime, and proves the following results:

* `mergeSort_sorted`: `mergeSort` produces a sorted list.
* `mergeSort_perm`: `mergeSort` is a permutation of the input list.
* `mergeSort_of_sorted`: `mergeSort` does not change a sorted list.
* `mergeSort_cons`: proves `mergeSort le (x :: xs) = l₁ ++ x :: l₂` for some `l₁, l₂`
  so that `mergeSort le xs = l₁ ++ l₂`, and no `a ∈ l₁` satisfies `le a x`.
* `mergeSort_stable`: if `c` is a sorted sublist of `l`, then `c` is still a sublist of `mergeSort le l`.

### DIFF
--- a/tests/lean/run/mergeSort.lean
+++ b/tests/lean/run/mergeSort.lean
@@ -1,4 +1,4 @@
-open List
+open List MergeSort Internal
 
 unseal mergeSort merge in
 example : mergeSort (· ≤ ·) [3, 1, 4, 1, 5, 9, 2, 6, 5, 3, 5] = [1, 1, 2, 3, 3, 4, 5, 5, 5, 6, 9] :=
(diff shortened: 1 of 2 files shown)


semorrison:
I am really trying hard to resist the temptation to state the result that "forgetting it is sorted" and "stable merge sort" form a Galois connection between sorted lists and lists, with the `<+` (sublist) partial order.

semorrison:
This PR will need to wait for me to verify that #5088 doesn't have any bad performance implications.

semorrison:
This is expected to break Mathlib as there is a (non-tail-recursive, non-stable) implementation of `mergeSort` there. I'll clean it up after this is merged.

PR #5095, opened 2024-08-19
refactor: in simp's unfold? function, the helper is not recursive

nomeata:
(it never was ever since it was introduced in #2042.)


nomeata:
Probably intentional, TODO: Leave comment.

PR #5113, opened 2024-08-21
chore: move LeanSAT logic to Std

hargoniX:
As discussed for bootstrapping reasons. The only new files here are `Std.Tactic` and `Std.Tactic.BVDecide`. The rest is move + renamespacing.


tobiasgrosser (approved):
This looks good to me.

PR #5119, opened 2024-08-21
fix: `get_elem_tactic_trivial` to not loop in the presence of mvars

### LINKED ISSUE #5061: Exponential behavior in compilation
### Prerequisites

### Description

On the following example, the lean runtime grows exponentially with the recursion depth. With default depth, lean is stuck. 

```lean
def fin_max {n: Nat} (f: Fin n → Nat) : Nat :=
  loop 0 0
where
  loop (i: Nat) (max_so_far: Nat): Nat :=
    if h: i < n then
      loop (i + 1) (max max_so_far (f ⟨i, h⟩))
    else
      max_so_far

set_option profiler true in
set_option maxRecDepth 74 in
def test (a: Array Nat) : Nat := fin_max (fun i => a[i])
```

Here's what I see as the depth increases:

    -----------------------------------------
       depth    | compilation time (in s)
    -----------------------------------------
       68       | 1.35
       70       | 2.68
       72       | 5.43
       74       | 10.8
       76       | 21.8
    -----------------------------------------

### Steps to Reproduce

```
> ~/.elan/toolchains/leanprover--lean4---v4.10.0/bin/lean slow.lean
compilation of test took 10.8s
slow.lean:12:51: error: maximum recursion depth has been reached
use `set_option maxRecDepth <num>` to increase limit
use `set_option diagnostics true` to get diagnostic information
cumulative profiling times:
        attribute application 0.00237ms
        compilation 10.8s
        tactic execution 6.63ms
        type checking 0.417ms
        typeclass inference 0.446ms
```

**Expected behavior:** 

Fail quickly. There's not enough info to infer the type of `i` in `fun i => a[I]`.

**Actual behavior:**

Doesn't terminate (unless one gives a small recursion depth limit).

### Versions

4.10.0

### Additional Information

### Impact

nomeata:
The goal at the crucial step is
```
a : Array Nat
i : Fin ?m.27
⊢ ↑i < a.size
```
and after the `apply Fin.val_lt_of_le;` we have
```
a : Array Nat
i : Fin ?m.27
⊢ ?m.27 ≤ a.size
```
and now `apply Fin.val_lt_of_le` applies again, due to accidential defeq. Adding `with_reducible` helps here.

fixes #5061

### DIFF
--- a/src/Init/GetElem.lean
+++ b/src/Init/GetElem.lean
@@ -184,7 +184,7 @@ instance [GetElem? cont Nat elem dom] [h : LawfulGetElem cont Nat elem dom] :
 @[simp] theorem getElem!_fin [GetElem? Cont Nat Elem Dom] (a : Cont) (i : Fin n) [Inhabited Elem] : a[i]! = a[i.1]! := rfl
 
 macro_rules
-  | `(tactic| get_elem_tactic_trivial) => `(tactic| apply Fin.val_lt_of_le; get_elem_tactic_trivial; done)
+  | `(tactic| get_elem_tactic_trivial) => `(tactic| (with_reducible apply Fin.val_lt_of_le); get_elem_tactic_trivial; done)
 
 end Fin
 
--- a/tests/lean/run/issue5061.lean
+++ b/tests/lean/run/issue5061.lean
@@ -1,0 +1,33 @@
+opaque fin_max {n: Nat} (f: Fin n → Nat) : Nat
+
+set_option maxRecDepth 30 in
+-- NB: Important to not specify the implicit argument of fin_max here
+def test (a: Array Nat) : Nat := @fin_max _ fun i =>
+  let h : i < a.size := by
+    -- this may apply once
+    with_reducible apply Fin.val_lt_of_le
+    fail_if_success with_reducible apply Fin.val_lt_of_le;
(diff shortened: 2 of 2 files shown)


src/Init/GetElem.lean
```diff
@@ -184,7 +184,7 @@ instance [GetElem? cont Nat elem dom] [h : LawfulGetElem cont Nat elem dom] :
 @[simp] theorem getElem!_fin [GetElem? Cont Nat Elem Dom] (a : Cont) (i : Fin n) [Inhabited Elem] : a[i]! = a[i.1]! := rfl
 
 macro_rules
-  | `(tactic| get_elem_tactic_trivial) => `(tactic| apply Fin.val_lt_of_le; get_elem_tactic_trivial; done)
+  | `(tactic| get_elem_tactic_trivial) => `(tactic| (with_reducible apply Fin.val_lt_of_le); get_elem_tactic_trivial; done)
```
Kha, line 187: An explanatory comment seems to be missing here :)

nomeata: Hmm, maybe. Although `with_reducible` itself already shuts loudly “don't apply this lemma unless the goals looks like it”, and brings this in line with similar tactic invocations

PR #5121, opened 2024-08-22
feat: use StateT.run instead of function application

kim-em:
This PR using `StateT.run` rather than the "defeq abuse" of function application. There remain many places where we still use function application for `ReaderT`, but I've updated this in the touched files.

(To really solve this, we would make `StateT` irreducible, but that is not happening here.)

### DIFF
--- a/src/lake/Lake/Build/Run.lean
+++ b/src/lake/Lake/Build/Run.lean
@@ -70,7 +70,7 @@ private abbrev MonitorM := ReaderT MonitorContext <| StateT MonitorState BaseIO
 @[inline] private def MonitorM.run
   (ctx : MonitorContext) (s : MonitorState) (self : MonitorM α)
 : BaseIO (α × MonitorState) :=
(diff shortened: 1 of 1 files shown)


tydeu:
If we are not treating `StateT` as see-through, are we also not going to treat `ReaderT` as see-through? I notice that in many of these cases this is now using `run` for the `StateT` component but not the `ReadertT` component of a monad stack. That strikes me as a bit confusing / inconsistent.

mhuisi (approved):
The changes to my parts of the code look good.

kim-em:
> If we are not treating `StateT` as see-through, are we also not going to treat `ReaderT` as see-through? I notice that in many of these cases this is now using `run` for the `StateT` component but not the `ReadertT` component of a monad stack. That strikes me as a bit confusing / inconsistent.

I haven't attempted a complete treatment of ReaderT, but now in all the touched files we use `run` with ReaderT, too.

PR #5125, opened 2024-08-22
feat: support for secure temporary files

hargoniX:


src/Init/System/IO.lean
```diff
@@ -455,6 +456,16 @@ see there for more information.
 -/
 @[extern "lean_io_rename"] opaque rename (old new : @& FilePath) : IO Unit
 
+/--
+Creates a temporary file in the most secure manner possible. There are no race conditions in the
+file’s creation. The file is readable and writable only by the creating user ID. Additionally
+on UNIX style platforms the file is executable by nobody. The function returns both a `Handle`
+to the already opened file as well as its `FilePath` in case that is of interest.
+
+Note that it is the caller's job to remove the file after use.
```
nomeata, line 465: I find “in case that is of interest” a bit confusing when the next line tell me I have to delete the file, so it should be of interest to every caller? 

Maybe just drop that half-sentence:

```suggestion
to the already opened file as well as its `FilePath`.

Note that it is the caller's job to remove the file after use.
```

src/runtime/io.cpp
```diff
@@ -828,6 +828,36 @@ extern "C" LEAN_EXPORT obj_res lean_io_rename(b_obj_arg from, b_obj_arg to, lean
...
+    if (retval > MAX_PATH || (retval == 0)) {
+        return io_result_mk_error((sstream() << GetLastError()).str());
+    }
+    // If the temp dir is so long that we can't put files into it something is seriously wrong.
+    const char* file_pattern = "tmp.XXXXXXXX";
+    const int file_pattern_size = strlen(file_pattern);
+    lean_always_assert(PATH_MAX >= strlen(path) + strnlen(file_pattern, file_pattern_size) + 1);
+    strncat(path, file_pattern, file_pattern_size);
+#else
+    const char* path_template = "/tmp/tmp.XXXXXXXX";
```
Kha, line 846: No love for `TMPDIR`? Is that what Rust does?

hargoniX: It would appear they do respect `TMPDIR` and default to `/tmp` otherwise, I'll change that.

PR #5130, opened 2024-08-22
feat: lemmas about List.join

kim-em:


src/Init/Data/List/Lemmas.lean
```diff
@@ -1771,6 +1776,55 @@ theorem join_concat (L : List (List α)) (l : List α) : join (L ++ [l]) = join
...
+      · obtain ⟨a', rfl, rfl⟩ := z
+        refine ⟨[], a', xs, ?_⟩
+        simp
+  · rintro ⟨as, bs, cs, rfl, h₁, rfl⟩
+    simp [join_eq_nil.mpr h₁]
+
+theorem join_eq_append (xs : List (List α)) (ys zs : List α) :
+    xs.join = ys ++ zs ↔
+      (∃ as bs, xs = as ++ bs ∧ ys = as.join ∧ zs = bs.join) ∨
+        ∃ as bs c cs ds, xs = as ++ (bs ++ c :: cs) :: ds ∧ ys = as.join ++ bs ∧
```
nomeata, line 1803: You could strengthen this by stating that `bs` is not empty, as else we are in the first case, right?

And maybe more natural to only have a nonempty `cs` rather than introducing `c::cs`?

Or if you like this style, use `b::bs` for the first list?

PR #5133, opened 2024-08-22
chore: running the simpNF linter over Lean

kim-em:
This should resolve nearly all of the simpNF lints. This is a follow-up to #4620.

### DIFF
--- a/src/Std/Data/DHashMap/Internal/List/Associative.lean
+++ b/src/Std/Data/DHashMap/Internal/List/Associative.lean
@@ -18,7 +18,6 @@ File contents: Verification of associative lists
 set_option linter.missingDocs true
 set_option autoImplicit false
 
-
 universe u v w
 
 variable {α : Type u} {β : α → Type v} {γ : α → Type w}


src/Std/Data/DHashMap/Internal/List/Associative.lean
```diff
@@ -26,6 +27,8 @@ open List (Perm)
 
 namespace Std.DHashMap.Internal.List
 
+attribute [-simp] List.isEmpty_eq_false
+
```
semorrison, line 31: @TwoFX, I'd like to have this lemma on globally, but it interferes with the way you've set things up here. I suspect your preference will be to turn it off locally, so you can keep your lemma names parallel with other files. Let me know if we should add a comment. (Feel free to push anything.)

src/Std/Data/DHashMap/Internal/List/Associative.lean
```diff
@@ -18,6 +18,7 @@ File contents: Verification of associative lists
 set_option linter.missingDocs true
 set_option autoImplicit false
 
+
```
semorrison, line 21:
```suggestion
```

PR #5140, opened 2024-08-23
feat: `debug.proofAsSorry`

Kha:
An experimental variant of `debug.byAsSorry`


semorrison:
I've only just had a look at this, but it is using sorry for all declarations, not just theorems! I pushed a (failing) test.

Kha:
It doesn't look like we'll need this

Issue #5050, opened 2024-08-15
Broken Lake linking


Kha:
Statically linking user executables against Lake is broken on Windows: https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/automatically.20generate.20ProjectName.2Elean/near/444253802. We should also enable dynamic linking while we're at it.

PR #5143, opened 2024-08-23
feat: Lake shared library

### LINKED ISSUE #2436: Lake native libraries include `main` symbol
see the issue at https://github.com/digama0/lean-sys/issues/5.

https://github.com/leanprover/lean4/blob/63d2bdd4908b4df7db381537af754a7319e334e4/src/lake/lakefile.lean#L6

In lake, it seems that, when `Main` resides in the lib source directory, the main symbol will also live inside the lib. Is this by design?

Kha:
Fixes #2436 #5050

Next step: when libLake_shared is in stage 0, --load-dynlib it when building stage 1 Lake

### DIFF
--- a/nix/bootstrap.nix
+++ b/nix/bootstrap.nix
@@ -134,7 +134,7 @@ lib.warn "The Nix-based build is deprecated" rec {
       mods = foldl' (mods: pkg: mods // pkg.mods) {} stdlib;
       print-paths = Lean.makePrintPathsFor [] mods;
       leanc = writeShellScriptBin "leanc" ''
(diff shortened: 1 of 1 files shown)


Kha:
@tydeu Moving `Lake.Main` out of Lake/ seemed like the easiest option to exclude it from the library outputs

tydeu:
@Kha Sounds good to me! 😄 Style-wise. I would probably just name it `Main.lean` unless that is an issues. As the file is already under `src/lake` and `Main.lean` is currently the standard naming convention used by executable templates.  Functionally-wise, the TOML configurations need updating to point the `lean_exe` to the new module (i.e., in `src/lake/lakefile.toml` and `src/lakefile.toml.in`).

Kha:
> The reason being the file is already under `src/lake`

I was hoping it would not stay there :)

Kha:
@tydeu I don't actively use the TOML files atm and don't want to touch them without testing, could you update them in a separate PR?

PR #5144, opened 2024-08-23
chore: CI: build 64-bit platforms consistently with GMP

Kha:
As proposed in https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Crashing.20bug.2C.20possibly.20related.20to.20UInt64.20arithmetic/near/463428033

Should be combined with the version bump in #5089

### DIFF
--- a/src/library/module.cpp
+++ b/src/library/module.cpp
@@ -52,7 +52,8 @@ struct olean_header {
     // 1 byte: version, incremented on structural changes to header
     uint8_t version = 2;
     // 1 byte of flags:
-    // - whether persisted bignums use GMP or Lean-native encoding
+    // * bit 0: whether persisted bignums use GMP or Lean-native encoding
+    // * bit 1-7: reserved
     uint8_t flags =
 #ifdef LEAN_USE_GMP
         0b1;
@@ -60,7 +61,7 @@ struct olean_header {
         0b0;
 #endif
     // 33 bytes: Lean version string, padded with '\0' to the right
-    // e.g. "4.12.0-nightly-2024-10-18"
+    // e.g. "4.12.0-nightly-2024-10-18". May not be null-terminated.
     char lean_version[33];
     // 81b008650766442a0dfa9faa796e4588c9d7d3a1
     // 40 bytes: build githash, padded with `\0` to the right


digama0:
While we're at it with the olean version bump, can we add a bit of data which explicitly indicates GMP-ness in the file? Also, it would be extremely useful to have the lean version string (i.e. `v4.12.0-rc1`) inside the file, with the field null or empty for custom, nightly, PR, fork toolchains. Right now the only version information available is the githash and this is useless for inequality queries.

eric-wieser:
Such information could be used in https://github.com/digama0/oleandump/pull/4, instead of the user having to tell `oleandump` which platform the oleans they are inspecting were built on.

src/library/module.cpp
```diff
@@ -49,17 +49,26 @@ namespace lean {
...
     char marker[5] = {'o', 'l', 'e', 'a', 'n'};
-    // 1 byte: version, currently always `1`
-    uint8_t version = 1;
-    // 42 bytes: build githash, padded with `\0` to the right
-    char githash[42];
+    // 1 byte: version, incremented on structural changes to header
+    uint8_t version = 2;
+    // 1 byte flags:
+    // - whether persisted bignums use GMP or Lean-native encoding
+    bool:1 uses_gmp =
```
Kha, line 56: @digama0 Here's the flag

src/library/module.cpp
```diff
@@ -49,17 +49,26 @@ namespace lean {
...
+#ifdef LEAN_USE_GMP
+        true;
+#else
+        false;
+#endif
+    bool:7 unused = 0;
+    // 41 bytes: build githash, padded with `\0` to the right
```
eric-wieser, line 63: This is the hex-encoded git hash rather than the raw 20-byte SHA, right?

digama0:
Thanks for adding the GMP flag. What are your thoughts on including the version information? (IIRC @eric-wieser also had a request for olean v2 but I forget what it is and it might be covered by my two requests.) If encoded as `(4u8, 13u8, 0u8, 1u8)` for `v4.13.0-rc1` RCs with `rc = 255` meaning the stable, then it can be done as pure lexicographic order, but encoding as a string is also an option which leaves more room for encoding unusual versions at the risk of underspecifying how version ordering should apply on these versions.

src/library/module.cpp
digama0: Yes

eric-wieser:
Perhaps you're thinking of me talking about #2908? This would be a change to what ends up in the olean, but not one that breaks any downstream readers, so is orthogonal to any versioning.

Kha:
@digama0 I added a Lean version string field and hope this is now enough .olean metadata for the next decade :) . Could you take a look?

Kha:
@digama0 @eric-wieser Final Comment Period :)

digama0:
Sorry for not sending my lgtm after your last message. I will get on implementing this in leangz.

eric-wieser:
My only comment here is whether we want to fix #2908 at the same time rather than bumping the olean version number twice.

src/library/module.cpp
```diff
@@ -49,17 +49,29 @@ namespace lean {
...
     // 5 bytes: magic number
     char marker[5] = {'o', 'l', 'e', 'a', 'n'};
-    // 1 byte: version, currently always `1`
-    uint8_t version = 1;
-    // 42 bytes: build githash, padded with `\0` to the right
-    char githash[42];
+    // 1 byte: version, incremented on structural changes to header
+    uint8_t version = 2;
+    // 1 byte of flags:
+    // - whether persisted bignums use GMP or Lean-native encoding
```
eric-wieser, line 55:
```suggestion
    // * bit 0: whether persisted bignums use GMP or Lean-native encoding
    // * bit 1-7: reserved
```

eric-wieser: Could potentially use
```
uint8_t uses_gmp : 1; 
uint8_t reserved_flags : 7;
```
instead

src/library/module.cpp
```diff
@@ -49,17 +49,29 @@ namespace lean {
...
+#ifdef LEAN_USE_GMP
+        0b1;
+#else
+        0b0;
+#endif
+    // 33 bytes: Lean version string, padded with '\0' to the right
+    // e.g. "4.12.0-nightly-2024-10-18"
+    char lean_version[33];
+    // 81b008650766442a0dfa9faa796e4588c9d7d3a1
+    // 40 bytes: build githash, padded with `\0` to the right
```
eric-wieser, line 66: There is no padding here, right?

src/library/module.cpp
```diff
@@ -49,17 +49,29 @@ namespace lean {
...
+    // 1 byte of flags:
+    // - whether persisted bignums use GMP or Lean-native encoding
+    uint8_t flags =
+#ifdef LEAN_USE_GMP
+        0b1;
+#else
+        0b0;
+#endif
+    // 33 bytes: Lean version string, padded with '\0' to the right
+    // e.g. "4.12.0-nightly-2024-10-18"
```
eric-wieser, line 63:
```suggestion
    // 33 bytes: Lean version string, padded with '\0' to the right
    // e.g. "4.12.0-nightly-2024-10-18". May not be null terminated.
```
according to the docs for strncopy.

src/library/module.cpp
```diff
@@ -49,17 +49,29 @@ namespace lean {
...
+    // 40 bytes: build githash, padded with `\0` to the right
+    char githash[40];
     // address at which the beginning of the file (including header) is attempted to be mmapped
     size_t base_addr;
     // payload, a serialize Lean object graph; `size_t` has same alignment requirements as Lean objects
     size_t data[];
 };
 // make sure we don't have any padding bytes, which also ensures `data` is properly aligned
-static_assert(sizeof(olean_header) == 5 + 1 + 42 + sizeof(size_t), "olean_header must be packed");
+static_assert(sizeof(olean_header) == 5 + 1 + 1 + 33 + 40 + sizeof(size_t), "olean_header must be packed");
```
eric-wieser, line 74: Perhaps having an `static_assert(offsetof(olean_header,data) % 40 = 0)` or whatever the actual alignment requirements are.

src/library/module.cpp
digama0: I'd prefer if it didn't, this adds an unnecessary dependency on [more C wibbly bits](https://stackoverflow.com/a/19376742/890016) than necessary

src/library/module.cpp
digama0: There can be, if the githash is empty. I don't think there is anything syntactically restricting it to be exactly 40 bytes, although it will be for actual lean releases.

digama0:
> My only comment here is whether we want to fix #2908 at the same time rather than bumping the olean version number twice.

I think #2908 can be fixed without a version bump, since the current implementation is nondeterministic anyway and this just resolves the nondeterminism.

src/library/module.cpp
eric-wieser: Ouch, I retract that suggestion then.

eric-wieser:
Bumping the version means you can easily check whether you need to consider determinism when inspecting the file.

Kha:
I won't worry about #2908 for now, given that it's only relevant for non-standard Lean builds now

digama0:
Which milestone is this change set for? I want to make sure I release there are no unpatched lean versions between when leangz releases a new version and when the patch comes out in an RC or regular release.

kim-em:
Oops, @Kha has sent this to the queue, but that won't work as it has stage0 updates.

I will rebase / redo stage0 now.

kim-em:
@digama0, I think if we merge in the next 24 hours, this will arrive in `v4.14.0-rc1`. Would that work for you for `leangz`?

digama0:
I have a version of leangz I would like to test, but this PR has no working pr-release AFAICT

digama0:
Okay, https://github.com/digama0/leangz/compare/olean2 seems to work based on a manually built lean executable using this PR. But I did notice one issue: Currently, manually built lean executables have the `lean_version` field set to e.g. `4.12.0`, even though they are not official releases. Can we do something so that `LEAN_VERSION_IS_RELEASE` is also included in that field? I would like to be able to differentiate between official releases and unofficial ones (and correctly ordering official releases) without looking at the githash.

kim-em:
> Okay, https://github.com/digama0/leangz/compare/olean2 seems to work based on a manually built lean executable using this PR. But I did notice one issue: Currently, manually built lean executables have the `lean_version` field set to e.g. `4.12.0`, even though they are not official releases. Can we do something so that `LEAN_VERSION_IS_RELEASE` is also included in that field? I would like to be able to differentiate between official releases and unofficial ones (and correctly ordering official releases) without looking at the githash. cc: @Kha

Oof, this is not exactly your question, but I see that I have screwed up the last two months and not advanced the `LEAN_VERSION_MINOR` on `master`.

digama0:
Note, to fix this issue I think it suffices to ensure that any non-official release has a `lean_version` string not matching the regex `[0-9]+.[0-9]+.[0-9]+(-rc[0-9]+)?`. For example, using the lean-toolchain name verbatim would work since it starts with a repo name. Other options include just leaving the field blank for these releases, or prefixing or suffixing with `custom`, etc.

PR #5145, opened 2024-08-23
feat: split `Lean.Kernel.Environment` from `Lean.Environment`

Kha:
This PR splits the environment used by the kernel from that used by the elaborator, providing the foundation for tracking of asynchronously elaborated declarations, which will exist as a concept only in the latter.

Minor changes:
* kernel diagnostics are moved from an environment extension to a direct environment as they are the only extension used directly by the kernel
* `initQuot` is moved from an environment header field to a direct environment as it is the only header field used by the kernel; this also makes the remaining header immutable after import


tobiasgrosser:
I got it. TCB = Trusted Code Base

Kha:
!bench

src/Lean/Message.lean
```diff
@@ -545,6 +542,15 @@ def toMessageData (e : KernelException) (opts : Options) : MessageData :=
...
   | deepRecursion                       => "(kernel) deep recursion detected"
   | interrupted                         => "(kernel) interrupted"
+where mkCtx (_env : Kernel.Environment) (lctx : LocalContext) (opts : Options) (msg : MessageData) : MessageData :=
+  -- In the split of `Lean.Kernel.Environment` from `Lean.Environment`, the environment returned in
+  -- kernel exceptions is not of the correct type anymore, and we do not want to pass the full
+  -- environment to the kernel. Thus the environment set here may be missing declarations added
+  -- during the kernel execution up to the exception. We could update `elabEnv.base` with this
+  -- environment but the environment type is opaque at this point. We claim that the added
+  -- declarations should rarely be relevant to the presentation of this less frequent exception.
+  MessageData.withContext { env := elabEnv, mctx := {}, lctx := lctx, opts := opts } msg
```
leodemoura, line 552: Delete unused argument `_env`?
Why do we still store `env` in the exception constructors above if they are not used anymore?
The comment above also looks fishy to me. The caller may call `toMessageData` with a. completely different `Environment` object (e.g., the empty one).

src/Lean/Environment.lean
```diff
@@ -88,11 +88,6 @@ structure EnvironmentHeader where
   -/
   trustLevel   : UInt32       := 0
```
leodemoura, line 89: We should move `trustLevel` to `Kernel.Environment`. It is not being used since we moved from Lean 3, but we will eventually be used again.

src/Lean/Environment.lean
```diff
@@ -106,24 +101,172 @@ structure EnvironmentHeader where
...
+An environment stores declarations provided by the user. The kernel currently supports different
+kinds of declarations such as definitions, theorems, and inductive families. Each has a unique
+identifier (i.e., `Name`), and can be parameterized by a sequence of universe parameters. A constant
+in Lean is just a reference to a `ConstantInfo` object. The main task of the kernel is to type check
+these declarations and refuse type incorrect ones. The kernel does not allow declarations containing
+metavariables and/or free variables to be added to an environment. Environments are never
+destructively updated.
+
+This type contains only the minimal data necessary for basic type checking. Other data used only by
+and for elaboration, as well data for the TCB extension of native reduction, is stored in
```
leodemoura, line 123: The comment regarding native reduction looks quite fishy to me. We discussed this issue before. As soon as we have the new code generator, the IR bytecode must be in the `Kernel.Environment`.

src/Lean/Environment.lean
```diff
@@ -106,24 +101,172 @@ structure EnvironmentHeader where
...
+  { env with quotInit := true }
+
+@[export lean_environment_quot_init]
+private def isQuotInit (env : Environment) : Bool :=
+  env.quotInit
+
+/--
+Type check given declaration and add it to the environment
+
+**NOTE**: This function does not implement `reduceBool`/`reduceNat` special reduction rules.
```
leodemoura, line 193: The lack of support for `reduceBool` / `reduceNat` in `Kernel.Environment` is creating too many issues and hacks. This API feels like a footgun.

leodemoura:
@Kha This PR has update stage0 commits. So, squash should not be used, and the commit messages should be fixed.

Kha:
!bench

PR #5149, opened 2024-08-24
fix: `conv => arg n` to handle `.subsingletonInst`

### LINKED ISSUE #4394: conv’s arg fails with subsingleton parameters
Consider
```
opaque foo (p : Prop) [Decidable p] : Prop

/--
error: tactic 'rfl' failed, equality expected
  Decidable ?m.58
p : Prop
inst✝ : Decidable p
⊢ Decidable ?m.58
-/
#guard_msgs in
example  [Decidable p] : foo p := by
  conv => arg 1
  sorry
```

The problem seems to be this code:

```lean
private def selectIdx (tacticName : String) (mvarIds : List (Option MVarId)) (i : Int) :
  TacticM Unit := do
  if i >= 0 then
    let i := i.toNat
    if h : i < mvarIds.length then
      for mvarId? in mvarIds, j in [:mvarIds.length] do
        match mvarId? with
        | none => pure ()
        | some mvarId =>
          if i != j then
            mvarId.refl
      match mvarIds[i] with
      | none => throwError "cannot select argument"
      | some mvarId => replaceMainGoal [mvarId]
      return ()
  throwError "invalid '{tacticName}' conv tactic, application has only {mvarIds.length} (nondependent) argument(s)"
```

This assumes that all goals in `mvarIds`, which is the result of 
```
def congr (mvarId : MVarId) (addImplicitArgs := false) (nameSubgoals := true) :
    MetaM (List (Option MVarId))
```
are equality goals, but due to
```
      | .subsingletonInst =>
        proof := mkApp proof arg
        let rhs ← mkFreshExprMVar (← whnf (← inferType proof)).bindingDomain!
        proof := mkApp proof rhs
        mvarIdsNewInsts := mvarIdsNewInsts.push (some rhs.mvarId!)
```
there this can also contain subsingleton goals.

Probably these should be solved, once the new `p` is known, using type class search?

### Versions

`4.10.0-nightly-2024-06-07`

### Impact

nomeata:
this fixes #4394, see there for an analysis.

### DIFF
--- a/tests/lean/run/issue4394.lean
+++ b/tests/lean/run/issue4394.lean
@@ -28,8 +28,31 @@ example [Decidable p] [Decidable (id p)] (h : foo p) : foo (id p) := by
   conv => arg 1; rw [id]
   exact h
 
+/--
+error: tactic 'fail' failed
+p : Prop
+inst✝ : Decidable p
+⊢ Decidable (id p)
+---
+error: unsolved goals
+p : Prop
+inst✝ : Decidable p
+⊢ foo (id p)
+-/
+#guard_msgs in
+example [Decidable p] : foo p := by
+  conv =>
+    arg 1
+    · rw [← id.eq_def p]
+    · tactic => fail
+
 example [PropClass n] [PropClass (id n)] (h : P1 n) : P1 (id n) := by
   conv => arg 1; rw [id]
+  exact h
+
+example [PropClass n] (h : P1 n) : P1 n := by
+  conv => arg 1; rw [← id.eq_def n]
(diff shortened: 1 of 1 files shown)


nomeata:
While it’s a bit fishy for non-conv-goals to come out of a conv tactic, this isn't the fault of `arg n` (it already comes out of `congr` that way), and it’s ceratinly better leave the goal there than to try `.refl` if it can’t work. So while this might not solve everything in this corner of the system, it’s likely an improvement. Will let it sit for a bit in case someone cares and has thoughts, and else merge.

PR #5153, opened 2024-08-25
feat: lemmas about List.attach/pmap

kim-em:


src/Init/Data/List/Attach.lean
```diff
@@ -73,6 +73,13 @@ theorem pmap_map {p : β → Prop} (g : ∀ b, p b → γ) (f : α → β) (l H)
   · rfl
   · simp only [*, pmap, map]
 
+@[simp] theorem attach_cons (x : α) (xs : List α) :
```
nomeata, line 76: Looking at this I wonder how useful it is to have `pmap` at all, or whether it could just always be simped to `.attach.map`, and only those get a simp API?

PR #5167, opened 2024-08-26
feat: allow users to disable `simpCtorEq` simproc 

### LINKED ISSUE #5046: RFC: remove `simpCtorEq` from the default simprocs
### Proposal

Currently `simp only` simplifies an equality `a = b` to `False`, if `a` and `b` are different constructors for an inductive type.  This is apparently done by the `Meta.Simp.simpCtorEq` simproc.

Would it be possible to turn off this `simproc` in `simp only`, while still keeping it in the set of rules used by `simp`?  Here are three arguments in favour of this change:

1) This simproc can cause unpredictable behaviour.  For example,
```lean
example : (1:Nat) = 0 := by
  simp only -- goal `False`

example : (1:Int) = 0 := by
  simp only -- "simp made no progress"

example : (-1:Int) = 0 := by
  simp only -- goal `False`
```
The `simpCtorEq` simproc is applicable to the first and last of these goals but not to the middle one.

2) This simproc can cause inefficient behaviour.  For example (from @kmill):
```lean
example : 2^10000 = 2^9999 := by
  simp (config := Lean.Meta.Simp.neutralConfig) only
/-
exponent 10000 exceeds the threshold 256, exponentiation operation was not evaluated,
use `set_option exponentiation.threshold <num>` to set a new threshold
-/
```

3) There are use cases where unambiguously this simproc is undesirable, so it's awkward to have it baked into `simp only`, where it is (AFAICT) impossible to turn off.  For example, the simp tactic is [used internally in the Mathlib `ring` tactic](https://github.com/leanprover-community/mathlib4/blob/08ecf4c7e62daf335fa2f9e42d6ad90574211eb5/Mathlib/Tactic/Ring/RingNF.lean#L162-L166) to clean up failures before they are presented to the user.  This simproc actually cleans up some failures too aggressively, so that the information to the user is lost!
```lean
import Mathlib

-- simproc does not act, user gets useful error message
example (x : ℤ) : x + 1 - x = 0 := by ring
/-
error: unsolved goals
⊢ 1 = 0
-/

-- simproc acts, user does not get useful error message
example (x : ℤ) : x - 1 - x = 0 := by ring
/-
error: unsolved goals
⊢ False
-/
```

### Community Feedback

Discussion at 
https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/simp.20only.20on.20numeric.20goals

### Impact

leodemoura:
`simp only` will not apply this simproc anymore. Users must now write `simp only [reduceCtorEq]`. See RFC #5046 for motivation.
This PR also renames simproc to `reduceCtorEq`. 

close #5046 

@semorrison A few `simp only ...` tactics will probably break in Mathlib. Fix: include `reduceCtorEq`.

### DIFF
--- a/src/Init/Data/List/Erase.lean
+++ b/src/Init/Data/List/Erase.lean
@@ -39,13 +39,13 @@ theorem eraseP_of_forall_not {l : List α} (h : ∀ a, a ∈ l → ¬p a) : l.er
   | cons x xs ih =>
     simp only [eraseP_cons, cond_eq_if]
     split <;> rename_i h
-    · simp only [false_or]
+    · simp
       constructor
       · rintro rfl
         simpa
       · rintro ⟨_, _, rfl, rfl⟩
         rfl
-    · simp only [cons.injEq, false_or, false_iff, not_exists, not_and]
+    · simp
       rintro x h' rfl
       simp_all
 
--- a/src/Init/Data/List/Nat/Range.lean
+++ b/src/Init/Data/List/Nat/Range.lean
@@ -200,7 +200,7 @@ theorem range'_eq_append_iff : range' s n = xs ++ ys ↔ ∃ k, k ≤ n ∧ xs =
       cases k with
       | zero => simp [range'_succ]
       | succ k =>
-        simp only [range'_succ, false_and, cons.injEq, true_and, ih, exists_eq_left', false_or]
+        simp [range'_succ, ih, exists_eq_left']
         refine ⟨k, ?_⟩
         simp_all
         omega


leodemoura:
TODO: we must update-stage0 after merging this PR, and then remove staging workarounds.

leodemoura:
Managed to build stage2 locally.

leodemoura:
@semorrison I had to add a few workarounds to build stage2. I will cleanup after merging the PR, and updating stage0.

nomeata:
About the naming of the simpproc: judging from the code it only works for equalities between different constructors, but doesn't apply injectivity? The name could make users believe that it would.

Also do we have a consistent naming scheme for simpprocs? reduceIte doesn't start with `simp`.

I think it's worth thinking about a good name for a moment here, as due to `simp?` this will be exposed to unsuspecting users, and if it looks too scarily internal (as `simpCtorEq` maybe does), it's not great for the early user experience.

Maybe to stick to existing jargon, why not `noConfusion` or `noConstrConfusion`? Or less jargony `constrDontOverlap`?

nomeata:
And while we are at it: I would find it useful if, when reading code (either as a learner or as a reviewer) I could quickly grasp which `simp` arguments are plain theorems, and which are simprocs.

So a consistent naming convention for simprocs may be useful. I’d say it needn’t have `simp`, as it’s always about `simp`.

I was thinking about using a `Proc` suffix, e.g. `noConfusionProc`, `iteProc`, `diteProc`…

Most (all?) of the simprocs in core are called `reduceFoo`, so just sticking to that (`reduceNoConfusion`) is also an feasible option.

leodemoura:
> Also do we have a consistent naming scheme for simprocs? `reduceIte` doesn't start with `simp`.

Note that the `simproc` was renamed to `reduceCtorEq`.

leodemoura:
> I was thinking about using a `Proc` suffix, e.g. `noConfusionProc`, `iteProc`, `diteProc`…

I am not opposing this suggestion, but it should be a different PR if users see a benefit on this change. Remark: it will break many proofs in Mathlib.

leodemoura:
> About the naming of the simproc: judging from the code it only works for equalities between different constructors, but doesn't apply injectivity?

We use the injectivity theorems for that. We can add a docstring to make it clear the simproc only handles the different constructor case.

PR #5170, opened 2024-08-26
fix: Syntax.unsetTrailing

### LINKED ISSUE #4958: Go to definition on import doesn't work on last module name
```
import Init
--^ works ^ works
import Init
--^ works ^ doesn't
```

Kha:
Fixes #4958


Kha:
We should figure out a few to test import requests

PR #5174, opened 2024-08-26
chore: use text-variant of ✝︎

### LINKED ISSUE #5015: RFC: Use emoji variant selector for unicode characters
### Proposal

Occasionally, Lean uses emojis, for example it uses `✅❌💥` for debugging. The proposal is to use the [variant selector](https://en.wikipedia.org/wiki/Variation_Selectors_(Unicode_block)) with `\uFE0F` followed:

In particular, use

```
emoji variant: ✅️❌️💥️
text variant: ✝︎▼︎▶︎
```
(which look the same, but are different to the currently used `✅❌💥✝▼▶`)

#### User Experience

Minor improvement, but helps with using custom fonts. See for example [live.lean-lang.org](https://live.lean-lang.org/#code=set_option%20trace.Meta.synthInstance%20true%0A%0A%23synth%20Decidable%20Nat%0A):

```lean
set_option trace.Meta.synthInstance true

#synth Decidable Nat
```
which displays

```
[Meta.synthInstance] ❌︎ CoeT Type Nat Prop ▼
  [] new goal CoeT Type Nat Prop ▶
  [] ❌︎ apply @instCoeT to CoeT Type Nat Prop ▶
  [] ✅︎ apply @instCoeTOfCoeDep to CoeT Type Nat Prop ▶
  [] ✅︎ apply @instCoeTOfCoeHTCT to CoeT Type Nat Prop ▶
  [] ❌︎ apply @instCoeHTCT to CoeHTCT Type Prop ▶
  [] ✅︎ apply @instCoeHTCTOfCoeHTC to CoeHTCT Type Prop ▶
  [] ❌︎ apply @instCoeHTC to CoeHTC Type Prop ▶
  [] ✅︎ apply @instCoeHTCOfCoeOTC to CoeHTC Type Prop 
```

instead of

```
[Meta.synthInstance] ❌️ CoeT Type Nat Prop ▼
  [] new goal CoeT Type Nat Prop ▶
  [] ❌️ apply @instCoeT to CoeT Type Nat Prop ▶
  [] ✅️ apply @instCoeTOfCoeDep to CoeT Type Nat Prop ▶
  [] ✅️ apply @instCoeTOfCoeHTCT to CoeT Type Nat Prop ▶
  [] ❌️ apply @instCoeHTCT to CoeHTCT Type Prop ▶
  [] ✅️ apply @instCoeHTCTOfCoeHTC to CoeHTCT Type Prop ▶
  [] ❌️ apply @instCoeHTC to CoeHTC Type Prop ▶
  [] ✅️ apply @instCoeHTCOfCoeOTC to CoeHTC Type Prop 
```

Further, using the wrong font, one might see `inst✝️` instead of `inst✝︎`, which is distracting.

#### Beneficiaries

Better support for custom fonts helps in particular the international community as for example Chinese, Japanese or Cyrillic script might require usage of a custom font. E.g. JuliaMono is defacto a solid choice and also used by live.lean-lang.org.

#### Maintainability

There is some potential issue to consider about unicode characters that look identical.

### Community Feedback

There was some private feedback by Joachim Breitner (@nomeata) that `❌︎` and `✅︎` in the trace output is not nice as they are harder to recognise.

[Zulip discussion](https://leanprover.zulipchat.com/#narrow/stream/113488-general/topic/Lean.20specific.20font) suggests this could be a valid approach if it doesn't create new pitfalls for the user.

### Impact

joneugster:
Second part of https://github.com/leanprover/lean4/issues/5015, using emoji variant of unicode symbols for ✝︎.

---

(Partially) closes #5015

### DIFF
--- a/src/Init/Meta.lean
+++ b/src/Init/Meta.lean
@@ -117,6 +117,7 @@ def getRoot : Name → Name
 def isInaccessibleUserName : Name → Bool
   -- note: this uses `✝` (`\u{271D}`) whereas everywhere else in the code
   -- we use the text-variant selector `✝︎` (`\u{271D}\u{FE0E}`).
+  -- This is easier simply because we don't have `s.hasSubstring` in core.
   | Name.str _ s   => s.contains '✝' || s == "_inaccessible"
   | Name.num p _   => isInaccessibleUserName p
   | _              => false


joneugster:
Note: This appends the text-variant selector `\u{FE0E}` to all `✝︎` `\u{271D}\u{FE0E}`.

Implications:
- Batteries: TODO
- VSCode-lean4/infoview: TODO
- Mathilb: TODO

src/Init/Meta.lean
```diff
@@ -115,7 +115,7 @@ def getRoot : Name → Name
 
 @[export lean_is_inaccessible_user_name]
 def isInaccessibleUserName : Name → Bool
-  | Name.str _ s   => s.contains '✝︎' || s == "_inaccessible"
+  | Name.str _ s   => s.contains '✝' || s == "_inaccessible"
```
joneugster, line 118: TODO: here it crashed when changing it.

nomeata: Maybe it's not a single character to lean's lexer?

joneugster: Indeed, it says "missing end of character literal". This makes me think that this change to `✝︎` might be a bad idea even though it probably works the way the PR is rn...

nomeata: That’s the same error you get with
```lean
#check 'ab'
```

So I’d say

* That error message should be improved (Maybe parse `'ab'` like `"ab"` and give a helpful error if it’s not a single unicode codepoint.
* For this PR, the `s.contains '✝︎'` has to become `s.hasSubString '✝'`, or (a bit hacky, leave comment) leave the code as it, since the old `✝` character appears in the new string `✝`, so this check will still work.

src/Init/Meta.lean
```diff
@@ -115,6 +115,8 @@ def getRoot : Name → Name
 
 @[export lean_is_inaccessible_user_name]
 def isInaccessibleUserName : Name → Bool
+  -- note: this uses `✝` (`\u{271D}`) whereas everywhere else in the code
+  -- we use the text-variant selector `✝︎` (`\u{271D}\u{FE0E}`).
   | Name.str _ s   => s.contains '✝' || s == "_inaccessible"
```
joneugster, line 120: I now added this comment. I am partially concerned that these unicode variant selectors will be forgotten and will confuse a future developer, but it's up to the maintainers how you want to document this correctly...

src/Init/Meta.lean
```diff
@@ -115,6 +115,8 @@ def getRoot : Name → Name
 
 @[export lean_is_inaccessible_user_name]
 def isInaccessibleUserName : Name → Bool
+  -- note: this uses `✝` (`\u{271D}`) whereas everywhere else in the code
+  -- we use the text-variant selector `✝︎` (`\u{271D}\u{FE0E}`).
```
nomeata, line 119: Maybe also say _why_ we use something else here?

```suggestion
  -- we use the text-variant selector `✝︎` (`\u{271D}\u{FE0E}`).
  -- This is easier simply because we don't have `s.hasSubstring` in core.
```

nomeata:
(Just labeling `awaiting-author` as per TODO list in first comment.)

joneugster:
I am not 100% confident, that this will not cause problems anywhere, and I don't strictly need this change. (having emojis marked as emojis means I can load the emoji font *behind* the non-emoji one and everything renders fine).

So I'm going to close this PR

PR #5182, opened 2024-08-27
feat: well-founded recursion: opaque well-foundedness proofs

### LINKED ISSUE #2171: Kernel tries to reduce WF definitions
The following theorem fails to type-check in the kernel:
```lean
theorem y : id (Nat.gcd 314159 10000) = Nat.gcd 314159 10000 := rfl
-- (kernel) deep recursion detected
```

Presumably this happens because we provide a too constructive proof of `Acc (· < ·) 314159`, which allows the kernel to unfold the definition 314159 times.  (theorized by @digama0)

as reported by @thorimur

nomeata:
This PR makes functions defined by well-founded recursion use an `opaque` well-founded proof by default. This reliably prevents kernel reduction of such definitions and proofs, which tends to be prohibitively slow (fixes #2171), and which regularly causes hard-to-debug kernel type-checking failures. This changes renders `unseal` ineffective for such definitions. To avoid the opaque proof, annotate the function definition with `@[semireducible]`.


nomeata:
Closing since #2171 was demoted to low priority.

nomeata:
May need more kernel changes (#7545)

nomeata:
Still not good enough:
```
 error: ././././Archive/Imo/Imo2024Q5.lean:922:6: (kernel) application type mismatch
argument has type
  (⟨(↑⟨1, hi✝⟩ + 1) / 2, fn1OfEdge0.proof_3 N ⟨1, hi✝⟩⟩, ⟨↑⟨1, hi✝⟩ / 2 + 1, fn1OfEdge0.proof_4 N ⟨1, hi✝⟩ h⟩) ∈
    MonsterData.monsterCells N m
but function has type
  (1, ⟨1, fn1OfEdge0.proof_4 N ⟨1, hi✝⟩ h⟩) ∈ MonsterData.monsterCells N m
```

Sigh. I guess defining `Nat.divCore` and `Nat.modCore` with fuel should help.

PR #5200, opened 2024-08-29
feat: proposed change to BitVec API

kim-em:
This renames `BitVec.getLsb` to `getLsbD` (`D` for "default" value, i.e. false), and introduces `getLsb?` and `getLsb'` (which we can rename to `getLsb` after a deprecation cycle).

(Similarly for `getMsb`.)

Also adds a `GetElem` class so we can use `x[i]` and `x[i]?` notation. 

Later, we will turn
```
theorem getLsbD_eq_getElem?_getD (x : BitVec w) (i : Nat) (h : i < w) :
    x.getLsbD i = x[i]?.getD false
```
on as a `@[simp]` lemma.

This PR doesn't attempt to demonstrate the benefits, but I think both arguments are going to get easier, and this will bring the BitVec API closer in line to List/Array, etc.

### DIFF
--- a/src/Init/Data/BitVec/Basic.lean
+++ b/src/Init/Data/BitVec/Basic.lean
@@ -121,22 +121,22 @@ Return the `i`-th least significant bit.
 
 This will be renamed `getLsb` after the existing deprecated alias is removed.
 -/
-@[inline] def getLsb' (x : BitVec w) (i : Nat) (_ : i < w) : Bool := x.toNat.testBit i
+@[inline] def getLsb' (x : BitVec w) (i : Fin w) : Bool := x.toNat.testBit i
 
 /-- Return the `i`-th least significant bit or `none` if `i ≥ w`. -/
 @[inline] def getLsb? (x : BitVec w) (i : Nat) : Option Bool :=
-  if h : i < w then some (getLsb' x i h) else none
+  if h : i < w then some (getLsb' x ⟨i, h⟩) else none
 
 /--
 Return the `i`-th most significant bit.
 
 This will be renamed `getMsb` after the existing deprecated alias is removed.
 -/
-@[inline] def getMsb' (x : BitVec w) (i : Nat) (_ : i < w) : Bool := x.getLsb' (w-1-i) (by omega)
+@[inline] def getMsb' (x : BitVec w) (i : Fin w) : Bool := x.getLsb' ⟨w-1-i, by omega⟩
 
 /-- Return the `i`-th most significant bit or `none` if `i ≥ w`. -/
 @[inline] def getMsb? (x : BitVec w) (i : Nat) : Option Bool :=
-  if h : i < w then some (getMsb' x i h) else none
+  if h : i < w then some (getMsb' x ⟨i, h⟩) else none
 
 /-- Return the `i`-th least significant bit or `false` if `i ≥ w`. -/
 @[inline] def getLsbD (x : BitVec w) (i : Nat) : Bool :=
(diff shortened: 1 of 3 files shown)


src/Init/Data/BitVec/Folds.lean
```diff
@@ -52,13 +52,13 @@ private theorem iunfoldr.eq_test
 
 theorem iunfoldr_getLsb' {f : Fin w → α → α × Bool} (state : Nat → α)
```
TwoFX, line 53:
```suggestion
theorem iunfoldr_getLsbD' {f : Fin w → α → α × Bool} (state : Nat → α)
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -478,13 +478,13 @@ protected theorem extractLsb_ofNat (x n : Nat) (hi lo : Nat) :
...
   (extractLsb hi lo x).toNat = (x.toNat >>> lo) % 2^(hi-lo+1) := rfl
 
-@[simp] theorem getLsb_extractLsb' (start len : Nat) (x : BitVec n) (i : Nat) :
-    (extractLsb' start len x).getLsb i = (i < len && x.getLsb (start+i)) := by
-  simp [getLsb, Nat.lt_succ]
+@[simp] theorem getLsbD_extractLsb' (start len : Nat) (x : BitVec n) (i : Nat) :
+    (extractLsb' start len x).getLsbD i = (i < len && x.getLsbD (start+i)) := by
+  simp [getLsbD, Nat.lt_succ]
 
 @[simp] theorem getLsb_extract (hi lo : Nat) (x : BitVec n) (i : Nat) :
```
TwoFX, line 485:
```suggestion
@[simp] theorem getLsbD_extract (hi lo : Nat) (x : BitVec n) (i : Nat) :
```

tobiasgrosser (approved):

src/Init/Data/BitVec/Basic.lean
```diff
@@ -116,17 +116,67 @@ end zero_allOnes
...
+@[simp] theorem getLsb'_eq_getElem (x : BitVec w) (i : Nat) (h : i < w) :
+    x.getLsb' i h = x[i] := rfl
+
+/-- We prefer `x[i]?` as the simp normal form for `getLsb?` -/
+@[simp] theorem getLsb?_eq_getElem? (x : BitVec w) (i : Nat) :
+    x.getLsb? i = x[i]? := rfl
+
+theorem getLsbD_eq_getElem?_getD (x : BitVec w) (i : Nat) (h : i < w) :
+    x.getLsbD i = x[i]?.getD false := by
+  simp only [getElem?, decidableGetElem?, getElem]
```
TwoFX, line 175: To avoid pain, never unfold `getElem?`, always use `getElem?_def` :)
```suggestion
  rw [getElem?_def]
```

TwoFX (approved):

hargoniX (approved):

src/Init/Data/BitVec/Basic.lean
```diff
@@ -116,17 +116,67 @@ end zero_allOnes
 
 section getXsb
 
+/--
+Return the `i`-th least significant bit.
+
+This will be renamed `getLsb` after the existing deprecated alias is removed.
+-/
+@[inline] def getLsb' (x : BitVec w) (i : Nat) (_ : i < w) : Bool := x.toNat.testBit i
```
alexkeizer, line 124: I think it might make sense to bundle the proof, and have this take an `i : Fin w`, akin to how [`List.get`](https://leanprover-community.github.io/mathlib4_docs/Init/Prelude.html#List.get) does it.
IIRC, there are some places we currently have `getLsb i` with `i : Fin w` (relying on the `Fin` to `Nat` coercion). If we bundle the proof in a single `Fin` argument, these call sites will just magically keep working

src/Init/Data/BitVec/Basic.lean
```diff
@@ -116,17 +116,67 @@ end zero_allOnes
...
+/-- We prefer `x[i]?` as the simp normal form for `getLsb?` -/
+@[simp] theorem getLsb?_eq_getElem? (x : BitVec w) (i : Nat) :
+    x.getLsb? i = x[i]? := rfl
+
+theorem getLsbD_eq_getElem?_getD (x : BitVec w) (i : Nat) (h : i < w) :
+    x.getLsbD i = x[i]?.getD false := by
+  rw [getElem?_def]
+  split <;> rfl
+
+end getElem
```
alexkeizer, line 178: We've previously had quite some discussion about not wanting to treat either endianness preferentially.
Is that something we should worry about with this `GetElem` instance? I guess there's not really a good way to have both an lsb- and msb-based instance without picking favourites?

alexkeizer (commented):
I like the getLsb -> getLsbD rename (in fact, I've proposed this exact change before, [back when we were still in now-batteries/then-std4](https://github.com/leanprover-community/batteries/pull/400))

src/Init/Data/BitVec/Basic.lean
semorrison: Yes, briefly discussed this with Tobias and Markus yesterday. We can only support one under `GetElem`. I think we will just have to document and follow through on our intention to build the API for `getMsb(|D|?)` too! It's already been falling behind of course, and for some purposes, e.g. bitblasting, it makes sense to pick a preferred one.

src/Init/Data/BitVec/Basic.lean
semorrison: Agreed.

PR #5205, opened 2024-08-29
feat: use incrementality for completion in tactic blocks

mhuisi:
This PR enables the use of incrementality for completion in tactic blocks. Consider the following example:
```lean
example : True := by
  have : True := T
  sleep 10000
```

Before this PR, in order to respond to a completion request after `T`, `sleep 10000` has to complete first since the command must be fully elaborated. After this PR, the completion request is responded to immediately.

### DIFF
--- a/src/Lean/Server/FileWorker/RequestHandling.lean
+++ b/src/Lean/Server/FileWorker/RequestHandling.lean
@@ -34,7 +34,7 @@ def findCompletionInfoTreeAtPos
     : Task (Option Elab.InfoTree) :=
   -- NOTE: use `+ 1` since we sometimes want to consider invalid input technically after the command,
   -- such as a trailing dot after an option name. This shouldn't be a problem since any subsequent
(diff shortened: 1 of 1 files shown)


src/Lean/Server/FileWorker/RequestHandling.lean
```diff
@@ -38,15 +38,13 @@ def handleCompletion (p : CompletionParams)
   -- NOTE: use `+ 1` since we sometimes want to consider invalid input technically after the command,
   -- such as a trailing dot after an option name. This shouldn't be a problem since any subsequent
   -- command starts with a keyword that (currently?) does not participate in completion.
```
Kha, line 40: This doesn't seem to be true anymore as the subsequent syntax can now also be a tactic keyword and they do participate in completion

mhuisi: I changed the implementation to use the syntax endpos now, which (I think) should not cause any issues.

PR #5213, opened 2024-08-30
feat: BitVec.eq_of_getElem_eq

tobiasgrosser:


src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -168,6 +168,17 @@ theorem getMsbD_eq_getMsb?_getD (x : BitVec w) (i : Nat) :
       intros
       omega
 
+theorem eq_of_getElem_eq {x y : BitVec w}
+    (pred : ∀ (i : Fin w), x[i] = y[i]) : x = y := by
```
TwoFX, line 172: I think the preferred spelling of this would be to use `Nat`  instead of `Fin` to be consistent with the bits of the API added in #5200. 
```suggestion
    (pred : ∀ (i : Nat) (_ : i < w), x[i] = y[i]) : x = y := by
```
See also the analogous lemma [`List.ext_getElem`](https://leanprover-community.github.io/mathlib4_docs/Init/Data/List/Lemmas.html#List.ext_getElem).

tobiasgrosser: It seems for `BitVec` we used historically:

```lean
theorem eq_of_getLsbD_eq {x y : BitVec w}
    (pred : ∀(i : Fin w), x.getLsbD i.val = y.getLsbD i.val) : x = y := by

theorem eq_of_getMsbD_eq {x y : BitVec w}
    (pred : ∀(i : Fin w), x.getMsbD i.val = y.getMsbD i.val) : x = y := by
```

I guess we should at the very least remain consistent within `BitVec`.

I now added both variants to this PR and ported one theorem to show the code impact. Happy to flip to Nat, but would appreciate a comment why this is preferable over Fin.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -107,6 +107,8 @@ theorem eq_of_toNat_eq {n} : ∀ {x y : BitVec n}, x.toNat = y.toNat → x = y
 
 theorem testBit_toNat (x : BitVec w) : x.toNat.testBit i = x.getLsbD i := rfl
 
+theorem testBit_toNat_getElem (x : BitVec w) (i : Fin w): x.toNat.testBit i = x[i.val] := rfl
+
```
tobiasgrosser, line 111:
```suggestion
theorem testBit_toNat_getElem (x : BitVec w) (i : Fin w) : x.toNat.testBit i = x[i.val] := rfl

```

semorrison:
I don't think it's a good idea to prove theorems about `zeroExtend` without having proved the corresponding theorems about `zeroExtend'` first.

src/Init/Data/BitVec/Basic.lean
```diff
@@ -170,6 +170,9 @@ instance : GetElem (BitVec w) Nat Bool fun _ i => i < w where
 @[simp] theorem getLsb?_eq_getElem? (x : BitVec w) (i : Nat) :
     x.getLsb? i = x[i]? := rfl
 
+theorem getElem_eq_toNat_testBit (x : BitVec w) (i : Fin w) :
+  x[i] = x.toNat.testBit i := rfl
```
semorrison, line 174: This should use separate `i : Nat` and `h : i < w` arguments, otherwise it's only applicable for `getElem` of the coercion of a Fin.

semorrison:
@tobiasgrosser, e.g. my first steps here would be #5247.

tobiasgrosser:
This was resolved in https://github.com/leanprover/lean4/pull/5247.

PR #5215, opened 2024-08-30
feat: List.erase_range

kim-em:


semorrison:
This really is scraping the bottom of the barrel, but I had it around, and it's not impossible someone will want it!

PR #5217, opened 2024-08-30
fix: BitVec benchmark after renaming of getLsb

hargoniX:
fallout from https://github.com/leanprover/lean4/pull/5200


hargoniX:
!bench

PR #5222, opened 2024-08-31
perf: reduce allocations in `Json.compress`

hargoniX:
1. Remove the need to allocate an intermediate `String` for literally every character in a JSON `String`.
2. Use a single `String` buffer in the entire `Json.compress` machinery.
3. Use `toListAppend`

Number 1 is doing most of the lifting in the perf diff, the rest are some minor but measurable improvements.


hargoniX:
!bench

hargoniX:
!bench

mhuisi (approved):

PR #5226, opened 2024-09-01
perf: maximize term sharing at `instantiateMVarDeclMVars`

leodemoura:
The `instantiateMVarDeclMVars` is used by `runTactic`.

This commit address a performance issue at `LinearCategory`.


leodemoura:
!bench

PR #5227, opened 2024-09-01
test: #5225 and #5226 on top of nightly-with-mathlib

leodemoura:
**SHOULD NOT BE MERGED.**
Testing purposes only


semorrison:
This gave great results: a 3.5% instruction count reduction on mathlib.

PR #5233, opened 2024-09-02
refactor: WF.fix: use WellFoundedRelation.fixBy

nomeata:
this defines
```
WellFoundedRelation.fixBy {α : Sort u} {β : Sort v} [WellFoundedRelation β] (f : α → β)
  {C : α → Sort w} (F : ∀ x, (∀ y, WellFoundedRelation.rel (f y) (f x) → C y) → C x) (x : α) : C x
```
based on `WellFounded.fix`, `WellFoundedRelation` and `InvImage`. This
moves some logic from meta code in WF.Fix to this definitions, so mostly
a conceptual simplicification, with the effect that `invImage` should
no longer show up to the user (even with #5016 disabled.)


nomeata:
There is some rather hackish way in which `WF.Fix` recognizes which termination goals belong to which function in the case of mutual recursion, and this break with this refactoring. I’ll put this on hold until I have a better solution there.

nomeata:
With #5234 for now off the roadmap, this refactoring isn’t so pressing anymore. Maybe some other time.

PR #5243, opened 2024-09-03
feat: upstream operation properties from mathlib `Init`

Parcly-Taxel:
This is part of an effort to clear out mathlib's `Init` folder. The new (bundled) classes in `Init.Core` correspond to (unbundled) definitions in `Mathlib.Init` as follows:
* `Std.SymmetricOp` <- `IsSymmOp` in `Mathlib.Init.Algebra.Classes`
* `Std.LeftCancellative` <- `IsLeftCancel`
* `Std.RightCancellative` <- `IsRightCancel`
* `Std.LeftCommutative` <- `LeftCommutative` in `Mathlib.Init.Logic`
* `Std.RightCommutative` <- `RightCommutative`
* `Std.leftComm` <- `left_comm`
* `Std.rightComm` <- `right_comm`

Cf. leanprover-community/mathlib4#16438.


semorrison:
@Parcly-Taxel, could you please just move these in Mathlib? There is no immediate need for these here.

PR #5244, opened 2024-09-03
feat: (`DHashMap`|`HashMap`|`HashSet`).(`getKey?`|`getKey`|`getKey!`|`getKeyD`)

hargoniX:


src/Std/Data/HashSet/Basic.lean
```diff
@@ -88,6 +88,13 @@ Equivalent to (but potentially faster than) calling `contains` followed by `inse
   let ⟨replaced, r⟩ := m.inner.containsThenInsertIfNew a ()
   ⟨replaced, ⟨r⟩⟩
 
+/--
+Checks if given key is contained and returns the key if it is, otherwise `none`.
+The result in the `some` case is guaranteed to be pointer equal to the key in the set.
+-/
+@[inline] def getKey? (m : HashSet α) (a : α) : Option α :=
```
TwoFX, line 95: What do you think, should this be called `get?`?

TwoFX, line 95: The idea is that the lemma files are structured like this pseudo-code:
```suggestion
for i in [0, num_ops):
  for j in [0, i]:
    print_lemmas_relating_op[i]_and_op[j]
```
It looks to me that this isn't the case right now, so you should reorder either the operations or the lemmas.

src/Std/Data/HashSet/Lemmas.lean
```diff
@@ -177,6 +189,158 @@ theorem size_le_size_erase [EquivBEq α] [LawfulHashable α] {k : α} :
...
+    (m.erase k).getKey? a = if k == a then none else m.getKey? a :=
+  HashMap.getKey?_erase
+
+@[simp]
+theorem getKey?_erase_self [EquivBEq α] [LawfulHashable α] {k : α} : (m.erase k).getKey? k = none :=
+  HashMap.getKey?_erase_self
+
+theorem getKey_insert [EquivBEq α] [LawfulHashable α] {k a : α} {h₁} :
+    (m.insert k).getKey a h₁ =
+      if h₂ : k == a ∧ ¬ k ∈ m then k else m.getKey a (mem_of_mem_insert' h₁ h₂) :=
```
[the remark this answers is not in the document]
hargoniX: Was only done here according to my grep-fu

src/Std/Data/HashSet/Basic.lean
hargoniX: Renamed!

PR #5265, opened 2024-09-05
chore: remove unused deriving handler argument syntax

david-christiansen:
As far as I can tell, the ability to pass a structure instance to a deriving handler is not actually used in practice. It didn't seem to be used in the test suite, at least.

Do we want to remove this, or do we want to use and document it? This PR removes it, but that's not something I feel strongly about - but seeing if it breaks Mathlib is a useful data point.


david-christiansen:
From what I can read here, the Mathlib breakage is only due to a deprecation warning being issued, and not that it doesn't compile, right?

If so, I take that as evidence that this syntax is indeed not used and can be removed.

kim-em:
> From what I can read here, the Mathlib breakage is only due to a deprecation warning being issued, and not that it doesn't compile, right?

Right.

Kha (approved):

kim-em:
Perhaps needs `master` merged? This failed on the queue.

david-christiansen:
Thanks for the poke, @kim-em!

PR #5270, opened 2024-09-06
chore: turn off [Inhabited β] : Inhabited (Sum α β)

kim-em:
This would disable the instances for `Inhabited (Sum α β)` from `Inhabited β`, because this can easily result in non-defeq Inhabited instances by competing with the instance from the left-summand.

I'm not sure whether this is a good idea or not: people might be surprised that this instance is missing.

An alternative would be an instance `[IsEmpty α] [Inhabited β] : Inhabited (Sum α β)` that could only fire when the other instances *couldn't*. However this would either have to happen in Mathlib, or we'd need to move `IsEmpty` up to Lean.

See [zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Inhabited.20.28Sum.20_.20_.29/near/467988659) discussion.

### DIFF
--- a/src/Init/Core.lean
+++ b/src/Init/Core.lean
@@ -165,13 +165,13 @@ inductive PSum (α : Sort u) (β : Sort v) where
 
 @[inherit_doc] infixr:30 " ⊕' " => PSum
 
-instance instInhabitedPSumLeft {α β} [Inhabited α] : Inhabited (PSum α β) := ⟨PSum.inl default⟩
+instance PSum.inhabitedLeft {α β} [Inhabited α] : Inhabited (PSum α β) := ⟨PSum.inl default⟩
 
 /--
 `PSum α β` is inhabited if `β` is inhabited.
 This is not an instance to avoid non-defeq instances.
 -/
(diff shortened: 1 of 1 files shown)


src/Init/Core.lean
```diff
@@ -165,9 +165,13 @@ inductive PSum (α : Sort u) (β : Sort v) where
...
 
-instance {α β} [Inhabited α] : Inhabited (PSum α β) := ⟨PSum.inl default⟩
+instance instInhabitedPSumLeft {α β} [Inhabited α] : Inhabited (PSum α β) := ⟨PSum.inl default⟩
 
-instance {α β} [Inhabited β] : Inhabited (PSum α β) := ⟨PSum.inr default⟩
+/--
+`PSum α β` is inhabited if `β` is inhabited.
+This is not an instance to avoid non-defeq instances.
+-/
+def instInhabitedPSumRight {α β} [Inhabited β] : Inhabited (PSum α β) := ⟨PSum.inr default⟩
```
urkud, line 174: Should this be `reducible` so that people can use it as a local or `letI` instance? Same below.

semorrison: Thanks, yes.

semorrison:
Closed in favour of #5284

PR #5276, opened 2024-09-06
fix: prevent `addPPExplicitToExposeDiff` from assigning metavariables

kmill:
Type mismatch errors have a nice feature where expressions are annotated with `pp.explicit` to expose differences via `isDefEq` checking. However, this procedure has side effects since `isDefEq` may assign metavariables. This PR wraps the procedure with `withoutModifyingState` to prevent assignments from escaping.

Assignments can lead to confusing behavior. For example, in the following a higher-order unification fails, but the difference-finding procedure unifies metavariables in a naive way, producing a baffling error message:
```lean
theorem test {f g : Nat → Nat} (n : Nat) (hfg : ∀a, f (g a) = a) :
    f (g n) = n := hfg n

example {g2 : ℕ → ℕ} (n2 : ℕ) : (λx => x * 2) (g2 n2) = n2 := by
  with_reducible refine test n2 ?_
  /-
  type mismatch
    test n2 ?m.648
  has type
    (fun x ↦ x * 2) (g2 n2) = n2 : Prop
  but is expected to have type
    (fun x ↦ x * 2) (g2 n2) = n2 : Prop
  -/
```
With the change, it now says `has type ?m.153 (?m.154 n2) = n2`.

Note: this uses `withoutModifyingState` instead of `withNewMCtxDepth` because we want to know something about where `isDefEq` failed — we are trying to simulate a very basic version of `isDefEq` for function applications, and we want the state at the point of failure to know which argument is "at fault".

### DIFF
--- a/tests/lean/run/addPPExplicitToExposeDiff.lean
+++ b/tests/lean/run/addPPExplicitToExposeDiff.lean
@@ -29,7 +29,7 @@ Error message shouldn't fake a higher-order unification. This next one used to g
     (fun x ↦ x * 2) (g2 n2) = n2 : Prop
 ```
 It now doesn't for the stronger reason that we don't let `addPPExplicitToExposeDiff` have side effects,
-but still it avoids doing higher-order unifications in its reasoning.
+but still it avoids doing incorrect higher-order unifications in its reasoning.
 -/
 
 theorem test {f g : Nat → Nat} (n : Nat) (hfg : ∀ a, f (g a) = a) :


tests/lean/run/4405.lean
```diff
@@ -15,9 +15,9 @@ def foo := fun n => (not_and_self_iff _).mp ⟨Nat.lt_irrefl _, Fin.is_lt _⟩
 
 /--
 error: type mismatch
-  Fin.is_lt ?m.185
+  Fin.is_lt ?m.181
 has type
-  ↑?m.185 < ?m.184 : Prop
```
nomeata, line 20: `set_option pp.mvars false`?

tests/lean/1870.lean.expected.out
```diff
@@ -1,7 +1,7 @@
 1870.lean:21:2-21:35: error: type mismatch
-  congrArg (@OfNat.ofNat Nat 0) (congrArg (@Zero.toOfNat0 Nat) ?_)
+  congrArg ?_ (congrArg ?_ ?_)
 has type
-  OfNat.ofNat 0 = OfNat.ofNat 0 : Prop
+  ?_ (?_ ?_) = ?_ (?_ ?_) : Prop
 but is expected to have type
   OfNat.ofNat 0 = OfNat.ofNat 1 : Prop
```
nomeata, line 6: Hmm, in this instance it’s not really an improvement. As a user I look at this and wonder why these metavariables couldn’t just be instantiated. But maybe this example is somewhat artificial.

Maybe add the test from the issue description to see one where it’s a UX improvement?

kmill: Yeah, this one is a bit obscure. I've added some tests.

I tried making a version that is consistent about partially assigning metavariables to mimic a half-successful isDefEq, but I think in the end that's more confusing, since you don't know what the real type is. It might be nice having a real isDefEq diagnostic function instead.

PR #5278, opened 2024-09-07
feat: add `BitVec.getElem_truncate`

tobiasgrosser:


tobiasgrosser:
@semorrison, we modelled this PR after #5247, but I am a bit confused that we would not have getElem on the RHS of these identities. Could you briefly explain why the RHS has getLsbD?

tobiasgrosser:
I chatted with @alexkeizer, and his suggestion was to add:

```lean4
theorem getLsbD_eq_getElem {x : BitVec w} {i : Nat} (h : i < w) :
    x.getLsbD i = x[i] := by
  simp [getLsbD, getElem_eq_testBit_toNat]
```

to get back to getElem after these rewrites to getLsbD, if the relevant hypothesis is flying around. Does this make sense from your perspective?

semorrison:
> I chatted with @alexkeizer, and his suggestion was to add:
> 
> ```lean
> theorem getLsbD_eq_getElem {x : BitVec w} {i : Nat} (h : i < w) :
>     x.getLsbD i = x[i] := by
>   simp [getLsbD, getElem_eq_testBit_toNat]
> ```
> 
> to get back to getElem after these rewrites to getLsbD, if the relevant hypothesis is flying around. Does this make sense from your perspective?

Yes, this looks good. I think it can even be a `simp` lemma.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -523,6 +523,14 @@ theorem getElem?_zeroExtend (m : Nat) (x : BitVec n) (i : Nat) :
   all_goals (first | apply getLsbD_ge | apply Eq.symm; apply getLsbD_ge)
     <;> omega
 
+theorem getElem_truncate (m : Nat) (x : BitVec n) (i : Nat) (hi : i < m) :
+    (truncate m x)[i] = x.getLsbD i := by
+  simp [getElem?_eq, getElem_zeroExtend]
```
semorrison, line 528: This one can just be `x[i]` on the right hand side.

My suggestion (not entirely sure I've followed it previously, but I think) is to prefer `x[i]` over `.getLsbD` on the RHS where possible --- but when it isn't possible without an `if` or `getD`, accepting `getLsbD`.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -523,6 +523,14 @@ theorem getElem?_zeroExtend (m : Nat) (x : BitVec n) (i : Nat) :
   all_goals (first | apply getLsbD_ge | apply Eq.symm; apply getLsbD_ge)
     <;> omega
 
+theorem getElem_truncate (m : Nat) (x : BitVec n) (i : Nat) (hi : i < m) :
```
semorrison, line 526: @[simp]

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -523,6 +523,14 @@ theorem getElem?_zeroExtend (m : Nat) (x : BitVec n) (i : Nat) :
...
+    (truncate m x)[i] = x.getLsbD i := by
+  simp [getElem?_eq, getElem_zeroExtend]
+
+theorem getElem?_truncate (m : Nat) (x : BitVec n) (i : Nat) :
+    (truncate m x)[i]? = if i < m then some (x.getLsbD i) else none :=
```
semorrison, line 531:
```suggestion
    (truncate m x)[i]? = if h : i < m then some x[i] else none :=
```

semorrison:
Apologies for the delay reviewing here.

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: Done.

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: AFAIU, this can only be `x[i]` on the RHS if I introduce `h : i < n`. For some reason, truncate is just an alias to `zeroExtend`, so there is no knowledge that `i < x`. Or maybe I miss something here?

I implemented this change by adding the `h : i < n` hypothesis. This does not feel that beautiful, though.

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: I am not sure why you suggest adding `h :` here. AFAIU the hypothesis is taken into account even without the `h`. What is missing here is `h : i < n` to allow for the use of getElem. I moved this to `getElem` but at the cost of the new `h` hypothesis.

tobiasgrosser:
@semorrison, I applied your suggestions. However, I am not use we want to implement them without adding a new hypothesis `h : i < n`. @alexkeizer's suggestion of introducing `getLsbD_eq_getElem` was precisely to have the other theorems to be able to have a weaker precondition and a weaker RHS, but allowing for an automatic translation to getElem, whenever a stronger precondition is available.

`getLsbD_eq_getElem` is not yet @[simp] as this will break the build without a flashed out getElem API. So I will enable it only after the API is complete.

tobiasgrosser:
awaiting-review

src/Init/Data/BitVec/Lemmas.lean
semorrison: Oh, sorry, I misread the theorem. Sorry, no change required here.

src/Init/Data/BitVec/Lemmas.lean
semorrison: I'm sorry, I'm wrong again here...

tobiasgrosser:
I moved this back to `getLsbD` on the RHS. @semorrison, feel invited to merge.

src/Init/Data/BitVec/Basic.lean
```diff
@@ -173,6 +173,11 @@ instance : GetElem (BitVec w) Nat Bool fun _ i => i < w where
 theorem getElem_eq_testBit_toNat (x : BitVec w) (i : Nat) (h : i < w) :
   x[i] = x.toNat.testBit i := rfl
 
+/-- This should become @[simp] after the getElem API has been completed. -/
```
semorrison, line 176:
```suggestion
```
Prefer to leave such comments out of the source.

Issue #5234, opened 2024-09-02
RFC: Bounded recursion instead of well-founded recursion on Nat


nomeata:
### Motivation

Right now, functions defined by well-founded recursion get compiled to (effectively)
```lean
fixBy {α : Sort u} {β : Sort v} [WellFoundedRelation β] (f : α → β)
  {C : α → Sort w} (F : ∀ x, (∀ y, WellFoundedRelation.rel (f y) (f x) → C y) → C x) (x : α) : C x
```

Internally, this function uses recursion on `Acc`, and is known to reduce very inefficiently in the kernel, sometimes causing odd errors #2171. We have made these functions irreducible to the meta code (https://github.com/leanprover/lean4/pull/4061) and am considering also making it irreducible to the kernel (#5192).

However, many recursive functions have a measure function that maps into `Nat`. These functions do not need the full power of well-founded recursion: That `Nat` simply indicates an upper bound on the number of recursive calls and can be implemented using the following fuel-based fixpoint combinator:

```lean
def Nat.fix (h : α → Nat) {C : α → Sort u}  (F : (x : α) → ((y : α) → h y < h x → motive y) → motive x) : (x : α) → C x :=
  let rec go : ∀ (fuel : Nat) (x : α), (h x < fuel) → motive x := Nat.rec 
    (fun _ hfuel => (not_succ_le_zero _ hfuel).elim)
    (fun _ ih x hfuel => F x (fun y hy => ih y (Nat.lt_of_lt_of_le hy (Nat.le_of_lt_add_one hfuel))))
  fun x => go (h x + 1) x (Nat.lt_add_one _)
```

This is essentially drop-in, so existing features of recursive functions (equational theorems, functional induction) should work as before. But with this fuel-based implementation, kernel reduction works fine.

For example, `Nat.sqrt 441` now reduces in 2m instead of 1400ms!

### Proposal

So we could let `WF.Fix` recognize when the `termination_by` argument indicates a termination argument of type `Nat`, and then use the different combinator, and *not* mark the functions as irreducible.

### Impact

Some functions would now behave well when reduced by the kernel. This would already help mathlib, which uses such reduction to decide primality of small numbers (see https://github.com/leanprover/lean4/issues/5192#issuecomment-2319061605 for more investigation about this function) and will become much more relevant when more proof-by-reflection happens.

[...]

PR #5280, opened 2024-09-07
refactor: WF.Fix: store original declName in recApp mdata

nomeata:
This removes some code smell in `WF.Fix`, and thus unblocks #5233.


nomeata:
With #5234 for now off the roadmap, this refactoring isn’t so pressing anymore. Will solve this puzzle some other time.

Issue #1867, opened 2022-11-21
passing instances to named arguments turns explicit args into implicit


digama0:
```lean
structure Bla : Type
class Foo (x : Bla) (y : Int) : Prop
opaque Bla.bla (x : Bla) (y : Int) [foo : Foo x y] : Nat

variable (x : Bla) (y : Int) (f : β → β)
#check x.bla y (foo := ⟨⟩) -- fails
#check x.bla (y := y) (foo := ⟨⟩) -- fails
-- function expected at
--   Bla.bla ?m.87500 ?m.87503
-- term has type
--   Nat
```

This is a recent regression caused by a5ab59a (the fix for #1851), and came up in the std4 nightly bump leanprover/std4#31.

Here is a proposal for how to handle app args that should fix both this issue and #1851:

* Process the arguments in left-to-right order, starting with the one before the dot.
* The argument before the dot triggers dot-notation lookup and always goes in for the argument of that function of the correct type.
* Whenever filling an argument, all arguments it depends on are marked as implicit.
* If you encounter a regular argument, it goes in for the first unfilled explicit argument.
* If you encounter a named argument, it goes in for that argument (and also marks all args it depends on as implicit if they are not already filled).
* When you run out of arguments, all unfilled implicit arguments up to the first semi-implicit or explicit argument are replaced with `_` and the rest are reverted.

(I believe the only difference from current behavior being suggested here is that named parameters should be processed in order with explicit args rather than all at the beginning.)

This has the following behavior:
* `x.bla y (foo := ⟨⟩)` turns into `@Bla.foo x y ⟨⟩`
* `x.bla (foo := ⟨⟩) y` is an error, because the `y` argument was made implicit
* `x.bla (y := y) (foo := ⟨⟩)` and `x.foo (foo := ⟨⟩) (y := y)` both work
* `x.bla` works and has type `∀ (y : Int) [foo : Foo x y], Nat`

and using the example from #1851:
* `f.val` works and results in `@Approx.val (α → β) f' (X → Y) f : X → Y`
* `f.val x.val` is an over-application because `f.val` has already filled all the explicit arguments, so this is handled as `(f.val) x.val` which works

kmill:
This issue was reported by Terry Tao in https://leanprover.zulipchat.com/#narrow/stream/217875-Is-there-code-for-X.3F/topic/Set.2EtoFinset.20without.20Fintype.20or.20Set.2EFinite/near/418374814

He found he had to write
```lean
[...]

Issue #1851, opened 2022-11-18
Dot notation with typeclass instances issue


lecopivo:
There seems to be an issue when using dot notation with type class instances.

mwe:
```
class Approx {α : Type} (a : α) (X : Type) : Type where
  val : X

variable {α β X Y : Type} {f' : α → β} {x' : α} [f : Approx f' (X → Y)] [x : Approx x' X]

-- fails
#check f.val x.val

-- works
#check let f'' := f.val
       let x'' := x.val
       f'' x''
```

The `f.val x.val` fails with
```
application type mismatch
  @Approx.val X (Approx.val x') ?m.70 f
argument
  f
has type
  Approx f' (X → Y) : Type
but is expected to have type
  Approx (Approx.val x') ?m.70 : Type
```

I would expect `f.val x.val` to typecheck

Tested on Lean (version 4.0.0-nightly-2022-11-18, commit 7529e86307a1, Release)

PR #5283, opened 2024-09-08
fix: refine how named arguments suppress explicit arguments

kmill:
**Note:** Most of the implementation of this PR is at #5512 due to a bad rebase before merging.

Recall that currently named arguments suppress all explicit parameters that are dependencies. This PR limits this feature to only apply to true structure projections, except in the case where it is triggered when there are no more positional arguments. This preserves the primary reason for generalizing this feature (issue #1851), while removing the generalized feature, which has led to numerous confusions (issue #1867). This also fixes a bug pointed out [on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/.40foo.20.28A.20.3A.3D.20bar.29.20_.20_/near/468564862) where in `@` mode, instance implicit parameter dependencies to named arguments would be suppressed unless the next positional argument was `_`.

More detail:
* The `NamedArg` structure now has a `suppressDeps : Bool` field. It is set to `true` for the `self` argument in structure projections. If there is such a `NamedArg`, explicit parameters that are dependencies to the named argument are turned into implicit arguments. The consequence is that *all* structure projections are treated as if their type parameters are implicit, even for class projections. This flag is *not* used for generalized field notation.
* We preserve the suppression feature when there are no positional arguments remaining. This feature pre-dates the fix to issue #1851, and it is useful when combining named arguments and the eta expansion feature, since dependencies of named arguments cannot be turned into eta arguments. Plus, there are examples of the form `rw [lem (h := foo)]` where `lem` has explicit arguments that `h` depends on.
* For instance implicit parameters in explicit mode, now `_` arguments register terminfo and are hoverable.
* Now `..` is respected in explicit mode.

This implements RFC #5397. The `suppressDeps` flag suggests a future possibility of a named argument syntax that can suppress dependencies.

### DIFF
--- a/src/Lean/Elab/App.lean
+++ b/src/Lean/Elab/App.lean
@@ -654,10 +654,10 @@ mutual
       if (← anyNamedArgDependsOnCurrent) then
         /-
         See the note in processExplicitArg about `anyNamedArgDependsOnCurrent`.
-        For consistency, instance implicit arguments should remain instance implicit if a named
-        argument depends on them.
+        For consistency, instance implicit arguments should always become implicit if a named argument depends on them.
         If we do not do this check here, then the `nextArgHole?` branch being before `processExplicitArg`
-        would result in counterintuitive behavior. For example, in the following the `_` is optional.
+        would result in counterintuitive behavior.
+        For example, without this block of code, in the following the `_` is optional.
         When it is omitted, `processExplicitArg` sees that `h` depends on the `Decidable` instance
         so makes the `Decidable` instance argument become implicit.
         When it is `_`, the `nextArgHole?` branch allows it to be present.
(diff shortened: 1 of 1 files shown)


kmill:
Originally this PR conservatively changed how `_` worked for instance arguments in explicit mode. Now it changes how named arguments suppress dependencies in general, starting from [a4af20a](https://github.com/leanprover/lean4/pull/5283/commits/a4af20a274b182b1009c627648c9154384403404).

src/Lean/Elab/App.lean
```diff
@@ -641,24 +664,30 @@ mutual
...
     if (← read).explicit then
-      if (← isNextArgHole) then
-        /- Recall that if '@' has been used, and the argument is '_', then we still use type class resolution -/
-        let arg ← mkFreshExprMVar (← getArgExpectedType) MetavarKind.synthetic
+      if let some stx ← nextArgHole? then
+        -- We still use typeclass resolution for `_` arguments.
+        -- This behavior can be suppressed with `(_)`.
+        let ty ← getArgExpectedType
+        let arg ← mkInstMVar ty
+        addTermInfo' stx arg ty (← getLCtx)
```
kmill, line 677:
```suggestion
        addTermInfo' stx arg ty
```

PR #5291, opened 2024-09-09
feat: set priority in monadic class instances

### LINKED ISSUE #4212: RFC: type class synthesis order when `outParam` has multiple possible values

JovanGerb:
This PR adds the `high` priority to some instances with monad transformers, most notably to `MonadExceptOf ε (ExceptT ε m)`. This means that in `ExceptT α MetaM`, `throw` will throw the `α` instead of the `Exception`.

These instances now have `high` priority:
- MonadReaderOf ρ (ReaderT ρ m)
- MonadWithReaderOf ρ (ReaderT ρ m)
- MonadStateOf σ (StateT σ m)
- MonadStateOf σ (EStateT ε σ m)
- MonadExceptOf ε (ExceptT ε m)
- MonadExceptOf ε (EStateT ε σ m)
- MonadExceptOf ε (ExceptCpsT ε m)
- MonadCallStackOf κ (CallStackT κ m)
- MonadCycleOf κ (CycleT κ m)

And these instance now have `low` priority:
- MonadExceptOf Unit (OptionT m)
- [MonadExcept ε m] {α : Type v} : OrElse (m α)

The last one makes sure that the `OrElse` instance has lower priority than `instOrElseOfAlternative` (this change doesn't affect the synthesis order, because the order of the files in which the instances were declared was already 'correct').

In Prelude.lean, instead of setting the priority to `high` or `low`, it has to be set to `10000` or `100`, which is the same.

The increased priority of `MonadExceptOf ε (ExceptT ε m)` affected a few files that had to be fixed.

Closes #4212

### DIFF
--- a/src/Init/Data/Nat/Bitwise/Lemmas.lean
+++ b/src/Init/Data/Nat/Bitwise/Lemmas.lean
@@ -711,6 +711,32 @@ theorem mul_add_lt_is_or {b : Nat} (b_lt : b < 2^i) (a : Nat) : 2^i * a + b = 2^
   rw [mod_two_eq_one_iff_testBit_zero, testBit_shiftLeft]
   simp
 
+theorem testBit_mul_two_pow (x i n : Nat) :
+    (x * 2 ^ n).testBit i = (decide (n ≤ i) && x.testBit (i - n)) := by
+  rw [← testBit_shiftLeft, shiftLeft_eq]
+
+theorem bitwise_mul_two_pow (of_false_false : f false false = false := by rfl) :
+    (bitwise f x y) * 2 ^ n = bitwise f (x * 2 ^ n) (y * 2 ^ n) := by
+  apply Nat.eq_of_testBit_eq
+  simp only [testBit_mul_two_pow, testBit_bitwise of_false_false, Bool.if_false_right]
+  intro i
+  by_cases hn : n ≤ i
+  · simp [hn]
+  · simp [hn, of_false_false]
+
+theorem shiftLeft_bitwise_distrib {a b : Nat} (of_false_false : f false false = false := by rfl) :
+    (bitwise f a b) <<< i = bitwise f (a <<< i) (b <<< i) := by
+  simp [shiftLeft_eq, bitwise_mul_two_pow of_false_false]
+
+theorem shiftLeft_and_distrib {a b : Nat} : (a &&& b) <<< i = a <<< i &&& b <<< i :=
+  shiftLeft_bitwise_distrib
+
+theorem shiftLeft_or_distrib {a b : Nat} : (a ||| b) <<< i = a <<< i ||| b <<< i :=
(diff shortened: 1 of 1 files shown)


JovanGerb:
The test that broke had the number from the name generator offset by 8.
The line `local infix:65 (priority := high) "*" => mul` previously used to increment the name generator by 1560, and now this became 1568. I don't understand where these 8 extra names come from, but I equally don't understand why a simple `infix` declaration needs to use so ridiculously many names.

kim-em:
!bench

kim-em:
I'm concerned that there's an unexplained performance regression revealed by `!bench`. Can you determine if that is real or not, and if so try to localize the exact change that causes it?

JovanGerb:
Can you benchmark it again? There seems to be nothing in the rbmap.lean benchmarking test related to the changes in this PR.

JovanGerb:
@kim-em, can you have a look at this one again? I don't know how to verify that the slowdowns weren't real.

I've now updated the PR to the new toolchain, but I'm not sure how to get it to build the accompanying Batteries and Mathlib.

kim-em:
@JovanGerb, you'll need to rebase this PR onto `origin/nightly-with-mathlib`. Then you can run `!bench` again.

JovanGerb:
It seems I don't understand how git works, so I've opened it as a new PR: #6725

JovanGerb:
> @JovanGerb, you'll need to rebase this PR onto `origin/nightly-with-mathlib`. Then you can run `!bench` again.

I tried to !bench the new PR, but it seems that I still cannot do that myself.

Kha:
Instructions at least are unchanged, most likely noise

PR #5292, opened 2024-09-09
refactor: `Lean.Elab.Deriving.FromToJson`

arthur-adjedj:
Refactors the derive handlers for `ToJson` and `FromJson` in preparation for #3160.
This splits up the different parts of the handler according to how other similar handlers are implemented while keeping the original logic intact. This makes the changes necessary to adapt the file in #3160 much easier.

### DIFF
--- a/src/Lean/Elab/Deriving/FromToJson.lean
+++ b/src/Lean/Elab/Deriving/FromToJson.lean
@@ -180,7 +180,7 @@ def mkFromJsonBody (ctx : Context) (e : Expr) : TermElabM Term := do
 def mkFromJsonAuxFunction (ctx : Context) (i : Nat) : TermElabM Command := do
   let auxFunName := ctx.auxFunNames[i]!
   let indval     := ctx.typeInfos[i]!
-  let header     ←  mkFromJsonHeader indval --TODO fix header info
+  let header     ←  mkFromJsonHeader indval
   let binders    := header.binders
   Term.elabBinders binders fun _ => do
   let type ← Term.elabTerm header.targetType none


arthur-adjedj:
WIP

arthur-adjedj:
awaiting-review

mhuisi:
!bench

mhuisi:
Ah, we unfortunately don't have a benchmark for `FromToJson` (the `.ilean` roundtrip one uses custom handlers). Since this PR might affect language server performance, it might be good to add one to `tests/bench/speedcenter.exec.velcom.yaml` in a separate PR to confirm that this will not be an issue.

nomeata:
>  Since this PR might affect language server performance,

Thanks for flagging! It wouldn’t affect performance if the generated code is going to be the same, and it’s just refactoring the meta-code, right?

mhuisi:
> > Since this PR might affect language server performance,
> 
> Thanks for flagging! It wouldn’t affect performance if the generated code is going to be the same, and it’s just refactoring the meta-code, right?

Yes, and the benchmark would help confirm that the generated code is identical.

arthur-adjedj:
- How good is the test coverage ? 
There are a few tests that check the correctness of outputs from the handler, namely [`toFromJson.lean`](https://github.com/leanprover/lean4/blob/master/tests/lean/run/toFromJson.lean). These derive handlers are also heavily used in the language server, which, IIRC, is tested fairly enough in the `interactive` tests.
- How confident are you ? 
I'm confident that the generated code is the exact same, the core logic behind the handler has stayed the same after all. Furthermore, the derive handler should behave very similarly in terms of performance. 
I'll dive into the code generated in the `toFromJson.lean` test, ensure it corresponds with `nightly-with-mathlib` and confirm that in a minute.

nomeata (approved):

arthur-adjedj:
Alright, here are the diffs for each ir code generated on each inductive/structure in [`toFromJson.lean`](https://github.com/leanprover/lean4/blob/master/tests/lean/run/toFromJson.lean). As far as I saw scrolling through, the only differences were in the hygienic naming, the code itself has stayed the same.
- [Foo](https://www.diffchecker.com/BppPZACf/)
- [WInfo](https://www.diffchecker.com/gON5jDxg/)
- [E](https://www.diffchecker.com/u8eVA6Cu/)
- [ERec](https://www.diffchecker.com/i8LGiiYI/)
- [ENest](https://www.diffchecker.com/Wk1jtuzd/)
- [EParam](https://www.diffchecker.com/PPWdJs5f/)
- [Minus](https://www.diffchecker.com/uMKmOVLw/)
- [MinusAsk](https://www.diffchecker.com/WrO5fAHY/)

nomeata:
Thanks for checking!

nomeata:
That said, a benchmark for the generated code would of course be great to have more assurance with subsequent changes. Would you be up to writing one?

arthur-adjedj:
> That said, a benchmark for the generated code would of course be great to have more assurance with subsequent changes. Would you be up to writing one?

I'm not sure how I should go about writing such a benchmark, I have yet to write one... any indications are welcome :)
I'm also not certain of my available bandwidth to work on this in the near-future, but if it's not too time-consuming, I can probably work on that soon enough.

nomeata:
I also wouldn't know. And not a requirement, just a suggestion :-)

PR #5309, opened 2024-09-11
chore: remove @[simp] from forall_const

kim-em:


semorrison:
This doesn't give much performance improvement over Mathlib, so is a bad idea.

PR #5317, opened 2024-09-12
chore: rename 13 simp lemmas

Parcly-Taxel:
Add `_eq` to the end of `(and,or,iff)_(true,false)`, `(true,false)_(and,or,iff)` and `iff_self`.

This resolves a TODO in [Mathlib](https://github.com/leanprover-community/mathlib4/blob/216c754008de909a8ecc89695c745846c9b736a3/Mathlib/Init/Logic.lean#L72).

### DIFF
--- a/src/Init/SimpLemmas.lean
+++ b/src/Init/SimpLemmas.lean
@@ -101,13 +101,13 @@ theorem dite_cond_eq_false {α : Sort u} {c : Prop} {_ : Decidable c} {t : c →
 end SimprocHelperLemmas
 @[simp] theorem ite_self {α : Sort u} {c : Prop} {d : Decidable c} (a : α) : ite c a a = a := by cases d <;> rfl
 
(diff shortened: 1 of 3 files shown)


YaelDillies:
Can you also rename `forall_congr` to `forall_congr_eq`?

src/Init/SimpLemmas.lean
```diff
@@ -101,13 +101,13 @@ theorem dite_cond_eq_false {α : Sort u} {c : Prop} {_ : Decidable c} {t : c →
 end SimprocHelperLemmas
 @[simp] theorem ite_self {α : Sort u} {c : Prop} {d : Decidable c} (a : α) : ite c a a = a := by cases d <;> rfl
 
-@[simp] theorem and_true (p : Prop) : (p ∧ True) = p := propext ⟨(·.1), (⟨·, trivial⟩)⟩
```
eric-wieser, line 104: Alternatively, should this be changed to be an `Iff` statement instead?

kim-em:

PR #5319, opened 2024-09-12
chore: restrict exists_prop' to Type

kim-em:


semorrison:
Sadly, Mathlib demonstrates this is a bad idea.

PR #5323, opened 2024-09-12
refactor: redefine unsigned fixed width integers in terms of BitVec

hargoniX:
I made a few choices so far that can probably be discussed:
- got rid of `modn` on `UInt`, nobody seems to use it apart from the definition of `shift` which can use normal `mod`
- removed the previous defeq optimized definition of `USize.size` in favor for a normal one. The motivation was to allow `OfNat` to work which doesn't seem to be necessary anymore afaict.
- Minimized uses of `.val`, should we maybe mark it deprecated?
- Mostly got rid of `.val` in basically all theorems as the proper next level of API would now be `.toBitVec`. We could probably re-prove them but it would be more annoying given the change of definition.
- Did not yet redefine `log2` in terms of `BitVec` as this would require a `log2` in `BitVec` as well, do we want this?
- I added a couple of theorems around the relation of `<` on `UInt` and `Nat`. These were previously not needed because defeq was used all over the place to save us. I did not yet generalize these to all types as I wasn't sure if they are the appropriate lemma that we want to have.

### DIFF
--- a/src/Init/Data/BitVec/Basic.lean
+++ b/src/Init/Data/BitVec/Basic.lean
@@ -11,10 +11,10 @@ import Init.Data.Int.Bitwise
 import Init.Data.BitVec.BasicAux
 
 /-!
-We define bitvectors. We choose the `Fin` representation over others for its relative efficiency
-(Lean has special support for `Nat`), alignment with `UIntXY` types which are also represented
-with `Fin`, and the fact that bitwise operations on `Fin` are already defined. Some other possible
-representations are `List Bool`, `{ l : List Bool // l.length = w }`, `Fin w → Bool`.
+We define the basic algebraic structure of bitvectors. We choose the `Fin` representation over
+others for its relative efficiency (Lean has special support for `Nat`),  and the fact that bitwise
+operations on `Fin` are already defined. Some other possible representations are `List Bool`,
+`{ l : List Bool // l.length = w }`, `Fin w → Bool`.
 
 We define many of the bitvector operations from the
 [`QF_BV` logic](https://smtlib.cs.uiowa.edu/logics-all.shtml#QF_BV).
--- a/src/Init/Data/BitVec/BasicAux.lean
+++ b/src/Init/Data/BitVec/BasicAux.lean
@@ -8,6 +8,11 @@ import Init.Data.Fin.Basic
 
 set_option linter.missingDocs true
 
+/-!
+This module exists to provide the very basic `BitVec` definitions required for
+`Init.Data.UInt.BasicAux`.
+-/
+
 namespace BitVec
 
 section Nat
--- a/src/Init/Data/UInt/Lemmas.lean
+++ b/src/Init/Data/UInt/Lemmas.lean
@@ -45,21 +45,13 @@ theorem lt_def {a b : $typeName} : a < b ↔ a.toBitVec < b.toBitVec := .rfl
 
 @[simp] protected theorem lt_irrefl (a : $typeName) : ¬ a < a := by simp
 
-protected theorem le_trans {a b c : $typeName} : a ≤ b → b ≤ c → a ≤ c := by
-  simp only [le_def, lt_def]
-  apply BitVec.le_trans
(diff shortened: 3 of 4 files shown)


semorrison:
Looks great.

Could you leave a deprecation in place for `modn`, just in case there is a downstream user?

Agree to marking `.val` as deprecated, but it can also happen later.

`BitVec.log2` sounds good, but probably deserves a different name, and can happen later.

Would you mind rebasing onto `nightly-with-mathlib` so we can check to see how much we have disturbed downstream. (Ideally nothing, but I'm be pleasantly surprised if that were the case.)

semorrison:
(I'm keen to have this in, as I'm thinking about `Fin` refactors.)

src/Init/Data/BitVec/Basic.lean
```diff
@@ -8,6 +8,7 @@ import Init.Data.Fin.Basic
 import Init.Data.Nat.Bitwise.Lemmas
 import Init.Data.Nat.Power2
 import Init.Data.Int.Bitwise
+import Init.Data.BitVec.BasicAux
 
 /-!
 We define bitvectors. We choose the `Fin` representation over others for its relative efficiency
```
TwoFX, line 14: There is a sentence in this module doc which is no longer true.

src/Init/Data/BitVec/BasicAux.lean
```diff
@@ -0,0 +1,47 @@
+/-
+Copyright (c) 2024 Lean FRO, LLC. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Joe Hendrix, Wojciech Nawrocki, Leonardo de Moura, Mario Carneiro, Alex Keizer, Harun Khan, Abdalrhman M Mohamed
+-/
+prelude
+import Init.Data.Fin.Basic
+
```
TwoFX, line 8: Could you add a module doc to this file explaining why the `Basic`/`BasicAux` split is needed in this case? Probably more or less the contents of https://lean-fro.zulipchat.com/#narrow/stream/398861-general/topic/Reworking.20import.20structure.20in.20Init/near/467753170

src/Init/Data/UInt/Lemmas.lean
```diff
@@ -17,48 +19,108 @@ instance : Inhabited $typeName where
...
-protected theorem val_eq_of_eq {a b : $typeName} (h : a = b) : a.val = b.val := h ▸ rfl
-protected theorem eq_of_val_eq {a b : $typeName} (h : a.val = b.val) : a = b := by cases a; cases b; simp at h; simp [h]
-open $typeName (val_eq_of_eq) in
-protected theorem ne_of_val_ne {a b : $typeName} (h : a.val ≠ b.val) : a ≠ b := fun h' => absurd (val_eq_of_eq h') h
-open $typeName (ne_of_val_ne) in
-protected theorem ne_of_lt {a b : $typeName} (h : a < b) : a ≠ b := ne_of_val_ne (Fin.ne_of_lt h)
+
+protected theorem le_trans {a b c : $typeName} : a ≤ b → b ≤ c → a ≤ c := by
+  simp only [le_def, lt_def]
+  apply BitVec.le_trans
```
TwoFX, line 50: This is fine, but
```suggestion
protected theorem le_trans {a b c : $typeName} : a ≤ b → b ≤ c → a ≤ c := BitVec.le_trans
```
would have been good also.

TwoFX (approved):
Looks good!

src/Init/Data/BitVec/Basic.lean
bollu, line 14:
```suggestion
We define the basic algebraic structure of bitvectors.
```

hargoniX:
I addressed all comments and everything in batteries and mathlib apart from a singular sorry here: https://github.com/leanprover-community/mathlib4/blob/lean-pr-testing-5323/Mathlib/Data/UInt.lean#L77 is fixed. This was previously `rfl` but is now not `rfl` anymore. It also appears there is very little API around this thing so I don't quite see how to do a nice proof here without definition bashing.

PR #5329, opened 2024-09-13
refactor: tag `Iff.refl` with `@[refl]`

Parcly-Taxel:
…and remove `exact Iff.rfl` from the `rfl` macro.

This upstreams a property found in [`Mathlib.Init.Logic`](https://github.com/leanprover-community/mathlib4/blob/4e40837aec257729b3c38dc3e635fc64a7a8a8c1/Mathlib/Init/Logic.lean#L63).

### DIFF
--- a/tests/lean/run/variable.lean
+++ b/tests/lean/run/variable.lean
@@ -138,6 +138,7 @@ variable [ToString α] in
 omit [ToString α] in
 theorem t13 (a : α) : toString a = toString a := rfl
 
+set_option pp.mvars false in
 /--
 error: application type mismatch
   ToString True
@@ -146,7 +147,7 @@ argument
 has type
   Prop : Type
 but is expected to have type
-  Type ?u.1754 : Type (?u.1754 + 1)
+  Type _ : Type (_ + 1)
 -/
 #guard_msgs in
 omit [ToString True]


nomeata:
Part of https://github.com/leanprover/lean4/pull/3714 as well, so 👍🏻 from me.

nomeata:
But shouldn't that make this line obsolete?

https://github.com/leanprover/lean4/blob/2080fc022187595020d519c1e08aa72340c4b5ff/src/Init/Core.lean#L826

Parcly-Taxel:
> But shouldn't that make this line obsolete?
> 
> https://github.com/leanprover/lean4/blob/2080fc022187595020d519c1e08aa72340c4b5ff/src/Init/Core.lean#L826

Note that that line has Iff.rfl, not Iff.r**e**fl.

nomeata:
That shouldn’t matter, they differ only in whether the argument is implicit or explicitly. Both this line and your attribute have the effect “`rfl` can solve `Iff` goals”, so one should suffice, shouldn’t it?

Parcly-Taxel:
> That shouldn’t matter, they differ only in whether the argument is implicit or explicitly. Both this line and your attribute have the effect “`rfl` can solve `Iff` goals”, so one should suffice, shouldn’t it?

I tried removing the aforementioned `@[refl]` attribute in mathlib. One lemma in `ModelTheory.Semantics` seems to actually require that attribute – as it is currently written.

nomeata:
Moving the attribute to core is the right thing to do – but at the same time, I assume we should be removing 
```
 macro_rules | `(tactic| rfl) => `(tactic| exact Iff.rfl) 
```
in Core, else `Iff.rfl` is tried twice whenever someone types `rfl`.


It would be good to understand why `ModelTheory.Semantics` needed the attribute. Is it using the `apply_rfl` tactic directly instead of the `rfl` tactic (unlikely). Else probably some transparency issue.

Parcly-Taxel:
> It would be good to understand why `ModelTheory.Semantics` needed the attribute while we (also) have the macro.

I made a new branch [`untag-refl`](https://github.com/leanprover-community/mathlib4/tree/untag-refl) to see what happens when the mathlib `@[refl]` attribute is removed. ([Build log here.](https://github.com/leanprover-community/mathlib4/actions/runs/10845829191/job/30097497405))

The failure in `ModelTheory.Semantics` is
```lean
@[simp]
theorem realize_relabel {m n : ℕ} {φ : L.BoundedFormula α n} {g : α → β ⊕ (Fin m)} {v : β → M}
    {xs : Fin (m + n) → M} :
    (φ.relabel g).Realize v xs ↔
      φ.Realize (Sum.elim v (xs ∘ Fin.castAdd n) ∘ g) (xs ∘ Fin.natAdd m) := by
  rw [relabel, realize_mapTermRel_add_castLe] <;> intros <;> simp
```

The failures are in four of the five `simp` calls at the end; `realize_mapTermRel_add_castLe` produces five subgoals. Changing the proof to just `apply realize_mapTermRel_add_castLe <;> simp` resolves all errors.

nomeata:
Oh, I remember that lemma, I think I had to deal with it in #3714, too.

Note that the macro above says `exact Iff.rfl`. This is not the same as `apply Iff.rfl`! The later runs `defEq` in “approximate mode”, which allows it to instantiate some higher order meta variables. And `apply_rfl` (i.e. the tactic that uses the `@[refl]` attribute) uses `apply` not `exact`. This explains why that lemma needs the attribute, and the macro isn’t enough. All very tricky.

Anyways, thanks for checking, so back to the present PR: If you add `@[refl]` to `Iff.refl` here, you should be able to remove
```
macro_rules | `(tactic| rfl) => `(tactic| exact Iff.rfl) 
```
at the same time, right? 

(Maybe it breaks proofs that use `rfl` on goals with relation that are not reducible but semi-reducible and reduce to `Iff` – but we have to check to know.)

Parcly-Taxel:
@nomeata OK, mathlib builds now.

PR #5332, opened 2024-09-13
chore: notation ^^ for Bool.xor

kim-em:
Not sure why this had been missing. Precedence is slightly higher than `||`, matching the precedence order we have for bitwise operators.


digama0:
FWIW C also does not have a boolean XOR. And as markus says we have `!=` already for a defeq version of Bool.xor.

semorrison:
Happy to take advice here. Options:
1. Make this change
2. Do nothing
3. Start ripping out all the existing uses of `Bool.xor` in favour of `Bool.bne`. 

(My preference order is 1>2>3.)

semorrison:
@TwoFX, shall I proceed with this, or abandon it? (Or think about replacing xor with bne??)

tobiasgrosser (approved):

TwoFX:
@semorrison I vote for proceeding with this.

PR #5338, opened 2024-09-13
perf: do not lint unused variables defined in tactics by default

Kha:
Should ensure we visit at most as many expr nodes as in the final expr instead of many possibly overlapping mvar assignments. This is likely the only way we can ensure acceptable performance in all cases.

### DIFF
--- a/src/Lean/Linter/UnusedVariables.lean
+++ b/src/Lean/Linter/UnusedVariables.lean
@@ -10,41 +10,55 @@ set_option linter.missingDocs true -- keep it documented
 
 /-! # Unused variable Linter
 
-This file implements the unused variable linter, which runs automatically on all commands
-and reports any local variables that are never referred to, using information from the info tree.
+This file implements the unused variable linter, which runs automatically on all
+commands and reports any local variables that are never referred to, using
+information from the info tree.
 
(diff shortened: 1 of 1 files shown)


src/Lean/Server/InfoUtils.lean
```diff
@@ -40,25 +40,30 @@ structure InfoWithCtx where
...
     (postNode : ContextInfo → Info → (children : PersistentArray InfoTree) → List (Option α) → m α)
     : InfoTree → m (Option α) :=
   go none
 where go
   | ctx?, context ctx t => go (ctx.mergeIntoOuter? ctx?) t
   | some ctx, node i cs => do
-    preNode ctx i cs
-    let as ← cs.toList.mapM (go <| i.updateContext? ctx)
-    postNode ctx i cs as
+    if !(← preNode ctx i cs) then
```
mhuisi, line 56:
```suggestion
    let visitChildren ← preNode ctx i cs
    if !visitChildren then
```

mhuisi (approved):

kim-em:
What is the plan here? It generates a massive number of warnings in Mathlib.

1. Reconsider the design?
2. Resolve them all in Mathlib. There are enough that it is probably worth automating.

kim-em (approved):

PR #5358, opened 2024-09-16
chore: introduce BitVec.setWidth to unify zeroExtend and truncate

tobiasgrosser:
We always simplify to setWidth to minimize the API surface.


tobiasgrosser:
This works, but for some reason, the simprocs for setWidth and family do not fire.

tobiasgrosser:
> This works, but for some reason, the simprocs for setWidth and family do not fire.

It required a stage-0 update. Now everything works.

src/Init/Data/BitVec/Basic.lean
```diff
@@ -456,7 +456,7 @@ def extractLsb (hi lo : Nat) (x : BitVec n) : BitVec (hi - lo + 1) := extractLsb
 /--
 A version of `zeroExtend` that requires a proof, but is a noop.
```
semorrison, line 457:
```suggestion
A version of `setWidth` that requires a proof, but is a noop.
```

semorrison:
@alexkeizer, @bollu, @TwoFX, any thoughts on this?

I like this (indeed, suggested it!) as we've been accumulating inconsistently duplicated API for `zeroExtend` and `truncate`. After this change, we will just be simplifying both of those abbreviations to `setWidth`, and having a single verification API.

Moreover, we can further (I think, but we'll experiment in a later PR) replace `cast (h : w = v)` with `setWidth v`.

semorrison (approved):

src/Init/Data/BitVec/Basic.lean
```diff
@@ -475,20 +475,26 @@ def shiftLeftZeroExtend (msbs : BitVec w) (m : Nat) : BitVec (w + m) :=
 /--
 Zero extend vector `x` of length `w` by adding zeros in the high bits until it has length `v`.
 If `v < w` then it truncates the high bits instead.
```
alexkeizer, line 477: I'd suggest we rephrase the doc a bit, to make sure we don't emphasize zeroExtension over truncation

```suggestion
/--
Transform `x` of length `w` into a bitvector of length `v`, by either:
- zero extending, that is, adding zeros in the high bits until it has length `v`, if `v > w`, or
- truncating the high bits, if `v < w`.
```

src/Init/Data/BitVec/Basic.lean
```diff
@@ -475,20 +475,26 @@ def shiftLeftZeroExtend (msbs : BitVec w) (m : Nat) : BitVec (w + m) :=
...
   else
     .ofNat v x.toNat
 
+/--
+Zero extend vector `x` of length `w` by adding zeros in the high bits until it has length `v`.
+If `v < w` then it truncates the high bits instead.
+
+SMT-Lib name: `zero_extend`.
+-/
+abbrev zeroExtend := @setWidth
```
alexkeizer, line 491:
```suggestion
@[inherit_doc]
abbrev zeroExtend := @setWidth
```

By inheriting the doc, it becomes more obvious that zeroExtend can also truncate, and vice-versa

src/Init/Data/BitVec/Basic.lean
```diff
@@ -475,20 +475,26 @@ def shiftLeftZeroExtend (msbs : BitVec w) (m : Nat) : BitVec (w + m) :=
...
+
 /--
 Truncate the high bits of bitvector `x` of length `w`, resulting in a vector of length `v`.
 If `v > w` then it zero-extends the vector instead.
 -/
```
alexkeizer, line 496:
```suggestion
@[inherit_doc]
```

alexkeizer (commented):
I like it! I've also been a bit bothered with the API duplication, picking a more neutrally named simp normal form makes a lot of sense

alexkeizer:
> Moreover, we can further (I think, but we'll experiment in a later PR) replace cast (h : w = v) with setWidth v.

I do have my reservation about this, won't it lose information? For example, could the following still work with such a redefinition?
```
example {w v : Nat} {x : BitVec w} (h : w = v) : 
    (x.cast h).cast h.symm = x := by
  simp
```

src/Init/Data/BitVec/Basic.lean
```diff
@@ -472,23 +472,30 @@ def shiftLeftZeroExtend (msbs : BitVec w) (m : Nat) : BitVec (w + m) :=
         exact (Nat.two_pow_pos m)
   (msbs.toNat <<< m)#'(shiftLeftLt msbs.isLt m)
 
+/--
+Transform `x` of length `w` into a bitvector of length `v`, by either:
+- zero extending, that is, adding zeros in the high bits until it has length `v`, if `v > w`, or
+- truncating the high bits, if `v < w`.
+-/
+def setWidth (v : Nat) (x : BitVec w) : BitVec v :=
```
tobiasgrosser, line 480:
```suggestion
- truncating the high bits, if `v < w`.

SMT-Lib name: `zero_extend`.
-/
def setWidth (v : Nat) (x : BitVec w) : BitVec v :=
```

src/Init/Data/BitVec/Basic.lean
tobiasgrosser: I had to remove @[inherit_doc] as it does not seem to be available at this point.

PR #5359, opened 2024-09-16
chore: ensure that the `rfl` tactic tries `Iff.rfl`

Parcly-Taxel:
Revert the removal of the macro containing `Iff.rfl` in #5329; it was causing errors in leanprover-community/mathlib4#16839.

### DIFF
--- a/src/Init/Core.lean
+++ b/src/Init/Core.lean
@@ -817,11 +817,13 @@ variable {a b c d : Prop}
 theorem iff_iff_implies_and_implies {a b : Prop} : (a ↔ b) ↔ (a → b) ∧ (b → a) :=
   Iff.intro (fun h => And.intro h.mp h.mpr) (fun h => Iff.intro h.left h.right)
 
-theorem Iff.refl (a : Prop) : a ↔ a :=
+@[refl] theorem Iff.refl (a : Prop) : a ↔ a :=
   Iff.intro (fun h => h) (fun h => h)
 
-@[refl] protected theorem Iff.rfl {a : Prop} : a ↔ a :=
+protected theorem Iff.rfl {a : Prop} : a ↔ a :=
   Iff.refl a
+
+macro_rules | `(tactic| rfl) => `(tactic| exact Iff.rfl)
 
 theorem Iff.of_eq (h : a = b) : a ↔ b := h ▸ Iff.rfl
 
--- a/tests/lean/run/5359.lean
+++ b/tests/lean/run/5359.lean
@@ -1,0 +1,19 @@
+/-!
+Test that `Iff.rfl` is tried by the `rfl` tactic.
+-/
+universe u v w
+
+class L (F : Sort u) (α : outParam (Sort v)) (β : outParam (α → Sort w)) where
+  coe : F → ∀ a : α, β a
+
+instance {F : Sort u} {α : Sort v} {β : α → Sort w} [L F α β] :
+    CoeFun F (fun _ ↦ ∀ a : α, β a) where coe := @L.coe _ _ β _
+
+instance {π : Nat → Type u} [∀ i, LE (π i)] : LE (∀ i, π i) where le x y := ∀ i, x i ≤ y i
+
+structure S (α : Nat → Type u) where
+variable {α : Nat → Type u} [∀ i, LE (α i)]
+instance : L (S α) Nat α := sorry
+instance : LE (S α) := ⟨fun f g ↦ ∀ i, f i ≤ g i⟩
+
+example : ∀ {a b : S α}, L.coe a ≤ L.coe b ↔ a ≤ b := by rfl


nomeata:
Just curious: When does the difference matter?

Parcly-Taxel:
> Just curious: When does the difference matter?

When it's in leanprover-community/mathlib4#16839. And I needed _no_ adaptation of mathlib at all here, unlike #5329.

nomeata:
Hmm, but can you explain why it makes a difference? Because I’d expect the argument is solved by unification with the goal in any case.

Compare (e.g. on https://live.lean-lang.org/, without mathlib and without #5329 yet) that both these work:

```lean
attribute [refl] Iff.refl

example : True ↔ True := by apply_rfl
```


```lean
attribute [refl] Iff.refl

example : True ↔ True := by apply_rfl
```

Parcly-Taxel:
> Hmm, but can you explain why it makes a difference? Because I’d expect the argument is solved by unification with the goal in any case.

For the simple reason that `Iff.refl` takes one more explicit argument. If I try to change the fragment in `Data.DFinsupp.Order` to
```lean
/-- The order on `DFinsupp`s over a partial order embeds into the order on functions -/
def orderEmbeddingToFun : (Π₀ i, α i) ↪o ∀ i, α i where
  toFun := DFunLike.coe
  inj' := DFunLike.coe_injective
  map_rel_iff' := by exact Iff.refl
```

I get the very explanatory error
```
type mismatch
  Iff.refl
has type
  ∀ (a : Prop), a ↔ a : Prop
but is expected to have type
  { toFun := DFunLike.coe, inj' := ⋯ } a✝ ≤ { toFun := DFunLike.coe, inj' := ⋯ } b✝ ↔ a✝ ≤ b✝ : Prop
the following variables have been introduced by the implicit lambda feature
  a✝ : Π₀ (i : ι), α i
  b✝ : Π₀ (i : ι), α i
you can disable implicit lambdas using `@` or writing a lambda expression with `{}` or `[]` binder annotations.
```

nomeata:
Ah, but `apply_rfl` uses `apply`, not `exact`, doesn’t it?

https://github.com/leanprover/lean4/blob/45af92fcd1846d7ec61125ab0fbd939f38ca68fe/src/Lean/Meta/Tactic/Rfl.lean#L68

(Sorry for being an annoying reviewer her, but I’d just make sure we are not missing something here. Also in light of the imminent changes to the `rfl` tactic in https://github.com/leanprover/lean4/pull/3714.)

Parcly-Taxel:
@nomeata `apply` also fails.
```lean
/-- The order on `DFinsupp`s over a partial order embeds into the order on functions -/
def orderEmbeddingToFun : (Π₀ i, α i) ↪o ∀ i, α i where
  toFun := DFunLike.coe
  inj' := DFunLike.coe_injective
  map_rel_iff' := by apply Iff.refl
```

```
  ∀ (a : Prop), a ↔ a
with
  ∀ {a b : Π₀ (i : ι), α i}, { toFun := DFunLike.coe, inj' := ⋯ } a ≤ { toFun := DFunLike.coe, inj' := ⋯ } b ↔ a ≤ b
```

nomeata:
Interesting, so it's the implicit variables in the goal that matter here! That might explain some of the adaption changes I had to do for my changes to `rfl`, that's good to know.

If with `apply`, neither works, are you sure that this PR fixes the problems? The comment in the mathlib PR that you linked said “hopeful”.

Maybe a test case should go with this PR to test this corner case behavior, and to make sure it isn't broken again in the future.

Parcly-Taxel:
> Maybe a test case should go with this PR to test this corner case behavior, and to make sure it isn't broken again in the future.

How am I to write a test case? What should I put in there?

Parcly-Taxel:
@nomeata I'm still confused. On current master this is what is working (without a `@[refl]` attribute on `Iff.refl` in mathlib):
```lean
/-- The order on `DFinsupp`s over a partial order embeds into the order on functions -/
def orderEmbeddingToFun : (Π₀ i, α i) ↪o ∀ i, α i where
  toFun := DFunLike.coe
  inj' := DFunLike.coe_injective
  map_rel_iff' := by rfl
```
In the testing branch for this PR it actually breaks.

Parcly-Taxel:
> Now I am confused, then what is this PR fixing?

In short, you were wrong to make me remove the macro in #5329.

nomeata:
Hmm, yes, that may be. Or maybe I was wrong to rely on “OK, mathlib builds now. The necessary modifications were very small.”

In any case, clearly the lean4 test suite is insufficient, else it is hard to tell which behaviours are accidential and which are a feature. Thanks for condensing the issue!

nomeata:
Thanks for the test case! That helped to understand the issue better; the underlying problem here is that `exact` is in a way more powerful than `apply` that I would consider surprising. I’ve opened https://github.com/leanprover/lean4/issues/5366 to discuss.

semorrison:
@nomeata, do you think it is reasonable to proceed with this PR in the meantime? It would allow us cleaning up the four regressions on Mathlib's nightlty-testing.

nomeata:
Hmm, wouldn’t it be better to understand if they are really regressions, or use of an “accidental feature” first?

nomeata:
But fair enough, we can change later. If `rfl` ends up using `exact` then some other proofs will break.

PR #5360, opened 2024-09-16
feat: lemmas about List.tail

kim-em:


src/Init/Data/List/Attach.lean
```diff
@@ -492,4 +488,22 @@ theorem getLast_attach {xs : List α} (h : xs.attach ≠ []) :
     xs.attach.getLast h = ⟨xs.getLast (by simpa using h), getLast_mem (by simpa using h)⟩ := by
   simp only [getLast_eq_head_reverse, reverse_attach, head_map, head_attach]
 
+theorem countP_attach (l : List α) (p : α → Bool) :
+    l.attach.countP (fun a : {x // x ∈ l} => p a) = l.countP p := by
+  simp only [← Function.comp_apply (g := Subtype.val), ← countP_map, attach_map_subtype_val]
+
```
nomeata, line 494: No `simp`?

src/Init/Data/List/Nat/Count.lean
```diff
@@ -28,4 +28,59 @@ theorem count_set [BEq α] (a b : α) (l : List α) (i : Nat) (h : i < l.length)
     (l.set i a).count b = l.count b - (if l[i] == b then 1 else 0) + (if a == b then 1 else 0) := by
   simp [count_eq_countP, countP_set, h]
 
+/--
+The number of elements satisfying a predicate in a sublist is at least the number of elements satisfying the predicate in the list,
+minus the difference in the lengths.
+-/
+theorem Sublist.le_countP (s : l₁ <+ l₂) (p) : countP p l₂ - (l₂.length - l₁.length) ≤ countP p l₁ := by
```
nomeata, line 35: Maybe avoiding `Nat.sub`, since that is the general advise, isn’t it?
```suggestion
theorem Sublist.le_countP (s : l₁ <+ l₂) (p) : countP p l₂ + l₁.length ≤ countP p l₁ + l₂.length := by
```
(This is based on a suggestion by @jcommelin on a somewhat similar lemma of mine in https://github.com/leanprover-community/mathlib4/pull/15383#discussion_r1718761193)

But I agree that reads much less natural that way.

semorrison: Yeah, I trust in `omega` for things like this.

PR #5374, opened 2024-09-17
feat: add Nonempty instances for products

david-christiansen:
After #5270, `partial` functions that use products of sums no longer compile with only `Nonempty` constraints on their arguments. These instances allow the compilation to work.


jcommelin (approved):
LGTM

PR #5375, opened 2024-09-17
feat: add `bv_decide` normalization rules for `ofBool (a.getLsbD i)` and `ofBool a[i]`

alexkeizer:
In LNSym we often use the pattern `ofBool (a.getLsbD i)` to pick out a specific bit (`i`) from a bitvector (`a`).

By adding a rewrite to `extractLsb` to `bv_decide`s normalization set, we can still automatically close goals that have this pattern. In the process, I also added a simp-lemma about the value of a `Fin 1`.

### DIFF
--- a/src/Std/Tactic/BVDecide/Normalize/BitVec.lean
+++ b/src/Std/Tactic/BVDecide/Normalize/BitVec.lean
@@ -215,7 +215,7 @@ theorem BitVec.ofBool_getLsbD (a : BitVec w) (i : Nat) :
 @[bv_normalize]
 theorem BitVec.ofBool_getElem (a : BitVec w) (i : Nat) (h : i < w) :
     BitVec.ofBool a[i] = a.extractLsb' i 1 := by
-  rw [show a[i] = a.getLsbD i from rfl]
+  rw [← BitVec.getLsbD_eq_getElem]
   apply ofBool_getLsbD
 
 end Normalize


src/Std/Tactic/BVDecide/Normalize/BitVec.lean
```diff
@@ -206,5 +206,17 @@ theorem BitVec.max_ult' (a : BitVec w) : (BitVec.ult (-1#w) a) = false := by
...
+@[bv_normalize]
+theorem BitVec.ofBool_getLsbD (a : BitVec w) (i : Nat) :
+    BitVec.ofBool (a.getLsbD i) = a.extractLsb' i 1 := by
+  ext j
+  simp
+
+@[bv_normalize]
+theorem BitVec.ofBool_getElem (a : BitVec w) (i : Nat) (h : i < w) :
+    BitVec.ofBool a[i] = a.extractLsb' i 1 := by
+  rw [show a[i] = a.getLsbD i from rfl]
```
hargoniX, line 218: There should probably exist a lemma for this? If it doesn't I think we should add it?

tests/lean/run/bv_extract.lean
```diff
@@ -13,3 +13,15 @@ theorem bv_extract_3 (h : x = 1#64) : extractLsb 63 32 x = 0#32 := by
...
 theorem bv_extract_4 (h : x = 1#64) : extractLsb 31 0 x = 1#32 := by
   bv_decide
+
+theorem bv_ofBool_1 (h : x = 1#64) : ofBool (x.getLsbD 0) = 1#1 := by
+  bv_decide
+
+theorem bv_ofBool_2 (h : x = 1#64) : ofBool (x.getLsbD 1) = 0#1 := by
+  bv_decide
+
+theorem bv_ofBool_3 (h : x = 1#64) : ofBool (x[0]) = 1#1 := by
```
hargoniX, line 23: No need for parenthesis here I think?

alexkeizer: Indeed not

src/Std/Tactic/BVDecide/Normalize/BitVec.lean
alexkeizer: You're right, I didn't find it cause my master was outdated.

I did golf the proof of the lemma while I was at it: it used to involve `simp`, but `rfl` is enough.

alexkeizer:
@hargoniX addressed your comments, another round of review please!

tobiasgrosser (approved):
Nice. Thank you.

hargoniX (approved):

PR #5376, opened 2024-09-18
fix: modify projection instance binder info

### LINKED ISSUE #5333: modifying TC synth order can give 2% speedup of mathlib
The mathlib PR https://github.com/leanprover-community/mathlib4/pull/16646 gives a 2% speedup of mathlib, coming from a 6% speedup of TC synth.
This is achieved by changing the TC synth order of certain instances, using mechanisms that were likely not intended as user-facing API, and deviating from the documented behaviour of TC synth.
(The PR was merged into mathlib, but later reverted, in order to allow for more research on this topic.)

The following MWE displays behaviour that we believe is analogous to what happens with mathlib's algebraic hierarchy.
(Code snippet adapted from https://leanprover.zulipchat.com/#narrow/stream/287929-mathlib4/topic/Ways.20to.20speed.20up.20Mathlib/near/469291144)
```lean
import Lean.Elab.Command
import Lean.Elab.MatchExpr
import Lean.Meta.Instances

-- # Example hierarchy

class A (n : Nat) where

instance [A n] : A n.succ where

class B [A 20050] where

class C [A 20000] extends B where

instance : A 20050 where

class D where

instance inst1 : D where
instance inst2 [B] : D where

-- # TC synth times out by default

/--
error: (deterministic) timeout at `typeclass`, maximum number of heartbeats (20000) has been reached
Use `set_option synthInstance.maxHeartbeats <num>` to set the limit.
Additional diagnostic information may be available using the `set_option diagnostics true` command.
-/
#guard_msgs in
#synth D
```

The beginning of the trace looks like
```
[Meta.synthInstance] 💥️ D ▼
  [] new goal D ▼
  [instances] #[inst1, @inst2] 
  [] ✅️ apply @inst2 to D ▼
  [tryResolve] ✅️ D ≟ D 
  [] new goal B ▼
  [instances] #[@C.toB] 
  [] ✅️ apply @C.toB to B ▼
  [tryResolve] ✅️ B ≟ B 
  [] new goal A 20000 ▼
  [instances] #[@instASucc] 
  [] ✅️ apply @instASucc to A 20000 ▼
  [tryResolve] ✅️ A 20000 ≟ A (Nat.succ 19999) 
  [] new goal A 19999 ▼
  [instances] #[@instASucc] 
  [] ✅️ apply @instASucc to A 19999 ▼
  [tryResolve] ✅️ A 19999 ≟ A (Nat.succ 19998) 
  [] new goal A 19998 ▶
  [] ✅️ apply @instASucc to A 19998 ▶
  [] ✅️ apply @instASucc to A 19997 ▶
<... snip>
```

When Lean applies `C.toB` it first attempts to find an instance for `A 20000`, and times out while doing so. So it never gets to look for an instance of `C`, which does not exist.

```lean
-- # Meta code to specify synth order

namespace Lean.Meta

/-- Specify the synthesization order for instances. -/
def Instances.setSynthOrder (declName : Name) (synthOrder : Array Nat) : CoreM Unit := do
  let some entry := Meta.instanceExtension.getState (← getEnv) |>.instanceNames.find? declName |
    throwError "'{declName}' does not have [instance] attribute"
  instanceExtension.add { entry with synthOrder } entry.attrKind

end Lean.Meta

namespace Lean.Elab.Command

open Meta

/-- Specify the synthesization order for instances. -/
elab "set_synth_order " name:ident synthOrder:term : command => do
  let q := Term.MatchExpr.toDoubleQuotedName name
  elabCommand (← `(command| run_meta Instances.setSynthOrder $q $synthOrder))

set_synth_order instAddNat #[]

end Lean.Elab.Command

-- # TC synth succeeds with a new order

set_synth_order C.toB #[1, 0]

/-- info: instD -/
#guard_msgs in
#synth D
```

Now the trace looks like
```
[Meta.synthInstance] ✅️ D ▼
  [] new goal D ▼
  [instances] #[inst1, @inst2] 
  [] ✅️ apply @inst2 to D ▼
  [tryResolve] ✅️ D ≟ D 
  [] new goal B ▼
  [instances] #[@C.toB] 
  [] ✅️ apply @C.toB to B ▼
  [tryResolve] ✅️ B ≟ B 
  [] no instances for C ▼
  [instances] #[] 
  [] ✅️ apply inst1 to D ▼
  [tryResolve] ✅️ D ≟ D 
  [answer] ✅️ D 
  [] result inst1 
```
where Lean immediately bails on `C.toB` and continues with `inst1`.

leodemoura:
closes #5333

This PR tries to address issue #5333.

My conjecture is that the binder annotations for `C.toB` and `Algebra.toSMul` are not ideal. `Algebra.toSMul` is one of declarations where the new command `set_synth_order` was used. Both classes, `C` and `Algebra`, are parametric over instances, and in both cases, the issue arises due to projection instances: `C.toB` and `Algebra.toSMul`. Let's focus on the binder annotations for `C.toB`. They are as follows:

```
C.toB [inst : A 20000] [self : @C inst] : @B ...
```

As a projection, it seems odd that `inst` is an instance-implicit argument instead of an implicit one, given that its value is fixed by `self`. We observe the same issue in `Algebra.toSMul`:

```
Algebra.toSMul {R : Type u} {A : Type v} [inst1 : CommSemiring R] [inst2 : Semiring A]
   [self : @Algebra R A inst1 inst2] : SMul R A
```

The PR changes the binder annotations as follows:

```
C.toB {inst : A 20000} [self : @C inst] : @B ...
```

and

```
Algebra.toSMul {R : Type u} {A : Type v} {inst1 : CommSemiring R} {inst2 : Semiring A}
    [self : @Algebra R A inst1 inst2] : SMul R A
```

In both cases, the `set_synth_order` is used to force `self` to be processed first.

In the MWE, there is no instance for `C ...`, and `C.toB` is quickly discarded. I suspect a similar issue occurs when trying to use `Algebra.toSMul`, where there is no `@Algebra R A ... ...`, but Lean spends unnecessary time trying to synthesize `CommSemiring R` and `Semiring A` instances. I believe the new binder annotations make sense, as if there is a way to synthesize `Algebra R A ... ...`, it will tell us how to retrieve the instance-implicit arguments.

TODO: 
- Impact on Mathlib.
- Document changes.

### DIFF
--- a/src/Lean/Meta/Instances.lean
+++ b/src/Lean/Meta/Instances.lean
@@ -114,7 +114,7 @@ For example:
 
 (The type of `inst` must not contain mvars.)
 
-Remark: `projInfo?` is some if the instance is a projection.
+Remark: `projInfo?` is `some` if the instance is a projection.
 We need this information because of the heuristic we use to annotate binder
 information in projections. See PR #5376 and issue #5333. Before PR
 #5376, given a class `C` at


leodemoura:
!bench

mattrobball:
Yes, this is exactly my understanding of the problem in Mathlib and this fix is far more sensible. Thanks @leodemoura !

src/Lean/Meta/Instances.lean
```diff
@@ -113,8 +113,34 @@ For example:
     (because A B are out-params and are only filled in once we synthesize 2)
 
 (The type of `inst` must not contain mvars.)
+
+Remark: `projInfo?` is some if the instance is a projection.
```
jcommelin, line 117:
```suggestion
Remark: `projInfo?` is `some` if the instance is a projection.
```

PR #5399, opened 2024-09-19
fix: upgrade instance synth order issues to hard errors

kmill:
Motivated [by a user's question](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Confusing.20instance.20error/near/471539940), this increases the severity of the "cannot find synthesization order" message from a log error to throwing an exception. This saves some confusion about whether the instance was added or not.

### DIFF
--- a/src/Lean/Meta/Instances.lean
+++ b/src/Lean/Meta/Instances.lean
@@ -208,11 +208,9 @@ private partial def computeSynthOrder (inst : Expr) (projInfo? : Option Projecti
           let typeLines := ("" : MessageData).joinSep <| Array.toList <| ← toSynth.mapM fun i => do
             let ty ← instantiateMVars (← inferType argMVars[i]!)
             return indentExpr (ty.setPPExplicit true)
-          logWarning m!"\
+          throwError m!"\
             cannot find synthesization order for instance {inst} with type{indentExpr instTy}\n\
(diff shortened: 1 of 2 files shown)


TomasPuverle (commented):
My two cents: As the originator of this question/issue, the new wording is more helpful and would have alleviated some of the confusion I expressed on Zulip.
Many thanks!

Kha:
> This isn't a hard error

Just checking, do you think this was intentional and should stay that way?

Kha:
Triage team agrees: making this a hard error seems like the safer option

TomasPuverle:
> Just checking, do you think this was intentional and should stay that way?

I am not sure if this question was aimed at me or @kmill.  I don't have a preference - ultimately, I was just confused when I saw "error: ..." but the code still ran.

kmill:
@Kha Should we leave the `synthInstance.checkSynthOrder` option? My thoughts are
* if we have the option, then since the check can be turned off, it's clearly safe to proceed, so we should log a warning like in this PR; but
* if we want it to be a hard error due to safety, then we shouldn't be able to proceed without the check.

digama0:
I mostly agree with you, but this is not the only error that can be turned off. `set_option hygiene` is another one that comes to mind.

PR #5402, opened 2024-09-20
feat: localize universe metavariable errors at `let` bindings and `fun` binders

### LINKED ISSUE #2058: Confusing error message with unassigned metavariables
```lean
def sameHere : Nat :=
  let x := [] -- invalid occurrence of universe level 'u_1' at 'sameThing', ...
  42
```

Ideally we'd:
 - complain about the unassigned metavariable for the type instead (like we would do if lists weren't universe polymorphic), and
 - add the `sameHere` definition to the environment anyhow (like we would do if lists weren't universe polymorphic).

kmill:
Modifies how the declaration command elaborator reports when there are unassigned metavariables. The visible effects are that (1) now errors like "don't know how to synthesize implicit argument" and "failed to infer 'let' declaration type" take precedence over universe level issues, (2) universe level metavariables are reported as metavariables (rather than as `u_1`, `u_2`, etc.), and (3) if the universe level metavariables appear in `let` binding types or `fun` binder types, the error is localized there.

Motivation: Reporting unsolved expression metavariables is more important than universe level issues (typically universe issues are from unsolved expression metavariables). Furthermore, `let` and `fun` binders can't introduce universe polymorphism, so we can "blame" such bindings for universe metavariables, if possible.

Example 1: Now the errors are on `x` and `none` (reporting expression metavariables) rather than on `example` (which reported universe level metavariables).
```lean
example : IO Unit := do
  let x := none
  pure ()
```

Example 2: Now there is a "failed to infer universe levels in 'let' declaration type" error on `PUnit`.
```lean
def foo : IO Unit := do
  let x : PUnit := PUnit.unit
  pure ()
```

In more detail:
* `elabMutualDef` used to turn all level mvars into fresh level parameters before doing an analysis for "hidden levels". This analysis turns out to be exactly the same as instead creating fresh parameters for level mvars in only pre-definitions' types and then looking for level metavariables in their bodies. With this PR, error messages refer to the same level metavariables in the Infoview, rather than obscure generated `u_1`, `u_2`, ... level parameters.
* This PR made it possible to push the "hidden levels" check into `addPreDefinitions`, after the checks for unassigned expression mvars. It used to be that if the "hidden levels" check produced an "invalid occurrence of universe level" error it would suppress errors for unassigned expression mvars, and now it is the other way around.
* There is now a list of `LevelMVarErrorInfo` objects in the `TermElabM` state. These record expressions that should receive a localized error if they still contain level metavariables. Currently `let` expressions and binder types in general register such info. Error messages make use of a new `exposeLevelMVars` function that adds pretty printer annotations that try to expose all universe level metavariables.
* When there are universe level metavariables, for error recovery the definition is still added to the environment after assigning each metavariable to level 0.
* There's a new `Lean.Util.CollectLevelMVars` module for collecting level metavariables from expressions.

Closes #2058

### DIFF
--- a/src/Lean/Elab/Binders.lean
+++ b/src/Lean/Elab/Binders.lean
@@ -172,7 +172,7 @@ private def toBinderViews (stx : Syntax) : TermElabM (Array BinderView) := do
 
 private def registerFailedToInferBinderTypeInfo (type : Expr) (ref : Syntax) : TermElabM Unit := do
   registerCustomErrorIfMVar type ref "failed to infer binder type"
-  registerLMVarErrorExprInfo type ref m!"failed to infer universe levels in binder type"
+  registerLevelMVarErrorExprInfo type ref m!"failed to infer universe levels in binder type"
 
 def addLocalVarInfo (stx : Syntax) (fvar : Expr) : TermElabM Unit :=
   addTermInfo' (isBinder := true) stx fvar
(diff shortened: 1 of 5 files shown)


Kha:
Are we using `L` for "Level" anywhere else? My first thought would be "Local"

kmill:
@Kha I'm copying LMVarId here. I don't have a strong preference, I could use LevelMVar when there's no Id.

PR #5404, opened 2024-09-20
feat: more of BitVec.getElem_*

tobiasgrosser:


src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -980,6 +1013,10 @@ theorem getLsbD_shiftLeft' {x : BitVec w₁} {y : BitVec w₂} {i : Nat} :
     (x <<< y).getLsbD i = (decide (i < w₁) && !decide (i < y.toNat) && x.getLsbD (i - y.toNat)) := by
   simp [shiftLeft_eq', getLsbD_shiftLeft]
 
+theorem getElem_shiftLeft' {x : BitVec w₁} {y : BitVec w₂} {i : Nat} (h : i < w₁) :
```
semorrison, line 1016: @[simp]?

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1082,6 +1119,15 @@ theorem sshiftRight_eq_of_msb_true {x : BitVec w} {s : Nat} (h : x.msb = true) :
         Nat.not_lt, decide_eq_true_eq]
       omega
 
+@[simp] theorem getElem_sshiftRight {x : BitVec w} {s i : Nat} (h : i < w) :
```
semorrison, line 1122: Skeptical this should be `@[simp]`: generally if the RHS starts with an if we don't let simp proceed.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1219,6 +1265,11 @@ theorem signExtend_eq_not_setWidth_not_of_msb_true {x : BitVec w} {v : Nat} (hms
   · rw [signExtend_eq_not_setWidth_not_of_msb_true hmsb]
     by_cases (i < v) <;> by_cases (i < w) <;> simp_all <;> omega
 
+@[simp] theorem getElem_signExtend {x  : BitVec w} {v i : Nat} (h : i < v) :
```
semorrison, line 1268: similarly here

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1247,6 +1298,13 @@ theorem append_def (x : BitVec v) (y : BitVec w) :
...
+@[simp] theorem getElem_append {x : BitVec n} {y : BitVec m} (h : i < n + m) :
+    (x ++ y)[i] = bif i < m then getLsbD y i else getLsbD x (i - m) := by
+  simp only [append_def, getElem_or, getElem_shiftLeftZeroExtend, getElem_setWidth']
+  by_cases h' : i < m
+  · simp [h']
+  · simp [h']; simp_all
+
 @[simp] theorem getMsbD_append {x : BitVec n} {y : BitVec m} :
     getMsbD (x ++ y) i = bif n ≤ i then getMsbD y (i - n) else getMsbD x i := by
   simp [append_def]
```
semorrison, line 1310: Oof, I see we are already far from consistent here. :-(

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1396,6 +1460,20 @@ theorem toNat_cons' {x : BitVec w} :
     have p2 : i - n ≠ 0 := by omega
     simp [p1, p2, Nat.testBit_bool_to_nat]
 
+@[simp] theorem getElem_cons {b : Bool} {n} {x : BitVec n} {i : Nat} (h : i < n + 1) :
+    (cons b x)[i] = if i = n then b else getLsbD x i := by
```
semorrison, line 1464: Probably remove this one.

src/Init/Data/BitVec/Lemmas.lean
semorrison: Could you check if we can remove `@[simp]` from getMsbD_append?

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: Done.

tobiasgrosser:
I addressed your comments by dropping `@[simp]` from the theorems that have an if on the RHS (and doing the same for corresponding getLsbD theorems). Historically, all getLsbD had simps independent of the RHS. I am happy to not have them simplify if ifs are needed on the RHS.

tobiasgrosser:
awaiting-review

PR #5410, opened 2024-09-20
feat: BitVec analogues of Nat.{mul_two, two_mul, mul_succ, succ_mul}

bollu:
As requested by @hargoniX .

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -1664,6 +1664,15 @@ theorem BitVec.mul_add {x y z : BitVec w} :
   rw [Nat.mul_mod, Nat.mod_mod (y.toNat + z.toNat),
     ← Nat.mul_mod, Nat.mul_add]
 
+theorem mul_succ {x y : BitVec w} : x * (y + 1#w) = x * y + x := by simp [BitVec.mul_add]
+theorem succ_mul {x y : BitVec w} : (x + 1#w) * y = x * y + y := by simp [BitVec.mul_comm, BitVec.mul_add]
+
+theorem mul_two {x : BitVec w} : x * 2#w = x + x := by
+  have : 2#w = 1#w + 1#w := by apply BitVec.eq_of_toNat_eq; simp
+  simp [this, mul_succ]
+
+theorem two_mul {x : BitVec w} : 2#w * x = x + x := by rw [BitVec.mul_comm, mul_two]
+
 @[simp, bv_toNat] theorem toInt_mul (x y : BitVec w) :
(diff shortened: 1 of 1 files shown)


src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1664,6 +1664,15 @@ theorem BitVec.mul_add {x y z : BitVec w} :
   rw [Nat.mul_mod, Nat.mod_mod (y.toNat + z.toNat),
     ← Nat.mul_mod, Nat.mul_add]
 
+theorem mul_succ {x y : BitVec w} : x * (y + 1#w) = x * y + x := by simp [BitVec.mul_add]
```
hargoniX, line 1667: Can we make these binders explicit for consistency with the respective `Nat` theorems?

bollu: It's not totally clear to me, because right above, there's `theorem BitVec.mul_add {x y z : BitVec w} : ...`. I made these implicit following the above, because `x, y` can be determined by the equality.

tobiasgrosser: This is generally treated inconsistently in the `BitVec` file and I would really like this to become consistent. @hargoniX, is there official guidance how to choose here -- at best with an explanation why a certain style is preferred?

semorrison: Generally:
* for `iff` goals any variable that appears on both sides of the `iff` should be implicit
* for `eq` goals, it is less clear. If the lemma will mostly be used for rewriting or closing a goal, variables appearing on both side should be implicit as a reasonable default, but if at a call site you find yourself then needing to use a named argument to specify the value, it is reasonable to make that variable explicit. In an ideal world there would always be a comment about this. We are far from ideal here.
* for other statements, for values in Type only arguments which can be inferred from later arguments should be implicit. For `Prop`, it is again less clear, but generally err more on the side of explicitness compared to `eq` goals.

Sorry this is not as definite as it should be. Guidelines for implicitness have been very hard to pin down.

tobiasgrosser: I guess this means the implicitness proposed by @bollu follows these guidelines and this PR can move forward.

tobiasgrosser (approved):

src/Init/Data/BitVec/Lemmas.lean
semorrison: (Generally, the `Nat` namespace in Lean is not really a good source of code conventions, as much of it was written a long time ago.)

PR #5411, opened 2024-09-20
feat: BitVec.toNat_{add,sub,mul_of_lt} for BitVector non-overflow reasoning

bollu:
These theorems are useful when one wants to simplify the goal state, under knowledge that the bitvector operations don't overflow. This can produce much smaller goal states that eventually allows `bv_omega` to quickly close the goal.

Note that the LHS of the theorem is *not* in `simp` normal form, since e.g. `(x + y).toNat` is normalized to `(x.toNat + y.toNat) % 2^w`. It's not immediately clear to me what should be done about this.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -25,7 +25,7 @@ Please put an X between the brackets as you perform the following steps:
 
 ### Context
 
-[Broader context that the issue occured in. If there was any prior discussion on [the Lean Zulip](https://leanprover.zulipchat.com), link it here as well.]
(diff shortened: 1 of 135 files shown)


tobiasgrosser:
Hey @bollu, which part of the theorem is not in simp normal form, the LHS or the hypothesis? With respect to the hypothesis, I am curious how you turn it into toNat in the first place?

bollu:
@tobiasgrosser updated the message to be more accurate: It's the LHS of the equality that's not in simp normal form.

semorrison:
It's fine if the LHS of a *non*-simp lemma is not in simp normal form --- it just means you'll only be able to rewrite by it prior to calling simp.

I'd consider the argument that `toNat_add` and so on should not be simp lemmas, but this would take some work to make viable, I think.

PR #5415, opened 2024-09-22
feat: implement `Repr Empty`

TomasPuverle:
Given the derived `Repr` instance for types with parameters, the absence of `Repr Empty` can cause `Repr` instance synthesis to fail.  For example, given
```lean
inductive Prim (special : Type) where
  | plus
  | other : special → Prim special
deriving Repr
```
this works:
```lean
#eval (Prim.plus : Prim Int)
```
but this fails:
```lean
#eval (Prim.plus : Prim Empty)
```

### DIFF
--- a/tests/lean/run/repr_empty.lean
+++ b/tests/lean/run/repr_empty.lean
@@ -26,4 +26,4 @@ open Prim in
 /-- info: Prim.minus -/
 #guard_msgs in
 open Prim in
-#eval (minus: Prim Empty)
+#eval (minus : Prim Empty)


tests/lean/repr_empty.lean
```diff
@@ -0,0 +1,22 @@
...
+-/
+
+inductive Prim (special : Type) where
+  | plus
+  | minus
+  | other : special → Prim special
+deriving Repr
+
+-- Check that both Repr's work
+open Prim in
```
kmill, line 18: Please move this test file into `tests/lean/run` and use `#guard_msgs` (you can see examples there).

tests/lean/repr_empty.lean
```diff
@@ -0,0 +1,22 @@
+/-!
+This test validates the Repr instance for `Empty`.
```
kmill, line 2:
```suggestion
# Validation of `Repr Empty` instance
```

src/Init/Data/Repr.lean
```diff
@@ -51,6 +51,12 @@ instance [Repr α] : Repr (id α) :=
 instance [Repr α] : Repr (Id α) :=
   inferInstanceAs (Repr α)
 
+/-
+This instance allows us to use `Empty` as a type parameter without causing automatic
+derivations of `Repr` to fail.
+-/
+deriving instance Repr for Empty
```
kmill, line 58: We may as well write the instance explicitly rather than rely on the deriving machinery.
```suggestion
instance : Repr Empty where
  reprPrec := nofun
```
I'm not sure we need the comment justifying the instance, but I suppose it doesn't hurt.

eric-wieser (commented):
Should we have `ToString` and `ToJson` instances too?

src/Init/Data/Repr.lean
TomasPuverle: Will fix.

Just for my understanding, when would I chose one approach over the other?  It seems `deriving` is used lower down in the same file.  Also, strictly speaking, I wouldn't have known how to write that instance because I'm not familiar with `nofun` (and `nomatch` which comes up when I hover).

I found the definition of `nofun` in `Match.lean` (the elaborator! ;) ) but I can't quite read that code yet.  Is there a short summary/explanation of the semantics you may be able to point me at, please?

TomasPuverle: > I'm not sure we need the comment justifying the instance

Please let me know if you have a preference.  I opted to be more explicit but that's just because I wanted to "do the right thing".  Comments which are obvious to experienced readers are a nuisance, but I'm not the right audience to ask whether the need for `Repr` on `Empty` is obvious or not.

TomasPuverle:
Hi @eric-wieser 

I'd be happy to implement those + tests if you (or someone else) can help with the "risk assessment" and also indicate whether such a PR is likely to get accepted (some of the pages I've read said "ask first").  

Would you care if I did it as a separate PR?  It's good for me to do it in smaller pieces so I can practice and "internalize" the expected development/PR workflows etc.

Thanks!

tests/lean/repr_empty.lean
TomasPuverle: Done.

tests/lean/run/repr_empty.lean
```diff
@@ -0,0 +1,27 @@
...
+The Empty type is used to exlude the `other` constructor from the `Prim` type.
+-/
+
+inductive Prim (special : Type) where
+  | plus
+  | minus
+  | other : special → Prim special
+deriving Repr
+
+-- Check that both Repr's work
```
kmill, line 17: I'm not sure this is written down anywhere, but my understanding of the style is to use module docstrings for explanations of tests. Given that you already wrote about it in the main module docstring, this seems superfluous, but in any case, here's the syntax.
```suggestion
/-!
Check that both Repr's work
-/
```

src/Init/Data/Repr.lean
kmill: Let's leave the comment, it's helpful knowing why we'd have a "useless" instance.

> when would I chose one approach over the other?

You generally want to use `deriving`, but I just figured that this one is so simple that we may as well give the definition. The `nofun` syntax is `fun | patt1 => ... | patt2 => ... | ...` syntax but with no patterns, which itself is short for `fun x => match x with | patt1 => ... | patt2 => ... | ...` syntax. You can also write `fun x => nomatch x` instead of `nofun`.

kmill:
@eric-wieser Yes, but we can't ask TomasPuverle to implement those. I think the scope of this PR is fine as it is.

The immediate motivation for `Repr` is that it's blocking `#eval`. There hasn't been motivation for the other typeclasses yet.

src/Init/Data/Repr.lean
TomasPuverle: Thanks for the thorough explanation.

kmill: Side note: I've updated the PR description to clarify what the change is for. The terminology is "typeclass inference" or "instance synthesis" instead of "deriving" for what causes the issue in the `#eval` command. Deriving is what occurs at the `deriving` clause.

tests/lean/run/repr_empty.lean
TomasPuverle: You're right, it's documented here which is how I knew to write the initial docstring: [Writing good tests](https://github.com/leanprover/lean4/blob/master/doc/dev/testing.md#writing-good-tests).  The second comment was a mistake - I'll be aware in the future.

src/Init/Data/Repr.lean
```diff
@@ -51,6 +51,13 @@ instance [Repr α] : Repr (id α) :=
 instance [Repr α] : Repr (Id α) :=
   inferInstanceAs (Repr α)
 
+/-
+This instance allows us to use `Empty` as a type parameter without causing automatic
+derivations of `Repr` to fail.
```
kmill, line 56: One more suggestion, let's reword this to something like the following
```suggestion
This instance allows us to use `Empty` as a type parameter without causing instance synthesis to fail.
```

src/Init/Data/Repr.lean
TomasPuverle: Noted with thanks.

tests/lean/run/repr_empty.lean
```diff
@@ -0,0 +1,29 @@
...
+deriving Repr
+
+/-!
+Check that both Repr's work
+-/
+
+/-- info: Prim.plus -/
+#guard_msgs in
+open Prim in
+#eval (plus: Prim Int)
```
eric-wieser, line 24: extreme style nit:
```suggestion
#eval (plus : Prim Int)
```

eric-wieser:
> I think the scope of this PR is fine as it is.

Agreed

> There hasn't been motivation for the other typeclasses yet.

Here's one for `Lean.ToJson`:
```lean
import Lean

structure Foo (α : Type) where
  y : Option α
deriving Lean.ToJson

#eval Lean.toJson (⟨some 1⟩ : Foo Nat) -- ok
#eval Lean.toJson (⟨none⟩ : Foo Empty) -- fails
```

Indeed `ToString` seems slightly less relevant, since `deriving` doesn't work with it.

tests/lean/run/repr_empty.lean
TomasPuverle: Indeed, this is "other languages" fingers auto-typing coming through.  Wish Lean had the equivalent of e.g. the 'black" formatter in Python.

TomasPuverle: (I will fix the other case, too)

tests/lean/run/repr_empty.lean
```diff
@@ -0,0 +1,29 @@
...
+
+/-- info: Prim.plus -/
+#guard_msgs in
+open Prim in
+#eval (plus : Prim Int)
+
+/-- info: Prim.minus -/
+#guard_msgs in
+open Prim in
+#eval (minus: Prim Empty)
```
TomasPuverle, line 29:
```suggestion
#eval (minus : Prim Empty)
```

TomasPuverle:
@eric-wieser 

Happy to (try?) take that on next and use your example as a starting point.

However, I am not sure what @kmill meant above by "but we can't ask TomasPuverle to implement those".  (Perhaps his assessment is that it's too involved/risky)?

semorrison:
> @eric-wieser
> 
> Happy to (try?) take that on next and use your example as a starting point.
> 
> However, I am not sure what @kmill meant above by "but we can't ask TomasPuverle to implement those". (Perhaps his assessment is that it's too involved/risky)?

He just meant that it would be unreasonable to ask you to do this. You would be welcome too, however!

kmill:
@TomasPuverle I just meant that we don't need those to make this a complete PR, so we shouldn't be asking you to do it in this PR. Feel free to create another PR with missing instances for `Empty` if you think it would be fun.

For the next ones, there's no need to make tests. The tests in the current PR are good because they're regression tests that the `Empty` instance won't be removed, and they remind a future person why it's there. The primary justification is "we need `#eval` to work for any sensible input" for learning and experimentation. The case for other typeclasses is less strong, but there's no reason *not* to include them, and they'll eventually help someone. However, all you would be testing with any test you could write is that typeclass inference works. We have plenty of tests about typeclass inference.

kmill:
(I know you're the code owner for Init.Data @semorrison — let me know if I shouldn't be merging simple missing instances like this!)

semorrison:
> (I know you're the code owner for Init.Data @semorrison — let me know if I shouldn't be merging simple missing instances like this!)

Please do!

PR #5416, opened 2024-09-22
feat: add LawCommIdentity + IdempotentOp for BitVec.[and|or|xor]

tobiasgrosser:
As these instances seemingly require explicit arguments, this PR also makes some arguments explicit.

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -1586,12 +1586,12 @@ protected theorem add_comm (x y : BitVec n) : x + y = y + x := by
   simp [add_def, Nat.add_comm]
 instance : Std.Commutative (α := BitVec n) (· + ·) := ⟨BitVec.add_comm⟩
 
-@[simp] protected theorem add_zero {x : BitVec n} : x + 0#n = x := by simp [add_def]
+@[simp] protected theorem add_zero (x : BitVec n) : x + 0#n = x := by simp [add_def]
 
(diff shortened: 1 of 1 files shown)


tobiasgrosser:
I feel we really need to understand better under which conditions arguments should be explicit or implicit and then should standardize this across the BitVec library.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -710,14 +710,20 @@ theorem or_comm (x y : BitVec w) :
...
 instance : Std.Commutative (fun (x y : BitVec w) => x ||| y) := ⟨BitVec.or_comm⟩
 
-@[simp] theorem or_self {x : BitVec w} : x ||| x = x := by
+@[simp] theorem or_self (x : BitVec w) : x ||| x = x := by
   ext i
   simp
 
-@[simp] theorem or_zero {x : BitVec w} : x ||| 0#w = x := by
+instance : Std.IdempotentOp (α := BitVec n) (· ||| · ) where
+  idempotent := BitVec.or_self
```
semorrison, line 718: Setting up Lawful classes is not sufficient reason to change the implicitness. You can use, e.g.
```
instance : Std.IdempotentOp (α := BitVec n) (· ||| · ) where
  idempotent _ := BitVec.or_self
```
here with the original implicitness.

tobiasgrosser: OK, that makes sense.

tobiasgrosser:
awaiting-review

PR #5418, opened 2024-09-23
feat: decidable quantifers for BitVec

kim-em:


semorrison:
@hargoniX, if you want to veto this on the basis that one should use `bv_decide`, I won't complain too much.

But:
* this has a smaller trust base
* and handles existential goals.

tobiasgrosser:
Interesting. I can see a small risk of confusion, but it is probably a good idea to have this for completeness.

hargoniX:
This looks fine to me :+1:. I guess people are just going to notice that it doesn't scale well very quickly for non trivial bit widths. It would be cool if we could tell them about `bv_decide` in the error message then but that's non trivial of course.

PR #5419, opened 2024-09-23
fix: must not reduce `ite` in the discriminant of match-expression when reducibility setting is `.reducible`

### LINKED ISSUE #5388: matching on an `if _ then _ else _` gets reduced to `Decidable.rec` by `simp only`, which it then can't further simplify
### Prerequisites

### Description

See the following MWE:
```
example (k : Prop)  (h : k) [Decidable k] : 
    let ⟨a, b⟩ := if k then (1, 0) else (0, 1)
    a = 1 := by
  simp [h]
  /-
  unsolved goals
  k : Prop
  h : k
  inst✝ : Decidable k
  ⊢ (Decidable.rec (fun h => (0, 1)) (fun h => (1, 0)) inst✝).fst = 1
  -/
```

When we match on the result of an `if _ then _ else _` (recalling that `let ⟨a, b⟩ := foo; ...` desugars to `match foo with | ⟨a, b⟩ => ...)`, then even `simp only` will already reduce this to an application of `Decidable.rec`.

This is unfortunate, because our simpset is not equiped to simplify `Decidable.rec`, even if the if-condition is known (as it is in the example), and even if we include a proof that the if-condition is true, this reduction seems to happen first, blocking any further simplification and preventing a seemingly trivial goal from being closed.

### Context

This showed up in LNSym, for example, in [our semantics for a `sub` instruction](https://github.com/leanprover/LNSym/blob/ba4a9016fec5f6d1bb7d8120c1ff3b45e6f7faae/Arm/Insts/DPI/Add_sub_imm.lean#L27)

### Steps to Reproduce

MWE:
```
example (k : Prop)  (h : k) [Decidable k] : 
    let ⟨a, b⟩ := if k then (1, 0) else (0, 1)
    a = 1 := by
  simp [h]
  /-
  unsolved goals
  k : Prop
  h : k
  inst✝ : Decidable k
  ⊢ (Decidable.rec (fun h => (0, 1)) (fun h => (1, 0)) inst✝).fst = 1
  -/
```

**Expected behavior:** I would've expected the goal to be automatically closed by `simp`, seeing that we gave it a proof that the condition of the `if` is true

**Actual behavior:** `simp` is unable to close the goal, getting stuck with 
```
/-
  unsolved goals
  k : Prop
  h : k
  inst✝ : Decidable k
  ⊢ (Decidable.rec (fun h => (0, 1)) (fun h => (1, 0)) inst✝).fst = 1
  -/
```

### Versions

4.12.0-nightly-2024-09-18, at https://live.lean-lang.org/#project=lean-nightly&codez=KYDwhgtgDgNsAEAKA1vAXPACgJwPZQEp4kALdeZIgbQBFgBjASwBMwAjOCgXXIChjicAC7xAF+RgANPDaBL8nQBeeIwBmFeEJLAAdkgCMUgAxFgMAM4JEBqboL8BYeIt0LpATzunG0eFRJc7APQAtHYArlqmuDAAbsDM8ADmuGBmdqgYOPh2ZBjIdowRQoC45OR0TKwcCHnEgEVESGUs7HAAdNgMSMrh8GTyAHxIVvA2RIidOj39iPrwRkQFpsUEzcoLjkN2QQG8dgDEwNEp8AAywGBazbHYnrhaAMpC2AUJQA

### Impact

leodemoura:
closes #5388

See updated comment for additional details.

### DIFF
--- a/src/Init/GetElem.lean
+++ b/src/Init/GetElem.lean
@@ -156,11 +156,11 @@ theorem getElem?_neg [GetElem? cont idx elem dom] [LawfulGetElem cont idx elem d
 theorem getElem!_pos [GetElem? cont idx elem dom] [LawfulGetElem cont idx elem dom]
     [Inhabited elem] (c : cont) (i : idx) (h : dom c i) [Decidable (dom c i)] :
     c[i]! = c[i]'h := by
-  simp only [getElem!_def, getElem?_def, h]
+  simp [getElem!_def, getElem?_def, h]
 
 theorem getElem!_neg [GetElem? cont idx elem dom] [LawfulGetElem cont idx elem dom]
     [Inhabited elem] (c : cont) (i : idx) (h : ¬dom c i) [Decidable (dom c i)] : c[i]! = default := by
-  simp only [getElem!_def, getElem?_def, h]
+  simp [getElem!_def, getElem?_def, h]
 
 namespace Fin
 
--- a/src/Lean/Meta/WHNF.lean
+++ b/src/Lean/Meta/WHNF.lean
@@ -13,6 +13,7 @@ import Lean.Meta.Offset
 import Lean.Meta.CtorRecognizer
 import Lean.Meta.Match.MatcherInfo
 import Lean.Meta.Match.MatchPatternAttr
+import Lean.Meta.Transform
 
 namespace Lean.Meta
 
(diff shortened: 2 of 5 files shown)


leodemoura:
@semorrison @jcommelin This PR will probably break Mathlib at `simp only [...]` occurrences, but the fixes should be similar to the one at [RupAddResult.lean](https://github.com/leanprover/lean4/pull/5419/files#diff-c43b9c516b0aaed55b15b0618c32489738539d05e1e297ef06d60bf44e3b35c4).

semorrison:
Magically no breakages. :-)

PR #5421, opened 2024-09-23
feat: implement `To/FromJSON Empty`

TomasPuverle:
Resolve cases when the `To/FromJSON` type classes are used with `Empty`, e.g. in the following motivating example.

```
import Lean

structure Foo (α : Type) where
  y : Option α
deriving Lean.ToJson

#eval Lean.toJson (⟨none⟩ : Foo Empty) -- fails
```

This is a follow-up to this PR https://github.com/leanprover/lean4/pull/5415, as suggested by @eric-wieser.  It expands on the original suggestion by also handling `FromJSON`.

### DIFF
--- a/src/Lean/Data/Json/FromToJson.lean
+++ b/src/Lean/Data/Json/FromToJson.lean
@@ -30,8 +30,8 @@ instance : ToJson JsonNumber := ⟨Json.num⟩
 
 instance : FromJson Empty where
   fromJson? j := throw (s!"type Empty has no constructor to match JSON value '{j}'. \
-                           deserializing a value for type Empty, e.g. at type \
-                           Option Empty with code for the 'some' constructor.")
+                           This occurs when deserializing a value for type Empty, \
+                           e.g. at type Option Empty with code for the 'some' constructor.")
 
 instance : ToJson Empty := ⟨nofun⟩
 -- looks like id, but there are coercions happening
--- a/tests/lean/run/json_empty.lean
+++ b/tests/lean/run/json_empty.lean
@@ -40,14 +40,14 @@ deriving Lean.ToJson, Lean.FromJson, Repr
 
 /-! Show that parsing fails -/
 /-- info: Except.error "type Empty has no constructor to match JSON value '\"Yo!\"'.
-deserializing a value for type Empty, e.g. at type Option Empty with code for the 'some'
-constructor." -/
+This occurs when deserializing a value for type Empty, e.g. at type Option Empty with code
+for the 'some' constructor." -/
 #guard_msgs in
 #eval Lean.fromJson? (α := Empty) <| json% "Yo!"
 
 /-! Show that parsing fails if we supply anything else but `null` -/
 /-- info: Except.error "Foo.y: type Empty has no constructor to match JSON value '1'.
-deserializing a value for type Empty, e.g. at type Option Empty with code for the 'some'
-constructor." -/
+This occurs when deserializing a value for type Empty, e.g. at type Option Empty with code
+for the 'some' constructor." -/
 #guard_msgs in
 #eval Lean.fromJson? (α := Foo Empty) <| json% {"y": 1}


TomasPuverle:
CC: @kmill in case interested/relevant.

tests/lean/run/json_empty.lean
```diff
@@ -0,0 +1,52 @@
...
+
+/-! First, two basic examples with α not-Empty -/
+/-- info: {"y": 1} -/
+#guard_msgs in
+#eval Lean.toJson (⟨some 1⟩ : Foo Nat)
+
+/-! Ensure we can parse the type back -/
+/-- info: Except.ok { y := some 1 } -/
+#guard_msgs in
+#eval fromStrHelper (Foo Nat) "{\"y\": 1}"
```
eric-wieser, line 24: Perhaps it's not in core, but somewhere there's a `json%` elaborator which should avoid the quoting awkwardness here.

src/Lean/Data/Json/FromToJson.lean
```diff
@@ -28,6 +28,9 @@ instance : ToJson Json := ⟨id⟩
 instance : FromJson JsonNumber := ⟨Json.getNum?⟩
 instance : ToJson JsonNumber := ⟨Json.num⟩
 
+instance : FromJson Empty where
+  fromJson? j := throw s!"no constructor matched JSON value '{j}'"
```
david-christiansen, line 32: As long as the string is a literal here, rather than something generated by `deriving`, it might save someone some debugging pain someday to write:
```suggestion
  fromJson? j := throw s!"type Empty has no constructor to match JSON value '{j}'"
```

tests/lean/run/json_empty.lean
david-christiansen: There's also raw strings:
```suggestion
#eval fromStrHelper (Foo Nat) r#"{"y": 1}"#
```

eric-wieser:
```suggestion
#eval Lean.fromJson? (α := Foo Nat) <| json% {"y": 1}
```

src/Lean/Data/Json/FromToJson.lean
tydeu: It might also be worth noting (in the error message) that the reason for reaching this error at all is that the JSON is ill-formed and the expected type is never meant to have JSON of this kind. That is, actually calling `fromJson?` or `toJson` on an `Empty` type at runtime is always an error.

tests/lean/run/json_empty.lean
TomasPuverle: @eric-wieser 
Thanks for the suggestion!  I will change.  Initially, I was trying to keep the test "as simple as possible", i.e. without involving additional Elab machinery such as `json%`

@david-christiansen 
Thanks, TIL.  I've not come across raw strings in FPiL yet 😄  FWIW, this (esp the PR preceding this one) is inspired because I wanted to `#eval`/`Repr` the `inductive Prim (special : Type)` in Section 5.2, so thanks for the motivation!

src/Lean/Data/Json/FromToJson.lean
TomasPuverle: Thanks, will re-phrase.  I wasn't sure how detailed to be in the error but I am big fan or helpful error messages so more than happy to expand it.

tests/lean/run/json_empty.lean
kmill: (Raw strings are less than a year old — you could take a look at #2929, which also included a bit of documentation about the lexical structure.)

TomasPuverle: Done.

src/Lean/Data/Json/FromToJson.lean
TomasPuverle: Done.

src/Lean/Data/Json/FromToJson.lean
```diff
@@ -28,6 +28,12 @@ instance : ToJson Json := ⟨id⟩
 instance : FromJson JsonNumber := ⟨Json.getNum?⟩
 instance : ToJson JsonNumber := ⟨Json.num⟩
 
+instance : FromJson Empty where
+  fromJson? j := throw (s!"type Empty has no constructor to match JSON value '{j}'. "
+  ++ "This is likely due to invalid JSON input, such as e.g. constructing an `Option Empty` "
+  ++ "with a non-null value.")
```
david-christiansen, line 34: I think that this isn't quite true - the problem isn't the validity of the JSON, it's the fact that it's being used to target a type that it shouldn't. And the bit about `null` leaks the encoding in a way that I think isn't so helpful.

What about this phrasing?
```suggestion
  ++ "This occurs when deserializing a value for type Empty, e.g. at type "
  ++ "Option Empty with code for the 'some' constructor.")
```

A [string gap](https://lean-lang.org/blog/2024-2-4-lean-450/) would also be nice here.

tests/lean/run/json_empty.lean
david-christiansen: Glad you liked the book! I hope to find time soon to update it for all the new things that Lean has become able to do since way back in 4.3 :-)

src/Lean/Data/Json/FromToJson.lean
TomasPuverle: Sure. I went back-and-forth with the wording, trying to decide if I should make the error relative to the input format or the target type.  Happy to change.

TomasPuverle:
🎉

PR #5427, opened 2024-09-23
chore: fix spelling mistakes in src/Init/

euprunin:


src/Init/Data/Fin/Iterate.lean
```diff
@@ -35,7 +35,7 @@ Concretely, `hIterate start stop f init` is equal to
   init |> f start _ |> f (start+1) _ ... |> f (end-1) _
 ```
 
-Because it is heterogenous and must return a value of type `P stop`,
+Because it is heterogeneous and must return a value of type `P stop`,
```
nomeata, line 38: Isn't both correct, depending on whether you are in the US or not? But no need to discuss, if a few non-typos switch idiom among really typo fixes I'm happy.

PR #5428, opened 2024-09-23
fix: worker: make sure to always report some recent range as progress when waiting

Kha:
On a document edit, it may be the case that the first nontrivial snapshot is e.g. for a macro-generated tactic call that does not have range information. In that case, instead of just displaying nothing, we should fall back to a previous range, in this case of the original tactic macro.


Kha:
Fixes a missing progress bar issue reported by @b-mehta

PR #5429, opened 2024-09-23
perf: avoid negative environment lookup

Kha:
Avoids some `Environment.find?` lookup misses that become especially expensive on the async branch

### DIFF
--- a/src/Lean/Meta/WHNF.lean
+++ b/src/Lean/Meta/WHNF.lean
@@ -831,10 +831,12 @@ mutual
           else
             unfoldDefault ()
     | .const declName lvls => do
+      let some cinfo ← getUnfoldableConstNoEx? declName | pure none
+      -- check smart unfolding only after `getUnfoldableConstNoEx?` because smart unfoldings have a
+      -- significant chance of not existing and `Environment.contains` misses are more costly
       if smartUnfolding.get (← getOptions) && (← getEnv).contains (mkSmartUnfoldingNameFor declName) then
         return none
       else
-        let some cinfo ← getUnfoldableConstNoEx? declName | pure none
         unless cinfo.hasValue do return none
         deltaDefinition cinfo lvls
           (fun _ => pure none)


Kha:
!bench

Kha:
!bench

src/Lean/Meta/WHNF.lean
```diff
@@ -75,7 +75,7 @@ def isAuxDef (constName : Name) : MetaM Bool := do
 -- ===========================
 
 private def getFirstCtor (d : Name) : MetaM (Option Name) := do
-  let some (ConstantInfo.inductInfo { ctors := ctor::_, ..}) ← getUnfoldableConstNoEx? d |
+  let some (ConstantInfo.inductInfo { ctors := ctor::_, ..}) := (← getEnv).find? d |
```
leodemoura, line 78: This is a good change, and we should merge it as a 'fix' in a separate PR. Using `getUnfoldedConstNoEx?` here looks very strange, so I decided to trace its origin. It originates from Lean 3, where a function for retrieving a constant without throwing an exception was used. Since then, the code has been modified and refactored a few  times, but this line was never addressed and has become increasingly odd.

src/Lean/Meta/GetUnfoldableConst.lean
```diff
@@ -39,7 +39,10 @@ def getUnfoldableConst? (constName : Name) : MetaM (Option ConstantInfo) := do
   match (← getEnv).find? constName with
   | some (info@(.thmInfo _))  => getTheoremInfo info
   | some (info@(.defnInfo _)) => if (← canUnfold info) then return info else return none
-  | some info                 => return some info
+  | some (info@(.recInfo _))  => return some info
```
leodemoura, line 42: The name of this function is strange. It is currently called `getUnfoldableConst?` but some of the cases do not even have something to unfold (e.g., `recInfo`). 
I found two use cases for this function: 
- `isDeltaCandidate?`: This one immediately checks whether the `info` has a value or not. We can optimize this for this case and have clean code.
- `matchConstAux`: which is used to handle orthogonal cases using the same API. We should cleanup this one. This function looks odd to me.

I prefer to cleanup and ensure the function names make sense. If we decide to not cleanup, we should add comments explaining why the function implementation is counterintuitive.

Kha:
@leodemoura Thanks for the additional context, I reduced this PR to the reordering of operations and will rework `getUnfoldableConst?` in a separate PR

Kha:
!bench

Kha:
Apparently this change is neutral until parallelism comes in, but I believe it is harmless enough now to go ahead with merging it

PR #5450, opened 2024-09-24
feat: add BitVec.toInt_[intMin|neg|neg_of_ne_intMin ]

tobiasgrosser:


tobiasgrosser:
I am not yet very happy with `two_pow_pred_sub_two_pow` and related functons:

- I currently have two versions of `two_pow_pred_sub_two_pow` and would really like to have one canonical version. The one without explicit casts looks a bit nicer, but then has these heterogenous `@HPow.hPow Int Nat Int instHPow 2 (w - 1) : Int` which do not match with any of the `2 ^ w` lemmas in Nat.

- Is there a reason `Int.natCast_pow` is not tagged as `norm_cast`?

- I had to introduce `theorem natCast_two_pow (n : Nat) : ((2^n : Nat) : Int) = (2: Int) ^ n := by` to make `2 ^ w` lemmas apply.
  Obvious, it would be nice to generalize this for any positive constant.
  
  I tried to tag this lemmas as norm_cast, but this just led to more heterogenous pows, rather than to turn heterogenous pows into nat-pows with int-casts at the root.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -2161,6 +2162,18 @@ theorem getLsbD_intMin (w : Nat) : (intMin w).getLsbD i = decide (i + 1 = w) :=
 theorem toNat_intMin : (intMin w).toNat = 2 ^ (w - 1) % 2 ^ w := by
   simp [intMin]
 
+@[simp]
+theorem toInt_intMin {w : Nat} :
+    (intMin w).toInt = -((2 ^ (w - 1) % 2 ^ w) : Nat) := by
```
kim-em, line 2167: I'm confused: why it there a mod on the RHS?

src/Init/Data/Int/Pow.lean
```diff
@@ -41,4 +42,22 @@ theorem natCast_pow (b n : Nat) : ((b^n : Nat) : Int) = (b : Int) ^ n := by
   | n + 1 =>
     simp only [Nat.pow_succ, Int.pow_succ, natCast_mul, natCast_pow _ n]
 
+theorem natCast_two_pow (n : Nat) : ((2^n : Nat) : Int) = (2: Int) ^ n := by
```
kim-em, line 45:
```suggestion
theorem natCast_two_pow (n : Nat) : ((2^n : Nat) : Int) = (2 : Int) ^ n := by
```

src/Init/Data/Int/Pow.lean
```diff
@@ -41,4 +42,22 @@ theorem natCast_pow (b n : Nat) : ((b^n : Nat) : Int) = (b : Int) ^ n := by
...
+  match n with
+  | 0 => rfl
+  | n + 1 =>
+    simp [Nat.pow_succ, natCast_mul, natCast_pow _ n, Nat.cast_ofNat_Int,
+      Int.pow_succ]
+
+@[simp]
+protected theorem two_pow_pred_sub_two_pow {w : Nat} (h : 0 < w) :
+    (2 ^ (w - 1) : Nat) - (2 ^ w : Nat) = - ((2 ^ (w - 1) : Nat) : Int) := by
```
kim-em, line 54: Let's write a `(... : Int)` around the LHS, for readability.

src/Init/Data/Int/Pow.lean
```diff
@@ -41,4 +42,22 @@ theorem natCast_pow (b n : Nat) : ((b^n : Nat) : Int) = (b : Int) ^ n := by
...
+  rw [← Nat.two_pow_pred_add_two_pow_pred h]
+  omega
+
+@[simp]
+protected theorem two_pow_pred_sub_two_pow' {w : Nat} (h : 0 < w) :
+    2 ^ (w - 1) - 2 ^ w = - 2 ^ (w - 1) := by
```
kim-em, line 60: Maybe write 
```suggestion
    (2 : Int) ^ (w - 1) - 2 ^ w = - 2 ^ (w - 1) := by
```
so a human reader can get the types right?

src/Init/Data/Int/Pow.lean
kim-em: You can just replace `2` with `k : Nat` here, and the proof still works.

kim-em: Indeed, `natCast_pow k n` works as the entire proof!

src/Init/Data/Int/Pow.lean
tobiasgrosser: Done.

src/Init/Data/Int/Pow.lean
tobiasgrosser: OK.

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: To ensure that for `w = 0` the expression evaluates to `0` instead of `1`. There is precedence of this pattern for here:

```lean
theorem toNat_intMin : (intMin w).toNat = 2 ^ (w - 1) % 2 ^ w := by
  simp [intMin]

File: src/Init/Data/Int/Pow.lean (hunk 2 above)
Comment (tobiasgrosser) [reply src/Init/Data/Int/Pow.lean]: Dropped.

File: src/Init/Data/BitVec/Lemmas.lean (hunk 1 above)
Comment (kim-em) [reply src/Init/Data/BitVec/Lemmas.lean:2268]: Let's add a comment on both of those so I'm not confused next time I encounter it. :-)

File: src/Init/Data/BitVec/Lemmas.lean [hunk 5]
```diff
@@ -2262,13 +2263,41 @@ theorem getLsbD_intMin (w : Nat) : (intMin w).getLsbD i = decide (i + 1 = w) :=
...
+    omega
+  by_cases w_0 : w = 0
+  · subst w_0
+    simp [BitVec.eq_nil x]
+  have : 0 < w := by omega
+  rw [Nat.two_pow_pred_mod_two_pow (by omega)] at rs
+  simp only [BitVec.toInt, BitVec.toNat_neg, BitVec.sub_toNat_mod_cancel x_zero]
+  have := @Nat.two_pow_pred_mul_two w (by omega)
+  split <;> split <;> omega
+
```
kim-em, line 2300: Could we also have the one involving `bmod`? Perhaps rename this to `toInt_neg_of_ne_intMin`.

I would like to have `toInt` lemmas for all the algebraic operations: these will all involve `bmod` on the RHS, so it's best if we have this pattern for `neg` as well.

tobiasgrosser: OK, I got it. Will be adding this for `neg` and then will look into the other algebraic operations.

PR #5451, opened 2024-09-24
chore: remove trailing and repeated spaces

euprunin:


nomeata:
Thanks for your typo fixes, it's nice to have them fixed.

As for whitespace the benefit is marginal, and there is a risk of conflict with open PRs and additional noise in `git blame`, so I'd prefer not to apply this. Nevertheless thanks for wanting to help out!

PR #5474, opened 2024-09-25
feat: have autoparams report parameter/field on failure

### LINKED ISSUE #2950: Bad error message when tactic for filling default field value fails
```lean
structure Foo where
  val : String
  len : Nat := val.length
  inv : val.length = len := by decide

def test2 : Foo := { -- The error message just says "failed to reduce to 'true'\n  false"
  val := "abc"
  len := 5
}
```
The error should say that Lean failed to initialize field `inv` using the specified tactic.

kmill:
Adds a mechanism where when an autoparam tactic fails to synthesize a parameter, the associated parameter name or field name for the autoparam is reported in an error.

Examples:
```text
could not synthesize default value for parameter 'h' using tactics

could not synthesize default value for field 'inv' of 'S' using tactics
```

Notes:
* Autoparams now run their tactics without any error recovery or error-to-sorry enabled. This enables catching the error and reporting the contextual information. This is justified on the grounds that autoparams are not interactive.
* Autoparams for applications now cleanup the autoParam annotation, bringing it in line with autoparams for structure fields.
* This preserves the old behavior that autoparams leave terminfo, but we will revisit this after some imminent improvements to the unused variable linter.

Closes #2950

### DIFF
--- a/src/Lean/Elab/App.lean
+++ b/src/Lean/Elab/App.lean
@@ -596,6 +596,9 @@ mutual
           let info := (← getRef).getHeadInfo
           let tacticBlock := tacticBlock.raw.rewriteBottomUp (·.setInfo info)
           let mvar ← mkTacticMVar argType.consumeTypeAnnotations tacticBlock (.autoParam argName)
+          -- Note(kmill): We are adding terminfo to simulate a previous implementation that elaborated `tacticBlock`.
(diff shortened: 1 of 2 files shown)


kim-em:
Really happy to see this!

Kha (approved):

PR #5478, opened 2024-09-26
feat: add BitVec.(shiftLeft_add_distrib, shiftLeft_ushiftRight)

luisacicolini:
Moved some Nat theorems from Mathlib

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -1512,7 +1512,6 @@ theorem shiftRight_add {w : Nat} (x : BitVec w) (n m : Nat) :
   ext i
   simp [Nat.add_assoc n m i]
 
-@[simp]
 theorem shiftLeft_ushiftRight {x : BitVec w} {n : Nat}:
     x >>> n <<< n = x &&& BitVec.allOnes w <<< n := by
   induction n generalizing x


src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1512,6 +1512,25 @@ theorem shiftRight_add {w : Nat} (x : BitVec w) (n m : Nat) :
   ext i
   simp [Nat.add_assoc n m i]
 
+@[simp]
+theorem shiftLeft_shiftRight {x : BitVec w} {n : Nat}:
+  x >>> n <<< n = x &&& BitVec.allOnes w <<< n := by
```
tobiasgrosser, line 1517:
```suggestion
theorem shiftLeft_ushiftRight {x : BitVec w} {n : Nat}:
  x >>> n <<< n = x &&& BitVec.allOnes w <<< n := by
```

tobiasgrosser (approved):
This looks good to me.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1512,6 +1512,25 @@ theorem shiftRight_add {w : Nat} (x : BitVec w) (n m : Nat) :
   ext i
   simp [Nat.add_assoc n m i]
 
+@[simp]
+theorem shiftLeft_ushiftRight {x : BitVec w} {n : Nat}:
+  x >>> n <<< n = x &&& BitVec.allOnes w <<< n := by
```
tobiasgrosser, line 1517:
```suggestion
theorem shiftLeft_ushiftRight {x : BitVec w} {n : Nat} :
  x >>> n <<< n = x &&& BitVec.allOnes w <<< n := by
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1512,6 +1512,25 @@ theorem shiftRight_add {w : Nat} (x : BitVec w) (n m : Nat) :
...
+  induction n generalizing x
+  case zero =>
+    ext; simp
+  case succ n ih =>
+    rw [BitVec.shiftLeft_add, Nat.add_comm, BitVec.shiftRight_add, ih,
+       Nat.add_comm, BitVec.shiftLeft_add, BitVec.shiftLeft_and_distrib]
+    ext i
+    have hi₁ (_: 0 < i.val) : 1 + (i.val - 1) = i := by
+      rw [Nat.add_comm, Nat.sub_add_cancel]
```
hargoniX, line 1526: Omega should be able to solve this goal without this `rw` I think?

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1512,6 +1512,25 @@ theorem shiftRight_add {w : Nat} (x : BitVec w) (n m : Nat) :
...
+      omega
+    by_cases hw : w = 0
+    · simp [hw]
+    by_cases hi₂ : i.val = 0
```
hargoniX, line 1530: I'm assuming this is the second case of the `by_cases` in 1528? If so there should be an additional center dot in front of this `by_cases` and then according indentation etc.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1512,6 +1512,25 @@ theorem shiftRight_add {w : Nat} (x : BitVec w) (n m : Nat) :
...
+    · simp [hi₂]
+    · simp [Nat.lt_one_iff, hi₂, hi₁ (show 0 < i.val by omega)]
```
hargoniX, line 1532: AFAICT this is the only use of `hi₁` right? Given the fact that it and its premise are provable by `omega` I'd almost be certain that you can just put the `have` without the additional premise etc. in this specific case and prove it `by omega`, then rewrite with it in this `simp`.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1737,6 +1756,15 @@ theorem ofInt_add {n} (x y : Int) : BitVec.ofInt n (x + y) =
...
   simp
 
+@[simp]
+theorem shiftLeft_add_distrib {x y : BitVec w} {n : Nat} :
+    (x + y) <<< n = x <<< n + y <<< n := by
+  induction n
+  case zero =>
+    simp
+  case succ n ih =>
+    simp [ih, toNat_eq, Nat.shiftLeft_eq_mul_pow, ←Nat.add_mul]
```
hargoniX, line 1766: Our style typically mandates a space after the `←`: `← Nat.add_mul`, same for left arrows in do notation.

src/Init/Data/Nat/Lemmas.lean
```diff
@@ -916,6 +919,8 @@ theorem mul_add_mod (m x y : Nat) : (m * x + y) % m = y % m := by
   · exact (m % 0).div_zero
   · case succ n => exact Nat.div_eq_of_lt (m.mod_lt n.succ_pos)
 
+theorem shiftLeft_eq_mul_pow (m) : ∀ n, m <<< n = m * 2 ^ n := shiftLeft_eq _
```
hargoniX, line 922: If you can prove this theorem just as an alias to `shiftLeft_eq` why is it restated under a new name here?

src/Init/Data/BitVec/Lemmas.lean
luisacicolini: not sure I understood exactly what you mean, but without the premise (`have hi₁ : 1 + (i.val - 1) = i := by...`) `omega` is not sufficient (it is otherwise)

hargoniX: Yes but if you prove the `have` in the context of the line that is marked here that hypothesis will be implicitly available so you should be able to directly show your fact as a `have` in this particular subgoal. Instead of proving the more general version above and then filling in the argument with another `by omega`

hargoniX (approved):

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1512,6 +1512,21 @@ theorem shiftRight_add {w : Nat} (x : BitVec w) (n m : Nat) :
   ext i
   simp [Nat.add_assoc n m i]
 
+theorem shiftLeft_ushiftRight {x : BitVec w} {n : Nat}:
```
bollu, line 1515:
```suggestion
/--
Shifting right by `n` to clear the low `n` bits and then shifting left
is the same as bitmasking with an all-ones mask, shifted to the left by `n`
to clear the *mask's* low `n` bits.
-/
theorem shiftLeft_ushiftRight {x : BitVec w} {n : Nat}:
```

kim-em:
`Nat.lt_one_iff` was in Mathlib with `@[simp]`, and was moved here without `simp`.

* probably that shouldn't have happened
* but this shouldn't have been merged without a `builds-mathlib` label at all. :-(

tobiasgrosser:
> `Nat.lt_one_iff` was in Mathlib with `@[simp]`, and was moved here without `simp`.
> 
> * probably that shouldn't have happened
> * but this shouldn't have been merged without a `builds-mathlib` label at all. :-(

Uh, sorry for this. I guess all builds that move theorems from Mathlib should have this label, no?

PR #5480, opened 2024-09-26
test: check that recusive functions do not apply attriubutes twices

nomeata:
I suspected a bug based on reading the code, but it seems there is no
bug.


nomeata:
Uff, that’s a typo-laden commit title. Sorry.

PR #5484, opened 2024-09-26
fix: handling BitVec.ofNat with Nat fvars in bv_decide

hargoniX:


bollu:
Discovered when reducing memory obligations to `bv_decide` in https://github.com/leanprover/LNSym/pull/197

PR #5488, opened 2024-09-27
feat: add DecidableRel instances for BitVec

bollu:
We follow the recipe of the `Decidable` instance for bitvectors and create a `DecidableRelation` instance,

This was motivated by a case-bash that was used to prove a `sdiv` lemma at width 1. However, due to file ordering constraints, we cannot use the `DecidableRel` machinery to implement the lemma `sdiv_eq_and`.

Regardless, we upstream the `DecidableRel` machinery for future use.


bollu:
Ah, right, these can just be inferred by type class synthesis. No need to define these explicitly.

PR #5491, opened 2024-09-27
feat: add BitVec.toNat_[abs|sdiv|smod]

tobiasgrosser:


src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1335,6 +1335,15 @@ theorem sdiv_eq (x y : BitVec w) : x.sdiv y =
   rw [BitVec.sdiv]
   rcases x.msb <;> rcases y.msb <;> simp
 
+theorem toNat_sdiv (x y : BitVec w) : (x.sdiv y).toNat =
```
luisacicolini, line 1338:
```suggestion
theorem toNat_sdiv {x y : BitVec w} : (x.sdiv y).toNat =
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1358,6 +1367,23 @@ theorem smod_eq (x y : BitVec w) : x.smod y =
   rw [BitVec.smod]
   rcases x.msb <;> rcases y.msb <;> simp
 
+theorem toNat_smod (x y : BitVec w) : (x.smod y).toNat =
```
luisacicolini, line 1370:
```suggestion
theorem toNat_smod {x y : BitVec w} : (x.smod y).toNat =
```

luisacicolini (approved):
lgtm

hargoniX (changes requested):
I'm assuming the reason that you didn't put these theorems in the `bv_toNat` simp set is that `omega` cannot handle `match` and `if` yes? Apart from that it looks fine to me.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1335,6 +1335,15 @@ theorem sdiv_eq (x y : BitVec w) : x.sdiv y =
   rw [BitVec.sdiv]
   rcases x.msb <;> rcases y.msb <;> simp
 
+theorem toNat_sdiv {x y : BitVec w} : (x.sdiv y).toNat =
+    match x.msb, y.msb with
+    | false, false => (udiv x y).toNat
+    | false, true  =>  (- (x.udiv (- y))).toNat
+    | true,  false => (- ((- x).udiv y)).toNat
+    | true,  true  => ((- x).udiv (- y)).toNat := by
+  simp [sdiv_eq]
```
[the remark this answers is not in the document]
tobiasgrosser: Done.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1358,6 +1367,23 @@ theorem smod_eq (x y : BitVec w) : x.smod y =
...
+    match x.msb, y.msb with
+    | false, false => (x.umod y).toNat
+    | false, true =>
+      let u := x.umod (- y)
+      (if u = 0#w then u.toNat else (u + y).toNat)
+    | true, false =>
+      let u := (-x).umod y
+      (if u = 0#w then u.toNat else (y - u).toNat)
+    | true, true => (- ((- x).umod (- y))).toNat := by
+  simp [smod_eq]
```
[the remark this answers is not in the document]
tobiasgrosser: Done.

tobiasgrosser:
> I'm assuming the reason that you didn't put these theorems in the `bv_toNat` simp set is that `omega` cannot handle `match` and `if` yes? Apart from that it looks fine to me.

Right. However, on reflection it still makes sense. The pattern will then just be to call bv_toNat, then split on the cases, use omega after.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1961,6 +1989,17 @@ theorem neg_ne_iff_ne_neg {x y : BitVec w} : -x ≠ y ↔ x ≠ -y := by
     subst h'
     simp at h
 
+/-! ### abs -/
+
+@[simp]
```
hargoniX, line 1994: If you agree that the pattern makes sense, why not `bv_toNat` this one as well?^^

tobiasgrosser: Right. I forgot to push this last change. Its in now.

hargoniX (approved):

PR #5497, opened 2024-09-27
doc: update documentation and tests for `toUIntX` functions

### LINKED ISSUE #5483: The conversion of `Float` to `UInt64` doesn't behave as described in the documentation.
### Prerequisites

### Description

The conversion of `Float` to `UInt64` (etc) doesn't behave as described in the documentation.

From `Data/Float.lean`:

```
/-- If the given float is positive, truncates the value to the nearest positive integer.
If negative or larger than the maximum value for UInt64, returns 0. -/
@[extern "lean_float_to_uint64"] opaque Float.toUInt64 : Float → UInt64
```

However, the following seems problematic:

```
def inf : Float := 1.0 / 0.0

-- All of the statements below return true
#eval Float.isInf inf
#eval inf.toUInt64.toNat == (UInt64.size - 1)
#eval (UInt64.size.toFloat * 2).toUInt64.toNat == (UInt64.size - 1)
```

### Context

As discussed on [Zulip]( https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/RFC.3A.20Float.2EtoNat.20.2B.20possible.20Float.20bug.3F/near/472733249)

### Steps to Reproduce

1.  See the code above

**Expected behavior:** 
- Either fix the code to return 0
- Update documentation to reflect that the "largest Uint64 value" is returned

**Actual behavior:** 
- Currently returns `UIntXX.size - 1`

### Versions

"4.12.0, commit 5dea30f169b91cdda9a3903a7a9d125e0f075605", MacOS

### Additional Information

I'm happy to create a PR, either to fix the behavior or the documentation.

TomasPuverle:
Update documentation on functions to reflect actual behavior.
Add tests to ensure said behavior is as documented.

Closes #5483


TomasPuverle:
CC: @david-christiansen since I noticed most people ask you for documentation/wording review for PRs.

david-christiansen:
Thanks for the ping! I was off away from the Internet for a few days. Other than marking code with backticks, I think the text is great :)

david-christiansen:
(and the presence of tests even greater)

TomasPuverle:
Thanks!  I'll know to use backticks in the future.

As for the tests, I considered moving them to the format using `#guard_msg` but thought perhaps that was too big of a change for just a doc update.  However, is it the general intention for all tests to be ported to the new "standard"?

PR #5498, opened 2024-09-27
feat: make `BitVec.getElem` the simp normal form and use it in `ext`

tobiasgrosser:
This PR makes `BitVec.getElem` the simp normal form in case a proof is available and changes `ext` to return `x[i]` + a hypothesis that proves that we are in-bounds. This aligns `BitVec` further with the API conventions of the Lean standard datatypes.

We move our proofs to this new normal form, which results in slightly smaller proofs.  With the exception of `getElem_ofFin`, no new API surface is added as the `getElem` API has already been completed over the previous months. We also move `getElem_shiftConcat_*` a bit higher as they are needed in earlier proofs. To keep the changeset small, we do not update the API of `BVDecide` but insert `← BitVec.getLsbD_eq_getElem` at the few locations where it is needed. Finally, we add a simproc for getElem, mirroring the existing ones for getLsbD/getMsdD.

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -93,6 +93,11 @@ theorem getLsbD_eq_getElem?_getD {x : BitVec w} {i : Nat} :
   · rfl
   · simp_all
 
+@[simp]
+theorem getElem_of_getLsbD_eq_true {x : BitVec w} {i : Nat} (h : x.getLsbD i = true) :
+    (x[i]'(lt_of_getLsbD h) = true) = True := by
+  simp [← BitVec.getLsbD_eq_getElem, h]
+
 /--
 This normalized a bitvec using `ofFin` to `ofNat`.
 -/
@@ -3728,8 +3733,7 @@ theorem setWidth_setWidth_succ_eq_setWidth_setWidth_or_twoPow_of_getLsbD_true
   simp only [getElem_setWidth, h, getElem_or, getElem_twoPow]
   by_cases hik : i = k
   · subst hik
-    simp only [h, getLsbD_eq_getElem] at hx
-    simp [h, hx]
+    simp [hx]
   · by_cases hik' : k < i + 1
     <;> simp [hik, hik', show ¬ (k = i) by omega]
     <;> omega


src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -3735,16 +3725,19 @@ theorem setWidth_setWidth_succ_eq_setWidth_setWidth_or_twoPow_of_getLsbD_true
     setWidth w (x.setWidth (i + 1)) =
       setWidth w (x.setWidth i) ||| (twoPow w i) := by
   ext k h
-  simp only [getLsbD_setWidth, h, decide_true, Bool.true_and, getLsbD_or, getLsbD_and]
   by_cases hik : i = k
   · subst hik
-    simp [hx, h]
-  · by_cases hik' : k < i + 1 <;> simp [hik, hik'] <;> omega
+    simp only [h, getLsbD_eq_getElem] at hx
+    simp [h, hx]
```
luisacicolini, line 3731: nonterminal simp

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -359,24 +359,25 @@ theorem msb_sub {x y: BitVec w} :
...
 
 theorem bit_not_add_self (x : BitVec w) :
-  ((iunfoldr (fun (i : Fin w) c => (c, !(x.getLsbD i)))) ()).snd + x  = -1 := by
+  ((iunfoldr (fun (i : Fin w) c => (c, !(x[i.val])))) ()).snd + x  = -1 := by
   simp only [add_eq_adc]
   apply iunfoldr_replace_snd (fun _ => false) (-1) false rfl
-  intro i; simp only [ BitVec.not, adcb, testBit_toNat]
-  rw [iunfoldr_replace_snd (fun _ => ()) (((iunfoldr (fun i c => (c, !(x.getLsbD i)))) ()).snd)]
-  <;> simp [bit_not_testBit, negOne_eq_allOnes, getLsbD_allOnes]
+  intro i; simp [adcb, Fin.is_lt, getLsbD_eq_getElem, atLeastTwo_false_right, bne_false,
```
luisacicolini, line 369: nonterminal simp

alexkeizer:
@tobiasgrosser I addded a `reduceGetElem` simproc, which (locally) makes the previously broken tests pass

src/Init/Data/BitVec/Bitblast.lean
tobiasgrosser: Changed.

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: Addressed.

tobiasgrosser:
changelog-library

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -178,9 +181,7 @@ theorem getMsbD_eq_getMsb?_getD (x : BitVec w) (i : Nat) :
       intros
       omega
 
--- We choose `eq_of_getLsbD_eq` as the `@[ext]` theorem for `BitVec`
--- somewhat arbitrarily over `eq_of_getMsbD_eq`.
-@[ext] theorem eq_of_getLsbD_eq {x y : BitVec w}
```
JamesGallicchio, line 183: I think the ext annotation autogenerates a `BitVec.eq_of_getLsbD_eq_iff` declaration, which probably should be added explicitly. Not sure if there is a "right" way to autogenerate the iff versions of ext lemmas for types with multiple reasonable ext lemmas.

luisacicolini (approved):

JamesGallicchio (approved):
I am not sure how to test the new simp set and simproc in my code before it lands in a RC, but the diff looks very reasonable to me!

tobiasgrosser:
> I am not sure how to test the new simp set and simproc in my code before it lands in a RC, but the diff looks very reasonable to me!

The Lean CI builds toolchains for all PRs. For this PR, the toolchain is available at `leanprover/lean4-pr-releases:pr-release-5498`. You can use it by updating your `lean-toolchain` file.

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: I added this.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -808,12 +821,9 @@ and the second `setWidth` is a non-trivial extension.
...
     setWidth w (setWidth v x) = setWidth w x := by
-  ext
-  simp_all only [getLsbD_setWidth, decide_true, Bool.true_and, Bool.and_iff_right_iff_imp,
-    decide_eq_true_eq]
-  intro h
-  replace h := lt_of_getLsbD h
-  omega
+  ext i ih
+  have h := @lt_of_getLsbD u x i
+  by_cases h : x.getLsbD i = true <;> simp [h] at * <;> omega
```
kim-em, line 826: Strange that both are called `h`. Drop one name?

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -2033,23 +2039,22 @@ theorem msb_append {x : BitVec w} {y : BitVec v} :
     simp [h, q, t, BitVec.msb, getMsbD]
 
 @[simp] theorem append_zero_width (x : BitVec w) (y : BitVec 0) : x ++ y = x := by
-  ext
-  rw [getLsbD_append] -- Why does this not work with `simp [getLsbD_append]`?
-  simp
+  ext i ih
+  rw [getElem_append] -- Why does this not work with `simp [getLsbD_append]`?
```
kim-em, line 2043: Update comment?

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: Addressed.

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: Addressed.

kim-em: I think we need to update `getElem_append`'s right-hand-side.

kim-em:
Can we update the RHS of getElem_append? (and maybe to a quick review of `_getElem` lemmas for the same issue)

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -2033,23 +2039,22 @@ theorem msb_append {x : BitVec w} {y : BitVec v} :
     simp [h, q, t, BitVec.msb, getMsbD]
 
 @[simp] theorem append_zero_width (x : BitVec w) (y : BitVec 0) : x ++ y = x := by
-  ext
-  rw [getLsbD_append] -- Why does this not work with `simp [getLsbD_append]`?
-  simp
+  ext i ih
+  rw [getElem_append] -- Why does this not work with `simp [getELem_append]`?
```
kim-em, line 2043: Typo in `getELem`

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -3633,9 +3636,9 @@ theorem msb_twoPow {i w: Nat} :
 
 theorem and_twoPow (x : BitVec w) (i : Nat) :
     x &&& (twoPow w i) = if x.getLsbD i then twoPow w i else 0#w := by
-  ext j
-  simp only [getLsbD_and, getLsbD_twoPow]
-  by_cases hj : i = j <;> by_cases hx : x.getLsbD i <;> simp_all
+  ext j h
+  simp only [←getLsbD_eq_getElem, getLsbD_and, getLsbD_twoPow]
```
kim-em, line 3640:
```suggestion
    simp only [getElem_and, getLsbD_twoPow]
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -3724,16 +3725,19 @@ theorem setWidth_setWidth_succ_eq_setWidth_setWidth_or_twoPow_of_getLsbD_true
     setWidth w (x.setWidth (i + 1)) =
       setWidth w (x.setWidth i) ||| (twoPow w i) := by
   ext k h
-  simp only [getLsbD_setWidth, h, decide_true, Bool.true_and, getLsbD_or, getLsbD_and]
+  simp only [getElem_setWidth, getLsbD_setWidth, h, getLsbD_eq_getElem, getElem_or, getElem_twoPow]
```
kim-em, line 3728:
```suggestion
  simp only [getElem_setWidth, h, getElem_or, getElem_twoPow]
```

tobiasgrosser:
> Can we update the RHS of getElem_append? (and maybe to a quick review of `_getElem` lemmas for the same issue)

To `if h: i < m then y[i] else x[i - m]`?

kim-em:
Also `getElem_setWidth`. Then review the proof of `setWidth_setWidth_succ_eq_setWidth_setWidth_or_twoPow_of_getLsbD_true`

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -3761,7 +3765,7 @@ theorem getLsbD_replicate {n w : Nat} {x : BitVec w} :
     by_cases hi : i < w * (n + 1)
     · simp only [hi, decide_true, Bool.true_and]
       by_cases hi' : i < w * n
-      · simp [hi', ih]
+      · simp [-getLsbD_eq_getElem, hi', ih]
```
kim-em, line 3768:
```suggestion
      · rw [ih]
        simp_all
```
(I see what's going on here, but having to remove simp lemmas from simp sets always worries me. The better fix might be to prove `getElem_replicate` *first*, and then derive this lemma from it, but not now.)

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -4116,14 +4124,14 @@ theorem msb_reverse {x : BitVec w} :
...
   · by_cases hw : w ≤ i
-    · simp [getMsbD_append, getLsbD_cast, getLsbD_append, getLsbD_reverse, hw]
-    · simp [getMsbD_append, getLsbD_cast, getLsbD_append, getLsbD_reverse, hw, show i < w by omega]
+    · simp [getMsbD_append, getLsbD_reverse, hw]
+    · simp [getMsbD_append, getElem_reverse, hw, show i < w by omega]
   · by_cases hw : w ≤ i
-    · simp [getMsbD_append, getLsbD_cast, getLsbD_append, hw, show ¬ i < w by omega, getLsbD_reverse]
-    · simp [getMsbD_append, getLsbD_cast, getLsbD_append, hw, show i < w by omega, getLsbD_reverse]
+    · simp [getMsbD_append, getLsbD_append, hw, show ¬ i < w by omega, getLsbD_reverse]
+    · simp [getMsbD_append, hw, show i < w by omega, getElem_reverse]
```
kim-em, line 4134:
```
  ext i h
  simp only [getElem_reverse, getElem_cast, getElem_append]
  by_cases hi : i < v
  · by_cases hw : w ≤ i
    · simp [getLsbD_reverse, hw]
    · simp [getElem_reverse, hw, show i < w by omega]
  · by_cases hw : w ≤ i
    · simp [hw, show ¬ i < w by omega, getLsbD_reverse]
    · simp [hw, show i < w by omega, getElem_reverse]
```
At some point the proof acquired some unnecessary explicit simp lemmas. May as well clean up.

kim-em:
> > Can we update the RHS of getElem_append? (and maybe to a quick review of `_getElem` lemmas for the same issue)
> 
> To `if h: i < m then y[i] else x[i - m]`?

Yes, hopefully! (With a space in `h:` :-)

kim-em:
Otherwise, LGTM.

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: Fixed.

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: Fixed.

tobiasgrosser:
> > Can we update the RHS of getElem_append? (and maybe to a quick review of `_getElem` lemmas for the same issue)
> 
> To `if h: i < m then y[i] else x[i - m]`?

I did this. The following changes may require from a quick check:

```lean
-theorem getElem_ofBool {b : Bool} : (ofBool b)[0] = b := by simp
+@[simp] theorem getElem_ofBool_zero {b : Bool} : (ofBool b)[0] = b := by simp
+
+@[simp]
+theorem getElem_ofBool {b : Bool} {h : i < 1}: (ofBool b)[i] = b := by
+  simp [← getLsbD_eq_getElem]
+  omega
```
I added these changes above as getElem_ofBool was previously too specific, and the generic version was needed for my proof.

I am not 100% sure what you have in mind with `getElem_setWidth`:

```lean
theorem getElem_setWidth (m : Nat) (x : BitVec n) (i : Nat) (h : i < m) :
    (setWidth m x)[i] = x.getLsbD i := by
```

I cannot only rewrite it to `(setWidth m x)[i] = x[i]` if I have `h' : i < n`. I feel the current form is more generic. x.getLSbD will simp to x[i] when `i < n` is in the hypothesis, no?

I will spend some time looking over the setWidth theorems. Otherwise, I addressed all comments.

tobiasgrosser:
> I will spend some time looking over the setWidth theorems. Otherwise, I addressed all comments.

I avoided the `lean4 simp only [h, getLsbD_eq_getElem]` in `setWidth_setWidth_succ_eq_setWidth_setWidth_or_twoPow_of_getLsbD_true` by introducing:

```lean
@[simp]
theorem getElem_of_getLsbD_eq_true {x : BitVec w} {i : Nat} (h : x.getLsbD i = true) :
    (x[i]'(lt_of_getLsbD h) = true) = True := by
  simp [← BitVec.getLsbD_eq_getElem, h]
```

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: Fixed.

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: Fixed.

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: Fixed.

src/Init/Data/BitVec/Lemmas.lean
tobiasgrosser: Fixed.

PR #5499, opened 2024-09-27
chore: BitVec.Lemmas - drop non-terminal simps

tobiasgrosser:
`BitVec.Lemmas` contained a couple of non-terminal simps. We turn non-terminal `simp$`, `simp [`, and `simp at` expressions into `simp only` to improve code maintainability.

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -1498,14 +1498,14 @@ theorem getLsbD_append {x : BitVec n} {y : BitVec m} :
   simp only [append_def, getLsbD_or, getLsbD_shiftLeftZeroExtend, getLsbD_setWidth']
   by_cases h : i < m
   · simp [h]
-  · simp [h]; simp_all
+  · simp_all [h]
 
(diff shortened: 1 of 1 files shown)


src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -901,7 +901,7 @@ theorem not_def {x : BitVec v} : ~~~x = allOnes v ^^^ x := rfl
     rw [Nat.testBit_two_pow_sub_succ (isLt _)]
     · cases w : decide (i < v)
       · simp at w
```
hargoniX, line 903: You missed at least one here

hargoniX (commented):
Is this PR supposed to address all of them or only the ones seen here? If it is all of them, did you verify with e.g. `grep simp | grep -v "simp only"` that you got all of them?

tobiasgrosser:
> Is this PR supposed to address all of them or only the ones seen here? If it is all of them, did you verify with, e.g., `grep simp | grep -v "simp only"` that you got all of them?

I searched now for `simp$`, `simp [`, and `simp at` and manually verified that they are terminal. While this was manual, I tried to be careful.

hargoniX (approved):

PR #5500, opened 2024-09-27
Git conflict experiment

kmill:
This is an experiment to have git conflict syntax be processed like comments.

The option `parser.git.useIncoming` controls whether to use the incoming or current versions.

@jcommelin @adomani

### DIFF
--- a/tests/lean/run/git_conflict.lean
+++ b/tests/lean/run/git_conflict.lean
@@ -1,0 +1,29 @@
+/--
+error: unsolved goals
+case right
+⊢ 0 = 0
+-/
+#guard_msgs in
+example : True ∧ 0 = 0 := by
+<<<<<<<
+  refine ⟨.intro, ?right⟩
+  done
+=======
+    refine ⟨?left, rfl⟩
+  done
+>>>>>>>
+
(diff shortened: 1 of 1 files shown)


src/Lean/Parser/Basic.lean
```diff
@@ -549,6 +549,56 @@ partial def finishCommentBlock (nesting : Nat) : ParserFn := fun c s =>
...
 
+def checkColZero (errorMsg : String) : ParserFn := fun c s =>
+    let pos      := c.fileMap.toPosition s.pos
+    if pos.column = 0 then s
+    else s.mkError errorMsg
+
+def parseStartDiff (errorMsg : String) : ParserFn :=
+  let p := satisfyFn (· == '<') errorMsg
+  andthenFn (checkColZero errorMsg)
+    (andthenFn p (andthenFn p (andthenFn p (andthenFn p (andthenFn p (andthenFn p (andthenFn p
```
adomani, line 560: What do you think of starting the diff with 6, rather than 7, `<`?

This means that you can remove the first, exclude all editor diff-related noise, and view a "clean" file?

PR #5501, opened 2024-09-27
fix: ensure `instantiateMVarsProfiling` adds a trace node

alexkeizer:
We add a new `Meta.instantiateMVars` trace node to the `instantiateMVarsProfiling` definition used in `elabMutualDef`, and we replace various uses of plain `instantiateMVars` with the profiled version (which necessitated pulling up the definition to be higher in the file).

This fixes a "time leak" when profiling large proofs, where instantiating the goal metavariable can take a significant amount of time, that previously would not be accounted for when using the trace profiler.

### DIFF
--- a/src/Lean/Elab/MutualDef.lean
+++ b/src/Lean/Elab/MutualDef.lean
@@ -22,6 +22,14 @@ open Lean.Parser.Term
 
 open Language
 
+builtin_initialize
+  registerTraceClass `Meta.instantiateMVars
+
+def instantiateMVarsProfiling (e : Expr) : MetaM Expr := do
+  profileitM Exception s!"instantiate metavars" (← getOptions) do
+  withTraceNode `Meta.instantiateMVars (fun _ => pure e) do
+    instantiateMVars e
+
 /-- `DefView` plus header elaboration data and snapshot. -/
 structure DefViewElabHeader extends DefView, DefViewElabHeaderData where
   /--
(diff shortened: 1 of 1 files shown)


alexkeizer:
Hmm, the CI error links to https://leanprover.github.io/lean4/doc/dev/commit_convention.html, which redirects to
https://www.lean-lang.org/lean4/doc/dev/commit_convention.html, which 404s.

hargoniX:
!bench to ensure the code doesn't introduce non trivial regressions.

alexkeizer:
> I'm also not exactly sure that this is the proper solution for the `instantiateMVars` time leak. This is only going to profile it in very specific situations, a more uniform solution seems more desirable to me? But also potentially more threatening to performance.

Fair point. It should be noted that `instantiateMVarsProfiling` was already a thing, I just expanded its use and ensured it has a trace node so that it shows up in the trace profiler also.

The proper solution would be to add a trace node to `instantiateMVars` itself, but that is a lot trickier, since `Lean.Util.Trace` (which defines the trace machinery) imports `Lean.MetavarContext` (where `instantiateMVars` lives).

hargoniX:
I do understand why you did what you did now. But I do not have enough knowledge about our meta internals to know what exactly would be required here. Someone else from the team would have to weigh in.

alexkeizer:
awaiting-review

Probably helps if I put that tag on. I'll be away for the next two/three weeks, though, so I won't respond to any review comments until then

alexkeizer:
awaiting-review

PR #5524, opened 2024-09-29
feat: add ac_nf and test [ac_nf|ac_rfl] for BitVec

tobiasgrosser:
ac_nf is a counterpart to ac_rfl, which normalizes bitvector expressions with respect to associativity and commutativity.

While there, also add test coverage for ac_rfl and ac_nf for BitVec, complementing the existing test coverage.

### DIFF
--- a/tests/lean/run/ac_rfl.lean
+++ b/tests/lean/run/ac_rfl.lean
@@ -138,3 +138,4 @@ example (a b c d : BitVec w) :
 
 example (a b c d : Nat) : a + b + c + d = d + (b + c) + a := by
   ac_nf
+  rfl


kim-em:
Would you mind adding a small number of tests, just to exercise that it works for one or two other types besides BitVec? `Int` would be fine.

tobiasgrosser:
> Would you mind adding a small number of tests, just to exercise that it works for one or two other types besides BitVec? `Int` would be fine.

When searching for ac_rfl uses, I missed to add the test folder to my search path. There are indeed test cases already available. I now moved the BitVec test cases in the existing test file.

PR #5528, opened 2024-09-30
feat: allow explicit mode with field notation

### LINKED ISSUE #5406: Structure update notation falls over optional parameters
### Description

In the following code, I’d expect `{ m with a := id }` to work, but the elaboration falls over copying the other field over.

```lean
structure Methods where
  a : Int → Int
  b : Nat → (opt : Nat := 42) → Nat


/--
error: type mismatch
  fun a => m.b a
has type
  Nat → Nat : Type
but is expected to have type
  Nat → optParam Nat 42 → Nat : Type
-/
#guard_msgs in
def f (m : Methods) : Methods := { m with a := id }

/--
error: type mismatch
  fun a => m.b a
has type
  Nat → Nat : Type
but is expected to have type
  Nat → optParam Nat 42 → Nat : Type
-/
#guard_msgs in
def g (m : Methods) : Methods := { m with a := id, b := m.b }

def h (m : Methods) : Methods := { m with a := id, b := @Methods.b m }
```

Maybe related to https://github.com/leanprover/lean4/issues/5397?

### Versions

4.12.0-nightly-2024-09-19

### Impact

kmill:
Now one can write `@x.f`, `@(x).f`, `@x.1`, `@(x).1`, and so on.

This fixes an issue where structure instance update notation (like `{x with a := a'}`) could fail if the field `a` had a type with implicit, optional, or auto parameters.

Closes #5406

### DIFF
--- a/src/Lean/Elab/App.lean
+++ b/src/Lean/Elab/App.lean
@@ -1394,8 +1394,6 @@ private def elabAppLValsAux (namedArgs : Array NamedArg) (args : Array Arg) (exp
 
 private def elabAppLVals (f : Expr) (lvals : List LVal) (namedArgs : Array NamedArg) (args : Array Arg)
     (expectedType? : Option Expr) (explicit ellipsis : Bool) : TermElabM Expr := do
-  if !lvals.isEmpty && explicit then
-    throwError "invalid use of field notation with `@` modifier"
   elabAppLValsAux namedArgs args expectedType? explicit ellipsis f lvals
 
 def elabExplicitUnivs (lvls : Array Syntax) : TermElabM (List Level) := do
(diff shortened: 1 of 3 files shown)


digama0:
isn't this ambiguous? If `x : {a : A} -> Foo a` and `Foo.f : {b : B} -> R`, does `@x.f` introduce `a` or `b` as explicit?

kmill:
@digama0 The current precedence rules have `@x.f` parse as `@` applied to `x.f`. Same for `@(x).f`, it's `@` applied to `(x).f`. On the other hand, `(@x).f` is `@` applied to `x`.

Dot notation doesn't work when `x` is a function, so `(@x).f` isn't useful as an interpretation.

kmill:
@alreadydone Discussion about that [here](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/Discussion.20for.20v4.2E14.2E0.20and.20v4.2E15.2E0-rc1/near/488678634)

PR #5529, opened 2024-09-30
chore: move `@[simp]` from `exists_prop'` to `exists_prop`

kmill:
The lemma `exists_const` already handles all real cases of `(∃ _ : α, p) ↔ p` for general types `α`. If there are no `Nonempty` instances and this lemma cannot apply, it seems unlikely that simp could make more progress with `(∃ _ : α, p) ↔ Nonempty α ∧ p`.

However, it is still worth simplifying `(∃ _ : p, q)` to `p ∧ q`.

Also adds a `Nonempty (Decidable a)` instance, which is used by Mathlib.


fgdorais:
It only breaks two lines from Mathlib. In one case, this actually simplifies the proof a bit. I have a patch for lean-pr-testing-5529 if you want it.

kmill:
@fgdorais Please do.

PR #5530, opened 2024-09-30
fix: automatic instance names about types with hygienic names should be hygienic

### LINKED ISSUE #2044: Auto-generated instance names do not inherit macro scopes
### Description

Suppose that the following metaprogram is executed in two different modules `A.lean` and `B.lean`, and that `C.lean` then imports both modules.

```lean
import Lean
open Lean Elab Command

#eval (do
  let stx ← `(
    structure Foo where
      field : String
    deriving ToJson)
  elabCommand stx
  : CommandElabM Unit
)
```

The duplication of the name `Foo` is fine and does not cause a clash thanks to macro hygiene which turns it into something like `` `Foo._@.A._hyg.7 `` and `` `Foo._@.B._hyg.7 `` in the respective module (by the way, there doesn't seem to be a `pp` option to show this - is there? I had to resort to using `Repr Syntax`) so the struct names are not really duplicate. *But* the automatic instance name ends up being `instToJsonFoo`, without macro scopes, in both cases. This causes an error on importing both into `C.lean`.

According to @Kha, this is not the desired behaviour and macro scopes should be preserved in the autogenerated name somehow. See also [Zulip](https://leanprover.zulipchat.com/#narrow/stream/341532-lean4-dev/topic/Should.20instance.20names.20inherit.20macro.20scopes.3F).

In practice, this issue arises in [the `deriving` handler](https://github.com/leanprover/lean4/blob/e477d41f3f34887df936bc29f53ce13565532aa8/src/Lean/Server/Rpc/Deriving.lean#L45) for `Server.RpcEncodable`.

### Steps to Reproduce

1. Create three modules as described above.

**Expected behavior:** Hygiene ensures that no names clash.

**Actual behavior:** We get a duplicate name error.

### Versions

`nightly-2023-01-14`

kmill:
Macros sometimes create auxiliary types and instances about them, and they rely on the instance name generate to create unique names in that case.

This modifies the automatic name generator to add a fresh macro scope to the generated name if any of the constants in the type of the instance themselves have macro scopes.

Closes #2044

### DIFF
--- a/src/Lean/Elab/DeclNameGen.lean
+++ b/src/Lean/Elab/DeclNameGen.lean
@@ -206,6 +206,9 @@ Uses heuristics to generate an informative but terse base name for a term of the
 Makes use of the current namespace.
 It tries to make these names relatively unique ecosystem-wide,
 and it adds suffixes using the current module if the resulting name doesn't refer to anything defined in this module.
+
+If any constant in `type` has a name with macro scopes, then the result will be a name with fresh macro scopes.
+While in this case we could skip the naming heuristics, we still want informative names for debugging purposes.
 -/
 def mkBaseNameWithSuffix (pre : String) (type : Expr) : MetaM Name := do
   let (name, st) ← mkBaseName type |>.run {}
@@ -222,7 +225,7 @@ def mkBaseNameWithSuffix (pre : String) (type : Expr) : MetaM Name := do
       s!"{name}{moduleToSuffix project}"
     else
       name
-  if Option.isSome <| type.find? fun e => if let .const n _ := e then n.hasMacroScopes else false then
+  if Option.isSome <| type.find? (fun e => if let .const n _ := e then n.hasMacroScopes else false) then
     mkFreshUserName name
   else
     return name


kmill:
@Kha What do you think about having the auto-generated instance names use a fresh macro scope if any of the types in the type of the instance themselves have macro scopes?

It's still using the name generator for debuggability, if you need to poke around the macro expansions.

src/Lean/Elab/DeclNameGen.lean
```diff
@@ -217,8 +217,13 @@ def mkBaseNameWithSuffix (pre : String) (type : Expr) : MetaM String := do
...
   -- We can also avoid adding the full module suffix if the instance refers to "project"-local names.
   let isProjectLocal := isModuleLocal || modules.any fun mod? => mod?.map (·.getRoot) == project
-  if !isProjectLocal then
-    return s!"{name}{moduleToSuffix project}"
+  let name := Name.mkSimple <|
+    if !isProjectLocal then
+      s!"{name}{moduleToSuffix project}"
+    else
+      name
+  if Option.isSome <| type.find? fun e => if let .const n _ := e then n.hasMacroScopes else false then
```
Kha, line 225: I would always put parentheses around such confusing lambdas

Kha (approved):

PR #5533, opened 2024-09-30
chore: deprecate variants other than `inductive ... where`

### LINKED ISSUE #5236: Deprecate and remove `inductive` ... `:=`
### Description

Today, Lean accepts all of the following:

```
inductive DogSound
  | woof
  | grr
```

```
inductive DogSound :=
  | woof
  | grr
```

```
inductive DogSound where
  | woof
  | grr
```

It seems that the overwhelming majority of declarations use the latter syntax (with `where`). In the interest of simplifying things, how about we deprecate the first two?

### Context

The broad context here is that I'm writing the new reference manual with tooling that can let us know about undocumented tokens. The documentation should be comprehensive and describe what the language does.

Deprecating these two alternate syntaxes, rather than documenting them, would make Lean code more uniform and reduce the language surface area.

I'm not advocating immediate removal - a long deprecation period won't hurt anything - but a warning would be useful.

### Impact

kmill:
**Current state:** This is a draft since there are some questions about whether no-`where` should be disallowed. #5542 deprecates just the `:=` variant.

Deprecates `inductive ... :=`. Also, makes `where` required if there is at least one constructor. Currently produces a warning, controlled by `linter.deprecated`.

We don't require `where` when there are no constructors.

Closes #5236

### DIFF
--- a/src/Init/Data/AC.lean
+++ b/src/Init/Data/AC.lean
@@ -9,7 +9,7 @@ import Init.Classical
 import Init.ByCases
 
 namespace Lean.Data.AC
-inductive Expr
+inductive Expr where
   | var (x : Nat)
   | op (lhs rhs : Expr)
   deriving Inhabited, Repr, BEq
--- a/src/Init/Core.lean
+++ b/src/Init/Core.lean
@@ -1124,7 +1124,7 @@ The transitive closure `TransGen r` of a relation `r` is the smallest relation w
 transitive and contains `r`. `TransGen r a z` if and only if there exists a sequence
 `a r b r ... r z` of length at least 1 connecting `a` to `z`.
 -/
-inductive Relation.TransGen {α : Sort u} (r : α → α → Prop) : α → α → Prop
+inductive Relation.TransGen {α : Sort u} (r : α → α → Prop) : α → α → Prop where
   /-- If `r a b` then `TransGen r a b`. This is the base case of the transitive closure. -/
   | single {a b} : r a b → TransGen r a b
   /-- The transitive closure is transitive. -/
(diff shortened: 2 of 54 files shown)


david-christiansen:
Out of curiosity, why should `where` be optional in the case of no constructors, rather than making it either mandatory or forbidden?

david-christiansen:
(where of course "forbidden" means "deprecated for a long long time" in this context)

digama0:
I think having it be optional is defensible:

* It should be allowed to be omitted because there is a legitimate aesthetic/readability argument against it (`inductive Foo where` looks like an unfinished sentence)
* It should be allowed to be included because it is more regular (it makes editing easier if lists of 0+ things can be added or removed without requiring surrounding edits as well)

That is, downstream style guides could go either way on this, and we shouldn't try to fight with either option.

Thermotronica:
Aw man

src/Init/Data/AC.lean
```diff
@@ -9,7 +9,7 @@ import Init.Classical
 import Init.ByCases
 
 namespace Lean.Data.AC
-inductive Expr
+inductive Expr where
   | var (x : Nat)
```
eric-wieser, line 13: In #5236 I thought the argument was leaning towards banning `where` for consistency with pattern-matched `def`s.

On the other hand, I guess `match _ with |` is already inconsistent when compared to `inductive _ where |` and `def _ : _ |`...

tydeu: @eric-wieser

> In #5236 I thought the argument was leaning towards banning `where` for consistency with pattern-matched `def`s.

I believe this PR has been superseded by #5542 for approximately this reason.

PR #5535, opened 2024-09-30
chore: rename Array.flatten to join, and upstream lemmas

kim-em:


david-christiansen:
What is the motivation for this? It's a closer match to Haskell, but I think most programmers would have an easy time guessing "flatten" and a difficult time guessing "join" for this operation.

digama0:
I second David's concern, `flatten` is a more intuitive name while `join` is a monad-brained name. (I think I've said similar things about `List.bind` vs `flatMap` in the past, which IMO is even worse.)

kim-em:
If everyone is happy with renaming `List.join` also to `List.flatten`, then I am perfectly happy to go the other way.

kim-em:
Replaced by #5551

PR #5541, opened 2024-09-30
fix: rm new shared libs before build for Windows

tydeu:
On Windows, shared libraries must be removed before linking. Otherwise, linking can fail with "Permission denied" when the libraries are in use. This ensures such removal is done for the new `libLake_shared.dll` and both parts of `libleanshared`.


tydeu:
@Kha As this touches the build code, a 👍 from you might be helpful.

Kha:
> @Kha As this touches the build code, a 👍 from you might be helpful.

(late :+1:)

PR #5542, opened 2024-09-30
chore: deprecate `:=` variants of `inductive` and `structure`

kmill:
Deprecates `inductive ... :=`, `structure ... :=`, and `class ... :=` in favor of the `... where` variant. Currently this syntax produces a warning, controlled by the `linter.deprecated` option.

Breaking change: modifies `Lean.Linter.logLintIf` to use `Lean.Linter.getLinterValue` to determine if a linter value is set. This means that the `linter.all` option now is taken into account when the linter option is not set.

Part of #5236


kmill:
@adomani @grunweg I noticed that `logLintIf` didn't take into account `linter.all`, and I modified the logic here. It seems to me that it was meant to respond to `linter.all`, but please let me know if it's not supposed to.

adomani:
Kyle, thanks for the heads up!  I think that I never used `logLintIf` (at least not directly), but I'll keep an eye out for linters misbehaving!  :smile:

PR #5550, opened 2024-10-01
feat: List.unattach and simp lemmas

kim-em:


src/Init/Data/List/Attach.lean
```diff
@@ -548,4 +548,131 @@ theorem count_attachWith [DecidableEq α] {p : α → Prop} (l : List α) (H : 
...
+We use it by providing a simp lemma `l.attach.unattach = l`, and simp lemmas which recognize higher order
+functions applied to `l : List { x // p x }` which only depend on the value, not the predicate, and rewrite these
+in terms of a simpler function applied to `l.unattach`.
+
+Further, we provide simp lemmas that push `unattach` inwards.
+-/
+
+/--
+A synonym for `l.map (·.val)`. Mostly this should not be needed by users.
+It is used introduced as in intermediate step by lemmas such as `map_subtype`,
```
nomeata, line 564:
```suggestion
It is introduced as an intermediate step by lemmas such as `map_subtype`,
```

src/Init/Data/List/Attach.lean
```diff
@@ -548,4 +548,131 @@ theorem count_attachWith [DecidableEq α] {p : α → Prop} (l : List α) (H : 
...
+
+@[simp] theorem unattach_attachWith {α : Type _} {p : α → Prop} {l : List α}
+    {H : ∀ a ∈ l, p a} :
+    (l.attachWith p H).unattach = l := by
+  unfold unattach
+  induction l with
+  | nil => simp
+  | cons a l ih => simp [ih, Function.comp_def]
+
+/-! ### Recognizing higher order functions using a function that only depends on the value. -/
```
nomeata, line 594:
```suggestion
/-! ### Recognizing higher order functions on subtypes using a function that only depends on the value. -/
```

src/Init/Data/List/Attach.lean
```diff
@@ -548,4 +548,131 @@ theorem count_attachWith [DecidableEq α] {p : α → Prop} (l : List α) (H : 
...
+
+/--
+This lemma identifies folds over lists of subtypes, where the function only depends on the value, not the proposition,
+and simplifies these to the function directly taking the value.
+-/
+@[simp] theorem foldl_subtype {p : α → Prop} {l : List { x // p x }}
+    {f : β → { x // p x } → β} {g : β → α → β} {x : β}
+    {hf : ∀ b x, f b x = g b x.1} :
```
nomeata, line 602: Just a gut feeling, but I'd expect this (in particular the final higher order matching done by simp) is more reliable if your bind `x, h` separately, and pass a concrete constructor to f (`f b <x, h>`)

tests/lean/run/nested_inductive.lean
```diff
@@ -0,0 +1,41 @@
...
+  | node ts => .node (ts.attach.reverse.map (fun ⟨t, _⟩ => t.rev))
+
+-- Note that `simp` now automatically removes the `attach`.
+@[simp] theorem rev_def (ts : List Tree) :
+    Tree.rev (.node ts) = .node (ts.reverse.map rev) := by
+  simp [Tree.rev]
+
+/-- Define `size` using a `foldl` over `attach`. -/
+def size : Tree → Nat
+  | node ts => 1 + ts.attach.foldl (fun acc ⟨t, _⟩ => acc + t.size) 0
```
nomeata, line 25: Let's maybe also test the variant where the user didn't use pattern matching but `x.val` in the definition.

src/Init/Data/List/Attach.lean
```diff
@@ -548,4 +548,131 @@ theorem count_attachWith [DecidableEq α] {p : α → Prop} (l : List α) (H : 
...
+-/
+
+/--
+A synonym for `l.map (·.val)`. Mostly this should not be needed by users.
+It is introduced as an intermediate step by lemmas such as `map_subtype`,
+and is ideally subsequently simplified away by `unattach_attach`.
+
+If not, usually the right approach is `simp [List.unattach, -List.map_subtype]` to unfold.
+-/
+def unattach {α : Type _} {p : α → Prop} (l : List { x // p x }) := l.map (·.val)
```
digama0, line 569: Like `attach`, this needs an `implemented_by` version which uses an `unsafeCast`

PR #5552, opened 2024-10-01
chore: switch obvious cases of array "bang"`[]!` indexing to rely on hypothesis

TomasPuverle:
Update certain uses of `arr[i]!` to use the "provably correct" version `arr[i]`, in order to use "best practices".

Some motivation and discussion on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Lean.20compiler.2Felaborator.20development.20question/near/472934715)

### DIFF
--- a/src/Lean/ParserCompiler.lean
+++ b/src/Lean/ParserCompiler.lean
@@ -73,7 +73,7 @@ partial def compileParserExpr (e : Expr) : MetaM Expr := do
       forallTelescope ty fun params _ => do
         let mut p := mkConst p
         let args  := e.getAppArgs
-        for h: i in [:Nat.min params.size args.size] do
+        for i in [:Nat.min params.size args.size] do
           let param := params[i]!
           let arg   := args[i]!
           let paramTy ← inferType param


TomasPuverle:
The changes themselves are trivial and purely mechanistic, and I deliberately avoided changing places where multiple arrays were involved - my plan is to address this separately.

However, working through this I noticed a few "paper cuts", which I posted about on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/Indexing.20hypothesis.20observations/near/473969520)

TomasPuverle:
I am happy to switch to any naming preference but there didn't seem to be a consistent convention that I could see.  Existing code typically used `h`, so I stuck with it.

nomeata:
I think it has been tried, and caused problems with large definitions, and maybe also with tactics like `case` that revert some hypotheses picking it up when it shouldn't. I wouldn't rule it out, but I don't thinks is an easy route.

TomasPuverle:
I had the same thought!  Just like with implicit arguments, I think it would be interesting to be able to introduce "implicit return values", e.g.

```
def foo (...) : Int { h: <some value> }
```

TomasPuverle:
> I think it has been tried, and caused problems with large definitions, and maybe also with tactics like `case` that revert some hypotheses picking it up when it shouldn't. I wouldn't rule it out, but I don't thinks is an easy route.

Interesting!

PR #5558, opened 2024-10-01
fix: make lean.h compile with MSVC

### LINKED ISSUE #5557: Lean.h cannot be compiled using MSVC
### Prerequisites

### Description

Basically I want to compile Lean script, and provide CL.exe as C compiler.

### Context

I know that this is probably not primary support target, but I lost in Mingw/Cygwin installations from different languages, so I prefer to use MSVC where possible.

### Steps to Reproduce

1. Compile script with C FFI.

**Expected behavior:** Want that there no errors from lean.h produced by MSVC.h

**Actual behavior:** .some crappy errors, can rollback changes and give exact errors.

### Versions

[Output of `#eval Lean.versionString`]
[OS version, if not using live.lean-lang.org.]

### Additional Information

### Impact

kant2002:
Closes #5557

### DIFF
--- a/src/include/lean/lean.h
+++ b/src/include/lean/lean.h
@@ -27,8 +27,11 @@ extern "C" {
 
 #ifdef _MSC_VER
 #define LEAN_ALLOCA(s) _alloca(s)
+#include <stdnoreturn.h>
+#define LEAN_NORETURN _Noreturn
 #else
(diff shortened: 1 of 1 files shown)


nomeata (approved):
Looks innocent enough

PR #5559, opened 2024-10-01
doc: simplify one step in installation instructions

fpvandoorn:
The ∀-symbol is always visible, so no need to create an empty file anymore before clicking it.


fpvandoorn:
awaiting-review

TwoFX:
>  The ∀-symbol is always visible

I don't think this is true. If no file is open, then there is no forall symbol:
image

fpvandoorn:
Ah, good point. When testing I had the `Welcome` screen or the `Extension: Lean 4` page open, which both count as "having a file open".
So in most cases the user will have a file open, but for full robustness, it's might be good to keep this step. Feel free to close this PR without merging.

PR #5563, opened 2024-10-01
fix: getFunInfo, inferType to use withAtLeastTransparency, not withTransparency

### LINKED ISSUE #2194: `infer_type` fails to unfold irreducible constants even when transparency is `all`
When checking whether a term is a function, `infer_type` performs `whnf` ([here](https://github.com/leanprover/lean4/blob/8a302e6135bc1b0f1f2901702664c56cd424ebc2/src/Lean/Meta/InferType.lean#L81)) at `default` transparency ([here](https://github.com/leanprover/lean4/blob/8a302e6135bc1b0f1f2901702664c56cd424ebc2/src/Lean/Meta/InferType.lean#L193)), regardless of the ambient transparency. This means that if the ambient transparency is `all`, `infer_type` still refuses to unfold irreducible constants, leading to failures like this one:

```lean
@[irreducible] def T := Nat → Nat

example (h : T) (n : Nat) : True := by
  with_unfolding_all have : Nat := h n
  /-
  function expected
    h n
  -/
```

I believe the solution would be to change the `withTransparency` to `withAtLeastTransparency`. I imagine we also need to execute `infer_type` with an empty `infer_type` cache if the ambient transparency is `all`.

Alternatively, Kyle Miller suggested that `infer_type` should maybe always run with `all` transparency, mirroring `check`.

Zulip discussion: https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/inferType.20does.20not.20respect.20ambient.20transparency

### LINKED ISSUE #2975: `invalid projection` from `Meta.reduce`
### Prerequisites

### Description

The following code snippet (thanks @kmill) unexpectedly produces the error `invalid projection s.1`:

```lean
import Lean
open Lean Meta Elab Term

elab "#reduce'" t:term : command => Elab.Command.runTermElabM fun _ => do
  let e ← withSynthesize <| elabTerm t none
  let e ← Meta.reduce e
  logInfo m!"{← inferType e}"

structure S where
  x : Nat

def S' := S
def S'.x (s : S') : Nat := S.x s

attribute [irreducible] S'

variable (s : S')
#reduce' s.x -- invalid projection s.1
```

### Context

This issue has been [discussed on Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/invalid.20projection). The problem originally arose in the following code snippet:

```lean
import Mathlib.AlgebraicTopology.DoldKan.SplitSimplicialObject
import Lean
open Lean Meta

elab "#fails" : term => do
  let info ← getConstInfo ``SimplicialObject.Splitting.πSummand_comp_ιSummand_comp_PInfty_eq_PInfty
  forallTelescopeReducing info.type fun _ body => do
    match body.eq? with
    | none => return mkNatLit 0
    | some (_, lhs, _) => 
      let _ ← Meta.reduceAll lhs
      return mkNatLit 0

#eval #fails -- invalid projection g.1
```

The intent was to get the type of `SimplicialObject.Splitting.πSummand_comp_ιSummand_comp_PInfty_eq_PInfty`, extract its underlying equality, and reduce the type of each side of the equality. This produces an error that is hard to understand as it references a term `g` whose origin isn't immediately clear. 

### Steps to Reproduce

Run the code snippet shown in the description using Lean version `4.3.0-rc2`.

**Expected behavior:** The command `#reduce' s.x` prints the type `Nat`.

**Actual behavior:** The command `#reduce' s.x` fails with the error `invalid projection s.1`.

### Versions

```lean
#eval Lean.versionString -- "4.3.0-rc2"
```

```
$ sw_vers
ProductName:		macOS
ProductVersion:		14.0
BuildVersion:		23A344
```

### Impact

### LINKED ISSUE #5562: Meta.inferType fails at transparency all
The `Meta.inferType` function sets the transparency to `.default`, even if it was set to `.all`, preventing “kernel-close” meta-programming where one wants all definitions to be transparent.

Presumably using `withAtLeastTransparency` is the right fix here (PR coming).

### Impact

nomeata:
when the transparency mode is `.all`, then one expects `getFunInfo` and
`inferType` to also work with that transparency mode.

Fixes #5562
Fixes #2975 
Fixes #2194

### DIFF
--- a/stage0/src/stdlib_flags.h
+++ b/stage0/src/stdlib_flags.h
@@ -6,7 +6,7 @@ options get_default_options() {
     // see https://lean-lang.org/lean4/doc/dev/bootstrap.html#further-bootstrapping-complications
 #if LEAN_IS_STAGE0 == 1
     // switch to `true` for ABI-breaking changes affecting meta code
-    opts = opts.update({"interpreter", "prefer_native"}, false);
+    opts = opts.update({"interpreter", "prefer_native"}, true);
     // switch to `true` for changing built-in parsers used in quotations
     opts = opts.update({"internal", "parseQuotWithCurrentStage"}, false);
     // toggling `parseQuotWithCurrentStage` may also require toggling the following option if macros/syntax


nomeata:
Ah, finally. Prefer_native helped.

PR #5564, opened 2024-10-01
fix: remove non-conforming size-0 arrays

eric-wieser:
In C, these are supported only as a vendor extension; they should instead use proper C99 flexible array members.

In C++, both `[]` and `[0]` are vendor extensions.

### DIFF
--- a/src/include/lean/lean.h
+++ b/src/include/lean/lean.h
@@ -148,7 +148,7 @@ typedef lean_object * b_lean_obj_res; /* Borrowed object result. */
 
 typedef struct {
     lean_object   m_header;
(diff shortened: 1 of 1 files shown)


tkoeppe:
We wanted to note separately that the header `lean.h`, despite best efforts, is not actually conforming as a C and C++ interop header, since neither zero-sized nor flexible array members are conforming C++.

This change is still useful to make non-conforming C into conforming C, but as an interop header, the header should not contain these type definitions at all, and only declare opaque functions that can be used from both languages. The struct types with flexible array members should only exist in the C code.

PR #5566, opened 2024-10-01
fix: use `MessageData.tagged` to mark maxHeartbeat exceptions

### LINKED ISSUE #5565: Exception.isMaxHeartbeat no longer works
### Prerequisites

### Description

`Exception.isMaxHeartbeats` returns false on the exception thrown by `Lean.Core.throwMaxHeartbeat`

### Context

This means that `try`/`catch` now catches heartbeat errors, which means many tactics now ignore the heartbeat counter.

### Steps to Reproduce

1. Run
```lean
import Lean

run_meta
  try
    Lean.Core.throwMaxHeartbeat `foo `bar 200
  catch e =>
    Lean.logWarning m!"{e.toMessageData}"
    unless e.isMaxHeartbeat do
      throwError "oh no"
```

**Expected behavior:** The exception should not be caught

**Actual behavior:** A heartbeat exception is thrown, but `e.isMaxHeartbeat` returns false.

### Versions

* 4.11.0
* 4.12.0-nightly-2024-10-01

### Additional Information

### Impact

eric-wieser:
Fixes #5565, by using tags instead of trying to string match on a `MessageData`. This ends up reverting some unwanted test output changes from #4781 too. 

This changes `isMaxRecDepth` for good measure too.

This was a regression in Lean 4.11.0, so may be worth backporting to 4.12.x, if not also 4.11.x.


eric-wieser:
I'm fighting slightly with reverting relevant test changes from #4781, as I don't have a gitpod build working; but other than that this should be reviewable.

kmill (approved):
@kim-em This PR fixes something I broke in #4781; I didn't realize that isMaxHeartbeat checked for the exact text of the error message, and it's an important part of the exception handling architecture. We probably should backport this to 4.12. At least the system normally avoids bumping into heartbeat errors, so it's not the highest of priorities, but users running into heartbeat errors might see confusing states.

PR #5569, opened 2024-10-01
chore: remove `matcherExt` cache

Kha:
Incompatible as is with parallelism; let's first check if it has any impact at all


Kha:
!bench

Kha:
The test failures are worrying and indicate that removing this extension may require changes to how `simp` (etc?) unifies matchers: structurally instead of by underlying name. In any case, my current plan is to leave things as is so that matchers are reused on the same elaboration branch (e.g. between `def` and `theorem` proof, until we want to make `def` parallel as well) but not between independent elaboration branches.

PR #5571, opened 2024-10-01
chore: remove `SplitIf.ext` cache

Kha:
Incompatible as is with parallelism; let's first check if it has any impact at all


Kha:
!bench

Kha:
Mathlib bench is insignificant as well: http://speed.lean-fro.org/mathlib4/run-detail/736f134c-465b-45d6-b096-bb5f2f5ac81c

src/Lean/LazyInitExtension.lean
```diff
@@ -1,42 +0,0 @@
...
-Copyright (c) 2021 Microsoft Corporation. All rights reserved.
-Released under Apache 2.0 license as described in the file LICENSE.
-Authors: Leonardo de Moura
--/
-prelude
-import Lean.MonadEnv
-
-namespace Lean
-
-structure LazyInitExtension (m : Type → Type) (α : Type) where
```
Kha, line 11: @leodemoura I removed this abstraction as it's now unused and doing lazy initialization *in an elaboration monad* seems like a bad idea in general under parallelism as the elaboration context will not be deterministic. Sounds good?

Issue #5544, opened 2024-09-30
bv_check builds a proof with an incorrect application of Lean.ofReduceBool


bollu:
### Description

Consider the MWE, where the LRAT cerificate comes from [leanprover/lnsym](https://github.com/leanprover/LNSym/blob/main/Arm/Memory/lrat_files/SeparateProofs.lean-mem_subset_refl-77-2.lrat)

```lean
import Lean

-- "4.12.0-nightly-2024-09-30"
#eval Lean.versionString

def mem_subset (a1 a2 b1 b2 : BitVec 64) : Bool :=
  (b2 - b1 = BitVec.ofNat 64 (2^64 - 1)) ||
  ((a2 - b1 <= b2 - b1 && a1 - b1 <= a2 - b1))

theorem mem_subset_refl : mem_subset a1 a2 a1 a2 := by
  unfold mem_subset
  bv_check "/Users/sidbhatz/temp/mem_subset_refl.lrat"
/-
application type mismatch
  Lean.ofReduceBool mem_subset_refl._reflection_def_1 true (Eq.refl true)
argument has type
  true = true
but function has type
  Lean.reduceBool mem_subset_refl._reflection_def_1 = true → mem_subset_refl._reflection_def_1 = true
-/
```

### Context

This occurred when bumping `leanprover/lnsym` to the latest nightly.

### Impact

Add :+1: to [issues you consider important](https://github.com/leanprover/lean4/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc). If others are impacted by this issue, please ask them to add :+1: to it.

hargoniX:
This is a UX bug, `bv_check` is doing the right thing but checking the certificate ends up yielding something wrong. I assume that this is because you have not updated the certificate after you updated to the new Lean version. What that means is that now the bitblaster generates a CNF based on the output of the improved rewriter and tries to validate that that CNF is UNSAT with the old certificate which was for an old CNF. The kernel then reports this rather confusing type checking issue because it doesn't really have a way to say what's wrong here apart from this.

PR #5574, opened 2024-10-01
feat: error message+suggestion when bv_check fails to build a legal Lean proof

bollu:
Closes https://github.com/leanprover/lean4/issues/5544, by instructing the user of what's likely going wrong. Picture of the new result:

image


hargoniX:
This patch would half performance for every user that does not have a failing cached certificate. This doesn't seem like the way to go.

PR #5580, opened 2024-10-02
fix: lake: do not fail `--no-build` on fetch

tydeu:
Lake's `--no-build` option will no longer exit with an error if the build action in question is simply a cache fetch.


tydeu:
Closing in favor of #5583.

PR #5585, opened 2024-10-02
feat: extend `if h: ` hypothesis scope after early return

TomasPuverle:
Extend the scope of hypothesis `if h: ...` inside of `do` notation after a `return`, `continue` and `break`.  This PR makes it easier to work with "early return" code.

The tests demonstrate the usefulness but here's a MWE:

```
def testSingleArm (val : Nat) : Nat := Id.run do
  if h: val != 0 then
    return 0
  -- h not in scope here
  return 1
```

```
def safeDiv (a b : Nat) (_ : b ≠ 0): Nat := a / b

def earlyReturn (a b : Nat) : Option Nat := Id.run do
  if zh : b = 0 then
    return none

  return safeDiv a b zh

def earlyBreak : Nat := Id.run do
  let mut sum := 0
  for val in [2, 0, 1] do
    if zh : val = 0 then
      break

    sum := sum + safeDiv 10 val zh

  return sum

def earlyContinue : Nat := Id.run do
  let mut sum := 0
  for val in [1, 0, 2] do
    if zh : val = 0 then
      continue

    sum := sum + safeDiv 10 val zh

  return sum
```

Here's some discussion on [Zulip](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/RFC.3A.20If.20hypothesis.20scope/near/474189980) and additional motivation in this RFC #5598

### DIFF
--- a/src/Lean/Elab/Do.lean
+++ b/src/Lean/Elab/Do.lean
@@ -1483,6 +1483,15 @@ mutual
     else
       throwError "unexpected kind of `do` reassignment"
 
+  private partial def emptyElse? (elseBranch : List Syntax) : M Bool  := do
+    if elseBranch.isEmpty then
+      return true
+    else if let some expr := isDoExpr? (elseBranch.head!) then
+      let pexp ← ``(pure PUnit.unit)
+      return expr == pexp.raw
+
+    return false
+
   /-- Generate `CodeBlock` for `doIf; doElems`
      `doIf` is of the form
      ```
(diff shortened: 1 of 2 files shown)


Kha:
This module is currently frozen pending a rewrite, so I'll close this PR. Please open an RFC prior to a PR to get feedback on whether a contribution is desired.

TomasPuverle:
Hi @Kha and @nomeata,

Thanks for your replies.  I think someone mentioned that this part of the code is frozen after I posted the PR.  I wasn't really expecting for the code to get merged but for the purpose of a RFC, I started a discussion on Zulip (linked above) and used this PR as a "proof of concept" for the idea (including the tests to demonstrate it)

If this code is getting rewritten, please LMK if I can help at some point!

TomasPuverle:
> Is that worth it, when you can just be explicit and use else.

Ideally yes but as I pointed out in the motivation & Zulip discussion, this doesn't seem to be the case, unfortunately, even the source of Lean itself.  (It's also sort of related to my RFC about `unless h: ...`: #5598 )

Kha:
>  I wasn't really expecting for the code to get merged but for the purpose of a RFC, I started a discussion on Zulip (linked above) and used this PR as a "proof of concept" for the idea (including the tests to demonstrate its benefits)

You can mark PRs as drafts to avoid any confusion on this point :)

PR #5589, opened 2024-10-02
fix: bug in apply? when using intro

kim-em:


kim-em:
Closing: the code has been significantly reworked on master since this branch was created, and the bug no longer reproduces. The `checkState?` validation mechanism added in the interim addresses suggestion correctness differently.

PR #5590, opened 2024-10-02
feat: adding Insert/Singleton/Union instances for HashMap/Set.Raw

kim-em:
These were missing from https://github.com/leanprover/lean4/pull/5581.

### DIFF
--- a/src/Std/Data/DHashMap/Basic.lean
+++ b/src/Std/Data/DHashMap/Basic.lean
@@ -79,6 +79,9 @@ instance : Singleton ((a : α) × β a) (DHashMap α β) := ⟨fun ⟨a, b⟩ =>
 
 instance : Insert ((a : α) × β a) (DHashMap α β) := ⟨fun ⟨a, b⟩ s => s.insert a b⟩
 
+instance : LawfulSingleton ((a : α) × β a) (DHashMap α β) :=
+  ⟨fun _ => rfl⟩
+
 @[inline, inherit_doc Raw.insertIfNew] def insertIfNew (m : DHashMap α β)
     (a : α) (b : β a) : DHashMap α β :=
   ⟨Raw₀.insertIfNew ⟨m.1, m.2.size_buckets_pos⟩ a b, .insertIfNew₀ m.2⟩
--- a/src/Std/Data/DHashMap/Raw.lean
+++ b/src/Std/Data/DHashMap/Raw.lean
@@ -71,6 +71,9 @@ instance [BEq α] [Hashable α] : Singleton ((a : α) × β a) (Raw α β) :=
 instance [BEq α] [Hashable α] : Insert ((a : α) × β a) (Raw α β) :=
   ⟨fun ⟨a, b⟩ s => s.insert a b⟩
 
+instance [BEq α] [Hashable α] : LawfulSingleton ((a : α) × β a) (Raw α β) :=
+  ⟨fun _ => rfl⟩
+
 /--
 If there is no mapping for the given key, inserts the given mapping into the map. Otherwise,
 returns the map unaltered.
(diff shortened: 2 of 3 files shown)


src/Std/Data/HashMap/Raw.lean
```diff
@@ -231,10 +235,20 @@ m.inner.values
...
+@[inline] def union [BEq α] [Hashable α] (m₁ m₂ : Raw α β) : Raw α β :=
+  m₂.fold (init := m₁) fun acc x => acc.insert x
+
+instance [BEq α] [Hashable α] : Union (Raw α β) := ⟨union⟩
+
 @[inline, inherit_doc DHashMap.Raw.Const.unitOfList] def unitOfList [BEq α] [Hashable α]
     (l : List α) : Raw α Unit :=
   ⟨DHashMap.Raw.Const.unitOfList l⟩
 
+@[inline, inherit_doc DHashMap.Raw.Const.unitOfArray] def unitOfArray [BEq α] [Hashable α]
```
TwoFX, line 248: Did this split in by accident?

kim-em: I noticed that `ofArray` was missing below -- just because I was putting `union` directly below `ofArray`, and then my anchor wasn't there.

src/Std/Data/DHashMap/Lemmas.lean
```diff
@@ -87,6 +87,12 @@ theorem isEmpty_iff_forall_not_mem [EquivBEq α] [LawfulHashable α] :
     m.isEmpty = true ↔ ∀ a, ¬a ∈ m := by
   simpa [mem_iff_contains] using isEmpty_iff_forall_contains
 
+@[simp] theorem insert_eq_insert {p : (a : α) × β a} : Insert.insert p m = m.insert p.1 p.2 := rfl
+
+@[simp] theorem singleton_eq_insert {p : (a : α) × β a} :
+    Singleton.singleton p = (∅ : DHashMap α β).insert p.1 p.2 :=
+  rfl
+
```
TwoFX, line 95: Sorry for scope-screeping this even more, but what do you think of providing a `LawfulSingleton` instance here? The law should hold by `rfl`.

src/Std/Data/HashMap/Raw.lean
TwoFX: Ah, I hadn't seen #5369 because I was on vacation. Makes sense, thanks.

src/Std/Data/HashMap/Raw.lean
```diff
@@ -74,6 +74,12 @@ set_option linter.unusedVariables false in
     (a : α) (b : β) : Raw α β :=
   ⟨m.inner.insert a b⟩
 
+instance [BEq α] [Hashable α] : Singleton (α × β) (Raw α β) := ⟨fun ⟨a, b⟩ => Raw.empty.insert a b⟩
+
+instance [BEq α] [Hashable α] : Insert (α × β) (Raw α β) := ⟨fun ⟨a, b⟩ s => s.insert a b⟩
+
+instance [BEq α] [Hashable α] : LawfulSingleton (α × β) (Raw α β) := ⟨fun _ => rfl⟩
```
TwoFX, line 81: It looks like you only added two `LawfulSingleton` instances, but there should be six in total.

src/Lean/Elab/Tactic/LibrarySearch.lean
```diff
@@ -24,8 +24,8 @@ Implementation of the `exact?` tactic.
 def exact? (ref : Syntax) (required : Option (Array (TSyntax `term))) (requireClose : Bool) :
     TacticM Unit := do
   let mvar ← getMainGoal
-  let (_, goal) ← (← getMainGoal).intros
-  goal.withContext do
+  let (_, goal) ← mvar.intros
+  mvar.withContext do
```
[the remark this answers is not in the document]
kim-em: Ugh, sorry. I've missed the boat for `v4.13.0-rc1` in any case, I'll deal with this in less of a hurry tomorrow.

PR #5592, opened 2024-10-02
chore: simp lemmas for HashMap Insert/Singleton

kim-em:


kim-em:
No point having this separate, I'll roll these changes into #5590.

PR #5594, opened 2024-10-02
chore: induction-friendly `List.min?_cons`

TwoFX:
@kim-em, I'm happy to keep any subset of `foldl_min`, `foldl_min_right`, `foldl_min_le`, `foldl_min_min_of_le` (should that one have been called `foldl_min_le_of_le`?). Which ones do you like?


kim-em:
I like `foldl_min`. The rest do seem a bit lame, and I guess can go.

PR #5604, opened 2024-10-03
feat: complete `BitVec.[getMsbD|getLsbD|msb]` for shifts

luisacicolini:


src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1181,6 +1209,7 @@ theorem ushiftRight_eq' (x : BitVec w₁) (y : BitVec w₂) :
 
 /-! ### sshiftRight -/
 
+
 theorem sshiftRight_eq {x : BitVec n} {i : Nat} :
```
tobiasgrosser, line 1213:
```suggestion
theorem sshiftRight_eq {x : BitVec n} {i : Nat} :
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1322,11 +1351,44 @@ theorem not_sshiftRight_not {x : BitVec w} {n : Nat} :
...
+  all_goals (simp [h, h₁, h₂, h₃, h₄]; try congr; try omega)
+  simp_all
+
 /-! ### sshiftRight reductions from BitVec to Nat -/
 
 @[simp]
 theorem sshiftRight_eq' (x : BitVec w) : x.sshiftRight' y = x.sshiftRight y.toNat := rfl
 
+theorem getLsbD_sshiftRight' (x y: BitVec w) {i : Nat} :
+    getLsbD (x.sshiftRight' y) i =
```
tobiasgrosser, line 1372:
```suggestion
theorem getLsbD_sshiftRight' {x y: BitVec w} {i : Nat} :
    getLsbD (x.sshiftRight' y) i =
```

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1910,6 +1976,11 @@ theorem shiftLeft_add_distrib {x y : BitVec w} {n : Nat} :
   case succ n ih =>
     simp [ih, toNat_eq, Nat.shiftLeft_eq, ← Nat.add_mul]
 
+theorem add_eq_xor {a b : BitVec 1} : a + b = a ^^^ b := by
+  have ha : a = 0 ∨ a = 1 := eq_zero_or_eq_one _
+  have hb : b = 0 ∨ b = 1 := eq_zero_or_eq_one _
+  rcases ha with h | h <;> (rcases hb with h' | h' <;> (simp [h, h']))
```
tobiasgrosser, line 1982: These and the following ones are unrelated, no? Maybe you need to merge with the latest master?

luisacicolini: right!

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -2859,4 +2940,7 @@ abbrev zeroExtend_truncate_succ_eq_zeroExtend_truncate_or_twoPow_of_getLsbD_true
 @[deprecated and_one_eq_setWidth_ofBool_getLsbD (since := "2024-09-18")]
 abbrev and_one_eq_zeroExtend_ofBool_getLsbD := @and_one_eq_setWidth_ofBool_getLsbD
 
+@[deprecated sshiftRight_msb_eq_msb (since := "2024-10-03")]
+abbrev msb_sshiftRight := @sshiftRight_msb_eq_msb
+
```
tobiasgrosser, line 2944: You want to rename the implementation of `sshiftRight_msb_eq_msb` to `msb_sshiftRight` and then express `sshiftRight_msb_eq_msb` as `msb_sshiftRight`.

tobiasgrosser:
Also, I feel you can have a slightly shorter title, e.g., `feat: complete bitVec.[getMsbD|getLsbD|msb] for shifts`

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -463,6 +463,16 @@ theorem toInt_pos_iff {w : Nat} {x : BitVec w} :
...
 
+theorem eq_zero_or_eq_one (a : BitVec 1) : a = 0#1 ∨ a = 1#1 := by
+  obtain ⟨a, ha⟩ := a
+  simp only [Nat.reducePow]
+  have acases : a = 0 ∨ a = 1 := by omega
+  rcases acases with ⟨rfl | rfl⟩
+  · simp
+  · case inr h =>
+    subst h
+    simp
```
tobiasgrosser, line 474: This is also unrelated.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1173,6 +1173,24 @@ theorem toNat_ushiftRight_lt (x : BitVec w) (n : Nat) (hn : n ≤ w) :
...
   · apply Nat.pow_pos (by decide)
 
+theorem getMsbD_ushiftRight {x : BitVec w} {i n : Nat} :
+    getMsbD (x.ushiftRight n) i = (decide (i < w) && if i < n then false else getMsbD x (i - n)) := by
+  simp only [getMsbD, Bool.if_false_left]
+  by_cases h : i < w
+  <;> by_cases h₁ : i < n
+  <;> by_cases h₂ : i - n < w
+  all_goals (simp [h, h₁, h₂]; try congr; try omega)
+  rw [BitVec.getLsbD_ge]
```
tobiasgrosser, line 1183: This simp is non-terminal. I am afraid you may want to turn it into a `simp only`.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1322,11 +1340,44 @@ theorem not_sshiftRight_not {x : BitVec w} {n : Nat} :
...
+    getMsbD (x.sshiftRight n) i = (decide (i < w) && if i < n then x.msb else getMsbD x (i - n)) := by
+  simp only [getMsbD]
+  rw [BitVec.getLsbD_sshiftRight]
+  by_cases h : i < w
+  <;> by_cases h₁ : w ≤ w - 1 - i
+  <;> by_cases h₂ : ¬(i < n)
+  <;> by_cases h₃ : n + (w - 1 - i) < w
+  <;> by_cases h₄ : i - n < w
+  all_goals (simp [h, h₁, h₂, h₃, h₄]; try congr; try omega)
+  simp_all
```
tobiasgrosser, line 1353: I am afraid this simp is non-terminal, so you may want to turn it into a simp only. You can try to be strategic and only include the minimal set of theorems needed to get it through.

Also, do you really need a simp_all at the end?

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -2914,7 +2914,7 @@ abbrev zeroExtend_truncate_succ_eq_zeroExtend_truncate_or_twoPow_of_getLsbD_true
 @[deprecated and_one_eq_setWidth_ofBool_getLsbD (since := "2024-09-18")]
 abbrev and_one_eq_zeroExtend_ofBool_getLsbD := @and_one_eq_setWidth_ofBool_getLsbD
 
-@[deprecated sshiftRight_msb_eq_msb (since := "2024-10-03")]
-abbrev msb_sshiftRight := @sshiftRight_msb_eq_msb
+@[deprecated msb_sshiftRight (since := "2024-10-03")]
```
tobiasgrosser, line 2917: Does this not deprecate `msb_sshiftRight`?

luisacicolini: not AFAIU from the docs: 
The attribute @[deprecated] on a declaration indicates that the declaration is discouraged for use in new code, and/or should be migrated away from in existing code. It may be removed in a future version of the library.
@[deprecated myBetterDef] means that myBetterDef is the suggested replacement.
@[deprecated myBetterDef "use myBetterDef instead"] allows customizing the deprecation message.
@[deprecated (since := "2024-04-21")] records when the deprecation was first applied.

tobiasgrosser: Right.

luisacicolini:
I am having some trouble refactoring the proofs of `getMsbD_sshiftRight` and `getMsbD_ushiftRight` to use `simp only` when non terminal. `getMsbD_sshiftRight_exp` and `getMsbD_ushiftRight_exp` are my attempts at this, which however made the proofs quite a lot longer and less elegant. I am wondering whether there are any proving strategies that I did not think about and would make these proofs better.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1173,6 +1173,35 @@ theorem toNat_ushiftRight_lt (x : BitVec w) (n : Nat) (hn : n ≤ w) :
...
+  <;> simp (discharger := omega) [h, h₁, h₂]
+  congr; omega
+
+theorem getMsbD_ushiftRight_exp {x : BitVec w} {i n : Nat} :
+    getMsbD (x.ushiftRight n) i = (decide (i < w) && if i < n then false else getMsbD x (i - n)) := by
+  simp only [getMsbD, Bool.if_false_left]
+  by_cases h : i < w
+  <;> by_cases h₁ : i < n
+  <;> by_cases h₂ : i - n < w
+  all_goals (simp only [h, decide_False, ushiftRight_eq, getLsbD_ushiftRight, Bool.false_and, h₁,
```
bollu, line 1191: This looks fairly aggressive, and wonder if there are higher-level lemmas to be extracted here.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1173,6 +1173,35 @@ theorem toNat_ushiftRight_lt (x : BitVec w) (n : Nat) (hn : n ≤ w) :
...
   · apply Nat.pow_pos (by decide)
 
+theorem getMsbD_ushiftRight {w} {x : BitVec w} {i n : Nat} :
+    getMsbD (x.ushiftRight n) i = (decide (i < w) && if i < n then false else getMsbD x (i - n)) := by
+  simp only [getMsbD, ushiftRight_eq, getLsbD_ushiftRight, Bool.if_false_left]
+  by_cases h : i < w
+  <;> by_cases h₁ : i < n
+  <;> by_cases h₂ : i - n < w
+  <;> simp (discharger := omega) [h, h₁, h₂]
+  congr; omega
```
tobiasgrosser, line 1183:
```suggestion
theorem getMsbD_ushiftRight {x : BitVec w} {i n : Nat} :
    getMsbD (x.ushiftRight n) i = (decide (i < w) && if i < n then false else getMsbD x (i - n)) := by
  simp only [getMsbD, Bool.if_false_left]
  by_cases h : i < n
  · simp [getLsbD_ge, show w ≤ (n + (w - 1 - i)) by omega]
    omega
  · by_cases h₁ : i < w <;> by_cases h₂ : i - n < w
    · (simp only [h, ushiftRight_eq, getLsbD_ushiftRight, h₁, h₂]; congr; omega)
    all_goals (simp [h, h₁, h₂]; try omega)
```

tobiasgrosser: Maybe you can refactor the other cases in similar ways?

luisacicolini:
I changed the proof `getMsbD_sshiftRight` and it seems to me slightly better (if I understood correctly the rationale behind `simp` and `simp only` - CC @hargoniX). I left the two other versions as well (`getMsbD_sshiftRight'` and `getMsbD_sshiftRight''`) in case either of the two is preferrable.

src/Init/Data/BitVec/Lemmas.lean
```diff
@@ -1183,6 +1183,47 @@ theorem toNat_ushiftRight_lt (x : BitVec w) (n : Nat) (hn : n ≤ w) :
...
+
+theorem getMsbD_ushiftRight_exp2 {x : BitVec w} {i n : Nat} :
+    getMsbD (x.ushiftRight n) i = (decide (i < w) && if i < n then false else getMsbD x (i - n)) := by
+  simp only [getMsbD, Bool.if_false_left]
+  by_cases h : i < n
+  · simp [getLsbD_ge, show w ≤ (n + (w - 1 - i)) by omega]
+    omega
+  · by_cases h₁ : i < w
+    · simp only [h₁, decide_True, ushiftRight_eq, getLsbD_ushiftRight, Bool.true_and, h,
+      decide_False, Bool.not_false, show i - n < w by omega]
```
tobiasgrosser, line 1214:
```suggestion
    ·  simp only [h, ushiftRight_eq, getLsbD_ushiftRight, show i - n < w by omega]
```

tobiasgrosser: I think this variant is now beautiful enough that we can eliminate the other variants, no?

hargoniX (approved):

PR #5605, opened 2024-10-03
perf: remove List.redLength

hargoniX:


hargoniX:
!bench

hargoniX:
!bench

hargoniX:
The number of branch misses increases but there's also 14M branches less so this should be fine.

PR #5608, opened 2024-10-03
chore: add bv_toNat tag for toNat_ofInt

bollu:
These were missing from the `bv_toNat` simp-set,
discovered when refactoring LNSym's simp-set:
https://github.com/leanprover/LNSym/pull/208


bollu:
Aha, I see that `BitVec.toNat_sub` is intentionally not part of the `bv_toNat` simp-set:https://github.com/leanprover/lean4/blob/master/src/Init/Data/BitVec/Lemmas.lean#L1935-L1938

```
-- We prefer this lemma to `toNat_sub` for the `bv_toNat` simp set.
-- For reasons we don't yet understand, unfolding via `toNat_sub` sometimes
-- results in `omega` generating proof terms that are very slow in the kernel.
@[bv_toNat] theorem toNat_sub' {n} (x y : BitVec n) :
    (x - y).toNat = ((x.toNat + (2^n - y.toNat)) % 2^n) := by
  rw [toNat_sub, Nat.add_comm]
```

PR #5618, opened 2024-10-04
feat: support `at` in `ac_nf` and use it in `bv_normalize`

tobiasgrosser:
... while at it also call `trivial` to close goals that can be trivially closed.

### DIFF
--- a/tests/lean/run/bv_decide_rewriter.lean
+++ b/tests/lean/run/bv_decide_rewriter.lean
@@ -20,6 +20,9 @@ example (x y z : BitVec 8) (h1 : x = z → False) (h2 : x = y) (h3 : y = z) : Fa
 example (x y : BitVec 256) : x * y = y * x := by
   bv_decide
 
+example {x y z : BitVec 64} : ~~~(x &&& (y * z)) = (~~~x ||| ~~~(z * y)) := by
+  bv_decide
+
 def mem_subset (a1 a2 b1 b2 : BitVec 64) : Bool :=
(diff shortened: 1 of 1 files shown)


src/Lean/Meta/Tactic/AC/Main.lean
```diff
@@ -170,21 +159,55 @@ where
...
 
-@[builtin_tactic acNf] def acNfTactic : Lean.Elab.Tactic.Tactic := fun _ => do
-  let goal ← getMainGoal
-  goal.withContext <| rewriteUnnormalizedNormalForm goal
+/-- Implementation of the `ac_nf` tactic when operating on the main goal. -/
+def acNfTarget : TacticM Unit :=
+  liftMetaTactic1 fun goal => do return (← rewriteUnnormalized goal)
+
+/-- Implementation of the `ac_nf` tactic when operating on a hypothesis. -/
+def acNfHyp (fvarId : FVarId) : TacticM Unit :=
```
bollu, line 185: Can we generalize this to `acNfExpr (e : Expr) : TacticM Unit` instead? Then we can write `acNFHyp` as a thin wrapper around this. Happy to do this if you want me to @tobiasgrosser .

hargoniX (changes requested):
I think it looks fine apart from this, remaingin TODOs:
- [x] change bv_ test cases so that goals that are now trivially solved by `ac_nf` are expanded upon with new tests that keep stressing the bitblaster on the same operations
- [ ] fix already existing test cases

hargoniX (approved):
Will wait until CI issues are resolved to merge.

PR #5281, opened 2024-09-08
feat: add udiv/umod bitblasting for bv_decide

bollu:
This PR adds the theorems

```
@[simp]
theorem divRec_zero (qr : DivModState w) :
  divRec w w 0 n d qr  = qr

@[simp]
theorem divRec_succ' (wn : Nat) (qr : DivModState w) :
    divRec w wr (wn + 1) n d qr =
    let r' := shiftConcat qr.r (n.getLsbD wn)
    let input : DivModState w :=
      if r' < d then ⟨qr.q.shiftConcat false, r'⟩ else ⟨qr.q.shiftConcat true, r' - d⟩
    divRec w (wr + 1) wn n d input
```

The final statements may need some masasging to interoperate with `bv_decide`. We prove the recurrence for unsigned division by building a shift-subtract circuit, and then showing that this circuit obeys the division algorithm's invariant. 

--- 

A `DivModState` is lawful if the remainder width `wr` plus the dividend width `wn` equals `w`,
and the bitvectors `r` and `n` have values in the bounds given by bitwidths `wr`, resp. `wn`.
This is a proof engineering choice: An alternative world could have
`r : BitVec wr` and `n : BitVec wn`, but this required much more dependent typing coercions.
Instead, we choose to declare all involved bitvectors as length `w`, and then prove that
the values are within their respective bounds.

PR #5620, opened 2024-10-04
feat: Document Bitblasting in a documentation comment

bollu:
As requested by @kim-em at https://github.com/leanprover/lean4/pull/5281#issuecomment-2376102963. We provide a high-level overview of the workflow for adding new bitblasting theorems, by using the `BitVec.mul` as a prototypical example.

### DIFF
--- a/src/Init/Data/BitVec/Bitblast.lean
+++ b/src/Init/Data/BitVec/Bitblast.lean
@@ -87,7 +87,7 @@ computes the correct value for multiplication.
 
 To zoom out, therefore, we follow two steps:
 First, we prove bitvector lemmas to unfold a high-level operation (such as multiplication)
-into already bitblstable operations (such as addition and left shift).
+into already bitblastable operations (such as addition and left shift).
 We then use these lemmas to prove the correctness of the circuit that `bv_decide` builds.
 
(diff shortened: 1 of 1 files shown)


hargoniX (approved):

src/Init/Data/BitVec/Bitblast.lean
```diff
@@ -18,6 +18,80 @@ as vectors of bits into proofs about Lean `BitVec` values.
...
+```
+
+The definition and theorem above are internal to `bv_decide`,
+and use `mulRec_{zero,succ}_eq` to prove that the circuit built by `bv_decide`
+computes the correct value for multiplication.
+
+To zoom out, therefore, we follow two steps:
+First, we prove bitvector lemmas to unfold a high-level operation (such as multiplication)
+into already bitblstable operations (such as addition and left shift).
+We then use these lemmas to prove the correctness of the circuit that `bv_decide` builds.
```
tobiasgrosser, line 91: bitblstable?

bollu: fixed.

tobiasgrosser (approved):
Thank you, @bollu.

PR #5609, opened 2024-10-03
feat: lemmas for Bitvector division when denominator is zero

bollu:
These lemmas explain what happens when the denominator is zero with `udiv`, `umod`, `sdiv`, `smod`. A follow-up PR will show what happens with `smtUDiv` and `smtSMod`, since these need some more bitvector theory. 
These lemmas will be used by `bv_decide` for bitblasting.

The theorems `{sdiv, smod}_zero` are located after `neg` theory has been built for the purpose of writing terse proofs.

PR #5623, opened 2024-10-06
chore: move BitVec.udiv/umod/sdiv/smod after add/sub/mul/lt

tobiasgrosser:
Divison proofs are more likely to depend on add/sub/mul proofs than the other way around. This cleans up https://github.com/leanprover/lean4/pull/5609, which added division proofs that rely on negation to already be defined.

### DIFF
--- a/src/Init/Data/BitVec/Lemmas.lean
+++ b/src/Init/Data/BitVec/Lemmas.lean
@@ -1445,104 +1445,6 @@ theorem msb_sshiftRight' {x y: BitVec w} :
     (x.sshiftRight' y).msb = x.msb := by
   simp [BitVec.sshiftRight', BitVec.msb_sshiftRight]
 
-/-! ### udiv -/
-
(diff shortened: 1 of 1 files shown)


bollu:
LGTM, and I ran a `diff`, and the only change was to make a terminal `simp` into a `simp only`:

```diff
bollu@paracompact-2 ~/temp [1]> diff a b
87c87
<   <;> simp [h'', h''']
---
>   <;> simp only [h'', h''', ↓reduceIte, toNat_ofNat, Nat.zero_mod, toNat_umod]
```

tobiasgrosser:
> LGTM, and I ran a `diff`, and the only change was to make a terminal `simp` into a `simp only`:
> 
> ```diff
> bollu@paracompact-2 ~/temp [1]> diff a b
> 87c87
> <   <;> simp [h'', h''']
> ---
> >   <;> simp only [h'', h''', ↓reduceIte, toNat_ofNat, Nat.zero_mod, toNat_umod]
> ```

This is intentional.

bollu (approved):
Since it was intentional, LGTM

PR #5631, opened 2024-10-07
feat: `decide!`

### LINKED ISSUE #5629: RFC: decide! tactic variant
### Proposal

In the `equational_theories` project there are many script-generated theorems with essentially `by decide` as the proof. In theses cases, we really know that the `decide` tactic will succeed (if not there is a bug somewhere), and it’s silly for `decide` to first run the decision procedure at elaboration time, and then again in the kernel.

Thus I propose to add a `decide!` variant of the tactic that causes this check to be skipped:

https://github.com/leanprover/lean4/blob/01d414ac36dc28f3e424dabd36d818873fea655c/src/Lean/Elab/Tactic/ElabTerm.lean#L404-L405

This halfs the processing time of such files.

(The equational_theories project is currently using a tactic that has this effect.)

### Community Feedback

(none yet)

### Impact

nomeata:
This pull request introduces a new variant of the `decide` tactic called `decide!` to speed up the processing of large proofs by skipping proposition checks at the time of elaboration.

This is particularly useful for auto-generated lean proofs.

Fixes #5629

### DIFF
--- a/src/Lean/Parser/Tactic.lean
+++ b/src/Lean/Parser/Tactic.lean
@@ -129,7 +129,7 @@ example : 1 + 1 = 2 := by rfl
 A variant of `decide` that does not check the proposition at the time of elaboration, and just
 assumes it is true. The proposition is of course still checked by the kernel.
 
-This is useful to speed up the processing of large proofs, by halfing the processing time.
+This is useful to speed up the processing of large proofs, by halving the processing time.
 -/
 @[builtin_tactic_parser] def decideBang := leading_parser
   nonReservedSymbol "decide!"


tobiasgrosser:
Interesting. We just run into this issue in https://github.com/opencompl/lean-mlir/pull/673/files. In particular, we use `native_decide` which fails and then run in a kernel error. For some reason `try native_decide` does not fail, but succeeds and we get an error only on the function overall. This PR is not fixing this, but maybe adds a data-point that there are also automated uses of native_decide that fail and where we trigger the same kernel errors.

bollu:
The same issue crops up at https://github.com/leanprover/lean4/pull/5574, where checking the proof to throw an error message when `bv_check` fails to check the LRAT proof effectively doubles the execution time. 

In an ideal world, there would be some sort of API that can request the kernel to check a proof from Meta code, and returns either a "checked proof handle", or an error, which can then be handled at meta-time.

It seems complex to implement such a feature, though, as the kernel would then need to maintain such "proof handles", which is bad for TCB...

alexkeizer:
> It seems complex to implement such a feature, though, as the kernel would then need to maintain such "checked proof handles", which is bad for TCB.

This function already mostly exists, it's called a theorem!

IIRC, a tactic can add new theorems to the environment, so we can have `decide` or `bv_decide` generate the proof term. Then, it adds that proof to the environment as a theorem, to get the kernel to check it. If type-checking fails (I assume we can catch that?) we can throw a meaningfull errro in the tactic. Otherwise, if type-checking succeeds, we can emit the new theorem in our original proof, so that we don't have to recheck the proof

nomeata:
That's clever! This would allow us to half the processing time of `decide` everywhere - at the expense of lots of theorems proving `0 < 1` over and over again.

Maybe this can hook into the ”abstract proofs” machinery used elsewhere.

alexkeizer:
We could think of some threshold heuristic, I can imagine that for small statements like `0 < 1` the overhead of adding a new theorem to the kernel could be more than just running the meta checker like we do currently. 

> Maybe this can hook into the ”abstract proofs” machinery used elsewhere.

I'm not sure I know what you're referring to here, what is this abstract proofs machinery?

nomeata:
> We could think of some threshold heuristic, I can imagine that for small statements like `0 < 1` the overhead of adding a new theorem to the kernel could be more than just running the meta checker like we do currently.

That's a possibility, if we can make it robust (run with a low number of heartbeats). But the incidental complexity and unpredictabiliy of such a  design is a bit worrysome.

> > Maybe this can hook into the ”abstract proofs” machinery used elsewhere.
> 
> I'm not sure I know what you're referring to here, what is this abstract proofs machinery?

I'll have to look for it myself, but I've seen `.proof_1` definitions show up. Likely something related to https://github.com/leanprover/lean4/blob/f1ff9cebf29edfc40fab70967274d3fb9b3cda1b/src/Lean/Meta/AbstractNestedProofs.lean#L78

nomeata:
Ah, but it's not the abstracted proofs where I saw lean use previously defined helpers from unrelated definitions, but the match compiler:
https://github.com/leanprover/lean4/blob/f1ff9cebf29edfc40fab70967274d3fb9b3cda1b/src/Lean/Meta/Match/Match.lean#L769

src/Lean/Parser/Tactic.lean
```diff
@@ -125,6 +125,15 @@ example : 1 + 1 = 2 := by rfl
 @[builtin_tactic_parser] def decide := leading_parser
   nonReservedSymbol "decide"
 
+/--
+A variant of `decide` that does not check the proposition at the time of elaboration, and just
+assumes it is true. The proposition is of course still checked by the kernel.
+
+This is useful to speed up the processing of large proofs, by halfing the processing time.
```
vihdzp, line 132:
```suggestion
This is useful to speed up the processing of large proofs, by halving the processing time.
```

nomeata:
Closing in favor of https://github.com/leanprover/lean4/pull/5665

PR #5638, opened 2024-10-07
feat: add `MonadLift Id m` instance

kmill:


kmill:
Closing. I don't have a pressing need for this, and it's easy enough to lift `Id` monads using `Id.run`, which is very rare to need to do.

The original motivation was in some automation in `#eval` where it was easier to make all values monadic, even pure ones, and then use `MonadLift` to bring all values into a common monad. The design changed, and this instance is no longer necessary.

kmill:
(Note: closed after discussion with @Kha)

PR #5643, opened 2024-10-07
feat: add option `pp.mvars.delayed`

kmill:
Where before we had
```lean
#check fun x : Nat => ?a
-- fun x ↦ ?m.7 x : (x : Nat) → ?m.6 x
```
Now by default we have
```lean
#check fun x : Nat => ?a
-- fun x => ?a : (x : Nat) → ?m.6 x
```
In particular, delayed assignment metavariables such as `?m.7` pretty print using the name of the metavariable they are delayed assigned to, suppressing the bound variables used in the delayed assignment (hence `?a` rather than `?a x`). Hovering over `?a` shows `?m.7 x`.

The benefit is that users can see the user-provided name in local contexts. A justification for this pretty printing choice is that `?m.7 x` is supposed to stand for `?a`, and furthermore it is just as opaque to assignment in defeq as `?a` is (however, when synthetic opaque metavariables are made assignable, delayed assignments can be a little less assignable than true synthetic opaque metavariables).

The original pretty printing behavior can be recovered using `set_option pp.mvars.delayed true`.

This PR also extends the documentation for holes and synthetic holes, with some technical details about what delayed assignments are. This likely should be moved to the reference manual, but for now it is included in this docstring.

(This PR is a simplified version of #3494, which has a round-trippable notation for delayed assignments. The pretty printing in this PR is unlikely to round trip, but it is better than the current situation, which is that delayed assignment metavariables never round trip, and plus it does not require introducing a new notation.)


src/Lean/Parser/Term.lean
```diff
@@ -140,11 +140,62 @@ def optSemicolon (p : Parser) : Parser :=
...
+Related concept: implicit parameters are automatically filled in with holes during the elaboration process.
+
+See also `?m` syntax (synthetic holes).
+-/
 @[builtin_term_parser] def hole := leading_parser
   "_"
-/-- Parses a "synthetic hole", that is, `?foo` or `?_`.
-This syntax is used to construct named metavariables. -/
+/--
+A *synthetic hole* (or *synthetic placeholder*), which stands for an unknown term that should should be synthesized using tactics.
```
nomeata, line 162:
```suggestion
A *synthetic hole* (or *synthetic placeholder*), which stands for an unknown term that should be synthesized using tactics.
```

src/Lean/Parser/Term.lean
```diff
@@ -140,11 +140,62 @@ def optSemicolon (p : Parser) : Parser :=
...
+- When synthetic holes appear under binders, they capture local variables using a more complicated mechanism known as delayed assignment.
+- During elaboration, unification will not solve for synthetic opaque metavariables, they are "opaque".
+  This is to prevent counterintuitive behavior such as disappearing goals.
+
+## Delayed assigned metavariables
+
+This section gives a high-level overview of technical details of synthetic holes, which you should feel free to skip.
+When a synthetic hole appears under a binding construct, such as for example `fun (x : α) (y : β) => ?s`,
+the system creates a construct called a *delayed assignment*. This consists of
+1. A metavariable `?m` of type `α → β → γ` and whole local context is from the local context outside the `fun`,
```
nomeata, line 181: ?
```suggestion
1. A metavariable `?m` of type `α → β → γ` and its local context is the local context outside the `fun`,
```

src/Lean/Parser/Term.lean
```diff
@@ -140,11 +140,62 @@ def optSemicolon (p : Parser) : Parser :=
...
+Synthetic holes are similar to holes in that `_` also creates metavariables,
+but synthetic opaque metavariables have some different properties:
+- In tactics such as `refine`, only synthetic holes yield new goals.
+- When synthetic holes appear under binders, they capture local variables using a more complicated mechanism known as delayed assignment.
+- During elaboration, unification will not solve for synthetic opaque metavariables, they are "opaque".
+  This is to prevent counterintuitive behavior such as disappearing goals.
+
+## Delayed assigned metavariables
+
+This section gives a high-level overview of technical details of synthetic holes, which you should feel free to skip.
```
nomeata, line 178: Good to write this up! But I’m wary of putting it anywhere where a beginner could accidentially stumble over it, such as hovers. In general, this is probably too long for a hover docstring. It should probably go into the reference manual and the docstring could indicate that there is more deeply technical information there.

Not sure if this should hold up the present PR, though; we could merge as it is right now and doc-refactor later.

nomeata (approved):

src/Lean/Parser/Term.lean
kmill: I agree. I've just added some more warnings not to read the section :-)

PR #5644, opened 2024-10-08
fix: now linters in general do not run on `#guard_msgs` itself

kmill:
The `#guard_msgs` command runs the command it is attached to as if it were a top-level command. This is because the top-level command elaborator runs linters, and we are interested in capturing linter warnings using `#guard_msgs`. However, the linters will run on `#guard_msgs` itself, leading sometimes to duplicate warnings (like for the unused variable linter).

Rather than special-casing `#guard_msgs` in every affected linter, this PR special-cases it in the top-level command elaborator itself. **Now linters are only run if the command doesn't contain `#guard_msgs`.** This way, the linters are only run on the sub-command that `#guard_msgs` runs itself. This rule also keeps linters from running multiple times in cases such as `set_option pp.mvars false in /-- ... -/ #guard_msgs in ...`.

### DIFF
--- a/src/Lean/Elab/Command.lean
+++ b/src/Lean/Elab/Command.lean
@@ -520,8 +520,12 @@ def elabCommandTopLevel (stx : Syntax) : CommandElabM Unit := withRef stx do pro
     -- recovery more coarse. In particular, If `c` in `set_option ... in $c` fails, the remaining
     -- `end` command of the `in` macro would be skipped and the option would be leaked to the outside!
     elabCommand stx
-    withLogging do
-      runLinters stx
+    -- Run the linters, unless `#guard_msgs` is present, which is special and runs `elabCommandTopLevel` itself,
+    -- so it is a "super-top-level" command. This is the only command that does this, so we just special case it here
+    -- rather than engineer a general solution.
+    unless (stx.find? (·.isOfKind ``Lean.guardMsgsCmd)).isSome do
+      withLogging do
+        runLinters stx
   finally
     -- note the order: first process current messages & info trees, then add back old messages & trees,
     -- then convert new traces to messages
--- a/src/Init/Notation.lean
+++ b/src/Init/Notation.lean
@@ -675,6 +675,13 @@ Message ordering:
 
 For example, `#guard_msgs (error, drop all) in cmd` means to check warnings and drop
 everything else.
+
+The command elaborator has special support for `#guard_msgs` for linting.
+The `#guard_msgs` itself wants to capture linter warnings,
+so it elaborates the command it is attached to as if it were a top-level command.
+However, the command elaborator runs linters for *all* top-level commands,
+which would include `#guard_msgs` itself, and would cause duplicate and/or uncaptured linter warnings.
+The top-level command elaborator only runs the linters if `#guard_msgs` is not present.
 -/
 syntax (name := guardMsgsCmd)
   (docComment)? "#guard_msgs" (ppSpace guardMsgsSpec)? " in" ppLine command : command
--- a/src/Lean/Elab/GuardMsgs.lean
+++ b/src/Lean/Elab/GuardMsgs.lean
@@ -140,6 +140,7 @@ def MessageOrdering.apply (mode : MessageOrdering) (msgs : List String) : List S
         |>.trim |> removeTrailingWhitespaceMarker
     let (whitespace, ordering, specFn) ← parseGuardMsgsSpec spec?
     let initMsgs ← modifyGet fun st => (st.messages, { st with messages := {} })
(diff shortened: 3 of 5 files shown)


src/Lean/Elab/Command.lean
```diff
@@ -520,8 +520,12 @@ def elabCommandTopLevel (stx : Syntax) : CommandElabM Unit := withRef stx do pro
     -- recovery more coarse. In particular, If `c` in `set_option ... in $c` fails, the remaining
     -- `end` command of the `in` macro would be skipped and the option would be leaked to the outside!
     elabCommand stx
-    withLogging do
-      runLinters stx
+    -- Run the linters, unless `#guard_msgs` is present, which is special and runs `elabCommandTopLevel` itself,
+    -- so it is a "super-top-level" command. This is the only command that does this, so we just special case it here
+    -- rather than engineer a general solution.
+    unless (stx.find? (·.isOfKind ``Lean.guardMsgsCmd)).isSome do
```
adomani, line 526: This means that linting for `#guard_msgs` or "`in`s that precede it" will become trickier, right?

E.g. the linters would only see `section` in
```lean
variable (x : Nat) in
open Lean in
#guard_msgs in
section
```
?

And something like
```lean
open Lean in
run_cmd
  let stx ← `(command| #guard_msgs in section)
  logInfo stx
```
would be completely unlinted?

Just to be explicit, I am happy with the change, just testing how far the consequences stretch!  :smile:

kmill: You are correct, this change is at odds with the `#`-command linter. One idea that comes to mind is that we could have a core linter against `#guard_msgs` itself and enable it for mathlib, but I need to think about this more.

adomani: This is not an issue for Mathlib: the `#`-command linter checks the kind of the command on the nos (no digging inside the syntax, as far as I remember).  Thus `#guard_msgs` actually appears as an `in` to that linter and `#guard_msgs` is *not* linted.

kmill: I'm a bit confused about what you're saying — could you take a look at the log for the mathlib testing branch? There's an error in `test/HashCommandLinter.lean` where `#guard_msgs` is supposed to be linted. This PR makes it so that `#guard_msgs` never has linters run on it, and this causes that test to fail.

I'm not sure what you mean about `#guard_msgs` appearing as an `in`; the `in` is part of its syntax, it's not `Lean.Parser.Command.in`.

grunweg:
Note to self/whoever gets to address this first: when this lands, batteries also has such as hack which can be removed now.

src/Lean/Elab/Command.lean
adomani: Part of my confusion was due to the fact that the error in the tests is not visible in VSCode, but only with `lake test`.

The other confusion is that I remembered that `in` shielded the `#`-command linter, but really it is the doc-string that usually accompanies `#guard_msgs` that makes the linter ignore `#guard_msgs`.

Hence, as is, it is hard for `#guard_msgs` to get caught in the `#`-command linter:
```lean
import Mathlib.Tactic.Linter

--flagged
#guard_msgs in
run_cmd return

-- not flagged
/---/
#guard_msgs in
run_cmd return

-- not flagged
/-- info: 0 -/
#guard_msgs in
#eval 0
```
This behaviour was partly by design.  I would be happy to have `#guard_msgs` flagged more consistently as a `#`-command, but also to make it maintain "special status".

kmill: Thanks for your feedback @adomani, it's helpful to know that you are ok with `#guard_msgs` being special like this. We can revisit being able to lint for `#guard_msgs` in mathlib if that ever comes up.

Would you mind fixing that first mathlib `HashCommandLinter.lean` test as a mathlib PR? That will help with the next nightly-testing.

adomani: I'm going to fix the test in the same `lean-pr-testing-5644` branch, right?

kmill: I don't think it's necessary to fix `lean-pr-testing-5644`. Maybe all we need to do is fix the nightly-testing branch of mathlib directly.

Pinging @jcommelin: My understanding is that in the next nightly, we'll have to fix the first test in HashCommandLinter.

adomani: Ok, I have added the fix to `lean-pr-testing-5644` and I also removed the "hacky pairs" of `set_option ... false #guard_msgs in set_option ... true in ...`.

I can cherry-pick these changes to the appropriate nightly PR, though.

kmill: Thanks @adomani!

PR #5653, opened 2024-10-08
perf: benchmark for modulo on bv_decide

hargoniX:
This verifies a bit hack from here: https://en.wikipedia.org/wiki/Lehmer_random_number_generator#Sample_C99_code

I previously ran the SMTLIB equivalent this with Bitwuzla in my crypto class and got the following numbers:
- 22s with Bitwuzla
- Z3 and CVC5 don't yet terminate after > 2min

Now with`bv_decide` the overall timing is 33.7s, consisting of:
- 5s of checking the LRAT cert
- 5s of trimming the LRAT cert from 800k to 300k proof steps
- remainder actual solving time

So running `bv_decide` like a normal SMT solver without verifying the result of the SAT solver would yield approximately ~24s.

### DIFF
--- a/tests/bench/bv_decide_mod.lean
+++ b/tests/bench/bv_decide_mod.lean
@@ -1,0 +1,39 @@
+import Std.Tactic.BVDecide
+
+/-
+Verify: https://en.wikipedia.org/wiki/Lehmer_random_number_generator#Sample_C99_code
+
+uint32_t lcg_parkmiller(uint32_t *state)
+{
+	return *state = (uint64_t)*state * 48271 % 0x7fffffff;
+}
+
+vs
+
+uint32_t lcg_parkmiller(uint32_t *state)
+{
+	uint64_t product = (uint64_t)*state * 48271;
+	uint32_t x = (product & 0x7fffffff) + (product >> 31);
+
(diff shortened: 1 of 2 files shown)


hargoniX:
!bench

tests/bench/bv_decide_mod.lean
```diff
@@ -0,0 +1,39 @@
...
+
+def opt (state : BitVec 32) : BitVec 32 :=
+  let product := (state.zeroExtend 64) * 48271
+  let x := ((product &&& 0x7fffffff) + (product >>> 31)).extractLsb 31 0
+  let x := (x &&& 0x7fffffff) + (x >>> 31)
+  x
+
+--set_option trace.Meta.Tactic.bv true in
+--set_option trace.Meta.Tactic.sat true in
+--set_option trace.profiler true in
```
tobiasgrosser, line 34: Do you want to leave these comments in?

tobiasgrosser (approved):
Nice.

tests/bench/bv_decide_mod.lean
hargoniX: Yes

PR #5656, opened 2024-10-08
test: add a benchmark that is slow to elaborate

atomb:
Add an example Lean file that includes an unusually large definition that takes a long time to elaborate.

It may be that it's difficult to process it more efficiently, but perhaps someone will discover a way to improve it if it's in the benchmark suite. Improved performance on this benchmark will likely make some program analysis and verification tasks within Lean more feasible.


atomb:
!bench

Kha:
!bernch

Kha:
!bench

Kha:
@atomb Thank you for the benchmarks! I can see the `big_do` being useful, I added some small commentary to it. The other benchmark I fear is far too large for the suite, none of them should take more than a few seconds. I also fear that this large number of `variable`s is not something we might want to optimize for and would better be modeled as more `axiom`s.

atomb:
> @atomb Thank you for the benchmarks! I can see the `big_do` being useful, I added some small commentary to it. The other benchmark I fear is far too large for the suite, none of them should take more than a few seconds. I also fear that this large number of `variable`s is not something we might want to optimize for and would better be modeled as more `axiom`s.

Why don't I trim this one down to just `big_do`, then? A variation on the other might be appropriate later on, but we can leave that for the future.

PR #5657, opened 2024-10-09
feat: Strict Ackermannization (bv_ac_eager) tactic for QF_UFBV

bollu:
This PR implements strict Ackermannization, which is an algorithmic technique to reduce `QF_UFBV` into `QF_BV`.

The implementation walks over the goals and the hypotheses. When it encounters a function application `(f x1 x2 ... xn)` where the type signature of `f` is `BitVec k1 -> BitVec k2 -> ... -> BitVec kn -> BitVec ko`, it creates a new variable `fAppX : BitVec k0`, and an equation `fAppX = f x1 ... xn`. Next, when it encounters another application of the same function `(f y1 y2 ... yn)`, it creates a new variable `fAppY : BitVec k0`, and another equation `fAppY = f y1 ... yn`.

After the walk, we loop over all pairs of applications of the function `f` we have abstracted. In this case, we have `fAppX` and `fAppY`. For these, we build an extensionality equation, which says that if

```
hAppXAppYExt: x1 = x2 -> y1 = y2 -> ... xn = yn -> fAppX = fAppY
```

Intuitively, this is the only fact that the theory UF can propagate, and we thus encode it directly as a SAT formula,
by abstracting out the actual function application into a new free variable.

This implementation creates the Ackermann variables (`fAppX, fAppY`) first and then, in a subsequent phase, adds all the Ackermann equations (`hAppXAppYExt`). This anticipates the next patches, which will implement *incremental* Ackermannization, where we will use the counterexample from the model to selectively add Ackerman equations.

### DIFF
--- a/src/Lean/Elab/Tactic/BVAckermannize.lean
+++ b/src/Lean/Elab/Tactic/BVAckermannize.lean
@@ -53,8 +53,8 @@ with `?w` a literal `Nat`.  -/
 def ofExpr? (e : Expr) : MetaM (Option BVTy) := ofExpr?Aux e |>.run where
   ofExpr?Aux (e : Expr) : OptionT MetaM BVTy :=
     match_expr e.consumeMData with
-    | bool => return bool
-    | bitVec w => do
+    | Bool => return bool
+    | BitVec w => do
        let w ← getNatValue? w
        return bitVec w
     | _ => OptionT.fail


src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+
+structure Config where
+
+structure Context extends Config where
+ config : Config
+
+ /-- Types that can be bitblasted by bv_decide -/
+ inductive BVTy
+ /-- booleans -/
+ | Bool
```
tobiasgrosser, line 42:
```suggestion
 /-- Booleans -/
 | Bool
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+  xTy : BVTy
+deriving Hashable, BEq, Inhabited
+
+namespace Argument
+
+instance : ToMessageData Argument where
+  toMessageData arg := m!"{arg.x} : {arg.xTy}"
+
+/-- Build an `Argument` from a raw expression -/
+def ofExpr? (e : Expr) : OptionT MetaM Argument := do
```
tobiasgrosser, line 85:
```suggestion
/-- Build an `Argument` from a raw expression. -/
def ofExpr? (e : Expr) : OptionT MetaM Argument := do
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+  let t ← BVTy.ofExpr? (← inferType e)
+  return { x := e, xTy := t}
+
+end Argument
+
+
```
tobiasgrosser, line 91:
```suggestion

```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+structure Function where
+  /-- The function -/
+  f : Expr
+  codTy : BVTy
+ deriving Hashable, BEq, Inhabited
+
+
```
tobiasgrosser, line 98:
```suggestion

```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+namespace Function
+
+instance : ToMessageData Function where
+  toMessageData fn := m!"{fn.f} (cod: {fn.codTy})"
+
+/--
+Reify an expression `e` of the kind 
+`f x₁ ... xₙ`, where all the arguments and the return type are a `BVTy` into an Ap
```
tobiasgrosser, line 106:
```suggestion
Reify an expression `e` of the kind `f x₁ ... xₙ`, where all the arguments and the return type are
a `BVTy` into an App.
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+  let xs := f.getAppArgs
+  /- We need at least one argument for this to be a function call we can ackermannize -/
+  guard <| xs.size > 0
+  let codTy ← BVTy.ofExpr? (← inferType f)
+  let args ← xs.mapM Argument.ofExpr?
+  let fn : Function := { f, codTy }
+  return (fn, args)
+end Function
+
+
```
tobiasgrosser, line 118:
```suggestion

```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+  toMessageData fn := m!"{fn.f} (cod: {fn.codTy})"
+
+/--
+Reify an expression `e` of the kind 
+`f x₁ ... xₙ`, where all the arguments and the return type are a `BVTy` into an Ap
+-/
+def reifyAp (f : Expr) : OptionT MetaM (Function × Array Argument) := do
+  let xs := f.getAppArgs
+  /- We need at least one argument for this to be a function call we can ackermannize -/
+  guard <| xs.size > 0
```
tobiasgrosser, line 111:
```suggestion
  /- We need at least one argument for this to be a function call we can ackermannize. -/
  guard <| xs.size > 0
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+
+
+/--
+NOTE: is it sensible to hash an array of arguments?
+We may want to use something like a trie to index these.
+Consider swiching to something like `Trie`.
+-/
+abbrev ArgumentList := Array Argument
+
+
```
tobiasgrosser, line 126:
```suggestion

```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+/--
+The data stored for an ackermannized call to allow us to build proofs.
+-/
+structure CallVal where
+  /-- the free variable `ack_fx₁...xₙ := (f x₁ x₂ ... xₙ)`. -/
+  fvar : FVarId
```
tobiasgrosser, line 132:
```suggestion
  /-- The free variable `ack_fx₁...xₙ := (f x₁ x₂ ... xₙ)`. -/
  fvar : FVarId 
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+deriving Hashable, BEq, Inhabited
+
+namespace CallVal
+
+instance : ToMessageData CallVal where
+  toMessageData cv := m!"{Expr.fvar cv.fvar} ({cv.heqProof})"
+
+end CallVal
+
+
```
tobiasgrosser, line 144:
```suggestion

```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+structure State where
+  /--
+  A maping from a function `f`, to a map of arguments `x₁ ... xₙ`, to the information stored about the call.
+  This is used to generate equations of the form `x₁ = y₁ → x₂ = y₂ → ... → xₙ = yₙ → ack_fx₁...xₙ = ack_fy₁...yₙ on-demand.
```
tobiasgrosser, line 148:
```suggestion
  A mapping from a function `f`, to a map of arguments `x₁ ... xₙ`, to the information stored about the call.
  This is used to generate equations of the form `x₁ = y₁ → x₂ = y₂ → ... → xₙ = yₙ → ack_fx₁...xₙ = ack_fy₁...yₙ on-demand.
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+  -/
+  fn2args2call : Std.HashMap Function (Std.HashMap ArgumentList CallVal) := {}
+  /-- A counter for generating fresh names. -/
+  gensymCounter : Nat := 0
+
+
```
tobiasgrosser, line 154:
```suggestion

```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+private def _insertCallVal (fn : Function) (args : ArgumentList) (cv : CallVal) : AckM Unit := do
+  let calls ← getCallMap fn
+  modify fun s => { s with fn2args2call := s.fn2args2call.insert fn (calls.insert args cv) }
+      
+
+/--
+Replace a call to the function `f` with an `fvar`. Since the `fvar` is defeq to the call,
+we can just replace occurrences of the call with the fvar `f`.
+
+We will later need to add another hypothesis with the equality that the `fvar = f x₁ ... xₙ`
```
tobiasgrosser, line 209:
```suggestion
We will later need to add another hypothesis with the equality that `fvar = f x₁ ... xₙ`
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+-/
+
+def replaceCallWithFVar (g : MVarId) (fn : Function) (args : ArgumentList) : AckM (CallVal × MVarId) := do
```
tobiasgrosser, line 212:
```suggestion
def replaceCallWithFVar (g : MVarId) (fn : Function) (args : ArgumentList) : AckM (CallVal × MVarId) := do
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+      return (val, g)
+    let e := mkAppN fn.f (args.map Argument.x)
+    let name ← gensym
+    let (introDef, g) ← introDefExt g name fn.codTy.toExpr e
+    let cv := { fvar := introDef.defn, heqProof := Expr.fvar introDef.eqProof : CallVal }
+    _insertCallVal fn args cv
+    return (cv, g)
+
+/-- create a trace node in trace class (i.e. `set_option traceClass true`),
+with header `header`, whose default collapsed state is `collapsed`. -/
```
tobiasgrosser, line 225:
```suggestion
/-- Create a trace node in trace class (i.e. `set_option traceClass true`),
with header `header`, whose default collapsed state is `collapsed`. -/
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+    -- NOTE: Rewrite as a fold? It's not too much cleaner, sadly.
+    let mut args := #[]
+    let mut g := g
+    for arg in e.getAppArgs do
+      let gArg ← introAckForExpr g arg
+      g := gArg.2
+      args := args.push gArg.1
+    return (mkAppN f args, g) -- NOTE: is there some way to use update to update this?
+    
+
```
tobiasgrosser, line 253:
```suggestion
    
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+            -- bail out, and recurse into the `app` since we can't ackermannize it.
+            trace[bv_ack] "{crossEmoji} {arg}"
+            return (← ackAppChildren g e)
+           
+        let (call, g) ← replaceCallWithFVar g fn ackArgs
+        trace[bv_ack] "{checkEmoji} {e} → {call}."
+        return (Expr.fvar call.fvar, g)
+end
+
+
```
tobiasgrosser, line 307:
```suggestion

```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,426 @@
...
+    (ackEqApp : fx1x2 = f x1 x2)
+    (ackEqApp' : fx1'x2' = f x1' x2') 
+    -- to be intros'd
+    (h1 : x1 = x1')
+    (h2 : x2 = x2') :
+  fx1x2 = fx1'x2' := 
+  let appEqApp : f x1 x2 = f x1' x2' := congr (congr (Eq.refl f) h1) h2
+  Eq.trans (Eq.trans ackEqApp appEqApp) (Eq.symm ackEqApp')
+
+
```
tobiasgrosser, line 338:
```suggestion

```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,409 @@
...
+    let (eqProof, g) ← g.intro1P
+    return ({ defn, eqProof}, g)
+
+/-- Insert the CallVal `cv` at `(fn, args)` into the state. -/
+private def _insertCallVal (fn : Function) (args : ArgumentList) (cv : CallVal) : AckM Unit := do
+  let calls ← getCallMap fn
+  modify fun s => { s with fn2args2call := s.fn2args2call.insert fn (calls.insert args cv) }
+      
+
+/--
```
tobiasgrosser, line 193:
```suggestion

/--
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,408 @@
+/-
+Copyright (c) 2024 Siddharth Bhat. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Siddharth Bhat
+
+This file implements lazy ackermannization [1, 2]
```
alexkeizer, line 6: It implements strict Ackermannization, not lazy, right?

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,408 @@
...
+open Lean Elab Meta Tactic
+
+initialize Lean.registerTraceClass `bv_ack
+
+namespace Ack
+
+structure Config where
+
+structure Context extends Config where
+ config : Config
```
alexkeizer, line 37: I'm extraordinarily confused as to what is happening here, `Config` is empty, then we have `Context` extend `Config`, but *also* take a `Config` member. But both are empty types, so `Context` is also empty?

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,408 @@
...
+
+instance : ToMessageData BVTy where
+  toMessageData 
+  | .Bool => m!"bool"
+  | .BitVec w => m!"BitVec {w}"
+
+namespace BVTy
+
+ /-- info: _root_.BitVec (w : Nat) : Type -/
+ #guard_msgs in #check _root_.BitVec
```
alexkeizer, line 55: What is this doing, asserting that `BitVec` exists?
I guess you want this to raise an error if the signature of `BitVec` changes, to update `ofExpr?`, if that's the intention, a comment is sorely needed. But I wonder how useful such a `#check` really is, why would BitVec change type signature

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,408 @@
...
+
+/-- Reify a raw expression into the types of bitvectors we can bitblast -/
```
alexkeizer, line 57:
```suggestion
/-- Reify a raw expression of type `Type` into the types of bitvectors we can bitblast,
returning `none` if `e` was not recognized as either `Bool` or `BitVec ?w`, 
with `?w` a literal `Nat`  -/
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,408 @@
...
+def ofExpr? (e : Expr) : OptionT MetaM BVTy :=
+  match_expr e.consumeMData with
+  | _root_.Bool => return Bool
+  | _root_.BitVec w => do
+     let w ← getNatValue? w
+     return .BitVec w
+  | _ => OptionT.fail
+
+def toExpr : BVTy → Expr
```
alexkeizer, line 66: docstring please!

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,408 @@
...
+| .Bool => mkConst ``_root_.Bool
+| .BitVec w => mkApp (mkConst ``_root_.BitVec) (mkNatLit w)
+
+end BVTy
+
+structure Argument where
```
alexkeizer, line 72: argument to what?

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,408 @@
...
+  /-- The expression corresponding to the argument -/
+  x : Expr
+  /-- The cached type of the expression x -/
+  xTy : BVTy
```
alexkeizer, line 76: why not `value` and `type` for the field names, that seems a bit more descriptive

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,408 @@
...
+deriving Hashable, BEq, Inhabited
+
+namespace Argument
+
+instance : ToMessageData Argument where
+  toMessageData arg := m!"{arg.x} : {arg.xTy}"
+
+/-- Build an `Argument` from a raw expression. -/
```
alexkeizer, line 84:
```suggestion
/-- Build an `Argument` from a raw expression of type `Bool` or `BitVec _`. -/
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,408 @@
...
+def ofExpr? (e : Expr) : OptionT MetaM Argument := do
+  let t ← BVTy.ofExpr? (← inferType e)
+  return { x := e, xTy := t}
+
+end Argument
+structure Function where
```
alexkeizer, line 90:
```suggestion

structure Function where
```

Also, docstring please!

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,408 @@
...
+  /-- The function -/
+  f : Expr
+  codTy : BVTy
```
alexkeizer, line 93: I find the following a bit clearer, well worth the few extra keystrokes

```suggestion
  codomain : BVTy
```

alexkeizer: Also, why do we store the codomain, but not the domain?

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,408 @@
...
+namespace Function
+
+instance : ToMessageData Function where
+  toMessageData fn := m!"{fn.f} (cod: {fn.codTy})"
+
+/--
+Reify an expression `e` of the kind `f x₁ ... xₙ`, where all the arguments and the return type are
+a `BVTy` into an App.
+-/
+def reifyAp (f : Expr) : OptionT MetaM (Function × Array Argument) := do
```
alexkeizer, line 104:
```suggestion
def reifyApp (f : Expr) : OptionT MetaM (Function × Array Argument) := do
```

alexkeizer (commented):
I did a partial pass, so far I'm missing some documentation to explain what's what, have not attempted to grok the correctness of the code yet

src/Std/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,12 @@
...
+Copyright (c) 2024 Siddharth Bhat. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Siddharth Bhat
+-/
+prelude
+import Std.Tactic.BVAckermannize.Syntax
+
+/-!
+This directory contains the lazy ackermannization tactic.
+This uses lean's inbuilt bitblaster to uninterpreted functions + bitvectors to SAT.
```
alexkeizer, line 11:
```suggestion
This uses lean's builtin bitblaster to reduce uninterpreted functions + bitvectors to SAT.
```

src/Lean/Elab/Tactic/BVAckermannize.lean
bollu: Deleted.

src/Lean/Elab/Tactic/BVAckermannize.lean
bollu: fixed.

src/Std/Tactic/BVAckermannize/Syntax.lean
```diff
@@ -0,0 +1,31 @@
...
+/--
+Close fixed-width `BitVec`, `Bool`, and uninterpreted function goals by obtaining a proof from an external SAT solver and
+verifying it inside Lean. The solvable goals are currently limited to the Lean equivalent of
+[`QF_BV`](https://smt-lib.org/logics-all.shtml#QF_BV) + [`QF_UF`](https://smt-lib.org/logics-all.shtml#QF_UF)
+
+```lean
+example : ∀ (a b : BitVec 64), (a &&& b) + (a ^^^ b) = a ||| b := by
+  intros
+  bv_decide
+```
```
tobiasgrosser, line 25: I don't see an uninterpreted function here.

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,407 @@
...
+open Lean Elab Meta Tactic
+
+initialize Lean.registerTraceClass `bv_ack
+
+namespace Ack
+
+ /-- Types that can be bitblasted by bv_decide -/
+ inductive BVTy
+ /-- Booleans -/
+ | Bool
```
hargoniX, line 37: Should be indented.

src/Lean/Elab/Tactic/BVAckermannize.lean
hargoniX: Why is this resolved? It should be `App`. The doc-string also talks about an expression `e` while we are working on an `f`?

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,407 @@
...
+
+instance : ToMessageData Function where
+  toMessageData fn := m!"{fn.f} (cod: {fn.codomain})"
+
+/--
+Reify an expression `e` of the kind `f x₁ ... xₙ`, where all the arguments and the return type are
+a `BVTy` into an App.
+-/
+def reifyAp (f : Expr) : OptionT MetaM (Function × Array Argument) := do
+  let xs := f.getAppArgs
```
hargoniX, line 107: what if we are working on a function like
```
f : Nat -> Bool -> Bool
```
applied to something like `f 1 b`. In this situation you would fully give up on reification even though it is very much possible to do this by using `(f 1)` as the function instead. Do we want this? And if so, how far do we want to push it? In theory we could even support things like `f : Nat -> Bool -> Nat -> Bool`.

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,407 @@
...
+  let args ← xs.mapM Argument.ofExpr?
+  let fn : Function := { f, codomain }
+  return (fn, args)
+end Function
+/--
+NOTE: is it sensible to hash an array of arguments?
+We may want to use something like a trie to index these.
+Consider swiching to something like `Trie`.
+-/
+abbrev ArgumentList := Array Argument
```
hargoniX, line 120: It's not a list. We should also check if there exist reasonable examples that make the note above relevant before merging this.

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,407 @@
...
+
+end CallVal
+structure State where
+  /--
+  A mapping from a function `f`, to a map of arguments `x₁ ... xₙ`, to the information stored about the call.
+  This is used to generate equations of the form `x₁ = y₁ → x₂ = y₂ → ... → xₙ = yₙ → ack_fx₁...xₙ = ack_fy₁...yₙ on-demand.
+  -/
+  fn2args2call : Std.HashMap Function (Std.HashMap ArgumentList CallVal) := {}
+  /-- A counter for generating fresh names. -/
+  gensymCounter : Nat := 0
```
hargoniX, line 144: Why can't we use the normal name generator for this? The hypotheses are going to be for machine consumption anyway so accessibility doesn't really matter.

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,407 @@
...
+def withContext (g : MVarId) (ma : AckM α) : AckM α := g.withContext ma
+
+/-- Get the calls to a function `fn`. -/
+def getCallMap (fn : Function) : AckM (Std.HashMap ArgumentList CallVal) := do
+  return (← get).fn2args2call.getD fn {}
+
+/-- Get the calls to a function `fn`. -/
+def getCallVal? (fn : Function) (args : Array Argument) : AckM (Option CallVal) := do
+  let calls ← getCallMap fn
+  if let .some val := calls.get? args then
```
hargoniX, line 168: Is this not just `return calls.get? args` ?

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,407 @@
...
+
+/-
+Introduce a new definition with name `name : hdefTy` into the local context,
+and return the FVarId of the new definition in the new goal (the MVarId) returned.
+-/
+def introDefExt (g : MVarId) (name : Name) (hdefTy : Expr) (hdefVal : Expr) : AckM (IntroDefResult × MVarId) := do
+  withContext g do
+    let g ← g.assertExt name (hName := Name.str name "value") hdefTy hdefVal
+    let (defn, g) ← g.intro1P
+    let (eqProof, g) ← g.intro1P
```
hargoniX, line 186: Does an API like `MVarId.assertHypotheses` make this easier?

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,407 @@
...
+    return ({ defn, eqProof}, g)
+
+/-- Insert the CallVal `cv` at `(fn, args)` into the state. -/
+private def _insertCallVal (fn : Function) (args : ArgumentList) (cv : CallVal) : AckM Unit := do
+  let calls ← getCallMap fn
+  modify fun s => { s with fn2args2call := s.fn2args2call.insert fn (calls.insert args cv) }
+
+/--
+Replace a call to the function `f` with an `fvar`. Since the `fvar` is defeq to the call,
```
hargoniX, line 195: It is not clear without looking at the implementation of the function that the `fvar` you are referring to here is the returned `CallVal`. The remark about having to add another hypothesis later is also not clear to me as `CallVal` already contains a proof of this, why don't we add it on the spot when the `CallVal` got created and store the `FVarId` in `CallVal`? After all the `CallVal` seems to be cached anyway so there should not be a risk of duplicating these theorems?

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,407 @@
...
+    let mut args := #[]
+    let mut g := g
+    for arg in e.getAppArgs do
+      let gArg ← introAckForExpr g arg
+      g := gArg.2
+      args := args.push gArg.1
+    return (mkAppN f args, g) -- NOTE: is there some way to use update to update this?
+
+
+partial def introAckForExpr (g : MVarId) (e : Expr) : AckM (Expr × MVarId) := do
```
hargoniX, line 242: Going directly under binders violates the invariant that we don't deal with loose bvars in `MetaM`, all important functions like `isDefEq`, `whnf` etc. rely on the fact that you do not do this. Is there a particular reason that you break the invariant here.

Furthermore despite being some of the larger functions in this file these are one of the only ones without a doc string.

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,407 @@
...
+    (x1 x1' : X1) (x2 x2' : X2)
+    (ackEqApp : fx1x2 = f x1 x2)
+    (ackEqApp' : fx1'x2' = f x1' x2')
+    -- to be intros'd
+    (h1 : x1 = x1')
+    (h2 : x2 = x2') :
+  fx1x2 = fx1'x2' :=
+  let appEqApp : f x1 x2 = f x1' x2' := congr (congr (Eq.refl f) h1) h2
+  Eq.trans (Eq.trans ackEqApp appEqApp) (Eq.symm ackEqApp')
+/-
```
hargoniX, line 322: Not a doc-string

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,407 @@
...
+Make the ackermannization theorem, which states that: `(∀ i, arg₁[i] = arg₂[i]) -> call₁ = call₂`.
+Formally, we build an expr such as `arg₁ = arg'₁ -> arg₂ = arg'₂ -> ... argₙ = arg'ₙ -> call₁ = call₂`,
+where the proof is by congruence over the equalities.
+-/
+def mkAckThm (g : MVarId) (fn : Function) (args args' : Array Argument) (call call' : CallVal) : AckM MVarId := do
+    trace[bv_ack] "making ack congr thm for '{fn}' '{args}' ~ '{args'}',  calls '{call}', '{call'}'"
```
hargoniX, line 328: Wrong indentation.

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,407 @@
...
+/-
+For every bitvector (x : BitVec w), for every function `(f : BitVec w → BitVec w')`,
+walk every function application (f x), and add a new fvar (fx : BitVec w').
+- Keep an equality that says `fx = f x`.
+- For function application of f, for each pair of bitvectors x, y,
+  add a hypothesis that says `x = y => fx = fy, with proof.
+-/
+def ack (g : MVarId) : AckM MVarId := do
+  g.withContext do
+    let (target', g) ← introAckForExpr g (← inferType (Expr.mvar g))
```
hargoniX, line 369: There is `Lean.MVarId.getType`

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,407 @@
...
+    for hyp in hyps do
+      g ← g.withContext do
+        withTraceNode m!"🎯 hyp '{← hyp.getType}'" (collapsed := false) do
+          let hypG ← introAckForExpr g (← hyp.getType)
+          pure hypG.2
+
+    for (fn, arg2call) in (← get).fn2args2call do
+      let argCallsArr := arg2call.toArray
+      for i in [0:argCallsArr.size] do
+        let (arg₁, call₁) := argCallsArr[i]!
```
hargoniX, line 383: If you are feeling fancy you can actually make these provably in bounds with the `for h : i in ...` syntax.

src/Lean/Elab/Tactic/BVAckermannize.lean
bollu: Done!

src/Lean/Elab/Tactic/BVAckermannize.lean
bollu: This was unused, so I deleted it.

src/Lean/Elab/Tactic/BVAckermannize.lean
bollu: Yes, that's a fair point, I do indeed bail out. I added a comment describing that it's currently very conservative, and that we can extend to handle a larger fragment.

src/Lean/Elab/Tactic/BVAckermannize.lean
bollu: Fixed, thanks!

src/Lean/Elab/Tactic/BVAckermannize.lean
bollu: Done!

src/Lean/Elab/Tactic/BVAckermannize.lean
bollu: Nice, thanks for the fanciness refactor

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,407 @@
...
+import Lean.Meta.InferType
+import Lean.Elab.Tactic.FalseOrByContra
+import Lean.Meta.Tactic.Assert
+import Lean.Meta.Tactic.Util
+import Lean.Meta.Tactic.Congr
+import Lean.Meta.Tactic.Replace
+
+open Lean Elab Meta Tactic
+
+initialize Lean.registerTraceClass `bv_ack
```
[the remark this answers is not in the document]
bollu: I moved it to `trace.Meta.Tactic.bv_ack`. Does that suffice?

src/Lean/Elab/Tactic/BVAckermannize.lean
bollu: Fixed, thanks!

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,434 @@
...
+  match_expr e with
+  | BitVec _ => true
+  | Bool => true
+  | _ => false
+
+
+/--
+The result of running ackermannization on an expression.
+Returns the ackermannized exprssion, as well as informing
+whether this expression can be used as a subexpression for further ackermannization -/
```
tobiasgrosser, line 222:
```suggestion
Returns the ackermannized expression, as well as informing
whether this expression can be used as a subexpression for further ackermannization -/
```

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,434 @@
...
+structure AckResult where
+  /-- The resulting expression from ackermannization -/
+  val : Expr
+  /--
+  Whether an expression has bound variables in it.
+  If it does, then it cannot be used as the subexpression of an ackermanized call
+  -/
```
tobiasgrosser, line 229:
```suggestion
  If it does, then it cannot be used as the subexpression of an ackermanized call.
  -/
```

src/Std/Tactic/BVAckermannize/Syntax.lean
bollu: Fixed, I've updated the docstring.

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -217,17 +217,18 @@ def isBitblastTy (e : Expr) : Bool :=
 
 
 /--
-The result of running ackermannization on an expression.
-Returns the ackermannized exprssion, as well as informing
-whether this expression can be used as a subexpression for further ackermannization -/
+The result of running ackermannization on an expression. Returns the ackermannized exprssion,
+as well as informing whether this expression can be used as a subexpression for further ackermannization.
```
tobiasgrosser, line 221:
```suggestion
The result of running ackermannization on an expression. Returns the ackermannized expression,
as well as informing whether this expression can be used as a subexpression for further ackermannization.
```

src/Lean/Elab/Tactic/BVAckermannize.lean
bollu: I prefer to generate nicer looking names :)

src/Lean/Elab/Tactic/BVAckermannize.lean
bollu: Fixed, thanks!

src/Lean/Elab/Tactic/BVAckermannize.lean
bollu: I rephrased this, and explain that we need to later add the *ackermannization* theorems for the given call, not the equation theorem. Thanks for the critique.

src/Lean/Elab/Tactic/BVAckermannize.lean
bollu: I don't think it saves much, I'm happy the way it is, unless you feel differently? It's not clear to me how to use the `MVarId.assertHypotheses` API to access the proof that `fvar = <defn>` without building the `eqRefl` myself.

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,446 @@
...
+import Lean.Meta.Basic
+import Lean.Elab.Tactic.Basic
+import Lean.Meta.LitValues
+import Lean.Meta.InferType
+import Lean.Elab.Tactic.FalseOrByContra
+import Lean.Meta.Tactic.Assert
+import Lean.Meta.Tactic.Util
+import Lean.Meta.Tactic.Congr
+import Lean.Meta.Tactic.Replace
+
```
alexkeizer, line 27:
```suggestion
/-
Copyright (c) 2024 Siddharth Bhat. All rights reserved.
Released under Apache 2.0 license as described in the file LICENSE.
Authors: Siddharth Bhat
-/
prelude
import Lean.Expr
import Lean.Message
import Std.Tactic.BVDecide.Bitblast
import Std.Tactic.BVAckermannize.Syntax
import Lean.Meta.Basic
import Lean.Elab.Tactic.Basic
import Lean.Meta.LitValues
import Lean.Meta.InferType
import Lean.Elab.Tactic.FalseOrByContra
import Lean.Meta.Tactic.Assert
import Lean.Meta.Tactic.Util
import Lean.Meta.Tactic.Congr
import Lean.Meta.Tactic.Replace

/!-
This file implements strict ackermannization [1, 2]

[1] https://lara.epfl.ch/w/_media/model-based.pdf
[2] https://leodemoura.github.io/files/oregon08.pdf
[3] https://github.com/Z3Prover/z3/blob/d047b86439ec209446d211f0f6b251ebfba070d8/src/ackermannization/lackr.cpp#L206
[4] https://github.com/Z3Prover/z3/blob/d047b86439ec209446d211f0f6b251ebfba070d8/src/ackermannization/lackr_model_constructor.cpp#L344
-/
```

Somewhat nit-picky, but I believe the standard style is to have copyright-blurb, imports, and then file-level documentation

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,446 @@
...
+namespace Ack
+
+ /-- Types that can be bitblasted by bv_decide. -/
+ inductive BVTy
+   /-- Booleans. -/
+   | Bool
+   /-- Bitvectors of a fixed width `w`. -/
+   | BitVec (w : Nat)
+   deriving Hashable, DecidableEq, Inhabited
```
alexkeizer, line 40:
```suggestion
 inductive BVTy
   /-- Booleans. -/
   | bool
   /-- Bitvectors of a fixed width `w`. -/
   | bitVec (w : Nat)
   deriving Hashable, DecidableEq, Inhabited
```

I'm not sure if this is a Mathlib-specific convention, but I'd expect the constructors of an inductive to follow snakeCase naming.

In this case, specifically, I see where you get the capitals from, but feel the similarity between `.Bool : BVTy` and  `Bool : Type` might cause unnecessary confusion as compared to `.bool : BVTy`.

bollu: Done

bollu:
CC @hargoniX :)

bollu:
awaiting-author

bollu:
awaiting-review

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,447 @@
...
+  return out
+
+mutual
+
+/--
+Try to ackermannize an application. Recursively ackermannize all of its subexpressions,
+and try to ackermannize the application if:
+- Neither the function `f` nor the arguments `x₁` ... `xₙ` have bound variables in them.
+- `f` has a type that is ackermannizable: all arguments and output are either `BitVec w` or `Bool`.
+-/
```
alexkeizer, line 235:
```suggestion
Try to ackermannize an application. 
First, recursively ackermannize all of its subexpressions,
then try to ackermannize the application itself if:
- Neither the function `f` nor the arguments `x₁` ... `xₙ` have bound variables in them.
- `f` has a type that is ackermannizable: all arguments and output are either `BitVec w` or `Bool`.
-/
```

I assume this is what's meant? The original phrasing confused me a bit, since I read it as: first ackermannize the application, then ackermannize the subexpressions, then ackermannize the application (huh?).

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,447 @@
...
+This is an optimization that we do not yet implement.
+-/
+def areArgListsTriviallyDifferent (_arg₁ _arg₂ : Array Argument) : AckM Bool := return false
+
+/-
+Return true if the argument lists are trivially the same.
+This is an optimization that we do not yet implement.
+If possible, return the simplified hypothesis of the equality of these expressions.
+TODO: -- def areArgListsTriviallySame (arg₁ arg₂ : Array Argument) : AckM (Option Expr) := return none
+-/
```
alexkeizer, line 338:
```suggestion
```

Do you have plans to implement this optimization soon? If not, adding these stubs just feels a bit too noisy for me, and I'd prefer a comment at the use-site explaining that we can bail out early if the arguments lists are different.

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,447 @@
...
+
+
+/-- info: congr.{u, v} {α : Sort u} {β : Sort v} {f₁ f₂ : α → β} {a₁ a₂ : α} (h₁ : f₁ = f₂) (h₂ : a₁ = a₂) : f₁ a₁ = f₂ a₂ -/
+#guard_msgs in #check congr
```
alexkeizer, line 342:
```suggestion
```

Why do you check for congr here?

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,447 @@
...
+    -- we have these already.
+    (x1 x1' : X1) (x2 x2' : X2)
+    (ackEqApp : fx1x2 = f x1 x2)
+    (ackEqApp' : fx1'x2' = f x1' x2')
+    -- to be intros'd
+    (h1 : x1 = x1')
+    (h2 : x2 = x2') :
+  fx1x2 = fx1'x2' :=
+  let appEqApp : f x1 x2 = f x1' x2' := congr (congr (Eq.refl f) h1) h2
+  Eq.trans (Eq.trans ackEqApp appEqApp) (Eq.symm ackEqApp')
```
alexkeizer, line 355: This seems like they were notes for the implementation, but do we want to keep them around?

src/Lean/Elab/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,447 @@
...
+[1] https://lara.epfl.ch/w/_media/model-based.pdf
+[2] https://leodemoura.github.io/files/oregon08.pdf
+[3] https://github.com/Z3Prover/z3/blob/d047b86439ec209446d211f0f6b251ebfba070d8/src/ackermannization/lackr.cpp#L206
+[4] https://github.com/Z3Prover/z3/blob/d047b86439ec209446d211f0f6b251ebfba070d8/src/ackermannization/lackr_model_constructor.cpp#L344
+-/
+open Lean Elab Meta Tactic
+
+initialize Lean.registerTraceClass `Meta.Tactic.bv_ack
+
+namespace Ack
```
alexkeizer, line 33: To me, "ack" means acknowledge. In the context of this file, the abbrevation makes sense, but for the namespace we might want to spell out Ackermannize fully

src/Std/Tactic/BVAckermannize/Syntax.lean
```diff
@@ -0,0 +1,57 @@
...
+it creates a new variable `fAppY : BitVec k0`, and another equation `fAppY = f y1 ... yn`,
+along with another equation (the congruence lemma), which states:
+
+```
+hAppXAppYExt: x1 = x2 -> y1 = y2 -> ... xn = yn -> fAppX = fAppY
+```
+
+Intuitively, this is the only fact that they theory UF can propagate,
+and we thus encode it directly as a SAT formula,
+by abstracting out the actual function application into a new free variable.
```
alexkeizer, line 32:
```suggestion
Intuitively, this is the only fact that the theory UF can propagate,
and we thus encode it directly as a SAT formula,
by abstracting out the actual function application into a new free variable.
```

src/Std/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,12 @@
+/-
+Copyright (c) 2024 Siddharth Bhat. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Siddharth Bhat
+-/
+prelude
+import Std.Tactic.BVAckermannize.Syntax
+
+/-!
+This directory contains the lazy ackermannization tactic.
```
alexkeizer, line 10:
```suggestion
This directory contains the eager ackermannization tactic.
```

src/Std/Tactic/BVAckermannize.lean
```diff
@@ -0,0 +1,12 @@
...
+This uses lean's builtin bitblaster to reduce uninterpreted functions + bitvectors to SAT.
```
alexkeizer, line 11: Does it actually integrate `bv_decide` into the tactic? I thought it rather reduces `uninterpreted functions + bitvectors` to `bitvectors` (which bv_decide can then solve by reduction to SAT)

